PR #2956, opened 2020-03-20
Upgrade netty and add Cookie SameSite attribute support

croudet:


http/src/main/java/io/micronaut/http/cookie/Cookie.java
```diff
@@ -78,6 +78,23 @@
      */
     long getMaxAge();
 
+    /**
+     * Checks to see if this {@link Cookie} can be sent along cross-site requests.
+     * For more information, please look
+     * <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05">here</a>
+     * @return The SameSite attribute of the cookie
+     */
+    @Nullable SameSite getSameSite();
```
graemerocher: Maybe prefer `Optional<SameSite>`

http/src/main/java/io/micronaut/http/cookie/Cookie.java
```diff
@@ -78,6 +78,23 @@
...
+
+    /**
+     * Determines if this this {@link Cookie} can be sent along cross-site requests.
+     * For more information, please look
+     *  <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05">here</a>
+     * @param sameSite SameSite value
+     * @return This cookie
+     */
+    @Nonnull Cookie sameSite(SameSite sameSite);
```
graemerocher: Argument should have `@Nullable`

http/src/main/java/io/micronaut/http/cookie/SameSite.java
```diff
@@ -0,0 +1,43 @@
...
+   * Return the enum value corresponding to the passed in same-site-flag, using a case insensitive comparison.
+   *
+   * @param name value for the SameSite Attribute
+   * @return enum value for the provided name or null
+   */
+  static SameSite of(String name) {
+      if (name != null) {
+          for (SameSite each : SameSite.class.getEnumConstants()) {
+              if (each.name().equalsIgnoreCase(name)) {
+                  return each;
```
graemerocher: Can you not just use `valueOf`? https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf(java.lang.Class,%20java.lang.String)

http/src/main/java/io/micronaut/http/cookie/Cookie.java
graemerocher: Also should be a default method that return `Optional.empty()` to avoid breaking API change

http/src/main/java/io/micronaut/http/cookie/Cookie.java
graemerocher: Make it a default no-op method to avoid breaking API change

http/src/main/java/io/micronaut/http/cookie/CookieConfiguration.java
```diff
@@ -63,4 +63,9 @@
      * @return The max age to use for the cookie
      */
     Optional<TemporalAmount> getCookieMaxAge();
+
+    /**
+     * @return return the SameSite to use for the cookie.
+     */
+    Optional<SameSite> getCookieSameSite();
```
graemerocher: Make a default method that returns `Optional.empty()` to avoid breaking API change

http/src/main/java/io/micronaut/http/cookie/SameSite.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.cookie;
+
+/**
+ * Possible values for the SameSite attribute.
+ * See <a href="https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05">changes to RFC6265bis</a>
+ */
```
graemerocher: add `@author` and `@since` tags

http/src/main/java/io/micronaut/http/cookie/SameSite.java
```diff
@@ -0,0 +1,43 @@
...
+public enum SameSite {
+  Lax,
+  Strict,
+  None;
```
graemerocher: javadoc for each constant please

http/src/main/java/io/micronaut/http/cookie/Cookie.java
croudet: Done

http/src/main/java/io/micronaut/http/cookie/Cookie.java
croudet: Done

http/src/main/java/io/micronaut/http/cookie/SameSite.java
croudet: Removed

http/src/main/java/io/micronaut/http/cookie/CookieConfiguration.java
croudet: Done

http/src/main/java/io/micronaut/http/cookie/SameSite.java
croudet: Done

http/src/main/java/io/micronaut/http/cookie/SameSite.java
croudet: Done

graemerocher:
Thanks for the contribution!

Issue #2973, opened 2020-03-24
HTTP 500 when throwing HttpStatusException from Futures.


ratcashdev:
### Steps to Reproduce

```
    @Get("future")
    public CompletableFuture<String> future() {
        // results in HTTP 500
        return CompletableFuture.supplyAsync(() -> {
            throw new HttpStatusException(HttpStatus.BAD_REQUEST, "world");
        });
    }

    @Get("direct")
    public String direct() {
        // results in HTTP 400
        throw new HttpStatusException(HttpStatus.BAD_REQUEST, "world");
    }
```

### Expected Behaviour

Both cases will return HTTP 400

### Actual Behaviour

future() returns HTTP 500

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** 1.3.3
- **JDK Version:** JDK11

ratcashdev:
a class like 
```
@Singleton
public class CompletionExceptionHandler implements ExceptionHandler<CompletionException, HttpResponse> {

    @Override
    public HttpResponse handle(HttpRequest request, CompletionException exception) {
        if (exception.getCause() instanceof HttpStatusException) {
            HttpStatusException cause = (HttpStatusException) exception.getCause();
            return HttpResponse.status(cause.getStatus()).body(cause.getMessage());
        }
        return HttpResponse.status(HttpStatus.INTERNAL_SERVER_ERROR, exception.getMessage());
    }

}
```
restores expected behavior.

PR #2974, opened 2020-03-24
Handler for wrapped HttpStatusExceptions 

ratcashdev:
wrapped by CompletionExceptions (used by CompletableFutures) to solve https://github.com/micronaut-projects/micronaut-core/issues/2973


graemerocher:
Thanks for the PR, can you provide an example of what problem this PR is trying to fix? When is `HttpStatusException` causing a problem?

ratcashdev:
@graemerocher see the linked issue for the problem description: https://github.com/micronaut-projects/micronaut-core/issues/2973

ratcashdev:
@jameskleeh does line 274 have any relation to https://github.com/micronaut-projects/micronaut-core/blob/ae5430bfec4cd2fe838e2df874509b0403158795/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java#L416 ? Or one is for sync calls and the other for async?

jameskleeh:
@ratcashdev That line 416 is only invoked if the exception handler throws an exception, so I don't think that has any bearing on this fix.

>  should we do a separate PR, or revert the unnecessary changes in this one?

I would say just revert the changes here and use the existing PR

> would the implemention in method exceptionCaughtInternal correctly call the handler

Yes it will

ratcashdev:
@jameskleeh so will this get into 1.3.4?

Issue #2970, opened 2020-03-24
Unable to configure netty  WRITE_BUFFER_WATER_MARK


ratcashdev:
Guides point to the `NettyHttpServerConfiguration` class as a reference for Netty's configuration.
Latest NETTY API recommends to use the WRITE_BUFFER_WATER_MARK `ChannelOption` to configure watermark thresholds. See https://netty.io/4.1/api/io/netty/channel/ChannelOption.html#WRITE_BUFFER_WATER_MARK

### Steps to Reproduce
Given the above, let's assume the following yaml config (based on the :
```
micronaut:
  server:
    netty:
      childOptions:
        write_buffer_water_mark: 
          # 256 Kbytes
          high: 262144
          # 64 kBytes
          low: 65536
```

Such options are processed in https://github.com/micronaut-projects/micronaut-core/blob/e9378a1adf1873bbb30889b74a5498c6cab89eca/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java#L581

### Expected Behaviour
Netty's `serverBootstrap::childOption` is called for the argument.

### Actual Behaviour
It seems however, that the line `environment.convert(value, arg);` returns an empty optional, hence NETTY never receives the values.

### Environment Information

- **Operating System**: Manjaro Linux 5.5.7
- **Micronaut Version:** 1.3.3
- **JDK Version:** JDK11

### Example Application
```
public class NettyConfigTest {

    NettyHttpServerConfiguration config;
    Environment environment;
    long callCount = 0;

    @BeforeMethod
    public void init() {
        Map<String, Object> params = Map.of(
                "micronaut.server.netty.childOptions.write_buffer_water_mark.high", 262143,
                "micronaut.server.netty.childOptions.write_buffer_water_mark.low", 65535
        );

        ApplicationContext ctx = ApplicationContext.run(params, (String) null);
        config = ctx.createBean(NettyHttpServerConfiguration.class);
        environment = ctx.getEnvironment();
    }

    @Test
    void testNettyConfig() throws IOException {
//        ServerBootstrap serverBootstrap = new ServerBootstrap();

        Map<String, Object> configValue
                = (Map<String, Object>) config.getChildOptions().get(ChannelOption.WRITE_BUFFER_WATER_MARK);
        assertEquals(configValue.get("high"), 262143);
        assertEquals(configValue.get("low"), 65535);

[...]

PR #2975, opened 2020-03-24
Support the WRITE_BUFFER_WATER_MARK ChannelOption of netty.

ratcashdev:
https://netty.io/4.1/api/io/netty/channel/ChannelOption.html#WRITE_BUFFER_WATER_MARK introduced a new way of configuring watermarks that requires a new TypeConverter in netty to be able to configure it through `NettyHttpServerConfiguration`.
The `WriteBufferWaterMarkConverterTest` is rather heavy (and java) but could not find a better way of ensuring that the TypeConverter actually kicks in and hands over the config values to `ServerBootstrap`. Feel free to amend.
Solves https://github.com/micronaut-projects/micronaut-core/issues/2970

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/WriteBufferWaterMarkConverter.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/WriteBufferWaterMarkConverter.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


ratcashdev:
@graemerocher code review please.

graemerocher:
The problematic part in this PR is in the test which is written in JUnit 4 and doesn't shutdown the context after using it. If you could update it to JUnit 5 and shutdown the server afterwards that would be good.

ratcashdev:
@graemerocher so I have to call explicitly `ctx.close()`? I thought that the finalizer takes care. I will do the changes.

ratcashdev:
@graemerocher  JUnit5 does not seem to be on the classpath for this module(?)

graemerocher:
Ok leave it in junit 4 we will migrate it later, still need to close the context though

graemerocher:
Thanks for the contribution

Issue #2902, opened 2020-03-09
Chunked response detection / switching


sgammon:
### Steps to Reproduce

1. Prepare a `@Controller` that returns a reactive type
2. Try to publish only one chunk as an HTTP response
3. It is still a chunked response, when it could be a regular response (or vice versa)

### Expected Behaviour

It would be awesome if Micronaut could "detect" a single-step emission, and turn that into a regular HTTP response, versus a multi-step emission, which it should turn into a chunked response.

### Actual Behaviour

Instead of the expected/desired behavior, Micronaut currently will only do one or the other per endpoint, based on the return type.

### Environment Information

- **Operating System**: N/A
- **Micronaut Version:** `1.3.2`
- **JDK Version:** N/A

### Example Application

This issue surfaced during work on `micronaut-views` (specifically, the Soy rendering layer), in micronaut-projects/micronaut-views#34. @graemerocher suggested filing this issue to track progress.

jameskleeh:
As far as I understand, the headers get written first. The body is then subscribed to and written. I don't think we can know only a single item will be emitted unless it's noted either through the `@Single` annotation or `@Get(single=true)`, or returning `Single` that the response will not be chunked.

graemerocher:
I think it may be possible, we would need to alter the logic to evaluate whether the emitted object is single or not. I did this for Micronaut Servlet so it may be possible with the Netty implementation too.

graemerocher:
This PR includes a potential implementation of this https://github.com/micronaut-projects/micronaut-core/pull/2976

Don't know if you want to try build that branch of Micronaut and let me know if it works for you.

sgammon:
@graemerocher i'll give it a shot, it may be next week though (we're pushing to get this little app into prod and for now it's patched to emit a single chunk)

Issue #2566, opened 2019-12-24
Reduce Reactive call stack


fabienrenaud:
The reactive call stack is still huge in micronaut 1.3.0.M2 and has a lot of overhead.

![Screen Shot 2019-12-24 at 9 39 11 AM](https://user-images.githubusercontent.com/3112233/71425729-403c7100-2655-11ea-9683-d5c5b0f934ef.png)

This call stack shows only about 10% of useful work is done for a total of 32%  cpu used for this call stack.

Profiled with a very simple controller:
```
public class App {

  @Controller("/service")
  public static class Control {

    @Get
    public CompletionStage<String> handle(Optional<String> str) {
      return CompletableFuture.completedFuture(str.orElse("nop"));
    }
  }

  public static void main(String[] args) {
    Micronaut.run(App.class, args);
  }
}
```

fabienrenaud:
@graemerocher Any hope of this getting done in the near future? Is there any work scheduled to improve micronaut's performance and move it upwards on these techempower benchmark charts?

graemerocher:
@fabienrenaud Yes this issue is still on my todo list to look at for Micronaut 2.x which we are working on now. There are definitely things we can improve and have been improving. See https://github.com/micronaut-projects/micronaut-core/commit/15a1fd6a316e293ca0d2076604c290b54ea48e6c for example of a recent commit.

There are many potential areas for optimization.

graemerocher:
btw what tool did you use to produce that call stack? I have been using YourKit for profiling

fabienrenaud:
I used jprofiler from ej-technologies.

> Micronaut 2.x

Nice! Is there a roadmap / discussion topic somewhere that I can follow?

graemerocher:
PR #2976 improves matters further, as for the road map you can see what we are working on here https://github.com/micronaut-projects/micronaut-core/milestone/43

graemerocher:
Closing this for now, the stack depth is smaller in master. I have pushed some performance optimizations to some hotspots as well.

PR #2976, opened 2020-03-25
Rewrite the RoutingInBoundHandler to simplify trace and allow chunking

graemerocher:
This PR is a rewrite of RoutingInBoundHandler to improve performance, reduce the reactive call stack (issue #2566) and allow response chunking (issue #2902) via filter.

### DIFF
--- a/core/src/main/java/io/micronaut/core/async/publisher/Publishers.java
+++ b/core/src/main/java/io/micronaut/core/async/publisher/Publishers.java
@@ -16,6 +16,7 @@
 package io.micronaut.core.async.publisher;
 
 import io.micronaut.core.annotation.Internal;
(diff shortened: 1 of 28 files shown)


core/src/main/java/io/micronaut/core/async/publisher/Publishers.java
```diff
@@ -44,10 +45,13 @@
     @SuppressWarnings("ConstantName")
     private static final List<Class<?>> SINGLE_TYPES = new ArrayList<>(3);
 
+    private static final List<Class<?>> COMPLETABLE_TYPES = new ArrayList<>(3);
```
croudet: May be call `ArrayList.trimToSize()` once the list have been filled.

PR #3008, opened 2020-04-01
Fix flaky test in ManualHttpServiceDefinitionSpec

renannprado:
The 'test working SSL configuration' was flaky due to the fact
that it was depending on the exact availability of the port 8443
in the host machine in order to pass.

The test is now using a random server port, which will allow us
to remove the `@Retry` that we had there before, avoiding failing
this test due to specific port unavailability.


graemerocher:
Thanks

PR #3018, opened 2020-04-03
Do not recreate Sharable channel handlers

croudet:


graemerocher:
Thanks for the contribution!

PR #3019, opened 2020-04-03
Fix #3013 - Add size validator for Object array

### LINKED ISSUE #3013: micronaut-validation does not respect `@Size` constraint on arrays
Perhaps this is expected as the documentation states that Micronaut's implementation is not fully compliant with the Bean Validator Spec, but it appears that Micronaut's implementation does not respect `@Size` constraints on arrays.

Things work as expected if I add the Hibernate Validator implementation.

### Task List

### Steps to Reproduce

1. Create a pojo with an array as a property
2. Annotate the array property with `@Size(min=3)`
3. Set an array with one element to the property in step 2
4. Validate the pojo

### Expected Behaviour

Validation should fail and throw a `ConstraintViolationException`

### Actual Behaviour

Validation succeeds

### Environment Information

- **Operating System**: Fedora Linux 30
- **Micronaut Version:** 1.3.3
- **JDK Version:** OpenJDK Runtime Environment (build 1.8.0_242-b08)

### Example Application

- https://github.com/JasonTypesCodes/mn-validation-array
See: PojoServiceTest.java

croudet:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/3013


graemerocher:
Thanks!

Issue #2615, opened 2020-01-09
Consul NewServiceEntry do not support meta


wenerme:
[Consul service](https://www.consul.io/api/catalog.html#service) has a meta field, but NewServiceEntry don't.

![image](https://user-images.githubusercontent.com/1777211/72059662-675d9880-330d-11ea-8b66-7b3b126bd686.png)

PR #3038, opened 2020-04-08
Allow configuration of metadata when registering an application in Consul

fnonnenmacher:
As #2615 describes, until now it was not possible to specify the attribute [meta](https://www.consul.io/api/agent/service.html#meta) in the  Consul registration customization.

This PR adds a new configuration property `consul.client.registration.meta` which allows the user to define custom metadata when registering a service in Consul.

 I checked the test which covers the tag property and I have tried to add corresponding tests for the metadata.

Different from the tag property, Consul [does not support](https://www.consul.io/api/health.html#parameters-2) simple querying based on the meta attribute. Therefore, I have not changed the `ConsulDiscoveryConfiguration`.

I have updated the configuration example in the documentation so that it includes now the new property. Running this example configuration results in the following Consul entry:
```
$ curl http://127.0.0.1:8500/v1/health/service/hello-world
[
    {
        "Node": {...},
        "Service": {
            "ID": "hello-world:8080",
            "Service": "hello-world",
            "Tags": [
                "hello",
                "world"
            ],
            "Address": "localhost",
            "Meta": {
                "instance-type": "t2.medium",
                "some": "value"
            },
            ...
        },
        "Checks": [...]
    }
]
```

### DIFF
--- a/discovery-client/src/main/java/io/micronaut/discovery/consul/ConsulConfiguration.java
+++ b/discovery-client/src/main/java/io/micronaut/discovery/consul/ConsulConfiguration.java
@@ -327,6 +327,7 @@ public class ConsulConfiguration extends DiscoveryClientConfiguration {
         public static final String PREFIX = ConsulConfiguration.PREFIX + "." + RegistrationConfiguration.PREFIX;
 
         private List<String> tags = Collections.emptyList();
+        private Map<String, String> meta = Collections.emptyMap();
         private CheckConfiguration check = new CheckConfiguration();
 
         /**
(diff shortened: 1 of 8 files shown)


fnonnenmacher:
During the [first build](https://github.com/micronaut-projects/micronaut-core/actions/runs/73939352) the following tests failed for Java 13:

* EurekaMockHeartbeatSpec > test that the server reports a heartbeat to Eureka
* EurekaContextPathSpec > test that the server reports a heartbeat to Eureka
* EurekaMockAutoRegistrationSpec > test that an application can be registered and de-registered with Eureka hyphenated

Rerunning the tests made them succeed. Do you have an idea? Are those tests failing from time to time e.g. when the Eureka container does not start in time?

graemerocher:
Due to the eventually consistent nature of Eureka those tests fail occasionally unfortunately.

graemerocher:
Thanks for the contribution!

PR #3045, opened 2020-04-10
Add Support For Oracle Cloud Vault

recursivecodes:
This feature adds support for Oracle Cloud Vault. Users can define a list of vaults in `bootstrap.yml` and each vault will be listed and the secrets within them retrieved and set in to configuration variables in the application. The feature

### DIFF
--- a/discovery-client/src/main/java/io/micronaut/discovery/oraclecloud/vault/config/OracleCloudVaultConfigurationClient.java
+++ b/discovery-client/src/main/java/io/micronaut/discovery/oraclecloud/vault/config/OracleCloudVaultConfigurationClient.java
@@ -79,6 +79,7 @@ public class OracleCloudVaultConfigurationClient implements ConfigurationClient 
      * @param oracleCloudVaultClientConfiguration   Oracle CloudVault Client Configuration
      * @param applicationConfiguration              The application configuration
      * @param executorService                       Executor Service
+     * @throws Exception                            If no configuration is provided
      */
     public OracleCloudVaultConfigurationClient(OracleCloudVaultClientConfiguration oracleCloudVaultClientConfiguration,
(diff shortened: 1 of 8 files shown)


graemerocher:
Thanks will take a look!

recursivecodes:
I've submitted an enhancement request internally to add support to our SDK for retrieving all secrets in a vault with a single call instead of N calls. If this gets implemented I will update this feature to use the new method.

discovery-client/src/main/java/io/micronaut/discovery/oraclecloud/vault/config/OracleCloudVaultConfigurationClient.java
```diff
@@ -0,0 +1,218 @@
...
+        this.executorService = executorService;
+
+        BasicAuthenticationDetailsProvider provider = null;
+        if (oracleCloudVaultClientConfiguration.isUseInstancePrincipal()) {
+            provider = InstancePrincipalsAuthenticationDetailsProvider.builder().build();
+        } else {
+            try {
+                provider = new ConfigFileAuthenticationDetailsProvider(oracleCloudVaultClientConfiguration.getPathToConfig(), oracleCloudVaultClientConfiguration.getProfile());
+            } catch (IOException e) {
+                e.printStackTrace();
```
graemerocher, line 98: Could you log this exception instead of dumping the stack trace?

recursivecodes: Fixed in https://github.com/micronaut-projects/micronaut-core/pull/3045/commits/a2b03f3d41e66db97edc2a67ee028a7fb1ae8d38

graemerocher:
Thanks for the contribution!

PR #3056, opened 2020-04-14
Fix #2066 - Add support for native transport channel options

### LINKED ISSUE #2066: Ability to use native transport options with Netty
### Steps to Reproduce

1. Try to use `TCP_CORK` or `TCP_QUICKACK`
2. Can't
3. See below in *Example Application* for code example

### Expected Behaviour

It would be really cool to be able to use higher-level options for `epoll` and `kqueue`

### Actual Behaviour

Instead, they are seen as "unknown options" on a given channel, even if `use-native-transport` is `true`. This is because Micronaut currently looks for socket options on the regular [`ChannelOption`](https://netty.io/4.0/api/io/netty/channel/ChannelOption.html), instead of knowing to look on the implementation-specific ones (like, say, [`EpollChannelOption`](https://netty.io/4.0/api/io/netty/channel/epoll/EpollChannelOption.html) and [`KQueueChannelOption`](https://netty.io/4.1/api/io/netty/channel/kqueue/KQueueChannelOption.html).

### Environment Information

- **Operating System**: `macOS 10.14.6` and `Alpine Linux`
- **Micronaut Version:** `master`
- **JDK Version:**
   ```
   java version "1.8.0_221"
   Java(TM) SE Runtime Environment (build 1.8.0_221-b11)
   Java HotSpot(TM) 64-Bit Server VM (build 25.221-b11, mixed mode)
   ```

### Example Application

**`application.yml`**
```yaml
micronaut:
  server:
    netty:
      use-native-transport: true
      childOptions:
        tcpNodelay: true
        tcpQuickack: true
````

*Note: I'm told one should never enable both `TCP_NODELAY` and `TCP_QUICKACK` at the same time, but I'm doing it here to illustrate that one is supported, and the other is not.*

Here is the output when trying to run an app with the above config:
```
17:25:25 [INFO]: Startup completed in 2152ms. Server Running: http://0.0.0.0:8080
17:25:40 [WARN]: Unknown channel option 'TCP_QUICKACK' for channel '[id: 0x23d15fe9, L:/0:0:0:0:0:0:0:1%0:8080 - R:/0:0:0:0:0:0:0:1%0:63150]'
[... app logs begin...]
```

croudet:
Options can be specified as regular options and childOptions.

Issue:  https://github.com/micronaut-projects/micronaut-core/issues/2066
alternative PR: https://github.com/micronaut-projects/micronaut-core/pull/2106

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupFactory.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupFactory.java
@@ -28,6 +28,7 @@ import io.netty.channel.socket.SocketChannel;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+
 import java.util.concurrent.Executor;
 import java.util.concurrent.ThreadFactory;
 
@@ -106,7 +107,6 @@ public class DefaultEventLoopGroupFactory implements EventLoopGroupFactory {
         } else {
             return this.defaultFactory.serverSocketChannelClass(configuration);
         }
-
     }
 
     @NonNull
@@ -118,4 +118,5 @@ public class DefaultEventLoopGroupFactory implements EventLoopGroupFactory {
             return this.defaultFactory.clientSocketChannelClass(configuration);
         }
     }
+
 }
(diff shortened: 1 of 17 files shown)


graemerocher:
Thanks for the contribution!

sgammon:
thank you for finishing this out :)

PR #3063, opened 2020-04-15
Add retry predicate support for @Retryable and @CircuitBreaker

Im-dex:


graemerocher:
So I am wondering, this predicate only has access to the exception that was thrown but not the `RetryState`. What if you wanted to programmatically customize retry based on the number of attempts etc.?

Im-dex:
> So I am wondering, this predicate only has access to the exception that was thrown but not the `RetryState`. ?

The predicate is just a more flexible alternative for includes/excludes and nothing else, it useful when you need to check exception properties before you make a decision: retry or not to retry.

> What if you wanted to programmatically customize retry based on the number of attempts etc.

The predicate designed as a decision-making tool, not as a retry logic customizer. Also, I can't imagine use case when I need to access attempts number to make a retry decision.

graemerocher:
Thanks for the contribution

PR #3066, opened 2020-04-16
Bug # 3065 - test case - TypeVisitorElement should not visit BaseTest…

croudet:
…Controller


croudet:
Here are the expected asserts
```
expect:
        AllElementsVisitor.VISITED_CLASS_ELEMENTS.size() == 1
        AllElementsVisitor.VISITED_CLASS_ELEMENTS[0].hasAnnotation(Controller)
```

Issue #3065, opened 2020-04-16
TypeElementVisitor visits classes that have not the required annotation


croudet:
### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

```java
    void "test visit inherited controller classes"() {
        buildBeanDefinition('test.TestController', '''
package test;

import io.micronaut.http.annotation.*;
import javax.inject.Inject;

@Controller("/test")
public class TestController extends BaseTestController {

    @Get("/getMethod")
    public String getMethod(int[] argument) {
        return null;
    }

}

class BaseTestController {

    public String hello(String name) {
        return name;
    }


}
''')
        expect:
        AllElementsVisitor.VISITED_CLASS_ELEMENTS[0].hasAnnotation(Controller)
        AllElementsVisitor.VISITED_CLASS_ELEMENTS[1].hasAnnotation(Controller) // should be false for one of the two
        AllElementsVisitor.VISITED_CLASS_ELEMENTS.size() == 1
    }
```

### Expected Behaviour

`BaseTestController` should not be visited.
`BaseTestController.hasAnnotation(Controller.class)` should be false

### Actual Behaviour

`BaseTestController` is visited.
`BaseTestController.hasAnnotation(Controller.class)` returns `true`.

### Environment Information

- **Operating System**: Windows 10
- **Micronaut Version:** 1.3.4
- **JDK Version:** jdk 14

### Example Application

Test case added:
- https://github.com/micronaut-projects/micronaut-core/pull/3066

jameskleeh:
Is that test case designed to fail or pass? The assertions don't make sense:

```
        AllElementsVisitor.VISITED_CLASS_ELEMENTS[0].hasAnnotation(Controller)
        AllElementsVisitor.VISITED_CLASS_ELEMENTS[1].hasAnnotation(Controller) // should be false for one of the two
        AllElementsVisitor.VISITED_CLASS_ELEMENTS.size() == 1
```

Why are you asserting there is only one element but also attempting to access index 1? The comment says one should be false but the test is asserting both are true. The test case should be designed so it passes if the behavior is changed.

croudet:
The `hasAnnotation` asserts must be removed. There are here to show that both classes returns true when it should not.

PR #3069, opened 2020-04-17
Issue #3065 - TypeElementVisitor visits classes that have not the required annotation

croudet:
https://github.com/micronaut-projects/micronaut-core/issues/3065

Potential fix.
Scan enclosed elements in type hierarchy and discards method overrides.


graemerocher:
Confirmed this PR doesn't cause issues with data. Thanks for the contribution

Issue #3072, opened 2020-04-20
Error deploying micronuat Function to Lambda


sdelamo:
```
io.micronaut.context.exceptions.BeanInstantiationException: Bean definition [io.micronaut.runtime.context.scope.refresh.RefreshInterceptor] could not be loaded: Failed to inject value for parameter [refreshScope] of class: io.micronaut.runtime.context.scope.refresh.RefreshInterceptor

Message: No bean of type [io.micronaut.runtime.context.scope.refresh.RefreshScope] exists. Make sure the bean is not disabled by bean requirements (enable trace logging for 'io.micronaut.context.condition' to check) and if the bean is enabled then ensure the class is declared a bean and annotation processing is enabled (for Java and Kotlin the 'micronaut-inject-java' dependency should be configured as an annotation processor).
Path Taken: new RefreshInterceptor([RefreshScope refreshScope])
at initializeContext (DefaultBeanContext.java:1365)
at initializeContext (DefaultApplicationContext.java:217)
at readAllBeanDefinitionClasses (DefaultBeanContext.java:2585)
at start (DefaultBeanContext.java:210)
at start (DefaultApplicationContext.java:169)
at startEnvironment (AbstractExecutor.java:120)
at buildApplicationContext (MicronautRequestHandler.java:82)
at handleRequest (MicronautRequestHandler.java:49)
at invoke0 (NativeMethodAccessorImpl.java:-2)
at invoke (NativeMethodAccessorImpl.java:-1)
at invoke (DelegatingMethodAccessorImpl.java:-1)
at invoke (Method.java:-1)```

`RefreshScope` is annotated with `@Requires(notEnv = {Environment.FUNCTION, Environment.ANDROID})`.

PR #3074, opened 2020-04-20
Don't load RefreshInterceptor in function and android environments.

sdelamo:
Don't load RefreshInterceptor in function and android environments.

Close: https://github.com/micronaut-projects/micronaut-core/issues/3072


graemerocher:
Thanks!

PR #3076, opened 2020-04-20
Make Java 14 buildable + CI build

szpak:
I wanted to play with the new Micronaut, but it turned out I was not able to build it with Java 14, which I (try to) use by default, due to the old Gradle version.

This PR fixes it and replaces Java 13 CI build with the one using Java 14.

Btw, I switched from `-all` to `-bin` for Gradle wrapper as it is more helpful while writing custom plugins. I may bring it back, if it was useful to develop micronaut-core.


graemerocher:
Thanks!

Issue #2901, opened 2020-03-09
Can't Introspect sources generated though an AnnotationProcessor


bidorffOL:
**EDIT**: Added example application

I have been trying to use an `AnnotationMapper` to add the `@Introspected` annotation to Immutables `@Generated` classes. However, it seems that `TypeElementVisitorProcessor` ignore those classes.

I believe that this is because `TypeElementVisitorProcessor` [only run on the first pass](https://github.com/micronaut-projects/micronaut-core/blob/479570f7a5af3e183bf46b04ef056957dbb1031b/inject-java/src/main/java/io/micronaut/annotation/processing/TypeElementVisitorProcessor.java#L62). However, classes generated through an AnnotationProcessor are not available until the second pass and therefore are ignored and can't be `@Introspected`.

### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

Create a simple `@Value.Immutable` class:

```
@Value.Immutable
abstract class MyClass {
    // No-op
}
```

And a `NamedAnnotationMapper`:
```
public class GeneratedAnnotationMapper implements NamedAnnotationMapper {

  @Nonnull
  @Override
  public String getName() {
    return "org.immutables.value.Generated";
  }

  @Override
  public List<AnnotationValue<?>> map(AnnotationValue<Annotation> annotation, VisitorContext visitorContext) {
    return Arrays.asList(AnnotationValue.builder("io.micronaut.core.annotation.Introspected").build(),
  }
}
```

### Expected Behaviour

A `BeanIntrospection` should be available for `ImmutableMyClass`.

### Actual Behaviour

No introspection is generated for this class.

### Environment Information

- **Operating System**: N/A
- **Micronaut Version:** 1.3.2
- **JDK Version:** 11.0.6

### Example Application

[https://github.com/bidorffOL/micronaut-issue-2901](https://github.com/bidorffOL/micronaut-issue-2901)

bidorffOL:
Any hope to see this issue resolved "soon" ? I'm having trouble finding a way to bypass that. I'm trying to use validation with Immutables, and the lack of support for @Introspected makes it difficult.

graemerocher:
Haven't had time to look into it yet, but PRs welcome!

bidorffOL:
[...]

PR #3090, opened 2020-04-23
Issue #2901 - Improve bean introspection

bidorffOL:
#2901 

This fix allows the `TypeElementVisitorProcessor` to be executed for each round. This implies several changes :

- the `LoadedVisitor` are initialized once and for all in the `init` method
- `javaVisitorContext.finish` is only called once every class has been generated and processed (this call is responsible for writing the services)
- some visitors needed to be modified to ensure that the files they generated were written only once
- a `@Generated` annotation was added to every generated classes, preventing them to be handled by the `TypeElementVisitorProcessor`

The PR may need some improvements (and an additional control that I didn't miss any Visitor), however the result seems to be OK on my end.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -291,6 +291,8 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
         String[] interfaces = getImplementedInterfaceInternalNames();
         classWriter.visit(V1_8, ACC_SYNTHETIC, className, null, !isInterface ? superType.getInternalName() : null, interfaces);
 
+        classWriter.visitAnnotation(TYPE_GENERATED.getDescriptor(), false);
+
         classWriter.visitField(ACC_FINAL | ACC_PRIVATE, FIELD_INTERCEPTORS, FIELD_TYPE_INTERCEPTORS.getDescriptor(), null, null);
         classWriter.visitField(ACC_FINAL | ACC_PRIVATE, FIELD_PROXY_METHODS, FIELD_TYPE_PROXY_METHODS.getDescriptor(), null, null);
     }
@@ -890,6 +892,8 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
                 null,
                 isInterface ? TYPE_OBJECT.getInternalName() : getTypeReference(targetClassFullName).getInternalName(),
                 interfaces);
+
+        proxyClassWriter.visitAnnotation(TYPE_GENERATED.getDescriptor(), false);
 
         // set $proxyMethods field
         proxyConstructorGenerator.loadThis();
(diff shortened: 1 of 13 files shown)


bidorffOL:
Unfortunately, some tests fail and I'm wondering what should be the expected behavior.

The `AllClassesVisitor` is expecting only user made classes/methods/fields to be listed, however, elements generated by Micronaut are listed as well (because they are not filtered by the processor on 2nd+ rounds) . Should this behavior be avoided (in which case I need to find a way to filter them) or is this an unexpected "bonus feature" (I doubt it, but who knows...).

graemerocher:
Yeah we don’t want to visit generated classes again

bidorffOL:
I though so. So my idea was to add a `@io.micronaut.core.annotation.Generated` with a `CLASS` retention (I couldn't find anything like that in the project) that would be added to every generated class and filtered out by the `TypeElementVisitorProcessor`, preventing any unwanted side effect.

I'll try to update my PR quickly, I'm open to any comment on this.

graemerocher:
Seems like a reasonable approach

graemerocher:
Sorry could you send the PR to the `master` branch for Micronaut 2.0? Maybe to big a change to go into `1.3.x`

bidorffOL:
Yes, I actually did set the `@since` to 2.0 because I wasn't expecting it to be merged before, it was a mistake on my end to put it in 1.3.x, I'll update it.

graemerocher:
Thanks, now the only thing I am concerned about is lack of a test

bidorffOL:
I'm new to Groovy testing, but I'll try to see what I can do about that!

graemerocher:
Test can be in Java too, we are not picky :)

bidorffOL:
Don't worry, its always nice to learn new stuff ^^

I have added a couple of tests (actually the same one in both Java & Groovy) and took the opportunity to ignore `@Generated` classes in Groovy too. I'll try to think about how I can test the effects of the multi round processing but right now I have no idea.

graemerocher:
We have another similar test which is https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/test-suite/src/test/java/io/micronaut/docs/inject/generated/VerifyDefinitionInjectionSpec.java

That uses https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/test-suite-helper/src/main/java/io/micronaut/testsuitehelper/TestGeneratingAnnotationProcessor.java

Note sure if that helps.

bidorffOL:
It does actually help a lot: I just need to build a similar file as `io.micronaut.test.generated.Example`, make it `@Introspected` and check if an Introspection exists.

graemerocher:
You done with the changes? if so looks good to me.

bidorffOL:
I think I am, I got everything I was think of in the PR

graemerocher:
Thanks for the contribution!

bidorffOL:
Thank you for your time and your advices ! :)

PR #3094, opened 2020-04-24
Protect instrumenters from chained invocation to prevent corrupting their internal state

### LINKED ISSUE #3093: Flowable.merge corrupts opentracing ScopeManager state
### Steps to Reproduce

1. When `tracer.activeSpan() == null`
2. Build and start a new span.
3. Execute a `Flowable.merge` operation in the scope of that span (in the same Thread).
4. Close the scope and finish the span.
5. Check `tracer.activeSpan()` afterwards.

### Expected Behaviour

The active span should be `null`.

### Actual Behaviour

The active span is not null, it contains infromation leaked from the merge operation's span context.

### Environment Information

- **Operating System**: macOS 10.15.4
- **JDK Version:** 11.0.5

### Example Application

I have created a [test case](https://github.com/Tesco/micronaut-core/blob/rxinstrumented-fix/tracing/src/test/groovy/io/micronaut/tracing/instrument/util/OpenTracingInvocationInstrumenterSpec.groovy) for this. The issue is generic though effecting potentially all `InvocationInstrumenter` implementations which do not tolerate chained invocation of their before/afterInvocation() methods, see [this generic test case](https://github.com/Tesco/micronaut-core/blob/rxinstrumented-fix/runtime/src/test/groovy/io/micronaut/reactive/rxjava2/RxChainedInstrumentationSpec.groovy). I posted this as a [question on gitter](https://gitter.im/micronautfw/questions?at=5e98177ce920022432aa1305) earlier on, as I was unsure whether this violates the contract of the `InvocationInstrumenter`. I strongly believe this should not be allowed as it is pretty hard to properly handle any internal state in the instrumenter instances if chained invocations can happen.

I will also post my proposed fix in a PR and link to this issue.

lgathy:
Fixes #3093 

Some `RxInstrumentedComponent` implementation classes may trigger chained invocations to their instrumenter instance which can result in corrupting the internal state of that instrumenter (test cases included).

Example: In case of `RxInstrumentedSubscriber` a call to its `onSubscribe()` can trigger a chained `onNext()` on the same instance. If chained instrumentation should not be supported (see: linked issue), then this `onNext()` should not call before/afterInvocation() on the instrumenter. Currently it does, but none of the instrumenters are protected against that.

The proposed change fixes this by tracking if the instrumentation is already active in a boolean field. Three things to consider:
1. Altough my tests only trigger this chained execution on `RxInstrumentedSubscriber` I have applied the same protective measures on other similar classes too where I believe similar situation can happen. Others like `RxInstrumentedSingle` do not seem suspicious to me.
2. I have only seen this chained behaviour happen on the same thread, so memory visibility of the `active` field should not cause problems
3. These classes are no longer thread-safe. I think it is safe to assume, that they will not be accessed concurrently, but I am not 100% sure about it.


graemerocher:
Thanks for doing the research and providing the fix. Could you move `OpenTracingInvocationInstrumenterSpec` to the `test-suite` subproject and remove the `jaeger` dependency. We have all of the `jaeger` tests in `test-suite` to isolate it from zipkin.

lgathy:
Sure. Good to know that, I was not aware.

lgathy:
Strange failure in `ConsulClientSpec`, no matter how I try to reproduce this, always passes for me. Can we rerun the checks? This does not seem to depend on what I have changed (I don't see instrumented components). Alternatively, is it possible to retrieve the test logs from these runs or maybe even the logs of the consul container this test used?

graemerocher:
Yeah we have trouble with that test failing intermittently on github actions. It doesn't relate to your changes no.

graemerocher:
Thanks for the contribution!

PR #3101, opened 2020-04-27
Fix overwritting of span operation name in OpenTracingServerFilter

### LINKED ISSUE #3084: Open tracing operation name set twice, overriding user customisation
Hi,

You are resolving name of request and setting operation name twice. Once is done after response which overrides any changes user might want to do on active scope.

The call that looks wrong is here:
https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/tracing/src/main/java/io/micronaut/tracing/instrument/http/OpenTracingServerFilter.java#L96

First, correct time is done here:
https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/tracing/src/main/java/io/micronaut/tracing/instrument/http/AbstractOpenTracingFilter.java#L109-L111

Which is called here:
https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/tracing/src/main/java/io/micronaut/tracing/instrument/http/OpenTracingServerFilter.java#L76

I think the second call should be removed.

KrzysztofKowalczyk:
Simple PR, just removed 2 lines that set span operation name second time which is not needed and make it impossible for user to change it. Added test to show it.

Fixes #3084 .

### DIFF
--- a/test-suite/src/test/groovy/io/micronaut/tracing/jaeger/HttpTracingSpec.groovy
+++ b/test-suite/src/test/groovy/io/micronaut/tracing/jaeger/HttpTracingSpec.groovy
@@ -334,6 +334,27 @@ class HttpTracingSpec extends Specification {
         context.close()
     }
 
+    void "tested customising span name"() {
+        given:
+        ApplicationContext context = buildContext()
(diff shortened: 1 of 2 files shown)


graemerocher:
Thanks for the contribution

PR #3105, opened 2020-04-28
Add support for @RequestBean

asodja:
This PR adds new Bindable annotation @RequestBean. With this annotation you can bind all request bindable values to one object. Example:

```
    @Controller('/rest')
    static class MyController {

        @Get("/{pathVariable}{?queryValue}")
        String getQueryValue(@RequestBean Bean bean) {
            return "Some result"
        }

        @Getter
        @AllArgsConstructor
        @Introspected
        static class Bean {
            @Nullable
            @QueryValue
            String queryValue

           @PathVariable
           String pathVariable

           @Nullable
           @Header("some-header")
           String header

        }
    } 
```

Motivation:
I was testing some Jax-rs application with Micronaut and noticed that there is no construct that is equal to `@BeanParams`. You can use exploded query params, but that is not the same. And in case you move some Jax-rs application to Micronaut and you are heavy user of `@BeanParams` this might not be easy since you have to change a lot of signatures and also micronaut-rs does not support `@BeanParams` (so with this micronaut-rs could support also `@BeanParams`).

Implementation details:
- Added new annotation @RequestBean and RequestBeanAnnotationBinder
- Every Bean marked with @RequestBean have to have `@Introspected` since object is constructed with Instrospection
- Added RequestBeanParameterRule that checks that object can be constructed via @Instropcted with all values. I added it as RouteValidationRule, although this should not be RouteValidationRule since user can disable that, but did not find where this is normally done.
- Added support in NullableParameterRule and MissingParameterRule
- Currently it supports Bindable arguments and HttpRequest. 

In case this feature is doable I already see some improvements:
- Add support for Typed Request Arguments (mapped with TypedRequestArgumentBinder), this should be easy to do but I was not able to test it. This would be cool since you could also have properties of type `Authentication` and friends in your bean
- Add annotation @RequestBeanConfiguration (or something shorter) that would be added to RequestBean class. Currently I would like to have all fields automatic marked as nullable. This will then affect validation (all fields are treated as optional) and bean initialization. This would be again really cool when moving some old app to Micronaut so you don't have to mark all fields with Nullable. Example:
```
 @RequestBeanConfiguration(autoNullable = true)
 class MyBean {
       // This is treated as nullable 
       @QueryValue
       String myValue
  }
```

### Things to consider
I am not familiar with Micronaut code base so I don't know what issues can this feature bring or if this will even work for all cases (from my tests it looks that it works). And I also don't know what should also be done to fully support this (probably at least support for OpenApi has to be added). But from this implementation it seems that it is not that complex.

So is this a feature that could get into the Micronaut or are there some issues that can arise?

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/aop/RequestBeanSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/aop/RequestBeanSpec.groovy
@@ -1,0 +1,257 @@
+package io.micronaut.http.client.aop
+
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.core.annotation.Introspected
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.Header
+import io.micronaut.http.annotation.PathVariable
+import io.micronaut.http.annotation.QueryValue
+import io.micronaut.http.annotation.RequestBean
+
+import io.micronaut.http.client.annotation.Client
+import io.micronaut.http.client.exceptions.HttpClientResponseException
+import io.micronaut.runtime.server.EmbeddedServer
+import spock.lang.AutoCleanup
+import spock.lang.Shared
+import spock.lang.Specification
+
+import javax.annotation.Nullable
+import javax.validation.Valid
+import javax.validation.constraints.Pattern
+
+class RequestBeanSpec extends Specification {
+
+    @Shared
+    @AutoCleanup
+    EmbeddedServer embeddedServer = ApplicationContext.run(EmbeddedServer)
+
+    @Shared
+    RequestBeanClient client = embeddedServer.getApplicationContext().getBean(RequestBeanClient)
+
+    void "test @QueryValue is injected to Bean"() {
+        expect:
+            client.getQueryValue("Riverside") == "Riverside"
+    }
+
+    void "test @QueryValue have empty Optional Value"() {
+        expect:
+            !client.getOptionalValue()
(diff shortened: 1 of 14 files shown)


graemerocher:
Wow, nice contribution! Thanks and will review soon

graemerocher:
@asodja in the meantime could you sign the CLA? Thanks

asodja:
Of course, signed.

asodja:
Any updates on reviewing this one? I am really looking forward for this feature.

One thing I noticed: 
In case RequestBean object is immutable I try to instantiate it with primary constructor by matching properties and constructor arguments by name (there is also compile time Rule that validates that this is possible). Although probably this is not possible when this class is from another jar and compiled without `-parameters` (similar issue as with controller method arguments).

Any recommendations how to deal with that? 

I could fallback to Reflection in that case, but that is probably not in the nature of Micronaut.

graemerocher:
Yes, will review today. Your approach with constructors is fine. My only doubt is the name of the annotation right now. In JAX-RS this is equivalent to https://docs.oracle.com/javaee/7/api/javax/ws/rs/BeanParam.html

asodja:
I am open for any suggestion. RequestBean feels a bit long anyway.

graemerocher:
I merged this and added some simplifications https://github.com/micronaut-projects/micronaut-core/commit/3f2c26654e086ab40413e24f81afa478dcf54bce

Thanks for the contribution. The feature could use some documentation. If you have some time to put together a PR for that it would be great.

asodja:
Nice, thank you. I can contribute docs also, I will make a PR in the next 7 days.

PR #3109, opened 2020-04-28
Fix references to reactive postgres module and add mysql module

benmccann:
See also https://github.com/micronaut-projects/micronaut-sql/pull/149


benmccann:
I'm not sure why the cache is failing. I closed and reopened the PR to run the CI again, but same result

graemerocher:
Yeah seems nothing to do with the PR

Issue #3117, opened 2020-04-29
io.micronaut.kotlin dependences not in build


sdelamo:
Dependencies such as

`io.micronaut.kotlin:micronaut-kotlin-runtime` are not in buildDependencies

https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/build.gradle#L46-L538

PR #3118, opened 2020-04-29
feat: add io.micronaut.kotlin modules

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-core/issues/3117

See: https://mvnrepository.com/search?q=io.micronaut.kotlin


graemerocher:
Fixed by https://github.com/micronaut-projects/micronaut-core/commit/a072aa733d5e3a2b01bd89d730f1832ac3c3eefb

For future reference all dependencies should be included together based on the group, not individual entries for each dependency

Issue #2614, opened 2020-01-09
Add a String to char[] converter


goeh:
Configuring password for redis results in an error as @chenjpu described in https://github.com/micronaut-projects/micronaut-core/issues/1800

### Steps to Reproduce

1. Create Micronaut application with redis-lettuce feature.
2. Configure redis.host and redis.password in application.yml
3. Start a Redis instance on localhost (docker run -p 6379:6379 redis redis-server  --requirepass secret)
4. Start the application
5. curl http://localhost:8080/foo

### Expected Behaviour

Application should start and use the configured password to connect to redis.
The endpoint should return 404 because key 'foo' does not exist in redis.

When the issue is fixed you should be able to set a value for a key and retrieve it.

```
curl -X POST -H 'Content-Type: text/plain' -d "hello world" http://localhost:8080/foo
curl http://localhost:8080/foo
hello world*
```

### Actual Behaviour

```
Message: Error resolving property value [redis.password]. Property doesn't exist
Path Taken: new RedisController([RedisService redisService]) --> new RedisService([StatefulRedisConnection redis]) --> StatefulRedisConnection.redisConnection([RedisClient redisClient]) --> RedisClient.redisClient([AbstractRedisConfiguration config]) --> DefaultRedisConfiguration.setPassword([char[] password])
        at io.micronaut.context.AbstractBeanDefinition.getValueForMethodArgument(AbstractBeanDefinition.java:795)
        at io.micronaut.configuration.lettuce.$DefaultRedisConfigurationDefinition.injectBean(Unknown Source)
        at io.micronaut.configuration.lettuce.$DefaultRedisConfigurationDefinition.build(Unknown Source)
        at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:1598)
        at io.micronaut.context.DefaultBeanContext.createAndRegisterSingleton(DefaultBeanContext.java:2307)
        at io.micronaut.context.DefaultBeanContext.getBeanForDefinition(DefaultBeanContext.java:1989)
        at io.micronaut.context.DefaultBeanContext.getBeanInternal(DefaultBeanContext.java:1963)
        at io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:1082)
        at io.micronaut.context.AbstractBeanDefinition.getBeanForConstructorArgument(AbstractBeanDefinition.java:1007)
[...]

PR #3121, opened 2020-04-29
implemented char converter

pollend:
This implemented a string to char[] converter into the default environment for #2614 and allows strings to be converted to a char array when assigned to a configuration. is this the intended behavior because the issue discusses some other problems with this i.e a method that accepts a string and another method for a char[]. would it have to be implemented through another means to avoid this case?

### DIFF
--- a/inject/src/main/java/io/micronaut/context/converters/StringToCharArray.java
+++ b/inject/src/main/java/io/micronaut/context/converters/StringToCharArray.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


inject/src/main/java/io/micronaut/context/converters/StringToCharArray.java
```diff
@@ -0,0 +1,13 @@
+package io.micronaut.context.converters;
+
+import io.micronaut.core.convert.ConversionContext;
+import io.micronaut.core.convert.TypeConverter;
+
+import java.util.Optional;
+
+public class StringToCharArray implements TypeConverter<CharSequence, char[]> {
```
jameskleeh, line 8: This class is missing the license header and javadoc

Issue #2820, opened 2020-02-17
Performance problems with DefaultSyncCache


phxql:
### Task List

- [x] Steps to reproduce provided

### Steps to Reproduce

1. Write a method which returns a huge map and annotate it with `@Cacheable`
1. Call that method
1. Call it again

### Expected Behaviour

2nd method call is at least as fast as 1st.

### Actual Behaviour

2nd method call is a lot slower.

### Environment Information

- **Operating System**: Fedora 31
- **Micronaut Version:** 1.3.0
- **JDK Version:** AdoptJDK 11.0.6

Hi,

the DefaultSyncCache uses [the conversion service](https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/runtime/src/main/java/io/micronaut/cache/DefaultSyncCache.java#L56) to [convert the stuff](https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/runtime/src/main/java/io/micronaut/cache/DefaultSyncCache.java#L124) coming from the actual cache implementation. The `DefaultConversionService` contains some code which iterates over all the keys from map (because it implements Iterable) and tries to convert them. I'm not sure why it's doing this, but this caused a huge performance hit in our application. We're using the default in-memory caffeine cache.

graemerocher:
Should be checking if conversion is needed before attempting conversion

jameskleeh:
@phxql So you understand the root of this behavior. When the map is stored in the cache, any information about its generics are lost. We don't know what the key type or value type of the map is. When we retrieve the map, it is converted to being a `Map<String, String>` (or whatever the return type is). What we can do is add an option to the `@CacheConfig` annotation to allow disabling of the conversion.

alvarosanchez:
What if we would allow to define key and value types ?

phxql:
> @phxql So you understand the root of this behavior. When the map is stored in the cache, any information about its generics are lost. We don't know what the key type or value type of the map is. When we retrieve the map, it is converted to being a `Map<String, String>` (or whatever the return type is). What we can do is add an option to the `@CacheConfig` annotation to allow disabling of the conversion.

Fine for me :)

jameskleeh:
This was implemented in the cache module

PR #3132, opened 2020-05-01
Make cache conversion configurable

jameskleeh:
Closes https://github.com/micronaut-projects/micronaut-core/issues/2820


graemerocher:
Hasn't this logic been moved to https://github.com/micronaut-projects/micronaut-cache/

Would this be better applied there?

Issue #2617, opened 2020-01-09
Add documentation for MessageSource features


DrewEaster:
### Steps to Reproduce

I looked up the Micronaut online documentation for information about localization support and could not find anything. I discovered via the source code that Micronaut does have support for this, but it's not documented anywhere. 

### Expected Behaviour

The online documentation should provide full details on how to introduce localization into a Micronaut application using MessageSources, including how to construct MessageContexts with variables, and how to parameterize externalized message strings that reference said variables.

### Actual Behaviour

The desired documentation does not exist.

wamiota:
@graemerocher 
Hi, I am new to this Micronaut Framework project which I am strongly interested in.
I want to contribute to this project and look for an easy issue.
I think this issue is good for me to start contributing. Can I try this one?

graemerocher:
Absolutely. Look forward to your contributions 👍

PR #3134, opened 2020-05-02
Modify some typos and make a line readable

wamiota:
During reading codes for working on [this issue](https://github.com/micronaut-projects/micronaut-core/issues/2617), I found two typos in `StaticMessageSource`.
I think the phrase `The the message` is `The message`, so I modified.
And I think [here](https://github.com/micronaut-projects/micronaut-core/blob/88a809f176ddf85c0c8456f45f4fd527cfbf8257/inject/src/main/java/io/micronaut/context/StaticMessageSource.java#L72-L74) is more readable by coding with one line.
`return Optional.of(msg);`


jameskleeh:
Thanks!

PR #3144, opened 2020-05-04
Changed the LoggersEndpoint to default to be disabled. Previously was enabled by default.

JoeBeeton:
Changed the LoggersEndpoint to default to be disabled. Previously was enabled by default.
To enable the LoggersEndpoint the following property needs to be set.
endpoints.loggers.enabled = true
This is due to the LoggersEndpoint allows unauthenticated users to modify the logging configuration of the service by default. This is inconsistent with the other management endpoints that allow modification of the internal state of the application, such as CachesEndpoint or ServerStopEndpoint which default to disabled.

Also updated the tests to validate the new behaviour and updated the documentation to reflect this change.

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/loggers/LoggersEndpoint.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/loggers/LoggersEndpoint.java
@@ -57,7 +57,7 @@ public class LoggersEndpoint {
     /**
      * Endpoint default sensitivity.
      */
-    public static final boolean DEFAULT_SENSITIVE = false;
+    public static final boolean DEFAULT_SENSITIVE = true;
 
     private final LoggingSystem loggingSystem;
     private final LoggersManager<Map<String, Object>> loggersManager;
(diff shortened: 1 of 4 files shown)


jameskleeh:
Making the loggers endpoint disabled by default does not change the core issue here that it allows unauthorized users to modify it. It should be sensitive by default. You can update this PR to make the endpoint sensitive by default, and also ensure to list this breaking change in `breaks.adoc`

jameskleeh:
Also please sign the CLA

JoeBeeton:
Hello, I'm waiting on approval to sign the CLA, that may take some time. But in the mean time I'll make the requested changes.

graemerocher:
@jameskleeh I think we should just independently apply this change and close the PR

JoeBeeton:
Hello. Apologies, we are still waiting on getting approval of the CLA.  Ill check what the status is today.

JoeBeeton:
Hi, still waiting, if this is going to hold up a release please do as you say "independently apply" the change and close the PR.

graemerocher:
Applied manually. Thanks anyway

JoeBeeton:
No problem, just glad the change is in place.

PR #3168, opened 2020-05-07
Generate resource-config.json for GraalVM

### LINKED ISSUE #3108: Modify GraalTypeElementVisitor to generate resource-config.json
Modify `GraalTypeElementVisitor` to generate the file `resource-config.json` and include all the files in `resource` directory.

ilopmar:
Closes #3108

### DIFF
--- a/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
+++ b/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
@@ -32,6 +32,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Writer;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
 import java.util.function.Function;
@@ -65,9 +66,21 @@ public class GraalTypeElementVisitor implements TypeElementVisitor<Object, Objec
      */
     protected static Set<String> arrays = new HashSet<>();
 
+    /**
+     * Files in resources directory to generate metadata.
+     */
+    protected static Set<String> resourceFiles = new HashSet<>();
+
     private static final TypeHint.AccessType[] DEFAULT_ACCESS_TYPE = {TypeHint.AccessType.ALL_DECLARED_CONSTRUCTORS};
     private static final String REFLECTION_CONFIG_JSON = "reflection-config.json";
     private static final String NATIVE_IMAGE_PROPERTIES = "native-image.properties";
+
+    private static final String BASE_RESOURCE_CONFIG_JSON = "src/main/graal/resource-config.json";
+    private static final String RESOURCE_CONFIG_JSON = "resource-config.json";
+    private static final String RESOURCES_DIR = "src/main/resources";
+    private static final String RESOURCES = "resources";
+    private static final String BUNDLES = "bundles";
+    private static final String PATTERN = "pattern";
 
     private static final String BASE_REFLECT_JSON = "src/main/graal/reflect.json";
 
@@ -218,6 +231,7 @@ public class GraalTypeElementVisitor implements TypeElementVisitor<Object, Objec
             executed = true;
 
             generateNativeImageProperties(visitorContext);
+            generateResourceConfig(visitorContext);
         }
     }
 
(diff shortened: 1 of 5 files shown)


graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -216,85 +225,175 @@ public final void finish(VisitorContext visitorContext) {
...
+
+        try {
+            String path = buildNativeImagePath();
+            String reflectFile = path + REFLECTION_CONFIG_JSON;
+            String propsFile = path + NATIVE_IMAGE_PROPERTIES;
+
+            visitorContext.visitMetaInfFile(propsFile).ifPresent(gf -> {
+                visitorContext.info("Writing " + NATIVE_IMAGE_PROPERTIES + " file to destination: " + gf.getName());
+                try (PrintWriter w = new PrintWriter(gf.openWriter())) {
+                    w.println("Args = -H:ReflectionConfigurationResources=${.}/reflection-config.json");
```
graemerocher, line 268: I don't think we need this business. If a file exists call `reflect-config.json` then GraalVM will pick it up automatically

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -216,85 +225,175 @@ public final void finish(VisitorContext visitorContext) {
...
+                } catch (IOException e) {
+                    e.printStackTrace();
```
graemerocher, line 270: Better error handling here please

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -216,85 +225,175 @@ public final void finish(VisitorContext visitorContext) {
...
+                }
+            });
+        } finally {
+            classes.clear();
+            arrays.clear();
+        }
+    }
+
+    private void generateResourceConfig(VisitorContext visitorContext) {
+        ObjectMapper mapper = new ObjectMapper();
```
graemerocher, line 301: We should be using a shared object mapper instead of creating a new one for each operation

inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
```diff
@@ -99,6 +104,39 @@
         return Collections.emptyList();
     }
 
+    /**
+     * Obtain the project directory.
+     *
+     * @return An optional wrapping the project directory
+     */
+    default Optional<Path> projectDir() {
```
graemerocher, line 112: Rename to `getProjectDir()` also should be implemented in the child implementations so that the computed `Path` is cached and the computation is not done each time this is called. This can be done by simply calling the super interface

inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
```diff
@@ -37,7 +40,10 @@
  * @since 1.0
  */
 public interface VisitorContext extends MutableConvertibleValues<Object>, ClassWriterOutputVisitor {
+
     String MICRONAUT_BASE_OPTION_NAME = "micronaut";
+    String MICRONAUT_PROJECT_DIR = "micronaut.project.dir";
```
graemerocher, line 45: Should be declared as an annotation processor property. See `@SupportedOptions` on `BeanDefinitionInjectProcessor`. Also I think we should be consistent and name the property `micronaut.processing.project.dir`

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -216,85 +225,175 @@ public final void finish(VisitorContext visitorContext) {
...
-            } finally {
-                classes.clear();
-                arrays.clear();
+            } catch (Exception e) {
+                // skip processing resources
+            }
+        }
+    }
+
+    private Set<String> findResourceFiles(File folder, List<String> filePath) {
```
graemerocher, line 352: Would probably be better implemented using a file tree walker https://docs.oracle.com/javase/tutorial/essential/io/walk.html

But ok

inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
ilopmar: I don't think it is necessary to implement the method on the child implementations. The first thing we do in it is:
```java
Optional<Path> projectDir = get(MICRONAUT_PROJECT_DIR, Path.class);
if (projectDir.isPresent()) {
    return projectDir;
}
```
That will get the cached directory once it has been calculated.

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
ilopmar: I need to revert the change. The code is necessary to tell GraalVM to pickup the generated file. Without it, all the files generated during Micronaut compilation are not picked up by native-image. I've tested with a sample app it failed.

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
ilopmar: I've give it a try and created a FileVisitor but it's not that simple to keep the current directory to prepend it to every file during the visit. As you are ok with this I think I'll keep the current implementation.

Issue #3124, opened 2020-04-30
Deadlock while trying to send event from another thread inside @PostConstruct


Im-dex:
Micronaut version: 1.3.4

It seems that micronaut stuck in deadlock at DefaultBeanContext:2588 when you trying to do some asynchronous work in @PostConstruct which leads to event sending. 

https://github.com/Im-dex/micronaut-deadlock

graemerocher:
This is probably #3070

Thanks for the reproducer, will take a look

graemerocher:
Added a `publishEventAsync` method which will solve your use case

Im-dex:
@graemerocher The issue is a little bit more complex than just sending events from user code. I have found out this issue while calling declarative http client (annotated with @Retryable) within @PostConstruct method. When retry happens it is trying to send event and stuck in a deadlock. I think that micronaut itself could contain a lot of code sending events in a synchronous manner and this fix does nothing with that.

graemerocher:
Whilst I agree, unfortunately great care needs to be taken when it comes to using `@PostConstruct` and resolving beans at the same time because when `@PostConstruct` is invoked the bean is currently in the middle of being created and is holding a lock on the bean registry until it is created to make sure that it is impossible to create duplicate singletons.

If users have complex logic within a `@PostConstruct` then I would recommend ensuring the beans involved in that method are annotated with `@Context` to ensure they are initialized prior  to `@PostConstruct` being called.

Alternatively in Micronaut 2.x you can consider using the eager initialization features.

PR #3184, opened 2020-05-07
Provide API and solution sending events async 

graemerocher:
Provides solution for #3124 

The actual problem is not solvable because a lock is placed on the bean registry to avoid duplicate singletons being created so removing this lock would break correctness. 

However the issue can be resolved by looking up the event listeners before firing events. This change adds an API to do that and provides a solution for this use case.

Debatable whether it should go into 1.3.x or not.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/event/ApplicationEventPublisher.java
+++ b/inject/src/main/java/io/micronaut/context/event/ApplicationEventPublisher.java
@@ -16,7 +16,6 @@
 package io.micronaut.context.event;
 
 import javax.annotation.Nonnull;
-import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


inject/src/main/java/io/micronaut/context/event/ApplicationEventPublisher.java
```diff
@@ -34,4 +35,15 @@
      * @param event The event to publish
      */
     void publishEvent(@Nonnull Object event);
+
+    /**
+     * Publish the given event. The event will be published synchronously and only return once all listeners have consumed the event.
```
jameskleeh, line 40: Javadoc needs updated

PR #3196, opened 2020-05-08
Move the lock to allow parallel bean creation. Fixes #2867

### LINKED ISSUE #2867: @Parallel not working for multiple factory beans
### Task List

### Steps to Reproduce

1. Checkout https://github.com/tobiasschaefer/micronaut-parallel-bug
2. Start the application
3. Look at startup time shown in console, e.g. "Startup completed in 10999ms"

### Expected Behaviour

In the class temp.Foo there are two factory beans, both annotated with @Parallel and a Thread.sleep(5000).

I expect a startup time of a bit more that 5 seconds because both should be executed in parallel.

### Actual Behaviour

I see a startup time of a bit more than 10 seconds.

@jameskleeh said "Yeah it looks like there is a synchronization block where there shouldn’t be
Can you file an issue?"

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** 1.3.2 (current)
- **JDK Version:** 11.0.2

### Example Application

https://github.com/tobiasschaefer/micronaut-parallel-bug

graemerocher:


graemerocher:
PR has issues that need to be considered

PR #3209, opened 2020-05-11
PR test bug report. Property with kebab case is missing in 1.3.5 (1.3.4 works)

dstepanov:
Property with kebab case is missing in `@ConfigurationProperties` bean.


croudet:
I think it is related to this commit:
https://github.com/micronaut-projects/micronaut-core/commit/bad3ea93713946afe23ae8072a369629590732a8#diff-9db4fb0351fdafdad56c738a217e0276

dstepanov:
@croudet Thanks, I found what changed.
@graemerocher What do you think about this fix?

graemerocher:
The issue with the fix is that `MapToObjectConverter` is also used for binding JSON in controllers. I believe we need to alter this so that `MapToObjectConverter` is no longer used for binding lists and maps in `@ConfigurationProperties`

jameskleeh:
@dstepanov Thanks for your work on this issue!

Issue #3243, opened 2020-05-14
Add a root bean into the validation context.


l0co:
This is a feature request for `micronaut-validation`.

You could extend `ConstraintValidatorContext` with method returning validated bean. It currently returns only a clock provider:

```java
public interface ConstraintValidatorContext {
    @NonNull ClockProvider getClockProvider();
}
```

But it could return also a `rootBean`: 

```java
public interface ConstraintValidatorContext {
    @NonNull ClockProvider getClockProvider();
    @Nullable Object getRootBean();
}
```

You have everything required beneath, just need to add this method to the interface and to (private final) `DefaultConstraintValidatorContext`.

Usecase: contextual validation on which one can check something additional about the root bean, for example if it implements some interface, etc.

Related discussions:

1. https://stackoverflow.com/questions/37750656/how-to-access-a-field-which-is-described-in-annotation-property
1. https://hibernate.atlassian.net/browse/BVAL-237

They aren't going to implement this in hibernate validator because "they aren't going to".

l0co:
In fact, there's a root bean accessible now from the context, but only with a reflection.

graemerocher:
Since this may be specific to your use case maybe you would like to put together a PR for what you have in mind.

jameskleeh:
@l0co Based on your example you included in the hibernate issue, it seems the validation annotation should be applied to the root object since it relies on multiple properties of that object

l0co:
@jameskleeh This kind of solution is ugly, dirty and not very fantastic from the performance point of view. To achieve simple goals you have to use reflection and iterate through fields to find what you want. See the example [here](https://stackoverflow.com/a/37814382/320761).

Having the root bean in validation context the implementation from the link above could look like this:

```java
@Entity public class User {

    @NotBlank
    private String password;

    @Match(field = "password")
    private String passwordConfirmation;
}

public class MatchValidator implements  ConstraintValidator<Match, Object> {

    @Override
    public boolean isValid(final Object o, final ConstraintValidatorContext context) {
        String fieldName = context.getAnnotation().field();
        Object rootBean = context.getRootBean();
        return o.equals(rootBean.getField(fieldName).get(rootBean));
    }
    
}
```

jameskleeh:
```
@Matches("password", "passwordConfirmation")
@Introspected
@Entity public class User {

    @NotBlank
    private String password;

    private String passwordConfirmation;
}

[...]

PR #3263, opened 2020-05-18
Provide root bean in ConstraintValidatorContext

l0co:
https://github.com/micronaut-projects/micronaut-core/issues/3243


graemerocher:
Thanks for the contribution

PR #3264, opened 2020-05-18
Changed Publisher to Maybe Conversion to use Flowable

philippreston:
Changed the conversion of Publisher to Maybe to use Flowable to allow support of the source emitting 0 items.  

In the current implementation, it uses `Single.fromPublisher(publisher)` which means that in case where the publisher emits 0 items, the Single will throw:

```
java.util.NoSuchElementException: The source Publisher is empty
```

For a Maybe - it's valid that the source can be empty.


philippreston:
Relates to issue https://github.com/micronaut-projects/micronaut-core/issues/3265

graemerocher:
Thanks for the contribution

PR #3268, opened 2020-05-18
Copy headers from original response when returning not modified.

croudet:
Otherwise headers added by filters are lost.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/FileTypeHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/FileTypeHandler.java
@@ -47,6 +47,9 @@ import java.util.Arrays;
 @Singleton
 public class FileTypeHandler implements NettyCustomizableResponseTypeHandler<Object> {
 
+    // sorted array of entity headers
+    // https://tools.ietf.org/html/rfc2616#section-7.1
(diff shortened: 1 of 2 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/FileTypeHandler.java
```diff
@@ -157,8 +157,13 @@ protected void setDateHeader(MutableHttpResponse response) {
         headers.date(now);
     }
 
-    private FullHttpResponse notModified() {
+    private static void copyHeaders(NettyMutableHttpResponse<?> from, NettyMutableHttpResponse to) {
+        from.getHeaders().forEachValue((header, value) -> to.getHeaders().set(header, value));
```
graemerocher, line 161: I think `set(..)` will override the header for multi value headers and `add` should be used instead

croudet: Done.

croudet:
Build failed:
```
Caused by:
        org.testcontainers.containers.ContainerFetchException: Can't get Docker image: 
```

graemerocher:
Yes test containers is down because of https://status.quay.io/

jameskleeh:
I don't think we should be copying all headers:
https://tools.ietf.org/html/rfc2616#section-10.3.5

> If the conditional GET used a strong cache validator (see Weak and Strong Validators), the response SHOULD NOT include other entity-headers. Otherwise (i.e., the conditional GET used a weak validator), the response MUST NOT include other entity-headers; this prevents inconsistencies between cached entity-bodies and updated headers.

https://tools.ietf.org/html/rfc2616#section-7.1

croudet:
I have changed the copy function to exclude entity headers.

ElArtista:
Well done!

PR #3273, opened 2020-05-19
Generate resource-config.json

ilopmar:
Improves #3108.

I've also updated the documentation to include GraalVM 20.1.0 which will be released today.


inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
```diff
@@ -39,10 +40,11 @@
  * @author Graeme Rocher
  * @since 1.0
  */
+@SupportedOptions({VisitorContext.MICRONAUT_PROCESSING_PROJECT_DIR})
```
graemerocher, line 43: `@SupportedOptions` should be on `GraalTypeElementVisitor`

ilopmar: Done.

PR #3278, opened 2020-05-19
Upgrade to Netty 4.1.50

ilopmar:


graemerocher:
Something seems to have changed in Netty that has broken some tests https://github.com/micronaut-projects/micronaut-core/runs/692170504?check_suite_focus=true#step:6:597

ilopmar:
After opening an issue https://github.com/netty/netty/pull/10311 it seems that there was a regression https://github.com/netty/netty/pull/10285 which has been already fixed. I've tried with Netty `4.1.51.Final-SNAPSHOT` and those tests that were failing, now pass.

I'll send a new PR once next Netty version is released.

PR #3298, opened 2020-05-20
Instrumentation rework

### LINKED ISSUE #3106: MDC is leaking out to other threads

### Expected Behaviour

### Actual Behaviour

### Environment Information

- **Operating System**: Mac
- **JDK Version:** 11

### Example Application

lgathy:
Adds built-in support for common publisher instrumentation scenarios common in `HttpServerFilter` implementations. Closes #3106

This changeset consists of three main parts:

1. `Instrumenters` provides public static methods (one-liners) for instrumenting tasks. These constructs are currently inlined whereever instrumentation takes place. It might worth considering if/how these methods sould be moved to the `InvocationInstrumenter` interface or merged with its current non-public `instrument(...)` methods. However they are not compatible in their current form as those assume that instrumentation is always done automatically via the registered factories (and not programatically as here, see next point), thus they unwrap any of their instrumented Runnable/Callable instances regardless of whether the instrumenters are different on purpose.
2. `Conditional*` classes support efficient programmatic instrumentation by wrapping assembled reactive constructs. An example usage for injecting data into MDC in a server filter is demonstrated in `MDCSpec`.
3. Internal implementation of certain `RxInstrumented*` classes are reworked to remove code duplication using `ConditionalInstrumenter` and `RunOnceIncovationInstrumenter`.

Remark: my general feeling is that the visibility of the instrumentation related codebase is too restricted in some places. I understand the benefits of that, however I believe it might be reconsidered to open up visibility of that API to a certain degree to provide accessible, documented constructs for real-world use-cases to discourage manual and potentially faulty reimplementations in projects.


dstepanov:
Instrumentation was rewritten like this to eliminate call stack, each lambda wrap adds two calls, simple run/call just one and all of it multiply by the number of instrumenters.

I think MDC issue can be fixed with:

```
                @Override
                public void afterInvocation(boolean cleanup) {
                    if (cleanup || oldContextMap == null || oldContextMap.isEmpty()) {
                        MDC.clear();
                    } else {
                        MDC.setContextMap(oldContextMap);
                    }
                }
```

graemerocher:
@dstepanov I did try that but did not have any luck, did you have any?

Another option to reduce the call stack is to make the whole instrumentation part optional and only have modules that depend on it (tracing etc.) pull it in. I think that would be a good improvement for 2.0

lgathy:
@dstepanov Ok, keeping the call stack smaller is a fair point. I have an idea about it, let me quickly update my PR. Regarding your proposed fix for the MDC: I don't see how that could do the cleanup in cases where the `doFinally(...)` has been pushed to another thread. I see no option to force returning to the same thread where `MDC.put(...)` was invoked and if MDC was empty prior to our call - like in this case - there will be no MdcInstrumenter active in the call stack due to MdcInstrumenter:47.

graemerocher:
@lgathy I merged @dstepanov PR which seems to resolves the original report. If you are still interested in contributing improvements to the instrumentation APIs please do so

lgathy:
@graemerocher  I will take a look. My first impression is that I can still break that solution, as doFinally(...) does not always run on the same thread where put was invoked. I will add a test-case for that. Still working on my fix proposal, should I open a new ticket or reopen this one instead?

graemerocher:
Reopening this one is fine. Thanks.

lgathy:
@dstepanov I could identify any misbehaviour with your fix in #3313 , since it's way more elegant than my proposal here, I merged this PR with your most recent changes and removed the things which were introduced to fix the original issue in my way. Kudos to you. Do you think this could be merged back to `1.3.x`? It would be really helpful and as I see would not introduce any new API or braking changes.

In the meantime I had some ongoing work regarding your previous comments on avoiding lambdas to reduce the call stack which I still believe might worth sharing. So I reworked this PR accordingly. The base idea is to instrument via `Instrumentation` objects created in try-with-resources blocks. Example usage:

```java
try (Instrumentation ignored = instrumenter.newInstrumentation()) {
    ...
}
```

This would avoid adding extra elements to the call stack and it would also provide an easy reusable and straightforward way to invoke instrumentation properly without code duplication. Moreover I believe this enables creating smarter instrumenter logic to be implemented i.e. to optimize performance by avoiding executing the same instrumentation multiple times, which could be beneficiary especially on the long run. It seems to me that I did not break any contracts, all tests succeed.

@graemerocher What do you think of this change? Do you think it would worth merging or do you see any pitfalls I missed?

dstepanov:
@lgathy It should be possible to backport it from the master.
I like try-with-resources but I'm not sure if it's worse it to have an extra allocation.

graemerocher:
The try with resources approach is certainly much cleaner and in terms of maintainability and correctness seems a better approach. I wonder if the extra allocation will add any actual real overhead. 

In terms of backporting, the challenge is these APIs have been changed significantly in 2.0.x and whilst these changes were needed to fix critical problems with the previous instrumentation approach the APIs were not marked with `@Internal` in 1.x hence if we are following semantic versioning strictly changing them in 1.x would constitute a breaking API change.

lgathy:
Creating short-lived object should be extremely cheap with a modern JVM / GC.

Regarding backporting: I only ment the commints in the two PR's of @dstepanov (#3313 and #3314). It currently shows only minor conflicts in `ServerRequestContextFilter`. Anything I shared in this PR is intended for v2.0.

graemerocher:
Right, if you wish to send a backport PR for 1.3.x we can consider that. I will merge this. Thanks

graemerocher:
@lgathy I may have to rollback these changes since merging this the `MdcSpec` is failing randomly https://github.com/micronaut-projects/micronaut-core/runs/699190252?check_suite_focus=true#step:7:1462

lgathy:
@graemerocher I will take a look...

graemerocher:
I have reverted in this commit https://github.com/micronaut-projects/micronaut-core/commit/77f460c2b3bfeb05d9b51e7bd9d6c5a8cad0b83c

We can reapply the changes if the underlying issue can be identified.

PR #3304, opened 2020-05-21
ObjectSerializer should deserialize using Argument<T>

dstepanov:
To allow deserializing generic collections when using "@Cacheable".
I will add PR for Redis cache to call the new method.


graemerocher:
Thanks for the contribution

PR #3305, opened 2020-05-21
Cleanup Mdc, Fixes #3106

### LINKED ISSUE #3106: MDC is leaking out to other threads

dstepanov:


graemerocher:
The MDC tests are failing still

dstepanov:
Runnable and Callable is now by default set to clean after itself which is correct when used in Executor and ThreadFactory. I find it's strange that all Runnable and Callable is required to celan by default, I think it would be better to call to clean only in specific places like Executor, ThreadFactory and Netty filter.

graemerocher:
Also we would need to add this test to the code base https://github.com/tom59/micronaut-mdc-issue/blob/master/src/test/java/mn/mdc/instrumenter/MDCTest.java

I was not able to get this test to pass with a similar fix to yours

dstepanov:
I have a new PR for 2.0 that uses the test.

PR #3313, opened 2020-05-21
Clean MDC context, Fixes #3106

### LINKED ISSUE #3106: MDC is leaking out to other threads

dstepanov:
The problem is a bit more complicated, MdcInstrumenter needs to be used all the time and reset context if somebody modified it. 

If somebody is setting MDC outside of MdcInstrumenter begin/after it will stay bound to the thread.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/context/ServerRequestContextFilter.java
+++ b/http-server/src/main/java/io/micronaut/http/server/context/ServerRequestContextFilter.java
@@ -19,14 +19,16 @@ import io.micronaut.core.annotation.Internal;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.http.annotation.Filter;
-import io.micronaut.http.context.ServerRequestContext;
-import io.micronaut.http.context.ServerRequestTracingPublisher;
 import io.micronaut.http.filter.HttpServerFilter;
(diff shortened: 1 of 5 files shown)


dstepanov:
@graemerocher In your test, you were setting MDC in doOnSubscribe which is called after request has been processed.

graemerocher:
@dstepanov Gotcha. Thanks for the PR and looking into this!

dstepanov:
@lgathy @graemerocher BTW `doFinally` is not needed, but simple `chain.proceed(request)` or `Flowable.fromPublisher(chain.proceed(request))` will not propagate MDC, some of the examples that propagate context:

```
Flowable.fromPublisher(chain.proceed(request)).doOnSubscribe({})
```
```
return Flowable
                    .fromPublisher(chain.proceed(request))
                    .doFinally{->
                    }
```
```
return Single.fromPublisher(chain.proceed(request)).toFlowable()
```


```
return Flowable.defer({ -> chain.proceed(request) })
```

lgathy:
@dstepanov I have tried your solution previously and it did not work. Now I see why: you have also changed `MdcInstrumenter` now, I believe this was causing the problem during my previous attempts. Let me check if I can think of any circumstances where this new solution could go wrong.

PR #3314, opened 2020-05-21
Clean MDC context 2

dstepanov:
Allow propagating MDC values set in doFilter.


graemerocher:
Thanks for the contribution

PR #3322, opened 2020-05-22
Add tracing where MDCSpecs randomly fail

lgathy:
@graemerocher I could not reproduce [this failure](https://github.com/micronaut-projects/micronaut-core/runs/699190252?check_suite_focus=true#step:7:1462), but analysing the logs how it failed I added some extra tracing to better see what might have gone wrong when the trace_id is not found. I am only creating this PR as a draft hoping that build will still be triggered. Since it surfaces randomly we might need to trigger the build multiple times, is that possible?

### DIFF
--- a/tracing/src/test/groovy/io/micronaut/tracing/instrument/util/MDCSpec.groovy
+++ b/tracing/src/test/groovy/io/micronaut/tracing/instrument/util/MDCSpec.groovy
@@ -46,9 +46,15 @@ class MDCSpec extends Specification {
 
         @Get("/mdc-test")
         HttpResponse<String> getMdc() {
-            Map<String, String> mdc = MDC.getCopyOfContextMap()
-            println ("Thread = " + Thread.currentThread().getName())
-            return HttpResponse.ok(mdc.get(RequestIdFilter.TRACE_ID_MDC_KEY))
+            Map<String, String> mdc = MDC.getCopyOfContextMap() ?: [:]
+            Thread currentThread = Thread.currentThread()
+            println ("Thread = " + currentThread.getName())
+            String traceId = mdc.get(RequestIdFilter.TRACE_ID_MDC_KEY)
+            if (traceId == null) {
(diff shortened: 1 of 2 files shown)


lgathy:
Relates to reverted PR #3298

graemerocher:
I think you can only restart builds that fail. Only a change can trigger another build I think, so you may need to add a comment or something

graemerocher:
Actually no I can restart them.. I restarted again

graemerocher:
Actually it looks like it may be unrelated to the PR, since it just failed again with  https://github.com/micronaut-projects/micronaut-core/pull/3323/checks?check_run_id=699694930#step:7:1341

Looks like we need to investigate why that is happening

lgathy:
It may help then to merge this PR as it will provide further detail when this fails again as the call stack of the failing thread will be logged.

graemerocher:
Looks like the trace ID is missing when it fails https://github.com/micronaut-projects/micronaut-core/runs/700013955?check_suite_focus=true#step:7:1482

So potentially there is still some thread safety issues with the current solution. @dstepanov Any idea why the impl could sometimes fail under load?

lgathy:
It always seems to be the very first invocation failing, which could indicate spinning up a new thread has something to do with it. Stack traces are identical to when it does not fail, Logs show that `MDC.put` has been successfully executed. If MDC adapter is not different, it leaves only one possibilty: `MDC.remove()` or `MDC.clear()` gets executed between the filter and the controller, I will try to find a way to track that, it could be MdcInstrumenter, since that part changed recently. Btw, filter and controller seem to get executed on the same thread.

@graemerocher I will add more tracing and submit it in a new PR. I was still not able to reproduce this locally no matter what I tried, so I see no alternatives atm.

graemerocher:
So yeah currently in Micronaut 2.x we don’t do thread branching and leave that up to the user to manage with scheduleOn etc

PR #3326, opened 2020-05-22
MDCSpec: more detailed tracing

lgathy:
Follow-up to #3322 and #3298 to investigate random failure of MDCSpec: verbose logging added to find out how MDC gets cleared between the filter and the controller.


lgathy:
@graemerocher No luck in trying to reproduce in 2 build attempts. If you see no risk in merging this, please do so.

graemerocher:
Merged. thanks

PR #3328, opened 2020-05-23
Fix MDC context cleanup

dstepanov:
@graemerocher @lgathy Fixes master test fails.
Long story short tracing caused MDC context to me empty instead of null and caused MdcInstrumenter to always clean it.


graemerocher:
Thanks so much for looking into it!

dstepanov:
It would be nice if Gradle had deterministic tests order, that would have pointed out the problem right away.

graemerocher:
I think it may be because we run tests with —parallel to improve the execution speed of the CI

PR #3329, opened 2020-05-23
RequestBean supports binding also Typed Values now

asodja:
I extended functionality of RequestBean so it supports binding also typed values. 

Now you can bind any typed value like BasicAuth, Authorization or any custom type to single POJO (previously there was just support for HttpRequest and values with `@Bindable` annotation).


graemerocher:
Thanks for the contribution

PR #3335, opened 2020-05-25
Backported dstepanov's solution to fix issue #3106

lgathy:
Related PRs: #3313 #3314 and #3328

@graemerocher As discussed earlier, backporting this to 1.3.x seems straightforward.

@dstepanov In relation to the fix in #3328 I was wondering if we would need an `MDC.clear()` on the `else` branch in `MdcInstrumenter.beforeInvocation()` symmetric to `afterInvocation()`? I could see this to be a problem when the executing thread already has data in MDC but our instrumented invocation requires MDC to be empty.


graemerocher:
Thanks

dstepanov:
@lgathy MDC instrumentation doesn't require context to be empty, I think the correct way would be for `MdcInstrumenter` to extend the current context and then return it to the previous state.

PR #3361, opened 2020-05-27
Update distributedConfigurationConsul.adoc

dancegit:
seems to use application-prod and not application,prod
in consul distributed configuration


jameskleeh:
This is not correct

PR #3362, opened 2020-05-27
Handle kotlin coroutines

croudet:
- skip last method parameter if it is a Continuation
- Continuation holds the return type.


graemerocher:
Awesome, thanks!

dave08:
@croudet Is it possible that using a `suspend fun` in a Controller like this:

```kotlin
@Delete
suspend fun evictCache() = withContext(Dispatchers.IO) { if (clearCache()) HttpStatus.RESET_CONTENT else HttpStatus.EXPECTATION_FAILED }
```
is not possible using the current implementation? It seems like it returns a string "RESET_CONTENT" with a 200 status code instead of the expected 205 status code...

Without the `withContext` it works fine.

croudet:
Try to return an http response.

dave08:
What's the difference? Is it only by HttpStatus? It would be nice to know if this might happen with other return types... thanks @croudet !

PR #3367, opened 2020-05-28
Use module/group if define for generated GraalVM files path.

### LINKED ISSUE #3274: Use GraalVM recommended groupID and artifactID
We need to generate GraalVM configuration files: `native-image.properties`,... in the right directory.

See [GraalVM docs](https://www.graalvm.org/docs/reference-manual/native-image/#native-image-configuration).

Add two new annotation processor options
`-Amicronaut.processing.group='foo.bar'` and `-Amicronaut.processing.module='foo-bar'`
and if present use those values to generate the path. It would be `META-INF/native-image/$group/$module`.

In Gradle see this: https://github.com/micronaut-projects/micronaut-starter/pull/230/files as an example on how to specify those.
In Maven it may be possible with the parent POM.

ilopmar:
Fixes #3274

After this is merged, merge https://github.com/micronaut-projects/micronaut-starter/pull/256

### DIFF
--- a/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
+++ b/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
@@ -46,7 +46,11 @@ import java.util.stream.Collectors;
  * @author Iván López
  * @since 1.1
  */
-@SupportedOptions({VisitorContext.MICRONAUT_PROCESSING_PROJECT_DIR})
+@SupportedOptions({
+        VisitorContext.MICRONAUT_PROCESSING_PROJECT_DIR,
+        VisitorContext.MICRONAUT_PROCESSING_GROUP,
+        VisitorContext.MICRONAUT_PROCESSING_MODULE
+})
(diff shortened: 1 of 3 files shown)


graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -46,7 +46,11 @@
  * @author Iván López
  * @since 1.1
  */
-@SupportedOptions({VisitorContext.MICRONAUT_PROCESSING_PROJECT_DIR})
+@SupportedOptions({
+        VisitorContext.MICRONAUT_PROCESSING_PROJECT_DIR,
+        VisitorContext.MICRONAUT_PROCESSING_GROUP,
+        VisitorContext.MICRONAUT_PROCESSING_MODULE
+})
```
graemerocher, line 53: Could you move these new declarations to `TypeElementVisitorProcessor` as I believe another processors (like OpenAPI) could potentially using these flags. Thanks.

ilopmar: Done

Issue #3334, opened 2020-05-25
IllegalStateException: No lock present for object


theHacker:
### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Open the demo project
2. Run the `MyServiceTest` inside your IDE to see the full Exception.

### Expected Behaviour

Test finishes ordinarily.

### Actual Behaviour

Micronaut will throw an exception
```
No lock present for object: MyClient(#1)
java.lang.IllegalStateException: No lock present for object: MyClient(#1)
	at io.micronaut.runtime.context.scope.refresh.RefreshScope.getLock(RefreshScope.java:158)
	at io.micronaut.runtime.context.scope.refresh.RefreshInterceptor.intercept(RefreshInterceptor.java:46)
	at io.micronaut.aop.chain.MethodInterceptorChain.proceed(MethodInterceptorChain.java:69)
	at micronaut.issue.$MyServiceTest$MyClientMock0Definition$Intercepted.getSomeFoos(Unknown Source)
	at micronaut.issue.MyService.getFoos(MyService.kt:9)
	at micronaut.issue.MyServiceTest$1.invokeSuspend(MyServiceTest.kt:24)
	at micronaut.issue.MyServiceTest$1.invoke(MyServiceTest.kt)
	at io.kotlintest.runner.jvm.TestCaseExecutor$executeTest$supervisorJob$1$invokeSuspend$$inlined$map$lambda$1.invokeSuspend(TestCaseExecutor.kt:121)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:233)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Environment Information

- **Operating System**: MacOS
- **Micronaut Version:** 1.3.0.RC1 - Project is created with `nm create-app --lang=kotlin`
- **JDK Version:** Kotlin 1.3.50

### Example Application

[...]

PR #3372, opened 2020-05-29
Correct refresh scope destroy bean keeping old instances

dstepanov:


graemerocher:
Does this fix #3334?

dstepanov:
I don't think so, this one just removes extra locks left.

graemerocher:
Thanks

Issue #3254, opened 2020-05-16
Determine path forward for complex conditions and null return in factory methods


jameskleeh:
1. Update bean retrieval in conditions to work properly (sometimes causes exceptions in JDK11)
2. Fix factory bean instantiation to swallow a more specific exception related to null return values, or remove the feature

graemerocher:
IMO we should completely remove the null return stuff it was a massive mistake adding it

graemerocher:
So currently this behaviour seems fundamentally broken to me. A class such as:

```
@EachBean(MyConfiguration.class)
public class MyBean {
    final MyConfiguration configuration;

    MyBean(MyConfiguration configuration) {
        this.configuration = configuration;
    }

    ...
}
```

Productes byte code that nulls out the whole bean when an exception occurs and hides the originating exception:

```java
    public MyBean build(BeanResolutionContext var1, BeanContext var2, BeanDefinition<MyBean> var3) {
        MyConfiguration var4 = null;

        try {
            var4 = (MyConfiguration)super.getBeanForConstructorArgument(var1, var2, 0);
        } catch (BeanContextException var6) {
            return null;
        }

        MyBean var5 = new MyBean(var4);
        var5 = (MyBean)this.injectBean(var1, var2, var5);
        return var5;
    }
```

Even though the parameter is not nullable. 

I have investigated various approaches such as introducing `DisabledBeanException` that can be thrown but all of them have issues because there is no way for a `BeanDefinition` itself to disabled via an exception.

I think we need to think about what motivates the desire for this behaviour and why it cannot be handled via regular conditions because currently there are all sorts of problems with the current approach.

PR #3387, opened 2020-06-02
Introduce DisabledBeanException, pass resolution context within conditions

graemerocher:
This change removes the ability to return null from factory methods and adds a DisabledBeanException that users can throw. It has a disavantage in that the only way to know the exception is going to be thrown is to invoke the fatory method hence the counts changed in the test.

Also changes ConditionContext to implement BeanLocator and PropertyResolver and pass the resolution context along to try to alleviate circular problems


graemerocher:
Candidate fix for #3254

PR #3414, opened 2020-06-04
EndpointDefaultConfiguration port should be nullable

lightoze:
This makes it possible to have "endpoints.all.port: null" in property source.


lightoze:
@graemerocher I fixed the annotation.

PR #3434, opened 2020-06-08
Utility class for implementing AutomaticFeatures

alvarosanchez:


graemerocher:
I'm not sure the `graal` subproject is the right place as that belongs only the annotation processor path. It may need to go into `core` or another module

ilopmar:
I think we could move it to `core` as all modules depends on it and we avoid to add a new module with just only one class.

alvarosanchez:
Done.

PR #3437, opened 2020-06-09
Prevent parallel invocation instrumentation 2

dstepanov:
@graemerocher @lgathy I have removed the mutable state from `MultipleInvocationInstrumenter` it was only there to prevent `after` to be called in a case when an exception is thrown during `begin` which we can just ignore.

It would be better to refactor again to something like:
- capture context = InvocationInstrumenter
- try catch = set/restore context

### DIFF
--- a/runtime/src/main/java/io/micronaut/scheduling/instrument/MultipleInvocationInstrumenter.java
+++ b/runtime/src/main/java/io/micronaut/scheduling/instrument/MultipleInvocationInstrumenter.java
@@ -19,9 +19,10 @@ import io.micronaut.core.annotation.Internal;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayDeque;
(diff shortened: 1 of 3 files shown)


graemerocher:
Thanks

lgathy:
This will make the exception go away, but most of the instrumenter implementations are not thread-safe atm and they will still be invoked concurrently. I will try to fabricate tests to prove my point later.

graemerocher:
@lgathy hence why the original change to add synchronization seems safer although my introduce thread contention, although not every interaction with RxJava involves multi threading especially if the user explicitly controls back pressure.

dstepanov:
@lgathy You are right, if we really want to fix it I think the right solution would be to have `synchronized subscribeActual` that's IMHO the only case when multiple calls can occur and this method should be fast.

I will add a PR with a refactoring to fix it in master but not sure what to do with 1.3.

PR #3438, opened 2020-06-09
Reduce RxJava2 wrappers

dstepanov:
This will change instrumentation to skip if previous one is already wrapped:
```
.just(value) -> wrap
.map ... -> skip
.filter... -> skip
```

```
externalSource 
.map ... -> wrap
.filter... -> skip
```


graemerocher:
Thanks

PR #3463, opened 2020-06-10
Support storing property files in config folder

dstepanov:
Not sure if it needs to be configured or it's fine to have it like this.

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/file/DefaultFileSystemResourceLoader.java
+++ b/core/src/main/java/io/micronaut/core/io/file/DefaultFileSystemResourceLoader.java
@@ -23,6 +23,8 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Files;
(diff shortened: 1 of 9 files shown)


graemerocher:
@dstepanov I'm not sure this should be done by default as it adds overhead. We already have `-Dmicronaut.config.files` which allow applies if specified either by an environment variable or a system property.

If we were to consider something like this (maybe `-Dmicronaut.config.dirs` it would need to be disabled by default)

dstepanov:
Maybe if I add a check if dir exists, that should't be a big overhead?

graemerocher:
It is disk I/O on startup which we are trying to minimize as much as possible for serverless. In general the PR is maybe a bit opinionated about the directory location, name etc. 

Another option instead of system properties is to modify `ApplicationContextBuilder` to allow this behaviour to be customized

dstepanov:
Added option to override default config location in a similar way as the boot.
Also, boot by default has `classpath:/config/` it can be easily added.

PR #3505, opened 2020-06-16
Adding micronaut-acme into bom at 1.0.0.RC3

zendern:
Give this one more go now that things are good to go when using the new build tools in micronaut-acme.


graemerocher:
@zendern just one thing, if we merge this `1.0.0` must be released before the 26th which is the GA date for Micronaut 2.0

zendern:
> @zendern just one thing, if we merge this `1.0.0` must be released before the 26th which is the GA date for Micronaut 2.0

I have no issues with that deadline. I plan on doing more testing on the latest release client and will cut a 1.0.0 probably this weekend.

graemerocher:
Ok merged into the BOM for now

PR #3506, opened 2020-06-16
Remove thread factory tracing

dstepanov:
Unfortunately, thread factory tracing is causing all kind of issues, the problem occurs when the thread is actually a long-running reusable.
The first invocation will set server context, MDC, and tracing of the first request, and in RC1 it's causing that first server request object being set again.


graemerocher:
Well this is what we have RCs for. It would be could if this was a test included that demonstrated the issue

dstepanov:
I didn't manage to reproduce it by a test in micronaut, in a private project I managed to point it out to only netty threads wrapping, I suspected client and server are sharing the same thread pool and when client is called from a controller context is propagated to a new runnable wrapper.

PR #3519, opened 2020-06-18
Handle class arrays in annotation metadata

### LINKED ISSUE #3481: No introspection data created/returned for array classes
With the following:

```java
@TypeHint(
    value = {UUID[].class, UUID.class},
    accessType = AccessType.ALL_PUBLIC_CONSTRUCTORS
)
```
If we try to get the `value`, for example in the `GraalTypeElementVisitor`: https://github.com/micronaut-projects/micronaut-core/blob/1.3.x/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java#L103

The `UUID[]` is not returned, only `UUID`.

### Expected Behaviour

When retrieving the element information doing `final String[] introspectedClasses = element.stringValues(TypeHint.class);` it should return both `UUID[]` and `UUID`

### Actual Behaviour

It only returns `UUID`.

### Workaround

Declare the classes as String using `typeNames`

### Environment Information

- **Operating System**: Linux Mint
- **Micronaut Version:** 1.3.6 and 2.0.0
- **JDK Version:** 1.8

jameskleeh:
Fixes #3481

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/annotation/AnnotationMetadataWriterSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/annotation/AnnotationMetadataWriterSpec.groovy
@@ -21,6 +21,7 @@ import io.micronaut.context.annotation.Requirements
 import io.micronaut.context.annotation.Requires
 import io.micronaut.core.annotation.AnnotationClassValue
 import io.micronaut.core.annotation.AnnotationMetadata
+import io.micronaut.core.annotation.TypeHint
 
 import javax.inject.Named
 import javax.inject.Qualifier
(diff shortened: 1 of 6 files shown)


inject-groovy/src/test/groovy/io/micronaut/inject/annotation/AnnotationMetadataWriterSpec.groovy
```diff
@@ -338,4 +339,32 @@ interface TestApi {
...
+
+@TypeHint([UUID[].class, UUID.class])
+class Test {
+
+}
+""")
+
+        expect:
+        metadata.stringValues(TypeHint).size() == 2
+        metadata.stringValues(TypeHint)[0] == '[Ljava.util.UUID;'
```
graemerocher, line 358: What happens when you do `metadata.classValues(TypeHint)[0]`? Need a test here

Issue #3537, opened 2020-06-22
ConcurrentModificationException from NettyMutableHttpResponse.getMessage()


KiranSathi:
We are getting ConcurrentModificationException from NettyMutableHttpResponse while calling getMessage() method. Internally we are using LinkedHashMap and is not thread safe. Is this a known issue?

Following is the stack trace.

  "file": "HashMap.java",
  "method": "computeIfAbsent",
  "class": "java.util.HashMap",
  "line": 1134,
  "exact": false,
  "location": "?",
  "version": "?"
},
{
  "file": "NettyMutableHttpResponse.java",
  "method": "getBody",
  "class": "io.micronaut.http.netty.NettyMutableHttpResponse",
  "line": 146,
  "exact": false,
  "location": "product-store-position-api-1588876073.0.623-z45ad911.jar",
  "version": "?"
},
{
  "file": "HttpClientResponseException.java",
  "method": "getMessage",
  "class": "io.micronaut.http.client.exceptions.HttpClientResponseException",
  "line": 70,
  "exact": false,
  "location": "product-store-position-api-1588876073.0.623-z45ad911.jar",
  "version": "?"
}

Can we make convertedBodies Map in NettyMutableHttpResponse to concurrent Map?

graemerocher:
Seems reasonable, want to send a PR?

KiranSathi:
Sure. Will do.

KiranSathi:
@graemerocher Raised PR https://github.com/micronaut-projects/micronaut-core/pull/3549. Please review.

graemerocher:
Thanks for the PR

PR #3549, opened 2020-06-23
Avoid sync issue in getBody

KiranSathi:
Avoid ConcurrentModificationException in NettyMutableHttpResponse.getBody by using ConcurrentLinkedHashMap instead of lInkedHashMap.

Issue Details: https://github.com/micronaut-projects/micronaut-core/issues/3537


graemerocher:
The PR is causing the build to hang, probably with a deadlock.

KiranSathi:
@graemerocher Updated with synchronizedMap. Please review.

PR #3584, opened 2020-06-28
Add support for Http2 settings.

croudet:


graemerocher:
Thanks for the contribution!

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -571,13 +733,6 @@ public void setPreferNativeTransport(boolean preferNativeTransport) {
             this.preferNativeTransport = preferNativeTransport;
         }
 
-        /**
-         * @return The number of threads to use
-         */
-        public int getNumOfThreads() {
-            return threads;
-        }
-
```
jameskleeh, line 580: Isn't this a breaking change? @graemerocher

graemerocher: @jameskleeh didn't spot that, could you restore the method?

PR #3586, opened 2020-06-29
Support `Optional` in interface config

David-16:
Previously, returning an `Optional` type in immutable / interface config would cause failures if the property did not exist

This PR amends the behavior to return `Optional.empty()` in the event the property does not exist similar to how `@Nullable` will cause it to return `null`


graemerocher:
Thanks for the contribution!

PR #3594, opened 2020-06-29
Added another example for maven group id renaming

romeoro:
In case for maven group id renaming for micronaut jdbc dependencies, the group was renamed from configuration to sql.


jameskleeh:
Thanks!

PR #3595, opened 2020-06-29
Clean-up http code.

croudet:


http-netty/src/main/java/io/micronaut/http/netty/reactive/HandlerPublisher.java
```diff
@@ -259,13 +259,13 @@ public void channelRegistered(ChannelHandlerContext ctx) {
     private void provideChannelContext(ChannelHandlerContext ctx) {
         switch (state) {
             case NO_SUBSCRIBER_OR_CONTEXT:
-                verifyRegisteredWithRightExecutor(ctx);
+                verifyRegisteredWithRightExecutor();
```
jameskleeh, line 262: I don't think this change is correct. The instance variable isn't set until after the method is called

croudet: `ctx` was not used in `verifyRegisteredWithRightExecutor`

jameskleeh: Ah I see. Github collapsed some code and it looked like it was. Thanks

jameskleeh:
Thanks!

PR #3597, opened 2020-06-29
Clean-up core code

croudet:


core/src/main/java/io/micronaut/core/io/service/DefaultServiceDefinition.java
```diff
@@ -56,7 +56,7 @@ public boolean isPresent() {
     public <X extends Throwable> S orElseThrow(Supplier<? extends X> exceptionSupplier) throws X {
         final Class<S> type = loadedClass.orElseThrow(exceptionSupplier);
         try {
-            return type.newInstance();
+            return type.getDeclaredConstructor().newInstance();
```
jameskleeh, line 59: What is the reasoning behind this change?

croudet: `newInstance` is deprecated in new version of jdk:
https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--

PR #3599, opened 2020-06-29
Clean-up inject code.

croudet:


inject-java/src/main/java/io/micronaut/annotation/processing/visitor/AbstractJavaElement.java
```diff
@@ -211,30 +211,23 @@ public String toString() {
                 return mirrorToClassElement(bound, visitorContext, genericsInfo);
             } else {
 
-                ClassElement classElement = mirrorToClassElement(upperBound, visitorContext, genericsInfo);
-                if (classElement != null) {
-                    return classElement;
-                } else {
-                    return mirrorToClassElement(tv.getLowerBound(), visitorContext, genericsInfo);
-                }
+                return mirrorToClassElement(upperBound, visitorContext, genericsInfo);
```
jameskleeh, line 214: This doesn't seem correct

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
```diff
@@ -211,37 +211,31 @@ protected void accept(DeclaredType declaringType, javax.lang.model.element.Eleme
                         getterReturnType = mirrorToClassElement(returnType, visitorContext, JavaClassElement.this.genericTypeInfo);
                     }
 
-                    if (getterReturnType != null) {
```
jameskleeh, line 214: Removing this check doesn't seem correct. The previous calls `mirrorToClassElement` could have returned null

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
```diff
@@ -211,37 +211,31 @@ protected void accept(DeclaredType declaringType, javax.lang.model.element.Eleme
...
+                        if (!setterParameterType.getName().equals(getterReturnType.getName())) {
+                            beanPropertyData.setter = null; // not a compatible setter
                         }
                     }
                 } else if (NameUtils.isSetterName(methodName) && executableElement.getParameters().size() == 1) {
                     String propertyName = NameUtils.getPropertyNameForSetter(methodName);
                     TypeMirror typeMirror = executableElement.getParameters().get(0).asType();
                     ClassElement setterParameterType = mirrorToClassElement(typeMirror, visitorContext, JavaClassElement.this.genericTypeInfo);
 
-                    if (setterParameterType != null) {
```
jameskleeh, line 233: Removing this check doesn't seem correct. The previous call to mirrorToClassElement could have returned null

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
```diff
@@ -415,9 +404,7 @@ public AnnotationMetadata getAnnotationMetadata() {
         while (tpi.hasNext()) {
             TypeParameterElement tpe = tpi.next();
             ClassElement classElement = mirrorToClassElement(tpe.asType(), visitorContext, this.genericTypeInfo);
-            if (classElement != null) {
-                map.put(tpe.toString(), classElement);
-            }
+            map.put(tpe.toString(), classElement);
```
jameskleeh, line 407: This also doesn't make sense to me. If the call returns null it shouldn't be put into the map with a null value

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
```diff
@@ -128,7 +128,7 @@ protected AbstractBeanDefinition(Class<T> producedType,
                                      boolean requiresReflection,
                                      Argument... arguments) {
 
-        AnnotationMetadata beanAnnotationMetadata = getAnnotationMetadata();
+        getAnnotationMetadata();
```
jameskleeh, line 131: What is the reasoning for this change?

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/AbstractJavaElement.java
croudet: `mirrorToClassElement` never returns null.

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
croudet: `beanAnnotationMetadata ` is not used.
Is the call mandatory?

jameskleeh: The only thing it will do is initialize the annotation metadata, which should not need done at that point in time. You can try removing the call and see if the build passes

croudet: Call removed.

PR #3602, opened 2020-06-29
Clean-up mangement, router and runtime code.

croudet:


router/src/main/java/io/micronaut/web/router/AnnotatedFilterRouteBuilder.java
```diff
@@ -41,27 +41,23 @@
...
      *
-     * @param beanContext The bean context
      * @param executionHandleLocator The execution handler locator
      * @param uriNamingStrategy The URI naming strategy
      * @param conversionService The conversion service
      * @param contextPathProvider The server context path provider
      */
     @Inject
     public AnnotatedFilterRouteBuilder(
-            BeanContext beanContext,
```
jameskleeh, line 58: This is a breaking change and thus must be reverted

runtime/src/main/java/io/micronaut/health/HeartbeatTask.java
```diff
@@ -37,17 +37,14 @@
...
-    private final HeartbeatConfiguration configuration;
     private final CurrentHealthStatus currentHealthStatus;
 
     /**
      * @param eventPublisher      To publish the events related to heartbeat
-     * @param configuration       The configurations for heartbeat
      * @param currentHealthStatus The current status of health indicator
      */
-    public HeartbeatTask(ApplicationEventPublisher eventPublisher, HeartbeatConfiguration configuration, CurrentHealthStatus currentHealthStatus) {
+    public HeartbeatTask(ApplicationEventPublisher eventPublisher, CurrentHealthStatus currentHealthStatus) {
```
jameskleeh, line 46: This is a breaking change and must be reverted

router/src/main/java/io/micronaut/web/router/AnnotatedFilterRouteBuilder.java
croudet: Argument restored.

runtime/src/main/java/io/micronaut/health/HeartbeatTask.java
croudet: Argument restored.

PR #3603, opened 2020-06-29
Fix query params in websocket url

hassomehide:
DefaultHttpClient copies an URI object, replacing its protocol with ws:// in the connectWebSocket method. However, query params are left missing. This commit copies query params as well, if present.

An example of a websocket url with query params is the docker attach api: https://docs.docker.com/engine/api/v1.40/#operation/ContainerAttachWebsocket


jameskleeh:
@hassomehide Are you interested in taking this PR forward?

jameskleeh:
Closed in favor of https://github.com/micronaut-projects/micronaut-core/pull/3802

PR #3604, opened 2020-06-30
Add missing Override annotations and

croudet:
other code clean-up.

### DIFF
--- a/tracing/src/main/java/io/micronaut/tracing/brave/BraveTracerFactory.java
+++ b/tracing/src/main/java/io/micronaut/tracing/brave/BraveTracerFactory.java
@@ -93,9 +93,7 @@ public class BraveTracerFactory {
     @Primary
     Tracer braveTracer(Tracing tracing) {
         BraveTracer braveTracer = BraveTracer.create(tracing);
-        if (!GlobalTracer.isRegistered()) {
-            GlobalTracer.register(braveTracer);
-        }
+        GlobalTracer.registerIfAbsent(braveTracer);
         return braveTracer;
     }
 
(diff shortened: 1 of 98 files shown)


tracing/src/main/java/io/micronaut/tracing/brave/BraveTracerFactory.java
```diff
@@ -94,7 +94,7 @@ SpanCustomizer spanCustomizer(Tracing tracing) {
     Tracer braveTracer(Tracing tracing) {
         BraveTracer braveTracer = BraveTracer.create(tracing);
         if (!GlobalTracer.isRegistered()) {
-            GlobalTracer.register(braveTracer);
+            GlobalTracer.registerIfAbsent(braveTracer);
```
jameskleeh, line 97: Shouldn't this be replacing the if statement?

validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
```diff
@@ -1056,7 +1056,7 @@ public void keyedValue(String nodeName, Object key, Object keyedValue) {
     @SuppressWarnings("unchecked")
     private <T> void validatePojoInternal(@NonNull Class<T> rootClass,
                                           @Nullable T object,
-                                          @NonNull Object[] argumentValues,
+                                          @Nullable Object[] argumentValues,
```
jameskleeh, line 1059: What is the reasoning behind this change?

croudet: https://github.com/micronaut-projects/micronaut-core/blob/8fdb9010e1753af34aece11a9dc1c58539f3cf3e/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java#L1122

Calls this method with null for that argument.

tracing/src/main/java/io/micronaut/tracing/brave/BraveTracerFactory.java
croudet: Done.

PR #3612, opened 2020-07-01
Improve Async and Provider performance

dstepanov:
- Reduced bean lookup in `AsyncInterceptor`
- Removed extra step for `Flowable` it should be enough to use `subscribeOn`
- Provider impl. can use `BeanDefinition` to resolve the instance and cache it if it's a singleton


graemerocher:
Thanks

Issue #1859, opened 2019-07-03
Don't generate beans if @Inject is present but the type is not declared a bean


graemerocher:
Currently we generate a bean definition if there is a method or field annotated with `@Inject` regardless if the type is declared a bean. This IMO is not good behaviour as the scope of the bean is undefined.

It is also inconsistent as there is no way to achieve the same with constructor injection and constructor injection really is preferred approach so this encourages bad practise.

### Steps to Reproduce

Define a class:

```java
package test;

import javax.annotation.PostConstruct;

class Test {

    @Inject
    void setFoo(Foo foo) {
    

    }
}
```

### Expected Behaviour

A bean should not be generated without an undefined scope (either `@Prototype`, `@Singleton` or `@Context` as it can lead to unpredictable behaviour

### Actual Behaviour

A bean is generated.

jameskleeh:
What about factory methods that are just annotated with `@Bean`? There is no explicit scope there either. Perhaps `@Bean` should be annotated with `@DefaultScope(Prototype.class)` ?

benmccann:
It's a little unclear to me what is being proposed here, so I just want to check that the following would still be valid:

```
@Controller("/hello") 
public class HelloController {

  @Inject FooBar foo;
  
  @Get 
  public String index() {
    return "Hello World " + foo; 
  }

}

public class FooBar {

  @Inject FooBar() {}

  public String toString() {
    return "foo world";
  }
  
}
```

graemerocher:
No that would not be valid. You would need to declare a bean scope to make it a bean such as Singleton

benmccann:
I would be very upset if this change were made and it probably would stop me from using Micronaut. It saves a lot of unnecessary annotations, visual clutter, and boilerplate code to default no scope to prototype scope. It also would be inconsistent with other DI frameworks such as Guice and Dagger leading to higher learning curve and higher migration costs (e.g. I have projects I'm considering migrating from Play to Micronaut). If it were implemented, I hope it would at least be with an option that could be disabled. (Or alternatively that DI systems were made pluggable so that I could use something like Dagger instead though that seems less feasible)

> It is also inconsistent as there is no way to achieve the same with constructor injection and constructor injection really is preferred approach so this encourages bad practise.

I was confused by this statement. You can write constructors with just an `@Inject` annotation and no declared scope today (i.e. defaulting to prototype scope).

[...]

PR #3617, opened 2020-07-01
Ensure bean definitions are only created for defined beans. Fixes #3561

### LINKED ISSUE #3561: Attempt to instantiate abstract class when it implements event listener and has an AOP advice
In `2.0.0.RC2`.

Starting an application with the following classes:

```java
@Validated
public abstract class AbstractListener implements BeanCreatedEventListener<AutoStartingBean> {

	public static final Logger logger = LoggerFactory.getLogger(AbstractListener.class);

	@Override
	public AutoStartingBean onCreated(BeanCreatedEvent<AutoStartingBean> event) {
		logger.debug("Listener called with: {}", event.getBean());

		return event.getBean();
	}
}

@Singleton
public class Listener extends AbstractListener {
}
```

Ends up with `Caused by: java.lang.InstantiationError: micronaut.general.AbstractListener`. This is because generated `$AbstractListenerDefinition` tries to instantiate abstract class:

```java
class $AbstractListenerDefinition extends AbstractBeanDefinition<AbstractListener> implements BeanFactory<AbstractListener> {

    // ...

    public AbstractListener build(BeanResolutionContext var1, BeanContext var2, BeanDefinition<AbstractListener> var3) {
        AbstractListener var4 = new AbstractListener();
        var4 = (AbstractListener)this.injectBean(var1, var2, var4);
        return var4;
    }

}
```

When I remove `@Validated` or remove `BeanCreatedEventListener` from `AbstractListener`, it works ok.

jameskleeh:
Note that this is technically a breaking change, however this is more of an extension of the work already done in 2.0.0 to ensure that bean definitions aren't created for things not explicitly defined as a bean. I think the potential impact is low


graemerocher:
I have concerns about the breaking nature of this change. Also the correct fix for the issue is to check if the bean is `isAbstract()` and filter it out from being used as a event listener

jameskleeh:
@graemerocher I think only applying this change to abstract classes leaves out many other undesired possibilities. For example:

Non declared beans can't have executable methods
The validated annotation will currently not be applied to non declared beans with constraints
Post construct will not work in non declared beans
Pre destroy will not work in non declared beans

I think explicitly excluding abstract classes is more of a bandaid around the larger issue. Groovy currently behaves today as Java would with these changes. I simply added the test case to the inject-groovy module and it passed without any further changes. 

The changes in our code base were classes that had either @Named or @Primary or both, but no other bean annotations. I think there is a possible impact, but I think its a minor one and is worth the fix compared to leaving the other possible errors out there until 3.0.0

graemerocher:
Ok if it is Groovy specific that changes matters. Java and Groovy should be consistent

jameskleeh:
The change is for the Java/Kotlin. I just pointed out that Groovy already works this way (requires beans to be declared)

graemerocher:
Hmm then I am unsure frankly, and the bug seems to be in Groovy. In general we should be aiming for less ceremony and having to declare a scope when just a `@Named(..)` would suffice seems a regression and a breaking change that we shouldn't contemplate. I'm at this stage -1 on this change in patch release.

You could argue that using `@Validated` (ie an AOP advice and which is the original reported issue) alone shouldn't result in a bean on an abstract class but this change seems to go beyond that.

ilopmar:
I think it makes sense to fix the current behavior if we consider it a bug in the next 2.0.1 release. Keeping in mind that Micronaut 2.0 was released less than 2 weeks ago, this won't probably affect any (or not many users). In any case, if some folks are affected, the solution for them is pretty simple, annotate the classes they want to make a bean with singleton/prototype as they need.

graemerocher:
@ilopmar whilst I agree the change here in the PR is more invasive than that and alters behaviour from 1.x. Why do you need to declare a scope for example if you have already declared a qualifier?

jameskleeh:
@graemerocher It's beyond just a qualifier though. You can apply AOP to a class and it will be a bean as well.

graemerocher:
That is behaviour that is questionable I agree. Since neither `@Validated` nor `@Around` are qualifiers or scopes these shouldn't trigger a bean I agree, but this PR as written goes way beyond that and disables bean generated for `@Named(..)` which is large breaking change IMO.

jameskleeh:
Just to give my opinion on the qualifier though - I think the issue is consistency. If we allowed beans to be created just by the presence of a qualifier then beans without qualifiers need a scope and beans with qualifiers don't. That isn't consistent and isn't documented. I think users should be explicit about whether the bean is a prototype or a singleton and not rely on framework defaults.

graemerocher:
I personally disagree, it adds way more ceremony to the framework than is needed. There are other areas where we have too much ceremony for example it shouldn't be necessary to use `@Inject` on fields when there is already a qualifier. We should be removing ceremony where possible and not adding it.

Regardless of my opinion however, the ship has already sailed with regards to 2.0.0 and if we are following semantic versioning this shouldn't be changed in a patch release.

jameskleeh:
We have always weighed the potential impact of a breaking change with the benefit and sometimes made changes in patch/minor releases, so I don't see a reason why that wouldn't happen here as well. 

Regarding the ceremony, in general I agree. Not requiring `@Inject` on where a qualifier is present makes sense because for example we don't require it on constructors. In addition it's not adding any value because there is no decision to make. 

That isn't the case with classes though. A class with `@Named` and no other annotations will result in a prototype bean, which is most likely not the desired outcome. I think most users would expect it to result in a singleton bean. I think if you would like to remove the ceremony for classes, we should change the default scope to singleton.

jameskleeh:
In one case even in our code we were using Groovy's `@Singleton` with `@Named` and it was resulting in a prototype bean https://github.com/micronaut-projects/micronaut-core/pull/3617/files#diff-13e9911334254d8a91eae0286448feb2R18

graemerocher:
I agree that is undesirable that that happened, however that code was in a test. 

Related to this we had #1859 open for a while and assigned to the 2.0 milestone and in the end didn't implement due to feedback from the community regarding consistency with other DI frameworks, see https://github.com/micronaut-projects/micronaut-core/issues/1859#issuecomment-620683454

If we now implement this in a patch release that would be a bad breaking change. If we want to take this forward I suggest altering the PR to change the behaviour of `isDeclaredBean` and adding `concreteClassMetadata.hasStereotype(Qualifier.class)` to it

PR #3632, opened 2020-07-03
Remove use of computeIfAbsent.

### LINKED ISSUE #3577: Recursive update during beans context initalization gets back
In `2.0.0.RC2`.

The problem is widely described in #2140 and #2099. It was fixed in `1.2.3` but it seems it's now present again in `2.0.0.RC2`.

This is **not** a problem of having circular dependencies between beans, because it is not 100% reproducible, maybe 10% tops. This is some race condition during beans context initialization, which was previously fixed successfully for some time.

My example exception is [contained in this comment](https://github.com/micronaut-projects/micronaut-core/issues/2140#issuecomment-650163797).

graemerocher:
This change removes the use of computeIfAbsent which for some cases leads to recursive update.
There may be a slight cost in terms of duplicate initialization of BeanDefinition caches dependency on the injection path,
but I don't think it is worth the cost of tring to retain atomicity of computeIfAbsent since these are just the caches
for the backing bean definitions and not the bean instances themselves. We may even want to attempt a periodic purge
of these caches to save memory, but that is a separate issue.

I have not been able to come up with a test that replicates the issue, but have verified it fixes in the problem
by testing in the Micronaut Security code base. Once Security upgrades to 2.0.1 the intermittent failure should be fixed there.

Fixes #3577

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -117,7 +117,8 @@ public class DefaultBeanContext implements BeanContext {
     private final Map<CharSequence, Object> attributes = Collections.synchronizedMap(new HashMap<>(5));
 
     private final Map<BeanKey, Collection<Object>> initializedObjectsByType = new ConcurrentHashMap<>(50);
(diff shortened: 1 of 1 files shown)


ilopmar:
@graemerocher do you think this will also solve this issue? https://github.com/micronaut-projects/micronaut-flyway/issues/11
The symptoms are pretty much the same and sometimes I see failures in Flyway GraalVM app. There are random and not to common but once this is merged I'll keep an eye on the GraalVM CI to see if the errors are gone.

graemerocher:
Yes I believe it will solve that issue

PR #3633, opened 2020-07-03
Define Jackson Java 8 libs as API dependencies

dstepanov:
Before v2 those dependencies were accessible in the user code.

### DIFF
--- a/runtime/build.gradle
+++ b/runtime/build.gradle
@@ -10,8 +10,8 @@ dependencies {
     api dependencyVersion("validation")
     api "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
     api dependencyVersion("rxjava2")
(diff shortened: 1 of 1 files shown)


graemerocher:
Is there a reason they need to be?

dstepanov:
We have some code is using Jackson Jdk8 modules directly, if there are already included why not allow to use them?

graemerocher:
In general it is better to reduce the API surface area of the framework as much as possible hence the change, it seemed a safe one as I had no idea folks used these modules directly. It maybe needs to be debated whether it makes sense for these to be api dependencies

dstepanov:
IMHO those modules should be part of the main Jackson library and the dependency rules for them should be the same. It's not a big deal to include them but it would be better to have consistent classpath rules.

PR #3646, opened 2020-07-05
Match url with trailing slash

dwlizlo:
Simply cutting off trailing slash to make it compatible with micronaut.


test-suite-groovy/src/test/groovy/io/micronaut/docs/server/uris/UriTemplateTest.groovy
```diff
@@ -15,4 +15,14 @@ class UriTemplateTest extends Specification {
         template.expand(["name": "John"]) == "/hello/John" // <2>
         // end::match[]
     }
+
+    void "test uri with slash before params"() {
+        // tag::match[]
```
graemerocher, line 20: Could you remove these `tag:`` and `end::`` comments from the new test as they may break documentation

PR #3652, opened 2020-07-06
Set plugin versions using Maven properties

alejandropg:
In Maven this is the preferred way because it allows any user of the pom parent to change the version simply overwriting the property in their own pom.xml


graemerocher:
Thanks!

alejandropg:
@graemerocher thanks to all of you.
I said to @alvarosanchez, the great work all the team are doing.
Really congrats! 👏👏👏

I will do some more proposals, I hope do you like, and any feedback is really welcomed.

alvarosanchez:
@alejandropg Keep going!

PR #3653, opened 2020-07-06
Migrate discovery-client to an external project

graemerocher:
New project located at https://github.com/micronaut-projects/micronaut-discovery-client


alvarosanchez:
Can we take the opportunity to split the modules so that service discovery and distributed configuration goes in 2 different modules?

graemerocher:
@alvarosanchez I think that would be a good idea for the 2.1.x line, for 2.0.x we should keep it the same but in principal I agree. Also I would say the `compileOnly` dependency on Oracle Cloud should be moved to a separate module

alvarosanchez:
2.1.x of core? That is what master is (or will be) soon.

In my opinion, spinning off core but keeping the same coupling might not be worth. If the structure is going to be the same for 2.0.x, you can leave the module in core.

graemerocher:
No what I meant was 2.0.x branch of https://github.com/micronaut-projects/micronaut-discovery-client/tree/2.0.x

PR #3655, opened 2020-07-06
Set sensible defaults in the pom parent

alejandropg:
Maven properties are used, so the user of the pom parent can easily override them in their own pom.xml

`jdk.version` property is set to 1.8 because this is the lower JDK compatibility supported by Micronaut at this moment.

The next used properties are supported officially by Maven:
- maven.compiler.source
- maven.compiler.target
- maven.compiler.release
- project.build.sourceEncoding
- project.reporting.outputEncoding

### DIFF
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -10,6 +10,13 @@ ext.startPomInfo = {
 }
 ext.extraPomInfo = {
     delegate.properties {
+        'jdk.version'('1.8')
+        'maven.compiler.source'('${jdk.version}')
+        'maven.compiler.target'('${jdk.version}')
+        'maven.compiler.release'('${jdk.version}')
+        'project.build.sourceEncoding'('UTF-8')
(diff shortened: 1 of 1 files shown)


alejandropg:
@graemerocher my intention is, if you like this, to modify the micronaut-profiles to simplify the generated poms.

I don't know if I should review some other project, maybe micronaut-starter?

graemerocher:
Yes Micronaut-starter is where the poms can be simplified

alejandropg:
But micronaut-profiles is active too, not? Is the proyecto used to generate de CLI, right?

jameskleeh:
@alejandropg We're not maintaining the 1.3 CLI so the starter project is where any changes should be made

alvarosanchez:
@alejandropg as far as the parent POM is concerned, is this PR ready?

alejandropg:
@alvarosanchez yes, this PR is ready 👌

graemerocher:
Thanks!

PR #3661, opened 2020-07-07
Use official maven-compiler-plugin's property `maven.compiler.paramet…

alejandropg:
…ers` to retain parameter names

The use of this property instead of explicit configuration is preferred because the user of the pom parent can change the plugin configuration and don't lose this behaviour. Even if she wants to change it, is easier to simply override the property in their own pom.xml

### DIFF
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -14,6 +14,7 @@ ext.extraPomInfo = {
         'maven.compiler.source'('${jdk.version}')
         'maven.compiler.target'('${jdk.version}')
         'maven.compiler.release'('${jdk.version}')
+        'maven.compiler.parameters'('true')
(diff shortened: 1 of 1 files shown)


alejandropg:
For now this is my last PR in the parent. Now I would like to take a look to the starter project to take account the changes of the parent.
@alvarosanchez please some advice about where I should look (just point me a directory for example).
Thanks!

alvarosanchez:
The `pom.xml` of created applications is generated out of https://github.com/micronaut-projects/micronaut-starter/blob/master/starter-core/src/main/java/io/micronaut/starter/feature/build/maven/templates/pom.rocker.raw

It has conditionals based on selected features, java version, language, test framework, etc. Let me know if you have further questions.

PR #3662, opened 2020-07-07
Remove consul version

debuglevel:
The documentation suggests that "only" consul 1.2 works with micronaut (actually meaning "1.2 or newer"). As other documentation parts about consul also do not reference any version, it might be just removed from the documentation.


graemerocher:
Thanks for the contribution!

PR #3665, opened 2020-07-07
Use micronaut-gradle-plugins:2.0.4 release version

alvarosanchez:


build.gradle
```diff
@@ -1,13 +1,12 @@
 buildscript {
     repositories {
-        mavenLocal()
         maven { url "https://oss.jfrog.org/oss-snapshot-local" }
```
graemerocher, line 3: I believe this snapshot repo can be removed as well?

alvarosanchez: Yeah, thought about removing it but wasn't sure if it was there because of something else.

I'll get rid of it and see what happens.

PR #3682, opened 2020-07-08
Reset the body after encoding the request.

### LINKED ISSUE #3619: @Retryable @Client sends empty/invalid body in the retries
### Steps to Reproduce

1. Declarative Http Client with @Retryable
2. An endpoint with POST that require non empty body
3. Send request using the client towards the endpoint which returns >399 response

### Expected Behaviour

Client retries due to non successful response, all requests contains same body.

### Actual Behaviour

First attempt body is correct, following attempts the body is:

`    "body" : {
      "contiguous" : true,
      "direct" : false,
      "readable" : false,
      "writable" : false,
      "readOnly" : false
    }`

### Environment Information

- **Operating System**: MacOsx
- **JDK Version:** 11

### Example Application

https://github.com/pgisslen/micronaut-client-retry-issue

graemerocher:
Fixes #3619

- Test for issue #3619
- Reset the body after encoding the request. Fixes #3619

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1475,7 +1475,6 @@ public class DefaultHttpClient implements
                 ByteBuf bodyContent = null;
                 if (hasBody) {
                     Object bodyValue = body.get();
-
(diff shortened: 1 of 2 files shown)


jameskleeh:
Seems there are checkstyle failures

PR #3687, opened 2020-07-08
Prioritize HTTP/2 over HTTP/1.1 when certificate is provided. Fixes #3299

### LINKED ISSUE #3299: HTTP/2 Problems When Specifying KeyStore
I am having trouble getting HTTP/2 to work using M3. I've enabled it, but Chrome and Firefox are both still using HTTP/1.1. There are a bunch of online tools that check whether HTTP/2 is enabled, but they give conflicting results:

- [http2.pro](https://http2.pro/) says **No**
- [gf.dev](https://gf.dev/http2-test) says **Yes**
- [nixcp.com](https://nixcp.com/tools/http2-test/) says **No**

### Environment Information

- **Operating System**: Amazon Linux 2.0
- **Micronaut Version:** 2.0.0.M3
- **JDK Version:** OpenJDK Runtime Environment Corretto-11.0.7.10.1

### Example

When putting together an example to illustrate the problem, I initially excluded `keyStore` and used `build-self-signed` instead so that it would be easier to run. I was a little surprised to see that it worked perfectly with `build-self-signed`. I still don't know exactly what the problem is, but it seems like it might somehow related to `keyStore` being specified. The TLS certificate I am using is whatever Let's Encrypt gives you by default for a request such as `--domains "example.com"  --domains "*.example.com"`.	

[Example.zip](https://github.com/micronaut-projects/micronaut-core/files/4659416/Example.zip)

RyanHoldren:
Fixes #3299


graemerocher:
Thanks for the contribution!

PR #3690, opened 2020-07-08
update docs specify multiple configuration params via command line - closes Issue 3622

### LINKED ISSUE #3622: Document how to supply configuration via command line
Thanks for reporting an issue for Micronaut, please review the task list below before submitting the

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (http://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Task List

### Steps to Reproduce

In official docs: https://docs.micronaut.io/latest/guide/config.html, there are many ways to config application. It is convinet to config application use different files in different environment.  And  there is an example like this:

```bash
java -Dmicronaut.environments=foo,bar -jar myapp.jar
```

I can use it to switch environment in gradle comman line:

```bash
./gradlew run -Dmicronaut.environments=test
```

But I can't use this format _-Dkey=value_ to pass other arguments to config applications like:

```bash
./gradlew run -Dlogger.levels.ai.inceptio=DEBUG
```

This adds a lot of work to Gitlab CI. I have to use extra enviroment application config yaml file and take some times to deploy them in a cloud env.

Now, I know that I can do this with environment variables......

But I still think you can consider about it. 

### Expected Behaviour

I want to pass arguments in command line with gradle like Spring Boot:

```bash
./gradlew bootRun -Pargs=--spring.main.banner-mode=off,--customArgument=custom
```

In micronaut, it should be:

```bash
./gradlew run -Pargs=logger.levels.root=DEBUG
```

or 

```bash
./gradlew run -Dlogger.levels.ai.inceptio=DEBUG
```

### Actual Behaviour

It seems that Micronaut doen't support config applications with  comman-line arguments now.

### Environment Information

- **Operating System**: Ubuntu 18.04
- **Micronaut Version:** 1.3.6
- **JDK Version:** 11

### Example Application

https://www.baeldung.com/spring-boot-command-line-arguments

ladybug:
Closes #3622


src/main/docs/guide/config/environments.adoc
```diff
@@ -16,6 +16,13 @@ $ java -Dmicronaut.environments=foo,bar -jar myapp.jar
 
 The above activates environments called `foo` and `bar`.
 
+To specify configuration via command line:
+[source,bash]
+----
+$ ./gradlew run --args="-endpoints.health.enabled=true -micronaut.environments=test"
```
jameskleeh, line 22: This isn't correct because `micronaut.environments` is not a configuration property. It's only read as a system property. 

The maven example will achieve the desired result, however isn't exactly the same thing as the gradle example because the maven one is just setting system properties. The gradle example will not result in a system property for `endpoints.health.enabled`.

The maven example should be using https://micronaut-projects.github.io/micronaut-maven-plugin/latest/run-mojo.html#appArguments

It also probably makes sense to include this at the end of the section titled "Included PropertySource Loaders"

jameskleeh: Also it's a good idea to add a note about how the args must be passed to `Micronaut.run`

ladybug: Ok, @jameskleeh I have pushed changes that *I think* address all your concerns... Please let me know how you normally reference a snippet in more than one place... I see the snippets tags, and I tried that and an include, but it doesn't seem to work for me

ladybug: fixed

src/main/docs/guide/config/commandLine.adoc
```diff
@@ -0,0 +1,21 @@
+To specify configuration via command line:
```
jameskleeh, line 1: Can this file be deleted?

ladybug: yes, it is garbage

PR #3698, opened 2020-07-09
Validate nullable is not used on primite parameters. Fixes #3666

### LINKED ISSUE #3666: Validate executable methods with primitive arguments with @Nullable
Adding `@Nullable` to a primitive argument allows Micronaut's generated code to pass a null value, which will fail at runtime. We can catch this error at compile time instead. This would apply to all executable methods

ilopmar:
Some comments:

- I needed to declare the TypeElementVisitor as `TypeElementVisitor<Object, Object>`. If I do `TypeElementVisitor<Object, Executable>` it will work for `@Controller` for not for `@Singleton`, because a `@Singleton` class is not `@Executable`.
- I've also tried with the declarative HTTP Client and the methods there are not being validated. Probably because TypeElemenVisitors only work for classes and not interfaces, right? Is there anything we can do this case?


validation/src/main/java/io/micronaut/validation/executable/ExecutableVisitor.java
```diff
@@ -0,0 +1,43 @@
...
+import io.micronaut.inject.visitor.VisitorContext;
+
+/**
+ * A {@link io.micronaut.inject.visitor.TypeElementVisitor} that validates primitive types can't have a Nullable annotation.
+ *
+ * @author Iván López
+ * @since 2.0.1
+ */
+@Internal
+public class ExecutableVisitor implements TypeElementVisitor<Object, Object> {
```
graemerocher, line 31: Should declare isolating:

```
    @NonNull
    @Override
    public VisitorKind getVisitorKind() {
        return VisitorKind.ISOLATING;
    }
```

jameskleeh:
@ilopmar Regarding clients, I don't think that is a concern because the user executes the method and not us

ilopmar:
So, this change has detected some places in which where using `@Nullable` with `int` :rofl: https://github.com/micronaut-projects/micronaut-core/blob/13c0e2990e55ef9bf068b8a596ddda758807683c/http-client/src/test/groovy/io/micronaut/http/client/aop/StreamSpec.groovy#L169

I'll fix them!

ilopmar:
Changes done and build is green now. Ready to review.

validation/src/main/java/io/micronaut/validation/executable/ExecutableVisitor.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @NonNull
+    @Override
+    public VisitorKind getVisitorKind() {
+        return VisitorKind.ISOLATING;
+    }
+
+    @Override
+    public void visitMethod(MethodElement element, VisitorContext context) {
+        ParameterElement[] parameters = element.getParameters();
```
jameskleeh, line 42: We need to ensure the `MethodElement` is annotated with `@Executable`. Can you look into why `TypeElementVisitor<Object, Executable>` is not working?

validation/src/test/groovy/io/micronaut/validation/executable/NullablePrimitiveSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+    String errorMsg = "Primitive types can not be null"
+
+    void 'test @Nullable is not allowed on primitive parameters'() {
+        when:
+        buildTypeElement("""
+package test;
+
+@javax.inject.Singleton
+class Foo {
+    void bar(@javax.annotation.Nullable boolean boolPrimitive) {
```
jameskleeh, line 18: This type of scenario should still compile because Micronaut will not be executing this method

validation/src/main/java/io/micronaut/validation/executable/ExecutableVisitor.java
graemerocher: If we can limit this as much as possible to a pair of meta annotations it is better for incremental compilation

validation/src/test/groovy/io/micronaut/validation/executable/NullablePrimitiveSpec.groovy
ilopmar: I'm sorry James, you are right. I was confused at the beginning and didn't really think about the difference on "user calling it's own code" vs "Micronaut calling user code" (like in a controller). It's fixed now.

validation/src/main/java/io/micronaut/validation/executable/ExecutableVisitor.java
ilopmar: @jameskleeh fixed now. See my comment: https://github.com/micronaut-projects/micronaut-core/pull/3698#discussion_r452714413 and https://github.com/micronaut-projects/micronaut-core/pull/3698/commits/5197f8a986e7d5bb34fd73ce5e679ee4fb8465d5.

@graemerocher Can you please explain what do you mean? I've pushed a commit that makes it work with `TypeElementVisitor<Object, Executable>`. Do you mean making it for something that it's not `Object`? Like  `TypeElementVisitor<Any-other-thing-here, Executable>?

graemerocher: If you have been able to limit it to just `@Executable` that is probably ok. In general `<Object, Object>` is bad for incremental compilation since it has to visit every annotation which breaks the way Gradle's incremental compilation works. Currently this is only a problem with `GraalTypeElementVisitor` I believe, which is the only one that visits everything (really we should try and change that visitor or restructure it to visit fewer annotations)

ilopmar: Thanks for the clarification.

Regarding the `GraalTypeElementVisitor`, if you have any idea or guidance in mind, please open an issue explaining what you have in mind and I can give it a try.

ilopmar:
I've updated the branch with the latest changes from master. I think then we can merge it, right?

PR #3699, opened 2020-07-09
Fix how suspend function results are mapped

### LINKED ISSUE #3667: Invalid status in metrics when an exception is thrown (Micronaut 2.0)
We have tried to upgrade our service to Micronaut 2.0 and found a strange bug with metrics.
we are using kotlin coroutine dispatcher for custom thread pool. Metrics were not being updated with the right status code when an exception is thrown after the upgrade.

### Task List

### Steps to Reproduce

I have created a sample project (https://gitlab.com/anilkd/micronaut-kotlin-sample) to demonstrate the issue.
Sample projects contain 3 endpoints
1. /products/executor{?productId}
returns expected status, message when an exception is thrown and metric `http.server.requests` is updated with right status code
2. /products/scope{?productId}
returns expected status, message when an exception is thrown and metric `http.server.requests` is updated with right status code
3. /products/dispatcher{?productId}
returns expected status, message when an exception is thrown but metric `http.server.requests` is updated with status code 200 rather than 400, This was working fine in previous version of Micronaut 1.3.6

### Expected Behaviour

Metric `http.server.requests` should be updated with the right status code when an exception is thrown

### Actual Behaviour
Metric `http.server.requests` is updated with 200 status code when an exception is thrown
Tell us what happens instead

### Environment Information

- **Operating System**: Mac
- **JDK Version:** 11

### Example Application
https://gitlab.com/anilkd/micronaut-kotlin-sample

jameskleeh:
Fixes #3667


jameskleeh:
Previously for suspend functions we were effectively returning `HttpResponse.OK.body(someFlowable)`. The filters that listened to the `onNext` were getting a response with `OK` even though the resulting body could throw an error and result in an error response being sent. This changes the result to be `Flowable<HttpResponse>` so that filters will have the correct response.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -1407,49 +1408,47 @@ protected void doOnError(Throwable t) {
...
+                                                response = response.body(o);
+                                            }
                                         }
-                                    });
+                                        response.setAttribute(HttpAttributes.ROUTE_MATCH, finalRoute);
+                                        emitter.onSuccess(response);
+                                    }
                                 });
-                            }
+                            }).toFlowable();
```
graemerocher, line 1431: Isn't it just more efficient to use `Flowable.create` instead of `Single.create` if we are just converting it to a flowable afterwards? Only different in the code would be you would have to call `onNext(..)` then `onComplete()`

jameskleeh: Probably - I'll update it

PR #3710, opened 2020-07-10
Add OpenAPI yml and UI to GraalVM resources if Swagger is present.

### LINKED ISSUE #3709: Configure OpenAPI resources for GraalVM automatically
If OpenAPI is present, `GraalTypeElementVisitor` should add the appropriate configuration to `resource-config.json` to expose the OpenAPI yml file and any possible UI.

This is what needs to be generated:

```
{
  "resources": [
    {"pattern":"\\QMETA-INF/swagger\\E"},
    {"pattern":".*/swagger/.*yml"},
    {"pattern":"\\QMETA-INF/swagger/views/rapidoc/index.html\\E"},
    {"pattern":"\\QMETA-INF/swagger/views/redoc/index.html\\E"}
    {"pattern":"\\QMETA-INF/swagger/views/swagger-ui/index.html\\E"}
  ]
}
```

ilopmar:
Fixes #3709

This PR supersedes https://github.com/micronaut-projects/micronaut-starter/pull/296

### DIFF
--- a/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
+++ b/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
@@ -74,7 +74,6 @@ public class GraalTypeElementVisitor implements TypeElementVisitor<Object, Objec
     private static final String RESOURCE_CONFIG_JSON = "resource-config.json";
     private static final String RESOURCES_DIR = "src/main/resources";
     private static final String RESOURCES = "resources";
-    private static final String BUNDLES = "bundles";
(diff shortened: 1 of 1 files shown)


jameskleeh:
Would this be better done in our openapi visitor?

graemerocher:
Right now a single resource config has to be written in order to make it possible for open api visitor to contribute to that we would have to add some extension API to allow adding additional resources in other visitors. That is maybe an improvement for the future, in the interim this is an ok workaround IMO as even if users are not using our swagger support the entries are pretty harmless.

ilopmar:
@jameskleeh if you agree with Graeme's reply, can we please merge this?

PR #3711, opened 2020-07-10
RecoveryInterceptor performance improvement

dstepanov:
I have noticed RecoveryInterceptor is called for all methods, PR caches some of the extractions and is able to skip it on the next call, not sure if MethodExecutionHandle is cacheable.

The other solution would be to set some boolean into context.setAttribute indicating that there is no fallback.

Should the execution of the fallback happen in a separate thread?


graemerocher:
I think an attribute makes more sense then have another global cache. I'm not sure it is possible to generalize what thread a fallback should be executed on.

dstepanov:
Changed, maybe it would be better to always require "@Recoverable" in the future major version.

graemerocher:
Yes I agree

jameskleeh:
@graemerocher @dstepanov Can one of you please file an issue to track the discussion around the proposed change to require `@Recoverable`?

dstepanov:
Created https://github.com/micronaut-projects/micronaut-core/issues/3719

PR #3720, opened 2020-07-14
Create threads named after its executor/event-loop

dstepanov:
It makes debugging much better, you can see what kind of threads are being blocked.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
@@ -23,6 +23,8 @@ import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.util.ArgumentUtils;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.util.concurrent.DefaultThreadFactory;
 
 import javax.inject.Named;
 import javax.inject.Singleton;
@@ -41,23 +43,29 @@ import java.util.concurrent.ThreadFactory;
 @BootstrapContextCompatible
 public class DefaultEventLoopGroupRegistry implements EventLoopGroupRegistry {
     private final EventLoopGroupFactory eventLoopGroupFactory;
-    private final ThreadFactory threadFactory;
     private final BeanLocator beanLocator;
 
     /**
      * Default constructor.
      *
      * @param eventLoopGroupFactory The event loop group factory
-     * @param threadFactory         The thread factory
      * @param beanLocator           The bean locator
      */
-    public DefaultEventLoopGroupRegistry(
-            EventLoopGroupFactory eventLoopGroupFactory,
-            @Named(NettyThreadFactory.NAME) ThreadFactory threadFactory,
-            BeanLocator beanLocator) {
+    public DefaultEventLoopGroupRegistry(EventLoopGroupFactory eventLoopGroupFactory, BeanLocator beanLocator) {
         this.eventLoopGroupFactory = eventLoopGroupFactory;
-        this.threadFactory = threadFactory;
         this.beanLocator = beanLocator;
+    }
+
+    /**
+     * Constructs an event loop group thread factory.
+     *
+     * @param configuration The configuration
+     * @return The thread factory
+     */
+    @EachBean(EventLoopGroupConfiguration.class)
+    @BootstrapContextCompatible
+    protected ThreadFactory eventLoopGroupThreadFactory(EventLoopGroupConfiguration configuration) {
+        return new DefaultThreadFactory(configuration.getName() + "-" + DefaultThreadFactory.toPoolName(NioEventLoopGroup.class));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


dstepanov:
Would be nice to be able to change the `@Named` value for `@EachBean`, something like `@EachBean(value = EventLoopGroupConfiguration.class, suffix = "-thread-factory")` or with a `prefix`.

http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
```diff
@@ -79,21 +87,23 @@ protected EventLoopGroup eventLoopGroup(EventLoopGroupConfiguration configuratio
                                       configuration.getIoRatio().orElse(null)
                               )).orElseThrow(() -> new ConfigurationException("No executor service configured for name: " + executor));
         } else {
+            ThreadFactory threadFactory = beanLocator.getBean(ThreadFactory.class, Qualifiers.byName(configuration.getName()));
```
graemerocher, line 90: This can be replaced by `@Parameter ThreadFactory threadFactory` I believe

http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
```diff
@@ -41,25 +43,31 @@
...
      *
      * @param eventLoopGroupFactory The event loop group factory
-     * @param threadFactory         The thread factory
      * @param beanLocator           The bean locator
      */
-    public DefaultEventLoopGroupRegistry(
-            EventLoopGroupFactory eventLoopGroupFactory,
-            @Named(NettyThreadFactory.NAME) ThreadFactory threadFactory,
-            BeanLocator beanLocator) {
+    public DefaultEventLoopGroupRegistry(EventLoopGroupFactory eventLoopGroupFactory, BeanLocator beanLocator) {
```
graemerocher, line 54: If you do as I suggested with using `@Parameter` then we don't need to change this constructor which is a breaking change. Otherwise we will need a new constructor and to annotate the new one with `@Inject` and deprecate the old to not break code

runtime/src/main/java/io/micronaut/scheduling/executor/ExecutorFactory.java
```diff
@@ -53,31 +69,30 @@ public ExecutorService executorService(ExecutorConfiguration executorConfigurati
...
-
+                return Executors.newCachedThreadPool(getThreadFactory(executorConfiguration));
             case SCHEDULED:
-                return executorConfiguration
-                    .getThreadFactoryClass()
-                    .flatMap(InstantiationUtils::tryInstantiate)
-                    .map(factory -> Executors.newScheduledThreadPool(executorConfiguration.getCorePoolSize(), factory))
-                    .orElse(Executors.newScheduledThreadPool(executorConfiguration.getCorePoolSize(), threadFactory));
-
+                return Executors.newScheduledThreadPool(executorConfiguration.getCorePoolSize(), getThreadFactory(executorConfiguration));
```
graemerocher, line 76: I believe we can just add `@Nullable @Parameter ThreadFactory threadFactory` to avoid having to do the manually bean locator lookups

runtime/src/main/java/io/micronaut/scheduling/executor/NamedThreadFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Creates new named threads.
+ *
+ * @author Denis Stepanov
+ * @since 2.0.1
+ */
+@Internal
+public class NamedThreadFactory implements ThreadFactory {
```
graemerocher, line 31: Can this be made package private?

http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
dstepanov: That will always inject a thread factory, we don't need one when the executor is provided.

http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
dstepanov: The class is annotated as `@Internal` so I assume no one should be using it.

runtime/src/main/java/io/micronaut/scheduling/executor/ExecutorFactory.java
dstepanov: SCHEDULED doesn't need a `threadFactory`

runtime/src/main/java/io/micronaut/scheduling/executor/NamedThreadFactory.java
dstepanov: Changed

http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
graemerocher: Ah ok

dstepanov:
@graemerocher Do you want some changes?

graemerocher:
Replaced by https://github.com/micronaut-projects/micronaut-core/pull/3761

PR #3732, opened 2020-07-15
Fix discovery-client group

ilopmar:
The new groupId is now `io.micronaut.discovery`: https://github.com/micronaut-projects/micronaut-discovery-client/blob/159641d2f25907583e36342964f8b25c810e00ab/build.gradle#L14

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -157,7 +157,7 @@ ext {
             ],
             'micronaut.discovery': [
                     version:micronautDiscoveryClientVersion,
-                    group:'io.micronaut',
+                    group:'io.micronaut.discovery',
                     name:'micronaut-discovery-client'
             ],            
             'micronaut.hibernate.validator': [


ilopmar:
Ummm, it seems we haven't published a release of `discover-client` with this new group. 

@graemerocher should we publish 2.0.0 with the new groupID? I've noticed that both `master` and `2.0.x` branch in `discovery-client` define the next version as `2.1.0.BUILD-SNAPSHOT`. Has this done on purpose?

ilopmar:
Actually we've already published snapshots only for `2.1.0`: http://oss.jfrog.org/oss-snapshot-local/io/micronaut/discovery/micronaut-discovery-client/

jameskleeh:
We shouldn't be changing the groupid for 2.0.x

ilopmar:
Makes sense, then we need to change it here: https://github.com/micronaut-projects/micronaut-discovery-client/blob/159641d2f25907583e36342964f8b25c810e00ab/build.gradle#L14 and release new version (we haven't released anything yet).

ilopmar:
@graemerocher thoughts?

jameskleeh:
@ilopmar Its not changed for the 2.0.x branch https://github.com/micronaut-projects/micronaut-discovery-client/blob/2.0.x/build.gradle#L14

ilopmar:
Ouch! I missed that. I think I checked that branch to but obviously I didn't. Sorry about the noise. I'll change my PR in discovery-client that fixes the GraalVM to target that branch.

Issue #3734, opened 2020-07-15
Random hotswap problems with classes re-genereated by kapt


l0co:
In `2.0.0`.

I'm trying to debug why hotswap doesn't work for my project, during development in Idea. I use gradle build and I run services using gradle. I also use kapt instead of standard annotation processor, because the project is partially written in kotlin. My configuration (important part) is the following:

```
build.gradle
------------

    kapt {
        arguments {
            arg("micronaut.processing.incremental", true)
            arg("micronaut.processing.annotations", "io.luna.*") // this is my main package
        }
    }

    dependencies {
        kapt platform("io.micronaut:micronaut-bom:$micronautVersion")
        kapt "io.micronaut:micronaut-inject-java"
        kapt "io.micronaut:micronaut-validation"
        kaptTest platform("io.micronaut:micronaut-bom:$micronautVersion")
        kaptTest "io.micronaut:micronaut-inject-java"

gradle.properties
-----------------

kapt.include.compile.classpath=false
kapt.use.worker.api=true
kotlinVersion=1.3.71
org.gradle.parallel=true
```

Now, I'm trying to test hotswap just by changing some string inside a bean method. It randomly works and randomly says "hotswap failed, hierarchy change not implemented". I've compared build output for successful and in a nutshell here's the difference in the failed hotswaps:

![image](https://user-images.githubusercontent.com/1880196/87540885-a7c51700-c6a0-11ea-888b-68e70bb36331.png)

It seems you use a `Set` or something similar to keep superclasses, and if the order is changed, the class is reloaded and for some reason this ends up with a failed hotswap.

The other difference I've found, which didn't result in failed hotswap, but it concerns the class which didn't change, is here:

![image](https://user-images.githubusercontent.com/1880196/87541339-62551980-c6a1-11ea-8083-fb2c924d85d2.png)

I'm not 100% sure the above is a reason for failed hotswap, I'm just trying to find a solution to the problem. I don't know also why kapt decides to process these classes again on recompilation, while they weren't changed (they only refer to the singleton which were changed).

However, I think, that for the consistency, if the class weren't changed, Micronaut annotation processor should generate exactly the same bytecode as in the previous run, and this would be the desired fix for me.

l0co:
I've checked in aside project that hotswap indeed doesn't work when the order of implemented interfaces changes (Micronaut is not involved in this project). So now I'm sure that this is at least a part of this problem.

graemerocher:
[...]

PR #3744, opened 2020-07-16
Predictible order for BeanDefinitionWriter.interfaceTypes & loadTypeM…

l0co:
Relates to #3734


l0co:
@graemerocher I'm not 100% sure this will work, because it still depends on the order of calling particular `interfaceTypes.add()` invocations, and there are a few of them in the `BeanDefinitionWriter` code. But this should be reviewed by someone who sits closer to this code...

Maybe using `TreeSet` with class name comparator will be safer?

l0co:
Build failed after switching from `HashSet` to `LinkedHashSet`? :)

ilopmar:
@l0co we had an issue with the Spock version. Please rebase master.

ilopmar:
@l0co I've merged master and the build is green now :-)

graemerocher:
I would say we should perform a sort by interface name as well yes

l0co:
@graemerocher  I've changed `LinkedHashSet` to `TreeSet` and now I think this will work in all screnarios.

@ilopmar Thanks for merging. However, now, after merging my own branch with current master it shows that `micronautSqlVersion` was increased from `2.3.8` to `2.3.9` in my branch. I don't know why, because this change is already done in master (?).

ilopmar:
@l0co you may want to try to rebase master

l0co:
@ilopmar I did that and it turned our wrong, because now a new commit has been created. I should just have merged master in my branch... Anyway, if you merge this change soon before next `micronautSqlVersion` bump it shouldn't be harmful. Otherwise I can close this PR and create another one, clean.

graemerocher:
Thanks for the contribution!

PR #3761, opened 2020-07-21
Create threads named after its executor/event-loop

graemerocher:


dstepanov:
There can be one problem if you have event loops and executors named the same, that’s why I suggested to be able to prefix or suffix the name of a created bean using EachBean

graemerocher:
Hmm yeah that is not possibly right now. Would require a further change. @jameskleeh thoughts?

runtime/src/main/java/io/micronaut/scheduling/executor/UserExecutorConfiguration.java
```diff
@@ -37,18 +41,53 @@
...
+     * @param name the name
+     * @param nThreads number of threads
+     * @param type the type
+     * @param parallelism the parallelism
+     * @param corePoolSize the core pool size
+     * @param threadFactoryClass the thread factory class
      */
-    protected UserExecutorConfiguration() {
+    @Inject
+    protected UserExecutorConfiguration(@Nullable @Parameter String name,
```
jameskleeh, line 73: Are you certain this works? I believe this has to be annotated with `@ConfigurationInject`?

jameskleeh:
Overall this seems like a pretty big change. This wouldn't go into a point release, perhaps 2.1. I don't know if there is any code out there that is relying on the existing names of the threads, so I'm not sure if this would be considered a breaking change or not. In general I'm in favor of having the thread names be more informative

dstepanov:
The conflict can occur only if somebody managed to set up a custom event loop in 2.0 with the same name as an executor. Probably not a big probability.

graemerocher:
I agree it is probably a good improvement for 2.1

graemerocher:
@dstepanov can you rebase this onto 2.1.x

dstepanov:
@graemerocher I don't think I can technically rebase your branch 😄 I would need to create a new PR

graemerocher:
ah damn, forgot that this was mine.. will handle it. Sorry for the noise 🤦‍♂️

graemerocher:
Thanks for the contribution!

PR #3770, opened 2020-07-22
added missing lombok dependency for the maven configuration

danielprinz:
The configuration example for gradle is correct, but for the maven configuration the dependency declaration is missing. See the documentation here: https://projectlombok.org/setup/maven

When adding the dependency lombok works correctly in my project. 🙂


graemerocher:
Thanks for the contribution!

PR #3783, opened 2020-07-23
i18n ticket - describe/doc how it works

ladybug:
closes Issue #2617


src/main/docs/guide/toc.yml
```diff
@@ -52,6 +52,8 @@ config:
   # validation: Validating Configuration
   jmx:
     title: JMX Support
+  i18n:
+    title: Internationalization
```
sdelamo, line 56: This should be a top level instead of nested inside configuration.

src/main/docs/guide/config/i18n.adoc
```diff
@@ -0,0 +1,31 @@
...
+        Micronaut.run(Application.class);
+        Locale defaultLocale = new Locale();
+        ResourceBundle messages = ResourceBundle.getBundle("messages", defaultLocale);
+        String msg1 = messages.getString("app.title");
+        System.out.println( msg1 ); // this will print "micronaut.demo"
+    }
+}
+
+Micronaut provides a class called
+link:{api}/io/micronaut/context/i18n/ResourceBundleMessageSource.html[ResourceBundleMessageSource] to read the user
```
sdelamo, line 28: `link:{api}/io/micronaut/context/i18n/ResourceBundleMessageSource.html[ResourceBundleMessageSource]`can be simplified as `api:context.i18n.ResourceBundleMessageSource[]`.

src/main/docs/guide/config/i18n.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+messages.properties - file will define key value pairs like this:
+
+[source,text]
+app.title=micronaut.demo
+
+To retrieve this message inside your application, you will use the ResourceBundle like this:
+
+[source,java]
+package com.example;
```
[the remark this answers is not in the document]
graemerocher: Also need to be multi language

sdelamo:
The pr contained an asciidoc file without extension `bundle` which was not referenced from the table of contents `toc.yml`. 

I have removed it. 

When you reference classes in micronaut use the ascidoctor api macro. 

That it is to say: 

Instead of `MessageSource` use  `api:context.MessageSource[]` You don’t need to specify `io.micronaut as the package preffix`.

Instead of `ResourceBundleMessageSource` use `api:context.i18n.ResourceBundleMessageSource[]`. 

You can create links to JavaSE documentation with e.g. 

`link:{javase}/java/util/ResourceBundle.html[Resource Bundles]` generates a link to [https://docs.oracle.com/javase/7/docs/api/java/util/ResourceBundle.html](https://docs.oracle.com/javase/7/docs/api/java/util/ResourceBundle.html). 

I have created `Internationalization` at a root level instead of inside `Configuration`. 

When you show code samples in the docs, we should strive to use code samples coming from real tests to ensure documentation keeps up to date and valid. 

See `snippet` macro and how it used to import code from Java, groovy and Kotlin tests. 

Moreover, I don’t think we should encourage to use: ` ResourceBundle messages = ResourceBundle.getBundle("messages", defaultLocale);´ but `new ResourceBundleMessageSource("io.micronaut.docs.i18n.messages”).getMessage("hello", MessageSource.MessageContext.of(new Locale("es"))).get()`

Moreover, `ResourceBundleMessageSource` should be instantiated in a Singleton instead of each time a message is retrieved.

PR #3789, opened 2020-07-24
Replace "INFO:" with "NOTE:" to show a notice box

debuglevel:
It seems that "INFO:" does not render into any notice box as "NOTE:" does.


graemerocher:
Thanks for the contribution!

Issue #3767, opened 2020-07-21
Allow for annotation mappers to supply stereotypes


tomas-langer:
### Expected Behaviour

When I add an existing third party annotation, I would like to add an interceptor that would intercept methods (or constructors or all if on class level) annotated with it.

Pseudocode (assuming my third party annotation is `@Timed`

It would be great to have a simple way to do this, such as (very very simple):
```
registerInterceptor(Timed.class, TimedInterceptor.class);
```

or providing the annotation metadata, such as:

```
addMetaAnnotation(Timed.class, emptyAnnotation(Around.class));
addMetaAnnotation(Timed.class, withValue(Type.class, TimedInterceptor.class))
``` 

### Actual Behaviour

Currently I need to implement a new annotation with the same signature that contains the desired meta annotations and create a `Mapper` or `Transformer` that would copy the values from the `Timed` annotation to my custom annotation.

### Environment Information

- **Operating System**: any
- **Micronaut Version:** 2.0.0
- **JDK Version:** 11.0.7

pollend:
I think it would be simpler to allow a mapped annotation to also specify a list of parent sterotypes. I was either thinking of returning a list from AnnotationMapper or updating AnnotationValue with a way to specify a parent annotations. The mapper just attaches the annotation so when aop looks at the Around annotation it can't find an associated Type annotation since Type isn't a sterotype of Around. what would be most appropriate? @graemerocher 

```
AnnotationValue.builder(Test.class)
.parent(Test1.class,Test2.class)
```

graemerocher:
A parent method makes sense to me. Should probably be called stereotype for consistency

pollend:
Thanks for clearing that up. I can work on a pr with some test cases.

pollend:
Just realized the sterotype is not quite enough for this to work how I was thinking. Type.class is picked up from the annotation.  

```
List<Class<? extends Annotation>> annotations = method.getAnnotationTypesByStereotype(annotationType);

        Set<Class> applicableClasses = new HashSet<>();

        for (Class<? extends Annotation> aClass: annotations) {
            if (annotationType == Around.class && aClass.getAnnotation(Introduction.class) != null) {
                continue;
            } else if (annotationType == Introduction.class && aClass.getAnnotation(Around.class) != null) {
                continue;
            }
            Type typeAnn = aClass.getAnnotation(Type.class);
            if (typeAnn != null) {
[...]

PR #3795, opened 2020-07-26
Allow for annotation mappers to supply stereotypes

pollend:
This adds a way to define a mapped stereotype for an annotation builder. I mainly want to be able to define an @Arround annotation with a Type so a targeted annotation could be used with AOP.  I explain the problem in the issue and PR's listed below for additional context. 

https://github.com/micronaut-projects/micronaut-core/issues/3767
https://github.com/resilience4j/resilience4j/pull/967
https://github.com/micronaut-projects/micronaut-ignite/pull/23

```
AnnotationValue.builder(Test.class)
.parent(Test1.class,Test2.class)
```

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/SourceAnnotationValue.java
+++ b/core/src/main/java/io/micronaut/core/annotation/SourceAnnotationValue.java
@@ -27,7 +27,7 @@ import java.util.Map;
 import java.util.function.Function;
 
 /**
- * A runtime representation of the an stereotype annotation and its values.
+ * A compile time representation of an annotation.
  *
  * <p>This class implements the {@link AnnotationValueResolver} interface and methods such as {@link AnnotationValueResolver#get(CharSequence, Class)} can be used to retrieve the values of annotation members.</p>
  *
--- a/core/src/main/java/io/micronaut/core/annotation/SourceAnnotationValue.java
+++ b/core/src/main/java/io/micronaut/core/annotation/SourceAnnotationValue.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.annotation;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.micronaut.core.convert.value.ConvertibleValues;
+import io.micronaut.core.util.ArgumentUtils;
+
+import java.lang.annotation.Annotation;
+import java.lang.annotation.RetentionPolicy;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * A runtime representation of the an stereotype annotation and its values.
+ *
+ * <p>This class implements the {@link AnnotationValueResolver} interface and methods such as {@link AnnotationValueResolver#get(CharSequence, Class)} can be used to retrieve the values of annotation members.</p>
+ *
+ * <p>This this class implements a list of {@link AnnotationValue} for parent stereotypes</p>
+ *
+ * <p>If a member is not present then the methods of the class will attempt to resolve the default value for a given annotation member. In this sense the behaviour of this class is similar to how
+ * a implementation of {@link Annotation} behaves.</p>
+ *
+ * @param <A> The annotation type
+ * @author Michael Pollind
+ * @since 1.0
+ */
+public class SourceAnnotationValue<A extends Annotation> extends AnnotationValue<A> {
+    private final List<AnnotationValue<? extends Annotation>> stereotypes;
+
+    /**
+     * @param annotationName the annotation name
+     * @param values         the values
+     */
+    public SourceAnnotationValue(String annotationName, Map<CharSequence, Object> values) {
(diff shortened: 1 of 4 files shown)


jameskleeh:
I don't think you can store the stereotypes just as names because they should be retrievable from the annotation metadata with values.

pollend:
I need to do something a bit different. its not possible to always have the stereotype data available which makes this kind of confusing.

jameskleeh:
@pollend Not sure what you mean. Can you share an example?

pollend:
This should be ready.

core/src/main/java/io/micronaut/core/annotation/SourceAnnotationValue.java
```diff
@@ -0,0 +1,167 @@
...
+
+import java.lang.annotation.Annotation;
+import java.lang.annotation.RetentionPolicy;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * A runtime representation of the an stereotype annotation and its values.
```
jameskleeh, line 30: This javadoc isn't right. Should also mention how this only used at compile time

inject-java/src/test/groovy/io/micronaut/inject/annotation/AnnotationMapperSpec.groovy
```diff
@@ -113,10 +145,31 @@ class Test {
...
+
+            AnnotationValueBuilder<CustomAround> customAroundValueBuilder =
+                AnnotationValue.builder(CustomAround.class);
+
+            AnnotationValueBuilder<Type> typeAnnotationValueBuilder =
+                AnnotationValue.builder(Type.class);
+
+            AnnotationValueBuilder<Around> aroundAnnotationValueBuilder =
+                AnnotationValue.builder(Around.class)
+                    .stereotype(typeAnnotationValueBuilder.build(), customAroundValueBuilder.build());
```
jameskleeh, line 160: Type should not be a stereotype of Around, Type and Around should both be stereotypes of CustomAround. The type should also reference a class. It would be good to build a context with the runtime compiled class and verify the advice is actually referenced

graemerocher:
Yeah I have my doubts this will actually work without further changes because of the way `resolveAroundInterceptors` is written

pollend:
@graemerocher I just realized this won't work. all this is done through reflection. 

```
  for (Class<? extends Annotation> aClass: annotations) {
            if (annotationType == Around.class && aClass.getAnnotation(Introduction.class) != null) {
                continue;
            } else if (annotationType == Introduction.class && aClass.getAnnotation(Around.class) != null) {
                continue;
            }

            Type typeAnn = aClass.getAnnotation(Type.class);
            if (typeAnn != null) {
                applicableClasses.addAll(Arrays.asList(typeAnn.value()));
            }
        }
```

graemerocher:
Right, that code would need to be adapted somehow as well.

pollend:
so I was testing this type of configuration.

```
static class CustomAround1Mapper implements AnnotationMapper<CustomAround1> {
        @Override
        List<AnnotationValue<?>> map(AnnotationValue<CustomAround1> annotation, VisitorContext visitorContext) {

            AnnotationValueBuilder<Type> typeAnnotationValueBuilder =
                AnnotationValue.builder(Type.class).member("value",
                    new AnnotationClassValue(DefaultRetryInterceptor.class));

            AnnotationValueBuilder<Around> aroundAnnotationValueBuilder =
                AnnotationValue.builder(Around.class);

            AnnotationValueBuilder<CustomAround1> customAroundValueBuilder =
                AnnotationValue.builder(CustomAround1.class).stereotype(typeAnnotationValueBuilder.build(), aroundAnnotationValueBuilder.build());


            return Arrays.asList(customAroundValueBuilder.build())
        }
    }

    static class CustomAroundMapper implements AnnotationMapper<CustomAround> {
        @Override
        List<AnnotationValue<?>> map(AnnotationValue<CustomAround> annotation, VisitorContext visitorContext) {

            AnnotationValueBuilder<Type> typeAnnotationValueBuilder =
                AnnotationValue.builder(Type.class).member("value",
                    new AnnotationClassValue("io.micronaut.aop.TraceInterceptor"));

            AnnotationValueBuilder<Around> aroundAnnotationValueBuilder =
                AnnotationValue.builder(Around.class);

            AnnotationValueBuilder<CustomAround> customAroundValueBuilder =
                AnnotationValue.builder(CustomAround.class).stereotype(typeAnnotationValueBuilder.build(), aroundAnnotationValueBuilder.build());


            return Arrays.asList(customAroundValueBuilder.build())
        }
    }
```

The sterotype gets overwritten by the first annotation. Sterotypes only link an annotation by name so there is no way to pick out a sterotype by a unique annotation. umm, not exactly sure what to do in this case.

graemerocher:
we already have `metadata.getAnnotationNamesByStereotype(..)` for that purpose

pollend:
What do you suggestion could be done? stereotypes have to be unique but if you have more then one annotation with a set of unique stereotypes but the same annotation, they will overwrite each other. Annotation one with a Type stereotype for one interceptor and another type annotation for another interceptor will be overwritten by the last type annotation.

jameskleeh:
@graemerocher Is this why the annotations are retrieved via reflection in the first place?

graemerocher:
I can't recall why we did it this way but it seems like something we should improve

jameskleeh:
Seems like we should have `@Types` which has multiple `@Type`. I think that would allow all `@Type` annotations to exist

pollend:
I'm trying to work out how would Types work in this case. Type is just an array of classes would would Types be an array of an array of classes? Just a little worried that I would end up changing the semantic. Around and Introduction are unique to their stereotype which is good and they are still sent to their associated intercepts. In the scope of this it doesn't matter if the type is associated with one or both in this case. Is Type used anywhere besides intercepts where this might be a problem? This change isn't as straightforward as I thought it would be.

pollend:
so If I group up types won't the same intercept be added multiple times? looks like it just picks up the Type and adds it to the list of intercepts. Different @Around annotations with reference the same list of types.

jameskleeh:
This will be included in Micronaut 3

PR #3802, opened 2020-07-27
Bind query values in websocket servers

dorian-yeager:


vinhbt:
I think it should be include fix miss build ssl context in buildSslContext function when scheme is wss.

jameskleeh:
@graemerocher I think the most correct fix for this is:

1. Change the request binder registry to add a specific binder for `@QueryValue`
2. Change the request binder registry to not return a default binder. A new method would be created to emulate the current behavior. This is a breaking change, however you could argue the DefaultRequestBinderRegistry should be internal.. The contract returns an optional, however an empty optional is never returned in the current implementation.
3. Remove the hacky logic in WebSocketStateBinderRegistry https://github.com/micronaut-projects/micronaut-core/blob/master/websocket/src/main/java/io/micronaut/websocket/bind/WebSocketStateBinderRegistry.java#L70
4. Create a new default binder for the websocket binder registry

The fix I have put in place keeps the query value binding logic in one place without changing the behavior of the request binder registry

graemerocher:
@jameskleeh sounds reasonable, in the meantime should be this PR be merged?

PR #3824, opened 2020-07-30
Fix "Recursive update" for scoped proxies

dstepanov:
Looks like there is one more problem with "Recursive update" update in 2.0.1
```
        java.lang.IllegalStateException: Recursive update
            at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1760)
            at io.micronaut.context.DefaultBeanContext.getScopedBeanForDefinition(DefaultBeanContext.java:2316)
            at io.micronaut.context.DefaultBeanContext.getBeanForDefinition(DefaultBeanContext.java:2298)
            at io.micronaut.context.DefaultBeanContext.getBeanInternal(DefaultBeanContext.java:2270)
            at io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:1240)
            at io.micronaut.context.AbstractBeanDefinition.getBeanForConstructorArgument(AbstractBeanDefinition.java:1013)
            at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:1889)
            ... 74 more
```

Trying to fix it with a `Provider` instance and computing the instance after the `computeIfAbsent` call.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -107,7 +107,7 @@ public class DefaultBeanContext implements BeanContext {
 
     final Map<BeanKey, BeanRegistration> singletonObjects = new ConcurrentHashMap<>(100);
     final Map<BeanIdentifier, Object> singlesInCreation = new ConcurrentHashMap<>(5);
-    final Map<BeanKey, Object> scopedProxies = new ConcurrentHashMap<>(20);
+    final Map<BeanKey, Provider<Object>> scopedProxies = new ConcurrentHashMap<>(20);
 
     Set<Map.Entry<Class, List<BeanInitializedEventListener>>> beanInitializedEventListeners;
 
@@ -2313,7 +2313,7 @@ public class DefaultBeanContext implements BeanContext {
             Class<?> proxiedType = resolveProxiedType(beanType, definition);
             BeanKey key = new BeanKey(proxiedType, qualifier);
             BeanDefinition<T> finalDefinition = definition;
-            return (T) scopedProxies.computeIfAbsent(key, (Function<BeanKey, T>) beanKey -> {
+            return (T) scopedProxies.computeIfAbsent(key, beanKey -> ProviderUtils.memoized(() -> {
                 Qualifier<T> q = qualifier;
                 if (q == null) {
                     q = finalDefinition.getDeclaredQualifier();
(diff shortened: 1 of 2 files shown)


inject/src/main/java/io/micronaut/context/LazyProvider.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.inject.Provider;
+
+/**
+ * A lazy provider.
+ *
+ * @param <T> The type
+ * @author Denis Stepanov
+ * @since 2.0.2
+ */
+abstract class LazyProvider<T> implements Provider<T> {
```
graemerocher, line 27: Maybe instead of introducing a new class `SupplierUtil.memoized` could be used instead?

dstepanov: It can be, I didn’t know about it, Provider might be useful for direct injection of it.
Do you want to use Supplier instead?

jameskleeh: @dstepanov If you want to create a `ProviderUtil.memoized` that would be fine. Either way the synchronization logic should be externalized for reusability. The supplier one works well so use that as an example

graemerocher: I think an issue there may be `Provider` is part of `java.inject` and not available in the `core` module

dstepanov: I have created an util class in the same package.

dstepanov: @jameskleeh What do you think about the latest version?

jameskleeh: @dstepanov The util doesn't follow the pattern of SupplierUtil. The crux of the implementation is the delegate gets overridden so it doesn't use synchronization after the first invocation. You can message me on gitter if you'd like more details

dstepanov: @jameskleeh Changed

jameskleeh:
Thanks for your contribution!

someshrathi02:
@dstepanov / @jameskleeh , 

Im also facing similar issue.  which build version should I test to validate this fix ? 2.0.2.BUILD-SNAPSHOT ?

jameskleeh:
2.0.2

PR #3826, opened 2020-07-30
Update quickStart.adoc

Deepak25101997:
Added info regarding Maven/Gradle configuration requirements. Will be helpful for new developers.


Deepak25101997:
Sir being new to Java, I thought it would be good to explicitly specify that, user must ensure Maven/Gradle is correctly configured in their machine (like all environment variables configured) before going on to make the app through CLI. So that when user runs the app through CLI, he/she faces no issue. I just thought of this change from a newbie perspective.

jameskleeh:
@Deepak25101997 Thanks for responding. The applications generated via our CLI come with Gradle and Maven wrappers, so its not even necessary to have Gradle or Maven installed on your machine to begin running the applications. Simply use `mvnw` or `gradlew`. Perhaps you can add that information to the docs so it will be more clear for others?

Deepak25101997:
Sure sir. I will add this to info to the relevant area in the docs and generate a pull request.

PR #3830, opened 2020-07-30
Changed Immage spelling to Image

ParhiShubhendu:


sdelamo:
Good catch! @ParhiShubhendu

PR #3835, opened 2020-07-31
Fix error routes, fixes #3834

### LINKED ISSUE #3834: Error routes aren't resolved according to an exception heirarchy
Regression in v2.

Error routes with `Exception`, `Throwable` etc should not be matched first when the error route for `MyException` is available.

Looks like `DefaultRouter` is not calling `findRouteMatch(Map<ErrorRoute, RouteMatch<T>> matchedRoutes, Throwable error)` like it was in the previous version.

dstepanov:


http-server-netty/src/test/groovy/io/micronaut/docs/http/server/exception/SimpleControllerSpec.groovy
```diff
@@ -87,11 +88,25 @@ class SimpleControllerSpec extends Specification {
...
             return HttpResponse.<JsonError>status(HttpStatus.BAD_REQUEST, "Fix it!")
                     .body(error)
         }
+
+        @Error
+        HttpResponse exceptionExceptionHandler(HttpRequest request, Exception throwable) {
+            JsonError error = new JsonError("Invalid: " + throwable.getMessage())
+            return HttpResponse.<JsonError>status(HttpStatus.BAD_REQUEST, "exceptionExceptionHandler")
+                    .body(error)
+        }
```
jameskleeh, line 109: What value is this method adding? Perhaps add a test that throws `Exception` to ensure the one that catches `Throwable` is not invoked?

dstepanov: I was just trying to surround the error handler so it will trigger the problem because I’m not sure what is the logic to find the handler.

PR #3843, opened 2020-08-02
Fix PropertySourcePropertyResolver.java when property key starts with [.

dave08:
This is a proposed fix for the following problem (the problem started in 2.0.0 when PropertyCatalog.NORMALIZED was added, but the null check on normalized wasn't there like for the entries, so it caused this problem):

Keycloak or jboss wildfly had a system property who's name is something like [foo]=... and in PropertySourcePropertyResolver, any key that contains [ is assumed to have a property name before, which in this case is null... so it crashes with:

```
java.lang.NullPointerException
    at io.micronaut.context.env.PropertySourcePropertyResolver.processPropertySource(PropertySourcePropertyResolver.java:539)
    at io.micronaut.context.env.DefaultEnvironment.readPropertySources(DefaultEnvironment.java:399)
    at io.micronaut.context.env.DefaultEnvironment.start(DefaultEnvironment.java:243)
    at io.micronaut.context.env.DefaultEnvironment.start(DefaultEnvironment.java:62)
    at io.micronaut.context.env.DefaultEnvironmentSpec.test system property source loader(DefaultEnvironmentSpec.groovy:101)
```

I just stuck the case in the test as a little side-check, if you want it to be more explicit, I could try. And I'm not sure whether I should have included the `first = false` in the `if` or not...

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -536,7 +536,9 @@ public class PropertySourcePropertyResolver implements PropertyResolver {
                         }
                         if (first) {
                             Map<String, Object> normalized = resolveEntriesForKey(resolvedProperty, true, PropertyCatalog.NORMALIZED);
-                            normalized.put(propertyName, value);
+                            if (normalized != null) {
+                                normalized.put(propertyName, value);
+                            }
                             first = false;
                         }
                     } else {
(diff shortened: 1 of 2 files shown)


dave08:
The current change is basically the same behavior as before 2.0.0 which also ignores such property names, this PR basically just fixes the crash that was caused by the changes in 2.0.0. 

It looks like [..] has a meaning in properties (a map/list key?), so I'm not sure how this would work? `[propName[indexNo]`, `[propName][indexNo]` should work? Or only when the property begins with `[abc]` with no indices? I really had no idea what was better, so I just concentrated on fixing the problem.

I wouldn't mind making a change if I would know what behavior makes more sense... any ideas @jameskleeh ?

jameskleeh:
@dave08 OK - this is definitely an improvement so will merge as is

PR #3850, opened 2020-08-03
Add support for multi dimensional arrays in introspections

### LINKED ISSUE #2573: Introspecting a class containing a multi-dimensional primitive array field (e.g. int[][]) will cause Already an array exception in annotation processor
### Task List

Introspecting a class with a primitive multidimensional array field causes an "Already an array" exception in an annotation processor during compile time with no stack trace

### Steps to Reproduce

 1. Create a project with `mn`.
 2. Create a class with a multidimensional array and attempt to introspect it.
 3. Observe you get an `Already an array` error.

```
git clone https://github.com/doctorpangloss/arrayintrospection-error-repro.git
cd arrayintrospection-error-repro
./gradlew compileJava
```

results in

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> java.lang.IllegalStateException: Already an array
```

### Expected Behaviour

Should introspect objects with multidimensional primitive arrays. Here's the class.

```
package micronaut.tests;

@Introspected
public class IntrospectedFailingClass {
	private int[][] failingField;

	public IntrospectedFailingClass() {
	}

	public int[][] getFailingField() {
		return failingField;
	}

	public IntrospectedFailingClass setFailingField(int[][] failingField) {
		this.failingField = failingField;
		return this;
	}
}
```

### Actual Behaviour

Compile errors.

### Environment Information

- **Operating System**: macOS (native-image run in docker)
- **Micronaut Version:** 1.3.0.M2
- **JDK Version:** OpenJDK 64-Bit Server VM GraalVM CE 19.3.0.2 (build 11.0.5+10-jvmci-19.3-b06, mixed mode, sharing)

### Example Application

https://github.com/doctorpangloss/arrayintrospection-error-repro.git

jameskleeh:
Fixes #2573


jameskleeh:
The only thing that could be considered a breaking change is that for a class element that is a primitive, the name will return the internal class name `I` instead of `int` for example. I could restore the behavior, but it would require another method either on `TypedElement`, or `PrimitiveElement`, which would then require casting in order to retrieve the internal name.

inject/src/main/java/io/micronaut/inject/ast/PrimitiveElement.java
```diff
@@ -0,0 +1,128 @@
...
+    public static final PrimitiveElement VOID = new PrimitiveElement("void", "V");
+
+    private static final PrimitiveElement INT = new PrimitiveElement("int", "I");
+    private static final PrimitiveElement CHAR = new PrimitiveElement("char", "C");
+    private static final PrimitiveElement BOOLEAN = new PrimitiveElement("boolean", "Z");
+    private static final PrimitiveElement LONG = new PrimitiveElement("long", "J");
+    private static final PrimitiveElement FLOAT = new PrimitiveElement("float", "F");
+    private static final PrimitiveElement DOUBLE = new PrimitiveElement("double", "D");
+    private static final PrimitiveElement SHORT = new PrimitiveElement("short", "S");
+    private static final PrimitiveElement BYTE = new PrimitiveElement("byte", "B");
```
ilopmar, line 32: Why declaring this here if they already exist in https://github.com/micronaut-projects/micronaut-core/blob/d116782bd252f2b544594b2b55740bf86c5eb8ed/inject/src/main/java/io/micronaut/inject/writer/AbstractClassFileWriter.java#L91-L100?

jameskleeh: Great catch - I didn't see that map there

PR #3859, opened 2020-08-04
Correct inconsistent mapping of camelCase vs kebab-case properties

dstepanov:
There should be some refactoring of properties processing so keys processing in Jackson registrar is not needed, this PR just fixed cases when kebab case properties were resolved as nulls.


graemerocher:
In general I would like to completely remove jackson from configuration binding in the future

dstepanov:
@graemerocher Can this one make it to the 2.0.2

graemerocher:
Scheduled to have a look

Issue #3837, opened 2020-07-31
@Retryable will execute for all exceptions if all includes are missing on classpath


altery:
# Context
Micronaut 2.0.1

# Description

If none of the Throwables of the "includes" field of @Retryable are on the runtime classpath, the retry is executed for every exception that occurs in the annotated method.

# Expected behaviour

I'd expect that the retry is never executed, because adding at least one Throwable to "includes" changes the semantics from "execute for every exception" to "execute only if the cause is of at least one of the Throwables. This semantic should be independent of the classpath. Alternatively, I'd also understand if an exception would be thrown if the includes/excludes Throwables could not be loaded.

# Analysis

This is caused by the DefaultRetryPredicate, which calculates 'hasIncludes' only based on the number of elements in the includes list. This list is only populated if the specified includes are on the classpath.

See AnnotationRetryStateBuilder#resolveIncludes and AnnotationValue#resolveClassValues

graemerocher:
Seems like you have provided good analysis of the problem, would you like to send a PR?

altery:
While working on the PR I noticed that things behave differently when compiling with JDK8. Micronaut doesn't seem to generate BeanDefinitions for Groovy-Methods that contains annotations not an Classpath. As a consequence, the Spock-Tests do not execute at all: https://github.com/micronaut-projects/micronaut-core/pull/3860/checks?check_run_id=951201393

The following happens when compiling the referenced PR on JDK8 (Amazon and OpenJDK):
```
> ./gradlew :runtime:testCompile --stacktrace
(...)
org.gradle.api.internal.tasks.testing.TestSuiteExecutionException: Could not complete execution for Gradle Test Executor 3.
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:63)
(...)
Caused by: org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to discover tests
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverEngineRoot(EngineDiscoveryOrchestrator.java:104)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:81)
	at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:83)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:66)
[...]

PR #3860, opened 2020-08-04
Do not fallback to 'include all Throwable' in @Retryable, if no includes could be resolved at runtime

altery:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/3837

### DIFF
--- a/runtime/src/main/java/io/micronaut/retry/intercept/AnnotationRetryStateBuilder.java
+++ b/runtime/src/main/java/io/micronaut/retry/intercept/AnnotationRetryStateBuilder.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.retry.intercept;
 
+import io.micronaut.core.annotation.AnnotationClassValue;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.AnnotationValue;
 import io.micronaut.core.reflect.InstantiationUtils;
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/io/micronaut/retry/intercept/AnnotationRetryStateBuilder.java
```diff
@@ -78,17 +78,21 @@ public RetryState build() {
 
     private static RetryPredicate createPredicate(Class<? extends RetryPredicate> predicateClass, AnnotationValue<Retryable> retry) {
         if (predicateClass.equals(DefaultRetryPredicate.class)) {
-            List<Class<? extends Throwable>> includes = resolveIncludes(retry, INCLUDES);
-            List<Class<? extends Throwable>> excludes = resolveIncludes(retry, EXCLUDES);
-            return new DefaultRetryPredicate(includes, excludes);
+            // annotationClassValues are independent from the runtime classpath
+            AnnotationClassValue<?>[] annotationClassValues = retry.annotationClassValues(INCLUDES);
```
jameskleeh, line 82: At this point if the `annotationClassValues` is not empty, but `includes` is, then you could simply return a predicate that always returns false because the thrown annotation could never be in the includes list.

jameskleeh: I don't think modifying the API of `DefaultRetryPredicate` is necessary. Let me know if I have misunderstood

altery: Thanks, that is correct and probably simplifies things. I changed the PR accordingly.

graemerocher:
@altery @jameskleeh status update on this? Seems to be failing tests

altery:
Thanks for triggering, I'll have a look within the next few days.

altery:
To test the behavior, there is an annotation class value present on a test that is not on the CP (OptimisticLockException):

`@Retryable(attempts = '5', delay = '5ms', includes = OptimisticLockException.class)`

It seems the test fails on Java 8 because of the following bug in the JDK:

https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-7183985

The annotation is loaded by the Junit/Jupiter platform, I'm unable to (directly) workaround this problem.

There might be a possibility to move the affected class/bean (CounterService) out of the Junit context, so that junit does not scan for that annotation. This would change the 'style' of the test, though. What do you think, @graemerocher ?

graemerocher:
@altery disable the test just for Java 8. See https://github.com/micronaut-projects/micronaut-core/pull/4043/files#diff-673675b341076a28bc30ccee0b7ca423R38 for an example

altery:
Thanks for the hint, but this doesn't seem to work. I tried both ignoring the test method and the whole test class. I can only speculate that annotation processing of Junit unit takes place before Spock processing of @IgnoreIf

altery:
So I tried a few workaround, s.t. the CounterService is not picked up by the Junit discovery engine, but I couldn't get it to work. The problem occurs very early and there are only a limited amount of conditions that prevent that code path from being taken (private or static methods are excluded, for example). Not sure if this is worth any further work, as the actual root issue is also a corner case.

PR #3866, opened 2020-08-05
Introduce IntrospectionBeanPropertyBinder as alternative to JacksonBeanPropertyBinder

dstepanov:
I had another look at properties binding, PR based on https://github.com/micronaut-projects/micronaut-core/pull/3859
Moved all the keys processing to `JacksonBeanPropertyBinder` with some refactoring.

Probably not for a minor release.

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/configproperties/ConfigurationPropertiesListBindingSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/configproperties/ConfigurationPropertiesListBindingSpec.groovy
@@ -4,15 +4,15 @@ import io.micronaut.context.ApplicationContext
 import io.micronaut.context.annotation.ConfigurationProperties
 import io.micronaut.context.env.PropertySource
 import io.micronaut.context.env.yaml.YamlPropertySourceLoader
+import io.micronaut.core.annotation.Introspected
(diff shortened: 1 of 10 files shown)


graemerocher:
Will need to be rebased. Scheduled to look at it for 2.1

dstepanov:
@graemerocher Rebased, `buildSourceMapNode` is producing a `Map` of `Map,List,String,Number,Boolean` so new introspected impl. can be added.

dstepanov:
@graemerocher My take to add alternative `BeanPropertyBinder` using `BeanIntrospection`.
There are a lot of cases when `ConfigurationProperties`'s beans need to be introspected also some HTTP tests fail for the same reason, it looks like there are some Map->Object conversions inside of HTTP processing.

graemerocher:
@dstepanov nice work, could you target 2.1.x branch? Think we can start off having this disabled and gradually deprecate the jackson one in 3.0

dstepanov:
@graemerocher There are still tests failing because of missing `@Intercepted` do you want to fix them and enable `IntrospectionBeanPropertyBinder` by default?

graemerocher:
We can't enable it by default in 2.x as that would be a breaking change, but we could potentially enable it for newly created apps

dstepanov:
I think `BeanPropertyBinderRegistrar` should throw exception in some of conversion errors.

PR #3873, opened 2020-08-06
Bump micronaut-discovery-client to 2.1.0

micronaut-build:


ilopmar:
:warning: **Do not merge this until Micronaut 2.1.0** because the groupId change to `io.micronaut.discovery`.

Also it's necessary to update the BOM with the new groupId

PR #3874, opened 2020-08-06
Add new utility methods for GraalVM and refactor

ilopmar:
See https://github.com/micronaut-projects/micronaut-sql/issues/275

Once this is merged and Micronaut 2.0.2 released I will merge https://github.com/micronaut-projects/micronaut-sql/pull/278 with the changes.

### DIFF
--- a/core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
+++ b/core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
@@ -24,18 +24,19 @@ import org.graalvm.nativeimage.hosted.Feature.BeforeAnalysisAccess;
 import org.graalvm.nativeimage.hosted.RuntimeClassInitialization;
 import org.graalvm.nativeimage.hosted.RuntimeReflection;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
-
 /**
  * Utility methods for implementing Graal's {@link com.oracle.svm.core.annotate.AutomaticFeature}.
  *
  * @author Álvaro Sánchez-Mariscal
  * @author graemerocher
+ * @author Iván López
  * @since 2.0.0
  */
 @Internal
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
```diff
@@ -44,67 +45,118 @@
...
      * Marks the given class to be initialized at build time, only if it is present.
      *
-     * @param access the {@link BeforeAnalysisAccess} instance
-     * @param className the class name
+     * @param access     The {@link BeforeAnalysisAccess} instance
+     * @param classNames The class names
      */
-    public static void initializeAtBuildTime(BeforeAnalysisAccess access, String className) {
-        findClass(access, className).ifPresent(RuntimeClassInitialization::initializeAtBuildTime);
+    public static void initializeAtBuildTime(BeforeAnalysisAccess access, String... classNames) {
```
sdelamo, line 51: Even if the class is flagged as internal. Should not we keep the method and annotate it with `@Deprecated` and in the java doc tell to use the new one: 

```java
    public static void initializeAtBuildTime(BeforeAnalysisAccess access, String className) {
```

From the old one we can the new one.

core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
```diff
@@ -44,67 +45,118 @@
...
 
     /**
      * Marks the given class to be initialized at runtime, only if it is present.
      *
-     * @param access the {@link BeforeAnalysisAccess} instance
-     * @param className the class name
+     * @param access     The {@link BeforeAnalysisAccess} instance
+     * @param classNames The class names
      */
-    public static void initializeAtRunTime(BeforeAnalysisAccess access, String className) {
```
sdelamo, line 60: Even if the class is flagged as internal. Should not we keep the method and annotate it with `@Deprecated` and in the java doc tell to use the new one: 

```java
    public static void initializeAtBuildTime(BeforeAnalysisAccess access, String className) {
```

From the old one we can the new one.

core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
ilopmar: No, we don't. I asked about this at one engineering meeting and it's totally ok to do "whatever" we want with `@Internal` classes. Besides that, both signatures are compatible.

core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
ilopmar: Same as before.

PR #3883, opened 2020-08-08
Add parsing cookies from Set-Cookie HTTP headers

### LINKED ISSUE #3082: [Feature Request] Parse `set-cookie` response headers into Cookies
Currently it's possible to get the `set-cookie` headers as a `List<String>`

```
HttpHeaders headers = response.getHeaders();
List<String> cookies = headers.getAll(HttpHeaderNames.SET_COOKIE)
```

But then in order to add those to subsequent requests, they must be converted into `io.micronaut.http.cookie.Cookie`

It would be nice to have a `response.getCookies()` that returns `Set<Cookie>` ready for use in follow-up requests.

jantosi:
Expose parsed cookies in HttpMessage

Fixes #3082

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/util/CookieUtilSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/util/CookieUtilSpec.groovy
@@ -33,7 +33,7 @@ class CookieUtilSpec extends Specification {
 
     void "test parsing cookie date from header #setCookieHeader"() {
         given:
-        Cookie cookie = CookieUtil.getCookieFromString("$setCookieHeader")
+        Cookie cookie = CookieUtil.getCookieFromString(setCookieHeader)
 
         expect:
         cookie.getMaxAge() == maxAge
--- a/http-server-netty/src/test/groovy/io/micronaut/http/util/CookieUtilSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/util/CookieUtilSpec.groovy
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.util
+
+import io.micronaut.http.cookie.Cookie
+import io.micronaut.http.cookie.SameSite
+import spock.lang.Specification
+
+import java.time.Duration
+import java.time.LocalDateTime
+import java.time.Month
+import java.time.temporal.ChronoUnit
+
+class CookieUtilSpec extends Specification {
+
+    static long durationInSeconds(Integer year, Month month, Integer day, Integer hour, Integer minute, Integer second){
+        return Duration.between(LocalDateTime.of(year, month, day, hour, minute, second), LocalDateTime.now())
+                .get(ChronoUnit.SECONDS)
+    }
+
+    void "test parsing cookie date from header #setCookieHeader"() {
+        given:
+        Cookie cookie = CookieUtil.getCookieFromString("$setCookieHeader")
+
+        expect:
+        cookie.getMaxAge() == maxAge
+
+        where:
+        setCookieHeader                                                     | maxAge
+        "Set-Cookie: name=value"                                            | 0
+        "Set-Cookie: name=value; max-age=10"                                | 10
+        "Set-Cookie: name=value; max-age=-10"                               | -10
+        "Set-Cookie: name=value; max-age;"                                  | 0
+        "Set-Cookie: name=value; expires=Thu, 04 Feb 2021 03:47:49 GMT;"    | durationInSeconds(2021, Month.FEBRUARY, 4, 3, 47, 49)
+        "Set-Cookie: name=value; expires=Thu, 04-Feb-2021 03:47:49 GMT;"    | durationInSeconds(2021, Month.FEBRUARY, 4, 3, 47, 49)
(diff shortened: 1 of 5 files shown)


http/src/main/java/io/micronaut/http/util/CookieUtil.java
```diff
@@ -0,0 +1,210 @@
...
+                    RFC_5322_RFC_850_FORMATTER,
+                    ANSI_C_ASCTIME_FORMATTER
+            }).map(dateTimeFormatter -> {
+                try {
+                    return LocalDateTime.parse(propertyValue, dateTimeFormatter);
+                } catch (DateTimeParseException e) {
+                    //ignored, try other patterns
+                    return null;
+                }
+            }).filter(Objects::nonNull).findFirst();
```
graemerocher, line 189: Seems a bit of unnecessary overhead to use streams here. Would it not be enough to simply do:

```java
DateTimeFormatter daateTimeFormatter = new DateTimeFormatter[]{
                    DateTimeFormatter.RFC_1123_DATE_TIME,
                    RFC_5322_RFC_850_FORMATTER,
                    ANSI_C_ASCTIME_FORMATTER
}
try {
    return Optional.of(LocalDateTime.parse(propertyValue, dateTimeFormatter));
} catch (DateTimeParseException e) {
    return Optional.empty();
}
```

http/src/main/java/io/micronaut/http/util/CookieUtil.java
```diff
@@ -0,0 +1,210 @@
...
+
+        @Override
+        @NonNull
+        public String getCookieName() {
+            return cookieName;
+        }
+
+        @Override
+        public Optional<String> getCookieDomain() {
+            return Optional.ofNullable(props.get(CookieProperties.DOMAIN.toLowerCase()));
```
graemerocher, line 144: It seems like you are always using the lower case versions of the constants. Doesn't it make sense to have the constants already in lower case?

http/src/main/java/io/micronaut/http/util/CookieUtil.java
```diff
@@ -0,0 +1,210 @@
...
+        @Override
+        public Optional<TemporalAmount> getCookieMaxAge() {
+            // https://tools.ietf.org/html/rfc6265#page-19 - Max-Age has precedence over Expires.
+            if (props.get(CookieProperties.MAX_AGE.toLowerCase()) != null) {
+                String propertyValue = props.get(CookieProperties.MAX_AGE.toLowerCase());
+                long maxAge = propertyValue.isEmpty() ? 0 : Long.parseLong(propertyValue);
+                return Optional.ofNullable(Duration.ofSeconds(maxAge));
+            } else if (props.get(CookieProperties.EXPIRES.toLowerCase()) != null) {
+                Optional<LocalDateTime> expiry = parseCookieDate(props.get(CookieProperties.EXPIRES.toLowerCase()));
+                return expiry.map($ -> Duration.between($, LocalDateTime.now()));
```
graemerocher, line 170: Could you use a variable name instead of `$`

http/src/main/java/io/micronaut/http/util/CookieUtil.java
```diff
@@ -0,0 +1,210 @@
...
+                    if (split[0].trim().equals("")) {
+                        return null;
+                    }
+                    if (split.length < 2) {
+                        return new AbstractMap.SimpleEntry<>(split[0].trim().toLowerCase(), "");
+                    }
+                    return new AbstractMap.SimpleEntry<>(split[0].trim().toLowerCase(), trimQuotes(split[1].trim()));
+                })
+                .filter(Objects::nonNull)
+                .collect(Collectors.toMap(AbstractMap.SimpleEntry::getKey, AbstractMap.SimpleEntry::getValue));
```
graemerocher, line 88: Again the use of streams seems overkill here when you are just populating a map

http/src/main/java/io/micronaut/http/HttpMessage.java
```diff
@@ -45,6 +53,19 @@
      */
     @NonNull HttpHeaders getHeaders();
 
+    /**
+     * @return the cookies of this HTTP message
+     */
+    default @NonNull Cookies getCookies() {
+        Map<CharSequence, Cookie> cookiesByName = getHeaders().getAll(SET_COOKIE).stream()
+                .map(CookieUtil::getCookieFromString)
+                .collect(Collectors.toMap(Cookie::getName, $ -> $));
```
graemerocher, line 62: Could you use a variable name instead of `$`

graemerocher (changes requested):
Thanks for the contribution and sorry for the delay reviewing! Please see the feedback. Thanks!

jameskleeh:
@graemerocher @eahrold Do we really need to implement custom logic for parsing the cookie in the simple case? The netty side should be using the built in Netty logic to do it. What are the current use cases for returning a SimpleHttpResponse from a client? This logic is only necessary for responses that come from a client, not created in a server. Currently SimpleHttpResponse is mutable so it doesn't seem like it should support this functionality at all.

http-server-netty/src/test/groovy/io/micronaut/http/util/CookieUtilSpec.groovy
```diff
@@ -0,0 +1,108 @@
...
+class CookieUtilSpec extends Specification {
+
+    static long durationInSeconds(Integer year, Month month, Integer day, Integer hour, Integer minute, Integer second){
+        return Duration.between(LocalDateTime.of(year, month, day, hour, minute, second), LocalDateTime.now())
+                .get(ChronoUnit.SECONDS)
+    }
+
+    void "test parsing cookie date from header #setCookieHeader"() {
+        given:
+        Cookie cookie = CookieUtil.getCookieFromString("$setCookieHeader")
```
saw303, line 36: since you are not creating something new with `setCookieHeader` there is no need to create a new GString from `setCookieHeader`.

```suggestion
        Cookie cookie = CookieUtil.getCookieFromString(setCookieHeader)
```

saw303 (commented):
Nice PR.

jameskleeh:
@jantosi We appreciate the effort put into this PR! I hope to see future contributions from you! Unfortunately there are already capabilities in Netty to accomplish this work so duplicating that effort would add additional maintenance for us going forward that isn't necessary. As of now I'm not aware of any use case for the SimpleHttpResponse that would require this code so I'm going to close this PR in favor of https://github.com/micronaut-projects/micronaut-core/pull/4000. In the future I encourage you to discuss solutions prior to putting in the work to ensure you have all of the necessary information.

jantosi:
Very disappointing experience for a `type:good-first-issue` ticket.

PR #3886, opened 2020-08-10
Add details about Netty's memory management config

adamkobor:
If you want to set the heap's maximum size in a native image, you could easily get an `OutOfMemoryError`, coming from Netty's `PooledByteBufAllocator`. The issue can be tackled by configuring Netty through the image's entrypoint, and it worth to mention in the FAQ section of the docs IMO.


graemerocher:
Thanks for the contribution!

PR #3892, opened 2020-08-11
collect most specific type

Danny02:
This is a fix for a regression introduced by #3200 . Before this merge, abstract methods where ignored in the introspection.

Introspection now wrongfully deduces `Object` as return type for classes which implements a generic interface with a property with a generic return type.

i.e.
```java
interface BaseEntity<ID> {
  ID getId();
}

@Entity
class Person implements BaseEntity<Long> {
...
}
```


jameskleeh:
Closing this in favor of https://github.com/micronaut-projects/micronaut-core/pull/3965. Some of the changes were not necessary and the current change worked around a larger issue

PR #3902, opened 2020-08-14
explicitly setting random port in application-test.yml

HoffiMuc:
if you have a micronaut.server.port in src/**main**/resources/application.yml and use a classpath scanning build tool like gradle, only the first test (starting up the embedded server) will succeed, all consecutive tests will fail with `Unable to start Micronaut server on port: xxxx` and `Address already in use`


jameskleeh:
Thanks!

PR #3909, opened 2020-08-17
Improve loggers endpoint docs

ilopmar:


alvarosanchez (approved):

dcalap:
Thanks for the fix guys! It was a pain fro troubleshooting to re-deploy in k8s every time modifying the logger configuration in `application.yml` instead of using the curl directly in the containers

PR #3920, opened 2020-08-19
Keep a collection of ServiceInstace to send a Heartbeat for all of them

ilopmar:
Fixes https://github.com/micronaut-projects/micronaut-grpc/issues/33

The problem is that `HeartbeatTask` only allows to store one `ServiceInstance` so when the application uses Netty and gRPC, the second one that is registered wins (usually Netty), and overrides the first one. During bootstrap both ports are registered in Consul but after the first heartbeat is sent only one is sent and the other one is displayed as fail in Consul.

@graemerocher I'm not really sure if my fix using a `Set` backed by a `ConcurrentHashMap` is the way to go or not. I don't know the decission about using an `AtomicReference` in the first place but I think it may not be necessary because we only called the `get` method and not use any of `getAndSet`, `getAndUpdate`,...; also the `set` method in only called once per `ServiceInstance`. If you think I should use another data structure, please let me know.

Regarding the second part of the issue. At this moment, both services (http and grpc) are registered in Consul using the same name. The user in the issue proposed to add the sufix `-grpc`, so Micronaut will register two different services: `hello:1234` for http and `hello-grpc:4321` for grpc. 
I think it kind of makes sense to use different names because even if the application is the same, they are not instances of the same thing because one uses http and the other uses grpc. I'm not sure if adding a suffix will break something or not, and it will probably be a breaking change. I did a quick test modifying this line https://github.com/micronaut-projects/micronaut-grpc/blob/2a22dd3253a3c9b8c6f423e4b5f1baabea556ba3/grpc-server-runtime/src/main/java/io/micronaut/grpc/server/GrpcServerInstance.java#L58 to `this.id = id + "-grpc";` and it works.
But as I said, not 100% sure about this. Can you please give me some advice on this?

image

### DIFF
--- a/runtime/src/main/java/io/micronaut/health/HeartbeatTask.java
+++ b/runtime/src/main/java/io/micronaut/health/HeartbeatTask.java
@@ -22,7 +22,8 @@ import io.micronaut.discovery.event.ServiceReadyEvent;
 import io.micronaut.scheduling.annotation.Scheduled;
 
 import javax.inject.Singleton;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A scheduled task that sends a periodic heartbeat whilst the server is active.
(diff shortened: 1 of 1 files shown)


graemerocher:
@ilopmar probably it is ok to use `-grpc` as an instance id however it will need to be made configurable, probably via some config props in the grpc project. Otherwise this PR seems ok to me

graemerocher (approved):

ilopmar:
@graemerocher Thanks! I'll merge this and work on the other PR :-)

PR #3921, opened 2020-08-19
Add API to make it easier to write method interceptors and support Kotlin suspend functions

dstepanov:
Added util classes to work with Kotlin's suspend functions in `MethodInterceptor`s and implemented to use it with `@Retryable`.

It should be possible to implement other interceptors like https://github.com/micronaut-projects/micronaut-cache/issues/183.

Personally I'm not a Kotlin user just found it interesting.

### DIFF
--- a/aop/build.gradle
+++ b/aop/build.gradle
@@ -1,7 +1,16 @@
+plugins {
+    id "org.jetbrains.kotlin.jvm" version "1.3.72"
+}
+
 ext {
     shadowJarEnabled = true
 }
 dependencies {
     api project(':inject')
     api project(':core')
+    compileOnly "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.8"
+}
+
+compileKotlin {
+    kotlinOptions.jvmTarget = "1.8"
 }
--- a/aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
+++ b/aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.aop;
+
+import io.micronaut.aop.internal.intercepted.InterceptedMethodUtil;
+import io.micronaut.context.exceptions.ConfigurationException;
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.type.Argument;
+import org.reactivestreams.Publisher;
+
+import java.util.concurrent.CompletionStage;
+
+/**
+ * The intercept method supporting intercepting different reactive invocations.
+ *
+ * @author Denis Stepanov
+ * @since 2.1.0
+ */
+@Experimental
+public interface InterceptedMethod {
+
+    /**
+     * Creates a new instance of intercept method supporting intercepting different reactive invocations.
+     *
+     * @param context The {@link MethodInvocationContext}
+     * @return The {@link InterceptedMethod}
+     */
+    static InterceptedMethod of(MethodInvocationContext<?, ?> context) {
+        return InterceptedMethodUtil.of(context);
+    }
+
+    /**
+     * Returns result type of the method.
+     *
+     * @return The {@link ResultType}
+     */
+    ResultType resultType();
+
+    /**
+     * Returns result type value.
+     *
+     * @return The return type value.
+     */
+    Argument<?> returnTypeValue();
+
+    /**
+     * Proceeds with invocation of {@link InvocationContext#proceed()} and converts result to appropriate type.
+     *
+     * @return The intercepted result
+     */
+    Object interceptResult();
+
+
+    /**
+     * Proceeds with invocation of {@link InvocationContext#proceed(Interceptor)} and converts result to appropriate type.
+     *
+     * @param from The interceptor to start from
+     * @return The intercepted result
+     */
+    Object interceptResult(Interceptor<?, ?> from);
+
+    /**
+     * Proceeds with invocation of {@link InvocationContext#proceed()} and converts result to {@link CompletionStage}.
+     *
+     * @return The intercepted result
+     */
+    default CompletionStage<?> interceptResultAsCompletionStage() {
+        if (resultType() != ResultType.COMPLETION_STAGE) {
+            throw new ConfigurationException("Cannot return `CompletionStage` result from '" + resultType() + "' interceptor");
+        }
+        return (CompletionStage<?>) interceptResult();
+    }
+
+    /**
+     * Proceeds with invocation of {@link InvocationContext#proceed()} and converts result to {@link Publisher}.
+     *
+     * @return The intercepted result
+     */
+    default Publisher<?> interceptResultAsPublisher() {
+        if (resultType() != ResultType.PUBLISHER) {
+            throw new ConfigurationException("Cannot return `Publisher` result from '" + resultType() + "' interceptor");
+        }
+        return (Publisher<?>) interceptResult();
+    }
+
+    /**
+     * Proceeds with invocation of {@link InvocationContext#proceed(Interceptor)} and converts result to {@link CompletionStage}.
+     *
+     * @param from The interceptor to start from
+     * @return The intercepted result
+     */
+    default CompletionStage<?> interceptResultAsCompletionStage(Interceptor<?, ?> from) {
+        if (resultType() != ResultType.COMPLETION_STAGE) {
+            throw new ConfigurationException("Cannot return `CompletionStage` result from '" + resultType() + "' interceptor");
+        }
+        return (CompletionStage<?>) interceptResult(from);
+    }
+
+    /**
+     * Proceeds with invocation of {@link InvocationContext#proceed(Interceptor)} and converts result to {@link Publisher}.
+     *
+     * @param from The interceptor to start from
+     * @return The intercepted result
+     */
+    default Publisher<?> interceptResultAsPublisher(Interceptor<?, ?> from) {
+        if (resultType() != ResultType.PUBLISHER) {
+            throw new ConfigurationException("Cannot return `Publisher` result from '" + resultType() + "' interceptor");
+        }
+        return (Publisher<?>) interceptResult(from);
+    }
+
+
+    /**
+     * Handle the value that should be the result of the invocation.
+     *
+     * @param result The result of the invocation
+     * @return The result of the invocation being returned from the interceptor
+     */
+    Object handleResult(Object result);
+
+    /**
+     * Handle the exception that should be thrown out of the invocation.
+     *
+     * @param exception The exception
+     * @param <E> Sneaky throws helper
+     * @return The result of the invocation being returned from the interceptor
+     * @throws E The exception
+     */
+    <E extends Throwable> Object handleException(Exception exception) throws E;
+
+    /**
+     * Indicated unsupported return type.
+     *
+     * @return The result of the invocation being returned from the interceptor
+     */
+    default Object unsupported() {
+        throw new ConfigurationException("Cannot intercept method invocation, missing '" + resultType() + "' interceptor configured");
+    }
+
+    /**
+     * Possible result types.
+     */
+    enum ResultType {
+        COMPLETION_STAGE, PUBLISHER, SYNCHRONOUS
+    }
+
+}
--- a/aop/src/main/java/io/micronaut/aop/internal/intercepted/CompletionStageInterceptedMethod.java
+++ b/aop/src/main/java/io/micronaut/aop/internal/intercepted/CompletionStageInterceptedMethod.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.aop.internal.intercepted;
+
+import io.micronaut.aop.InterceptedMethod;
+import io.micronaut.aop.Interceptor;
+import io.micronaut.aop.MethodInvocationContext;
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.convert.ConversionService;
+import io.micronaut.core.type.Argument;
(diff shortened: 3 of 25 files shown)


aop/src/main/java/io/micronaut/aop/util/KotlinInvocationContextUtils.java
```diff
@@ -0,0 +1,152 @@
...
+     * Checks if the method invocation is a Kotlin coroutine.
+     *
+     * @param context {@link MethodInvocationContext}
+     * @return true if Kotlin coroutine
+     */
+    public static boolean isKotlinCoroutine(MethodInvocationContext<Object, Object> context) {
+        if (!KotlinUtils.KOTLIN_COROUTINES_SUPPORTED || !context.getExecutableMethod().isSuspend()) {
+            return false;
+        }
+        Collection<MutableArgumentValue<?>> arguments = context.getParameters().values();
```
graemerocher: Think it would be more efficient to use `getParameterValues()` and check the array since this will not build the `MutableArgumentValue` objects and there is no requirement here to use the `MutableArgumentValue` interface

aop/src/main/java/io/micronaut/aop/util/KotlinInvocationContextUtils.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Handle Kotlin coroutine invocation represented as {@link CompletableFuture}s.
+     *
+     * @param context  {@link MethodInvocationContext}
+     * @param consumer helper
+     * @return invocation return value
+     */
+    public static Object handleKotlinCoroutine(MethodInvocationContext<Object, Object> context, Consumer<KotlinCoroutineInvocation> consumer) {
+        Collection<MutableArgumentValue<?>> arguments = context.getParameters().values();
```
graemerocher: Same thing here I thin you can use `getParameterValues()` as well

graemerocher (changes requested):
Nice work! Definitely a 2.1 feature this

aop/src/main/java/io/micronaut/aop/util/KotlinInvocationContextUtils.java
dstepanov: I need `MutableArgumentValue` to change the value `lastArgument.setValue(completableFutureContinuation);`

graemerocher: even then will be more efficient to the array element directly since you are not using anything the `Argument` interface provides (like the annotation metadata or the argument name)

dstepanov:
I think there should be a better way to do interception instead of the current code that is almost the same everywhere: "check if publisher", "check if future" and now this will require adding "check if kotlin suspend".

graemerocher:
@dstepanov yes we need some generalized abstraction for handling all these cases I agree

dstepanov:
Extracted a helper class, looks much better now.

graemerocher:
Seem to be some checkstyle violations

graemerocher:
The only doubt I have about this is I am unsure if it is going into `micronaut-aop` that we should couple it to RxJava 2

In general I think we need to move rxjava 2 to a module like the other frameworks and add a dependency on the modules that use it to RxJava 2 to make sure we have a road to move to rxjava 3 in the future

dstepanov:
I have removed RxJava 2 dependency, only one use case with publishers/futures that is not changed is Http client.

dstepanov:
@graemerocher I have made some refactoring to introduce reactive introduce call API, the core should have all interceptors refactored to use new API. Let me know what do you think.

graemerocher:
@dstepanov this is really awesome work and will definitely help avoid duplication! There are a few things I am unsure about:

* I am not sure if `reactive()` is the correct name here since the helper handles more than just reactive but also completion stage and synchronous. Maybe `interceptBuilder()` or something would be more appropriate. I would also rename `ReactiveMethodInvocationInterceptor` to `InterceptBuilder`  
* I am confused by why retry handling exists for completion stage in this helper https://github.com/micronaut-projects/micronaut-core/pull/3921/files#diff-35c23af827950f588d03002d3b502d97R62 it seems like something that should not be part of the core API?

dstepanov:
@graemerocher To retry the method call the interceptor needs to be supplied `context.proceed(interceptor)`, not sure how to represent it in the API of the reactive interceptor. Maybe to have just one `Supplier` which would call `context.proceed()` for the first invocation and `context.proceed(interceptor)` next. WDYT?

graemerocher:
ah ok that confused me, maybe it should be called `restart` or `reset` instead as you are restarting from the beginning of the interceptor chain?

dstepanov:
Naming is not the best, for the users of the API it might not be very clear what is the interceptor chain etc. that's why `retry` is most clear but it does hide some impl. details.

dstepanov:
@graemerocher How would you name `InterceptCompletionStageSupplierWithRetry`?

graemerocher:
@dstepanov maybe `InterceptCompletionStageRestartSupplier`?

aop/src/main/java/io/micronaut/aop/util/ReactiveMethodInterceptorHelper.java
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * @author Denis stepanov
+ * @since 2.1.0
+ *
+ * @param <T> The declaring type
+ * @param <R> The result of the method call
+ */
+@Internal
+@Experimental
+public class ReactiveMethodInterceptorHelper<T, R> implements InterceptBuilder<T, R> {
```
graemerocher: Does this need to public? Probably needs renaming too

dstepanov: Renamed it also, do you want to rebase?

graemerocher: Makes sense to rebase yes

dstepanov:
Refactored `ValidatingInterceptor` to use the builder

dstepanov:
Looks like I also missed `ClientWebSocketInterceptor`

dstepanov:
I had to add one more method 'provideCompletableFuture` to the builder for `ClientWebSocketInterceptor` to support returning `java.util.concurrent.Future`

graemerocher:
Cool. I'm just wondering, the current logic just falls through if any hook is `null`, maybe there should be a mode that either throws an exception if you don't provide all the hooks otherwise could be easy to miss one.

dstepanov:
Maybe, I have noticed `TraceInterceptor` doesn't intercept `CompletionStage` for `ContinueSpan`

graemerocher:
@dstepanov yeah there are probably other cases of missing features like this. Maybe a strict mode where it validates the hooks are provided?

dstepanov:
Yeah, added strict mode and redid `Future` integration, now it should be possible to use it everywhere, but if it's some strange custom type it should fail because you cannot easily convert it to CompletableFuture without pulling.

graemerocher:
One last thing I am wondering, which is from a performance perspective since the new implementation will be slightly less efficient from a performance perspective. Currently if you look at the following logic:

```java
        return context.interceptBuilder()
                .interceptPublisherSupplier((valueType, publisherSupplier) -> invokeFn(body, functionName, functionDefinition, valueType))
                .interceptCompletionStageSupplier((valueType, completionStageSupplier) -> toCompletableFuture(
                        invokeFn(body, functionName, functionDefinition, valueType)
                ))
                .interceptSynchronous(() -> {
                    FunctionDefinition def = functionDefinition.blockingFirst();
                    FunctionInvoker functionInvoker = functionInvokerChooser.choose(def).orElseThrow(() -> new FunctionNotFoundException(def.getName()));
                    return functionInvoker.invoke(def, body, context.getReturnType().asArgument());
                }).intercept(this);
```

What this means is that we have for every method execution 3 method calls and 3 lambdas being constructed plus a call to `intercept(this)`.

It could be more efficient if instead it worked like this:

```java
        return context.interceptBuilder(this, (builder)-> {
                builder.interceptPublisherSupplier((valueType, publisherSupplier) -> invokeFn(body, functionName, functionDefinition, valueType))
                .interceptCompletionStageSupplier((valueType, completionStageSupplier) -> toCompletableFuture(
                        invokeFn(body, functionName, functionDefinition, valueType)
                ))
                .interceptSynchronous(() -> {
                    FunctionDefinition def = functionDefinition.blockingFirst();
                    FunctionInvoker functionInvoker = functionInvokerChooser.choose(def).orElseThrow(() -> new FunctionNotFoundException(def.getName()));
                    return functionInvoker.invoke(def, body, context.getReturnType().asArgument());
                }
        });
```

What would happen here is that if `interceptPublisherSupplier` matches the context we throw a flow control exception such as `ExecuteInterceptorException` that holds a reference to the lambda that calls `invokeFn` and overrides `fillStackTrace` such that the exception costs nothing. The `interceptBuilder` would catch this exception and return the `interceptPublisherSupplier` lambda to continue execution.

This would mean for the most common case (the first one), only a single method call and a single lambda is constructed. What do you think?

dstepanov:
@graemerocher Cannot imagine the code with `ExecuteInterceptorException`. Can you maybe provide an example?

graemerocher:
Would look something like this:

```java
Object interceptBuilder(MethodInvocationContext context, Consumer<InterceptBuilder> builderConsumer) {
      try {
         builderConsumer.accept(new InterceptBuilder(context))
          // no interception occurred, proceed as normal
         return context.proceed();
      } catch (ExecutionInterceptorException e) {
          return e.getNewResult();
      } 
}
```

`ExecutionInterceptorException` would look like:

```java
// flow control exception 
class ExecutionInterceptorException<T> extends RuntimeException {
       private final T newResult;
       ExecutionInterceptorException(T newResult) { this.newResult = newResult; }
       
        // optimize fill stack trace to avoid performance cost
       @Override public Throwable fillInStackTrace() { return this; }
}
```

The logic that is currently in the `intercept` method for each interception type would be moved to the actual method something like:

```java
@Override
public InterceptBuilder<T, R> interceptPublisherSupplier(InterceptPublisherSupplier interceptPublisherSupplier) {
       if (Publisher.class.isAssignableFrom(returnType)) {
             Argument valueType = returnType.asArgument().getFirstTypeVariable().orElse(Argument.OBJECT_ARGUMENT);
             Publisher publisherResult = interceptPublisherSupplier.intercept(valueType, () -> convertToPublisher(context.proceed()));
             Object finalPublisher = convertPublisherToReactiveType(returnType, publisherResult);
             Publisher newResult = convertFinalPublisher(finalPublisher);
             throw new ExecutionInterceptorException(newResult);
       }
       return this;
}
```

So if the `interceptPublisherSupplier` method matches it would handle the logic and throw `ExecutionInterceptorException` with the new result to return otherwise it will return `this` allowing the next registered interceptor the opportunity to handle it.

This would avoid having multiple method invocations and multiple lambda instantiations for the most commonly supported interception type which is different for each interceptor. For example for `@Transactional` synchronous is most common.

dstepanov:
@graemerocher How would you hande other methods like `strictMode` and `convertPublisherToReactiveType`?

graemerocher:
yeah I guess it is in conflict with strict mode 🤔

Maybe the overhead is not enough to justify the optimization

dstepanov:
The only other solution would be to go back to anonymous impl. of interface with default methods returning null except `interceptSynchronous`.

graemerocher:
I think we should probably leave it like it is for now

graemerocher:
@dstepanov one final change. I think we should not include this as a method on `MethodInvocationContext` and instead have a static factory method on the interface. I can image wanting to create subclasses of this to support other reactive frameworks and maybe the Flow API in Java 9

dstepanov:
@graemerocher something like `static InterceptBuilder<T, R> of(MethodInvocationContext<T, R> context)` on `InterceptBuilder`?

graemerocher:
yes something like that

dstepanov:
@graemerocher Renamed

dstepanov:
I was thinking maybe it would be better to replace null checks with instanceof, impl. would need to pass an instance that implements 2-3 interfaces.

```
     return InterceptHelper.intercept(new ReactiveIntercept() {
            @Override
            public CompletionStage<Object> intercept(Argument<Object> valueType, Supplier<CompletionStage<Object>> resultSupplier, Supplier<CompletionStage<Object>> retryResultSupplier) {
                ...
            }

            @Override
            public Publisher<Object> intercept(Argument<Object> valueType, Supplier<Publisher<Object>> publisherSupplier) {
                ...
            }

            @Override
            public Object intercept() {
                ...
            }
        }, this);
    }
    
    private abstract class ReactiveRetryIntercept implements InterceptPublisherSupplier, InterceptCompletionStageRestartSupplier, InterceptSynchronous {
    }

```

The API is a bit strange because Java doesn't allow it to implement multiple interfaces for anonymous classes, but the performance should be better.

Also I would add `MethodInvocationContext<Object, Object> context` to every new intercept interface to look it more like `MethodInterceptor<Object, Object>`.

WDYT?

dstepanov:
@graemerocher Should I try to refactor it?

graemerocher:
@dstepanov I wonder if it is worth providing a worse API in the name of performance.

dstepanov:
Probably not

jameskleeh:
@dstepanov I'd like to get this merged in pretty soon so I can look into coroutine support for the client. Overall it looks to be a big improvement over what we have currently. I am also concerned about the performance implication and I'd like to propose an alternative.

What if instead of having a builder style, an object was returned `InterceptedMethod.of(context)` which would then have methods on it to return the type (an enum), among others. The calling code could then have a switch block to handle the different types with a fallback default of `interceptedMethod.unsupported()` or something of the sort. This wouldn't require the multiple lambdas to be created for every method call.

I haven't dug into this logic very deeply so perhaps this wouldn't work for some reason, but it seems like it wouldn't complicate the usage of the API very much.

dstepanov:
@jameskleeh That's one of the ways to do it with the disadvantage of moving the switch outside to every interceptor and removing the strick check.

It would be possible to completely remove the builder:
```
switch(InterceptMethodUtils.invocationType(context)) { 
    case COMPLETION_STAGE:
        return InterceptMethodUtils.interceptCompletionStage(context, (..) -> {});
    case PUBLISHER:
         return ..
    case SYNCHRONOUS:
        return ..
}
```

This PR also adds coroutine support for the http client:

https://github.com/micronaut-projects/micronaut-core/blob/880b2f4a5ecd54b8762715b11b46c195a28d7bdf/test-suite-kotlin/src/test/kotlin/io/micronaut/docs/server/suspend/SuspendClient.kt#L18-L47

jameskleeh:
@dstepanov I see now that coroutines are handled as part of the completion stage handlers. Awesome!

@graemerocher Can you give your thoughts? Perhaps we should have an idea of the performance implications before moving to a less friendly API.

graemerocher:
@jameskleeh only way to know the performance implications is benchmarking, however in general it is safe to assume a switch style would be much faster and more efficient in terms of memory consumption since there is less object creation to perform method dispatch.

dstepanov:
Should I make some changes?

jameskleeh:
@dstepanov You can, however I think the code you showed above wouldn't result in increased performance:

```
switch(InterceptMethodUtils.invocationType(context)) { 
    case COMPLETION_STAGE:
        return InterceptMethodUtils.interceptCompletionStage(context, (..) -> {}); //still using a lambda here
```

I think it would need to be more like

```
InterceptedMethod interceptedMethod = InterceptedMethod.of(context);
switch (interceptedMethod.returnType()) {
    case COMPLETION_STAGE:
        CompletableFuture f = toCompletableFuture(invokeFn(body, functionName, functionDefinition, valueType))
        return interceptedMethod.handleResult(f);
        break;
    default: //strict mode, up to the caller if the call this
        interceptedMethod.unsupported(); //throws exception
```

If that seems doable and you can get to it within the next week or so, that would be fantastic. If not, I will make sure it gets taken care of before 2.1

dstepanov:
@jameskleeh @graemerocher Please review

graemerocher:
That looks great to me. @jameskleeh ?

jameskleeh:
@graemerocher Looks good to me. Really great contribution @dstepanov!

dstepanov:
@jameskleeh Thanks! You had a good suggestion for the API

PR #3930, opened 2020-08-19
Oracle updates

aalmiray:


sdelamo (approved):

ilopmar:
This is a breaking change and, at least should be documented as such. Any application using the old coordinates would fail, like the GraalVM tests just failed: https://gitlab.com/micronaut-projects/micronaut-graal-tests/-/jobs/695117552#L56

graemerocher:
@ilopmar in that case we would need to maintain in the BOM the old coordinates and point to the older version of the driver

graemerocher:
ie version `19.3.0.0`

ilopmar:
I think we should definitely do that, specially for a patch release. We did that when the GraalVM team modified the coordinates, updated to the new one but keep the old for compatibility (until 2.0 that we removed it)

I'll submit a PR with the old version.

aalmiray:
Would it be possible to keep both old and new coordinates in the BOM?

ilopmar:
@aalmiray https://github.com/micronaut-projects/micronaut-core/pull/3934

PR #3934, opened 2020-08-20
Add old Oracle driver coordinates to the BOM

ilopmar:
Follow up on #3930

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -507,7 +507,7 @@ ext {
                 group  : 'com.oracle.database.jdbc',
                 name   : 'ojdbc8'
             ],
-            // Deprecated: Remove in the next major version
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -507,6 +507,12 @@ ext {
                 group  : 'com.oracle.database.jdbc',
                 name   : 'ojdbc8'
             ],
+            // Deprecated: Remove in the next major version
```
alvarosanchez, line 510:
```suggestion
            // TODO Deprecated: Remove in the next major version
```

This way at least appears in Intellij's TODO window.

ilopmar: Done. Good catch!

alvarosanchez (approved):

PR #4009, opened 2020-09-02
Fix for method argument annotation visitors.

### LINKED ISSUE #4008: Method argument annotations from processor not working
### Task List

### Steps to Reproduce

1. See https://github.com/tomas-langer/reproduce-micronaut-visitor

### Expected Behaviour

When I have a `TypeElementVisitor` implemented that adds an annotation to a method parameter, that annotation should be visible at runtime in an executable method

The reproducer main method should successfully execute.

### Actual Behaviour

The annotation is not visible (nor is it in the generated class of the executable method).

Exception from reproducer:
```
Caused by: io.micronaut.core.bind.exceptions.UnsatisfiedArgumentException: Required argument [FirstBean paramInjection] not specified
	at io.micronaut.core.bind.DefaultExecutableBinder.bind(DefaultExecutableBinder.java:96)
	at io.reproducers.reproducer.impl.ExecMethodProcessor.process(ExecMethodProcessor.java:37)
	at io.reproducers.reproducer.impl.ExecMethodProcessor.process(ExecMethodProcessor.java:17)
	at io.micronaut.context.AnnotationProcessorListener.onCreated(AnnotationProcessorListener.java:89)
	... 10 more
```

### Environment Information

- **Operating System**: MacOS 
- **Micronaut Version:** 2.0.1
- **JDK Version:** 11

### Example Application

https://github.com/tomas-langer/reproduce-micronaut-visitor

tomas-langer:
Resolves #4008 

This fix makes sure the updated annotation metadata is stored in the modified cache under the same key it is retrieved later on.

This PR also includes a test verifying the annotation is correctly added.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/AbstractJavaElement.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/AbstractJavaElement.java
@@ -25,6 +25,7 @@ import io.micronaut.inject.ast.ClassElement;
 import io.micronaut.inject.ast.MemberElement;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import io.micronaut.inject.ast.ParameterElement;
 import io.micronaut.inject.ast.PrimitiveElement;
 
 import javax.lang.model.element.Element;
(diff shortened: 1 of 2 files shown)


jameskleeh (approved):
Thanks for the PR!

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/AbstractJavaElement.java
```diff
@@ -76,7 +77,20 @@
                 .newAnnotationBuilder()
                 .annotate(annotationMetadata, av);
 
-        String declaringTypeName = this instanceof MemberElement ? ((MemberElement) this).getOwningType().getName() : getName();
+        String declaringTypeName;
+        if (this instanceof MemberElement) {
+            declaringTypeName = ((MemberElement) this).getOwningType().getName();
+        } else if (this instanceof ParameterElement) {
+            TypeElement typeElement = visitorContext.getModelUtils().classElementFor((Element) this.getNativeType());
```
tomas-langer, line 84: The only part I am not sure about is if this cast is safe, or if there is another way to obtain the declaring type...

graemerocher: Yes it is safe

PR #4026, opened 2020-09-04
Bump micronaut-openapi to 2.0.0

micronaut-build:


ilopmar (approved):

graemerocher:
Seems the group needs updating and for backwards compatibility we should probably retain the old group referring to the previous version

ilopmar:
Done!

ilopmar:
Superseded by #4028

PR #4027, opened 2020-09-04
Add support for a default environment

### LINKED ISSUE #2417: Add ability to provide default environment
In current implementation configuration is taken from `application.yml` by default, so it seems appropriate to use it for local ("dev") environment. 
However, in this case it is necessary to remove the `application.yml` from artifact before publishing it so no "dev" configs are used on testing/production environment. 

A way to solve this issue is to store dev configuration in environment specific file (`application-dev.yml`) and use it if no other environment is provided. In that case it would be necessary to provide a name of default environment(s?) in main class:

```java
Micronaut.build((String[]) null)
               .defaultEnvironments("dev")
               .start();
```

jameskleeh:
Closes #2417


graemerocher:
Just needs an entry in the `whatsNew` doc

graemerocher (approved):

PR #4028, opened 2020-09-04
Bump micronaut-openapi to 2.1.0

micronaut-build:


ilopmar:
Supersedes #4026

ilopmar (approved):

graemerocher:
Also needs updating https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/src/main/java/io/micronaut/gradle/MicronautLibraryPlugin.java#L37

ilopmar:
The gradle plugin wasn't on my radar yet. I need to take a closer look at it

I've submitted this PR https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/18 although we can't merge it until we have a Micronaut 2.1 non-snapshot version that includes the group in the bom.

In any case I think we can merge because the plugin will use the old group until we merge the mentioned PR.

PR #4031, opened 2020-09-04
Add a URI context so binders can participate in the URI creation

jameskleeh:


graemerocher:
I think in general we need to rethink this API a bit because there are other Micronaut clients that are not HTTP that could benefit from binding such as `@KafkaClient` and so on. Is there some way we could refactor this so as not have to duplicate this API for other client implementations? It worries me that is specific just to HTTP clients.

jameskleeh:
@graemerocher Perhaps it could be extracted to a common API, however the only code that would be duplicated currently is the registry. This feature is currently marked as experimental so we can make changes to it in the future.

PR #4058, opened 2020-09-09
Use rootBean as a leafBean for validator (#4030)

### LINKED ISSUE #4030: Micronaut validator is not compatible with Hibernate BeanValidationEventListener
In `2.0.1`.

Micronaut's `DefaultValidator` is not compatible with Hibernate `BeanValidationEventListener`, because produced `ConstraintViolation`-s frequently contains null `leafBean` and [this line](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationEventListener.java#L124) crashes with NPE. Please note, that I don't use Hibernate Validator, only Hibernate with Micronaut validator.

Looking at `DefaultValidator` and spots where it produces `DefaultConstraintViolation`, there are three cases where `leafBean` is `null`, and only one where is not. For example if you define your own validator with `@Validation` annotation to be used directly on entity class:

```
@Validation class MyEntity {}
```

This will crash with NPE from within Hibernate. This scenario is documented on `ConstraintViolation.getLeafBean()`: *Returns the bean instance the constraint is applied on if it is a bean constraint*. So, at least this kind of produced `ConstraintViolation` is wrong, and in this scenario `leafBean` should be set to `rootBean`, but there're two other usages of `DefaultConstraintViolation` constructor where `leafBean` is also `null`, which should be reviewed separately.

Refs:

1. Line which crashes in Hibernate [here](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/cfg/beanvalidation/BeanValidationEventListener.java#L124)
1. What should be set as a `leafBean` in `ConstraintViolation` in different scenarios is described [here](https://docs.oracle.com/javaee/7/api/javax/validation/ConstraintViolation.html#getLeafBean--)
1. Spots where there's a constraint violation created with `null` `leafBean`: [one](https://github.com/micronaut-projects/micronaut-core/blob/2.0.x/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java#L861), [two](https://github.com/micronaut-projects/micronaut-core/blob/2.0.x/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java#L1031), [three](https://github.com/micronaut-projects/micronaut-core/blob/2.0.x/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java#L1084)

l0co:
Changes to fix #4030

### DIFF
--- a/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
+++ b/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
@@ -15,6 +15,8 @@
  */
 package io.micronaut.validation.validator;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import io.micronaut.aop.Intercepted;
 import io.micronaut.context.BeanResolutionContext;
 import io.micronaut.context.ExecutionHandleLocator;
@@ -55,8 +57,6 @@ import io.micronaut.validation.validator.extractors.ValueExtractorRegistry;
 import io.reactivex.Flowable;
 import org.reactivestreams.Publisher;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
-import edu.umd.cs.findbugs.annotations.Nullable;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.validation.*;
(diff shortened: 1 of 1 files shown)


graemerocher:
Could you change the base branch to 2.0.x?

graemerocher:
Would be nice to have a test too 🤔

l0co:
Generally, I don't know how to trigger the other cases than using validation annotation on a root bean. This probably better will be done by `DefaultValidator` author. This fix can be treated as a quick temporary workaround.

PR #4061, opened 2020-09-09
Upgrade to Netty 4.1.52

### LINKED ISSUE #3842: Upgrade netty to 4.1.52.Final
https://netty.io/news/2020/07/09/4-1-51-Final.html mentions some changes to accommodate openssl and GraalVM

ilopmar:
Fixes #3842


graemerocher:
Causing build to hang?

ilopmar:
Yes, it seems so. I'll try to find locally which test hangs...

ilopmar:
@graemerocher fixed! 

This was because of [this change](https://github.com/netty/netty/pull/10407) introduced in Netty that makes necessary to bump `tcnative` to 2.0.33. Is this something we should highlight in the "what's new" section in the docs?

graemerocher:
Probably worth mentioning yes. Nice find!

ilopmar:
Done

PR #4065, opened 2020-09-09
- Changed order of responsePublisher declaration, allows downstream f…

viniciusccarvalho:
Bumped into this issue when trying to create a downstream filter that would rely on the SpanContext from the HttpRequest.

Turns out that because the responsePublisher is being invoked prior the request configuration, downstream filters would never be able to see the request attributes.

Changing the declaration order fixes the issue.


viniciusccarvalho:
Not needed sorry about noise

PR #4067, opened 2020-09-09
Handle both BeanInstantiationException and DependencyInjectionException in error handling. 

### LINKED ISSUE #4045: Controller initialisation loop leading to CPU usage due to missing configuration value
#4020 # Steps to Reproduce

1. Create micronaut application with a Controller and a test stream endpoint
```
 @ExecuteOn(TaskExecutors.IO)
    @Get("test/stream", produces = [MediaType.TEXT_EVENT_STREAM])
    fun testStream(): Publisher<Event<Counter>> {
        logger.debug { "New test stream connected" }
        return Flowable.intervalRange(
            0,
            11,
            5,
            5,
            TimeUnit.SECONDS
        ).map {
            logger.debug { "New test stream event $it" }
            createCounterEvent(it, Counter(it))
        }
    }
```
2. Create bean with a configuration property
```
@Singleton
class ServiceImpl(
    @Value("\${propertyt}")
    val property: String
) : Service {
```
3. Fail to provide property in configuration

### Stracktrace
```
{"time":"2020-09-08T10:56:28.401+01:00","@version":"1","message":"Found matching exception handler for exception [Failed to inject value for parameter [property] of class: com.example.ServiceImpl\n\nMessage: Error resolving property value [${property}]. Property doesn't exist\nPath Taken: new Controller(EventConsumer eventConsumer,[Service service]) --> new ServiceImpl(ServiceStub serviceStub,[String property])]:  Throwable -> Controller#HttpResponse error(HttpRequest request,Throwable e)","logger_name":"io.micronaut.http.server.netty.RoutingInBoundHandler","thread_name":"nioEventLoopGroup-1-2","level":"DEBUG","level_value":10000}
```
### Expected Behaviour
Http call returns a 500 error and stops trying to initialise bean.

### Actual Behaviour
Http call hangs trying to initialise bean in an infinite loop leading to CPU usage.

### Environment Information

- **Operating System**: adoptopenjdk:11-jre
- **Micronaut Version:** 2.0.1
- **JDK Version:** 11

jameskleeh:
Fixes #4045

### DIFF
--- a/inject/src/main/java/io/micronaut/context/exceptions/BeanCreationException.java
+++ b/inject/src/main/java/io/micronaut/context/exceptions/BeanCreationException.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package io.micronaut.context.exceptions;
 
 import io.micronaut.context.BeanResolutionContext;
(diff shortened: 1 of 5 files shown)


inject/src/main/java/io/micronaut/context/exceptions/BeanCreationException.java
```diff
@@ -0,0 +1,104 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.context.exceptions;
+
+import io.micronaut.context.BeanResolutionContext;
+import io.micronaut.inject.BeanType;
+
+import java.util.Optional;
+
+public class BeanCreationException extends BeanContextException {
```
graemerocher: Needs javadoc including `author` and `since` tags

graemerocher (changes requested):
Only one small change

PR #4073, opened 2020-09-10
Trigger exception error route before triggering status error route

dstepanov:
@graemerocher @jameskleeh 
This is for discussion because of the behavior change.

There is a problem when someone wants to have a custom error type and map all the exceptions to it.

In our case we are using global error catch controller methods:
```
    @Error(status = HttpStatus.BAD_REQUEST, global=true)
    public HttpResponse badRequest(HttpRequest request) {
         // Transform to custom error bean
    }

   // catch known exceptions

    @Error(global=true)
    public HttpResponse catchUnknown(HttpRequest request, Throwable e) {
         // Transform unknown
    }
```
This should catch everything we care about.

The problem occurs when `UnsatisfiedRouteException` is thrown, it's always caught by `badRequest` and we are losing information that can help to track the problem. The same problem affects `HttpStatusException` because it can have a body defined.

The solution can be to use `ExceptionHandler` for a global exception but handler cannot be used to catch:

```
   HttpResponse bad() {
       HttpResponse.badRequest()
   }
```

For that we need:

```
    @Error(status = HttpStatus.BAD_REQUEST, global=true)
    public HttpResponse badRequest(HttpRequest request) {
    }
```

The PR changes the priority and the exception error route is triggered first and then the status error route is triggered if needed.
In the example above it will cause `catchUnknown` to receive `UnsatisfiedRouteException` and `HttpStatusException` in a case of `@Status(HttpStatus.BAD_REQUEST)` if `catch(UnsatisfiedRouteException/HttpStatusException)` is not added .

BTW: At the moment it's not possible to catch `NOT_FOUND`, I have added a test to show it.


jameskleeh:
@dstepanov The change makes sense to me, however I don't think we can make this change in Micronaut 2. This has the potential to break a large percentage of applications.

dstepanov:
Makes sense, let's add it to 3.0

PR #4076, opened 2020-09-10
Include jackson in parent.

### LINKED ISSUE #4072: micronaut-parent should import micronaut-bom and also any other supported BOMs
The current arrangement doesn't seem to work like we think it works for Maven.

The BOM imports applied in `micronaut-bom` have no impact whatsoever and conflicting versions of things like Jackson are possible.

According to https://stackoverflow.com/a/40545876/259167 multiple inheritance is not possible for parent POMs so it looks like the only option is to repeat all of the BOM imports present in `micronaut-bom` inside `micronaut-parent` as well unless there are any other ideas.

graemerocher:
This provides a fix for the jackson conflict issue. Once this is merged this code can be removed from the gradle PR: https://github.com/micronaut-projects/micronaut-starter/pull/439/files#diff-aca5dd72ca236ffc337a4144027b7d18R70-R77

- Use invoker beta2
- Bump Jackson version to 2.11.2
- Include jackson in parent. Fixes #4072


graemerocher:
@alvarosanchez this being merged?

graemerocher:
@alvarosanchez @jameskleeh Tried to use exclusions instead but using the exclusions doesn’t seem to work seemingly because the oracle cloud BOM inherits these jackson dependencies from a parent POM.

So if declaring them again in our parent is a no go then the only other option we have is to not import the oracle cloud bom and force users to specify the version when declaring the Oracle Cloud SDK.

We would also have to hard code a version of the Oracle Cloud SDK in launch and make sure to keep updating it somehow.

How to proceed?

alvarosanchez:
If that is the case, I would go ahead with this PR

alvarosanchez (approved):

graemerocher:
@alvarosanchez I am going to try one more thing which is to manually parse the oracle bom and extract only the SDK dependencies

graemerocher:
Worked around this with https://github.com/micronaut-projects/micronaut-oracle-cloud/commit/1dbba5245303ac374e04049d04835e4cfe1e776b

PR #4079, opened 2020-09-10
Preserve byte[] of the body in FullNettyClientHttpResponse to allow more body conversions

dstepanov:
Solving the problem when the response body didn't support returning different body representations after the response has been completed.

All media type codecs are extracting bytes for `ByteBuffer` already, I have moved it a level up and keeping it in `FullNettyClientHttpResponse`.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
@@ -71,6 +71,8 @@ public class FullNettyClientHttpResponse<B> implements HttpResponse<B>, Completa
     private final ByteBufferFactory<ByteBufAllocator, ByteBuf> byteBufferFactory;
     private final B body;
     private boolean complete;
+    private byte[] bodyBytes;
+
 
     /**
      * @param fullHttpResponse       The full Http response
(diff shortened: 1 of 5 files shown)


jameskleeh:
@dstepanov What is the use case? Parsing the body into different pojos based on some data in the response? Storing the bytes will essentially duplicate the memory used to store the body and won't be cleaned up until a GC

dstepanov:
Yes, sometimes error responses are different, depending on the content type.

In our case, we also have logging filter and it would be nice to log the actual response as a string, right now it isn't possible.

The bytes are already there in the memory so there is no extra data, they just going to have a bit longer lifetime.

jameskleeh:
@dstepanov I'm relatively certain that `ByteBufUtil.getBytes(content)` copies the data, thus causing additional memory consumption

dstepanov:
@jameskleeh Correct, this is what `JacksonMediaTypeCodec` doing:
```
public <T> T decode(Argument<T> type, ByteBuffer<?> buffer) throws CodecException {
        try {
            if (CharSequence.class.isAssignableFrom(type.getType())) {
                return (T) buffer.toString(applicationConfiguration.getDefaultCharset());
            } else if (type.hasTypeVariables()) {
                JavaType javaType = constructJavaType(type);
                return getObjectMapper().readValue(buffer.toByteArray(), javaType);
            } else {
                return getObjectMapper().readValue(buffer.toByteArray(), type.getType());
            }
        } catch (IOException e) {
            throw new CodecException("Error decoding stream for type [" + type.getType() + "]: " + e.getMessage(), e);
        }
    }
```
And this is what inside `toByteArray`:
```
    public byte[] toByteArray() {
        return ByteBufUtil.getBytes(delegate);
    }
```
I haven't added additional memory consumption just moved the bytes to be stored inside the response.

jameskleeh:
Thanks for pointing that out @dstepanov. What are your thoughts on this @graemerocher ?

graemerocher:
Seems ok

dstepanov:
The only user-case I can come up with why this can be a bad change is if in the future there would be a way to process `netty buffer -> jackson -> object` without doing the byte array copy.

mliberato:
I think this is going to fix as well a **trick**, **specific** problem of getting body from response from `HttpClientResponseException` when dealing with kotlin coroutines, because of this code in DefaultHttpClient:

```java
	try {
	    emitter.tryOnError(clientError);
	} finally {
	    response.onComplete();
	}
```
The emitter dispatches an error, but because of context switch, `response.onComplete()` happens before exception handling, and `complete` flag inside `FullNettyClientHttpResponse` prevents one from getting the response after that.
I'll try to create a reproducible unit of code with this and report/fix separately, if needed.

PR #4099, opened 2020-09-14
Add option to set Response Status for absent value

### LINKED ISSUE #4098: Enhancement request: Add option to set Response Status for absent value
By default Micronaut returns 404 on Optional.empty() or null value. But sometimes it is desired to change this behaviour, for example one would want to return 204 (No Content) in case of no value. 

There is already `@Status` annotation that handles response code when body is present or when method is void. It would be nice to be able to modify also response code when body is null or Optional.empy or Maybe.empty().

asodja:
Closes #4098.

`@Status` can now be used also for null, Optional.empty() or Maybe.empty(), e.g.:
`@Status(absentValue =  HttpStatus.NO_CONTENT)` which sets response code to 204 if some absent value is returned.

What do you think?

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/docs/httpstatus/HttpStatusSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/docs/httpstatus/HttpStatusSpec.groovy
@@ -101,4 +101,39 @@ class HttpStatusSpec extends Specification {
         e.message == "success"
         e.status == HttpStatus.NOT_FOUND
     }
+
+    @Unroll("#description")
+    void "verify default HTTP status when absent value"() {
+        when:
+        client.toBlocking().exchange(HttpRequest.GET(uri), String)
+
+        then:
+        HttpClientResponseException e = thrown()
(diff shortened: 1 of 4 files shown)


graemerocher:
I see the use case. The main issue is currently the `value()` of the `Status` annotation is required, and we would have to make it optional and then ensure we handle the cases where `value()` is not specified. Thoughts @jameskleeh ?

jameskleeh:
I personally see this as a bit of feature creep. If you want to return a different status based on conditions you can just return a response. In general I'd like to avoid multiple ways of doing the same thing, even if the existing way is slightly less convenient.

asodja:
I see this as a nice complement to current `@Status` annotation. With this change both cases (when body is returned or when body is null or empty) can be handled in the same way, and not in two different ways.

So when somebody asks, how to change a status code, you can say: just use `@Status`. Right now it is like: use `@Status` except if there is null or Optional.empty body returned, then you will have to use a response. :)

If you don't share the same view then you can close the pull request, no problem.

PR #4110, opened 2020-09-16
Jdk 15

ilopmar:
Updates the support for JDK15 that was released yesterday (Sep 15th 2020)

I've modified directly the workflow file changing JDK14 with 15 to make sure it works on CI. If everything works and we merge this I will update the workflow file in the template so all projects are switched to 15.


ilopmar:
The build passes with JDK15. Converting the PR from Draft to Ready to review.

ilopmar:
@graemerocher JDK15 has been merged in all the module. Are you ok with merging this here in core?

Issue #3983, opened 2020-08-27
Server filter unable to modify response headers for reactive controllers


lightoze:
Looks like server filter cannot modify response headers when controller has a signature of `Publisher<MutableHttpResponse<?>>`. As a result, for example HTTP session cannot be created during such controller invocation because no cookie is returned to the client.

I believe this is also causing https://github.com/micronaut-projects/micronaut-security/issues/319

Here is a Gist with simple code to reproduce the problem: https://gist.github.com/lightoze/32f6f43b4a8fd8f3a12c2d288489f07c

Micronaut version: 2.0.1

lightoze:
Here's my temporary workaround:
```java
@Filter("/**")
public class ReactiveResponseFixFilter extends OncePerRequestHttpServerFilter {

    @Override
    public int getOrder() {
        return ServerFilterPhase.RENDERING.before();
    }

    @Override
    protected Publisher<MutableHttpResponse<?>> doFilterOnce(HttpRequest<?> request, ServerFilterChain chain) {
        RouteMatch<?> routeMatch = request.getAttribute(HttpAttributes.ROUTE_MATCH, RouteMatch.class).orElse(null);

        if (routeMatch != null && routeMatch.getReturnType().isReactive() && routeMatch.getReturnType().getFirstTypeVariable()
                .stream().allMatch(arg -> MutableHttpResponse.class.isAssignableFrom(arg.getType()))) {
            return Publishers.map(chain.proceed(request), response -> {
                Object body = response.getBody().orElse(null);
                if (body instanceof Publisher) {
                    Object bodyResult = Flowable.fromPublisher((Publisher<?>) body).blockingSingle();
                    if (bodyResult instanceof MutableHttpResponse) {
                        return (MutableHttpResponse<?>) bodyResult;
                    } else {
                        response.body(bodyResult);
                    }
                }
                return response;
            });
        } else {
            return chain.proceed(request);
        }
    }

}
```

jameskleeh:
@lightoze Can you test this with 2.0.2?

lightoze:
@jameskleeh The bug is still there - `/test` and `/body` return responses without a cookie, although filter method is invoked.

[...]

PR #4117, opened 2020-09-17
test: ResponseChangingFilterSpec  #3983

lightoze:
This is an attempt to add a comprehensive test for filter-controller interaction, more specifically response modification. It uncovers a few more problems in addition to #3983 and represents my understanding of how filters should work. If some of the cases work as desired or aren't supported, it might be worth documenting them.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/interceptor/ResponseChangingFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/interceptor/ResponseChangingFilterSpec.groovy
@@ -15,10 +15,10 @@ import io.micronaut.http.sse.Event
 import io.micronaut.test.annotation.MicronautTest
 import io.reactivex.Flowable
 import io.reactivex.Single
+import jakarta.inject.Inject
 import org.reactivestreams.Publisher
 import spock.lang.Specification
 
-import javax.inject.Inject
 import java.util.concurrent.CompletableFuture
 
 @MicronautTest
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/interceptor/ResponseChangingFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/interceptor/ResponseChangingFilterSpec.groovy
@@ -1,0 +1,357 @@
+package io.micronaut.http.server.netty.interceptor
+
+import io.micronaut.context.annotation.Property
+import io.micronaut.context.annotation.Requires
+import io.micronaut.core.async.publisher.Publishers
+import io.micronaut.http.*
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Filter
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.client.RxStreamingHttpClient
+import io.micronaut.http.client.annotation.Client
+import io.micronaut.http.filter.HttpServerFilter
+import io.micronaut.http.filter.ServerFilterChain
+import io.micronaut.http.sse.Event
+import io.micronaut.test.annotation.MicronautTest
+import io.reactivex.Flowable
+import io.reactivex.Single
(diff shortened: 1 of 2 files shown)


jameskleeh:
@lightoze Can you merge 2.2.x into your branch?

lightoze:
@jameskleeh Rebase done. Most of the failures are due to the fact that `RoutingInBoundHandler` assumes that response processing mode (in terms of publisher/stream handling) should be always as defined for the route, but when `HttpServerFilter` replaces or modifies the response this cannot be true. Imagine some streaming response controller being intercepted by authentication filter and returning static login page instead, or exception handling filter kicking in and responding with JSON body instead of initially expected byte chunk stream. More extreme case would be if static controller response is being replaced with a stream by a filter.

`SessionCreationSpec`, on the other hand, is currently processed so that `RoutingInBoundHandler.buildResultEmitter` returns Flowable(MutableHttpResponse(body=Flowable(HttpResponse.ok()))) which should not happen, and it breaks things.

lightoze:
I'm very glad to see this is being taken into account for 3.0. I fixed compilation error because of moved `@Inject` class so the test is usable again.

I have also marked `Publisher<HttpResponse> reactive()` controller method as a single result and it improved test outcome. It is quite counter-intuitive, though, that you have to do that. `Publisher<HttpResponse>` seem to always imply a single result, and even if you attempt to return multiple responses only the first one is used.

All the issues seem to arise from how the response publisher processing mode is being derived from the controller method, even when the response is being entirely replaced by a filter. With 3.0, have you considered moving these controls to `MutableHttpResponse`, with defaults deriving from controller signature before the filter chain is invoked? Basically, copy all necessary info into the response object so that route/controller metadata is not used at later response encoding anymore.

jameskleeh:
> Publisher<HttpResponse> seem to always imply a single result

Not for HTTP/2

> have you considered moving these controls to MutableHttpResponse, with defaults deriving from controller signature before the filter chain is invoked? Basically, copy all necessary info into the response object so that route/controller metadata is not used at later response encoding anymore.

Can you give an example of what info you're referring to?

lightoze:
> Not for HTTP/2

Good point )

> Can you give an example of what info you're referring to?

Prime example is to create a "single vs streaming result" attribute for `MutableHttpResponse` to replace current "isSingle" logic.

The current processing chain looks like the following:
- Call the controller method and wrap result into `MutableHttpResponse` (`createExecuteRoutePublisher`, `emitRouteResponse`
methods)
- Apply filter chain (`filterPublisher` method)
- If body is reactive type, do some processing (`executeRoute` method):
  - When route is marked as "single", replaces response body with such single value
  - As a special case, when body is `HttpResponse`, replaces the whole response discarding outer response object entirely. This is the reason why e.g. session creation does not work on `Publisher<HitpResponse> reactive()` method without `single = true` - cookie headers are being added to outer response object and then discarded.
  - For streaming (not "single") responses it serializes each chunk to `HttpContent` object.
- Resulting response is sent to the client (`handleRouteMatch`, `encodeHttpResponse` methods)

What I was suggesting is to:
- Call the controller method and wrap result into `MutableHttpResponse`. Handle `Publisher<HttpResponse>` case here, so it is
not possible to end up with `MutableHttpResponse<Publisher<HttpResponse>>`.
- Apply route/controller metadata to the response object, unless it is already explicitly set. If body is reactive type, it will be
marked as "single" if necessary, response media type will be applied, etc.
  `MutableHttpResponse` API can be expanded with methods like `bodySingle(boolean)` to set single body marker and
`bodySingle(Publisher<?>)` to set reactive body alongside with the marker. I would personally prefer to assume single body by
default in this API and call `bodyStream(Publisher<?>)` when I want the opposite. Having streaming response content type would also default it to expect multiple results from a publisher, of course. But changing defaults may be too much of a breaking change, so whatever.
- Apply fitter chain
- Do the rest of processing as before, but instead of `RouteMatch` object all necessary information should already be in
`MutableHttpResponse` object

PR #4134, opened 2020-09-18
Fix dependency resolution happening at configuration time

melix:
The BOM check tasks were not using `doLast`, which means that
everything was done at configuration time, which is particularly
bad in this case because it resolves _all_ configurations, and
also requires _all_ artifacts to be downloaded, for every build.

This is a major performance issue, which this commit just fixes
by wrapping everything into an appropriate `doLast`.

Ideally this code should be moved in a `buildSrc` plugin, in order
to avoid bloating the build script, and which would have made
this evident (because the code would have lived in a `@TaskAction`).

I noticed the problem because after updating a local branch and starting a build, it spent more than 10 minutes **downloading jars** during configuration time, which should _never_ happen.

Before, just running `tasks`:
   - dependency resolution: https://scans.gradle.com/s/fqguly56fsz6c/performance/suggestions?open=WyJwYXJhbGxlbCIsInJlc29sdXRpb25EdXJpbmdDb25maWd1cmF0aW9uIiwib3V0T2ZEYXRlR3JhZGxlVmVyc2lvbkNoZWNrIiwibm9uQ2FjaGVhYmxlVGFza091dHB1dCJd#resolution-during-configuration
   - downloads: https://scans.gradle.com/s/fqguly56fsz6c/performance/networkActivity

After:

https://scans.gradle.com/s/d663hiitaorqo/performance/resolution

I didn't optimize anything, it's just a quickfix. In particular, it also looks like you're resolving way too many configurations.

### DIFF
--- a/bom-check/build.gradle
+++ b/bom-check/build.gradle
@@ -83,42 +83,46 @@ publishing {
 
 publish.enabled = false
 
-task validateMicronautBom() {
-    configurations.each { cfg ->
-        if (cfg.isCanBeResolved()) {
-            cfg.resolve()
-        }
+task validateMicronautBom {
+    doLast {
+		configurations.each { cfg ->
+		    if (cfg.isCanBeResolved()) {
+		        cfg.resolve()
+		    }
+		}
     }
 }
 
 task validateThirdPartyBoms(dependsOn:['validateMicronautBom']) {
-    for (dep in bomVersions) {
-        def info = dep.value
-        if (info.group.startsWith("io.micronaut")) continue
+    doLast {
+		for (dep in bomVersions) {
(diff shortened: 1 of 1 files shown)


jameskleeh:
Thanks @melix !

jameskleeh:
@alvarosanchez This doesn't need changed in the template as well right?

alvarosanchez:
No, the BOM checker thing exists only in core.

That being said, @melix, any suggestions are welcome. The idea behind it was to check Micronaut BOM so that we could make sure that it references existing versions. We've had typos and mistakes like that in the past that we are trying to avoid with that thing.

PR #4151, opened 2020-09-23
Fix missing interface methods when @Introduction is mapped

dstepanov:


graemerocher:
Thanks for the contribution!

Issue #4141, opened 2020-09-21
Reactive retry doesn't work as expected


Im-dex:
### Steps to Reproduce

Run test

### Expected Behaviour

All the test successfully completes

### Actual Behaviour

The reactive test does fail. The reactive retryable method doesn't retry while blocking and async methods successfully retry.

### Environment Information

- **Operating System**: TODO
- **Micronaut Version:** 2.0.2
- **JDK Version:** TODO

### Example Application

https://github.com/Im-dex/micronaut-rx-retryable-issue

dstepanov:
Reactive retry is done by re-subscribing so the method is invoked only once.

Im-dex:
> Reactive retry is done by re-subscribing so the method is invoked only once.

It is quite unobvious, error-prone, and inconsistent behavior I think, isn't it?

graemerocher:
Generally for code like you have presented in the example there is no value in using reactive programming 

If you are just doing logic and returning `just` or `error` then you may as well use imperative style programming and be done with it in which case retry behaviour would work like you expect.

Im-dex:
> Generally for code like you have presented in the example there is no value in using reactive programming
> 
> If you are just doing logic and returning `just` or `error` then you may as well use imperative style programming and be done with it in which case retry behaviour would work like you expect.

The code is just a minimal example. Let's take something like this:
```java
@Singleton
class Bean {
    @Inject
    DbReader reader;

    @Retryable(attempts = "5", delay = "1ms")
    public Mono<Integer> read(String name) {
        return reader.readFromDatabase("SELECT * FROM ....", name)
            .map(record -> record.getColumn("age"));
    }
};
```

Here I should don't forget to wrap the body into `Mono.defer` or nothing will be retried silently.
So the code becomes ugly:

```java
@Singleton
class Bean {
    @Inject
    DbReader reader;

    @Retryable(attempts = "5", delay = "1ms")
    public Mono<Integer> read(String name) {
        return Mono.defer(() ->
            reader.readFromDatabase("SELECT * FROM ....", name)
                .map(record -> record.getColumn("age"))
        );
    }
};
```

But micronaut already knows how to call the method, so why the users need to do unnecessary wrapping?

graemerocher:
Seems a bug in `dbReader` which is not allowing to be subscribed to twice?

Im-dex:
> Seems a bug in `dbReader` which is not allowing to be subscribed to twice?

[...]

PR #4154, opened 2020-09-23
Retry method call before reactive type is returned

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/4141


graemerocher:
Thanks for the contribution!

PR #4157, opened 2020-09-23
Fix introspecting external @Entity classes

dstepanov:
It's not enough to check if the class has `@Introspected`, external classes present the annotation because if the mapper.

Not sure if the fix is correct.

### DIFF
--- a/inject-groovy/build.gradle
+++ b/inject-groovy/build.gradle
@@ -14,6 +14,8 @@ dependencies {
     testImplementation project(":validation")
     testImplementation 'org.neo4j.driver:neo4j-java-driver:1.4.5'
     testImplementation dependencyModuleVersion("groovy", "groovy-json")
+    testImplementation "com.blazebit:blaze-persistence-core-impl:1.5.0-Alpha1"
+
 }
 
 test {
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/micronaut/core/io/service/SoftServiceLoader.java
```diff
@@ -197,23 +198,20 @@ public boolean hasNext() {
...
+                        .map(line -> {
+                            int i = line.indexOf('#');
+                            if (i > -1) {
+                                line = line.substring(0, i);
+                            }
+                            return line;
+                        })
+                        .collect(Collectors.toList());
+                    unprocessed = lines.iterator();
+                } catch (IOException | UncheckedIOException e) {
```
dstepanov: Not related, somehow I would get `UncheckedIOException` a lot during Gradle invocation

graemerocher:
Yeah this seems like a hack, must be a better a way 🤔

jameskleeh:
@dstepanov Can you more clearly describe the problem?

dstepanov:
@jameskleeh The problem here is that the mapper adds `@Introspected` to all `@Entity` but external classes are not processed. That causes `@Introspected(classes = ValuesEntity.class)` to skip the class because of the founded `@Introspected` annotation.

graemerocher:
seems the correct solution would be to make sure mappers are applied to external classes as well?

dstepanov:
Mapper is applied correctly but `IntrospectedTypeElementVisitor` doesn't process classes in a jar.
Manually defining `classes`/`package` skips the class because annotation is present because of the mapper.

graemerocher:
There are only tests for Groovy code in this PR, the change needs to add tests for Java as well

dstepanov:
@graemerocher Added

dstepanov:
@graemerocher Should this be merged?

graemerocher:
Thanks for the contribution!

PR #4168, opened 2020-09-24
Allow to use @Around and @Introduction(interfaces) together

dstepanov:
I have a use case where I want to implement an interface and also intercept all the methods, at the moment is not possible.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/InterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/InterceptorChain.java
@@ -285,9 +285,9 @@ public class InterceptorChain<B, R> implements InvocationContext<B, R> {
         Set<Class> applicableClasses = new HashSet<>();
 
         for (Class<? extends Annotation> aClass: annotations) {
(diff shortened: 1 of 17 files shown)


dstepanov:
@jameskleeh @graemerocher What do you think about this one? It's a small change with a big potential 😄

graemerocher:
looks good to me. @jameskleeh ?

jameskleeh:
This seems like it would only work if the `@Type` is the same. If you had a different interceptor for around and introduction I think only one of them would be invoked because the annotation specified later would override the value for `@Type`. I'm having trouble understanding the use case as well.

graemerocher:
seems like the test is using different interceptors unless I am missing something

dstepanov:
@jameskleeh I have updated the test to show that different interceptors are invoked.

Use-case is to implement a proxy, we need to interceptor for all the methods and we want to introduce a new interface that provides some functionality related to the proxy implementation.

dstepanov:
@jameskleeh @graemerocher Can this one make to 2.1.0?

PR #4186, opened 2020-09-25
Correct ExecutableMethod cast and allow to change the target in ExecutableMethod generated by `@Around`

dstepanov:
Sorry for another PR 🤣 
I have been working on POC to make compile-time Hibernate proxies without any bytecode modifications.

https://github.com/dstepanov/micronaut-sql/commit/89f54fe14e51b084ddb5efca0bacf7cd4f1e5bb1

I think this is the last thing that is needed, it looks like I need all of the AOP features: introduce, around + executable.
It surprised me that it isn't possible to take `ExecutableMethod` from `MethodInvocationContext` and simply invoke it with a different target, looks like the target is hard-wired and the only way is to have an executable method for every bean method 🤔 

Maybe there is a better way @graemerocher @jameskleeh ?

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -571,11 +571,22 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
                         invokeMethodVisitor.visitMaxs(AbstractClassFileWriter.DEFAULT_MAX_STACK, 1);
                         invokeMethodVisitor.visitEnd();
                     } else {
+                        Label invokeTargetBlock = new Label();
+
                         // load this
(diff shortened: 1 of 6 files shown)


dstepanov:
I have investigated it a bit more, it's possible to fix hard-wired target in `@Around` methods by checking the target.

Now `invokeInternal` looks like this:
```
    public Object invokeInternal(Object var1, Object[] var2) {
        Intercepted var10000 = this.$parent;
        if (var10000 == var1) {
            Intercepted.$$access2(var10000, (Long)var2[0]);
            return null;
        } else {
            ((MyBean6)var1).setId((Long)var2[0]);
            return null;
        }
    }
```

With this change I think it should be possible to add those proxy executable methods as ordinary executable methods with null `parent`.

graemerocher:
nice work

PR #4190, opened 2020-09-28
Move cache endpoint to micronaut-cache

### LINKED ISSUE #4172: Docs for caches endpoint need moved out
We still have documentation for the caches endpoint (listed as one of the built in endpoints provided OOTB) and have documentation for it here: https://docs.micronaut.io/latest/guide/index.html#cachesEndpoint. The docs should be moved to the micronaut-cache project

ilopmar:
Fixes #4172

This PR needs to be merged with https://github.com/micronaut-projects/micronaut-cache/pull/219


graemerocher:
Rather than just deleting it which may breaks links does it make sense to add a link to the new location?

ilopmar:
Yes, you're right. Restored the file and linked to the micronaut-cache docs.

PR #4193, opened 2020-09-28
Issue 3718 - provide a mechanism/api for TypeElementVisitors to contr…

wetted:
Issue 3718 - provide a mechanism/api for TypeElementVisitors to contribute resources for optional consumption by subsequently processed TypeElementVisitors (e.g. OpenAPI generates files that need to be included by GraalVM).

### DIFF
--- a/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyVisitorContext.java
+++ b/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyVisitorContext.java
@@ -15,6 +15,8 @@
  */
 package io.micronaut.ast.groovy.visitor;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import groovy.lang.GroovyClassLoader;
 import io.micronaut.ast.groovy.utils.AstAnnotationUtils;
 import io.micronaut.ast.groovy.utils.InMemoryByteCodeGroovyClassLoader;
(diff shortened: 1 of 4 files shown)


graemerocher:
IMO this is not ready for 2.1 

There are aspects to consider here since generated resources need to report their originating element for the purpose of incremental compilation and we need to consider that when adding resources to be written.

wetted:
This PR doesn't result in new code generating any resources that weren't already being generated, nor causing them to be generated. It's an API that allows TypeElementVisitors to report information about resources they already generate (e.g. OpenApi), and for other downstream TypeElementVisitors (e.g. graal) to consume that information. The participating TypeElementVisitors should already take into account considerations for incremental compiling, regardless of the code under consideration here.

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyVisitorContext.java
```diff
@@ -309,4 +310,14 @@ private SyntaxErrorMessage buildErrorMessage(String message, Element element) {
...
     }
 
+    @Override
+    public Collection<String> getGeneratedResources() {
+        return Collections.unmodifiableCollection(generatedResources);
+    }
+
+    @Override
+    public void addGeneratedResource(@NonNull String resource) {
+        info("Adding generated resource: " + resource, null);
```
graemerocher, line 320: Why is this info statement here seems like debug output?

wetted: I suppose it doesn't need to be. Do we have some guidelines for when to use `info` messages during TypeElement visitor processing?

wetted: It's in the `JavaVisitorContext` too.

graemerocher: For me this adds no value to the user. Output should be generated to point user to a generated file but not this kind of debut output IMO

wetted: OK. I'll remove it from the java and Groovy visitor contexts.

PR #4204, opened 2020-09-29
Reuse executable methods

dstepanov:
This PR changes how Around/Introduce generates executable methods, instead of generating inner classes it generates executable methods with extended ability.

The executable method class now knows about possible interception proxy and allows it to delegate to a target or bypass it by a new boolean field.

I have tested it on private projects:
P1: *.class count 2986 -> 2825
P2: *.class count 4106 -> 3831

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -33,7 +33,6 @@ import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.naming.NameUtils;
 import io.micronaut.core.reflect.ReflectionUtils;
 import io.micronaut.core.util.ArrayUtils;
-import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.value.OptionalValues;
 import io.micronaut.inject.BeanDefinition;
 import io.micronaut.inject.ExecutableMethod;
@@ -56,7 +55,17 @@ import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.lang.reflect.Constructor;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -505,8 +514,6 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
                                    boolean isDefault,
                                    boolean isInterface) {
 
-        boolean interfaceMethod = isInterface || this.isInterface;
-
         List<Object> argumentTypeList = new ArrayList<>(argumentTypes.values());
         int argumentCount = argumentTypes.size();
         Type returnTypeObject = getTypeReference(returnType);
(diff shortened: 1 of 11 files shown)


graemerocher:
Nice work! I think this is probably late to get into 2.1 as it needs a more thorough review and we are releasing 2.1 this week but I like this idea and it is definitely a win in terms of class count which will help memory consumption etc. We need to think about whether this change is backwards compatible from a byte code perspective.

dstepanov:
Yeah, there is no added missing value to rush the merge, it would be nice if somebody reviewed it carefully.
I think it's backward compatible, everything is regenerated in a new version.

graemerocher:
@dstepanov could you rebase to 2.2.x?

dstepanov:
@graemerocher Did you have time to take a look at it?

graemerocher:
@dstepanov with regards to the generated byte code can you paste a before and after for the generated code? Want to make 100% sure I understand what I think this is doing.

dstepanov:
The main idea is to generate one implementation of `AbstractExecutableMethod` to be used in the AOP proxy and the bean definition.

```java
// inherited annotations:
@Around
@Introduction(interfaces = CustomProxy.class)
public class MyBean1 {

    private Long id;
    private String name;

    public MyBean1() {
    }

    public Long getId() {
        return id;
    }

    public void setId(Long id) {
        this.id = id;
    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }


}
```

Shared `AbstractExecutableMethod` implementation 
```java
@Generated
class $MyBean1$InterceptedDefinition$$exec2 extends AbstractExecutableMethod {
    private final boolean $interceptable;

    protected AnnotationMetadata resolveAnnotationMetadata() {
        return InterceptedDefinitionClass.$ANNOTATION_METADATA;
    }

    public $MyBean1$InterceptedDefinition$$exec2() {
        this(false);
    }

    public $MyBean1$InterceptedDefinition$$exec2(boolean var1) {
        super(MyBean1.class, "getId", Argument.of(Long.class, "getId"));
        this.$interceptable = var1;
    }

    public final boolean isAbstract() {
        return false;
    }

    public final boolean isSuspend() {
        return false;
    }

    public Object invokeInternal(Object var1, Object[] var2) {
        return this.$interceptable && var1 instanceof Intercepted ? ((Intercepted)var1).$$access$$getId() : ((MyBean1)var1).getId();
    }

    public final Method resolveTargetMethod() {
        return ReflectionUtils.getRequiredMethod(MyBean1.class, "getId", ReflectionUtils.EMPTY_CLASS_ARRAY);
    }
}
```
Bean definition
```java
@Generated
class $MyBean1$InterceptedDefinition extends AbstractBeanDefinition<Intercepted> implements BeanFactory<Intercepted> {
    protected $MyBean1$InterceptedDefinition(Class var1, AnnotationMetadata var2, boolean var3, Argument[] var4) {
        super(var1, var2, var3, var4);
        super.addExecutableMethod(new $MyBean1$InterceptedDefinition$$exec1());
        super.addExecutableMethod(new $MyBean1$InterceptedDefinition$$exec2());
        super.addExecutableMethod(new $MyBean1$InterceptedDefinition$$exec3());
        super.addExecutableMethod(new $MyBean1$InterceptedDefinition$$exec4());
        super.addExecutableMethod(new $MyBean1$InterceptedDefinition$$exec5());
    }
...
```
AOP proxy
```java
@Generated
class MyBean1$Intercepted extends MyBean1 implements CustomProxy, Introduced {
    private final Interceptor[][] $interceptors = new Interceptor[5][];
    private final ExecutableMethod[] $proxyMethods = new ExecutableMethod[5];

    ...

    public MyBean1$Intercepted(BeanContext var1, Qualifier var2, @Type({ProxyIntroduction.class, ProxyAround.class}) Interceptor[] var3) {
        this.$proxyMethods[0] = new exec1(); // implemented interface method
        this.$interceptors[0] = InterceptorChain.resolveIntroductionInterceptors(var1, this.$proxyMethods[0], var3);
        this.$proxyMethods[1] = new exec2(true);
        this.$interceptors[1] = InterceptorChain.resolveAroundInterceptors(var1, this.$proxyMethods[1], var3);
        this.$proxyMethods[2] = new exec3(true);
        this.$interceptors[2] = InterceptorChain.resolveAroundInterceptors(var1, this.$proxyMethods[2], var3);
        this.$proxyMethods[3] = new exec4(true);
        this.$interceptors[3] = InterceptorChain.resolveAroundInterceptors(var1, this.$proxyMethods[3], var3);
        this.$proxyMethods[4] = new exec5(true);
        this.$interceptors[4] = InterceptorChain.resolveAroundInterceptors(var1, this.$proxyMethods[4], var3);
    }
}
```

The worst scenario I saw: `@Controller` with X methods where Z of them are http methods and Y of them are annotated with `@Around` advice - `@Async`, `@Transactional` etc. it generates:
- Z executable methods added to the bean definition
- Y executable methods added to the proxy bean definition
- X executable methods (inner class) added to AOP proxy

With this change, it only generates max X executable methods instead of worst scenario 3x executors per method.

graemerocher:
Thanks for the contribution!

PR #4205, opened 2020-09-29
Save Micronaut version in properties and read it from there

### LINKED ISSUE #4100: VersionUtils class is not compatible with GraalVM
The method used to get the Micronaut version https://github.com/micronaut-projects/micronaut-core/blob/2.0.x/core/src/main/java/io/micronaut/core/version/VersionUtils.java#L46 doesn't work with GraalVM.

ilopmar:
Fixes #4100

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -37,3 +37,8 @@ spotless {
     }
 }
 
+processResources.dependsOn('micronautVersionInfo')
+task micronautVersionInfo(type: WriteProperties) {
+    outputFile = "${buildDir}/resources/main/micronaut-version.properties"
+    property 'micronaut.version', projectVersion
+}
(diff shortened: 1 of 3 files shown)


graemerocher:
For this to work on GraalVM you would need to declare the file no?

ilopmar:
No because that code is executed at build time and we can access those resources without declaring them. I tried yesterday and it worked.

Having said that I've tried again with the code as it is in 2.0.x branch and it is failing because after my test I run the checkstyle and I needed to reorder the code and declare the `public static final String MICRONAUT_VERSION` before the `static` block, so now it is executed before it and it fails.

I've submitted this PR to fix it (again): https://github.com/micronaut-projects/micronaut-core/pull/4211

ilopmar:
Once this new PR is merged in `2.0.x` and we merge it up to  `2.1.x` I'll add a new test in the GraalVM basic-app tests

Issue #4165, opened 2020-09-24
Declarative client with CompletableFuture<Optional<T>> returns null on 404 instead of Optional.empty()


dave08:
Given:
```kotlin
@Client("some-service")
interface SomeService {
    @Get("/")
    fun index(): CompletableFuture<Optional<String>>
}
```

and I get a 404, shouldn't I get an empty optional? Here it seems like it would return null... : https://github.com/micronaut-projects/micronaut-core/blob/c5699e4817c62d577c26ef30eeb088efc596f4e7/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L497

This behavior seems to contradict the documentation over here: https://docs.micronaut.io/latest/guide/index.html#clientError

PR #4212, opened 2020-09-30
Support CompletableFuture<Optional<T>> in Http Server/Client

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/4165


graemerocher:
Thanks for the contribution!

PR #4230, opened 2020-10-01
Fix default ProxySelector

### LINKED ISSUE #4229: Make the default ProxySelector work with HttpClient
### Steps to Reproduce

1. Set a working HTTP proxy via standard Java system properties (see: https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) using a DNS hostname (not IP address) and port.
2. Upon application startup set proxySelector to `java.net.ProxySelector.getDefault()` in `DefaultHttpClientConfiguration`.
3. Using the default `HttpClient` send an HTTP request to a public website or API accessible by the configured proxy server.

### Expected Behaviour

The HTTP request goes through and the response received is the same as without the proxy.

### Actual Behaviour

`HttpClientException` is thrown. Caused by: java.nio.channels.UnresolvedAddressException: null.

### Environment Information

- **Operating System**: macOS 10.15.7
- **Micronaut Version:** 2.0.3
- **JDK Version:** 11.0.7

### Example Application

I have put together a test running squid as a proxy in a testcontainer attempting to download the `micronaut.io` webpage. The test spec is located in this [branch](https://github.com/Tesco/micronaut-core/tree/fix-default-selector) alongside with my proposed fix. I will raise a PR.

The problem is that the proxy selector is returning (potentially) unresolved addresses and the client is not attempting to resolve those before passing it to netty. When a fix `proxyAddress` is set in the `HttpClientConfiguration` this does not cause issue since the `ConversionService` always creates resolved `InetSocketAddress` instances. Since the configuration instance is a long-lived object, this is not such a good idea, as the underlying IP address may change in the DNS over time. Moreover, hostname resolution in `InetSocketAddress` is done in a best effort basis, if it fails the exception gets swallowed, and the configuration may end up with an unresolved proxyAddress anyway. We have seen this happening in long running services.

lgathy:
Fixes #4229 

Besides fixing the proxyAddress resolution I also made `SocketAddress` instances returned by the `ConversionService` to be unresolved to avoid storing the IP indefinitely and `ProxySelector.getDefault()` configurable from yaml:

```yaml
micronaut.http.client.proxy-selector: 'default'
```

This did not seem to cause any issues here in micronaut-core, however it may need to be tested in separate modules to avoid side-effects. Also, it would worth mentioning in the docs, however I did not attempt to make that change.

I am not sure if testing this kind of public page download is the best way to test this behaviour, I could not come up with anything more elegant.

Unfortunately testcontainers does not seem to work for me on Mac, however it seems to build on Linux and Windows.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1363,6 +1363,13 @@ public class DefaultHttpClient implements
         String username = configuration.getProxyUsername().orElse(null);
         String password = configuration.getProxyPassword().orElse(null);
 
+        if (proxyAddress instanceof InetSocketAddress) {
+            InetSocketAddress isa = (InetSocketAddress) proxyAddress;
+            if (isa.isUnresolved()) {
+                proxyAddress = new InetSocketAddress(isa.getHostString(), isa.getPort());
+            }
+        }
+
         if (StringUtils.isNotEmpty(username) && StringUtils.isNotEmpty(password)) {
             switch (proxyType) {
                 case HTTP:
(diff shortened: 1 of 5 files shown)


graemerocher:
Thanks for the contribution!

lgathy:
@graemerocher Could this be merged back to `2.0.x` too before the new patch release? As far as I recall there were no conflicts. Should I raise a PR for that or will you do it automatically?

graemerocher:
There may or may not be a new patch release of 2.0.x .. the next patch release scheduled is 2.1.1

PR #4237, opened 2020-10-02
Change method "add" to append header values if not null

### LINKED ISSUE #4126: SimpleHttpHeaders.add() is setting the header instead of adding
[This line](https://github.com/micronaut-projects/micronaut-core/blob/master/http/src/main/java/io/micronaut/http/simple/SimpleHttpHeaders.java#L90) is **setting** the header instead of **adding** a new header value. The entire implementation looks wrong as well, because Http headers can accept multi values..

I have found it by chance while running unit tests in a project without `micronaut-http-server-netty`. If netty project is added, calls to `io.micronaut.http.HttpResponse.ok()` would return a netty implementation at low level that doesn`t have the bug.

ze12augusto:
Fix #4126


graemerocher:
This PR doesn't make a great deal of sense. The implementation should be modified to use a multi value map not split and merge strings internally

graemerocher:
Could you target the 2.1.x branch instead of master?

jameskleeh:
@ze12augusto Can you please re-apply your changes on top of the 2.1.x branch? I see additional commits in this PR that aren't yours and in the past that has caused changes to be reverted.

ze12augusto:
Closed by another pr #4267 with correct branch

PR #4246, opened 2020-10-05
Add support for array types of @Introspected annotated classes in graalvm

mliberato:
It is practical to annotate your model classes (such as models parsed from JSON over remote servers) with `@Introspected`. 
When dealing with reflection, some libs may want to create Array types reflectively by calling `Array.newInstance(Class)`.
That is the case of kotlin-reflect + jackson kotlin module. During class member discover, the library try to create an Array instance of the type using the above method.

Currently `GraalTypeElementVisitor` generates this snippet for a sample `LoginRequest` annotated class:
```
{
  "name" : "com.example.LoginRequest",
  "allDeclaredFields" : true,
  "allPublicMethods" : true,
  "allDeclaredConstructors" : true
}
```

To make things work, you have to add to `reflection-config.json` file a syntax like this:
```
{
  "name" : "com.example.LoginRequest[]"
}
```

You need this even if you want to do `Array.newInstance` by yourself.

With that, I find reasonable to add this patch to register by default array types of `@Introspected` annotated classes.
This patch adds this behavior only for the current annotated class, but it may be extended to add to classes under `@Introspected.classes` attribute as well.


mliberato:
I think the broken build is not on me. It seems to be something related to SSL.

ilopmar:
@mliberato yeah, it's not you one of the certificates used in tests expired. Please rebase `2.1.x`

mliberato:
> @mliberato yeah, it's not you one of the certificates used in tests expired. Please rebase `2.1.x`

Still failing..

ilopmar:
It seems there are some checkstyle issues...

ilopmar (approved):

Issue #4244, opened 2020-10-05
2.1.0 App doesn't start in AWS


dstepanov:
```
│ Caused by: java.lang.UnsupportedOperationException                                                                                                                                                                                         │
│     at java.base/java.util.AbstractList.add(AbstractList.java:153)                                                                                                                                                                         │
│     at java.base/java.util.AbstractList.addAll(AbstractList.java:270)                                                                                                                                                                      │
│     at io.micronaut.context.env.DefaultEnvironment.<init>(DefaultEnvironment.java:117)                                                                                                                                                     │
│     at io.micronaut.context.DefaultApplicationContext$BootstrapEnvironment.<init>(DefaultApplicationContext.java:517)                                                                                                                      │
│     at io.micronaut.context.DefaultApplicationContext$RuntimeConfiguredEnvironment.createBootstrapEnvironment(DefaultApplicationContext.java:745)                                                                                          │
│     at io.micronaut.context.DefaultApplicationContext$RuntimeConfiguredEnvironment.start(DefaultApplicationContext.java:695)                                                                                                               │
│     at io.micronaut.context.DefaultApplicationContext$RuntimeConfiguredEnvironment.start(DefaultApplicationContext.java:665)                                                                                                               │
│     at io.micronaut.context.DefaultApplicationContext.startEnvironment(DefaultApplicationContext.java:213)                                                                                                                                 │
│     at io.micronaut.context.DefaultApplicationContext.start(DefaultApplicationContext.java:165)                                                                                                                                            │
│     at io.micronaut.runtime.Micronaut.start(Micronaut.java:64)   
```

safarmer:
This issue is coming from setting the current environment (either through the environment variable `MICRONAUT_ENVIRONMENTS` or system property `micronaut.environments`). The `BootstrapEnvironment` passes an immutable list of environments to `DefaultEnvironment` which then attempts to add to this list.

See PR #4247 for a fix.

jameskleeh:
@safarmer Do you have an example app that reproduces the issue? The bootstrap environment uses `Arrays.asList` as the return for `getEnvironments`, which should be mutable. https://github.com/micronaut-projects/micronaut-core/blob/2.1.x/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java#L532

Edit: n/m. I saw in the source it creates an `ArrayList`, but that is `java.utils.Arrays.ArrayList`, not the same as `java.util.ArrayList`.

[...]

PR #4247, opened 2020-10-06
Issue 4244 - ensure that the environments list is mutable #4244

safarmer:
Since the BootstrapEnvironment calls the contructor with an implementation of `getEnvironments`
that is immutable, as could others, the DefaultEnvironment class was updated to convert the list
to a mutable implementation.

The test case for getting Micronauts environments from the environment variables has been updated
to reflect the case of setting the bootstrap environment when the environment variable is used.

Fixes: issue #4244


jameskleeh:
@safarmer Can you merge 2.1.x into your branch? That should get the build to pass

safarmer:
@jameskleeh Done. Thanks!

graemerocher:
Thanks for the contribution!

PR #4259, opened 2020-10-06
Add AdvisedBeanType interface

### LINKED ISSUE #4217: BeanDefinitionReference does not have full access to metadata
### Task List

- [ ] Stacktrace (if present) provided - not relevant
- [ ] Example that reproduces the problem uploaded to Github - steps are described below

### Steps to Reproduce

1. Load metadata using the service loader
Let's assume I am interested in `Singleton` only:
```java
        SoftServiceLoader.load(BeanDefinitionReference.class)
                .forEach(list::add);

        List<BeanDefinitionReference> collect = list.parallelStream()
                .filter(ServiceDefinition::isPresent)
                .map(ServiceDefinition::load)
                .filter(BeanDefinitionReference::isPresent)
                // we only care for singletons
                .filter(BeanDefinitionReference::isSingleton)
                .collect(Collectors.toList());
```
2. Find all types that are supported by Micronaut
```java
for (BeanDefinitionReference defRef : collect) {
            micronautBeansMap.computeIfAbsent(defRef.getBeanType(), it -> new LinkedList<>())
                    .add(defRef);
}
```
3. A problem
Now I would expect the `getBeanType()` to return the actual type of the bean (e.g. the class that is present in my source code), or I would expect a method that would do so. 
Currently the `getBeanType()` returns the `$Intercepted` class, instead of the original

### Expected Behaviour

I would expect a way to get to the original class without the need to resort to scanning implemented interfaces and superclasses of each bean type. 

### Actual Behaviour
I cannot reach the original class except through reflection.

- **Operating System**: MacOs
- **Micronaut Version:** 2.0.3
- **JDK Version:** 11

The source files I use ("the beans"):
```java
@JdbcRepository(dialect = Dialect.H2)
public interface DbOwnerRepository extends CrudRepository<Owner, Long> {
    @Override
    List<Owner> findAll();

    Optional<Owner> findByName(String name);
}

@JdbcRepository(dialect = Dialect.H2)
public abstract class DbPetRepository implements PageableRepository<Pet, UUID> {

    public abstract List<NameDTO> list(Pageable pageable);

    @Join("owner")
    public abstract Optional<Pet> findByName(String name);
}
```

And I need to have a mapping as follows (without the semantic understanding that these are Micronaut Data repositories, or any other types of beans):
`Class<DbOwnerRepository> -> BeanDefinitionReference`
`Class<DbPetRepository> -> BeanDefinitionReference`

This is for integration purposes with another framework, where I need to map CDI beans to Micronaut beans

graemerocher:
Introduces a new interface called AdvisedBeanType that allows access to the intercepted type from a bean definition reference or a bean definition.

Fixes #4217


jameskleeh:
@graemerocher This seems like a 2.2 thing for me. Is there a reason this needs to go into 2.1?

tomas-langer:
I have tried with the current version and it works as expected. Thanks

Issue #4263, opened 2020-10-06
SAM method generated from @Adapter annotation with multiple arguments can have getTypeArguments() in wrong order


willemsst:
I've been fighting this morning with the use of the `@Adapter` annotation and using a `Qualifiers.byTypeArguments(..)` to lookup my generated _Single Abstract Method_ **(SAM)** classes.

For some weird reason, the order of types in the generated types, is not the same as the one in my interface that is referenced by the @Adapter annotation.

I've been trying to find the magic of the AOP code generation in the source code, but couldn't spot it yet, so, if anyone can clarify how the order is determined, I would like to look into it to understand what's going on. Because, it feels like a bug that the order is changed, but as this is already at version 2.1.0, I might just do something stupid as well.

_Originally posted by @willemsst in https://github.com/micronaut-projects/micronaut-core/discussions/4254_

willemsst:
I found the issue and resolved it as I commented in the original question: https://github.com/micronaut-projects/micronaut-core/discussions/4254#discussioncomment-91618

the problem is that 
https://github.com/micronaut-projects/micronaut-core/blob/1c2f8783cb00da30890330aaf8fd267a1cf208b4/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java#L1474
should actually be a `LinkedHashMap` instead of a `HashMap` as this variable is later used to generate the `getTypeArguments()` code of the generated proxy class. With the HashMap, as you surely know, the order is not respected.

https://github.com/micronaut-projects/micronaut-core/blob/1c2f8783cb00da30890330aaf8fd267a1cf208b4/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java#L1527-L1535

attached a patch containing some test code and the bugfix.

[SAM_TypeArguments.patch.zip](https://github.com/micronaut-projects/micronaut-core/files/5336434/SAM_TypeArguments.patch.zip)

gbkishan:
Hi @willemsst. Good Day. It seems as per your patch code, was implemented custom Event handler (CqrsEventBusHandler). Can I have that piece of code how this is implemented, where i have the requirement of implementation of triggering the events, with the one which has the custom annotation. I see this can be implemented using @Adapter as per documentation. Somehow I am unable to implement the same. Can you please help me with piece of code on this.

willemsst:
[...]

PR #4265, opened 2020-10-07
fixed issue #4263: HashMap should be a LinkedHashMap to respect order …

willemsst:
…of method arguments in generated getTypeArguments()


graemerocher:
Thanks for the CLA you need to do the commits with the same account as your submit the PR with. The commits are with `stevenwillems` whilst your PR is with `willemsst`

stevenwillems:
> Thanks for the CLA you need to do the commits with the same account as your submit the PR with. The commits are with `stevenwillems` whilst your PR is with `willemsst`

hmm.. not sure how that happened, but I also signed with the stevenwillems account. should be good now.

graemerocher:
Thanks for the contribution!

PR #4267, opened 2020-10-07
Change SimpleHttpHeaders implementation to use multimap

### LINKED ISSUE #4126: SimpleHttpHeaders.add() is setting the header instead of adding

ze12augusto:
Fix #4126

### DIFF
--- a/http/src/main/java/io/micronaut/http/simple/SimpleHttpHeaders.java
+++ b/http/src/main/java/io/micronaut/http/simple/SimpleHttpHeaders.java
@@ -57,10 +57,7 @@ public class SimpleHttpHeaders implements MutableHttpHeaders {
     @SuppressWarnings("unchecked")
     public <T> Optional<T> get(CharSequence name, ArgumentConversionContext<T> conversionContext) {
         Optional<String> value = headers.getFirst(name.toString());
-        if (value.isPresent()) {
-            return conversionService.convert(value, conversionContext);
-        }
-        return Optional.empty();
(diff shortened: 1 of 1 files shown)


http/src/main/java/io/micronaut/http/simple/SimpleHttpHeaders.java
```diff
@@ -53,31 +56,26 @@ public SimpleHttpHeaders(ConversionService conversionService) {
     @Override
     @SuppressWarnings("unchecked")
     public <T> Optional<T> get(CharSequence name, ArgumentConversionContext<T> conversionContext) {
-        String value = headers.get(name.toString());
-        if (value != null) {
+        Optional<String> value = headers.getFirst(name.toString());
+        if (value.isPresent()) {
             return conversionService.convert(value, conversionContext);
         }
         return Optional.empty();
```
jameskleeh: This can be changed to use `.map()`

jameskleeh (changes requested):

graemerocher:
Thanks for the contribution!

Issue #4260, opened 2020-10-06
camelCase http service/client name doesn't apply config read-timeout


arudge:
### Steps to Reproduce

1. Create a mn http service with a camel case name, like 'camelCaseV1' and attempt to set the `read-timeout` value for this service.
2. Make an http request that would exceed the timeout

### Expected Behaviour

A `ReadTimeout` exception should be thrown

### Actual Behaviour

The default netty timeout of 10 seconds is still used.

### Environment Information

- **Operating System**: osx and alpine 
- **Micronaut Version:** `2.0.0` and `2.0.1`
- **JDK Version:** `11`

### Example Application

This test documents the issue: https://github.com/arudge/mn-http-service-names/blob/master/src/test/groovy/mn/http/service/name/NonPropertyAwareClientSpec.groovy#L28

This test shows a work around: https://github.com/arudge/mn-http-service-names/blob/master/src/test/groovy/mn/http/service/name/PropertyAwareClientSpec.groovy#L27

graemerocher:
Service IDs have to be in kebab case pretty much universally in Micronaut, this probably needs clarification in the documentation.

PR #4269, opened 2020-10-07
Closes https://github.com/micronaut-projects/micronaut-core/issues/4260

arudge:


jameskleeh:
@arudge This is a bit misleading. The configuration doesn't have to be kebab case, however the id in the `@Client` annotation does.

arudge:
@jameskleeh updated to be specific to the `@Client` `id`

PR #4283, opened 2020-10-09
Enable changing logger levels from properties when using Log4j

xpmatteo:
When using the log4j2 feature, setting log levels via application properties does not work.  It turns out it doesn't because there is no implementation of LoggingSystem for Log4j.  This pull request fixes the problem, I think.

### DIFF
--- a/runtime/build.gradle
+++ b/runtime/build.gradle
@@ -26,6 +26,7 @@ dependencies {
         transitive = false
     }
 
+    compileOnly "org.apache.logging.log4j:log4j-core:2.12.1"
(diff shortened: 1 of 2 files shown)


xpmatteo:
I did sign the CLA... I don't know why it still says I havent

ilopmar:
@xpmatteo It's because you did the commit with user `xpmatteo-tw` but send the PR with `xpmatteo`.

xpmatteo:
Thanks @ilopmar ! I will close this PR and open another with the correct user.  I don't see a simpler way to fix it.

PR #4286, opened 2020-10-10
Enable changing logger levels from properties when using Log4j

xpmatteo:
When using the log4j2 feature, setting log levels via application properties does not work. It turns out it doesn't because there is no implementation of LoggingSystem for Log4j. This pull request fixes the problem.

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/loggers/impl/Log4jLoggingSystem.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/loggers/impl/Log4jLoggingSystem.java
@@ -37,6 +37,7 @@ import static java.util.stream.Collectors.toList;
  * An implementation of {@link ManagedLoggingSystem} that works with logback.
  *
  * @author Matteo Vaccari, Matthew Moss
+ * @since 2.2.0
  */
 @Singleton
 @Requires(beans = LoggersEndpoint.class)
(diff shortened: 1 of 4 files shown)


graemerocher:
Thanks for the contribution!

xpmatteo:
I also added a second LoggingSystem implementation to be used when the micronaut-management module is loaded.  With this, the /loggers endpoint works with log4j2

runtime/src/main/java/io/micronaut/logging/impl/Log4jLoggingSystem.java
```diff
@@ -0,0 +1,53 @@
...
+import io.micronaut.logging.LoggingSystem;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.config.Configurator;
+
+import javax.inject.Singleton;
+
+/**
+ * An implementation of {@link LoggingSystem} that works with Log4j.
+ *
+ * @author Matteo Vaccari
```
jameskleeh, line 30: Add @since 2.2.0 here

management/src/main/java/io/micronaut/management/endpoint/loggers/impl/Log4jLoggingSystem.java
```diff
@@ -0,0 +1,110 @@
...
+
+import javax.inject.Singleton;
+import java.util.Collection;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * An implementation of {@link ManagedLoggingSystem} that works with logback.
+ *
+ * @author Matteo Vaccari, Matthew Moss
```
jameskleeh, line 39: Add @since 2.2.0 here

graemerocher:
Thanks for the contribution!

PR #4314, opened 2020-10-13
Fix regression in jaeger tracing when controller throws exception

### LINKED ISSUE #4313: Regression in tracing error response with Jaeger
Due to changes in #4112 the span of a failed server request is not tracked correctly, since a new span is initiated for the error handler.

### Steps to Reproduce

1. Create a Micronaut application providing an HTTP endpoint with a long-running operation that fails with an exception after a while.
2. Enable Jaeger tracing and reporting configured.
3. Call the endpoint.
4. Inspect the reported span details, i.e. duration.

### Expected Behaviour

The reported span should accurately contain the duration of the whole request processing.

### Actual Behaviour

The reported span only account for the duration of the error handler, not the whole request processing. Furthermore, there are unfinished spans created by Jaeger (observable via metrics).

### Environment Information

- **Operating System**: MacOS 10.15.7
- **Micronaut Version:** 2.0.3 (also present with 2.1.1)
- **JDK Version:** 11.0.8

### Example Application

- I included a test case in  `HttpTracingSpec` demonstrating this in the following [branch](https://github.com/Tesco/micronaut-core/tree/jaeger-error-regression). I also have a fix for it, I will raise a PR.

Further remarks: it is also possible that this causes some sort of memory leak due to the unfinished spans opened for the original request processing.

lgathy:
Closes #4313

### DIFF
--- a/test-suite/src/test/groovy/io/micronaut/tracing/jaeger/HttpTracingSpec.groovy
+++ b/test-suite/src/test/groovy/io/micronaut/tracing/jaeger/HttpTracingSpec.groovy
@@ -16,6 +16,7 @@
 package io.micronaut.tracing.jaeger
 
 import io.jaegertracing.internal.JaegerTracer
+import io.jaegertracing.internal.metrics.InMemoryMetricsFactory
 import io.jaegertracing.internal.reporters.InMemoryReporter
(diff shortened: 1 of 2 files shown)


graemerocher:
Thanks for the contribution

PR #4319, opened 2020-10-14
Issue 3582 configure Netty shutdown quiet period and shutdown timeout

### LINKED ISSUE #3582: Add configuration to control shutdown behavior of event loop groups
### Task List

### Description

I experienced one side effect of using a commonly shared event loop group for handling incoming requests and sending requests using the declarative HTTP client in the command-line application scenario. The default [`shutdownGracefully` method implementation](https://github.com/netty/netty/blob/netty-4.1.48.Final/common/src/main/java/io/netty/util/concurrent/AbstractEventExecutorGroup.java#L70) uses 2 seconds shutdown quiet period. It does not affect much long-running microservice like applications. However, it adds an unnecessary 2 seconds lag to the short-living apps like the command-line app or the serverless function. I created a demo app that calls https://api.chucknorris.io/jokes/random endpoint using Micronaut's declarative HTTP client as well as a regular Java 11 HTTP client. While the second one responds in ~300ms, the first one keeps the process running for 2 seconds after the output gets printed to the console.

[asciicast](https://asciinema.org/a/KcQU8bL8cxKLkTcEwrj5Nx6a3)

I checked if there is a simple way to override this behavior in the current state of the framework, but I couldn't find a single easy to apply solution. I tried to configure a separate `eventLoopGroup` for the `http.client` and then inject extended `EventLoopGroup` that overrides the default behavior of the `shutdownGracefully` method. I was thinking about injecting the default `eventLoopGrooup` and decorating it with the custom implementation of that interface, but when I've seen tons of methods that had to be decorated, I said "ok, enough."

I would like to volunteer and create a PR with a working solution, but I would like to ask what do you think about it in the first place. Micronaut's `@Bean(preDestroy)` allows us to define a method via string and it expects that the destroy method does not take any parameters. I see two potential options:

* extend `@Bean(preDestroy)` to support a parameterized method, so one can pre-configure the bean to use zero seconds quite period when necessary,
* provide Micronaut's wrapper for the `EventLoopGroup` with configurable options passed to `shutdownGracefully` method.

The declarative HTTP client is powerful, and I think people love it so much that they would like to use it in the command-line apps or serverless functions. However, when every cli/function execution adds 2 seconds to something that takes less than a second, it feels like something could be improved in this area.

### Steps to Reproduce
```
git clone https://github.com/wololock/micronaut-declarative-http-client-app-demo.git

cd micronaut-declarative-http-client-app-demo

./gradlew --no-daemon assemble

sdk use java 20.1.0.r11-grl

native-image --no-server -cp build/libs/demo-cli-0.1-all.jar

time ./demo-cli dynamic-http-client

time ./demo-cli java-http-client
```

### Expected Behaviour

There should be a way to configure an event loop group to shut down instantly.

### Actual Behaviour

There is no simple way to override the default `shutdownGracefully` method to shut down an even loop group instantly.

### Environment Information

- **Operating System**: Fedora Linux 32 (kernel 5.6.19-300.fc32.x86_64)
- **JDK Version:** GraalVM CE 20.1.0 (build 11.0.7+10-jvmci-20.1-b02)

### Example Application

- https://github.com/wololock/micronaut-declarative-http-client-app-demo

wetted:
Fixes #3582 - adding feature to configure the shutdown quiet period and shutdown timeout for Netty event loop groups and http server

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
@@ -233,4 +233,16 @@ public interface HttpClient extends Closeable, LifeCycle<HttpClient> {
     static HttpClient create(@Nullable URL url) {
         return HttpClientConfiguration.createClient(url);
     }
+
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration the client configuration
+     * @return The client
+     */
+    static HttpClient create(@Nullable URL url, HttpClientConfiguration configuration) {
+        return HttpClientConfiguration.createClient(url, configuration);
+    }
 }
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
@@ -241,6 +241,7 @@ public interface HttpClient extends Closeable, LifeCycle<HttpClient> {
      * @param url The base URL
      * @param configuration the client configuration
      * @return The client
+     * @since 2.2.0
      */
     static HttpClient create(@Nullable URL url, HttpClientConfiguration configuration) {
         return HttpClientConfiguration.createClient(url, configuration);
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -17,6 +17,7 @@ package io.micronaut.http.client;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.context.ApplicationContext;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.convert.format.ReadableBytes;
 import io.micronaut.core.util.ArgumentUtils;
@@ -27,6 +28,7 @@ import io.micronaut.http.ssl.SslConfiguration;
 import io.micronaut.logging.LogLevel;
 import io.micronaut.runtime.ApplicationConfiguration;
 
+import javax.inject.Inject;
 import java.net.*;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
@@ -36,7 +38,10 @@ import java.util.*;
 import java.util.concurrent.ThreadFactory;
 
 /**
- * Configuration for the {@link HttpClient}.
+ * Configuration for the {@link HttpClient}. This configuration only takes affect for {@link HttpClient}
+ * instances created outside the application context using {@link HttpClient#create(URL, HttpClientConfiguration)}.
+ * For clients created within the context using, e.g. {@link Inject} or
+ * {@link ApplicationContext#createBean(Class)}, use event loop group configuration.
  *
  * @author Graeme Rocher
  * @since 1.0
@@ -54,6 +59,12 @@ public abstract class HttpClientConfiguration {
      */
     @SuppressWarnings("WeakerAccess")
     public static final long DEFAULT_READ_IDLE_TIMEOUT_MINUTES = 5;
+
+    /**
+     * The default shutdown timeout in millis.
+     */
+    @SuppressWarnings("WeakerAccess")
+    public static final long DEFAULT_SHUTDOWN_QUIET_PERIOD_MILLISECONDS = 1;
 
     /**
      * The default shutdown timeout in millis.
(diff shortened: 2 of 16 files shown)


http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
```diff
@@ -56,29 +66,49 @@ public DefaultEventLoopGroupRegistry(EventLoopGroupFactory eventLoopGroupFactory
...
+     */
+    @PreDestroy
+    void shutdown() {
+        eventLoopGroups.forEach((eventLoopGroup, configuration) -> {
+            try {
+                long quietPeriod = configuration.getShutdownQuietPeriod().toMillis();
+                long timeout = configuration.getShutdownTimeout().toMillis();
+                eventLoopGroup.shutdownGracefully(quietPeriod, timeout, TimeUnit.MILLISECONDS);
+            } catch (Throwable t) {
+                LOGGER.warn("Error shutting down EventLoopGroup: {}", t.getMessage(), t);
```
jameskleeh: This should be surrounded with `if (LOGGER.isWarnEnabled())`

http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
```diff
@@ -42,9 +48,13 @@
 @Internal
 @BootstrapContextCompatible
 public class DefaultEventLoopGroupRegistry implements EventLoopGroupRegistry {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultEventLoopGroupRegistry.class);
```
jameskleeh: `LOG` here is generally preferred over `LOGGER`

http-client-core/src/main/java/io/micronaut/http/client/RxStreamingHttpClient.java
```diff
@@ -59,4 +59,16 @@
...
+
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration The client configuration
+     * @return The client
+     */
+    static RxStreamingHttpClient create(URL url, HttpClientConfiguration configuration) {
```
jameskleeh, line 72: Please add @since 2.2.0 here

http-client-core/src/main/java/io/micronaut/http/client/RxHttpClientFactory.java
```diff
@@ -43,4 +53,14 @@
...
+
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration The client configuration
+     * @return The client
+     */
+    RxStreamingHttpClient createStreamingClient(@Nullable URL url, HttpClientConfiguration configuration);
```
jameskleeh, line 67: Please add @since 2.2.0 here

http-client-core/src/main/java/io/micronaut/http/client/RxHttpClientFactory.java
```diff
@@ -35,6 +35,16 @@
...
 
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration the client configuration
+     * @return The client
+     */
+    RxHttpClient createClient(@Nullable URL url, HttpClientConfiguration configuration);
```
jameskleeh, line 47: Please add @since 2.2.0 here

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -609,6 +649,26 @@ static RxHttpClient createClient(@Nullable URL url) {
...
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration The client configuration
+     * @return The client
+     */
+    @Internal
+    static RxStreamingHttpClient createStreamingClient(@NonNull URL url, HttpClientConfiguration configuration) {
```
jameskleeh, line 667: Please add @since 2.2.0 here

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -586,19 +619,26 @@ public Proxy resolveProxy(boolean isSsl, String host, int port) {
...
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration the client configuration
+     * @return The client
+     */
+    @Internal
+    static RxHttpClient createClient(@Nullable URL url, HttpClientConfiguration configuration) {
```
jameskleeh, line 637: Please add @since 2.2.0 here

http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
```diff
@@ -233,4 +233,16 @@ default HttpClient refresh() {
...
+
+    /**
+     * Create a new {@link HttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link javax.inject.Inject} to inject a client instead
+     *
+     * @param url The base URL
+     * @param configuration the client configuration
+     * @return The client
+     */
+    static HttpClient create(@Nullable URL url, HttpClientConfiguration configuration) {
```
jameskleeh, line 246: Please add @since 2.2.0 here

jameskleeh (changes requested):

PR #4333, opened 2020-10-16
Fix proxy generation of inherited default method

dstepanov:
The problem introduced in 2.1.
Bridge proxy should't use the default method's declared class to make "super" call, which doesn't work when the method is inherited from another interface, solution is to use actual type.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -638,7 +638,7 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
                 bridgeGenerator.loadArg(i);
             }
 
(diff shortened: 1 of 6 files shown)


dstepanov:
Actually the problem was introduced in 2.0.

@graemerocher @jameskleeh Any idea what what does this means:
`
java.lang.IncompatibleClassChangeError: Method 'java.lang.String io.micronaut.http.client.DefaultMethodClient3.defaultMethod()' must be Methodref constant
`
?
Only occurs in JDK15. Probably something is missing in bytecode generation.

graemerocher:
I do not, maybe compare the byte code JDK 15 generates to what Micronaut generates to see what the differences are

dstepanov:
Looks like JDK 15 is more strict 🤯

graemerocher (approved):

PR #4340, opened 2020-10-19
Use Spock MicronautTest for Groovy example

jonnybot0:
This is a very minor doc change. I noticed that the Groovy example was using a deprecated annotation while setting up my first app, figured I could submit a quick fix. If it's more complicated than I'm making it out to be, no hard feelings if you need to decline. :)


graemerocher:
Thanks for the contribution!

PR #4341, opened 2020-10-19
Adds a new marker interface HttpRequestTenantResolver

sdelamo:
- cookie
- subdomain
- principal 
- http header 
- session

resolvers implement this marker. 

This allows to resolve the current tenant without using `ServerRequestContext.currentRequest();`


multitenancy/src/main/java/io/micronaut/multitenancy/tenantresolver/HttpRequestTenantResolver.java
```diff
@@ -0,0 +1,39 @@
...
+ * @since 2.1.2
+ */
+public interface HttpRequestTenantResolver {
+    /**
+     * Resolves the current tenant identifier.
+     * @param request The HTTP request
+     * @return The tenant identifier
+     * @throws TenantNotFoundException if tenant not found
+     */
+    Serializable resolveTenantIdentifier(@NonNull @NotNull HttpRequest<?> request) throws TenantNotFoundException;
```
graemerocher, line 40: Needs to expression whether the return type is nullable or not

graemerocher (changes requested):

multitenancy/src/main/java/io/micronaut/multitenancy/tenantresolver/HttpRequestTenantResolver.java
sdelamo: I've added the `@NonNull` annotation to both `TenantResolver::resolveIdentifier`  and `HttpRequestTenantResolver::resolveIdentifier(HttpRequest)` methods.

sdelamo:
Moreover, I've annotated  both `TenantResolver`  and `HttpRequestTenantResolver` with `@FunctionalInterface`

jameskleeh:
@sdelamo This should target 2.2

PR #4358, opened 2020-10-21
Add support for obtaining the interception target

graemerocher:
This PR adds support for obtaining the target of an interceptor once when the target is created. This always doing any initialization upfront avoiding delaying it until the when the method interceptor executes which can eliminate the need for creating caches in certain circumstances.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/aop/targetaware/TypeSpecificTargetAwareInterceptor.java
+++ b/inject-java/src/test/groovy/io/micronaut/aop/targetaware/TypeSpecificTargetAwareInterceptor.java
@@ -1,0 +1,35 @@
+package io.micronaut.aop.targetaware;
+
+import io.micronaut.aop.MethodInvocationContext;
+import io.micronaut.aop.TargetAwareMethodInterceptor;
+import io.micronaut.context.Qualifier;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.spockframework.util.Assert;
+
+import javax.inject.Singleton;
+
+@Singleton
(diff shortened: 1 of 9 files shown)


inject-java/src/test/groovy/io/micronaut/aop/targetaware/TypeSpecificTargetAwareInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+    private int count;
+    @Override
+    public Object intercept(MethodInvocationContext<TestBean, Object> context) {
+        Assert.notNull(target);
+        return context.proceed();
+    }
+
+    @Override
+    public void newTarget(@Nullable Qualifier<TestBean> qualifier, @NotNull TestBean target) {
+        count++;
```
JaroslavTulach, line 24: Nice. Yes, this would be helpful for initialization of [proto and micro fields](https://github.com/jtulach/micronautui/blob/04ff91230ffa470536e7c0a88d2d3d20bfa3ab5a/beaninfo-api/src/main/java/com/dukescript/impl/javafx/beans/ObservableInterceptor.java#L22) as soon as the `ObservableInterceptor` instance is created.

graemerocher:
Think this needs further thought with regards to when the target bean is destroyed

JaroslavTulach:
After additional hint by @graemerocher it turned out that it is better to process all annotated beans on context initialization. Done in [6956c690](https://github.com/jtulach/micronautui/commit/6956c690bc4f8d0356461aeb9ac17a551f8bd73e) and it works great. My work no longer needs here in proposed changes.

PR #4375, opened 2020-10-23
Fix continuation being encoded in the body of suspend client methods

### LINKED ISSUE #4360: Using suspend fun in HttpClient sends the continuation info in the json payload
Given a client:
```kotlin
@Client(id = "some-service", path = "/")
interface SomeService {
    @Put("/something/{requestId}")
    suspend fun setState(@PathVariable requestId: Int, newState: SomeEnum): HttpStatus
}
```
wiremock logs report something like:
```
Accept: [application/json]
Connection: [close]
Host: [localhost:9613]
Content-Length: [156]
Content-Type: [application/json]
{"newState":"sent","continuation":{"completableFuture":{"completedExceptionally":false,"numberOfDependents":1,"done":false,"cancelled":false},"context":{}}}
```
I was expecting only:
`{"newState":"sent"}` to be in the body.

### Environment Information

- **Operating System**: linux mint
- **Micronaut Version:** 2.1.1
- **JDK Version:** 1.8

jameskleeh:
Fixes #4360

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/MethodInterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/MethodInterceptorChain.java
@@ -38,7 +38,7 @@ import java.lang.reflect.Method;
 @Internal
 public final class MethodInterceptorChain<T, R> extends InterceptorChain<T, R> implements MethodInvocationContext<T, R> {
 
-    private final static Object[] EMPTY_ARRAY = new Object[0];
+    private static final Object[] EMPTY_ARRAY = new Object[0];
 
     /**
      * Constructor for empty parameters.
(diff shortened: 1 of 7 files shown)


graemerocher:
Wouldn't it be far simpler to just skip the last argument if `method.isSuspend()` returns true here? https://github.com/micronaut-projects/micronaut-core/blob/2.1.x/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L230

jameskleeh:
Doing that would require the duplication of logic that the continuation argument is the last argument

graemerocher:
Where would be this duplication?

PR #3831, opened 2020-07-30
request input...

ladybug:
Closes #3425. Committing what I have for input and review feedback... Am I on the right track with this?

PR #4378, opened 2020-10-24
argument binder documentation created

### LINKED ISSUE #3425: Document request argument binders

meyerinj:
Intended to close #3425 .

Like Cindy's work on this in #3831 I am interested in input on this as I am interested in input.


src/main/docs/guide/httpServer/customArgumentBinding.adoc
```diff
@@ -0,0 +1,60 @@
...
+TIP: It is common to use api:io.micronaut.core.convert.ConversionService[ConversionService] to convert the data to the type of argument.
+
+=== TypedRequestArgumentBinders
+
+* api:http.bind.binders.TypedRequestArgumentBinder[TypedRequestArgumentBinder]
+
+- An interface for classes that bind an api:core.type.Argument[Argument] from an api:http.HttpRequest[HttpRequest] driven by
+the argument's type.
+
+Binders are called by the api:io.micronaut.http.annotation.Controller[Controller] once the request is received and the argument
```
jameskleeh: This paragraph is confusing to me. The filters aren't called by controllers. Both controllers and the binders are called by the server logic. I think the next paragraph is a great description, however I think an example should be given (the filters example). I would move the generic descriptions of argument binders above the specific information for typed and annotated otherwise it seems the generic information is specific to typed. The code samples need to be snippets and available in all languages and need to have tests to verify the code actually works.

meyerinj: If I understood your recommendations correctly this new push should do what you suggested. Fun fact: I had never written Kotlin before this!

Issue #4306, opened 2020-10-13
Bean validation should fail if @Valid field is not introspected


dstepanov:
2.1.0

If you are using `@Valid` on the body parameter there is a runtime error if the field type is not `@Introspected` (BTW: It would be great have compile-time error but I guess that wouldn't be correct for Hibernate validator)
That is not the case if you have a nested field with a type without introspection, the field is not validated and you are wondering was is wrong.

jameskleeh:
@dstepanov Are you going to be submitting a PR? If not, can you upload an example app?

dstepanov:
I haven't looked into it. The example is a simple case of:

```
@Introspected
class A {
   @Valid
   private List<B> bees;
   ...
}

// Not introspected
class B {
   @NotNull
   private String name;
   ...
}

@Post 
void submit(@Body @Valid A body) { ... }

`{"bees": {}}` is OK if `B` is not introspected

```

jeffscottbrown:
> Bean validation should fail if @Valid field is not introspected

Should that be the requirement or should the requirement be that validation should work even if the class is not marked introspected?

dstepanov:
An exception is thrown if `@Body` object class is not introspected it should also fail if nested objects aren't introspected and you requested them to be validated by `@Valid` annotation.

dstepanov:
@wetted Are you working on it?

wetted:
@dstepanov among other things, yes, so that non-introspected nested objects will fail the `@Valid` validation.

dstepanov:
I have come up with this fix https://github.com/micronaut-projects/micronaut-core/pull/4405

wetted:
I submitted [PR 4411](https://github.com/micronaut-projects/micronaut-core/pull/4411) for the basic case where just the @Valid annotation is on a container, where the container element type is not introspected. That fixes this case. 

However, it doesn't address bogus cases like `@Valid @SomeConstraint List<Foo> foosNotInstrospected` where `@SomeConstraint` doesn't support `Foo` directly. This is a bit more complicated since something like `@Valid @Size(min=1, max=2) List<String> strs` is perfectly valid, even though String isn't introspected. The fix is not so straightforward since any number of constraints are legitimate with @Valid on a container, as long as the container element type is a supported type for all the constraints. This needs to accommodate not only built-in constraints, but custom ones as well.

PR #4405, opened 2020-10-29
Fail if object is not introspected for validation

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/4306

### DIFF
--- a/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
+++ b/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
@@ -1330,7 +1330,13 @@ public class DefaultValidator implements Validator, ExecutableMethodValidator, R
             Object propertyValue,
             @Nullable DefaultPropertyNode container) {
 
-        final BeanIntrospection<Object> beanIntrospection = getBeanIntrospection(propertyValue);
+        BeanIntrospection<Object> beanIntrospection = null;
+        if (propertyValue != null && !ClassUtils.isJavaLangType(propertyValue.getClass())) {
+            beanIntrospection = getBeanIntrospection(propertyValue);
+            if (beanIntrospection == null) {
+                throw new ValidationException("Passed object [" + propertyValue + "] cannot be introspected. Please annotate with @Introspected");
+            }
+        }
 
         if (beanIntrospection != null) {
             if (container != null) {
--- a/validation/src/test/groovy/io/micronaut/validation/validator/ValidatorSpec.groovy
+++ b/validation/src/test/groovy/io/micronaut/validation/validator/ValidatorSpec.groovy
@@ -10,8 +10,10 @@ import spock.lang.Shared
 import spock.lang.Specification
 
 import javax.inject.Singleton
+import javax.validation.ConstraintViolation
 import javax.validation.ElementKind
 import javax.validation.Valid
+import javax.validation.ValidationException
 import javax.validation.ValidatorFactory
 import javax.validation.constraints.*
 import javax.validation.metadata.BeanDescriptor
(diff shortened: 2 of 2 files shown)


validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
```diff
@@ -1330,7 +1330,13 @@ public void keyedValue(String nodeName, Object key, Object keyedValue) {
             Object propertyValue,
             @Nullable DefaultPropertyNode container) {
 
-        final BeanIntrospection<Object> beanIntrospection = getBeanIntrospection(propertyValue);
+        BeanIntrospection<Object> beanIntrospection = null;
+        if (propertyValue != null && !ClassUtils.isJavaLangType(propertyValue.getClass())) {
```
jameskleeh, line 1334: I'm not sure if `!ClassUtils.isJavaLangType(propertyValue.getClass()))` is enough. This could break other cases where `@Valid` is being applied. Perhaps a `List<Date>`. You could get around that with `ClassUtils.isJavaBasicType`, however even then I think there is a high potential for this to break applications

dstepanov:
I think the main problem is how Micronaut implements  `@Valid`.

According to the bean validation spec:
```
In addition to supporting instance validation, validation of graphs of objects is also supported. 
The result of a graph validation is returned as a unified set of constraint violations. 
@Valid is used to express validation traversal of an association.
```

I would expect `@Valid` to be used to traverse to the object and validate its fields etc, but Micronaut also uses `@Valid` to traverse containers and apply constraints applied to the container.
This case is not valid in Hibernate validation:
```
@Introspected
@Prototype
class ArrayTest {
    @Valid
    @Max(20l)
    @NotNull
    int[] integers
...
}
```
- Spec doesn't support arrays
- There is no implementation of `@Max` for `int[]`
- Only way is to use `List<@Max(20l) Integer>`

Micronaut will validate the array with `@Max` for every element of the container but only if it's annotated with `@Valid`.

I think the behaviour should change to drop the requirement for `@Valid` when applying a constrait to the container elements and require every `@Valid` object to be introspected.

WDYT? @jameskleeh @graemerocher

graemerocher:
My understanding of the spec is that `@Valid` is used to control cascading validation so in 1.0.0 of the spec (which is what we support right now), the following:

```
    @Valid
    @Max(20l)
    @NotNull
    int[] integers
```

Would apply `@Max` and `@NotNull` to each element of the array. If you remove `@Valid` it would try to apply those to the array itself.

I don't think the requirement to add `@Valid` in order to cascade should be dropped at all however, as it also leave the user in control of when validation should cascade and if cascading were to happen automatically (which I believe is not what the spec says should happen) then you are risk to stack overflows etc.

Now 2.0 of the spec is different because that introduces support for annotations in generic type arguments which we don't currently support at compilation time and we need to add support for.

dstepanov:
Does it mean that in this case:
```
    @Valid
    @Max(20l)
    @NotNull
    List<Integer> integers
```
`integers` can be null because constraints should be applied to the container element?

graemerocher:
I don't believe in that case integers can be null, since `@NotNull` would apply to each container element in bean validation 1.0.0

validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
wetted: I agree with @jameskleeh that `isjavaLangType(Class)` doesn't cover everything. Any parameterized type supported by a build-in or custom constraint definition doesn't require introspection. That's easy enough to implement for the simple case of `@Valid List<NotIntrospected> list` but is a bit more complicated for something like `@Valid @FooConstraint(…) List<Foo> list`.

wetted:
The implementation is broken for the following case:
```
    @Valid
    @Size(min=1, max=3)
    @NotNull
    List<String> strs // or Set<String> or String[]
```
because `@Size` is applied to each string in the list., but also to the list itself. This is clearly not right since the size of the list itself is constrained the same as the size of each element in the list. This is likely the case for any constraint that is applicable to a container as well as the element type of the container.

dstepanov:
I think of solutions:
- Fail only if no constraint is applied in 2.2
- Remove constraint propagation in the major version, support type constraints and always fail on Valid

WDYT?

graemerocher:
@wetted it is a bug to apply the validation to the `list` itself. The use of `@Valid` should mean the constraints are only applied to the elements of the list

jameskleeh:
I think this should be the plan:

2.2: If the field has only @Valid and no other annotation, fail validation if the class is not introspected
3.0: Conform to the 2.0 spec

wetted:
> @wetted it is a bug to apply the validation to the `list` itself. The use of `@Valid` should mean the constraints are only applied to the elements of the list

I agree that's the bug. I encountered it while researching the bug we're addressing currently. I have a failing spock test that demonstrates it. With @Valid no constraints should apply to the container itself, but that is what's happening currently. 

- I have a fix as @jameskleeh noted above for 2.2. 
- I will create an issue for the case of a constraint applied erroneously to a container when @Valid is present

dstepanov:
@wetted Do you have a fix to allow it only when no constraints are applied?

wetted:
> @wetted Do you have a fix to allow it only when no constraints are applied?

Yes, I've been working this already. I have a PR coming for it soon.

dstepanov:
Great

wetted:
> > @wetted Do you have a fix to allow it only when no constraints are applied?
> 
> Yes, I've been working this already. I have a PR coming for it soon.

On [PR 4411](https://github.com/micronaut-projects/micronaut-core/pull/4411)

PR #4411, opened 2020-10-29
Issue 4306 - Bean validation should fail if @Valid field is not introspected

### LINKED ISSUE #4306: Bean validation should fail if @Valid field is not introspected

wetted:
Issue 4306 - Bean validation should fail if @Valid field is not introspected
- emit validation failure when `@Valid` alone is applied to a container of elements where the element type is not `@Introspected`.
- this does not address cases with additional constraint(s), since that might still be legal if the element type is supported by those constraints directly (e.g. `@Valid @Size(min=1, max=2) List<String> strs`), even though something like `@Valid @Size(min=1, max=2) List<NotIntrospected> oops` might be bogus.

### DIFF
--- a/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
+++ b/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
@@ -1321,8 +1321,7 @@ public class DefaultValidator implements Validator, ExecutableMethodValidator, R
 
         final BeanIntrospection<Object> beanIntrospection = getBeanIntrospection(propertyValue);
         AnnotationMetadata annotationMetadata = cascadeProperty.getAnnotationMetadata();
-        if (beanIntrospection == null && annotationMetadata.getDeclaredAnnotationNames().equals(
-            Collections.singleton(javax.validation.Valid.class.getName()))) {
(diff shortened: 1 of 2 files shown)


validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
```diff
@@ -1331,6 +1320,14 @@ public void keyedValue(String nodeName, Object key, Object keyedValue) {
             @Nullable DefaultPropertyNode container) {
 
         final BeanIntrospection<Object> beanIntrospection = getBeanIntrospection(propertyValue);
+        AnnotationMetadata annotationMetadata = cascadeProperty.getAnnotationMetadata();
+        if (beanIntrospection == null && annotationMetadata.getDeclaredAnnotationNames().equals(
```
jameskleeh: `@Valid` is not annotated with `@Constraint` so you can just look for the presence of that annotation instead of comparing the annotation names

PR #4416, opened 2020-10-30
Issue 4315 - health readiness/liveness bug

### LINKED ISSUE #4315: Health readiness and liveness endpoints responding with 200 OK even when status is DOWN
From [this PR](https://github.com/micronaut-projects/micronaut-core/pull/4012) included on the 2.1.0 release, now the `/health` endpoint includes `/health/liveness` and `/health/readiness` endpoints. But integrating them, both liveness and readiness are responding with `200 OK` no matter the status, as opposed to the health one which returns `200 OK`, `503 SERVICE UNAVAILABLE`.

### Task List

### Steps to Reproduce

Add the following configuration to the service configuration yml file:
```yml
endpoints:
  health:
    enabled: true
    sensitive: false
    details-visible: ANONYMOUS
    status:
      http-mapping:
        DOWN: 503
    disk-space:
      enabled: true
      threshold: 9000000MB
```
Assuming you don't have 9TB of disk space (increase this value if needed to make the service unhealthy), it will make the service unhealthy. This is using the provided `DiskSpaceIndicator` health indicator. From the docs:

`A health indicator is provided that determines the health of the application based on the amount of free disk space. Configuration for the disk space health indicator can be provided under the endpoints.health.disk-space key.`

The indicator works just fine, as well as the `/health` endpoint, but, the `/readiness` (and `/liveness` with a custom `@Liveness` health indicator) endpoint will respond as follows:

```json
{
  "name": "hello-world",
  "status": "DOWN",
  "details": {
    "compositeDiscoveryClient()": {
      "name": "hello-world",
      "status": "UP"
    },
    "service": {
      "name": "hello-world",
      "status": "UP"
    },
    "diskSpace": {
      "name": "hello-world",
      "status": "DOWN",
      "details": {
        "error": "Free disk space below threshold. Available: 21631291392 bytes (threshold: 9437184000000 bytes)"
      }
    }
  }
}
```
`The "worst" status will be returned as the overall status`

The worst status is `DOWN`, as it is also in the response, but the HTTP response status is `200 OK`. This happens also with the `liveness` operation if you create a custom health indicator (as stated on the GitHub repository example)

### Expected Behaviour

If the status is `DOWN` the HTTP response status should follow the configuration mapping or the default values instead (503 on the above configuration)

### Actual Behaviour

Both liveness and readiness endpoint return 200 OK no matter the status.

### How to fix it

One solution could be on `io.micronaut.management.endpoint.health.filter.HealthResultFilter` there is a filter which only applies for the health endpoint. It is filtering requests under the path `${endpoints.all.path:/}${endpoints.health.id:health}` and setting the HTTP status based on the status of the health response, using the provided mapping on the service configuration (with default values for `DOWN` 503). This does not include `readiness` and `liveness` operations. Including the same path but ended as `/*` as a new value of the `@Filter` annotation could work:

```java
@Filter({DEFAULT_MAPPING, HEALTH_INTERNAL_OPERATIONS_MAPPING})
@Requires(beans = HealthEndpoint.class)
public class HealthResultFilter extends OncePerRequestHttpServerFilter {

    /**
     * Configurable default mapping for filter.
     */
    public static final String DEFAULT_MAPPING =
        "${" + EndpointDefaultConfiguration.PREFIX + ".path:" +
            EndpointDefaultConfiguration.DEFAULT_ENDPOINT_BASE_PATH + "}${" +
            HealthEndpoint.PREFIX + ".id:health}";

    static final String HEALTH_INTERNAL_OPERATIONS_MAPPING = DEFAULT_MAPPING + "/*";
```

### Environment Information

- **Operating System**: macOS
- **Micronaut Version:** 2.1.1
- **JDK Version:** 11

### Example Application

https://github.com/victorcastanogutierrez/micronaut-health-issue

This application has a `CustomHealthIndicator` returning always `DOWN` for the liveness operation. On the readiness, the disk space health indicator is enabled with 9TB as threshold

wetted:
- fixing issue where HttpStatus is 200 by default – instead of 503 – when readiness and/or liveness status is DOWN

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/health/filter/HealthResultFilter.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/health/filter/HealthResultFilter.java
@@ -37,7 +37,11 @@ import org.reactivestreams.Publisher;
  * @since 1.0
  *
  */
-@Filter(HealthResultFilter.DEFAULT_MAPPING)
+@Filter({
+    HealthResultFilter.DEFAULT_MAPPING,
+    HealthResultFilter.LIVENESS_PROBE_MAPPING,
+    HealthResultFilter.READINESS_PROBE_MAPPING
+})
 @Requires(beans = HealthEndpoint.class)
 public class HealthResultFilter extends OncePerRequestHttpServerFilter {
 
@@ -48,6 +52,8 @@ public class HealthResultFilter extends OncePerRequestHttpServerFilter {
             "${" + EndpointDefaultConfiguration.PREFIX + ".path:" +
                     EndpointDefaultConfiguration.DEFAULT_ENDPOINT_BASE_PATH + "}${" +
                     HealthEndpoint.PREFIX + ".id:health}";
+    public static final String LIVENESS_PROBE_MAPPING = DEFAULT_MAPPING + "/liveness";
+    public static final String READINESS_PROBE_MAPPING = DEFAULT_MAPPING + "/readiness";
 
     private final HealthEndpoint healthEndpoint;
 
--- a/management/src/test/groovy/io/micronaut/management/endpoint/health/HealthEndpointSpec.groovy
+++ b/management/src/test/groovy/io/micronaut/management/endpoint/health/HealthEndpointSpec.groovy
@@ -23,9 +23,9 @@ import io.micronaut.health.HealthStatus
 import io.micronaut.http.HttpRequest
 import io.micronaut.http.HttpResponse
 import io.micronaut.http.HttpStatus
+import io.micronaut.http.bind.binders.TypedRequestArgumentBinder
 import io.micronaut.http.client.RxHttpClient
 import io.micronaut.http.client.exceptions.HttpClientResponseException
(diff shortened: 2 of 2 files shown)


mvalensky:
The fix doesn't fix the issue
The filter mapping lacks slash, so instead of 
` "${" + EndpointDefaultConfiguration.PREFIX + ".path:" +
                    EndpointDefaultConfiguration.DEFAULT_ENDPOINT_BASE_PATH + "}${" +
                    HealthEndpoint.PREFIX + ".id:health}";`
it should be 
` "${" + EndpointDefaultConfiguration.PREFIX + ".path:" +
                    EndpointDefaultConfiguration.DEFAULT_ENDPOINT_BASE_PATH + "}/${" +
                    HealthEndpoint.PREFIX + ".id:health}";`

wetted:
@mvalensky Can you please provide a sample project that demonstrates why you think this is still broken? I updated my original development spike to 3.5.x and it still passes as expected.

If you are using the `endpoints.all.path` management configuration property then it should end with a `/` character since the default value is just `/`. Your proposed change would result in `//` as part of the URL otherwise, which is incorrect.

PR #4424, opened 2020-11-03
Allow retrieving array element types

rdesgroppes:
Follow-up of micronaut-projects/micronaut-openapi#368.

This is to ease post-processing without adhoc helpers such as:
https://github.com/micronaut-projects/micronaut-openapi/blob/a4524550d45c36d78061ab392bdf41d274188cf6/openapi/src/main/java/io/micronaut/openapi/visitor/AbstractOpenApiVisitor.java#L1553

It consists in providing a `fromArray` method to all "final" classes for which `toArray` has to be specialized:
* `PrimitiveElement`
* `GroovyClassElement`
  * `GroovyEnumElement`
* `JavaVoidElement` (deprecated)
* `JavaClassElement`
  * `JavaEnumElement`
    (the latter was missing a specialization of `toArray` and therefore used to return a `JavaClassElement`, which is now fixed)


rdesgroppes:
The reason why CI just failed happens to be unrelated to changes brought by the PR:
```java
Micronaut Server Error - No request state present. Cause: javax.net.ssl.SSLHandshakeException: Received fatal alert: certificate_expired
```
Does anyone on the line know, by chance, where the problem is?

graemerocher:
Think this error is fixed in the other branches. Please target the the 2.1.x branch for this change

rdesgroppes:
> Think this error is fixed in the other branches. Please target the the 2.1.x branch for this change

On it, thanks!

Now targeting 2.1.x.

rdesgroppes:
Rebased on top of branch 2.2.x as per above changes (target & milestone).

graemerocher:
Thanks for the contribution!

PR #4443, opened 2020-11-06
Fix Jackson bean introspection constructor initialization

dstepanov:
This fix make it work in the same way as ordinary Jackson.

BTW: I have observed unexpected behavior: it looks like bean introspection will choose the first constructor without any logic around.

I this case no args constructor is choosen: 
```java
    public Notification() {
    }

    public Notification(Long id, MessageTemplate template) {
        this.id = id;
        this.template = template;
    }
```

I this case all args constructor is choosen: 
```java
    public Notification(Long id, MessageTemplate template) {
        this.id = id;
        this.template = template;
    }

    public Notification() {
    }
```


graemerocher:
Thanks, is there an associated issue report?

dstepanov:
I haven't created one and didn't find anything similar to a problem I had.

graemerocher:
Thanks for the contribution!

Issue #4438, opened 2020-11-05
java.lang.ClassCastException when using @Retryable with coroutine


xiscosc:
`@Retryable` annotation is not working well with Kotlin coroutines. It does its work when some error is thrown but it fails when it's trying to return the correct value. I found two kind of errors:

1. ```java.lang.IllegalStateException: Not a Kotlin coroutine```
2. ```java.lang.ClassCastException: class io.micronaut.aop.util.CompletableFutureContinuation cannot be cast to class kotlin.coroutines.jvm.internal.CoroutineStackFrame (io.micronaut.aop.util.CompletableFutureContinuation and kotlin.coroutines.jvm.internal.CoroutineStackFrame are in unnamed module of loader 'app')```

I created a project with several tests in order to reproduce them, it can be found here https://github.com/xiscosc/micronaut-kotlin-retry-coroutines

#### Not a Kotlin coroutine error
1. Create a  class with `@Retryable` and a method that returns from a coroutine
```
@Singleton
@Retryable(attempts = "1")
class HomeService {
    suspend fun test2(): String {
        return coroutineScope {
            "hi"
        }
    }
}
```
2. Call the method
```
@Controller("/home")
class HomeController(val homeService: HomeService) {

    @Get
    suspend fun getTest2(): HttpResponse<String> {
        return HttpResponse.ok(homeService.test2())
    }
}
```
#### Cast error
1. Create a  class with `@Retryable` and a method that returns from a coroutine with a delay
```
@Singleton
@Retryable(attempts = "1")
class HomeService {
    suspend fun test2(): String {
        return coroutineScope {
           delay(1L)
            "hi"
        }
    }
}
```
2. Call the method
```
@Controller("/home")
class HomeController(val homeService: HomeService) {

    @Get
    suspend fun getTest2(): HttpResponse<String> {
        return HttpResponse.ok(homeService.test2())
    }
}
```
### Expected Behaviour

Method `test2()` should return string `"hi"`

#### Not a Kotlin coroutine error
```
java.lang.IllegalStateException: Not a Kotlin coroutine
	at io.micronaut.aop.internal.intercepted.KotlinInterceptedMethod.interceptResultAsCompletionStage(KotlinInterceptedMethod.java:126)
	at io.micronaut.aop.internal.intercepted.KotlinInterceptedMethod.interceptResult(KotlinInterceptedMethod.java:138)
[...]

PR #4453, opened 2020-11-09
Avoid IllegalStateException "Not a Kotlin coroutine" when suspend function returns a value

xiscosc:
As commented in https://github.com/micronaut-projects/micronaut-core/issues/4438 coroutines intercepted with `@Retryable` annotation fail when they return values. 

This PR changes that and allows to return the values instead of throwing an exception.


jameskleeh:
This needs a test

graemerocher:
Thanks for the contribution!

PR #4464, opened 2020-11-11
Make HttpRequest.getPath() return the non-decoded path

### LINKED ISSUE #656: Lambda proxy is routing and passing decoded path variable when it should be encoded
### Context:
Let's say we have the following user class with its respective controller:

```groovy
package com.example

import groovy.transform.CompileStatic

@CompileStatic
@Introspected
class User {

    String email
}
```

```groovy
package com.example

import groovy.transform.CompileStatic
import io.micronaut.http.annotation.Controller
import io.micronaut.http.annotation.Get
import io.micronaut.http.annotation.PathVariable

@Controller('/users')
@CompileStatic
class UserController {

    Set<User> users = [
        new User(email: 'email+alias@gmail.com')
    ].toSet()

    @Get('/{+email}')
    User show(@PathVariable String email) {
        users.find { it.email == email }
    }
}
```

If we run the application purely **without using `micronaut-aws` and its Lambda proxy**, the `UserController.show` action will only find the user `email+alias@gmail.com` if we encode the email. If we don't encode the email Micronaut will interpret the plus sign (`+`) as a whitespace (because it automatically decodes the path variable), therefore the user won't be found.

That means:

1. `GET - /users/email%2Balias%40gmail.com` works as expected, user will be found
2. `GET - /users/email+alias@gmail.com` doesn't work, user won't be found (Micronaut will interpret the path parameter as `email alias@gmail.com`)

So far everything looks good, we just need to make sure emails are encoded when making the request

### Problem:

Once you add the Lambda proxy, as explained in this guide https://guides.micronaut.io/mn-application-aws-lambda-java11-groovy/guide/index.html, the issue starts to happen. The following test for example will fail when it shouldn't:

```groovy
package com.example

import com.amazonaws.serverless.proxy.internal.testutils.AwsProxyRequestBuilder
import com.amazonaws.serverless.proxy.internal.testutils.MockLambdaContext
import com.amazonaws.serverless.proxy.model.AwsProxyRequest
import com.amazonaws.serverless.proxy.model.AwsProxyResponse
import com.amazonaws.services.lambda.runtime.Context
import io.micronaut.function.aws.proxy.MicronautLambdaHandler
import io.micronaut.http.HttpMethod
import io.micronaut.http.HttpStatus
import spock.lang.AutoCleanup
import spock.lang.Shared

import java.nio.charset.StandardCharsets

class UserControllerSpec extends Specification {

    @Shared
    @AutoCleanup
    MicronautLambdaHandler handler = new MicronautLambdaHandler()

    @Shared
    Context lambdaContext = new MockLambdaContext()

    void 'show should find the user'() {
        given:
        String email = 'email+alias@gmail.com'
        String encoded = URLEncoder.encode(email, StandardCharsets.UTF_8.toString())
        String path = "/users/${encoded}"

        when:
        AwsProxyRequest request = new AwsProxyRequestBuilder(path, HttpMethod.GET.toString()).build()
        AwsProxyResponse response = handler.handleRequest(request, lambdaContext)

        then:
        response.statusCode == HttpStatus.OK.code
    }
}
```

`response.statusCode` will be `404` when it should be `200`.

image
image
image

We can also confirm that there is a bug in there when we encode the email twice like this:

```groovy
package com.example

import com.amazonaws.serverless.proxy.internal.testutils.AwsProxyRequestBuilder
import com.amazonaws.serverless.proxy.internal.testutils.MockLambdaContext
import com.amazonaws.serverless.proxy.model.AwsProxyRequest
import com.amazonaws.serverless.proxy.model.AwsProxyResponse
import com.amazonaws.services.lambda.runtime.Context
import io.micronaut.function.aws.proxy.MicronautLambdaHandler
import io.micronaut.http.HttpMethod
import io.micronaut.http.HttpStatus
import spock.lang.AutoCleanup
import spock.lang.Shared

import java.nio.charset.StandardCharsets

class UserControllerSpec extends Specification {

    @Shared
    @AutoCleanup
    MicronautLambdaHandler handler = new MicronautLambdaHandler()

    @Shared
    Context lambdaContext = new MockLambdaContext()

    void 'show should find the user'() {
        given:
        String email = 'email+alias@gmail.com'
        String encoded = URLEncoder.encode(email, StandardCharsets.UTF_8.toString())
        String encodedTwice = URLEncoder.encode(encoded, StandardCharsets.UTF_8.toString())
        String path = "/users/${encodedTwice}"

        when:
        AwsProxyRequest request = new AwsProxyRequestBuilder(path, HttpMethod.GET.toString()).build()
        AwsProxyResponse response = handler.handleRequest(request, lambdaContext)

        then:
        response.statusCode == HttpStatus.OK.code
    }
}
```

Now the test passes and the user is found.

Another thing to note is that if we change the route mapping in the `UserController` like this:

```groovy
...
class UserControllerSpec extends Specification {
    ...
    @Get('/{email}') // plus sign removed
    User show(@PathVariable String email) {
    ....
```

The route won't even match, which makes me believe the problem is happening when routing and also when passing URL parameters.

### Application example:

[demo.zip](https://github.com/micronaut-projects/micronaut-aws/files/5446065/demo.zip)

ttzn:
Fixes https://github.com/micronaut-projects/micronaut-aws/issues/656

The default implementation of `getPath()` returns the URI-decoded path, while `NettyHttpRequest` returns the raw path:

```java
    // In AbstractNettyHttpRequest
    private String decodePath(String uri) {
        QueryStringDecoder queryStringDecoder = createDecoder(uri);
        return queryStringDecoder.rawPath();
    }
```

It's probably best to align the default impl with the most widely-used implementation. In any case, the Javadoc should also state the expected format.

### DIFF
--- a/http/src/main/java/io/micronaut/http/HttpRequest.java
+++ b/http/src/main/java/io/micronaut/http/HttpRequest.java
@@ -127,10 +127,10 @@ public interface HttpRequest<B> extends HttpMessage<B> {
     }
 
     /**
-     * @return Get the path without any parameters
+     * @return Get the raw, percent-encoded path without any parameters
      */
     default @NonNull String getPath() {
-        return getUri().getPath();
+        return getUri().getRawPath();
     }
 
     /**
--- a/http/src/test/groovy/io/micronaut/http/HttpRequestSpec.groovy
+++ b/http/src/test/groovy/io/micronaut/http/HttpRequestSpec.groovy
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017-2019 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http
+
+import spock.lang.Specification
+
+import java.nio.charset.Charset
+
+class HttpRequestSpec extends Specification {
+
+    def "HttpRequest.getPath() returns the non-decoded URI path component"() {
+        when:
+        String pathSegment = "?bar"
+        String encodedPathSegment = URLEncoder.encode(pathSegment, Charset.defaultCharset().name())
+        String path = "http://www.example.org/foo/$encodedPathSegment?queryParam=true"
+        HttpRequest request = HttpRequest.GET(path)
+
+        then:
+        request.getPath() == "/foo/%3Fbar"
+    }
+
+}


graemerocher:
probably needs a test

graemerocher (approved):

PR #4470, opened 2020-11-12
Fix isRoot logic. Fixes #4463

### LINKED ISSUE #4463: Endpoints on root context don't match
After upgrading our apps to micronaut 2.1.* we encountered that applications that run 
on root context "/" no longer serve our list endpoint as it doesn't find the route.
Changing the controller root uri to anything else makes it work again.

### Task List

### Steps to Reproduce

1. Create a controller on as below:

```groovy
@Controller("/")
class RootController {

    @Get("/{?paginationCommand*}")
    HttpResponse<Collection<String>> list(@Valid @Nullable PaginationCommand paginationCommand) {
        HttpResponse.ok(["will", "this", "work"])
    }
}
``` 
2. Call the list endpoint

`curl -v http://localhost:8080/`

It will respond with:

```
{"message":"Not Found","_links":{"self":{"href":"/","templated":false}}}[sf@sf-lenovo root-context-test]$ curl -v http://localhost:8080/
*   Trying ::1:8080...
* connect to ::1 port 8080 failed: Verbindungsaufbau abgelehnt
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Date: Mon, 09 Nov 2020 12:22:34 GMT
< Content-Type: application/json
< Transfer-Encoding: chunked
< Server: Jetty(9.4.33.v20201020)
< 
* Connection #0 to host localhost left intact
{"message":"Not Found","_links":{"self":{"href":"/","templated":false}}}
```
3. Cange the controler to serve on `/test`

```groovy
@Controller("/test")
class RootController {
 …
}
```
4.  Call the list endpoint

`curl -v http://localhost:8080/`

It will respond with:

```
   Trying ::1:8080...
* connect to ::1 port 8080 failed: Verbindungsaufbau abgelehnt
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /test HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Mon, 09 Nov 2020 12:25:47 GMT
< Content-Type: application/json
< Content-Length: 22
< Server: Jetty(9.4.33.v20201020)
< 
* Connection #0 to host localhost left intact
```

### Expected Behaviour

Calling controller methods should work on all controller base urls.

### Actual Behaviour

Calling controller with base url "/" dows not work when it has get params…

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** 2.1.3
- **JDK Version:** 11

### Example Application

https://github.com/safri-net/mn-root-context

jameskleeh:


http/src/main/java/io/micronaut/http/uri/UriMatchTemplate.java
```diff
@@ -202,6 +198,29 @@ protected UriTemplateParser createParser(String templateString, Object... parser
...
+                }
+            } else {
+                if (rawSegment == null) {
+                    rawSegment = segment;
+                } else {
+                    return false;
+                }
+            }
+        }
+        if (rawSegment == null) {
```
graemerocher, line 216: I don't understand this check, would `rawSegment` not always be null?

jameskleeh:
It would be assigned to the first raw segment

http/src/main/java/io/micronaut/http/uri/UriMatchTemplate.java
jameskleeh: It could be null if there are no raw segments. It could not be null if there is a single raw segment combined with only query segments.

graemerocher: yeah but the assignment on line 209 only assigns it if it is `null` so what is the case where it is not `null`?

jameskleeh: its not null if it there is more than 1 raw segment

PR #4427, opened 2020-11-04
Revert "#4396 Allow classloader to be passed as parameter to load annotation type (#4406)"

ilopmar:
This reverts commit 0389f4d17a0f9ce6e9820e1941848166e3e9ea53.

Fixes #4425

PR #4471, opened 2020-11-12
Allow classloader for retrieving annotations

jameskleeh:
This is a new PR for https://github.com/micronaut-projects/micronaut-core/pull/4427 that was reverted.


ilopmar (approved):
GraalVM works with these changes.

Issue #4484, opened 2020-11-14
Cannot obtain bean data from an associated @Prototype


JaroslavTulach:
### Task List

- [ x ] Steps to reproduce provided
- [ NA ] Stacktrace (if present) provided
- [ x ] Example that reproduces the problem uploaded to Github
- [ x ] Full description of the issue provided (see below)

### Steps to Reproduce

I am writing a [UI extension](https://github.com/jtulach/micronautui/blob/5a336657c4a0be74b3451ea2381b803422a58470/README.md) based on Micronaut `@Introspected` annotation. Things are working fine and my [reactive UI framework](https://github.com/jtulach/micronautui/) is capable to implement classical TODO example with [Micronaut concepts](https://github.com/jtulach/micronautui/blob/b0bbbbbe522ec9c47f48ee388aab109e293fed5f/micronautui-demo/src/main/java/io/micronaut/ui/demo/Demo.java). However essential aspect of my solution is to keep additional data in my [prototype object](https://github.com/jtulach/micronautui/blob/b0bbbbbe522ec9c47f48ee388aab109e293fed5f/micronautui-api/src/main/java/io/micronaut/ui/impl/ObservableInterceptor.java) and I need to find the data when I have pointer to the bean only. I haven't found a simple way to do it.

### Expected Behaviour

The PR #4485 shows one possible solution.

### Actual Behaviour

Every [UI reactive](https://github.com/jtulach/micronautui/) class in my system has to provide a dummy override of `equals`:
```
    @Override
    public boolean equals(Object obj) {
        return super.equals(obj);
    }
```

It would be more natural if [code like this one](https://github.com/jtulach/micronautui/blob/b0bbbbbe522ec9c47f48ee388aab109e293fed5f/micronautui-demo/src/main/java/io/micronaut/ui/demo/Demo.java#L101) could be avoided.

### Environment Information

- **Micronaut Version: 2.1.2

### Example Application

https://github.com/jtulach/micronautui/

jameskleeh:
Are you creating classes in the `io.micronaut` package (`import io.micronaut.ui.ObservableUI;`)? If so, you shouldn't do that. I don't understand the problem. Can you describe the problem you're facing more clearly?

JaroslavTulach:
[...]

PR #4485, opened 2020-11-14
Let @Around advice intercept default Object methods.

JaroslavTulach:
This PR is one possible solution to #4484. There may be other solutions as such I am reporting the issue and the fix separately. If this PR is found acceptable, one can consider it a good fix for to necessary `equals` override as well.

Currently the #4484 problem is being solved by [explicitly overriding equals](https://github.com/jtulach/micronautui/blob/b0bbbbbe522ec9c47f48ee388aab109e293fed5f/micronautui-demo/src/main/java/io/micronaut/ui/demo/Demo.java#L101) in each "UI-ready class" and intercepting such `equals` calls later by the [ObservableInterceptor](https://github.com/jtulach/micronautui/blob/5a336657c4a0be74b3451ea2381b803422a58470/micronautui-api/src/main/java/io/micronaut/ui/impl/ObservableInterceptor.java#L57). It is easy to forget to override `equals` and then everything falls apart. It would be better to avoid the need of overriding `equals`.

This PR introduces new `wrapObjectMethods` attribute to the `@Around` annotation. With its help I shall be able to change my [ObservableUI](https://github.com/jtulach/micronautui/blob/b0bbbbbe522ec9c47f48ee388aab109e293fed5f/micronautui-api/src/main/java/io/micronaut/ui/ObservableUI.java#L40) annotation to use 

```java
@Around(wrapObjectMethods = true)
```
and then my code will be able to [intercept calls to equals](https://github.com/jtulach/micronautui/blob/b0bbbbbe522ec9c47f48ee388aab109e293fed5f/micronautui-api/src/main/java/io/micronaut/ui/impl/QueryProto.java#L49) without forcing client code to override

```java
    @Override
    public boolean equals(Object obj) {
        return super.equals(obj);
    }
```

Thanks for considering this enhancement.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/Around.java
+++ b/aop/src/main/java/io/micronaut/aop/Around.java
@@ -80,4 +80,13 @@ public @interface Around {
      * @return True if the proxy target should be resolved lazily
      */
     boolean lazy() default false;
+
+    /**
+     * Process the {@link Object} non-final method by
+     * the {@code @Around} handlers or not.
+     *
+     * @return true if {@code Object} methods shall be processed
+     * @since 2.2
+     */
+    boolean wrapObjectMethods() default false;
 }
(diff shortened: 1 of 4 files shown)


graemerocher:
Thanks Jaroslav. Could you target the 2.2.x branch? 2.1.x is only receiving bug fixes atm

JaroslavTulach:
FYI: I no longer need this PR to solve my inquiry. Denis [found another solution](https://github.com/jtulach/micronautui/commit/d7e2fbeed1ce35946659aea540e6b06c1a43b548) to my problem. This PR can be closed without merging.

graemerocher:
Thanks for confirming!

PR #4507, opened 2020-11-16
Set kotlinx coroutines version to variable. Upgrade to 1.4.1

jameskleeh:


graemerocher:
Note: a dependency that we have integration with and hence should be in the BOM

PR #4575, opened 2020-11-20
fix: protect against npe thrown by nullable bean context

sdelamo:
I am getting a NPE after upgrading from 2.14. to 2.2.0 here: 

```
Caused by: java.lang.NullPointerException: null
	at io.micronaut.aop.chain.InterceptorChain.resolveAroundInterceptors(InterceptorChain.java:246)
	at uk.co.metadataworks.auth.repository.$UserRepositoryProviderImplDefinition$Intercepted.<init>(Unknown Source)
```

This is an extract from `UserRepositoryProviderImpl` bean:

```
@Singleton
public class UserRepositoryProviderImpl implements UserRepositoryProvider {
    private final BeanContext beanContext;
    private final TenantResolver tenantResolver;
    private final Map<String, UserRepository> userRepository = new ConcurrentHashMap<>();

    public UserRepositoryProviderImpl(BeanContext beanContext, TenantResolver tenantResolver) {
        this.beanContext = beanContext;
        this.tenantResolver = tenantResolver;
    }
```

This PR protects against NPE after these changes https://github.com/micronaut-projects/micronaut-core/pull/4471

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/InterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/InterceptorChain.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.aop.chain;
 
+import edu.umd.cs.findbugs.annotations.Nullable;
 import io.micronaut.aop.*;
 import io.micronaut.aop.exceptions.UnimplementedAdviceException;
 import io.micronaut.context.ApplicationContext;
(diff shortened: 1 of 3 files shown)


jameskleeh (changes requested):
We shouldn't change the methods to start accepting nulls when they previously did not. Instead pass `getClass().getClassLoader()` if bean context is null

sdelamo:
I can do what you suggest but I think given that this was a [7 days ago change](https://github.com/micronaut-projects/micronaut-core/pull/4471/files). I personally think it is a bug and we should  considering fixing the method signature to reflect the nullability. 

Do you feel strongly that `classLoader`  should be `@NonNull` at `default Optional<Class<? extends Annotation>> getAnnotationType(@NonNull String name, @NonNull ClassLoader classLoader)`?

jameskleeh:
The signature that should be changed is the one that takes a `BeanContext`. It currently does not have `@Nullable` even though null is possible. That is why this issue wasn't caught.

sdelamo:
I cannot call `getClass().getClassLoader());` in `InterceptorChain::resolveIntroductionInterceptors` because it is a static method. I've moved classLoader resolution to  `AnnotationMetadataDelegate` the first non static method in the chain.

graemerocher:
This is the incorrect fix, not all implementations of `AnnotationMetadata` would necessarily implement `AnnotationMetadataDelegate`

You can use `InterceptorChain.class.getClassLoader()` instead of `getClass().getClassLoader()`

sdelamo:
I've done the change you suggested @graemerocher

PR #4596, opened 2020-11-23
Fix @RequestBean field values that depend on filter result

### LINKED ISSUE #3869: RequestBean does not bind Principal or Authentication
As discussed on Gitter, the @RequestBean annotated bean was not binding the Principal or Authentication instances. HttpRequest was binded succesfully.

### Steps to Reproduce

Try to bind the Principal or Authentication beans to an @RequestBean annotated parameter bean on an controller method, using ```@Get()```

Bean
```java
package user.api;

import java.security.Principal;

import lombok.Getter;
import lombok.Setter;

@Getter
@Setter
@Introspected
public class GetMeRequest {

    private Principal principal;

}
```

Controller
```java
    @Get("/me")
    public Maybe<GetMeResponse> me(@RequestBean GetMeRequest request) {
        // ... 
    }   
```
I've tried using template portion on url too, same problem.
```java
    @Get("/me{?request*}")
    public Maybe<GetMeResponse> me(@RequestBean GetMeRequest request) {
        // ... 
    }   
```

### Expected Behaviour

The principal must be binded with current authenticated user.

### Actual Behaviour

Returns an error

```
Required argument [Principal principal] not specified
```
### Environment Information

- **Operating System**: MacOS Big Sur 
- **JDK Version:** 11

asodja:
This is proposal to fix https://github.com/micronaut-projects/micronaut-core/issues/3869.

Note that I am not sure if this is correct way, since Micronaut currently binds some arguments before filters and unsatisfied arguments after filters. This PR changes behavior that `@RequestBean` is always resolved after filters. 

And since `@RequestBean` can contain other Bindable values (e.g. @Header), means that these values can be resolved differently when they are part of RequestBean vs when they are method parameters. This can be fixed, but adds some lines of code.

What is the reason that arguments are bound before filters?

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/aop/RequestBeanSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/aop/RequestBeanSpec.groovy
@@ -6,7 +6,9 @@ import io.micronaut.core.annotation.Introspected
 import io.micronaut.core.convert.ArgumentConversionContext
 import io.micronaut.core.type.Argument
 import io.micronaut.http.HttpRequest
+import io.micronaut.http.MutableHttpResponse
 import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Filter
 import io.micronaut.http.annotation.Get
 import io.micronaut.http.annotation.Header
 import io.micronaut.http.annotation.PathVariable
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #4610, opened 2020-11-25
Support @RequestBean in @Client methods

dstepanov:
I expected this to work, maybe I'm missing something.
Also, I found a problem with Optional as a query parameter.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
@@ -17,6 +17,8 @@ package io.micronaut.http.client.bind;
 
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.beans.BeanIntrospection;
+import io.micronaut.core.beans.BeanProperty;
 import io.micronaut.core.bind.annotation.Bindable;
 import io.micronaut.core.convert.ConversionContext;
 import io.micronaut.core.convert.ConversionService;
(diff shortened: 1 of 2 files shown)


jameskleeh:
This needs tests for Optional with all of the annotation based binders and if we want to support it properly we need duplicated type based binders for Optional<Cookie> and the like. Probably better to separate that out into its own PR

dstepanov:
@jameskleeh Yeah, probably this is not a correct fix for that `Optional` issue, can we keep this one as it is to introduce the support for `@RequestBean` in the clients?

jameskleeh:
@dstepanov Sure - remove the code for optional handling

dstepanov:
@jameskleeh Done

http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
```diff
@@ -122,6 +124,17 @@ protected DefaultHttpClientBinderRegistry(ConversionService<?> conversionService
         byAnnotation.put(Body.class, (ClientArgumentRequestBinder<Object>) (context, uriContext, value, request) -> {
             request.body(value);
         });
+        byAnnotation.put(RequestBean.class, (ClientArgumentRequestBinder<Object>) (context, uriContext, value, request) -> {
+            BeanIntrospection<Object> introspection = BeanIntrospection.getIntrospection(context.getArgument().getType());
+            for (BeanProperty<Object, Object> beanProperty : introspection.getBeanProperties()) {
+                findArgumentBinder(beanProperty.asArgument()).ifPresent(binder -> {
+                    Object propertyValue = beanProperty.get(value);
+                    if (propertyValue != null) {
+                        binder.bind(ConversionContext.of(beanProperty.asArgument()), uriContext, propertyValue, request);
```
jameskleeh: This should be `context.with(beanProperty.asArgument())` so that errors are combined in the same context

dstepanov: Done

graemerocher (approved):

PR #4611, opened 2020-11-25
Log Micronaut version at startup

alvarosanchez:


graemerocher:
What is the motivation? I think the less output  per line the better maybe just combine the version in the line that output the Server URL like  " Micronaut 2.2.1 startup completed in 926ms. Server Running: http://localhost:8080"

alvarosanchez:
> What is the motivation? 

We were discussing the other day the biggest downside of the framework: the lack of startup banners 😜. Until we get there, we thought that displaying the Micronaut version would be at least really valuable.

> I think the less output per line the better maybe just combine the version in the line that output the Server URL like " Micronaut 2.2.1 startup completed in 926ms. Server Running: http://localhost:8080"

Thought about that, but wasn't sure whether it would break something that was parsing the current output. Can change it like you said, though

alvarosanchez:
Closed favouring #4805

PR #4617, opened 2020-11-27
Publish a new inject-groovy-test module

alvarosanchez:
To allow using `AbstractBeanDefinitionSpec` from other modules


graemerocher:
Looks good, one problem we have is that the build currently doesn't package the Groovy sources in the sources JAR. Might be good to fix that as part of this PR

alvarosanchez:
That should be done in the micronaut-build plugin no?

https://docs.gradle.org/current/samples/sample_building_groovy_libraries.html#generating_sources_jar

graemerocher:
Yes

PR #4618, opened 2020-11-27
GraalVM build profile for Maven users

### LINKED ISSUE #4578: Research about adding a GraalVM profile to the parent POM
While validating https://github.com/micronaut-projects/micronaut-core/issues/4576, we realised about the situation that users can still generate a GraalVM native image even if they don't select the `graalvm` feature. This leads to generating native images that are incomplete, because the `micronaut-graal` annotation processor is not in the classpath.

One possibility we could explore is to add a profile to the parent POM like this:

```xml
<profile>
  <id>graalvm</id>
  <activation>
    <file>
      <exists>${env.JAVA_HOME}/bin/native-image</exists>
    </file>
  </activation>
  <dependencies>
    <dependency>
      <groupId>org.graalvm.sdk</groupId>
      <artifactId>graal-sdk</artifactId>
      <version>${graal.version}</version>
      <scope>provided</scope>
    </dependency>
  </dependencies>
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>${maven-compiler-plugin.version}</version>
        <configuration>
          <annotationProcessorPaths combine.children="append">
            <path>
              <groupId>io.micronaut</groupId>
              <artifactId>micronaut-graal</artifactId>
              <version>${micronaut.version}</version>
            </path>
          </annotationProcessorPaths>
        </configuration>
      </plugin>
    </plugins>
  </build>
</profile>
```

alvarosanchez:
Fixes #4578

### DIFF
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -56,6 +56,11 @@ ext.extraPomInfo = {
                     delegate.version '${graal.version}'
                     scope "provided"
                 }
+                delegate.dependency {
+                    groupId "org.graalvm.nativeimage"
+                    artifactId "svm"
+                    scope "provided"
+                }
             }
             delegate.build {
                 delegate.plugins {
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -42,6 +42,40 @@ ext.extraPomInfo = {
                 'maven.compiler.release'('${release.version}')
             }
         }
+        delegate.profile {
+            id('graalvm')
+            delegate.activation {
+                delegate.file {
+                    exists('${env.JAVA_HOME}/bin/native-image')
+                }
+            }
+            delegate.dependencies {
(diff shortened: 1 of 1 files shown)


graemerocher:
One thing, maybe this should be back ported to 2.2.x?

alvarosanchez:
I purposely sent this to 2.3.x because it's a companion of https://github.com/micronaut-projects/micronaut-starter/pull/583, which is a new feature that should go into 2.3

graemerocher:
I see, and what are we doing for 2.2.x users who are confused by the problem with static resources?

alvarosanchez:
They need to select the `graalvm` feature.

PR #4633, opened 2020-11-30
Publish new snapshots following semantic versioning.

ilopmar:
See https://github.com/micronaut-projects/github-actions/pull/7


ilopmar:
Once this is merged I'll send a PR to Starter 2.3.x to use the new snapshot.

graemerocher:
Should probably send a PR for the 2.2.x branch has well

ilopmar:
I think changing that in 2.2.x can be a breaking change in a way that users using snapshots from that branch won't get updated versions anymore. That's why I target 2.3.x branch in this PR.

PR #4639, opened 2020-12-01
Ensure interface based config props can be validated

### LINKED ISSUE #4606: Interface based @EachProperty with validation doesn't work
If you use `@EachPropery` on an interface and apply validation rules the configuration fails to load this is because validation is applied before the bean is registered with the context as a singleton

### Steps to Reproduce

Define:

```java

import io.micronaut.context.annotation.EachProperty;

import javax.validation.constraints.Min;
import javax.validation.constraints.NotBlank;

@EachProperty("owners")
public interface OwnerConfiguration {
    @NotBlank
    String getName();
    @Min(18)
    int getAge();
}
```

And add configuration properties:

```
    public Map<String, String> getProperties() {
        return Map.of(
                "owners.fred.name", "Fred",
                "owners.fred.age", "35",
                "owners.barney.name", "Barney",
                "owners.barney.age", "30"
        );
    }
```

### Expected Behaviour

Should work

### Actual Behaviour

Fails with:

```
Caused by: io.micronaut.core.value.PropertyNotFoundException: No property found for name [owners.*.name] and type [java.lang.String]
	at io.micronaut.runtime.context.env.ConfigurationIntroductionAdvice.lambda$intercept$1(ConfigurationIntroductionAdvice.java:109)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at 
```

### Environment Information

- **Micronaut Version:** 2.2.0

graemerocher:
Fixes #4606

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/configproperties/itfce/MyConfig.java
+++ b/inject-java/src/test/groovy/io/micronaut/inject/configproperties/itfce/MyConfig.java
@@ -1,0 +1,13 @@
+package io.micronaut.inject.configproperties.itfce;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.context.annotation.Requires;
+
+import javax.validation.constraints.NotBlank;
+
+@ConfigurationProperties("my.config")
+@Requires(property = "my.config")
+public interface MyConfig {
+    @NotBlank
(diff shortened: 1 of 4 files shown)


graemerocher:
@jameskleeh this one is ready to review/merge too

PR #4644, opened 2020-12-01
Standard behavior for ConstraintViolationException

### LINKED ISSUE #4616: ConstraintExceptionHandler confuse behavior for 1 and N violations
### Task List

Hello!

Long time ago there were some changes on the [ConstraintExceptionHandler](https://github.com/micronaut-projects/micronaut-core/blob/2.2.x/validation/src/main/java/io/micronaut/validation/exceptions/ConstraintExceptionHandler.java) class through the issue #980.

The confusing part is how the handler creates a response when only 1 rule is violated and how different it is when N rules are violated. These 2 ways to handle the same issue (request rules violation) causes an overhead on the frontend, because the frontend now have to understand these 2 types of violation output. I understand that the frontend must do the same validation on the client side and not rely just on the backend validation, but a backend is not only consumed by the frontend. Another backend, third party services and so on will have to be prepared for these 2 types of bad request output.

To explain what is going on with some examples, let's use a create user request with some validation:

```java
@Introspected
public class UserCreateRequest {

    @NotBlank
    private String name;

    @NotBlank
    private String email;
}
```

If I send a request without the `name` field I get this response:

```json
{
  "message": "userCreateRequest.name: must not be blank",
  "_links": {
    "self": {
      "href": "/users",
      "templated": false
    }
  }
}
```

If I send a request without the `name` and the `email` fields, I get this response:

```json
{
  "message": "Bad Request",
  "_links": {
    "self": {
      "href": "/users",
      "templated": false
    }
  },
  "_embedded": {
    "errors": [
      {
        "message": "userCreateRequest.email: must not be blank"
      },
      {
        "message": "userCreateRequest.name: must not be blank"
      }
    ]
  }
}
```

So, in the first response the `message` output has the rule violation message, but in the second response the `message` is a default bad request message, and the rules violations are instead on the `_embedded.errors` field.

Is there a reason for the first output example don't be like this?

```json
{
  "message": "Bad Request",
  "_links": {
    "self": {
      "href": "/users",
      "templated": false
    }
  },
  "_embedded": {
    "errors": [
      {
        "message": "userCreateRequest.name: must not be blank"
      }
    ]
  }
}
```

No matter if 1 rule is violated or N rules are violated, the response schema would be the same.

I understand that we can create our own ExceptionHandler to override the default ConstraintViolationException, but shouldn't this be the default behavior for micronaut?

The ConstraintExceptionHandler.handle could be changed to something like this:

```java
@Override
public HttpResponse<JsonError> handle(HttpRequest request, ConstraintViolationException exception) {
    Set<ConstraintViolation<?>> constraintViolations = exception.getConstraintViolations();

    JsonError error;

    if (exception.getMessage() == null) {
        error = new JsonError(HttpStatus.BAD_REQUEST.getReason());
    } else {
        error = new JsonError(exception.getMessage());
    }

    if (constraintViolations != null && !constraintViolations.isEmpty()) {
        List<Resource> errors = new ArrayList<>();

        for (ConstraintViolation<?> violation : constraintViolations) {
            errors.add(new JsonError(buildMessage(violation)));
        }

        error.embedded("errors", errors);
    }

    error.link(Link.SELF, Link.of(request.getUri()));

    return HttpResponse.badRequest(error);
}
```

### Steps to Reproduce

1. Send a request with 1 violated rule
1. Send a request with 2+ violated rules

### Expected Behaviour

To receive the same JSON schema showing the violated rule or rules.

### Actual Behaviour

One dedicated JSON for 1 violated rule and another dedicated JSON for 2+ violated rules.

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** 2.2.0
- **JDK Version:** 11

rafaelrenanpacheco:
To avoid breaking changes, this PR creates a new property:
- `jackson.always-serialize-errors-as-list`

By default it will be `false`, resulting in no changes on how `ConstraintExceptionHandler` and `OptionalValuesSerializer` serializes `_embedded.errors`.

On the other hand, if set to `true`, there will be a standard behavior for the `_embedded.errors` serialization.
So, when `ConstraintExceptionHandler` is thrown, all violation messages will be serialized in the `_embedded.errors` array, like this:

```json
{
  "message": "Bad Request",
  "_links": {
    "self": {
      "href": "/users",
      "templated": false
    }
  },
  "_embedded": {
    "errors": [
      {
        "message": "userCreateRequest.name: must not be blank"
      }
    ]
  }
}
```

```json
{
  "message": "Bad Request",
  "_links": {
    "self": {
      "href": "/users",
      "templated": false
    }
  },
  "_embedded": {
    "errors": [
      {
        "message": "userCreateRequest.name: must not be blank"
      },
      {
        "message": "userCreateRequest.email: must not be blank"
      }
    ]
  }
}
```

For more details take a loot at #4616.

- Fixes #4616

### DIFF
--- a/runtime/src/main/java/io/micronaut/jackson/JacksonConfiguration.java
+++ b/runtime/src/main/java/io/micronaut/jackson/JacksonConfiguration.java
@@ -329,24 +329,6 @@ public class JacksonConfiguration {
         this.hateoas = hateoas;
     }
 
-    @ConfigurationProperties("hateoas")
-    @Requires("jackson.hateoas")
-    public static class Hateoas {
-        /**
-         * If true _embedded.errors will always be an array. When set to false, _embedded.errors will be serialized
-         * as an object for 1 error or serialized as an array for 2 or more errors.
-         */
-        private boolean alwaysSerializeErrorsAsList = false;
-
-        public boolean isAlwaysSerializeErrorsAsList() {
-            return alwaysSerializeErrorsAsList;
-        }
-
-        public void setAlwaysSerializeErrorsAsList(boolean alwaysSerializeErrorsAsList) {
-            this.alwaysSerializeErrorsAsList = alwaysSerializeErrorsAsList;
-        }
-    }
-
     /**
      * Constructors a JavaType for the given argument and type factory.
      * @param type The type
@@ -396,5 +378,23 @@ public class JacksonConfiguration {
             }
         }
         return javaTypes.toArray(new JavaType[0]);
+    }
+
+    @ConfigurationProperties("hateoas")
+    @Requires("jackson.hateoas")
+    public static final class Hateoas {
+        /**
+         * If true _embedded.errors will always be an array. When set to false, _embedded.errors will be serialized
+         * as an object for 1 error or serialized as an array for 2 or more errors.
+         */
+        private boolean alwaysSerializeErrorsAsList = false;
+
+        public boolean isAlwaysSerializeErrorsAsList() {
+            return alwaysSerializeErrorsAsList;
+        }
+
+        public void setAlwaysSerializeErrorsAsList(boolean alwaysSerializeErrorsAsList) {
+            this.alwaysSerializeErrorsAsList = alwaysSerializeErrorsAsList;
+        }
     }
 }
--- a/runtime/src/main/java/io/micronaut/jackson/JacksonConfiguration.java
+++ b/runtime/src/main/java/io/micronaut/jackson/JacksonConfiguration.java
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 import io.micronaut.context.annotation.ConfigurationProperties;
-import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Experimental;
 import io.micronaut.core.annotation.TypeHint;
 import io.micronaut.core.type.Argument;
@@ -31,8 +30,6 @@ import io.micronaut.core.util.CollectionUtils;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 
-import javax.annotation.Nullable;
-import javax.inject.Inject;
 import java.util.*;
 
 /**
@@ -79,7 +76,7 @@ public class JacksonConfiguration {
     private JsonInclude.Include serializationInclusion = JsonInclude.Include.NON_EMPTY;
     private ObjectMapper.DefaultTyping defaultTyping = null;
     private PropertyNamingStrategy propertyNamingStrategy = null;
-    private Hateoas hateoas;
+    private boolean alwaysSerializeErrorsAsList = false;
 
     /**
      * Whether the {@link io.micronaut.core.beans.BeanIntrospection} should be used for reflection free object serialialization/deserialialization.
@@ -202,10 +199,13 @@ public class JacksonConfiguration {
     }
 
     /**
-     * @return Settings for the hateoas configuration
+     * Whether _embedded.errors should always be serialized as list. If set to false, _embedded.errors
+     * with 1 element will be serialized as an object.
+     *
+     * @return True if _embedded.errors should always be serialized as list.
      */
-    public Hateoas getHateoas() {
-        return hateoas;
+    public boolean isAlwaysSerializeErrorsAsList() {
+        return alwaysSerializeErrorsAsList;
     }
 
     /**
@@ -320,13 +320,13 @@ public class JacksonConfiguration {
     }
 
     /**
-     * Sets the property for hateoas configuration.
+     * Sets whether _embedded.errors should always be serialized as list (defaults to false).
+     * If set to false, _embedded.errors with 1 element will be serialized as an object.
      *
-     * @param hateoas The hateoas configuration
+     * @param alwaysSerializeErrorsAsList True if _embedded.errors should always be serialized as list.
      */
-    @Inject
-    void setHateoas(@Nullable Hateoas hateoas) {
-        this.hateoas = hateoas;
+    public void setAlwaysSerializeErrorsAsList(boolean alwaysSerializeErrorsAsList) {
+        this.alwaysSerializeErrorsAsList = alwaysSerializeErrorsAsList;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


graemerocher:
This could be regarded as a breaking change for users that parse errors although it is clearly a more consistent fix. Thoughts @jameskleeh ?

jameskleeh:
@graemerocher I think its better to gate this behind a config option and then remove the option or change the default for 3.0

graemerocher:
Right

rafaelrenanpacheco:
Hello!

In order to gate this behind a config option (application.properties), is there an specific way to read this config from the micronaut-core point of view?

I know how to read properties from an user point of view, a simple `@Value` or property mapper can achieve that, but I don't know if this is the expected way to do on micronaut-core.

jameskleeh:
@rafaelrenanpacheco We use `@ConfigurationProperties` beans to read config

rafaelrenanpacheco:
Thanks @jameskleeh!

I'll look for some examples using `@ConfigurationProperties` and update this PR later today.

rafaelrenanpacheco:
I think it is ready.

sweating

validation/src/main/java/io/micronaut/validation/exceptions/ConstraintExceptionHandler.java
```diff
@@ -46,14 +47,20 @@
 @Requires(classes = {ConstraintViolationException.class, ExceptionHandler.class})
 public class ConstraintExceptionHandler implements ExceptionHandler<ConstraintViolationException, HttpResponse<JsonError>> {
 
+    private final JacksonConfiguration jacksonConfiguration;
+
+    public ConstraintExceptionHandler(JacksonConfiguration jacksonConfiguration) {
+        this.jacksonConfiguration = jacksonConfiguration;
+    }
```
graemerocher: To avoid a breaking API change please restore the no argument constructor and annotate this one with `@Inject`

runtime/src/main/java/io/micronaut/jackson/serialize/OptionalValuesSerializer.java
```diff
@@ -35,6 +37,12 @@
 @Singleton
 public class OptionalValuesSerializer extends JsonSerializer<OptionalValues<?>> {
 
+    private final JacksonConfiguration jacksonConfiguration;
+
+    public OptionalValuesSerializer(JacksonConfiguration jacksonConfiguration) {
+        this.jacksonConfiguration = jacksonConfiguration;
+    }
```
graemerocher: To avoid a breaking API change please restore the no argument constructor and annotate this one with `@Inject`

runtime/src/main/java/io/micronaut/jackson/serialize/OptionalValuesSerializer.java
```diff
@@ -64,4 +73,8 @@ public void serialize(OptionalValues<?> value, JsonGenerator gen, SerializerProv
         }
         gen.writeEndObject();
     }
+
+    private boolean canSerializeElementAsObject(Class<?> type) {
+        return type != JsonError.class || !jacksonConfiguration.getHateoas().isAlwaysSerializeErrorsAsList();
```
graemerocher: Please assign `jacksonConfiguration.getHateoas().isAlwaysSerializeErrorsAsList()` to a final local field to avoid repeated method calls

validation/src/main/java/io/micronaut/validation/exceptions/ConstraintExceptionHandler.java
```diff
@@ -46,14 +47,20 @@
...
+
     @Override
     public HttpResponse<JsonError> handle(HttpRequest request, ConstraintViolationException exception) {
         Set<ConstraintViolation<?>> constraintViolations = exception.getConstraintViolations();
         if (constraintViolations == null || constraintViolations.isEmpty()) {
             JsonError error = new JsonError(exception.getMessage() == null ? HttpStatus.BAD_REQUEST.getReason() : exception.getMessage());
             error.link(Link.SELF, Link.of(request.getUri()));
             return HttpResponse.badRequest(error);
-        } else if (constraintViolations.size() == 1) {
+        } else if (constraintViolations.size() == 1 && !jacksonConfiguration.getHateoas().isAlwaysSerializeErrorsAsList()) {
```
graemerocher: Please assign `jacksonConfiguration.getHateoas().isAlwaysSerializeErrorsAsList()` to a final local field to avoid repeated method calls

runtime/src/main/java/io/micronaut/jackson/JacksonConfiguration.java
```diff
@@ -357,4 +379,22 @@ public void setPropertyNamingStrategy(PropertyNamingStrategy propertyNamingStrat
...
+        private boolean alwaysSerializeErrorsAsList = false;
+
+        public boolean isAlwaysSerializeErrorsAsList() {
+            return alwaysSerializeErrorsAsList;
+        }
+
+        public void setAlwaysSerializeErrorsAsList(boolean alwaysSerializeErrorsAsList) {
+            this.alwaysSerializeErrorsAsList = alwaysSerializeErrorsAsList;
+        }
+    }
```
graemerocher: I don't think we need a whole new nested config props just for this. Adding the field at the top level is fine

graemerocher (changes requested):

validation/src/main/java/io/micronaut/validation/exceptions/ConstraintExceptionHandler.java
```diff
@@ -46,14 +48,19 @@
 @Requires(classes = {ConstraintViolationException.class, ExceptionHandler.class})
 public class ConstraintExceptionHandler implements ExceptionHandler<ConstraintViolationException, HttpResponse<JsonError>> {
 
+    @Inject
+    private JacksonConfiguration jacksonConfiguration;
```
graemerocher: private field injection requires reflection, please change this to a field representing the configuration and use constructor injection and then add an additional constructor to maintain compatibility. Something like:

```java
private final boolean alwaysSerializeErrorsAsList;

@Inject
public ConstraintExceptionHandler(JacksonConfiguration jacksonConfiguration) {
       this. alwaysSerializeErrorsAsList = jacksonConfiguration.isAlwaysSerializeErrorsAsList();
}

public ConstraintExceptionHandler() {
      this.alwaysSerializeErrorsAsList = true
}
```

runtime/src/main/java/io/micronaut/jackson/serialize/OptionalValuesSerializer.java
```diff
@@ -35,13 +38,18 @@
 @Singleton
 public class OptionalValuesSerializer extends JsonSerializer<OptionalValues<?>> {
 
+    @Inject
+    private JacksonConfiguration jacksonConfiguration;
```
graemerocher: private field injection requires reflection, please change this to a field representing the configuration and use constructor injection and then add an additional constructor to maintain compatibility. Something like:

```java
private final boolean alwaysSerializeErrorsAsList;

@Inject
public OptionalValuesSerializer(JacksonConfiguration jacksonConfiguration) {
       this. alwaysSerializeErrorsAsList = jacksonConfiguration.isAlwaysSerializeErrorsAsList();
}

public OptionalValuesSerializer() {
      this.alwaysSerializeErrorsAsList = true
}
```

graemerocher (changes requested):

graemerocher:
Thanks for the contribution!

PR #4677, opened 2020-12-07
Fix misleading error 'empty body' for responses with a body that could not be decoded. 

### LINKED ISSUE #4613: HttpClient falsely claiming "Empty body" when server gives no content-type
### Task List

### Description
Take an  armpit rolled server which provides **no Content-type header** although responses bodies are correct json payloads. We had this with a heavily customized Magento1 instance recently.

In that case Micronaut falsely complains the body is empty which is a source of confusion at the very least.

The reason is the Micronaut client has no clue as of how to decode the response without a content-type, so the resulting decoded object ends up empty, which is the reason for the exception message.
The exception should be triggered upfront as soon as it's seen there is no suitable decoder for a given payload.
Or maybe have a way to specify a "fallback" content-type.

Workarounds: ask the server maintainers to return a content-type (which was not possible in our case)
or retrieve as String and decode by hand with Jackson. Yay.

### Steps to Reproduce

1. Checkout https://github.com/fabmars/no-content-type-issue
2. Run /server/server-bad.sh which fires up a custom nginx which never returns any content-type header.
3. Run NoContentTypeTest unit test

You may also reproduce those steps using /server/server-good.json for a "nominal" behavior (with a content-type)

### Expected Behaviour

At the very least a consistent HttpClientResponseException message (eg: _"unable to decode body for content type {}"_)
Or a fallback (to decoding application/json for example)

### Actual Behaviour

HttpClientResponseException with misleading "Empty body" message.

### Environment Information

- **Operating System**: Ubuntu 20.04
- **Micronaut Version:** 2.2.0 and 1.3.7
- **JDK Version:** 1.8

### Example Application

https://github.com/fabmars/no-content-type-issue

 ### Stacktrace
[stacktrace.log](https://github.com/micronaut-projects/micronaut-core/files/5600134/stacktrace.log)

jameskleeh:
Fixes #4613

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/BlockingHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/BlockingHttpClient.java
@@ -16,6 +16,7 @@
 package io.micronaut.http.client;
 
 import io.micronaut.core.annotation.Blocking;
+import io.micronaut.core.io.buffer.ByteBuffer;
 import io.micronaut.core.type.Argument;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpResponse;
@@ -23,6 +24,7 @@ import io.micronaut.http.HttpStatus;
 import io.micronaut.http.client.exceptions.HttpClientResponseException;
 
 import java.io.Closeable;
+import java.util.Optional;
 
 /**
  * A blocking HTTP client interface that features a subset of the operations provided by {@link HttpClient} and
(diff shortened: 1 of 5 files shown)


http-client-core/src/main/java/io/micronaut/http/client/BlockingHttpClient.java
```diff
@@ -140,12 +142,19 @@
...
         } else {
-            return response
-                    .getBody()
-                    .orElseThrow(() -> new HttpClientResponseException(
-                            "Empty body",
-                            response
-                    ));
+            Optional<O> body = response.getBody();
+            Optional<byte[]> bodyBytes = response.getBody(Argument.of(byte[].class));
+            if (!body.isPresent() && bodyBytes.isPresent()) {
```
graemerocher: Should be:

```java
if (!body.isPresent() && response.getBody(Argument.of(byte[].class)).isPresent()) {
```

To avoid the extra object creation and method call

graemerocher (changes requested):

PR #4679, opened 2020-12-07
Trim strings when deserializing

### LINKED ISSUE #4675: Feature request - Trim request strings
Hello!

First of all, is there a way to trim string fields from requests globally? Usually there is no reason to accept a string field with blank spaces before and after the string, which leads to a codebase full of this:

```java
String email = userCreateRequest.getEmail().trim();
String name = userCreateRequest.getName().trim();
```

The above case is null safe because the request payload has the `@NotBlank` rule. Otherwise it would require a null check or the use of some helper like `trimToNull`.

Some frameworks like fastify or laravel allow to trim all strings through middleware, which leads to a cleaner codebase.

If there is no such thing in Micronaut, did you consider some kind of strategy to handle this "issue"? For example:

**1 - Global property**
- Add to `application.properties` something like: `micronaut.mvc.request.strings=trim` or `micronaut.mvc.request.strings=trim-to-null`.

**2 - Request type annotation**
```java
@Introspected
@Trimmed(toNull = false)
public class UserCreateRequest {

    @NotBlank
    private String name;

    @NotBlank
    @Email
    private String email;
}
```

**3 - Request field annotation**
```java
@Introspected
public class UserCreateRequest {

    @NotBlank
    @Trimmed
    private String name;

    @NotBlank
    @Email
    @Trimmed
    private String email;
}
```

If this kind of feature isn't suitable for micronaut-core, what's the best way to achieve this from an user point of view? If it is a good feature to have in micronaut, I can help with a PR, just need some lead where to code this feature.

rafaelrenanpacheco:
Hello!

This PR creates a new property that enables an automatic trim on strings during deserialization.

If `jackson.trim-strings` is set to `true`, then the following payload

```json
{
  "id": 1,
  "email": " abc ",
  "name": " Micronaut ",
  "city": " ",
  "active": true,
  "createdAt": "2020-01-31"
}
```

will have the strings trimmed to null before attributing them to a pojo. It would be the equivalent of sending the following request:

```json
{
  "id": 1,
  "email": "abc",
  "name": "Micronaut",
  "city": null,
  "active": true,
  "createdAt": "2020-01-31"
}
```

By default this feature is disabled. If trim is desired, then `jackson.trim-strings` must be set to `true`.

Thanks @jameskleeh for the code snippet.

- Fixes #4675


rafaelrenanpacheco:
I had to add the following dependencies to test the string trim through a json request to a controller:

```groovy
testImplementation project(":http-server-netty")
testImplementation project(":http-client")
```

This broke the [ExecutorServiceConfigSpec](https://github.com/micronaut-projects/micronaut-core/blob/2.2.x/runtime/src/test/groovy/io/micronaut/runtime/executor/ExecutorServiceConfigSpec.groovy) tests:

- `executorServices.size() == 4` is now `5`
- `ctx.getBean(ThreadPoolExecutor, Qualifiers.byName("one"))` don't find any bean with this name.

Any hints on what went wrong here?

runtime/src/main/java/io/micronaut/jackson/deserialize/StringTrimModule.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.databind.deser.std.StringDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.core.util.StringUtils;
+
+import javax.inject.Singleton;
+import java.io.IOException;
+
+@Singleton
+@Requires(property = "jackson.trim-strings", value = StringUtils.TRUE)
```
jameskleeh: There needs to be a field/getter/setter for this in `JacksonConfiguration`

runtime/src/main/java/io/micronaut/jackson/deserialize/StringTrimModule.java
```diff
@@ -0,0 +1,43 @@
...
+public class StringTrimModule extends SimpleModule {
+
+    public StringTrimModule() {
+        addDeserializer(String.class, new StringDeserializer() {
```
jameskleeh: This should probably be done within the `ObjectMapperFactory` and then you can check for `jacksonConfiguration.getTrimStrings()`

jameskleeh (changes requested):
Instead of having a Pojo and adding dependencies to the server and client, simply add a test in `JacksonSetupSpec` that passes in JSON with strings and assert they are trimmed

rafaelrenanpacheco:
Hello @jameskleeh, thanks for the suggestions, things are much simpler now.

But, there is a problem.

The deserializer is being beaned and being [picked here](https://github.com/micronaut-projects/micronaut-core/blob/186d434c66e3d9215a40de6647ccf513b7b54907/runtime/src/main/java/io/micronaut/jackson/ObjectMapperFactory.java#L120).

The problem is, this deserializer is not being using during [this test](https://github.com/micronaut-projects/micronaut-core/blob/9f8b6814eea60edffcc2d3a0582ee74da7f1108d/runtime/src/test/groovy/io/micronaut/jackson/JacksonSetupSpec.groovy#L136-L151).

I tried extending `StdDeserializer`, `JsonDeserializer`, and even using `extends SimpleModule`, the latter being [picked here](https://github.com/micronaut-projects/micronaut-core/blob/186d434c66e3d9215a40de6647ccf513b7b54907/runtime/src/main/java/io/micronaut/jackson/ObjectMapperFactory.java#L90), but none of them actually run the `deserialize()` method.

I tried to debbug the `ObjectMapper` as well, but didn't find any place that tries to choose these deserializers, even tho they are configured through the `ObjectMapperFactory`.

There is a `BigDecimalDeserializer extends JsonDeserializer<BigDecimal>` in the micronaut-core that is just a `@Singleton` and I suppose it works just fine.

Any clue on what I might be missing?

Thanks.

jameskleeh:
@rafaelrenanpacheco I took the PR and moved it forward here https://github.com/micronaut-projects/micronaut-core/pull/4696

rafaelrenanpacheco:
Thanks @jameskleeh!

PR #4693, opened 2020-12-09
Upgrade to Netty 4.1.55.Final

ilopmar:
Without the extra condition in the `if` [this test](https://github.com/micronaut-projects/micronaut-core/blob/29adc6a8b94bffc131ca12bde21705e362578b70/http-client/src/test/groovy/io/micronaut/http/client/DataStreamSpec.groovy#L183) fails with a NPE [here](https://github.com/micronaut-projects/micronaut-core/blob/ebc39d4f6e3872d030763b3d5d36913262b4a062/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpDataReference.java#L130).

I'm not sure why it fails. This are [the release notes](https://netty.io/news/2020/12/08/4-1-55-Final.html) for the new Netty version, I haven't found anything that can be related.

Maybe @jameskleeh or @graemerocher want to take a look before merging this.

P.S: GraalVM integration works.


jameskleeh:
Can't upgrade due to https://github.com/netty/netty/issues/10851

PR #4717, opened 2020-12-15
Locale resolver

jameskleeh:
Close: https://github.com/micronaut-projects/micronaut-core/issues/3787

PR #4727, opened 2020-12-17
refactor: changes to locale resolver

### LINKED ISSUE #3787: LocaleResolver bean
I think it would be good to create a `LocaleResolver` contract to help with Locale resolution. 

Multiple implementations should be provided. Some for resolution within an HTTP Request (Accept Header, Cookie, Session) but also for environments such as Alexa Skills. 

Relates to: https://github.com/micronaut-projects/micronaut-core/issues/2617

sdelamo:
This PR address most of the requested changes to https://github.com/micronaut-projects/micronaut-core/pull/4717

- I have fixed the missing Javadoc. 
- I have implemented a test for `SessionLocaleResolver` and added extra tests.
- I moved the main logic for default and fixed resolution outside of the `http-server` module. So that they can be used without an `http-server`. 
- I created configuration interfaces so that parts of configuration can be used outside HttpServerConfiguration. 
- I have sorted the locale resolvers in `CompositeLocaleResolver`. 
- I moved locale resolution into its own packages. 

See: https://github.com/micronaut-projects/micronaut-core/issues/3787


core/src/main/java/io/micronaut/core/util/localeresolution/FixedLocaleResolver.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class FixedLocaleResolver<T> extends DefaultLocaleResolver<T> {
+
+    protected final Locale locale;
+
+    /**
+     *
+     * @param localeResolutionConfiguration Locale Resolution configuration
+     */
+    public FixedLocaleResolver(LocaleResolutionConfiguration localeResolutionConfiguration) {
```
jameskleeh: if this is a standalone class it doesn't make sense to tie it to configuration. It should simply receive the locale as the argument

core/src/main/java/io/micronaut/core/util/localeresolution/LocaleResolutionConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Locale;
+import java.util.Optional;
+
+/**
+ * Locale resolution configuration.
+ *
+ * @author Sergio del Amo
+ * @since 2.3.0
+ */
+public interface LocaleResolutionConfiguration {
```
jameskleeh: I don't think we need a separate interface for this

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
```diff
@@ -13,31 +13,41 @@
...
 
+    /**
+     * @param localeResolvers HTTP Locale Resolvers
+     * @param httpLocaleResolutionConfiguration Locale Resolution configuration for HTTP Requests
+     */
     public CompositeHttpLocaleResolver(HttpLocaleResolver[] localeResolvers,
-                                       HttpServerConfiguration serverConfiguration) {
+                                       HttpLocaleResolutionConfiguration httpLocaleResolutionConfiguration) {
+        super(httpLocaleResolutionConfiguration);
+        OrderUtil.sort(localeResolvers);
```
jameskleeh: This is redundant

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
```diff
@@ -50,7 +60,7 @@ public CompositeHttpLocaleResolver(HttpLocaleResolver[] localeResolvers,
     }
 
     @Override
-    public Locale resolveOrDefault(@NonNull HttpRequest<?> request) {
-        return resolve(request).orElse(defaultLocale);
+    public int getOrder() {
+        return Ordered.LOWEST_PRECEDENCE;
```
jameskleeh: Setting an order here doesn't make sense because this is the primary bean

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/HttpFixedLocaleResolver.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * @param localeResolutionConfiguration Locale Resolution configuration
+     */
+    public HttpFixedLocaleResolver(LocaleResolutionConfiguration localeResolutionConfiguration) {
+        super(localeResolutionConfiguration);
+    }
+
+    @Override
+    public int getOrder() {
+        return Ordered.HIGHEST_PRECEDENCE;
```
jameskleeh: We shouldn't use HIGHEST_PRECEDENCE because then its impossible to create one with a higher precedence

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/HttpLocaleResolutionConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Optional;
+
+/**
+ * Configuration for Locale Resolution in a HTTP Request.
+ *
+ * @author Sergio del Amo
+ * @since 2.3.0
+ */
+@Indexed(HttpLocaleResolutionConfiguration.class)
```
jameskleeh: This should not be indexed. There will only be 1 bean of this type

jameskleeh (changes requested):

core/src/main/java/io/micronaut/core/util/localeresolution/DefaultLocaleResolver.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Locale;
+
+/**
+ * Provides an abstract class which implements {@link LocaleResolver} and handles default locale resolution.
+ *
+ * @author Sergio del Amo
+ * @since 2.3.0
+ * @param <T> The context object which will be used to resolve the locale
+ */
+public abstract class DefaultLocaleResolver<T> implements LocaleResolver<T> {
```
jameskleeh: This should be `AbstractLocaleResolver`. Calling it default means to me that its the default implementation, which isn't true

core/src/main/java/io/micronaut/core/util/localeresolution/LocaleResolutionConfiguration.java
sdelamo: I think it is best to abstract with an interface. Then in an application, which does not use the http-server module (E.g. Alexa), I could implement the interface and support fixed locale and default resolution without effort.

jameskleeh: You an also do so without effort without this interface. The fixed and default can be direct args instead of having a separate interface

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
sdelamo: But if you inject every `HttpLocaleResolver` with `Collection<LocaleResolver>` and you sort the
 resolver,  you would want the `CompositeHttpLocaleResolver` to have the lowest order. Specially, since we don't provide a way to disable `CompositeHttpLocaleResolver`.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/HttpLocaleResolutionConfiguration.java
sdelamo: Is not `@Indexed` necessary because there might be multiple beans of type `LocaleResolutionConfiguration`?

core/src/main/java/io/micronaut/core/util/localeresolution/FixedLocaleResolver.java
sdelamo: The configuration is the contract, the interface. It is not a `@ConfigurationProperties` implementation. It needs `LocaleResolutionConfiguration` to pass it to the parent class.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
jameskleeh: There is no use case to inject all locale resolvers outside of the composite one. The composite one here is specific to HTTP so there shouldn't be a need to disable it.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/HttpLocaleResolutionConfiguration.java
jameskleeh: The chance of multiple beans here is very low and the there is a cost to indexed so it shouldn't be used in the case of 1 or 2 beans.

core/src/main/java/io/micronaut/core/util/localeresolution/DefaultLocaleResolver.java
sdelamo: I've renamed it.

core/src/main/java/io/micronaut/core/util/localeresolution/FixedLocaleResolver.java
jameskleeh: I don't think having the configuration here as a contract makes sense. Fixed does not need a default because its fixed. There is no chance for it to not have a value

core/src/main/java/io/micronaut/core/util/localeresolution/LocaleResolutionConfiguration.java
sdelamo: If we have an interface, then it forces us to use the use same keys in configuration. 

I could have: 

```
alexa:
  skills:
    myskill:
      skill-id 'xxxx-yaaa-zz123' 
      locale-resolution: 
          default-locale: 'es_ES'
```      

Having the interface forces us to reuse the same `default-locale` key in both `http-server` configuration and other configurations.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
sdelamo: Are bean's arrays injected sorted?

core/src/main/java/io/micronaut/core/util/localeresolution/LocaleResolutionConfiguration.java
jameskleeh: If you want to leave the interface for use in other places thats OK, but the fixed resolver shouldn't require it. Users should be able to pull the data from anywhere and not have to conform to our contract

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
jameskleeh: Yes

sdelamo: I've removed the sorting.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/HttpFixedLocaleResolver.java
sdelamo: I've added a padding.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/HttpLocaleResolutionConfiguration.java
sdelamo: I've removed it.

http-server/src/main/java/io/micronaut/http/server/util/localeresolution/CompositeHttpLocaleResolver.java
sdelamo: I've removed the `getOrder`override. I still think it is more correct that `CompositeHttpLocaleResolver`
has the lowest order in a list of `LocaleResolver`,

core/src/main/java/io/micronaut/core/util/localeresolution/LocaleResolutionConfiguration.java
sdelamo: I've splitted the configuration in two interfaces. Now, I am injecting only the `fixed` configuration in the `fixed` resolver.

core/src/main/java/io/micronaut/core/util/localeresolution/FixedLocaleResolver.java
sdelamo: I've splitted the configuration in two interfaces. Now, I am injecting only the `fixed` configuration in the `fixed` resolver.

sdelamo: I've changed FixedLocaleResolver to take a Locale in the constructor.

PR #4755, opened 2020-12-27
Allow setting log level through environment variables

### LINKED ISSUE #4580: Log level using an environment variable ignored
The documentation [mentions](https://docs.micronaut.io/latest/guide/index.html#_controlling_log_levels_with_properties) it is possible to set property using environment variables.

Nevertheless, `LOGGER_LEVELS` nor `LOG_LEVEL` prefixes works maybe I'm missing something but the issue seems to come from the [PropertiesLoggingLevelsConfigurer line 81](https://github.com/micronaut-projects/micronaut-core/blob/8d8e1b680225ca2077f49283d731c327d5d4f559/runtime/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java#L81) class where the iteration is performed environment using the [StringConvention#RAW](https://github.com/micronaut-projects/micronaut-core/blob/8d8e1b680225ca2077f49283d731c327d5d4f559/core/src/main/java/io/micronaut/core/naming/conventions/StringConvention.java#L51) so environment variables delimited by underscores are excluded.

To reproduce you can checkout the [this project](https://github.com/equidis/micronaut-grpc-users-service) and try changing the logger `<logger name="com.github.jntakpe" level="DEBUG"/>` to INFO or whatever using an environment variable

ttzn:
Fixes #4580 

For some reason `PropertiesLoggingLevelsConfigurer` was only loading raw keys.

### DIFF
--- a/runtime/src/test/groovy/io/micronaut/logging/LogbackLogLevelConfigurerSpec.groovy
+++ b/runtime/src/test/groovy/io/micronaut/logging/LogbackLogLevelConfigurerSpec.groovy
@@ -16,6 +16,7 @@ class LogbackLogLevelConfigurerSpec extends Specification {
             ((Logger) LoggerFactory.getLogger('foo.bar1')).setLevel(Level.DEBUG)
             ((Logger) LoggerFactory.getLogger('foo.bar2')).setLevel(Level.DEBUG)
             ((Logger) LoggerFactory.getLogger('foo.bar3')).setLevel(Level.ERROR)
+            ((Logger) LoggerFactory.getLogger('foo.barBaz')).setLevel(Level.WARN)
 
         when:
             ApplicationContext context = ApplicationContext.run(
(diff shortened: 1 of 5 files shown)


jameskleeh:
This could have implications if the intent is to configure the log level of a non conventional package name. For example if someone had `com.myApp` as a package name, the resulting value would now be `com.my-app`.

graemerocher:
@jameskleeh Java package names should be lower case so I think that is non-issue

ttzn:
@jameskleeh good catch, this MR introduces a regression for those kinds of package names. Maybe we can load both raw keys for compatibility and generated keys to allow environment variables, with a little warning in the doc.

runtime/src/test/groovy/io/micronaut/logging/LogbackLogLevelConfigurerSpec.groovy
```diff
@@ -37,6 +40,32 @@ class LogbackLogLevelConfigurerSpec extends Specification {
             'foo.bar2'    | Level.INFO
             'foo.bar3'    | null
             'aaa.bbb.ccc' | Level.ERROR
+            'foo.barBaz'    | Level.WARN
```
jameskleeh: Shouldn't this be INFO?

ttzn: Absolutely, fixed

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #4756, opened 2020-12-28
Bump micronaut-views to version 2.2.0

laurentpellegrino:
This PR bumps the version of _micronaut-views_ to the latest one. There might be a need to check why @micronaut-build has not detected the new version. Unfortunately, I don't know about the Micronaut build pipeline to help regarding this last point.

Using _micronaut-bom_ 2.2.1 or 2.2.2, _micronaut-views_ in version 2.1.0 is fetched. It seems that @micronaut-build is not detecting updates for this module since months. That's an issue because, as a consequence, the transitive module _micronaut-views-soy_ is using a very old version of closure-templates.


jameskleeh:
Views 2.2 is slated for Micronaut 2.3 and is already included in that branch

PR #4759, opened 2020-12-29
Update javadoc for MutableConvertibleValues

vasac:
Looks like all values are being returned in the various implementations.


jameskleeh:
Thanks!

PR #4771, opened 2021-01-04
Allow Filter annotations to be stacked in a third one

### LINKED ISSUE #4767: Filter Annotations cannot be stacked on @Client (at least for HttpClientFilter)
### Task List

### Steps to Reproduce

1. Create two Filter Annotations
```java
@FilterMatcher
public @interface FilterA {}
@FilterMatcher
public @interface FilterB {}
```
2. Create two Filters implementations
```java
@FilterA
@Singleton
public class MyFilterA implements HttpClientFilter {
	@Override
	public Publisher<? extends HttpResponse<?>> doFilter(MutableHttpRequest<?> request, ClientFilterChain chain) {
		// do something
                return chain.proceed(request);
	}
}
@FilterB
@Singleton
public class MyFilterB implements HttpClientFilter {
	@Override
	public Publisher<? extends HttpResponse<?>> doFilter(MutableHttpRequest<?> request, ClientFilterChain chain) {
		// do something else
                return chain.proceed(request);
	}
}
```
3. Create an annotation to group the two previous one
```java
@FilterA
@FilterB
public @interface AllFilters {}
```

4. Create a Client
```java
@AllFilters
@Client("/path")
public interface FiltersNotWorkingDemoClient {}
```
### Expected Behaviour

Filters should be associated with the client and applied

### Actual Behaviour

Filters are only applied if both individually declared
```java
@FilterA
@FilterB
@Client("/path")
public interface FiltersWorkingDemoClient {}
```
### Environment Information
Not relevant

### Example Application

- demo here : https://github.com/ggerbaud/mn-filter-demo

### Solution
It can be solved (if it should) at this line : https://github.com/micronaut-projects/micronaut-core/blob/2.2.x/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java#L88
replacing
```java
matches = context.getAnnotationMetadata().hasAnnotation(filterAnnotation);
```
by
```java
 matches = context.getAnnotationMetadata().getAnnotationNamesByStereotype(FilterMatcher.class).contains(filterAnnotation);
```

What do you think ?

ggerbaud:
Fixes #4767

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
@@ -85,7 +85,7 @@ public class DefaultHttpClientFilterResolver implements HttpClientFilterResolver
                     boolean matches = !annotationMetadata.hasStereotype(FilterMatcher.class);
                     String filterAnnotation = annotationMetadata.getAnnotationNameByStereotype(FilterMatcher.class).orElse(null);
                     if (filterAnnotation != null && !matches) {
-                        matches = context.getAnnotationMetadata().hasAnnotation(filterAnnotation);
+                        matches = context.getAnnotationMetadata().getAnnotationNamesByStereotype(FilterMatcher.class).contains(filterAnnotation);
                     }
 
                     if (matches) {
--- a/http-client/src/test/groovy/io/micronaut/http/client/filter/AnotherMarkerStereotypeAnnotation.java
+++ b/http-client/src/test/groovy/io/micronaut/http/client/filter/AnotherMarkerStereotypeAnnotation.java
@@ -15,12 +15,18 @@
  */
 package io.micronaut.http.client.filter;
 
+import io.micronaut.context.annotation.AliasFor;
+import io.micronaut.http.HttpMethod;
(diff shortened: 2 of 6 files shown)


jameskleeh:
@ggerbaud This needs a test. Should probably live in HttpFilterSpec

ggerbaud:
@jameskleeh Ok, so I add some tests and dit the changes on client and server filters

jameskleeh:
@ggerbaud Appreciate the effort here! I think the test needs to also reflect the original issue, which was the use of multiple annotations on the container annotation.

ggerbaud:
@jameskleeh My issue was related to a use case (grouping multiple filters behind one annotation to simplify thier declaration) but the test illustrated the technical reason (filters were searched only in direct annotations, not through stereotypes).
But I add more tests to illustrate all of this. 

I also add tests with two annotations filters directly on the client.
Note that in this case, annotation declaration order matters because the clients are the same except for filters. If they have the same first filter, they end up with the same key, due to [this line](https://github.com/micronaut-projects/micronaut-core/blob/2.2.x/http-client/src/main/java/io/micronaut/http/client/netty/RxNettyHttpClientRegistry.java#L407).

jameskleeh:
Thank you for your contribution!

PR #4780, opened 2021-01-06
adding optional idle timeout to http client connection pool channels

drmaas:


http-client/src/main/java/io/micronaut/http/client/netty/IdleTimeoutHandler.java
```diff
@@ -0,0 +1,45 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandler;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+
+/**
+ * This class is responsible for detecting idle timeout events, upon which the channel in the pool is closed.
+ */
+@ChannelHandler.Sharable
+public final class IdleTimeoutHandler extends ChannelDuplexHandler {
```
graemerocher: Missing `@author` and `@since` tags and I think this class can be package/private and annotated with `@Internal`

http-client/src/main/java/io/micronaut/http/client/netty/IdlingConnectionHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandler;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.util.ReferenceCountUtil;
+
+/**
+ * This handler prevents reading a channel when it is not being used by the connection pool.
+ */
+@ChannelHandler.Sharable
+public final class IdlingConnectionHandler extends ChannelInboundHandlerAdapter {
```
graemerocher: Missing `@author` and `@since` tags and I think this class can be package/private and annotated with `@Internal`

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -2579,6 +2589,19 @@ public void channelReleased(Channel ch) {
                     }
                 }
             }
+
+            @Override
+            public void channelAcquired(Channel ch) throws Exception {
+                if (ch.pipeline().context(IdlingConnectionHandler.INSTANCE) != null) {
```
graemerocher: Maybe extra `pipeline()` to a local variable to avoid multiple calls

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -2564,6 +2564,16 @@ protected void addFinalHandler(ChannelPipeline pipeline) {
 
             @Override
             public void channelReleased(Channel ch) {
+                Duration idleTimeout = configuration.getConnectionPoolIdleTimeout().orElse(Duration.ofNanos(0));
+                if (ch.isOpen()) {
+                    ch.config().setAutoRead(true);
+                    ch.pipeline().addLast(IdlingConnectionHandler.INSTANCE);
```
graemerocher: Maybe extra `pipeline()` to a local variable to avoid multiple calls

graemerocher (changes requested):
A few changes, and looks like there is a failing test. Probably a concurrency issue

drmaas:
@graemerocher updated per comments. It seems like on each push one of the java builds randomly fails (last time was 15, now 11) and the failing tests don't seem to be related to the PR

graemerocher:
Thanks for the contribution!

PR #4798, opened 2021-01-11
Bump micronaut-test to 2.3.2

micronaut-build:


ilopmar:
Converted to draft because the Maven Central Sync job is failing when checking out the repository: https://github.com/micronaut-projects/micronaut-test/runs/1681090309?check_suite_focus=true#step:2:32

I've retried a few times. I'll retry it later.

PR #4810, opened 2021-01-14
Support for copy-constructors on immutable introspections

graemerocher:
This PR adds a new `withValue(..)` method on `BeanProperty` that allows using a copy-constructor approach on immutable introspections.

This will help support java records, Kotlin data classes, immutable types etc. better in Micronaut Data

### DIFF
--- a/core/src/main/java/io/micronaut/core/beans/BeanProperty.java
+++ b/core/src/main/java/io/micronaut/core/beans/BeanProperty.java
@@ -27,6 +27,8 @@ import io.micronaut.core.util.ArgumentUtils;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import javax.annotation.concurrent.Immutable;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/micronaut/core/beans/BeanProperty.java
```diff
@@ -121,6 +123,75 @@
...
+     *     <li>If there is no way for the property to be mutated then an {@link UnsupportedOperationException} is thrown</li>
+     * </ul>
+     *
+     * @param bean The bean
+     * @param value The new value
+     * @return Either the existing instance or the property is mutable or a newly created instance via the copy constructor pattern.
+     * @throws UnsupportedOperationException if the property cannot be mutated
+     * @since 2.3.0
+     */
+    default B withValue(@NonNull B bean, @Nullable T value) {
```
dstepanov, line 156: It would be nice to skip modification if value is the same:
```java
if (get(bean) == value) {
     return bean;
}
```

inject/src/main/java/io/micronaut/inject/beans/visitor/BeanPropertyWriter.java
```diff
@@ -155,22 +167,165 @@ public void accept(ClassWriterOutputVisitor classWriterOutputVisitor) throws IOE
...
+                // override isWriteOnly method
+                final GeneratorAdapter isWriteOnly = startPublicFinalMethodZeroArgs(classWriter, boolean.class, "isWriteOnly");
                 isWriteOnly.push(true);
                 isWriteOnly.returnValue();
                 isWriteOnly.visitMaxs(1, 1);
                 isWriteOnly.endMethod();
             }
 
+            // override isMutable method
+            boolean isMutable = this.isMutable;
```
dstepanov, line 187: Like above It would be nice to skip modification if the value is the same.
Maybe it can even delegate to an existing method to modify immutable objects like generated by https://projectlombok.org/features/Value

```java
ValueExample withAge(int age) {
    return this.age == age ? this : new ValueExample(name, age, score, tags);
  }
```

core/src/main/java/io/micronaut/core/beans/BeanProperty.java
```diff
@@ -121,6 +123,75 @@
...
+                if (prop == null) {
+                    throw new UnsupportedOperationException("Cannot create copy of type [" + declaringBean.getBeanType() + "]. Constructor contains argument [" + argumentName + "] that is not a readable property");
+                } else if (prop == this) {
+                    found = true;
+                    values[i] = value;
+                } else {
+                    values[i] = prop.get(bean);
+                }
+            }
+            if (found) {
```
dstepanov, line 180: This doesn't look correct, `found == true` should just instantiate and return otherwise it should look for the setter and fail if it doesn't find one 😉

inject/src/main/java/io/micronaut/inject/beans/visitor/BeanPropertyWriter.java
graemerocher: I think delegating to existing methods is a nice idea and may be possible to implement once #4813 is merged as we can use the new `ElementQuery` API to find the methods

PR #4813, opened 2021-01-15
Support for executable methods on introspections

graemerocher:
I did this on top of #4810 so that needs merging first, then I can rebase this


dstepanov:
Might be a bit confusing that getter/setter methods aren't included in "bean methods". Would a getter/setter be included if it's annotated as executable?

graemerocher:
it would be included yes

graemerocher:
FWIW I think we need to keep the separation between properties and methods because it may be for example that we want to also support public fields, so there may not always be a method behind a property.

The other reason as well is to ensure a property can represent both field and method metadata, so for example:

```
@NotBlank
private String name;

@Size(max=20)
public String getName() {
    return this.name;
}
```

Here we can fuse together the metadata from the field and the method or likewise if there were a constructor argument and a getter we can fuse those together.

graemerocher:
@jameskleeh rebased this

PR #4821, opened 2021-01-19
Respect multipart.enabled for Netty

### LINKED ISSUE #4773: Configuation micronaut.server.multipart.enabled seems to do nothing
There is a configuration setting `micronaut.server.multipart.enabled` which defaults to `false`.
As fas as I see, only this line evaluates the setting:
https://github.com/micronaut-projects/micronaut-core/blob/ebc39d4f6e3872d030763b3d5d36913262b4a062/http-server-netty/src/main/java/io/micronaut/http/server/netty/FormDataHttpContentProcessor.java#L72
which might save a `false` into this field:
https://github.com/micronaut-projects/micronaut-core/blob/ebc39d4f6e3872d030763b3d5d36913262b4a062/http-server-netty/src/main/java/io/micronaut/http/server/netty/FormDataHttpContentProcessor.java#L47

There is a getter
https://github.com/micronaut-projects/micronaut-core/blob/ebc39d4f6e3872d030763b3d5d36913262b4a062/http-server-netty/src/main/java/io/micronaut/http/server/netty/FormDataHttpContentProcessor.java#L77
but it does not seem to get accesses ever (at least my IntelliJ debugger did not pause on any access).

Actually, I'm using multipart forms on my project and everything is working fine with this setting set to false (default) :-).

jameskleeh:


http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -539,7 +540,18 @@ public long getMaxFileSize() {
...
+        }
+
+        /**
+         * The default multipart enabled setting is false for servlets, but effectively
+         * true for Netty. This method returns the value only if set by the user.
+         *
+         * @return The enabled setting.
+         */
+        @Internal
+        public Optional<Boolean> getEnabled() {
```
jameskleeh, line 553: I had to create this method because the default is false. We can't change the default for Netty so I needed to get access to the property without a default to know if the user explicitly set the value to false or not. I think the default should change to true for 3.0 then we can remove this method.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -443,6 +443,27 @@ protected void channelRead0(ChannelHandlerContext ctx, io.micronaut.http.HttpReq
         RouteMatch<?> route;
 
         final String requestMethodName = request.getMethodName();
+        MediaType contentType = request.getContentType().orElse(null);
+
+        Optional<Boolean> multipartEnabled = serverConfiguration.getMultipart().getEnabled();
```
graemerocher: thing this should be unwrapped from the optional and assigned to a field to avoid the performance overhead for each request

graemerocher (changes requested):

PR #4822, opened 2021-01-19
Exclude Kubernetes-specific variables

alvarosanchez:
Fixes https://github.com/micronaut-projects/micronaut-kubernetes/issues/209

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/KubernetesEnvironmentPropertySource.java
+++ b/inject/src/main/java/io/micronaut/context/env/KubernetesEnvironmentPropertySource.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context.env;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static io.micronaut.context.env.EnvironmentPropertySource.getEnv;
+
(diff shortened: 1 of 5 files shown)


jameskleeh:
@alvarosanchez It seems likely to me that some users may be relying on this data, and thus it should be opt in somehow. Do you think it's unreasonable to believe some folks are reading this data from config?

inject/src/main/java/io/micronaut/context/env/KubernetesEnvironmentPropertySource.java
```diff
@@ -0,0 +1,85 @@
...
+                    && !entry.getKey().endsWith("_TCP_PORT")
+                    && !entry.getKey().endsWith("_TCP_PROTO")
+                    && !entry.getKey().endsWith("_TCP_ADDR")
+                    && !entry.getKey().endsWith("_UDP_PORT")
+                    && !entry.getKey().endsWith("_UDP_PROTO")
+                    && !entry.getKey().endsWith("_UDP_ADDR")
+                    && !entry.getKey().endsWith("_SERVICE_PORT")
+                    && !entry.getKey().endsWith("_SERVICE_PORT_HTTP")
+                    && !entry.getKey().endsWith("_SERVICE_HOST")
+                    && (!entry.getKey().endsWith("_PORT") && !entry.getValue().startsWith("tcp://"))
```
jameskleeh: Some of these checks are redundant. `entry.getKey().endsWith("_PORT")` should cover `_TCP_PORT`, `_UDP_PORT`, `_SERVICE_PORT`

jameskleeh: Perhaps these string suffixes should be put in a list and you can just loop

jameskleeh (changes requested):

inject/src/main/java/io/micronaut/context/env/KubernetesEnvironmentPropertySource.java
alvarosanchez: Yeah I copied the code from the original poster of the issue. Will refactor it a bit

alvarosanchez:
@jameskleeh the old `KubernetesDiscoveryClient`, which relied on those environment variables, was deprecated in Micronaut 1.x and removed in 2.x. The Micronaut Kubernetes modules doesn't use environment variables.

I don't see a reason why anyone would access those variables directly, but in any case, they will always be available via `System.getenv()`.

graemerocher:
yeah typically you would use the Kubernetes module which would use those variables so I don't think there is any reason not to merge this change

PR #4827, opened 2021-01-20
adding the version for micronaut-jms

burtbeckwith:


build.gradle
```diff
@@ -665,7 +665,13 @@ ext {
                 version: lombokVersion,
                 group  : 'org.projectlombok',
                 name   : 'lombok'
-            ]
+            ],
+            'micronaut.jms': [
+                    version:micronautJmsVersion,
+                    group:'io.micronaut.jms',
+                    name:'micronaut-jms',
+                    modules:['jms-core','jms-activemq-classic','jms-activemq-artemis','jms-sqs']
```
graemerocher: Think these will need to be prefixed with `micronaut-`

graemerocher (changes requested):

graemerocher (approved):

build.gradle
```diff
@@ -665,7 +665,13 @@ ext {
                 version: lombokVersion,
                 group  : 'org.projectlombok',
                 name   : 'lombok'
-            ]
+            ],
+            'micronaut.jms': [
+                    version:micronautJmsVersion,
+                    group:'io.micronaut.jms',
+                    name:'micronaut-jms',
```
alvarosanchez: Is there a `io.micronaut.jms:micronaut-jms` artifact? I don't think so.

As far as I understand this, the `name` property is for modules with just one artifact, where `modules` is for registering additional artifacts.

@graemerocher correct me if I'm wrong

alvarosanchez (changes requested):

build.gradle
graemerocher: Yes you are right `name` should be `micronaut-jms-core` and the core module removed from the other modules

alvarosanchez (approved):

PR #4869, opened 2021-01-29
Revert bugfix around setting properties and nullability

### LINKED ISSUE #4866: BeanWrapper.setProperty validates nullability starting 2.3.0, it did not in 2.2.3
I have been using BeanIntrospection to load an excel file into a Micronaut application. 

The following test works with 2.2.3 but fails in 2.3.0 with: 

```

Null values not supported by property: title
java.lang.IllegalArgumentException: Null values not supported by property: title
	at io.micronaut.core.beans.AbstractBeanProperty.set(AbstractBeanProperty.java:147)
	at io.micronaut.core.beans.BeanProperty.convertAndSet(BeanProperty.java:243)
	at io.micronaut.core.beans.BeanWrapper.lambda$setProperty$5(BeanWrapper.java:193)
	at java.util.Optional.ifPresent(Optional.java:159)
	at io.micronaut.core.beans.BeanWrapper.setProperty(BeanWrapper.java:193)
	at example.BookSpec.introspection does not throw exception if constraints(BookSpec.groovy:23)
```

```java
package example;

import edu.umd.cs.findbugs.annotations.NonNull;

import javax.validation.constraints.NotBlank;
import javax.validation.constraints.NotNull;
import java.util.Objects;

@Introspected
public class Book {

    @NonNull
    @NotBlank
    private String title;

    @NonNull
    @NotNull
    private Integer pages;

    public Book() {
    }

    public Book(@NonNull String title, @NonNull Integer pages) {
        this.title = title;
        this.pages = pages;
    }

    @NonNull
    public String getTitle() {
        return title;
    }

    public void setTitle(@NonNull String title) {
        this.title = title;
    }

    @NonNull
    public Integer getPages() {
        return pages;
    }

    public void setPages(@NonNull Integer pages) {
        this.pages = pages;
    }

    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        Book book = (Book) o;
        return Objects.equals(title, book.title) && Objects.equals(pages, book.pages);
    }

    @Override
    public int hashCode() {
        return Objects.hash(title, pages);
    }
}

```

```groovy
package example

import io.micronaut.core.beans.BeanIntrospection
import io.micronaut.core.beans.BeanWrapper

class BookSpec extends Specification {

    void "introspection does not throw exception if constraints"() {

        List<String> headers = ['title', 'pages']
        List<List<String>> lines = [
                ['Raspberry Pi Guide',120],
                [null,50],
        ]

        when:
        List<BeanWrapper> beanWrapperList = []
        for (List<String> line : lines) {
            BeanIntrospection introspection = BeanIntrospection.getIntrospection(Book)
            BeanWrapper bw = BeanWrapper.getWrapper(BeanIntrospection.getIntrospection(Book.class).instantiate());
            for (int i = 0; i < headers.size(); i++) {
                bw.setProperty(headers[i], line[i])
            }
            beanWrapperList << bw
        }

        then:
        beanWrapperList.any {bw -> bw.bean == new Book('Raspberry Pi Guide', 120) }
        beanWrapperList.any {bw -> bw.bean == new Book(null, 50) }
    }
}

```

I think this change of behaviour is a regression. Moreover, I think it is a worst behaviour. I want to have control around when to the validate the objects after I load them to create a complete report. 

Sample project:

[micronaut-two-three-zero-introspection.zip](https://github.com/micronaut-projects/micronaut-core/files/5887905/micronaut-two-three-zero-introspection.zip)

jameskleeh:
Closes #4866

### DIFF
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
@@ -25,4 +25,20 @@ class BeanWrapperSpec extends Specification {
         wrapper.getRequiredProperty("name", String) == 'Fred'
         wrapper.getRequiredProperty("age", Integer.class) == 10
     }
+
+    void "test setting a non null with null"() {
+        when:"A wrapper is obtained"
+        def bean = new NullabilityBean()
+        BeanWrapper<NullabilityBean> wrapper = BeanWrapper.getWrapper(bean)
+
+        then:"it is correct"
+        wrapper.bean == bean
+
+        when:
+        wrapper.setProperty("name", null)
+
+        then:
+        noExceptionThrown()
+        bean.name == null
+    }
 }
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
@@ -41,4 +41,5 @@ class BeanWrapperSpec extends Specification {
         noExceptionThrown()
         bean.name == null
     }
+
 }
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
@@ -1,6 +1,7 @@
 package io.micronaut.inject.visitor.beans
 
 import io.micronaut.core.beans.BeanWrapper
+import spock.lang.PendingFeature
 import spock.lang.Specification
 
 class BeanWrapperSpec extends Specification {
@@ -26,6 +27,7 @@ class BeanWrapperSpec extends Specification {
         wrapper.getRequiredProperty("age", Integer.class) == 10
     }
 
+    @PendingFeature
     void "test setting a non null with null"() {
         when:"A wrapper is obtained"
         def bean = new NullabilityBean()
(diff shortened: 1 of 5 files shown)


inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
```diff
@@ -25,4 +25,20 @@ class BeanWrapperSpec extends Specification {
...
+        then:"it is correct"
+        wrapper.bean == bean
+
+        when:
+        wrapper.setProperty("name", null)
+
+        then:
+        noExceptionThrown()
+        bean.name == null
+    }
```
graemerocher: I would change this test to assert the opposite of what it is asserting now then annotate it with `@PendingFeature` and explain in the annotation that this behaviour will be changed in 3.0

graemerocher (changes requested):

inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanWrapperSpec.groovy
jameskleeh: I prefer this because then we have a test in place to ensure the behavior works as it should in 2.x and then the test will fail when the behavior changes. It may be forgotten or missed when the functionality changes if it was marked with `@PendingFeature`. I forgot to add a test for the constructor part, I'll do that

graemerocher: That is the whole point of `@PendingFeature` so that you know when the functionality changes it breaks the build hence it is not forgotten or missed :)

PR #4885, opened 2021-02-03
Making the parent HTTP Request available in handlers of the DefaultHttpClient

stepanv:
The aim of this PR is to start a discussion and get confirmation whether this is idea and the implementation is in the right direction. 

This PR has no tests which must be improved.

When executing the business logic in a scope of an incoming parent http request, the `DefaultHttpClient` executes part of its code on the original thread and in such case the `ServerRequestContext.currentRequest()` is available. Most of the logic, however, is executed as handlers on Netty event loop and the original parent request is not available. This PR propagates the original parent request to all the handlers.

As a result, when using the MDC [*] approach to propagate information (such as Correlation ID) between logging messages, all the logging statements logged by the DefaultHttpClient may contain the information from MDC and thanks to that the analysis of the logs becomes much easier.

The resulting log messages can look like:
```
2021-02-02 CET 12:54:07,438 [nioEventLoopGroup-1-5] [correlationId=4400b5bbfaf94515907b84df3e7fc69b] [principal=stepan@appd] INFO  com.appd.myservice - Hello everyone!
2021-02-02 CET 12:54:08,294 [nioEventLoopGroup-1-8] [correlationId=4400b5bbfaf94515907b84df3e7fc69b] [principal=stepan@appd] DEBUG io.micronaut.http.client.netty.DefaultHttpClient - Sending HTTP Request: POST /my-backend/v1.0
2021-02-02 CET 12:55:04,788 [nioEventLoopGroup-1-7] [correlationId=4400b5bbfaf94515907b84df3e7fc69b] [principal=stepan@appd] TRACE io.micronaut.http.client.netty.DefaultHttpClient - HTTP Client Response Received for Request: POST http://localhost:62003/my-backend/v1.0
2021-02-02 CET 12:55:08,448 [nioEventLoopGroup-1-7] [correlationId=4400b5bbfaf94515907b84df3e7fc69b] [principal=stepan@appd] TRACE io.micronaut.http.client.netty.DefaultHttpClient - Status Code: 200 OK
2021-02-02 CET 12:55:08,536 [parallel-1] [correlationId=4400b5bbfaf94515907b84df3e7fc69b] [principal=stepan@appd] INFO  com.appd.logging.info.SLOW_QUERY - Threshold of 30000 (millis) exceeded, execution took = 60617 (millis)
2021-02-02 CET 12:55:12,367 [nioEventLoopGroup-1-5] [correlationId=4400b5bbfaf94515907b84df3e7fc69b] [principal=stepan@appd] INFO  com.appd.logging.info.HTTP_ACCESS - 127.0.0.1 - - "POST /my-service/v1 HTTP/1.1" 200 327 35,741 61,662 ms [X-Forwarded-For: ""]
```

[*] The thread-local approach of using [MDC ](http://www.slf4j.org/api/org/slf4j/MDC.html)(i.e., `MDC.put()`, etc) is not the best in this case. Much better approach is through [Log4j2's](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/ContextDataInjector.html) `ContextDataInjector` and provide the resolved values from the `ServerRequestContext.currentRequest()` as a `Map` such as:
```
public class ServerRequestDiagnosticContextDataInjector implements ContextDataInjector {

    @Override
    public ReadOnlyStringMap rawContextData() {
        return ServerRequestContext.currentRequest()
                .map(request -> new SortedArrayStringMap(
                        Map.of(DiagnosticContext.CORRELATION_ID,
                        request.getAttribute(DiagnosticContext.CORRELATION_ID_ATTR, String.class)
                                .orElse("-- no correlation ID --"))))
                .orElse(EMPTY_CONTEXT_DATA);
    }
}

```

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -35,6 +35,7 @@ import io.micronaut.core.reflect.InstantiationUtils;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.util.ArgumentUtils;
 import io.micronaut.core.util.ArrayUtils;
+import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.*;
 import io.micronaut.http.HttpResponse;
@@ -73,6 +74,9 @@ import io.micronaut.jackson.codec.JsonMediaTypeCodec;
 import io.micronaut.jackson.codec.JsonStreamMediaTypeCodec;
 import io.micronaut.jackson.parser.JacksonProcessor;
 import io.micronaut.runtime.ApplicationConfiguration;
+import io.micronaut.scheduling.instrument.Instrumentation;
+import io.micronaut.scheduling.instrument.InvocationInstrumenter;
+import io.micronaut.scheduling.instrument.InvocationInstrumenterFactory;
 import io.micronaut.websocket.RxWebSocketClient;
 import io.micronaut.websocket.annotation.ClientWebSocket;
 import io.micronaut.websocket.annotation.OnMessage;
@@ -141,7 +145,6 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.*;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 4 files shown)


graemerocher:
Would be better to wrap the calls in `InvocationInstrumenter`. See ServerRequestContextFilter for an example on how to do this.

stepanv:
@graemerocher , do you mean something like this change? https://github.com/micronaut-projects/micronaut-core/commit/43e55d19a2e255cb4faff3365a4f8f92bf8bc931

stepanv:
@graemerocher , I think I understand what you meant ;) can you please check if the new commit is in the right direction?
The nice thing is that it automatically propagates Zipkin,MDC correlation data if enabled. The downside is that these are managed by Micronaut so I had to somehow propagate the `List<InvocationInstrumenterFactory>` to the Client impl.

graemerocher:
It is the right direction yes, you should in this case just use a no-op instrumented if there list is null and not throw an exception.

graemerocher:
@stepanv are you interested in taking this forward?

stepanv:
Yes @graemerocher , we will move it forward in next few days, thx for checking out!

stepanv:
@graemerocher, when you have time, please take a look at the diff and let me know if there's anything you disagree with.
Please let me know to what extent you expect this functionality to be tested. 
And also if it's possible to add (more) tests in a different PR.
Thx!

Issue #3552, opened 2020-06-24
Allow @Nullable as a Meta-annotation when using DefaultExecutableBinder


elliottpope:
### Task List

- [ ] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [ ] Full description of the issue provided (see below)

### Steps to Reproduce

1. Define an annotation that is itself annotated with `@Nullable` or `@NotNull` i.e.
```java
@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.PARAMETER)
@Nullable
public @interface MyAnnotation {
    String value();
}
```
2. Use the `DefaultExecutableBinder` to bind that argument to a method call where an argument has the annotation on a parameter:
```java
public class MyClass {
   public void myMethod(@MyAnnotation String arg) {
      System.err.println("Argument is: " + arg);
   } 

   public void test() {
      final DefaultExecutableBinder<String> binder = new DefaultExecutableBinder<>();
      BoundExecutable boundExecutable = binder.bind(this::myMethod, new TestArgumentBinderRegistry(), null);
      boundExecutable.invoke(this);
   }
}

public class TestArgumentBinderRegistry implements ArgumentBinderRegistry<String> {
   @Override
    public <T> Optional<ArgumentBinder<T, String>> findArgumentBinder(Argument<T> argument, String source) {
        return Optional.of((context, source) -> {
            if (context.hasAnnotation(MyAnnotation.class)) {
               return () -> Optional.ofNullable(source);
            }
            throw new RuntimeException("There shouldn't be any arguments in this test not annotated with @MyAnnotation.");
        });
    }
}
```

### Expected Behaviour

If the meta-annotation has the `@Nullable` annotation and the argument if the provided binding is `null` then the method should be invoked with `null`. If the `@Nullable` annotation is not present then the default behaviour should continue

### Actual Behaviour

Currently, if `@MyAnnotation` has `@Nullable` and the provided argument is `null` then the method invocation silently fails. 

I believe this is due to the fact that in `io.micronaut.core.bind.DefaultExecutableBinder.java:81`the underlying `argument.isNullable()` executes `getAnnotationMetadata().hasAnnotation(AnnotationUtil.NULLABLE)` (`io.micronaut.core.annotation.AnnotatedElement.java:39`)

[...]

PR #4907, opened 2021-02-09
Introduce nullability annotations and remove hard dependency on spotbugs

### LINKED ISSUE #4334: micronaut-core still has transitive dependency on findbugs:jsr305 (com.google.code.findbugs:jsr305:3.0.2)
As of Micronaut 2.1, there is still a transitive dependency on `findbugs:jsr305`

```
+--- io.micronaut:micronaut-inject:2.1.1
     +--- org.slf4j:slf4j-api:1.7.26 -> 1.7.30
     +--- javax.annotation:javax.annotation-api:1.3.2
     +--- javax.inject:javax.inject:1
     +--- io.micronaut:micronaut-core:2.1.1
     |    +--- org.slf4j:slf4j-api:1.7.26 -> 1.7.30
     |    +--- org.reactivestreams:reactive-streams:1.0.3
     |    \--- com.github.spotbugs:spotbugs-annotations:4.0.3
     |         \--- com.google.code.findbugs:jsr305:3.0.2
     \--- org.yaml:snakeyaml:1.26
```

The transitive dependency results in Java Module System split-package issues:

```
error: the unnamed module reads package javax.annotation from both java.annotation and jsr305
error: module io.micronaut.core reads package javax.annotation from both jsr305 and java.annotation
error: module javafx.controlsEmpty reads package javax.annotation from both jsr305 and java.annotation
error: module javafx.graphicsEmpty reads package javax.annotation from both jsr305 and java.annotation
error: module javafx.baseEmpty reads package javax.annotation from both jsr305 and java.annotation
error: module java.annotation reads package javax.annotation from both jsr305 and java.annotation
error: module org.yaml.snakeyaml reads package javax.annotation from both jsr305 and java.annotation
error: module org.reactivestreams reads package javax.annotation from both jsr305 and java.annotation
error: module com.github.spotbugs.annotations reads package javax.annotation from both jsr305 and java.annotation
error: module jsr305 reads package javax.annotation from both jsr305 and java.annotation
error: module io.micronaut.inject reads package javax.annotation from both jsr305 and java.annotation
error: module org.slf4j reads package javax.annotation from both jsr305 and java.annotation
error: module javax.inject reads package javax.annotation from both jsr305 and java.annotation
```

 I work around this with the Java Modularity Gradle Modules Plugin as follows:

```
patchModules.config = [
        "java.annotation=jsr305-3.0.2.jar"
]
```

Note that the [documentation](https://github.com/micronaut-projects/micronaut-core/blob/760c8117e7bd1f3a38b2060241ed5025dd7f24b1/src/main/docs/guide/appendix/breaks.adoc#spot-bugs-instead-of-jsr-305-nullablenonnull-annotations) is incorrect and says

> The JSR-305 annotations library (com.google.code.findbugs:jsr305) is no longer a dependency (replaced by spotbugs-annotations). If you need this library you will need to add it manually.

graemerocher:
Note that with this change if a user uses the annotations in their code they will likely receive this compiler warning:

```
  reason: class file for javax.annotation.meta.When not found
```

This can be mitigated by adding:

```
compileOnly('com.google.code.findbugs:jsr305')
```

Which should probably be in the project template for 2.4.0 or added to the Gradle / Maven plugins so it can be configured automatically

Fixes #4334 as well as the issue with spot bugs being LGPL


core/src/main/java/io/micronaut/core/annotation/NonNull.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * <p>Should be used at parameter, return value, and field level. Method overrides should repeat parent {@code @NonNull} annotations unless
+ * they behave differently.</p>
+ *
+ * <p>Use {@code @NonNullApi} (scope = parameters + return values) to set the default behavior to non-nullable in order to avoid annotating
+ * your whole codebase with {@code @NonNull}.</p>
+ *
+ * @since 2.4
+ * @author graemerocher
+ * @see Nullable
```
jameskleeh, line 35: Should see non null annotation

core/src/main/java/io/micronaut/core/annotation/NonNull.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Target({ElementType.METHOD, ElementType.PARAMETER, ElementType.FIELD})
```
jameskleeh: The target of annotation should be added so we can support https://github.com/micronaut-projects/micronaut-core/issues/3552

inject-java-test/src/test/groovy/io/micronaut/inject/annotation/AnnotationTransformerSpec.groovy
```diff
@@ -12,7 +12,8 @@ package test;
 @io.micronaut.inject.annotation.ToTransform
 @javax.inject.Singleton
 class Test {
-
+    void test(@io.micronaut.core.annotation.Nullable String test) {
+    }
```
jameskleeh: How is this change relevant for this test?

runtime/src/main/java/io/micronaut/health/HealthStatus.java
```diff
@@ -29,7 +28,6 @@
  * @author Graeme Rocher
  * @since 1.0
  */
-@Immutable
```
jameskleeh, line 32: Why was this removed?

validation/src/test/groovy/io/micronaut/validation/executable/NullablePrimitiveSpec.groovy
```diff
@@ -47,7 +47,7 @@ class NullablePrimitiveSpec extends AbstractTypeElementSpec {
         output.contains(warning)
 
         where:
-        annotation << ['@javax.annotation.Nullable', '@edu.umd.cs.findbugs.annotations.Nullable']
+        annotation << ['@javax.annotation.Nullable', '@io.micronaut.core.annotation.Nullable']
```
jameskleeh: Perhaps add the micronaut annotation here instead of replacing the spotbugs one

core/src/main/java/io/micronaut/core/annotation/NonNull.java
graemerocher: this javadoc is on the `NonNull` annotation, so pointing to itself doesn't make sense to me

inject-java-test/src/test/groovy/io/micronaut/inject/annotation/AnnotationTransformerSpec.groovy
graemerocher: Don't think it is, will cleanup it up

runtime/src/main/java/io/micronaut/health/HealthStatus.java
graemerocher: `@Immutable` is inside the the findbugs jar file which we can no longer use. The alternatively is to leave this annotation there and make find bugs `compileOnly`

core/src/main/java/io/micronaut/core/annotation/NonNull.java
jameskleeh: Ah I see. I thought it was pointing to the javax ones. Nevermind then

PR #4910, opened 2021-02-09
Decouple core and inject from reactive streams

graemerocher:
This PR decouples `core` and `inject` from reactive streams which we would probably need to do at some point anyway to support the `Flow` API in JDK 11 etc.

It also allows people to use Micronaut DI with one less dependency if they are not interested in the reactive part


core/build.gradle
```diff
@@ -8,11 +8,9 @@ shadowJar {
 
 dependencies {
     api dependencyVersion("slf4j")
-    api dependencyVersion("reactive.streams")
```
jameskleeh, line 11: Are you reliant on this being brought in transitively through rxjava2? Can you envision a scenario where removing this would break an app?

graemerocher: I added it as a dependency of runtime and http so don’t believe from a user perspective this will break anything

PR #4911, opened 2021-02-09
Fix request URI usage for stream requests

### LINKED ISSUE #4908: HttpClientFilter not working for streaming requests
### Task List

### Steps to Reproduce

1. Create a catch all HttpClientFilter that adds a uri param to any request
```
@Filter("/**")
public class SampleFilter implements HttpClientFilter {
    @Override
    public Publisher<? extends HttpResponse<?>> doFilter(MutableHttpRequest<?> request, ClientFilterChain chain) {
        return chain.proceed(request.uri((ub) -> ub.queryParam("uid", "valueFromFilter")));
    }
}
```

2. Create a declarative HttpClient that returns a stream
```
@Client("http://localhost:8088")
public interface SampleClient {
    @Get(value = "/flowable", processes = MediaType.APPLICATION_OCTET_STREAM)
    Flowable<ByteBuffer<?>> getFlowable();
}
```

### Expected Behaviour

Query param `uid` with value `valueFromFilter` is added to the request.

### Actual Behaviour

The request is sent unmodified.

### Environment Information

- **Operating System**:  macOS 11.2
- **Micronaut Version:** 2.2.0 in the example application, also tested with 2.3.1 with same results
- **JDK Version:** openjdk version "1.8.0_232"

### Example Application

- https://github.com/Sandmania/micronaut-http-client-filter-bug
- The example application consists of a client and a server. The server expects to receive a `uid`uri param for every request. The client attempts to achieve this with a catch all HttpClientFilter but it fails to work with request returning Flowable.

ttzn:
Fixes #4908

URI modifications performed in HTTP client filters are not picked up when building streaming requests.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1002,7 +1002,6 @@ public class DefaultHttpClient implements
                             request.setAttribute(NettyClientHttpRequest.CHANNEL, channel);
                             streamRequestThroughChannel(
                                     parentRequest,
-                                    requestURI,
                                     requestWrapper,
                                     emitter,
                                     channel,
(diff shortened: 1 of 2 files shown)


graemerocher:
Thanks for the contribution!

PR #4914, opened 2021-02-10
Introduce micronaut-context subproject that micronaut-runtime depends on

graemerocher:
Currently if you want to use just DI from Micronaut if you depend on `micronaut-inject` then certain container features like scheduling, immutable configuration properties, event listeners do not work. Also the `Micronaut` class becomes inaccessible.

However if you use `micronaut-runtime` is includes a whole additional set of requirements like Jackson, cloud features, HTTP related functionality etc.

This is a start on modularizing things a bit better by moving the following functionality with the `micronaut-context` subproject and making `micronaut-runtime` depend on it:

* Job scheduling
* Immutable config properties
* event listeners
* The `Micronaut` main class

This means you get a more complete DI solution if you want to just use DI in Micronaut by depending on `micronaut-context`

In terms of the name this is similar to the way Spring is arranged with the `spring-context` subproject (with `spring-beans` providing core di)


context/src/test/groovy/io/micronaut/scheduling/ScheduledFixedRateSpec.groovy
```diff
@@ -62,24 +61,6 @@ class ScheduledFixedRateSpec extends Specification {
         }
     }
 
-    void 'test scheduled annotation with retry'() {
```
jameskleeh, line 65: Why was this removed?

runtime/src/test/groovy/io/micronaut/runtime/event/annotation/EventListenerSpec.groovy
```diff
@@ -26,18 +11,11 @@ class EventListenerSpec extends Specification {
         ApplicationContext ctx = ApplicationContext.run()
 
         when:
-        TestListener t = ctx.getBean(TestListener)
-        GroovyListener g = ctx.getBean(GroovyListener)
         AsyncListener a = ctx.getBean(AsyncListener)
         PollingConditions conditions = new PollingConditions(timeout: 1)
 
         then:
-        t.invoked
```
jameskleeh, line 35: Why were these assertions removed?

context/src/test/groovy/io/micronaut/scheduling/ScheduledFixedRateSpec.groovy
graemerocher: No test had to be split since it tested functionality from both modules and new tests were added to replace removed ones

runtime/src/test/groovy/io/micronaut/runtime/event/annotation/EventListenerSpec.groovy
graemerocher: No see above

PR #4915, opened 2021-02-10
Fix for distributed tracing instrumentation on blocking controllers

### LINKED ISSUE #4383: Distributed tracing instrumentation doesn't work with @ExecuteOn IO
The distributed tracing instrumentation doesn't work the same when executing call on Netty's event loop and on the IO thread pool.

### Steps to Reproduce

- `git clone https://github.com/ilopmar/jconf-peru-2020-micronaut-en-accion`
- `cd jconf-peru-2020-micronaut-en-accion`
- Start Consul: `docker run -it --rm -p 8500:8500 consul:latest`
- Start Zipkin: `docker run -it --rm -p 9411:9411 openzipkin/zipkin:2.21.1`
- Start both encrypt-service and gateway
  - `cd encrypt-service && ./gradlew run`
  - `cd gateway && ./gradlew run`
- Send request: `curl localhost:8080/api/v1/encrypt/Hello`
- Open Zipkin UI: http://localhost:9411. It is possible to find all the tracing across both services:

image

- Stop `gateway` and annotate `GatewayController` with `@ExecuteOn(TaskExecutors.IO)` and start it again.
- Send again the a new curl request: `curl localhost:8080/api/v1/encrypt/Hello2`

### Expected Behaviour

Same requests should be logged on Zipkin

### Actual Behaviour

The request is splitted in two different requests on Zipkin:

image

image

### Environment Information

- **Operating System**: Linux Mint
- **Micronaut Version:** 2.1.2
- **JDK Version:** Open JDK 11

lgathy:
Closes #4383 

The problem seemed to be that the resultEmitter is constructed before the http filter chain was assembled and executed, thus the deffered Flowable could not catch all the changes to the context done by any of the filters or the publishers created by them. I think the only valid option is to delay this defer until the subscribe() of the publisher produced by the innermost http filter is called. This is only done after subscribing to all publishers created by all the other http filters. However, I needed to add an extra `Flowable.defer(...)` operation to achive this. I am not sure if this is the best way, but it does not seem to break any other tests.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
@@ -365,12 +365,12 @@ class RoutingInBoundHandler extends SimpleChannelInboundHandler<io.micronaut.htt
                 ExceptionHandler handler = exceptionHandler.get();
                 MediaType defaultResponseMediaType = MediaType.fromType(handler.getClass()).orElse(MediaType.APPLICATION_JSON_TYPE);
                 try {
-                    Publisher<MutableHttpResponse<?>> routePublisher = Flowable.fromCallable(() -> {
+                    Flowable<MutableHttpResponse<?>> routePublisher = Flowable.fromCallable(() -> {
                         Object result = handler.handle(nettyHttpRequest, cause);
                         return errorResultToResponse(result);
                     });
-
-                    filterPublisher(new AtomicReference<>(nettyHttpRequest), routePublisher, ctx.executor(), nettyException)
+                    Supplier<Flowable<? extends MutableHttpResponse<?>>> publisherSupplier = () -> routePublisher;
+                    filterPublisher(new AtomicReference<HttpRequest<?>>(nettyHttpRequest), publisherSupplier, ctx.executor(), nettyException)
                             .firstOrError()
                             .subscribe((mutableHttpResponse, throwable) -> {
                                 if (throwable != null) {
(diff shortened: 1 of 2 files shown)


lgathy:
@ilopmar @graemerocher Is it possible to merge this before the `2.3.2` release, if you agree with the approach I have taken? Luckily, only one private method signature change was needed.

jameskleeh:
A broken test showing the issue would be a better place to start. At a glance these changes don't make sense to me

lgathy:
@jameskleeh The first commit changes the built-in test so that it breaks. Isn't that sufficient enough?

jameskleeh:
@lgathy If the existing tests break then I would assume that existing users applications might break. Do you think that is the case? If you need to change the tests in order for them to break then you are effectively no longer testing the previous case and thus new tests should be added instead of changing old ones.

graemerocher:
@lgathy Thanks for the PR!

 I think the change makes sense however I agree with @jameskleeh that we need to verify the other case works as well so we would need an additional set of tests. Would you be able to modify the test to test both cases? Thanks.

lgathy:
@jameskleeh Ok, I get your point.

Actually I see these changes as fixing that test to actually test what its original intention was. Those controller methods intended to be blocking methods back in `1.x`, however in `2.0.0` [this breaking change](https://docs.micronaut.io/latest/guide/index.html#_thread_pool_selection_now_manual) was introduced, and since this test controller was not adopted to that change those methods were no longer executed on the IO threads as before. So applying the `@ExecuteOn` annotations only makes them `2.x` compatible.

Nonetheless, we could create new additional test cases if you say that's better. Honestly I thought that having a blocking method without the `@ExecuteOn` annotation is actually misusage so that's why I went for adding the annotation on the existing methods instead. What do you think?

lgathy:
@graemerocher Sure, I will do that. Thanks for the feedback.

graemerocher:
@lgathy I understand your point though in just because a method returns `String` doesn't mean it blocks the event loop. Some of the methods in the test actually do block the event loop but others do not I/O so don't block. I think having a set of tests that cover both cases would be a good thing.

lgathy:
@graemerocher That is indeed a valid use case, I slipped through that and I also didn't notice that @jameskleeh was refering to that still being a valid use-case, tbh.

I modified the PR now, reverted the changes to the existing methods and added separate ones for the blocking use-case alongside with test cases. There would be a couple more possible combinations - i.e. to call blocking from non-blocking and vica versa in the nested test cases. I am not sure if those would make any difference in the execution. Do you think we would need them?

graemerocher:
I think this is probably enough for now. Thanks for the contribution

PR #4917, opened 2021-02-10
Support for Jakarta inject annotations

jameskleeh:


core/src/main/java/io/micronaut/core/annotation/AnnotationValueBuilder.java
```diff
@@ -421,4 +421,17 @@
...
         return this;
     }
+
+    /**
+     * Sets the members from the provided map
+     *
+     * @param members The map of members
+     * @return This builder
+     */
+    public AnnotationValueBuilder<T> members(Map<CharSequence, Object> members) {
```
graemerocher: Should probably add `@NotNull` to return type and arguments

core/src/main/java/io/micronaut/core/annotation/AnnotationValueBuilder.java
```diff
@@ -421,4 +421,17 @@
         }
         return this;
     }
+
+    /**
+     * Sets the members from the provided map
+     *
+     * @param members The map of members
+     * @return This builder
```
graemerocher, line 461: Would be good to add `@since 2.4.0`

core/src/main/java/io/micronaut/core/annotation/AnnotationValueBuilder.java
```diff
@@ -421,4 +421,17 @@
...
+     */
+    public AnnotationValueBuilder<T> members(Map<CharSequence, Object> members) {
+        if (members != null) {
+            values.putAll(members);
```
graemerocher: Since this new API now allows arbitrary values I think we may need some checking that the objects in the map are either primitives, primitive arrays, strings, or annotation values otherwise passing something invalid could lead to odd compilation errors downstream

inject/src/main/java/io/micronaut/context/Provider.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.context;
+
+/**
+ * A provider contract.
+ *
+ * @param <T> The type
+ * @author James Kleeh
+ * @since 2.3.0
```
graemerocher: 2.4.0?

inject/src/main/java/io/micronaut/context/Provider.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public interface Provider<T> {
```
graemerocher: Maybe name it `BeanProvider` to more easily distinguish it from the other interfaces of the same name

inject/src/main/java/io/micronaut/context/Provider.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * @return A fully-constructed and injected instance of T.
+     */
+    T get();
```
graemerocher: `@NotNull` annotation?

inject/src/main/java/io/micronaut/context/ProviderFactory.java
```diff
@@ -0,0 +1,88 @@
...
+            if (provider.isAssignableFrom(clazz)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public static Set<Class> getProviders() {
+        return providers.keySet();
+    }
```
graemerocher: javadoc on methods

graemerocher (changes requested):
Nice work, just a few changes.

PR #4925, opened 2021-02-11
test-suite java code cleanup

burtbeckwith:


test-suite/src/test/java/io/micronaut/docs/annotation/PetOperations.java
```diff
@@ -30,7 +29,7 @@
 public interface PetOperations {
     // tag::save[]
     @Post
-    Single<Pet> save(@NotBlank String name, @Min(1L) int age);
+    Single<Pet> save(@NotBlank String name, @Min(1) int age);
```
jameskleeh: The value to `Min` is a long so the previous code makes more sense to me

PR #4937, opened 2021-02-14
Messaging annotation updates

graemerocher:
* Deprecate Body/Header/Headers to avoid confusion with HTTP versions
* Make MessageListener applicable to type to allow registering listeners without coupling to messaging system
* Add MessageProducer to meta annotation interfaces that produce messages


messaging/src/main/java/io/micronaut/messaging/annotation/MessageBody.java
```diff
@@ -0,0 +1,45 @@
...
+ * An annotation that can be applied to method argument to indicate that the method argument is bound from message body.
+ *
+ * @author Graeme Rocher
+ * @since 1.0
+ */
+@Documented
+@Retention(RUNTIME)
+@Target({ElementType.PARAMETER})
+@Bindable
+public @interface MessageBody {
```
jameskleeh, line 37: I don't think this is backwards compatible for use cases like https://github.com/micronaut-projects/micronaut-rabbitmq/blob/7a41c432cc57c1a0974b507faff7b293328ca638/rabbitmq/src/main/java/io/micronaut/rabbitmq/bind/RabbitBinderRegistry.java#L76. If someone switched to using this with RabbitMQ the annotation type would be `MessageBody` and there is no binder setup for that. It seems either we do a big bang for 3.0, or introduce the new annotations now but do not deprecate the existing ones. We can then update the modules to support both and then deprecate the old ones for 2.5, with removal in 3.0.

graemerocher: There is an alias for on the annotation so it should activate the `@Body` annotation as well and not break anything, the alternative is an annotation transformer to maintain compatibility

jameskleeh: The alias only applies if the value is set, and its common to not set the value for `@Body`. `getAnnotationTypeByStereotype` would choose the first annotation, so even in the case where the value is set, it's unclear which annotation it would pick.

graemerocher: yeah hence why I think a annotation transformer would solve that. Will write a test

graemerocher: @jameskleeh I added a transformer to make the annotations backwards compatible now. Can you check and merge if you agree?

PR #4941, opened 2021-02-15
Ensure singleton bean is not created twice

### LINKED ISSUE #4891: Singleton bean created twice when taken from ApplicationContext.getBeansOfType() by interface
In `2.2.3`.

I have the following classes:

```java
public abstract class BaseService<T> implements IBeanValidator<T>  {

	// validates entities marked with @BeanValidation
	public boolean isValid(@Nonnull T entity,
						   @Nonnull AnnotationValue<BeanValidation> annotationMetadata,
						   @Nonnull ConstraintValidatorContext context) {
		return true;
	}

}
```

```kotlin
@TransactionalAdvice @Singleton
class BankService(...): BaseService<Bank> {} // not direct inheritance, but for simplicity ....
```

The `BankSevice` bean is created properly by the container and as I can see in `DefaultBeanContext.getBeanInternal()` its key is `io.luna.service.main.finance.bank.service.BankService`.

However, this code created another instance of this bean (and all beans derived from the `BaseService`):

```java
@Singleton
public class BeanValidatorDelegate implements ConstraintValidator<BeanValidation, Object> {

	@Inject
	protected ApplicationContext applicationContext;

	@Override
	public boolean isValid(@Nullable Object value, @Nonnull AnnotationValue<BeanValidation> annotationMetadata,
						   @Nonnull ConstraintValidatorContext context) {
		if (value==null)
			return true;

		return applicationContext
			.getBeansOfType(IBeanValidator.class, Qualifiers.byTypeArgumentsClosest(value.getClass()))
			.stream()
			.map(bean -> bean.isValid(value, annotationMetadata, context))
			.reduce((b1, b2) -> b1 && b2)
			.orElse(true);
	}

}
```

In `DefaultBeanContext.getBeansOfTypeInternal()` I can see the `existing` bean is evaluated to `null` because of different key used here which is `<Bank> io.luna.feature.core.validation.IBeanValidator`.

I understand the underlying logic, however this breaks the rule of `@Singleton` bean in the app.

jameskleeh:
Fixes #4891

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/beans/inheritance/AbstractService.java
+++ b/inject-java/src/test/groovy/io/micronaut/inject/beans/inheritance/AbstractService.java
@@ -1,0 +1,4 @@
+package io.micronaut.inject.beans.inheritance;
+
+public abstract class AbstractService<T> implements ServiceContract<T> {
+}
--- a/inject-java/src/test/groovy/io/micronaut/inject/beans/inheritance/BankService.java
+++ b/inject-java/src/test/groovy/io/micronaut/inject/beans/inheritance/BankService.java
@@ -1,0 +1,7 @@
+package io.micronaut.inject.beans.inheritance;
+
+import javax.inject.Singleton;
+
+@Singleton
+public class BankService extends AbstractService<String> {
+}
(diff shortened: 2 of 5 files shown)


jameskleeh:
I noticed that DefaultBeanContext is not internal. Should I deprecate the existing method instead of directly adding the param?

graemerocher:
Strictly yes the method signature should stay the same and deprecation introduced.

Issue #4658, opened 2020-12-03
CDI Alignment Tasks


graemerocher:
The following issue covers the tasks required to assign Micronaut's DI system with CDI such that future compatibility with CDI lite can be achieved. This is an epic issue covering many other sub issues that need to implement to correctly support a proposed future CDI lite specification.

### Task List

- [x] Support for Annotations on Generic Type Arguments. (Fixed in https://github.com/micronaut-projects/micronaut-core/pull/4703)
- [x] Support for `@NonBinding` and using members in qualifiers. (See https://github.com/micronaut-projects/micronaut-core/issues/4769)
- [x] Support for Narrowing Injection by Type Arguments and Qualifiers on Type Arguments (PR #5082)
- [x] Support for Constructor Interceptors (PR https://github.com/micronaut-projects/micronaut-core/pull/5239)
- [x] Support for `@Named` without a value https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#named_at_injection_point (PR #5386)
- [x] Support for `@Any` qualifier (Issue #5155, PR #5156)
- [x] Support for `@Dependent` and `@TransientReference` https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#dependent_destruction (PR #5612)
- [x] Allow controlling annotation inheritance with `java.lang.annotation.Inherited` and possibly a member of `@Bean` (PR #5643)
- [x] Support for `@Factory` to define [bean fields](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#producer_field) (See https://github.com/micronaut-projects/micronaut-core/issues/4797, PR https://github.com/micronaut-projects/micronaut-core/pull/5343)
- [x] [Altering the exposed type](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#restricting_bean_types) via `@Typed` (Issue #5098, PR https://github.com/micronaut-projects/micronaut-core/pull/5106)
- [x] Allow `@Ordered` (later mapped at compilation to `@Priority`) to be used as fallback in order to support [bean priority and alternatives](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#declaring_selected_alternatives) (PR #5433)
[...]

PR #4988, opened 2021-02-24
Support more flexible Interceptor binding

graemerocher:
Adds a new `@InterceptorBinding` annotation and lookup strategy for interceptors that is backwards compatible with the old approach.

Benefits include:

* Now possible to have multiple interceptors for a single annotation
* An annotation mapper can add AOP to existing annotations without the existence of a Micronaut AOP annotation
* The approach here is more closely aligned with CDI (see #4658)


dstepanov:
Interesting, does it make sense to have `@InterceptorBinding(kind = InterceptorKind.INTRODUCTION)` on the interceptor type? I can imagine a scenario where you can have a projection of a type with some additional interface: `@GenerateProxy` on an `@Entity` will produce an instance with implemented interface HibernateProxy etc. and this is separated from different interface introductions. 

I feel like it would be better to have `@InterceptorBinding` just a simple marker annotation without any kind and have `@Interceptor(MyAnn.class)` as an actual interceptor type annotation.

graemerocher:
I tried to follow the as close as possible the design of CDI since we will most likely support a lighter version of that in the future. See https://docs.jboss.org/cdi/spec/1.0/html/interceptors.html

dstepanov:
I would have something like this:
```java
@Inherited
@InterceptorBinding
@Target({TYPE, METHOD})
@Retention(RUNTIME)
public @interface Transactional {}

@Interceptor(Transactional.class)
public class TransactionInterceptor implements MethodInterceptor<Object, Object> {
}

// Existing cases:

@Documented
@Retention(RUNTIME)
@Scope
@Introduction
//@Type(KafkaClientIntroductionAdvice.class)
@InterceptorBinding
@Recoverable
@Singleton
public @interface KafkaClient {...}

@Documented
@Retention(RUNTIME)
@Target({ElementType.METHOD, ElementType.TYPE, ElementType.ANNOTATION_TYPE})
//@Around
//@Type(RecoveryInterceptor.class)
@InterceptorBinding
public @interface Recoverable {}

@Documented
@Retention(RUNTIME)
@Target({ElementType.METHOD, ElementType.TYPE, ElementType.ANNOTATION_TYPE})
@Around(lazy=true) // custom settings
//@Type(RecoveryInterceptor.class)
@InterceptorBinding
public @interface CustomRecoverable {}

```

graemerocher:
@dstepanov that already works, the syntax is:

```java
@Inherited
@InterceptorBinding
@Target({TYPE, METHOD})
@Retention(RUNTIME)
public @interface Transactional {}

@InterceptorBinding(Transactional.class)
public class TransactionInterceptor implements MethodInterceptor<Object, Object> {
}
```

dstepanov:
Yes, but IMHO it's better to have a separate annotation for an interceptor. Otherwise, you can include `kind` and `interceptorType`, which I don't think makes sense at the interceptor.

Personally, I think that `@Around` and `@Introduction` should be replaced with something like `@AopProxy` to specify what exactly is happening and avoid using it in the new API.

graemerocher:
Using `kind` is still possibly useful on the interceptor to limit where the interceptor can be applied and whilst I agree that `interceptorType` is not really applicable like I said we are trying to stay at least a bit consistent with `javax.interceptor` (see https://docs.oracle.com/javaee/6/tutorial/doc/gkhjx.html)

In terms of `@Around` and `@Introduction` we can't really go back on those now as that would be a major breaking change for users. This PR is targeting 2.4.x which must follow semantic versioning. It is possible that we can re-arrange things for 3.0 but even then large breaking changes need to be strongly justified.

dstepanov:
@graemerocher I'm not sure what exactly you are referring to. My suggestion was just to have two annotations instead of one.

Future `@AopProxy` can be used in `@Around` and `@Introduction`, which would inherit it for backward compatibility.

graemerocher:
@dstepanov a common meta annotation for that makes sense yes

gradle.properties
```diff
@@ -25,6 +25,7 @@ testswebsocket=http-server-netty/src/test/groovy/io/micronaut/http/server/netty/
 metricscore=configurations/micrometer-core/src/main/java/io/micronaut/configuration/metrics
 examples=examples
 jdkapi=https://docs.oracle.com/javase/8/docs/api
+jdkapi=https://docs.oracle.com/javase/8/docs/api
```
jameskleeh: Seems like a mistake?

inject-java/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
 
     testImplementation project(':aop')
     testAnnotationProcessor project(":inject-java")
+    testImplementation project(":inject-java-test")
```
jameskleeh, line 13: inject-java-test has an api dependency on inject-java, thus creating a cycle?

inject-java/src/test/resources/logback.xml
```diff
@@ -11,4 +11,5 @@
     <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
+    <logger name="io.micronaut.aop.chain" level="trace" />
```
jameskleeh: Needs cleaned up

inject-java/build.gradle
graemerocher: Seems to work fine 🤷‍♂️

graemerocher:
@jameskleeh added `@InterceptorBean` which requires fixing some bugs in `@AliasFor` support for repeatable annotations as well as treating meta annotations with `@Bean` as a declared stereotype as beans.

Can we try get this merged soon?

PR #4997, opened 2021-02-25
Only build environment metadata if property expressions exist

graemerocher:
Currently we wrap all metadata in environment specific handling which is immensely wasteful if there are no property expressions. This PR which needs to be merged after the interceptor binding PR checks the metadata for property expressions at compilation time and then only wraps the metadata if they are present.


graemerocher:
Once https://github.com/micronaut-projects/micronaut-core/pull/4988 is merged I will shift this over to target 2.4.x

jameskleeh (approved):

PR #5008, opened 2021-02-26
Bump micronaut-kubernetes to 2.3.0

micronaut-build:


build.gradle
```diff
@@ -616,7 +616,7 @@ ext {
             'micronaut.kubernetes': [
                     version:micronautKubernetesVersion,
                     group:'io.micronaut.kubernetes',
-                    name:'micronaut-kubernetes-discovery-client'
+                    modules:['micronaut-kubernetes-discovery-client', 'micronaut-kubernetes-client', 'micronaut-kubernetes-client-processor']
```
alvarosanchez, line 619: Did you test how the BOM is published? Not sure if you can omit a `name` key. As far as I remember, all the entries have at least one artifact under `name`, then the rest in `modules`

alvarosanchez (approved):
Minor comment which I don't know whether it's an issue or not

build.gradle
pgressa: i have just followed the same pattern like it is above and below this module.. the name is not there as well, but I didn't check how the bom is generated

PR #5021, opened 2021-03-01
Fix for #5020

### LINKED ISSUE #5020: Micronaut parent POM breaks exec.executable CLI override in maven, breaks execution from NetBeans IDE
### Steps to Reproduce

1. Generate a Micronaut project, using the [Micronaut Launch](https://micronaut.io/launch/). Choose:
1.2. Micronaut Applicaiton, Java: 11, Base package: com.example (default), Name: micronaut-maven-exec-bug
1.3. MIcronaut version: 2.2.3, Language: Java, Build: Maven, Test Framework: JUnit
2. Setup system so that Java 8 or Java 9 is listed first in PATH and pointed to by JAVA_HOME env variable: this should be the  setup of default JDK for most tools.
3. Let `/space/java/11` be the path to the Java11 JDK installation (generated project now requires Java11). Execute the following command: `JAVA_HOME=/space/java/11 mvn -DskipTests -Dexec.executable=/space/java/11/bin/java install exec:exec`

### Expected Behaviour

The project should compile with the selected (different than system-default) JDK. That JDK runtime will be used to run the Micronaut application. The JVM that should be used as runtime (`/space/java/11/bin/java`) is specified on commandline, according to [maven-exec-plugin documentation](https://www.mojohaus.org/exec-maven-plugin/exec-mojo.html#executable). 

### Actual Behaviour

The project **is compiled** by Java11 JDK as expected. But the application **is run by JDK8**, the system-default JDK (first listed on PATH). The first JDK on PATH is used *despite* `JAVA_HOME` (used by maven-compiler-plugin) *and despite* `exec.executable` explicit override passed on commandline, according to maven documentation. The project's POM contains no `exec.excutable` property definition, or `executable` property configured for exec-maven-plugin, so the ignorance of a maven-documented property is unexpected.

### Environment Information

- **Operating System**: Ubuntu Linux 19.04
- **Micronaut Version:** 2.2.3
- **JDK Version:** Oracle JDK 8, 11, custom-installed from tar.gz (i.e. `/space/java/{8,11}`)

Note: custom JDK installation is not a requirement, just a setup of local environment used for the testing. Installation from DEB packages is also OK, provided that different JDK than 11 is selected as the default using _alternatives_.

### Example Application

- git@github.com:sdedic/micronaut-maven-exec-bug.git
- Note: this is just a boilerplate application generated at micronaut.io (steps above) and pushed to my git repo.

### Further information - impact, analysis / proposed fix

The behaviour breaks executon on a JDK different than system default. NetBeans IDE uses `JAVA_HOME` and `exec.executable` properties to run maven projects on user-specified JDK (jdk can be specified globally or in project's configuration). 

The core cause is that Micronaut's parent-pom (maven artifact `io.micronaut:micronaut-parent`)  defines literal `java` for executabe property of the `exec:exec` plugin:
```
        <plugin>
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>exec-maven-plugin</artifactId>
          <version>${exec-maven-plugin.version}</version>
          <configuration>
            <executable>java</executable>
            <!-- the rest of cfg omitted -->
          </configuration>
        </plugin>
 ```
The literal `java` breaks the contract specified in Maven documentation that `exec.executable` property in project's POM or commandline argument is used by default. 

I suggest that the `exec-maven-plugin` configuration references ${exec.executable} property, or does not specify any executable (which is the default) and the `exec.executable` property default is set up instead:
```
  <properties>
    <exec.executable>java</exec.executable>
  </properties>
  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>exec-maven-plugin</artifactId>
          <version>${exec-maven-plugin.version}</version>
          <configuration>
            <!-- Possibly omit entirely, as ${exec.executable} is the default for the MOJO -->
            <executable>${exec.executable}</executable>
            <!-- the rest of cfg omitted -->
          </configuration>
        </plugin>
 ```

It could be also possible to workaround in the IDE by e.g. adding the selected JDK to PATH variable first (see [NETBEANS-5393](https://issues.apache.org/jira/browse/NETBEANS-5393)), but IMHO Micronaut setup breaks the maven contract and should be fixed as well (or preferrably).

alvarosanchez:
Fixes #5020

### DIFF
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -125,7 +125,6 @@ ext.extraPomInfo = {
                     artifactId "exec-maven-plugin"
                     delegate.version '${exec-maven-plugin.version}'
                     delegate.configuration {
-                        executable("java")
                         arguments {
                             argument("-classpath")
                             classpath()
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -17,6 +17,7 @@ ext.extraPomInfo = {
         'maven.compiler.parameters'('true')
         'project.build.sourceEncoding'('UTF-8')
         'project.reporting.outputEncoding'('UTF-8')
+        'exec.executable'('java')
 
         "micronaut.version"(projectVersion)
         'micronaut-maven-plugin.version'(micronautMavenPluginVersion)


parent/build.gradle
```diff
@@ -125,7 +125,6 @@ ext.extraPomInfo = {
                     artifactId "exec-maven-plugin"
                     delegate.version '${exec-maven-plugin.version}'
                     delegate.configuration {
-                        executable("java")
```
sdedic, line 128: The default value for `exec.executable` project property (`<properties>` section) has to be defined for the application's POM to work out-of-the box. 

Maven does not specify any default for it (exec plugin seems to be designed as generic, although it falls back on `jdk` toolchain ;)), so if only java is omitted from exec plugin's config, the `mvn exec:exec` will fail with an error. 

Please see the pom snippet in #5020.

alvarosanchez: The idea was that whoever wants to use `exec:exec`, needs to pass `-Dexec.executable=/path/to/java`.

We can also add a `<property>`.

sdedic: Understood, I would vote in favour of adding the `<property>` to keep the user experience of working with the Micronaut projects, both current and the newly generated: if both proposed changes are made, current projects (even with the newer parent POM) will continue to work 'as they are' when run from the commandline. The Micronaut Launch web wizard also does not specify exec.executable in the project's pom.

alvarosanchez: Done.

graemerocher (approved):

PR #5038, opened 2021-03-03
Resolve generic type for bean properties

dstepanov:
Not really sure if it's the right way.

Fixes https://github.com/micronaut-projects/micronaut-data/issues/896

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/visitors/ClassElementSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/visitors/ClassElementSpec.groovy
@@ -210,6 +210,93 @@ interface AnotherInterface {
         accessibleFields*.name as Set == ['s1', 't1'] as Set
     }
 
+    void "test generic beans"() {
+        given:
+            ClassElement classElement = buildClassElement('''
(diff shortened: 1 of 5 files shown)


graemerocher:
Seems to cause a failing test

inject-java/src/test/groovy/io/micronaut/visitors/ClassElementSpec.groovy
```diff
@@ -210,6 +210,93 @@ interface AnotherInterface {
...
+        expect:
+            classElement.getBeanProperties().find(prop -> prop.name == "id").getType().getName() == "java.lang.Long"
+    }
+
+    void "test generic beans 2"() {
+        given:
+            ClassElement classElement = buildClassElement('''
+package genericbeans;
+
+class TaskGenericEntityY extends TaskGenericEntityX {
```
graemerocher, line 245: Seems to be invalid syntax

graemerocher (changes requested):

dstepanov:
Actually the solution is to use `getGenericType` on `TypedElement`.

Issue #5037, opened 2021-03-03
@CircuitBreaker(includes) does not work


l0co:
In `2.3.4`.

```java
@Singleton
public class DefaultRetryInterceptor implements MethodInterceptor<Object, Object> {

    private <T> Function<? super Throwable, ? extends Publisher<? extends T>> retryFlowable(MethodInvocationContext<Object, Object> context, MutableRetryState retryState, Flowable<Object> observable) {
        return exception -> {
            if (retryState.canRetry(exception)) {
                Flowable retryObservable = observable.onErrorResumeNext(retryFlowable(context, retryState, observable));
                long delay = retryState.nextDelay();
                if (eventPublisher != null) {
                    try {
                        eventPublisher.publishEvent(new RetryEvent(context, retryState, exception));
                    } catch (Exception e1) {
                        LOG.error("Error occurred publishing RetryEvent: " + e1.getMessage(), e1);
                    }
                }
                if (LOG.isDebugEnabled()) {
                    LOG.debug("Retrying execution for method [{}] after delay of {}ms for exception: {}", context, delay, exception.getMessage(), exception);
                }
                return retryObservable.delaySubscription(delay, TimeUnit.MILLISECONDS);
            } else {
                if (LOG.isDebugEnabled()) {
                    LOG.debug("Cannot retry anymore. Rethrowing original exception for method: {}", context);
                }
                retryState.close(exception);
                return Flowable.error(exception);
            }
        };
    }

}
```

This code shows that `CircuitBreaker(includes)` is not considered when the decision about opening circuit is about to be made (what happens in `retryState.close()`). It's only considered for underlying `@Retryable`. In the result, when I use `includes` with `CircuitBreaker`, and the thrown exception is not on the list, the retry policy won't be applied, but the circuit will be opened anyway.

l0co:
It's a simple fix. I've sent a PR.

PR #5039, opened 2021-03-03
Consider retryable predicate in circuit breaker

l0co:
Fix for #5037

### DIFF
--- a/runtime/src/main/java/io/micronaut/retry/RetryState.java
+++ b/runtime/src/main/java/io/micronaut/retry/RetryState.java
@@ -16,6 +16,8 @@
 package io.micronaut.retry;
 
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.retry.annotation.RetryPredicate;
+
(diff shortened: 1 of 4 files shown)


jameskleeh:
In general modifying existing tests is not preferred because it makes it impossible to tell if the changes broke the existing tests or not. That is an indicator that the change might be considered to be breaking. Just FYI for the future. You have contributed a lot to the framework and we really appreciate it! 🥇

jameskleeh:
I've created a new PR that targets 2.4 and makes the interface change backwards compatible

l0co:
OK, I understand the requirement but tests were written having the previous logic in the mind, ie. `includes` does not affect `@CircutBreaker`, only underlying `@Retryable`, so they failed. If I will send a PR with a breaking change like this in the future - should I left failed tests to be resolved by you?

jameskleeh:
You don't have to leave the test in a failing state, but instead change it the minimal amount possible to get it to pass so its easy to see what is changing

PR #5042, opened 2021-03-03
Json error format

jameskleeh:


jameskleeh:
The only thing missing is a config value that can be used to control the format. Then we can support multiple formats

graemerocher:
I'm just wondering what issue / discussion motivated this change? Otherwise looks fine in principal to me

jameskleeh:
@graemerocher Currently in order to use JSON API or Problem JSON for errors you have to replace 10+ beans. With this change that number is now 1.

PR #5068, opened 2021-03-08
feat: add problem+json support

sdelamo:
@jameskleeh this adds support for [Problem+Json](https://tools.ietf.org/html/rfc7807#section-3.1) on top of your changes to error handle. 

I have also extracted logic to @deprecated methods so that it is easy to remove them in Micronaut 3. 

I have addressed also the missing checkstyle.


graemerocher:
Wouldn't this be better in an external module?

sdelamo:
> Wouldn't this be better in an external module?

We will move this to an external module.

PR #5082, opened 2021-03-09
Initial support for dependency injection narrowing by generic types

### LINKED ISSUE #5079: Implement support for narrowing bean injection by generic type arguments
Micronaut doesn't support narrowing injection by generic type arguments yet, for example:

ie

```java
interface Foo<T> {}

class FooString implements Foo<String> {}
class FooInteger implements Foo<Integer> {}

@Inject Foo<String> fooString;
```

The above will currently fail with a non-unique bean exception.

graemerocher:
Fixes #5079

Probably needs more tests for more complex cases but this handles the basic case.

The fundamental change is bean keys are now `Argument` instances and the type and the generic types are factored into bean identity.


graemerocher:
Note this could break existing code if people were defining generic types in their injection points but not on the exposed beans.

jameskleeh:
@graemerocher I think this is just missing documentation. Should be in whatsNew, breaks, and the general dependency injection docs.

graemerocher:
@jameskleeh I provided docs and merged for now, if there is any other feedback will address in another PR

PR #5085, opened 2021-03-09
Improve performance of writing generated classes

dstepanov:
This speeds up writing classes significantly.

Micronaut JDBC data recompile goes from 1m 53s to 36s

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
@@ -16,6 +16,7 @@
 package io.micronaut.annotation.processing;
 
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.micronaut.core.reflect.ReflectionUtils;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.inject.writer.AbstractClassWriterOutputVisitor;
 import io.micronaut.inject.writer.ClassGenerationException;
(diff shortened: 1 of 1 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
```diff
@@ -96,7 +99,50 @@ public OutputStream visitClass(String classname, io.micronaut.inject.ast.Element
...
+        OutputStream os = javaFileObject.openOutputStream();
+        return unwrapFilterOutputStream(os);
+    }
+
+    private OutputStream unwrapFilterOutputStream(OutputStream os) {
+        // FilterOutputStream and JavacFiler$FilerOutputStream is always using write(int) and killing performance, unwrap if possible
+        if (os instanceof FilterOutputStream) {
+            try {
+                Field outField = ReflectionUtils.getRequiredField(FilterOutputStream.class, "out");
+                outField.setAccessible(true);
```
jameskleeh: I think this would be called for each file we create. Could refactor this so we only do the reflection bit once

graemerocher:
Root cause https://bugs.openjdk.java.net/browse/JDK-8255729

inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
dstepanov: Done

PR #5091, opened 2021-03-10
refactor: deprecated logic to @Deprecated methods

sdelamo:
Moves the deprecated logic to private method annoatated with `@Deprecated` so that it is easy to remove all that dead code in the future. 

This was part of https://github.com/micronaut-projects/micronaut-core/pull/5068 but it was not merged. Thus, I have extracted it to a separate PR.


jameskleeh (changes requested):
Is the goal of this to just make the code more organized? Deprecating private methods doesn't make much sense to me and given this code is unlikely to change I don't see the benefit of these changes

sdelamo:
Yes, the goal is to encapsulate deprecated functionality in private method and flag them so that they are easy to spot and to remove in micronaut 3.0.

PR #5096, opened 2021-03-10
add logic for when your configure usernames to endpoint sensitive health

oscarcitoz:
all detail about health. (PR RELATED micronaut-projects/micronaut-security#589)

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/health/HealthEndpoint.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/health/HealthEndpoint.java
@@ -16,6 +16,7 @@
 package io.micronaut.management.endpoint.health;
 
 import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.core.value.PropertyResolver;
 import io.micronaut.health.HealthStatus;
 import io.micronaut.http.HttpStatus;
 import io.micronaut.management.endpoint.annotation.Endpoint;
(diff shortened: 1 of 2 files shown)


Casvad (approved):

jameskleeh:
I'm not against adding in something that allows this functionality to be extensible, however this seems too specific of a feature to be added in to the core of the framework

oscarcitoz:
> I'm not against adding in something that allows this functionality to be extensible, however this seems too specific of a feature to be added in to the core of the framework

I understand your point, but this function is very necessary (in my opinion) because all authorized users can be sensitive to endpoints. However, this feature is optional and would not affect versions that do not have the configuration (it is very good to depend on the configuration), I still leave the pull request open in case you see the gain that can be obtained later (on security). It would be great if micronaut with only one configuration already lets you see the sensitive one (but not all users). 
that's why also my pull request micronaut-projects/micronaut-security#589

jameskleeh:
@oscarcitoz As I've said, I think this is too specific. You may want to restrict the details based on username, others may want to restrict details based on some other attribute of the user, some may want to do a database call to retrieve the information, etc... There is no clear sensible default, thus the only thing we should do in this case is to allow each application to control the behavior through some extensible API. If you're interested in contributing something like that, you are welcome to submit an initial plan for discussion in a github issue, and then follow up with a PR when the plan is agreed upon. I appreciate you making the effort to contribute the framework!

PR #5116, opened 2021-03-12
Doesn't enable WebSocketVisitor is http classes not present

### LINKED ISSUE #5114: Simple websocket app fails to compile in 2.4.0
See the example app at https://github.com/burtbeckwith/websocket_240

If you clone it and run `./gradlew compileJava` it fails with

```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':compileJava'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$3(ExecuteActionsTaskExecuter.java:186)
        ...
Caused by: java.lang.RuntimeException: java.lang.NoClassDefFoundError: io/micronaut/http/uri/UriMatchTemplate
        at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:163)
        ...
Caused by: java.lang.NoClassDefFoundError: io/micronaut/http/uri/UriMatchTemplate
        at io.micronaut.validation.websocket.WebSocketVisitor.visitMethod(WebSocketVisitor.java:73)
        ...
Caused by: java.lang.ClassNotFoundException: io.micronaut.http.uri.UriMatchTemplate
        ... 187 more
```

If you downgrade the app to 2.3.4 by changing the version in gradle.properties and dropping the application plugin version to 1.3.4 it compiles and runs successfully.

jameskleeh:
Fixes #5114

### DIFF
--- a/parent/build.gradle
+++ b/parent/build.gradle
@@ -167,6 +167,11 @@ ext.extraPomInfo = {
                                 artifactId "micronaut-validation"
                                 delegate.version '${micronaut.version}'
                             }
+                            delegate.path {
+                                groupId "io.micronaut"
+                                artifactId "micronaut-http"
+                                delegate.version '${micronaut.version}'
+                            }
                         }
                     }
                 }
--- a/validation/build.gradle
+++ b/validation/build.gradle
@@ -1,11 +1,11 @@
 dependencies {
     annotationProcessor project(":inject-java")
-    
+
     api project(":inject")
     api project(":core-reactive")
     api dependencyVersion("validation")
+    api project(":http")
 
-    compileOnly project(":http")
     compileOnly dependencyVersion("gorm")
     compileOnly project(":http-server")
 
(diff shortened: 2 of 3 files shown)


graemerocher:
I altered to PR to include 2 changes:

1. If the http classes are not present then the WebSocketVisitor is disabled. This fixes the compilation problem
2. I added micronaut-http to the parent POM which re-enables route/websocket validation by default since change 1) fixes the problem but alters the behaviour slightly in that routes will no longer be validated by default. I am actually not sure if this is desirable it may be better to include `micronaut-http` in generated project templates so the user has the option to remove route/websocket validation.

graemerocher:
Note that if we agree that 2.) should be included then I will alter the Gradle plugin to include `micronaut-http` but I am not sure that this is the right thing to do.

PR #5137, opened 2021-03-17
Groovy properties should expose read/write methods. Fixes #5129

### LINKED ISSUE #5129: 2.4 `withValue` is incorrectly generated for Groovy
```
 public Object withValueInternal(Object var1, Object var2) {
        return this.getDeclaringBean().instantiate(new Object[0]);
    }
```

graemerocher:
Currently `GroovyPropertyElement` doesn't expose read/write methods for the property which means they have to be handled differently.

This alters `GroovyPropertyElement` with new synthetic read/write methods so the behaviour is consistent with when getters/setters are defined.

Fixes #5129

### DIFF
--- a/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java
+++ b/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java
@@ -15,25 +15,19 @@
  */
 package io.micronaut.ast.groovy.visitor;
 
-import groovy.lang.MetaProperty;
-import io.micronaut.core.annotation.Nullable;
 import io.micronaut.ast.groovy.annotation.GroovyAnnotationMetadataBuilder;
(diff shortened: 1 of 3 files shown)


inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java
```diff
@@ -523,6 +529,43 @@ public ClassElement getType() {
...
+                                GroovyClassElement.this,
+                                annotationMetadata,
+                                getType(),
+                                getGenericType(),
+                                getGetterName(propertyName, getType())
+                        ));
+                    }
+
+                    private String getGetterName(String propertyName, ClassElement type) {
+                        final String prefix;
```
jameskleeh: There are methods in NameUtils to do this

graemerocher: Changed to use them, hopefully they are consistent with how Groovy behaves

PR #5140, opened 2021-03-17
Handle LogLevel.OFF in LoggerLevelsConfigurer

brianwyka:
Properly handle `OFF` as a valid `LogLevel` for disabling a logger.

### DIFF
--- a/runtime/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java
+++ b/runtime/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java
@@ -22,6 +22,7 @@ import io.micronaut.context.env.Environment;
 import io.micronaut.context.event.ApplicationEventListener;
 import io.micronaut.context.exceptions.ConfigurationException;
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.convert.ConversionService;
 import io.micronaut.core.naming.conventions.StringConvention;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.runtime.context.scope.refresh.RefreshEvent;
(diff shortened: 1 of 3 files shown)


jameskleeh:
"Otherwise, this will be incorrectly converted to boolean false." That seems like something that should be fixed rather than documented.

brianwyka:
@jameskleeh, this is part of default conversion service processing.  Here is a test that demonstrates:

https://github.com/micronaut-projects/micronaut-core/blob/2.4.x/core/src/test/groovy/io/micronaut/core/convert/DefaultConversionServiceSpec.groovy#L55

I feel like it would be a breaking change to fix it, unless there is a way to read the configuration properties differently in the `LoggerLeversConfigurer` without using that conversion service...

https://github.com/micronaut-projects/micronaut-core/blob/2.4.x/runtime/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java#L83

jameskleeh:
@brianwyka Yes it makes sense if the goal is to convert it to a boolean, but that doesn't seem like the intent here

brianwyka:
Correct. I'll see if I can work some magic in the `LoggerLeversConfigurer` to avoid the issue then, since it is very tightly scoped.  I'll update this PR if I am able to get a reasonable fix for it established.

jameskleeh:
@brianwyka Even if the issue can't be resolved upstream, the configureLogLevelForPrefix method can check for false

brianwyka:
Good call @jameskleeh, I'll employ that workaround if necessary.

jameskleeh:
This false value is coming from the YAML Parser directly. We never receive "OFF".

brianwyka:
Alright, yes, just a did a quick google search and seems to be part of `SnakeYml` :
https://bitbucket.org/asomov/snakeyaml/src/master/src/main/java/org/yaml/snakeyaml/resolver/Resolver.java#lines-32

brianwyka:
Thanks @jameskleeh.  Let me know if the tests look good.

jameskleeh:
@brianwyka Looks good, just need to fix the checkstyle

brianwyka:
Fixed @jameskleeh

jameskleeh:
Thanks @brianwyka

PR #5141, opened 2021-03-18
Docs - Fixed wrong annotation reference and name

grisotto:
In the docs, in the Http Client Filter section has wrong reference and name to an annotation. There is reference to @HttpFilterStereotype annotation, but the correct is @FilterMatcher annotation.

You can check this at one line above to this link in the docs: https://docs.micronaut.io/snapshot/guide/index.html#clientSample


jameskleeh:
Thanks!

PR #5156, opened 2021-03-20
Add support for Any annotation and improve BeanProvider interface

### LINKED ISSUE #5155: Add support for Any qualifier
CDI supports the notion of a `@Any` qualifier to inject bean providers and also obtain any instance without throwing non unique bean exceptions:

https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#programmatic_lookup

We need to add support for this as part of https://github.com/micronaut-projects/micronaut-core/issues/4658

graemerocher:
* Addresses problems with generic type resolution
* Makes Bean Provider mechanism pluggable
* Supports `@Any` annotation to allow injecting providers without throwing NoSuchBeanException or NonUniqueBeanException and meet CDI requirement
 * Deprecates / removes support for automatic factory beans from `Provider` implementations

Fixes #5155


inject/src/main/java/io/micronaut/context/BeanProvider.java
```diff
@@ -16,19 +16,101 @@
...
+     */
+    default boolean isUnique() {
+        return true;
+    }
+
+    /**
+     * <p>Determines if there is no bean that matches the required type and qualifiers.</p>
+     *
+     * @since 3.0.0
+     * @return true if no bean matches.
```
jameskleeh: This javadoc is wrong it seems.

jameskleeh (changes requested):
I think this needs tests to assert the behavior of `@Any` for non BeanProvider use cases to assert what happens with multiple beans of different orders, same order, and no beans.

inject/src/main/java/io/micronaut/context/BeanProvider.java
```diff
@@ -16,19 +16,101 @@
...
+     */
+    default boolean isPresent() {
+        return true;
+    }
+
+    /**
+     * Is the bean resolvable.
```
jameskleeh: Perhaps expand on this to indicate what resolvable means

inject/src/main/java/io/micronaut/context/BeanProvider.java
```diff
@@ -16,19 +16,101 @@
...
+     * @return True if it is
+     * @since 3.0.0
+     */
+    default boolean isResolvable() {
+        return isUnique() && isPresent();
+    }
+
+    /**
+     * Executes the given logic if the bean is present.
```
jameskleeh: This should probably explain that in the case of multiple beans only the first one is passed

graemerocher:
I have not yet implemented the part where `@Order` is taken into account to select the bean that will come in a second PR. 

Addressed the other feedback and for the moment merging this one.

PR #5167, opened 2021-03-23
Support fragmented frames

jameskleeh:


http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -315,9 +321,31 @@ protected void handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame ms
                         CloseReason.UNSUPPORTED_DATA
                 );
             } else {
+                ByteBuf msgContent = msg.content().retain();
+                if (!msg.isFinalFragment()) {
+                    frameBuffer.updateAndGet((buffer) -> {
+                        if (buffer == null) {
+                            buffer = UnpooledByteBufAllocator.DEFAULT.compositeBuffer();
```
jameskleeh: @graemerocher If there is a different allocator I should be using here please lmk

graemerocher: Should use `ctx.alloc().compositeBuffer()`

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -315,9 +321,31 @@ protected void handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame ms
                         CloseReason.UNSUPPORTED_DATA
                 );
             } else {
+                ByteBuf msgContent = msg.content().retain();
```
graemerocher, line 324: In general to avoid memory leaks this PR is going to also have to cleanup the `frameBuffer` reference inside `handlerRemoved` and possibly `exceptionCaught` otherwise this could trigger a memory leak I fear

graemerocher (changes requested):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/websocket/BinaryChatClientWebSocket.java
```diff
@@ -71,4 +73,10 @@ public void onMessage(
     public abstract Future<ByteBuf> sendAsync(ByteBuf message);
 
     public abstract Single<ByteBuffer> sendRx(ByteBuffer message);
+
+    public void sendMultiple() {
+        session.sendSync(new TextWebSocketFrame(false, 0, "hello"));
+        session.sendSync(new ContinuationWebSocketFrame(false, 0, " "));
+        session.sendSync(new ContinuationWebSocketFrame(true, 0, "world"));
```
graemerocher, line 80: One more change is I think a test needs to be added for more than 2 continuation frames since the way the code is written behaves differently depending if there are 2 or more which isn't covered by test I don't think.

graemerocher (changes requested):

Issue #5160, opened 2021-03-21
Json serialization for camelCased bean properties does not respect JsonView when snake_case setting is turned on


igorlev:
It looks like currently the `BeanIntrospectionModule` doesn't copy the **_filteredProperties** field when creating a new `BeanSerializerBuilder` from the original. This means that if an existing `BeanPropertyWriter` is used in the new builder if will ignore any JsonView information set (on the bean to be serialized) thereby potentially leaking fields that should be omitted. 

The reason this is not caught by the existing unit tests is because in most cases the original `BeanPropertyWriter`(s) are not actually used in the new builder but are replaced with `BeanIntrospectionPropertyWriter`(s) as intended. 

However, the way the `BeanIntrospectionModule` looks for properties to replace (either via JsonProperty or `BeanIntrospection`) fails if the user _also_ has the Jackson snake_case field name serialization option turned on. In that case the BeanIntrospection property name is keyed by camelCase while the field to be serialized is looked up by snake_case. 

The original `BeanPropertyWriter` is then preserved but without the **_filteredProperties** value it will ignore JsonView filters  and write the linked property out. 

So there are, potentially, two issues here:
1. **_filteredProperties** should be copied to the new builder
2. and/or bean introspection lookup should not use snake_case field names. 

### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Enable jackson.json-view, jackson.bean-introspection-module, jackson.property-naming-strategy: SNAKE_CASE
2. Create a class to be used with JsonView
3. Create bean with the above JsonView class set on some properties that should not be serialized
4. Serialize the bean with the micronaut configured ObjectMapper

### Expected Behaviour

Bean properties marked with the JsonView should not be serialized while all the others should be serialized.

### Actual Behaviour

Any Bean property that does not have JsonProperty set on it but does have JsonView and is camelCased will be written out even when it should be filtered.

### Environment Information

- **Operating System**: Big Sur 11.0.1
- **Micronaut Version:** 2.3.4
- **JDK Version:** 11

igorlev:
Can replicate this in my code but will put together a standalone unit test that highlights this issue. The _filteredProperties copy is easy to fix but the snake_case field lookup might be a bit more work.

igorlev:
[...]

PR #5170, opened 2021-03-24
 Fix JsonView filtering when snake_case and bean introspection are enabled together (#5160)

igorlev:
* Added test case for #5160
* Copy over filteredProperties from original BeanSerializerBuilder when creating a new one in BeanInstrospectionModule


graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #5194, opened 2021-03-30
docs(ioc): fix conum

simon04:
<1> and <2> are not rendered...


jameskleeh:
This isn't the right fix. It should be `subs="attributes+"`. Appreciate the PR though!

jameskleeh:
Fixed via https://github.com/micronaut-projects/micronaut-core/commit/d83c6323a02529af65f7cc352f9dee3da67481e2

PR #5197, opened 2021-03-31
feat: text/vnd.turbo-stream.html media type

sdelamo:
See: https://turbo.hotwire.dev/handbook/streams#streaming-from-http-responses

> Turbo knows to automatically attach <turbo-stream> elements when they arrive in response to <form> submissions that declare a MIME type of text/vnd.turbo-stream.html.


jameskleeh:
Is this content type used by any feature of Micronaut?

sdelamo:
I want to write turbo integration: 

https://github.com/micronaut-projects/micronaut-views/issues/156

graemerocher:
I think the constant can exist I the subproject. This is not common enough for core

PR #5198, opened 2021-03-31
feat: Media type problem json

sdelamo:
>  when formatted as a JSON document, it uses the "application/problem+json" media type.
see: https://tools.ietf.org/html/rfc7807#section-6.1
see: https://tools.ietf.org/html/rfc7807#section-6.2


jameskleeh:
Going to close because I don't think this is commonly used enough to go into core. It can exist in the subproject

PR #5200, opened 2021-03-31
feat: handle UnsupportedOperationException returning 501

sdelamo:


jameskleeh:
I can definitely see how 501 could be viewed as the appropriate response type for UnsupportedOperationException, however I'm not sure it meets the technical definition:

"The server does not support the functionality required to fulfill the request. This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource."

If 501 is only intended for request methods that are not recognized then I would say it doesn't match for UnsupportedOperationException. For curiosity's sake I tried this with boot and it returns 500.

jameskleeh:
@sdelamo Do you have any response to my thoughts?

sdelamo:
It is fine. We can close this PR.

PR #5202, opened 2021-03-31
Support for Custom Validator Message

brianwyka:
This allows for custom validator implementations to override the message template of a given annotation and to supply its own to the context.

This can be useful if validation is at class-level and is validating multiple fields, where the validation message needs to be different.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
@@ -47,23 +47,23 @@ import java.util.*;
  */
 public class AnnotationProcessingOutputVisitor extends AbstractClassWriterOutputVisitor {
 
+    private static final Field FILTER_OUTPUT_STREAM_OUT = ReflectionUtils.findField(FilterOutputStream.class, "out")
+        .map(field -> {
+            try {
+                addOpenJavaModules(FilterOutputStream.class, AnnotationProcessingOutputVisitor.class);
+                field.setAccessible(true);
+                return field;
+            } catch (Exception e) {
+                return null;
+            }
+        })
+        .orElse(null);
+
     private final Filer filer;
     private final Map<String, Optional<GeneratedFile>> metaInfFiles = new LinkedHashMap<>();
     private final Map<String, FileObject> openedFiles = new LinkedHashMap<>();
     private final Map<String, Optional<GeneratedFile>> generatedFiles = new LinkedHashMap<>();
     private final boolean isGradleFiler;
-
-    private static final Field FILTER_OUTPUT_STREAM_OUT = ReflectionUtils.findField(FilterOutputStream.class, "out")
-            .map(field -> {
-                try {
-                    addOpenJavaModules(FilterOutputStream.class, AnnotationProcessingOutputVisitor.class);
-                    field.setAccessible(true);
-                    return field;
-                } catch (Exception e) {
-                    return null;
-                }
-            })
-            .orElse(null);
 
     /**
      * @param filer The {@link Filer} for creating new files
--- a/runtime/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
+++ b/runtime/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
@@ -15,15 +15,33 @@
  */
 package io.micronaut.jackson.modules;
 
-import com.fasterxml.jackson.annotation.*;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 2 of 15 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/AnnotationProcessingOutputVisitor.java
```diff
@@ -47,24 +47,24 @@
...
+            try {
+                addOpenJavaModules(FilterOutputStream.class, AnnotationProcessingOutputVisitor.class);
+                field.setAccessible(true);
+                return field;
+            } catch (Exception e) {
+                return null;
+            }
+        })
+        .orElse(null);
+
```
brianwyka, line 61: Unrelated checkstyle fix...

runtime/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
```diff
@@ -55,7 +73,14 @@
...
 import java.lang.annotation.Annotation;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
```
brianwyka, line 83: Unrelated checkstyle fix...

validation/src/main/java/io/micronaut/validation/validator/constraints/ConstraintValidatorContext.java
```diff
@@ -48,5 +48,21 @@
      * @return The root bean under validation.
      */
     @Nullable Object getRootBean();
+
+    /**
+     * Sets a message template to be used for the validation error to override default.
+     *
+     * @param messageTemplateOverride the message to override default
+     */
+    default void overrideMessageTemplate(@Nullable final String messageTemplateOverride) { }
```
jameskleeh: this should throw UnsupportedOperationException

jameskleeh (changes requested):
Given this is a new feature it should target the 2.5.x branch. I'll let others weigh in on whether this approach is a good idea or not

brianwyka:
> Given this is a new feature it should target the 2.5.x branch. I'll let others weigh in on whether this approach is a good idea or not

Changed base branch to `2.5.x`

validation/src/main/java/io/micronaut/validation/validator/constraints/ConstraintValidatorContext.java
brianwyka: Done.

brianwyka:
@jameskleeh, who are the others here? Should we tag them for review?

validation/src/main/java/io/micronaut/validation/validator/constraints/ConstraintValidatorContext.java
```diff
@@ -48,5 +48,23 @@
      * @return The root bean under validation.
      */
     @Nullable Object getRootBean();
+
+    /**
+     * Sets a message template to be used for the validation error to override default.
+     *
+     * @param messageTemplateOverride the message to override default
+     */
+    default void overrideMessageTemplate(@Nullable final String messageTemplateOverride) {
```
graemerocher: Maybe just call this `messageTemplate`, explicitly prefixing with `override` seems an implementation detail

validation/src/main/java/io/micronaut/validation/validator/constraints/ConstraintValidatorContext.java
```diff
@@ -48,5 +48,23 @@
      * @return The root bean under validation.
      */
     @Nullable Object getRootBean();
+
+    /**
+     * Sets a message template to be used for the validation error to override default.
+     *
+     * @param messageTemplateOverride the message to override default
```
graemerocher: This method needs `@since 2.5.0`

validation/src/main/java/io/micronaut/validation/validator/constraints/ConstraintValidatorContext.java
```diff
@@ -48,5 +48,23 @@
...
+    default void overrideMessageTemplate(@Nullable final String messageTemplateOverride) {
+        throw new UnsupportedOperationException("not implemented");
+    }
+
+    /**
+     * Get the message template override.
+     *
+     * @return the message override
+     */
+    @Nullable default String getMessageTemplateOverride() {
```
graemerocher: Does this need to be `public` API? Seems it is only used internally. Either way it should return `Optional` and if it must be public be annotated with `@since 2.5.0`

graemerocher (changes requested):

brianwyka:
Ready for re-review @graemerocher

graemerocher (approved):

Issue #3508, opened 2020-06-17
Problem with @Recoverable with @Fallback


l0co:
In `2.0.0.RC1` after migration from `2.0.0.M3`.

I have the following test which now fails:

```java
@MicronautTest
public class SafeDeclarativeClientTest extends BaseIntegrationTest {

	public interface IFallbackTest2 {
		int getTwo();
	}

	@Singleton
	@Recoverable
	public static class PrimaryFallbackBean2 implements IFallbackTest2 {

		@Override
		@CircuitBreaker(includes = RuntimeException.class, attempts = "3", delay = "100ms", reset = "500ms")
		public int getTwo() {
			throw new RuntimeException();
		}
	}

	@Fallback(includes = RuntimeException.class)
	public static class SecondaryFallbackBean2 implements IFallbackTest2 {

		@Override
		public int getTwo() {
			return 2;
		}

	}

	@Inject protected PrimaryFallbackBean2 primaryFallbackBean2;

	@Test
	public void testCircuitBreakerWithFallback2() {
		assertEquals(2, primaryFallbackBean2.getTwo());
	}

}
```

For the above code in `2.0.0.M3` `RecoveryInterceptor.findFallbackMethod()` properly resolves declaring type to `IFallbackTest2` interface, while in `2.0.0.RC1` it's resolved to `SafeDeclarativeClientTest$PrimaryFallbackBean2`. The same happens for this version:

```java
@MicronautTest
public class SafeDeclarativeClientTest extends BaseIntegrationTest {

	@Recoverable
	public interface IFallbackTest2 {
		@CircuitBreaker(includes = RuntimeException.class, attempts = "3", delay = "100ms", reset = "500ms")
		int getTwo();
	}

	@Singleton
	public static class PrimaryFallbackBean2 implements IFallbackTest2 {

		@Override
		public int getTwo() {
			throw ReadTimeoutException.TIMEOUT_EXCEPTION;
		}
	}

	@Fallback(includes = RuntimeException.class)
	public static class SecondaryFallbackBean2 implements IFallbackTest2 {

		@Override
		public int getTwo() {
			return 2;
		}

	}

	@Inject protected IFallbackTest2 primaryFallbackBean2;

	@Test
	public void testCircuitBreakerWithFallback2() {
		assertEquals(2, primaryFallbackBean2.getTwo());
	}

}
```

graemerocher:
[...]

PR #5207, opened 2021-04-02
feat(documentation): Update clientFallback.adoc

thisdotnull:
Related to issue [#3508](https://github.com/micronaut-projects/micronaut-core/issues/3508). @Fallbacks are not invoked with declarative client if `api` value is not set. Adding this information to `TIP`.


ilopmar (approved):

ilopmar:
Thanks!

PR #5220, opened 2021-04-06
Update types.adoc

David-16:
Added information regarding injecting lists of beans in an aggregation pattern.  Not sure it needs its own section, so I made it a note


David-16:
Build failed in `micronaut-core/micronaut-core/tracing/build/test-results/` but I only changed a documentation file

If the two are related, I would be glad to fix it, but currently I do not know what to do

graemerocher:
unrelated failure. Thanks for the PR

Issue #5236, opened 2021-04-12
Parameter name lost during Gradle's incremental compilation


martijndwars:
### Task List

- [x] Steps to reproduce provided
- [x] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. `git clone https://github.com/MartijnDwars/mn-test`
2. `cd mn-test`
3. `./gradlew compileJava`
4. Update in `src/main/java/mn/test/ByeController.java` the text `"Bye $n"` to `"Byte $n+1"`.
5. `./gradlew compileJava`

The first `compileJava` runs fine, but the second `compileJava` fails with the following error:

```
> Task :compileJava FAILED
error: The route declares a uri variable named [id], but no corresponding method argument is present
Note: Creating bean classes for 2 type elements
1 error

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
1 actionable task: 1 executed
```

For what it's worth, if you run `compileJava` a third time it would succeed. Also, if you run `clean compileJava` you do not run into this problem, so this seems specific to some incremental compilation. I stumbled upon this issue when building through IntelliJ, which internally delegates to Gradle's `compileJava`.

### Expected Behaviour

Both `compileJava` tasks should succeed.

### Actual Behaviour

The second `compileJava` fails with the following error: `The route declares a uri variable named [id], but no corresponding method argument is present`. I believe the following is happening:

1. I change `ByeController.java`, so incremental compilation triggers only the recompilation of `ByeController.java`.
2. The Java compiler loads the already-compiled `HelloController.class` file.
3. The Java compiler ignores parameter names in this class file, even though the parameter names are present.
4. The annotation processor runs on `HelloController` (even though this class is already compiled and not changed).
5. The `MissingParameterRule` is unable to match the `arg0` parameter to the route and produces an error.

### Environment Information

- **Operating System**: macOS 10.15.7
- **Micronaut Version:** Using io.micronaut.application Gradle plugin version 1.4.2 (Micronaut 2.4.2).
- **JDK Version:** 1.8.0_281

### Example Application

See https://github.com/martijndwars/mn-test

[...]

PR #5248, opened 2021-04-14
Disable route validation on Java 8 incremental (#5236)

martijndwars:
As discussed in #5236. When the Java 8 compiler loads a _compiled_ class, the method parameters are named arg0, arg1, ... (even though the real names are available in the class file), and the route validation processor incorrectly produces an error. With this PR route validation is skipped when running on Java 8 or lower and incremental compilation is enabled.

Feedback welcome!

### DIFF
--- a/validation/src/main/java/io/micronaut/validation/routes/RouteValidationVisitor.java
+++ b/validation/src/main/java/io/micronaut/validation/routes/RouteValidationVisitor.java
@@ -48,6 +48,7 @@ import java.util.stream.Collectors;
 @SupportedOptions(RouteValidationVisitor.VALIDATION_OPTION)
 public class RouteValidationVisitor implements TypeElementVisitor<Object, Object> {
 
+    static final String MICRONAUT_PROCESSING_INCREMENTAL = "micronaut.processing.incremental";
     static final String VALIDATION_OPTION = "micronaut.route.validation";
     private static final String METHOD_MAPPING_ANN = "io.micronaut.http.annotation.HttpMethodMapping";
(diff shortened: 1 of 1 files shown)


validation/src/main/java/io/micronaut/validation/routes/RouteValidationVisitor.java
```diff
@@ -111,10 +112,44 @@ public void visitMethod(MethodElement element, VisitorContext context) {
...
+        String version = System.getProperty("java.version");
+        if (version.startsWith("1.")) {
+            version = version.substring(2, 3);
+        } else {
+            int dot = version.indexOf(".");
+            if (dot != -1) {
+                version = version.substring(0, dot);
+            }
+        }
+        return Integer.parseInt(version);
```
graemerocher: This could lead to a NPE if one of the conditions fail. They shouldn't fail, but you never know. Best to wrap in an a try/catch and return -1 if an exception occurs.

graemerocher (changes requested):

graemerocher:
Thanks, one small change and it can be merged

martijndwars:
Thanks! I considered adding a test case to avoid regressions, but I wasn't sure where to start. Are there any tests that exercise incremental compilation that I can use as inspiration?

graemerocher:
yeah not really, the incremental part is really hard to test :/

Maybe a test could be written here https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/src/test/groovy/io/micronaut/gradle/MicronautLibraryPluginSpec.groovy

But it would require us to first release a version of micronaut with this fix

PR #5260, opened 2021-04-16
issue 5255 fix

### LINKED ISSUE #5255: CircularDependencyException in HttpClientSenderFactory when instrument-threads is true
Thanks for reporting an issue, please review the task list below before submitting the

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (https://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Task List

### Steps to Reproduce

The issue occurs when using zipkin http tracing and setting `tracing.instrument-threads` configuration to true.

1. Pull [example application](https://github.com/tom-rauk/zipkin-http-sender)
2. ./gradlew run

### Expected Behaviour

Application starts up and traces sent via http send.

### Actual Behaviour

Application fails to start up with CircularDependencyException exception. 

```
Message: Circular dependency detected
Path Taken: 
new ExecutorServiceInstrumenter([List invocationInstrumenterFactories]) --> new ThreadTracingInvocationInstrumenterFactory([TracingInvocationInstrumenterFactory tracingInvocationInstrumenterFactory]) --> new BraveTracingInvocationInstrumenter([Tracing tracing]) --> Tracing.braveTracing([Reporter reporter]) --> AsyncReporter.asyncReporter([AsyncReporterConfiguration configuration]) --> new AsyncReporterConfiguration(BraveTracerConfiguration configuration,[Sender sender]) --> new HttpClientSenderFactory(HttpClientSenderConfiguration configuration,[List invocationInstrumenterFactories])
```

See [stack](https://raw.githubusercontent.com/tom-rauk/zipkin-http-sender/main/stacktrace.txt) for more details.

### Environment Information

- **Operating System**: MacOS Catalina 10.15.7 (19H512)
- **Micronaut Version:** 2.4.X
- **JDK Version:** 11.0.9.1

### Example Application

- https://github.com/tom-rauk/zipkin-http-sender

tom-rauk:
changed the http client sender factory to use Providers for the instrumentation factories for late bean resolution to resolve circular dependency introduced

This is to resolve: #5255

### DIFF
--- a/tracing/src/main/java/io/micronaut/tracing/brave/sender/HttpClientSenderFactory.java
+++ b/tracing/src/main/java/io/micronaut/tracing/brave/sender/HttpClientSenderFactory.java
@@ -23,6 +23,7 @@ import io.micronaut.tracing.brave.BraveTracerConfiguration;
 import zipkin2.reporter.Sender;
 
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
@@ -37,7 +38,7 @@ import javax.inject.Singleton;
 public class HttpClientSenderFactory {
 
     private final BraveTracerConfiguration.HttpClientSenderConfiguration configuration;
-    private final List<InvocationInstrumenterFactory> invocationInstrumenterFactories;
+    private final List<Provider<InvocationInstrumenterFactory>> invocationInstrumenterFactories;
 
     /**
      * Initialize the factory for creating Zipkin {@link Sender} with configurations.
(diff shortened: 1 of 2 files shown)


tracing/src/main/java/io/micronaut/tracing/brave/sender/HttpClientSenderFactory.java
```diff
@@ -46,8 +47,8 @@
      * @param invocationInstrumenterFactories The invocation instrumeter factories to instrument http client netty handlers execution with
      */
     protected HttpClientSenderFactory(
-            BraveTracerConfiguration.HttpClientSenderConfiguration configuration,
-            List<InvocationInstrumenterFactory> invocationInstrumenterFactories) {
+        BraveTracerConfiguration.HttpClientSenderConfiguration configuration,
+        List<Provider<InvocationInstrumenterFactory>> invocationInstrumenterFactories) {
```
jameskleeh, line 51: Given the class is public and this is protected this is a breaking change. Unfortunately its not as simple as adding a new constructor since the generic is just changing. This definitely shouldn't go into 2.4.x, however it could be argued the cost/benefit works out for 2.5. Thoughts @graemerocher ?

graemerocher: I think it is highly unlikely anyone is using this API and really it should probably have been marked as `@Internal` 

That said as you say it is a breaking change although things are broken as is without this change so if we can get it into 2.5.x we should

jameskleeh:
Really appreciate the PR @tom-rauk !

PR #5283, opened 2021-04-21
Annotation processing should let a visitor with a higher priority to process possible elements first

dstepanov:
`TypeElement`s can arrive in a random order, we need to make sure that a visitor with a higher priority is going to process all supported elements first.

Micronaut Data use-case: EntityMapper with a higher priority needs to process entities first before RepositoryMapper is going to process repositories and read entities.


dstepanov:
`ElementVisitor` is not working with only one visitor.

PR #5297, opened 2021-04-22
Performance/memory optimizations

dstepanov:


http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -446,13 +468,64 @@ public MediaType(String name, String extension, Map<String, String> params) {
...
             }
         }
-        if (params != null) {
-            parameters.putAll(params);
-        }
-
         this.strRepr = toString0();
     }
 
+    public static MediaType of(String name) {
```
graemerocher: Needs javadoc. Also parameter and return type should be annotated with `@NonNull`

http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -587,8 +660,14 @@ private String toString0() {
         if (parameters.isEmpty()) {
             return name;
         } else {
-            return name + ";" + parameters.entrySet().stream().map(Object::toString)
-                    .collect(Collectors.joining(";"));
+            StringBuilder sb = new StringBuilder(name);
+            for (Map.Entry<CharSequence, String> e : parameters.entrySet()) {
```
graemerocher: Would `parameters.forEach((key, value))` not be more efficient? As far as I understand that removes the need to create the  `Entry` instances

http/src/main/java/io/micronaut/http/MutableHttpHeaders.java
```diff
@@ -37,6 +37,8 @@
  */
 public interface MutableHttpHeaders extends MutableHeaders, HttpHeaders  {
 
+    ZoneId GMT = ZoneId.of("GMT");
```
graemerocher, line 43: needs javadoc

graemerocher (changes requested):
Overall looks good, just a few small changes

http/src/main/java/io/micronaut/http/MediaType.java
dstepanov: Yeah good catch

graemerocher (approved):

PR #5306, opened 2021-04-24
Init/Bean context performance improvements

dstepanov:


inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -81,7 +82,7 @@
     private static final String ORACLE_CLOUD_ASSET_TAG_FILE = "/sys/devices/virtual/dmi/id/chassis_asset_tag";
     private static final String ORACLE_CLOUD_WINDOWS_ASSET_TAG_CMD = "wmic systemenclosure get smbiosassettag";
     private static final String DO_SYS_VENDOR_FILE = "/sys/devices/virtual/dmi/id/sys_vendor";
-    private static final Boolean DEDUCE_ENVIRONMENT_DEFAULT = true;
+    private static final Boolean DEDUCE_ENVIRONMENT_DEFAULT = GraphicsEnvironment.isHeadless();
```
dstepanov: This is up for discussion

graemerocher: I don't think we should depend on AWT

We can make a decision to default to `false` for 3.0 or at least default to not make the network calls for certain clouds

jameskleeh: Perhaps move this change and any other changes related to environment deduction to a separate PR so we can merge the rest of this?

graemerocher: Agreed

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
```diff
@@ -94,7 +94,7 @@
      */
     @UsedByGeneratedCode
     public static Map<String, Object> getDefaultValues(String annotation) {
-        return ANNOTATION_DEFAULTS.computeIfAbsent(annotation, s -> Collections.emptyMap());
+        return ANNOTATION_DEFAULTS.getOrDefault(annotation, Collections.emptyMap());
```
dstepanov, line 97: This doesn't look like it was correct, inserting an immutable empty map where the value can be modified and some of the code checked if default values are present.

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
dstepanov: I have removed it for now

PR #5307, opened 2021-04-24
Added Support for specifying range on random number in properties

### LINKED ISSUE #5291: [Feature] support "configurable" random in PropertyResolver / or allow custom implementations
Feature request:
The property "${random.int}" does not help, because it returns any integer (that can be also negative,-2147483648 to 2147483647). I would like to extend the expression, allowing upper and lower bounds to the random, like used in spring. There might be situations, where you need some random values in a given range.

Background:
As an example, where this new "${random.int}" can be used, is for example a cron expression, like `0 ${random.int[0,20]} 2 * * *`, that allows to run the cron at a random hour between 0h and 22h. If you want to have multiple instances of your micronaut microservice, it makes sense, that the cron for example does not run on every instance at the same moment. With the random, you could better achieve, that the processing does not run at same time.

The changes must be done here: https://github.com/micronaut-projects/micronaut-core/blob/2.4.x/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java

### Ideas

1. allow giving/configuring the range of randoms in properties, like `${random.int[0,20]}`
2. I would also like the approach, if you can add own "property resolver" here, that would allow custom implementations as well. But for now, all related methods are private and not reuseable for a more common approach.

### Expected Behaviour

the property `${random.int[1,7]}` would return a value between 1 and 7 (would also work with long, etc.)

Happy to hear, what you might think about the request.

utsav0209:
Solves issue #5291 

The range can be defined on  `int`, `integer`, `long` and `float`

There are two ways to specify range now as follows: 
1. `random.integer(number)`
2. `random.integer[lower_bound, upper_bound]`

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -903,7 +903,7 @@ public class PropertySourcePropertyResolver implements PropertyResolver {
             int upperBound = Integer.parseInt(tokens[1]);
             return lowerBound + (int) (Math.random() * (upperBound - lowerBound));
         } catch (NumberFormatException ex) {
-            throw new ValueException("Invalid range: `" + range + "` found for type Integer while parsing property: " + property);
+            throw new ValueException("Invalid range: `" + range + "` found for type Integer while parsing property: " + property, ex);
         }
     }
 
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -53,7 +53,10 @@ public class PropertySourcePropertyResolver implements PropertyResolver {
 
     private static final Logger LOG = ClassUtils.getLogger(PropertySourcePropertyResolver.class);
     private static final Pattern DOT_PATTERN = Pattern.compile("\\.");
-    private static final Pattern RANDOM_PATTERN = Pattern.compile("\\$\\{\\s?random\\.(\\S+?)\\}");
+    private static final String RANDOM_PREFIX = "\\s?random\\.(\\S+?)";
+    private static final String RANDOM_UPPER_LIMIT = "(\\(-?\\d+(\\.\\d+)?\\))";
+    private static final String RANDOM_RANGE = "(\\[-?\\d+(\\.\\d+)?,-?\\d+(\\.\\d+)?])";
+    private static final Pattern RANDOM_PATTERN = Pattern.compile("\\$\\{" + RANDOM_PREFIX + "(" + RANDOM_UPPER_LIMIT + "|" + RANDOM_RANGE + ")?}");
     private static final char[] DOT_DASH = new char[] {'.', '-'};
     private static final Object NO_VALUE = new Object();
     private static final PropertyCatalog[] CONVENTIONS = {PropertyCatalog.GENERATED, PropertyCatalog.RAW};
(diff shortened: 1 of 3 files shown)


graemerocher:
Nice work, could you target 2.5.x since this is a new feature and it will be out soon?

utsav0209:
@graemerocher changed the target branch.

utsav0209:
@graemerocher added docs for the same

src/main/docs/guide/config/propertySource.adoc
```diff
@@ -160,6 +160,20 @@ micronaut:
...
+- `[min,max]` where min being inclusive and max being exclusive values.
+
+[source,yaml]
+----
+instance:
+  id: ${random.int[5,10]}
+  count: ${random.int(5)}
+----
+
+NOTE: The range could very from negative to positive as well.
```
graemerocher: Should be `vary` instead of `very`

graemerocher (changes requested):

graemerocher (approved):

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -885,6 +892,36 @@ private void fill(List list, Integer toIndex, Object value) {
...
+    private float getNextFloatInRange(String range) {
+        String[] tokens = range.split(",");
+        float lowerBound = Float.parseFloat(tokens[0]);
+        if (tokens.length == 1) {
+            return (float) (Math.random() * (lowerBound));
+        }
+        float upperBound = Float.parseFloat(tokens[1]);
+        return lowerBound + (float) (Math.random() * (upperBound - lowerBound));
+    }
+
```
jameskleeh: I think the only thing missing here is better error handling. If the user doesn't use the feature correctly they would get `NumberFormatException`, etc... however that wouldn't be a very helpful error message. This logic should catch any exceptions thrown by `parse` methods and report a more helpful message to indicate to the user what they did wrong.

utsav0209: @jameskleeh is the last commit I pushed is something you want?

jameskleeh: @utsav0209 I would instead throw a ValueException with the original passed in as the cause

utsav0209: @jameskleeh  Okay understood, will make the change and should I keep the Log traces?

jameskleeh: The exception will be logged so no need for additional logs

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -885,6 +893,48 @@ private void fill(List list, Integer toIndex, Object value) {
...
+        try {
+            String[] tokens = range.split(",");
+            int lowerBound = Integer.parseInt(tokens[0]);
+            if (tokens.length == 1) {
+                return lowerBound >= 0 ? 1 : -1  * (random.nextInt(Math.abs(lowerBound)));
+            }
+            int upperBound = Integer.parseInt(tokens[1]);
+            return lowerBound + (int) (Math.random() * (upperBound - lowerBound));
+        } catch (NumberFormatException ex) {
+            throw new ValueException("Invalid range: `" + range + "` found for type Integer while parsing property: " + property);
```
jameskleeh: This is missing the cause

graemerocher:
Thanks for the contribution!

Issue #5056, opened 2021-03-04
Micronaut URL Pattern documentation


tandel-pratik:
I couldn't find good examples of how the `@Filter` annotation takes in as values. `/**` doesn't compile using `java.util.regex.Pattern.compile`. Are there examples of what values are acceptable to pass in the `@Filter` annotation? 

On a related note is it possible to exclude certain patterns instead?

### Task List

- [ ] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [ ] Full description of the issue provided (see below)

### Steps to Reproduce

1. TODO
2. TODO
3. TODO

### Expected Behaviour

Tell us what should happen

### Actual Behaviour

Tell us what happens instead

### Environment Information

- **Operating System**: TODO
- **Micronaut Version:** TODO
- **JDK Version:** TODO

### Example Application

- TODO: link to github repository with example that reproduces the issue

pfyod:
Micronaut uses ant-style pattern matching. You can find some documentation here: https://docs.micronaut.io/latest/api/io/micronaut/core/util/AntPathMatcher.html

I do agree though that it needs to be properly documented within the main micronaut docs.

PR #5316, opened 2021-04-26
Added doc for Filter annotation path pattern matching

utsav0209:
Solves issue #5056


graemerocher:
Thanks for the contribution!

PR #5329, opened 2021-04-27
Refactor reactive routing

### LINKED ISSUE #5348: No longer possible to return Single<HttpStatus> in Micronaut 2.4 and above
This used to be supported in a controller:

```java
Single<HttpStatus> foo() {
     return HttpStatus.OK;
}
```

But has been broken since 2.4 and what happens is the `HttpStatus` object is encoded as JSON and a 200 is returned.

A test in discovery is failing due this regression.

### Expected Behaviour

Should emit the correct HTTP status

### Actual Behaviour

Encodes the enum as JSON and returns 200

### Environment Information

- **Micronaut Version:** 2.5
- **JDK Version:** any

### Example Application

n/a

dstepanov:


core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
```diff
@@ -225,6 +231,53 @@ protected void doOnComplete() {
...
+    /**
+     * Map the result from a publisher using the given mapper or supply empty value.
+     *
+     * @param publisher The publisher
+     * @param mapOrSupplyEmpty    The mapOrSupplyEmpty
+     * @param <T>       The generic type
+     * @param <R>       The result type
+     * @return The mapped publisher
+     */
+    public static <T, R> Publisher<R> mapOrSupplyEmpty(Publisher<T> publisher, MapOrSupplyEmpty<T, R> mapOrSupplyEmpty) {
```
graemerocher, line 245: Needs `@since`

core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
```diff
@@ -401,6 +454,32 @@ public static boolean isCompletable(Class<?> type) {
...
+     */
+    public interface MapOrSupplyEmpty<T, R> {
+
+        /**
+         * Maps next result.
+         *
+         * @param result The next value.
+         * @return The mapped value.
+         */
+        R map(T result);
```
graemerocher: Needs nullability annotations

core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
```diff
@@ -401,6 +454,32 @@ public static boolean isCompletable(Class<?> type) {
...
+
+        /**
+         * Supplies an empty value if there is no next value.
+         *
+         * @return The result.
+         */
+        R supplyEmpty();
```
graemerocher: Needs nullability annotations

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -386,13 +390,28 @@ private void exceptionCaughtInternal(ChannelHandlerContext ctx,
                         Object result = handler.handle(nettyHttpRequest, cause);
                         return errorResultToResponse(result);
                     });
-                    Supplier<Flowable<? extends MutableHttpResponse<?>>> publisherSupplier = () -> routePublisher;
```
graemerocher, line 389: We need to be careful here not to create regressions for https://github.com/micronaut-projects/micronaut-core/commit/be1bc41b4ac037dab48a10eeb25ec6df6059fc2f

However if this tests pass probably ok

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -425,6 +444,17 @@ private void exceptionCaughtInternal(ChannelHandlerContext ctx,
         }
     }
 
+    private List<InvocationInstrumenter> getInvocationInstrumenter() {
+        List<InvocationInstrumenter> instrumenters = new ArrayList<>(invocationInstrumenterFactories.size() + 1);
```
graemerocher: If there are no instruments can we return empty or will there always be one?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -1142,110 +1193,167 @@ protected void onComplete(MutableHttpResponse<?> message) {
...
-                                        ), throwable -> exceptionCaughtInternal(
-                                                context,
-                                                throwable,
-                                                request,
-                                                false
-                                        ));
+                                        bodyPublisher.subscribe(new CompletionAwareSubscriber<HttpResponse<?>>() {
+                                            @Override
+                                            protected void doOnSubscribe(Subscription subscription) {
+                                                subscription.request(Long.MAX_VALUE);
```
graemerocher: Do we not need to control back pressure here requesting 1 writing the message and then requesting the next?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -1383,165 +1491,228 @@ protected void doOnError(Throwable t) {
...
+                                MutableHttpResponse<?> singleResponse;
+                                if (o instanceof Optional) {
+                                    Optional optional = (Optional) o;
+                                    if (optional.isPresent()) {
+                                        o = ((Optional<?>) o).get();
+                                    } else {
+                                        return supplyEmpty();
+                                    }
+                                }
+                                if (o instanceof HttpResponse) {
```
graemerocher, line 1577: Need to check if `o` instanceof `HttpStatus`

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
dstepanov: Refactoring surprisingly eliminated the need for deferred publishers.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
dstepanov: Actually, we don't need instruments here at all.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
dstepanov: Original code didn't do it, but I think it makes sense to have it.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
dstepanov: Sound like a new feature, but it makes sense

jameskleeh:
Can you explain what code allows for `ServerRequestContextFilter.java` to be deleted?

dstepanov:
Code moved to `filterPublisher`:
```
            private void doSubscribe(Subscriber<? super MutableHttpResponse<?>> actualSubscriber, HttpRequest<?> request) {
                try {
                    ServerRequestContext.set(request);
                    upstreamResponsePublisher.subscribe(actualSubscriber);
                } finally {
                    ServerRequestContext.set(null);
                }
            }
```

dstepanov:
I think this can be merged @graemerocher @jameskleeh

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
lgathy: @dstepanov @graemerocher This might still have caused some regression as per my investigation of a bug appeared in v2.5.1+

The current request in `ServerRequestContext` was no longer set properly in filters. Now as I see this has been fixed for HTTP requests in v2.5.2 but not for WebSocket requests, I still see this when upgrading our projects through @RequestScoped beans throwing exceptions in filters.

I just managed to track down the regression to this change, I will try to put together some sort of test to reproduce the issue tomorrow to raise the issue, but in case you would have any hints on where this distinction between HTTP and WebSocket requests is located in the codebase or point me to the right direction on how to put together a websocket based test, that would be helpful.

jameskleeh: @lgathy This is where filters are executed for websocket upgrade requests https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java#L150

lgathy: @jameskleeh Thanks for the help. I found a way to reproduce this behaviour in tests (might not be the nicest way), so I raised an issue #5900 and soon will submit a draft PR to demonstrate it.

PR #5333, opened 2021-04-28
Reduce `stream` usage

dstepanov:


graemerocher (approved):

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -390,7 +395,10 @@ default boolean isDeclaredAnnotationPresent(@NonNull Class<? extends Annotation>
      * @return The annotation name
      */
     default Optional<String> getAnnotationNameByStereotype(@Nullable String stereotype) {
-        return getAnnotationNamesByStereotype(stereotype).stream().findFirst();
+        for (String s : getAnnotationNamesByStereotype(stereotype)) {
+            return Optional.of(s);
+        }
+        return Optional.empty();
```
jameskleeh: Is looping even beneficial here? You can just assign to a variable, `if (size > 0) return .get(0) else empty`

dstepanov: Looping is a equivalent to: get iterator; if (it.hasNext) return it.next

dstepanov: @jameskleeh `get(0)` is probably better because of not creating an iterator but then it looks a bit strange

jameskleeh: @dstepanov I don't think it would be strange. If the goal is to improve performance it seems reasonable to me

jameskleeh: Having said that I don't know if it would even perform better since you have to check the size. Would be interested to know

dstepanov: Iterator also needs to check the size at least once, `get` call should reduce the iterator allocation. I have modified a few more cases in the class.

dstepanov:
```
io.micronaut.docs.streaming.HeadlineControllerSpec testStreamingClient FAILED (3.1s)

  java.lang.AssertionError: Asynchronous error occurred: TimeoutException
      at org.junit.Assert.fail(Assert.java:89)
      at io.micronaut.docs.streaming.HeadlineControllerSpec.testStreamingClient(HeadlineControllerSpec.java:93)
      at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
      at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
      at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
      at java.lang.reflect.Method.invoke(Method.java:498)
      at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
```
Seems like unrelated timeout, publishers refactoring is not merged is it's not related to my changes.

PR #5335, opened 2021-04-28
Bean introspection: store constructor arguments in a field instead of recreating the array

dstepanov:
This is now exposing a mutable array, not sure if it makes sense to guard it by making a copy


graemerocher:
I guess creating a copy would add overhead?

dstepanov:
Not a big one

graemerocher:
probably best to create one in that case since could lead to weird bugs

dstepanov:
I have added an array copy

PR #5340, opened 2021-04-28
Introduce typed event publishers

dstepanov:


graemerocher:
I think given this is a new feature it is better if it targets 3.0.x and then we modify the existing `ApplicationEventPublisher` to include generic type arguments rather than introducing a whole new interface

dstepanov:
Replaced by https://github.com/micronaut-projects/micronaut-core/pull/5809

Issue #4797, opened 2021-01-11
@Factory should support producing beans from non-private fields


graemerocher:
Example:

```java
@Factory
class FooFactory {
     @Bean Foo foo = new Foo();
} 
```

Relates to #4658

graemerocher:
Thinking about this, there would have to be some another annotation to dictate a field as otherwise there would be trouble distinguishing between for example `@Client` and a bean producer

graemerocher:
Triaged as Feature and resolved. Support for `@Bean` fields was added by merged #5343, which explicitly fixes this issue, and the current factory documentation states that public and package-protected non-primitive fields are supported. Closing as completed with `closed: completed`.

PR #5343, opened 2021-04-29
Add support for producing beans from fields

graemerocher:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/4797


graemerocher:
@jameskleeh The failures are unrelated and seems to be because the `3.0.x` branch is currently broken due to this merge commit https://github.com/micronaut-projects/micronaut-core/runs/2448437879

PR #5352, opened 2021-04-30
Improve bean context startup performance

dstepanov:
Custom fork-join implementation performs much better than a parallel stream.
On my machine, a simple http app startup goes from best ~650ms to ~550ms.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
@@ -37,6 +37,7 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 
 import java.util.*;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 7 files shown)


dstepanov:
Parallel processing might not be needed for Graal

graemerocher (approved):

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -120,11 +121,20 @@ public DefaultApplicationContext(@NonNull ApplicationContextConfiguration config
     }
 
     @Override
-    protected @NonNull List<BeanDefinitionReference> resolveBeanDefinitionReferences() {
-        if (resolvedBeanReferences != null) {
+    protected @NonNull List<BeanDefinitionReference> resolveBeanDefinitionReferences(Predicate<BeanDefinitionReference> predicate) {
```
graemerocher, line 132: Can we retain the original signature and delegate just in case to avoid the breaking change

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -217,8 +227,8 @@ protected void initializeContext(List<BeanDefinitionReference> contextScopeBeans
     }
 
     @Override
-    protected <T> Collection<BeanDefinition<T>> findBeanCandidates(BeanResolutionContext resolutionContext, Class<T> beanType, BeanDefinition<?> filter, boolean filterProxied) {
-        Collection<BeanDefinition<T>> candidates = super.findBeanCandidates(resolutionContext, beanType, filter, filterProxied);
+    protected <T> Collection<BeanDefinition<T>> findBeanCandidates(BeanResolutionContext resolutionContext, Class<T> beanType, boolean filterProxied, Predicate<BeanDefinition<T>> predicate) {
+        Collection<BeanDefinition<T>> candidates = super.findBeanCandidates(resolutionContext, beanType, filterProxied, predicate);
```
graemerocher, line 245: Same here

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -609,14 +619,18 @@ protected BootstrapEnvironment createEnvironment(@NonNull ApplicationContextConf
         }
 
         @Override
-        protected @NonNull List<BeanDefinitionReference> resolveBeanDefinitionReferences() {
-            List<BeanDefinitionReference> refs = super.resolveBeanDefinitionReferences();
+        protected @NonNull List<BeanDefinitionReference> resolveBeanDefinitionReferences(Predicate<BeanDefinitionReference> predicate) {
+            List<BeanDefinitionReference> refs = super.resolveBeanDefinitionReferences(null);
```
graemerocher, line 637: Same here

graemerocher (changes requested):

PR #5354, opened 2021-04-30
Create Netty socket instance instead of using reflection to create

dstepanov:


graemerocher:
Seem to be breaking API changes in here, might need to be 3.0.x branch

dstepanov:
Which one? `DefaultHttpClient` is internal and the factory just introduces default methods.

graemerocher (approved):

graemerocher:
Good point

PR #5360, opened 2021-05-03
Support for introspections on public / package protected fields

graemerocher:


graemerocher:
Test failure is unrelated

jameskleeh:
I copied the tests over to inject-groovy and found some bugs with the element query logic. Unfortunately it seems that both `@PackageScope` and introspections occur in the same phase of AST thus the `@PackageScope` is not applied yet and it appears to the class node that the field is private. I reached out to Paul to see if anything has been done to allow ordering of transforms within a phase.

graemerocher:
Ok looks good PackageScope annotation is a bit niche so maybe not worth holding up the PR for it

PR #5361, opened 2021-05-03
Use `Map.of` and `List.of` for generated annotation metadata if available

dstepanov:


graemerocher:
Is there a way we can base this on the target byte code version? What if the user uses Java 11 but wants to compile java 8 compatible byte code?

dstepanov:
I think this is already working, when I compile core with Java 15 I don't get `of`s code.

graemerocher:
Have you tried compiling with JDK 11 with target =8 then running in 8?

jameskleeh:
I think this type of change would need to be really justified by significant performance gains in order to consider. The issues with the use of reflection and varying byte code based on conditions are points for unexpected failures.

dstepanov:
@graemerocher I did and it works.

@jameskleeh The best would be to switch to Java 9 but it's not an option yet.

Here is the quote from the official documentation describing the benefits of new immutable collections in Java 9:

> For optimal performance, the immutable collections store a data set that never changes. However, you may be able to take advantage of the performance and space-saving benefits even if your data is subject to change. These collections may provide better performance than the mutable collections, even if your data changes occasionally.
If you have a large number of values, you may consider storing them in a HashMap. If you are constantly adding and removing entries, then this is a good choice. But, if you have a set of values that never change, or rarely change, and you read from that set a lot, then the immutable Map is a more efficient choice. If the data set is read frequently, and the values change only rarely, then you may find that the overall speed is faster, even when you include the performance impact of destroying and rebuilding an immutable Map when a value changes.


https://docs.oracle.com/javase/9/core/creating-immutable-lists-sets-and-maps.htm

graemerocher:
@dstepanov One more thing, can you verify no reflective access warnings appear from the compiler in any JDK version

dstepanov:
The warning is only for accessing private/internal methods, I'm only accessing public API.

graemerocher:
@dstepanov can you rebase this for the 3.0.x branch. We should merge it into 3.0.x

dstepanov:
Alternative changes included in https://github.com/micronaut-projects/micronaut-core/pull/5672

PR #5374, opened 2021-05-04
Use latest h2 driver

dstepanov:


graemerocher:
we need to check if this version works in native image

jameskleeh:
I believe there are breaking changes in this version

EDIT: https://github.com/h2database/h2database/issues/2198

dstepanov:
This version is more than a year old I don't know why is it still not updated. Some of the Data and SQL projects have the newest version hardcoded

jameskleeh:
> This version is more than a year old I don't know why is it still not updated

Because it contains breaking changes.

dstepanov:
Looks like they just don't allow to use of an old configuration option that did nothing before anyway

graemerocher:
We would have updated it some time back but I believe it didn't work on native image hence why we didn't.

As for breaking changes we update other third party libraries that contain breaking changes in 2.x releases so I don't believe that is a justification for not updating it.

jameskleeh:
@graemerocher It definitely is a justification for not updating per https://github.com/micronaut-projects/micronaut-core/wiki/Module-Upgrade-Strategy. "If it contains breaking changes that have a real possibility of impacting our users, it should be included in the next major release of the module"

jameskleeh:
@dstepanov Imparting the burden on the users to discover why something was not failing previously and is failing now is the thing we need to avoid in patch and minor releases. Regardless of whether the change is inconsequential the user has to discover the issue, find out what has changed, determine the correct fix, and make the change in potentially many code bases. This isn't something users should have to do in a patch or minor release upgrade.

graemerocher:
Right well then we have committed dozens of such violations already I am afraid with various modules that have upgraded to new minor versions which have breaking behavior in them

ilopmar:
Before merging this we need to fix https://github.com/micronaut-projects/micronaut-data/issues/1007 so we can have a green CI.

ilopmar:
As Graeme mentioned, it seems that 1.4.200 (or .201) are not compatible with GraalVM: https://github.com/h2database/h2database/issues/2207

graemerocher:
Updated in the 3.0.x branch

PR #5377, opened 2021-05-04
Correct location of micronaut-validation native-image configuration

brianwyka:
Moved the `native-image.properties` for the `validation` module into the correct `META-INF` resources location.

This will fix the following issue which it was initially intended to:

```
Warning: class initialization of class io.micronaut.validation.exceptions.ValidationExceptionHandler failed with exception java.lang.NoClassDefFoundError: org/grails/datastore/mapping/validation/ValidationException. This class will be initialized at run time because option --allow-incomplete-classpath is used for image building. Use the option --initialize-at-run-time=io.micronaut.validation.exceptions.ValidationExceptionHandler to explicitly request delayed initialization of this class.
```


ilopmar (approved):
Good catch!
:man_facepalming: for me when fixing this sometime ago.

PR #5381, opened 2021-05-05
added doc text for using logback

andreashe:
As it did not use logback out of the box but it was mentioned, I wanted to add the solution to get it work.

### DIFF
--- a/src/main/docs/guide/config/propertySource.adoc
+++ b/src/main/docs/guide/config/propertySource.adoc
@@ -207,3 +207,28 @@ logger:
 
 
 Note that the ability to control log levels via config is controlled via the api:logging.LoggingSystem[] interface. Currently, Micronaut includes a single implementation that allows setting log levels for the Logback library. If you use another library, you should provide a bean that implements this interface.
+
+==== Enable logback usage
+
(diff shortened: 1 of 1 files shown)


jameskleeh:
I think this is a little out of place. It makes sense to me to create a top level section in the docs called "Logging" that describes the default setup and how to change it. The configuration for the loggers could be placed there as well

andreashe:
> I think this is a little out of place. It makes sense to me to create a top level section in the docs called "Logging" that describes the default setup and how to change it. The configuration for the loggers could be placed there as well

Good point. Do you mean an menu entry in the main document or a separate one?

jameskleeh:
@andreashe A top level menu item, like `13 Logging` that has a document explaining that we use Slf4j and the default implementation is logback, how to configure, maybe show an example of how to switch to log4j2 etc..

andreashe:
> @andreashe A top level menu item, like `13 Logging` that has a document explaining that we use Slf4j and the default implementation is logback, how to configure, maybe show an example of how to switch to log4j2 etc..

OK fine. I can create one with the knowledge I have. How to switch to log4j, I don't know. I guess it makes sense to move also the http logging to that page? I would also describe usage of lombok for logging.

Issue #5379, opened 2021-05-04
NoSuchBeanException is thrown on tests with MockBean and Named singletons


ednesic:
Some tests are throwing `NoSuchBeanException` when some Bean is mocked with `@MockBean` and is a dependency from a bean with `Named` qualifier. When the application tries to instantiate the named singleton, it also tries to instantiate the mock with the same Named qualifier and it fails.

### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [x] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Mock some bean that is a dependency from another bean with a Named qualifier
2. The test will try to instantiate this mocked bean with the same Named qualifier even it doesn't have this annotation

### Expected Behaviour

Should try to instantiate the mock without the Named qualifier

### Actual Behaviour

Tries to instantiate the mock with the Named qualifier of the dependent and fails because it really doesn't exists.

### Environment Information

- **Operating System**: OSX catalina
- **Micronaut Version:** 2.5.0
- **JDK Version:** 11

### Example Application

https://github.com/ednesic/micronaut-no-such-bean-wrong-qualifier

PR #5382, opened 2021-05-05
add test for issue #5379

graemerocher:
Seems like an erroneous qualifier is injected in the interceptor. Note sure what the solution is just yet.


graemerocher:
This is the only proposed fix I could think of. Maybe you want to take a look @jameskleeh if you have other ideas

PR #5386, opened 2021-05-05
Support implicit bean names for factories and injection points

graemerocher:
Support for `@Named` without a value https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#named_at_injection_point

And alignment with Spring/CDI to make beans from factories inherit the method / field name as the bean name.

Other changes in this PR were related to bugs that this change uncovered in the bean destruction logic


graemerocher:
@jameskleeh I'm going to merge this as I have other work I need to layer on top. If there is anymore feedback let me know and I will address it.

PR #5394, opened 2021-05-07
Improve HttpStreamsHandler

dstepanov:
My knowledge of the Netty pipelines is limited but this code is strange, I don't think all those `addListener` are needed, and especially triggering `fireChannelWritabilityChanged` doesn't makes sense. Reducing callbacks and pipeline triggers improves performance.


graemerocher:
@dstepanov can you test this under load? This code has been sensitive to threading issues before and want to make sure there are no regressions

dstepanov:
I did and everything is fine.

PR #5403, opened 2021-05-10
Disable downloading javadoc and sources in IntelliJ IDEA.

donbeave:
Otherwise, it takes quite a long time to download all sources and javadocs. But for those who need to use javadocs and sources they can download them from IDEA directly per one dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -897,6 +897,9 @@ allprojects {
         module {
             outputDir file('build/classes/java/main')
             testOutputDir file('build/classes/groovy/test')
+
+            downloadJavadoc = false
+            downloadSources = false
         }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -422,7 +422,7 @@ ext {
             picocli                    : [
                     version: picocliVersion,
                     group  : 'info.picocli',
-                    modules: ['picocli', 'picocli-codegen']                    
+                    modules: ['picocli', 'picocli-codegen']
             ],
             lettuce                    : [
                     version: lettuceVersion,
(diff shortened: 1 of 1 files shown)


jameskleeh:
I don't think this is something we should do. If you want to change how IDEA works it seems that should be done within IDEA

donbeave:
@jameskleeh you can not do this from IDEA UI, there is no option to change it. It's controlled by the Gradle IDEA plugin (https://docs.gradle.org/current/userguide/idea_plugin.html), which Micronaut is using, that's why during project importing IDEA downloading javadocs and sources because it's default values. But with these default values, for newbie who just cloned `micronaut-core` module and imported it in IDEA, they will also need to wait for a quite a long time to download all related javadocs and sources. If you know any other way how to do it, please let me know, I can not find an option to change it from the IDEA.

jameskleeh:
> they can download them from IDEA directly per one dependency.

@donbeave Can you elaborate on this further? How would one revert to the previous behavior?

donbeave:
I think can do it this way. Be default disable downloading javadoc/sources, and use [global Gradle properties](https://blog.mrhaki.com/2015/10/gradle-goodness-setting-global.html) to manipulate this behaviour, for those who really need download all these files, they will able to re-enabled these properties by editing their local `$HOME/.gradle/gradle.properties` file or can use system environments as well.

So default (if you don't set system env or global Gradle property) will be download sources by clicking the button from IDEA:

CleanShot 2021-05-12 at 00 09 58@2x

And you will be able to see the source class:

CleanShot 2021-05-12 at 00 11 30@2x

jameskleeh:
>they will able to re-enabled these properties by editing their local $HOME/.gradle/gradle.properties 

What property can be set to control this? Also what about Maven?

donbeave:
@jameskleeh there are 2 options:

1) system environments:

```
MICRONAUT_DOWNLOAD_JAVADOC=true
MICRONAUT_DOWNLOAD_SOURCES=true
```

2) gradle properties `$HOME/.gradle/gradle.properties`:

```
micronautDownloadJavadoc=true
micronautDownloadSources=true
```

jameskleeh:
I don't think that will work. Those properties would be read when Micronaut gets built, not when you pull the dependency via IDEA

donbeave:
@jameskleeh It's not. These properties will read when Gradle initializes for this project. When you import the Gradle project into IDEA, IDEA will start Gradle daemon, which will initialize `build.gradle` file and activates that idea module, because it's initialized in your project, which configure settings for IDEA which IDEA will use to import. That's all because you have such lines in your `build.gradle` file:

```
allprojects {
    // you are applying IDEA Gradle plugin for all submodules
    apply plugin: 'idea'

    idea {
        module {
           // by default downloadSources will be true here, it's default value, you can verify it in official Gradle github repo:
           // https://github.com/gradle/gradle/blob/06a728bc89c235edc621d02f0656b4ae244c3dd9/subprojects/ide/src/main/java/org/gradle/plugins/ide/idea/model/IdeaModule.java#L166
           ...
        }
    }
}
```

Source: https://github.com/micronaut-projects/micronaut-core/blob/b4a0265224d673cc70907b87c2ff4886e31e9698/build.gradle#L895

If you don't believe me, we can have a Zoom call and I will share with you how it works, and how these properties affect IDEA.

> Also what about Maven?

It's not related to Maven in any way. It's only about Gradle and IDEA.

PR #5405, opened 2021-05-10
More performance/memory optimizations

dstepanov:


router/src/main/java/io/micronaut/web/router/RouteMatch.java
```diff
@@ -226,4 +228,26 @@ default boolean isSatisfied(String name) {
...
     }
+
+    /**
+     * Finds predefined route http status or uses default.
+     *
+     * @param defaultStatus The default status
+     * @return The status
+     * @since 2.5.2
+     */
+    default HttpStatus findStatus(HttpStatus defaultStatus) {
```
graemerocher, line 241: Needs `@NonNull` annotations on the return type and parameter

router/src/main/java/io/micronaut/web/router/DefaultRouteBuilder.java
```diff
@@ -461,6 +477,37 @@ protected UriRoute buildBeanRoute(String httpMethodName, HttpMethod httpMethod,
...
+
+        private void setConsumesMediaTypesContainsAll() {
+            this.consumesMediaTypesContainsAll = consumesMediaTypes == null || consumesMediaTypes.isEmpty() || consumesMediaTypes.contains(MediaType.ALL_TYPE);
+        }
+
+        private void setProducesMediaTypesContainsAll() {
+            this.producesMediaTypesContainsAll = producesMediaTypes == null || producesMediaTypes.isEmpty() || producesMediaTypes.contains(MediaType.ALL_TYPE);
+        }
+
+        protected String resolveInputName(Argument requiredArgument) {
```
graemerocher: protected method needs javadoc and nullable annotations

graemerocher (changes requested):
Couple of small changes, otherwise looks good

dstepanov:
Some strange error
```
io.micronaut.scheduling.ScheduledCustomExecutorSpec test scheduled with a custom executor FAILED (10.2s)

  Condition not satisfied after 10.00 seconds and 101 attempts
      at spock.util.concurrent.PollingConditions.within(PollingConditions.java:170)
      at spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:140)
      at io.micronaut.scheduling.ScheduledCustomExecutorSpec.test scheduled with a custom executor(ScheduledCustomExecutorSpec.groovy:25)
  Caused by: Condition not satisfied:
  
  bean.cronEvents.get() == 2
  |    |          |     |
  |    6          6     false
  <io.micronaut.scheduling.Scheduled
```

graemerocher:
That is fixed by https://github.com/micronaut-projects/micronaut-core/pull/5438 which is waiting on @jameskleeh to decide how to take it forward

dstepanov:
Might be worse to fix it in 2.5 too

http-netty/src/main/java/io/micronaut/http/netty/NettyMutableHttpResponse.java
```diff
@@ -165,63 +264,38 @@ public HttpStatus getStatus() {
...
         message = message == null ? status.getReason() : message;
-        nettyResponse.setStatus(new HttpResponseStatus(status.getCode(), message.toString()));
+        httpResponseStatus = new HttpResponseStatus(status.getCode(), message.toString());
         return this;
     }
 
-    /**
-     * @return The Netty {@link FullHttpResponse}
-     */
-    public FullHttpResponse getNativeResponse() {
```
lgathy, line 187: @dstepanov @graemerocher Removing this public method breaks binary compatibility between patch versions. This makes it extremely difficult for external libraries to provide compatibility across at least different patch releases of the same minor version.

I know this is an internal class, but still, couldn't we leave it there marked as deprecated and returning `toFullHttpResponse()` instead? I believe the `replace(ByteBuf)` method may also suffer from the same issue.

Btw, the reason I needed to use this method was that this class does not override the `getReason()` method, thus I saw no other option to get any previously set custom reason phrase.

jameskleeh: > the reason I needed to use this method was that this class does not override the getReason() method

The thing to do would have been to file an issue to get your use case supported instead of relying on internal behavior. Internal classes are not part of the public API and treating them as such would defeat the purpose of having them in the first place

dstepanov: @lgathy I understand your pain, binary compatible might have been worse solution here as semantics changed and you aren’t getting stored instance but a new one. The best solution is to create a public api interface with netty specific methods.

lgathy: @jameskleeh you are absolutely right about raising an issue, I will do that. I understand using internal API is dangerous, in this instance, I had no other choice as I needed to support older Micronaut versions too.

lgathy: Raised an issue with an example app to demonstrate the current behaviour: https://github.com/micronaut-projects/micronaut-core/issues/5544

PR #5414, opened 2021-05-12
Fix regression with beans that implement iterable and filter order

jameskleeh:


core/src/main/java/io/micronaut/core/util/CollectionUtils.java
```diff
@@ -113,6 +113,18 @@ public static boolean isNotEmpty(@Nullable Collection collection) {
         return Optional.empty();
     }
 
+    public static <T> Optional<Iterable<T>> convertCollection(Class<? extends Iterable<T>> iterableType, Iterable<T> collection) {
```
graemerocher: Should be called `convertIterable` and include javadoc with `@since`

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
```diff
@@ -888,7 +888,7 @@ protected final Object getBeanForMethodArgument(BeanResolutionContext resolution
      */
     @SuppressWarnings("WeakerAccess")
     @Internal
-    protected final Collection getBeansOfTypeForMethodArgument(BeanResolutionContext resolutionContext, BeanContext context, MethodInjectionPoint injectionPoint, Argument argument) {
+    protected final Iterable getBeansOfTypeForMethodArgument(BeanResolutionContext resolutionContext, BeanContext context, MethodInjectionPoint injectionPoint, Argument argument) {
```
graemerocher: This is a binary incompatible change, not sure this belongs in a patch release

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
```diff
@@ -2253,14 +2248,23 @@ private Object resolveOptionalObject(Optional value) {
...
             return CollectionUtils.convertCollection(collectionType, beansOfType).orElse(null);
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private T[] coerceCollectionToArray(Class<T> componentType, Iterable<T> beansOfType) {
+        List<T> list = new ArrayList<>();
+        for (T t : beansOfType) {
+            list.add(t);
+        }
```
graemerocher: Shouldn't this use `convertIterable` the new static method added earlier?

graemerocher (changes requested):

PR #5420, opened 2021-05-13
Fix problems with generics injection / deprecate concrete type lookup

graemerocher:
This fixes a bunch of problems with generic lookup which was too complicated with the composite qualifier stuff so it is no simplified and a `isCandidateType` now just works with the generics which is better overall.

I also removed the behaviour that let you lookup a bean based on its concrete type but only after it had been created. This behaviour is inconsistent and wrong IMO and an explanation has been added to the `breaks.adoc`


graemerocher:
Test failures are unrelated

graemerocher:
@jameskleeh I'm going to merge this as I have other work I need to layer on top. If there is anymore feedback let me know and I will address it.

Issue #5418, opened 2021-05-13
Uninformative exception message in exception when provide unsupported properties file


aleksandr-chermashentsev:
### Steps to Reproduce

1. provide unsupported properties file format using VM argument for instance: -Dmicronaut.config.files=application.tbd

### Expected Behaviour

Exception which points that .tbd is unsupported format for properties file. List of supported formats also could be useful.

### Actual Behaviour

Exception "Unsupported properties file format: application" 

### Stacktrace
```io.micronaut.context.exceptions.ConfigurationException: Unsupported properties file format: application
	at io.micronaut.context.env.DefaultEnvironment.lambda$readPropertySourceListFromFiles$4(DefaultEnvironment.java:456) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at java.util.Optional.ifPresent(Optional.java:176) ~[?:?]
	at io.micronaut.context.env.DefaultEnvironment.readPropertySourceListFromFiles(DefaultEnvironment.java:438) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.context.env.DefaultEnvironment.readPropertySources(DefaultEnvironment.java:396) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.context.env.DefaultEnvironment.start(DefaultEnvironment.java:254) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.context.DefaultApplicationContext$RuntimeConfiguredEnvironment.start(DefaultApplicationContext.java:696) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.context.DefaultApplicationContext$RuntimeConfiguredEnvironment.start(DefaultApplicationContext.java:664) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.context.DefaultApplicationContext.startEnvironment(DefaultApplicationContext.java:212) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.context.DefaultApplicationContext.start(DefaultApplicationContext.java:164) ~[micronaut-inject-2.4.1.jar:2.4.1]
	at io.micronaut.runtime.Micronaut.start(Micronaut.java:71) [micronaut-context-2.4.1.jar:2.4.1]
	at io.micronaut.runtime.Micronaut.run(Micronaut.java:311) [micronaut-context-2.4.1.jar:2.4.1]
[...]

PR #5423, opened 2021-05-13
Use 'extension' field in exception message because it more informative then 'fileName'

aleksandr-chermashentsev:
Fix for issue https://github.com/micronaut-projects/micronaut-core/issues/5418

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
+++ b/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
@@ -453,7 +453,8 @@ public class DefaultEnvironment extends PropertySourcePropertyResolver implement
                             }
                             order++;
                         } else {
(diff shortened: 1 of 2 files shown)


graemerocher:
The tests will need updating too

jameskleeh:
The filename is shown because the extension doesn't tell you what file had the issue. You could improve this by showing both

aleksandr-chermashentsev:
> The filename is shown because the extension doesn't tell you what file had the issue. You could improve this by showing both

Fixed with using absolute path of configuration file in exception

jameskleeh:
I can see how it may be undesirable for an absolute file path to be logged. Please just use the file name

aleksandr-chermashentsev:
I have ran tests locally and they are good. What's wrong with build action?

jameskleeh:
Another PR made this change

PR #5433, opened 2021-05-17
Select the bean with the highest precedence as specified by Order annotation

graemerocher:
Alters the logic that would otherwise throw a `NonUniqueBeanException` to make a last effort to pick a bean based on the highest precedence specified.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -2349,7 +2349,7 @@ public class DefaultBeanContext implements BeanContext {
      *
      * @param beanType   The bean type
      * @param qualifier  The qualifier
-     * @param candidates The candidates
+     * @param candidates The candidates, always more than 1
      * @param <T>        The generic time
      * @return The concrete bean definition
      */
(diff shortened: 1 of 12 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -3158,13 +3155,40 @@ public boolean test(BeanDefinition<T> candidate) {
...
-                if (exactMatches.size() == 1) {
-                    definition = exactMatches.iterator().next();
-                } else if (throwNonUnique) {
-                    definition = findConcreteCandidate(beanClass, specifiedQualifier, candidates);
+                if (candidates.stream().anyMatch(candidate -> candidate.hasAnnotation(Order.class))) {
+                    // pick the bean with the highest priority
+                    final Iterator<BeanDefinition<T>> i = candidates.stream()
+                            .sorted((bean1, bean2) -> {
+                                int order1 = OrderUtil.getOrder(bean1, bean1);
+                                int order2 = OrderUtil.getOrder(bean2, bean1);
```
mattmoss: Is this a typo?  `bean2, bean1`

Should this be:  `int order2 = OrderUtil.getOrder(bean2, bean2);`

mattmoss (commented):
Potential typo

PR #5438, opened 2021-05-17
Propose fix

graemerocher:
This test keeps failing intermittently but I am not sure what this test is testing, please take a look @jameskleeh


jameskleeh:
I made this change

PR #5458, opened 2021-05-19
Generate fewer executable methods

graemerocher:
Moves the annotation stereotypes to the method level annotation to reduce the number of methods we generate


graemerocher:
Merging this, if there are any other concerns let me know.

PR #5463, opened 2021-05-19
Revert "Revert "More performance/memory optimizations""

jameskleeh:
Reverts micronaut-projects/micronaut-core#5454


ilopmar:
I'm waiting for a `Revert "Revert "Revert "More performance/memory optimizations"""` after this :joy:

jameskleeh:
@ilopmar I'll turn this revert train right around!

PR #5464, opened 2021-05-19
feat: HttpHeadersUtils to normalize HTTP Header name

sdelamo:
see: https://github.com/micronaut-projects/micronaut-azure/issues/183
see: https://github.com/micronaut-projects/micronaut-azure/pull/205

Not sure if this should be pointing to 3.0.x since don't have a 2.6.x branch.


jameskleeh:
@sdelamo Let's discuss this at the next engineering meeting. Seems wrong to list every header out manually

jameskleeh:
The headers should be stored in a data structure that is case insensitive instead of storing all of this data in a static map

PR #5487, opened 2021-05-23
Bug when a filter modify a request

johnowl:
When the first filter modifies a request, the next filter should receive the modified request. But it was receiving the original one.


johnowl:
Closing this PR because requests are immutable in Micronaut.

Issue #5425, opened 2021-05-14
Micronaut unjustly refuses WebSocket connections from Firefox 


BenTels:
Thanks for reporting an issue, please review the task list below before submitting the
issue. Your issue report will be closed if the issue is incomplete and the below tasks not completed.

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (https://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Task List

- [X] Steps to reproduce provided
- [X] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [X] Full description of the issue provided (see below)

### Steps to Reproduce

1. Set up a web socket server in Micronaut
2. Connect to it with Firefox 78 or 88 (and probably versions in between as well), or simulate the Firefox request in curl
3. Watch the connection be refused

### Expected Behaviour

The web socket session should be established

### Actual Behaviour

The connection is refused as a 400 bad request and the error message is {"message":"Not a WebSocket request","_links":{"self":{"href":"/topics/person","templated":false}}}

### Environment Information

- **Operating System**: Linux, kernel 5.3.18-lp152.75-preempt/ OpenSuSE LEAP 15.2
- **Micronaut Version:** 2.5.1 and 2.5.3
- **JDK Version:** 16.0.1
- **Firefox version:** 78.10.0esr (64-bit) and 88.0.1 (64-bit)

### Description

Experimenting with curl, and comparing Firefox to Chrome, I think the problem is in the connection header. This is the curl simulation for Firefox:

curl --include \
     --no-buffer \
     --header "Host: localhost:8080" \
     --header "User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:88.0) Gecko/20100101 Firefox/88.0" \
     --header "Accept: */*" \
     --header "Accept-Language: en-US,en;q=0.5" \
     --header "Accept-Encoding: gzip, deflate" \
     --header "Sec-WebSocket-Version: 13" \
     --header "Origin: http://localhost:3000" \
     --header "Sec-WebSocket-Extensions: permessage-deflate" \
     --header "Sec-WebSocket-Key: CLHpIeywCglp4c7R+/LIfw==" \
     --header "Connection: keep-alive, Upgrade" \
     --header "Pragma: no-cache" \
     --header "Cache-Control: no-cache" \
[...]

PR #5510, opened 2021-05-29
Check for multi-valued headers in a header map that encodes headers as strings

BenTels:
This is a suggested fix for issue #5425 , which is currently closed but is not fixed.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
@@ -117,8 +117,17 @@ public class NettyServerWebSocketUpgradeHandler extends SimpleChannelInboundHand
     public boolean acceptInboundMessage(Object msg) {
         if (msg instanceof NettyHttpRequest) {
             NettyHttpRequest<?> request = (NettyHttpRequest) msg;
-            if (request.getNativeRequest().headers().contains(HttpHeaderNames.CONNECTION.toString(), HttpHeaderValues.UPGRADE, true)) {
-                return request.getNativeRequest().headers().containsValue(HttpHeaderNames.UPGRADE, WEB_SOCKET_HEADER_VALUE, true);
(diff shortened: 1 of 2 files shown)


dstepanov:
Would you be able to add a test to one of the existing WS specs like `PojoWebSocketSpec` that replicated the problem?

BenTels:
> Would you be able to add a test to one of the existing WS specs like `PojoWebSocketSpec` that replicated the problem?

Yeah, sort of...

I've added a Spock version of a very nasty, low-level test I set up when working on the fix idea. However, I cannot run it because the test suite doesn't run on my machine. It wants to copy a file .../micronaut-core/http-server-netty/src/test/resources/public/??????-0.0.yml which sort of doesn't give me enough information to fix it. Still, I hope it helps...

By the way, on a totally unrelated note, my fix doesn't correct for the fact that the value of Upgrade might actually also be multi-valued. That will most likely not occur, but theoretically....

And on another, totally unrelated, note: have you guys taken into account that in HTTP/2 the WebSocket start doesn't use the Upgrade header?

dstepanov:
You should be able to send a request to the running test server without mocking anything

BenTels:
Like I said, I wouldn't know -- the test suite won't even start on my system. 

In any case, the problem is in the way Netty handles headers. The Connection header isn't parsed at all, so if you look for a Connection header with a value of "upgrade", you won't find it; in the case of Firefox, the value will be "keep-alive, Upgrade".

BenTels:
Sorry, but I give up. I figured out what the problem was in the processTestResources task (it's the yml file in the http-netty-server test resources with the Japanese name; gradle in my environment doesn't play well with UTF-16 filenames). With a lot of massaging and skipping of tests I can get the whole source tree to compile and I can verify the test I wrote. But I can't get the test with the test server in PojoWebSocketSpec working, it fails to start with a failure to inject a bean of type io.micronaut.http.server.codec.TextStreamCodec .

The test I created may be ugly, but it proves the point. I'm not going to get any further than that.

dstepanov:
No worries, I will try to add the test

jameskleeh:
I fixed this by reverting to the code that was in 2.4.x

PR #5514, opened 2021-06-01
Add simple integration with JBang

graemerocher:
This is a fairly non invasive change so I am targeting 2.5.x as it changes no existing code. With this in place you can define scripts like the following and execute them with JBang. See https://www.jbang.dev:

```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS io.micronaut:micronaut-bom:2.5.5-SNAPSHOT@pom
//DEPS io.micronaut:micronaut-http-server-netty
//DEPS io.micronaut:micronaut-inject-java
//DEPS io.micronaut.data:micronaut-data-jdbc
//DEPS io.micronaut.data:micronaut-data-processor
//DEPS io.micronaut.sql:micronaut-jdbc-hikari
//DEPS com.h2database:h2
//DEPS org.slf4j:slf4j-simple
//M:CONFIG datasources.default.url=jdbc:h2:mem:devDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE
//M:CONFIG datasources.default.schema-generate=CREATE_DROP
//M:CONFIG datasources.default.dialect=H2

package app;

import io.micronaut.http.annotation.*;
import io.micronaut.data.annotation.*;
import io.micronaut.core.annotation.*;
import io.micronaut.data.jdbc.annotation.*;
import io.micronaut.data.repository.CrudRepository;

@Controller("/books")
class BookController {
    private final BookRepository repository;
    BookController(BookRepository repository) {
        this.repository = repository;
    }
    @Get("/")
    Iterable<Book> list() {
        return repository.findAll();
    }
}

@MappedEntity
record Book(
    @Id @GeneratedValue @Nullable Long id,
    String title,
    int pages
) {}

@JdbcRepository(dialectName="H2")
interface BookRepository extends CrudRepository<Book, Long> {}
```

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/Micronaut.java
+++ b/context/src/main/java/io/micronaut/runtime/Micronaut.java
@@ -287,6 +287,16 @@ public class Micronaut extends DefaultApplicationContextBuilder implements Appli
     }
 
     /**
+     * Run the application for the given arguments. Classes for the application will be discovered automatically
+     *
+     * @param args The arguments
+     * @since 2.5.5
+     */
+    public static void main(String... args) {
+        run(new Class[0], args);
+    }
+
+    /**
      * Run the application for the given arguments.
      *
      * @param cls  The application class
--- a/context/src/main/java/io/micronaut/runtime/jbang/JBangIntegration.java
+++ b/context/src/main/java/io/micronaut/runtime/jbang/JBangIntegration.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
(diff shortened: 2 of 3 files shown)


jameskleeh:
Even though this is super simple, it seems to me it should probably be in an external module. We should have tests for the JBangIntegration class and could have a class with a main method in that module instead of adding one to Micronaut. In addition that would be a place to add the necessary documentation

graemerocher:
Yeah, I agree. At least this PR demonstrates what needs to be done for a module. Closing this one.

Issue #5527, opened 2021-06-02
Tracing library restricts setting JAEGER_SERVICE_NAME via environment variable


SubhadeepSen:
I have been using the micronaut tracing library with jaeger for getting the traces. The application name property has not been set in application.yml file. Now I'm trying to set the JAEGER_SERVICE_NAME via environment variable but it is never set. The value is always **application**.

Now when I update my application.yml file with application name property the JAEGER_SERVICE_NAME is also getting set to application name.

After going through the code, I've understood that setting JAEGER_SERVICE_NAME via environment variable will never be possible.

**_Reason:_**
1. It has been checked if JAEGER_SERVICE_NAME is set via System property. If it is not set, then System property is set for JAEGER_SERVICE_NAME with either application name or the default value which is **application**.

Reference: https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/tracing/src/main/java/io/micronaut/tracing/jaeger/JaegerConfiguration.java#L67

2. Now a method has been called to configure from environment variables. This method is present in jaeger-core library.

Reference: https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/tracing/src/main/java/io/micronaut/tracing/jaeger/JaegerConfiguration.java#L69

3.  fromEnv() method internally calls getProperty() method. Here it returns the JAEGER_SERVICE_NAME set in System property or environment variable value as default if System property is not set.

Reference: https://github.com/jaegertracing/jaeger-client-java/blob/318dad2808454f483785db270e066fed4e60160a/jaeger-core/src/main/java/io/jaegertracing/Configuration.java#L730

**_Conclusion:_** 
It will never consider the environment variable value as the System property was set by micronaut tracing library.

Can we also consider setting JAEGER_SERVICE_NAME via environment variable ?

if (StringUtils.isEmpty(System.getProperty(JAEGER_SERVICE_NAME)) && StringUtils.isEmpty(System.getenv(JAEGER_SERVICE_NAME))) {
 // existing code 
}

graemerocher:
Sure seems reasonable. PRs welcome.

SubhadeepSen:
Thanks, I'll work on this and raise the PR.

SubhadeepSen:
@graemerocher could you please help by approving the workflow(s)

PR #5529, opened 2021-06-02
setting JAEGER_SERVICE_NAME system prop if not present in system prop and env var

SubhadeepSen:
The PR contains the changes to allow user to set JAEGER_SERVICE_NAME via environment variable. More details can be found in the following link of the issue.

**_ISSUE-5527:_** https://github.com/micronaut-projects/micronaut-core/issues/5527


graemerocher:
Thanks for the contribution!

Issue #5535, opened 2021-06-03
Type element visitors that visit annotations not processed by Micronaut do not work incrementally


graemerocher:
If you have a `TypeElementVisitor` that processes an annotation not by default handled by Micronaut it is never invoked when incremental annotation processing is on.

This is because the annotation names to process only take into annotation mappers and transformers but not visitors

A workaround is available which is to define a no-op mapper but that is a hack.

PR #5537, opened 2021-06-03
Ensure visited annotations for visitors can be visited downstream

graemerocher:
if incremental compilation is enabled and a visitor adds a bean defining annotation to a ClassElement via a visitor with annotate(..) then that type is not subsequently processed downstream when incremental compilation is enabled. The annotations handled by visitors should be included in this list of supported annotations for the bean definition processor. This change fixes that.


graemerocher:
Fixes #5535

jameskleeh:
Is this only a problem for 3 or can it target 2.x?

graemerocher:
It could target 2.x I guess

graemerocher:
Looks like there is an issue with this approach since new visitors are created on each cycle. Guess I will have to try load only the classes. Working on a revised PR

PR #5540, opened 2021-06-04
Update config.adoc

nikbucher:
Provide a hint, where to store the application config file.

### DIFF
--- a/src/main/docs/guide/config.adoc
+++ b/src/main/docs/guide/config.adoc
@@ -1,6 +1,6 @@
 Configuration in Micronaut takes inspiration from both Spring Boot and Grails, integrating configuration properties from multiple sources directly into the core IoC container.
 
-Configuration can by default be provided in Java properties, YAML, JSON, or Groovy files. The convention is to search for a file named `application.yml`, `application.properties`, `application.json` or `application.groovy`.
+Configuration can by default be provided in Java properties, YAML, JSON, or Groovy files. The convention is to search for a file named `application.yml`, `application.properties`, `application.json` or `application.groovy` in the `config` directory from the `user.dir`.
 
 In addition, like Spring and Grails, Micronaut allows overriding any property via system properties or environment variables.
 


jameskleeh:
This isn't true. It looks in the `src/{sourceset}/resources` directory

nikbucher:
@jameskleeh thanks for reviewing. We're both partially right. It looks in the classpath (compile-time/preset/default config) as in the `$(user.dir)/config`dir (runtime config) as well. Sadly the config locations and the corresponing resolving-strategy isn't (well) documented, as far as I've seen. Or did I miss something?

jameskleeh:
@nikbucher Where are you seeing that it looks at `$(user.dir)/config`?

nikbucher:
I didn't find it in the documentation anywhere, so I tried until it worked because I couldn't believe there is no other way than `src/{sourceset}/resources`. Eventually I remembered Spring Boot looks in the `$(user.dir)/config` dir for configuration files too. (see https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.files) And it worked.

And here I just found the corresponding code in micronaut 
https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java#L84

jameskleeh:
@nikbucher That refers to the config folder at the root of the project, not at `user.dir`. It could be you are overriding the location in your project and incorrectly assumed that was the default

nikbucher:
@jameskleeh I wasn't sure, so I tried this in my project to be sure:

1. `./gradlew clean build installDist`
2. `cd build/install/<app-name>`
3. `./bin/<app-name>`-> error
4.  created `./config/application.yml` with the necessary entries
5. `./bin/<app-name>`-> worked

It seems it works for `user.dir` too. 
afaik `Paths.get("config/")` returns the 'config' dir from the `user.dir` path. (https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java#L477-L479)

jameskleeh:
> afaik Paths.get("config/") returns the 'config' dir from the user.dir path

I don't think that is true. It returns it from the current working directory. That may happen to be `user.dir` if you run an app from your home directory

nikbucher:
@jameskleeh Ah sorry, here we go. :-)  With `user.dir` I meant the system property (`System.getProperty('user.dir')`) which returns the current working directory.

jameskleeh:
For some reason I thought `user.dir` was the users home directory. Sorry about that. I think we've both been saying the same thing but I didn't understand. If the documentation isn't clear that that location is searched by default then we should make it so. I'm not sure that main file is the place to put that info and it should instead reference "the current working directory" instead of `user.dir` so others don't suffer the same fate I just did

nikbucher:
No problem, thank you a lot for your patience! I didn't realize it too that my `user.dir`-thing is cause for confusion. Sorry from my side too. Yes "the current working directory" is much clearer. I'd suggest something similar like here https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.files which shows the order of resolving too. Which would be the right place to document that?

nikbucher:
Great would be a list like https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config too. But I don't know, how this would look like for micronaut.

jameskleeh:
@nikbucher https://docs.micronaut.io/latest/guide/#propertySource under "Included PropertySource Loaders". There lists the different property sources. You can expand on the last 2 bullets to denote where the files are searched at.

PR #5543, opened 2021-06-06
make SimpleCookie#toString return a Set-Cookie compliant value

### LINKED ISSUE #5542: SimpleCookie#toString doesn't generate a valid value for Set-Cookie
Method toString of SimpleCookie currently returns a value similar to what would be possible to use on the Set-Cookie header but with some differences. It would be better to return a String value that is compliant with the HTTP specification.

### Task List

### Steps to Reproduce

```java
import io.micronaut.http.cookie.SameSite;
import io.micronaut.http.simple.cookies.SimpleCookie;

public class Application {
        System.out.println(
            SimpleCookie("NAME", "VALUE")
                .sameSite(SameSite.Lax)
                .maxAge(1)
                .domain(".micronaut.io")
                .path("/")
                .secure(true)
                .httpOnly(true)
                .toString()
        );
    }
}
```

### Expected Behaviour

```
NAME=VALUE; Domain=.micronaut.io; Path=/; Max-Age=1; Secure; HttpOnly; SameSite=Lax
```

### Actual Behaviour

```
NAME=VALUE, domain=.micronaut.io, path=/, maxAge=1s, secure, HTTPOnly, SameSite=Lax
```

### Environment Information

- **Operating System**: macOS Big Sur 11.3.1 (20E241)
- **Micronaut Version:** 2.5.4
- **JDK Version:** openjdk 11.0.11 2021-04-20 LTS

t0rr3sp3dr0:
Closes #5542

### DIFF
--- a/http/src/test/groovy/io/micronaut/http/simple/cookies/SimpleCookie.groovy
+++ b/http/src/test/groovy/io/micronaut/http/simple/cookies/SimpleCookie.groovy
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.simple.cookies
+
+import io.micronaut.http.cookie.Cookie
+import io.micronaut.http.cookie.SameSite
+import spock.lang.Specification
+
+class SimpleCookieSpec extends Specification {
+
+    def "SimpleCookie.toString() returns a set-cookie compliant string"() {
+        when:
+        String name = "NAME"
+        String value = "VALUE"
+        String domain = "DOMAIN"
+        String path = "PATH"
+        boolean httpOnly = true
+        boolean secure = true
+        long maxAge = 1
+        SameSite sameSite = SameSite.Lax
+
+        Cookie cookie = new SimpleCookie(name, value)
+                .domain(domain)
+                .path(path)
+                .httpOnly(httpOnly)
+                .secure(secure)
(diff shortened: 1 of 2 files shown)


t0rr3sp3dr0:
Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

graemerocher:
Needs a test

t0rr3sp3dr0:
@graemerocher, I've updated my commit with the requested test

http/src/test/groovy/io/micronaut/http/simple/cookies/SimpleCookie.groovy
```diff
@@ -0,0 +1,61 @@
...
+ */
+package io.micronaut.http.simple.cookies
+
+import io.micronaut.http.cookie.Cookie
+import io.micronaut.http.cookie.SameSite
+import spock.lang.Specification
+
+class SimpleCookieSpec extends Specification {
+
+    def "SimpleCookie.toString() returns a set-cookie compliant string"() {
```
jameskleeh, line 24: This test isn't testing that the string is compliant with the spec. This should instead pass the string into a decoder that would normally fail if it was malformed.

jameskleeh:
What is the use case here? In general we try to stay away from trying to write spec compliant code when solutions already exist to do so

t0rr3sp3dr0:
@jameskleeh, Micronaut is being used to implement the Envoy Authentication Service. It is a gRPC server that intercepts HTTP requests of services using Envoy and allows the requests and responses to be modified.

We are using it to provide OIDC authentication to our internal services and we need to manage Cookies manually, setting the value of the `Set-Cookie` header. I'm converting the Envoy objects to Micronaut objects (like HttpRequest, HttpResponse, and Cookie) to create controllers (with `@Singleton` instead of `@Controller`) so it is easier and more familiar to write code for it.

http/src/test/groovy/io/micronaut/http/simple/cookies/SimpleCookie.groovy
t0rr3sp3dr0: It would be possible to use `io.netty.handler.codec.http.cookie.ServerCookieDecoder` to do it, but it is not available in the classpath. Am I allowed to include it as a test dependency?

yawkat:
I don't think we should guarantee the format of `toString`. A separate method if anything. And that method should probably be in `NettyCookie` and be based on netty logic, not on our own.

You can always use netty's `ServerCookieEncoder` directly if you want, too.

PR #5550, opened 2021-06-07
Filter rework

jameskleeh:


core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
```diff
@@ -298,8 +298,8 @@ protected void doOnSubscribe(Subscription subscription) {
             @Override
             protected void doOnNext(T message) {
                 try {
-                    actual.onNext(message);
                     consumer.accept(message);
+                    actual.onNext(message);
```
jameskleeh, line 302: This is debatable whether it should be done. The `then` in the method implies to me it happens after the `onNext` but the javadoc isn't clear

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -639,57 +621,37 @@ private void handleStatusError(
...
-            NettyHttpRequest nettyHttpRequest,
+            NettyHttpRequest<?> request,
             MutableHttpResponse<?> finalResponse,
-            MediaType defaultResponseMediaType,
-            boolean skipOncePerRequest) {
+            MediaType defaultResponseMediaType) {
+
         AtomicReference<HttpRequest<?>> requestReference = new AtomicReference<>(request);
-        filterPublisher(
+        Flowable<MutableHttpResponse<?>> responsePublisher = Flowable.fromPublisher(filterPublisher(
```
dstepanov: Please, use a custom subscriber instead of returning Flowables code

jameskleeh: Why? We would have to implement a custom `onErrorReturn`

dstepanov:
Looks like this removes a lot kf what I have done for performance improvements

jameskleeh:
> Looks like this removes a lot kf what I have done for performance improvements

I don't think so

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
dstepanov: This change was made to improve performance by removing flowables flow

jameskleeh: I don't think we should be in the position of having to create a custom reactive library

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -1107,362 +1116,207 @@ private boolean isSingle(RouteMatch<?> finalRoute, Class<?> bodyClass) {
...
+                        boolean isReactive = routeMatch.isAsyncOrReactive() || Publishers.isConvertibleToPublisher(body);
+                        if (isReactive && Publishers.isConvertibleToPublisher(body)) {
+                            message.body(null);
+
+                            Class<?> bodyClass = body.getClass();
+                            boolean isSingle = isSingle(routeMatch, bodyClass);
+                            boolean isCompletable = !isSingle && routeMatch.isVoid() && Publishers.isCompletable(bodyClass);
+                            if (isSingle || isCompletable) {
+                                // full response case
+                                return ((Flowable<Object>) Publishers.convertPublisher(body, Flowable.class))
```
dstepanov, line 1167: This looks like the code that was before I removed

PR #5558, opened 2021-06-08
Delay emission of response headers until at least 1 item is emitted in a streaming response

### LINKED ISSUE #5488: Controller returning Flowable with error as a 1st item should/could return a non-200 response http code.
Hi Micronaut team, I wonder if there is a workaround for this.
This is possibly a feature request.

I have a Controller rest endpoint returning `Flowable` and I want to get non-200 in case the first item in the `Flowable` is actually an error (the same for Reactor `Flux`).
It looks like Micronaut "blindly" sends 200 without waiting for the first item which could be an error.

In case of `Single` or `Mono` Micronaut waits for the item and correctly returns 200 or executes the error mappers in case of an error.

### Steps to Reproduce

See the `ServerErrorSpec` modification at https://github.com/stepanv/micronaut-core/commit/81f3c0b6e2cab54da4da291caf00ef7d6aee890d

### Expected Behaviour

I need to be able to control the http error code if something gets wrong (in a reactive fashion) before I start to emit the reactive items.
I cannot use the approach with returning `HttpResponse<Flowable<?>>` described at https://github.com/micronaut-projects/micronaut-core/issues/5344 as I detect the error state during the execution of the `Flowable`, so the controller's method is already executed and returned.

### Actual Behaviour

It looks like Micronaut "blindly" sends 200 without waiting for the first item which could be an error.

### Environment Information

- **Micronaut Version:** 2.5.x

jameskleeh:
Fixes #5488

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/stream/ArrayBracketSubscriber.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/stream/ArrayBracketSubscriber.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.netty.stream;
+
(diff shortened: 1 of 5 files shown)


http-netty/src/main/java/io/micronaut/http/netty/stream/ArrayBracketSubscriber.java
```diff
@@ -0,0 +1,73 @@
...
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.http.netty.content.HttpContentUtil;
+import io.netty.handler.codec.http.HttpContent;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@Internal
+public class ArrayBracketSubscriber implements Subscriber<HttpContent> {
```
graemerocher: Make `final`. Does it need to be `public`?

graemerocher (changes requested):

PR #5563, opened 2021-06-09
Bump micronaut-grpc to 2.5.0

micronaut-build:


ilopmar:
We are upgrading a module minor version in a Micronaut patch release, but that's because the module is broken. See https://github.com/micronaut-projects/micronaut-grpc/pull/362

ilopmar:
Another option if we don't want to merge this PR is close this PR so we don't include the version by default. Users would need to upgrade manually to it, but, as it is broken I think we should merge it.
I'll leave the decision to @jameskleeh

PR #5565, opened 2021-06-09
adding config to add default keep alive header

leenanimbekar:


jameskleeh:
The configuration should be in `NettyHttpServerConfiguration` and the default value should maintain the current behavior

jameskleeh:
Also the "leenanimbekar1" github account needs to sign the CLA

leenanimbekar:
@jameskleeh  done, please re-review it.

graemerocher:
The setting seems confusingly named to me. Would be better if it was called `keepAliveOnServerError` and the javadoc updated appropriately to describe what it does since it only handles a particular scenario

leenanimbekar:
@graemerocher please check now, updated java docs and config name

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -1885,7 +1885,7 @@ private void writeFinalNettyResponse(MutableHttpResponse<?> message, HttpRequest
             if (!isHttp2) {
                 if (!nettyHeaders.contains(HttpHeaderNames.CONNECTION)) {
                     boolean expectKeepAlive = nettyResponse.protocolVersion().isKeepAliveDefault() || request.getHeaders().isKeepAlive();
-                    if (!expectKeepAlive || httpStatus.getCode() > 499) {
+                    if ((!expectKeepAlive || httpStatus.getCode() > 499) && !serverConfiguration.isKeepAliveOnServerError()) {
```
jameskleeh: This seems wrong because if the setting was set to true it would would always be keep alive, even on non server error states.

It should probably be `if (expectKeepAlive || httpStatus.getCode() < 500 || serverConfiguration.isKeepAliveOnServerError()) { KEEP_ALIVE } else { CLOSE }`

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/HttpResponseSpec.groovy
```diff
@@ -200,13 +207,102 @@ class HttpResponseSpec extends AbstractMicronautSpec {
...
+        def e = thrown(HttpClientResponseException)
+        e.response.status == HttpStatus.INTERNAL_SERVER_ERROR
+        e.response.header(HttpHeaders.CONNECTION) == 'keep-alive'
+
+        cleanup:
+        client.stop()
+        applicationContext.stop()
+    }
+
+    void "test connection header is not defaulted to keep-alive  when configured to false for > 499 response"() {
```
jameskleeh: I don't think this test adds value. Specifying the default and not specifying is the same thing

leenanimbekar: removed this one.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
leenanimbekar: updated. @jameskleeh agree this one is more clear.

jameskleeh:
Closing in favor of https://github.com/micronaut-projects/micronaut-core/pull/5585

PR #5569, opened 2021-06-09
Generate error when private methods are annotated with @Around advice

### LINKED ISSUE #5522: Warn or error out when AOP annotations are placed on private methods
We had some code that got refactored and a `@Transactional` annotation got placed on a private method accidentally. We only discovered it thanks to some failing tests. It'd be nice if Micronaut's annotation processors refused to compile or at least emitted a warning in such a case. Thanks!

mattmoss:
Fixes #5522

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/aop/simple/AnotherClass.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/aop/simple/AnotherClass.groovy
@@ -29,11 +29,6 @@ class AnotherClass<A> {
         return "Name is " + name
     }
 
-    // protected methods not proxied
-    private String testPrivate(String name) {
-        return "Name is " + name
-    }
-
     String test(String name) {
         return "Name is " + name
     }
--- a/inject-groovy/src/test/groovy/io/micronaut/aop/simple/AnotherClass.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/aop/simple/AnotherClass.groovy
@@ -29,6 +29,11 @@ class AnotherClass<A> {
         return "Name is " + name
     }
 
+    // protected methods not proxied
+    private String testPrivate(String name) {
+        return "Name is " + name
+    }
+
     String test(String name) {
         return "Name is " + name
     }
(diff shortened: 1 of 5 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -804,6 +804,10 @@ public Object visitExecutable(ExecutableElement method, Object o) {
                             (isExecutableThroughType(method.getEnclosingElement(), methodAnnotationMetadata, annotationMetadata, modifiers, isPublic) ||
                                     hasAroundStereotype(annotationMetadata));
 
+            if (javaMethodElement.isPrivate() && hasAroundStereotype(annotationMetadata)) {
```
jameskleeh: Perhaps using `isExecutable` here is more complete?

jameskleeh: Oh I see that variable includes `!hasInvalidModifiers`. I would say to remove that condition then check it inside the `if (isDeclaredBean && isExecutable) {` condition. This check also needs done for Groovy

mattmoss: I had initially replicated this in Groovy AST code, but private methods don't show up as private.

inject-groovy/src/test/groovy/io/micronaut/aop/simple/AnotherClass.groovy
```diff
@@ -29,11 +29,6 @@ class AnotherClass<A> {
         return "Name is " + name
     }
 
-    // protected methods not proxied
```
jameskleeh: Can you revert this? This method should not cause a problem

mattmoss: Reverted.

dstepanov:
Is it fixing https://github.com/micronaut-projects/micronaut-data/issues/882 ?

jameskleeh:
@dstepanov Yep

Issue #5557, opened 2021-06-08
Explore removing generating of GraalVM reflect config with for `@Introspected`


ilopmar:
Explore if our GraalVM tests applications work without the GraalTypeElementVisitor generated GraalVM reflect-config configuration for classes annotated with `@Introspected`.

This was mandatory at the very beginning for Jackson serialization but using Micronaut Jackson module shouldn't be necessary.
If that works we should remove that code from the Graal visitor and expand the `@ReflectiveAccess` annotation to allow annotate classes and have the same behaviour that we have now with `@Introspected` for GraalVM.
This way `@Introspected` will only be used for generation Bean Introspection Metadata but not for anything related to GraalVM.

ilopmar:
I published a snapshot version without the `GraalTypeElementVisitor` included `@Introspected` classes in the GraalVM `reflect-config.json` file and then I ran the GraalVM CI: https://gitlab.com/micronaut-projects/micronaut-graal-tests/-/pipelines/317529342

There are some know failures like rabbitmq, mqtt, cache,... that we know don't work at this moment with GraalVM until we do some changes. For the rest of the applications only the Micronaut Data JPA and the jOOQ apps failed. It happens because in Micronaut Data we automatically make all `@Entity` to be `@Introspected` and in the jOOQ application we have configure it to [automatically generate JPA annotations](https://github.com/micronaut-graal-tests/micronaut-jooq-graal/blob/24750a130259e54c2514fe6d9be8de594cc2b8cd/build.gradle#L85).
We need to modify the Micronaut Data mapper to make the entities annotated `@ReflectiveAccess` and then modify the visitor to take care of if and generate the appropriate GraalVM config.

graemerocher:
The logic is part of core not data https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/inject/src/main/java/io/micronaut/inject/beans/visitor/EntityIntrospectedAnnotationMapper.java

ilopmar:
That probably explains why the mapping I'm trying in https://github.com/micronaut-projects/micronaut-data/blob/5b7feab2be9736f7f72aa37f7066db3e19ee769c/data-processor/src/main/java/io/micronaut/data/processor/mappers/MappedEntityMapper.java doesn't work :sweat_smile: 

Thanks for the hint!

ilopmar:
Closed by #5574

PR #5570, opened 2021-06-10
Test for checking if a ClassElement is an enum

ilopmar:
The test fails checking if a `ClassElement` is an enum. I need to check this for #5557


jameskleeh:
Should be a test for GroovyClassElement too

ilopmar:
done

ilopmar (commented):
Tried in my branch and it works! :-)

PR #5574, opened 2021-06-11
Do not use `@Introspected` for generating GraalVM reflect config. Use `@ReflectiveAccess`

### LINKED ISSUE #5557: Explore removing generating of GraalVM reflect config with for `@Introspected`

ilopmar:
Fixes #5557

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/ReflectiveAccess.java
+++ b/core/src/main/java/io/micronaut/core/annotation/ReflectiveAccess.java
@@ -26,7 +26,7 @@ import java.lang.annotation.Target;
  * @author graemerocher
  * @since 1.1
  */
-@Target({ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.TYPE})
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.TYPE, ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 @Experimental
 public @interface ReflectiveAccess {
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/micronaut/core/annotation/ReflectiveAccess.java
```diff
@@ -21,15 +21,14 @@
...
 
 /**
- * Descriptive annotation that can be used to declare a field, method, constructor etc for reflective access.
+ * Descriptive annotation that can be used to declare a field, method, constructor and types for reflective access.
  *
  * @author graemerocher
  * @since 1.1
  */
-@Target({ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR})
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.TYPE})
```
graemerocher: Maybe add `ElementType.ANNOTATION_TYPE` as well so it can be a meta annotation

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/annotation/ReflectiveAccess.java
ilopmar: Done

PR #5576, opened 2021-06-11
Provide better error message with Adapter on Kotlin suspend method.

### LINKED ISSUE #5548: Better MethodAdapters error handling for annotating suspend functions
I tried annotating a `suspending` function with `@EventListener` and `@TransactionalEventListener` and it is not actually possible to do this. But the error log did not make this easy to find out.

```kotlin
@Singleton
open class TopScoreWinnerService {
    @EventListener
    open suspend fun onClassWinnersEvent(event: ClassWinnersEvent) {
       ...
    }
}
```

Error log:
```
Caused by:
java.lang.IllegalStateException: Cannot adapt method [void onApplicationEvent(Object event)]. Target method [onClassWinnersEvent] not found on bean class com.beable.services.TopScoreWinnerService
at io.micronaut.aop.chain.AdapterIntroduction.lambda$new$1(AdapterIntroduction.java:75)
```

The fix is to remove the `suspend` keyword:
```kotlin
@Singleton
open class TopScoreWinnerService {
    @EventListener
    open fun onClassWinnersEvent(event: ClassWinnersEvent) = runBlocking {
       ...
    }
}
```

It would be nicer if Micronaut told me explicitly that suspending functions were not allowed. Thanks.

mattmoss:
Fixes #5548

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2021 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -1258,6 +1258,11 @@ protected void accept(DeclaredType type, Element element, AopProxyWriter aopProx
                             int paramLen = targetParameters.size();
                             if (paramLen == sourceParams.length) {
```
jameskleeh, line 1259: I was under the impression this check would return false for the example in the issue because the number of parameters in the method is different than the number of methods in the interface method

jameskleeh: Did some looking into. It appears the JavaMethodElement separates out the continuation argument thus getParameters() returns all other params

mattmoss: Yup. I initially had a "fix" in a different location, then found the same thing when it wasn't hitting my change.

PR #5580, opened 2021-06-14
fix: incorrect logger

sdelamo:


jameskleeh:
I think this was done intentionally. It should probably just be `DefaultHttpClient` for Micronaut 3

PR #5590, opened 2021-06-14
Bring back stereotype and annotation mapper changes from 3.x. Add stereotype processing for repeatable

### LINKED ISSUE #5588: Annotation types by stereotype not respecting annotation mapping
Using the Micronaut 2.5.5 version with rabbitMQ the listener receiving the same value for both the parameter

### Task List

### Steps to Reproduce

1. Clone the repo and build and run the application using Gradle
2. On the browser enter the URL http://localhost:8080/message
3. On the `Consumer` class put a break point `consumeMessage` and check the value for `messageFirst` & `messageSecond`

### Expected Behaviour

The value for `messageFirst & messageSecond` in `Consumer` class for the method `consumeMessage` should contain `first & second`

### Actual Behaviour

The value for `messageFirst & messageSecond` in `Consumer` class for the method `consumeMessage` is  contain `second & second`

### Environment Information

- **Operating System**: Mac 
- **Micronaut Version:** 2.5.5
- **JDK Version:** 16

### Example Application

- https://github.com/anandjaisy/micronautRabbitMq

jameskleeh:
Fixes #5588


graemerocher (approved):
If this is primarily back port then it looks ok to me

PR #5597, opened 2021-06-15
Fix multiple factory replacements

PhilHardwick:
The performance improvements in 2.5.0 changed existing functionality for when you have a factory replacing another factory and a further one replacing that one. This fixes this change to be in line with 2.4.x behaviour. Whether this is the desired behaviour it up for debate, but I think it's intuitive to be able to chain @Replaces annotations.


graemerocher:
@dstepanov can you take a look?

dstepanov:
Looks good! Thanks, @PhilHardwick!

dstepanov (approved):

Issue #5615, opened 2021-06-18
When injecting a collection of strings, spaces are not trimmed.


eutkin:
When we inject a collection of strings from the configuration into a bean, there are spaces in the elements of the collection.

Example:

application.yml
```yml
untrimmed: one, two
```

Bean:

```kotlin
@Singleton
class Test(@Value("\${untrimmed}") val untrimmed : Set<String>)
```

In field of Test contains `one`, ` two`.

### Steps to Reproduce

1. Create application based by Micronaut
2. Put in the `application.yml` property with value `one, two`
3. Inject this value into bean as `Collection<String>`
4. assert second collection element with `two`

### Expected Behaviour

Collection must be contains `one`, `two`

### Actual Behaviour

Collection contains `one`, ` two`

### Environment Information

- **Operating System**: Mac OS
- **Micronaut Version:** 2.5.6
- **JDK Version:** 11

### Example Application

- [Example](https://github.com/eutkin/untrimmed-collection-item)

jameskleeh:
I don't necessarily think we should do this. If you don't want a space don't put one in the config

jameskleeh:
Closing this because we feel the current behavior is a more sensible default

sdelamo:
@eutkin could you use?

```yaml
untrimmed: 
    - one
    - two
```

PR #5617, opened 2021-06-18
Trim element collection in env property

eutkin:
Issue #5615


jameskleeh:
As I said in the issue I'm not sure we can assume this is the desired behavior. It is easily avoided by simply not putting spaces in the config. This is also not backwards compatible

jeffscottbrown:
I see that this has been moved from `2.5.x` to `3.0.x`.  If this is brought into 3.0, will there still be a way to express leading and trailing whitespace in a config value?

eutkin:
> I see that this has been moved from `2.5.x` to `3.0.x`. If this is brought into 3.0, will there still be a way to express leading and trailing whitespace in a config value?

only:

```yml
property:
  - 'one'
  - ' two'
 ```

jeffscottbrown:
@eutkin I see.  I didn't run it but from looking at the PR it looked like that probably wouldn't work.  Sorry for the noise.

PR #5619, opened 2021-06-18
Include Maven configuration for javaParameters option

lelerabino:
This can save people from (hard)troubleshooting especially because the micronaut-gradle-plugin automatically set the flag to `true` but not the micronaut-maven-plugin.
Working with micronaut-data and JdbcRepository results in compiler error because of this.

### DIFF
--- a/src/main/docs/guide/languageSupport/kotlin/kotlinretainparamnames.adoc
+++ b/src/main/docs/guide/languageSupport/kotlin/kotlinretainparamnames.adoc
@@ -2,7 +2,7 @@ Like with Java, the parameter name data for method parameters is not retained at
 
 To enable retention of parameter name data with Kotlin, set the `javaParameters` option to `true` in your `build.gradle`:
 
-.build.gradle
+.configuration in Gradle
 [source,groovy]
 ----
 compileTestKotlin {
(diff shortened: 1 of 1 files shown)


graemerocher:
For me this is a bug.

This should go into the `pluginManagement` block of the parent POM imo https://github.com/micronaut-projects/micronaut-core/blob/f183e79e2506d41b035de877c3d95548aadc74d7/parent/build.gradle#L87

Seems the config for Groovy is missing as well and is hard coded into launch:

https://micronaut.io/launch/?javaVersion=JDK_11&lang=GROOVY&build=MAVEN&test=JUNIT&name=demo&package=com.example&type=DEFAULT&version=2.5.6&activity=preview&showing=/pom.xml

graemerocher:
Seems in general the Kotlin config could benefit from some parent POM management love for Kotlin as it is all hard coded in launch https://micronaut.io/launch?javaVersion=JDK_11&lang=KOTLIN&build=MAVEN&test=JUNIT&name=demo&package=com.example&type=DEFAULT&version=2.5.6&activity=preview&showing=/pom.xml

lelerabino:
I am quite new to this community and the overall jvm landscape, so it's totally possible that I miss some basic assumption.

Regarding `pluginManagement`: the decision whether to put this plugin configuration (in `pluginManagement` or in the actual plugin configuration) is totally out of scope here imo. The purpose of this documentation is not "teaching Maven" :)

By the way, again, I am quite green in this project, the main purpose of this PR is to push on the bad experience for people moving from gradle plugin to maven plugin working on a micronaut-data project (I'm mentioning here because it rely on keeping parameter names during compilation). Gradle plugin is [automatically setting](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/401e3603173dcdb07452252b3d2969a723d31101/src/main/java/io/micronaut/gradle/MicronautKotlinSupport.java#L176) the `javaParameters` flag, and in any case in this documentation there is a special section about this setting for Gradle.

I just thought it should somehow be documented also for Maven.

graemerocher:
Yeah that is fine. Comment wasn’t directed at the change you made more a general comment that we (the Micronaut team) should fix the root cause

graemerocher:
Thanks for the contribution!

Issue #4791, opened 2021-01-10
Allow custom HTTP response status codes


debuglevel:
### Steps to Reproduce
`HttpResponse.ok().status(599)`

### Expected Behaviour
Some REST services may use custom HTTP status codes to indicate specific states. Let's say `599 Sorry, still processing your request check again in some minutes` or something like that.

Although that's probably quite a bad practise to use custom HTTP status codes, there are legacy services which use them. If somebody wants to migrate such a service to Micronaut, he cannot fulfill the REST API and must change the REST API and their clients first.

I'd totally understand if you do not want to support this, and I'd be fine with that ;-). I just stumbled across such a service and it gave me some headache during migration.

### Actual Behaviour
`HttpResponse.ok().status(599)` throws an Exception, as no such pre-defined HTTP status code is found.

### Environment Information
- **Micronaut Version:** 2.1
- **JDK Version:** 1.8

utsav0209:
@graemerocher [HttpStatus](https://github.com/utsav0209/micronaut-core/blob/b5cc40ea85d020d875f5a95e570188d4a47890a1/http/src/main/java/io/micronaut/http/HttpStatus.java) is an ENUM, how do we allow this then?

graemerocher:
would have to be an internal implementation of for example `NettyHttpResponseFactory`

utsav0209:
@graemerocher, you mean to say that we should create a new `CustomerHttpresponseFactory` for this just the way we have `NettyHttpResponseFactory`?

graemerocher:
Currently there are only methods variants that take the enum:

https://github.com/micronaut-projects/micronaut-core/blob/a237e29cf0baf2628a51c2e20d265bfb1f003b0e/http/src/main/java/io/micronaut/http/HttpResponseFactory.java#L48

We would need to add new methods like:

```java
   /**
     * Return a response for the given status.
     *
     * @param status The status code
     * @param reason An alternatively reason message
     * @param <T>    The response type
     * @return The response
     */
    default <T> MutableHttpResponse<T> status(int statusCode, String reason) {
            return status(HttpStatus.valueOf(statusCode), reason);
    }
```

And then implement these in `NettyHttpResponseFactory` by overriding the method to produce the underlying Netty status object

utsav0209:
@graemerocher  If we add a default method as you suggested above, and override it like: 
```java
    @Override
    public MutableHttpResponse status(int statusCode, String reason) {
[...]

PR #5621, opened 2021-06-18
Custom Http response status codes

hernancerm:
Related issue: #4791

### Changes

Add custom response status support for implementations of `HttpResponse`:

- `NettyMutableHttpResponse`
- `SimpleHttpResponse`

Also for implementations of `HttpResponseFactory`:

- `NettyHttpResponseFactory`
- `SimpleHttpResponseFactory`

The previous changes allow invocations such as the following, creating custom response statuses:

```java
HttpResponse.status(480);                    // Response status: 480 Client Error (480)
HttpResponse.status(480, "My Reason");       // Response status: 480 My Reason
HttpResponse.ok().status(480);               // Response status: 480 Client Error (480)
HttpResponse.ok().status(480, "My Reason");  // Response status: 480 My Reason
```

Implement tests for the changes, in `CustomResponseStatusSpec.groovy`

### Servlet runtimes

My changes work on the Netty runtime, however the response status is not consistent in Tomcat and Jetty. I haven't tested the other servlet runtimes but I'm thinking they wouldn't be consistent as well.

**Tomcat**

|`HttpResponse`|Expected status|Actual status|
|:---|:---|:---|
|`.status(480)`|`480 Client Error (480)`|`480`|
|`.status(480, "My Reason")`|`480 My Reason`|`480`|
|`.ok().status(480)`|`480 Client Error (480)`|`480`|
|`.ok().status(480, "My Reason")`|`480 My Reason`|`480`|

**Jetty**

|`HttpResponse`|Expected status|Actual status|
|:---|:---|:---|
|`.status(480)`|`480 Client Error (480)`|`480 480`|
|`.status(480, "My Reason")`|`480 My Reason`|`480 480`|
|`.ok().status(480)`|`480 Client Error (480)`|`480 480`|
|`.ok().status(480, "My Reason")`|`480 My Reason`|`480 480`|


http/src/main/java/io/micronaut/http/HttpStatus.java
```diff
@@ -262,10 +280,37 @@ public static HttpStatus valueOf(int code) {
...
+     * @return The {@link HttpStatus}
+     */
+    public static HttpStatus custom(int statusCode, String reason) {
+        if (Objects.isNull(reason)) {
+            return valueOf(statusCode);
+        }
+        CUSTOM_STATUS.code = statusCode;
+        CUSTOM_STATUS.reason = reason;
+        return CUSTOM_STATUS;
+    }
```
graemerocher: This approach will not work, it is not thread safe and sharing static state is not viable

hernancerm: From this [StackOverflow post](https://stackoverflow.com/a/9614325/12591546) it seems that it is **not** possible to create a new instance of an enum, which I thought could solve the shared state issue.

Regardless, I believe that an enum is still required for the annotation [`@Status`](https://github.com/micronaut-projects/micronaut-core/blob/2.5.x/http/src/main/java/io/micronaut/http/annotation/Status.java) to work, since the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.1) doesn't allow classes or interfaces as a member type of an annotation.

I have made changes to the PR to meet two criteria:

- Keep the annotation `@Status` functional.
- Make the implementation thread safe.

I changed many files, the big-picture idea is shown in the following diagram:

<p align="center">
Screen Shot 2021-06-28 at 0 31 08
</p>

This change is about refactoring `HttpStatus` from an enum to an interface. The implementation is thread safe beacause custom status codes are now instances of `HttpStatusCustom`, which are value objects.

**Shortcoming**: Since the name of the enum is changed, `@Status` is now used as follows:

```java
@Status(HttpStatusStandard.CREATED)
```

Instead of:

```java
@Status(HttpStatus.CREATED)
```

What are your thoughts on this approach?

mliberato: A possible workaround (without a huge refactor) is to ship the same HttpStatus enum class (same micronaut package) with custom statuses added on your project source directory. If you distribute your application as a "fat jar", your local class will overwrite framework's class.

It's not as elegant as it should be, but it works and you can start using your custom statuses.

Either way, I think the possibility of using custom statuses would be a nice add to the framework.

yawkat:
This PR changes HttpStatus to be an interface, with the enum HttpStatusStandard containing the standard codes. For convenience, `HttpStatus` has fields for the standard codes also, that are initialized similar to: `HttpStatus OK = HttpStatusStandard.OK`.

This circular dependency will cause deadlocks. If one thread loads HttpStatus while another loads HttpStatusStandard, they will wait for each other forever. We can't use this approach.

Additionally, there seem to be some issues in kotlin with using an enum constant when the superinterface has a field of the same name.

This PR could still be "rescued" by going with a interface BaseHttpStatus, enum HttpStatus, class CustomHttpStatus structure instead. BaseHttpStatus would not have any constants for the standard status codes. WDYT @graemerocher ?

graemerocher:
can the constants not exist within the `HttpStatus` interface? Why is `HttpStatusStandard` necessary?

yawkat:
@graemerocher we can only use enum constants inside annotations, so we need some enum (unless we want to switch to strings...)

yawkat:
Sorry it took so long. We decided not to go with this approach due to the problems introduced by the enum split. There seems to be no satisfactory approach that involves changing the HttpStatus api.

I've made a PR with a more limited solution in #8147.

PR #5627, opened 2021-06-21
Fix for custom reason phrase is inaccessible in filters when using Netty runtime

### LINKED ISSUE #5544: Custom reason phrase is inaccessible in filters when using Netty runtime
### Task List

### Steps to Reproduce

1. Create a Micronaut project using Netty runtime.
2. Create an HTTP endpoint and set a custom reason phrase in the response.
3. Create an `HttpServerFilter`, wrap the publisher to access the set reason phrase in the `onNext` method - i.e. copy its value into a custom HTTP header.

### Expected Behaviour

The custom reason phrase is expected to be set in the response header.

### Actual Behaviour

The value of the HTTP response header set in the filter contains the standard reason string defined in the `io.micronaut.http.HttpStatus` enum.

The root cause of the problem is that `io.micronaut.http.netty.NettyMutableHttpResponse` does not override the default implementation of the `reason()` method defined in the `HttpResponse` interface, so it uses `getStatus().getReason()` which always returns the default reason phrase predefined in the `HttpStatus` enum.

To fix this, override this method in `NettyMutableHttpRespnonse` as:

```java
  @Override
  public String reason() {
    return httpResponseStatus.reasonPhrase();
  }
```

### Environment Information

- **Operating System**: Windows 10
- **Micronaut Version:** 2.5.5
- **JDK Version:** OpenJDK 11.0.11

### Example Application

https://github.com/lgathy/micronaut-netty-reason-phrase-example

lgathy:
Closes #5544

Regression test is added to demonstrate the failure first. Fix will follow soon.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/NettyMutableHttpResponse.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/NettyMutableHttpResponse.java
@@ -229,6 +229,16 @@ public class NettyMutableHttpResponse<B> implements MutableHttpResponse<B>, Nett
     }
 
     @Override
+    public int code() {
+        return httpResponseStatus.code();
+    }
+
+    @Override
+    public String reason() {
+        return httpResponseStatus.reasonPhrase();
+    }
+
+    @Override
     public MutableHttpResponse<B> cookie(Cookie cookie) {
         if (cookie instanceof NettyCookie) {
             NettyCookie nettyCookie = (NettyCookie) cookie;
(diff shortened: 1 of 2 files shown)


jameskleeh:
Please remove `@MicronautTest` from the test as its spinning up a context that won't be used

lgathy:
> Please remove `@MicronautTest` from the test as its spinning up a context that won't be used

Right, sry. It was just a leftover from my initial approach that did not work.

jameskleeh (approved):

Issue #5528, opened 2021-06-02
Review inheritance of class annotations


jameskleeh:
Should `@Requires` be inherited?
Should qualifiers be inherited?
Should `io.micronaut.http.annotation` annotations be inherited?

graemerocher:
no and no

I plan to implement support for `@Inherited` to allow inheriting annotations to be controlled by stereotypes. This is a CDI requirement for the work I am doing

Issue #2616, opened 2020-01-09
Allow extending a bean without inherting conditions


wenerme:
### Steps to Reproduce

```java
@Context
public class IssuesTest {
  @Inject private ConsulClient consulClient;

  public static void main(String[] args) throws MalformedURLException {
    ConsulConfiguration configuration = new ConsulConfiguration();
    configuration.setDefaultZone(Arrays.asList(new URL("http://localhost:8200")));
    ApplicationContextBuilder builder =
        Micronaut.build(args)
            .mainClass(IssuesTest.class)
            .deduceEnvironment(true)
            .singletons(configuration);

    // this is required - but should works with just ConsulConfiguration
    builder.propertySources(PropertySource.of("application", ImmutableMap.of("consul.client.enabled", "true")));
    builder.start();
  }
}
```

### Expected Behaviour

Works without properties

### Actual Behaviour

Throw exception, condition failed, consul.client null

### Environment Information

- **Operating System**: macOS
- **Micronaut Version:** 1.2.8
- **JDK Version:** 11

### Why
I use micronaut as middleware, want to initialize micronaut by typesafe conf classes, don't want to inject properties file in env, also map based property is error prone.

graemerocher:
You can still do that. 

```java
@Singleton
@Replaces(ConsulConfiguration.class)
@BootstrapContextCompatible
class MyConsulConfig extends ConsulConfiguration {
      MyConsulConfig() {
           setDefaultZone(Arrays.asList(new URL("http://localhost:8200")));
      }
}
```

wenerme:
Still get error

```java
@Context
public class IssuesTest {
  @Inject private ConsulClient consulClient;

  public static void main(String[] args) throws MalformedURLException {
//    ConsulConfiguration configuration = new ConsulConfiguration();
//    configuration.setDefaultZone(Arrays.asList(new URL("http://localhost:8200")));
    ApplicationContextBuilder builder =
        Micronaut.build(args)
            .mainClass(IssuesTest.class)
            .deduceEnvironment(true)
//            .singletons(configuration)
      ;

    // this is required
//    builder.propertySources(PropertySource.of("application", ImmutableMap.of("consul.client.enabled", "true")));
    builder.start();
  }

  @Singleton
  @Replaces(ConsulConfiguration.class)
  @BootstrapContextCompatible
  class MyConsulConfig extends ConsulConfiguration {
    MyConsulConfig() throws MalformedURLException {
[...]

PR #5643, opened 2021-06-23
Allow controlling annotation inheritance with @Inherited

### LINKED ISSUE #5628: Respect @Inherited in annotation metadata builder & review inheritance of annotations
Currently the annotation metadata builders don't take into account `@Inherited` and inheritance is not consistent across the different annotations Micronaut supports

graemerocher:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/5528 and Fixes https://github.com/micronaut-projects/micronaut-core/issues/2616 and Fixes #5628

This PR alters Micronaut so that an annotation is not inherited into the `AnnotationMetadata` from a parent type or overridden method unless it is annotated with `@Inherited` following the same rules for types as https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/AnnotatedElement.html

Such that annotations directly declared on a type, method or field are available via the `getDeclaredAnnotation*` etc. methods whilst annotations inherited with `@Inherited` are available via `getAnnotation*` methods.

The core additional tests are here:

https://github.com/micronaut-projects/micronaut-core/blob/1337e0c3a32089acf44588fb73c55f352614c7f9/inject-java/src/test/groovy/io/micronaut/inject/annotation/AnnotationInheritanceSpec.groovy

Scopes, qualifiers and requirements are no longer inherited by default.

I have revised the annotations that are part of core and made the following changes:

https://github.com/micronaut-projects/micronaut-core/blob/0072ccca4f1a2d35eeb164ca651013cedcd06312/src/main/docs/guide/appendix/breaks.adoc#annotation-inheritance


inject-java/src/test/groovy/io/micronaut/inject/annotation/BeanDefinitionAnnotationMetadataSpec.groovy
```diff
@@ -282,7 +282,7 @@ interface Foo {
         BeanDefinition definition = classLoader.loadClass('test.$Test$Foo0Definition').newInstance()
         expect:
         definition != null
-        definition.hasStereotype(AnnotationUtil.SINGLETON)
+        definition.hasAnnotation(AnnotationUtil.SINGLETON)
```
jameskleeh, line 285: If singleton isn't inherited the factory definition shouldn't have the annotation, right? The bean should be a prototype in this case

jameskleeh: I think in this case a factory should be considered "extending" the return type, otherwise things like `@Requires` couldn't be changed via a factory

graemerocher: That is the current behavior. I haven't changed anything there. It is imo debatable whether annotations on the type returned should be taken into account at all.

jameskleeh: Right I think this is one of the areas that should be addressed as part of this. If you prefer to resolve it after this PR that is OK with me. Please bring it up at the next meeting. In my opinion I don't think any annotations should be taken into account.

graemerocher: I agree with you :)

PR #5651, opened 2021-06-24
Fix dead link in graal.adoc

FelixSchuSi:
There is a dead link in graal.adoc.
With the wayback machine I found out that the page used to look like [this](http://web.archive.org/web/20180420043824/https://www.graalvm.org/docs/reference-manual/aot-compilation/).
It seems like the graal docs have been refactored, I figured that [this](https://www.graalvm.org/reference-manual/native-image/) is probably the best page to link to now.


sdelamo (approved):
Thanks @FelixSchuSi

PR #5666, opened 2021-06-27
Extract http compile time validation to a new module

jameskleeh:


graemerocher:
I think this is fine, but we need to update Launch as well to add this module by default to the annotation processor path if you create a web app

graemerocher:
I created https://github.com/micronaut-projects/micronaut-starter/issues/844 to track this

PR #5672, opened 2021-06-28
Improve v3 class generating

dstepanov:
Introducing `AbstractBeanDefinition2`, `AbstractBeanDefinitionReference2` with improvements:
- passing most of the parameters in the constructor instead of overriding methods
- initializing method/field injection in the static block allowing Graal to perform build-time optimizations
- generating annotation metadata with shared annotation default variables, reducing significantly duplicate calls mostly for config classes
- improving how annotation metadata are generated
- use try-resource syntax for bean context push/pop
- `AbstractBeanDefinitionReference2` includes more precompilated variables at compile-time
- bean references are generated with prefix `$Reference` instead of the previous `Class` for better understanding what does the class represent
- generate all executable methods in one class prefixed "$Exec" instead of the previous one class per method
- changed to generate bean definition classes with prefix "$Definition" instead of previous "Definition" to better distinguish generated class
- use new map implementation `ImmutableSortedStringsArrayMap` for annotation classes instead of `Map.of`. The implementation is using a sorted `String` keys array for lookup and an extra array for keys. The size of the collection should be the same as in `Map.of`
- Refactored a lot of original `AbstractBeanDefinition`, most of the resolve code is now shared between field/constructor/methods methods
- Moved most of the code that decided how to resolve the argument to the compile-time
- Including `AbstractParametrizedBeanDefinition` main methods inside `AbstractBeanDefinition2` for the benefit of simplicity to have one super-class
- Resolving generic types, known qualifiers, and inner configuration types at the compile-time

### DIFF
--- a/core/src/main/java/io/micronaut/core/util/ArrayUtils.java
+++ b/core/src/main/java/io/micronaut/core/util/ArrayUtils.java
@@ -16,6 +16,8 @@
 package io.micronaut.core.util;
 
 import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.annotation.UsedByGeneratedCode;
+
 import java.lang.reflect.Array;
 import java.util.*;
 import java.util.function.IntFunction;
@@ -31,6 +33,7 @@ public class ArrayUtils {
     /**
      * An empty object array.
      */
+    @UsedByGeneratedCode
     public static final Object[] EMPTY_OBJECT_ARRAY = new Object[0];
 
     /**
@@ -170,6 +173,18 @@ public class ArrayUtils {
     public static <T> T[] toArray(Collection<T> collection, IntFunction<T[]> createArrayFn) {
         T[] array = createArrayFn.apply(collection.size());
         return collection.toArray(array);
+    }
+
+    /**
+     * Returns an array containing all of the elements in this collection, using the item class.
+     *
+     * @param collection The collection
+     * @param arrayItemClass The array item class
+     * @param <T> The type of the array
+     * @return The array
+     */
+    public static <T> T[] toArray(Collection<T> collection, Class<T> arrayItemClass) {
+        return (T[]) collection.toArray((Object[]) Array.newInstance(arrayItemClass, collection.size()));
     }
 
     /**
--- a/core/src/main/java/io/micronaut/core/util/ArrayUtils.java
+++ b/core/src/main/java/io/micronaut/core/util/ArrayUtils.java
@@ -182,6 +182,7 @@ public class ArrayUtils {
      * @param arrayItemClass The array item class
      * @param <T> The type of the array
      * @return The array
+     * @since 3.0
      */
     public static <T> T[] toArray(Collection<T> collection, Class<T> arrayItemClass) {
         return (T[]) collection.toArray((Object[]) Array.newInstance(arrayItemClass, collection.size()));
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
@@ -2,17 +2,14 @@ package io.micronaut.inject.field
 
 import io.micronaut.context.ApplicationContext
 import io.micronaut.context.BeanContext
-import io.micronaut.context.DefaultApplicationContext
-import io.micronaut.context.DefaultBeanContext
 import io.micronaut.context.annotation.Property
 import io.micronaut.context.annotation.Value
-import io.micronaut.context.event.BeanContextEvent
 import io.micronaut.context.exceptions.BeanContextException
+import jakarta.inject.Inject
+import jakarta.inject.Singleton
 import spock.lang.Specification
 
 import javax.annotation.Nullable
-import jakarta.inject.Inject
-import jakarta.inject.Singleton
 
 class FieldInjectionSpec extends Specification {
 
@@ -66,8 +63,8 @@ class FieldInjectionSpec extends Specification {
 
         then:
         e.d == null
-        e.value == "Default greeting"
-        e.property == "Default greeting"
+        e.value == null
+        e.property == null
 
         when:
         context.getBean(C2)
@@ -86,6 +83,21 @@ class FieldInjectionSpec extends Specification {
 
         then:
         thrown(BeanContextException)
+
+        cleanup:
+        context.close()
+    }
+
+    void "test injection with no bean/property found and not nullable and protected fields"() {
+        BeanContext context = ApplicationContext.run()
+
+        when:
+        E e = context.getBean(E)
+
+        then:
+        e.d == null
+        e.value == null
+        e.property == null
 
         cleanup:
         context.close()
@@ -148,6 +160,32 @@ class FieldInjectionSpec extends Specification {
         @Nullable
         @Value('${greeting}')
         private String value = "Default greeting"
+
+        @Nullable
+        @Property(name = 'greeting')
+        private String property = "Default greeting"
+
+        D getD() {
+            return d
+        }
+
+        String getValue() {
+            return value
+        }
+
+        String getProperty() {
+            return property
+        }
+    }
+
+    static class F {
+        @Inject
+        @Nullable
+        protected D d
+
+        @Nullable
+        @Value('${greeting}')
+        protected String value = "Default greeting"
 
         @Nullable
         @Property(name = 'greeting')
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
@@ -92,7 +92,7 @@ class FieldInjectionSpec extends Specification {
         BeanContext context = ApplicationContext.run()
 
         when:
-        E e = context.getBean(E)
+        F e = context.getBean(F)
 
         then:
         e.d == null
--- a/inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
+++ b/inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
@@ -1,0 +1,1846 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.context.annotation.EachBean;
+import io.micronaut.context.annotation.EachProperty;
+import io.micronaut.context.annotation.Parameter;
+import io.micronaut.context.annotation.Property;
+import io.micronaut.context.annotation.Value;
+import io.micronaut.context.env.Environment;
+import io.micronaut.context.event.BeanInitializedEventListener;
+import io.micronaut.context.event.BeanInitializingEvent;
+import io.micronaut.context.exceptions.BeanContextException;
+import io.micronaut.context.exceptions.BeanInstantiationException;
+import io.micronaut.context.exceptions.DependencyInjectionException;
+import io.micronaut.context.exceptions.DisabledBeanException;
+import io.micronaut.context.exceptions.NoSuchBeanException;
+import io.micronaut.core.annotation.AnnotationMetadata;
+import io.micronaut.core.annotation.AnnotationUtil;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.annotation.UsedByGeneratedCode;
+import io.micronaut.core.bind.annotation.Bindable;
+import io.micronaut.core.convert.ArgumentConversionContext;
+import io.micronaut.core.convert.ConversionContext;
+import io.micronaut.core.convert.ConversionService;
+import io.micronaut.core.naming.Named;
+import io.micronaut.core.reflect.ClassUtils;
+import io.micronaut.core.reflect.ReflectionUtils;
+import io.micronaut.core.type.Argument;
+import io.micronaut.core.type.DefaultArgument;
+import io.micronaut.core.util.CollectionUtils;
+import io.micronaut.core.value.PropertyResolver;
+import io.micronaut.inject.BeanDefinition;
+import io.micronaut.inject.ConstructorInjectionPoint;
+import io.micronaut.inject.ExecutableMethod;
+import io.micronaut.inject.ExecutableMethodsDefinition;
+import io.micronaut.inject.FieldInjectionPoint;
+import io.micronaut.inject.MethodInjectionPoint;
+import io.micronaut.inject.ValidatedBeanDefinition;
+import io.micronaut.inject.annotation.AbstractEnvironmentAnnotationMetadata;
+import io.micronaut.inject.qualifiers.InterceptorBindingQualifier;
+import io.micronaut.inject.qualifiers.Qualifiers;
+import io.micronaut.inject.qualifiers.TypeAnnotationQualifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+/**
+ * <p>Default implementation of the {@link BeanDefinition} interface. This class is generally not used directly in user
+ * code.
+ * Instead a build time tool does analysis of source code and dynamically produces subclasses of this class containing
+ * information about the available injection points for a given class.</p>
+ * <p>
+ * <p>For technical reasons the class has to be marked as public, but is regarded as internal and should be used by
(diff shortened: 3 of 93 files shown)


dstepanov:
Not sure what to do with `containsProperties` I have discovered that it's not correctly implemented in the current version:
```
@UsedByGeneratedCode
    protected final boolean containsProperties(@SuppressWarnings("unused") BeanResolutionContext resolutionContext, BeanContext context, String subProperty) {
        boolean isSubProperty = StringUtils.isNotEmpty(subProperty);
// This hack is to skip the `if` if the configuration contains bean injects
        if (!isSubProperty && !requiredComponents.isEmpty()) {
            // if the bean requires dependency injection we disable this optimization
            return true;
        }
        if (isConfigurationProperties && context instanceof ApplicationContext) {
            AnnotationMetadata annotationMetadata = getAnnotationMetadata();
            ApplicationContext appCtx = (ApplicationContext) context;
// This is always true because it return empty array
            if (annotationMetadata.getValue(ConfigurationProperties.class, "cliPrefix").isPresent()) {
                return true;
            } else {
                String path = getConfigurationPropertiesPath(resolutionContext);
                return appCtx.containsProperties(path);
            }

        }
        return false;
    }
```
In the new version I have made `requiredComponents` lazy, so not sure what would be the best to implement this method, for now, it return the value if `isConfigureProperties`
The best solution would be to generate `if (containsProperties` only for the group of property injection but that requires collecting all of the injections and generating the method later.

graemerocher:
Thanks @dstepanov will review on Monday, can you provide some data on what the improvements are like for:

- Memory consumption (GraalVM & JIT)
- Startup (if any?)
- Image size for native

I think as well before we introduce a new metadata format we should address some problems with the old one, I will elaborate more on this in my review

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,2338 @@
...
+            Collection beansOfType = getBeansOfTypeForConstructorArgument(resolutionContext, context, argument);
+            return beansOfType.toArray((Object[]) Array.newInstance(beanType.getComponentType(), beansOfType.size()));
+        } else if (Collection.class.isAssignableFrom(beanType)) {
+            Collection beansOfType = getBeansOfTypeForConstructorArgument(resolutionContext, context, argument);
+            return coerceCollectionToCorrectType(beanType, beansOfType);
+        } else if (Stream.class.isAssignableFrom(beanType)) {
+            return streamOfTypeForConstructorArgument(resolutionContext, context, argument);
+        } else if (argument.isOptional()) {
+            return findBeanForConstructorArgument(resolutionContext, context, argument);
+        } else {
```
graemerocher: Can we precompute this at compilation time and just dispatch to the right method? All of this `isAssingableFrom` checks can be calculated at build time. What I would like to see if we are going to switch metadata format is more fine grained methods like `getBeanArrayForConstructorArgument` that are directly dispatched to

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,2338 @@
...
+    private boolean isInnerOfAnySuperclass(Class argumentType) {
+        Class beanType = getBeanType();
+        while (beanType != null) {
+            if ((beanType.getName() + "$" + argumentType.getSimpleName()).equals(argumentType.getName())) {
+                return true;
+            }
+            beanType = beanType.getSuperclass();
+        }
+        return false;
+    }
```
graemerocher: Can we precompute these at build time? We should know at build time whether an argument is an inner configuration and not require this logic for the `v2` format

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,2338 @@
...
+
+    private Argument resolveGenericType(Argument argument, Class argumentType) {
+        Argument genericType;
+        if (argument.isArray()) {
+            genericType = Argument.of(argumentType.getComponentType());
+        } else {
+            return argument.getFirstTypeVariable().orElse(null);
+        }
+        return genericType;
+    }
```
graemerocher: Can we precompute this build time? There should be no reason to resolve the generic type if we already know it at build time

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,2338 @@
...
+                        qualifier = currentQualifier;
+                    } else {
+                        final Optional<String> n = resolutionContext.get(NAMED_ATTRIBUTE, ConversionContext.STRING);
+                        qualifier = n.map(Qualifiers::byName).orElse(null);
+                    }
+                }
+            }
+            return qualifier;
+        }
+    }
```
graemerocher: I would like to see some of this logic resolved at build time. For the simple case it should not be necessary to have all of this logic and we should be able to compute the qualifier at build time and only fall back for the more complex logic if we know it is iterable (EachBean/EachProperty)

inject/src/main/java/io/micronaut/context/AbstractParametrizedBeanDefinition2.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param requiredArgumentValues The required arguments
+     * @return The built instance
+     */
+    protected abstract T doBuild(BeanResolutionContext resolutionContext, BeanContext context, BeanDefinition<T> definition, Map<String, Object> requiredArgumentValues);
+
+    private Argument[] resolveRequiredArguments() {
+        return Arrays.stream(getConstructor().getArguments())
+            .filter(arg -> {
+                Optional<Class<? extends Annotation>> qualifierType = arg.getAnnotationMetadata().getAnnotationTypeByStereotype(AnnotationUtil.QUALIFIER);
+                return qualifierType.isPresent() && qualifierType.get() == Parameter.class;
```
graemerocher: Should be able to use `getAnnotationNameByStereotype` here instead which will be cheaper

inject/src/main/java/io/micronaut/inject/ExecutableMethodsDefinition.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.stream.Stream;
+
+/**
+ * Collects a set of executable methods {@link ExecutableMethod}.
+ *
+ * @param <T> The bean type
+ * @author Denis Stepanov
+ * @since 3.0
+ */
+public interface ExecutableMethodsDefinition<T> {
```
graemerocher, line 33: Needs `@Nullable` / `@NonNull` on the method signatures also do we want this to be public API? Seems internal and should be `@Internal`?

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -15,36 +15,9 @@
...
-import io.micronaut.core.annotation.AnnotationMetadata;
-import io.micronaut.core.annotation.AnnotationMetadataProvider;
-import io.micronaut.core.annotation.AnnotationUtil;
-import io.micronaut.core.annotation.AnnotationValue;
-import io.micronaut.core.annotation.Internal;
-import io.micronaut.core.annotation.NonNull;
-import io.micronaut.core.annotation.Nullable;
+import io.micronaut.context.*;
+import io.micronaut.context.annotation.*;
+import io.micronaut.core.annotation.*;
```
graemerocher: Can you remove all changes that use `*` imports and revert to individual imports

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,2338 @@
...
+        }
+
+        @Nullable
+        @Override
+        protected Environment getEnvironment() {
+            return environment;
+        }
+    }
+
+    public static final class MethodReference extends MethodOrFieldReference {
```
graemerocher: Needs javadoc, also we already have an interface called `MethodReference` so needs to be named something else

core/src/main/java/io/micronaut/core/util/ImmutableSortedStringsArrayMap.java
```diff
@@ -0,0 +1,540 @@
...
+ * The lookup is implemented using {@link Arrays#binarySearch(Object[], Object)}
+ * which should guarantee performance of O(log N).
+ *
+ * @author Denis Stepanov
+ * @since 3.0
+ */
+@SuppressWarnings("ParameterNumber")
+@Internal
+@UsedByGeneratedCode
+public final class ImmutableSortedStringsArrayMap {
```
graemerocher: Make package private and have a factory method in `AnnotationUtil` create this

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
```diff
@@ -450,19 +530,21 @@ private ClassWriter generateClassBytes() {
         GeneratorAdapter constructor = startConstructor(classWriter);
         DefaultAnnotationMetadata annotationMetadata = (DefaultAnnotationMetadata) this.annotationMetadata;
 
+        Map<String, Integer> defaultsStorage = new HashMap<>(3);
```
graemerocher, line 480: Can you explain what this `defaultsStorage` is used for?

inject/src/main/java/io/micronaut/context/AbstractExecutableMethodsDefinition.java
```diff
@@ -0,0 +1,381 @@
...
+    /**
+     * Checks if the method at index matches name and argument types.
+     *
+     * @param index         The method index
+     * @param name          The method name
+     * @param argumentTypes The method arguments
+     * @return true if matches
+     */
+    @UsedByGeneratedCode
+    protected final boolean methodAtIndexMatches(int index, String name, Class[] argumentTypes) {
```
graemerocher, line 169: Where/how is this used?

graemerocher (changes requested):
In general I think if we are going to introduce a new metadata format this is something that should happen very rarely and we have to make sure we get it right and fix any problems in the previous format. With that in mind I have added some questions and comments on what needs to be improved.

I think that we need to reduce the complexity in `AbstractBeanDefinition` and `AbstractBeanDefinition2` (maybe we need a better name) should aim for that. I have a feeling we should have a more specialized `AbstractConfigPropsBeanDefinition` that moves all the logic for `@EachProperty / @EachBean / @ConfigProp` to a child class and reduces the complexity necessary in `AbstractBeanDefinition` to support configuration properties.

We also need to use the build time knowledge better and dispatch directly to the appropriate method rather than having too much complexity in the generic `getBeanForConstructorArgument` methods etc.

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
dstepanov: It should be possible

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
dstepanov: It's for sharing annotation defaults instead of calling it multiple times in the generated code.

inject/src/main/java/io/micronaut/context/AbstractExecutableMethodsDefinition.java
dstepanov: That's part of the `getMethod` "string switch" like code to check if the method at index matches the requested name + arguments.

dstepanov:
I think I agree with most of the suggestions to limit runtime evaluation. Not sure about `AbstractConfigPropsBeanDefinition` the idea is nice but if you check there is no so much of the code that is needed only for config-prop beans. We would also need a new config-specific class extending `AbstractParametrizedBeanDefinition2`.

One thing regarding `ParametrizedBeanFactory` I have noticed there is a code inside `DefaultBeanContext` that does resolve value for the arguments inside `doCreateBean`. Do you know why does it need to be there instead of in `AbstractParametrizedBeanDefinition`?

dstepanov:
Regarding the native image size, I have observed in the GraalVM dashboard:
- `io.micronaut.*` code size in M2: 5.259 MB vs PR: 3.558 MB
- heap size with PR: -400 kB I see big amount of `HashMap$Node` entries are gone
- native image binary is smaller by 2.115MB

dstepanov:
I have addressed all of the comments, moved a lot of the resolutions to the compile-time.

Regarding the speed improvements on JVM, with the first changes, I have seen around 400ms to 390ms drop in the startup time, I would expect more improvements with the final changes.

I have updated the summary with the changes included.

inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
```diff
@@ -166,5 +178,31 @@ class FieldInjectionSpec extends Specification {
         }
     }
 
+    static class F {
```
jameskleeh, line 181: Is this class used somewhere else? Doesn't seem to be used in this spec

inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
```diff
@@ -66,8 +63,8 @@ class FieldInjectionSpec extends Specification {
 
         then:
         e.d == null
-        e.value == "Default greeting"
-        e.property == "Default greeting"
+        e.value == null
```
jameskleeh, line 66: I think we should maintain this behavior if possible

dstepanov: That’s only Groovy related, Java code is returning nulls.
Do we want to have those inconsistencies?

jameskleeh: No all good then. Should make a note in the breaks.adoc about it though

inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
dstepanov: Good catch, I have corrected the test

inject-groovy/src/test/groovy/io/micronaut/inject/field/FieldInjectionSpec.groovy
dstepanov: I have added the note to the breaking changes

jameskleeh (approved):

jameskleeh:
@dstepanov Have you done a smoke test to ensure classes compiled with the old abstract classes still function correctly with the changes in this PR?

dstepanov:
@jameskleeh I didn't, but I have checked the changes again I don't see any runtime changes that can break it.

graemerocher:
@jameskleeh I will verify that

core/src/main/java/io/micronaut/core/util/ArrayUtils.java
```diff
@@ -172,6 +175,18 @@ public static String toString(String delimiter, @Nullable Object[] array) {
...
 
+    /**
+     * Returns an array containing all of the elements in this collection, using the item class.
+     *
+     * @param collection The collection
+     * @param arrayItemClass The array item class
+     * @param <T> The type of the array
+     * @return The array
+     */
+    public static <T> T[] toArray(Collection<T> collection, Class<T> arrayItemClass) {
```
graemerocher, line 187: Add `@UsedByGeneratedCode`

core/src/main/java/io/micronaut/core/util/ArrayUtils.java
```diff
@@ -172,6 +175,18 @@ public static String toString(String delimiter, @Nullable Object[] array) {
         return collection.toArray(array);
     }
 
+    /**
+     * Returns an array containing all of the elements in this collection, using the item class.
+     *
+     * @param collection The collection
+     * @param arrayItemClass The array item class
+     * @param <T> The type of the array
+     * @return The array
```
graemerocher, line 184: Add `@since`

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,1846 @@
...
+ * compile or runtime</p>
+ *
+ * @param <T> The Bean definition type
+ * @author Graeme Rocher
+ * @author Denis Stepanov
+ * @see io.micronaut.inject.writer.BeanDefinitionWriter
+ * @since 3.0
+ */
+@Internal
+public class AbstractBeanDefinition2<T> extends AbstractBeanContextConditional implements BeanDefinition<T>, EnvironmentConfigurable {
```
graemerocher: Let's name this `AbstractInitializableBeanDefinition` since it is more descriptive of what it does vs the previous one

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,1846 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractBeanDefinition2.class);
+    private static final String NAMED_ATTRIBUTE = Named.class.getName();
+
+    private final Class<T> type;
+    private final AnnotationMetadata annotationMetadata;
+    private final Optional<String> scope;
```
graemerocher: Can we just store a nullable string here? It would be more consistent with the other parameters that are annotated with `@Nullable`

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition2.java
```diff
@@ -0,0 +1,1846 @@
...
+    @Nullable
+    private final FieldReference[] fieldInjection;
+    @Nullable
+    private final ExecutableMethodsDefinition<T> executableMethodsDefinition;
+    @Nullable
+    private final Map<String, Argument<?>[]> typeArgumentsMap;
+    @Nullable
+    private Environment environment;
+    private Set<Class<?>> exposedTypes;
+    private Optional<Argument<?>> containerElement;
```
graemerocher: Same here do we need to store the optoinal? Would it not be sufficient to just make this Nullable?

graemerocher (changes requested):
This is amazing work @dstepanov well done! I tested it mixing with other old byte code and there is no problem as far as I can tell also in JIT mode it reduces RSS memory by between 30-40MB and the JAR size is over 2MB smaller

I have a few small changes to request but then we should merge this. Good job!

core/src/main/java/io/micronaut/core/util/ArrayUtils.java
dstepanov: Actually, I end up using generated `.toArray` instead of this method. I can delete it or I will just add `@since`

graemerocher (approved):

Issue #1885, opened 2019-07-10
Register beans with external annotations (aka package level or mixin annotations)


agentgt:
(I asked on gitter if it was alright to file an issue about this. Here is the comment thread: https://gitter.im/micronautfw/questions?at=5d25f4a417cc7b05caa55674).

Currently to register a bean for creation in Micronaut you have roughly three choices:

 1. Use a `javax.injection` annotation directly on the bean (e.g. `@Singleton`)
 2. Use a `@Factory` and manually wire up the bean
 3. `ApplicationContext#registerSingleton` but that requires the bean already be instantiated.

In our code base that used Spring we leveraged Spring XML to wire beans that have a single constructor with all of its collaborators as arguments aka constructor based injection. Those collaborators are `private final` and thus the bean cannot be created with a default constructor (for the most part). 

In Spring XML (since Spring 2) you can list the beans in whatever order like:

```xml
<bean class="my.package.SomeClass" /> <!-- no need to list args -->
<bean class="my.package.SomeOtherClassMayDependOnSomeClass" />
```

Spring will automatically wire up the beans.

For various reasons we did not use `@Inject` as `@Inject` while not as bad as `@Autowire` still required a dependency. Furthermore putting `@Inject` (and friends) on a bean sort of violates the single responsibility in that the class now knows its going to be used by dependency injection but I understand how many could see that as a weak argument. (Also back in the day annotation registration was slow for Spring so just listing all the beans in XML was faster but that is a moot point with micronaut).

Anyway we are migrating from Spring to Micronaut and now I have the choice of littering all of our classes with `@Singleton` or doing tedious instantiating of the beans in a `@Factory` which requires me to manually fill each constructor.

Thus what I'm proposing is to allow registration of beans with out requiring annotations directly on the bean. This could be a package level annotation (package-info.java) or some configuration class. Another approach is similar to what Jackson does: https://github.com/FasterXML/jackson-docs/wiki/JacksonMixInAnnotations but probably don't need that level of configuration like Jackson does.

I understand that there are pitfalls to this approach in that if the beans come from other pre-compiled projects they will need to have the annotation processors run on them prior and thus would require me to put the list of all beans for each project in some package level annotation but this would only require one file change for each project instead of changing each class (to add `@Singleton`).

graemerocher:
Probably possible, we already have something similar with bean introspections with  `@Introspected(packages="foo.bar")` which will import introspections from the given package

PR #5673, opened 2021-06-28
Bean Import support for JSR 330

graemerocher:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/1885


graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -91,15 +99,36 @@ public int getOrder() {
...
+                originatingElements.add(element);
                 packages.add(element.getPackageName());
                 final String beanName = element.getName();
                 addBean(beanName);
-                resolveClassData(beanName + "[]");
+                if (element.isArray()) {
+                    resolveClassData(beanName + "[]");
+                } else {
+                    resolveClassData(beanName);
+                }
```
graemerocher: @ilopmar this appeared to be a bug? Please confirm

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -160,8 +217,9 @@ public void visitMethod(MethodElement element, VisitorContext context) {
     @Override
     public void visitConstructor(ConstructorElement element, VisitorContext context) {
         if (!isSubclass) {
-            if (element.hasAnnotation(Creator.class)) {
+            if (element.hasAnnotation(Creator.class) && element.isPrivate()) {
```
graemerocher, line 215: @ilopmar again I am not sure why we were creating reflective data is the creator didn't require reflection

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
ilopmar: Someone submitted a PR sometime ago to also include arrays in the reflection config for all the declared classes. Apparently they needed for their use case

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
ilopmar: I don't remember, probably it was necessary sometime ago and maybe it's not now.

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -122,11 +122,7 @@ public void visitClass(ClassElement element, VisitorContext context) {
                 packages.add(element.getPackageName());
                 final String beanName = element.getName();
                 addBean(beanName);
-                if (element.isArray()) {
-                    resolveClassData(beanName + "[]");
-                } else {
-                    resolveClassData(beanName);
-                }
+                resolveClassData(beanName + "[]");
```
ilopmar: @graemerocher shouldn't this be the other way? I mean remove the one with the array and keep the original?

graemerocher:
> shouldn't this be the other way? I mean remove the one with the array and keep the original?

I just reverted to the original behaviour

inject/src/main/java/io/micronaut/context/annotation/Import.java
```diff
@@ -0,0 +1,58 @@
...
+     * A list of package names to import.
+     *
+     * <p>Note that only types with a bean defining annotation will be imported.</p>
+     *
+     * @return The packages to import.
+     */
+    String[] packages() default {};
+
+    /**
+     * The annotations types to include in a search. If set to "*" will include all. Defaults to only included types annotated with scopes or qualifiers.
```
jameskleeh: Perhaps make this more clear that this only works with packages also being set

Issue #5536, opened 2021-06-03
Add more customization for query parameters & Document modifying uri with client argument binders


andriy-dmytruk:
# Description

Unable to bind parameters to query with `AnnotatedClientArgumentRequestBinder ` (Using bind to add headers is described in [docs](https://docs.micronaut.io/latest/guide/#clientParameters))

# Reproduce

**The code is available in a [github repo](https://github.com/andriy-dmytruk/query-param-binder/tree/master/src/main/java/com/example).**

To add a query parameter, I create a custom annotation:
```java
@Documented
@Retention(RUNTIME)
@Target(PARAMETER)
@Bindable
public @interface NameQueryParam {
}
```

Then add the annotation to a `Client`:
```java
@io.micronaut.http.client.annotation.Client("/")
public interface Api {
    @Get("/person")
    void customQuery(@NameQueryParam String name);
}
```
(Making `@Get("/person{?name}")` does not fix it either.

The binder looks the following:
```java
@Singleton
public class NameQueryParamBinder implements AnnotatedClientArgumentRequestBinder<NameQueryParam> {
    @NotNull
    @Override
    public Class<NameQueryParam> getAnnotationType() {
        return NameQueryParam.class;
    }

    @Override
    public void bind(@NotNull ArgumentConversionContext<Object> context,
                     @NonNull ClientRequestUriContext uriContext,
                     @NotNull Object value,
                     @NotNull MutableHttpRequest<?> request
    ) {
        request.getParameters().add("name", value.toString());
    }
}
```

# Result

The request is then lacking the described query parameter and is simply `/person`.

This is likely caused because inside `HttpClientIntroductionAdvice` the uri is set on line [292](https://github.com/micronaut-projects/micronaut-core/blob/3f33453ec8b10a3bbcdd1af71ca4189cb923d87b/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L292). 

It overwrites any changes to query parameters or uri which are made on line [241](https://github.com/micronaut-projects/micronaut-core/blob/3f33453ec8b10a3bbcdd1af71ca4189cb923d87b/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L241).

The expected behaviour is to be able to add queryParameters (likely for both `{?name}` being specified as query and not). This is important if user wants to customize format for collections in query (like `?names=A,B` or `?name=A|B`, etc.).

[...]

PR #5678, opened 2021-06-29
Allow ClientArgumentBinders that can bind the whole request & make QueryValue more flexible

andriy-dmytruk:
## QueryValue(format) parameter

format parameter of [QueryValue](https://github.com/andriy-dmytruk/micronaut-core/blob/a7aa4339c9c2b44ae2bb5def411d39097d1d87c3/http/src/main/java/io/micronaut/http/annotation/QueryValue.java#L60) annotation allows to specify how the parameter is formatted in the request.

A [QueryValueClientArgumentBinder](https://github.com/andriy-dmytruk/micronaut-core/blob/query-parameter-formats/http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentBinder.java) is added to accomodate for the requirements. `ConversionService` is used for conversion. `BeanWrapper` is used to support the `DEEP_OBJECT` format, while other formats cannot encode Objects with properties, only Lists and convertible values. If user needs some other behavior - then can overwrite the implementation with their implementation of `AnnotatedClientArgumentRequestBinder<QueryValue>` or define new annotation.

For the binders `uri.queryParameters` can now be used to change them (additionally to the `uriContext.parameters` that could be used before). This option is more versatile as supports adding unencoded or repeated parameters. (More problem details can be seen in [this issue](https://github.com/micronaut-projects/micronaut-core/issues/5536))

## ClientRequestBinder

Added [AnnotatedClientRequetBinder](https://github.com/andriy-dmytruk/micronaut-core/blob/query-parameter-formats/http-client-core/src/main/java/io/micronaut/http/client/bind/AnnotatedClientRequestBinder.java) that allows to modify the request by applying an annotation to the method.

(An example is given in the documentation)
(It might be worth renaming it - I am not sure)

@graemerocher @jameskleeh


graemerocher:
@andriy-dmytruk could you fix the test failures and check style problems before review. Thanks.

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentBinder.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    @Override
+    public void bind(@NotNull ArgumentConversionContext<Object> context,
+                     @NonNull ClientRequestUriContext uriContext,
+                     @NonNull Object value,
+                     @NonNull MutableHttpRequest<?> request
+    ) {
+        AnnotationValue<QueryValue> annotation = context.getAnnotationMetadata()
+                .getAnnotationValuesByType(QueryValue.class).get(0);
```
graemerocher: Remove this line

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentBinder.java
```diff
@@ -0,0 +1,185 @@
...
+        String unencodedKey = context.getAnnotationMetadata().stringValue(QueryValue.class)
+                    .filter(StringUtils::isNotEmpty)
+                    .orElse(context.getArgument().getName());
+        String key = encodeURIComponent(unencodedKey);
+
+        QueryValue.Format format = annotation.get("format", QueryValue.Format.class)
```
graemerocher: Use `annotationMetadata.enumValue(..)` instead

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -250,6 +241,14 @@ public Object intercept(MethodInvocationContext<Object, Object> context) {
                 }
             };
 
+            // Apply all the method binders
+            List<Class<? extends Annotation>> methodBinderTypes = context.getAnnotationTypesByStereotype(Bindable.class);
+            for (Class<? extends Annotation> binderType: methodBinderTypes) {
```
graemerocher: check if empty before looping

http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
```diff
@@ -178,6 +168,11 @@ protected DefaultHttpClientBinderRegistry(ConversionService<?> conversionService
         }
     }
 
+    @Override
+    public Optional<AnnotatedClientRequestBinder<?>> findBinder(Class<?> annotationType) {
```
graemerocher: Name this something more descriptive since `findBinder` could refer to argument binders or request binders

http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
```diff
@@ -201,6 +196,10 @@ protected DefaultHttpClientBinderRegistry(ConversionService<?> conversionService
         }
     }
 
+    public void addBinder(AnnotatedClientRequestBinder<?> binder) {
```
graemerocher: Some here a more descriptive method name needed

http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
```diff
@@ -63,13 +54,16 @@
...
     /**
      * @param conversionService The conversion service
-     * @param binders           The request argument binders
+     * @param argumentBinders   The request argument binders
+     * @param binders           The method binders
      */
     protected DefaultHttpClientBinderRegistry(ConversionService<?> conversionService,
-                                              List<ClientArgumentRequestBinder<?>> binders) {
+                                              List<ClientArgumentRequestBinder<?>> argumentBinders,
+                                              List<AnnotatedClientRequestBinder<?>> binders) {
```
graemerocher: Think it makes sense to introduce a super interface for both `ClientArgumentRequestBinder` and `AnnotatedClientRequestBinder` maybe called `ClientRequestBinder` and inject just a single list to avoid duplicate lookups

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentBinder.java
```diff
@@ -0,0 +1,185 @@
...
+                i++;
+            }
+        } else if (value != null) {
+            StringBuilder builder = new StringBuilder(key);
+            // noinspection unechecked
+            BeanWrapper wrapper = BeanWrapper.getWrapper(value);
+            // noinspection unchecked
+            Collection<BeanProperty<Object,Object>> properties = wrapper.getBeanProperties();
+            for (BeanProperty<Object, Object> property: properties) {
+                Object item = property.get(value);
```
graemerocher, line 147: Need to check if `property.isWriteOnly()` and skip it if it is otherwise this will throw an exception

graemerocher (changes requested):

jameskleeh:
Please wait to merge until I can review next week

andriy-dmytruk:
I noticed there is also some formatting inside the `UriTemplate.expand()`, for example `'{?param*}'` would be the multi format for an `Iterable` and would expand a `Map`. The query parameters still can be formatted with this functionality if user ommitts the `@QueryValue` annotation completely. Or should I add another way to achieve the fallback behaviour?

http-client/src/test/groovy/io/micronaut/http/client/HttpGetSpec.groovy
```diff
@@ -860,17 +860,19 @@ class HttpGetSpec extends Specification {
...
-        String multipleExplodedQueryParams(@QueryValue("bar") List<String> foo, @QueryValue("tag") String label)
+        @Get("/multipleExplodedQueryParams{?bar,tag}")
+        String multipleExplodedQueryParams(@QueryValue(value="bar", format=QueryValue.Format.MULTI) List<String> foo,
+                                           @QueryValue(value="tag", format=QueryValue.Format.MULTI) String label)
 
-        @Get("/multipleExplodedQueryParams{?bar*,tag}")
-        String multipleExplodedQueryParams2(@QueryValue List<String> bar, @QueryValue String tag)
+        @Get("/multipleExplodedQueryParams{?bar,tag}")
+        String multipleExplodedQueryParams2(@QueryValue(format=QueryValue.Format.MULTI) List<String> bar,
+                                            @QueryValue(format=QueryValue.Format.MULTI) String tag)
```
jameskleeh: Don't change existing tests. Add new ones instead

http-client/src/test/groovy/io/micronaut/http/client/aop/ParameterSpec.groovy
```diff
@@ -64,23 +65,23 @@ class ParameterSpec extends Specification {
...
 
-        @Get('/user/{X-username}')
-        User get(@QueryValue('X-username') String username)
+        @Get('/user{?X-username}')
+        User get(@QueryValue('X-username') @Nullable String username)
 
-        @Get('/user/age/{userAge}')
-        User findByAge(@QueryValue('userAge') Integer age)
+        @Get('/user/age{?userAge}')
+        User findByAge(@QueryValue('userAge') @Nullable Integer age)
```
jameskleeh, line 84: Same as above, don't change existing tests

graemerocher:
> Please wait to merge until I can review next week

No worries no rush to merge this but it would be nice if it can make it into 3.0

http-client/src/test/groovy/io/micronaut/http/client/HttpGetSpec.groovy
jameskleeh: Unless of course the old behavior should not be supported, but I don't think that is the case here

jameskleeh:
This needs new tests to cover the other queryvalue format cases

http-client/src/test/groovy/io/micronaut/http/client/HttpGetSpec.groovy
andriy-dmytruk: I changed these tests because in the `{?bar*}` the `*` means the same as to specify `@QueryValue(format=QueryValue.Format.MULTI)`. Without `*` it would be `COMMA_DELIMITED` format.

For the way I implemented it, one can either ommit `@QueryValue` completely - and the parameter will be found by name and then added to query with the `UriTemplate.expand` which will parse the path template and format everything according to it.  Or the format needs to be specified inside the annotation `@QueryValue(format=...)` and then path is not read.

Now that I think of it, maybe it would be better to store the formats for query variables somewhere in the `uriContext` object, and then change the `UriTemplate.expand` to support it, as it would add the format feature everywhere instead of just for Declerative Client. But that sounds hard and I am not sure if needed.

andriy-dmytruk:
> This needs new tests to cover the other queryvalue format cases

Yes, there are some tests in [ClientBindSpec](https://github.com/andriy-dmytruk/micronaut-core/blob/8211fca16fd89ae4a92e16745d5bac39889f73ad/http-client/src/test/groovy/io/micronaut/http/client/bind/ClientBindSpec.groovy#L32-L77), but maybe some more are needed.

http-client/src/test/groovy/io/micronaut/http/client/HttpGetSpec.groovy
jameskleeh: The existing tests need to stay as is to assert the previous behavior is maintained

http-client-core/src/main/java/io/micronaut/http/client/bind/ClientRequestBinder.java
```diff
@@ -8,7 +8,7 @@
 /**
  * A binder that binds to a {@link MutableHttpRequest}.
  * This class is extended by {@link ClientArgumentRequestBinder} and {@link AnnotatedClientRequestBinder}
- * to account for different binding behaviors
+ * to account for different binding behaviours
```
jameskleeh, line 25: Don't actively switch from American to British english

http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
```diff
@@ -29,24 +29,15 @@
...
 import io.micronaut.http.BasicAuth;
 import io.micronaut.http.HttpHeaders;
-import io.micronaut.http.annotation.Body;
-import io.micronaut.http.annotation.CookieValue;
-import io.micronaut.http.annotation.Header;
-import io.micronaut.http.annotation.PathVariable;
-import io.micronaut.http.annotation.QueryValue;
-import io.micronaut.http.annotation.RequestAttribute;
-import io.micronaut.http.annotation.RequestBean;
+import io.micronaut.http.annotation.*;
```
graemerocher: Don't replace with `*` imports, use individual imports

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentBinder.java
```diff
@@ -0,0 +1,183 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Optional;
+
+/**
+ * Implementation of the binder for the {@link QueryValue} annotation
+ *
+ * @author Andriy Dmytruk
```
graemerocher, line 41: Needs a `@since 3.0` declaration

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -38,26 +38,9 @@
...
-import io.micronaut.http.annotation.RequestAttribute;
-import io.micronaut.http.client.BlockingHttpClient;
-import io.micronaut.http.client.HttpClient;
-import io.micronaut.http.client.ReactiveClientResultTransformer;
-import io.micronaut.http.client.RxHttpClient;
-import io.micronaut.http.client.RxHttpClientRegistry;
-import io.micronaut.http.client.StreamingHttpClient;
+import io.micronaut.http.*;
+import io.micronaut.http.annotation.*;
+import io.micronaut.http.client.*;
```
graemerocher: Again doesn't replace star imports

graemerocher (changes requested):

graemerocher:
@jameskleeh this has been cleaned up and feedback addressed by @andriy-dmytruk can you do final review and approve/merge?

jameskleeh:
I don't think we should support adding query parameters via the URI of the request as well as the uri context. The context query paramters should be updated to support a List as the value instead of a single string

andriy-dmytruk:
> I don't think we should support adding query parameters via the URI of the request as well as the uri context. The context query paramters should be updated to support a List as the value instead of a single string

There is still one minor thing that is left if I do this:
The query keys and values are component-encoded, which means that all `','` and other symbols which are allowed in query but might not be allowed in other url components. Then the comma-separated format would be actually separated by `'%2C'` and not `','` and would mean that user cannot use commas inside the iterable values causing ambiguity.

@jameskleeh I am not sure if this is important or if there is a way to solve this. Can you comment on this please?

jameskleeh:
@andriy-dmytruk Are you certain the `,` getting encoded is a problem? It should be decoded on the other side

andriy-dmytruk:
@jameskleeh I am not certain.
But URL format supports using both `','` and `'%2C'`, so if `','` is used for delimiting and commas in user Strings become `'%2C'` we can perfectly retreive the List even if there were values containing commas like `["hello", "hi, Joe"]`. Again, I am unsure if this is needed.
Probably I can remove this, as the server side doesn't seem to account for this.

jameskleeh:
@andriy-dmytruk If a user supplies multiple items, it should result in x=item1&x=item2. If they supply a single value with a comma we should assume it's a single value and thus encoding makes sense to me. We cannot attempt to ascertain if the user intended to supply multiple values or not with the comma, as the server is responsible for determining if the value is intended to be a list or not.

andriy-dmytruk:
@jameskleeh  Yes, but user can supply `@QueryValue(format=QueryValue.Format.CSV) List<String> a` with `a = ["item1", "item,withcomma"]`, which would be incorrectly encoded as something like "a=item1%2Citem%2cwithcomma". It would be correctly encoded with `Format.MULTI` as "a=item1&a=item%2cwithcomma".
How the `QueryValue` adds the values for all the formats is on my side. Of course, we can ban user from using `CSV` format and commas in the list.
Anyway, I think I can do whatever you and graeme think is correct)

jameskleeh:
I don't think it makes sense for a user to supply a comma in a value and also say the format is CSV, regardless of the encoding.

`@QueryValue(format=QueryValue.Format.CSV) List<String> a` with `["item1", "item,withcomma"]` should result in `a=item1,item,withcomma`, not multiple values. I thought we were discussing the API of the URI context supporting a list.

andriy-dmytruk:
Ok, I understand. I will rewrite it to `List<String>` in `context.queryParameters` and stop using `request.uri`.
Thanks.

andriy-dmytruk:
@jameskleeh Can you review one more time then, please?

jameskleeh:
@andriy-dmytruk Can you split out the changes to the uri context and the method annotation binders to another PR? I think the query value format stuff needs a lot more consideration given that it doesn't seem to be applicable to the server as well as the potential combinations of data types with formats not taken into account.

graemerocher:
@jameskleeh Andriy is working on a separate PR to address the server side of the changes to query value. I think we have made him go back and forth and revise the code several times and asking this to be revised again after what is seemingly the 10th cycle of review is a bit much don't you think?

jameskleeh:
@graemerocher No I don't think. If there are changes to the annotation they should be addressed in a single PR. I'm not asking for a revision, but the code that is not applicable to query format value to be split into a separate PR

graemerocher:
Can I asked why this was not requested before in the initial review instead of the endless back and forth?

jameskleeh:
The code in question was just added after the last review

graemerocher:
Only the URI context stuff was added recently

jameskleeh:
And that is what I'm asking to be split out. It also makes sense to include the other code not related to query format value to be split out as well, imo.

graemerocher:
Ok but you also asked Andriy to revise this PR with the changes in the previous review without mentioning the need for a separate PR for the client context changes . In the name of efficiency if we can try and be more clear on requirements early on in the review process so time is not wasted needlessly

jameskleeh:
The best thing to avoid wasting time would have been to propose such a feature `(QueryValue(format=))` prior to submitting a PR. The original issue this sprung out of was about changing the URI context to support a list. In general having features of an annotation that only apply to certain data types creates confusion about their intent, so I'm not particularly in favor of that specific change. That being said I'm not against it if it's implemented in a comprehensive way, which is not here in this PR due to the previous things I've mentioned. I'd like to get the URI context change into Micronaut 3 and I think this will likely have to wait until 3.1.

graemerocher:
That is great I am just trying to understand how we can improve communication such that this PR can comprehensively cover all your concerns.

If that means we need more formal process like a JEP before doing work then I think that should happen but then it needs to happen for all new features. Can you document somewhere in the contributing docs what the process should be to avoid time wasting?

The process here of being drip fed requirements for acceptance that are only in your head is clearly not working. 

FWIW Andriy is working on support for open api generator and identified missing functionality that makes it not possible to fully support without these changes  https://openapi-generator.tech/docs/generators/java

PR #5713, opened 2021-07-07
Discovery of management endpoints

### LINKED ISSUE #5006: Discovery view for management endpoints
Endpoints are described here https://docs.micronaut.io/2.3.3/guide/index.html#providedEndpoints. Boot has such a view at /actuator, enabled via _EndpointLinksResolver_ and _WebMvcEndpointHandlerMapping_. A similar view would be very useful in Micronaut.
In the meantime - until this does or does not become available - is there a way to discover all enabled @Endpoint-annotated classes, so I could build such a view myself?

### Task List

### Steps to Reproduce

1. Run any sample application with _endpoints.all.enabled_ set to true.

### Expected Behaviour

A management discovery endpoint, say /management, listing all available endpoints, together with their links, is available.

### Actual Behaviour

This is not available.

### Environment Information

- **Operating System**: TODO
- **Micronaut Version:** TODO
- **JDK Version:** TODO

### Example Application

- TODO: link to github repository with example that reproduces the issue

hernancerm:
Solves #5006

Created the endpoint `/management` to reveal mangament routes in a HATEOAS format. This is the format that Spring Boot's `/actuator` uses. Example output:

```json
{
  "_links": {
    "self": {
      "href": "http://localhost:8080/management",
      "templated": false
    },
    "threaddump": {
      "href": "http://localhost:8080/threaddump",
      "templated": false
    },
    "loggers": [
      {
        "href": "http://localhost:8080/loggers/{name}",
        "templated": true
      },
      {
        "href": "http://localhost:8080/loggers",
        "templated": false
      }
    ]
}
```

Considerations:

- Duplicate route paths are not shown because the output does not provide HTTP method information to differentiate them. For example, if a route accepts GET and HEAD, a single path is shown (not one per method).
- Regardless of the HTTP method of the route, it is shown if it is a management endpoint. This means that to use some links it is not enough to click on them. I allowed this to keep the design similar to Spring Boot Actuator, since at `/actuator` it can show `/actuator/shutdown` which expects a POST, not a GET.
- The implementation also shows custom endpoints defined by the user.

### Work that I think is pending

- [x] Implement tests for `ManagementController`


management/src/main/java/io/micronaut/management/endpoint/management/ManagementEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+@Endpoint(ManagementEndpoint.ENDPOINT_ID)
+public class ManagementEndpoint {
+
+    static final String ENDPOINT_ID = "management";
+
+    private final ManagementRoutesResolver managementRoutesResolver;
+    private final ManagementLinkCollector managementLinkCollector;
+
+    public ManagementEndpoint(ManagementRoutesResolver managementRoutesResolver,
+                              ManagementLinkCollector managementLinkCollector) {
```
jameskleeh: It shouldn't be assumed the data being returned is links. Users may want to replace the implementation with something that does something entirely different. You could rename this to `ManagementDataCollector` to be generic

management/src/main/java/io/micronaut/management/endpoint/management/ManagementLinkCollector.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.stream.Stream;
+
+/**
+ * <p>Collect routes in a {@link Resource} to respond in {@link ManagementEndpoint}.</p>
+ *
+ * @author Hernán Cervera
+ * @since 2.5
+ * */
+public interface ManagementLinkCollector {
+    Resource collectLinks(Stream<UriRoute> routes, String selfEndpointId);
```
jameskleeh: Similar to above, the interface shouldn't assume `Resource` or `links`. This should also return a `Publisher`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementLinkCollector.java
```diff
@@ -0,0 +1,85 @@
...
+     * Http methods are not shown on a resource, which makes identical route paths seemingly the same. Each entry
+     * has the id of the Endpoint as key, except for the management route, which its key is <code>self</code>.</p>
+     *
+     * @param routes management routes to collect.
+     * @param selfEndpointId id of the {@link Endpoint} which returns the {@link Resource}.
+     * @return the {@link Resource} with the links to the management endpoints.
+     */
+    @Override
+    public Resource collectLinks(Stream<UriRoute> routes, String selfEndpointId) {
+        String routeBase = String.format("%s://%s:%d",
```
jameskleeh: Getting the route base from the embeddedServer is incorrect since it wouldn't reflect the actual host name. Given endpoints can be executed outside the context of a request I don't think you can include the base URL in the link

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementLinkCollector.java
```diff
@@ -0,0 +1,85 @@
...
+
+        AbstractResource<ManagementResponse> resource = new ManagementResponse();
+
+        Set<String> collectedPaths = new HashSet<>();
+        routes.forEach(route -> {
+            String path = route.getUriMatchTemplate().toString();
+
+            // Since Http methods are not shown on a Resource, avoid identical
+            // URLs, since they only differ on their Http method.
+            if (collectedPaths.contains(route.getUriMatchTemplate().toString())) {
```
jameskleeh: Reduce the number of calls to `route.getUriMatchTemplate()`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementLinkCollector.java
```diff
@@ -0,0 +1,85 @@
...
+
+            collectedPaths.add(path);
+        });
+
+        return resource;
+    }
+
+    private String getEndpointId(UriRoute route) {
+        Class<?> declaringType = ((MethodBasedRoute) route).getTargetMethod().getDeclaringType();
+        Endpoint endpoint = declaringType.getAnnotation(Endpoint.class);
```
jameskleeh: The annotation metadata is available on the `MethodBasedRoute` that should be used to determine if the annotation exists. `annotationMetadata.hasAnnotation(..)`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementRoutesResolver.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    @Override
+    public Stream<UriRoute> getRoutes() {
+        return router.uriRoutes().filter(this::isManagementRoute);
+    }
+
+    private boolean isManagementRoute(UriRoute route) {
+        if (route instanceof MethodBasedRoute) {
+            Class<?> declaringType = ((MethodBasedRoute) route).getTargetMethod().getDeclaringType();
+            Endpoint endpoint = declaringType.getAnnotation(Endpoint.class);
```
jameskleeh: Same as above, use the annotation metadata on the route

management/src/main/java/io/micronaut/management/endpoint/management/impl/ManagementResponse.java
```diff
@@ -0,0 +1,13 @@
...
+import io.micronaut.http.hateoas.AbstractResource;
+import io.micronaut.management.endpoint.management.ManagementEndpoint;
+
+/**
+ * <p>Allow {@link ManagementEndpoint} to provide a response in HATEOAS architecture</p>
+ *
+ * @author Hernán Cervera
+ * @since 2.5
+ */
+public class ManagementResponse extends AbstractResource<ManagementResponse> {
```
jameskleeh: Class name is a bit generic. Change to ManagementEndpointResponse

jameskleeh (changes requested):
This should target 3.0.x. Change @since to 3.0.0

jameskleeh:
Thinking about this more it seems like it would be better done with a controller if the goal is to discover the HTTP links

hernancerm:
> Thinking about this more it seems like it would be better done with a controller if the goal is to discover the HTTP links

@jameskleeh By using a controller, would an `EmbeddedServer` injected in the controller return the correct scheme, host and port to create the links? I think that having clickable links as the author of the issue mentions would be helpful for navigation

jameskleeh:
> > Thinking about this more it seems like it would be better done with a controller if the goal is to discover the HTTP links
> 
> @jameskleeh By using a controller, would an `EmbeddedServer` injected in the controller return the correct scheme, host and port to create the links? I think that having clickable links as the author of the issue mentions would be helpful for navigation

No you would have to inject the HttpHostResolver

management/src/main/java/io/micronaut/management/endpoint/management/ManagementEndpoint.java
hernancerm: I've renamed `ManagementLinkCollector` to `ManagementDataCollector`

management/src/main/java/io/micronaut/management/endpoint/management/ManagementLinkCollector.java
hernancerm: Now it returns a `Publisher`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementLinkCollector.java
hernancerm: I followed the approach you mentioned of using `@Controller` instead of `@Endpoint`. Now `ManagementController` uses `HttpHostResolver` to get the base route

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementLinkCollector.java
hernancerm: I've refactored to code to use what you mention

management/src/main/java/io/micronaut/management/endpoint/management/impl/ManagementResponse.java
hernancerm: I changed it to `ManagementControllerResponse`, since the class is now a controller, not an endpoint

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementRoutesResolver.java
hernancerm: I also did the refactoring here to use your suggestion

hernancerm:
@jameskleeh I've addressed your suggestions, now I'm also targeting branch 3.0.x

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementLinkCollector.java
hernancerm: Done

management/build.gradle
```diff
@@ -4,6 +4,7 @@ dependencies {
 
     api project(":router")
     api project(":runtime")
+    api project(":http-server")
```
jameskleeh: This should not be an api dependency. You can have a compileOnly dependency on the `http` project and ensure any http specific beans require the presence of an embeddedserver bean

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementDataCollector.java
```diff
@@ -0,0 +1,79 @@
...
+            resource.link(endpointId, Link.build(routeBase + path).templated(isTemplated).build());
+
+            collectedPaths.add(path);
+        });
+
+        return Flowable.fromArray(new Resource[]{resource}).firstOrError().toFlowable();
+    }
+
+    private String getEndpointId(UriRoute route) {
+        AnnotationValue<Endpoint> endpoint = ((MethodBasedRoute) route).getTargetMethod()
```
jameskleeh: You shouldn't cast here and instead check for instance of

management/build.gradle
hernancerm: Shouldn't the compileOnly dependency be on `http-server` (rather than `http`)?

The controller uses `HttpHostResolver` which is defined in `http-server`, and I don't see that `http` has a direct (or transitive) dependency on this submodule. Depending on `http` makes the build fail, I think because it does not contain such interface

jameskleeh: Yep you're right

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementDataCollector.java
hernancerm: Now I do the instanceof check before doing the casting, and throw an exception if the check fails

management/build.gradle
hernancerm: Great, thanks for the response, now it depends as compileOnly on `http-server`

hernancerm:
@jameskleeh I've implemented the tests for the management controller.

Locally `./gradlew check` passes all tests

hernancerm:
Hi @jameskleeh, are there additional changes that I should make?

jameskleeh:
@HerCerM Overall it looks OK however we are hard at work getting the next milestone of 3.0 out so we'll get this merged afterwards

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementRoutesResolver.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public Stream<UriRoute> getRoutes() {
+        return router.uriRoutes().filter(this::isManagementRoute);
+    }
+
+    private boolean isManagementRoute(UriRoute route) {
+        if (route instanceof MethodBasedRoute) {
+            AnnotationValue<Endpoint> endpoint = ((MethodBasedRoute) route).getTargetMethod()
+                    .getAnnotationMetadata().getAnnotation(Endpoint.class);
+            return Objects.nonNull(endpoint);
```
graemerocher: Can be replaced with:

```java
        return ((MethodBasedRoute) route).getTargetMethod()
                    .getAnnotationMetadata().hasAnnotation(Endpoint.class);
```

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementDataCollector.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            CharSequence endpointId = getEndpointId(route);
+            boolean isTemplated = !uriMatchTemplate.getVariables().isEmpty();
+            resource.link(endpointId, Link.build(routeBase + path).templated(isTemplated).build());
+
+            collectedPaths.add(path);
+        });
+
+        return Flowable.fromArray(new Resource[]{resource}).firstOrError().toFlowable();
```
graemerocher: Replace with `Mono.just(resource)`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementDataCollector.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private String getEndpointId(UriRoute route) {
+        if (route instanceof MethodBasedRoute) {
+            AnnotationValue<Endpoint> endpoint = ((MethodBasedRoute) route).getTargetMethod()
+                    .getAnnotationMetadata().getAnnotation(Endpoint.class);
```
graemerocher: replace with 

```java
return ((MethodBasedRoute) route).getTargetMethod()
                    .getAnnotationMetadata().stringValue(Endpoint.class).orElse(null);
```

You don't need to handle both `value` and `id` since they are aliased, but you do need to either error or ignore the endpoint if a `null` id is specified

management/src/main/java/io/micronaut/management/endpoint/management/impl/ManagementControllerResponse.java
```diff
@@ -0,0 +1,28 @@
...
+import io.micronaut.http.hateoas.AbstractResource;
+import io.micronaut.management.endpoint.management.ManagementController;
+
+/**
+ * <p>Allow {@link ManagementController} to provide a response in HATEOAS architecture.</p>
+ *
+ * @author Hernán Cervera
+ * @since 3.0.0
+ */
+public class ManagementControllerResponse extends AbstractResource<ManagementControllerResponse> {
```
graemerocher: Rename to `AvailableEndpoints` and annotate with `@Introspected`

management/src/main/java/io/micronaut/management/endpoint/management/ManagementController.java
```diff
@@ -0,0 +1,61 @@
...
+        this.httpHostResolver = httpHostResolver;
+        this.managementRoutesResolver = managementRoutesResolver;
+        this.managementDataCollector = managementDataCollector;
+    }
+
+    @Get("management")
+    public Single<?> getManagementRoutes(HttpRequest<?> httpRequest) {
+        String routeBase = httpHostResolver.resolve(httpRequest);
+        String managementDiscoveryPath = httpRequest.getPath();
+        Stream<UriRoute> uriRoutes = managementRoutesResolver.getRoutes();
```
graemerocher, line 57: * Replace `Single` with `Mono` and don't use RxJava 2 but instead use Reactor
* I believe this should also be an `@Endpoint` to take advantage of the security / port binding features
* I don't think the current path or route base needs resolving, this can be discovered from https://github.com/micronaut-projects/micronaut-core/blob/f2e1d122efc4ee7bb2e7249c20dc1c4f2ebccdf1/management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java and it needs to factor in the path and the port

management/src/main/java/io/micronaut/management/endpoint/management/ManagementDataCollector.java
```diff
@@ -0,0 +1,47 @@
...
+     * Collect management data.
+     *
+     * @param routes management routes.
+     * @param routeBase base route of the endpoints.
+     * @param managementDiscoveryPath path of the management endpoint exposed by {@link ManagementController}.
+     * @param isManagementDiscoveryPathTemplated whether the discovery endpoint path is templated.
+     * @return the {@link Publisher}.
+     */
+    Publisher<T> collectData(Stream<UriRoute> routes, String routeBase,
+                             String managementDiscoveryPath, boolean isManagementDiscoveryPathTemplated);
```
graemerocher: It is probably unnecessary for this to return a `Publisher`

graemerocher (changes requested):
Please rebase and target the 3.1.x branch

hernancerm:
Hi @graemerocher, thanks for the review and sorry for the delay. I expect to address the changes during the next week, since currently I have other things at hand

management/src/main/java/io/micronaut/management/endpoint/management/ManagementController.java
hernancerm: This is a `@Controller` because in [this comment](https://github.com/micronaut-projects/micronaut-core/pull/5713#issuecomment-875735950) James suggested to me that a `@Controller` would be more appropriate to discover HTTP links (since [the author of the issue was hoping for clickable URLs](https://github.com/micronaut-projects/micronaut-core/issues/5006#issuecomment-786981171)); something of the following format:

```text
http://localhost:8080/health
```

Building such a URL requires knowing: protocol, host and port.

I have learned that the proper way to discover this data is through the `HttpHostResolver`, but it doesn't seem it can be used in an `@Endpoint` since there is no access to the request.

I think that making this an `@Endpoint` and using `EndpointDefaultConfiguration` as you suggest makes it possible to create links of the following format, wich are not clickable:

```text
:9090/health    # When the port is configured through "endpoints.all.port".
/health         # When the port is the same as the application's port.
```

Would you approve the previous link format, or should it be clickable (contain protocol, host, port)? If you think the latter is required, I would appreciate some pointers into how to discover the app's protocol and port using an `@Endpoint`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementRoutesResolver.java
hernancerm: I've done the replacement

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementDataCollector.java
hernancerm: I've done the code replacement. I ignore endpoints which id is null

management/src/main/java/io/micronaut/management/endpoint/management/impl/ManagementControllerResponse.java
hernancerm: I've done the changes

management/src/main/java/io/micronaut/management/endpoint/management/ManagementDataCollector.java
hernancerm: Now I just return the generic type `T`

management/src/main/java/io/micronaut/management/endpoint/management/impl/DefaultManagementDataCollector.java
hernancerm: Since now I just return the generic type `T` from this method, as my response to [this comment](https://github.com/micronaut-projects/micronaut-core/pull/5713#discussion_r713700094), I've replaced this line with: `return resource;`

management/src/main/java/io/micronaut/management/endpoint/management/ManagementController.java
graemerocher: Yeah I am unconvinced this should be `@Controller` since I would expect all of the configuration under `EndpointDefaultConfiguration` to apply to this endpoint as well (the port it is exposed under, the sensitivity, the path exposed under) and as written none of this is respected. Maybe @jameskleeh can elaborate more what he had in mind with making it a `@Controller`

jameskleeh: @graemerocher I suggested it be a controller because the notion of a URI is outside the realm of `@Endpoint`. In the context of JMX there is no URI so having an endpoint dedicated to producing URIs doesn't make sense to me. As @HerCerM said above, the need to use HTTP based services (HttpHostResolver) require a dependency on HTTP and thus should not be an endpoint.

graemerocher: @jameskleeh Right, but my question remains, how do you get this new controller to respect the port management is exposed under, the sensitivity and the path management is exposed under.

Right now this would expose this information under `/management` with seemingly no way to configure that and it doesn't respect `endpoints.all.port` so you can't send it to another port either.

As it written it is a breaking change if anyone has an existing URI mapped to `/management`, plus many users run the management endpoints on a different port to the actual application for security reasons.

jameskleeh: @graemerocher I don't think the requirement exists for it to respect that configuration because it's not an endpoint. I'm in support for the URI to be configurable and be able to be turned off.

I don't consider a new feature that maps to a given endpoint to be a breaking change, otherwise we wouldn't be able to add many features. For example we couldn't add a new exception handler because one may already exist for that exception. The security module couldn't add any endpoints because it may already be mapped to another controller (this example). We wouldn't be able to add new type converters, etc... the list goes on.

graemerocher: I don't think it makes any sense whatsoever to have this behave differently and is a potential security vulnerability if you expose the port where the management endpoints do exist over the default port.

Spring Boot treats this as part of the endpoints and respects the path / port configuration (the equivalent of `endpoints.all.port` and `endpoints.all.path` https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.hypermedia

I don't also buy the argument that in the context of JMX there is no URI because in the context of JMX `endpoints.all.path` is not relevant either.

jameskleeh: > if you expose the port where the management endpoints do exist over the default port

The management endpoints are exposed over the default port afaik. I think this discovery api should create links with the configured endpoint port though.

I don't think it's relevant what spring boot does.

The management module supports the notion of a configurable base URI because we expose the endpoints over the web by default. So correct its not relevant to JMX and this new "endpoint" isn't either. I'm not sure the point you're trying to make here.

jameskleeh: Spring has the notion of `@WebEndpoint` and endpoints specific to the web. I don't understand the point of doing so given the main point of endpoints is to not be web specific. Perhaps the thing to do would be to understand why they decided to do that and then determine if something similar makes sense here.

graemerocher: > The management endpoints are exposed over the default port afaik.

They are not if you specify `endpoints.all.port` to expose them over a different port and you can customize the path with `endpoints.all.path`

As a very least if `endpoints.all.path` is specified the controller should map to that and if `endpoints.all.port` is specified then `Controller(port)` for this controller should map to that

Issue #5608, opened 2021-06-17
Support for jakarta.annotation for PreDestroy / PostConstruct


graemerocher:
IMO supporting to these and moving to them as default is as important as moving to `jakarta.inject`

PR #5714, opened 2021-07-07
Switch from javax.annotation to jakarta.annotation

graemerocher:
Same as the switch for `jakarta.inject` this is for `jakarta.annotation`


graemerocher:
Fixes  #5608

jameskleeh:
Are there any tests left which use the javax annotations? If not, worth adding some to test-suite-javax-inject

PR #5715, opened 2021-07-07
Required components improvements

jameskleeh:
Add Argument.isProvider and change required components to extract the type arguments from collections and providers

### DIFF
--- a/inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
+++ b/inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
@@ -85,7 +85,6 @@ public class AbstractBeanDefinition<T> extends AbstractBeanContextConditional im
     private final Class<?> declaringType;
     private final ConstructorInjectionPoint<T> constructor;
     private final Collection<Class<?>> requiredComponents = new HashSet<>(3);
-    private final Collection<Argument<?>> requiredComponentArguments = new HashSet<>(3);
     private AnnotationMetadata beanAnnotationMetadata;
     private Environment environment;
     private Set<Class<?>> exposedTypes;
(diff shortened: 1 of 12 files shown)


inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
```diff
@@ -2253,7 +2255,13 @@ private Object coerceCollectionToCorrectType(Class collectionType, Collection be
...
         if (arguments != null) {
             for (Argument argument : arguments) {
-                requiredComponents.add(argument.getType());
+                if (argument.isContainerType() || argument.isProvider()) {
+                    argument.getFirstTypeVariable()
+                            .map(Argument::getType)
+                            .ifPresent(requiredComponents::add);
+                } else {
+                    requiredComponents.add(argument.getType());
+                }
```
graemerocher, line 2264: The `isProvider` changes look ok, with regards to this change we are trying to eliminate all runtime evaluation in `AbstractBeanDefinition` with @dstepanov's PR so I would not like to add more here and this should be instead computed at build time by building on the work Denis has done.

Out of interest what is the use case? Since `getRequiredComponents` is very rarely used.

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/AbstractBeanDefinition.java
jameskleeh: This is used for shutdown behavior to ensure injected beans are shutdown after the bean they are injected into

dstepanov: I would have to port it to my pr

graemerocher: @dstepanov can you do that?

dstepanov: Sure

dstepanov: We can merge it so the tests are included

PR #5717, opened 2021-07-08
Remove support for runtime classpath scanning with ASM

graemerocher:
This change removes the support for runtime classpath scanning with ASM. This functionality is only used by the Hibernate/JPA module and in all recent versions of Micronaut 2.x entities are located via introspections and not via this API so it is no longer necessary.

Keeping this around significantly increases the native image size for Micronaut since it makes ASM reachable and pulls in a large amount of unnecessary code at runtime.


dstepanov (changes requested):
I think we can remove asm dependency from core

graemerocher:
probably makes sense to just make ASM a direct dependency of `inject-java` and `inject-groovy`. Really we should be moving the `BeanDefinitionWriter` etc. classes to another module outside `inject` as they aren't needed on the runtime classpath

dstepanov (approved):

PR #5735, opened 2021-07-12
Register inner classes of classes annotated with @ReflectiveAccess

graemerocher:
In previous versions of Micronaut the handling of `@Introspected` produced reflective metadata for inner enums.

Since the changes to remove reflective access from introspected this is no longer the case. This restores the previous behaviour for `@Entity` with inner classes that is causing the Micronaut SQL tests to fail.


graemerocher:
See https://github.com/micronaut-projects/micronaut-sql/runs/3044686172?check_suite_focus=true#step:7:11769

PR #5742, opened 2021-07-12
Remove deprecations

jameskleeh:


jameskleeh:
This is broken due to micronaut test using a deprecated API. Need to release that first

PR #5753, opened 2021-07-13
Bump micronaut-groovy to 3.0.0-M1

micronaut-build:


jameskleeh:
Some gorm releases are missing from central, thus this error. I'm going to see about getting them in central. If that can't happen I'll roll back to releases that are

PR #5790, opened 2021-07-20
Make resource loader fix GraalVM compatible. Fixes #5765

### LINKED ISSUE #5765: Regression with 2.5.9 and static resources in GraalVM
This commit (https://github.com/micronaut-projects/micronaut-core/commit/a0cfeb13bf1ef5d692d16d4a3b91b34b7456bb11) has broken static resource support for GraalVM in Micronaut 2.5.9. It works properly using the JVM.

### Steps to Reproduce

- `git clone https://github.com/micronaut-graal-tests/micronaut-views-graal`
- `cd micronaut-views-graal`
- `git checkout issue-static-resources-259`
- `sdk use java 21.1.0.r11-grl`
- `./gradlew nativeImage`
- `build/native-image/views-velocity`

Send the following curl requests:
- `curl localhost:8080`
- `curl localhost:8080/folder1/link1.html`

### Expected Behaviour

Both `curl` requests should return HTML.

### Actual Behaviour

They both return 404.

### Environment Information

- **Operating System**: Linux Mint 20.1
- **Micronaut Version:** 2.5.9
- **JDK Version:** Jdk 11

graemerocher:


ilopmar (approved):
The GraalVM app works with this patch without needed to declare the directory in resource config.

jameskleeh:
@graemerocher So with this change relative paths will be blocked under all conditions (GraalVM or not)? The boolean flag on the constructor is just for the case where a user wants to customize the behavior and is not used anywhere within Micronaut?

graemerocher:
That is correct

PR #5794, opened 2021-07-20
move BasicAuthArgumentBinder to http-server

sdelamo:
This moves the `BasicAuthArgumentBinder` to `http-server` since it does not need anything jetty related and it was in `http-server-netty`


http-server/src/main/java/io/micronaut/http/server/binding/BasicAuthBinderRegistrar.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A binder registrar that registers Reactive related binders.
+ *
+ * @author Sergio del Amo
+ * @since 3.0.0
+ */
+@Singleton
+@Internal
+class BasicAuthBinderRegistrar implements BeanCreatedEventListener<RequestBinderRegistry> {
```
jameskleeh: This doesn't need its own registrar. There is already one for http-server. It can just be registered there

sdelamo: Which `BinderRegistrar` already exists in `http-server`? I don't find it.

graemerocher: it is in `http`

PR #5795, opened 2021-07-20
register object to publisher converter

sdelamo:
> To be used for apps using http-server but not http-server-netty. E.g. AWS Lambda with Controllers.


jameskleeh:
This really doesn't have anything to do with the http-server, so perhaps it is better done in the `context` project

PR #5809, opened 2021-07-21
Events improvements

dstepanov:


http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -252,19 +256,17 @@ private void cleanupRequest(ChannelHandlerContext ctx, NettyHttpRequest request)
...
-                try {
-                    beanContext.publishEvent(
-                            new HttpRequestTerminatedEvent(
-                                    request
-                            )
-                    );
-                } catch (Exception e) {
-                    if (LOG.isErrorEnabled()) {
-                        LOG.error("Error publishing request terminated event: " + e.getMessage(), e);
+            if (terminateEventPublisher != ApplicationEventPublisher.NO_OP) {
```
jameskleeh, line 259: Perhaps add a method to event publisher to determine if there are listeners instead. It would be pretty difficult for users to find out this solution

dstepanov: This is a performance hack but I guess it can be more official

src/main/docs/guide/appendix/breaks.adoc
```diff
@@ -379,6 +379,10 @@ public class AProvider implements Provider<A> {
 }
 ----
 
+==== `ApplicationEventPublisher` has a generic event type and it's no longer possible to register an event listener after context has started
```
jameskleeh: The act of registering a listener post context startup is pretty unlikely so imo not worth mentioning here. More comprehensive documentation about way it was done before and how it should be done now should be reflected in the whatsNew and upgrading docs.

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/context/events/SampleEventEmitterBean.kt
```diff
@@ -24,7 +24,7 @@ import jakarta.inject.Singleton
 class SampleEventEmitterBean {
```
jameskleeh, line 24: The groovy and java ones also need updating

dstepanov: done

src/main/docs/guide/appendix/breaks.adoc
dstepanov: Added whatsNew entry

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
dstepanov: Added `ApplicationEventPublisher#noOp()`

jameskleeh:
We should probably deprecate the `ApplicationEventPublisher` methods when used on `BeanContext` since that isn't the recommended way

dstepanov:
> We should probably deprecate the `ApplicationEventPublisher` methods when used on `BeanContext` since that isn't the recommended way

I did deprecate them

inject/src/main/java/io/micronaut/context/BeanContext.java
```diff
@@ -44,10 +45,20 @@
         ExecutionHandleLocator,
         BeanLocator,
         BeanDefinitionRegistry,
-        ApplicationEventPublisher,
+        ApplicationEventPublisher<Object>,
         AnnotationMetadataResolver,
         MutableAttributeHolder {
 
+    @Override
+    @Deprecated
```
jameskleeh, line 59: Add `@deprecated` docs to show what to do instead

graemerocher:
@dstepanov looks good but could you update the docs https://docs.micronaut.io/latest/guide/#contextEvents which use the deprecated way?

dstepanov:
Docs added

PR #5884, opened 2021-07-26
Select constructor with most arguments

### LINKED ISSUE #5852: Kotlin immutable classes cannod be modified in v3
Looks like something is changed and for the cases like this when Kotlin generates multiple constructors, the empty one is being picked up in v3, preventing modifying the bean with `withValue` method.
```
@MappedEntity
data class SomeEntity(
    @field:GeneratedValue
    @field:Id
    val id: Long? = null,
    val something: Int? = null
)
```
The workaround is to remove the defaults:
```
@MappedEntity
data class SomeEntity(
    @field:GeneratedValue
    @field:Id
    val id: Long?
    val something: Int?
)
```

I think we should change how we select the bean constructor, we should prefer the constructor that is able to initialize the maximum about of the immutable properties.

graemerocher:
The issue is that Kotlin 1.5 changed the constructor order in KAPT and now the no args constructor is first. This changes the logic so the constructor with the most arguments is selected, however this has implications since some code may break (I had to change `NettyByteBufferFactory` therefore probably needs discussion.

Fixes #5852

### DIFF
--- a/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
+++ b/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
@@ -23,6 +23,7 @@ import io.micronaut.core.io.buffer.ByteBufferFactory;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.Unpooled;
+import jakarta.inject.Inject;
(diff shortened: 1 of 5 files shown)


jameskleeh:
@graemerocher Please verify the workaround works:

```
import io.micronaut.core.annotation.Creator

@Introspected
data class SomeEntity @Creator constructor(
        val id: Long? = null,
        val something: String? = null
)
```

If not, I think this needs to be a member of the `@Introspected` annotation

PR #5888, opened 2021-07-26
Add formatting to query values via the @Format annotation

andriy-dmytruk:
The support of `@Format` annotation to be used alongside with `@QueryValue` is added for both server and client.

The supported values for the `Format.value` attribute are:
* `"CSV"`, `"SSV"`, `"PIPES"` - delimited values
* `"MULTI"` - repeating values
* `"DEEP_OBJECT"` - parameter with attributes as `param[attr]=value`

This is very similar to [Open API v3 query parameter style attribute](https://swagger.io/docs/specification/serialization/) (The only difference is that `style=form`, `expand=false` is `"CSV"` here and `style=form`, `expand=true` is `"MULTI"` to avoid adding expand attribute.

The following types are supported:
* `Iterable`
* `Map`
* POJO `Object` with `@Introspected` annotation

Other types should not be used with any of the above formats, as it will result in NotIntrospected error.

The functionality is implemented by adding `Converter<ConvertibleMultiValues, DesiredType>` on server side and `Converter<GivenType, ConvertibleMultiValues>` on client side. The converters will return `Optional.empty()` if the format is not one of the specified above. 

The binders then firstly try to convert with `ConvertibleMultiValues` type (which could allow user to define custom converters if needed). If it fails, the binders will try to convert with `String`.

@jameskleeh Please, take a look.


http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentRequestBinder.java
```diff
@@ -0,0 +1,400 @@
...
+
+    private final ConversionService<?> conversionService;
+
+    public QueryValueClientArgumentRequestBinder(ConversionService<?> conversionService) {
+        this.conversionService = conversionService;
+
+        // Add the converters
+        conversionService.addConverter(Iterable.class, ConvertibleMultiValues.class, new IterableConverter());
+        conversionService.addConverter(Map.class, ConvertibleMultiValues.class, new MapConverter());
+        conversionService.addConverter(Object.class, ConvertibleMultiValues.class, new ObjectConverter());
```
jameskleeh: It's not appropriate to add the converters here. We have a converter registrar interface you can use which would make them available before this class gets instantiated

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentRequestBinder.java
```diff
@@ -0,0 +1,400 @@
...
+        ArgumentConversionContext<ConvertibleMultiValues> conversionContext = ConversionContext.of(
+                Argument.of(ConvertibleMultiValues.class, context.getArgument().getName(), context.getAnnotationMetadata()));
+        Optional<ConvertibleMultiValues> convertedValue = conversionService.convert(value, conversionContext);
+
+        if (convertedValue.isPresent()) {
+            // TODO How should I check if it is truely a List of Strings?
+            ConvertibleMultiValues<String> multiValues;
+            try {
+                // noinspection unchecked
+                multiValues = convertedValue.get();
```
jameskleeh: .get() should never throw an exception if `isPresent()` is true

http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
```diff
@@ -0,0 +1,176 @@
...
+
+    void "test csv format all values"() {
+        expect:
+        client.csvFormattedObject(param) == query
+
+        where:
+        param                                  | query
+        ["a", "b", "c"]                        | "csv=a,b,c"
+        ["name": "Bob", "age": 12]             | "csv=name,Bob,age,12"
+        new Cafe(name: "AAA", address: "here") | "csv=name,AAA,address,here"
```
jameskleeh, line 51: These use cases don't make sense to me. It seems really odd to want to convert a map to a csv where the keys of the map are part of the values. Do you have some reference behavior in some other framework or documentation that was used as the inspiration?

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/QueryParameterFormattingSpec.groovy
```diff
@@ -0,0 +1,183 @@
...
+        where:
+        uri                                                           | result
+        // List
+        '/formatted/csv?param=a,b,c'                                  | ["a", "b", "c"].inspect()
+        '/formatted/ssv?v=a b d'                                      | ["a", "b", "d"].inspect()
+        '/formatted/pipes?param=what|is|life'                         | ["what", "is", "life"].inspect()
+        '/formatted/multi?value=one&value=two&value=3'                | ["one", "two", "3"].inspect()
+        '/formatted/deep?param[0]=a&param[1]=b'                       | ["a", "b"].inspect()
+        '/formatted/csv'                                              | [].inspect()
+        '/formatted/csv?param='                                       | [].inspect()
```
jameskleeh, line 42: It would make more sense to me if the `?param=` was omitted in this case

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/QueryParameterFormattingSpec.groovy
```diff
@@ -0,0 +1,183 @@
...
+        '/formatted/pipes?param=|two'                                 | ['', 'two'].inspect()
+        // Map
+        '/formatted/m/csv?p=key,value,a,b,c,d'                        | ["key": "value", "a": "b", "c": "d"].toSorted().inspect()
+        '/formatted/m/ssv?param=start 0 end 1'                        | ["start": 0, "end": 1].toSorted().inspect()
+        '/formatted/m/pipes?param=a|b|c|d|e'                          | ["a": "b", "c": "d"].toSorted().inspect()
+        '/formatted/m/multi?k=1&val=2&c=3'                            | ["k": "1", "val": "2", "c": "3"].toSorted().inspect()
+        '/formatted/m/deep?v[start]=0&v[end]=2&v[middle]=3'           | ["start": 0, "end": 2, "middle": 3].toSorted().inspect()
+        '/formatted/m/ssv'                                            | [:].inspect()
+        '/formatted/m/ssv?param='                                     | [:].inspect()
```
jameskleeh, line 51: Same as above, for empty it makes sense to me to not include the key

src/main/docs/guide/httpServer/binding.adoc
```diff
@@ -67,6 +67,44 @@ Instead of binding from a single section of the request, it may be desirable to
...
+
+The format can only be applied to `java.lang.Iterable`, `java.util.Map` or POJO with link:{api}/io/micronaut/core/annotation/Introspected.html[Introspected] annotation. Examples of how different values will be formatted are given in the table below:
+
+|===
+| Format | Iterable example | Map example | Object example
+
+| `"CSV"`
+| `"param=Mike,Adam,Kate"`
+| `"param=name,Mike,age,30"`
+| `"param=name,Mike,age,30"`
```
jameskleeh: You can probably condense this to just one column (Map/Pojo example)

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentRequestBinder.java
```diff
@@ -0,0 +1,400 @@
...
+     * delimited with commas.
+     * <table borer="0">
+     *     <tr> <th><b> Type </b></th>      <th><b> Example value </b></th>                     <th><b> Example representation </b></th> </tr>
+     *     <tr> <td> Iterable &emsp </td>   <td> param=["Mike", "Adam", "Kate"] </td>           <td> "param=Mike,Adam,Kate" </td></tr>
+     *     <tr> <td> Map </td>              <td> param=["name": "Mike", "age": "30"] &emsp</td> <td> "param=name,Mike,age,30" </td> </tr>
+     *     <tr> <td> Object </td>           <td> param={name: "Mike", age: 30} </td>            <td> "param=name,Mike,age,30" </td> </tr>
+     * </table>
+     * Note that ambiguity may arise when the values contain commas themselves after being converted to String.
+     */
+    public static final String QUERY_VALUE_FORMAT_CSV = "CSV";
```
jameskleeh: I think these names should be lowercase, or at least match against what is put into the `@Format` annotation case insensitive

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentRequestBinder.java
andriy-dmytruk: I wasn't sure that the converters would be useful if someone is not using the micronaut-http or micronaut-client. Should I add them to the converter register anyway?

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentRequestBinder.java
andriy-dmytruk: The try-catch was in case the type was some ConvertibleMultiValues<T> and not ConvertibleMultiValues<String>, but since java won't throw an exception in this case, I will probably have to check the values inside.

http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
andriy-dmytruk: It is here if you scroll to the Query parameters and see the "form" style in the table: [Open api v3 parameter serialization](https://swagger.io/docs/specification/serialization/).
But I do agree that it doesn't make much sense.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/QueryParameterFormattingSpec.groovy
andriy-dmytruk: I did it to test how it works in corner cases. I don't think there should be any kind of exception if user does write it in this way.

http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
jameskleeh: In that table it shows N/A for `spaceDelimited` and `pipeDelimited` with objects. So this sort of test seems to go against that source `new Cafe(name: "AAA", address: "here") | "param=name|AAA|address|here"`

jameskleeh: It may just make sense to support it though to be consistent. Seems odd its N/A on their side

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/QueryParameterFormattingSpec.groovy
jameskleeh: I agree there shouldn't be an exception. I was asserting the `?param=` should not be present in the URI

http-client-core/src/main/java/io/micronaut/http/client/bind/binders/QueryValueClientArgumentRequestBinder.java
jameskleeh: I think they are useful outside of those contexts. For example in configuration, websockets, etc

jameskleeh:
Superseded by https://github.com/micronaut-projects/micronaut-core/pull/6061

PR #5894, opened 2021-07-27
Create inject-kotlin-test module for directly testing kapt processing in unit tests

yawkat:
This new module adds a `AbstractKotlinCompilerSpec` that calls directly into the kotlin compiler, similar to the java `AbstractTypeElementSpec`. Right now, only `buildClasspath` is supported, along with `buildContext` and `buildBeanIntrospection` built on that.

The code that calls into the compiler is adapted from [kotlin project integration tests](https://github.com/JetBrains/kotlin/blob/master/plugins/kapt3/kapt3-compiler/test/org/jetbrains/kotlin/kapt3/test/AbstractKotlinKapt3IntegrationTest.kt). It goes quite deep into compiler internals, so it may not be very compatible between kotlin versions.

Another potential snag is the classpath scanning: The compiler requires a list of all dependencies, and this code builds that list from the current classpath. Because of changes to system ClassLoader infrastructure since java 9, we have to rely on parsing the `java.class.path` system property. This works fine on linux java 11, but I can't test on other OS. It'll also break if it were called in more complicated classloader setups, but that should not be a concern for code that is only used in unit/integration tests.

### DIFF
--- a/inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt
+++ b/inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt
@@ -1,0 +1,240 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.annotation.processing.test
+
+import org.jetbrains.kotlin.base.kapt3.DetectMemoryLeaksMode
+import org.jetbrains.kotlin.base.kapt3.KaptOptions
+import org.jetbrains.kotlin.cli.common.CLIConfigurationKeys
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSourceLocation
+import org.jetbrains.kotlin.cli.common.messages.MessageCollector
+import org.jetbrains.kotlin.cli.jvm.compiler.CliBindingTrace
+import org.jetbrains.kotlin.cli.jvm.compiler.EnvironmentConfigFiles
+import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
+import org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM
+import org.jetbrains.kotlin.cli.jvm.config.JvmClasspathRoot
+import org.jetbrains.kotlin.codegen.ClassBuilderMode
+import org.jetbrains.kotlin.codegen.DefaultCodegenFactory
+import org.jetbrains.kotlin.codegen.KotlinCodegenFacade
+import org.jetbrains.kotlin.codegen.OriginCollectingClassBuilderFactory
+import org.jetbrains.kotlin.codegen.state.GenerationState
+import org.jetbrains.kotlin.com.intellij.psi.PsiFileFactory
+import org.jetbrains.kotlin.com.intellij.psi.impl.PsiFileFactoryImpl
+import org.jetbrains.kotlin.com.intellij.testFramework.LightVirtualFile
+import org.jetbrains.kotlin.config.CommonConfigurationKeys
+import org.jetbrains.kotlin.config.CompilerConfiguration
+import org.jetbrains.kotlin.config.JVMConfigurationKeys
+import org.jetbrains.kotlin.idea.KotlinLanguage
+import org.jetbrains.kotlin.kapt3.AbstractKapt3Extension
+import org.jetbrains.kotlin.kapt3.base.LoadedProcessors
+import org.jetbrains.kotlin.kapt3.base.incremental.DeclaredProcType
+import org.jetbrains.kotlin.kapt3.base.incremental.IncrementalProcessor
+import org.jetbrains.kotlin.kapt3.util.MessageCollectorBackedKaptLogger
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.AnalyzingUtils
+import org.jetbrains.kotlin.resolve.jvm.extensions.AnalysisHandlerExtension
+import java.io.File
+import java.io.IOException
+import java.net.URL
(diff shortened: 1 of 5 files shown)


graemerocher:
This PR applies to Micronaut 2.5.x and Kotlin 1.4

A separate PR is being prepared for Micronaut 3.0.x and Kotlin 1.5

yawkat:
Works in local testing with kotlin 1.5 too. There are some changes needed for mn3 compatibility though. They are the same changes that are already present in `AbstractTypeElementSpec`, and some javax->jakarta stuff.

```patch
===================================================================
diff --git a/inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/AbstractKotlinCompilerSpec.groovy b/inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/AbstractKotlinCompilerSpec.groovy
--- a/inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/AbstractKotlinCompilerSpec.groovy	(revision 38cd274ba7da3af23aa85a08cc0d433752cbe045)
+++ b/inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/AbstractKotlinCompilerSpec.groovy	(date 1627372950160)
@@ -15,6 +15,7 @@
  */
 package io.micronaut.annotation.processing.test
 
+import io.micronaut.aop.internal.InterceptorRegistryBean
 import io.micronaut.context.ApplicationContext
 import io.micronaut.context.DefaultApplicationContext
 import io.micronaut.context.Qualifier
@@ -23,6 +24,8 @@
 import io.micronaut.core.naming.NameUtils
 import io.micronaut.inject.BeanDefinition
 import io.micronaut.inject.BeanDefinitionReference
+import io.micronaut.inject.provider.BeanProviderDefinition
+import io.micronaut.inject.provider.JakartaProviderBeanDefinition
 import org.intellij.lang.annotations.Language
 
@@ -54,7 +57,7 @@
      *
      * @return the introspection if it is correct
      */
-    protected ApplicationContext buildContext(String className, @Language("kotlin") String cls) {
+    protected ApplicationContext buildContext(String className, @Language("kotlin") String cls, boolean includeAllBeans = false) {
         def result = KotlinCompileHelper.INSTANCE.run(className, cls)
         ClassLoader classLoader = result.classLoader
 
@@ -63,10 +66,16 @@
             protected List<BeanDefinitionReference> resolveBeanDefinitionReferences(Predicate<BeanDefinitionReference> predicate) {
                 // we want only the definitions we just compiled
                 def stream = result.fileNames.stream()
-                        .filter(s -> s.endsWith("DefinitionClass.class"))
+                        .filter(s -> s.endsWith('$Definition$Reference.class'))
                         .map(n -> classLoader.loadClass(n.substring(0, n.size() - 6).replace('/', '.')).newInstance())
                 if (predicate != null) stream = stream.filter(predicate)
-                return stream.collect(Collectors.toList())
+                return stream.collect(Collectors.toList()) + (includeAllBeans ? super.resolveBeanDefinitionReferences(predicate) : [
+                        new InterceptorRegistryBean(),
+                        new BeanProviderDefinition(),
+                        new JakartaProviderBeanDefinition(),
+                        Class.forName('io.micronaut.runtime.event.$ApplicationEventPublisherFactory$Build0$Definition$Reference').newInstance(),
+                        Class.forName('io.micronaut.runtime.event.$ApplicationEventPublisherFactory$Definition$Reference').newInstance()
+                ])
             }
         }.start()
     }
@@ -76,7 +85,7 @@
     }
 
     protected BeanDefinition buildBeanDefinition(String className, @Language("kotlin") String cls) {
-        def beanDefName= '$' + NameUtils.getSimpleName(className) + 'Definition'
+        def beanDefName= '$' + NameUtils.getSimpleName(className) + '$Definition'
         def packageName = NameUtils.getPackageName(className)
         String beanFullName = "${packageName}.${beanDefName}"
 
Index: inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt b/inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt
--- a/inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt	(revision 38cd274ba7da3af23aa85a08cc0d433752cbe045)
+++ b/inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt	(date 1627372350556)
@@ -196,7 +196,7 @@
             cl.files[outDir.relativize(p).toString()] = Files.readAllBytes(p)
         }
 
-        return Result(URLClassLoader(arrayOf(outDir.toUri().toURL()), cl), cl.files.keys)
+        return Result(cl, cl.files.keys)
     }
 
     private fun getClasspath(cl: ClassLoader): List<File> =
Index: inject-kotlin-test/src/test/groovy/io/micronaut/annotation/processing/test/Test.groovy
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/inject-kotlin-test/src/test/groovy/io/micronaut/annotation/processing/test/Test.groovy b/inject-kotlin-test/src/test/groovy/io/micronaut/annotation/processing/test/Test.groovy
--- a/inject-kotlin-test/src/test/groovy/io/micronaut/annotation/processing/test/Test.groovy	(revision 38cd274ba7da3af23aa85a08cc0d433752cbe045)
+++ b/inject-kotlin-test/src/test/groovy/io/micronaut/annotation/processing/test/Test.groovy	(date 1627372350564)
@@ -33,8 +33,8 @@
         def context = buildContext('example.Foo', '''
 package example
 
-import javax.inject.Inject
-import javax.inject.Singleton
+import jakarta.inject.Inject
 
 @Singleton
 class Foo {
@@ -56,7 +56,7 @@
         given:
         def definition = buildBeanDefinition('test.Test', '''
 package test;
-import javax.inject.Singleton;
 
 @Singleton
 class Test {
```

graemerocher (approved):

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/KotlinCompileHelper.kt
```diff
@@ -0,0 +1,240 @@
...
+import javax.annotation.processing.Processor
+
+object KotlinCompileHelper {
+    init {
+        System.setProperty("idea.ignore.disabled.plugins", "true")
+        System.setProperty("idea.io.use.nio2", "true")
+    }
+
+    fun run(className: String, code: String): Result {
+        val tmp = Files.createTempDirectory("abc")
```
jameskleeh: You can pass null here or provide a more useful name

yawkat: yes, will do, forgot to change that :)

PR #5895, opened 2021-07-27
Set Micronaut Data to 2.4.7

ilopmar:
Downgrade Micronaut Data to 2.4.7 for Micronaut 2.5.x. Version 2.5.0+ will target Micronaut 3.0


graemerocher:
We need a new version for Micronaut 3.0.x so Data 2.5.x won't

ilopmar:
This is to avoid upgrading to a new minor version of Data (that could potentially bring new version of dependencies) for 2.5.x users. https://github.com/micronaut-projects/micronaut-core/wiki/Module-Upgrade-Strategy

ilopmar:
As agreed during engineering meeting we'll keep Micronaut Data 2.5.0 because it only updates some patch versions of Hibernate and Reactor, fixes some issues and adds a new feature.

PR #5901, opened 2021-07-28
Fix regression: ServerRequestContext not accessible in WebSocket filters

### LINKED ISSUE #5900: Regression ServerRequestContext::currentRequest not working in HTTP filters of WebSocket requests
### Steps to Reproduce

1. Create an `HttpFilter` and annotate it with `@Filter(...)` to match a WebSocket endpoint
2. Call `ServerRequestContext.currentRequest().orElseThrow()` to verify that the current request is set. 
3. Call the WebSocket endpoint to trigger the filter.

### Expected Behaviour

The filter should not throw the exception as the current request is expected to be set in `ServerRequestContext`.

### Actual Behaviour

The filter throws the excpetion.

### Environment Information

- **Operating System**: macOS 11.5
- **Micronaut Version:** 2.5.1
- **JDK Version:** OpenJDK 11.0.11

### Example Application

I will soon raise a draft PR to demonstrate the regression in the built-in WebSocket tests by adding such a filter.

Related to #5400

lgathy:
Closes #5900

Posting this as draft for now to demonstrate the regression in the tests.
Note: I know this is not the nicest way to extend the tests as I did not create new cases but modified the existing test suite to demonstrate this behaviour. I was happy I could reproduce this, but I am open to suggestions on how to separate it properly into new test cases, if needed.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
@@ -21,6 +21,7 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.*;
 import io.micronaut.http.codec.MediaTypeCodecRegistry;
+import io.micronaut.http.context.ServerRequestContext;
 import io.micronaut.http.exceptions.HttpStatusException;
 import io.micronaut.http.filter.HttpFilter;
 import io.micronaut.http.filter.HttpServerFilter;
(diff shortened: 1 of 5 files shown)


dstepanov:
Please check if there is an existing test that tests the presence of the server request context in web socket controllers/handlers.
We would need to test web-socket filters as well.

lgathy:
@dstepanov Just to clarify, you would like me to add tests where the current request is asserted in the web socket handlers as well? If so, how about creating a copy of the chat server/client as let's say `/context/chat/**` which responds with the current request uri to every open/close/message events? Btw, I searched for the term `ServerRequestContext.currentRequest()` in the whole repo and did not find any web socket tests already doing this.
Regarding web socket filters: isn't the newly added `WebSocketContextValidationFilter` enough to prove that point? Btw, I could restrict that to the new `/context/chat/**` endpoints to limit its effect only on that test case, what do you think?

jameskleeh:
@lgathy A test that was separate from the others would be better

dstepanov:
@lgathy If you can add the check to the existing tests that the request context is propagating it would be great

jameskleeh:
> @lgathy If you can add the check to the existing tests that the request context is propagating it would be great

That works too

lgathy:
@jameskleeh @dstepanov How about this separate test? I copied the text based chat test and appended the request uri to each message broadcasted. This way the filter only has an effect on the new endpoints.

jameskleeh:
@lgathy You made commits from 2 separate github accounts so you'll need to sign the CLA with both

lgathy:
> You made commits from 2 separate github accounts so you'll need to sign the CLA with both

That was a mistake, but I managed to fix it 1m later by force pushing to the feature branch. Thanks for merging this.

PR #5908, opened 2021-07-30
Makes the behaviour of a missing provider configurable

graemerocher:
In CDI a missing provider does not result an exception on injection but instead only when the provider's get() method is called. This differs from Micronaut and in order to make it compliant this behaviour needs to be configurable.


jameskleeh:
What do you think about `builder.allowEmptyProviders(false)`?

graemerocher:
Fine by me

PR #5909, opened 2021-07-30
Move ApplicationEventPublisherFactory to inject module

graemerocher:
With the changes to ApplicationEventPublisher the `micronaut-inject` module doesn't work standalone anymore unless `micronaut-context` is added. This fixes that.


msgilligan:
Thanks for this fix, @graemerocher!  Supernaut.FX is using `micronaut-inject` standalone.

PR #5918, opened 2021-08-02
Routing refactor

jameskleeh:


http/src/main/java/io/micronaut/http/HttpRequest.java
```diff
@@ -35,6 +35,8 @@
 @SuppressWarnings("MethodName")
 public interface HttpRequest<B> extends HttpMessage<B> {
 
+    String KEY = "micronaut.http.request";
```
graemerocher, line 38: Should be `SERVER_KEY` with a value of `micronaut.http.server-request` since you will want to store both the current server request and current client request in the reactive flow

graemerocher (changes requested):
Looks good, only change I would make is the key used for the request in the context

http/src/main/java/io/micronaut/http/HttpRequest.java
jameskleeh: Reactor doesn't allow modifications to the context to be available downstream, so populating the request in a client publisher response won't have any effect.

PR #5921, opened 2021-08-03
System and environment properties not overriding application.yml for logger configuration

David-16:
The original logging code putAlls the generated properties (those returned by getProperties(String)) first and then putAlls the raw properties (those returned by getProperties(String, RAW))

This causes properties set in application.yaml to override environment variables which violates expectations

This fix gets both properties and putAlls them in opposite order

Still working on a test (am getting environment issues when running locally); will update once I get it worked out

### DIFF
--- a/runtime/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java
+++ b/runtime/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java
@@ -80,9 +80,14 @@ final class PropertiesLoggingLevelsConfigurer implements ApplicationEventListene
     }
 
     private void configureLogLevels() {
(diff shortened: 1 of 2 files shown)


David-16:
Added test.  Confirmed test fails without the fix

David-16:
@jameskleeh Sorry to bother, but is there any action item waiting on my end in order to have this merged?

jameskleeh:
@David-16 No, however I'm not sure this should be merged. The generated properties can contain things that would not be considered normal package names and may cause issues. This also makes the assumption that generated properties have priority over the raw which could be problematic.

Edit: Nevermind - I misread the diff. This should be OK I think. I'll bring it up for discussion to make sure there are no concerns

PR #5925, opened 2021-08-05
Mark some jackson-related classes as internal

yawkat:
This PR makes some classes related to jackson-databind internal. Getting this into 3.0 will make it easier to replace jackson-databind without needing compatibility classes in the future.

Using github code search, I have verified that none of these classes are used in the micronaut-projects organization, outside of micronaut-core. Should someone rely on these classes not being internal, they will get a warning so they can either migrate away or notify us so that we can make them usable again.

@graemerocher : I mentioned most of these classes yesterday, but I added some more: The SPI property source loader implementations and ObjectNodeConvertibleValues.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonViewCodecResolver.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonViewCodecResolver.java
@@ -24,7 +24,7 @@ import io.micronaut.jackson.codec.JsonMediaTypeCodec;
  * @author graemerocher
  * @since 1.1
  */
-public interface JsonViewCodecResolver {
+interface JsonViewCodecResolver {
     /**
      * Resolves a {@link JsonMediaTypeCodec} for the view class (specified as the JsonView annotation value).
      * @param viewClass The view class
--- a/runtime/src/main/java/io/micronaut/jackson/bind/JacksonBeanPropertyBinder.java
+++ b/runtime/src/main/java/io/micronaut/jackson/bind/JacksonBeanPropertyBinder.java
@@ -51,7 +51,7 @@ import java.util.Set;
  */
 @Singleton
 @Primary
-public class JacksonBeanPropertyBinder implements BeanPropertyBinder {
+class JacksonBeanPropertyBinder implements BeanPropertyBinder {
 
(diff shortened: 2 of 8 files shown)


graemerocher (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonViewCodecResolver.java
```diff
@@ -24,6 +25,7 @@
  * @author graemerocher
  * @since 1.1
  */
+@Internal
 public interface JsonViewCodecResolver {
```
jameskleeh: This can just be made package private. The method is already package private so it wouldn't be feasible to use the interface in any meaningful way

runtime/src/main/java/io/micronaut/jackson/bind/JacksonBeanPropertyBinder.java
```diff
@@ -51,6 +52,7 @@
  */
 @Singleton
 @Primary
+@Internal
 public class JacksonBeanPropertyBinder implements BeanPropertyBinder {
```
jameskleeh: Same thing here. Make the class and its constructor package private

http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonViewCodecResolver.java
yawkat: which method is package-private?

jameskleeh: Nevermind - confused it for a class. Still though I think we can make the class package private (not saying we shouldn't add @Internal)

graemerocher: it is an interface it is public

yawkat: in that case, I will also make the only implementation package-private. I assume `@Internal` is not required then either?

yawkat:
Build works locally on java 11, so probably a spurious failure?

PR #5936, opened 2021-08-06
Add generics aware `isAssignableFrom` to `Argument`

dstepanov:


core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -174,6 +174,37 @@ default boolean isAssignableFrom(@NonNull Class<?> candidateType) {
         return getType().isAssignableFrom(Objects.requireNonNull(candidateType, "Candidate type cannot be null"));
     }
 
+    /**
+     * Checks if the argument can be assigned to this argument.
+     * @param candidateArgument The candidate argument
+     * @return True if it is assignable from.
+     * @since 3.0.0
+     */
+    default boolean isAssignableFrom(@NonNull Argument candidateArgument) {
```
graemerocher: Should be `Argument<?>` with the wild card

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -174,6 +174,37 @@ default boolean isAssignableFrom(@NonNull Class<?> candidateType) {
...
+            return candidateArgumentTypeParameters.length >= 0;
+        }
+        if (candidateArgumentTypeParameters.length == 0) {
+            // Wildcard
+            return false;
+        }
+        for (int i = 0; i < typeParameters.length; i++) {
+            Argument typeParameter = typeParameters[i];
+            Argument candidateArgumentTypeParameter = candidateArgumentTypeParameters[i];
+            if (!typeParameter.isAssignableFrom(candidateArgumentTypeParameter)) {
```
graemerocher, line 201: Do we need to check `isTypeVariable`?

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/type/Argument.java
dstepanov: We know it's coming from the type parameters so I don't think so

core/src/main/java/io/micronaut/core/type/Argument.java
dstepanov: Added

PR #5939, opened 2021-08-06
Add utility method to convert `Annotation[]` to `Qualifier`

dstepanov:
Maybe the new method and `byAnnotation` should ignore any kind of annotation that isn't annotated with `@Qualifier`


graemerocher:
I'm not sure we should promote using `Annotation` as a type because it requires reflection get the members unless we only support `Annotation` instances created from `synthesize` (and a proxy interface to retrieve the values)

dstepanov:
I can have it in some internal utility class

PR #5942, opened 2021-08-06
Bump micronaut-openapi to 2.6.1

micronaut-build:


ilopmar:
Micronaut OpenAPI 2.5.0 wasn't included in any Micronaut version but it only contains some bug fixes and minor upgrades. Then 2.6.0 targeted Micronaut 3.0.x but it's compiled with Micronaut 2.5.x and it didn't contain any breaking change.

I discussed this with James and we agreed that this new 2.6.1 contains some important bug fixes and is compatible with Micronaut 2.5.x so it should be safe for users to upgrade to this even in a Micronaut patch version.

PR #5945, opened 2021-08-06
Provide unmatched MediaType extension from fileUpload

rawilder:
When a fileUpload has a contentType that doesn't have a matching MediaType, it will create a MediaType with only the contentType such that extension is defaulted to contentType.

This PR proposes checking the filename for an extension and providing it for use in creating the new MediaType if no match is found.

### DIFF
--- a/http/src/main/java/io/micronaut/http/MediaType.java
+++ b/http/src/main/java/io/micronaut/http/MediaType.java
@@ -469,6 +469,61 @@ public class MediaType implements CharSequence {
         this.strRepr = toString0();
     }
 
+    private static Optional<MediaType> optionalOf(String mediaType) {
+        switch (mediaType) {
+            case ALL:
+                return Optional.of(ALL_TYPE);
+            case APPLICATION_FORM_URLENCODED:
+                return Optional.of(APPLICATION_FORM_URLENCODED_TYPE);
+            case MULTIPART_FORM_DATA:
+                return Optional.of(MULTIPART_FORM_DATA_TYPE);
+            case TEXT_HTML:
+                return Optional.of(TEXT_HTML_TYPE);
+            case APPLICATION_XHTML:
+                return Optional.of(APPLICATION_XHTML_TYPE);
+            case APPLICATION_XML:
+                return Optional.of(APPLICATION_XML_TYPE);
+            case APPLICATION_JSON:
+                return Optional.of(APPLICATION_JSON_TYPE);
+            case APPLICATION_YAML:
+                return Optional.of(APPLICATION_YAML_TYPE);
+            case TEXT_XML:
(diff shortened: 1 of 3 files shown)


jameskleeh:
There is a `NameUtils.extension(filename)` method

jameskleeh:
There is also the `MediaType.forExtension` method

rawilder:
Let me take another look. I still think the idea here is right, but the methods you mentioned would be a much better choice than my rearrangement of the MediaType class. I think my mind on `forExtension` and `forFilename` went to "returns only what types are defined in the MediaType class" but upon further reading I see it references a `mime.types`. 

It still offers up the problem of to use contentType or filename. If contentType defaults to a generic MediaType  wouldn't we want to try filename instead? A kind of hierarchy and once we've tried both use the MediaType(name) constructor or MediaType(name, ext) if an `ext` exists.

http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -274,7 +274,7 @@
      * PDF: application/pdf.
      */
     public static final MediaType APPLICATION_PDF_TYPE = new MediaType(APPLICATION_PDF);
-
```
rawilder, line 277: sneaky whitespace

http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -817,27 +818,62 @@ public static MediaType forFilename(String filename) {
         return MediaType.TEXT_PLAIN_TYPE;
     }
 
+    /**
+     * Resolve the {@link MediaType} for the given media/content type text.
+     *
+     * @param mediaType The media/content type text
+     * @return An {@link Optional} of {@link MediaType}
+     */
+    public static Optional<MediaType> forMediaType(String mediaType) {
```
rawilder: Since the mime file has more definitions than the switch case, this can be utilized to access that file's definitions instead. I use it in the FileUpload implementations to make them keep using the browser provided file type.

http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -850,7 +886,11 @@ public static MediaType forFilename(String filename) {
                 String[] tokens = formattedLine.split("\\|");
                 for (int i = 1; i < tokens.length; i++) {
                     String fileExtension = tokens[i].toLowerCase(Locale.ENGLISH);
-                    result.put(fileExtension, tokens[0]);
+                    if (keyedByExtension) {
```
rawilder: Have to rethink this since there can be multiple extensions per media type, can't just choose a random one.

jameskleeh:
Superseded by #6106

PR #5954, opened 2021-08-09
Support regex-style patterns in `@Filter`

### LINKED ISSUE #5842: Support for regexes in @Filter for HttpServerFilter
Hi all!

When using `@Filter` annotations with `HttpServerFilter`, currently the `pattern` attribute only supports Ant-style patterns. Because of it, we can't use more complicated patterns (using regex, for example). A simple case that regex can support but Ant can't is accepting all paths except one.

This question was raised on gitter (discussed briefly with @jameskleeh), and I suggested that we support something like `@Filter(pattern="/mypath", patternStyle=ANT)` or `@Filter(pattern="/mypath", patternStyle=REGEX)`, with the default being Ant, to avoid breaking changes.

One thing to note is that using regexes is less performant than using Ant-style patterns. This should be communicated properly to users.

If this is fine, I can come up with a PR.

### Expected Behaviour

Micronaut should support more complex patterns for `@Filter`

### Actual Behaviour

Micronaut only supports simple Ant-style patterns for `@Filter`

racevedoo:
Adds a attribute to the `Filter` annotation, specifying which style of
pattern to use, keeping Ant as the default (avoiding breaking changes).

Fixes #5842

### DIFF
--- a/core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
+++ b/core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.util;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+
+/**
+ * PathMatcher implementation for regex-style patterns.
+ * @author Rafael Acevedo
+ * @since 3.1
+ */
+public class RegexPathMatcher implements PathMatcher {
+  private final Map<String, Pattern> compiledPatterns = new ConcurrentHashMap<>();
+
+  @Override
+  public boolean matches(String pattern, String source) {
+    if (pattern == null || source == null) return false;
+    return compiledPatterns.computeIfAbsent(pattern, Pattern::compile).matcher(source).matches();
+  }
+}
--- a/http/src/main/java/io/micronaut/http/filter/HttpFilterResolver.java
+++ b/http/src/main/java/io/micronaut/http/filter/HttpFilterResolver.java
@@ -78,6 +78,13 @@ public interface HttpFilterResolver<F extends HttpFilter, T extends AnnotationMe
         @NonNull String[] getPatterns();
 
         /**
+         * @return The filter patterns
+         */
+        default FilterPatternStyle getPatternStyle() {
+            return FilterPatternStyle.defaultStyle();
+        }
+
+        /**
          * @return Does the entry define any methods.
          */
         default boolean hasMethods() {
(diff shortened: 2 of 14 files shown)


core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * PathMatcher implementation for regex-style patterns.
+ * @author Rafael Acevedo
+ * @since 3.0
+ */
+public class RegexPathMatcher implements PathMatcher {
+
+  @Override
+  public boolean matches(String pattern, String source) {
+    Pattern compiledPattern = Pattern.compile(pattern);
```
racevedoo: Is it necessary to keep a cache of pre-compiled patterns here? @jameskleeh

core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.core.util;
+
+import java.util.regex.Pattern;
+
+/**
+ * PathMatcher implementation for regex-style patterns.
+ * @author Rafael Acevedo
+ * @since 3.0
```
racevedoo: should it be 3.0 here? or 3.1?

jameskleeh: 3.1

core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
jameskleeh: I think so yes

http/src/main/java/io/micronaut/http/filter/HttpFilterResolver.java
```diff
@@ -77,6 +77,11 @@
          */
         @NonNull String[] getPatterns();
 
+        /**
+         * @return The filter patterns
+         */
+        @NonNull FilterPatternStyle getPatternStyle();
```
jameskleeh: This should be a default method

http/src/main/java/io/micronaut/http/filter/HttpFilterResolver.java
```diff
@@ -96,18 +101,21 @@ default boolean hasPatterns() {
          * @param filter The filter
          * @param annotationMetadata The annotation metadata
          * @param methods The methods
+         * @param patternStyle the pattern style
          * @param patterns The patterns
          * @return The filter entry
          * @param <FT> the filter type
          */
         static <FT extends HttpFilter> FilterEntry<FT> of(
```
jameskleeh, line 110: Need to create a new method here instead of changing the existing one

core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
racevedoo: done

core/src/main/java/io/micronaut/core/util/RegexPathMatcher.java
racevedoo: done!

http/src/main/java/io/micronaut/http/filter/HttpFilterResolver.java
racevedoo: done

http/src/main/java/io/micronaut/http/filter/HttpFilterResolver.java
racevedoo: done!

racevedoo:
@jameskleeh should we wait for a 3.1.x branch before merging this? what's usually done in cases like this?

jameskleeh:
@racevedoo Yep should be soon

jameskleeh:
@racevedoo We can merge this now that we have 3.1.x, however this is missing documentation

racevedoo:
> @racevedoo We can merge this now that we have 3.1.x, however this is missing documentation

Great! I'll add some documentation in this PR

racevedoo:
@jameskleeh just added documentation

jameskleeh:
Thanks for the contribution!

PR #5960, opened 2021-08-11
Add New Relic Micrometer registry to BOM

MT-Jacobs:
New Relic's Micrometer registry is documented here:

https://micronaut-projects.github.io/micronaut-micrometer/latest/guide/#metricsAndReportersNewRelicTelemetry

But it's not included in the BOM like all of the other registries.


MT-Jacobs:
Here's the original PR that added it - it's only six months old, so that might explain why it's missing from the BOM:

https://github.com/micronaut-projects/micronaut-micrometer/pull/220

jameskleeh:
Thanks!

PR #5984, opened 2021-08-13
Don't use emitter after it's been cancelled

mattmoss:
If the emitter has already been cancelled/completed when the future is notified, just log a message instead of attempting to use the emitter. (At this point, the emitter is already done, likely with an error which the client will receive.)

This PR is for 1.3.x based on the original issue. Will do a separate PR for 2.5.x.


jameskleeh:
We don’t do separate PRs for separate branches. We merge up. Having said
that the PR should only target 2.x because we don’t maintain 1.x

mattmoss:
Okay... I'll close this PR and redo for 2.x.

PR #5993, opened 2021-08-17
feat: show validation index

rafaelrenanpacheco:
1 - Adds the violated index to the error message when using `@Valid` on a collection payload.
2 - Sorts the violations by message to avoid "random" responses.


jameskleeh:
Technically a breaking change so I rebased for 3.0 and created a new PR https://github.com/micronaut-projects/micronaut-core/pull/6004

PR #6049, opened 2021-08-19
Moving `NotNull` annotation in front of FQN.

tkvangorder:
In a "wow, didn't know you could do that in Java" moment, we had a parsing error in OpenRewrite. Apparently, you can embed annotations INSIDE a fully qualified name. I suspect this was just an IDE refactor gone wrong. This PR just moves the annotations outside the FQN for readability.


jameskleeh:
I also just had a moment, but mine was more like "how? why?"

PR #6058, opened 2021-08-24
Delegate bean resolution context when resolving proxy target

dstepanov:
Added a check not to cache proxies in `DefaultBeanContext` `scopedProxies` for `@Any` bean producers, not sure if we should cache anything as it can leak the instance for some short-lived scopes.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
@@ -181,16 +181,16 @@ public final class ConstructorInterceptorChain<T> extends AbstractInterceptorCha
 
         if (beanDefinition instanceof AdvisedBeanType) {
 
-            // intercepted bean constructors include 3 additional arguments in
+            // intercepted bean constructors include 4 additional arguments in
             // addition to the arguments declared in the bean
             // Here we subtract these from the parameters made visible to the interceptor consumer
-            if (originalParameters.length < 3) {
+            if (originalParameters.length < 4) {
                 throw new IllegalStateException("Invalid intercepted bean constructor. This should never happen. Report an issue to the project maintainers.");
             }
             return Arrays.copyOfRange(
                     originalParameters,
                     0,
-                    originalParameters.length - 3
+                    originalParameters.length - 4
             );
         }
         return originalParameters;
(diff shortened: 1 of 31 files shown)


aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
```diff
@@ -181,16 +181,16 @@ public T invoke(T instance, Object... arguments) {
...
             // Here we subtract these from the parameters made visible to the interceptor consumer
-            if (originalParameters.length < 3) {
+            if (originalParameters.length < 4) {
                 throw new IllegalStateException("Invalid intercepted bean constructor. This should never happen. Report an issue to the project maintainers.");
             }
             return Arrays.copyOfRange(
                     originalParameters,
                     0,
-                    originalParameters.length - 3
+                    originalParameters.length - 4
```
jameskleeh: How would this work with old style bean definitions? Couldn't this break with those definitions?

dstepanov:
I have added a way to pass the number of added constructor parameters in proxies that should solve the backward compatibility issue.

Implemented bean resolution context storing and propagating for lazy interceptor target beans.

Also added double-check locking to cache the target in lazy beans: 

```java
    public Object interceptedTarget() {
        ProxiedSomeBeanConsumer var1 = this.$target;
        if (var1 == null) {
            synchronized(this) {
                var1 = this.$target;
                if (var1 == null) {
                    this.$target = (ProxiedSomeBeanConsumer)((DefaultBeanContext)this.$beanLocator).getProxyTargetBean(this.$beanResolutionContext, Argument.of(ProxiedSomeBeanConsumer.class, $ProxiedSomeBeanConsumer$Definition$Intercepted$Definition$Reference.$ANNOTATION_METADATA, new Class[0]), this.$beanQualifier);
                    this.$beanResolutionContext = null;
                }
            }
        }

        return this.$target;
    }
 ```

**Comment by graemerocher** [comment]
@dstepanov tests are failing, those need addressing before review

**Comment by dstepanov** [comment]
It looks like caching target cannot be done by default.
I have added an option `@Around(cacheableLazyTarget)` to preserve the work I did

File: inject/src/main/java/io/micronaut/context/DefaultBeanContext.java [hunk 2]
```diff
@@ -4248,4 +4328,34 @@ public void close() {
             }
         }
     }
+
+    private class DefaultBeanResolutionContext extends AbstractBeanResolutionContext {
```
graemerocher: Rename to `SingletonBeanResolutionContext` since there is already a `DefaultBeanResolutionContext` in `io.micronaut.context`

graemerocher (changes requested):

PR #6067, opened 2021-08-27
Initial work to modernize the Gradle build

melix:
This PR is work in progress and needs to be discussed with the team. I'm creating it now so that we can track progress and start discussion.

Please note that it requires changes to the micronaut-build project which I didn't include in this PR, so it will _not_ build as is.


donbeave:
Do you have a plan to switch to Gradle Kotlin DSL?

graemerocher:
@donbeave no

melix:
This PR is now in a good shape, but depends on https://github.com/micronaut-projects/micronaut-build/pull/187 first.

melix:
@alvarosanchez when refactoring this code, I had to update the bom checker. Do you remember why, for the BOMs which are imported, we check that their dependencies in the dependencyManagement have a group which starts with the group of the BOM? In other words, why do we check, say for org.testcontainers, that the dependencies in dependencyManagement of the test containers BOM, only have dependencies of the org.testcontainers group? Why do we need to check this? (off tipic, sorry for cross posting, wrong tab ;))

graemerocher:
@alvarosanchez answered in another thread but the reason was because some BOMs we imported were not "modules only" boms and included api dependencies in the BOMs which then overrode versions specified in our BOM when imported.

This was the case with the Oracle Cloud BOM. However since then we have moved to generating out own BOM for Oracle Cloud which works around this issue https://github.com/micronaut-projects/micronaut-oracle-cloud/blob/master/oraclecloud-bom/build.gradle

melix:
Thanks! Do you have the GAV coordinates of this BOM which was causing issues? I suppose it's the dreaded "parent POM published as BOM" issue.

graemerocher:
Believe it was this one https://search.maven.org/artifact/com.oracle.oci.sdk/oci-java-sdk-bom/2.4.2/pom and the parent POM defines the conflicting dependencies

ilopmar:
Another thing we need to change is the GitHub action we use to send PRs from modules and update versions in core: https://github.com/micronaut-projects/github-actions/tree/master/update-bom

melix:
> Believe it was this one https://search.maven.org/artifact/com.oracle.oci.sdk/oci-java-sdk-bom/2.4.2/pom and the parent POM defines the conflicting dependencies

Alright so this is exactly the problem: a parent pom defines _dependencies_ and then the BOM inherits from them.

> Another thing we need to change is the GitHub action we use to send PRs from modules and update versions in core: 

Thanks @ilopmar I'll take a look!

melix:
This PR is ready for review. As discussed during the presentation of this work, we're going to publish a version catalog alongside the BOM. While we will probably not advertise this while the Gradle feature is in preview, it's worth discussing naming.

If you look at the [libs.versions.toml](gradle/libs.versions.toml) file, some entries start with `managed-`, for example:

```toml
managed-azure-function = "1.4.2"
```

in the `[versions]` section or

```toml
managed-dekorate-jaeger-annotations = { module = "io.dekorate:jaeger-annotations", version.ref = "managed-dekorate" }
```

for the `[libraries]` section. Those names are not neutral. By convention _in our build_, anything which starts with `managed-` will appear in the BOM. In our build files, using a library such as the above means declaring:

```gradle
implementation libs.managed.dekorate.jaeger.annotations
```

Note how the dashes are converted to dots in the DSL. This isn't neutral, as if you use the Kotlin DSL, **completion is available**, which means that if you start typing `libs.managed.deko<CTRL+SPACE>`, the IDE would complete to `libs.managed.dekorate` and show you the children of that node.

As I mentioned, we're going to _publish_ a version catalog that users can import in their builds. If they do so, we remove all the "managed" parts, so for a user this would become:

```gradle
implementation libs.dekorate.jaeger.annotations
```

Once published, it's a breaking change to rename the alias, so we must take particular care of naming. I would therefore suggest that we review carefully the `libs.versions.toml` file and make sure that all aliases make sense, and that their grouping makes sense. The advantage of dotted notations being that we can group libraries together.

bom-check/build.gradle
```diff
@@ -11,86 +8,9 @@ repositories {
...
-
-            }
-        }
-    }
-    if (pom == null) {
-        errors << "Error validating POM for dependency [${moduleName}]: POM Not Found or not Parseable".toString()
+tasks.withType(io.micronaut.build.internal.pom.PomChecker).configureEach {
+    onlyIf {
+        // We only perform validation on releases
+        !micronautSnapshot || providers.gradleProperty("force.check.bom").isPresent()
```
graemerocher, line 14: Will we be able to catch errors in the BOM before doing a release?

graal/build.gradle
```diff
@@ -1,12 +1,12 @@
 dependencies {
     annotationProcessor project(":inject-java")
     api project(":inject")
-    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
+    implementation libs.jackson.databind
```
graemerocher: jackson databind should be managed, seems like a bug it was not before

bom-check/build.gradle
melix: Yes. We can use the flag `force.check.bom` or we can remove the condition altogether: I just replicated what existed before.

graemerocher (approved):

PR #6070, opened 2021-08-30
Document bootstrap.yml

### LINKED ISSUE #6066: Document bootstrap.yml
There are a few mentions of using bootstrap.yml in the reference docs and some of the module docs, but there's no explanation for why you'd put properties in bootstrap.yml instead of application.yml, whether you should move everything from application.yml to bootstrap.yml, etc.

Also, phrasing like `add the following configuration to your bootstrap.yml file` (https://docs.micronaut.io/latest/guide/#distributedConfigurationVault, https://docs.micronaut.io/latest/guide/#distributedConfigurationSpringCloud) is misleading because it implies that the file already exists.

burtbeckwith:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/6066

### DIFF
--- a/src/main/docs/guide/config/bootstrap.adoc
+++ b/src/main/docs/guide/config/bootstrap.adoc
@@ -1,21 +1,20 @@
 Most application configuration is stored in `application.yml`, environment-specific files like `application-{environment}.{extension}`, environment and system properties, etc.
 These configure the application context.
-But during application startup, before the application context is created, a "bootstrap" context is created as its parent context.
+But during application startup, before the application context is created, a "bootstrap" context can be created to store configuration necessary to retrieve additional configuration for the main context. Typically that additional configuration is in some remote source.
 
-Configuration properties that must be present before application context configuration properties are resolved, for example when using distributed configuration, are stored in a bootstrap configuration file.
-By default this file is named `bootstrap.yml` or `bootstrap.properties` (JSON and Groovy files are not supported) and be in `src/main/resources`.
-The name is configurable with the link:{api}/io/micronaut/context/env/Environment.html#BOOTSTRAP_NAME_PROPERTY[micronaut.bootstrap.name^] system property, and additionally you can have environment-specific bootstrap config files, e.g. `bootstrap-oraclecloud.yml`.
+The bootstrap context is enabled if one of the following conditions are met:
 
-Note however that `bootstrap.yml` must exist for environment-specific bootstrap config files to be loaded.
+- `src/main/resources/bootstrap.yml` is present
+- `src/main/resources/bootstrap.properties` is present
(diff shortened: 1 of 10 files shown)


burtbeckwith:
Should `@BootstrapContextCompatible` be documented here also?

src/main/docs/guide/config/bootstrap.adoc
```diff
@@ -0,0 +1,19 @@
+Most application configuration is stored in `application.yml`, environment-specific files like `application-{environment}.{extension}`, environment and system properties, etc.
+These configure the application context.
+But during application startup, before the application context is created, a "bootstrap" context is created as its parent context.
+
+Configuration properties that must be present before application context configuration properties are resolved, for example when using distributed configuration, are stored in a bootstrap configuration file.
+By default this file is named `bootstrap.yml` or `bootstrap.properties` (JSON and Groovy files are not supported) and be in `src/main/resources`.
```
jameskleeh: That seems more like an oversight (JSON and Groovy not being supported)

src/main/docs/guide/config/bootstrap.adoc
```diff
@@ -6,6 +6,8 @@ Configuration properties that must be present before application context configu
 By default this file is named `bootstrap.yml` or `bootstrap.properties` (JSON and Groovy files are not supported) and be in `src/main/resources`.
 The name is configurable with the link:{api}/io/micronaut/context/env/Environment.html#BOOTSTRAP_NAME_PROPERTY[micronaut.bootstrap.name^] system property, and additionally you can have environment-specific bootstrap config files, e.g. `bootstrap-oraclecloud.yml`.
 
+Note however that `bootstrap.yml` must exist for environment-specific bootstrap config files to be loaded.
```
jameskleeh: This seems like a bug

jameskleeh:
The issues I've pointed out are due to how its determined if the bootstrap environment should be created. JSON and Groovy are supported and its not necessary for bootstrap.yml to exist, given its correctly determined the bootstrap environment should be created. For example if you set the system property `Environment.BOOTSTRAP_CONTEXT_PROPERTY` to true then the above issues go away. Currently it checks for that system property, or if `bootstrap.yml` exists, or if `bootstrap.properties` exists.

I think the most correct solution would be to have a flag in the context builder `bootstrapEnvironment(true)` instead of relying on classpath lookups, however that is a Micronaut 4 change. In the meantime, we could look up every combination of `bootstrap-{env}.{extension}` and that would resolve the issue, however that would lead to more work being done at startup. 

Thoughts @graemerocher ?

jameskleeh:
> Should @BootstrapContextCompatible be documented here also?

I think so

Issue #6062, opened 2021-08-25
Kotlin coroutines reactor lib and StreamingFileUpload causes first request to endpoint to hang


rawilder:
### Expected Behavior

Using StreamingFileUpload and Publisher<StreamingFileUpload> with `org.jetbrains.kotlinx:kotlinx-coroutines-reactor` should run without request hanging.

### Actual Behaviour

The first request hangs, but requests after first do not.

### Steps To Reproduce

Clone example project
`./gradlew clean run`
Request using cURL or other rest request tool

### Environment Information

OS: MacOS
JDK: tested with corretto 11 and openjdk 11

### Example Application

https://github.com/rawilder/mn-multi-file-upload

### Version

3.0.0

rawilder:
For the record, this was not happening on Micronaut 2.x, this was introduced in 3.x as far as my memory serves.

rawilder:
After many debugging sessions, I've narrowed it down to something with reading the bytes inside the map. Reading the bytes inside the map is sometimes causing the NettyPartData to release the byteBuf (`AbstractReferenceCountedByteBuf`). This deallocates the byteBuf setting it to null or causing the refCount to go into a bad state for the following.

`FormDataHttpContentProcessor` calls `postRequestDecoder.offer(httpContent)` which down the line attempts to use byteBuf. There is a weird thing happening where it gets to an if check at `AbstractMemoryHttpData:addContent`. The check sees a non-null byteBuf that doesn't fit any checks so it drops into the else, but then becomes null from the deallocation which ultimately causes the failure and cancellation. The cancellation is for some reason non propagating back to the controller to cancel the request which is causing the hang. 

There is some more legwork to do with regards to why it's not propagating and why the map and `.bytes` is executing in between the if statement check and else statement processing (this is likely inconsistent due to reactive code). 


Update: 

On first request it looks like the PartData publisher is publishing messages or the flow is consuming them in such a way, that two messages go through their onNext process before it's getting consumed by the flow. 

Update 2: 

Okay that last discovery sent me down the road of looking at the consuming part of kotlin's `Publisher.asFlow`. My best guess right now is the flow consumer is completely separated from the publisher in such a way that the publisher can get into buffered amounts of data. Since getBytes() is a destructive operations, this is probably not ideal. 

[...]

PR #6074, opened 2021-08-31
the netty unicast processor is not emitting errors for downstream subscribers

rawilder:
related to: https://github.com/micronaut-projects/micronaut-core/issues/6062


graemerocher:
Needs a test

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -827,16 +827,17 @@ protected void doOnNext(Object message) {
...
                 protected void doOnError(Throwable t) {
                     s.cancel();
+                    subjects.values().forEach(subject -> subject.onError(t));
                     emitter.error(t);
                 }
 
                 @Override
                 protected void doOnComplete() {
-                    for (UnicastProcessor subject : subjects.values()) {
+                    subjects.values().forEach(subject -> {
```
jameskleeh: Don't change to functional style

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -827,16 +827,17 @@ protected void doOnNext(Object message) {
                 @Override
                 protected void doOnError(Throwable t) {
                     s.cancel();
+                    subjects.values().forEach(subject -> subject.onError(t));
```
jameskleeh: A simple for loop as is below is preferred here

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
rawilder: Sorry about that, force of habit

rawilder:
This will sound crazy but I can't replicate the kotlin failure via a test (see here: https://github.com/rawilder/mn-multi-file-upload/pull/1/files).

It's only happening via server start -> manual request. Which means I'd have to inject a failure somehow into the onNext function of `RoutingInBoundHandler:buildSubscriber` which I really am not sure how to do. I was trying to find some component I could mock and make throw but nothing stood out as obvious or not breaking other bits. Any ideas?

jameskleeh:
@rawilder Try moving the test which should fail into its own class. Because tests are executing before the new tests they are not the first request to the server which is the case where we saw the failure

rawilder:
@jameskleeh I moved the two tests to their own class locally and both succeeded. Would you like me to still push that to separate them out?

PR #6077, opened 2021-09-01
refactor: extract static method for shouldDeduceEnvironment

sdelamo:
- test disable environment deduction via environment variable

This is related to https://github.com/micronaut-projects/micronaut-aws/issues/1171


jameskleeh:
I don't think we should expose this statically. If the goal is to test that environment deduction is turned off you can just create a context in a test and assert the test environment is not present, then turn it back on and assert it is present

sdelamo:
The goal is to have a static method which can be called from other module. See: https://github.com/micronaut-projects/micronaut-aws/issues/1171
 
 Maybe we could extract the logic to  a `EnvironmentDeductionsUtils` class

PR #6078, opened 2021-09-01
Fix names of inner annotations in metadata

### LINKED ISSUE #6045: Nested annotations handling
### Expected Behavior

It's expected that ApplicationContext finds a bean annotated with nested annotation:

```
class ExtractorFactoriesTest {

    @Inherited
    @Documented
    @Retention(RetentionPolicy.RUNTIME)
    public @interface TestExtractor {
    }

    @TestExtractor
    @Singleton
    public static class TestExtractorFactory implements ExtractorFactory {
    }
}
```

This code should return `TestExtractorFactory` bean.
```
ctx.findBean(ExtractorFactory.class, new FactoryQualifier<>(TestExtractor))
```

### Actual Behaviour

ApplicationContext is unable to find a bean as deep down `annotationMetadata.declaredAnnotations` map has a key `..ExtractorFactoriesTest.TestExtractor` instead of `..ExtractorFactoriesTest$TestExtractor`.

This happens with Micronaut 3.0.0 while 3.0.0-RC1 works as expected.

### Steps To Reproduce

_No response_

### Environment Information

macOS 10.15.7
JDK 1.8.0_301

### Example Application

_No response_

### Version

3.0.0

graemerocher:
Annotation names are currently stored using the canonical name of the type instead of the binary name. This makes them impossible to retrieve if they are inner classes.

This PR changes Micronaut so annotation names are stored using the binary class name so that lookup with `Outer.Inner.class.getName()` works as expected.

Fixes #6045

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/annotation/AnnotationMetadataWriterSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/annotation/AnnotationMetadataWriterSpec.groovy
@@ -35,6 +35,28 @@ import java.lang.annotation.Retention
  */
 class AnnotationMetadataWriterSpec extends AbstractBeanDefinitionSpec {
 
+    void "test inner annotations in metadata"() {
+        given:
+        def annotationMetadata = buildTypeAnnotationMetadata("inneranntest.Test","""
+package inneranntest;
+
+import java.lang.annotation.Retention;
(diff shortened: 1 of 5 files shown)


graemerocher:
This is actually a regression from 2.5.x as it does work there and we appear to have been storing annotation names with the binary name in 2.5.x

dstepanov (approved):

PR #6079, opened 2021-09-01
Extract Jackson dependencies to their own modules 

yawkat:
This PR moves most dependencies on jackson-core and jackson-databind into their own modules. Three new modules are introduced:

- **json-core** contains basic json support features, without a dependency on jackson. Central to this is a `JsonCodec` interface that exposes similar API as jacksons `ObjectMapper`. There is also a new `JsonNode` API in this module, replacing the `JsonNode` that is part of jackson-databind.
- **jackson-core** contains functionality that only depends on jackson-core, not on jackson-databind. This includes the reactive json -> `JsonNode` parser and other classes related to translating `JsonNode`s to jackson streams.
- **jackson-databind** contains functionality that depends on jackson-databind. This includes most of the jackson-related classes that were previously in the runtime module, and the jackson-databind implementation of `JsonCodec`.

There are still some decisions I'm not sure about. e.g. `JsonNode` is an overloaded name (open to alternative suggestions), and `JsonCodec` is a fairly big interface as it stands right now.

Since this PR is supposed to go into a minor version, there can't be any incompatibilities with existing code. All tests in micronaut-core pass unchanged, and there are no changes to method or class signatures in public, non-internal classes from what I can tell, but a second look would be appreciated.

Right now this PR has 3.0.x as the base branch, but it will be retargeted for 3.1.


jackson-core/src/main/java/io/micronaut/jackson/core/tree/JsonStreamTransfer.java
```diff
@@ -0,0 +1,124 @@
...
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.json.JsonConfig;
+
+import java.io.IOException;
+
+@Internal
+public final class JsonStreamTransfer {
```
graemerocher, line 32: Needs javadoc author/since tags

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+import com.fasterxml.jackson.core.*;
+import io.micronaut.json.JsonConfig;
+import io.micronaut.json.tree.JsonNode;
+
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.*;
+
+public final class MicronautTreeCodec {
```
graemerocher: Should this be annotated internal? Needs javadoc author/since tags

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+    private static final MicronautTreeCodec INSTANCE = new MicronautTreeCodec(JsonConfig.DEFAULT);
+
+    private final JsonConfig config;
+
+    private MicronautTreeCodec(JsonConfig config) {
+        this.config = config;
+    }
+
+    public static MicronautTreeCodec getInstance() {
```
graemerocher: Needs javadoc if public API

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+        return INSTANCE;
+    }
+
+    public MicronautTreeCodec withConfig(JsonConfig config) {
```
graemerocher: Needs javadoc if public API

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+        return new MicronautTreeCodec(config);
+    }
+
+    public JsonNode readTree(JsonParser p) throws IOException {
```
graemerocher: Needs javadoc if public API

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+            case VALUE_FALSE:
+                return JsonNode.createBooleanNode(false);
+            case VALUE_NULL:
+                return JsonNode.nullNode();
+            default:
+                throw new UnsupportedOperationException("Unsupported token: " + p.currentToken());
+        }
+    }
+
+    public void writeTree(JsonGenerator generator, JsonNode tree) throws IOException, JsonProcessingException {
```
graemerocher: Needs javadoc if public API

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+        if (tree.isObject()) {
+            generator.writeStartObject();
+            for (Map.Entry<String, JsonNode> entry : tree.entries()) {
+                generator.writeFieldName(entry.getKey());
```
graemerocher: Only the `key` is needed, can `keySet()` be used?

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+                writeTree(generator, tree);
```
graemerocher: This recursion seems unusual, what breaks this loop?

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+                generator.writeNumber(value.floatValue());
+            } else if (value instanceof Integer) {
+                generator.writeNumber(value.intValue());
+            } else if (value instanceof Byte || value instanceof Short) {
+                generator.writeNumber(value.shortValue());
+            } else if (value instanceof Long) {
+                generator.writeNumber(value.longValue());
+            } else if (value instanceof BigInteger) {
+                generator.writeNumber((BigInteger) value);
+            } else {
```
graemerocher: order `instanceof` checks by probability. `Integer` and `Long` are the most commonly used should be at the top

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+                throw new IllegalStateException("Unknown number type " + value.getClass().getName());
+            }
+        }
+    }
+
+    public JsonParser treeAsTokens(JsonNode node) {
```
graemerocher: javadoc

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,134 @@
...
+        return new TraversingParser(node);
+    }
+
+    public TreeGenerator createTreeGenerator() {
```
graemerocher: javadoc

jackson-core/src/main/java/io/micronaut/jackson/core/tree/TraversingParser.java
```diff
@@ -0,0 +1,507 @@
...
+
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.Map;
+
+class TraversingParser extends ParserMinimalBase {
```
graemerocher: Can this be `final` and annotated with `@Internal`?

json-core/src/main/java/io/micronaut/json/codec/JsonCodecMediaTypeCodec.java
```diff
@@ -47,26 +44,27 @@
  * @author svishnyakov
  * @since 1.3.0
  */
-public abstract class JacksonMediaTypeCodec implements MediaTypeCodec {
+public abstract class JsonCodecMediaTypeCodec implements MediaTypeCodec {
```
graemerocher: The duplication in the name makes this confusion. Probably should be just called `JsonMediaTypeCodec` or `JsonNodeMediaTypeCodec`

json-core/src/main/java/io/micronaut/json/JsonCodec.java
```diff
@@ -0,0 +1,94 @@
...
+import org.reactivestreams.Processor;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+public interface JsonCodec {
```
graemerocher: I think naming this `JsonCodec` is just going to confuse matters with our existing codec API so maybe a different name is needed here like `JsonMapper` or whatever.

Eitherway this is new public API and needs javadoc for the type and all public methods that are not marked with `@Internal`

Each method parameter should also be annotated with `@Nullable` or `@NotNull` to express nullability which helps kotlin users  as well

json-core/src/main/java/io/micronaut/json/JsonCodec.java
```diff
@@ -0,0 +1,94 @@
...
+    <T> T readValueFromTree(JsonNode tree, Argument<T> type) throws IOException;
+
+    default <T> T readValueFromTree(JsonNode tree, Class<T> type) throws IOException {
+        return readValueFromTree(tree, Argument.of(type));
+    }
+
+    JsonNode writeValueToTree(Object value) throws IOException;
+
+    @Internal
+    default void updateValueFromTree(Object value, JsonNode tree) throws IOException {
```
graemerocher: Perhaps create a separate internal interface that that this interface extends from rather than having all the methods in the public API. Or even better another interface that we cast down to in the cases where these methods are needed. Mixing internal and public methods in the same API should be avoided if possible

json-core/src/main/java/io/micronaut/json/bind/JsonBeanPropertyBinder.java
```diff
@@ -0,0 +1,284 @@
...
+
+/**
+ * An {@link io.micronaut.core.bind.ArgumentBinder} capable of binding from an object from a map.
+ *
+ * @author Graeme Rocher
+ * @since 1.0
+ */
+@Singleton
+@Primary
+class JsonBeanPropertyBinder implements BeanPropertyBinder {
```
graemerocher: Can this be declared final?

jackson-databind/src/main/java/io/micronaut/jackson/databind/convert/JacksonConverterRegistrar.java
```diff
@@ -13,15 +13,11 @@
...
+package io.micronaut.jackson.databind.convert;
 
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.PropertyNamingStrategies;
-import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.*;
```
graemerocher: Avoid the use of star imports please

json-core/src/main/java/io/micronaut/json/JsonConfig.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json;
+
+public final class JsonConfig {
```
graemerocher: How is this configured? Does it make sense to make it `@ConfigurationProperties` with defaults instead of static state?

graemerocher: Alao needs javadoc if it is public API

graemerocher (changes requested):

graemerocher:
This PR should target the 3.1.x branch

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
yawkat: This should have the same access as `JsonNode`. Right now, `JsonNode` is internal, but I forgot to make this internal as well. It might be better to make JsonNode public API though – it appears quite a bit in other APIs as well. Maybe it could be experimental?

json-core/src/main/java/io/micronaut/json/codec/JsonCodecMediaTypeCodec.java
yawkat: JsonMediaTypeCodec is already used by a subclass, and JsonNodeMediaTypeCodec isn't right because it's not only for json nodes. I will try to think of a better name

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
yawkat: Yes, this is a bug from copy-pasting. It's concerning that no tests caught this, seems like they didn't transfer over. I will need to check the test coverage of the PR.

json-core/src/main/java/io/micronaut/json/JsonCodec.java
yawkat: going the cast route, that would require all implementors to implement the second interface. is there a way to mark the JsonCodec so that extension by users is unsupported?

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
graemerocher: I would say mark everything public as `@Experimental` for the moment until we are satisfied with it

jackson-core/src/main/java/io/micronaut/jackson/core/tree/TraversingParser.java
yawkat: made it final, but it's package-private so it should be fine without `@Internal`, right?

json-core/src/main/java/io/micronaut/json/codec/JsonCodecMediaTypeCodec.java
yawkat: Renamed `JsonCodec` to `JsonMapper`, so I renamed this class to `MapperMediaTypeCodec`.

json-core/src/main/java/io/micronaut/json/JsonCodec.java
yawkat: There are a few different methods here. `updateValueFromTree` is supported by jackson, but I want to avoid implementing this with the generated serializers, so the method may be removed at some point if I can replace all the uses. I've marked it experimental, instead.

I've removed `@Internal` from some of the other APIs.

I will move `newConversionError` to its own interface. It has only a single use anyway.

json-core/src/main/java/io/micronaut/json/JsonConfig.java
yawkat: I've renamed this to `JsonStreamConfig` and added docs.

Right now, this config is constructed from the `ObjectMapper`. That means that it comes indirectly from the `JacksonConfiguration`. However, in some cases (e.g. when jackson config is changed through `@JacksonConfiguration`), the settings of the object mapper change and so the properties of this class also change. I don't think it should be made `@ConfigurationProperties` for that reason.

jackson-databind/src/main/java/io/micronaut/jackson/databind/convert/JacksonConverterRegistrar.java
graemerocher: wasn't resolved? If you are using intellij you change this in your intellij settings

json-core/src/main/java/io/micronaut/json/bind/JsonBeanPropertyBinder.java
```diff
@@ -0,0 +1,308 @@
...
+    private final JsonMapper jsonMapper;
+    private final int arraySizeThreshhold;
+    private final Iterable<JsonBeanPropertyBinderExceptionHandler> exceptionHandlers;
+
+    /**
+     * @param jsonMapper    To read/write JSON
+     * @param configuration The configuration for Jackson JSON parser
+     * @param locator       Locator for exception handlers.
+     */
+    JsonBeanPropertyBinder(JsonMapper jsonMapper, JsonConfiguration configuration, BeanLocator locator) {
```
graemerocher: The `BeanLocator` is not necessary here, you can just inject `Collection<JsonBeanPropertyBinderExceptionHandler>`

graemerocher: I would also say it is best to avoid initializing these beans until an exception actually occurs so it is probably better to inject `BeanProvider<JsonBeanPropertyBinderExceptionHandler>` which allows iteration

json-core/src/main/java/io/micronaut/json/bind/JsonBeanPropertyBinder.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private static class ArrayBuilder implements ValueBuilder {
+        final List<ValueBuilder> values = new ArrayList<>();
+
+        @Override
+        public JsonNode build() {
+            return JsonNode.createArrayNode(values.stream().map(ValueBuilder::build).collect(Collectors.toList()));
+        }
+    }
```
graemerocher: This inner classes can be made final

json-core/src/main/java/io/micronaut/json/bind/JsonBeanPropertyBinder.java
```diff
@@ -0,0 +1,308 @@
...
+            return value;
+        }
+    }
+
+    private static class ObjectBuilder implements ValueBuilder {
+        final Map<String, ValueBuilder> values = new LinkedHashMap<>();
+
+        @Override
+        public JsonNode build() {
+            Map<String, JsonNode> built = new LinkedHashMap<>();
```
graemerocher: pass the size() of the `values` map to the `built` map to avoid map resizing/extra allocation

yawkat:
Okay, I've done the japicmp checks. Two things:

- `JacksonMediaTypeCodec` has a method `cloneWithFeatures` that takes a `JacksonFeatures` parameter. `JacksonFeatures` is internal, so I had moved it, but this broke compat for `JacksonMediaTypeCodec`. I've moved `JacksonFeatures` back to its old package to preserve compatibility with the xml module.
- Some of the methods from `JacksonMediaTypeCodec` were moved to its new superclass. This gives warnings (not errors) in japicmp, but it should be fine according to the JLS: "Using their scheme, here is a list of some important binary compatible changes that the Java programming language supports: [...] Moving a method upward in the class hierarchy."

Other than that, there were no issues.

I've also removed `@Deprecated` from some classes where the replacements are still experimental.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -2552,10 +2549,11 @@ private void traceHeaders(HttpHeaders headers) {
     }
 
     private static MediaTypeCodecRegistry createDefaultMediaTypeRegistry() {
-        ObjectMapper objectMapper = new ObjectMapperFactory().objectMapper(null, null);
-        ApplicationConfiguration applicationConfiguration = new ApplicationConfiguration();
+        JsonMapper mapper = new JacksonDatabindMapper();
```
graemerocher, line 2556: does this need decoupling into a service loader?

jackson-core/src/main/java/io/micronaut/jackson/core/tree/MicronautTreeCodec.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.Map;
+
+/**
+ * Codec for transforming {@link JsonNode} from and to json streams.
+ *
+ * @author Jonas Konrad
+ * @since 3.1
+ */
+@Experimental
+public final class MicronautTreeCodec {
```
graemerocher: Rename to `JsonNodeTreeCodec`

jackson-core/src/main/java/io/micronaut/jackson/core/tree/TraversingParser.java
```diff
@@ -0,0 +1,514 @@
...
+
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.Map;
+
+final class TraversingParser extends ParserMinimalBase {
```
graemerocher: Rename to `JsonNodeTraversingParser` and add javadoc with `@since`

runtime/src/main/java/io/micronaut/discovery/cloud/ComputeInstanceMetadataResolverUtils.java
```diff
@@ -64,35 +67,26 @@ public static JsonNode readMetadataUrl(URL url, int connectionTimeoutMs, int rea
...
-
     /**
      * Populates the instance instance metadata's {@link AbstractComputeInstanceMetadata#setMetadata(Map)} property.
+     *
      * @param instanceMetadata The instance metadata
-     * @param metadata A map of metadata
+     * @param metadata         A json object of metadata
      */
-    public static void populateMetadata(AbstractComputeInstanceMetadata instanceMetadata, Map<?, ?> metadata) {
+    public static void populateMetadata(AbstractComputeInstanceMetadata instanceMetadata, JsonNode metadata) {
```
graemerocher, line 105: These methods seem to be used in modules and changing them will break compatibility. See:

https://github.com/micronaut-projects/micronaut-oracle-cloud/blob/ee5fe9110797e473c044ae20c36ee8987e2929bf/oraclecloud-common/src/main/java/io/micronaut/discovery/cloud/oraclecloud/OracleCloudMetadataResolver.java#L42-L43

json-core/src/main/java/io/micronaut/json/tree/JsonString.java
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json.tree;
+
+import io.micronaut.core.annotation.NonNull;
+
+class JsonString extends JsonScalar {
```
graemerocher: Add `@Internal` and javadoc with `since` / `author`

json-core/src/main/java/io/micronaut/json/tree/JsonScalar.java
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json.tree;
+
+import io.micronaut.core.annotation.NonNull;
+
+import java.util.Map;
+
+abstract class JsonScalar extends JsonNode {
```
graemerocher, line 28: Add `@Internal` and javadoc with `since` / `author`

json-core/src/main/java/io/micronaut/json/tree/JsonNumber.java
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json.tree;
+
+import io.micronaut.core.annotation.NonNull;
+
+class JsonNumber extends JsonScalar {
```
graemerocher: Add `@Internal` and javadoc with `since` / `author`

json-core/src/main/java/io/micronaut/json/tree/JsonNull.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json.tree;
+
+import io.micronaut.core.annotation.NonNull;
+
+final class JsonNull extends JsonScalar {
```
graemerocher, line 26: Add `@Internal` and javadoc with `since` / `author`

json-core/src/main/java/io/micronaut/json/tree/JsonContainer.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json.tree;
+
+abstract class JsonContainer extends JsonNode {
```
graemerocher, line 24: Add `@Internal` and javadoc with `since` / `author`

json-core/src/main/java/io/micronaut/json/tree/JsonBoolean.java
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.json.tree;
+
+import io.micronaut.core.annotation.NonNull;
+
+final class JsonBoolean extends JsonScalar {
```
graemerocher, line 26: Add `@Internal` and javadoc with `since` / `author`

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: eventually yes. however, the SPI implementation of the generator-based mapper is still a bit weird (constructs an entire app context right now, to locate serializers), so i think for now it is best to just keep using the jackson module directly here. there is another case of this, `DigitalOceanMetadataResolver`, where the old constructor takes an `ObjectMapper` and a `JacksonDatabindMapper` is constructed from it. That case can't be fixed without removing the constructor, so I feel like this one isn't too bad either.

runtime/src/main/java/io/micronaut/discovery/cloud/ComputeInstanceMetadataResolverUtils.java
yawkat: ok i will readd the old methods for now and mark the new ones as experimental. once the api has settled, we can migrate depending modules to the new methods.

yawkat:
Can I somehow prevent DI from injecting the old `JsonMediaTypeCodec` when injecting a `List<MediaTypeCodec>`? I have to keep the old classes around for compatibility, but I don't want them to be used unless they are explicitly asked for.

graemerocher:
Only way is to put some kind of the qualifier on the "good" beans and leave it off JsonMediaTypeCodec but I am not sure if that is a great solution.

yawkat:
@graemerocher dedbcfd9b0ee53492afd1e295266e09ae97a425b seems to work, it just uses the explicit `@Bean(typed = ...)` – or does that have any adverse impact?

graemerocher:
@jameskleeh can you look through this and provide feedback?

yawkat:
this will need changes for #6067 but they shouldn't affect the review. I'll do them tomorrow

graemerocher:
@yawkat can you resolve the conflicts again

yawkat:
looks like spurious failure (`Address already in use`)

graemerocher:
Yeah test failure is a unrelated race condition in port binding. I pushed a fix for it https://github.com/micronaut-projects/micronaut-core/commit/77a426dff1286ea188de69322a540688c54fbfe1

yawkat:
thanks!

jameskleeh (approved):

graemerocher (approved):

PR #6093, opened 2021-09-03
feat: allow SameSite case insensitive binding

### LINKED ISSUE #6092: Binding to `SameSite` via configuration is case sensitive
### Expected Behavior

Binding to `SameSite` via configuration is case sensitive

### Actual Behaviour

It is case insensitive. 

Moreover, IntelliJ IDEA autocompletes values for `SameSite` as lowercase. 

e.g. 

`micronaut.security.token.jwt.cookie.cookie-same-site=strict` 

but that it is not bound. 

`micronaut.security.token.jwt.cookie.cookie-same-site=Strict` works though.  

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.0.0

sdelamo:
Close #6092

enum `SameSite` is:

```java
package io.micronaut.http.cookie;

public enum SameSite {
  Lax,
  Strict,
  None;
}
```

I think it should have been something like:

```java
package io.micronaut.http.cookie;

import com.fasterxml.jackson.annotation.JsonValue;

public enum SameSite {
    LAX("Lax"),
    STRICT("Strict"),
    NONE("None");

    private final String value;
    SameSite(String value) {
        this.value = value;
    }

    @JsonValue
    public String toString() {
        return this.value;
    }
}
```

However, changing the enum is a breaking change. Henche, I have introduced a type converter to allow for lowercase binding.

IntelliJ IDEA suggests lowercase values.

Users type a lowercase value as their IDE suggests and they are confused because it does not work.

See: https://github.com/micronaut-projects/micronaut-security/issues/755

Also, test SameSite JSON rendering


jameskleeh:
I think this can go into 3.0.x given that the current case is a failure and thus this is a bug fix

sdelamo:
@jameskleeh  I have created https://github.com/micronaut-projects/micronaut-core/pull/6101 which has 3.0.x as base

PR #6094, opened 2021-09-03
Add support for StreamingFileUpload.transferTo(OutputStream)

### LINKED ISSUE #6084: Allow writing a StreamingFileUpload to an OutputStream instead of just a File
### Feature description

Currently the the StreamingFileUpload interface has methods

```java
Publisher<Boolean> transferTo(File destination);
Publisher<Boolean> transferTo(String location);
```

In our use case of writing the stream file to a MongoDB GridFS OutputStream it would be helpful to have a method
```
Publisher<Boolean> transferTo(OutputStream out);
```

This would also be presumably would be helpful for other use case like streaming to Amazon S3.

jameskleeh:
Closes #6084


src/main/docs/guide/httpServer/uploads.adoc
```diff
@@ -24,13 +24,23 @@ Route arguments of type rs:Publisher[Publisher<PartData>] are treated as intende
...
-snippet::io.micronaut.docs.server.upload.UploadController[tags="class", indent=0, title="Streaming file upload"]
+snippet::io.micronaut.docs.server.upload.UploadController[tags="class,file", indent=0, title="Streaming file upload"]
 
 <1> The method consumes api:http.MediaType#MULTIPART_FORM_DATA[]
 <2> The method parameters match form attribute names. In this case `file` will match for example an `<input type="file" name="file">`
-<3> The api:http.multipart.StreamingFileUpload.transferTo(java.lang.String)[] method transfers the file to the server. The method returns a rs:Publisher[]
+<3> The api:http.multipart.StreamingFileUpload.transferTo(File)[] method transfers the file to the server. The method returns a rs:Publisher[]
 <4> The returned reactor:Mono[] subscribes to the rs:Publisher[] and outputs a response once the upload is complete, without blocking.
 
+It is also possible to pass an output stream with the `transferTo` method.
```
graemerocher, line 34: Clarify that the actual transfer will happen on the I/O thread pool

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyStreamingFileUpload.java
```diff
@@ -212,4 +159,67 @@ public void subscribe(Subscriber<? super PartData> s) {
     public void discard() {
         fileUpload.release();
     }
+
+    private Publisher<Boolean> transferTo(ThrowingSupplier<OutputStream, IOException> outputStreamSupplier) {
+        return Mono.<Boolean>create(emitter ->
+
+                subject.subscribeOn(Schedulers.fromExecutorService(ioExecutor))
```
graemerocher, line 162: Seems this should be `publishOn` as well like https://github.com/micronaut-projects/micronaut-core/pull/6120/files#diff-5e0dd1237da04622a601621ff253c396e94d189c262f0f9975c89540878ba5bbR98 ?

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyStreamingFileUpload.java
jameskleeh: subscribeOn should be sufficient as it applies from the source and down the chain of operators. I verified the thread is indeed IO when the write occurs

PR #6097, opened 2021-09-04
Fix for custom validators on nested pojos

tbodor:
Fix for issue #5986 (validator not handling nested custom validations). Note that the example linked from that issue is also missing the @Valid annotation on the list of nested POJOs.

PR #6107, opened 2021-09-06
Fix custom validator with non object class during cascade from a list

tchiotludo:
When using a custom validation at class Level and a type different than Object, the validator is never triggered.

Example : 
```java 
@Retention(RetentionPolicy.RUNTIME)
@Constraint(validatedBy = { })
public @interface TaskDefault {
    String message() default "invalid taskDefaults expression ({validatedValue})";
}

@Value
@Introspected
public class Flow {
    @Valid
    private List<TaskDefault> taskDefaults;
}

@Value
@io.kestra.core.validations.TaskDefault
@Introspected
public class TaskDefault {
    private final String type;
}

ConstraintValidator<TaskDefault, io.kestra.core.models.flows.TaskDefault> taskDefaultValidator() {
    return (value, annotationMetadata, context) -> 
        return true;
    };
}
```
`taskDefaultValidator` will be never triggered when I do `modelValidator.isValid(flow) `

As I understand, the bug is here : 
https://github.com/micronaut-projects/micronaut-core/blob/764eb376a1213c436d7dec75cdeecc53cfeea596/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java#L1123-L1134

it's the 6 args `rootBeanClass` that is passed and must be the current object type `introspection.getBeanType()` in order to trigger the validation.

Tell what do you think ? or if it's a mistake from my side on Bean Validation specs

### DIFF
--- a/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
+++ b/validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
@@ -1112,7 +1112,7 @@ public class DefaultValidator implements Validator, ExecutableMethodValidator, R
                     null,
                     context,
                     overallViolations,
-                    rootBeanClass,
+                    introspection.getBeanType(),
                     object,
                     pojoConstraint,
                     introspection.getAnnotation(pojoConstraint));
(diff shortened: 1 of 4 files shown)


jameskleeh:
Is this the same fix as https://github.com/micronaut-projects/micronaut-core/pull/6097?

tchiotludo:
yes it's the same, I didn't see the other PR

PR #6120, opened 2021-09-08
Don't block the event loop when reading inputstream. 

### LINKED ISSUE #6100: @Body InputStream Blocks nioEventloop when its not read
### Expected Behavior

When using large payloads say approx 5 mb, Expected to be body read to happen on IO threadpool. Tried Both micronaut 2.5.5 and 2.5.12

application.yaml
```
micronaut:
  server:
    port: 9090
    thread-selection: io
```

Method Signature
```
@Post(uris = {
            "/hello"}
    )
    @Produces({MediaType.APPLICATION_JSON})
    @Consumes({MediaType.ALL})
    @ExecuteOn(TaskExecutors.IO)  public MutableHttpResponse<String> stream(HttpRequest httpRequest, @Body @Nullable InputStream payload) throws IOException {
....
}
``` 

### Actual Behaviour

When using large payloads say approx 5 mb, body read Happen on NIO thread pool  blocking the event loop.  This cause stuck threads in our application.

application.yaml
```
micronaut:
  server:
    port: 9090
    thread-selection: io
```

Method Signature
```
@Post(uris = {
            "/hello"}
    )
    @Produces({MediaType.APPLICATION_JSON})
    @Consumes({MediaType.ALL})
    @ExecuteOn(TaskExecutors.IO)  public MutableHttpResponse<String> stream(HttpRequest httpRequest, @Body @Nullable InputStream payload) throws IOException {
....
}
``` 

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

2.5.5, 2.5.12

graemerocher:
Fixes #6100

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/InputStreamBodyBinder.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/InputStreamBodyBinder.java
@@ -89,65 +89,84 @@ public class InputStreamBodyBinder implements NonBlockingBodyArgumentBinder<Inpu
             NettyHttpRequest nettyHttpRequest = (NettyHttpRequest) source;
             io.netty.handler.codec.http.HttpRequest nativeRequest = nettyHttpRequest.getNativeRequest();
             if (nativeRequest instanceof StreamedHttpRequest) {
-                HttpContentProcessor<ByteBufHolder> processor = (HttpContentProcessor<ByteBufHolder>) processorResolver.resolve(nettyHttpRequest, context.getArgument());
                 PipedOutputStream outputStream = new PipedOutputStream();
                 try {
-                    PipedInputStream inputStream = new PipedInputStream(outputStream);
-                    Flux.from(processor)
-                        .onBackpressureBuffer()
-                            .publishOn(ioExecutor)
-                            .subscribe(new CompletionAwareSubscriber<ByteBufHolder>() {
 
-                                @Override
-                                protected void doOnSubscribe(Subscription subscription) {
-                                    subscription.request(1);
-                                }
+                    PipedInputStream inputStream = new PipedInputStream(outputStream) {
+                        private volatile HttpContentProcessor<ByteBufHolder> processor;
+                        @Override
+                        public synchronized int read() throws IOException {
+                            init();
+                            return super.read();
+                        }
 
-                                @Override
-                                protected synchronized void doOnNext(ByteBufHolder message) {
-                                    if (LOG.isTraceEnabled()) {
-                                        LOG.trace("Server received streaming message for argument [{}]: {}", context.getArgument(), message);
-                                    }
-                                    ByteBuf content = message.content();
-                                    if (!(content instanceof EmptyByteBuf)) {
-                                        try {
-                                            byte[] bytes = ByteBufUtil.getBytes(content);
-                                            outputStream.write(bytes, 0, bytes.length);
-                                        } catch (IOException e) {
-                                            subscription.cancel();
-                                            return;
-                                        } finally {
-                                            content.release();
-                                        }
-                                    }
-                                    subscription.request(1);
-                                }
+                        private void init() {
+                            if (processor == null) {
+                                processor = (HttpContentProcessor<ByteBufHolder>) processorResolver.resolve(nettyHttpRequest, context.getArgument());
+                                Flux.from(processor)
+                                        .publishOn(ioExecutor)
+                                        .subscribe(new CompletionAwareSubscriber<ByteBufHolder>() {
 
-                                @Override
-                                protected synchronized void doOnError(Throwable t) {
-                                    if (LOG.isTraceEnabled()) {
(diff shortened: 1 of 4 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/InputStreamBodyBinder.java
```diff
@@ -77,59 +93,61 @@ public InputStreamBodyBinder(HttpContentProcessorResolver processorResolver) {
                 PipedOutputStream outputStream = new PipedOutputStream();
                 try {
                     PipedInputStream inputStream = new PipedInputStream(outputStream);
-                    processor.subscribe(new CompletionAwareSubscriber<ByteBufHolder>() {
+                    Flux.from(processor)
+                        .onBackpressureBuffer()
```
jameskleeh: Wouldn't this lead to OOM?

graemerocher: InputStream is not thread safe, so I don't see an alternative

jameskleeh:
Basically my thoughts are that if the stream is not read, etc.. we shouldn't buffer the data as that will lead to OOM exceptions. It's my understanding that in order to not buffer the event loop thread has to wait for the consumer to read the data.

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/InputStreamBodyBinder.java
jameskleeh: Then I don't think this is an improvement

graemerocher:
Feel free to amend the PR if you find a solution that is both thread safe and doesn't need buffering. Right now the current implementation is blocking the event loop

jameskleeh:
It's only blocking the event loop if the data isn't read. If the data isn't read with this change it could lead to OOM. I see this as a documentation issue

graemerocher:
Also note that we are already buffering the data in `FlowControlHandler` so I don't buy the OOM argument, if you can provide a test where you can trigger an OOM then amend the test to demonstrate that.

graemerocher:
@jameskleeh that is not true, the test reads the data and it is blocking the event loop

jameskleeh:
Yes it is read after it requires a thread from the event loop. If it is read before that it would not be an issue

graemerocher:
Set a break point on the code that writes to the output stream and you will note that all writes to the output stream (a blocking API) occur on the event loop thread without my change. After my change they occur on the I/O thread.

If that isn't blocking the event loop I don't know what is 🤷‍♂️

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/InputStreamBodyBinder.java
graemerocher: Provide a reproducer for your OOM theory in that case. "I don't think" is not a valid argument opposing a bug fix

jameskleeh: I will try to create the scenario. In the meantime, relax with your tone

graemerocher:
> Basically my thoughts are that if the stream is not read, etc.. we shouldn't buffer the data as that will lead to OOM exceptions. It's my understanding that in order to not buffer the event loop thread has to wait for the consumer to read the data.

I'm confused by what you are saying here, looking at the code prior to my change, the `InputStream` is bound and immediately on the same thread the code subscribes to the processor regardless whether any data is actually read from the `InputStream`:

https://github.com/micronaut-projects/micronaut-core/blob/8da7f725d6b750d9a48dc3b0e2a5678763a1040b/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/InputStreamBodyBinder.java#L79-L80

At this point the `PipedOutputStream` will be buffering the data anyway so I have no idea what you are referring to with regards to buffering being a problem with the change, since buffering is already happening within the `PipedOutputStream`

jameskleeh:
When you write to a PipedOutputStream, it blocks the current thread to wait for the reader to read the data. See the `awaitSpace` method in PipedInputStream

graemerocher:
Well exactly, so why are we writing to it on the event loop thread?

jameskleeh:
Perhaps I'm misunderstanding something here so please tell me if so

If we should not buffer data then the event loop thread has to wait for the data to be read before requesting more. If the server is continuously reading data from the client and the data isn't being read then it has to go somewhere (buffered), or block.

graemerocher:
Indeed you have to block somewhere, but that somewhere should not be the event loop which is what is currently being blocked.

jameskleeh:
I was able to reproduce the OOM in my example app before the latest changes

The request is hung with the latest changes (JDK 11). Could be some daemon/cache/etc issue on my machine

https://github.com/jameskleeh/delay-read-stream

Changing version back to 3.0.1 makes it work again

PR #6122, opened 2021-09-08
Correctly handle absence of request in RequestScope

### LINKED ISSUE #6096: Micronaut 3.0.0: @RequestScope throws IllegalMonitorStateException with Redis sessions enabled
### Expected Behavior

I tried upgrading my Micronaut 2.4.0 application to Micronaut 3.0.0, but found that my `@RequestScope` beans no longer work.

I created a blank app using Micronaut Launch and was able to narrow the issue down to having Redis enabled for HTTP sessions.

The attached GitHub repo has a minimal reproducible case in `com.example.Application`.

### Actual Behaviour

When hitting an authenticated endpoint with Redis enabled, attempting to access a `@RequestScope` bean throws an exception:
```
java.lang.IllegalMonitorStateException: attempt to unlock read lock, not locked by current thread
```

Not sure if related, but I've noticed that with Redis session enabled, my endpoint methods run in the `lettuce-nioEventLoop` thread pool (e.g. `Thread[lettuce-nioEventLoop-4-2,5,main]`). With redis sessions disabled, the endpoint runs in the `default-nioEventLoopGroup` pool. If I use `@ExecuteOn(TaskExecutors.IO)`, the endpoint runs in the `io-executor-thread` pool as expected.

### Steps To Reproduce

Clone https://github.com/chrisparton1991/micronaut-3-redis-session-scope-issue

1. Run a Redis instance on `localhost:6739`
2. Start the Micronaut app
3. Hit `GET http://localhost:8080/test` with an `Authorization: basic` header (doesn't matter what the value is)
4. Receive an HTTP 500 error, and a `java.lang.IllegalMonitorStateException`
5. In `application.yml`, set `micronaut.session.http.redis.enabled` to `false` and the error will no longer occur.

### Environment Information

- Operating System: Windows 10 Pro
- JDK version: AdoptOpenJDK 11.0.7

### Example Application

https://github.com/chrisparton1991/micronaut-3-redis-session-scope-issue

### Version

3.0.0

graemerocher:
If there is no request then request scope shouldn't silently fail. This throws an exception which reveals the true problem in #6096

In additional the RouteExecutor should wrap the route execution in the current request.

Fixes #6096

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -24,6 +24,7 @@ import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.ReturnType;
 import io.micronaut.http.*;
 import io.micronaut.http.bind.binders.ContinuationArgumentBinder;
+import io.micronaut.http.context.ServerRequestContext;
 import io.micronaut.http.exceptions.HttpStatusException;
 import io.micronaut.http.filter.HttpFilter;
 import io.micronaut.http.filter.ServerFilterChain;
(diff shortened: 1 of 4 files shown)


http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -595,19 +584,20 @@ private boolean isSingle(RouteInfo<?> finalRoute, Class<?> bodyClass) {
...
                 if (!routeMatch.isExecutable()) {
                     finalRoute = requestArgumentSatisfier
-                            .fulfillArgumentRequirements(routeMatch, requestReference.get(), true);
+                            .fulfillArgumentRequirements(routeMatch, httpRequest, true);
                 } else {
                     finalRoute = routeMatch;
                 }
 
-                Object body = finalRoute.execute();
+                Object body = ServerRequestContext.with(httpRequest, (Supplier<Object>) finalRoute::execute);
```
jameskleeh, line 595: Did you find a case where the request wasn't present? If so please add a test

graemerocher: It is not present in this example app linked to the issue https://github.com/chrisparton1991/micronaut-3-redis-session-scope-issue/blob/master/src/main/java/com/example/Application.java

It seems to be lost when Micronaut Security is involved. I tried but I cannot seem to come up an example that doesn't involve engaging Micronaut security like in the sample application.

Issue #6139, opened 2021-09-09
HttpHostResolver reuturns port -1 if no host header is set


jmongard:
### Expected Behavior

The port should not be part of the result if it is the default port. 

### Actual Behaviour

The host resolver add -1 to the end e.g. "https://www.example.com:-1".

### Steps To Reproduce

Create a new default project using Micronaut Launch. 
Run the following test:

```
package com.example;

import io.micronaut.http.HttpHeaders;
import io.micronaut.http.HttpRequest;
import io.micronaut.http.server.util.HttpHostResolver;
import io.micronaut.test.extensions.junit5.annotation.MicronautTest;
import jakarta.inject.Inject;
import org.junit.jupiter.api.Test;

import static org.junit.jupiter.api.Assertions.*;

@MicronautTest
class TestHttpHostResolver {
    @Inject
    HttpHostResolver resolver;

    @Test
    void resolveHost_without_host_header() {
        var expected = "https://www.example.com";

        var actual = resolver.resolve(HttpRequest.GET(expected));

        assertEquals(expected, actual);
    }
}
```

### Environment Information

Windows
JDK11

### Example Application

https://github.com/jmongard/demo

### Version

3.0.1

jmongard:
I created a PR with an easy fix for this: https://github.com/micronaut-projects/micronaut-core/pull/6140

PR #6140, opened 2021-09-09
DefaultHttpHostResolver should not return port -1 when HOST header is missing (#6139)

jmongard:
Fixes issue (#6139) with the DefaultHttpHostResolver returning port -1 when there is no HOST header set e.g.  "https://www.example.com:-1".


graemerocher:
Thanks for the contribution!

PR #6141, opened 2021-09-10
Only include JDBC url in detail map. 

### LINKED ISSUE #6136: JdbcIndicator adds url to HealthIndicator on SqlException causing it to print the url to the log
### Expected Behavior

When you get an SQLException you should NOT see the connection string in the log since it can contain the user and password in clear text.

### Actual Behaviour

HealthResult prints out the connection url in the log since JdbcIndicator adds url as name in the HealthResult builder.
`try {
                    key = dataSource.getClass().getMethod("getUrl").invoke(dataSource).toString();
                } catch (Exception n) {
                    key = dataSource.getClass().getName() + "@" + Integer.toHexString(dataSource.hashCode());
                }
HealthResult.Builder builder = HealthResult.builder(key);`

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

2.4.2

graemerocher:
Fixes #6136

### DIFF
--- a/management/src/main/java/io/micronaut/management/health/indicator/jdbc/JdbcIndicator.java
+++ b/management/src/main/java/io/micronaut/management/health/indicator/jdbc/JdbcIndicator.java
@@ -15,17 +15,20 @@
  */
 package io.micronaut.management.health.indicator.jdbc;
 
+import io.micronaut.context.BeanRegistration;
 import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.async.publisher.AsyncSingleResultPublisher;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.health.HealthStatus;
+import io.micronaut.inject.BeanIdentifier;
 import io.micronaut.jdbc.DataSourceResolver;
 import io.micronaut.management.endpoint.health.HealthEndpoint;
 import io.micronaut.management.health.aggregator.HealthAggregator;
 import io.micronaut.management.health.indicator.HealthIndicator;
 import io.micronaut.management.health.indicator.HealthResult;
 import io.micronaut.scheduling.TaskExecutors;
+import jakarta.inject.Inject;
 import jakarta.inject.Named;
 import jakarta.inject.Singleton;
 import org.reactivestreams.Publisher;
@@ -36,6 +39,8 @@ import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


management/src/main/java/io/micronaut/management/health/indicator/jdbc/JdbcIndicator.java
```diff
@@ -60,54 +65,96 @@
...
+        for (BeanRegistration<DataSource> dataSource : dataSources) {
+            DataSource ds = dataSource.bean();
+            if (dataSourceResolver != null) {
+                ds = dataSourceResolver.resolve(ds);
+            }
+            final BeanIdentifier id = dataSource.id();
+            if (id != null) {
+                dataSourceMap.put(id.getName(), ds);
+            } else {
+                dataSourceMap.put("default", ds);
```
jameskleeh, line 118: I think this could result in some datasources not being reported if the beans are created without a named qualifier

jameskleeh:
The URL was used as the key for the reason above. This seems like a breaking change as well

graemerocher:
It may be breaking but the underlying issue is a potential security vulnerability. Passwords should never be logged. If you have a better idea to fix the issue let me know. The other alternative is to never print the name of the indicatorbut even then using the URL as the name seems like in general a bad idea since this could be exposed as more easily and seems to violate the level of detail enum

management/src/main/java/io/micronaut/management/health/indicator/jdbc/JdbcIndicator.java
graemerocher: A data source without a qualifier is the default datasource, if more than one datasource existed without a qualifier then you would get non unique bean exceptions all over the place prior to even hitting this code.

jameskleeh:
What was changed in this PR to not allow passwords to be logged?

management/src/main/java/io/micronaut/management/health/indicator/jdbc/JdbcIndicator.java
jameskleeh: Where would you get a non unique exception specifically? Is that with the assumption they are using our libraries for managing the creation of the beans?

graemerocher: If you have more than one data source and no way to pick which one to use in your application then you must be a super hero regardless if you use our libraries or not

graemerocher:
The health indicator logs the name of each health result which potentially includes the password. 

We could remove that part but then we degrade error diagnosis for all indicators

management/src/main/java/io/micronaut/management/health/indicator/jdbc/JdbcIndicator.java
jameskleeh: Do we not allow injecting a single instance based on the highest order? Does that require super hero powers?

graemerocher:
Based of feedback to this PR I will leave you to come to a solution that resolves the underlying issue since you feel strongly about theoretical problems that are unlikely to actually occur in the wild

PR #6152, opened 2021-09-13
Support an API to start multiple netty servers

graemerocher:
* Refactor NettyHttpServer and introduce NettyEmbeddedServices since the number of constructor parameters had started to make the code a bit of a cluster
* Deprecate `netty.responses.file` config since it made no sense that this wasn't grouped under `micronaut.server.netty`
* Reduce the number of beans since internal APIs like `NettyCustomizableResponseTypeHandlerRegistry` don't need to be beans since they are not extension points for the framework
* Introduce `NettyEmbeddedServerFactory` which allows building an instance of `NettyEmbeddedServer` programmatically if a user wants to build one or more additional servers


http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyEmbeddedServices.java
```diff
@@ -0,0 +1,181 @@
...
+import io.micronaut.http.server.binding.RequestArgumentSatisfier;
+import io.micronaut.http.server.netty.ssl.ServerSslBuilder;
+import io.micronaut.scheduling.executor.ExecutorSelector;
+import io.micronaut.web.router.Router;
+import io.micronaut.web.router.resource.StaticResourceResolver;
+import io.micronaut.websocket.context.WebSocketBeanRegistry;
+import io.netty.channel.ChannelOutboundHandler;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.socket.ServerSocketChannel;
+/*
```
dstepanov: Extra license

inject/src/main/java/io/micronaut/context/BeanContext.java
```diff
@@ -56,6 +56,18 @@
...
+    /**
+     * Obtain an {@link io.micronaut.context.event.ApplicationEventPublisher} for the given even type.
+     * @param eventType The event type
+     * @param <E> The event generic type
+     * @return The event publisher, never {@code null}
+     */
+    @SuppressWarnings("unchecked")
+    default @NonNull <E> ApplicationEventPublisher<E> getEventPublisher(@NonNull Class<E> eventType) {
+        Objects.requireNonNull(eventType, "Event type cannot be null");
+        return getBean(Argument.of(ApplicationEventPublisher.class, eventType));
```
dstepanov, line 68: Correct way is to use:
```
 beanContext.getBeansOfType(ApplicationEventListener.class, Qualifiers.byTypeArguments(eventType.getType()))
```
In the same way as in `ApplicationEventPublisherFactory`, if I remember correctly `Argument.of(ApplicationEventPublisher.class, eventType)` would not trigger super class event listeners.

dstepanov (changes requested):

inject/src/main/java/io/micronaut/context/BeanContext.java
graemerocher: the call to `getBean` ends up calling `ApplicationEventPublisherFactory` so I don't think anything needs to be done here

dstepanov (approved):

PR #6167, opened 2021-09-16
Do not forward onComplete before all elements have been processed

yawkat:
This fixes another issue that I believe to be a bug with `SingleThreadedBufferingSubscriber`: Even if `onComplete` arrives before all data has been requested by the downstream (i.e. when we're buffering), the `onComplete` would be forwarded immediately and the buffered elements would be dropped (`processDemand` does nothing in state `DONE`).

This patch delays the `doOnComplete` call until all elements have been flushed.

Unlike #6147, which was clearly a bug, this patch is technically a behavior change. But it seems like the behavior implementors would expect.


jameskleeh:
Superseded by #6171

PR #6168, opened 2021-09-16
Replace use of @Overrides with caching

### LINKED ISSUE #6158: overlapping routes doesn't work in 3.0.1
### Expected Behavior

When having two controller with overlapping routes like

- /first/{id}
- /first/{id}/second/{id2}

I expect that both Controller can be called by there routes

### Actual Behaviour

In 3.0.1 only the second route is registered anymore. a Request to the first one will result in a 404

### Steps To Reproduce

- clone https://github.com/volkerrichert/overlappingRoutes
- build & run
- call http://localhost:8080/first/123 -> 404
- call http://localhost:8080/first/123/second/456 -> 200
- comment all the MN stuff in API2 & controller2
- clear
- rebuild & run
- call http://localhost:8080/first/123 -> 202

You can also check /routes

### Environment Information

- Windows 11
- JDK 11
- MN 3.0.1

MN 2.5 works as expected

### Example Application

https://github.com/volkerrichert/overlappingRoutes

### Version

3.0.1

graemerocher:
The root cause of #6158 is that incremental compilation doesn't work properly with source retention annotations. The reason for this is that Gradle materializes an element from compiled byte code when compiling incrementally and this element no longer has the source level annotations like `@Overrides`.

This is a bummer since it means that source level annotations cannot be used in general with incremental compilation. I am not sure if we should be printing warnings if incremental is enabled and any source level retention annotations are used.

Nevertheless we were relying on `@Overrides` to detect when to inherit annotations so when compiling incrementally these annotations disappeared from the incremental result.

The fix here replaces the reliance of `@Overrides` with a cache and optimizes the code to be more performant. This cache is cleared at the end of each compilation round to ensure it is not stale.

Fixes #6158

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
@@ -270,6 +270,7 @@ public class BeanDefinitionInjectProcessor extends AbstractInjectAnnotationProce
             } finally {
                 AnnotationUtils.invalidateCache();
                 AbstractAnnotationMetadataBuilder.clearMutated();
+                JavaAnnotationMetadataBuilder.clearCaches();
             }
         }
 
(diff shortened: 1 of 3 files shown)


melix (approved):
This LGTM but I'd try to add a test exercising the problem it solves.

graemerocher:
Would love to but I have no idea how to write incremental tests for Gradle 🤔

PR #6170, opened 2021-09-16
Fix support nullable inputstream 

### LINKED ISSUE #6125: @io.micronaut.core.annotation.Nullable doesn't work with Inputstream and reactive types like Flowable<byte[]> payload
### Expected Behavior

Request without body to work if method in controller has @Nullable annotation on Body where body type is inputstream or reactive types.

### Actual Behaviour

Getting 400 bad request
{
    "message": "Required Body [payload] not specified",
    "path": "/payload",
    "_links": {
        "self": {
            "href": "/stream/hello",
            "templated": false
        }
    }
}

This works if body is of type byte[]

### Steps To Reproduce

Create controller methods with signature like below, expected following curl to work

```
@Post(uris = {
            "/hello"}
    )
    @Produces({MediaType.APPLICATION_JSON})
    @Consumes({MediaType.ALL})
    @ExecuteOn(TaskExecutors.IO)
    public MutableHttpResponse<String> stream1(HttpRequest httpRequest, @Nullable @Body Flowable<byte[]> payload)
```

or
```
 @Post(uris = {
            "/hello"}
    )
    @Produces({MediaType.APPLICATION_JSON})
    @Consumes({MediaType.ALL})
    @ExecuteOn(TaskExecutors.IO)
    public MutableHttpResponse<String> stream1(HttpRequest httpRequest, @Nullable @Body Inputstream payload) throws IOException, InterruptedException {
```

Send below curl

```
curl --location --request POST ‘http://localhost:9090/stream/hello’
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

2.5.5,2.5.12

graemerocher:
Fixes #6125

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.http.server.netty;
 
+import io.micronaut.core.async.publisher.Publishers;
 import io.micronaut.core.util.Toggleable;
 import io.netty.buffer.ByteBufHolder;
 import org.reactivestreams.Publisher;
@@ -28,5 +29,5 @@ import org.reactivestreams.Subscriber;
  * @author Graeme Rocher
  * @since 1.0
  */
-public interface HttpContentProcessor<T> extends Publisher<T>, Subscriber<ByteBufHolder>, Toggleable {
+public interface HttpContentProcessor<T> extends Publishers.MicronautPublisher<T>, Subscriber<ByteBufHolder>, Toggleable {
 }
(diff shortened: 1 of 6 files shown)


jameskleeh:
This introduces a breaking change in behavior where previously a request without a body to a method like `@Body Publisher<byte[]> body` would produce a bad request response:

`{"message":"Bad Request","_links":{"self":{"href":"/","templated":false}},"_embedded":{"errors":[{"message":"Required Body [body] not specified","path":"/body"}]}}`

But now the method is executed with an empty publisher. Because the user did not add `@Nullable` to the argument I think the previous behavior is more desirable because now it is up to the user to determine what should happen if no data is produced.

Edit: Reverting the one line that was changed from returning EMPTY to an empty flux fixes the original issue

graemerocher:
I will update the Pr to make it configurable

jameskleeh:
@graemerocher I would say just revert that one line for this PR and introduce another PR that targets 3.1.x if you want to introduce a new feature to the binder

PR #6173, opened 2021-09-17
Using @ConfigurationProperties methods in @Requires annotation

### LINKED ISSUE #5350: Feature request: possibility to use values of @ConfigurationProperties in annotations
Currently it is possible to describe all of the configuration of application in a type-safe way with [@ConfigurationProperties](https://docs.micronaut.io/latest/guide/index.html#configurationProperties), but it is impossible to use it's methods in annotation like `@Scheduled`, `@Requres`, etc. It creates an inconsistency in configuration as part of it either not described in properties classes, or should be duplicated. This inconsistency prevents a user from having a single configuration point which could be through out all application and validated on startup.

Theoretically, it could look like this:
```java
@ConfigurationProperties("app.scheduling")
public class SchedulingProperties
{
    private boolean enabled = true;

    @NotBlank
    private String schedule;

    public boolean isEnabled()
    {
        return enabled;
    }

    public void setEnabled(boolean enabled)
    {
        this.enabled = enabled;
    }

    public String getSchedule()
    {
        return schedule;
    }

    public void setSchedule(String schedule)
    {
        this.schedule = schedule;
    }
}

public class Service
{
    @Requires(config = SchedulingProperties.class, method="isEnabled")
    @Scheduled(config = SchedulingProperties.class, method="getSchedule")
    public void action()
    {

    }
}

```

GavrilovSV:
This is a possible solution to resolve #5350 
I'd like someone to review this PR to find out whether this solution is ok, or some sort of fail fast validation is expected in case of mistyping the method name or providing method with wrong return type. Then I can proceed with the implementation of similar functionality in other annotations like @Scheduled etc


dstepanov:
Micronaut doesn't use reflection, to avoid that you would need to have the method `@Executable`
But I don't think that is something that should be supported, having a method defined by its name as a string is not the best idea, and how is it better than using the property path?

I would suggest using `Condition` and `@Requires(condition=MyCondition.class)`.
Or event creating custom annotations and use `@SchedulingProperties.Requires` and `@SchedulingProperties.Scheduled`

graemerocher:
I guess the only advantage of this approach that could be is if some kind of compilation time checking is supported, but even then I don't think it should support methods but only properties. Something like:

```
@Requires(configuration=TestConfig.class, property="enabled") 
```

And if `TestConfig` doesn't define an `enabled` property then fail compilation. We would also probably want to support `value` like:

```
@Requires(configuration=TestConfig.class, property="enabled", value="true")
```

But there is lots of think about (what about nested properties "foo.bar", what about when `property` is not specified etc.)

GavrilovSV:
@graemerocher 
I think compilation time checking can be implemented through the usage of recently introduced `BeanElementVisitor` as a part of `validation` module
 
Since the original issue was about avoiding duplication, I would also suggest usage of `@ConfigurationProperties` prefixes to resolve full path to specified property. For example:
```
@ConfigurationProperties("outer")
public class OuterConfig {

     @ConfigurationProperties("inner")
     public static class InnerConfig {
            private String innerProperty; 
     }
}
```
then we could do something like:
```
@Requires(configProperties = OuterConfig.InnerConfig.class, configProperty = "inner-property")
```
instead of 
```
@Requires(property = "outer.inner.inner-property")
```

The problem of this approach is handling default values of properties specified inside `@ConfigurationProperties` classes. For example:
```
@ConfigurationProperties("config")
public class Config {

     private String property = "defaultValue";

     public void setProperty(String property) {
          this.property = property;
     }
 
     public String getProperty() {
         return this.property;
     }
}

@Requires(configProperties = Config.class, configProperty = "property")
public class ConditionalBean
```
In this case I would expect the conditional bean to be created even if I don't explicitly specify the property in external configuration since the default value is already provided inside the class. The same problem arises if I want the configuration property value to equal/not equal certain value:
```
@Requires(configProperties = Config.class, configProperty = "property", notEquals = "defaultValue")
```
The only solution I can see is using bean introspection for all `@ConfigurationProperties` classes, not only those which are explicitly annotated with `@Introspected` or require constraint validations (as it works at the moment, I guess), but I'm not sure whether extra overhead of such approach worth introducing this feature

graemerocher:
Yes I think it would have to refer to the non-normalized name of properties only like `fooBar` (camel case)

GavrilovSV:
@graemerocher 
I've added support for config properties in `@Requires` and implemented compile time validation. Could someone pls review this PR?

graemerocher:
@GavrilovSV will do, I think it needs some consideration as it will essentially generate introspections for all configuration properties right? This could have memory implications

graemerocher:
@dstepanov what do you think? A lot of work has gone into this PR (Thanks!) perhaps there is a way to generate an optimized resolution handle to get the value of the property without an introspection or perhaps generate an introspection only if needed

GavrilovSV:
Yes, this implementation generates introspections for all configuration properties, even when those classes are not referenced in @Requires anywhere. On the other hand it seems to me like the main advantage of configuration properties is validation support, which requires introspections anyway. But if a way to generate introspections only when needed (or avoiding them at all) could be proposed, that would be great

dstepanov:
Looks good, I think this should also support checking if the configuration is implementing `io.micronaut.core.util.Toggleable`.
What is going to happen if the configuration is not introspected?

inject/src/main/java/io/micronaut/context/RequiresCondition.java
```diff
@@ -647,4 +658,77 @@ private boolean matchesCurrentOs(ConditionContext context, AnnotationValue<Requi
         }
         return true;
     }
+
+    private boolean matchesConfigurationProperties(ConditionContext context, AnnotationValue<Requires> requirements) {
+        if (requirements.contains(MEMBER_CONFIGURATION_PROPERTIES)) {
+            Class configPropertiesClass = requirements.classValue(MEMBER_CONFIGURATION_PROPERTIES).orElse(null);
+            String configProperty = requirements.stringValue(MEMBER_CONFIGURATION_PROPERTY).orElse(null);
+
+            if (configPropertiesClass != null) {
```
dstepanov: I'm not a fan of multiple nested if, it makes code hard to read.
I would write something like:
```java
if (configPropertiesClass == null) {
    ...
    return
}
if (!beanContext.containsBean(configPropertiesClass)) {
    ...
    return
}
if (StringUtils.isEmpty(configProperty)) {
    ...
    return
}
if (beanProperty == null) {
    ...
    return
}
```

GavrilovSV: Agree. Just tried to follow the style other methods of `RequiresCondition` are written in. I'll refactor it to be more flat

GavrilovSV:
@dstepanov 
> I think this should also support checking if the configuration is implementing `io.micronaut.core.util.Toggleable`. 

We can log warning on compilation if `ConfigurationProperties` class does not implement `io.micronaut.core.util.Toggleable`.  But I think it should be a separate validator in `validation` module. In this case I would also suggest mentioning in documenation that `ConfigurationProperties` are expected to implement `io.micronaut.core.util.Toggleable`

> What is going to happen if the configuration is not introspected?

If introspection for config properties class can not be loaded, the conditional bean referencing it will not be loaded as well

dstepanov:
The idea behind `io.micronaut.core.util.Toggleable` is that the configuration can define an `enabled` property that will disable the configuration. So I would expect that the required condition will fail also if the configuration implements `Toggleable` and returns enabled false.

> If introspection for config properties class can not be loaded, the conditional bean referencing it will not be loaded as well

This will be confusing. I would prefer an exception if the property cannot be accessed because of the missing introspection.
Maybe it will be possible to annotate the configuration by the annotation processor.

graemerocher:
Also I am unsure we should be generating introspections for all configuration properties. A user  can always use `@Introspected(classes=..)` to add any missing introspections

GavrilovSV:
@dstepanov 
I've considered your suggestion regarding `Toggeable` interface and I've got couple of issues with it:
1. I don't see a way of determining that `ConfigurationProperties` class implements `Toggeable` interface in runtime without using reflection (maybe I'm just not aware of the right way to do it)
2. I'm not sure, whether disabling conditional bean in case `enabled` property on referenced configuration is false would be expected since `Toggeable` presence doesn't seem to have such impact anywhere else at the moment (correct me if I'm wrong), so I think it may be confusing if it affects only `ConfigurationProperties` and only if it's referenced in `Requires` annotation. Anyway, user can just specify desired value for `enabled` property directy using `value` or `notEquals`

@graemerocher 
Denis suggested throwing an exception in case introspection is not available for `ConfigurationProperties` class. I agree that it may be the preferred behaviour, so I think making a user responsible for providing introspections is okay in case it is well documented and a meaningful exception message is provided when introspection is missing

dstepanov:
1. You can use `instanceof` after `Object bean = beanContext.getBean(beanDefinition);`
2. I think that's the idea behind `Toggleable` - `An interface for components or configurations that can be toggled on or off.`

GavrilovSV:
@graemerocher @dstepanov 
I've updated the branch, added support for `Toggleable` feature and refactored a bit. Now an exception is throws in case no introspection is available for configuration properties. I've also reverted changes in `IntrospectedTypeElementVisitor`, so in current implementation introspections are not created for all configuraiton properties by default. Tests for new features were added as well

jameskleeh:
Given there isn't anything in this feature that actually requires `@ConfigurationProperties`, couldn't we reuse `@Requires(beans = ConfigPropsBean, property = "someProp")`. The implementation doesn't need to care if it's a config props or not.

Edit: I suppose the above example could be a breaking change if people are using the annotation incorrectly. Perhaps add a `bean` member.

graemerocher:
One issue that we have to be careful about is that currently `@Requires(beans=FooBar.class)` doesn't actually materialize the bean it just checks that there is a bean like that in the context. Adding this feature and any variant that has `property` means that we have the materialize the bean inside of `RequiresCondition` this can be problematic as it means requires evaluation can result in bean initialization which can lead to circular bean reference problems.

If we isolate it to just configuration properties maybe those can be mitigated somewhat but I am not super keen on adding bean initialization to the `@Requires` logic 🤔

jameskleeh:
We should be able to put together the property string from the annotations and convert it to a property check

graemerocher:
What about `@Requires(beans = ConfigPropsBean, property = "someProp", value="blah")` ? In order to do the value evaluation you have to init the bean.

graemerocher:
The current implementation is in fact already doing that 

https://github.com/micronaut-projects/micronaut-core/pull/6173/files#diff-abfb539ae66f2056d09796347f04c8e9104dcc8a1a4841a6c712fe7080334953R689-R690

And since the resolution context is not passed it is likely to cause issues

GavrilovSV:
@graemerocher @dstepanov 
I was thinking about the ways of incorporating instantiation of beans referenced in `Requires(configProperties=...)` into existing bean creation flow and tried to implement the injection of such beans directly into the bean definition class similar to the way field injection works at the moment. This approach allows to prevent aforementioned circular dependency issues and would not require to initiate bean creation inside of `RequiresCondition`.

I was also thinking that it could eliminate the requirement to annotate `@ConfigurationProperties` classes with `@Introspected` since property getter invocation and value comparision could be generated at compilation time.

I've updated the branch with a proof of concept implementation which supports property presence check and equality comparision for value (in fact, there seems to be nothing requiring the usage of this feature with `@ConfigurationProperties` only and any bean property can be referenced this way). The code is far from perfect and there are a lot of details which should be taken into consideration. Some optimizations are missing as well (e.g. beans referenced multiple times are listed multiple times, the required bean is requested from context multiple times as well), but I didn't want to dive deeper into details before making sure that such approach has a chance and wouldn't itself lead to the problems which I'm not aware of at the moment

In a nutshell, the current implementation introduces `shouldLoad` method returning boolean which is invoked before the invocation of `build(...)` method on bean definition. The code inside `shouldLoad` method is generated at compilation time and looks like this:
```
public boolean shouldLoad(BeanResolutionContext beanResolutionContext, BeanContext beanContext) {
     String property = ((ConfigPropertyClass) super.getBeanForAnnotation(beanResolutionContext, beanContext, 0, (Qualifier)null)).getConfigProperty();
     // Not quite sure whether the usage of SHARED ConversionService is acceptable here
     String convertedProperty = (String)ConversionService.SHARED.convert(property, ConversionContext.of(String.class)).orElse((Object)null);
     return convertedProperty != null && convertedProperty.equals("requiredValue");
}
```

ivyazmitinov:
@dstepanov @graemerocher is there any chance you can take a look at the last changes? We have more and more pressing need in unified configuration point and really hope for this feature to be available :)

inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -608,6 +610,45 @@ public BeanDefinitionVisitor getOrCreateBeanDefinitionWriter(TypeElement classEl
...
         }
 
+        private void visitAnnotationMetadata(AnnotationMetadata annotationMetadata) {
+            if (annotationMetadata.hasAnnotation(Requires.class)) {
+                for (io.micronaut.core.annotation.AnnotationValue<Requires> annotation: currentClassMetadata.getAnnotationValuesByType(Requires.class)) {
+                    String configProperty = annotation.stringValue(RequiresCondition.MEMBER_CONFIGURATION_PROPERTY).orElse(null);
+
+                    TypeElement typeElement =
+                        annotation.stringValue(RequiresCondition.MEMBER_CONFIGURATION_PROPERTIES)
+                            .map(name -> name.replace("$", "."))
```
graemerocher: You can use the replace method that takes a char instead

inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -608,6 +610,45 @@ public BeanDefinitionVisitor getOrCreateBeanDefinitionWriter(TypeElement classEl
...
+                            .map(elementUtils::getTypeElement)
+                            .orElse(null);
```
graemerocher: You can use `orElseGet(() -> elementFactory.newClassElement(element, annotationUtils.getAnnotationMetadata(element)))` and delete the code that uses optional that follows in line 624

inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -608,6 +610,45 @@ public BeanDefinitionVisitor getOrCreateBeanDefinitionWriter(TypeElement classEl
...
+                    if (typeElement != null && configProperty != null) {
+                        BeanDefinitionVisitor writer = getOrCreateBeanDefinitionWriter(concreteClass, concreteClass.getQualifiedName());
+
+                        ExecutableElement propertyGetter =
+                            ElementFilter.methodsIn(elementUtils.getAllMembers(typeElement))
+                                .stream()
+                                .filter(method -> NameUtils.isGetterName(method.getSimpleName().toString()))
+                                .filter(method -> NameUtils.getPropertyNameForGetter(method.getSimpleName().toString()).equals(configProperty))
+                                .findFirst()
+                                .orElse(null);
```
graemerocher: This should use the `classElement` and instead use `getBeanProperties()` so that support for records works and support for the new `AccessStyle` annotation. See https://github.com/micronaut-projects/micronaut-core/pull/6620

inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -608,6 +610,45 @@ public BeanDefinitionVisitor getOrCreateBeanDefinitionWriter(TypeElement classEl
...
+
+                        if (propertyGetter != null) {
+                            JavaMethodElement javaMethodElement = elementFactory.newMethodElement(classElement,
+                                propertyGetter, annotationUtils.getAnnotationMetadata(propertyGetter));
+
+                            String value = annotation.stringValue().orElse(null);
+                            writer.visitAnnotationPropertyInjectionPoint(classElement, javaMethodElement, value);
```
graemerocher: This is good for `inject-java` but we will also need to update `inject-groovy` for Groovy support. It probably makes sense to move much of the logic into the writer and use the `ElementQuery` API as described above which is portable across Groovy & Java

inject/src/main/java/io/micronaut/context/AbstractBeanResolutionContext.java
```diff
@@ -585,6 +596,34 @@ public AnnotationMetadata getAnnotationMetadata() {
         }
     }
 
+    /**
+     * A segment that represents annotation.
+     */
+    public static class AnnotationSegment extends AbstractSegment implements InjectionPoint {
```
graemerocher: Make `final` and add `@since 3.3.0`

graemerocher: This needs to implement the `toString()` method so that errors are formatted correctly. Should also add a test for the error case

inject/src/main/java/io/micronaut/context/BeanResolutionContext.java
```diff
@@ -223,6 +223,8 @@ default void close() {
          */
         Path pushFieldResolve(BeanDefinition declaringType, Argument fieldAsArgument, boolean requiresReflection);
 
+        Path pushAnnotationResolve(BeanDefinition beanDefinition, Argument annotationBeanAsArgument);
```
graemerocher: Needs javadoc with `@since`

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
```diff
@@ -303,6 +303,16 @@ default T inject(BeanResolutionContext resolutionContext, BeanContext context, T
         return bean;
     }
 
+    /**
+     * Decide whether the bean should be loaded.
+     * @param resolutionContext - the resolution context
+     * @param context - the bean context
+     * @return should be be loaded
+     */
+    default boolean shouldLoad(BeanResolutionContext resolutionContext, BeanContext context) {
```
graemerocher: Add `@since` and mark as `@Internal` since `BeanResolutionContext` is internal

The two arguments should be annotated with `@NonNull`

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionVisitor.java
```diff
@@ -251,6 +251,17 @@ void visitFieldInjectionPoint(TypedElement declaringType,
...
                                   boolean requiresReflection);
 
+    /**
+     * Visits an annotation injection point.
+     *
+     * @param injectedBeanType   The type of the injected bean
+     * @param getterMethod       The getter method which has to be invoked on injected bean to obtain property value
+     * @param expectedValue      The value which is compared to the value returned by getterMethod
+     */
+    void visitAnnotationPropertyInjectionPoint(TypedElement injectedBeanType,
```
graemerocher: Add `@since`

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1026,6 +1067,82 @@ public void accept(Map<String, ClassElement> stringClassElementMap) {
...
+            adapter.loadThis();
+            // 1st argument load BeanResolutionContext
+            adapter.loadArg(0);
+            // 2nd argument load BeanContext
+            adapter.loadArg(1);
+            // 2rd argument the field index
+            adapter.push(currentIndex);
+            // push qualifier
+            pushQualifier(adapter, visitData.annotationBeanType, () -> resolveAnnotationArgument(adapter, currentIndex));
+            // invoke getBeanForField
```
graemerocher: incorrect comment

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -2113,6 +2238,16 @@ private void resolveBeanOrValueForSetter(TypedElement declaringType, MethodEleme
         }
     }
 
+    private void visitShouldLoadMethodDefinition() {
+        String desc = getMethodDescriptor("boolean", BeanResolutionContext.class.getName(), BeanContext.class.getName());
+        this.shouldLoadMethodVisitor = new GeneratorAdapter(classWriter.visitMethod(
+            ACC_PUBLIC,
```
graemerocher: Should be `final`

inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
```diff
@@ -151,6 +153,7 @@ protected AbstractInitializableBeanDefinition(
             @Nullable AnnotationMetadata annotationMetadata,
             @Nullable MethodReference[] methodInjection,
             @Nullable FieldReference[] fieldInjection,
+            @Nullable AnnotationReference[] annotationInjection,
```
graemerocher: This is a breaking binary incompatible change. Revert the change to the constructor and add a new constructor with this argument.

inject/src/main/java/io/micronaut/context/annotation/Requires.java
```diff
@@ -204,6 +204,25 @@
...
 
+    /**
+     * Expresses that configuration properties of the given class should be
+     * available for the bean or configuration to load. Can be used in combination
+     * with {@link #configProperty()} to specify the required configuration
+     * property
+     *
+     * @return The configuration properties class
+     */
+    Class configProperties() default void.class;
```
graemerocher: add `@since 3.3.0`

inject/src/main/java/io/micronaut/context/annotation/Requires.java
```diff
@@ -204,6 +204,25 @@
...
+
+    /**
+     * Used in combination with {@link #configProperties()} to
+     * express that the given configuration property should be
+     * set for the bean to load.
+     *
+     * @return The configuration property that should be set.
+     */
+    String configProperty() default "";
```
graemerocher: add `@since 3.3.0`

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1026,6 +1067,82 @@ public void accept(Map<String, ClassElement> stringClassElementMap) {
...
+            // storing getter return value to local variable
+            int propertyValueIndex = adapter.newLocal(JavaModelUtils.getTypeReference(getterVisitData.getMethodElement().getReturnType()));
+            adapter.storeLocal(propertyValueIndex);
+            // loading conversion service instance
+            adapter.getStatic(Type.getType(ConversionService.class), "SHARED", Type.getType(ConversionService.class));
+
+            adapter.loadLocal(propertyValueIndex);
+
+            // creating String conversion context
+            adapter.push(Type.getType(String.class));
```
graemerocher: Extract `Type.getType(String.class)` into a variable it is used a lot

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1026,6 +1067,82 @@ public void accept(Map<String, ClassElement> stringClassElementMap) {
...
+            pushCastToType(adapter, getterVisitData.getBeanType().getType());
+            // invoke property getter
+            adapter.invokeVirtual(JavaModelUtils.getTypeReference(getterVisitData.getBeanType()),
+                org.objectweb.asm.commons.Method.getMethod(getterVisitData.getMethodElement().getDescription(false)));
+
+            // storing getter return value to local variable
+            int propertyValueIndex = adapter.newLocal(JavaModelUtils.getTypeReference(getterVisitData.getMethodElement().getReturnType()));
+            adapter.storeLocal(propertyValueIndex);
+            // loading conversion service instance
+            adapter.getStatic(Type.getType(ConversionService.class), "SHARED", Type.getType(ConversionService.class));
```
graemerocher: Extract `Type.getType(ConversionService.class)` into a variable it is used a lot

validation/src/main/java/io/micronaut/validation/requires/RequiresConfigPropertiesVisitor.java
```diff
@@ -0,0 +1,141 @@
...
+            if (!isConfigPropertyBean) {
+                fail("Specified class [" + configPropertiesClass + "] must be configuration properties bean", visitorContext, conditionalBean);
+                return;
+            }
+
+            configPropertiesClass = configPropertiesClass.replace("$", ".");
+            if (configProperty != null) {
+                boolean propertyPresent = false;
+                for (PropertyMetadata propertyMetadata: metadataBuilder.getProperties()) {
+                    if (propertyMetadata.getDeclaringType().equals(configPropertiesClass) && propertyMetadata.getName().equals(configProperty)) {
```
graemerocher: I'm not sure this is sufficient with regards to Java visibility rules. What if the config property is package private and in a different package and therefore inaccessible? I think only public configuration properties should be considered.

graemerocher (changes requested):
This seems like a better direction, I have added some feedback.

In general I think we need to think about accessibility rules etc. The code is seemingly primarily testing public getters but it needs test for other cases like:

- Java 17 records
- non-public getters (if classes are in the same package they are accessible otherwise not)

graemerocher:
Could you also resolve the conflicts and get the tests passing

graemerocher:
Plus needs to target 3.3.x branch

GavrilovSV:
@graemerocher 
I've updated the implementation. I think there's no real need to limit the usage of this feature to `@ConfigurationProperties` only, so I've replaced `@Requires(configProperties=RequiredBean.class, configProperty = "property")` with `@Requires(bean = RequiredBean.class, beanProperty = "beanProperty")`. In case no `beanProperty` is specified, then `bean` annotation member behaves the same way as current `@Requires(beans={RequiredBean.class})` .  It also supports `value` and `notEquals` annotation members. 
The logic is implemented inside `checkIfShouldLoad` mehtod which is invoked and generated only if needed. In case bean is disabled, the `DisabledBeanException` is thrown from the `checkIfShouldLoad` method with descriptive message. 

It works fine with:
- Java 17 records
- `@AccessorStyle` modifiers
- Beans created inside `@Factory` classes
- Both privimitives and wrapper property types
- Different types of `@ConfigurationProperties` including nested properties, properties with inheritance and interface properties

Regarding properties visibility rules, since I've not limited the usage to `@ConfigurationProperties` only, this implementation allows referencing package-private and protected properties in case the referencing class is located in the same package  
I've also added tests both for groovy and java

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1127,6 +1163,150 @@ public void accept(Map<String, ClassElement> stringClassElementMap) {
...
+                // label visited when no exception is thrown
+                adapter.visitLabel(furtherCheckLabel);
+                // if check is over returning from method
+                if (isLastProperty) {
+                    adapter.returnValue();
+                }
+            }
+        }
+    }
+
```
graemerocher: I admire that you managed to generate working byte code for this logic, but it is quite a bit of logic. I think it makes more sense to move the logic into a method that you call with the data and implement the logic in Java, it will make it more maintainable. We try in general to keep us much logic out of the byte code generation as possible.

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1686,6 +1866,51 @@ private boolean isInnerType(ClassElement genericType) {
...
+            final String[] readPrefixes = annotationMetadata.getValue(AccessorsStyle.class, "readPrefixes", String[].class)
+                .orElse(new String[]{AccessorsStyle.DEFAULT_READ_PREFIX});
+
+            memberPropertyGetter = annotationMemberClassElement.getEnclosedElements(ElementQuery.ALL_METHODS)
+                .stream()
+                .filter(methodElement -> NameUtils.getPropertyNameForGetter(methodElement.getSimpleName(), readPrefixes).equals(annotationMemberProperty))
+                .filter(this::isMethodAccessible)
+                .findFirst()
+                .orElse(null);
+        }
```
graemerocher: I'm not sure if this is necessary, since `getBeanProperties()` already has the logic to handle `readPrefixes`

inject/src/main/java/io/micronaut/context/annotation/Requires.java
```diff
@@ -204,6 +204,25 @@
...
+
+    /**
+     * Used in combination with {@link #bean()} to
+     * express that the given bean property should be
+     * set for the bean to load.
+     *
+     * @return The configuration property that should be set.
+     * @since 3.3.0
+     */
+    String beanProperty() default "";
```
graemerocher, line 224: Perhaps this member is unnecessary and it could expressed as just `@Requires(bean=Foo, property="bar")` without adding a new member. Thoughts @jameskleeh ?

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/annotation/Requires.java
jameskleeh: In general I would agree but I think that would change the current behavior, thus making it a 4.0 thing.

`@Requires(bean=X, property="y")` today means the `X` bean has to be present and the configuration property `y` has to be present. Reusing `property` for this case changes the semantics.

graemerocher: Ok leave it as is then

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
GavrilovSV: Simplified this part. Property `value` and `notEquals` comparision is now implemented in separate java method in `AbstractInitializableBeanDefinition`

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
GavrilovSV: This part is required because it allows to reference package-private and protected properties from classes in sames packages while usage of `getBeanProperties()` allows to reference public properties only. I've also faced an issue with `@ConfigurationProperties` on groovy interfaces - `getBeanProperties()` doesn't consider abstract getters as properties because of line 935 in `GroovyClassElement` (there is no such issue with JavaClassElement):

```boolean validModifiers = node.isPublic() && !node.isStatic() && !node.isSynthetic() && !node.isAbstract();```

Comment (graemerocher) [reply inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java]: I see, the latter issue with Groovy seems like a bug.

Comment (GavrilovSV) [reply inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java]: At least, it seems like removing `!node.isAbstract()` doesn't break any tests. I can remove it in this PR. However, the question about usage of this feature with package-private and protected properties is still open

**Comment by GavrilovSV** [comment]
@graemerocher 
The original intent of this PR was not only about allowing to use `@ConfigurationProperties` in `@Requires`, but about the opportunity to make `@ConfigurationProperties` a unified configuration point. At least, we would also like to add similar opportunity to reference bean properties in `@Client` since some external urls in our services are located in configs, and `@Scheduled` (there may also be some other points where it could be useful). I can try to implement these extra features as part of this PR (but I'm not sure how much time it may take) or leave it for a separate PR

**Comment by graemerocher** [comment]
@jameskleeh please can you take another look, looks ok to me

**Comment by GavrilovSV** [comment]
@graemerocher 
Hey, since 3.3 is released, should I rebase it to 3.4.x branch? Could you pls also change milestone

**Comment by graemerocher** [comment]
@GavrilovSV please yes target 3.4.x

Sorry this didn't make it but @jameskleeh had a full plate and needs to do a final pass. Will definitely go into 3.4.

File: inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java [hunk 21]
```diff
@@ -1686,6 +1794,52 @@ private boolean isInnerType(ClassElement genericType) {
...
+
+        if (memberPropertyGetter == null) {
+            final String[] readPrefixes = annotationMemberBeanType.getAnnotationMetadata()
+                .getValue(AccessorsStyle.class, "readPrefixes", String[].class)
+                .orElse(new String[]{AccessorsStyle.DEFAULT_READ_PREFIX});
+
+            memberPropertyGetter = annotationMemberClassElement.getEnclosedElements(ElementQuery.ALL_METHODS)
+                .stream()
+                .filter(methodElement -> NameUtils.getPropertyNameForGetter(methodElement.getSimpleName(), readPrefixes).equals(annotationMemberProperty))
+                .filter(this::isMethodAccessible)
```
jameskleeh: This is incorrect. There is an `isAccessible` method on the element query API you should use instead

GavrilovSV: fixed

GavrilovSV:
@graemerocher @jameskleeh 
Should anything else be done as part of this PR? The CI pipeline has failed but it seems like its not because of any issues with this PR

graemerocher:
@jameskleeh ?

jameskleeh (approved):

GavrilovSV:
@graemerocher 
Can this be merged?

graemerocher:
Thanks for the contribution!

Issue #6113, opened 2021-09-07
MDC lost when processes are resumed on a different thread after hitting a suspension point


22dasi1bwi:
### Expected Behavior

Under heavy load MDC keeps correct information after first suspension point.

### Actual Behaviour

Under heavy load MDC loses correct information after first suspension point. See [example application](https://github.com/22dasi1bwi/micronaut-coroutines-context-loss-on-thread-switch) for more information.

Is there a preferred way on how to correctly do this with Reactor in Micronaut 3.0?

### Steps To Reproduce

All required steps to reproduce the behavior are provided in the [example application](https://github.com/22dasi1bwi/micronaut-coroutines-context-loss-on-thread-switch) . Also the background information are described there.

See [related issue](https://github.com/micronaut-projects/micronaut-core/issues/5656) for even more details.

### Environment Information

Operating System: Arch Linux
JDK Version: 11.0.12

### Example Application

https://github.com/22dasi1bwi/micronaut-coroutines-context-loss-on-thread-switch

### Version

3.0.0

dstepanov:
Can you please explain:
```
After upgrading to Micronaut 3.0 the two classes required to get things working in Micronaut 2.5.x com.example.ServerRequestContextFilterCopy.kt and com.example.ServerRequestContextInstrumenterCopy.kt are preventing the MDC to have proper information once processes are resumed on a different thread after hitting a suspension point.

The removal of those two classes makes stress testing (see Howto and execute_benchmark.sh) only work at a rate of approximately 40%.
```

Are you saying that it does work with those classes added but without it works only 40%?

22dasi1bwi:
It worked with those two classes in Micronaut version 2.x (100%) . But it has an error rate of 100% if I keep those two classes in Micronaut version 3.x . 
In Micronaut version 3.x I had to remove those two classes to get context propagation work at all. With those two classes gone the success rate raised from 0% to like 40%.

dstepanov:
This is fixed now in 3.1 and BTW `client.exchange(request, String::class.java).awaitFirst()` can be non-blocking `client.exchange(request, String::class.java).asFlow().single()`

22dasi1bwi:
@dstepanov 
Thanks for the fix. Can't wait to try it out! :smiley: 

I thought `awaitFirst()` would be non-blocking, too? That's at least what the documentation says:

[...]

PR #6183, opened 2021-09-20
Fix reactor instrumentation

dstepanov:
At this moment for request body routes this code:
```java
        if (!route.isExecutable() &&
                io.micronaut.http.HttpMethod.permitsRequestBody(request.getMethod()) &&
                nativeRequest instanceof StreamedHttpRequest &&
                (!bodyArgument.isPresent() || !route.isSatisfied(bodyArgument.get().getName()))) {
            routeMatchPublisher = Mono.<RouteMatch<?>>create(emitter -> httpContentProcessorResolver.resolve(request, route)
                    .subscribe(buildSubscriber(request, route, emitter))
            ).flux();
```
It is executing the route on a different thread than the filters and causing lost context.

I have tried different approaches, but this one looks like the only way.

Fixes https://github.com/micronaut-projects/micronaut-core/issues/6113

Additionally we can also introduce Reactor context-aware `ReactiveInvocationInstrumenterFactory` and `InvocationInstrumenter` and for example to have `ServerRequestContextInvocationInstrumenter` which would store the request in the context.

Plus, I think it would be nice to propagate the reactor context to the coroutine context by default if `kotlinx-coroutines-reactor` is present on the classpath.


graemerocher:
@dstepanov maybe those last two items you could send a PR for 3.1 to address? Those do seem like good improvements

dstepanov:
I will create separate PRs

graemerocher (approved):

PR #6186, opened 2021-09-21
docs: clarify return values in isEmpty/isNotEmpty/hasText

TimothyJones:
Firstly, thanks for all your hard work on Micronaut! It's much appreciated.

I noticed that the `StringUtils` javadoc isn't clear about how nulls are handled, so I ended up reading the source for the version we're using.

This PR makes the return values clearer in the documentation so that reading the source isn't necessary in the future.


TimothyJones:
The build failures seem unrelated to this PR, but please let me know if you need changes.

jameskleeh:
Thanks!

PR #6194, opened 2021-09-21
#6181 Handle exception thrown from http filter.

### LINKED ISSUE #6181: Request is blocking when filter throw exception
### Expected Behavior

image

image

image

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

windows 10
adopt11-hotspot

### Example Application

_No response_

### Version

3.0.1

ashishkujoy:
When the exception thrown from filter is not handle, the request does not complete and client get timeout.
Wrapping the filter call in `Flux.defer` wrap the exception thrown from function call into a Flux which
completes with error signal and `onErrorResume` then handle that.

closes: #6181

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -58,10 +58,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
@@ -340,13 +337,16 @@ public final class RouteExecutor {
                     return upstreamResponsePublisher;
                 }
                 HttpFilter httpFilter = filters.get(pos);
-                return Flux.defer(() -> (Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.getAndSet(request), this))
+
+                return wrapAsFlux(() ->
+                        (Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.getAndSet(request), this)
+                )
                         .flatMap(handleStatusException)
                         .onErrorResume(onError);
             }
         };
         HttpFilter httpFilter = filters.get(0);
-        return Flux.defer(() -> (Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.get(), filterChain))
+        return wrapAsFlux(() -> (Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.get(), filterChain))
                 .flatMap(handleStatusException)
                 .onErrorResume(onError);
     }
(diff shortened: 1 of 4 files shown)


http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -343,13 +340,13 @@ public MediaType resolveDefaultResponseContentType(HttpRequest<?> request, Route
                     return upstreamResponsePublisher;
                 }
                 HttpFilter httpFilter = filters.get(pos);
-                return Flux.from((Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.getAndSet(request), this))
+                return Flux.defer(() -> (Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.getAndSet(request), this))
```
ashishkujoy: Using `Flux.defer` here somehow changing the thread pool which is getting used while subscription of elements from `default-nioEventLoopGroup` to `io-executor-thread-1` resulting in test failure, I looked into java doc of `defer` it does not mentions anything about change in execution threads.

Another, but ugly way to fix this is, please do let me know if there is a better way to handle this.
```kotlin
Flux<MutableHttpResponse<?>> responseFromFilter;
try {
    responseFromFilter = Flux.from((Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.getAndSet(request), this));
} catch (Throwable t) {
    responseFromFilter = Flux.error(t);
}
return responseFromFilter
        .flatMap(handleStatusException)
        .onErrorResume(onError);
```

jameskleeh: Catching the exception seems reasonable to me. You would just call `onError` though instead of applying the flatMap and onErrorResume

ashishkujoy: The `flatMap` handling was present before my change, I think it is required, as it is handling the response of status code above `400`, while the `onErrorResponse` is handling other errors.

I added the commit to handle error and now test are passing.
Please trigger the github action.

jameskleeh: Handling a status code requires a response which we don't  have, so the onError is the only thing that needs to be called if the filter throws an exception. You can just return `onError.call(t)`

ashishkujoy: In `flatMap` we do get `response`.
The try catch was added to handle the exception thrown from(not the error signal which is getting handled by onError)
`(Publisher<MutableHttpResponse<?>>) httpFilter.doFilter(requestReference.getAndSet(request), this)`

`Flux.from` needs a publisher but if the above function call fails as mentioned in the issue card no `Flux` was getting created hence i add to make sure a `Flux` is getting created.

jameskleeh: flatMap would never be called because it processes emitted items and in this case there is no item being emitted

ashishkujoy: Okay so you mean if status `400` and above the publisher emit error signal instead of response as next element signal.

jameskleeh: No - I mean `Flux.error(t).flatMap(() -> this never gets called)`

ashishkujoy: Oh understood, implemented the suggested feedback.

jameskleeh:
Superseded by #6208

PR #6201, opened 2021-09-22
Support for repeatable qualifiers

graemerocher:
* Optimizes named qualifier creation
* Optimizes annotation qualifier creation
* Optimizes composite qualifier creation
* Adds support for repeatable qualifiers

Repeatable qualifiers are a CDI requirement https://github.com/micronaut-projects/micronaut-core/issues/4658

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
@@ -32,6 +32,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.reflect.Array;
 import java.util.*;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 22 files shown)


dstepanov:
Can we have a simple test that checks that it's possible to read repeatable annotations from `InjectionPoint`?

graemerocher:
There is no `getQualifier()` method on `InjectionPoint`?

dstepanov:
I'm thinking of a scenario where would you like to extract all variations of `@Location(...)` and inject something accordingly.

graemerocher:
There is no general API today for doing detecting the type of qualifier and materializing a `Qualifier` instance from the metadata not for repeatable annotations, nor for any other type of qualifier.

Any code that uses the `InjectionPoint` to materialize qualifiers would have to current explicitly handle each case accordingly. I don't see that as in scope for this PR, perhaps it belongs as part of the repeatable annotation in metadata PR but not this one.

dstepanov (approved):

PR #6206, opened 2021-09-23
Applying @Format to query param duplicated in URI

### LINKED ISSUE #5223: @Format is not applied to query parameter in declarative http client if uri declares it
### Steps to Reproduce

Consider the following setup:

```java
public interface CurrencyAPI
{
    @Get("/currency/rates/range{?to,from}")
    Mono<Map<LocalDateTime, Map<String, BigDecimal>>> getRates(
        @QueryValue("from") @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime from,
        @QueryValue("to") @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime to);
}

@Client("currency-service")
public interface CurrencyClient extends CurrencyAPI
{
    @Override
    Mono<Map<LocalDateTime, Map<String, BigDecimal>>> getRates(
        @QueryValue("from") @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime from,
        @QueryValue("to") @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime to);
}

@Controller
public class CurrencyController implements CurrencyAPI
{
    @Override
    public Mono<Map<LocalDateTime, Map<String, BigDecimal>>> getRates(
        @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime from,
        @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime to)
    {
        return null;
    }
}
```

### Expected Behaviour

Call `currencyClient.getRates()` should complete successfully

### Actual Behaviour

Call `currencyClient.getRates()` errors out with parsing exception on server side, because the parameter is formatted by default formatter.

This happens because of logic in `HttpClientIntroductionAdvice#intercept`, lines 289-290
```
uri = uriTemplate.expand(paramMap);
uriVariables.forEach(queryParams::remove);
```

### Workaround

Duplicate the route annotation from the `*API` class without query params defined:
```java
@Client("currency-service")
public interface CurrencyClient extends CurrencyAPI
{
    @Override
    @Get("/currency/rates/range")
    Mono<Map<LocalDateTime, Map<String, BigDecimal>>> getRates(
        @QueryValue("from") @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime from,
        @QueryValue("to") @Format("yyyy-MM-dd'T'HH:mm:ss.SSSSSS") LocalDateTime to);
}
``` 

### Environment 
Micronaut version 2.4.2

GavrilovSV:
fixes #5223

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
@@ -24,6 +24,7 @@ import io.micronaut.core.beans.BeanProperty;
 import io.micronaut.core.bind.annotation.Bindable;
 import io.micronaut.core.convert.ConversionContext;
 import io.micronaut.core.convert.ConversionService;
+import io.micronaut.core.convert.format.Format;
 import io.micronaut.core.naming.NameUtils;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.util.CollectionUtils;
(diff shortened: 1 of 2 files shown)


jameskleeh:
Your test would pass without the changes you made because the format chosen was the same as the output of LDT.toString(). I also removed the check for `@Format` since that didn't seem relevant. Superseded by #6209

PR #6212, opened 2021-09-24
Add link for CLA agreement

ashishkujoy:
CLA is a mandatory step for doing any code contribution to micronaut, but this information is not mentioned in CONTRIBUTING.md, I think it will be better for new contributors to know this upfront so that it does not come up as surprise when they submit pull request.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # Contributing Code or Documentation to Micronaut
 
+Sign the [Contributor License Agreement (CLA)](https://cla-assistant.io/micronaut-projects/micronaut-core). This is required before any of your code or pull-requests are accepted.
+
 ## Finding Issues to Work on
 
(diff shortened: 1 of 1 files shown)


graemerocher:
Perhaps worth updating the project template as well https://github.com/micronaut-projects/micronaut-project-template/blob/master/CONTRIBUTING.md

ashishkujoy:
I added in the project template and realized that CLA is per repo base, so may we need to some github action will trigger when some create a new project from template, that GitHub action will update the CLA link.

ashishkujoy:
Jetbrains does something similar around this for plugin templates. https://github.com/JetBrains/intellij-platform-plugin-template/blob/main/.github/workflows/template-cleanup.yml 

Maybe i can add something similar to update CLA url for each new project created with that template.

ashishkujoy:
Added GitHub action to update CLA URL for each new project created using the template. 
https://github.com/micronaut-projects/micronaut-project-template/pull/142

ashishkujoy:
Is this good to merge, or are is there is any feedback on this to be implemented?

PR #6213, opened 2021-09-24
Add ElementType.ANNOTATION_TYPE to all tracing annotations

ajtuss:
Add `ElementType.ANNOTATION_TYPE` to all tracing annotations. Discussion: #6207


graemerocher:
Thanks for the contribution!

PR #6223, opened 2021-09-24
Store repeatable annotations info and avoid reflection call

dstepanov:
`DefaultAnnotationMetadata` needs to have a flag for backward compatibility


graemerocher:
Rather than the additional flag you can just overrride the new methods in `AnnotationMetadataWriter` for newly compiled byte code

dstepanov:
If I understand it correctly, that would create a new class per `DefaultAnnotationMetadata` which I don't think is a good idea

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1488,6 +1493,47 @@ default boolean hasDeclaredStereotype(@Nullable Class<? extends Annotation>... a
         return false;
     }
 
+    /**
+     * Is repeatable annotation?
+     * @param annotation The annotation
+     * @return true if repeatable
+     * @since 3.1
+     */
+    default boolean isRepeatableAnnotation(Class<? extends Annotation> annotation) {
```
graemerocher: Add nullability annotations to `annotation` parameter and if it can be `null` add null check or return false when `null`

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1459,9 +1464,9 @@ default boolean hasDeclaredAnnotation(@Nullable Class<? extends Annotation> anno
      */
     default boolean hasDeclaredStereotype(@Nullable Class<? extends Annotation> stereotype) {
         if (stereotype != null) {
-            Repeatable repeatable = stereotype.getAnnotation(Repeatable.class);
-            if (repeatable != null) {
-                return hasDeclaredStereotype(repeatable.value().getName());
+            Optional<String> repeatable = findRepeatableAnnotation(stereotype);
```
graemerocher: For these internal calls maybe we can avoid optional wrapping?

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1488,6 +1493,47 @@ default boolean hasDeclaredStereotype(@Nullable Class<? extends Annotation>... a
...
+    }
+
+    /**
+     * Find repeatable annotation container.
+     * @param annotation The annotation
+     * @return optional repeatable annotation container
+     * @since 3.1
+     */
+    default Optional<String> findRepeatableAnnotation(Class<? extends Annotation> annotation) {
+        return Optional.ofNullable(annotation.getAnnotation(Repeatable.class))
```
graemerocher, line 1519: Add nullability annotations to `annotation` parameter and if it can be `null` add null check or return `Optional.empty()` when `null`

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1488,6 +1493,47 @@ default boolean hasDeclaredStereotype(@Nullable Class<? extends Annotation>... a
...
+                .map(repeatable -> repeatable.value().getName());
+    }
+
+    /**
+     * Find repeatable annotation container.
+     * @param annotation The annotation
+     * @return optional repeatable annotation container
+     * @since 3.1
+     */
+    default Optional<String> findRepeatableAnnotation(String annotation) {
```
graemerocher: Add nullability annotations to `annotation` parameter

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
```diff
@@ -102,6 +121,14 @@
         return ANNOTATION_DEFAULTS.getOrDefault(annotation, Collections.emptyMap());
     }
 
+    /**
+     * @param annotation The annotation
+     * @return The repeatable annotation container.
+     */
+    public static String getRepeatableAnnotation(String annotation) {
```
graemerocher, line 129: Annotate with `@Internal`

inject/src/main/java/io/micronaut/inject/annotation/DefaultAnnotationMetadata.java
```diff
@@ -1934,18 +1999,27 @@ public static void contributeDefaults(AnnotationMetadata target, AnnotationMetad
...
-                final Map<String, Map<CharSequence, Object>> additionalDefaults = ((DefaultAnnotationMetadata) source).annotationDefaultValues;
+                final Map<String, Map<CharSequence, Object>> additionalDefaults = damSource.annotationDefaultValues;
                 if (additionalDefaults != null) {
-                    additionalDefaults.forEach(((DefaultAnnotationMetadata) target)::addDefaultAnnotationValues);
+                    additionalDefaults.forEach(damTarget::addDefaultAnnotationValues);
+                }
+            }
+            if (damSource.repeated != null && !damSource.repeated.isEmpty()) {
+                if (damTarget.repeated == null) {
+                    damTarget.repeated = new HashMap<>(damSource.repeated);
```
graemerocher, line 2020: should be a linked hashmap to retain order?

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
```diff
@@ -413,6 +422,17 @@ public static void writeAnnotationDefaults(
                 }
 //                staticInit.visitLabel(falseCondition);
             }
+            if (annotationMetadata.repeated != null && !annotationMetadata.repeated.isEmpty()) {
+                Map<String, String> repeated = new HashMap<>();
```
graemerocher, line 426: Should be linked to retain order and incremental compile?

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
```diff
@@ -213,6 +239,26 @@ static void registerAnnotationType(AnnotationClassValue<?> annotationClassValue)
         }
     }
 
+    /**
+     * Registers repeatable annotations.
+     *
+     * @param repeatableAnnotations the repeatable annotations
+     */
+    static void registerRepeatableAnnotations(Map<String, String> repeatableAnnotations) {
```
graemerocher, line 249: Annotate with `@Internal`

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
```diff
@@ -213,6 +239,26 @@ static void registerAnnotationType(AnnotationClassValue<?> annotationClassValue)
...
+        REPEATABLE_ANNOTATIONS.putAll(repeatableAnnotations);
+    }
+
+    /**
+     * Remove Core repeatable annotations.
+     *
+     * @param repeatableAnnotations the repeatable annotations
+     */
+    static void removeCoreRepeatableAnnotations(@NotNull Map<String, String> repeatableAnnotations) {
```
graemerocher, line 259: Annotate with `@Internal`

graemerocher (changes requested):

graemerocher:
yeah good point, I guess the flag is needed, spotted a few other things

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
dstepanov: It's a bit problematic as it's an interface and we cannot have private methods. I can call `isRepeatableAnnotation` before getting it considering in most cases it's probably not repeatable. WDYT?

inject/src/main/java/io/micronaut/inject/annotation/DefaultAnnotationMetadata.java
dstepanov: No, it's converted to sorted string array

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
dstepanov: No, it's converted to a sorted string array

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
graemerocher: Sounds good

PR #6224, opened 2021-09-24
Look through interfaces for pre destroy methods

### LINKED ISSUE #6149: @Bean(preDestroy=..) broken for Groovy and interfaces
### Expected Behavior

Given interface:

```groovy
interface Foo extends AutoCloseable {}
```

The following should work:

```groovy
@Factory
class FooFactory {
    @Bean(preDestroy="close")
     Foo foo() {
         ...
     }
}
```

### Actual Behaviour

Fails with "@Bean method defines a preDestroy method that does not exist" compilation error because `ClassNode.getMethod(..)` in the Groovy AST only traverses super classes not super interfaces

### Steps To Reproduce

See above

### Environment Information

All

### Example Application

N/A

### Version

3.0.1

jameskleeh:
Fixes #6149


graemerocher:
Needs to check the method is zero arguments

PR #6227, opened 2021-09-27
Improve Bean Introspection compilation

dstepanov:
I have analyzed the new introspection classes in Micronaut Data and come up with a few improvements:
- Introduce `propertyIndexOf` that is well optimized by the generated class and can be reused for direct property mapping.
- Improve and extract string switch writer
- Store static indexes

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/writer/StringSwitchWriter.java
+++ b/inject/src/main/java/io/micronaut/inject/writer/StringSwitchWriter.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


inject/src/main/java/io/micronaut/inject/writer/StringSwitchWriter.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * String switch writer.
+ *
+ * @author Denis Stepanov
+ * @since 3.1
+ */
+public abstract class StringSwitchWriter {
```
graemerocher, line 38: Annotate with `@Internal`

graemerocher (changes requested):

PR #6230, opened 2021-09-27
Expose more generics information in ClassElement API

yawkat:
This PR adds an API to `ClassElement` that exposes more information about the generics declared in a source file.

The first large addition is a `getBoundTypeArguments` method to `ClassElement`. It differs from `getTypeArguments` mostly in that it does not respect the `genericsInfo` of the enclosing type. Also see the javadoc of the new method. 

---

As an aside: Reusing `getTypeArguments` was not possible, because it does some weird things with erasing wildcards and type arguments. For example, I found this bug in generics handling:

```groovy
    def 'test'() {
        given:
        def element = buildClassElement("""
package example;

import java.util.*;

class Wrapper<U extends Integer> {
    Test<List<U>> foo;
}
class Test<T> {
    List<T> bar;
}
""")
        def field = element.fields[0].genericType.fields[0]

        expect:
        field.genericType.getFirstTypeArgument().get().name == 'java.lang.Integer' // should be java.util.List
    }
```

I think it may be possible to unify these two approaches to generics (`getTypeArguments` and `getBoundTypeArguments`) at some point, but it would require changing some of the `getTypeArguments` behavior and would make guaranteeing compatibility difficult. Maybe in 4.x.

---

As it stands, the data backing `getBoundTypeArguments` has no influence on the existing generics handling, ensuring compatibility.

The other big API addition is two subinterfaces of `ClassElement`, `FreeTypeVariableElement` and `WildcardElement`. These interfaces expose additional information about type variables and wildcards respectively, while still maintaining the old erasure behavior when used as `ClassElement`s. For example, a type variable exposes all its bounds using the new `getBounds` method, but the super `JavaClassElement` constructor is called with the first bound, matching previous behavior.

There is one incompatibility here that comes up in the changed test case of `BeanDefinitionSpec`. Previously, multiple bounds on a type variable would lead to the variable being replaced by `Object`. Now, we replace it with the first bound, which is still correct.

The final addition is a few "copy mutators", similar to the existing `withArrayDimensions` methods:

- A method to get the raw type
- A method to get a new parameterized version of a type
- A method to recursively run a transform op on a type ("fold")

The reflective `ClassElement` implementations have been expanded to support these new types as well. Groovy support is not yet implemented, due to time constraints.


dstepanov:
Can this be used to check if method parameter is: `Event<?>`, `Event<Object>` or `Event`?

yawkat:
> Can this be used to check if method parameter is: `Event<?>`, `Event<Object>` or `Event`?

Yes. The `getBoundTypeArguments` is: `WildcardElement` with a single upper bound of Object, `ClassElement` for `Object`, and an empty list of type arguments, respectively.

graemerocher (approved):

graemerocher:
@dstepanov can you have a look through the latest changes and provide your thoughts

yawkat:
Changes look good to me.

graemerocher:
@yawkat whilst the reviews are ongoing please progress forward Groovy support

dstepanov:
Can you please add a test case that would use a type visitor and distinguish `SomeType<?>`, `SomeType<Object>` or `SomeType`, you might have it in the existing test suite but I would prefer for test-cases to use real API.

yawkat:
@dstepanov I added a short test case to `JavaReconstructionSpec` for that.

dstepanov (approved):

graemerocher:
I don't see tests for the Groovy code that was added

yawkat:
There's a GroovyReconstructionSpec that is basically a copy of the java one

graemerocher (approved):

graemerocher:
@jameskleeh this is ready for a final review, it has been reviewed by myself and @dstepanov 

It is needed for Micronaut Data, the CDI impl and the upcoming JSON impl

PR #6234, opened 2021-09-27
Attempt to load all possible property sources to determine if the bootstrap env should be started

### LINKED ISSUE #6066: Document bootstrap.yml

### LINKED ISSUE #6226: bootstrap.yaml not recognized
### Expected Behavior

whatever the file extension name (yaml or yml), the file should be loaded

### Actual Behaviour

When the bootstrap file extension is **yaml**, it is not recognized and so not loaded when starting app

### Steps To Reproduce

1. Create a simple micronaut project
2. add a _bootstrap.yml_ file
3. launch the app and see that the file is recognized

> [main] INFO  i.m.context.DefaultBeanContext - Reading Startup environment from bootstrap.yml

4. rename the file from _bootstrap.yml_ to _bootstrap.yaml_
5. see that the bootstrap is not taken into account

### Environment Information

- Operating System: MacOS
- Micronaut Version: 3.0.1
- JDK Version: 11
- gradle 7.2
- micronaut-gradle-plugin 2.0.4 

### Example Application

_No response_

### Version

3.0.1

jameskleeh:
Fixes #6226 
Fixes #6066 

This should attempt to load all variants of `bootstrap.{ext}` and `bootstrap-{env}.{ext}`. If any are found then the bootstrap environment will be started. The loading of those files is then cached so it won't be done twice. This fixes the hack that was in place that only checked for `bootstrap.yml` and `bootstrap.properties`.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
@@ -695,10 +695,12 @@ public class DefaultApplicationContext extends DefaultBeanContext implements App
         private final ApplicationContextConfiguration configuration;
         private BootstrapPropertySourceLocator bootstrapPropertySourceLocator;
         private BootstrapEnvironment bootstrapEnvironment;
+        private boolean bootstrapEnabled;
 
         RuntimeConfiguredEnvironment(ApplicationContextConfiguration configuration) {
             super(configuration);
             this.configuration = configuration;
+            this.bootstrapEnabled = configuration.isBootstrapEnvironmentEnabled();
         }
 
         @Override
@@ -713,10 +715,17 @@ public class DefaultApplicationContext extends DefaultBeanContext implements App
         public Environment start() {
             String bootstrapContextProp = System.getProperty(Environment.BOOTSTRAP_CONTEXT_PROPERTY);
             Boolean bootstrapEnabled = bootstrapContextProp == null ? null : Boolean.parseBoolean(bootstrapContextProp);
-            if (bootstrapEnvironment == null && (bootstrapEnabled == null || bootstrapEnabled)) {
+            if (bootstrapEnabled == null) {
+                bootstrapEnabled = this.bootstrapEnabled;
+            }
+            if (bootstrapEnvironment == null && bootstrapEnabled) {
                 BootstrapEnvironment bootstrapEnv = createBootstrapEnvironment(getActiveNames().toArray(new String[0]));
-                if (bootstrapEnabled != null || CollectionUtils.isNotEmpty(bootstrapEnv.readPropertySourceList(bootstrapEnv.getPropertySourceRootName()))) {
+                if (CollectionUtils.isNotEmpty(bootstrapEnv.readPropertySourceList(bootstrapEnv.getPropertySourceRootName()))) {
                     bootstrapEnvironment = startBootstrapEnvironment(bootstrapEnv);
+                } else {
+                    if (LOG.isInfoEnabled()) {
+                        LOG.info("The bootstrap environment is enabled but no local property sources were found. Consider disabling the bootstrap environment to gain startup time.");
+                    }
                 }
             }
             return super.start();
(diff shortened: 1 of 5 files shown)


graemerocher:
What is the impact on startup time?

jameskleeh:
For the case where the bootstrap env should be loaded, no impact.

For the case where it should not, the difference is 2 classpath lookups vs 3 * (number of environments + 1). So for the case of a single environment, 6 lookups.

graemerocher:
I am unconvinced it is worth the cost. Perhaps something to be discussed at the team meeting

jameskleeh:
I did a test by calculating the time it takes to do `resourceLoader.loadResourceAsStream`. Each call on my machine takes approximately half of a millisecond, so even if we consider say 3 environments and the json loader is in play that would work out to 10ms.

graemerocher:
That is on fast developer hardware with SSDs, that 10ms could be 100ms or more in certain environments. I would prefer to roll a fix for this into the work @melix is doing to pre-compute at build time what the possible files are to be loaded rather than trying to work it out again at runtime.

graemerocher:
Since we didn't get to finish the conversation on this topic, I will clarify my position. 

1. The "hack" that this PR "fixes" was intentionally put in place to avoid the additional I/O lookups that hurt startup time and has been there since Micronaut 1.0, so I am not buying that this is an urgent issue that needs a fast, potentially poorly thought out fix from a performance perspective. 
2. There are also multiple existing ways to solve the problem, since you can use environment variables to supply the config without even needing bootstrap.yml. The poor user experience can be mitigated by better documentation.
3. Anything that degrades startup time is a non-starter. What is 10ms on fast developer hardware could be 100ms or more in more constrained environments. We already had to deal with the cluster of environments with lots of environment variables and k8s and the last thing we need is to be explaining to people how to fix performance issues with special flags.
4. Any potential fix for this issue has to be opt-in and not automatic behavior that is documented 

Something simple which is documented like:

```java
Micronaut.builder()
     .useBootstrap(Environments.AWS, "bootstrap-aws.yml")
     .useDefaultBootstrap("bootstrap.yaml")
     .start();
```

Would suffice to provide a solution that doesn't degrade performance and simply needs to be documented in an appropriate manner.

Performance is a key metric developers are looking at when comparing Micronaut to the competition and any change that is not a performance improvement cannot be merged and I am vehemently against merging this PR in its current state.

graemerocher (changes requested):
See feedback above

jameskleeh:
It's probably better to refer to the numbers as a percentage of the current startup time, since that should be relatively constant across machines. On my pc an application starts up in 670ms. 10ms in addition to that represents a 1.4% increase.

Unfortunately due to the current behavior being automatically detecting the files, we can't move away from that without introducing a breaking change. We can fix the current behavior and also make it easier to disable the lookups and produce an info message to users that they should do so if no bootstrap files are found.

graemerocher:
James there is nothing broken today. The current implementation is working as designed and is merely lacking documentation which Burt provided a PR for and this is nothing more than a "improvement" that may improve on the behaviour, but introduces slowdown the the startup time at a time when we are heavily focused on startup time optimizations.

A "fix" that involves a message that is always printed out is also not acceptable IMO.

I suggest we discuss this and vote at the next engineering meeting since the decision is not yours alone.

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -712,18 +713,29 @@ public Environment stop() {
...
+            if (bootstrapEnabled == null) {
+                bootstrapEnabled = this.bootstrapEnabled;
+            }
+            if (bootstrapEnvironment == null && (bootstrapEnabled == null || bootstrapEnabled)) {
+                BootstrapEnvironment bootstrapEnv = createBootstrapEnvironment(getActiveNames().toArray(new String[0]));
+                if (bootstrapEnabled != null || CollectionUtils.isNotEmpty(bootstrapEnv.readPropertySourceList(bootstrapEnv.getPropertySourceRootName()))) {
+                    bootstrapEnvironment = startBootstrapEnvironment(bootstrapEnv);
+                } else {
+                    if (LOG.isInfoEnabled()) {
+                        LOG.info("The bootstrap environment is enabled but no local property sources were found. Consider disabling the bootstrap environment to gain startup time.");
```
graemerocher, line 727: The message needs a link to the reference documentation that explains how this can be disabled. From a user perspective most users probably don't even know what the boostrap environment is let alone how to disable it.

graemerocher (changes requested):

PR #6235, opened 2021-09-27
Refactors defaultBeanContext

sapanparikh18:
@graemerocher @ilopmar, will you be open to IDE-based automated refactoring like this? The names selected for the extracted methods may not be best suited, but in principle, they should have high RoI later.
DefaultBeanContext.java has a few methods with very high cognitive complexity plus is marked as a hotspot. A hotspot is a class that sees a lot of activity + is complex + it's big. The number of bugs found in the code is a function of these parameters. So if you can make it simple, smaller, and give fewer reasons to change, it should have lesser bugs in the future.


jameskleeh:
Sorry we haven't responded to this, however I don't think we should refactor without some need to do so. The code in question may have high complexity however it is tested more than any other code in the repo

PR #6243, opened 2021-09-29
feat: disable endpoints filter via configuration

sdelamo:
This is useful if you want to avoid loading the endpoints filter because you are going to handle security with your own filter.

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/EndpointsFilter.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/EndpointsFilter.java
@@ -15,7 +15,9 @@
  */
 package io.micronaut.management.endpoint;
 
+import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.async.publisher.Publishers;
+import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.HttpRequest;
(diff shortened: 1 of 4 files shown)


graemerocher:
I don't think we need a whole new configuration properties bean just for this, also it is not even used in the code.

sdelamo:
The advantage of having a `@ConfigurationProperties` file is that it generates documentation automatically. Thus, users are able to discover this feature.

graemerocher:
There is already https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java which can be reused rather than adding a whole new configuration for 1 property

sdelamo:
I thought about adding it to `EndpointDefaultConfiguration`. However, I thought the prefix `endpoints.all` may confuse users. I think `endpoints.filter.enabled` is nicer than `endpoints.all.filter-enabled` or `endpoints.all.filter` .

jameskleeh:
We already have a way to effectively disable the filter and adding this configuration creates two ways to do the same thing. For most users the fact that this is done via a filter is an implementation detail so adding the config exposes the implementation. I don't see an issue with the current process so I think we should close this

PR #6247, opened 2021-09-29
Allow socket address converter to handle urls with schemes

### LINKED ISSUE #5797: Proxy address fails w/ http URLs
When setting up a proxy for a given micronaut HttpClient, I tested using a full URL ('http://...') instead of an IP address or direct domain name (without http/https).

Example:

```
micronaut:
    http:
        services:
            vuln-db:
                url: ${myApp.baseUrl}
                proxy-type: HTTP                    
                proxy-address: 'http://www.my-app.com:80'
```

The application fails to start with a message saying the proxyAddress property cannot be found.

After changing to the IP address for the given domain, the proxyAddress setting worked and the proxy functioned as expected without issue.

jameskleeh:
Fixes #5797


jameskleeh:
Closing this because I got the commits messed up

PR #6248, opened 2021-09-30
Updates toString of CronExpression to return the underlying cron expression as string.

### LINKED ISSUE #5520: Getter for CronExpression.expr
I notice that there is no getter for `CronExpression.expr`. I think I expected `toString()` to return the original expression, but it returns `getClass().getSimpleName() + "<" + expr + ">"` instead.

ashishkujoy:
closes: #5520 


graemerocher:
Changing the `toString()` is perhaps a breaking change in a patch release. To solve this we would need to add an access method I fear.

ashishkujoy:
> Changing the `toString()` is perhaps a breaking change in a patch release. To solve this we would need to add an access method I fear.

Looking at original implementation of `toString()` I doubt anyone would be using that.

But to be on safer side I will make change to keep the `toString()` as it is and will add a getter for the expression.

jameskleeh:
This should target 3.1.x. Please merge that branch into yours

ashishkujoy:
closing this PR will create a new one targeting 3.1.x

ashishkujoy:
Opened: #6258

PR #6255, opened 2021-09-30
Delegate overridden method in AOP proxy

dstepanov:
This changed the AOP class to skip executable method for overridden method and delegate to the override just like it's generated by Javac:

```java
public class CustomCrudRepoImpl implements CustomCrudRepo {
   public Optional<String> findById(Long aLong) {
      return Optional.empty();
   }

   // $FF: synthetic method
   // $FF: bridge method
   public Optional findById(Object var1) {
      return this.findById((Long)var1);
   }
}
```

Fixes issues where the behavior was different when repositories were accessed by the superclass.

Implemented only for Java. I don't know how to implement overridden checks for Groovy methods.

Fixes https://github.com/micronaut-projects/micronaut-data/issues/1186

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -528,15 +528,13 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
     public void visitAroundMethod(TypedElement beanType,
                                   MethodElement methodElement) {
 
-        final List<MethodElement> overridden = methodElement.getOwningType()
-                .getEnclosedElements(ElementQuery.ALL_METHODS
-                        .filter(el -> el.getName().equals(methodElement.getName()) && el.overrides(methodElement)));
+        final Optional<MethodElement> overridden = methodElement.getOwningType()
+                .getEnclosedElement(ElementQuery.ALL_METHODS
+                        .named(name -> name.equals(methodElement.getName()))
+                        .filter(el -> el.overrides(methodElement)));
 
-        if (!overridden.isEmpty()) {
-            if (overridden.size() != 1) {
-                throw new IllegalStateException("Expected exactly one overridden method!");
-            }
-            MethodElement overriddenBy = overridden.iterator().next();
+        if (!overridden.isPresent()) {
+            MethodElement overriddenBy = overridden.get();
 
             String methodElementKey = methodElement.getName() +
                     Arrays.stream(methodElement.getSuspendParameters())
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -530,6 +530,7 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
 
         final Optional<MethodElement> overridden = methodElement.getOwningType()
                 .getEnclosedElement(ElementQuery.ALL_METHODS
+                        .onlyInstance()
                         .named(name -> name.equals(methodElement.getName()))
                         .filter(el -> el.overrides(methodElement)));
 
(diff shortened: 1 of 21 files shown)


aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
```diff
@@ -527,6 +528,31 @@ public void visitIntroductionMethod(TypedElement declaringBean,
     public void visitAroundMethod(TypedElement beanType,
                                   MethodElement methodElement) {
 
+        final List<MethodElement> overridden = methodElement.getOwningType()
+                .getEnclosedElements(ElementQuery.ALL_METHODS
+                        .filter(el -> el.getName().equals(methodElement.getName()) && el.overrides(methodElement)));
```
graemerocher: add `.named(n -> n.equals(methodElement.getName())` instead of the 2 conditions in the filter since it will be more efficient as it won't need to materialize the `MethodElement`

aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
```diff
@@ -527,6 +528,31 @@ public void visitIntroductionMethod(TypedElement declaringBean,
     public void visitAroundMethod(TypedElement beanType,
                                   MethodElement methodElement) {
 
+        final List<MethodElement> overridden = methodElement.getOwningType()
+                .getEnclosedElements(ElementQuery.ALL_METHODS
```
graemerocher: You can use `getEnclosedElement` instead of `getEnclosedElements` which returns an `Optional` that would simplify the code

inject/src/main/java/io/micronaut/inject/ast/MethodElement.java
```diff
@@ -120,6 +120,16 @@ default boolean isDefault() {
         return typeString + " " + getName() + "(" + args + ")";
     }
 
+    /**
+     * Checks if this method element overrides another.
+     * @param overridden Possible overridden method
+     * @return true if this overrides passed method element
+     * @since 3.1
+     */
+    default boolean overrides(MethodElement overridden) {
```
graemerocher: Add `@NonNull` to `overridden` parameter

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
```diff
@@ -75,6 +75,15 @@ public boolean isDefault() {
         return executableElement.isDefault();
     }
 
+    @Override
+    public boolean overrides(MethodElement methodElement) {
+        return visitorContext.getElements().overrides(
```
graemerocher: surround with `instanceof JavaMethodElement` to avoid case where `null` is passed and or a non `JavaMethodElement`

graemerocher (changes requested):

graemerocher:
Can you link the Micronaut Data issue this fixes as well?

dstepanov:
Looks like Groovy doesn't have the same problem

graemerocher (approved):

aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
```diff
@@ -527,6 +528,29 @@ public void visitIntroductionMethod(TypedElement declaringBean,
     public void visitAroundMethod(TypedElement beanType,
                                   MethodElement methodElement) {
 
+        final Optional<MethodElement> overridden = methodElement.getOwningType()
+                .getEnclosedElement(ElementQuery.ALL_METHODS
+                        .named(name -> name.equals(methodElement.getName()))
```
graemerocher, line 534: Add `onlyInstance()` otherwise this will search static methods

graemerocher (changes requested):
Sorry noticed one other thing

graemerocher (approved):

PR #6257, opened 2021-09-30
Bump micronaut-kubernetes to 3.1.0

pgressa:


pgressa:
@jameskleeh I'm not able to judge what to keep in the whatsNew. So here are few lines on the informer.

PR #6258, opened 2021-10-01
Adds a getter for cron expression

ashishkujoy:
closes: #5520 

### DIFF
--- a/context/src/main/java/io/micronaut/scheduling/cron/CronExpression.java
+++ b/context/src/main/java/io/micronaut/scheduling/cron/CronExpression.java
@@ -290,6 +290,15 @@ public final class CronExpression {
         }
     }
 
+    /**
+     * Returns String expression.
+     *
+     * @return The underlying cron expression as string.
+     */
+    public String getExpression() {
+        return expr;
+    }
+
     @Override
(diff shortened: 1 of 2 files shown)


context/src/main/java/io/micronaut/scheduling/cron/CronExpression.java
```diff
@@ -290,6 +290,15 @@ private static void checkIfDateTimeBarrierIsReached(ZonedDateTime nextTime, Zone
         }
     }
 
+    /**
+     * Returns String expression.
+     *
+     * @return The underlying cron expression as string.
```
graemerocher, line 297: Please add `@since 3.1.0` since this is a new method

graemerocher (changes requested):

context/src/main/java/io/micronaut/scheduling/cron/CronExpression.java
ashishkujoy: done, @graemerocher If it is good to merge, can u please add `hacktoberfest-accepted` label to this PR

jameskleeh:
Thanks!

PR #6268, opened 2021-10-04
Introduce an index in immutable sorted maps

melix:
The `ImmutableSortedStringsArrayMap` implementation is more memory
efficient than a traditional hashmap, but key lookups are typically
slower, because they rely on a binary search, which is O(log(n))
where in a hash map lookup would be O(1).

When optimizing native images, it became obvious that this lookup
was causing some slowdowns in bean initialization because of many
calls to `hasStereotype`. Profiling shows that most of those calls
actually fail, which means that we do a binary search for nothing.

Most of the maps are also actually pretty small, with just a few
entries (2 or 3). Therefore, this commit introduces a "bloom-filter"
like idea, where we precompute an index based on the strings hash
codes. For every key in the map, we compute a "reduced" hash which
is from 0 to the size of the index map. Then we store at the reduced
hash position of the index, the index of the key itself.

This allows us to reduce most lookups to constant time lookups: when
a key is searched, we compute the reduced hash and look into the index
if there's a positive value, in which case it means that something
has the same reduced hash. All we need to do then is to check if the
key found at this index is the same. If not, then there are 2 possible
outcomes:

- either the reduced hash for key we search for has a collision with
a key in the map
- or two keys in the map have the same reduced hashes

In both cases, a binary search will determine if the key is present
or not.

Last but not least, this optimization is more efficient when the
index is sparse, that is to say that it can quickly determine if a
key is _not_ in the map. For that reason we construct an index which
size is 8 times the number of entries in the map. Because we use
integers, it means that the cost of this optimization for a map of 3
entries is 3*8*4 (3 ints) + 1 (size of array) = 97 bytes. For a map
of 100 entries, that would be 3200 bytes, which is obviously more but
unlikely to happen and probably less than what a hashmap of the same
size would use.

All in all, this compromise algorith has both reduced memory footprint
_and_ better performance, as I was able to eliminate all binary searches
from the startup of a Micronaut app (in other words the hotspot is gone
with this change).

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/ImmutableSortedStringsArrayMap.java
+++ b/core/src/main/java/io/micronaut/core/annotation/ImmutableSortedStringsArrayMap.java
@@ -44,17 +44,50 @@ import java.util.function.BiConsumer;
 @UsedByGeneratedCode
 final class ImmutableSortedStringsArrayMap<V> implements Map<String, V> {
 
+    private final int[] index;
     private final String[] keys;
     private final Object[] values;
 
     ImmutableSortedStringsArrayMap(String[] keys, Object[] values) {
+        this.index = computeIndex(keys);
         this.keys = keys;
         this.values = values;
+    }
+
+    private static int[] computeIndex(String[] keys) {
+        int len = keys.length;
+        int[] filter = new int[8 * len];
+        Arrays.fill(filter, (byte) -1);
+        for (int i = 0; i < keys.length; i++) {
+            String key = keys[i];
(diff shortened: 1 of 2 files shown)


dstepanov:
Can you identify which `hasStereotype` calls are actually causing it? Most of the bean definition values should be pre-calculated.

dstepanov (approved):

melix:
I didn't check all calls but the ones I saw were checking for `io.micronaut.context.annotation.Requirements` and `io.micronaut.aop.Adapter`

PR #6270, opened 2021-10-04
Forward h2c upgrade request downstream so that it gets a response [Fixes #5005]

### LINKED ISSUE #5005: Micronaut server is not working for clear text http2 (h2c)
I am trying to use h2c protocol on the Micronaut however my endpoint is not responding the request when I use http2. This is my configuration for the h2c

```
micronaut:
  application:
    name: test
  server:
    http-version: 2.0
```    

### Task List

### Steps to Reproduce

Use the sample application from https://github.com/umutkocasarac/micronaut-h2c-test and run the application with **./gradlew run** then make a curl request to the application
```
curl -v  --http2 http://localhost:8080/test
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /test HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
> Connection: Upgrade, HTTP2-Settings
> Upgrade: h2c
> HTTP2-Settings: AAMAAABkAARAAAAAAAIAAAAA
>
< HTTP/1.1 101 Switching Protocols
< connection: upgrade
< upgrade: h2c
* Received 101
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Connection state changed (MAX_CONCURRENT_STREAMS == 4294967295)!
```
or this one
```
curl -v --http2-prior-knowledge  http://localhost:8080/test                                                                                                                                                         
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7fe61a80aa00)
> GET /test HTTP/2
> Host: localhost:8080
> User-Agent: curl/7.64.1
> Accept: */*
>
* Connection state changed (MAX_CONCURRENT_STREAMS == 4294967295)!
```
### Expected Behaviour

It should upgrade the protocol to http2 and then respond the client

### Actual Behaviour

Application freezes after the following log
Connection state changed (MAX_CONCURRENT_STREAMS == 4294967295)!

### Environment Information

- **Micronaut Version:** 2.3.3
- **JDK Version:** openjdk 11.0.8 2020-07-14

### Example Application

https://github.com/umutkocasarac/micronaut-h2c-test

yawkat:
From NettyServerUpgradeHandler: "The upgrade was successful, remove the message from the output list so that it's not propagated to the next handler. This request will be propagated as a user event instead."

So, the first request is discarded. This patch sends it downstream for normal handling.

---

Build is failing locally but that seems unrelated, so we'll see what CI says.

I'm not 100% on whether this patch could duplicate the request, but since it is a `FullHttpRequest` that is being forwarded, it should be safe – no partial data to worry about or anything. The fact that the `upgradeRequest` will still contain headers related to the upgrade may also be a concern, since it allows an attacker to send these headers further downstream than before. But I think that after the upgrade is complete, an attacker could send the same headers anyway.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1129,7 +1129,8 @@ public class DefaultHttpClient implements
                 boolean multipart = MediaType.MULTIPART_FORM_DATA_TYPE.equals(request.getContentType().orElse(null));
                 if (poolMap != null && !multipart) {
                     try {
-                        ChannelPool channelPool = poolMap.get(new RequestKey(requestURI));
+                        RequestKey requestKey = new RequestKey(requestURI);
+                        ChannelPool channelPool = poolMap.get(requestKey);
                         Future<Channel> channelFuture = channelPool.acquire();
                         addInstrumentedListener(channelFuture, future -> {
                             if (future.isSuccess()) {
@@ -1141,6 +1142,7 @@ public class DefaultHttpClient implements
                                             errorType,
                                             emitter,
                                             channel,
+                                            requestKey.isSecure(),
                                             channelPool
                                     );
                                 } catch (Exception e) {
@@ -1177,6 +1179,7 @@ public class DefaultHttpClient implements
                                         errorType,
                                         emitter,
                                         connectionFuture.channel(),
+                                        sslContext != null,
                                         null);
                             } catch (Throwable e) {
                                 emitter.error(e);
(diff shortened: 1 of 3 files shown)


graemerocher:
Can you add to your comments which issue this fixes with. Example `Fixes #...`

graemerocher:
Oh I see it is in the first line of the commit which was a bit too long so wrapped around

yawkat:
Converting this to draft for now. Turns out the client will also require changes. [The spec](https://httpwg.org/specs/rfc7540.html#rfc.section.3.2) says: 

> After the empty line that terminates the 101 response, the server can begin sending HTTP/2 frames. These frames MUST include a response to the request that initiated the upgrade.

This is what my patch does. However, the micronaut h2c client does not expect a response for the first request that triggers the upgrade. Changing this has compatibility implications that I need to explore further.

yawkat:
@graemerocher This PR is ready for now. It only fixes the h2c server and normal client. The streaming client does not work yet, it never gets a response, I will work on that separately. But that was broken before this change anyway.

If possible, this PR should go into 3.1.0.

graemerocher (approved):

PR #6273, opened 2021-10-04
Add support for reading the body for streamed requests when the response is an error

### LINKED ISSUE #6162: The body is not read for streaming http client responses that produce an error
### Expected Behavior

Having seen this issue: https://github.com/micronaut-projects/micronaut-core/issues/2324

Response body should be properly decoded, more specifically We want to map response body to a `Problem`.

https://github.com/nyilmaz/mcnt-demo/blob/master/src/main/java/com/migros/exception/HttpClientExceptionHandler.java#L28-L33

### Actual Behaviour

`exception.response.getBody(class)` ends up Optional.empty

### Steps To Reproduce

1. clone https://github.com/nyilmaz/mcnt-demo
2. run project
3. curl -v http://localhost:8080/test-api-v2/test
4. check logs

### Environment Information

macOS 10.15.7
```
openjdk 16.0.1 2021-04-20
OpenJDK Runtime Environment Homebrew (build 16.0.1+0)
OpenJDK 64-Bit Server VM Homebrew (build 16.0.1+0, mixed mode, sharing)
```

### Example Application

https://github.com/nyilmaz/mcnt-demo

### Version

2.5.13

jameskleeh:
Closes #6162

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
@@ -46,6 +46,7 @@ public interface StreamingHttpClient extends HttpClient {
     /**
      * Request a stream of data where each emitted item is a {@link ByteBuffer} instance.
      *
+     * @since 3.1.0
      * @param request   The request
      * @param errorType The type that the response body should be coerced into if the server responds with an error
      * @param <I>       The request body type
@@ -67,6 +68,7 @@ public interface StreamingHttpClient extends HttpClient {
      * Requests a stream data where each emitted item is a {@link ByteBuffer} wrapped in the {@link HttpResponse} object
      * (which remains the same for each emitted item).
      *
+     * @since 3.1.0
      * @param request   The {@link HttpRequest}
      * @param errorType The type that the response body should be coerced into if the server responds with an error
      * @param <I>       The request body type
@@ -106,6 +108,7 @@ public interface StreamingHttpClient extends HttpClient {
      * <p>The downstream {@link org.reactivestreams.Subscriber} can regulate demand via the subscription. Incoming data
      * is buffered.</p>
      *
+     * @since 3.1.0
      * @param request   The {@link HttpRequest} to execute
      * @param type      The type of object to convert the JSON into
      * @param errorType The type that the response body should be coerced into if the server responds with an error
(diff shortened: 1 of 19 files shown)


http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
```diff
@@ -42,6 +42,18 @@
...
+    /**
+     * Request a stream of data where each emitted item is a {@link ByteBuffer} instance.
+     *
+     * @since 3.1.0
+     * @param request   The request
+     * @param errorType The type that the response body should be coerced into if the server responds with an error
+     * @param <I>       The request body type
+     * @return A {@link Publisher} that emits a stream of {@link ByteBuffer} instances
+     */
+    <I> Publisher<ByteBuffer<?>> dataStream(HttpRequest<I> request, Argument<?> errorType);
```
graemerocher: I know it is missing from the rest of the methods, but it seems this whole API is missing `@Nullable` / `@NonNull` the presence of which would help kotlin users

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1805,6 +1871,63 @@ HttpToHttp2ConnectionHandlerBuilder newHttp2ConnectionHandlerBuilder(
...
+
+                                @Override
+                                public void onNext(HttpContent httpContent) {
+                                    buffer.addComponent(true, httpContent.content());
+                                    s.request(1);
+                                }
+
+                                @Override
+                                public void onError(Throwable t) {
+                                    emitter.error(t);
```
graemerocher, line 1901: Seems the buffer cleanup logic would need to occur `onError` as well since `onComplete` may not be called?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1805,6 +1871,63 @@ HttpToHttp2ConnectionHandlerBuilder newHttp2ConnectionHandlerBuilder(
...
+                                }
+
+                                @Override
+                                public void onComplete() {
+                                    FullHttpResponse fullHttpResponse = new DefaultFullHttpResponse(nettyResponse.protocolVersion(), nettyResponse.status(), buffer, nettyResponse.headers(), new DefaultHttpHeaders(true));
+                                    final FullNettyClientHttpResponse<Object> fullNettyClientHttpResponse = new FullNettyClientHttpResponse<>(fullHttpResponse, response.status(), mediaTypeCodecRegistry, byteBufferFactory, (Argument<Object>) errorType, true);
+                                    fullNettyClientHttpResponse.onComplete();
+                                    buffer.release();
```
graemerocher: `release()` in the finally block?

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
jameskleeh: Right on

PR #6281, opened 2021-10-06
Small bean context perf improvement

dstepanov:


graemerocher:
Seems to break stuff

dstepanov:
Seems fine now

graemerocher:
So how much does it improve in terms of perf?

dstepanov:
Not sure really, but I noticed that we have compile time info if the bean is iterable so we don't need to check annotations being present.

graemerocher:
Ok let's wait until after 3.1 for this and do more testing since I want to avoid regression

graemerocher (approved):

Issue #6085, opened 2021-09-01
EmbeddedServer java.net.BindException: Address already in use


savage-matt-genie:
### Expected Behavior

Spinning up `EmbeddedServer` should never throw `java.net.BindException: Address already in use`

### Actual Behaviour

Spinning up `EmbeddedServer` sometimes throws `java.net.BindException: Address already in use`.

My guess is that sometimes another process binds to the random port it has chosen before it has a chance to?

* `SocketUtils.findAvailableTcpPort()` starts looking at port `1024` rather than `5000`, increasing the probability of a clash with another container spun up during the test
* In any case, random ports should be generated by binding to port `0` and accepting whatever the OS gives us- all other implementations are buggy. The port is only guaranteed to be free at the point `SocketUtils` checks, not by the time the server tries to listen.

### Version

2.5.5

jameskleeh:
I think it makes sense to do, however we would have to figure out how to get the actual bound port to make that available in the `getPort` method.

jameskleeh:
@savage-matt-genie PR welcome

cbeams:
I just upgraded my application from 3.0.3 to 3.1.0-SNAPSHOT while testing @yawkat's fix for #5005, and I noticed that my tests started to fail, saying that they couldn't bind to port `-1`.

Strange, I thought, because I had specifically set up my `application-test.yml` file to override the default `micronaut.server.port` value to `-1` in order to ensure my tests' `EmbeddedServer`s bind to a random port, a feature advertised at https://docs.micronaut.io/3.0.3/guide/index.html#runningSpecificPort.

This arrangement had been working perfectly, but stopped under 3.1.0-SNAPSHOT. Indeed, looking at @graemerocher's changes in https://github.com/micronaut-projects/micronaut-core/pull/6128/files#diff-ae5ec3499c014b9efa0ec7b4b72092983d393bdb159a15f14f2cb1a8376a62c8L300-L309, it appears the `-1` support has been removed entirely.

I tried changing my `application-test.yml` file to set `micronaut.server.port` to `0` instead of `1`, but this did not work either. The `0` value is taken literally, and my tests fail as they are unable to bind to it.

Perhaps I missed something? In any case, I guess the docs should be updated to reflect the new situation.

graemerocher:
@cbeams I am struggling to reproduce the behaviour you mention. These are the steps I undertook:

1. Modify `application.yml` as follows:

```yaml
micronaut:
  application:
[...]

PR #6290, opened 2021-10-06
EmbeddedServer.getPort improvement

jameskleeh:
This is up for discussion. I can definitely see that some folks would prefer to just get the -1 rather than handle an exception

relates to https://github.com/micronaut-projects/micronaut-core/issues/6085


graemerocher:
+1 to this change

graemerocher (approved):

Issue #5680, opened 2021-06-29
Why PongWebSocketFrame is never processed as a Message?


tkgreg:
I was trying to implement WS ping-pong functionality in case when the server measures the latency between clients(Server send ping and expects pong frame from the client). I need this for detection of stalled connections. Unfortunately, the current implementation does not allow this. It simply ignores the PongWebSocketFrame

https://github.com/micronaut-projects/micronaut-core/blob/a0d5619f6c23d909159503f5ed80d44625e49fe3/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java#L425

What was the reason to implement it like this? Do I try to do something unusual and stupid? 😸

graemerocher:
Seems like an oversight 

PRs welcome

tkgreg:
We already implemented our own ping-pong solution which is not that hard but its nice to have it out of the box.

noam-alchemy:
What's the correct behavior for propagating the pong frame back to the serverside listener?

graemerocher:
@yawkat another one to your list :)

yawkat:
@graemerocher This needs new API for receiving websocket pongs. `javax.websocket` does this by allowing an `@OnMessage` method that accepts a [`PongMessage`](https://docs.oracle.com/javaee/7/api/javax/websocket/PongMessage.html). However this is difficult to implement given our existing `OnMessage` handling that currently only allows one such method. I thought about implementing it as a `WebSocketEvent`, but those are only for the server. What approach is the best, here? I could add a new annotation, but this would diverge further from `javax.websocket`.

graemerocher:
seems like you should be able to define an overridden `@OnMessage` handler that takes `Ping/PongMessage` the `WebSocketBean` API then needs extending to with additional methods like:

```
    /**
     * Returns the method annotated with {@link io.micronaut.websocket.annotation.OnMessage}.
     *
     * @return the method
     */
    Optional<MethodExecutionHandle<T, ?>> pingMethod();

    /**
     * Returns the method annotated with {@link io.micronaut.websocket.annotation.OnMessage}.
     *
     * @return the method
     */
    Optional<MethodExecutionHandle<T, ?>> pongMethod();
```

Which can be disambiguated by the presence of the appropriate argument

PR #6294, opened 2021-10-07
WebSocket ping API [Adds #5680]

yawkat:
This API is mostly parallel to the normal message API. To stay consistent with javax.websocket, methods annotated with `@OnMessage` are used as pong handlers if they have a `WebSocketPongMessage` argument.

A `sendPingAsync` method was added to `WebSocketSession` to allow sending a ping.

Neither of these APIs does the type conversion that the normal message handling does. In my opinion it does not make much sense for ping/pong messages – you might want to put a timestamp into the message, but that's about it.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/websocket/NettyWebSocketClientHandler.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/websocket/NettyWebSocketClientHandler.java
@@ -33,6 +33,7 @@ import io.micronaut.http.uri.UriMatchInfo;
 import io.micronaut.http.uri.UriMatchTemplate;
 import io.micronaut.inject.MethodExecutionHandle;
 import io.micronaut.websocket.CloseReason;
+import io.micronaut.websocket.WebSocketPongMessage;
 import io.micronaut.websocket.annotation.ClientWebSocket;
 import io.micronaut.websocket.bind.WebSocketState;
 import io.micronaut.websocket.bind.WebSocketStateBinderRegistry;
@@ -78,6 +79,7 @@ public class NettyWebSocketClientHandler<T> extends AbstractNettyWebSocketHandle
     private final WebSocketStateBinderRegistry webSocketStateBinderRegistry;
     private FullHttpResponse handshakeResponse;
     private Argument<?> clientBodyArgument;
+    private Argument<?> clientPongArgument;
 
     /**
      * Default constructor.
@@ -123,6 +125,11 @@ public class NettyWebSocketClientHandler<T> extends AbstractNettyWebSocketHandle
     @Override
     public Argument<?> getBodyArgument() {
         return clientBodyArgument;
+    }
+
+    @Override
+    public Argument<?> getPongArgument() {
+        return clientPongArgument;
     }
 
     @Override
@@ -187,6 +194,26 @@ public class NettyWebSocketClientHandler<T> extends AbstractNettyWebSocketHandle
                     }
                 }
                 return;
+            }
+
+            if (pongHandler != null) {
+                BoundExecutable<?, ?> boundPong = binder.tryBind(pongHandler.getExecutableMethod(), webSocketBinder, new WebSocketState(clientSession, originatingRequest));
+                List<Argument<?>> unboundPongArguments = boundPong.getUnboundArguments();
+
+                if (unboundPongArguments.size() == 1 && unboundPongArguments.get(0).isAssignableFrom(WebSocketPongMessage.class)) {
+                    this.clientPongArgument = unboundPongArguments.get(0);
+                } else {
+                    this.clientPongArgument = null;
+
+                    try {
+                        emitter.error(new WebSocketClientException("WebSocket @OnMessage pong handler method " + targetBean.getClass().getSimpleName() + "." + messageHandler.getExecutableMethod() + " should define exactly 1 pong message parameter, but found: " + unboundArguments));
+                    } finally {
+                        if (getSession().isOpen()) {
+                            getSession().close(CloseReason.INTERNAL_ERROR);
+                        }
+                    }
+                    return;
+                }
             }
 
             Optional<? extends MethodExecutionHandle<?, ?>> opt = webSocketBean.openMethod();
--- a/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
@@ -34,6 +34,7 @@ import io.micronaut.http.codec.MediaTypeCodecRegistry;
 import io.micronaut.inject.ExecutableMethod;
 import io.micronaut.inject.MethodExecutionHandle;
 import io.micronaut.websocket.CloseReason;
+import io.micronaut.websocket.WebSocketPongMessage;
 import io.micronaut.websocket.bind.WebSocketState;
 import io.micronaut.websocket.bind.WebSocketStateBinderRegistry;
 import io.micronaut.websocket.context.WebSocketBean;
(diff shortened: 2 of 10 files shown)


http-netty/src/main/java/io/micronaut/http/netty/websocket/NettyWebSocketSession.java
```diff
@@ -217,6 +219,26 @@ public void sendSync(Object message, MediaType mediaType) {
         }, FluxSink.OverflowStrategy.ERROR);
     }
 
+    @Override
+    public CompletableFuture<?> sendPingAsync(byte[] content) {
+        if (isOpen()) {
+            ByteBuf messageBuffer = channel.alloc().buffer(content.length);
```
graemerocher, line 227: Does this need releasing  when the future completes?

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -154,6 +158,25 @@ protected AbstractNettyWebSocketHandler(
...
+                } else {
+                    this.pongArgument = null;
+                    if (LOG.isErrorEnabled()) {
+                        LOG.error("WebSocket @OnMessage pong handler method " + webSocketBean.getTarget() + "." + pongHandler.getExecutableMethod() + " should define exactly 1 message parameter assignable from a WebSocketPongMessage, but found: " + unboundArguments);
+                    }
+
+                    if (session.isOpen()) {
+                        session.close(CloseReason.INTERNAL_ERROR);
+                    }
+                }
```
graemerocher, line 175: Please add compilation time validation as well. See https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/http-validation/src/main/java/io/micronaut/validation/websocket/WebSocketVisitor.java

http-client/src/main/java/io/micronaut/http/client/netty/websocket/NettyWebSocketClientHandler.java
```diff
@@ -189,6 +196,26 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) {
...
+                        emitter.error(new WebSocketClientException("WebSocket @OnMessage pong handler method " + targetBean.getClass().getSimpleName() + "." + messageHandler.getExecutableMethod() + " should define exactly 1 pong message parameter, but found: " + unboundArguments));
+                    } finally {
+                        if (getSession().isOpen()) {
+                            getSession().close(CloseReason.INTERNAL_ERROR);
+                        }
+                    }
+                    return;
+                }
+            }
+
```
graemerocher, line 218: Please add compilation time validation as well. See https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/http-validation/src/main/java/io/micronaut/validation/websocket/WebSocketVisitor.java

websocket/src/main/java/io/micronaut/websocket/WebSocketPongMessage.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package io.micronaut.websocket;
+
+import io.micronaut.core.io.buffer.ByteBuffer;
+
+/**
+ * Special message class that can be accepted by a {@link io.micronaut.websocket.annotation.OnMessage @OnMessage}
+ * method to listen to WebSocket pongs.
+ */
+public final class WebSocketPongMessage {
```
graemerocher, line 30: Add `@since`

websocket/src/main/java/io/micronaut/websocket/WebSocketPongMessage.java
```diff
@@ -0,0 +1,40 @@
...
+    private final ByteBuffer<?> content;
+
+    /**
+     * @param content The content of the pong message.
+     */
+    public WebSocketPongMessage(ByteBuffer<?> content) {
```
graemerocher: Maybe a no arguments constructor with an empty content is required as well, seems like sending a body is optional?

websocket/src/main/java/io/micronaut/websocket/WebSocketPongMessage.java
```diff
@@ -0,0 +1,40 @@
...
+        this.content = content;
+    }
+
+    /**
+     * @return The content of the pong message. This buffer may be released after the message handler has completed.
+     */
+    public ByteBuffer<?> getContent() {
```
graemerocher, line 45: Add nullability annotations

websocket/src/main/java/io/micronaut/websocket/WebSocketSession.java
```diff
@@ -162,6 +162,19 @@ default void sendSync(Object message) {
...
+    /**
+     * Send a ping through this WebSocket. The pong reply can be intercepted using a
+     * {@link io.micronaut.websocket.annotation.OnMessage @OnMessage} method that accepts a
+     * {@link WebSocketPongMessage}.
+     *
+     * @param content The content of the ping. The remote should return the same content in its
+     * {@link WebSocketPongMessage}.
+     * @return A future that completes when the ping has been sent. (Not when the pong has been received!)
+     */
+    default CompletableFuture<?> sendPingAsync(byte[] content) {
```
graemerocher: Add nullable annotations and `@since`

graemerocher: Do we need `ByteBuffer` variants?

websocket/src/main/java/io/micronaut/websocket/context/WebSocketBean.java
```diff
@@ -41,12 +41,23 @@
...
      */
     Optional<MethodExecutionHandle<T, ?>> messageMethod();
 
+    /**
+     * Returns the method annotated with {@link io.micronaut.websocket.annotation.OnMessage} responsible for pong
+     * messages.
+     *
+     * @return the method
+     */
+    default Optional<MethodExecutionHandle<T, ?>> pongMethod() {
```
graemerocher, line 58: Add `@since`

graemerocher (changes requested):

http-netty/src/main/java/io/micronaut/http/netty/websocket/NettyWebSocketSession.java
yawkat: The channel should take care of it I think, since it's being written

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
yawkat: Will do, will be fairly limited though because OnMessage validation should already do it

yawkat: Actually, the existing OnMessage checks already cover this completely. The only thing this code adds is check that the parameter is WebSocketPongMessage, but if it isn't, it'll just be used as a normal message handler

websocket/src/main/java/io/micronaut/websocket/WebSocketPongMessage.java
yawkat: Yes, netty does this. How can I create an empty ByteBuffer though, since this is in the module without netty?

It's also not that important, because this class is really only constructed by the actual websocket handler, not by users.

websocket/src/main/java/io/micronaut/websocket/WebSocketSession.java
yawkat: Not sure, the only thing the implementation could do with a ByteBuffer is copy it using toByteArray anyway. There could be a special case for a ByteBuffer<ByteBuf> but right now I do a defensive copy anyway (is that sensible?)

websocket/src/main/java/io/micronaut/websocket/WebSocketPongMessage.java
graemerocher: One option would be to allow content to be `@Nullable` and return an `Optional`

graemerocher: But yes I get your point about it being not that important, does it make sense to mark this class as `@Internal` in that case?

yawkat: Well the class is still used as a parameter type, users just wouldn't create instances on their own. The constructor could be internal

jameskleeh:
Needs documentation

yawkat:
Converted to draft, should go into 3.2.x branch once it's created.

@jameskleeh Do you mean more javadoc, or actual website docs? I'd need to read up how to do it for the latter.

jameskleeh:
Website docs

jameskleeh:
Rebase against 3.2.x

yawkat:
I've added a note for the `WebSocketPongMessage` parameter for `@OnMessage` methods.

In my opinion, adding the ping api to the example snippet in the guide is not sensible, because it is fairly niche and would just add noise for most users. But as-is, this API isn't very discoverable, so not sure how to proceed here.

PR #6311, opened 2021-10-10
Added more configuration options to websockets

BrianEstrada:
The following parameters from the `WebSocketDecoderConfig` were not configureable from the `ServerWebSocket` 

```
boolean expectMaskedFrames;
boolean allowMaskMismatch;
boolean allowExtensions;
boolean closeOnProtocolViolation;
boolean withUTF8Validator;
```

So I've added the possibility to configure them from the `OnMessage` annotation let me know if anything needs to be changed I tried to follow the code style as close as possible 😄

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
@@ -54,6 +54,7 @@ import io.netty.handler.codec.http.DefaultHttpHeaders;
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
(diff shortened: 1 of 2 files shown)


jameskleeh:
Please rebase against 3.2.x

This also needs tests

BrianEstrada:
@jameskleeh I don't see a branch called `3.2.x` I only see a `3.1.x` is that the one you meant?

jameskleeh:
There is a 3.2.x branch https://github.com/micronaut-projects/micronaut-core/tree/3.2.x

BrianEstrada:
ah okay, thanks. I've done that now I'll work on getting some tests in sometime this week 🙂

jameskleeh:
So I'm a little hesitant about this because while even though the websocket stuff only works with Netty, the project itself has no dependency on Netty and adding these Netty specific options would make it more difficult to add other implementations in the future.

graemerocher:
I don't think these options belong on the annotation, they seem more like configuration options

BrianEstrada:
@graemerocher I agree but I was trying to stick to the existing style, but I guess this might be nicer to have a configuration per websocket server if you have multiple?

PR #6313, opened 2021-10-11
Introduce static optimizations registry

melix:
This is preliminary work for Micronaut AOT, a module which is
going to compute optimizations at build time. For this to work,
we need to introduce a class called `StaticOptimizations` which
acts as a registry for "static" optimizations. The registry
works by statically (in the sense of a static field) injecting
optimization data, that _specific_ classes care about.

The pattern is simple: if a class can benefit from precomputed
data at build time, then it should create an inner class which
is going to store this data as an "optimization". The AOT module
will eventually be responsible for computing the data and injecting
it. This means that by themselves, the changes in this PR do nothing:
the optimizations need to be injected at runtime.

As an example, let's imagine that we have a class called `Primes` which
computes prime numbers. An obvious optimization that we can do is to
precompute a number of primes at build time, and inject them to the
`Primes` class. Therefore, the pseudo-code would look like this:

```
public class Primes {
    public static Optimizations {
       private final List<Integer> firstPrimes;
       ...
    }

    private List<Integer> primes = StaticOptimizations.get(Optimizations)
        .map(opt -> new ArrayList<>(opt.firstPrimes)
        .orElse(new ArrayList<>);
}
```

The pattern is to pre-initialize the fields by looking if a static optimization
is available.

The Micronaut AOT module will then be responsible for _somehow_ calling the
`StaticOptimizations#set` method at runtime, typically when the application
starts.

This mechanism is interesting for different reasons:

- it makes all static optimizations discoverable by searching for usages of
the `StaticOptimizations.get()` method
- it's transparent, in the sense that if no optimizations are injected at
runtime, the behavior of the application is rigourously the same as before
- it doesn't matter how the optimizations are injected. Micronaut AOT is one
way, but nothing prevents user code from injecting optimizations themselves
- it's runtime independent: distinct optimizations can be injected for
different contexts. For example, we can inject optimizations specifically
for the JIT mode, and different optimizations for native images. This is
typically the case in this commit for the service loading mechanism.


core-reactive/src/main/java/io/micronaut/core/async/publisher/CompletableFuturePublisher.java
```diff
@@ -30,7 +30,7 @@
  * @author Graeme Rocher
  * @since 1.0
  */
-class CompletableFuturePublisher<T> implements Publishers.MicronautPublisher<T> {
+public class CompletableFuturePublisher<T> implements Publishers.MicronautPublisher<T> {
```
graemerocher, line 35: If this is being made public maybe mark it as `@Internal`

core-reactive/src/main/java/io/micronaut/core/async/publisher/PublishersOptimizations.java
```diff
@@ -0,0 +1,43 @@
...
+        return reactiveTypes;
+    }
+
+    public List<Class<?>> getSingleTypes() {
+        return singleTypes;
+    }
+
+    public List<Class<?>> getCompletableTypes() {
+        return completableTypes;
+    }
```
graemerocher: Do these accessors need to be public?

core-reactive/src/main/java/io/micronaut/core/async/publisher/PublishersOptimizations.java
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.async.publisher;
+
+import java.util.List;
+
+public final class PublishersOptimizations {
```
graemerocher, line 30: Add `@Internal` and javadoc with `@since`

core/src/main/java/io/micronaut/core/io/service/SoftServiceLoader.java
```diff
@@ -135,16 +154,57 @@ private SoftServiceLoader(Class<S> serviceType, ClassLoader classLoader, Predica
...
+    private static void measure(String label, Runnable op) {
+        long sd = System.nanoTime();
+        try {
+            op.run();
+        } finally {
+            long dur = System.nanoTime() - sd;
+            LOGGER.debug(label + " took " + TimeUnit.MILLISECONDS.convert(dur, TimeUnit.NANOSECONDS) + "ms");
+        }
+    }
+
```
graemerocher, line 166: Do we need these measurements on the final commit?

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * This class is a generic container for pre-computed data
+ * which can be injected at initialization time. Every class
+ * which needs static optimizations should go through this
+ * class to get its static state.
+ */
+@SuppressWarnings("unchecked")
+@Internal
+public abstract class StaticOptimizations {
```
graemerocher, line 37: add `@since` to javadoc

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static <T> void set(T value) {
+        Class<?> optimizationClass = value.getClass();
+        LOGGER.debug("Setting optimizations for {}", optimizationClass);
+        OPTIMIZATIONS.put(optimizationClass, value);
+    }
+
+    public static boolean isEnvironmentCached() {
+        return cacheEnvironment;
+    }
```
graemerocher: Need javadoc on these methods

core/src/main/java/io/micronaut/core/reflect/ClassUtils.java
```diff
@@ -334,4 +352,17 @@ private static void populateHierarchyInterfaces(Class<?> superclass, List<Class>
             populateHierarchyInterfaces(aClass, hierarchy);
         }
     }
+
+    public static final class Optimizations {
```
graemerocher, line 358: Add `@Internal`

core/src/main/java/io/micronaut/core/util/EnvironmentProperties.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+@Internal
+public final class EnvironmentProperties {
```
graemerocher, line 37: Needs javadoc with `@since`

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package io.micronaut.context.env;
+
+import io.micronaut.core.optim.StaticOptimizations;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CachedEnvironment {
```
graemerocher, line 43: Needs javadoc with a description of the behaviour and `@since`

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final boolean LOCKED = StaticOptimizations.isEnvironmentCached();
+    private static final Map<String, String> CACHED_ENVIRONMENT = Collections.unmodifiableMap(new HashMap<>(System.getenv()));
+    private static final Map<Object, String> CACHED_PROPERTIES;
+
+    static {
+        Map<Object, String> props = new HashMap<>();
+        System.getProperties()
+                .forEach((key, value) -> props.put(key, String.valueOf(value)));
+        CACHED_PROPERTIES = Collections.unmodifiableMap(props);
```
graemerocher: Seems like this should only be intialized with values if `LOCKED = true`?

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
```diff
@@ -0,0 +1,47 @@
...
+
+import io.micronaut.core.optim.StaticOptimizations;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CachedEnvironment {
+    private static final boolean LOCKED = StaticOptimizations.isEnvironmentCached();
+    private static final Map<String, String> CACHED_ENVIRONMENT = Collections.unmodifiableMap(new HashMap<>(System.getenv()));
```
graemerocher: Seems like this should only be intialized with values if LOCKED = true?

inject/src/main/java/io/micronaut/context/env/ConstantPropertySources.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.context.env;
+
+import io.micronaut.core.annotation.Internal;
+
+import java.util.List;
+
+@Internal
+public final class ConstantPropertySources {
```
graemerocher, line 32: Needs javadoc with `@since`

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final Map<Object, String> CACHED_PROPERTIES;
+
+    static {
+        Map<Object, String> props = new HashMap<>();
+        System.getProperties()
+                .forEach((key, value) -> props.put(key, String.valueOf(value)));
+        CACHED_PROPERTIES = Collections.unmodifiableMap(props);
+    }
+
+    public static String getenv(String name) {
```
graemerocher, line 69: Add nullability annotations

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
```diff
@@ -0,0 +1,47 @@
...
+        return LOCKED ? CACHED_ENVIRONMENT.get(name) : System.getenv(name);
+    }
+
+    public static Map<String, String> getenv() {
```
graemerocher, line 80: Add nullability annotations

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
```diff
@@ -0,0 +1,47 @@
...
+        return LOCKED ? CACHED_ENVIRONMENT : System.getenv();
+    }
+
+    public static String getProperty(String name) {
```
graemerocher, line 92: Add nullability annotations

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
graemerocher: Add javadoc

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
graemerocher: Add javadoc

core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
```diff
@@ -497,7 +540,7 @@ public static boolean isCompletable(Class<?> type) {
      *
      * @param <T> The type
      */
-    private static class JustPublisher<T> implements MicronautPublisher<T> {
+    public static class JustPublisher<T> implements MicronautPublisher<T> {
```
graemerocher, line 544: If this is being made public please add `@Internal`

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(StaticOptimizations.class);
+
+    private static final Map<Class<?>, Object> OPTIMIZATIONS = new ConcurrentHashMap<>();
+    private static boolean cacheEnvironment = false;
+
+    public static void cacheEnvironment() {
+        cacheEnvironment = true;
+    }
+
+    public static <T> Optional<T> get(Class<T> optimizationClass) {
```
graemerocher: Add nullability annotations and javadoc

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
```diff
@@ -0,0 +1,63 @@
...
+        T value = (T) OPTIMIZATIONS.get(optimizationClass);
+        if (value != null) {
+            LOGGER.debug("Found optimizations {}", optimizationClass);
+        } else {
+            LOGGER.debug("No optimizations {} found", optimizationClass);
+        }
+        return Optional.ofNullable(value);
+    }
+
+    public static <T> void set(T value) {
```
graemerocher: Add nullability annotations and javadoc

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@SuppressWarnings("unchecked")
+@Internal
+public abstract class StaticOptimizations {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StaticOptimizations.class);
+
+    private static final Map<Class<?>, Object> OPTIMIZATIONS = new ConcurrentHashMap<>();
+    private static boolean cacheEnvironment = false;
+
+    public static void cacheEnvironment() {
```
graemerocher, line 47: add javadoc

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
```diff
@@ -0,0 +1,63 @@
...
+        return Optional.ofNullable(value);
+    }
+
+    public static <T> void set(T value) {
+        Class<?> optimizationClass = value.getClass();
+        LOGGER.debug("Setting optimizations for {}", optimizationClass);
+        OPTIMIZATIONS.put(optimizationClass, value);
+    }
+
+    public static boolean isEnvironmentCached() {
```
graemerocher, line 90: add javadoc

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/io/service/SoftServiceLoader.java
melix: I was wondering myself, but since I added/removed this several times during development I decided to keep it eventually, since it has no measurable overhead.

core-reactive/src/main/java/io/micronaut/core/async/publisher/PublishersOptimizations.java
melix: added

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
melix: done

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
melix: added

core/src/main/java/io/micronaut/core/reflect/ClassUtils.java
melix: added

core/src/main/java/io/micronaut/core/util/EnvironmentProperties.java
melix: added

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
melix: added

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
melix: added

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
melix: Out of curiosity: `Optional` doesn't imply `@NonNull`?

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
melix: added

core-reactive/src/main/java/io/micronaut/core/async/publisher/PublishersOptimizations.java
melix: Nope, made package private

core-reactive/src/main/java/io/micronaut/core/async/publisher/CompletableFuturePublisher.java
melix: done, this is indeed because otherwise they have to be instantiated via reflection.

core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
melix: done

core/src/main/java/io/micronaut/core/optim/StaticOptimizations.java
melix: added

inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
melix: It wasn't strictly necessary but I made the change for clarity.

graemerocher (approved):

PR #6329, opened 2021-10-14
Upgrade to Netty 4.1.69 and use its BOM

ilopmar:
This PR upgrades to Netty 4.1.69 and uses its bom for tcnative dependency so we don't need to worry anymore about keep them in sync.

Three tests failed for me when running them locally (https://ge.micronaut.io/s/ev6fuqordezci) but running only those 3 modules worked (https://ge.micronaut.io/s/edywycznjiuj6), so not sure if this is something on my side or some race condition.

Let's see what happens on CI...


ilopmar:
@melix can you please confirm this is the right way to apply Netty's BOM?

melix:
LGTM :+1:

PR #6350, opened 2021-10-18
Configure content length for h2c upgrade request [Fixes #6299]

### LINKED ISSUE #6299: POST under h2c returns 413 Request Entity Too Large
### Expected Behavior

The following is taken from a commit that will be linked to this issue shortly after its creation:

    This commit sets Micronaut's http-version to 1.1, such that the
    following steps work as expected:

    1. Run the application

        ./gradlew run

    2. Run

        curl -v -H "Content-Type: application/json" :2040/peers \
          --data '{"address":"http://localhost:2041"}'

    3. Notice that the response is 201 Created as expected.

### Actual Behaviour

The following is taken from a commit that will be linked to this issue shortly after its creation:

    This commit sets Micronaut's http-version to 2.0, in order to reproduce
    the reported issue.

    Repro steps when requesting with HTTP/1.1:

    1. Run the application:

        ./gradlew run

    2. Run the following (note that we are *not* passing `--http2` here):

        curl -v -H "Content-Type: application/json" :2040/peers \
          --data '{"address":"http://localhost:2041"}'

    3. Notice that the response hangs indefinitely

    Repro steps when requesting an HTTP/2 upgrade

    1. Reuse the same application already running from the steps above.

    2. Run the following (note that we *are* passing `--http2` this time):

        curl -v --http2 -H "Content-Type: application/json" :2040/peers \
          --data '{"address":"http://localhost:2041"}'

    3. Notice that the response comes back 413 Request Entity Too Large

### Steps To Reproduce

See above.

### Environment Information

JDK 16

### Example Application

https://github.com/cbeams/xnd

### Version

3.1.0-SNAPSHOT

yawkat:
Also small modification to `HttpRequestDecoder` to handle `FullHttpRequest`s properly as returned by the upgrade handler.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/decoders/HttpRequestDecoder.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/decoders/HttpRequestDecoder.java
@@ -28,6 +28,7 @@ import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.MessageToMessageDecoder;
 import io.netty.handler.codec.http.DefaultHttpRequest;
+import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/decoders/HttpRequestDecoder.java
```diff
@@ -77,6 +78,9 @@ protected void decode(ChannelHandlerContext ctx, HttpRequest msg, List<Object> o
         }
         try {
             NettyHttpRequest<Object> request = new NettyHttpRequest<>(msg, ctx, conversionService, configuration);
+            if (msg instanceof HttpContent && ((HttpContent) msg).content().readableBytes() > 0) {
+                request.addContent(((HttpContent) msg).retain());
```
graemerocher, line 82: Does this content get cleared up somewhere?

yawkat: `NettyHttpRequest.release` handles it

graemerocher (approved):

PR #6357, opened 2021-10-18
Store client cookies in a single header

### LINKED ISSUE #6356: Multiple `@CookieValue` client arguments generates multiple `Cookie` headers
### Issue description

According to https://www.rfc-editor.org/rfc/rfc6265#section-5.4 there must only be a single header

This restriction is relaxed for HTTP/2 

https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.5

jameskleeh:
Fixes #6356


jameskleeh:
It's debatable whether the `cookies` method should ignore the previous ones. Open to feedback

PR #6369, opened 2021-10-20
HTTP2 server push API

### LINKED ISSUE #4456: HTTP/2 Server Push
Any plans to add support for [HTTP/2 server push](https://en.wikipedia.org/wiki/HTTP/2_Server_Push) to Micronaut?

yawkat:
This is a proof of concept for the server push API. The test cases pass, but the implementation isn't complete yet.

There's still some outstanding questions on the API:

- Since server pushes are supposed to preempt a request, the server sends *request* headers in the `PUSH_PROMISE` frame as well. Those can't be customized right now. Should we support this? According to the wiki, there's some use cases related to cache headers, but I'm not sure if it's worth the extra confusion in the API.
- Right now, there's only a way to attach additional `HttpResponse`s to the main response, but maybe there should also be an API that triggers a fake request to another endpoint? For example, there could be a method `response.preemptWithServerPush(HttpRequest.GET("/resource1"))` that triggers normal request processing and sends the response of that endpoint to the client.

@graemerocher Thoughts on the API?

todo:

- [x] test with actual browsers
- [x] unit test for right stream assignment


graemerocher:
@yawkat interesting initial, looking at prior art in this area it seems that the Servlet Spec associates pushes with the request and the not the response and does some via some kind of builder API 

https://dzone.com/articles/first-look-at-http2-server-push-in-java-servlet-40-1

Maybe we don't need the build because we already have the `MutableHttpResponse` API which is similar.

With regards to headers it does seem like you would want to the send cache headers to the browser if the primary use case is sending additional static resources down and tell the browser to cache those. I would also be interested to how this interacts with the `StreamedFile` API since most of the use cases seems to be about sending additional content down the same connection (files, css, js etc.)

yawkat:
It looks like that servlet API also uses the approach in my second point. Maybe that's why their push builder is associated with the request, not the response.

graemerocher:
I'm not really show what you mean by triggering a fake request to another endpoint, maybe you need to elaborate on that point. How does the servlet API do that?

yawkat:
if you look at your example:

```
pushBuilder
                .path("images/kodedu-logo.png")
                .addHeader("content-type", "image/png")
                .push();
```

They don't actually include a full response in the push builder.

What happens internally (presumably) is that they're creating a second request with the data provided in that snippet, they're executing that request, and sending the response to that request as a server push.

graemerocher:
> if you look at your example:

The example is not mine, it was pulled from a Google search 🤔

> They don't actually include a full response in the push builder.

With servlet there is always a response, it is written in this snippet:

```
      try(PrintWriter respWriter = resp.getWriter();){
            respWriter.write("<html>" +
                    "<img src='images/kodedu-logo.png'>" + (3)
                    "</html>");
        }
```

But I assume what you mean is the servlet API does not provide a response object but instead a subset (`PushBuilder` in this case) and what you pass to is reads the file at the path `"images/kodedu-logo.png"` and sends the data back. I doubt very much an internal request is created, it probably looks up the file from the static assets configured and available in the server and streams it back. Or am I missing something?

yawkat:
At least undertow seems to execute the full request logic for the PushBuilder API:

- [PushBuilder.push](https://code.yawk.at/io.undertow/undertow-servlet/2.2.3.Final/io/undertow/servlet/spec/PushBuilderImpl.java#io.undertow.servlet.spec.PushBuilderImpl%23push()) calls into ServerConnection
- [Http2ServerConnection.pushResource](https://code.yawk.at/io.undertow/undertow-core/2.2.3.Final/io/undertow/server/protocol/http2/Http2ServerConnection.java#io.undertow.server.protocol.http2.Http2ServerConnection%23pushResource%28java.lang.String%2Cio.undertow.util.HttpString%2Cio.undertow.util.HeaderMap%2Cio.undertow.server.HttpHandler%29) builds a new HttpServerExchange and calls executeRootHandler using that exchange and the root handler

But looking at that code, there's a lot more logic there too, like copying headers. Some of this is also mentioned in the servlet docs, that pushbuilder "inherits" some of the headers from the request the push originates from. Not sure how much of this makes sense for us to duplicate.

yawkat:
I've added a second API that is similar to the servlet one. It's actually a simpler implementation than the first API, but higher level. I'm also not certain how it will interact with filters and such, since it doesn't copy headers right now.

http/src/main/java/io/micronaut/http/HttpResponse.java
```diff
@@ -418,4 +421,9 @@ default Cookies getCookies() {
     default Optional<Cookie> getCookie(String name) {
         throw new UnsupportedOperationException("Operation not supported on a " + this.getClass() + " response.");
     }
+
+    @NonNull
+    default Map<URI, HttpResponse<?>> getServerPushResponses() {
+        return Collections.emptyMap();
+    }
```
graemerocher: I am not sure this should be public API seems like something internal to the framework. Can probably modeled in the netty impl of this interface. In particular we would need to potentially implement this API in `micronaut-servlet` and I am not sure if it is accessible there.

http/src/main/java/io/micronaut/http/HttpResponseWrapper.java
```diff
@@ -15,6 +15,11 @@
  */
 package io.micronaut.http;
 
+import org.jetbrains.annotations.NotNull;
```
graemerocher: Use Micronaut nullable

http/src/main/java/io/micronaut/http/HttpRequest.java
```diff
@@ -198,6 +198,13 @@ default boolean isSecure() {
         return this.getAttribute(HttpAttributes.X509_CERTIFICATE, Certificate.class);
     }
 
+    default boolean isServerPushSupported() {
+        return false;
+    }
+
+    default void serverPush(HttpRequest<?> request) {
```
graemerocher: Should probably throw `UnsupportedOperationException` if it is unsupported by the impl

http/src/main/java/io/micronaut/http/MutableHttpResponse.java
```diff
@@ -185,4 +188,9 @@
     default MutableHttpResponse<B> attribute(CharSequence name, Object value) {
         return (MutableHttpResponse<B>) setAttribute(name, value);
     }
+
+    @NonNull
+    default MutableHttpResponse<B> serverPush(@NonNull URI uri, @NonNull HttpResponse<?> pushResponse) {
+        return this;
```
graemerocher: Should throw `UnsupportedOperationException` if not implemented

graemerocher:
I think header inheritance is going to be necessary for the use case this is trying solve because I imagine when a JS client is involved you will almost certainly require the use of CORS for cross origin resource sharing therefore if you need to push additional resources by issuing new requests (I think I understand how this works now 😄) you have to pass the CORS headers on to new request such that the CORS filter can do its thing and send the right CORS headers back?

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/HttpRequest.java
yawkat: My thought was that these apis are "best-effort" anyway and should fail silently if unsupported. But maybe not, because you want to avoid the computation of the parameters if the client doesn't support server push anyway.

http/src/main/java/io/micronaut/http/HttpResponse.java
yawkat: What about DefaultHttpResponse? Do I need to support that or will the framework automatically pick the NettyHttpResponse if it's on the classpath?

Either way, this API (on the response side) will probably be dropped, so that solves that design issue.

yawkat:
@graemerocher what did you mean by "returning the response" for the push api?

http/src/main/java/io/micronaut/http/HttpResponse.java
graemerocher: If netty is on the classpath it will be picked up, otherwise it can just fail. The check to see if server push is supported as assert that the response is a NettyHttpResponse. 

Alternatively is to create some internal interface that you can downcast to that implementors can optionally implement. Like `HttpPushCapableResponse`

http/src/main/java/io/micronaut/http/HttpRequest.java
graemerocher: there is already a method to check if push is supported. If the developer calls this first they can still implement best effort without being confused why it doesn't work

graemerocher:
> @graemerocher what did you mean by "returning the response" for the push api?

I'm not sure what you are referring to

yawkat:
Okay, I thought I remembered something from yesterday.

graemerocher:
One thing I am thinking is maybe it is time to introduction `ServerHttpRequest` since none of these new methods apply to the client. Thoughts @jameskleeh ?

yawkat:
- [x] todo: header copying
- [x] move api to subinterface

http/src/main/java/io/micronaut/http/bind/DefaultRequestBinderRegistry.java
```diff
@@ -290,4 +308,29 @@ public int hashCode() {
             return result;
         }
     }
+
+    private static class PushCapableFullHttpRequest<B> extends FullHttpRequest<B> implements PushCapableHttpRequest<B> {
```
graemerocher: Make `final`

http/src/main/java/io/micronaut/http/PushCapableHttpRequest.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Initiate a HTTP2 server push for the given request. The information from the given request (i.e. path, headers)
+     * will be passed on to the client immediately so that it does not send the request itself. Then, the given request
+     * will be handled as if it was initiated by the client, and the response will be passed back to the client.
+     *
+     * @param request The request to respond to using a server push.
+     * @throws UnsupportedOperationException if the client does not support server push. Check beforehand using
+     * {@link #isServerPushSupported()}.
+     */
+    void serverPush(HttpRequest<?> request);
```
graemerocher: Add `@NonNull` to the argument and maybe return `PushCapableHttpRequest<B>` so you can use the API fluently

graemerocher (changes requested):

yawkat:
I've implemented the header copying, including a note in the javadoc on how this is security-relevant for auth headers.

graemerocher:
@jameskleeh any update on this one?

http/src/main/java/io/micronaut/http/bind/DefaultRequestBinderRegistry.java
```diff
@@ -93,6 +94,24 @@ public DefaultRequestBinderRegistry(ConversionService conversionService, Request
...
+        byType.put(Argument.of(PushCapableHttpRequest.class).typeHashCode(), (RequestArgumentBinder<PushCapableHttpRequest>) (argument, source) -> {
+            if (source instanceof PushCapableHttpRequest) {
+                Optional<Argument<?>> typeVariable = argument.getFirstTypeVariable()
+                        .filter(arg -> arg.getType() != Object.class)
+                        .filter(arg -> arg.getType() != Void.class);
+                if (typeVariable.isPresent() && HttpMethod.permitsRequestBody(source.getMethod())) {
+                    if (source.getBody().isPresent()) {
+                        return () -> Optional.of(new PushCapableFullHttpRequest((PushCapableHttpRequest) source, typeVariable.get()));
+                    } else {
+                        return ArgumentBinder.BindingResult.UNSATISFIED;
```
jameskleeh: UNSATISFIED means that argument binding should be attempted multiple times. I think in this case if it fails the first time there is no point in retrying. If that is true, this should be EMPTY

jameskleeh (changes requested):
Edit: n/m

PR #6385, opened 2021-10-21
Properly consume content of full HTTP requests [Fixes #6299]

### LINKED ISSUE #6299: POST under h2c returns 413 Request Entity Too Large

yawkat:
The h2c upgrade requests case from #6350 seems to be the only case where our server has to handle FullHttpRequests that have non-empty content. My [previous fix](https://github.com/micronaut-projects/micronaut-core/pull/6350/files#diff-80171c172c037b0610a42407a946eda5504eeefb4db5e63e31867c5a8532fecaR81-R83) simply used `NettyHttpRequest.addContent` to add the data to the micronaut request. This works fine for `@Body` parameters that can convert directly from `ByteBuf`, but it seems to bypass the `HttpContentProcessor` logic that is also responsible for json parsing.

This patch replaces the `addContent` approach from the previous PR. Instead, further downstream in the `HttpStreamsHandler`, it creates a "fake" `StreamedHttpRequest` that, as the input publisher, just uses the body already present on the request.

I'm not 100% sure why the `inMsg` does not need an additional `retain` call when also used as the message body, but it seems to work fine this way, and it matches the behavior of `HandlerPublisher`.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/stream/HttpStreamsHandler.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/stream/HttpStreamsHandler.java
@@ -27,6 +27,7 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.handler.codec.http.DefaultHttpResponse;
 import io.netty.handler.codec.http.FullHttpMessage;
+import io.netty.handler.codec.http.FullHttpRequest;
 import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpMessage;
 import io.netty.handler.codec.http.HttpResponseStatus;
@@ -202,7 +203,7 @@ abstract class HttpStreamsHandler<In extends HttpMessage, Out extends HttpMessag
 
             if (inMsg instanceof FullHttpMessage) {
                 FullHttpMessage fullMessage = (FullHttpMessage) inMsg;
-                if (fullMessage.content().readableBytes() == 0) {
+                if (!(fullMessage instanceof FullHttpRequest) || fullMessage.content().readableBytes() == 0) {
                     // Forward as is
                     ctx.fireChannelRead(inMsg);
                 } else {
--- a/http-netty/src/main/java/io/micronaut/http/netty/stream/HttpStreamsHandler.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/stream/HttpStreamsHandler.java
@@ -37,6 +37,7 @@ import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Flux;
 
 import java.util.LinkedList;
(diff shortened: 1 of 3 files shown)


graemerocher (approved):
This looks like a better approach anyway IMO

yawkat:
Looks like the approach breaks the client though. There seems to be full response handling on that end...

graemerocher (approved):

cbeams:
Thanks, @yawkat. I've just tried out the fix in my repro branch for #6299, and while I no longer see the 413 Error, I now see 

    Required Body [asset] not specified

when attempting to post via curl with `--http2`.

The same post works fine when `--http2` is omitted. So it appears that the post body is still not making it through under h2c.

The repro steps are exactly the same as they were in my original commit at https://github.com/cbeams/xnd/commit/c3303e2. You can simply check out my `repro-6299` branch and follow the instructions that (latest) commit.

yawkat:
@cbeams I can't reproduce this anymore. From the new error it sounds like it's the `/assets` endpoint, but I can't reproduce with that endpoint either.

cbeams:
Confirmed from my side as well. I'm no longer able to reproduce the original 413 issue, nor the subsequent "required body not specified" issue described above. Perhaps I tried reproducing this too quickly yesterday, and I hadn't yet picked up the 3.2.0 snapshot that actually contained your fix.

With that said, I just upgraded my application in its current form to 3.2.0-SNAPSHOT and set `http-version` to `2.0` and started seeing various errors in my integration tests. I haven't fully investigated yet, and will file any issues when I do, but please feel free to mark this particular issue as resolved in the meantime, as it appears that the errors that I'm seeing are not directly related to this. Thanks!

PR #6386, opened 2021-10-21
Introduce `@ContextConfigurer` annotation

melix:
This new annotation allows application context builder customizers
automatically. A class implementing `ApplicationContextCustomizer`
which is _also_ annotated with `@ContextConfigurer` will automatically
be registered as a service, and loaded when an application builder
is created.

There are several advantages of doing this:

- the "things" which are responsible for configuring the application
context become discoverable to other tools, in particular the new
Micronaut AOT module (so it can execute them on its own)
- the configuration blocks can be shared, or even published, and
would be automatically applied to all contexts created as long as
the service is found on classpath

Hence, a recommended pattern to configure an application would be
to rewrite this:

```java
class Application {
    public static void main(String... args) {
          Micronaut.builder(args)
             .mainClass(Application.class)
             .deduceEnvironment(false)
             .start();
    }
}
```

You would now write:

```java
@ContextConfigurer
public class Application implements ApplicationContextCustomizer {
     @Override
     public void configure(ApplicationContextBuilder builder) {
         builder.deduceEnvironment(false);
     }

     public static void main(String... args) {
         Micronaut.run(args, Application.class);
     }
}
```

While being longer to write, this patterns clearly
separates the configuration phase from the execution
phase.

If the application class is also used as a bean, a
recommended pattern would be to use an inner class:

```java
@Singleton
public class Application {
    @ContextConfigurer
    public static class Configurer implements ApplicationContextCustomizer {
        @Override
        public void configure(ApplicationContextBuilder builder) { ... }
    }

    ...
}
```

There are 2 annotations which currently have the same semantics:

- `@ContextConfigurer` is the main one, used to annotate any
application context builder configurer
- `@MicronautApplication`, which is used when annotating an
application class. While it's also possible to annotate such
a class with `@ContextConfigurer`, there's a semantics difference.
In the future this annotation may be used to generate more code,
in particular entry points.


inject/src/main/java/io/micronaut/context/ApplicationContextCustomizer.java
```diff
@@ -0,0 +1,48 @@
...
+     * A default customizer which does nothing.
+     */
+    ApplicationContextCustomizer NO_OP = new ApplicationContextCustomizer() {
+    };
+
+    /**
+     * Customizes the application context builder.
+     * @param builder the builder to customize
+     */
+    default void customize(ApplicationContextBuilder builder) {
```
graemerocher: Add `NonNull` to parameter

inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
```diff
@@ -354,4 +366,34 @@ protected ApplicationContext newApplicationContext() {
...
+     * customizers found on classpath via service loading.
+     * @return an application customizer
+     */
+    @NonNull
+    private static ApplicationContextCustomizer loadApplicationContextCustomizer() {
+        ServiceLoader<ApplicationContextCustomizer> loader = ServiceLoader.load(
+                ApplicationContextCustomizer.class
+        );
+        List<ApplicationContextCustomizer> customizers = StreamSupport.stream(loader.spliterator(), false)
+                .collect(Collectors.toList());
```
graemerocher, line 382: Perhaps call `OrderUtils.sort` to allow sorting of customizers if one needs to run before another (thinking about overridding properties etc.

inject/src/main/java/io/micronaut/context/annotation/MicronautApplication.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * }</pre>
+ *
+ * @since 3.2
+ */
+@Documented
+@Retention(SOURCE)
+@Target(ElementType.TYPE)
+@ContextConfigurer
+public @interface MicronautApplication {
```
graemerocher: Do we still need this if we have `@ContextConfigurer`?

inject/src/main/java/io/micronaut/context/visitor/ContextConfigurerVisitor.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * This visitor is responsible for generating service files for classes
+ * annotated with {@link ContextConfigurer} or {@link io.micronaut.context.annotation.MicronautApplication}.
+ *
+ * @since 3.2
+ */
+public class ContextConfigurerVisitor implements TypeElementVisitor<ContextConfigurer, Object> {
```
graemerocher, line 36: So this is an aggregating visitor correct?

graemerocher (changes requested):
Looks like a good start, I added some comments. Also we will need documentation for this feature.

inject/src/main/java/io/micronaut/context/visitor/ContextConfigurerVisitor.java
melix: Yes it is: you can have multiple classes annotated with `@ContextConfigurer` which would contribute to a single service file.

inject/src/main/java/io/micronaut/context/annotation/MicronautApplication.java
melix: Strictly speaking, at this stage, no. However as described in the commit message I think we can make a semantic difference between the two:

- you would annotate an application class with `@MicronautApplication`, which would _imply_ `@ContextConfigurer`
- and later we could have APs/codegen to generate an entry point, so that you can't make the mistake of _also_ configuring in a `main(...)` method

inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
melix: Good idea.

inject/src/main/java/io/micronaut/context/ApplicationContextCustomizer.java
melix: added

melix:
For documentation, I'd be keen on doing this once we are more or less sure the API is stable. We might need some changes during the AOT work.

graemerocher:
ok can you annotate everything new and public with `@Experimental` up until that point then?

jameskleeh:
ApplicationContextCustomizer should be ordered. Given that builder methods return a builder we shouldn't make the assumption that it is the same instance. That assumption is being made by the `configure` method returning void.

Is the only purpose of `@ContextConfigurer` to make it visible to a type element visitor?

Given that this is experimental and there is no need for `@MicronautApplication` at this point I'd prefer to leave it out and add it later if the need arises.

This opens up the possibility for libraries to customize how the context is created for applications. There are definitely pros and cons with that but its something that needs to be considered

melix:
> ApplicationContextCustomizer should be ordered. 

Does it _have_ to be? I'm sorting the application context customizers using `OrderUtils` so _if_ they implement ordered they would. I guess it's no harm to make them implement `Ordered`, though.

> Given that builder methods return a builder we shouldn't make the assumption that it is the same instance. That assumption is being made by the configure method returning void.

This is not what the javadoc contract says: the builder methods on `ApplicationContextBuilder` explicitly states the builder method should return "this builder". If we assume that a _new builder_ can be returned, then things become significantly more complicated to implement, especially because there are subclasses to `DefaultApplicationContextBuilder` (in particular `Micronaut`) and a particular customizer cannot know what return type it should use. That is, when a customizer which is implemented by a 3rd party is applied, we don't know if it's applied to a `Micronaut` instance or something else. It means that every single subclass of `DefaultApplicationContextBuilder` would have to apply the customizers in some form, and that it cannot be done at construct time anymore (since the constructor would have to potentially return a different instance than itself).

While I agree that in general you cannot assume that _a_ builder returns _this_, in Micronaut the contract says so.

> Given that this is experimental and there is no need for @MicronautApplication at this point I'd prefer to leave it out and add it later if the need arises.

Right.

> This opens up the possibility for libraries to customize how the context is created for applications. There are definitely pros and cons with that but its something that needs to be considered

Agreed.

inject/src/main/java/io/micronaut/context/annotation/MicronautApplication.java
melix: Removed

graemerocher:
> Given that this is experimental and there is no need for @MicronautApplication at this point I'd prefer to leave it out and add it later if the need arises.

Yeah I raised this concern with Cedric before and generally agree

> This opens up the possibility for libraries to customize how the context is created for applications. There are definitely pros and cons with that but its something that needs to be considered

I see this as a massive benefit actually, in the future we could for example move the environment detection logic to the AWS/GCP/Oracle Cloud modules with this feature as they could define context customizers that do that detection there instead of this being hard coded in core for every project.

graemerocher:
Also I agree with @melix the current javadoc for `ApplicationContextBuilder` states that the methods return `This builder` and in fact the code is not designed in a way whereby a different impl can be returned so this is non-issue

jameskleeh:
@graemerocher @melix The javadoc is not consistent:

```
    /**
     * Specifies to eager init the given annotated types.
     *
     * @param annotations The annotation stereotypes
     * @return The context builder
     * @since 2.0
     */
    @NonNull ApplicationContextBuilder eagerInitAnnotated(Class<? extends Annotation>... annotations);

    /**
     * Override default config locations.
     *
     * @param configLocations The config locations
     * @return This environment
     * @since 2.0
     */
    @NonNull ApplicationContextBuilder overrideConfigLocations(String... configLocations);

    /**
     * Set the command line arguments.
     *
     * @param args The arguments
     * @return This application
     */
    default @NonNull ApplicationContextBuilder args(@Nullable String... args) {
        return this;
    }
```

I don't understand why allowing a different instance to be returned is an issue. Wouldn't it just be a matter of doing the following?

```
      return new ApplicationContextCustomizer() {

            @Override
            public ApplicationContextBuilder customize(ApplicationContextBuilder builder) {
                for (ApplicationContextCustomizer customizer : customizers) {
                    builder = customizer.customize(builder);
                }
                return builder;
            }

        };
```

jameskleeh:
Also what about the creation of the builder in the first place? Should we not allow that to be extensible?

graemerocher:
> Also what about the creation of the builder in the first place? Should we not allow that to be extensible?

It is not extensible today so how is that relevant to this PR?

> The javadoc is not consistent

Then that is a bug in the javadoc, and the javadoc is incorrect. As the original author of this code I will send another PR to the documentation to clarify that `this` should always be returned, also as above if it is not extensible today the issue or overriding is a non-issue.

I don't believe we should be supporting returning a different instance from each method, that will complicate things a great deal and mean we will have to factor in delegation and provide a delegating impl that users can extend all of which we don't do today and I don't see how it is relevant to this PR

melix:
> I don't understand why allowing a different instance to be returned is an issue

Typically because `Micronaut` _extends from_ `DefaultApplicationBuilder`. Currently the way it's integrated is by calling the customizers when the default application builder is instantiated, which means that `Micronaut` automatically gets the customizations.

If you change the semantics from "mutates this builder" to "returns a new builder", then it means that _any_ call to `new Micronaut().foo()` would return a potentially new builder. So either you'd have to instrument bytecode to capture all `Micronaut()` constructor calls, and have the customizers aware of this subtype so that they return `Micronaut` instances, or we break the contract.

Alternatively, we would have to change `Micronaut` so that it doesn't _extend from_ `DefaultApplicationContextBuilder`, but instead _delegates to_ an internal builder. It also means we'd have to capture every single call site where a new builder is created. If `Micronaut` was using a factory pattern like `ApplicationContext#builder(...)` we could do something, but this isn't the case: you cannot change the returned instance from a constructor...

jameskleeh:
@melix I understand now - thanks

melix:
Great, I think I addressed all the comments above. Are you good if I merge this?

inject/src/main/java/io/micronaut/context/ApplicationContextCustomizer.java
```diff
@@ -0,0 +1,53 @@
...
+     * A default customizer which does nothing.
+     */
+    ApplicationContextCustomizer NO_OP = new ApplicationContextCustomizer() {
+    };
+
+    /**
+     * Customizes the application context builder.
+     * @param builder the builder to customize
+     */
+    default void customize(@NonNull ApplicationContextBuilder builder) {
```
jameskleeh, line 50: Is there a reason why this method has to be `default`?

inject-java/src/main/java/io/micronaut/annotation/processing/ServiceDescriptionProcessor.java
```diff
@@ -87,6 +102,39 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
     }
 
+    private void processStartupAnnotation(List<io.micronaut.inject.ast.Element> originatingElements, Element element, TypeElement typeElement, String name) {
+        Optional<AnnotationValue<ContextConfigurer>> ann = annotationUtils.getAnnotationMetadata(element).findAnnotation(ContextConfigurer.class);
+        if (ann.isPresent()) {
+            List<? extends TypeMirror> interfaces = typeElement.getInterfaces();
+            for (TypeMirror interfaceType : interfaces) {
+                if (interfaceType instanceof DeclaredType) {
+                    String serviceName = modelUtils.resolveTypeName(interfaceType);
+                    String serviceImpl = modelUtils.resolveTypeName(element.asType());
```
graemerocher, line 116: Perhaps fail compilation if the type has any constructor arguments this will catch cases where people adding constructor args to `Application`

graemerocher (changes requested):
One small change @melix

inject/src/main/java/io/micronaut/context/ApplicationContextCustomizer.java
melix: Yes: this will be used for Micronaut AOT to discover the entry point, so it's useful even if there's actually no customization at all.

jameskleeh: Can you explain how you envision that working?

melix: Sure. The prototype is already working, the current state of the code is on this PR: https://github.com/micronaut-projects/micronaut-aot/pull/2

So if you want to use AOT, you'd have to write, minimally:

```
public class Application {
    public static void main(String... args) {
         Micronaut.run(Application.class, args)
    }
}
```

This is enough for AOT to "know" that there are no customizations to the application context. If you actually want to customize the context, then you _have to_ use the new pattern:

```
@ContextConfigurer
public class Application implements ApplicationContextConfigurer {
    @Override
    public void customize(ApplicationContextBuilder builder) {
       builder.deduceEnvironment(false);
    }

    public static void main(String... args) {
         Micronaut.run(Application.class, args)
    }
}
```

The AOT module is, at build time, going to scan for the `@ContextConfigurer` annotation (via the Micronaut core runtime), except that instead of starting the application, it creates its own `ApplicationContextBuilder` which is going to be "customized by" the `Application` class (in the example above).

Then AOT can introspect the resulting application context to infer a number of things (environments being used, beans, ...) and perform a number of optimizations.

The advantage of this technique is that we're not dependent on the `Application` use case: this would work similarly for anything which will eventually create an application context, such as AWS functions.

jameskleeh: So if someone wants to customize the context they would implement the interface. What use case is there for not implementing the customize method?

>> This is enough for AOT to "know" that there are no customizations to the application context

How can it know that? Are you scanning the code of the method to know that `Micronaut.build` is not being executed? What if it is?

melix: No, it's just that we will document that AOT will only work properly if you follow the new pattern. So if we can't find any `@ContextConfigurer`, we _assume_ that there's no configuration. That is, if you write:

```
@ContextConfigurer
public class Application implements ApplicationContextConfigurer {
    @Override
    public void customize(ApplicationContextBuilder builder) {
       builder.deduceEnvironment(false);
    }

    public static void main(String... args) {
         Micronaut.build(args)
            .mainClass(Application.class)
            .environments("...") // This would not be noticed by AOT!
            .start()
    }
}
```

Then the customization done in the `main()` method is transparent to AOT. Eventually we'll probably want to deprecate configuration that we cannot introspect.

jameskleeh: Ok then so in the case where the user is not customizing the context, and its assumed there are no customizations if there are no services found, what is the use case for implementing the interface without implementing the method?

melix: This would allow us to scan the classpath and identify all configurers. If there's none we can warn the user that their app might not be compatible with AOT. If they supply one, even without configuration, we can assume they are compatible.

inject-java/src/main/java/io/micronaut/annotation/processing/ServiceDescriptionProcessor.java
melix: added

jameskleeh: Instead of failing if there is a constructor, it should fail if there is not a no-arg constructor. It's somewhat common for people to use the application class as a bean. This change would make it so that wouldn't be possible if the bean has dependents.

melix: The pattern if you want to use the Application as a bean would be the one exercised in this test: https://github.com/melix/micronaut-core/blob/63d8cec35e9309ebc83ec7235b7749e38513880a/test-suite/src/functionalTest/groovy/io/micronaut/context/ApplicationContextCustomizerSpec.groovy#L42-L79

Basically you don't annotate the `Application` with `@ContextConfigurer`, you just create a separate configurer.

jameskleeh: OK can you update the error message to indicate to the user what they should do?

melix: Would something like this work?:

> demo.app.Application is annotated with @ContextConfigurer but has at least one constructor with arguments, which isn't supported. You can workaround by creating a public static inner class instead, annotated with @ContextConfigurer, which sole role is configuring the application context.

jameskleeh: That would work for Java and Groovy users, however I'm not sure how Kotlin users would interpret that

melix: I'm not sure I can detect if this error message is thrown in a Kotlin environment, in which case you'd just use a companion object or a `@JvmStatic` I guess.

graemerocher: Perhaps remove the reference to `public static inner` since it can be any class nested inside or outside of the `Application`. Just write:

> demo.app.Application is annotated with @ContextConfigurer but has at least one constructor with arguments, which isn't supported. To resolve this create a separate class with no constructor arguments annotated with @ContextConfigurer, which sole role is configuring the application context.

graemerocher (approved):

jameskleeh:
I think the final thing I'd like to see considered is how this impacts `@MicronautTest`. Currently the behavior is the configuration in the application class has no impact on the test environment. For some users this is an issue and they use a custom `ApplicationContextBuilder` class that gets instantiated via the test extensions. 

For those that prefer the current behavior there is no way to separate the test and main contexts while still benefiting from the optimizations. What plan is there to support this use case?

melix:
>  For some users this is an issue and they use a custom ApplicationContextBuilder class that gets instantiated via the test extensions.

For this use case you'd just have to create a `@ContextConfigurer` which appears on the test classpath. Because the "main" classpath is also visible to the test classpath (in general), it means that customizations which apply to the `Application` case would also apply to tests. There are several things you can do to avoid this, if you want to:

1. you can separate your application entry point, in a separate source set or separate module, so that the main logic isn't "customized" while the runtime app is
2. you could have a flag in the "main" customizer which, based on some system property or something else, decides _not_ to apply the customizations
3. you could have the test customizer override whatever the main one does (this is assuming that the builder provides _override_ methods, not simply addition ones)
4. you could have separate modules for the customizers, which are added on demand on classpath (via build tools)

I don't see any of those as blockers today, but they may indeed impact the implementation, which is why I'm keen on keeping it `@Experimental` for at least a few releases. I just need to make progress on this PR so that I have a base to work with Micronaut AOT.

graemerocher:
This is actually major benefit to this as feature because there are also cases whereby you do want to have shared code that runs for both the application and test and currently there is no way to do that with `@MicronautTest` since it doesn't invoke directly the `Application` class and the user has to use the awkward approach is modifying the context builder via the `@MicronautTest` annotation.

PR #6408, opened 2021-10-25
Allow customization of http clients created through HttpClientRegistry 

### LINKED ISSUE #4829: ChannelPipelineCustomizer not invoked for DefaultClient when using declarative clients
### Task List

### Steps to Reproduce

1. create client ChannelPipelineCustomizer
2. create declarative client with `@Client('client-name')` annotation 
3. adjust ClientChannel

### Expected Behaviour

`BeanCreatedEventListener<ChannelPipelineCustomizer>` gets invoked for DefaultHttpClient

### Actual Behaviour

does not get invoked

### Environment Information

- **Operating System**: macosx 11.1
- **Micronaut Version:** 2.2.3
- **JDK Version:** openjdk 11.0.9.1 2020-11-04

### Example Application

https://github.com/jaecktec/micronaut-channel-customizer-bug

yawkat:
The bug is targeted for a patch release but not sure if that's a good idea for this patch since this is kind of a feature.

Fixes #4829


yawkat:
Okay, the test failure looks like now sometimes the same client is "customized" twice. I *think* the problem is that `DefaultNettyHttpClientRegistry` can create http clients both through the application context (it's a `@Factory`) and through the `HttpClientRegistry` API. The former approach (presumably) leads to the app context calling the customizer, while the latter approach does not. So, the two approaches have to be handled differently.

Added difficulty is that http clients are cached by `ClientKey`, so the same client may be used with both approaches at different use sites.

yawkat:
Duplicate calls to the customizer are even a problem without this patch:

```groovy
    def 'duplicate customization for normal client'() {
        given:
        def ctx = ApplicationContext.run([
                'spec.name': 'DefaultNettyHttpClientRegistrySpec',
                'micronaut.http.services.test-client.url': 'https://micronaut.io'
        ])
        def customizer = ctx.getBean(MyCustomizer)

        when:
        def client1 = ctx.getBean(ClientHolder).client
        def client2 = ctx.getBean(ClientHolder2).client
        try {
            client1.toBlocking().retrieve('/')
            client2.toBlocking().retrieve('/')
        } catch (Exception e) {
            // usually onConnect should still be called, but print anyway in case this test fails
            e.printStackTrace()
        }

        then:'no duplicate customization'
        !customizer.duplicate
    }

    @Requires(property = 'spec.name', value = 'DefaultNettyHttpClientRegistrySpec')
    @Singleton
    static class MyCustomizer implements BeanCreatedEventListener<ChannelPipelineCustomizer> {
        static final AttributeKey<Boolean> CUSTOMIZED = AttributeKey.valueOf('micronaut.test.customized')

        def duplicate = false

        @Override
        ChannelPipelineCustomizer onCreated(BeanCreatedEvent<ChannelPipelineCustomizer> event) {
            event.bean.doOnConnect {
                if (it.channel().hasAttr(CUSTOMIZED)) {
                    duplicate = true
                }
                it.channel().attr(CUSTOMIZED).set(true)
                return it
            }
            return event.bean
        }
    }

    @Requires(property = 'spec.name', value = 'DefaultNettyHttpClientRegistrySpec')
    @Singleton
    static class ClientHolder {
        @Inject
        @Client('test-client')
        HttpClient client;
    }

    @Requires(property = 'spec.name', value = 'DefaultNettyHttpClientRegistrySpec')
    @Singleton
    static class ClientHolder2 {
        @Inject
        @Client('test-client')
        HttpClient client;
    }
```

This test fails, because `MyCustomizer` is called twice for the same client (once for each injection site). So there is *some* precedent that customizers should be able to deal with this case. However I suspect that most users of customizers don't consider this, like the `LogbookPipelineCustomizer` from the guides (which is the failing test in CI). 

I'm not sure how to solve this. The simple fix for this particular patch would be to "lock" the `DefaultHttpClient` after construction so no more customizers can be added (any further calls to `doOnConnect` are ignored). That seems hacky though.

@graemerocher thoughts? I think we should discuss this in the meeting tomorrow.

graemerocher (approved):

graemerocher:
locking the client seems the right way forward

yawkat:
@graemerocher would it be possible to make the `DefaultHttpClient` simply not implement `ChannelPipelineCustomizer`? If it's only constructed in the `DefaultNettyHttpClientRegistry` anyway, that would be a better solution

graemerocher:
Would that not be a breaking change since then it would not be possible to customize clients created through `@Inject HttpClient client`?

yawkat:
I'm not sure whether those are constructed through the registry too

graemerocher:
They should be yes, so maybe it is not a breaking change. but needs testing

yawkat:
okay, i'll amend this PR and see what CI says :)

jameskleeh (approved):
Great idea

yawkat:
is this still patch release level? I will merge in that case, or otherwise rebase onto 3.2

jameskleeh:
@yawkat Given that `DefaultHttpClient` is internal I don't see any breaking changes here so I'm in favor of 3.1.x. The combination of clients potentially getting customized twice and some not getting customized at all seems only like a bug to me.

PR #6410, opened 2021-10-25
Enrich object encoding with more generic information

graemerocher:
This PR adds the ability for a `JsonMapper` to obtain richer generic information about the object that is to be encoded.

Currently the Jackson implementation doesn't use this information and re-materialized everything via reflection downstream but a future object encoder could use this information.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1630,6 +1630,7 @@ public class DefaultHttpClient implements
      * @param requestURI             The URI of the request
      * @param requestContentType     The request content type
      * @param permitsBody            Whether permits body
+     * @param bodyType               The body type
      * @param onError                Called when the body publisher encounters an error
      * @param closeChannelAfterWrite Whether to close the channel. For stream requests we don't close the channel until disposed of.
      * @return A {@link NettyRequestWriter}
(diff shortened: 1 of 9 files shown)


yawkat:
is it possible that the type from `ROUTE_INFO` could diverge from the actual body type, maybe due to filters?

graemerocher:
It is possible perhaps an `isInstance` check is needed prior to picking the method that takes `Argument`

yawkat:
That may still not be enough for 100% of cases, e.g. with optional where an isInstance would return the same. Maybe the type could be carried in the HttpResponse instead and dropped when the body is modified? Is that feasible? Otherwise this looks good

yawkat (approved):

jameskleeh:
I agree with @yawkat that the case of the body being set in a filter needs to be handled. As suggested we could try to remove the route info when the body is set

graemerocher:
I think that could be a potential breaking change though and also means we have to add tracking to the response for every call that sets the body which will have to be maintained across implementations which I don't like.

What is the issue with just verifying that the message object is an instance of the emitted type? 

With regards to wrapper types like `Optional` these are already handled by `RouteInfo.getBodyType` which has logic to unwrap the actual type.

Ultimately this is non-invasive change that doesn't alter the current codepath in anyway and is merely for downstream codec implementations to potentially benefit from additional type information. These future downstream codec impls can even add checks that the passed type is an instance and fail if necessary rather than us trying to complicate and potentially break existing usages

yawkat:
Yea fair point, it would be a more risky change if the jackson mapper already used the additional type info (e.g. `writerFor(JavaType)`), but it doesn't so this won't break anything. And if Optional is already handled, it's probably fine with the instance check – can't think of other realistic cases where this would matter

graemerocher:
@jameskleeh can you approve this one? It impacts performance of the serialization library if we don't have this type info

graemerocher:
test failure is unrelated

PR #6415, opened 2021-10-25
Add method getFluentBeanProperties to JavaClassElement

ilopmar:
See https://github.com/micronaut-projects/micronaut-openapi/issues/593

This PR does two things:
- Adds the method `getFluentBeanProperties` to `ClassElement` to return the bean properties declared in a "fluent" way, like when using Lombok. It only adds implementation for Java because it doesn't make sense for Groovy.
- Extract some private methods with code used in `getBeanProperties` that is now used for both `getBeanProperties` and `getFluentBeanProperties` method.

Basically in the current micronaut-openapi implementation there is a helper class that duplicates all the code extracted here to private methods. With this PR merged in core (and some changes I've already done in micronaut-openapi) I can remove that class and avoid using core's internal classes. 
Without this merged micronaut-openapi won't use now internal classes but as mentioned before we will need to duplicate a lot of code and keep it in sync.

One final thought. I wasn't sure about including this new method in `ClassElement` instead of in `JavaClassElement` so I can change it to the child class if necessary. Being in `ClassElement` simplifies even more micronaut-openapi because there is no need to check if code is Java or Groovy, though.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
@@ -35,6 +35,7 @@ import javax.lang.model.element.*;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.type.TypeVariable;
+import javax.lang.model.util.ElementFilter;
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 import java.util.*;
@@ -76,12 +77,15 @@ public class JavaClassElement extends AbstractJavaElement implements ArrayableCl
     /**
      * Used by OpenAPI.
      *
+     * Deprecated. This can be removed in Micronaut 3.2.0
+     *
      * @param classElement       The {@link TypeElement}
      * @param annotationMetadata The annotation metadata
      * @param visitorContext     The visitor context
      * @param genericsInfo       The generic type info
      * @param arrayDimensions    The number of array dimensions
      */
+    @Deprecated
     JavaClassElement(
             TypeElement classElement,
             AnnotationMetadata annotationMetadata,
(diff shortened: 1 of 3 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
```diff
@@ -76,12 +77,15 @@ public JavaClassElement(TypeElement classElement, AnnotationMetadata annotationM
...
      *
+     * Deprecated. This can be removed in Micronaut 3.2.0
+     *
      * @param classElement       The {@link TypeElement}
      * @param annotationMetadata The annotation metadata
      * @param visitorContext     The visitor context
      * @param genericsInfo       The generic type info
      * @param arrayDimensions    The number of array dimensions
      */
+    @Deprecated
```
ilopmar, line 88: If this PR is merged I will send a new PR to remove this for Micronaut 3.2.0 when I release a new version of micronaut-openapi

jameskleeh:
Can you explain why it doesn't make sense for Groovy or Kotlin?
Does the `getFluentBeanProperties` include only the fluent properties? What is the use case for caring if the property is fluent or not?

graemerocher:
Personally I think we should instead add support for the element query API and allow:

`ClassElement.getEnclosedElements(ElementQuery.of(PropertyElement.class))` 

Rather than adding yet another method

ilopmar:
The `JavaClassElementExt` class was created sometime ago in micronaut-openapi and currently extends from the `@Internal` `JavaClassElement` was introduced originally to support fluent-like getters and setters like the ones created with Lombok. The test I've added to this PR is the same that was introduced in micronaut-openapi (and is the only one that is failing at this moment in my work-in-progress changes).

As far as I know Lombok only works with Java. Maybe is debatable that this pattern can be used with Groovy or Kotlin. From a Groovy point of view I've never used that patter before, I've only used Groovy's `@Builder` annotation when needed something different. I don't know too much about Kotlin but I think Kotlin users are used to use data classes that don't generate these setters/getters.

ilopmar:
I will explore @graemerocher's suggestion as it seems a better approach :)

ilopmar:
Sorry, I've merged this by mistake :man_facepalming: . Reverting it now in #6477

Issue #6269, opened 2021-10-04
Error subscribing to @OnClose handler for WebSocket bean [..Controller]  Broadcast Failure: null


mrSingh007:
### Actual Behaviour

When in a browser more then 4/5 tabs are opened and connected with socket. On Closing browser this following exception is thrown.

```
09:11:23.136 [globalEventExecutor-2-6488] ERROR i.m.h.s.n.w.NettyServerWebSocketHandler - Error subscribing to @OnClose handler for WebSocket bean [notification.hub.web
socket.controller.WebsocketController@4c5f9154]: Broadcast Failure: null
io.micronaut.websocket.exceptions.WebSocketSessionException: Broadcast Failure: null
        at io.micronaut.http.netty.websocket.NettyServerWebSocketBroadcaster.lambda$null$2(NettyServerWebSocketBroadcaster.java:84)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:578)
        at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:552)
        at io.netty.util.concurrent.DefaultPromise.access$200(DefaultPromise.java:35)
        at io.netty.util.concurrent.DefaultPromise$1.run(DefaultPromise.java:502)
        at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:240)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:831)
Caused by: io.netty.channel.group.ChannelGroupException: null
        at io.netty.channel.group.DefaultChannelGroupFuture$1.operationComplete(DefaultChannelGroupFuture.java:73)
        at io.netty.channel.group.DefaultChannelGroupFuture$1.operationComplete(DefaultChannelGroupFuture.java:48)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:578)
        at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:552)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:491)
        at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:616)
[...]

PR #6419, opened 2021-10-26
Don't attempt to write ABNORMAL_CLOSURE to closed websocket

yawkat:
Appeared while writing tests for #6269.

Test case is as follows:

```groovy
    def 'closing channels simultaneously does not log an error'() {
        given:
        def ctx = ApplicationContext.run([
                'spec.name'            : 'BroadcasterSpec',
                'micronaut.server.netty.worker.threads': 1,
                'micronaut.server.port': 1234, // -1 todo
        ])
        def embeddedServer = ctx.getBean(EmbeddedServer)
        embeddedServer.start()

        def client = ctx.getBean(WebSocketClient)

        def uri = "ws://$embeddedServer.host:$embeddedServer.port/ws"

        def sockets = new ArrayList<Client>()
        for (int i = 0; i < 1; i++) {
            sockets.add(Flux.from(client.<Client>connect(Client, uri)).blockFirst())
        }
        for (def socket : sockets) {
            socket.close()
        }

        expect:
        ctx.getBean(Server).errors.isEmpty()

        cleanup:
        ctx.close()
    }

    @Singleton
    @Requires(property = 'spec.name', value = 'BroadcasterSpec')
    @ServerWebSocket("/ws")
    static class Server {
        @Inject WebSocketBroadcaster broadcaster
        List<Throwable> errors = Collections.synchronizedList(new ArrayList<>())

        @OnMessage
        def onMessage(String msg) {}

        @OnClose
        def onClose() {
            broadcaster.broadcast('foo').subscribe(new Subscriber<String>() {
                @Override
                void onSubscribe(Subscription s) {
                }

                @Override
                void onNext(String s) {
                }

                @Override
                void onError(Throwable t) {
                    errors.add(t)
                }

                @Override
                void onComplete() {
                }
            })
        }
    }

    @ClientWebSocket
    static abstract class Client implements AutoCloseable {
        @OnMessage
        def onMessage(String msg) {}
    }
```

Without this patch, the test case passes, but logs the following error:

```
08:59:16.413 [default-nioEventLoopGroup-1-16] WARN  i.n.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.channel.ChannelPipelineException: io.micronaut.http.client.netty.websocket.NettyWebSocketClientHandler.handlerRemoved() has thrown an exception.
	at io.netty.channel.DefaultChannelPipeline.callHandlerRemoved0(DefaultChannelPipeline.java:640)
	at io.netty.channel.DefaultChannelPipeline.destroyDown(DefaultChannelPipeline.java:876)
	at io.netty.channel.DefaultChannelPipeline.destroyUp(DefaultChannelPipeline.java:844)
	at io.netty.channel.DefaultChannelPipeline.destroy(DefaultChannelPipeline.java:836)
	at io.netty.channel.DefaultChannelPipeline.access$700(DefaultChannelPipeline.java:46)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelUnregistered(DefaultChannelPipeline.java:1392)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:198)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:184)
	at io.netty.channel.DefaultChannelPipeline.fireChannelUnregistered(DefaultChannelPipeline.java:821)
	at io.netty.channel.AbstractChannel$AbstractUnsafe$8.run(AbstractChannel.java:839)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:469)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:500)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.IllegalArgumentException: WebSocket close status code does NOT comply with RFC-6455: 1006
	at io.netty.handler.codec.http.websocketx.CloseWebSocketFrame.requireValidStatusCode(CloseWebSocketFrame.java:209)
	at io.netty.handler.codec.http.websocketx.CloseWebSocketFrame.<init>(CloseWebSocketFrame.java:69)
	at io.micronaut.http.netty.websocket.AbstractNettyWebSocketHandler.writeCloseFrameAndTerminate(AbstractNettyWebSocketHandler.java:579)
	at io.micronaut.http.netty.websocket.AbstractNettyWebSocketHandler.writeCloseFrameAndTerminate(AbstractNettyWebSocketHandler.java:460)
	at io.micronaut.http.netty.websocket.AbstractNettyWebSocketHandler.handleCloseReason(AbstractNettyWebSocketHandler.java:496)
	at io.micronaut.http.netty.websocket.AbstractNettyWebSocketHandler.handlerRemoved(AbstractNettyWebSocketHandler.java:271)
	at io.netty.channel.AbstractChannelHandlerContext.callHandlerRemoved(AbstractChannelHandlerContext.java:946)
	at io.netty.channel.DefaultChannelPipeline.callHandlerRemoved0(DefaultChannelPipeline.java:637)
	... 16 common frames omitted
```

The `close()` call leads to `handlerRemoved` being called, which tries to `handleCloseReason` with the reason `ABNORMAL_CLOSURE`. This would be correct if there was an `@OnClose` method, but since there isn't, `handleCloseReason` attempts to write back the close to the remote. However, `ABNORMAL_CLOSURE` isn't an allowed close reason on the wire, so there is an exception. The write would fail anyway, because the channel is already closed at this point.

This patch simply disables the write branch for the `ABNORMAL_CLOSURE` case.

PR doesn't include a unit test yet, because I don't know how to intercept this error. It's one of those netty errors that only pollutes the log but doesn't actually cause any API call to fail or any broken network traffic.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
@@ -268,7 +268,9 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
 
     @Override
     public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
-        handleCloseReason(ctx, CloseReason.ABNORMAL_CLOSURE);
+        // don't write this close reason, only call the @OnClose handler.
+        // ABNORMAL_CLOSURE isn't allowed on the wire anyway
+        handleCloseReason(ctx, CloseReason.ABNORMAL_CLOSURE, false);
     }
 
     /**
@@ -464,8 +466,9 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
      * Used to close thee session with a given reason.
      * @param ctx The context
      * @param cr The reason
+     * @param writeCloseReason Whether to allow writing the close reason to the remote
      */
-    private void handleCloseReason(ChannelHandlerContext ctx, CloseReason cr) {
+    private void handleCloseReason(ChannelHandlerContext ctx, CloseReason cr, boolean writeCloseReason) {
         cleanupBuffer();
         if (closed.compareAndSet(false, true)) {
             if (LOG.isDebugEnabled()) {
@@ -491,14 +494,16 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
                     }
                 }
             } else {
-                writeCloseFrameAndTerminate(ctx, cr);
+                if (writeCloseReason) {
+                    writeCloseFrameAndTerminate(ctx, cr);
+                }
             }
         }
     }
 
     private void handleCloseFrame(ChannelHandlerContext ctx, CloseWebSocketFrame cwsf) {
         CloseReason cr = new CloseReason(cwsf.statusCode(), cwsf.reasonText());
-        handleCloseReason(ctx, cr);
+        handleCloseReason(ctx, cr, true);
     }
 
     private void invokeAndClose(ChannelHandlerContext ctx, Object target, BoundExecutable boundExecutable, MethodExecutionHandle<?, ?> methodExecutionHandle, boolean isClose) {


yawkat:
`ChannelPipelineCustomizer` can't be used either to test this patch. I tried adding a handler to the end of the pipeline using the customizer to catch the error, but since the error happens when the pipeline is cleaned up, that handler is removed before the error happens.

I think I'll just have to skip unit testing on this one. This only fixes a warning anyway, so there won't be high impact if there is ever a regression on this.

graemerocher (approved):

graemerocher:
@yawkat could you update the PR body so the merge commit looks prettier

PR #6425, opened 2021-10-26
Support public-only fields and methods for introspections

graemerocher:
Adds a new `Visibility` enum such that it can be configured that only public fields and methods are taken into account.


graemerocher:
Currently only setting `PUBLIC` or `DEFAULT` is supported, other visibility modifiers can be added in the future, but will likely require the addition of reflective calls

PR #6432, opened 2021-10-27
Don't copy content headers for redirect from POST request

### LINKED ISSUE #6417: ReadTimeoutException testing authentication: cookie using junit5
### Expected Behavior

Using the following application.yml:

```
micronaut:
  security:
    authentication: cookie
```
and the following AuthenticationProvider

```
@Singleton
public class AuthenticationProviderUserPassword implements AuthenticationProvider {
    @Override
    public Publisher<AuthenticationResponse> authenticate(@Nullable HttpRequest<?> httpRequest,
                                                          AuthenticationRequest<?, ?> authenticationRequest) {
        return Flux.create(emitter -> {
                emitter.next(AuthenticationResponse.success((String) authenticationRequest.getIdentity()));
                emitter.complete();
        }, FluxSink.OverflowStrategy.ERROR);
    }
}
```

Running the following unit test:

```
    @Test
    void testAuth() {
        HttpRequest request = HttpRequest.POST("/login", new UsernamePasswordCredentials("user", "pass"));
        HttpResponse<BearerAccessRefreshToken> rsp = client.toBlocking().exchange(request, BearerAccessRefreshToken.class);
        assertEquals(HttpStatus.OK, rsp.getStatus());
    }
```
The expected behavior is the test passes successfully.

### Actual Behaviour

The above code results in the following error:

`io.micronaut.http.client.exceptions.ReadTimeoutException: Read Timeout
`

Note it runs successfully with authentication: bearer.

This appears to be due to the netty event loop.  The bearer authentication method only requests /login where the cookie authentication method requests /login and then re-directs to the success page.  The timeout appears to be waiting for the success page.

I attempted to create a secondary netty server for the HttpClient as described in the following without success

https://docs.micronaut.io/latest/guide/#secondaryServers

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.0.1

yawkat:
Exclude headers like Content-Type and Content-Length from copying request headers when the first request gets a redirect. There is no spec on which headers should be copied, so this is a blocklist with a few relevant headers, for now.
Fixes #6417

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -223,6 +223,27 @@ public class DefaultHttpClient implements
     private static final int DEFAULT_HTTP_PORT = 80;
     private static final int DEFAULT_HTTPS_PORT = 443;
 
+    /**
+     * Which headers <i>not</i> to copy from the first request when redirecting to a second request. There doesn't
+     * appear to be a spec for this. {@link java.net.HttpURLConnection} seems to drop all headers, but that would be a
+     * breaking change.
+     * <p>
+     * Stored as a {@link HttpHeaders} with empty values because presumably someone thought about optimizing those
+     * already.
+     */
+    private static final HttpHeaders REDIRECT_HEADER_BLOCKLIST;
+
+    static {
+        REDIRECT_HEADER_BLOCKLIST = new DefaultHttpHeaders();
+        // The host should be recalculated based on the location
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.HOST, "");
+        // post body headers
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.CONTENT_TYPE, "");
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.CONTENT_LENGTH, "");
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.TRANSFER_ENCODING, "");
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.CONNECTION, "");
+    }
+
     protected final Bootstrap bootstrap;
     protected EventLoopGroup group;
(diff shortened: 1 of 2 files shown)


http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -223,6 +223,27 @@
...
+    static {
+        REDIRECT_HEADER_BLOCKLIST = new DefaultHttpHeaders();
+        // The host should be recalculated based on the location
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.HOST, "");
+        // post body headers
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.CONTENT_TYPE, "");
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.CONTENT_LENGTH, "");
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.TRANSFER_ENCODING, "");
+        REDIRECT_HEADER_BLOCKLIST.add(HttpHeaderNames.CONNECTION, "");
+    }
```
graemerocher, line 245: A simple `Set<String>` would be sufficient here no?

yawkat: HttpHeaders is case-insensitive too. Generally it should have the right semantics for key comparison.

I used this approach in the server push api because HttpHeaders also presumably has optimizations for netty AsciiString, but that isn't relevant here because we only look up by String anyway.

asautins: Consider

```
    private static final HttpHeaders REDIRECT_HEADER_BLOCKLIST = new DefaultHttpHeaders()
            // The host should be recalculated based on the location
            .add(HttpHeaderNames.HOST, "")
            // post body headers
            .add(HttpHeaderNames.CONTENT_TYPE, "")
            .add(HttpHeaderNames.CONTENT_LENGTH, "")
            .add(HttpHeaderNames.TRANSFER_ENCODING, "")
            .add(HttpHeaderNames.CONNECTION, "");
```

PR #6438, opened 2021-10-27
Bump micronaut-servlet to 3.1.0

micronaut-build:


ilopmar:
@jameskleeh does micronaut-servlet 3.1.0 fixes something fundamentally broken? I'm asking because we shouldn't do minor version upgrades in patch releases. Unless, of course, there is something really wrong.

PR #6439, opened 2021-10-27
Sort service file entries

### LINKED ISSUE #6437: Consistent META-INF/services/io.micronaut.inject.BeanDefinitionReference generation
### Feature description

I am trying to get my build to repeatably build the same docker image. When repeating a build, lines in META-INF/services/io.micronaut.inject.BeanDefinitionReference change order randomly causing the jar file hash to change and creating a different docker layer rather than matching the previously built one. It would be great (and probably relatively easy) to sort the contents of this file to make builds repeatable.

Thank you for the consideration

jameskleeh:
Fixes #6437

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/writer/AbstractClassWriterOutputVisitor.java
+++ b/inject/src/main/java/io/micronaut/inject/writer/AbstractClassWriterOutputVisitor.java
@@ -25,6 +25,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


jameskleeh:
@dstepanov If you can take a look at this I would appreciate it. Somehow sorting the service file entries is causing ReactorContextPropagationSpec to fail

Edit: The service sorting is a red herring. The order of the filters is non deterministic in the test and if the order is changed the test fails without any changes on 3.1.x

jameskleeh:
See https://github.com/micronaut-projects/micronaut-core/pull/6454

dstepanov:
It looks like Kotlin's implementation sets the context instead of merging it:
```kotlin
internal class ReactorContextInjector : ContextInjector {

    override fun <T> injectCoroutineContext(publisher: Publisher<T>, coroutineContext: CoroutineContext): Publisher<T> {
        val reactorContext = coroutineContext[ReactorContext]?.context ?: return publisher
        return when(publisher) {
            is Mono -> publisher.contextWrite(reactorContext)
            is Flux -> publisher.contextWrite(reactorContext)
            else -> publisher
        }
    }
```

I would add the order for the test to pass and submit an issue for the Kotlin Reactor support.

jameskleeh:
@dstepanov Are you going to file the issue?

dstepanov:
I will

inject/src/main/java/io/micronaut/inject/writer/AbstractClassWriterOutputVisitor.java
```diff
@@ -94,7 +98,7 @@ public final void finish() {
     public void writeServiceEntries(Map<String, Set<String>> serviceEntries, Element... originatingElements) {
         for (Map.Entry<String, Set<String>> entry : serviceEntries.entrySet()) {
             String serviceName = entry.getKey();
-            Set<String> serviceTypes = entry.getValue();
+            Set<String> serviceTypes = new HashSet<>(entry.getValue());
```
graemerocher: Would it not be easier to use a TreeSet instead of copying into an array and sorting? Or am I missing something?

PR #6442, opened 2021-10-28
Support for tracing propagation with coroutines.

konrad-kaminski:
Tracing context propagation does not work with coroutines. This PR adds support for it. The tracing context (if exists) will be propagated in controller actions which are `suspend` functions. For other cases one must create `CoroutineTracingDispatcher` using `Supplier<CoroutineTracingDispatcher>` and use it in their `async/withContext/runBlocking/etc`.

Since `CoroutineTracingDispatcher` must be used by `ContinuationArgumentBinder` static function I decided to make the `Supplier<CoroutineTracingDispatcher>` a `@Context` bean and upon creation it will set a global variable (in `ContinuationArgumentBinder`) to itself (since there is no `http` -> `tracing` dependency and we probably don't want one). Alternatively we could have `ContinuationArgumentBinder` as a regular bean - that, however, would be a bigger change and not the one that would make the code more clear.


graemerocher:
Could you target the 3.2.x branch for this PR? Thanks

dstepanov:
I think it would be better for `ContinuationArgumentBinder` to have access to `BeanContext` (add it to `DefaultRequestBinderRegistry` ->  `registerDefaultAnnotationBinders`) and introduce some interface for allowing to extend the context. `ContinuationArgumentBinder` would use `BeanContext` to get all the implementations including the tracing contributor.

konrad-kaminski:
@dstepanov What you suggest would require a lot of dependency propagation. I changed my proposal a bit towards what you suggest, but retaining most of the code. Let me know what you think.

dstepanov:
@konrad-kaminski I did some changes of removing the static context, also I have changed to have only one factory instead of two. Is it something that makes sense? BTW Do you need to have it as context scoped beans?

dstepanov (approved):

dstepanov:
Made a bit more modifications to make the tests pass.

@graemerocher Can you also take a look?

graemerocher:
Looks ok, only change necessary is we need to add runtime init entries for the new beans that depend on kotlin on the classpath to avoid graal warnings, like:

https://github.com/micronaut-projects/micronaut-core/blob/a647fd76804d5c1706f1cdf563b66aba05582d3c/http/src/main/resources/META-INF/native-image/io.micronaut.http/micronaut-http/native-image.properties#L17

dstepanov:
Added

graemerocher (approved):

http/src/main/kotlin/io/micronaut/http/bind/binders/ContinuationArgumentBinder.kt
```diff
@@ -50,13 +50,16 @@ class ContinuationArgumentBinder : TypedRequestArgumentBinder<Continuation<*>> {
         private val reactorContextPresent: Boolean = ClassUtils.isPresent("kotlinx.coroutines.reactor.ReactorContext", null);
 
         @JvmStatic
-        fun setupCoroutineContext(source: HttpRequest<*>, contextView: ContextView) {
+        fun setupCoroutineContext(source: HttpRequest<*>, contextView: ContextView, continuationArgumentBinderCoroutineContextFactories: Collection<HttpCoroutineContextFactory<*>>) {
```
jameskleeh, line 53: I believe this is public correct? If so then this is a breaking change

jameskleeh (changes requested):

http/src/main/kotlin/io/micronaut/http/bind/binders/ContinuationArgumentBinder.kt
graemerocher: Denis is probably gone for the day. A simple revision could be to add another method with overloaded arguments. This class should really have been `@Internal`

jameskleeh: OK I'll update it

jameskleeh:
Closed in favor of #6552

PR #6450, opened 2021-10-29
Document BigDecimal deserialization behavior

### LINKED ISSUE #6429: BigDecimal precision is lost in integer cases (regression)
### Expected Behavior

When passing in BigDecimal values on POJOs the precision should not be lost. For some business use cases the precision and rounding of decimals can be extremely important. (I noticed this on the upgrade from 2.x -> 3.1 as some of our test cases started to fail.)

### Actual Behaviour

Precision is lost in integer cases (it increases the precision by 1 by converting for example 100 to 100.0)

This appears to be caused by this line, converting Integer to BigDecimal (if it hit the Long case above it would work, but it does not since it seems to be Integer):
https://github.com/micronaut-projects/micronaut-core/blob/813923882bcfed158d136a27bf8010edf70bb613/json-core/src/main/java/io/micronaut/json/tree/JsonNode.java#L233

### Steps To Reproduce

1. Create a POJO/DTO with a BigDecimal field
2. Send in an integer value such as new BigDecimal("100")
3. Try to maintain the precision and return the same value as was sent in (it cannot be done)

See sample app:

expected: <100> but was: <100.0>
Expected :100
Actual   :100.0

### Environment Information

Windows

### Example Application

https://github.com/zdegner/int-to-big-decimal

### Version

3.1.1

yawkat:
JacksonCoreProcessor does not use BigDecimal by default for storing numeric values for efficiency reasons. This patch documents this behavior.

Closes #6429

### DIFF
--- a/src/main/docs/guide/httpServer/jsonBinding.adoc
+++ b/src/main/docs/guide/httpServer/jsonBinding.adoc
@@ -111,3 +111,17 @@ In addition to configuration, beans can be registered to customize Jackson. All 
 === Service Loader
 
 Any modules registered via the service loader are also added to the default object mapper.
+
+=== Number Precision
+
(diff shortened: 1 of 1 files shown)


burtbeckwith (approved):
lgtm

PR #6452, opened 2021-10-29
`@JsonSerialize` support for BeanIntrospectionModule

### LINKED ISSUE #6309: @JsonSeralize annotation is ignored by GraalVM native image
### Expected Behavior

Putting a

```java
@JsonSerialize(using = ToStringSerializer.class)
```

on a field or should work as expected both in JVM and a GraalVM native image

### Actual Behaviour

When put on a field the annotation is ignored by GraalVM

### Steps To Reproduce

Annotate a field with `@JsonSerialize(using = ToStringSerializer.class)` and try to serailize it

### Environment Information

- Operating System: Windows (Native image was generated in WSL)
- JDK version: 11

### Example Application

https://github.com/r-darwish/jsonserialize-graal

### Version

3.0.3

yawkat:
Parse annotations for custom serializers when in non-reflective mode. Also adds support for `@JsonDeserialize`.
Uses introspections to instantiate serializers.

Fixes #6309

---

Introspections seem like the best way to construct the serializers, here. It would be better to use the SerializerProvider because that's cached by jackson and such, but it's not exposed. (It is exposed to JsonSerializer itself so we could add a "delegating" JsonSerializer that constructs the actual serializer late, but this doesn't seem ideal either)

I've also looked at implementing TypeSerializer support, but the logic there is more complicated still.

### DIFF
--- a/jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
+++ b/jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
@@ -25,6 +25,8 @@ import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.SerializableString;
 import com.fasterxml.jackson.core.io.SerializedString;
 import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.deser.BeanDeserializerBuilder;
 import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier;
 import com.fasterxml.jackson.databind.deser.CreatorProperty;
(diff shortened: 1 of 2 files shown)


jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
```diff
@@ -165,6 +168,28 @@ private static Object suppressableValue(JsonInclude.Value inclusion) {
...
     }
 
+    /**
+     * Parse a {@link JsonSerialize} or {@link JsonDeserialize} annotation.
+     */
+    private <T> T findSerializerFromAnnotation(BeanProperty<?, ?> beanProperty, Class<? extends Annotation> annotationType) {
+        AnnotationValue<?> jsonSerializeAnnotation = beanProperty.getAnnotation(annotationType);
+        if (jsonSerializeAnnotation != null) {
+            // ideally, we'd use SerializerProvider here, but it's not exposed to the BeanSerializerModifier
+            Class using = jsonSerializeAnnotation.get("using", Class.class).orElse(null);
```
graemerocher: Use `jsonSerializeAnnotation.classValue("using")` instead

jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
```diff
@@ -165,6 +168,28 @@ private static Object suppressableValue(JsonInclude.Value inclusion) {
...
+            if (using != null) {
+                BeanIntrospection<Object> usingIntrospection = findIntrospection(using);
+                if (usingIntrospection != null) {
+                    return (T) usingIntrospection.instantiate();
+                } else {
+                    if (LOG.isWarnEnabled()) {
+                        LOG.warn("Cannot construct {}, please enable introspection for that class", using.getName());
```
graemerocher: Change from `enable introspection for that class` to `add the @Introspected annotation to the class declaration`

graemerocher (changes requested):

graemerocher (approved):

PR #6459, opened 2021-11-01
Omit injected points for configuration property beans

### LINKED ISSUE #6305: Add a compiler flag to produce more optimized and smaller bean metadata
### Feature description

Currently a lot of bean metadata is rarely used at runtime (injection points etc.), we should add a flag to disable this metadata for user projects if they do not need it.

dstepanov:
This PR refactors a bit how `@Property`, `@Value`, and inner beans were resolved in generated bean definitions:
- `@Property`, `@Value` annotation value, and possible CLI property value is resolved at build time
- For setters, fields injection no injection point is created by default and can be enabled by an env variable

Fixes #6305

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,7 @@ subprojects { Project subproject ->
     tasks.withType(JavaCompile).configureEach {
        options.compilerArgs.add("-Amicronaut.processing.group=$project.group")
        options.compilerArgs.add("-Amicronaut.processing.module=micronaut-$project.name")
-       options.compilerArgs.add("-Amicronaut.omit.confprop.injectpoints=true")
+       options.compilerArgs.add("-Amicronaut.processing.omit.confprop.injectpoints=true")
        options.forkOptions.memoryMaximumSize = "2g"
     }
     tasks.withType(GroovyCompile).configureEach {
--- a/build.gradle
+++ b/build.gradle
@@ -105,6 +105,7 @@ subprojects { Project subproject ->
     tasks.withType(JavaCompile).configureEach {
        options.compilerArgs.add("-Amicronaut.processing.group=$project.group")
        options.compilerArgs.add("-Amicronaut.processing.module=micronaut-$project.name")
+       options.compilerArgs.add("-Amicronaut.omit.confprop.injectpoints=true")
        options.forkOptions.memoryMaximumSize = "2g"
     }
     tasks.withType(GroovyCompile).configureEach {
--- a/inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
+++ b/inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
@@ -15,9 +15,9 @@
  */
 package io.micronaut.inject.writer;
 
-import io.micronaut.context.AbstractInitializableBeanDefinition;
 import io.micronaut.context.AbstractConstructorInjectionPoint;
 import io.micronaut.context.AbstractExecutableMethod;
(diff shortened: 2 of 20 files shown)


dstepanov:
The possible future improvement we can parse `@Value` annotation value during compile-time and build a string

jameskleeh:
Is adding the system property to those tests necessary for them to pass? This seems like it should be true by default

dstepanov:
I can change it, it would be better to have the new implementation enabled, I don't think anyone was using injecting points.

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -424,6 +513,13 @@
     private boolean constructorRequiresReflection;
     private boolean disabled = false;
 
+    private final boolean keepConfPropInjectPoints;
+
+    {
+        String property = System.getProperty("micronaut.keep-conf-prop-inject-points");
+        keepConfPropInjectPoints = property != null && property.toLowerCase(Locale.ROOT).equals("true");
```
graemerocher: I don't think we should be using system properties here, if anything they should be annotation processor inputs.

However in this case I think we can do something simpler and just have `@ConfigurationProperties(metadata=false)` to disable build time metadata or true to include it since this applies only to configuration propeties. This will probably need to be a meta annotation with `@AliasFor` pointing to `@ConfigurationReader` and also be added to `@EachProperty` so it is consistent.

We can then perhaps debate whether this should default to disable metadata or not. Personally I don't think anyone is using this metadata in real world apps, but it is perhaps a breaking change.

graemerocher (changes requested):

inject/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
dstepanov: I think it would be better to have a custom annotation because it’s more of internal implementation.

graemerocher: Perhaps, but if it is a custom annotation then we would have to make it work for other cases and not just configuration properties, like for example if I don't need the injection points for regular `@Inject` fields / methods

dstepanov:
Added compile option.

New generated BD for configurations format: https://gist.github.com/dstepanov/2f6b538248daabb1dfc8cf882f559163
New generated BD for configurations with omitted injection points format: https://gist.github.com/dstepanov/48808e92b2c0ca1fdebd383eb6c2c784

build.gradle
```diff
@@ -105,6 +105,7 @@ subprojects { Project subproject ->
     tasks.withType(JavaCompile).configureEach {
        options.compilerArgs.add("-Amicronaut.processing.group=$project.group")
        options.compilerArgs.add("-Amicronaut.processing.module=micronaut-$project.name")
+       options.compilerArgs.add("-Amicronaut.omit.confprop.injectpoints=true")
```
graemerocher: Rename to `micronaut.processing.omit.confprop.injectpoints` and I believe you will need to add it to https://github.com/micronaut-projects/micronaut-core/blob/6af280483f86f1c42ce8b1c010aa1ddd90176b5e/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java#L77

graemerocher (changes requested):

build.gradle
dstepanov: Done

jameskleeh (approved):

graemerocher (approved):

PR #6460, opened 2021-11-01
Consider `@JsonView` annotation for request body deserialization

### LINKED ISSUE #5046: @JsonView annotation restriction does not work for request-body
### Task List

### Steps to Reproduce

I' am currently working on Micronaut Version 2.1.2 and I ran in to the 
problem that, I cannot restrict the request-body parameter of my application with the JsonView annotation, like in this example: 

``` java
  @Post("/ressources")
  Ressource createService(@JsonView(Views.Writable.class) @Body Ressource ressource);
```

Strangely the annotation works fine , if I use it to restrict the response-body/return-value.

### Expected Behaviour

All fields which are not included in the Views.Writable class should be not taken into account. I also looked up my Jackson version, because I read in [this](https://stackoverflow.com/questions/32292237/jsonview-how-to-use-with-requestbody) stack-overflow post, that Jackson-Version 2.5 is requierred for this feature. But I use Jackson-Version 2.11.3, so I think this shouldn't be a problem.

### Actual Behaviour

All fields which are not included in the Views.Writable class are taken into account.

### Environment Information

- **Operating System**: Ubuntu 20.04.2 LTS
- **Micronaut Version:** 2.1.2
- **JDK Version:** Corretto-15.0.2.7.1

### Example Application

yawkat:
Propagate the annotations on a `@Body` parameter to a controller through the deserialization stack, and use that annotation information in JacksonDatabindMapper to select the json view used for deserialization.

Resolves #5046

---

To propagate the `@JsonView` all the way to the mapper, I had to make some of the intermediates use `Argument` properly. This may have unintended consequences.

### DIFF
--- a/jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
+++ b/jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
@@ -22,7 +22,6 @@ import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.micronaut.context.annotation.BootstrapContextCompatible;
 import io.micronaut.core.annotation.AnnotationMetadata;
-import io.micronaut.core.annotation.AnnotationValue;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
@@ -86,7 +85,7 @@ public final class JacksonDatabindMapper implements JsonMapper {
     public <T> T readValueFromTree(@NonNull JsonNode tree, @NonNull Argument<T> type) throws IOException {
         JsonParser tokens = treeCodec.treeAsTokens(tree);
         JavaType javaType = JacksonConfiguration.constructType(type, objectMapper.getTypeFactory());
-        Optional<Class<?>> view = type.findAnnotation(JsonView.class).flatMap(AnnotationValue::classValue);
+        Optional<Class> view = type.getAnnotationMetadata().classValue(JsonView.class);
         if (view.isPresent()) {
             return objectMapper.readerWithView(view.get()).readValue(tokens, javaType);
         } else {
--- a/jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
+++ b/jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
@@ -15,10 +15,14 @@
  */
 package io.micronaut.jackson.databind;
 
+import com.fasterxml.jackson.annotation.JsonView;
+import com.fasterxml.jackson.core.JsonParser;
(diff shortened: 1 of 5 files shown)


yawkat:
Yea there it is: removing the `targetType.getTypeParameters().length != 0` check broke `io.micronaut.http.client.StreamRequestSpec#test stream post request with numbers`...

yawkat:
okay this change seems to solve the tests. It isn't great, but it's no worse than the previous approach imo.

json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
```diff
@@ -211,8 +211,11 @@ private Object correctKeys(Object o) {
                     return Optional.of(new String(objectCodec.get().writeValueAsBytes(node), StandardCharsets.UTF_8));
                 } else {
                     Argument<?> argument = null;
-                    if (context instanceof ArgumentConversionContext && targetType.getTypeParameters().length != 0) {
+                    if (context instanceof ArgumentConversionContext) {
```
jameskleeh, line 214: Do you know what the purpose of the check for no type parameters was? Why did it need to be removed?

graemerocher: I believe it was a performance optimization since there was no need to use the `Argument` if there was no generic type information and `targetType` could just be used

jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
```diff
@@ -80,7 +84,14 @@ public ObjectMapper getObjectMapper() {
 
     @Override
     public <T> T readValueFromTree(@NonNull JsonNode tree, @NonNull Argument<T> type) throws IOException {
-        return objectMapper.readValue(treeCodec.treeAsTokens(tree), JacksonConfiguration.constructType(type, objectMapper.getTypeFactory()));
+        JsonParser tokens = treeCodec.treeAsTokens(tree);
+        JavaType javaType = JacksonConfiguration.constructType(type, objectMapper.getTypeFactory());
+        Optional<Class<?>> view = type.findAnnotation(JsonView.class).flatMap(AnnotationValue::classValue);
```
graemerocher: Can you use `type.getAnnotationMetadat().classValue(JsonView.class)` instead. It will be more efficient as it doesn't have to create the whole annotation instance

graemerocher (changes requested):

json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
yawkat: This is the check that failed the first CI run, see the comment above. There seem to be cases where the argument in the context and the targetType diverge, similar to the response body argument problem graeme had

jameskleeh (approved):

graemerocher (approved):

PR #6466, opened 2021-11-02
Keeps the http method same while following redirection

### LINKED ISSUE #6434: Do not change HTTP method when handling 307 redirects
### Expected Behavior

The [HTTP spec](https://httpwg.org/specs/rfc7231.html#status.307) says:

> the user agent MUST NOT change the request method if it performs an automatic redirection to that URI

When a POST request returns a 307, the micronaut http client should follow the redirect using a second POST request.

### Actual Behaviour

The micronaut http client only follows redirects using GET even for 307.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.1.1

ashishkujoy:
closes: #6434


jameskleeh (changes requested):
The behavior for keeping the original request method should only happen for 307, not all redirects. Changing it for other methods would be a breaking change

ashishkujoy:
> The behavior for keeping the original request method should only happen for 307, not all redirects. Changing it for other methods would be a breaking change

Corrected.

graemerocher (approved):

PR #6482, opened 2021-11-05
Add new `@AccessorsStyle` annotation

ilopmar:
As discussed yesterday during the engineering meeting instead of trying to be smart to detect this kind of situations, we introduce a new experimental annotation so users can tell us that the accessors don't follow the standard pattern with `get` and `set` prefixes.
This add supports for treating Lombok "fluent" accessors like Records when generating the bean properties. In the future we may want to add more styles or any other features.

See https://github.com/micronaut-projects/micronaut-openapi/issues/593

P.S: Open to change the names to more appropriates


graemerocher:
I'm not sure about this for the following reasons:

* We already have precedent for another approach which is to specify a prefix like in https://github.com/micronaut-projects/micronaut-core/blob/6af280483f86f1c42ce8b1c010aa1ddd90176b5e/inject/src/main/java/io/micronaut/context/annotation/ConfigurationBuilder.java#L43-L53
* If we introduce this annotation then I would expect it to impact the behaviour of configuration properties, introspections, configuration builder (see above) as well and not do nothing
* This change introduces an enum, what if I want a write prefix of "with" and a read prefix of ""? What would we call that? Are we going to have fixed enum values for all variations? What if I want a custom strategy?
* Maybe AccessorStyle should be an interface with multiple implementations and users can extend with additional implementations (granted these implementations would need to be on the annotation processor classpath)

ilopmar:
This was something @jameskleeh suggested yesterday to make users tell us about this particular use case we want to fix in micronaut-openapi. We discussed only about this use case and no custom prefix or things like that.
I'll leave James to explain it better but what he suggested was have a proper support in core instead of the ugly hack we have in openapi with lot of code copy-pasted from core and still using one internal class (`JavaMethodElement`)

graemerocher:
Right, but the point remains that currently the added annotation doesn't impact introspections, configuration properties, configuration builder and everywhere else where you would expect it to impact so that may end up just creating more confusion

ilopmar:
I see your point, but If we make explicit that this is only for one specific use case... :shrug: Maybe we can call it `@LombokAccessorsHack` or something like that :-P

graemerocher:
I don't think this issue is important enough to introduce hacks in core, but I will take that as a joke

jameskleeh:
We do need a way for users to tell us what style their properties are following. Perhaps something like 

```
public @interface AccessorsStyle {

    String readPrefix() default "get"

    String writePrefix() default "set"

}
```

Then users can use `"with"` or `""` or any combination. This could then be allowed to work on all of the things Graeme mentioned.

ilopmar:
As we discussed the other day, I pushed a new commit that adds a new `@AccessorsStyle` annotation with the posibility of defining custom prefixes for getters and setters.

I'm submitting this earlier to get feedback (I'm also switching the PR to draft). With the current code we can "fix" the Lombok-style issue in micronaut-openapi.

If this approach looks good I can continue with the changes to make it work for introspections, configuration properties and configuration builder. There are more changes to be done in `JavaClassElement` (and `GroovyClassElement`) to take the prefixes into account, but I'll make them with the rest of the changes to avoid breaking tests.

ilopmar:
I was thinking that I'm not completely sure if allowing more than one read/write prefix makes sense. There are methods `getterNameFor(propertyName)` and `setterNameFor(propertyName)` in `NameUtils` that I would need to modify, but calling those methods like `getterNameFor("foo", ["set", "with"])` doesn't really make sense, because I need to return only one setter, so in this case I would always return `setFoo`, i.e, using the first prefix.
For other methods I've already modified: `isWriterName`, `getPropertyNameForSetter`,... it does make sense to pass more prefixes because I can iterate over the prefixes until one matches (as I've already done), but for getting the setter/getter name for a property, I can't see how that would work.

Can we go back to
```java
public @interface AccessorsStyle {

    String readPrefix() default "get"

    String writePrefix() default "set"
}
```

As James originally suggested?

jameskleeh:
The original motivation for using an array is that the `@ConfigurationBuilder` annotation allows for an array. There are also cases like inheritance where multiple styles should be supported. We would pick the first one that matches as they are defined in the annotation

ilopmar:
> We would pick the first one that matches as they are defined in the annotation

Then, for the case of `getterNameFor("foo", ["set", "with"])` (and the same for `setterNameFor`) we would always return `setFoo` because in this case there is no such thing as match.

ilopmar:
I've pushed a new commit https://github.com/micronaut-projects/micronaut-core/pull/6482/commits/346dfa075c735d20264024bd3e323a8bd3865e73 that adds support for `@ConfigurationProperties` with classes and interfaces.

I still need to add docs for this and also make the changes for `@ConfigurationBuilder`. In the mean time I would like to get some feedback about the approach I've followed and how this PR looks like.

Another thing is that I'm using a code like this:

```java
final String[] readPrefixes = declaringElement.findAnnotation(AccessorsStyle.class)
    .filter(annotationValue -> !ArrayUtils.isEmpty(annotationValue.stringValues("readPrefixes")))
    .map(annotationValue -> annotationValue.stringValues("readPrefixes"))
    .orElse(new String[]{AccessorsStyle.DEFAULT_READ_PREFIX});
final String[] writePrefixes = declaringElement.findAnnotation(AccessorsStyle.class)
    .filter(annotationValue -> !ArrayUtils.isEmpty(annotationValue.stringValues("writePrefixes")))
    .map(annotationValue -> annotationValue.stringValues("writePrefixes"))
    .orElse(new String[]{AccessorsStyle.DEFAULT_WRITE_PREFIX});
```

In a lot of places. I was looking to move it to some utility class and just pass the `declaringElement`, `classMethod`,... or whatever thing I need as a parameter. I'm not sure which class is a good candidate to add it, or maybe I need to add a new utility class for this.

ilopmar:
I've found out that I can do:

```java
final String[] readPrefixes = declaringElement.getValue(AccessorsStyle.class, "readPrefixes", String[].class)
                                .orElse(new String[]{AccessorsStyle.DEFAULT_READ_PREFIX});
```

which improve the readability a lot so I probably don't need the utility class anymore (although there is still a bit of code duplication)

ilopmar:
Ready for review @graemerocher @jameskleeh

ilopmar:
I've messed everything up when changing the base branch. Closing this PR and submitting https://github.com/micronaut-projects/micronaut-core/pull/6620

PR #6487, opened 2021-11-08
Support WebSocket permessage-deflate extension

### LINKED ISSUE #6069: Websocket Server not handling permessage-deflat WebSocket-Extension
### Feature description

When a client sends a websocket connection request it can request a permessage-deflate extension. This enables gzip compression on both sides per message during a websocket transaction. This is not enabled in netty by default without adding a WebSocketServerCompressionHandler to the pipeline

https://stackoverflow.com/questions/32187071/websocket-compression-in-netty/32194798

This causes some bandwidth overhead that could be reduced

yawkat:
Both for client and server.
On the server side, this requires a change in the pipeline setup logic. Because the permessage-deflate codecs are placed after the websocket codecs, but aren't always present, we can't simply place the `NettyServerWebSocketHandler` after the `wsdecoder` anymore. Instead, we place it at the old location of the upgrade handler.

Resolves #6069


graemerocher (approved):

yawkat:
Test failure seems to be caused by the compression codec consuming the input `ByteBuf` data when performing the compression. This isn't surprising ofc, but the test relies on the buffer still having the same reader indices after the write is complete. I've changed the code to send a slice of the ByteBuf down the pipeline instead, but imo we shouldn't guarantee that the buffer remains unmodified anyway.

PR #6489, opened 2021-11-08
Copy setter annotations for bean properties

### LINKED ISSUE #6472: Jackson deserialization ignoring JsonProperty annotations
### Expected Behavior

Given the class listed below, Jackson should properly deserialize the `access_token` and `expires_in` properties provided in the JSON input being deserialized, using the `BeanIntrospection` class generated through the `@Introspected` annotation.

```java
@Data @Introspected @JsonIgnoreProperties(ignoreUnknown = true)
public final class TokenResponse {
	private String accessToken;
	private long expiresAt;
	public String getAccessToken() {
		return accessToken;
	}
	@JsonProperty("access_token")
	public void setAccessToken(String accessToken) {
		this.accessToken = accessToken;
	}
	@JsonProperty("expires_in")
	public void setExpiresIn(long expiresIn) {
		this.expiresAt = new Date().getTime()+((expiresIn-5)*1000);
	}
	public long getExpiresAt() {
		return expiresAt;
	}
	@JsonIgnore public boolean isExpired() {
		return new Date().getTime() > expiresAt;
	}
}
```

### Actual Behaviour

Jackson correctly deserializes the JSON data when running in a normal JVM, but not in a GraalVM native image. When running as a native image, it seems like Jackson doesn't see the `@JsonProperty` annotations and therefore the setters are not being invoked.

Not sure whether it's relevant, but the generated `$TokenResponse$IntrospectionRef` class only lists class-level annotations and not method-level annotations.

### Steps To Reproduce

1. Write a small sample program that deserializes a JSON object containing `access_token` and `expires_in` properties to an instance of the Java class listed under 'Expected Behavior', using the Micronaut-provided `ObjectMapper`
2. Run this program in a normal JVM, and see Jackson properly deserializing the JSON string
3. Run this program as a native GraalVM image, and see Jackson fail to properly deserialize the JSON string, failing to invoke the setters and leaving the instance fields at their default value

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.1.0

yawkat:
Copy annotations for bean properties not only from the field and getter, but also from the setter.

Fixes #6472

### DIFF
--- a/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/utils/AstAnnotationUtils.groovy
+++ b/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/utils/AstAnnotationUtils.groovy
@@ -69,12 +69,12 @@ class AstAnnotationUtils {
      * Get the {@link AnnotationMetadata} for the given annotated node
      *
      * @param sourceUnit the source unit
-     * @param parent the parent
+     * @param parents the parents
      * @param annotatedNode The node
      * @return The metadata
      */
-    static AnnotationMetadata getAnnotationMetadata(SourceUnit sourceUnit, CompilationUnit compilationUnit, AnnotatedNode parent, AnnotatedNode annotatedNode) {
-        new GroovyAnnotationMetadataBuilder(sourceUnit, compilationUnit).buildForParent(parent, annotatedNode)
+    static AnnotationMetadata getAnnotationMetadata(SourceUnit sourceUnit, CompilationUnit compilationUnit, List<AnnotatedNode> parents, AnnotatedNode annotatedNode) {
+        new GroovyAnnotationMetadataBuilder(sourceUnit, compilationUnit).buildForParents(parents, annotatedNode)
     }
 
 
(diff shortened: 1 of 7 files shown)


yawkat:
Test still fails because it's groovy...

yawkat:
Okay, this is the best approach that I could come up with. I'm not a huge fan, but I don't want to / can't do a refactor of the annotation handling code.

graemerocher:
We probably need a test to verify the behaviour that this change introduces which is that annotations on the getter take precedence over annotations on a setter. Essentially instead of:

* Getter -> overrides -> field
 
It is now

* Getter -> overrides -> Setter -> overrides -> field

yawkat:
Apparently it's actually getter -> field -> setter... Though I guess that's the better behavior for compatibility, anyway :)

graemerocher:
yes that is better for compatibility for sure

jameskleeh (approved):

jameskleeh (approved):

graemerocher (approved):

PR #6494, opened 2021-11-09
Move to URI for HTTP client API to support ws/wss schemes

### LINKED ISSUE #5711: using RxWebSocketClient with wss protocol
It is not possible to create an RxWebSocketClient with wss protocol.
It fails in
`io.micronaut.http.client.DefaultLoadBalancerResolver.resolve(String...)`
on the constructor-call to new URL(reference)

```
            try {
                URL url = new URL(reference);
                return Optional.of(LoadBalancer.fixed(url));
            } catch (MalformedURLException e) {
                return Optional.empty();
            }
```

`java.net.MalformedURLException: unknown protocol: wss`

Or is it me doing something fundamentally wrong?

### Steps to Reproduce

1. create an RxWebSocketClient using an url with wss-protocol
i tested with following url which works fine with in a browser-test (js websocket)
> wss://stream.binance.com:9443/stream?streams=btcusdt@kline_5m/adausdt@kline_5m

### Expected Behaviour
Websocket should provide data

### Actual Behaviour
```
Caused by: io.micronaut.http.client.exceptions.HttpClientException: Invalid service reference [wss://stream.binance.com:9443] specified to @Client
	at io.micronaut.http.client.netty.RxNettyHttpClientRegistry.lambda$null$0(RxNettyHttpClientRegistry.java:235)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at io.micronaut.http.client.netty.RxNettyHttpClientRegistry.lambda$getClient$1(RxNettyHttpClientRegistry.java:234)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
	at io.micronaut.http.client.netty.RxNettyHttpClientRegistry.getClient(RxNettyHttpClientRegistry.java:197)
	at io.micronaut.http.client.netty.RxNettyHttpClientRegistry.getClient(RxNettyHttpClientRegistry.java:148)
	at io.micronaut.http.client.netty.RxNettyHttpClientRegistry.resolveClient(RxNettyHttpClientRegistry.java:392)
	at io.micronaut.http.client.netty.RxNettyHttpClientRegistry.httpClient(RxNettyHttpClientRegistry.java:193)
	at io.micronaut.http.client.netty.$RxNettyHttpClientRegistry$HttpClient0Definition.doBuild(Unknown Source)
	at io.micronaut.context.AbstractParametrizedBeanDefinition.build(AbstractParametrizedBeanDefinition.java:118)
	at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:1930)
	... 27 common frames omitted
```

### Environment Information

- **Operating System**: win10
- **Micronaut Version:** 2.41
- **JDK Version:** 11

yawkat:
- `java.net.URL` does not support the ws/wss schemes out-of-the-box. We could implement a mock `URLStreamHandler` for them, but this seems hacky, since we never actually use the `URLConnection` support `URL` provides. So instead, this patch migrates all client APIs to `URI`s, deprecating the old `URL`-based methods.
- Some checks in `DefaultHttpClient` already supported ws/wss, but others did not. To unify this, this patch extracts the relevant check to an `isSecureScheme` method.
- Otherwise, ws/wss are completely identical to http/https (https://fetch.spec.whatwg.org/#websocket-opening-handshake : "There is no real reason for WebSocket to have distinct schemes, it’s a legacy artefact."), so no further modification of the client should be necessary.

Resolves #5711

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
@@ -147,6 +147,7 @@ public interface StreamingHttpClient extends HttpClient {
      * @deprecated Use {@link #create(URI)} instead
      */
     @Deprecated
+    @NonNull
     static StreamingHttpClient create(@Nullable URL url) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(url);
     }
@@ -162,6 +163,7 @@ public interface StreamingHttpClient extends HttpClient {
      * @deprecated Use {@link #create(URI, HttpClientConfiguration)} instead
      */
     @Deprecated
+    @NonNull
     static StreamingHttpClient create(@Nullable URL url, @NonNull HttpClientConfiguration configuration) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(url, configuration);
     }
@@ -175,7 +177,9 @@ public interface StreamingHttpClient extends HttpClient {
      *
      * @param uri The base URI
      * @return The client
+     * @since 3.2.0
      */
+    @NonNull
     static StreamingHttpClient create(@Nullable URI uri) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(uri);
     }
@@ -187,8 +191,9 @@ public interface StreamingHttpClient extends HttpClient {
      * @param uri The base URI
      * @param configuration the client configuration
      * @return The client
-     * @since 2.2.0
+     * @since 3.2.0
      */
+    @NonNull
     static StreamingHttpClient create(@Nullable URI uri, @NonNull HttpClientConfiguration configuration) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(uri, configuration);
     }
--- a/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
@@ -23,7 +23,6 @@ import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpResponse;
 import org.reactivestreams.Publisher;
 
-import java.net.URI;
 import java.net.URL;
 import java.util.Map;
 
@@ -144,10 +143,7 @@ public interface StreamingHttpClient extends HttpClient {
      *
      * @param url The base URL
      * @return The client
-     * @deprecated Use {@link #create(URI)} instead
      */
-    @Deprecated
-    @NonNull
     static StreamingHttpClient create(@Nullable URL url) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(url);
     }
@@ -160,41 +156,8 @@ public interface StreamingHttpClient extends HttpClient {
      * @param configuration the client configuration
      * @return The client
      * @since 2.2.0
-     * @deprecated Use {@link #create(URI, HttpClientConfiguration)} instead
      */
-    @Deprecated
-    @NonNull
     static StreamingHttpClient create(@Nullable URL url, @NonNull HttpClientConfiguration configuration) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(url, configuration);
-    }
-
-    /**
-     * Create a new {@link StreamingHttpClient}.
-     * Note that this method should only be used outside of the context of a Micronaut application.
-     * The returned {@link StreamingHttpClient} is not subject to dependency injection.
-     * The creator is responsible for closing the client to avoid leaking connections.
-     * Within a Micronaut application use {@link jakarta.inject.Inject} to inject a client instead.
-     *
-     * @param uri The base URI
-     * @return The client
-     * @since 3.2.0
-     */
-    @NonNull
-    static StreamingHttpClient create(@Nullable URI uri) {
-        return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(uri);
-    }
-
-    /**
-     * Create a new {@link StreamingHttpClient} with the specified configuration. Note that this method should only be used
-     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
-     *
-     * @param uri The base URI
-     * @param configuration the client configuration
-     * @return The client
-     * @since 3.2.0
-     */
-    @NonNull
-    static StreamingHttpClient create(@Nullable URI uri, @NonNull HttpClientConfiguration configuration) {
-        return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(uri, configuration);
     }
 }
--- a/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
@@ -23,6 +23,7 @@ import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpResponse;
 import org.reactivestreams.Publisher;
 
+import java.net.URI;
 import java.net.URL;
 import java.util.Map;
 
@@ -143,7 +144,9 @@ public interface StreamingHttpClient extends HttpClient {
      *
      * @param url The base URL
      * @return The client
+     * @deprecated Use {@link #create(URI)} instead
      */
+    @Deprecated
     static StreamingHttpClient create(@Nullable URL url) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(url);
     }
@@ -156,8 +159,37 @@ public interface StreamingHttpClient extends HttpClient {
      * @param configuration the client configuration
      * @return The client
      * @since 2.2.0
+     * @deprecated Use {@link #create(URI, HttpClientConfiguration)} instead
      */
+    @Deprecated
     static StreamingHttpClient create(@Nullable URL url, @NonNull HttpClientConfiguration configuration) {
         return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(url, configuration);
+    }
+
+    /**
+     * Create a new {@link StreamingHttpClient}.
+     * Note that this method should only be used outside of the context of a Micronaut application.
+     * The returned {@link StreamingHttpClient} is not subject to dependency injection.
+     * The creator is responsible for closing the client to avoid leaking connections.
+     * Within a Micronaut application use {@link jakarta.inject.Inject} to inject a client instead.
+     *
+     * @param uri The base URI
+     * @return The client
+     */
+    static StreamingHttpClient create(@Nullable URI uri) {
+        return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(uri);
+    }
+
+    /**
+     * Create a new {@link StreamingHttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration the client configuration
+     * @return The client
+     * @since 2.2.0
+     */
+    static StreamingHttpClient create(@Nullable URI uri, @NonNull HttpClientConfiguration configuration) {
+        return StreamingHttpClientFactoryResolver.getFactory().createStreamingClient(uri, configuration);
     }
 }
--- a/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
@@ -18,9 +18,6 @@ package io.micronaut.http.client;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 
-import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
 
 /**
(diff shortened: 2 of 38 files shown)


yawkat:
Not sure what to do about the `create(null)` case. The test failures come from that, because it can't decide between the URI and the URL factory methods. This is only a source compatibility issue, not a binary compatibility one, and it's easy to fix in client code. Does it still mean we have to fix this? One alternative would be giving the new methods a different name.

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
```diff
@@ -156,8 +159,37 @@ static StreamingHttpClient create(@Nullable URL url) {
...
+
+    /**
+     * Create a new {@link StreamingHttpClient}.
+     * Note that this method should only be used outside of the context of a Micronaut application.
+     * The returned {@link StreamingHttpClient} is not subject to dependency injection.
+     * The creator is responsible for closing the client to avoid leaking connections.
+     * Within a Micronaut application use {@link jakarta.inject.Inject} to inject a client instead.
+     *
+     * @param uri The base URI
+     * @return The client
```
graemerocher: add `@since` and `@NonNull` to return type

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
```diff
@@ -156,8 +159,37 @@ static StreamingHttpClient create(@Nullable URL url) {
...
+    }
+
+    /**
+     * Create a new {@link StreamingHttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration the client configuration
+     * @return The client
+     * @since 2.2.0
```
graemerocher: `@since` seems incorrect here

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClient.java
```diff
@@ -156,8 +159,37 @@ static StreamingHttpClient create(@Nullable URL url) {
...
+     */
+    static StreamingHttpClient create(@Nullable URI uri, @NonNull HttpClientConfiguration configuration) {
```
graemerocher: Add `@NonNull` to return type

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
```diff
@@ -47,8 +58,50 @@
...
+    }
+
+    /**
+     * Create a new {@link StreamingHttpClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration The client configuration
+     * @return The client
+     * @since 2.2.0
```
graemerocher: incorrect `@since`

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
```diff
@@ -47,8 +58,50 @@
...
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    /**
+     * Create a new {@link StreamingHttpClient}. Note that this method should only be used outside of the context of an application. Within Micronaut use
+     * {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @return The client
```
graemerocher: add `@since`

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
```diff
@@ -47,8 +58,50 @@
...
+     */
+    @NonNull
+    default StreamingHttpClient createStreamingClient(@Nullable URI uri) {
+        try {
+            return createStreamingClient(uri != null ? uri.toURL() : null);
+        } catch (MalformedURLException e) {
+            throw new UnsupportedOperationException(e);
```
graemerocher: better to throw `HttpClientException` or `IllegalArgumentException`

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
```diff
@@ -47,8 +58,50 @@
...
+     * @param configuration The client configuration
+     * @return The client
+     * @since 2.2.0
+     */
+    @NonNull
+    default StreamingHttpClient createStreamingClient(@Nullable URI uri, @NonNull HttpClientConfiguration configuration) {
+        try {
+            return createStreamingClient(uri != null ? uri.toURL() : null, configuration);
+        } catch (MalformedURLException e) {
+            throw new UnsupportedOperationException(e);
```
graemerocher: better to throw `HttpClientException` or `IllegalArgumentException`

http-client-core/src/main/java/io/micronaut/http/client/sse/SseClient.java
```diff
@@ -138,8 +141,37 @@ static SseClient create(@Nullable URL url) {
...
+
+    /**
+     * Create a new {@link SseClient}.
+     * Note that this method should only be used outside of the context of a Micronaut application.
+     * The returned {@link SseClient} is not subject to dependency injection.
+     * The creator is responsible for closing the client to avoid leaking connections.
+     * Within a Micronaut application use {@link jakarta.inject.Inject} to inject a client instead.
+     *
+     * @param uri The base URI
+     * @return The client
```
graemerocher: Add `@since` and `@NonNull` to return type

http-client-core/src/main/java/io/micronaut/http/client/sse/SseClient.java
```diff
@@ -138,8 +141,37 @@ static SseClient create(@Nullable URL url) {
...
+    }
+
+    /**
+     * Create a new {@link SseClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration the client configuration
+     * @return The client
+     * @since 2.2.0
```
graemerocher: incorrect `@since`

http-client-core/src/main/java/io/micronaut/http/client/sse/SseClientFactory.java
```diff
@@ -48,8 +59,50 @@
...
+     *
+     * @param uri The base URI
+     * @return The client
+     */
+    @NonNull
+    default SseClient createSseClient(@Nullable URI uri) {
+        try {
+            return createSseClient(uri != null ? uri.toURL() : null);
+        } catch (MalformedURLException e) {
+            throw new UnsupportedOperationException(e);
```
graemerocher: better to throw `HttpClientException` or `IllegalArgumentException`

http-client-core/src/main/java/io/micronaut/http/client/sse/SseClientFactory.java
```diff
@@ -48,8 +59,50 @@
...
+     * @param configuration the client configuration
+     * @return The client
+     * @since 2.2.0
+     */
+    @NonNull
+    default SseClient createSseClient(@Nullable URI uri, @NonNull HttpClientConfiguration configuration) {
+        try {
+            return createSseClient(uri != null ? uri.toURL() : null, configuration);
+        } catch (MalformedURLException e) {
+            throw new UnsupportedOperationException(e);
```
graemerocher: better to throw `HttpClientException` or `IllegalArgumentException`

http-client-core/src/main/java/io/micronaut/http/client/sse/SseClientFactory.java
```diff
@@ -48,8 +59,50 @@
...
+    }
+
+    /**
+     * Create a new {@link SseClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration the client configuration
+     * @return The client
+     * @since 2.2.0
```
graemerocher: incorrect `@since`

http-client-core/src/main/java/io/micronaut/http/client/sse/SseClientFactory.java
```diff
@@ -48,8 +59,50 @@
...
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    /**
+     * Create a new {@link SseClient}. Note that this method should only be used outside of the context of an application. Within Micronaut use
+     * {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @return The client
```
graemerocher: add `@since`

websocket/src/main/java/io/micronaut/websocket/WebSocketClient.java
```diff
@@ -125,9 +127,38 @@ static WebSocketClient create(@Nullable URL url) {
...
+    }
+
+    /**
+     * Create a new {@link WebSocketClient} with the specified configuration. Note that this method should only be used
+     * outside of the context of an application. Within Micronaut use {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration the client configuration
+     * @return The client
+     * @since 2.2.0
```
graemerocher: Incorrect `@since` and add `@NonNull` to return type

websocket/src/main/java/io/micronaut/websocket/WebSocketClient.java
```diff
@@ -125,9 +127,38 @@ static WebSocketClient create(@Nullable URL url) {
...
 
+    /**
+     * Create a new {@link WebSocketClient}.
+     * Note that this method should only be used outside of the context of a Micronaut application.
+     * The returned {@link WebSocketClient} is not subject to dependency injection.
+     * The creator is responsible for closing the client to avoid leaking connections.
+     * Within a Micronaut application use {@link jakarta.inject.Inject} to inject a client instead.
+     *
+     * @param uri The base URI
+     * @return The client
```
graemerocher, line 148: add `@since`

websocket/src/main/java/io/micronaut/websocket/WebSocketClientFactory.java
```diff
@@ -47,8 +58,49 @@
...
+     *
+     * @param uri The base URI
+     * @return The client
+     */
+    @NonNull
+    default WebSocketClient createWebSocketClient(@Nullable URI uri) {
+        try {
+            return createWebSocketClient(uri != null ? uri.toURL() : null);
+        } catch (MalformedURLException e) {
+            throw new UnsupportedOperationException(e);
```
graemerocher, line 86: better to throw `HttpClientException` or `IllegalArgumentException`

websocket/src/main/java/io/micronaut/websocket/WebSocketClientFactory.java
```diff
@@ -47,8 +58,49 @@
...
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    /**
+     * Create a new {@link WebSocketClient}. Note that this method should only be used outside of the context of an application. Within Micronaut use
+     * {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @return The client
```
graemerocher, line 78: add `@since`

websocket/src/main/java/io/micronaut/websocket/WebSocketClientFactory.java
```diff
@@ -47,8 +58,49 @@
...
+        }
+    }
+
+    /**
+     * Create a new {@link WebSocketClient} with the specified configuration. Note that this method should only be used outside of the context of an application. Within Micronaut use
+     * {@link jakarta.inject.Inject} to inject a client instead
+     *
+     * @param uri The base URI
+     * @param configuration The client configuration
+     * @return The client
```
graemerocher, line 96: add `@since`

websocket/src/main/java/io/micronaut/websocket/WebSocketClientFactory.java
```diff
@@ -36,9 +39,17 @@
      *
      * @param url The base URL
      * @return The client
+     * @deprecated Use {@link #createWebSocketClient(URI)} instead
```
graemerocher, line 42: add `@since`

graemerocher (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/StreamingHttpClientFactory.java
yawkat: I think `UnsupportedOperationException` is better here because this will only happen if `createStreamingClient(URI)` hasn't been implemented yet, and the scheme is ws/wss

jameskleeh:
Why do we need the `create(@Nullable URI uri) {` methods in anything other than the WebSocketClient?

yawkat:
We don't strictly need them, but it seems sensible to move the API to URI eventually, for consistency.

sdelamo:
We have `create(URL)` and `create(URL, HttpClientConfiguration)` static methods in [rxjava2](https://github.com/micronaut-projects/micronaut-rxjava2), [rxjava3](https://github.com/micronaut-projects/micronaut-rxjava3) and [reactor](https://github.com/micronaut-projects/micronaut-reactor) modules. We should do this change in the reactive libraries as well.

jameskleeh (changes requested):
Given the possibility for breakage with passing a null I think we should remove the URI conversion for the static methods not in WebsocketClient

yawkat:
Is the breakage acceptable for `WebSocketClient`?

jameskleeh:
I think the cost/benefit works out in my opinion for `WebSocketClient`

graemerocher (approved):

jameskleeh (changes requested):
I would like to see some tests using ws:// and wss://. Also please add to breaks.adoc the source compatibility with `create(null)`

graemerocher:
There are already tests:

https://github.com/micronaut-projects/micronaut-core/pull/6494/files#diff-50c9ef7968abe9afc19ca01d98e5041ec63797f25042f6540bbe6ae8e73e1f9dR109

and 

https://github.com/micronaut-projects/micronaut-core/pull/6494/files#diff-50c9ef7968abe9afc19ca01d98e5041ec63797f25042f6540bbe6ae8e73e1f9dR194

Docs make sense though

jameskleeh:
Ah OK missed that

PR #6507, opened 2021-11-11
Fix AbstractTypeElementSpec on Java 17

graemerocher:
AbstractTypeElementSpec is broken on Java 17, this fixes it. Note, I had to move some code to Java due to seeming bugs in Groovy.

Also closes  #6325


graemerocher:
I am going to merge this since @jameskleeh is ill, and this is holding me up. I don't believe there is anything controversial here

PR #6510, opened 2021-11-12
Add support for GenericPlaceholder at runtime

graemerocher:
Currently it is not possible at runtime to know what a variable name is for a generic type use. A new `GenericPlaceholder` interface is added that exposes a `getVariableName()` method that retains a reference to the variable name used. This is different to the actual type variable name and different to the field / parameter name for a generic type use.

Note that I have omitted adding support for multiple upper bounds since Micronaut really just supports one upper bound currently. A future revision of this new interface could in theory add support for multiple upper bounds.


graemerocher:
This is needed by the serialization project.

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -381,12 +437,14 @@ default boolean isAssignableFrom(@NonNull Argument<?> candidateArgument) {
...
      * Allows coercing a {@link Type} instance to an {@link Argument}.
+     *
      * @param type The type
      * @return The argument
      * @throws IllegalArgumentException If the type cannot be coerced
      * @since 3.0.0
      */
-    static @NonNull Argument<?> of(@NonNull Type type) {
+    static @NonNull
+    Argument<?> of(@NonNull Type type) {
```
yawkat, line 447: should this have an implementation for TypeVariable, now?

yawkat (approved):

core/src/main/java/io/micronaut/core/type/Argument.java
graemerocher: Perhaps, but our runtime reconstruction from reflection data is largely missing and not sure we need it or want it

yawkat (approved):

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -286,15 +293,45 @@ default boolean isAssignableFrom(@NonNull Argument<?> candidateArgument) {
...
+     * @param typeParameters     the type parameters
+     * @param <T>                The generic type
+     * @return The argument instance
+     * @since 3.2.0
+     */
+    @UsedByGeneratedCode
+    @NonNull
+    static <T> Argument<T> ofTypeVariable(
+            @NonNull Class<T> type,
+            @Nullable String name,
```
jameskleeh: Perhaps change this to be argumentName

jameskleeh: There is one other example below as well

Issue #6508, opened 2021-11-11
Repeatable annotations broken in type arguments of generics


graemerocher:
### Expected Behavior

This test is failing when running on JDK 17.

https://github.com/micronaut-projects/micronaut-core/blob/35af8faed6cb00ce5d0d5cd994e128b21e3b4c45/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy#L483-L506

I don't understand why it is not failing in the JDK 15 we have running it seems that the test is ignored.

and it should pass

### Actual Behaviour

Test fails because `@Min` is not regarded as a repeated annotation. This appears to be because the new precompiled logic for repeated annotations is not factoring in annotations on generics and annotations on anything nested (fields, methods) within an introspection because the defaults are not being tracked across all the arguments like we do for default annotation values.

### Steps To Reproduce

Run the mentioned test on JDK 17

### Version

3.1.4

graemerocher:
@dstepanov could you take a look at this, it is related to the changes to repeated annotation handling. They don't appear to be computed correctly.

PR #6514, opened 2021-11-15
Clear caches from previous compilation round to avoid hiding test failures

graemerocher:
This PR should be part of the resolution for #6508 since what is happening is tests are passing incorrectly because previous compilation phases are populating static state that is then present in later tests around known default annotations and repeated annotations. This is ugly but I don't know of a better solution given what we have today.


graemerocher:
I'm not sure if this will work actually, but yeah the problem remains that it is easy to create tests you think pass

PR #6517, opened 2021-11-15
Introduce handshakeTimeout option to SslConfiguration

### LINKED ISSUE #5501: Provide a way to set SSL handshake timeout value
### Task List

### Steps to Reproduce

N/A

### Expected Behaviour

Micronaut HTTP client configuration should provide a way to specify SSL handshake timeout.

### Actual Behaviour

See `io.netty.handler.ssl.SslHandler#handshakeTimeoutMillis`: default value of 10_000ms is always used. If an HTTPS connection handshake takes longer than that, the exception is thrown:
```
io.netty.handler.ssl.SslHandshakeTimeoutException: handshake timed out after 10000ms
    at io.netty.handler.ssl.SslHandler$7.run(SslHandler.java:2029)
    at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
    at io.netty.util.concurrent.ScheduledFutureTask.run(ScheduledFutureTask.java:170)
    at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
    at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:472)
    at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:500)
    at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
    at java.base/java.lang.Thread.run(Thread.java:829)
```

### Environment Information

- **Micronaut Version:** 2.5.x

### Example Application

N/A

yawkat:
Add an option for configuring the SSL handshake timeout for both the HTTP client and server.
Resolves #5501

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/SslSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/SslSpec.groovy
@@ -16,12 +16,30 @@
 package io.micronaut.http.client
 
 import io.micronaut.context.ApplicationContext
+import io.micronaut.runtime.server.EmbeddedServer
+import io.netty.bootstrap.Bootstrap
+import io.netty.bootstrap.ServerBootstrap
+import io.netty.channel.Channel
+import io.netty.channel.ChannelHandlerContext
+import io.netty.channel.ChannelInboundHandlerAdapter
+import io.netty.channel.ChannelInitializer
+import io.netty.channel.ServerChannel
+import io.netty.channel.nio.NioEventLoopGroup
+import io.netty.channel.socket.nio.NioServerSocketChannel
+import io.netty.channel.socket.nio.NioSocketChannel
+import io.netty.handler.ssl.SslHandshakeTimeoutException
+import org.jetbrains.annotations.NotNull
+import reactor.core.publisher.Flux
 import spock.lang.Ignore
 import spock.lang.Specification
 
+import java.time.Duration
+import java.util.concurrent.TimeUnit
+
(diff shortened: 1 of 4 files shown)


http-client/src/test/groovy/io/micronaut/http/client/SslSpec.groovy
```diff
@@ -34,4 +52,59 @@ class SslSpec extends Specification {
...
+        def embeddedServer = context.getBean(EmbeddedServer)
+        embeddedServer.start()
+        def channel = new Bootstrap()
+                .channel(NioSocketChannel)
+                .group(group)
+                .handler(new IgnoringChannelInitializer())
+                .connect(embeddedServer.host, embeddedServer.port).sync().channel()
+        expect:
+        // returns false on timeout
+        channel.closeFuture().await(5, TimeUnit.SECONDS)
```
jameskleeh, line 94: This is testing the timeout failure right? Wouldn't this test fail if it returned false?

yawkat: The channel should be closed after one second by the server because of the server timeout, and then the await on the client shouldn't timeout

jameskleeh: I see, thanks

jameskleeh (approved):

PR #6529, opened 2021-11-17
Make JsonNode converters more robust for different target types

### LINKED ISSUE #6526: Crash + Leak of implementation details : "class java.util.ArrayList cannot be cast to class io.micronaut.core.convert.value.ConvertibleValues" on invalid request body
### Expected Behavior

HTTP 400 response indicating that required body was not submitted. 

### Issues
1. Code crash with 500 http response
2. Error message reveals implementation details:
```
class java.util.ArrayList cannot be cast to class io.micronaut.core.convert.value.ConvertibleValues  .....
```
(full stack trace is presented at the bottom) 

### Investigation 
After short debugging session the following explanation for the crash was found:
1. Arguments are attempted to be bound using `RequestArgumentSatisfier.getValueForArgument` method
2. Inside this method private `RequestBinderRegistry` bean is queried using `findBinder` method for appropriate instance of `ArgumentBinder` for each parameter 
3. Then, for annotated parameters `ParameterAnnotationBinder` is returned and its `bind` method is called
4. The binder converts the submitted request body to an instance of `Optional<ConvertableValues>` landing at `DefaultConversionService` code, where appropriate converter is picked.  For our example case, the `JsonArray` input type is mapped to `JsonConverterRegistar` converter, because both `JsonArray` and `ConvertableValues` both have `Iterable` in their class hierarchy (see `DefaultConversionService.findTypeConverter` method)
7. `JsonConverterRegistar` returns `Optional<List>`. For our example case, the list ends up empty as submitted data cannot be parsed to `Flock` model.
8. Eventually, `ParameterAnnotationBinder` crashes on `Optional.get` with `ClassCastException` as in step 4 it was expecting instance of optional `ConvertibleValues` and received optional carrying list.

### Stacktrace
```
java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class io.micronaut.core.convert.value.ConvertibleValues (java.util.ArrayList is in module java.base of loader 'bootstrap'; io.micronaut.core.convert.value.ConvertibleValues is in unnamed module of loader 'app')
	at io.micronaut.http.bind.binders.ParameterAnnotationBinder.bind(ParameterAnnotationBinder.java:87)
	at io.micronaut.http.bind.binders.ParameterAnnotationBinder.bind(ParameterAnnotationBinder.java:39)
	at io.micronaut.http.server.binding.RequestArgumentSatisfier.lambda$getValueForArgument$3(RequestArgumentSatisfier.java:147)
	at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:265)
	at io.micronaut.web.router.RouteMatch.execute(RouteMatch.java:111)
	at io.micronaut.http.context.ServerRequestContext.with(ServerRequestContext.java:103)
	at io.micronaut.http.server.RouteExecutor.lambda$executeRoute$14(RouteExecutor.java:636)
	at reactor.core.publisher.FluxDeferContextual.subscribe(FluxDeferContextual.java:49)
	at reactor.core.publisher.Flux.subscribe(Flux.java:8402)
/* REST OMITTED */
```

### Steps To Reproduce

1. **Have a project for a service powered by micronaut 3.1.3**
2. **With request handler like:**
```java
@Post(value = "/feedBirds", processes = MediaType.APPLICATION_JSON)
public CompletableFuture<Boolean> feedBirds(
      @RequestBody(content = @Content(schema = @Schema(implementation = Flock.class)), required = true) 
      @Valid
      final Flock flock
) {
    return feedBirds(flock.getBirds());
}
```
**And toy models like:**
```java
 public final class Flock {
        private Bird[] birds;
        public Bird[] getBirds() { return birds; }
        public void setBirds(Bird[] birds) { this.birds = birds; }
 }
 public final class Bird {
        private String name;
        public String getName() { return name; }
        public void setName(String name) { this.name = name; }
  }
```
2. **Submit the following request**
```bash
curl -X POST 'http://localhost:8080/feedBirds' \
-H 'accept: application/json' -H 'Content-Type: application/json' \
-d '[ { "name": "eagle" }, { "name": "hen" } ]'
'
```

3. Note that here we **intentionally** submit wrong model:  `json array`  of birds (invalid) instead of `{ "birds": /*..same array..*/ }` (valid). 

### Reproduction steps generic formula (probably?)
Submit json array in the request body where bound request body argument is expected to be something not related to Iterable

### Workaround
Expected behavior could be achieved if `RequestBinderRegistry` bean is substituted with custom `DefaultRequestBinderRegistry`-delegating implementation introducing `catch-all` logic in `findArgumentBinder` method.

### Environment Information

Operating System: Ubuntu 20.04
Micronaut Version: v3.1.3
JDK Version: OpenJDK Runtime Environment (build 17.0.1+12-39)

### Example Application

_No response_

### Version

v3.1.3

yawkat:
This patch changes the JsonNode converters so that:
- ArrayNode->Iterable conversion picks a collection type that is compatible with the requested `targetType`, and fails if the `targetType` isn't workable
- JsonNode->ConvertibleValues fails gracefully when attempted on a non-object JsonNode
Fixes #6526

### DIFF
--- a/json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
+++ b/json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
@@ -122,7 +122,6 @@ public final class JsonConverterRegistrar implements TypeConverterRegistrar {
      */
     public TypeConverter<JsonArray, Iterable> arrayNodeToIterableConverter() {
         return (node, targetType, context) -> {
-            Map<String, Argument<?>> typeVariables = context.getTypeVariables();
             Collection<Object> results;
             if (targetType.isAssignableFrom(ArrayList.class)) {
                 results = new ArrayList<>();
@@ -132,6 +131,7 @@ public final class JsonConverterRegistrar implements TypeConverterRegistrar {
                 // don't know how to convert to that collection type
                 return Optional.empty();
             }
+            Map<String, Argument<?>> typeVariables = context.getTypeVariables();
             Class elementType = typeVariables.isEmpty() ? Map.class : typeVariables.values().iterator().next().getType();
             for (int i = 0; i < node.size(); i++) {
                 Optional<?> converted = conversionService.convert(node.get(i), elementType, context);
--- a/json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
+++ b/json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
@@ -37,7 +37,9 @@ import jakarta.inject.Singleton;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -74,14 +76,14 @@ public final class JsonConverterRegistrar implements TypeConverterRegistrar {
                 arrayNodeToObjectConverter()
         );
         conversionService.addConverter(
-                JsonArray.class,
-                Iterable.class,
-                arrayNodeToIterableConverter()
-        );
-        conversionService.addConverter(
                 JsonNode.class,
                 ConvertibleValues.class,
                 objectNodeToConvertibleValuesConverter()
+        );
+        conversionService.addConverter(
+                JsonArray.class,
+                Iterable.class,
+                arrayNodeToIterableConverter()
         );
         conversionService.addConverter(
                 JsonNode.class,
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

json-core/src/main/java/io/micronaut/json/convert/JsonConverterRegistrar.java
```diff
@@ -114,13 +123,19 @@ public void register(ConversionService<?> conversionService) {
     public TypeConverter<JsonArray, Iterable> arrayNodeToIterableConverter() {
         return (node, targetType, context) -> {
             Map<String, Argument<?>> typeVariables = context.getTypeVariables();
```
rdesgroppes: minor: `typeVariables` may be moved down because it's not needed in the 3rd/default case.

jameskleeh (changes requested):
This needs a test for `Set`

PR #6531, opened 2021-11-18
Support NaN and Infinity for JsonNodeTreeCodec

yawkat:
Add support for NaN and Infinity even when `useBigDecimalForFloats` is enabled (`BigDecimal` does not support these numbers). The input `JsonParser` still has to have support enabled (`JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS`).

### DIFF
--- a/jackson-core/src/main/java/io/micronaut/jackson/core/tree/JsonNodeTreeCodec.java
+++ b/jackson-core/src/main/java/io/micronaut/jackson/core/tree/JsonNodeTreeCodec.java
@@ -94,7 +94,12 @@ public final class JsonNodeTreeCodec {
                 }
             case VALUE_NUMBER_FLOAT:
                 if (config.useBigDecimalForFloats()) {
-                    return JsonNode.createNumberNode(p.getDecimalValue());
+                    // NaN and Inf can't be represented by BigDecimal. Note: isNaN returns true for Inf, too.
+                    if (p.isNaN()) {
+                        return JsonNode.createNumberNode(p.getFloatValue());
+                    } else {
+                        return JsonNode.createNumberNode(p.getDecimalValue());
+                    }
                 } else {
                     // technically, we could get an unsupported number value here.
                     return JsonNode.createNumberNodeImpl(p.getNumberValue());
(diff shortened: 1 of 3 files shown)


jackson-core/src/main/java/io/micronaut/jackson/core/tree/JsonNodeTreeCodec.java
```diff
@@ -94,7 +94,12 @@ public JsonNode readTree(JsonParser p) throws IOException {
                 }
             case VALUE_NUMBER_FLOAT:
                 if (config.useBigDecimalForFloats()) {
-                    return JsonNode.createNumberNode(p.getDecimalValue());
+                    // NaN and Inf can't be represented by BigDecimal. Note: isNaN returns true for Inf, too.
+                    if (p.isNaN()) {
```
ilopmar, line 98: `NaN`... are we javascript developers now? :rofl:

yawkat: NaNNaNNaNNaN... Batman!

ilopmar: LOL

jackson-core/src/test/groovy/io/micronaut/jackson/core/tree/JsonNodeTreeCodecSpec.groovy
```diff
@@ -42,7 +44,7 @@ class JsonNodeTreeCodecSpec extends Specification {
         def treeCodec = JsonNodeTreeCodec.getInstance().withConfig(JsonStreamConfig.DEFAULT
                 .withUseBigIntegerForInts(true)
                 .withUseBigDecimalForFloats(true))
-        def factory = new JsonFactory()
+        def factory = JsonFactory.builder().enable(JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS).disable(JsonWriteFeature.WRITE_NAN_AS_STRINGS).build()
```
jameskleeh: Create a new test for the NaN instead of modifying the existing one

jameskleeh (changes requested):

PR #6539, opened 2021-11-19
Complete all downstream subscribers in RoutingInBoundHandler

### LINKED ISSUE #6532: StreamingFileUpload subscriber "onComplete" not invoked sometimes
### Expected Behavior

- uploading file thru Controller
- subscribe to StreamingFileUpload
- when Publisher is done, onComplete is invoked on subscriber

### Actual Behaviour

- uploading file thru Controller
- subscribe to StreamingFileUpload
- onComplete() is never invoked

### Steps To Reproduce

Issue is happening randomly once in a while (e.g. every 20th REST call) and only! with file size of approx 32KB when first `onNext(pd: PartData) { ... }` receives already all data sent by client.

Here is a very simplified code in Kotlin I have:

Controller can accept 1 or more files (issues happening when 1 file is uploaded):
```
@Post("/upload")
@Consumes(MediaType.MULTIPART_FORM_DATA)
open fun upload(files: Publisher<StreamingFileUpload>): Mono<MutableHttpResponse<String>> {
  return Flux.from(files)
    .subscribeOn(Schedulers.boundedElastic())
    .flatMap { file ->
      Mono.create<String> { sink ->
        file.subscribe(MySubscriber(sink))
      }.map { "Uploaded FooBar" }
    }
    .reduce { _, _ -> "Upload summary FooBar }
    .map { HttpResponse.ok(it) }
}
```
Subscriber implementation:
```
class MySubscriber(private val sink: MonoSink<String>) : Subscriber<PartData> {
  private lateinit var subscription: Subscription
  private val stream: OutputStream = ByteArrayOutputStream()  // In reality I have Alex Mojaki's MultiPartOutputStream here, this is just for simplification

  override fun onSubscribe(s: Subscription) {
        subscription = s
        subscription.request(1)
  }

  override fun onNext(pd: PartData) {
    val bytes = pd.bytes
    stream.write(bytes)
    subscription.request(1)
  }

  override fun onError(t: Throwable) {
    sink.error(t)
  }

  override fun onComplete() {
    stream.close()
    sink.success("Finished")
  }
}
```

When added logging into `onNext` method on `PardData.bytes` size, when PartData size equals to whole upload file size, onComplete is not called (whole file content is sent within first `onNext`), HTTP client waits till timeout for any HTTP response, which is not arriving, as the Flow is not finishing.

Basically the code is the same since Micronaut 1.x.x, when migrated to Micronaut 3.0.x, later 3.1.x, this issue appeared.

Also I am not able to reproduce this behaviour with different file sizes (MB, GB ...), with smaller files the frequency of the issue is way less, like once in few hours, when doing serial continuous invocation of the API.

### Environment Information

Kotlin 1.5
JVM 11

### Example Application

_No response_

### Version

3.1.4

yawkat:
Before this patch, the `UnicastProcessor childSubject` would never have its `onComplete` method called, because it is not part of the `subjects` map. This patch adds a separate collection for forwarding `onComplete` and `onError` events so that they also reach nested subscribers.

Fixes #6532

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
@@ -102,6 +102,7 @@ import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxProcessor;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.MonoSink;
 import reactor.core.publisher.UnicastProcessor;
@@ -113,6 +114,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.channels.ClosedChannelException;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


yawkat:
Also there is a race condition that makes this work sometimes: `HttpPostMultipartRequestDecoder` can update a `HttpData` returned by our `next()` call, even though it is still in flight in our reactive processing. I think `RoutingInBoundHandler` will need to be rewritten to some extent to move the `HttpData` processing into the netty event loop, at some point

d35h:
Hello @yawkat, is there a chance that it will be released soon? We're blocked by this issue :(

yawkat:
Hi @d35h , it should already be part of the 3.2.0 release.

d35h:
@yawkat thanks for the reply; however, we use 3.2.0 and the issue is still there, it can be seen from the release notes that #6532 is not there: https://github.com/micronaut-projects/micronaut-core/releases/tag/v3.2.0. Any way how to accelerate it? :)

yawkat:
You're right, looks like 3.1.x wasn't merged into 3.2.x in time. You'll have to wait for 3.2.1 then, but that should come fairly soon.

PR #6548, opened 2021-11-22
Add a way to access Kotlin coroutine context from an interceptor

dstepanov:
I would like to introduce the transaction context propagation in Micronaut Data and need a way to access the coroutine's context.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
+++ b/aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
@@ -164,9 +164,9 @@ public interface InterceptedMethod {
     }
 
     /**
-     * Get native method context.
+     * Get native context from the suspension point.
      *
-     * @return The native context for the suspension point or null
+     * @return The native context from the suspension point or null
      * @since 3.2
      */
     @Nullable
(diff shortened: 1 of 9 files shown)


aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
```diff
@@ -163,6 +163,27 @@ default Object unsupported() {
...
+
+    /**
+     * Update native context for the suspension point.
+     *
+     * @param context The native context
+     * @since 3.2
+     */
+    default void updateNativeContext(@Nullable Object context) {
+        throw new IllegalStateException("Updating native context not supported for interceptor: " + this);
+    }
```
graemerocher: I think these should be placed behind an interface that you use an `instanceof` check on on the interceptor. Perhaps called it `SuspendedMethod`

graemerocher (changes requested):

aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
dstepanov: Added a new interface just for Kotlin methods

graemerocher (approved):

PR #6558, opened 2021-11-23
Fix input file normalization for functional tests

melix:
See the first commit message for details. This is **not** a blocker for the 3.2 release, it's a build optimization, so this can be merged after the release.

@wolfs kindly asking you as a reviewer as suggested by Etienne during the GE trial.

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.functional-test.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.functional-test.gradle
@@ -1,0 +1,44 @@
+import groovy.transform.CompileStatic
+
+plugins {
+    id 'java-library'
+}
+
+sourceSets {
+    functionalTest
+}
+
+configurations {
+    functionalTestImplementation {
+        extendsFrom(testImplementation)
+    }
+}
+
+def functionalTest = tasks.register("functionalTest", FunctionalTest) {
+    appUnderTestCompileClasspath.from(configurations.testCompileClasspath)
+    appUnderTestAnnotationProcessor.from(configurations.testAnnotationProcessor)
+    testClassesDirs = sourceSets.functionalTest.output.classesDirs
+    classpath = sourceSets.functionalTest.runtimeClasspath
+}
+
+tasks.named("check") {
+    dependsOn functionalTest
+}
+
+@CacheableTask
+@CompileStatic
+abstract class FunctionalTest extends Test {
+    @CompileClasspath
+    abstract ConfigurableFileCollection getAppUnderTestCompileClasspath()
+
+    @CompileClasspath
+    abstract ConfigurableFileCollection getAppUnderTestAnnotationProcessor()
+
+    @Override
+    @TaskAction
+    void executeTests() {
+        systemProperty("testapp.compile.classpath", appUnderTestCompileClasspath.asPath)
+        systemProperty("testapp.annotationprocessor.path", appUnderTestAnnotationProcessor.asPath)
+        super.executeTests()
+    }
+}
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/groovy/io.micronaut.build.internal.functional-test.gradle
```diff
@@ -0,0 +1,44 @@
...
+    @CompileClasspath
+    abstract ConfigurableFileCollection getAppUnderTestCompileClasspath()
+
+    @CompileClasspath
+    abstract ConfigurableFileCollection getAppUnderTestAnnotationProcessor()
+
+    @Override
+    @TaskAction
+    void executeTests() {
+        systemProperty("testapp.compile.classpath", appUnderTestCompileClasspath.asPath)
```
wolfs, line 40: I think it would be better to use a `CommandLineArgumentProvider`, which declares the two file collections as compile classpath inputs and then adds the system properties. This way you don't need to modify the inputs of the task during task execution. Moreover, that removes the need for a subclass of `Test`.

melix: I was actually saying I'm not sure if it's better. If I use a command line argument provider, then I need to build `-Dfoo=bar` manually, which is the role of the system properties DSL. It also potentially mixes system properties defined using the `systemProperties` DSL from others defined via the CLI argument provider. Also I'm replacing one custom type (the `FunctionalTest` type) with another (the `CommandLineArgumentProvider`).

Ideally, I'd prefer something like this:

```
tasks.register("functionalTest", Test) {
     systemProperty("testapp.compile.classpath", configurations.testCompileClasspath.map { it.asPath })
}
```

but then you're asking "what about the normalization". Well there are 2 options: either declare it explicitly, via `normalize(CompileClasspath, configurations.testCompileClasspath)` or infer it from the configuration attributes (because it represents a compile classpath).

wolfs: For me the advantage of using a `CommandLineArgumentProvider` is that is was built for extension and the `Test` task was not built for extending it. So using `CommandLineArgumentProvider` is less likely to break. I also really dislike the modification of the inputs in the task action. That is something which is most likely to break when we make those things immutable during task execution.

From the effect of your changes, I don't think there is a difference between your solution and the solution to use `CommandLineArgumentProvider`.

melix: My problem is less about extension than it is about semantics. A command-line arg provider is agnostic, while I'm "strongly typed" with `systemProperty`. The `systemProperty` will take care of escaping values, which I will have to care about (think spaces in paths). Also I don't know on top of my head how the CL arg provider will _combine_ with existing `systemProperty` calls (or regular args).

wolfs: Hmm..., it doesn't seem like `systemProperty` takes care of escaping values, maybe I am missing something: https://github.com/gradle/gradle/blob/38207db9fb8c1892d97cfa6aed785e75d38ed45a/subprojects/core/src/main/java/org/gradle/process/internal/JvmOptions.java#L104-L112

I agree that it should be easier to add system properties with normalization to a `Test` task.

graemerocher (approved):

buildSrc/src/main/groovy/io.micronaut.build.internal.functional-test.gradle
melix: I'm going to leave the code as is for now since none of the solution really pleases me and I feel less "dirty" with the way it is now.

PR #6562, opened 2021-11-23
Enable client SSL certificate checking by default

### LINKED ISSUE #4116: HTTP client does not reject untrusted certificates unless trust store is configured
### Steps to Reproduce

1. Create a Micronaut application with feature http-client
2. Create a controller
3. Configure the server to enable ssl and use a self signed certificate
4. Create a test that calls the controller with an http client that points to the controller's https endpoint

### Expected Behaviour

An SSLException is thrown due to an unknown certificate

### Actual Behaviour

The client accepts the certificate and gets a valid response

### Environment Information

- **Operating System**: macOS
- **Micronaut Version:** tested with 1.3.5 and 2.0.2
- **JDK Version:** 8

### Example Application

- https://github.com/myles-anderson-ox/micronaut-http-ssl-bug
Run `gradle test` and see 1 out of 3 tests fail. This may require an update to the trust store configuration in `HttpSslBugTest.java` if your jdk default cacerts has been modified.

yawkat:
This patch enables SSL cert checking by default. The old behavior without checking can be enabled using the `insecureTrustAllCertificates` option.
Fixes #4116

### DIFF
--- a/http/src/main/java/io/micronaut/http/ssl/ClientSslConfiguration.java
+++ b/http/src/main/java/io/micronaut/http/ssl/ClientSslConfiguration.java
@@ -103,9 +103,9 @@ public class ClientSslConfiguration extends SslConfiguration {
     }
 
     /**
-     * @return Whether the client should disable checking of the remote SSL certificate. <b>Note: This makes the SSL
-     * connection insecure, and should only be used for testing. If you are using a self-signed certificate, set up a
-     * trust store instead.</b>
+     * @return Whether the client should disable checking of the remote SSL certificate. Only applies if no trust store
+     * is configured. <b>Note: This makes the SSL connection insecure, and should only be used for testing. If you are
+     * using a self-signed certificate, set up a trust store instead.</b>
      */
     public boolean isInsecureTrustAllCertificates() {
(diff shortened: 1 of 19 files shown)


yawkat:
This is the core change required. I will look at the necessary test changes now.

http/src/main/java/io/micronaut/http/ssl/ClientSslConfiguration.java
```diff
@@ -100,6 +102,25 @@ void setTrustStore(@Nullable DefaultTrustStoreConfiguration trustStore) {
...
+     * @return Whether the client should disable checking of the remote SSL certificate. <b>Note: This makes the SSL
+     * connection insecure, and should only be used for testing. If you are using a self-signed certificate, set up a
+     * trust store instead.</b>
+     */
+    public boolean isInsecureTrustAllCertificates() {
+        return insecureTrustAllCertificates;
+    }
+
+    /**
+     * @param insecureTrustAllCertificates Whether the client should disable checking of the remote SSL certificate.
```
jameskleeh: Add some documentation that says this only applies if no trust store is configured

PR #6566, opened 2021-11-24
Don't initialize SLF4J in the compiler

graemerocher:
During compilation `Failed to load class "org.slf4j.impl.StaticLoggerBinder".` is printed because we are using logging in classes that are used within the compiler and shouldn't be. For these classes another logging strategy will need to be put in place.


melix (commented):
mmm, this is annoying as it's super handy to get those log messages when developing AOT, to make sure that they are actually loaded. Should we use JUL instead?

graemerocher:
An alternative is to add `slf4j-simple` as a dependency for the compiler modules

melix:
> An alternative is to add slf4j-simple as a dependency for the compiler modules

Any drawback in doing this?

graemerocher:
Well generally you are not supposed to print to standard out from annotation processors, that is what the `Messager` API is for and with this there may still be standard output appearing, but I guess if it is debug level and off by default it is probably ok

yawkat:
When there is a solution for this we should also use it for handling the SoftServiceLoader errors that can appear when developing micronaut-core.

PR #6572, opened 2021-11-25
Add slf4j-simple as a dependency to inject-java and inject-groovy

### LINKED ISSUE #6570: SLF4J binding warnings raised during compilation
### Expected Behavior

Importing [SLF4J](http://www.slf4j.org/) should raise no warnings at compile time. Seen from the Micronaut Launch project using Micronaut 3.1.4:

```
gw clean build

> Task :compileTestJava
Note: Creating bean classes for 1 type elements

BUILD SUCCESSFUL in 14s
15 actionable tasks: 15 executed
```

### Actual Behaviour

Importing SLF4J raises warnings during `:compileJava` and `:compileTestJava` regarding a failed class load. Seen from the Micronaut Launch project using Micronaut 3.2.0:

```
gw clean build

> Task :compileJava
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

> Task :compileTestJava
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Note: Creating bean classes for 1 type elements

BUILD SUCCESSFUL in 15s
15 actionable tasks: 15 executed
```

This warning occurs when [no SLF4J binding can be found on the classpath](http://www.slf4j.org/codes.html#StaticLoggerBinder). However, the Launch project ships `runtimeOnly("ch.qos.logback:logback-classic")`, which [should be sufficient as an SLF4J backend](http://www.slf4j.org/manual.html#swapping).

It also isn't clear why a runtime dependency would be initialized during compilation.

I saw an older [outstanding bug with the Groovy compiler](https://github.com/micronaut-projects/micronaut-core/issues/676) that showed a similar behavior. I also saw[ a bug around Graal native images ](https://github.com/micronaut-projects/micronaut-core/issues/3683#issuecomment-881713816)that may be related. However, I have also seen this warning message behavior in a private project that uses neither Groovy nor Graal.

### Steps To Reproduce

- Download demonstration project from [Micronaut Launch ](https://micronaut.io/launch/)with default settings: Micronaut 3.2.0, Java 11, Gradle build tool, etc.
- Add `implementation("org.slf4j:slf4j-api:1.7.32")` to dependencies in build.gradle
- `./gradlew clean build`

### Environment Information

- Java 11
- Gradle 7.2
- Micronaut 3.2.0

### Example Application

https://micronaut.io/launch/

### Version

3.2.0

graemerocher:
This is an alternative fix that adds slf4j-simple to the compiler modules. Note that this has one issue in that if either `inject-java` or `inject-groovy` appear on the runtime classpath (which they shouldn't) then you can get multiple bindings errors from SLF4J

Fixes #6570

### DIFF
--- a/inject-groovy/build.gradle
+++ b/inject-groovy/build.gradle
@@ -27,6 +27,7 @@ dependencies {
     api project(":inject")
     api project(':aop')
     api libs.managed.groovy
+    implementation(libs.slf4j.simple)
 
 //    testImplementation 'javax.validation:validation-api:1.1.0.Final'
     testImplementation project(":context")
--- a/inject-groovy/build.gradle
+++ b/inject-groovy/build.gradle
@@ -27,7 +27,7 @@ dependencies {
     api project(":inject")
     api project(':aop')
     api libs.managed.groovy
-    implementation(libs.slf4j.simple)
+    implementation(libs.managed.slf4j.simple)
 
 //    testImplementation 'javax.validation:validation-api:1.1.0.Final'
     testImplementation project(":context")
--- a/inject-java/build.gradle
+++ b/inject-java/build.gradle
@@ -2,7 +2,7 @@ dependencies {
     api project(":inject")
     api project(':aop')
 
-    implementation(libs.slf4j.simple
+    implementation(libs.managed.slf4j.simple)
     if (!JavaVersion.current().isJava9Compatible()) {
         compileOnly files(org.gradle.internal.jvm.Jvm.current().toolsJar)
     }
--- a/inject-java/build.gradle
+++ b/inject-java/build.gradle
@@ -2,6 +2,7 @@ dependencies {
     api project(":inject")
     api project(':aop')
 
+    implementation(libs.slf4j.simple
     if (!JavaVersion.current().isJava9Compatible()) {
         compileOnly files(org.gradle.internal.jvm.Jvm.current().toolsJar)
     }


melix:
Are `inject-java` and `inject-groovy` compiler only? They shouldn't be found on the application classpath, right?

graemerocher:
`inject-groovy` should always be `compileOnly` and `inject-java` should exist only on the annotation processor classpath. However some of our own modules that testImplementation dependencies on `inject-java-test` for testing will need to exclude slf4j-simple most likely

melix (approved):

devmage:
> `inject-groovy` should always be `compileOnly` and `inject-java` should exist only on the annotation processor classpath.

`inject-java` is needed at `compileOnly` for Android projects, per current docs: https://docs.micronaut.io/latest/guide/#android

There are also notes on including `inject-java` in `kapt` scope for Kotlin projects: https://docs.micronaut.io/latest/guide/#introspection and https://docs.micronaut.io/latest/guide/#kotlin

Finally there is a reference to using `inject-java` as a generic AOP processor. Dependency scopes are left to the implementer: https://docs.micronaut.io/latest/guide/#faq

I don't believe these uses would meaningfully change the impact of including `slf4j-simple` to `inject-java` though.

devmage:
Also worth remarking, `slf4j-simple` has exactly `slf4j-api` as a dependency, which is nice from the perspective of limiting transitive dependencies: https://repo1.maven.org/maven2/org/slf4j/slf4j-simple/1.7.32/slf4j-simple-1.7.32.pom

PR #6576, opened 2021-11-26
Java time default formatters

ckosmowski:
The Java Time classes have default formatters built in which use the ISO formats. So if in micronaut somewhere a string is to be converted to one of those classes we expect the default format to be the default format of the classes.

Up to now micronaut used RFC_1123_DATE_TIME as the default format for converting but this is a rather human readable format and does not use the defaults that are already there.

Another advantage of the built in defaults is, that they the classes do use the same format in .toString(). So i.e. OffsetDateTime.toString() is compatible with OffsetDateTime.parse() without thinking about the format.

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -145,7 +145,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 try {
                     return Optional.of(resolveFormatter(context)
                             .map(formatter -> LocalDateTime.parse(object, formatter))
-                            .orElse(LocalDateTime.parse(object)));
+                            .orElseGet(() -> LocalDateTime.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
@@ -158,7 +158,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
             try {
                 return Optional.of(resolveFormatter(context)
                         .map(formatter -> formatter.format(object))
-                        .orElse(object.toString()));
+                        .orElseGet(() -> object.toString()));
             } catch (DateTimeParseException e) {
                 context.reject(object, e);
                 return Optional.empty();
@@ -178,7 +178,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 try {
                     return Optional.of(resolveFormatter(context)
                             .map(formatter -> LocalDate.parse(object, formatter))
-                            .orElse(LocalDate.parse(object)));
+                            .orElseGet(() -> LocalDate.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
@@ -195,7 +195,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 try {
                     return Optional.of(resolveFormatter(context)
                             .map(formatter -> ZonedDateTime.parse(object, formatter))
-                            .orElse(ZonedDateTime.parse(object)));
+                            .orElseGet(() -> ZonedDateTime.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
@@ -211,7 +211,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                     try {
                         return Optional.of(resolveFormatter(context)
                                 .map(formatter -> OffsetDateTime.parse(object, formatter))
-                                .orElse(OffsetDateTime.parse(object)));
+                                .orElseGet(() -> OffsetDateTime.parse(object)));
                     } catch (DateTimeParseException e) {
                         context.reject(object, e);
                         return Optional.empty();
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -143,9 +143,9 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
             LocalDateTime.class,
             (object, targetType, context) -> {
                 try {
-                    DateTimeFormatter formatter = resolveFormatter(context);
-                    LocalDateTime result = LocalDateTime.parse(object, formatter);
-                    return Optional.of(result);
+                    return Optional.of(resolveFormatter(context)
+                            .map(formatter -> LocalDateTime.parse(object, formatter))
+                            .orElse(LocalDateTime.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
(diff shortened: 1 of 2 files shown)


jameskleeh:
I believe this is a breaking change. I think I agree with the change but it must wait until Micronaut 4

ckosmowski:
Yes I agree. Is there anything that I have to do, to make the change appear in micronaut 4 or can it just stay as it is?

jameskleeh:
You can just leave this PR as is and we will review/merge when the time comes

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -143,9 +143,9 @@ public void register(ConversionService<?> conversionService) {
             LocalDateTime.class,
             (object, targetType, context) -> {
                 try {
-                    DateTimeFormatter formatter = resolveFormatter(context);
-                    LocalDateTime result = LocalDateTime.parse(object, formatter);
-                    return Optional.of(result);
+                    return Optional.of(resolveFormatter(context)
+                            .map(formatter -> LocalDateTime.parse(object, formatter))
+                            .orElse(LocalDateTime.parse(object)));
```
aleksandy: I believe `.orElseGet(() -> LocalDateTime.parse(object))` would be more appropriate.

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -156,8 +156,9 @@ public void register(ConversionService<?> conversionService) {
         // TemporalAccessor - CharSequence
         final TypeConverter<TemporalAccessor, CharSequence> temporalConverter = (object, targetType, context) -> {
             try {
-                DateTimeFormatter formatter = resolveFormatter(context);
-                return Optional.of(formatter.format(object));
+                return Optional.of(resolveFormatter(context)
+                        .map(formatter -> formatter.format(object))
+                        .orElse(object.toString()));
```
aleksandy: And here.

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -175,9 +176,9 @@ public void register(ConversionService<?> conversionService) {
             LocalDate.class,
             (object, targetType, context) -> {
                 try {
-                    DateTimeFormatter formatter = resolveFormatter(context);
-                    LocalDate result = LocalDate.parse(object, formatter);
-                    return Optional.of(result);
+                    return Optional.of(resolveFormatter(context)
+                            .map(formatter -> LocalDate.parse(object, formatter))
+                            .orElse(LocalDate.parse(object)));
```
aleksandy: And here

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -192,9 +193,9 @@ public void register(ConversionService<?> conversionService) {
             ZonedDateTime.class,
             (object, targetType, context) -> {
                 try {
-                    DateTimeFormatter formatter = resolveFormatter(context);
-                    ZonedDateTime result = ZonedDateTime.parse(object, formatter);
-                    return Optional.of(result);
+                    return Optional.of(resolveFormatter(context)
+                            .map(formatter -> ZonedDateTime.parse(object, formatter))
+                            .orElse(ZonedDateTime.parse(object)));
```
aleksandy: Here too

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -208,9 +209,9 @@ public void register(ConversionService<?> conversionService) {
                 OffsetDateTime.class,
                 (object, targetType, context) -> {
                     try {
-                        DateTimeFormatter formatter = resolveFormatter(context);
-                        OffsetDateTime result = OffsetDateTime.parse(object, formatter);
-                        return Optional.of(result);
+                        return Optional.of(resolveFormatter(context)
+                                .map(formatter -> OffsetDateTime.parse(object, formatter))
+                                .orElse(OffsetDateTime.parse(object)));
                     } catch (DateTimeParseException e) {
```
aleksandy, line 215: And here

jameskleeh:
@ckosmowski Can you sign the CLA with other account "Christian Kosmowski" and rebase this to the 4.0.x branch?

ckosmowski:
> @ckosmowski Can you sign the CLA with other account "Christian Kosmowski" and rebase this to the 4.0.x branch?
Done

jameskleeh:
No idea why the CI isn't running, however this causes some tests failures because date headers conform to the old default format. We need to have some way of retaining that behavior. I'm considering adding something on the conversion context to handle format more officially

ckosmowski:
Can you clarify which date headers you are referring to? Our could you give
a pointer to a failing test? The whole point was that in case of the easy
case where you just want to serde a Offsetdatetime that brings its own
default format shouldn't be that hard to configure or even to find out the
problem.

It's not easy to understand why there is this nowhere used weird  default
RFC Format. And if I can I want to avoid format information to business
model classes.

James Kleeh ***@***.***> schrieb am Di., 8. Feb. 2022, 17:10:

jameskleeh:
io.micronaut.docs.server.binding.BindingControllerTest#testHeaderDateBinding is failing. Having an argument to a controller like `@Header OffsetDateTime date` should not require the user to specify `@Format` because date headers follow a standard format (the old default). The binder that handles headers has no way of applying that format by default because there is no API in the conversion context to do so. That is what I think will need to be added to change this behavior

ckosmowski:
I don't agree with that point of view. For me OffsetDatetime brings its own String representation internally (The ISO format it uses). That's what you put in into OffsetDatime.parse() good feature of the "new" Java DateTime API Imho.

The test you are referring to is testing the things that are returned from the HeaderBinding spec. This is also the place where the format "RFC_1123_DATE_TIME" is specified. 

I think the use case of using OffsetDateTime inside of the data exchanged by rest APIs would be very more common than reading the HTTP headers from the requests. The above format only applies for this special case (Only the standard HTTP headers). Even if someone specifies custom headers containing dates it would be very unlikely that this format is used. It would instead be likely that if you expect an OffsetDateTime you would use the OffsetDateTime internal format.

So what about considering the Standard Http-Header case the special case here and either let the user specify the format (which i would find okay) or use the format for the standard http headers implicitly.  Because this format actually only applies to standard Http headers and to nothing else i have seen anywhere, so making this the default seems wrong to me.

On top of that in case of adding a format to Headers i am adding this format to a micronaut controller class which is also better imho than adding a format specification inside a business model object that i try to deserialize from an API call Response.

jameskleeh:
I think you are misunderstanding me. I agree the default format should be changed in accordance with this PR 

> So what about considering the Standard Http-Header case the special case here and either let the user specify the format (which i would find okay) or use the format for the standard http headers implicitly. Because this format actually only applies to standard Http headers and to nothing else i have seen anywhere

That is exactly what my position is, that it should be implicit for `@Header`. The user should not have to specify format in that case because the format is already known

jameskleeh:
Closing in favor of https://github.com/micronaut-projects/micronaut-core/pull/6916

PR #6578, opened 2021-11-28
Dependency upgrades

micronaut-build:
Upgrades dependencies to their latest versions


sdelamo:
I don't think this is valid anymore

PR #6581, opened 2021-11-29
Check for static file resolution before returning route not found

### LINKED ISSUE #6580: Method not allowed is returned for static resources if there is another endpoint with different http method
### Expected Behavior

Get requests to static resources should not be returning method not allowed if there is another http method that maps to request. Router should check if there are static resources before returning error response.

### Actual Behaviour

A method not allowed response with available methods is returned instead of the static resource.

### Steps To Reproduce

1. Using micronaut 3.1.3
2. Serve some static resources in /test path with following config in application.yml
```
micronaut:
  router:
    static-resources:
      my-static-resource:
        paths: <path_to_static_resource>
        mapping: /my-static-resource/**
```
3.Create a controller with a different http method than GET.
```
@Controller("/")
public class OptionsController {

  @Options("/my-static-resource")
  void handleOptions() {
    
  }
}
```

With the above configuration, requests to /my-static-resource return Method Not Allowed with allow: Options header instead of the actual resource.
### Environment Information

- OS: MacOs 11.6
- JDK: 17

### Example Application

_No response_

### Version

3.1.3

segocago:


segocago:
To fix this issue: https://github.com/micronaut-projects/micronaut-core/issues/6580

yawkat:
Please add a test

yawkat:
The build is also failing because of checkstyle. You can run checkstyle (and the whole test suite) locally to find out why: `./gradlew check`

segocago:
Are there any tests for the routing that i can use as an example? I saw some tests with StaticResourceResolver but could not find any for the RoutingInBoundHandler that could trigger this case?

yawkat:
You could add a test case to the existing `StaticResourceResolutionSpec`. The tests are basically integration tests, they set up a full client + server, so it shouldn't be too difficult to get all the necessary components together. 

There should also be a test that the "proper" controller can still be called when the method for that controller is used (i.e. it's not overridden by the resource handler).

If you don't know how or aren't comfortable with spock, I can write a test for this PR. Just tell me, so we don't do duplicate work :)

segocago:
I'm not familiar with spock testing, could give it a try over the weekend but i would really appreciate if you can provide the tests for it.

yawkat (approved):

PR #6587, opened 2021-11-30
Add JsonMapper-based readMetadataUrl method to ComputeInstanceMetadataResolverUtils

yawkat:
This allows loading metadata with only the json-core module. The other methods require jackson-core or jackson-databind.

Required for micronaut-aws migration.


runtime/src/main/java/io/micronaut/discovery/cloud/ComputeInstanceMetadataResolverUtils.java
```diff
@@ -78,6 +80,24 @@ public static JsonNode readMetadataUrl(URL url, int connectionTimeoutMs, int rea
...
+    /**
+     * Reads the result of a URL and parses it using the given {@link JsonMapper}.
+     *
+     * @param url                 the URL to read
+     * @param connectionTimeoutMs connection timeout, in milliseconds
+     * @param readTimeoutMs       read timeout, in milliseconds
+     * @param jsonMapper          JSON mapper to use for parsing
+     * @param requestProperties   any request properties to pass
+     * @return a {@link JsonNode} instance
+     * @throws IOException if any I/O error occurs
```
graemerocher, line 92: Add `@since`

graemerocher (changes requested):

graemerocher (approved):

jameskleeh:
This should go into 3.3

graemerocher:
@jameskleeh normally I would agree but it would be nice if the all the modules were compatible with serialization for the release of Micronaut 3.3.

Without this method then AWS module will not work with Micronaut Serialization and is coupled to Jackson until 3.3 is out and we will have to do a patch release of AWS afterwards

If you think that is better then we do that, but seems a pretty minor change to wait all the way until 3.3.1

Issue #6519, opened 2021-11-16
Setting custom message via "messageTemplate" in Micronaut Validator overrides all constraint violation messages of fields


nrsina:
### Expected Behavior

Assume we have a DTO like below:
```
@Introspected
@TestValidation
public class TestRequest {
    @Min(1)
    private long personNo;

    @Size(max = 20)
    private String personalId;

    private int someField;

    @Size(max = 2)
    private String code;
    ...
}
```
And I want to set custom message inside TestValidation's validator:

```
@Override
public boolean isValid(TestRequest value,
                       AnnotationValue<TestValidation> annotationMetadata,
                       ConstraintValidatorContext context) {
    if (value.getPersonalId().equals("1234") && value.getSomeField() == 22) {
        context.messageTemplate("personalId '1234' and someField '22' is wrong!");
        return false;
    }
    return true;
}
```
The expected behavior when throwing and handling **ConstraintViolationException** is that it includes both messages for custom messages inside TestValidation's validator and other messages for field validations. for example:
```
private void printViolations(ConstraintViolationException exception) {
    exception.getConstraintViolations()
            .forEach(constraintViolation -> 
                    System.out.println(constraintViolation.getPropertyPath().toString() + " : " + constraintViolation.getMessage()));
}
```
Now, if we validate this DTO with below values:
```
personNo = 0
personalId = "1234"
someField = "22"
code = "333"
```
we will have three constraint violations (one for DTO and two for fields), the result of the above method should be:
```
test.testRequest : personalId '1234' and someField '22' is wrong!
test.testRequest.personNo : must be greater than or equal to 1
test.testRequest.code : size must be between 0 and 2
```

### Actual Behaviour

But the actual result is:
```
test.testRequest : personalId '1234' and someField '22' is wrong!
test.testRequest.personNo : personalId '1234' and someField '22' is wrong!
test.testRequest.code : personalId '1234' and someField '22' is wrong!
```

### Environment Information

Tested with both JDK 11 and 17

### Version

3.1.4

agugan:
@graemerocher Can I pick this up?

graemerocher:
@agugan contributions welcome!

nrsina:
Thanks for working on this issue @agugan, but I still have some issues with Micronaut v3.2.7

I have two DTOs like below:
```
@Introspected
@TestValidation
public class TestRequest {
    @Min(1)
    private long personNo;

    @Size(max = 20)
    private String personalId;

[...]

PR #6607, opened 2021-12-06
Reset previous validated messages in default validation context to null

agugan:
Fixes (#6519)


validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
```diff
@@ -1052,6 +1052,7 @@ public void keyedValue(String nodeName, Object key, Object keyedValue) {
             }
         } finally {
             context.removeLast();
+            context.messageTemplate(null);
```
jameskleeh: Instead of setting to null shouldn't this reset it to what it was previously?

jameskleeh (changes requested):

validation/src/main/java/io/micronaut/validation/validator/DefaultValidator.java
agugan: Sure, We can do that. Thanks. I thought that since we are checking for the context's message template, if null found then we move on to annotation's default messages., I thought it would make sense to set the message template to null. Can you help me understand more on the scenarios, where resetting it to previous message wold help us?

jameskleeh: @agugan Truthfully I'm not sure, but setting it to what it was prior reduces the chances of breaking some existing use case

agugan: @jameskleeh Sorry I had to take some timeoff for personal reasons. I have updated as per your suggestions. Please do let me know if any changes required.

graemerocher:
Thanks for the contribution!

PR #6613, opened 2021-12-07
Update testcontainers to 1.16.2 to fix race condition bug

danoliver1:
There is a fairly major bug in the testcontainers 1.16.1 version used by the latest version of Micronaut, so this PR updates to the patched version.

`1.16.2 fixes a race condition that was inadvertently added in 1.16.1. This bug can potentially cause unstable builds in some environments, manifesting as port wait timeouts at container startup.`

https://github.com/testcontainers/testcontainers-java/releases/tag/1.16.2


ilopmar (approved):

ilopmar:
Thanks!

PR #6616, opened 2021-12-08
Initialize netty classes at runtime

### LINKED ISSUE #6619: Netty fails with OutOfMemoryError when allocating direct memory on native-image
### Issue description

In 3.2, a basic netty application will fail with an OutOfMemoryError when trying to allocate direct memory, when running on native-image.

Example project: https://github.com/krisfoster/mn-stress

Stack trace:

```
java.lang.OutOfMemoryError: Cannot reserve 16777216 bytes of direct buffer memory (allocated: 285216775, limit: 299892736)
	at java.nio.Bits.reserveMemory(Bits.java:178)
	at java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:121)
	at java.nio.ByteBuffer.allocateDirect(ByteBuffer.java:332)
	at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:648)
	at io.netty.buffer.PoolArena$DirectArena.newChunk(PoolArena.java:623)
	at io.netty.buffer.PoolArena.allocateNormal(PoolArena.java:202)
	at io.netty.buffer.PoolArena.tcacheAllocateSmall(PoolArena.java:172)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:134)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:126)
	at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:395)
	at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
	at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:179)
	at io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:140)
	at io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator$MaxMessageHandle.allocate(DefaultMaxMessagesRecvByteBufAllocator.java:120)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:150)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:719)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:655)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:581)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:493)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986)
	at java.lang.Thread.run(Thread.java:833)
	at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:596)
	at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)
```

yawkat:
In 3.2, there is a bug where netty will claim more direct memory than is available in native-image. This is because the netty class `PlatformDependent` is initialized at runtime and thus sees the memory limits of the native-image build environment, not those of the runtime.

The recommendation of the native-image team is to move the entire `io.netty` package to runtime initialization. This patch implements that recommendation.


melix (approved):
LGTM. Maybe worth checking with @gradinac what overlaps with the new configuration format.

graemerocher:
@yawkat could you create an associated issue report as well that this PR resolves just for tracking purposes

gradinac:
Regarding the configuration changes, the PR seems on spot - the conditional configuration format currently only applies to `.json` config files like `reflection-config.json` etc.

One note - I'm not sure if it's possible to override all netty initialization options if the netty jars with those options end up on the classpath as well (if they have a `native-image.properties` file that forces a class to be build time initialized and we tell native-image to initialize it at runtime later, this may lead to errors).

There was an effort to runtime initialize all netty classes IIRC (something I'll also be looking into in the future) - however, this impacted the size of the resulting native image - it might be worth checking the image size of a Micronaut app that uses netty and see if it changes with this PR

yawkat:
@gradinac If class lists maintained by the netty project become available, it will be better to use those anyway. I hope we can remove this special handling then. Is such official support planned in the (near) future?

I've taken a look at the size difference, and for the test project in the issue, it looks like the images are roughly the same size (64M w/o vs 63M with this patch).

graemerocher (approved):

ilopmar (approved):
I've locally tested this patch with our GraalvM basic-app that uses Netty and our HTTP Client and everything works. Regarding the image size, it's the same than before applying the patch.
I've also run the whole GraalVM suite against a snapshot published from this branch and again, everything worked.

PR #6634, opened 2021-12-13
Downgrade Netty to 4.1.70

ilopmar:
See https://github.com/netty/netty/issues/11908


ilopmar:
Netty 4.1.72 is out: https://netty.io/news/2021/12/13/4-1-72-Final.html. I'll send a new PR

PR #6635, opened 2021-12-13
Introduce an internal interface for finalizing context configuration

melix:
During debugging of Micronaut AOT, we discovered that some bean requirements
weren't met, because bean configurations were not read. This happened because
those beans are only read when the context is actually started. In AOT, we
don't want to start the context, only introspect it, so we missed an operation
to perform the finalization of configuration, while keeping the context in
an unstarted state.

For this purpose, we introduce a new internal interface, which AOT will be
able to check for, and call this method instead of using reflection to
call the `readAllXXX` methods.

This shouldn't be a breaking change so I'm tentatively setting 3.2.2 as the milestone version.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/IntrospectableBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/IntrospectableBeanContext.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017-2021 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


inject/src/main/java/io/micronaut/context/IntrospectableBeanContext.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A marker interface for {@link BeanContext} implementations that can be introspected,
+ * that is to say for context which can be created and need to be fully configured,
+ * but not necessarily started yet.
+ *
+ * @since 3.2.2
+ */
+@Internal
+public interface IntrospectableBeanContext extends BeanContext {
```
graemerocher: Perhaps this name could be confused with bean introspections. Maybe `ConfigurableBeanContext` is more appropriate? Or `InitializableBeanContext`?

melix: Makes sense, let me change the name.

PR #6638, opened 2021-12-13
Implement content negotiation for Accept header

### LINKED ISSUE #6637: Improve content negotiation support for `Accept`
### Feature description

Right now, `Accept` header matching can only handle some very basic matching rules (i.e. `*/*`, but not `text/*`). We should support more of the spec.

yawkat:
This patch adds support for proper content negotiation using the `Accept` header. This includes new support for:
- "Partial" matches, like text/*
- Quality parameters
- Correct matching by precedence, e.g. matching text/plain over \*/*.

Compatibility considerations:
- MediaType now considers parameters in `matches`, if specified
- Overloaded routes that relied on the old, incorrect matching behavior will now match differently.

Resolves #6637

### DIFF
--- a/http/src/main/java/io/micronaut/http/MediaType.java
+++ b/http/src/main/java/io/micronaut/http/MediaType.java
@@ -370,6 +370,13 @@ public class MediaType implements CharSequence {
      */
     public static final String V_PARAMETER = "v";
 
+    /**
+     * Comparator that sorts media types by their match precedence, highest precedence first. For example,
+     * {@code text/*} has higher precedence that {@code *}{@code /*}, but {@code text/plain} has higher precedence than
+     * either.
+     */
+    public static final Comparator<MediaType> PRECEDENCE_COMPARATOR = Comparator.comparing((MediaType mt) -> -mt.matchWeight);
+
     @Internal
     static final Argument<MediaType> ARGUMENT = Argument.of(MediaType.class);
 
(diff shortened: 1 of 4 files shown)


yawkat:
The `ContentNegotiationSpec` still fails because of a test case that has `Accept: application/json, application/xml`, and expects that to only match the controller for `application/json`. So basically it relies on the order of `Accept` headers.

I don't think we should support this. [The spec](https://httpwg.org/specs/rfc7231.html#header.accept) makes no mention of order being relevant. Browsers [seem to](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values) use quality values properly and not rely on order either. Edge is the exception there, I will look around if I can find more details on Edge, but even that browser will work with this PR (I try to match the "most specific" controller).

@graemerocher thoughts?

yawkat:
Apparently others have noticed the Edge thing too, and they come to the same conclusion: https://github.com/willdurand/Negotiation/issues/98#issuecomment-390625746

They also propose the same approach of prioritizing non-wildcard accept types that I use in this patch. That fixes edge, but it does not fix the test case that is failing for this PR.

graemerocher:
@yawkat FWIW many frameworks implement it this way (based on order), both Rails and Grails use the order to prioritize. Also because of how inconsistently this is implemented across browsers and there being no real standard this was in fact disabled all together by default in Grails (see https://docs.grails.org/3.0.x/ref/Controllers/withFormat.html) with users having to enable explicitly

jameskleeh:
Without being gated behind a config option, this would have to wait for 4.x

yawkat:
yea i think it's not worth a config option. I'll let the PR sit and we can revisit it for 4.0

PR #6657, opened 2021-12-14
Create new netty-graal module for native image configuration

### LINKED ISSUE #6650: Generalize `HttpNettyFeature` for other uses of netty
### Issue description

Other micronaut projects that use netty but do not depend on micronaut-http-netty need the same graal configuration that `HttpNettyFeature` provides. The issue where this came up is https://github.com/micronaut-projects/micronaut-grpc/pull/435 . I can think of a few ways to solve this:

- Create a new micronaut-netty module that contains a new `NettyFeature` with the relevant functionality. Make downstream projects that use netty depend on this module.
- Add the `NettyFeature` to the core module, with a `compileOnly` dependency on netty.
- Add the runtime initializations to `native-image.properties` in the core module.

My favorite is number 2, because:

- Only runtime initializations aren't sufficient to get best netty performance, we also need the other code in `HttpNettyFeature`.
- A full module seems unnecessary. It also means that if we ever decide to create a general-purpose netty framework for micronaut (not http-specific), we would need to use a different name. Though using a name like `micronaut-netty-graal` would avoid this.

@graemerocher Thoughts? If it seems reasonable to you, I will implement 2 today, so it can be reviewed by james. Otherwise one of you will have to implement it, or it will have to wait over the holidays.

yawkat:
The graal configuration that netty needs was previously only present in the http-netty module (HttpNettyFeature). This patch creates a new netty-graal module for that configuration, so that other modules that use netty but not http-netty (in particular micronaut-grpc) can use the same configuration.

Resolves #6650

### DIFF
--- a/netty-graal/src/main/java/io/micronaut/netty/NettyFeature.java
+++ b/netty-graal/src/main/java/io/micronaut/netty/NettyFeature.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.netty;
+
+import com.oracle.svm.core.annotate.AutomaticFeature;
+import com.oracle.svm.core.jdk.SystemPropertiesSupport;
+import io.micronaut.core.graal.AutomaticFeatureUtils;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+import io.netty.util.internal.logging.Slf4JLoggerFactory;
+import org.graalvm.nativeimage.ImageSingletons;
+import org.graalvm.nativeimage.hosted.Feature;
+import org.graalvm.nativeimage.hosted.RuntimeClassInitialization;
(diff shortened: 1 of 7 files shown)


netty-graal/src/main/java/io/micronaut/netty/NettyFeature.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Arrays;
+
+/**
+ * A native image feature that configures the netty library.
+ *
+ * @author Jonas Konrad
+ * @since 3.3.0
+ */
+@AutomaticFeature
+final class NettyFeature implements Feature {
```
ilopmar: I'd add here `@Internal`

settings.gradle
```diff
@@ -52,6 +52,7 @@ include "jackson-databind"
 include "json-core"
 include "management"
 include "messaging"
+include 'netty-graal'
```
ilopmar: Sorry to be picky here but we are using double-quotes `"` instead of single-quote `'` :see_no_evil:

ilopmar (changes requested):
I've tried this locally with the gRPC module and everything works as expected

ilopmar:
I've created https://github.com/micronaut-projects/micronaut-grpc/pull/437 in gRPC module with the changes to use this once we release Micronaut 3.3.0

graemerocher:
I am still unconvinced a new module needed. Especially if https://github.com/micronaut-projects/micronaut-core/pull/6661 is merged in which case that module will be need to allow configurability of the Netty buffer allocator

yawkat:
Then #6661 should also be in the new module

graemerocher:
Why? it relates to configuring the buffer allocator and is in the netty-buffer module. What more appropriate module would it live in? Configuring the buffer allocator seems something that would be desirable in a gRPC app as well

yawkat:
imo the purpose of the netty-buffer module is to provide the netty implementation of `io.micronaut.core.io.buffer.ByteBuffer`, general netty configuration is unrelated

yawkat:
moved this to buffer-netty too, as per #6661

PR #6662, opened 2021-12-15
Upgrade to build plugins 5.1.0

melix:
The upgrade to workflows are intentional, because changes to the
templates are not yet merged.


melix:
Remote build cache is enabled :tada:

gradle.properties
```diff
@@ -28,12 +28,12 @@ jee7api=https://docs.oracle.com/javaee/7/api
...
-micronautribbonapi=https://micronaut-projects.github.io/micronaut-netflix/latest/api/io/micronaut/configuration/ribbon
-micronautcacheapi=https://micronaut-projects.github.io/micronaut-cache/latest/api/io/micronaut/cache
-micronautreactorapi=https://micronaut-projects.github.io/micronaut-reactor/latest/api/io/micronaut/reactor
-micronautspringapi=https://micronaut-projects.github.io/micronaut-spring/latest/api/io/micronaut/spring
+micronautdiscoveryapi=https://micronaut-projects.github.io/micronaut-discovery-client/latest/api
+micronauthystrixapi=https://micronaut-projects.github.io/micronaut-netflix/latest/api
+micronautribbonapi=https://micronaut-projects.github.io/micronaut-netflix/latest/api
+micronautcacheapi=https://micronaut-projects.github.io/micronaut-cache/latest/api
+micronautreactorapi=https://micronaut-projects.github.io/micronaut-reactor/latest/api
+micronautspringapi=https://micronaut-projects.github.io/micronaut-spring/latest/api
```
jameskleeh, line 36: This would break the existing docs, wouldn't it?

jameskleeh (changes requested):

gradle.properties
melix: That's actually the other way around: I was puzzled that I got error messages when I built locally, that the links were not valid. I then realized that they needed to end at `api` otherwise javadoc won't find the package list.

graemerocher (approved):

graemerocher:
Combined into https://github.com/micronaut-projects/micronaut-core/pull/6603

PR #6668, opened 2021-12-16
Constant property sources shouldn't all be loaded

melix:
This commit fixes a bug in how "constant" property sources are loaded.
Constant property sources are property sources which are generated by
Micronaut AOT, for example by converting YAML configuration files to
Java.

Once they are converted, they are added as property sources. However,
if more than one configuration file is converted, for example if
we convert `application.yml` and `application-test.yml`, then both
configurations would be applied at startup, even if the `test`
environment isn't set.

This commit fixes the bug by making sure that the constant property
sources are filtered by active environments.

See https://github.com/micronaut-projects/micronaut-aot/pull/37

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
+++ b/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
@@ -99,6 +99,7 @@ public class DefaultEnvironment extends PropertySourcePropertyResolver implement
     private static final String DO_SYS_VENDOR_FILE = "/sys/devices/virtual/dmi/id/sys_vendor";
     private static final Boolean DEDUCE_ENVIRONMENT_DEFAULT = true;
     private static final List<String> DEFAULT_CONFIG_LOCATIONS = Arrays.asList("classpath:/", "file:config/");
+    private static final String DEFAULT_ENVIRONMENT_NAME = "default";
     protected final ClassPathResourceLoader resourceLoader;
     protected final List<PropertySource> refreshablePropertySources = new ArrayList<>(10);
 
@@ -163,7 +164,6 @@ public class DefaultEnvironment extends PropertySourcePropertyResolver implement
         // Search config locations in reverse order
         Collections.reverse(configLocations);
         this.configLocations = configLocations;
-        CONSTANT_PROPERTY_SOURCES.forEach(p -> propertySources.put(p.getName(), p));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


graemerocher:
Is there a way to test this in core?

melix:
There is probably a way, can you point me at existing tests which verify what environment is loaded?

graemerocher:
In here makes sense https://github.com/micronaut-projects/micronaut-core/tree/3.2.x/inject/src/test/groovy/io/micronaut/context/env

graemerocher (approved):

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -99,6 +99,7 @@
     private static final String DO_SYS_VENDOR_FILE = "/sys/devices/virtual/dmi/id/sys_vendor";
     private static final Boolean DEDUCE_ENVIRONMENT_DEFAULT = true;
     private static final List<String> DEFAULT_CONFIG_LOCATIONS = Arrays.asList("classpath:/", "file:config/");
+    private static final String DEFAULT_ENVIRONMENT_NAME = "default";
```
jameskleeh: Can you explain this? Why would the constant property source be called default over application?

melix: `application` is not the name of the environment, but the name of the _resource_ (that is to say it corresponds to the base name in `application-[env].yml`). There is special casing for the "default" case where the `-[env]` part is not used when searching for resources. So the complete pattern is `[resource name]-[env name]`, except for the default environment.

jameskleeh: There is no default environment, unless the user specifies one. That is why this doesn't make sense to me

melix: The "default" application configuration is always loaded. However in the context of AOT, we need to make a difference between "the user didn't configure anything" and "the user explicitly chose to limit AOT to the default environment, ignoring `-dev`, `-aws` and all). For this we need a way to denote the "default" environment name, which is, the one with no name.

graemerocher: The default environment is always loaded though so perhaps this is not necessary and the logic should instead always load the default env?

melix: I removed the special case for `default` and rebased onto `3.3.x` instead.

PR #6670, opened 2021-12-17
Support for interceptor binding with annotation members

graemerocher:
This PR adds support for interceptor binding with annotation members. This is a requirement to support CDI lite.

Unlike with qualifiers we can't enable interceptor binding with members by default because that would be a breaking change, so this adds a new `bindMembers` option to `@InterceptorBinding` which developers can use to enable interceptor binding with members.

When enabled the `@NonBinding` annotation can be used to exclude members that shouldn't be factored in to the binding.


graemerocher:
This works for Micronaut but not for the CDI project I am working on so still needs work

dstepanov (approved):

graemerocher:
@jameskleeh any progress on this review?

PR #6689, opened 2021-12-23
Nullable body spec

### LINKED ISSUE #3064: HttpRequest<Optional<ByteBuffer<*>>> still requires a request body when it should be optional
Per a conversation in gitter, `HttpRequest<Optional<ByteBuffer<*>>>` should not require a request body in order to call the controller endpoint.

### Steps to Reproduce

Controller:
```
@Post("/test")
fun post(request: HttpRequest<Optional<ByteBuffer<*>>>) = request.body.isPresent
```
curl with empty body:
```
$ curl localhost:8080/test -XPOST
```

### Expected Behavior

```
false
```

### Actual Behavior

```
{"message":"Required argument [HttpRequest request] not specified","path":"/request","_links":{"self":{"href":"/test","templated":false}}}
```

### Environment Information

- **Operating System**: MacOS Catalina
- **Micronaut Version:** 1.3.2
- **JDK Version:** 11

mattmoss:
Fixes #3064


http/src/main/java/io/micronaut/http/bind/DefaultRequestBinderRegistry.java
```diff
@@ -88,7 +100,12 @@ public DefaultRequestBinderRegistry(ConversionService conversionService, Request
                 if (source.getBody().isPresent()) {
                     return () -> Optional.of(new FullHttpRequest(source, typeVariable.get()));
                 } else {
-                    return ArgumentBinder.BindingResult.UNSATISFIED;
+                    if (source.getAttribute(REQUEST_BINDING_UNSATISFIED).isPresent() && typeVariable.get().isNullable()) {
+                        return () -> Optional.of(source);
+                    } else {
+                        source.setAttribute(REQUEST_BINDING_UNSATISFIED, true);
+                        return ArgumentBinder.BindingResult.UNSATISFIED;
+                    }
```
graemerocher, line 108: This seems like a hack, can you explain this change?

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConverters.java
```diff
@@ -347,13 +341,6 @@ public void register(ConversionService<?> conversionService) {
         };
     }
 
-    /**
-     * @return A converter that returns bytebufs to objects
-     */
-    protected TypeConverter<ByteBuf, Object> byteBufToObjectConverter() {
-        return (object, targetType, context) -> conversionService.convert(object.toString(context.getCharset()), targetType, context);
-    }
-
```
graemerocher, line 356: Is this not a breaking change?

http/src/main/java/io/micronaut/http/bind/DefaultRequestBinderRegistry.java
mattmoss: This chunk of code is here to handle `HttpRequest<@Nullable T>` and similar.

For a request with an empty body, the first time the argument binder is called, the body is not yet decoded. The previous version of the code would simply return `UNSATISFIED`. When the second pass (w/ a decoded body) comes around, the situation is still that body is still null, and returns `UNSATISFIED` again, even though the the request has `@Nullable T` body type.

The first solution to try was just:
```
                  if (typeVariable.get().isNullable()) {
                        return () -> Optional.of(source);
                    } else {
                        return ArgumentBinder.BindingResult.UNSATISFIED;
                    }
```

First attempt to find would see the body was empty and `typeVariable` nullable, so it returned the lambda `() -> Optional.of(source)`. This works for when the body is null. But if the request has a non-null body, the first pass sees the type variable as nullable, and returns the lambda, but this interferes with the second pass (after non-null body is decoded) which should return `new FullHttpRequest(...)`.

To get around this, I use a custom attribute on the first pass to maintain existing behavior (pre-body-decode). But for the second pass, if the body is still null (determined because the attribute exists) AND the type variable is nullable, then returns the lambda `() -> Optional.of(source)`, which makes all `@Nullable T` cases happy without messing up other cases/tests.

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConverters.java
mattmoss: I couldn't find any uses of this (which does a strange `toString` conversion that interferes with other more useful conversions). No tests broke, and no one seemed to think this was a problem when it was brought up during last engineering meeting.

jameskleeh: @graemerocher The presence of the converter prevents proper conversion of ByteBuf to container types. In general I don't think it makes sense because if we don't know the target type we shouldn't assume that a string based conversion will succeed or will be desirable

Issue #6705, opened 2022-01-03
Multipart handling errors


morki:
### Expected Behavior

No error are present when using `multipart.disk = true` and `multipart.mixed = true`.

### Actual Behaviour

Errors presented below are present.

### Steps To Reproduce

There are 2 described issues:

On route `/` there is login form for micronaut security login endpoint. You can write anything in those fields and click submit. After that, error `Internal Server Error: refCnt: 0, decrement: 1` is thrown. But when you remove configuration `mixed: true` and `disk: true`, everything is running fine.

On route `/upload` there is multipart form with file selection. Please select file `big.txt` contained in the root of this repository and click submit. Request in browser will hang. But, when you stop the server, the `Entity too large` response is sent.

### Example Application

https://github.com/morki/micronaut-multipart-issue

### Version

3.2.3

yawkat:
thanks

yawkat:
fyi `multipart.mixed` does nothing when `multipart.disk` is also set. You should probably only use the former

yawkat:
i think the hang is fixed by #6696, i will check once that is merged

morki:
Thank you very much @yawkat, it seems to be fixed :)

morki:
Ok, so it only resolved the issue with `multipart.disk` but `multipart.mixed` now throws this on login endpoint:

![image](https://user-images.githubusercontent.com/3584190/148051107-4dd05f35-66b1-46b1-9133-fde2ae10c507.png)

PR #6707, opened 2022-01-03
Fix double release of form attributes

yawkat:
Attributes returned by `postRequestDecoder.next()` are released twice when the decoder is destroyed:
- Once because it's stored in the `bodyListHttpData` list
- A second time because the `HttpDataFactory` keeps a copy around.
This means that even if user code retains the attribute in question once, it may still be collected.

In the test case, this issue is visible because the attribute is only parsed to a map later downstream in an IO thread, when the `FormDataHttpContentProcessor` has already completed.

This patch uses `removeHttpDataFromClean` to avoid the duplicate release for HTTP data where it could happen. `removeHttpDataFromClean` is a no-op for data that isn't created by the `HttpDataFactory`, so this patch doesn't break in those cases.

This patch risks exposing downstream buffer leaks: If the user code doesn't properly release attributes it retains, they would previously still have been collected because of this duplicate release. After this change, they may leak, potentially even leaving files on disk. However, imo this patch implements the "right" behavior, and we should deal with other issues as they appear.

I've done some testing, but it's inconclusive. I ran `./gradlew check` with and without this patch, and in both cases, attribute files were retained in `/tmp`. Can't tell whether this patch introduced new leaks.

Addresses #6705

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/FormDataDiskSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/FormDataDiskSpec.groovy
@@ -1,0 +1,53 @@
+package io.micronaut.http.server.netty.binding
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.HttpStatus
+import io.micronaut.http.MediaType
+import io.micronaut.http.annotation.Body
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Post
+import io.micronaut.http.client.HttpClient
+import io.micronaut.runtime.server.EmbeddedServer
+import reactor.core.publisher.Flux
+import spock.lang.Issue
+import spock.lang.PendingFeature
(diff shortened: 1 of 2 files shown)


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/FormDataDiskSpec.groovy
```diff
@@ -0,0 +1,53 @@
...
+import io.micronaut.http.client.HttpClient
+import io.micronaut.runtime.server.EmbeddedServer
+import reactor.core.publisher.Flux
+import spock.lang.Issue
+import spock.lang.PendingFeature
+import spock.lang.Specification
+
+class FormDataDiskSpec extends Specification {
+    @Issue('https://github.com/micronaut-projects/micronaut-core/issues/6705')
+    @PendingFeature
```
graemerocher: Why is the test annotated with `@PendingFeature`?

yawkat: oops, forgot to remove that when i tested for Attr file leaks pre-patch.

PR #6711, opened 2022-01-05
Configurable masking and hiding for the env endpoint

### LINKED ISSUE #6663: The Environment Endpoint security issue
### Issue description

Current design of the [environment endpoint](https://docs.micronaut.io/latest/guide/#environmentEndpoint) poses a threat for a critical security issue, since it provides all the configuration values.
The problem consist of two parts:

1. The endpoint claims to mask sensitive config values, but it doesn't provide a clear definition what it considers "sensitive". The only way to figure this out is to look at the source of `EnvironmentEndpoint` and find hard-coded list of key words. It is far from ideal, since
    - A user can define any property (`db-pass` instead of `password`)
    - A user is prone to typos (`pssword` instead of `password`)
1. By default, the endpoint can be accessed by **any** authorized user, without any checks

Given that, consider the following, quite common, use-case:

1. A developer creates an web app with regular authentication for users (via login form, for example)
1. For internal use of the app the developer sets defines property `external-service-creds : <token>` containing sensitive token 
1. After a while the developer decides to enable metrics and plugs in `management` for the `/metrics` endpoint with `senstive: false`, since they consider metrics non-sensitive information

Since the environment endpoint is enabled by default and management endpoints are exposed over the same port as the application, any authorized user now have access to the token.

I think, it is much better from the security standpoint  to: 
1. Disable all management endpoints by default
1. Make the list of key words configurable 
2. Force developers to implement custom `SensitiveEndpointRule` if they use the environment endpoint

Example is located here: https://github.com/ivyazmitinov/env-endpoint-issue

timyates:
- Changes the env endpoint to be disabled by default (requires user to enable in config)
- If enabled, all env endpoint values are masked with *****
- Allow user to add a bean implementing EnvironmentEndpointFilter which:
  - allows legacy operation to be allowed so it behaves as before
  - can be set to mask or reveal all
  - allows patterns for masking to be defined if reveal all is set
  - allows patterns for plain-text values if mask all is set

The call to the filter also contains the Principal if there is one, so the user can choose
to alter visibility of items based on the currently logged in user


management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
```diff
@@ -17,104 +17,116 @@
```
timyates, line 17: This is a breaking change as it is a modification to the public API.  Discussed with @jameskleeh and we decided we should be ok, as we couldn't see why someone would be creating one of these themselves, nor extending it for their own use.

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public EnvironmentFilterSpecification legacyMasking() {
+        allMasked = false;
+        maskedPatterns.addAll(
+                Arrays.stream(PROPERTY_NAMES_TO_MASK)
+                        .map(s -> Pattern.compile(".*" + s + ".*", Pattern.CASE_INSENSITIVE))
+                        .collect(Collectors.toList())
+        );
+        return this;
+    }
```
timyates: Are we ok compiling these patterns each request?  We could compile them at startup, or could use something like an AtomicReference compile them on demand and cache them

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
```diff
@@ -17,104 +17,116 @@
...
     private final Environment environment;
-    private final List<Pattern> maskPatterns;
+    private final EnvironmentEndpointFilter environmentFilter;
 
     /**
      * @param environment The {@link Environment}
+     * @param environmentFilter The registered {@link EnvironmentEndpointFilter} bean if one is registered
      */
-    public EnvironmentEndpoint(Environment environment) {
+    public EnvironmentEndpoint(Environment environment, @Nullable EnvironmentEndpointFilter environmentFilter) {
```
graemerocher, line 67: Need to add a new constructor and not modify this one to avoid breaking binary compatiblity

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpointFilter.java
```diff
@@ -0,0 +1,33 @@
...
+ * @since 3.2.5
+ */
+public interface EnvironmentEndpointFilter {
+
+    /**
+     * Configure the filtering of PropertySources for the environment endpoint.
+     *
+     * @param specification a specification of which properties are masked or hidden from the endpoint.
+     */
+    void specifyFiltering(EnvironmentFilterSpecification specification);
```
graemerocher: Add `@NonNull` to argument

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,121 @@
...
+    public Principal getPrincipal() {
+        return principal;
+    }
+
+    /**
+     * Turn on global masking. Items can be unmasked by adding their {@link Pattern} to {@link EnvironmentFilterSpecification#maskPatterns(Pattern...)}.
+     *
+     * @return itself for chaining calls.
+     */
+    public EnvironmentFilterSpecification maskAll() {
```
graemerocher: Add `@NonNull` to return type

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,121 @@
...
+        allMasked = true;
+        return this;
+    }
+
+    /**
+     * Turn off global masking. Items can be masked by adding their {@link Pattern} to {@link EnvironmentFilterSpecification#maskPatterns(Pattern...)} (Pattern...)}.
+     *
+     * @return itself for chaining calls.
+     */
+    public EnvironmentFilterSpecification maskNone() {
```
graemerocher: Add `@NonNull` to return type

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Adds patterns to the list of known mask patterns.  If the {@link EnvironmentFilterSpecification#maskAll()} flag is set
+     * then this list will be used for exceptions (values in clear-text).  If the {@link EnvironmentFilterSpecification#maskNone()}
+     * flag is set then this list will be those values that are masked.
+     *
+     * @param propertySourceNameMask one or more Patterns to match against property names for masking.
+     * @return itself for chaining calls.
+     */
+    public EnvironmentFilterSpecification maskPatterns(Pattern... propertySourceNameMask) {
```
graemerocher: Add `@NonNull` to return type

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Adds the masking rules from legacy pre 3.3.0.  Effectively the same as calling {@link EnvironmentFilterSpecification#maskNone()}
+     * and adding patterns via {@link EnvironmentFilterSpecification#maskPatterns(Pattern...)} to mask anything containing the words
+     * {@code password, credential, certificate, key, secret} or {@code token}.
+     *
+     * @return itself for chaining calls.
+     */
+    public EnvironmentFilterSpecification legacyMasking() {
```
graemerocher: Add `@NonNull` to return type

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
graemerocher: Makes sense to compile then on startup

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * This class is passed to an instance of {@link EnvironmentEndpointFilter} (if one is defined) each time the {@link EnvironmentEndpoint}
+ * is invoked.
+ *
+ * @author Tim Yates
+ * @since 3.3.0
+ */
+public class EnvironmentFilterSpecification {
```
graemerocher: Looks like this can be made final

src/main/docs/guide/management/providedEndpoints/environmentEndpoint.adoc
```diff
@@ -1,20 +1,66 @@
...
+            .maskNone() // All values will be in plain-text apart from the supplied patterns
+            .maskPatterns(
+                Pattern.compile(".*secret.*", Pattern.CASE_INSENSITIVE),
+                Pattern.compile(".*key.*", Pattern.CASE_INSENSITIVE)
+            );
+    }
+}
+----
+
+The link:{api}/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.html[EnvironmentFilterSpecification] class
```
graemerocher: You can use `api:management.endpoint.env.EnvironmentFilterSpecification[]`

src/main/docs/guide/management/providedEndpoints/environmentEndpoint.adoc
```diff
@@ -1,20 +1,66 @@
...
+also contains the current `Principal` if one exists, so this can be used to further limit access based on the
+currently logged-in user.
+
+Sensible defaults can be applied by calling the link:{api}/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.html#legacyMasking--[`legacyMasking()`] method.
```
graemerocher: Same here use `api:` macro

src/main/docs/guide/management/providedEndpoints/environmentEndpoint.adoc
```diff
@@ -1,20 +1,66 @@
...
 ----
 endpoints:
   env:
-    enabled: Boolean    # default: true
+    enabled: Boolean    # default: false
     sensitive: Boolean  # default: true
 ----
 
+By default the endpoint will mask all values.
+To customize this masking you need to supply a Bean that implements link:{api}/io/micronaut/management/endpoint/env/EnvironmentEndpointFilter.html[EnvironmentEndpointFilter].
```
graemerocher: You can use `api:management.endpoint.env.EnvironmentEndpointFilter[]`

graemerocher (changes requested):

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpointFilter.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.management.endpoint.env;
+
+/**
+ * A bean interface that allows hiding or masking of parts of the environment and its property sources when they are
+ * displayed in the {@link EnvironmentEndpoint}.
+ *
+ * @author Tim Yates
+ * @since 3.2.5
```
ilopmar: Since 3.3.0

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
timyates: I tried with both, and the `EnvironmentEndpointSpec` tests failed (I guess because the one arg constructor got used even though an `EnvironmentEndpointFilter` was defined as a bean)

Is there a way to get it to prefer the two-arg constructor?  Would `@Inject` help?  I'll see

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpointFilter.java
timyates: Done 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: 👍

src/main/docs/guide/management/providedEndpoints/environmentEndpoint.adoc
timyates: Nice!

src/main/docs/guide/management/providedEndpoints/environmentEndpoint.adoc
timyates: 👍

src/main/docs/guide/management/providedEndpoints/environmentEndpoint.adoc
timyates: 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpointFilter.java
timyates: Nice catch!  Teach me to guess the version before asking 😉

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: Done.  I used a Stream to compile them all.  Just a plain repetitive array might be quicker, but with only 6 entries, I'm not sure by how much...

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
timyates: I think `@Inject` does it! 😎  [`7bc7553` (#6711)](https://github.com/micronaut-projects/micronaut-core/pull/6711/commits/7bc7553cef4cd65382f892ae950339dcd7042495)

timyates:
@graemerocher Hope I addressed all your feedback 🤞

graemerocher:
@timyates from my side yeah, @jameskleeh needs to review next

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Adds patterns to the list of known mask patterns.  If the {@link EnvironmentFilterSpecification#maskAll()} flag is set
+     * then this list will be used for exceptions (values in clear-text).  If the {@link EnvironmentFilterSpecification#maskNone()}
+     * flag is set then this list will be those values that are masked.
+     *
+     * @param propertySourceNameMask one or more Patterns to match against property names for masking.
+     * @return itself for chaining calls.
+     */
+    public @NotNull EnvironmentFilterSpecification maskPatterns(Pattern... propertySourceNameMask) {
+        maskedPatterns.addAll(Arrays.asList(propertySourceNameMask));
```
jameskleeh: This would have no effect without `allMasked = false;` so that should probably be set here

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Adds the masking rules from legacy pre 3.3.0.  Effectively the same as calling {@link EnvironmentFilterSpecification#maskNone()}
+     * and adding patterns via {@link EnvironmentFilterSpecification#maskPatterns(Pattern...)} to mask anything containing the words
+     * {@code password, credential, certificate, key, secret} or {@code token}.
+     *
+     * @return itself for chaining calls.
+     */
+    public @NotNull EnvironmentFilterSpecification legacyMasking() {
+        allMasked = false;
+        maskedPatterns.addAll(LEGACY_MASK_PATTERNS);
```
jameskleeh: This should just call `maskPatterns`

jameskleeh (changes requested):
I think this is missing:

- the ability to mask all properties except a set a patterns
- the ability to mask all properties except a set of static keys
- the ability to mask a set of static keys

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: If `allMasked = true`, then this sets the patterns that are allowed to be in the clear

https://github.com/micronaut-projects/micronaut-core/pull/6711/files#diff-4e7e5d0b9696543f54847a8d042130fc33e81143e23d6c414504d106ec0fa4c7R15

`maskPatterns` is probably a rubbish name...  But `toggleMask` was wrong too, as it's not a toggle, it's additive

jameskleeh: I think `maskPatterns` should define the patterns that are masked, thus only making it useful if not all keys are masked

timyates: With a separate list for unmasked patterns, that only come into play if allMasked is true?

Or we could name it "excludedPatterns" which are excluded from whichever mode you're in... And change the method to exclude() ?

jameskleeh: That sounds OK to me. Need to support static keys as well

jameskleeh: Also the ability to supply predicates. Everything could be converted to predicates

timyates: Switch to using `Predicate<String>` for all `exclusions`

To avoid `unchecked generic array creation for varargs parameter` warnings in user code (due to the `Predicate<String>` varargs), introduced `EnvironmentFilterNamePredicate` as a typealias

Also added a factory method to generate a predicate from a pattern and optional flags

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: Renamed to `exclusions` as discussed 👍

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
```diff
@@ -17,104 +17,125 @@
...
     }
 
     /**
      * @param propertySourceName The {@link PropertySource} name
+     * @param principal The current {@link Principal} if one exists
      * @return a map with all the properties defined in the property source if it exists; null otherwise.
      */
     @Read
-    public Map<String, Object> getProperties(@Selector String propertySourceName) {
+    public Map<String, Object> getProperties(@Selector String propertySourceName, @Nullable Principal principal) {
```
graemerocher: This is breaking binary compatibility. Add another deprecated method with the original signature.

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
```diff
@@ -17,104 +17,125 @@
...
     }
 
     /**
+     * @param principal The current {@link Principal} if one exists
      * @return The environment information as a map with the following keys: activeEnvironments, packages and
      * propertySources.
      */
     @Read
-    public Map<String, Object> getEnvironmentInfo() {
+    public Map<String, Object> getEnvironmentInfo(@Nullable Principal principal) {
```
graemerocher: This is breaking binary compatibility. Add another deprecated method with the original signature.

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public static EnvironmentFilterNamePredicate regularExpressionPredicate(String pattern) {
+        return regularExpressionPredicate(pattern, 0);
+    }
+
+    public static EnvironmentFilterNamePredicate regularExpressionPredicate(String pattern, int flags) {
+        return s -> Pattern.compile(pattern, flags).matcher(s).matches();
+    }
+
+    public interface EnvironmentFilterNamePredicate extends Predicate<String> {
```
graemerocher: This appears to be now part of the public API, therefore needs javadoc

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
```diff
@@ -0,0 +1,131 @@
...
+        return allMasked ? FilterResult.MASK : FilterResult.PLAIN;
+    }
+
+    public static EnvironmentFilterNamePredicate regularExpressionPredicate(String pattern) {
+        return regularExpressionPredicate(pattern, 0);
+    }
+
+    public static EnvironmentFilterNamePredicate regularExpressionPredicate(String pattern, int flags) {
+        return s -> Pattern.compile(pattern, flags).matcher(s).matches();
+    }
```
graemerocher: These appear now to be part of the public API? if it is the case they need javadoc

graemerocher (changes requested):

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: Done.  Would you say they were necessary?  I added them to make it simpler to create the Predicates, but can see that they may add noise 🤔

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentFilterSpecification.java
timyates: Done

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
timyates: Done

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
timyates: Done

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
jameskleeh: I would rather we just not handle the principal rather than adding a new method. Even adding the method is a breaking change because the original method and any extensions of it would never be called

timyates:
Also, having the `EnvironmentFilterNamePredicate` type-alias as an inner-class makes the javadoc quite long 😐 Should it be made top-level?  Or does someone know of a better way around the `unchecked generic array creation for varargs parameter` warning?

timyates:
After a sync up with James, we have a better API... Incoming asap

jameskleeh (approved):

jameskleeh:
@graemerocher Can you please give this a final look?

graemerocher (approved):

PR #6718, opened 2022-01-07
Do not insert `@RequestAttribute`s as path params

### LINKED ISSUE #6717: @RequestAttribute in HTTP Client is being treated as a Query Parameter in 3.2.x
### Expected Behavior

> Request attributes are intended for internal data sharing only, and are not attached to the outbound request.

https://docs.micronaut.io/latest/api/io/micronaut/http/annotation/RequestAttribute.html

### Actual Behaviour

They're being bound as Query Parameters, for example

```kotlin
@Client(id = "example")
interface ExampleClient {
  @Post("example/uri")
  suspend fun postExample(@Body body: requestExample, @RequestAttribute("example") metadata: Int): ExampleResponse
}
```

Expected URI:

```
https://example.com/example/uri
```

Actual URI:

```
https://example.com/example/uri?example=5
```

## Real Example:

GitHub Action Error:

https://github.com/NickHackman/micronaut-request-attributes-bug/runs/4730047911?check_suite_focus=true#step:4:41

Code:

https://github.com/NickHackman/micronaut-request-attributes-bug/blob/main/src/test/kotlin/example/request/attribute/bug/ExampleRequestAttributeBugTest.kt#L50-L82

### Steps To Reproduce

Above code provided in "Actual Behavior" or use example Application provided.

### Environment Information

- Operating System: MacOS
- Java: 11
- Kotlin: 1.6.0

### Example Application

https://github.com/NickHackman/micronaut-request-attributes-bug

### Version

3.2.x

yawkat:
a3c46be7bf2cb6fc91a90985e767fb4f19782b8d introduced the possibility of using `@RequestAttribute` parameters as path parameters. However, path parameters that aren't explicitly bound are inserted as query parameters, so the `@RequestAttribute` is inadvertently leaked.

This change reverts back to the old behavior.

Fixes #6717

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/bind/DefaultHttpClientBinderRegistry.java
@@ -124,11 +124,20 @@ public class DefaultHttpClientBinderRegistry implements HttpClientBinderRegistry
         });
         byAnnotation.put(RequestAttribute.class, (context, uriContext, value, request) -> {
             AnnotationMetadata annotationMetadata = context.getAnnotationMetadata();
+            String name = context.getArgument().getName();
             String attributeName = annotationMetadata
                     .stringValue(RequestAttribute.class)
                     .filter(StringUtils::isNotEmpty)
-                    .orElse(NameUtils.hyphenate(context.getArgument().getName()));
+                    .orElse(NameUtils.hyphenate(name));
             request.getAttributes().put(attributeName, value);
+
+            conversionService.convert(value, ConversionContext.STRING.with(context.getAnnotationMetadata()))
+                    .filter(StringUtils::isNotEmpty)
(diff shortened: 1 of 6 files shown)


graemerocher (approved):

jameskleeh (changes requested):
In general I agree with the change but I think instead of breaking the existing use case we can maintain the current behavior and fix the current bug by checking if a variable exists in the template.

```
            conversionService.convert(value, ConversionContext.STRING.with(context.getAnnotationMetadata()))
                    .filter(StringUtils::isNotEmpty)
                    .ifPresent(param -> {
                        String pathName = context.getArgument().getName();
                        if (uriContext.getUriTemplate().getVariableNames().contains(pathName)) {
                            uriContext.getPathParameters().put(pathName, param);
                        }
                    });
```

yawkat:
good idea, done.

jameskleeh (approved):

PR #6719, opened 2022-01-07
Retry flaky test

yawkat:
Can't make this test fail locally for debugging, unfortunately. CI logs show 'RTE outer', which makes sense for the failure, but I'm not sure why it would time out.


graemerocher (approved):

timyates:
Nice.  The junit-report action doesn't handle retried and successful tests, so I sent a PR to hopefully improve this https://github.com/mikepenz/action-junit-report/pull/479

PR #6721, opened 2022-01-07
Fix bug that makes it impossible to annotate inner types

graemerocher:
When dynamically annotating a type the key used to store the metadata uses the binary name whilst lookups using the canonical name. This PR updates Micronaut to always use the canonical name for mutated metadata keys.

In addition it shifts logic away from using our custom code to resolve the binary name which may not be correct and it is better to use the built in method within the JDK to achieve this.


jameskleeh:
In what cases would `elementUtils.getBinaryName` and `JavaModelUtils.getClassName` return different results? I'm concerned about compatibility with previously compiled annotation metadata

graemerocher:
rolled back the other changes for now, to address just the bug in question. We can leave the binary name stuff for now although I think it is wrong we are trying to compute this manually

PR #6728, opened 2022-01-10
Allow file watcher restarts for messaging apps

### LINKED ISSUE #2583: File watch not working for plain kafka project
Thanks for reporting an issue for Micronaut, please review the task list below before submitting the

NOTE: If you are unsure about something and the issue is more of a question a better place to ask questions is on Stack Overflow (http://stackoverflow.com/tags/micronaut) or Gitter (https://gitter.im/micronautfw/). DO NOT use the issue tracker to ask questions.

### Task List

### Steps to Reproduce

1. Create a new micronaut project without HTTP service, in my case project was created with Kafka only.
2. Add file watch dependencies in gradle. (since I am using mac os).
    developmentOnly "io.micronaut:micronaut-runtime-osx"
    developmentOnly "net.java.dev.jna:jna"
    developmentOnly "io.methvin:directory-watcher"

3. ./gradlew clean run - while the project is running, make any change in source code,

### Expected Behaviour

Expectation is that the application should shutdown (or restart if used --continuous) as soon as any file is modified from src folder.

### Actual Behaviour

Application does not shutdown OR restart after file change

### Environment Information

- **Operating System**:  MacOS
- **Micronaut Version:** Micronaut Version: 1.1.3
| JVM Version: 1.8.0_212
- **JDK Version:** 8.0.212-zulu 

### Example Application

None: But when I added Http dependencies in my gradle, it started watching the src/main folder.

My expectation is, any project type (without http in my case) should watch file changes.

timyates:
The file watcher required an EmbeddedServer to be enabled.

If you create a Kafka app with no http enabled, ie:

```
mn create-messaging-app messaging-test --features kafka
```

then you get a MessagingApplication in place of an EmbeddedServer.

This change moves the requirement up to an EmbeddedApplication<?> instead of a specific
EmbeddedServer implementation of it, so we listen for restart events for that type.

Targeting v3.3.0 as it's a breaking change to the public API

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/server/watch/event/FileWatchRestartListener.java
+++ b/context/src/main/java/io/micronaut/runtime/server/watch/event/FileWatchRestartListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


timyates:
This is for #2583

timyates:
Can't find that issue in the Linked Issues dropdown 🤔

graemerocher (approved):

PR #6734, opened 2022-01-10
Clear annotation defaults cache to prevent pollution from one language to another

jameskleeh:
This also reorders the reference classes to register annotation defaults before creating annotation metadata because the creation references the defaults.


dstepanov:
I have checked this PR fixes errors in Micronaut Data

graemerocher:
fails a lot of tests though :/

graemerocher (approved):

dstepanov:
Looks like it fixed some compile errors but then 3.2.5 has introduced other problems. Cases with `ap.getRequiredValue(AutoPopulated.UPDATEABLE, Boolean.class)` fail on required value and if I rewrite them `./gradlew clean :data-jdbc:test` will fail on missing compilation info at the runtime.

https://github.com/micronaut-projects/micronaut-data/compare/err

graemerocher:
Hmm I guess the changes added between when you verified the PR and it was merged caused the issues

dstepanov:
Actually, I have only checked the `:data-processor:test` expecting that runtime will be fine

PR #6737, opened 2022-01-11
Allow configuring more logging options using properties

dstepanov:


inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -269,11 +270,20 @@ public Environment start() {
...
         return this;
     }
 
+    private void applyLogging() {
+        List<LoggingConfigurer> list = new ArrayList<>(3);
+        SoftServiceLoader.load(LoggingConfigurer.class, getClassLoader()).collectAll(list);
+        for (LoggingConfigurer loggingConfigurer : list) {
+            loggingConfigurer.apply(this);
+        }
+    }
```
graemerocher: Can we run this logic only if config is present in the environment otherwise we are adding more startup cost

inject/src/main/java/io/micronaut/context/logging/DefaultLogbackLoggingConfigurer.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Logback default implementation of {@link LoggingConfigurer}.
+ *
+ * @author Denis Stepanov
+ * @since 3.3
+ */
+public class DefaultLogbackLoggingConfigurer implements LoggingConfigurer {
+
+    @Override
+    public void apply(Environment environment) {
```
graemerocher: what happens if logbook is not on the path?

dstepanov: I would assume that `SoftServiceLoader` is going to exclude the entry.

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
dstepanov: I can check if `logger` prefix exists.

inject/src/main/java/io/micronaut/context/logging/DefaultLogbackLoggingConfigurer.java
graemerocher: probably need to test that

Issue #6433, opened 2021-10-27
Micronaut caching elements from previous compiler runs which consumes lots of memory


thurka:
### Actual Behaviour

Scenario: I have running VS Code for hours on end hacking my way through a app (which was initially created by Burt for his ATP guide, link below). After a while, I noticed that the system was coming to a crawl. So, poking around in Task Manager, Net Beans Server was taking up lots of memory. see picture
![image](https://user-images.githubusercontent.com/22048069/139058531-9b076e4b-5105-4ac0-8859-a8e0a6c5bc9d.jpeg)

Guide: 
https://guides.micronaut.io/latest/micronaut-oracle-autonomous-db-gradle-java.html

### Environment Information

My System: Windows 10 Pro. Has 16 GB of Memory.
Windows Version 10.0.19043 Build 19043Version 10.0.19043 Build 19043

### Example Application

https://guides.micronaut.io/latest/micronaut-oracle-autonomous-db-gradle-java.html

### Version

3.0.1

thurka:
I have got the heap dump from `nbcode64` process. Heap dump has around 200M and it looks like 15% (around 30M) is occupied by static hash map `io.micronaut.annotation.processing.JavaAnnotationMetadataBuilder.ANNOTATION_DEFAULTS`. This map contains FQN of annotation type as key and list of default values for all annotation members as value. The list of default values uses instances of `javax.lang.model.element.ExecutableElement` to reference particular annotation type members and it means that this element holds all parsing data used at the time annotation type is added to `ANNOTATION_DEFAULTS` map. Since `nbcode64` process re-uses `JavaAnnotationMetadataBuilder` annotation processor, the `ANNOTATION_DEFAULTS` map accumulates references to `javax.lang.model.element.ExecutableElement` instances from different java compiler runs. This leads to unnecessary large size of `ANNOTATION_DEFAULTS` map. The number of annotation types cached in the map is the same as number of annotation types used in the particular project. However if the user opens already written project, annotation types are added to the map when project is indexed, so all cached `javax.lang.model.element.ExecutableElement` instances are from the same java compiler run, which is not that bad. The worst scenario occurs when annotation types are added to the project source files one-by-one (when the user actually writes the java code and uses annotations). This way annotation types are added to `ANNOTATION_DEFAULTS` map by different java compiler runs and thus such map occupy more java heap. Below is the textual dump of `ANNOTATION_DEFAULTS` map from attached heap dump. 

```
[...]

PR #6749, opened 2022-01-12
Fix mutated metadata being cleared before it was written in Groovy classes

### LINKED ISSUE #6744: Error compiling Data JDBC interface
### Expected Behavior

Micronaut Data JDBC applications should compile.

### Actual Behaviour

All the Data JDBC test applications in GraalVM CI have failed during compilation today with the error:

```
> Task :compileJava FAILED
  reason: class file for javax.annotation.meta.When not found
error: Unable to implement Repository method: PetRepository.update(Object entity). No value available for annotation member @io.micronaut.data.annotation.AutoPopulated[updateable] of type: class java.lang.Boolean
Note: Creating bean classes for 7 type elements
  reason: class file for javax.annotation.meta.When not found
1 error
2 warnings
```

They worked yesterday and nothing has changed in Micronaut Data so the error should be here in core.

### Steps To Reproduce

- `git clone https://github.com/micronaut-graal-tests/micronaut-data-jdbc-graal`
- `cd micronaut-data-jdbc-graal`
- `git checkout 3.2.x_h2`
- `./gradlew compileJava`

### Environment Information

Linux Mint 20.3
JDK 8

### Example Application

_No response_

### Version

3.2.5-SNAPSHOT

jameskleeh:
@dstepanov Please give this a try with data


dstepanov:
It fixes runtime errors but there are still some compile problems with `getRequiredValue` not returning default values. You can reproduce them by reverting modified Java files.

jameskleeh:
Right because the code to retrieve the default values only looks at the runtime source of that information. If we should not leak the compile time data to runtime, which was the cause of the original issue, then either default value data will not be available at compile time or the methods to get the defaults needs to look at the compile time defaults.

dstepanov:
We have code that relies on compile-time default values so we need to fix it somehow 🤔

graemerocher:
@jameskleeh can you link to the original issue, the original PR that broke everything to solve that issue and so on. None of the PRs that tried to mitigate this issue include any references to the issues they solve so it is hard to trace back the history of the issue

dstepanov:
The problem in 3.1.4 was caused by https://github.com/micronaut-projects/micronaut-core/pull/6455 and the fix that broke more stuff https://github.com/micronaut-projects/micronaut-core/pull/6734

jameskleeh:
@graemerocher The original issue is https://github.com/micronaut-projects/micronaut-core/issues/6433

The fix I created for that was https://github.com/micronaut-projects/micronaut-core/pull/6455 which basically reused the runtime cache at compile time. Because there was a singular cache the data from Groovy was leaking to Java. Groovy is creating proxies for annotation instances and the Java annotation metadata stuff is not designed to handle them.

I tried to solve that issue by clearing the cache after compilation but that caused lots of issues with compiling code at runtime in our tests, so it was evident the runtime cache could not be cleared.

The next step I took was to separate the compile time from runtime default value caches so the compile time cache could be cleared after compilation preventing one languages values leaking to another. https://github.com/micronaut-projects/micronaut-core/pull/6734

Unfortunately I mistakenly cleared the mutated metadata too early for Groovy so that caused runtime issues of mutated metadata being missing. It also had the side effect of annotation default values not being available at compile time because the metadata only looks at the runtime cache and we aren't populating it

dstepanov:
I can confirm these changes fix the problems in Micronaut Data

PR #6752, opened 2022-01-13
Fix HTTP/2 304 responses for files don't associate the correct stream ID

graemerocher:
When a 304 response is returned from `FileTypeHandler` and HTTP/2 is enabled the stream ID is not correctly associated between the request and the response resulting in some browsers (like Firefox) hanging and unable to handle the response


yawkat (approved):

yawkat:
@graemerocher I also changed NettyStreamedCustomizableResponseType to use the new method

graemerocher:
@jameskleeh could you prioritize merging this, would like to get a snapshot for external testing

PR #6753, opened 2022-01-13
Access log can exclude defined regular expression patterns

### LINKED ISSUE #4653: Disable access logger for management endpoints
Access Logging described here: https://docs.micronaut.io/latest/guide/index.html#accessLogger is turned on for both management and regular HTTP endpoints. For services that have scrapers that hit the health / info endpoints this introduces a lot of logging and it'd be ideal to disable those logs.

### Task List

### Steps to Reproduce

* Turn on access logging with instructions in the docs: https://docs.micronaut.io/latest/guide/index.html#accessLogger
* Send request to any management endpoint

### Expected Behaviour

* There should be a flag for disabling logging for management endpoints

### Actual Behaviour

* Expected behavior doesn't exist

### Environment Information

- **Operating System**: TODO
- **Micronaut Version:** TODO
- **JDK Version:** TODO

### Example Application

- TODO: link to github repository with example that reproduces the issue

timyates:
This change allows the user to specify a list of regular expression patterns that are compiled
and matched against the current request URI.

If the request URI matches one of them, then the request does not appear in the access log.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
@@ -107,10 +107,10 @@ public class HttpAccessLogHandler extends ChannelDuplexHandler {
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Http2Exception {
         if (logger.isInfoEnabled() && msg instanceof HttpRequest) {
+            final SocketChannel channel = (SocketChannel) ctx.channel();
             final HttpRequest request = (HttpRequest) msg;
+            AccessLog accessLog = accessLog(channel);
             if (uriInclusion == null || uriInclusion.test(request.uri())) {
-
-                final SocketChannel channel = (SocketChannel) ctx.channel();
                 final HttpHeaders headers = request.headers();
                 // Trying to detect http/2
                 String protocol;
@@ -119,9 +119,9 @@ public class HttpAccessLogHandler extends ChannelDuplexHandler {
                 } else {
                     protocol = request.protocolVersion().text();
                 }
-                accessLog(channel).onRequestHeaders(channel, request.method().name(), request.headers(), request.uri(), protocol);
+                accessLog.onRequestHeaders(channel, request.method().name(), request.headers(), request.uri(), protocol);
             } else {
-                ctx.pipeline().remove(this);
+                accessLog.exclude();
             }
         }
         ctx.fireChannelRead(msg);
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
@@ -107,10 +107,10 @@ public class HttpAccessLogHandler extends ChannelDuplexHandler {
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Http2Exception {
         if (logger.isInfoEnabled() && msg instanceof HttpRequest) {
-            final SocketChannel channel = (SocketChannel) ctx.channel();
             final HttpRequest request = (HttpRequest) msg;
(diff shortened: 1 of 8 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
```diff
@@ -50,11 +52,17 @@
      */
     public static final String HTTP_ACCESS_LOGGER = "HTTP_ACCESS_LOGGER";
 
+    /**
+     * A lenient filter that allows all uri's to pass.
+     */
+    public static final Predicate<String> INCLUDE_ALL = uri -> true;
```
graemerocher: Perhaps use `null` as the value for `INCLUDE_ALL` as it will add a little less overhead

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
timyates: Good shout.  Done 👍

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
```diff
@@ -73,25 +87,42 @@ public HttpAccessLogHandler(String loggerName, String spec) {
...
+                String protocol;
+                if (headers.contains(ExtensionHeaderNames.STREAM_ID.text()) || headers.contains(ExtensionHeaderNames.SCHEME.text())) {
+                    protocol = H2_PROTOCOL_NAME;
+                } else {
+                    protocol = request.protocolVersion().text();
+                }
+                accessLog.onRequestHeaders(channel, request.method().name(), request.headers(), request.uri(), protocol);
             } else {
-                protocol = request.protocolVersion().text();
+                accessLog.exclude();
```
jameskleeh, line 124: @graemerocher Would it be safe here to remove the handler from the pipeline? Setting an excluded flag and checking it is extra overhead

timyates: Nice idea!

```
                ctx.pipeline().remove(this);
```

Seems to work in the test...  Checking, and if there's no issues I can find I'll push

timyates: Load tested it with a filtered url and a non-filtered url

```
ab -k -n 100000 -r -c 5 http://localhost:8080/path/one & ab -k -r -n 100000 -c 5 http://localhost:8080/something
```

And we didn't get any errors.  I'm not a Netty-guru (yet) though, so @graemerocher may know if this is a dangerous thing to be doing

Anyway, pushed the change 👍

timyates: Right, so that doesn't work with keep-alive connections (all logs get dropped after the first excluded one removes the handler from the pipeline)

Reverted back to the field

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
```diff
@@ -55,6 +57,7 @@
 
     private final Logger logger;
     private final AccessLogFormatParser accessLogFormatParser;
+    private final Predicate<String> uriInclusion;
```
yawkat, line 60: `@Nullable`

yawkat (approved):

yawkat:
@jameskleeh can you merge this before monday so i can fix up #6785 in time for 3.3?

jameskleeh:
@yawkat Yeah I wasn't sure if you wanted to make the edits on this PR or not. If you're good with it I'll merge

yawkat:
@jameskleeh i don't have any changes to make to this PR. I've put them all in #6785, because they're unrelated to the exclusions feature.

PR #6757, opened 2022-01-14
Add jaxrs-server-security to catalog

pgressa:
This one is missing from the https://github.com/micronaut-projects/micronaut-jaxrs


graemerocher:
Ideally in the future the module should publish a BOM

pgressa:
Maybe we should have automatically add bom support to all projects while working on this new build plugins. But for now, the launcher with jaxrs-security just fails since it doesn't have proper version set.

melix:
I'm not sure adding a BOM to all modules makes sense. It probably only makes sense for multi-module projects. Note that we can also improve our BOMs for Gradle, to enable version alignment.

graemerocher (approved):

pgressa:
> I'm not sure adding a BOM to all modules makes sense. It probably only makes sense for multi-module projects. Note that we can also improve our BOMs for Gradle, to enable version alignment.

Yes, I was referring to multi-module projects only.

PR #6768, opened 2022-01-17
Allow exclusion of http requests from tracing

timyates:
This change allows configuration of a list of regular-expression patterns for tracing.

If the current path matches one of these patterns, then tracing is not performed on that
request.

Follows the same config naming strategy as #6753


graemerocher:
@timyates the only issue is that @burtbeckwith moved this code to https://github.com/micronaut-projects/micronaut-tracing

timyates:
> @timyates the only issue is that @burtbeckwith moved this code to https://github.com/micronaut-projects/micronaut-tracing

Ooops...  Is this code here in core just for legacy compatibility?

If I make the changes in `micronaut-projects/micronaut-tracing` do I then need to make changes here?  Or should it just be done over there?

So many questions 😀

timyates:
Closed in favour of https://github.com/micronaut-projects/micronaut-tracing/pull/12

PR #6776, opened 2022-01-19
build: bump non-micronaut managed patch dependencies

pgressa:


gradle/libs.versions.toml
```diff
@@ -16,42 +16,42 @@ jakarta-inject-api = "2.0.1"
...
 # that they will appear in the Micronaut BOM as <properties>
 #
 managed-azure-function = "1.4.2"
-managed-brave-instrumentation = "5.13.3"
+managed-brave-instrumentation = "5.13.7"
 managed-brave-opentracing = "1.0.0"
 managed-commons-dbcp = "2.8.0"
 managed-dekorate = "1.0.3"
-managed-elasticsearch = "7.16.1"
+managed-elasticsearch = "7.16.3"
```
ilopmar: This is managed by micronaut-elasticsearch https://github.com/micronaut-projects/micronaut-elasticsearch/blob/master/gradle.properties#L19

gradle/libs.versions.toml
```diff
@@ -64,18 +64,18 @@ managed-jaeger = "1.6.0"
...
-managed-jcache = "1.1.0"
+managed-jcache = "1.1.1"
 managed-jna = "5.7.0"
 managed-jsr305 = "3.0.2"
-managed-kafka = "2.8.0"
-managed-ktor = "1.6.1"
-managed-lettuce = "6.1.1.RELEASE"
+managed-kafka = "2.8.1"
+managed-ktor = "1.6.7"
+managed-lettuce = "6.1.6.RELEASE"
```
ilopmar: These are managed by micronaut kafka, kotlin and redis modules

gradle/libs.versions.toml
```diff
@@ -64,18 +64,18 @@ managed-jaeger = "1.6.0"
...
 managed-liquibase = "4.6.1"
-managed-logback = "1.2.3"
-managed-lombok = "1.18.20"
+managed-logback = "1.2.9"
+managed-lombok = "1.18.22"
 managed-maven-native-plugin = "0.9.7.1"
 managed-methvin-directory-watcher = "0.15.0"
-managed-micrometer = "1.8.0"
+managed-micrometer = "1.8.2"
```
ilopmar: Managed by micronaut-micrometer

gradle/libs.versions.toml
```diff
@@ -127,32 +127,32 @@ managed-micronaut-views = "3.0.4"
 managed-micronaut-xml = "3.0.0"
 managed-mongo = "4.3.0"
 managed-mongo-reactive = "4.3.0"
-managed-neo4j = "3.5.29"
+managed-neo4j = "3.5.30"
```
ilopmar: Managed by micronaut-neo4j

gradle/libs.versions.toml
```diff
@@ -127,32 +127,32 @@ managed-micronaut-views = "3.0.4"
...
 managed-neo4j-java-driver = "4.2.7"
 managed-netty = "4.1.73.Final"
 managed-opentracing = "0.33.0"
 managed-paho-v3 = "1.2.5"
 managed-paho-v5 = "1.2.5"
-managed-picocli = "4.6.1"
+managed-picocli = "4.6.2"
```
ilopmar: Managed by  micronaut-picocli

gradle/libs.versions.toml
```diff
@@ -127,32 +127,32 @@ managed-micronaut-views = "3.0.4"
...
 managed-protobuf = "3.17.2"
 managed-reactive-pg-client = "0.11.4"
 managed-reactive-streams = "1.0.3"
-managed-reactor = "3.4.8"
+managed-reactor = "3.4.14"
 managed-rxjava1 = "1.3.8"
 managed-rxjava1-interop = "0.13.7"
-managed-rxjava2 = "2.2.10"
+managed-rxjava2 = "2.2.21"
```
ilopmar: Managed by  micronaut-rxjava2 module

gradle/libs.versions.toml
```diff
@@ -127,32 +127,32 @@ managed-micronaut-views = "3.0.4"
...
 managed-slf4j = "1.7.29"
 managed-spock = "2.0-groovy-3.0"
 managed-spotbugs = "4.0.3"
-managed-spring = "5.3.9"
+managed-spring = "5.3.15"
 managed-springboot = "2.5.3"
-managed-swagger = "2.1.10"
+managed-swagger = "2.1.12"
```
ilopmar: Managed by  micronaut-openapi module

gradle/libs.versions.toml
```diff
@@ -127,32 +127,32 @@ managed-micronaut-views = "3.0.4"
...
+managed-reactor = "3.4.14"
 managed-rxjava1 = "1.3.8"
 managed-rxjava1-interop = "0.13.7"
-managed-rxjava2 = "2.2.10"
+managed-rxjava2 = "2.2.21"
 managed-slf4j = "1.7.29"
 managed-spock = "2.0-groovy-3.0"
 managed-spotbugs = "4.0.3"
-managed-spring = "5.3.9"
+managed-spring = "5.3.15"
```
ilopmar: Managed by  micronaut-spring module

ilopmar (changes requested):
Pretty much all your changes shouldn't be done here but in the appropriate micronaut modules.

pgressa:
@ilopmar how to reliably identify such modules? And does that mean that our current dependency upgrades job is broken right?

graemerocher:
@melix thoughts?

ilopmar:
> @ilopmar how to reliably identify such modules? And does that mean that our current dependency upgrades job is broken right?

We may miss something but if there is a micronaut-xxxxxx module with the same/related name that's probably the one responsible for managing the upgrade of the libraries. 
I agree that some times it can be tricky because modules like micronaut-sql handle different dependencies: https://github.com/micronaut-projects/micronaut-sql/blob/master/gradle.properties#L36

melix:
It's difficult to figure out what manages what, because there's a circular dependency: `micronaut-core` defines the "master" BOM, but also includes other BOMs, which are from modules which depend on `core` and may be using dependencies defined in the `micronaut-core` BOM.

I guess something we can do is to have a job or task which actually checks if a dependency is defined both in the micronaut core BOM and any of the modules it includes. If so, then we must decide whether it makes sense to include it in the core BOM (because, a dependency may want to be managed by core independently, if it makes sense in multiple modules).

graemerocher:
yeah I think personally the solution is that these versions shouldn't be managed by core and the module that updates the version in the core BOM should instead express it in their own BOM

gradle/libs.versions.toml
graemerocher: This version should be defined in https://github.com/micronaut-projects/micronaut-micrometer/blob/master/micrometer-bom/build.gradle

gradle/libs.versions.toml
graemerocher: Micronaut Spring should provide a BOM

gradle/libs.versions.toml
graemerocher: Micronaut OpenAPI should provide a BOM

gradle/libs.versions.toml
graemerocher: Micronaut RxJava 2 should provide a BOM

gradle/libs.versions.toml
graemerocher: Micronaut PicoCLI should provide a BOM

gradle/libs.versions.toml
graemerocher: Micronaut Neo4j should provide a BOM

gradle/libs.versions.toml
graemerocher: Micronaut Kotlin, Kafka and Redis should provide BOMs

ilopmar:
So that means that every micronaut module should publish its own bom? even for modules with one "exposed/managed" dependency (flyway, liquibase,...)

graemerocher:
yes ultimately those versions are only tested in the modules so the only way to know compatibility with the version is in the sub module anyway which makes these PRs close to useless

ilopmar (approved):

PR #6777, opened 2022-01-19
Bump discovery-client to 3.1.0

graemerocher:
Also corrects the ID of the managed dependencies since the update BOM task is expecting `discovery-client`


gradle/libs.versions.toml
```diff
@@ -83,7 +83,7 @@ managed-micronaut-cache = "3.0.0"
 managed-micronaut-cassandra = "4.0.0"
 managed-micronaut-coherence = "3.0.1"
 managed-micronaut-data = "3.2.2"
-managed-micronaut-discovery = "3.0.0"
+managed-micronaut-discovery-client = "3.1.0"
```
melix: This is a potentially breaking change: if someone uses the `micronaut.discovery` property in their POM file, or `libs.micronaut.discovery` (in Gradle, unlikely since we didn't promote the feature). Not necessarily blocking, but wanted to mention this.

jameskleeh: @graemerocher Can we have both until 4.0?

graemerocher: yeah probably makes sense

PR #6785, opened 2022-01-21
Support interleaved requests in access logger

### LINKED ISSUE #6782: Fix access logger for interleaved requests
### Feature description

Whilst supporting filtering in the access logger, it was noticed that this probably fails or emits garbage when there are interleaved requests

https://github.com/micronaut-projects/micronaut-core/pull/6753

We need to investigate if this is the case, and if so provide a fix

yawkat:
`HttpAccessLogHandler` used a channel attribute to manage the `AccessLog` instance for a request. Because `AccessLog` does not support concurrent requests, this breaks on HTTP1 pipelining and on concurrent HTTP2 streams.

This patch multiplexes the requests onto multiple `AccessLog` instances.

Fixes #6782

---

I built this PR based on the branch of #6753 to avoid merge conflicts, that needs to be merged first. I'm making this a separate PR because it's a bug that's independent of the exclusion feature.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
@@ -183,7 +183,8 @@ public class HttpAccessLogHandler extends ChannelDuplexHandler {
             if (!createIfMissing) {
                 return null;
             }
-            attr.set(holder = new AccessLogHolder());
+            holder = new AccessLogHolder();
+            attr.set(holder);
         }
         return holder;
     }
@@ -237,7 +238,8 @@ public class HttpAccessLogHandler extends ChannelDuplexHandler {
                     currentPendingResponseStreamId++;
                 }
             } else {
-                currentPendingResponseStreamId = streamId = Long.parseLong(streamIdHeader);
+                streamId = Long.parseLong(streamIdHeader);
+                currentPendingResponseStreamId = streamId; // in case future HttpContent objects arrive without a stream_id header
             }
             if (finishResponse) {
                 AccessLog accessLog = liveAccessLogsByStreamId.remove(streamId);
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.http.server.netty.handler.accesslog;
 
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.http.server.netty.handler.accesslog.element.AccessLog;
 import io.micronaut.http.server.netty.handler.accesslog.element.AccessLogFormatParser;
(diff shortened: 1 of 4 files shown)


yawkat:
The GH changeset is ugly, will fix that once #6753 is in. The relevant commit is 4906ff2d5bb1396df33e06182e0ff749e0865202

yawkat:
Okay there was another issue: It seems like we don't routinely add the stream ID to `HttpContent` messages we write to the channel. I think this causes bugs with concurrent responses in HTTP2, but they're another issue entirely. I've changed the HttpAccessLogHandler to match the behavior of HttpToHttp2ConnectionHandler in this case, so the tests pass now. This can be merged in time for 3.3.0.

I will look at the HTTP2 concurrency issues further, but it won't make it into 3.3.0.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
```diff
@@ -156,32 +149,105 @@ private void log(ChannelHandlerContext ctx, Object msg, ChannelPromise promise,
...
-        final AccessLog accessLogger = ctx.channel().attr(ACCESS_LOGGER).get();
-        if (accessLogger != null) {
-            if (msg instanceof HttpResponse && processHttpResponse((HttpResponse) msg, accessLogger, ctx, promise)) {
-                return;
+    /**
+     * Holder for {@link AccessLog} instances. {@link AccessLog} can only handle one concurrent request at a time, this
+     * class multiplexes access where necessary.
+     */
+    private final class AccessLogHolder {
+        private final Map<Long, AccessLog> liveAccessLogsByStreamId = new HashMap<>();
```
graemerocher, line 196: can this be concurrently accessed?

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
```diff
@@ -156,32 +149,105 @@ private void log(ChannelHandlerContext ctx, Object msg, ChannelPromise promise,
...
+                } else if (msg instanceof ByteBuf) {
+                    accessLogger.onResponseWrite(((ByteBuf) msg).readableBytes());
+                }
+            }
+        }
+        super.write(ctx, msg, promise);
+    }
+
+    @Nullable
+    @Contract("_, true -> !null") // can only return null when createIfMissing is false
```
graemerocher: Is this necessary?

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
yawkat: no, the holder is local to the channel, and the handler is only called on the event loop

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/HttpAccessLogHandler.java
yawkat: it removes a nullability warning in intellij, i can remove it if it's a problem

jameskleeh (approved):

graemerocher (approved):

PR #6786, opened 2022-01-21
Document and test literal placeholder syntax with no expansion

### LINKED ISSUE #6783: Add option to escape `${looks-like-a-variable}` in application.yml
### Feature description

Right now there is no option to escape a value in an `application.yml` file that looks like a variable, so micronaut won't try to resolve it.

Below is my simplified use case. I want to write variables to an `.env` file with content from an  `application.yml`.

Example `application.yml`
```yml
my-section:
  var: ${looks-like-a-variable}
```

Example service (groovy) that writes the `.env` file.

```groovy
@Singleton
class Test {
    
    final protected String myVar
    
    Test(@Value('${my-section.var}') String myVar) {
        this.myVar = myVar
    }
    
    void writeShellScript() {
        new File(".env").text = """TEST=${myVar}
        """
    } 
}
```

timyates:
We didn't document how to add a literal ${value} to the configuration.

This adds the docs, plus a test to ensure it keeps working.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/DefaultPropertyPlaceholderResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/DefaultPropertyPlaceholderResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -109,8 +109,22 @@ public class DefaultPropertyPlaceholderResolver implements PropertyPlaceholderRe
         String value = str;
         int i = value.indexOf(PREFIX);
         while (i > -1) {
-            //the text before the prefix
-            if (i > 0) {
+            if (isDoubleEscaped(value, i)) {
+                RawSegment rawSegment = new RawSegment(value.substring(0, i - 1));
+                segments.add(rawSegment);
+                value = value.substring(i);
+                i = value.indexOf(PREFIX);
+                continue;
+            }
+            else if (isEscaped(value, i)) {
+                RawSegment rawSegment = new RawSegment(value.substring(0, i - 1));
+                segments.add(rawSegment);
+                segments.add(new RawSegment(PREFIX));
+                value = value.substring(i + PREFIX.length());
+                i = value.indexOf(PREFIX);
+                continue;
+            } else if (i > 0) {
+                //the text before the prefix
                 String rawSegment = value.substring(0, i);
                 segments.add(new RawSegment(rawSegment));
             }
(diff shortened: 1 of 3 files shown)


jameskleeh:
I think we should instead implement an actual escape mechanism instead of misusing the placeholder feature to generate a `$` to avoid the resolution of the string

timyates:
Something like `\${raw-value}`?  Or would you introduce a global function like `${dollar('raw-value')}`

jameskleeh:
@timyates It would probably be easy to implement `\${`. The logic could just check the previous character and if its a `\` remove the `\` character and move on

timyates:
👍  on it

timyates:
As someone may have a property that is (for example) a windows file path, I'll check for an escape of the escape

```
folder: some-name
path: \path\to\\${folder}\readme.txt
```

So that will resolve to `\path\to\some-name\readme.txt`

jameskleeh:
You could use `$` for the escape instead

timyates:
Pushed in 241c3b1 to show what it looks like with `$${value}` for escaping expansion (and resulting in `${value}`

And allowing `$$${value}` for people who used to use `$${value}` and expect `$result` to be the end result.

I think I covered it will with the test...

Avoided using `\` for escaping because of windows paths...

It may be simpler to consider an expansion method that wraps the argument in `${...}` I'll bring it up at our meeting tomorrow

graemerocher:
Shouldn't this target 3.3.x?

timyates:
@graemerocher TL;DR:  I will move this to 3.3.x, though it may require 4.0.x

---

Originally, this was to document the workaround of `${:$}{some-value}` to generate `${some-value}` and so was targetted to 3.2.x

However, we were not happy with exploiting the placeholder feature to do this, so the idea was to allow escaping with additional dollars (avoiding backslash as it could already be used for windows paths), so `$${some-value}` to render `${some-value}` (without lookup)

There are cases however where people may currently be using `$$` for example:

```
values:
  lookup: xxx
  result: $${lookup}
```

To generate `$xxx`

So I introduced double escaping to allow this with `result: $$${lookup}`

There is an open question as to whether we should just change to `${dollar('some-value')}` to do the escaping instead.  I was going to raise it in the engineering meeting.

timyates:
Moved to 3.3.x 👍

yawkat:
imo the workaround is fine, we don't need a dedicated escape mechanism. It would be a rarely used feature anyway, and the workaround is fairly easy to understand

timyates:
I'd vote for document the placeholder workaround for now, and raise an issue for 4.0 to see if it gets any votes before we get there

jameskleeh:
@timyates Alright lets just do that then. Can you revert this PR back to where it was just documented?

timyates:
@jameskleeh done, and raised https://github.com/micronaut-projects/micronaut-core/issues/6834

timyates:
Another workaround https://github.com/micronaut-projects/micronaut-core/issues/6783#issuecomment-1027843476 not sure if we prefer this one before this gets documented?

PR #6789, opened 2022-01-21
Updated VSCode instructions

### LINKED ISSUE #3269: VSCode setup instructions are Mac-specific
The instructions link to https://code.visualstudio.com/docs/setup/mac and talk about setting up a terminal integration. However, when I go to the VSCode Linux setup instructions the word "terminal" cannot be found, so it's unclear how I should proceed

timyates:
It's only for OS X you need to explicitly install the `code` command line tool, so I
moved the docs around this to a NOTE:

For Windows, this tool is automatically added to the %PATH% (so long as the installer
had a checkbox checked at installation time)

For linux, stackoverflow shows that some package managers add this code symlink automatically,
though there are some reports of a manual `ln -s` step being required.

Closes #3269


timyates:
Issue: https://github.com/micronaut-projects/micronaut-core/issues/3269

PR #6797, opened 2022-01-24
Merge field annotation metadata into setter argument annotation metadata

### LINKED ISSUE #4308: @MapFormat(keyFormat = RAW) not working when used with @ConfigurationProperties
### Steps to Reproduce

1. Simple configuration class (uses lombok)
```
@Data
@ConfigurationProperties("conf")
public class MyConfig {
    @MapFormat(keyFormat = RAW)
    private Map<String, String> animals;
}
```
2. Corresponding config in `application.yml`
```
conf:
  animals:
    VERY_FAST: rabbit
    VERY_SLOW: turtle
```

### Expected Behaviour
When accessing `MyConfig#getAnimals()`, the map should contain keys: `VERY_FAST` and `VERY_SLOW`

### Actual Behaviour
`MyConfig#getAnimals()` actually contains keys: `very-fast` and `very-slow`

### Environment Information

- **Operating System**: macOS 10.15.7
- **Micronaut Version**: 2.1.0 
- **JDK Version**: openjdk version "11.0.6" 2020-01-14 LTS

### Example Application

- https://github.com/john-pan9/micronaut-key-format-issue

mattmoss:
Fixes #4308

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
@@ -23,13 +23,13 @@ import io.micronaut.core.naming.NameUtils;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.inject.ast.ClassElement;
+import io.micronaut.inject.ast.ElementQuery;
 import io.micronaut.inject.ast.GenericPlaceholderElement;
 import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.ast.ParameterElement;
 import io.micronaut.inject.ast.PrimitiveElement;
 
 import javax.lang.model.element.Element;
-import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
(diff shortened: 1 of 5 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
```diff
@@ -270,4 +274,29 @@ private boolean isSuspend(VariableElement ve) {
...
+
+        final String expectedFieldName = NameUtils.getPropertyNameForSetter(methodName, writerPrefixes);
+
+        // Return the field corresponding to this writer.
+        return classElement.getEnclosedElements()
+                .stream()
+                .filter(element -> element.getKind() == ElementKind.FIELD)
+                .filter(field -> field.getSimpleName().contentEquals(expectedFieldName))
+                .findFirst()
+                .orElse(null);
```
graemerocher, line 298: This search doesn't make sense, should be use the `PropertyElement` API and `getBeanProperties`

mattmoss: @graemerocher , I don't follow. I don't see how to extract a field `javax.lang.model.element.Element` from the `PropertyElement`/`getBeanProperties` API. Maybe I'm looking in the wrong place...?

Issue #6791, opened 2022-01-24
`:core:test` spends most of the time initializing test classes


alextu:
### Issue description

`:core:test` task takes [about 1min](https://ge.micronaut.io/s/hhzofcpjbcmci/timeline?details=m6g5y4lib3e3u) but the executed tests themselves take [about 10s](https://ge.micronaut.io/s/hhzofcpjbcmci/tests/overview?task-path=:core:test&toggled=WyI6Y29yZTp0ZXN0LTE2LWlvLm1pY3JvbmF1dC5jb3JlLmFubm90YXRpb24uaW1tdXRhYmxlc29ydGVkc3RyaW5nc2FycmF5bWFwdGVzdC0yMCIsIjpjb3JlOnRlc3QtMTYiXQ#:core:test-16)

The bottleneck seems to be `ImmutableSortedStringsArrayMapTest` spending most of the time in class initialization, `Set<String> ABSENT = KEYS - PRESENT` seems to be the culprit spending most of the time in metaclass calls. Replacing it with 
```        
        KEYS.removeAll(PRESENT)
        ABSENT = KEYS
``` 
does the trick and init time is unoticeable. I can submit a PR.

yawkat:
Good catch, please make a PR

timyates:
Awesome, fixed by https://github.com/micronaut-projects/micronaut-core/pull/6799 👍

PR #6799, opened 2022-01-25
Fix slow test initialization ImmutableSortedStringsArrayMapTest (#6791)

alextu:
This also moves the init logic in Spock's `setupSpec()` hook, exposing only the 2 fields necessary for the tests
The actual speed up bit is there:

```
        keys.removeAll(present)
```


graemerocher:
Thanks for the contribution!

PR #6808, opened 2022-01-25
Allow discovery of silent conversion errors for config and QueryValue

### LINKED ISSUE #2068: Silently omitting wrong enum values in properies
Based on the documentation `4.4 Configuration Properties
` we can do Property conversion with Lists and Enums. However, if you define a list of enums and provide non-existing values for the list's values, they will get silently ignored and there is no way to force the application to crash.

Steps to reproduce:

1. Create a new app via mn (for example: `mn create-app property-validation`)
2. Define Enum class:
```
public enum DataEnum {
    DATA_1, DATA_2;
}
```
3. Define Configuration Properies class:
```
@Context
@ConfigurationProperties("custom")
public class DataConfigurationProperties {
    List<DataEnum> data;

    @PostConstruct
    public void onInit() {
        System.out.println("Data: " + data);
    }
}
```
4. Finally try to run the application with configation:
```
custom:
  data:
  - DATA_1
  - DATA_3
```

The application starts normally and the output is: `Data: [DATA_1]`. If you start it in log level `trace`, the relevant output is:
```
19:12:55.966 [main] TRACE i.m.c.e.PropertySourcePropertyResolver - Resolved value [[DATA_1]] for property: custom.data
Data: [DATA_1]
```

Now, this is a sound default behavior, however in my experience, most of the times if someone misspells configuration enum it is better to force the application to crash early in order to indicate that there is a problem with the configuration. From what I've seen, looking into the Micronaut's code, if the Enum conversion throws IllegalStateException, it gets [silently omitted](https://github.com/micronaut-projects/micronaut-core/blob/44ef42fbf0d6175d7a06992d9882c2f896d29c4f/core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java#L592) without throwing any exception. Could that be changed so that we can do something like annotating the property in a way which triggers stricter validation?

timyates:
previously if you had a list of values (especially enums) and one or more of them
were invalid, they were silently dropped, and the app started as per normal.

This change adds an ApplicationContextBuilder method failOnConfigConversionError which
if set to true when the app starts up will thrown an exception if one of these conversion
errors is seen.

This also adds a debug logger to request query resolution so calls with invalid data to @Nullable or Optional @QueryValues can be found by turning on the debug level logging for `io.micronaut.http.server.binding`

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/Micronaut.java
+++ b/context/src/main/java/io/micronaut/runtime/Micronaut.java
@@ -180,6 +180,11 @@ public class Micronaut extends DefaultApplicationContextBuilder implements Appli
         return (Micronaut) super.banner(isEnabled);
     }
 
+    @Override
+    public @NonNull Micronaut failOnConfigConversionError(boolean fail) {
+        return (Micronaut) super.failOnConfigConversionError(fail);
+    }
+
     /**
      * Add classes to be included in the initialization of the application.
      *
--- a/core/src/main/java/io/micronaut/core/convert/ArgumentConversionContext.java
+++ b/core/src/main/java/io/micronaut/core/convert/ArgumentConversionContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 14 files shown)


graemerocher:
We need a configuration option to enable failures to not be silent for request binders /cc @yawkat

PR #6809, opened 2022-01-25
remove tracing and add new micronaut-tracing versions and bom

burtbeckwith:


dstepanov:
I think it's because of the change in the dependencies in the test suite and the introduction of `libs.managed.micronaut.tracing.zipkin` that introduced `OpenTracingServerFilter` returning `TracingPublisher` which doesn't support propagation of the reactor context. 

The fix might be for it to implement `CorePublisher` and propagate the context.

graemerocher:
seems it is has identified a bug in `TracingPublisher`

graemerocher:
@jameskleeh perhaps this one can still make it into 3.3

PR #6821, opened 2022-01-26
Add capturedException on Retry in order to be able to retry also on Throwable

### LINKED ISSUE #6814: @Retryable on interface
### Expected Behavior

Since micronaut 3.0, `@Retryable` is not inherited following [this table](https://docs.micronaut.io/latest/guide/index.html#breaks) 

But on the docs [here](https://docs.micronaut.io/latest/guide/index.html#retry): 
> Note also that the @Retryable annotation can be applied on interfaces, and the behaviour is inherited through annotation metadata. The implication of this is that @Retryable can be used in combination with Introduction Advice such as the HTTP Client annotation.
This suggest that `@Retryable` is inherited.

What is the expected behaviour here ? Does the error is on the doc or the behaviour ? 

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.2.6

tchiotludo:
close #6814

### DIFF
--- a/runtime/src/main/java/io/micronaut/retry/RetryState.java
+++ b/runtime/src/main/java/io/micronaut/retry/RetryState.java
@@ -77,6 +77,11 @@ public interface RetryState {
     }
 
     /**
+     * @return The captured exception type (default to {@link RuntimeException}
+     */
+    Class<? extends Throwable> getCapturedException();
+
+    /**
      * Opens the retry state.
      */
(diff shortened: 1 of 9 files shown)


graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #6841, opened 2022-01-31
Fix AnnotationValue#getAnnotation

bene1618:
The comparison compared different types and thus always evaluates to "false"; as an effect, the `ErrorStrategy` from micronaut-kafka's `KafkaListener` ceased to work.


graemerocher:
Thanks, could you sign the CLA?

bene1618:
Yes, sure :)

jameskleeh:
Needs a test

bene1618:
Added a test for this case and for the case of more than one value

PR #6848, opened 2022-02-01
Add config option for strict request argument error checking

### LINKED ISSUE #5135: @Nullable @QueryValue is not reporting conversion errors
### Steps to Reproduce

1. Create a controller that accepts a simple query value UUID parameter
```
@Controller("/test")
public class TestController {

    @Get
    public String test(@Nullable @QueryValue UUID id) {
        return "Id is " + id;
    }
}
```
2. Call endpoint with an invalid UUID value
`curl localhost:8080/test?id=123456`

### Expected Behaviour

Caller should receive an invalid request response as the parameter passed was not valid

### Actual Behaviour

The handler method is called with the 'id' parameter of null and no error is reported to the caller

### Environment Information

- **Operating System**: MacOS 11.2.1
- **Micronaut Version:** 2.4.0
- **JDK Version:** OpenJDK 11.0.2

### Example Application

[https://github.com/chapmonkey/micronaut-issue](https://github.com/chapmonkey/micronaut-issue)

yawkat:
When the value of a request argument (e.g. a @QueryValue) is invalid, but the parameter is marked @Nullable, strict conversion checking will produce an error. Without this option enabled, the parameter will simply be set to null.

I've added a parameter to the `RequestArgumentSatisfier` constructor. The class is `@Internal` and a code search of the micronaut-projects org shows no subclasses outside micronaut-core, so this seems safe.

Fixes #5135


timyates:
Should we also be failing for Optionals?

Adding a test of

```
        HttpMethod.GET  | '/parameterNullable/optional-named?maximum=inv'         | null      | HttpStatus.BAD_REQUEST
```

and a Get handler of

```
        @Get("/optional-named")
        String optionalNamed(@QueryValue('maximum') Optional<Integer> max) {
            "Parameter Value: $max"
        }
```

Fails as max is set to `Optional.empty()`, and the request succeeds

timyates:
Also, given:

```
        HttpMethod.GET  | '/parameterNullable/named-list?max=inv&max=2'           | null                  | HttpStatus.BAD_REQUEST
```

With a handler of:

```
        @Get("/named-list")
        String namedList(@QueryValue('max') @Nullable List<Integer> maxList) {
            "Parameter Value: $maxList"
        }
```

We get an OK response with a result of `Parameter Value: [2]`

~~I _*think*_ we need to make this change https://github.com/micronaut-projects/micronaut-core/pull/6808/files#diff-fb96e6a08d20e10f1121573768dd86c523b72869333bf2278a20278a155b9c16 so the collection errors get passed up~~

*EDIT:* Nope...that's not enough 🤔  posted the fix (I think) in the review comment

http-server/src/main/java/io/micronaut/http/server/binding/RequestArgumentSatisfier.java
```diff
@@ -140,7 +144,7 @@ public RequestBinderRegistry getBinderRegistry() {
                     }
                 } else if (bindingResult.isPresentAndSatisfied()) {
```
timyates: Think we also need to check here for the collection of values case ie: `@QueryParam("item") List<Integer> items` called with `item=1&item=tim`

```
                } else if (bindingResult.isPresentAndSatisfied() && (!conversionContext.hasErrors() || !httpServerConfiguration.isStrictArgumentConversion())) {
```

timyates (changes requested):

timyates (approved):
LGTM 👍 

Just the question on whether this should fail for Optional conversion, but I'm not sure how that could be done 🤔

yawkat:
Optional should be easy, it's just the code path above the ones that already changed. Will discuss in the meeting

yawkat:
also decided to test making this a ApplicationContextBuilder-level setting instead, will test.

graemerocher:
This should go into 3.3.x IMO as although this is a new configuration option, the existing behaviour is broken and inconsistent. I understand the motivation to not break existing applications but it makes no sense that the following:

```java
@Get("/{num}")
String get(@Nullable Long num) {
    //
}
```

With a request of `/junk` results in an exception, whilst the following:

```java
@Get("/{?num}")
String get(@QueryValue @Nullable Long num) {

}
```

With a request of `/?num=junk` does not and ignores the error silently.

yawkat:
@jameskleeh one problem with doing this as a more general setting is that we have to decide *now* where to implement this strict conversion checking. We can't add the stricter checks to other places later, without breaking compatibility.

kevin-wise:
any idea when this might get merged?

graemerocher:
@yawkat can you rebase this and maybe we should make this the default for 4.x?

yawkat:
Yea I made this PR draft because I wanted to try to change it at a higher level (in ConversionService) but I dont think thats viable, so this PR should be fine once rebased.

sdelamo (approved):
I have added a test for the default and string checking false. I think we should change the default in Micronaut Framework 4.0 to string checking true.

inject/src/main/java/io/micronaut/context/ApplicationContextBuilder.java
```diff
@@ -224,6 +224,16 @@ public interface ApplicationContextBuilder {
      */
     @NonNull ApplicationContextBuilder allowEmptyProviders(boolean shouldAllow);
 
+    /**
+     * Enable strict conversion error checking.
+     * @param strict Whether to enable strict conversion error checking
+     * @return This application
+     * @since 3.9.0
+     */
+    @NonNull default ApplicationContextBuilder strictConversionChecking(boolean strict) {
```
graemerocher, line 233: why is something related to the request binding including in `ApplicationContextBuilder`? This should be in `HttpServerConfiguraiton` surely?

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/ApplicationContextBuilder.java
sdelamo: @yawkat can you address this change?

sdelamo:
@yawkat should we close this?

yawkat:
no it's still on my todo. just not this approach

PR #6850, opened 2022-02-02
Allow deserializing _embedded values of HATEOAS resources

### LINKED ISSUE #2863: hateoas.AbstractResource: embedded is not deserialized
Using `JsonError jsonError = e.getResponse().getBody(JsonError.class).get();` in unit test does not return `embedded` in `JsonError`. 

The server is returning this json : 
```json
{
  "message": "Invalid entity: id: must match \"[a-zA-Z0-9_-]+\"",
  "_links": {
    "self": {
      "href": "/api/v1/flows/org.kestra.tests/each-sequential-nested",
      "templated": false
    }
  },
  "_embedded": {
    "errors": {
      "message": "must match \"[a-zA-Z0-9_-]+\"",
      "path": "id"
    }
  }
}
```

If I used the links attributes, I've the link provide by server.
If I try with embedded, nothing is return and size 0.

Maybe need the same Jackson setter for links ? : 
 
https://github.com/micronaut-projects/micronaut-core/blob/d1c6e7c5d7b94798049fd0f420ed03c7abbc1d6d/http/src/main/java/io/micronaut/http/hateoas/AbstractResource.java#L134-L150

yawkat:
Adds a `GenericResource` class that acts as a deserialization target for the embedded resources. We can't use "real" resource classes because there's no way to tell which implementation is the right one.

It's possible that we could use jackson "deduction" subtype detection to support common implementations like `JsonError`, to give the right resource type in at least some cases. However this jackson feature isn't supported by serde, so I've not looked at this.

Fixes #2863


yawkat:
This needs a test for serde, but that test should be in the micronaut-serialization repo

graemerocher:
failing tests

yawkat:
found a workaround that doesnt rely on `@JsonDeserialize` :)

graemerocher (approved):

sdelamo (approved):

PR #6865, opened 2022-02-03
build: use micronaut rxjava2 bom

### LINKED ISSUE #6858: Update Library Catalogue to use rxjava2 bom included in 1.2.0
### Issue description

[Micronaut RxJava2 1.2.0](https://github.com/micronaut-projects/micronaut-rxjava2/releases/tag/v1.2.0) introduced a bom module.

timyates:
Closes #6858

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -114,7 +114,7 @@ managed-micronaut-reactor = "2.1.1"
 managed-micronaut-redis = "5.1.0"
 managed-micronaut-rss = "3.0.0"
 managed-micronaut-rxjava1 = "1.0.0"
-managed-micronaut-rxjava2 = "1.1.0"
+managed-micronaut-rxjava2 = "1.2.0"
 managed-micronaut-rxjava3 = "2.1.1"
 managed-micronaut-security = "3.3.0"
 managed-micronaut-serialization = "1.0.0-M6"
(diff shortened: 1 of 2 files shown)


dstepanov:
After this change I cannot use reference:
- `mn.micronaut.reactor`
- `mn.micronaut.rxjava2`
- `mn.reactor`
-  `mn.rxjava2`

Those references were used in the Micronaut Data

@melix Is there a different syntax?

dstepanov:
IMHO we should keep the reference without the version.

melix:
That's correct, it's the drawback of moving everything to BOMs. Possibly we could automate the creation of those "nested BOMs" in catalogs if the included BOMs are themselves published as catalogs (we could parse the TOML file and deduce what elements to add).

sdelamo:
> After this change I cannot use reference:
> 
> * `mn.micronaut.reactor`
> * `mn.micronaut.rxjava2`
> * `mn.reactor`
> * `mn.rxjava2`
> 
> Those references were used in the Micronaut Data
> 
> @melix Is there a different syntax?

can you not use? 

```
implementation(platform("io.micronaut.reactor:micronaut-reactor-bom"))
implementation("io.micronaut.reactor:micronaut-reactor")
```

melix:
> can you not use?

You can, but it defeats the concept of catalogs and type-safe accessors.

PR #6870, opened 2022-02-04
feat: adds LocalizedMessageSource

sdelamo:
- It makes MessageSource Ordered since CompositeMessageSource attempts to order the MessageSource beans.
- It shows the recommended approach of using a factory to create a MessageSource with a ResourceBundleMessageSource
- It adds extra default methods to MessageSource to make it easier to work with the API.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
+++ b/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
@@ -33,11 +33,15 @@ import java.util.Locale;
 public class HttpLocalizedMessageSource extends AbstractLocalizedMessageSource<HttpRequest<?>> implements RequestAware {
     private Locale locale;
     /**
+     * @param httpLocaleResolutionConfiguration The locale resolution configuration
      * @param localeResolver The locale resolver
      * @param messageSource  The message source
      */
-    public HttpLocalizedMessageSource(LocaleResolver<HttpRequest<?>> localeResolver, MessageSource messageSource) {
+    public HttpLocalizedMessageSource(HttpLocaleResolutionConfiguration httpLocaleResolutionConfiguration,
+                                      LocaleResolver<HttpRequest<?>> localeResolver,
+                                      MessageSource messageSource) {
         super(localeResolver, messageSource);
+        locale = httpLocaleResolutionConfiguration.getDefaultLocale();
     }
 
     @Override
--- a/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
+++ b/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
@@ -32,6 +32,7 @@ import java.util.Locale;
 @RequestScope
 public class HttpLocalizedMessageSource extends AbstractLocalizedMessageSource<HttpRequest<?>> implements RequestAware {
     private Locale locale;
+    
     /**
      * @param localeResolver The locale resolver
      * @param messageSource  The message source
@@ -43,6 +44,9 @@ public class HttpLocalizedMessageSource extends AbstractLocalizedMessageSource<H
     @Override
     @NonNull
     protected Locale getLocale() {
+        if (locale == null) {
+            throw new IllegalStateException("RequestAware::setRequest should have set the locale");
+        }
         return locale;
     }
 
--- a/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
+++ b/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
@@ -27,7 +27,7 @@ import java.util.Locale;
 /**
  * A {@link RequestScope} which uses the current {@link HttpRequest} to resolve the locale and hence return the localized messages.
  * @author Sergio del Amo
- * @since 3.4.x
+ * @since 3.4.0
  */
 @RequestScope
 public class HttpLocalizedMessageSource extends AbstractLocalizedMessageSource<HttpRequest<?>> implements RequestAware {
--- a/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
+++ b/http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
@@ -33,15 +33,11 @@ import java.util.Locale;
 public class HttpLocalizedMessageSource extends AbstractLocalizedMessageSource<HttpRequest<?>> implements RequestAware {
     private Locale locale;
     /**
-     * @param httpLocaleResolutionConfiguration The locale resolution configuration
      * @param localeResolver The locale resolver
      * @param messageSource  The message source
      */
-    public HttpLocalizedMessageSource(HttpLocaleResolutionConfiguration httpLocaleResolutionConfiguration,
-                                      LocaleResolver<HttpRequest<?>> localeResolver,
-                                      MessageSource messageSource) {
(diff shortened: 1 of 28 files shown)


http/build.gradle
```diff
@@ -7,12 +7,6 @@ internalSanityChecks {
     expectedServiceCount.put('io.micronaut.inject.BeanDefinitionReference', 21)
 }
 
-
-internalSanityChecks {
-    expectedServiceCount.put('io.micronaut.inject.BeanDefinitionReference', 21)
-    expectedServiceCount.put('io.micronaut.core.beans.BeanIntrospectionReference', 4)
-}           
-
```
jameskleeh, line 15: Why was this block removed?

timyates (approved):
LGTM 👍 

I [tweaked the whitespace](https://github.com/micronaut-projects/micronaut-core/pull/6870/commits/de7a93515a862191b6a3b8ff92a8caa92452894a) a bit in one of the tests, hope that's ok

http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param messageSource  The message source
+     */
+    public HttpLocalizedMessageSource(LocaleResolver<HttpRequest<?>> localeResolver, MessageSource messageSource) {
+        super(localeResolver, messageSource);
+    }
+
+    @Override
+    @NonNull
+    protected Locale getLocale() {
+        return locale;
```
graemerocher, line 50: The return value should never be `null` but the code as written it could be, probably need to be able to return a default locale if non is available

inject/src/main/java/io/micronaut/context/LocalizedMessageSource.java
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.context;
+
+import io.micronaut.core.annotation.NonNull;
+
+import java.util.Map;
+import java.util.Optional;
+
+public interface LocalizedMessageSource {
```
graemerocher, line 28: Needs type level javadoc. 

I am not sure I understand the need for this interface when we already have the `MessageSource` API which supports passing in a Locale. What is problematic about that API that requires a whole new interface?

inject/src/main/java/io/micronaut/context/MessageSourceUtils.java
```diff
@@ -0,0 +1,45 @@
...
+     * E.g. for "Sergio", "John" the map ["0" => "Sergio", "1" => "John"] is returned
+     * @param args variables
+     * @return The variables map.
+     */
+    @NonNull
+    public static Map<String, Object> variables(@NonNull Object... args) {
+        Map<String, Object> variables = new HashMap<>();
+        int count = 0;
+        for (Object value : args) {
+            variables.put("" + count, value);
```
jameskleeh: Why not use `String.valueOf`  ?

inject/src/main/java/io/micronaut/context/MessageSourceUtils.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.micronaut.core.annotation.NonNull;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Utility class used by {@link MessageSource} to create variables maps.
+ * @author Sergio del Amo
+ * @since 3.4.x
```
jameskleeh: Should be 3.4.0

jameskleeh (changes requested):

http/build.gradle
sdelamo: it was duplicated.

inject/src/main/java/io/micronaut/context/MessageSourceUtils.java
sdelamo: Fixed via https://github.com/micronaut-projects/micronaut-core/pull/6870/commits/f848583afa66effe2a2274bfaddfee7882d21e11

inject/src/main/java/io/micronaut/context/MessageSourceUtils.java
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-core/pull/6870/commits/e47ba8bcf985e69a8068d7a1c604d1711beff48d

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/LocalizedMessageSource.java
sdelamo: The motivation for this interface is to create a nicer API to work with. 

It is much nicer from a developer point of view to do in a controller: 

```java
@Controller 
class Books {
    private LocalizedMessageSource messageSource;
    public Books(LocalizedMessageSource messageSource) {
        this.messageSource = messageSource;
    }

   @GET
   String index() {
		return messageSource.get(“book.title”, “Books”);

   }
}
```

than:

```java
@Controller 
class Books {
    private MessageSource messageSource;
    private HttpLocaleResolver httpLocaleResolver;
    public Books(MessageSource messageSource,
                 HttpLocaleResolver httpLocaleResolver) {
        this.messageSource = messageSource;
        this.httpLocaleResolver = httpLocaleResolver;
    }

   @GET
   String index(HttpRequest<?> request) {
		return messageSource.get(“book.title”, “Books”, httpLocaleResolver.resolveOrDefault(request));

   }
}
```

Don’t you think?

http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
sdelamo: I injected `HttpLocaleResolutionConfiguration` to set a default locale in the constructor.

 As I understood by being `RequestAware` the `setRequest` method would be called when the object was created and hence the locale was set. 

But I think now it is safer

inject/src/main/java/io/micronaut/context/LocalizedMessageSource.java
sdelamo: > Needs type level javadoc.
Done

http-server/src/main/java/io/micronaut/http/server/util/locale/HttpLocalizedMessageSource.java
jameskleeh: We know the request will be set so I don't see the nullability as a concern. If it is null you could throw an exception as that should never happen and keep the contract as NonNull

sdelamo: I kept the contract as `@NonNull`. What I have done is set in the constructor as well.

jameskleeh: Right but that is wrong imo. We can't assume the default is coming from that configuration. We can only rely on the contract of the interface

sdelamo: What exception do you want me to throw if request is null?

jameskleeh: IllegalStateException

sdelamo: fixed via https://github.com/micronaut-projects/micronaut-core/pull/6870/commits/a7359f01a21447439d4d1469b9466ad2e603a290

jameskleeh (approved):

inject/src/main/java/io/micronaut/context/LocalizedMessageSource.java
sdelamo: @graemerocher are you ok with this?

PR #6877, opened 2022-02-07
Introduce `UnsafeBeanProperty` and fix performance of setter

dstepanov:


graemerocher:
Looks good, if you fixed the tests etc.

It will probably need to target 3.4.x as well I imagine

dstepanov:
I think it can be a part of 3.3 as there is only an experimental interface added. If not I can backport the perf. optimization.

graemerocher (approved):

core/src/main/java/io/micronaut/core/beans/UnsafeBeanProperty.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    T getUnsafe(@NonNull B bean);
+
+    /**
+     * Unsafe version of {@link #withValue(Object, Object)}.
+     *
+     * @param bean  The bean
+     * @param value The new value
+     * @return Either the existing instance or the property is mutable or a newly created instance via the copy constructor pattern.
+     * @since 2.3.0
```
jameskleeh: since is incorrect

jameskleeh (changes requested):

core/src/main/java/io/micronaut/core/beans/AbstractBeanProperty.java
```diff
@@ -152,6 +164,10 @@ public final void set(@NonNull B bean, @Nullable P value) {
         writeInternal(bean, value);
     }
 
+    @Override
+    public void setUnsafe(B bean, P value) {
```
graemerocher: Can we make this `final`?

core/src/main/java/io/micronaut/core/beans/AbstractBeanProperty.java
```diff
@@ -117,14 +119,24 @@ public final P get(@NonNull B bean) {
...
 
         if (!beanType.isInstance(bean)) {
             throw new IllegalArgumentException("Invalid bean [" + bean + "] for type: " + introspection.getBeanType());
         }
-        if (value == get(bean)) {
+        return withValueUnsafe(bean, value);
+    }
+
+    @Override
+    public B withValueUnsafe(B bean, P value) {
```
graemerocher: Can we make this `final`?

core/src/main/java/io/micronaut/core/beans/AbstractBeanProperty.java
```diff
@@ -117,14 +119,24 @@ public final P get(@NonNull B bean) {
         return readInternal(bean);
     }
 
+    @Override
+    public P getUnsafe(B bean) {
```
graemerocher: Can we make this `final`?

graemerocher (changes requested):

graemerocher (approved):

PR #6886, opened 2022-02-08
Write out status code and message from HttpStatusException

### LINKED ISSUE #4862: Rest call with Flowable return results in OK status and wrong response exception when Flowable emits error.
A `Flowable` that raises an exception is returning a different exception with a 200 status and a different message.

This problem was first exposed to me by the `DiffController` in `micronaut-starter` while investigating https://github.com/micronaut-projects/micronaut-starter/issues/619. I discovered that when the emitter of the call to `Flowable.create` returns an exception (e.g. `HttpStatusException(HttpStatus.BAD_REQUEST, e.getMessage())`) via `FlowableEmitter.onError` the client actually gets a response with `HttpStatusException` with a `HttpStatus.OK ` status and message 'Empty body' 

The following test illustrates it. It seems like the streaming stuff is sending the initial response immediately and not waiting for at least 1 chunk of data?

```groovy
@MicronautTest
class FlowableErrorResponseSpec extends Specification {
    @Inject
    FlowableClient flowableClient

    void 'test flowable error'() {
        when:
        flowableClient.flowableError()

        then:
        def e = thrown(HttpClientResponseException)
        verifyAll {
            e.status == HttpStatus.BAD_REQUEST      // actual is HttpStatus.OK
            e.message == 'expected flowable error'    // actual is "Empty body"
        }
    }

    @Client("/flowable")
    static interface FlowableClient {
        @Get(uri = "/error", consumes = MediaType.TEXT_PLAIN)
        Foo flowableError()
    }

    @Controller("/flowable")
    static class FlowableErrorController {

        Logger logger = LoggerFactory.getLogger(FlowableErrorController.class)

        @Get(uri = "/error", produces = MediaType.TEXT_PLAIN)
        Flowable<Foo> flowableError() {
            return getError().doOnError(throwable ->
                    logger.error("flowableError endpoint: {}", throwable.getMessage())
            )
        }

        private static Flowable<Foo> getError() {
            return Flowable.error(
                    new HttpStatusException(HttpStatus.BAD_REQUEST, "expected flowable error")
            )
        }
    }

    static class Foo {}
}
```

mattmoss:
Fixes #4862


http-netty/src/main/java/io/micronaut/http/netty/stream/HttpStreamsHandler.java
```diff
@@ -410,8 +412,15 @@ protected void error(Throwable error) {
...
-                        ctx.writeAndFlush(new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.INTERNAL_SERVER_ERROR))
-                           .addListener(ChannelFutureListener.CLOSE);
+                        if (error instanceof HttpStatusException) {
+                            final HttpStatus status = ((HttpStatusException) error).getStatus();
+                            final String message = error.getMessage();
+                            ctx.writeAndFlush(new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.valueOf(status.getCode(), message)))
+                                            .addListener(ChannelFutureListener.CLOSE);
+                        } else {
+                            ctx.writeAndFlush(new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.INTERNAL_SERVER_ERROR))
+                                    .addListener(ChannelFutureListener.CLOSE);
```
jameskleeh: Please refactor to not duplicate the logic for writing the response

jameskleeh (changes requested):

PR #6712, opened 2022-01-05
Avoid double-free in content branch of NettyHttpRequest.buildBody

yawkat:
When concatenating `receivedContent` data into a CompositeByteBuf, we call `addComponent` with the content buffers of the individual `receivedContent` items. `addComponent` takes "ownership" of the buffers, so we need to retain them once so that freeing the composite buffer does not free the `receivedContent` items.

PR #6891, opened 2022-02-09
Release unwrapped body in NettyHttpRequest

yawkat:
This is an attempt at a fix for the flaky FuzzyInputSpec. Can't reproduce that locally, but this code is obviously buggy, so this might help.

`body` is a memoized supplier, but `NettyHttpRequest.release` checks whether it's a reference counted object. The supplier is never reference counted, but the object it supplies might be. This patch stores the unwrapped value as a separate field so that it can be released properly.

The test for #6712 still passes, so this doesn't seem to regress the issue that the patch that caused this leak fixed.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
@@ -148,6 +148,7 @@ public class NettyHttpRequest<T> extends AbstractNettyHttpRequest<T> implements 
     private List<ByteBufHolder> receivedContent = new ArrayList<>();
     private Map<IdentityWrapper, HttpData> receivedData = new LinkedHashMap<>();
 
+    private T bodyUnwrapped;
     private Supplier<Optional<T>> body;
     private RouteMatch<?> matchedRoute;
     private boolean bodyRequired;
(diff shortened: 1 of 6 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -366,8 +371,8 @@ public void release() {
         getBody().ifPresent(releaseIfNecessary);
         receivedContent.forEach(releaseIfNecessary);
         receivedData.values().forEach(releaseIfNecessary);
-        if (this.body instanceof ReferenceCounted) {
-            ReferenceCounted referenceCounted = (ReferenceCounted) this.body;
+        if (bodyUnwrapped instanceof ReferenceCounted) {
```
jameskleeh, line 374: bodyUnwrapped would need to be set in the `setBody(T body)` method as well no?

jameskleeh (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: yes will fix

rbhambhra:
@[yawkat](https://github.com/yawkat)
I get the following netty error in version 3.3.1 (not in 3.3.0) and eventually a crash due to OOM

`2022-02-12T03:33:52.301 41624 ERROR [default-nioEventLoopGroup-1-7] [default-nioEventLoopGroup-1-7] [] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
`
Perhaps its fixed by your commit - will try it out when it is released

thanks
Rahul

yawkat:
Buffer leaks can happen due to many things, not necessarily related to this PR. They shouldn't cause an OOM though.

PR #6883, opened 2022-02-08
bugfix: when request HEAD only body needs to be null

jorgeaguileraseqera:
fix issue https://github.com/micronaut-projects/micronaut-core/issues/3685

PR #6894, opened 2022-02-09
Don't clear contentType for HEAD requests

mattmoss:
Replaces #6883


mattmoss:
@jameskleeh I looked for other locations where `contentType` was being set, and unless I missed something, they all:

- didn't affect HEAD requests
- handled error conditions where the endpoint didn't participate

If you think there are others, I'm open to discuss. I moved the original PRs test into HttpHeadSpec and added a couple additional tests.

mattmoss:
Will redo/augment on the original PR

PR #6895, opened 2022-02-10
Fix RandomAccessFile leak in NettySystemFileCustomizableResponseType

yawkat:
If `write` is never reached, the file is never closed. This patch does two things:
- Introduces netty leak detection for the `RandomAccessFile`. This is used for the test case.
- Delay the opening of the file until the write method. This fixes the leak itself. It's technically not allowed to do blocking IO on the netty event loop, and opening the file probably falls under that, but this approach isn't any worse than the old behavior because the constructor would also be called in the event loop.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
@@ -139,7 +139,7 @@ public class NettySystemFileCustomizableResponseType extends SystemFile implemen
      * Wrapper class around {@link RandomAccessFile} with two purposes: Leak detection, and implementation of
      * {@link ChannelFutureListener} that closes the file when called.
      */
-    private static class FileHolder implements ChannelFutureListener {
+    private static final class FileHolder implements ChannelFutureListener {
         private static final ResourceLeakDetector<RandomAccessFile> LEAK_DETECTOR =
                 ResourceLeakDetectorFactory.instance().newResourceLeakDetector(RandomAccessFile.class);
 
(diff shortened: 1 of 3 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
```diff
@@ -153,4 +135,51 @@ public void write(HttpRequest<?> request, MutableHttpResponse<?> response, Chann
         }
     }
 
+    /**
+     * Wrapper class around {@link RandomAccessFile} with two purposes: Leak detection, and implementation of
+     * {@link ChannelFutureListener} that closes the file when called.
+     */
+    private static class FileHolder implements ChannelFutureListener {
```
graemerocher: Declare `final`

http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
```diff
@@ -85,11 +78,6 @@ public NettySystemFileCustomizableResponseType(SystemFile delegate) {
         this.delegate = Optional.of(delegate);
     }
 
-    @Override
-    public long getLength() {
-        return rafLength;
-    }
-
```
graemerocher, line 92: This could be a breaking change, can we use `file.length()`?

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
yawkat: It should be identical unless the file changes between the constructor and the write call, I think. Unfortunately not possible to "freeze" this value without opening an FD

yawkat:
I think some of these CI failures would be fixed by #6892

graemerocher (approved):

yawkat:
rebased onto 3.3, hopefully tests are fixed now

graemerocher (approved):

graemerocher:
@jameskleeh would you mind reviewing this soon?

PR #6896, opened 2022-02-10
Enable leak detection for all tests, fix various leaks

### LINKED ISSUE #7098: Netty memory leak with server sent events
We have an error that is raised several time per day on several project : 
`LEAK: ByteBuf.release() was not called before it's garbage-collected`: 

The project are : 
- https://github.com/tchiotludo/akhq with [this issue](https://github.com/tchiotludo/akhq/issues/1037)
- https://github.com/kestra-io/kestra

I can't make it easily reproduce but I'm sure that this error is located on Server Sent Event.

The stacktrace : 
```
LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
  io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:401)
  io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
  io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:174)
  io.netty.buffer.AbstractByteBufAllocator.buffer(AbstractByteBufAllocator.java:108)
  io.micronaut.buffer.netty.NettyByteBufferFactory.buffer(NettyByteBufferFactory.java:77)
  io.micronaut.json.codec.MapperMediaTypeCodec.encode(MapperMediaTypeCodec.java:255)
  io.micronaut.http.server.codec.TextStreamCodec.encode(TextStreamCodec.java:158)
  io.micronaut.http.codec.MediaTypeCodec.encode(MediaTypeCodec.java:128)
  io.micronaut.http.server.netty.RoutingInBoundHandler.lambda$mapToHttpContent$7(RoutingInBoundHandler.java:1046)
  reactor.core.publisher.FluxMapFuseable$MapFuseableSubscriber.onNext(FluxMapFuseable.java:113)
  io.micronaut.reactive.reactor.instrument.ReactorSubscriber.onNext(ReactorSubscriber.java:57)
  reactor.core.publisher.FluxHide$SuppressFuseableSubscriber.onNext(FluxHide.java:137)
  reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.runAsync(FluxPublishOn.java:440)
  reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.run(FluxPublishOn.java:527)
  io.micronaut.reactive.reactor.instrument.ReactorInstrumentation.lambda$null$0(ReactorInstrumentation.java:62)
  reactor.core.scheduler.WorkerTask.call(WorkerTask.java:84)
  reactor.core.scheduler.WorkerTask.call(WorkerTask.java:37)
  io.micronaut.scheduling.instrument.InvocationInstrumenterWrappedCallable.call(InvocationInstrumenterWrappedCallable.java:53)
  io.micrometer.core.instrument.composite.CompositeTimer.recordCallable(CompositeTimer.java:68)
  io.micrometer.core.instrument.Timer.lambda$wrap$1(Timer.java:171)
  java.base/java.util.concurrent.FutureTask.run(Unknown Source)
  java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
  java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
  java.base/java.lang.Thread.run(Unknown Source)
```

Do you need more information or only the stacktrace will help you ?

### Version

3.3.4

yawkat:
This PR adds a spock extension to http-server-netty so that any buffer leaks detected in the test suite lead to a test failure. (note: since BufferLeakDetector is not 100% reliable, it is possible that a failure appears in the wrong test. There seems to be no good way to avoid this. However with netty 4.1.75, this code will print the spec name that caused the allocation along with the leak detection stack trace.)

The patch also fixes various leaks uncovered by this extension. In some cases these changes are pretty wide-spread, but they are not random: They follow the pattern that for every retain, there should be a clearly defined corresponding release. There are also some changes that ensure this release is always reached (eg in error conditions).

Unfortunately it was too difficult to split up this PR, because there are interdependencies that may lead to regressions if only parts are applied (see eg #6970).


yawkat:
All tests pass now, except the one fixed by https://github.com/micronaut-projects/micronaut-jackson-xml/commit/3e7f2ff8abf55633101c3caf5d74b1beff3e9875 which is still pending on 3.4.x.

The buffer leak detection code is still not ideal, have to wait for a release with https://github.com/netty/netty/commit/86603872776e3ff5a60dea3c104358e486eed588 to improve that, but that will take a while. Since all leaks are fixed, this isn't flaky anymore, so this can be merged

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -846,6 +863,18 @@ protected void doOnNext(Object message) {
                 @Override
                 protected void doOnError(Throwable t) {
                     s.cancel();
+                    // discard parameters that have already been bound
+                    for (Object toDiscard : routeMatch.getVariableValues().values()) {
+                        if (toDiscard instanceof ReferenceCounted) {
+                            ((ReferenceCounted) toDiscard).release();
+                        }
+                        if (toDiscard instanceof io.netty.util.ReferenceCounted) {
```
graemerocher, line 872: should this be `else if`? Not sure it makes sense to call `release()` multiple times?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -894,6 +923,7 @@ protected void doOnNext(Object message) {
                         ((NettyHttpRequest) request).setBody(message);
                         s.request(1);
                     }
+                    ReferenceCountUtil.release(message);
```
graemerocher, line 927: Release in `finally` block?

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormDataHttpContentProcessor.java
```diff
@@ -102,12 +115,28 @@ public void request(long n) {
             @Override
             public void cancel() {
                 subscription.cancel();
+                pleaseDestroy = true;
+                destroyIfRequested();
```
graemerocher, line 119: should this be within `finally`?

graemerocher (commented):
There are still failing tests?

jameskleeh:
Assuming the `.retain()` will be added to this PR that was reverted in 3.3.x, prior to merging this we need to understand exactly what was causing the leak in https://github.com/micronaut-projects/micronaut-core/issues/6942 and ensure it is not re-introduced with this PR

yawkat:
it's impossible to know without a test case. There were lots of leaks in that code.

jameskleeh:
@yawkat I think it should be possible to create a test case for that specific leak. There are only a couple paths where that line was executed. There may be other leaks but I'm concerned about the one we introduced in 3.3.1.

yawkat:
@jameskleeh The problem is not the retain itself, the retain has a corresponding release in `NettyHttpRequest.release`, but the question is why that's not enough. There are a few fixes in this PR that relate to that, e.g. making sure that `NettyHttpRequest.release` is called in all cases.

yawkat:
CLA check stuck :/

graemerocher:
Is this ready for review otherwise?

yawkat:
yes, all tests pass now, the only thing missing was the breakage from #6941 that triggered the leak detection

graemerocher:
could you write up a more detailed commentary on the changes in the comment here https://github.com/micronaut-projects/micronaut-core/pull/6896#issue-1130353555 so when we merge the changes are described more broadly

yawkat:
done

graemerocher (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/MaxRequestSizeSpec.groovy
```diff
@@ -320,11 +320,12 @@ class MaxRequestSizeSpec extends Specification {
         }
 
         cleanup:
+        responses.forEach(r -> r.release())
         channel.close()
         embeddedServer.close()
     }
 
-    @PendingFeature
+    @Ignore
```
sdelamo, line 328: This test is broken? or not yet implemented?

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/ByteBufferSpec.groovy
```diff
@@ -32,6 +33,7 @@ class ByteBufferSpec extends Specification {
 
     @Inject
     @Client("/")
+    @AutoCleanup
```
sdelamo, line 36: should we recommend always adding `@AutoCleanup` when injecting an `HttpClient` in tests to close it?  We are not doing that. E.g. tests here: https://guides.micronaut.io/latest/micronaut-http-client-gradle-groovy.html#tests

sdelamo (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/MaxRequestSizeSpec.groovy
yawkat: it has a buffer leak which causes the leak extension to fail even with `@PendingFeature`

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/ByteBufferSpec.groovy
yawkat: it isnt really a problem unless you do leak detection

sdelamo:
I have run the tests with the `BufferLeakDetectionExtension.java` introduced in this PR and the following failed: 
Screenshot 2022-03-24 at 10 40 15

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/ByteBufferSpec.groovy
sdelamo: but we should do it? it does not cost us anything to change the best practices in guides.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/MaxRequestSizeSpec.groovy
sdelamo: I see, `@PendingFeature` executes the tests and if the tests fails it considers it green, if it passes it fails the test to tell you that you can remove the annotation.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/ByteBufferSpec.groovy
yawkat: sure

graemerocher: `@AutoCleanup` is not needed here the client is shutdown when the context is stopped

PR #6916, opened 2022-02-14
Java time default formatters

### LINKED ISSUE #2972: Date formatting issues with @Introspected
https://github.com/micronaut-projects/micronaut-core/blob/ae5430bfec4cd2fe838e2df874509b0403158795/runtime/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java#L210-L214

We use micronaut-kafka to deserialize JSON messages to Java Objects. We are also (recently) using the micronaut-validation mechanism for bean validation on these Java Beans.

However since we annotated our beans with "@Introspected" it is not possible anymore to deserialize Json Objects contains LocalDateTime objects. The Strings in our JSON are formatted in ISO format (Like LocalDateTime.parse is able to parse).

Since we added @Introspected the whole code seems to take a different way through execution and goes through the above snippet which uses a totally different Formatter than the Jackson Formatter does that is used in Micronaut Kafka or in the kafka HTTP Server.

It seems to me, that the default format should be ISO, or otherwise when parsing Strings to LocalDateTime here:

https://github.com/micronaut-projects/micronaut-core/blob/ae5430bfec4cd2fe838e2df874509b0403158795/runtime/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java#L134-L136

the formatter should not be used.

Anyway with the @Introspected we get very different results than without it.

timyates:
Originally https://github.com/micronaut-projects/micronaut-core/pull/6576

> The Java Time classes have default formatters built in which use the ISO formats. So if in micronaut somewhere a string is to be converted to one of those classes we expect the default format to be the default format of the classes.
>
> Up to now micronaut used RFC_1123_DATE_TIME as the default format for converting but this is a rather human readable format and does not use the defaults that are already there.
>
> Another advantage of the built in defaults is, that they the classes do use the same format in .toString(). So i.e. OffsetDateTime.toString() is compatible with OffsetDateTime.parse() without thinking about the format.

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,7 +23,6 @@ import io.micronaut.core.convert.ConversionContext;
 import io.micronaut.core.convert.ConversionService;
 import io.micronaut.core.convert.TypeConverter;
 import io.micronaut.core.convert.TypeConverterRegistrar;
-import io.micronaut.core.convert.format.Format;
 import io.micronaut.core.util.StringUtils;
 import jakarta.inject.Singleton;
 
@@ -221,8 +220,8 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
     }
 
     private Optional<DateTimeFormatter> resolveFormatter(ConversionContext context) {
-        Optional<String> format = context.getAnnotationMetadata().stringValue(Format.class);
-        return format
-            .map(pattern -> DateTimeFormatter.ofPattern(pattern, context.getLocale()));
+        return context.getFormat().map(pattern -> ConversionContext.RFC_1123_FORMAT.equals(pattern) ?
+                DateTimeFormatter.RFC_1123_DATE_TIME :
+                DateTimeFormatter.ofPattern(pattern, context.getLocale()));
     }
 }
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -145,7 +145,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 try {
                     return Optional.of(resolveFormatter(context)
                             .map(formatter -> LocalDateTime.parse(object, formatter))
-                            .orElse(LocalDateTime.parse(object)));
+                            .orElseGet(() -> LocalDateTime.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
@@ -158,7 +158,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
             try {
                 return Optional.of(resolveFormatter(context)
                         .map(formatter -> formatter.format(object))
-                        .orElse(object.toString()));
+                        .orElseGet(() -> object.toString()));
             } catch (DateTimeParseException e) {
                 context.reject(object, e);
                 return Optional.empty();
@@ -178,7 +178,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 try {
                     return Optional.of(resolveFormatter(context)
                             .map(formatter -> LocalDate.parse(object, formatter))
-                            .orElse(LocalDate.parse(object)));
+                            .orElseGet(() -> LocalDate.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
@@ -195,7 +195,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 try {
                     return Optional.of(resolveFormatter(context)
                             .map(formatter -> ZonedDateTime.parse(object, formatter))
-                            .orElse(ZonedDateTime.parse(object)));
+                            .orElseGet(() -> ZonedDateTime.parse(object)));
                 } catch (DateTimeParseException e) {
                     context.reject(object, e);
                     return Optional.empty();
@@ -211,7 +211,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                     try {
                         return Optional.of(resolveFormatter(context)
                                 .map(formatter -> OffsetDateTime.parse(object, formatter))
-                                .orElse(OffsetDateTime.parse(object)));
+                                .orElseGet(() -> OffsetDateTime.parse(object)));
                     } catch (DateTimeParseException e) {
                         context.reject(object, e);
                         return Optional.empty();
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -186,6 +186,22 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                 temporalConverter
         );
 
+        // ZonedDateTime -> CharSequence
+        conversionService.addConverter(
+                ZonedDateTime.class,
+                CharSequence.class,
+                (object, targetType, context) -> {
+                    try {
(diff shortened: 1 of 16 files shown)


core/src/main/java/io/micronaut/core/convert/ConversionContext.java
```diff
@@ -229,4 +235,24 @@ static ConversionContext of(Map<String, Argument<?>> typeVariables) {
...
         return new DefaultArgumentConversionContext<>(argument, finalLocale, finalCharset);
     }
+
+    /**
+     * Gets the default format for the current conversion.  Can be overridden if required (for example Headers have a specific
+     * timestamp format no matter what the attribute type).
+     *
+     * @return the default format for this context
+     */
+    default Optional<String> defaultFormat() {
```
jameskleeh: Should be `getDefaultFormat`

timyates: 👍

core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java
```diff
@@ -968,7 +968,7 @@ protected void registerDefaultConverters() {
 
     private SimpleDateFormat resolveFormat(ConversionContext context) {
         AnnotationMetadata annotationMetadata = context.getAnnotationMetadata();
-        Optional<String> format = annotationMetadata.stringValue(Format.class);
+        Optional<String> format = context.getFormat();
```
jameskleeh, line 980: I think we should take this opportunity to also review this default format. We should probably change it to whatever Date.toString returns. I looked into it and it seems like that format is `EEE MMM dd HH:mm:ss zzz yyyy`. We also need to support the RFC format here as well since `@Header` can bind to a `java.util.Date`. We should add tests for that

jameskleeh (changes requested):

core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java
timyates: Changed it as suggested, and added a test for the different Temporal classes

I added `OffsetTime`, `LocalTime` and `Instant` to `TimeConverterRegistrar`

There are still a couple we don't have converter for in there, but I'm not sure how pressing `Year` and `YearMonth` are 🤔

PTAL

core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java
```diff
@@ -68,8 +68,11 @@
  */
 public class DefaultConversionService implements ConversionService<DefaultConversionService> {
 
+    public static final SimpleDateFormat RFC_1123_SIMPLE_DATE_FORMAT = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss 'GMT'");
```
jameskleeh: I don't think this is right

Where `instant` and `date` point to the same point in time:

```
System.out.println(DateTimeFormatter.RFC_1123_DATE_TIME.format(ZonedDateTime.ofInstant(instant, ZoneId.of("GMT"))))
Thu, 17 Feb 2022 17:30:46 GMT


System.out.println(new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss 'GMT'").format(start))
Thu, 17 Feb 2022 12:30:46 GMT
```

jameskleeh: Having a static string `GMT` there is wrong because the date is still formatted with the default time zone

timyates: But this is only used for Header requests where it is in the spec that it must be GMT

timyates: Added a test for a Date query param with a timezone

jameskleeh: @timyates Right it must be GMT but putting `GMT` in the format string doesn't make it GMT

jameskleeh:
```
System.out.println(DateTimeFormatter.RFC_1123_DATE_TIME.format(ZonedDateTime.ofInstant(instant, ZoneId.of("GMT"))))
Thu, 17 Feb 2022 18:24:37 GMT


Defined field SimpleDateFormat df = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss z");
df.setTimeZone(TimeZone.getTimeZone("GMT"))

System.out.println(df.format(start))
Thu, 17 Feb 2022 18:24:37 GMT
```

timyates:
```
DefaultConversionService.RFC_1123_SIMPLE_DATE_FORMAT.parse("Thu, 17 Feb 2022 18:24:37 PST")
java.text.ParseException: Unparseable date: "Thu, 17 Feb 2022 18:24:37 PST"
```

So a header coming in with an invalid format will be unparseable...  Which is ok isn't it?

jameskleeh: Hard to say. Lots of folks do things they shouldn't and expect us to support them

timyates: You're right of course 😀  You'd hope these were just coming from spec compliant browsers, but could be coming from anywhere I guess...

We could leave it like this, and let people define a `@Format` when they don't follow the spec

Or change it to accept a timezone with `zzz` 🤔

jameskleeh: The parse works with the format I linked above:

```
SimpleDateFormat df = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss z");
df.setTimeZone(TimeZone.getTimeZone("GMT"));
System.out.println(df.parse("Thu, 17 Feb 2022 18:24:37 PST"));

Thu Feb 17 21:24:37 EST 2022
```

jameskleeh:
We also need tests for declarative clients that send date headers

core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java
timyates: Cool, updated 👍

I remembered SimpleDateFormat isn't threadsafe as well, so stopped using it as a static

timyates: Now lets see if Github finds errors in my GMT based testing 😉

timyates:
Added some tests to a declarative client with a header (once I eventually worked out how to make one) ☺️

jameskleeh:
@timyates This needs documentation, otherwise good to go imo

Edit: I think you misunderstood the declarative client tests. The tests should be sending headers via arguments to the client methods. For example `@Header OffsetDateTime date`. None of the declarative client methods take arguments

timyates:
@jameskleeh Ahhh, so I moved the header into the client:

```
        @Get('/header-localdatetime')
        String headerLocalDateTime(@Header('X-Test') LocalDateTime date)
```

And you're right...  It seems to be sending it through in the wrong format 🤔 Good catch, I'm looking in to it

timyates:
Right, I think that handles the declarative client conversions...  Need to make sure it works on a non-GMT machine though...  The perils of living on GMT

timyates:
@jameskleeh Probably worth you running `DateTimeConversionSpec` locally to check the conversion to GMT works for you

jameskleeh (approved):

timyates:
That's awesome 😍

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -208,21 +244,63 @@ public void register(ConversionService<?> conversionService) {
...
+                                      Function<T, Y> defaultFormat) {
+        return Optional.of(context.getFormat().map(pattern -> {
+            if (ConversionContext.RFC_1123_FORMAT.equals(pattern)) {
+                T converted = value;
+                if (converted instanceof TemporalAccessor) {
+                    converted = (T) toGmt((TemporalAccessor) converted);
+                }
+                return formatDate.apply(converted, DateTimeFormatter.RFC_1123_DATE_TIME);
+            } else {
+                return formatDate.apply(value, DateTimeFormatter.ofPattern(pattern, context.getLocale()));
```
altro3, line 288: Is it normal to recreate formatter for every field parse? Maybe it's better to save the already created formatters in some kind of map?

altro3: @graemerocher ?

PR #6923, opened 2022-02-16
NettyHttpServer pipeline setup refactor

yawkat:
This PR refactors the NettyHttpServer to have a more straight-forward pipeline setup in a separate class (HttpPipelineBuilder).

This is part of the #6842 PR, cherry-picked so that the pipeline stays the same. This removes the compatibility issues, so it can be merged earlier. It will also make the diff for #6842 easier to review.


jameskleeh:
Why do you need to change the tests to specify `'micronaut.server.ssl.port': -1,`?

yawkat:
@jameskleeh apparently it binds to 8443 by default, so it breaks when doing parallel tests. It's not related to the refactor

jameskleeh (approved):

PR #6924, opened 2022-02-16
Set an identifier on collections of BeanRegistrations

### LINKED ISSUE #6878: Injected BeanRegistration objects have a null Identifier
### Expected Behavior

I want to have multiple beans of the same interface type injected, to select which ones to use based on their "identifiers", e.g., a qualifier specified with the `@Named` annotation.

Micronaut currently doesn't offer a map-based injection, but I **can** have a Collection of `BeanRegistration`s injected (yet undocumented, see #6872).

As the `BeanRegistration` class provides a `getIdentifier()` method returning a usable bean identifier, I expect this method to return a non-null identifier, derived e.g. from a used `jakarta.inject.Named` annotation on the bean's class.

### Actual Behaviour

I get a collection of `BeanRegistration`s injected, with correct `BeanDefinition`s, but all `BeanIdentifier`s of the registrations are `null`, although `jakarta.inject.Named` annotations are present.

### Steps To Reproduce

1. Create a new project using e.g. Micronaut Launch, with default settings and Java 11 and Kotlin language.
2. Create a new bean class `MyBean`, annotate it with `@Named("someBean")` and `@Singleton`
3. Create a Controller class, handling `/` requests and getting the BeanRegistrations of `MyBean` injected
4. Let the controller e.g. return a special string if the `identifier` of the bean registration is `null` (see example project)
5. Run the app and invoke the controller's endpoint.

### Environment Information

- JDK 11.0.4
- OS: Windows 10
- Maven

### Example Application

https://gist.github.com/albrechtflo-hg/9a56bf0ac3cd59e43db082ce3ad3ace5

### Version

3.3.0

timyates:
Closes #6878 

This does not fix the case where someone is injecting a single `BeanRegistration<MyClass>`, but I'm not sure what the use case of doing that would be.

All the tests pass locally, but I am not sure of the wider damage that this change might inflict.

I've targeted 3.3.x as it's not an API change

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -3739,7 +3739,7 @@ public class DefaultBeanContext implements InitializableBeanContext {
                     ));
                 }
             } else {
-                beansOfTypeList.add(new BeanRegistration(null, candidate, bean));
+                beansOfTypeList.add(new BeanRegistration(new BeanKey(beanType, candidate.getDeclaredQualifier()), candidate, bean));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

timyates:
The failure is a flakey fuzzing test, and we're not up to the "flaky friendly" junit report action on 3.3.x, I think we should be ok

PR #6928, opened 2022-02-16
Add @Inherited to @Version annotation

sascha-frinken:
This adds the `@Inherited` to the `@Version` annotation. 
Without that the  `@Version` annotation had to be duplicated instead of using the one defined on an abstract Controller class.


graemerocher:
Needs a test

sdelamo (commented):
Is not this a breaking change?

Moreover, we will need to update the table "Annotation inheritance in Micronaut 3 and above".

PR #6930, opened 2022-02-17
Make leak detection more reliable

yawkat:
Adapted from https://github.com/netty/netty/pull/12091 , let's see if CI is more consistent with this.


yawkat:
apparently not

yawkat:
should still be merged, but to actually fix this i think i'll need https://github.com/netty/netty/pull/12091

PR #6931, opened 2022-02-17
Remove unused and deprecated import in documentation

sascha-frinken:
`io.micronaut.http.filter.OncePerRequestHttpServerFilter` is deprecated therefore there shouldn't be any imports in the documentation samples.


jameskleeh:
Thanks!

PR #6940, opened 2022-02-19
Remove redudant boxing in TimeConverterRegistrar

SplotyCode:
While modern hotspot will optimise this on the fly there is no reason for the integer boxing


jameskleeh (approved):
Note that `parseInt` does behave slightly differently, however for this use case I don't think its a concern

SplotyCode:
To be honest I hear that for the first time. Do you mean parse wise? I'm just curious

jameskleeh:
@SplotyCode `valueOf` can take a char and return the unicode value. Like I said though I don't think that functionality would be used for this case

PR #6939, opened 2022-02-18
Support @Scheduled cron timezones

MT-Jacobs:
Fills my need in https://github.com/micronaut-projects/micronaut-core/issues/6938.

Essentially - I want to (optionally) support using a non-default time zone for `@Scheduled(cron = ...)` tasks.

Is this enough testing? Can anyone think of a clean way to test that the time zone supplied is _actually_ being used over UTC?

Issue #6938, opened 2022-02-18
Add time zone support for `@Scheduled` annotation using cron


MT-Jacobs:
### Feature description

Here's a hypothetical cron task - 

``` kotlin
@Scheduled(cron = "0 4 * * *")
fun doImportantTimeBasedThing() {
    TODO()
}
```

Next comes the question - what time zone does this use? Unfortunately, it appears to use the default time zone for the application - meaning this will run at one time on a local machine and run at a different time on a cloud server that uses UTC.

Much like how [Spring supports cron time zone overrides](https://www.baeldung.com/spring-scheduled-tasks#schedule-a-task-using-cron-expressions), it would be very nice if Micronaut were to support the same.

MT-Jacobs:
Done in https://github.com/micronaut-projects/micronaut-core/pull/6945 - it's part of the 3.5.0 release!

PR #6945, opened 2022-02-23
Support @Scheduled cron timezones

MT-Jacobs:
This is a repeat of https://github.com/micronaut-projects/micronaut-core/pull/6939 but starting from 3.4.x as the base to avoid merge conflicts.

Original PR text is repeated:

---

Fills my need in https://github.com/micronaut-projects/micronaut-core/issues/6938.

Essentially - I want to (optionally) support using a non-default time zone for `@Scheduled(cron = ...)` tasks.

Is this enough testing? Can anyone think of a clean way to test that the time zone supplied is actually being used over UTC?


context/src/main/java/io/micronaut/scheduling/processor/ScheduledMethodProcessor.java
```diff
@@ -141,13 +142,16 @@ public void process(BeanDefinition<?> beanDefinition, ExecutableMethod<?, ?> met
...
             String fixedDelay = scheduledAnnotation.get(MEMBER_FIXED_DELAY, String.class).orElse(null);
 
             if (StringUtils.isNotEmpty(cronExpr)) {
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("Scheduling cron task [{}] for method: {}", cronExpr, method);
                 }
-                taskScheduler.schedule(cronExpr, task);
+
+                ScheduledFuture<?> scheduledFuture = taskScheduler.schedule(cronExpr, zoneIdStr, task);
+                scheduledTasks.add(scheduledFuture);
```
MT-Jacobs, line 155: I neglected to point this out last time.

This line, `scheduledTasks.add(scheduledFuture);` is a notable difference from the original code. I noticed that this is the only `if` block that doesn't add the `scheduledFuture` to the list of `scheduledTasks`.

`scheduledTasks` are used in `close` to cancel in-flight tasks, so I believe this fixes a bug.

context/src/main/java/io/micronaut/scheduling/ScheduledExecutorTaskScheduler.java
```diff
@@ -78,6 +80,39 @@ public ScheduledExecutorTaskScheduler(@Named(TaskExecutors.SCHEDULED) ExecutorSe
         return new ReschedulingTask<>(command, this, delaySupplier);
     }
 
+    @Override
+    public ScheduledFuture<?> schedule(String cron, String zoneId, Runnable command) {
```
MT-Jacobs: New methods are required. Per the comment in https://github.com/micronaut-projects/micronaut-core/pull/6939/files#r811273670 (where I attempted to update the original methods) -

> This is a public method and thus cannot be changed. A new method will have to be created instead

MT-Jacobs:
https://github.com/micronaut-projects/micronaut-core/runs/5298008459?check_suite_focus=true#step:8:6224

Seeing failures in `http-server-netty`. I'm not sure that my changes are the cause here - that module's build.gradle doesn't seem to use `context`.

context/src/main/java/io/micronaut/scheduling/ScheduledExecutorTaskScheduler.java
jameskleeh: The old methods should just call this new one to avoid duplication

context/src/main/java/io/micronaut/scheduling/TaskScheduler.java
```diff
@@ -59,6 +59,39 @@
...
+     * @param zoneId The zoneId to base the cron expression on. Defaults to system time zone
+     * @param command  the task to execute
+     * @return a ScheduledFuture representing pending completion of
+     * the task and whose {@code get()} method will return
+     * {@code null} upon completion
+     * @throws java.util.concurrent.RejectedExecutionException if the task cannot be
+     *                                                         scheduled for execution
+     * @throws NullPointerException                            if command or delay is null
+     */
+    ScheduledFuture<?> schedule(String cron, String zoneId, Runnable command);
```
jameskleeh: These new methods need to be default methods. We cannot add abstract methods to a public interface. The default implementation can just call the original method without the zoneId

jameskleeh (changes requested):

context/src/main/java/io/micronaut/scheduling/ScheduledExecutorTaskScheduler.java
MT-Jacobs: Is this a reversal? In any event, happy to do it. Sorry on the delay.

MT-Jacobs:
@jameskleeh I attempted to make your requested changes, though I got a bit confused with some of the instructions.

In the original PR, I was asked to not change the public methods, and in this PR I was asked to point the old public methods to a new method. I went with the latest instruction and had the old public methods point to the new method.

Also, I was asked to not add abstract methods to a public interface - instead, I should call the original method without a zone id. I don't totally understand that because it sounds like I would be discarding the zone id in that case. In the end I converted the method that simply converts a Runnable to a Callable into a default method.

Let me know if I got anything wrong here and I'll get it cleaned up!

MT-Jacobs:
@sdelamo originally I was asked to point this to the 3.4.x branch. With you adding a 3.5.0 milestone tag to this, should I repoint this PR to 3.5.x?

sdelamo:
> @sdelamo originally I was asked to point this to the 3.4.x branch. With you adding a 3.5.0 milestone tag to this, should I repoint this PR to 3.5.x?

Yes, we have not created a 3.5.x branch yet. I will do that once 3.4.x is out. We can change the PR base then.

MT-Jacobs:
Hi, is there anything else I need to do here for this PR? Am I still waiting on 3.5.x?

context/src/test/groovy/io/micronaut/scheduling/ScheduledFixedRateSpec.groovy
```diff
@@ -79,8 +80,11 @@ class ScheduledFixedRateSpec extends Specification {
         MyOtherTask myTask = beanContext.getBean(MyOtherTask)
 
         then:
+        // Cron tasks run once every (30/31/32) seconds, starting with the application startup.
+        // A timeout of 20 seconds ensures that each task triggers once.
         conditions.eventually {
-            myTask.cronEvents.get() == 2
```
MT-Jacobs, line 83: This test has a bug. The timeout is 20 seconds, but the cron tasks run once every 3 seconds and once every 4 seconds, respectively. It is possible for the `cronEvents` counter to jump from 1 to 3 between the polling conditions.

I refactored it to ensure that each of these cron tasks trigger exactly once within the time window.

MT-Jacobs: Side note - if one of the two original scheduled cron tasks broke, this test wouldn't catch that. Instead the one working test would increment the counter once every 3 or 4 seconds, causing this original condition to resolve as true.

context/src/test/groovy/io/micronaut/scheduling/ScheduledFixedRateSpec.groovy
```diff
@@ -131,13 +134,15 @@ class ScheduledFixedRateSpec extends Specification {
     }
 
     @Singleton
+    @Requires(property = 'spec.name', value = 'ScheduledFixedRateSpecMyOtherTask')
```
MT-Jacobs, line 137: Without this annotation, this Singleton is active for other tests that use `scheduled-test.task.enabled` too!

MT-Jacobs:
Okay! The tests related to this change didn't fail, it's an issue with `japiCmp`.

```
Execution failed for task ':context:japiCmp'.
619 actionable tasks: 163 executed, 200 from cache, 256 up-to-date
> A failure occurred while executing me.champeau.gradle.japicmp.JApiCmpWorkAction
   > Detected binary changes.
         - current: context-3.5.0-SNAPSHOT-all.jar
         - baseline: micronaut-context-3.4.1.jar.
     
     See failure report at file:///home/runner/work/micronaut-core/micronaut-core/context/build/reports/binary-compatibility-context.html
```

Everything else looks good. We're pointing to 3.5.X and I believe I addressed the change request (though parts of it seemed contradictory but I may have misunderstood).

@jameskleeh is there anything else needed here?

sdelamo (approved):

sdelamo:
> Okay! The tests related to this change didn't fail, it's an issue with `japiCmp`.
> 
> ```
> Execution failed for task ':context:japiCmp'.
> 619 actionable tasks: 163 executed, 200 from cache, 256 up-to-date
> > A failure occurred while executing me.champeau.gradle.japicmp.JApiCmpWorkAction
>    > Detected binary changes.
>          - current: context-3.5.0-SNAPSHOT-all.jar
>          - baseline: micronaut-context-3.4.1.jar.
>      
>      See failure report at file:///home/runner/work/micronaut-core/micronaut-core/context/build/reports/binary-compatibility-context.html
> ```
> 
> Everything else looks good. We're pointing to 3.5.X and I believe I addressed the change request (though parts of it seemed contradictory but I may have misunderstood).

I have changed both new methods to be default and accepted the api breaking change. 

We have now a gradle task to detect binary compatibility changes. 

Thanks a lot for the contribution @MT-Jacobs

dstepanov (approved):

PR #6948, opened 2022-02-23
Prevent class annotations being stored in method annotation metadata + fix removing annotations

dstepanov:
The `buildForParent` was being called with the class reference coping the annotations into the method metadata. Also, the same method was recreating removed annotations, it can be skipped for an empty parent collection but otherwise, it's still incorrect. Not sure how to test all of it.

Discovered this when one of the big test repositories in Micronaut Data started to fail to compile because of the size of the methods definition constructor.

Another thing I see is that return type class annotations are being included in the return argument metadata :(


graemerocher:
This could break existing behaviour I fear. For example what happens if you have:

```
@Transactional
class Foo {
  void someMethod() {}
}
```

And you want to retrieve the transactional from the type in an interceptor?

dstepanov:
You can see in the commit that `AnnotationMetadataHierarchy` is there and it's based on the method and the class metadata, but before the method metadata also redundantly included the class metadata.
The generated code is:
```
new AnnotationMetadataHierarchy(new AnnotationMetadata[]{Reference.$ANNOTATION_METADATA, new DefaultAnnotationMetadata...
```

graemerocher (approved):

PR #6949, opened 2022-02-23
Do not request demand for completed HandlerPublisher

yawkat:
When a HandlerPublisher is removed, it is completed. If at that point it is in DEMANDING state, this will trigger a `flushBuffer()`, which in turn can trigger a `requestDemand()` and a `context.read()`. This read will happen before the removal of the HandlerPublisher has finished, which can lead to a race condition upstream in HttpStreamsHandler: currentlyStreamedMessage will be set to null *after* the removal of the publisher, at which point a new publisher has already been created from the `context.read()`.

This patch fixes this issue in two ways: the HandlerPublisher won't trigger a read anymore if it's already completed, and the HttpStreamsHandler will unset the currentlyStreamedMessage *before* triggering downstream handler removal.

The test case for this also includes a new `EmbeddedChannel` approach to testing the server, which removes the need for a dedicated event loop, and thus removes any race conditions (as long as the IO pool isn't used). The basic `EmbeddedChannel` factory method is placed in `NettyHttpServer` so that I can use it in fuzzing.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/reactive/HandlerPublisher.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/reactive/HandlerPublisher.java
@@ -160,9 +160,7 @@ public class HandlerPublisher<T> extends ChannelDuplexHandler implements HotObse
             LOG.trace("Demand received for next message (state = " + state + "). Calling context.read()");
         }
 
-        if (!completed.get()) {
-            ctx.read();
-        }
+        ctx.read();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


yawkat:
for some reason this made `RequestScopeSpec.test request scope no request` flaky? super weird, that method doesn't even make any requests, i guess somewhere the set(null) isn't being performed properly

yawkat:
oh it's the new test that's not closing correctly. guess that's another bug.

yawkat:
ok it was just because of the EmbeddedChannel approach to testing

graemerocher (approved):

PR #6951, opened 2022-02-23
Do not delete temp file before writing to it

yawkat:
This patch changes NettyStreamingFileUpload to not delete the temp file it creates before writing to it. It is not necessary to delete the file, and deleting it weakens the security, because another user may recreate the temp file with lax permissions between the deletion and transferTo. This could potentially allow that process to read the file contents.

This temp file code path is somewhat pointless, though. The temp file name is not predictable, and the path is exposed to the API user nowhere, so there is actually no way for the user to get at the file if they call `transferTo(String)` without configuring the `micronaut.server.multipart.location`. We should just throw an error instead, but that's a breaking change.

Because there's no way to get at the file, the test case also doesn't actually verify that the file has the right permissions, even though it really should. I checked manually and confirmed that on linux, `createTempFile` sets the permissions to 600 by default.

I added a test to verify that the temp file transferTo works at all (doesn't error), but the content can't be verified.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyStreamingFileUpload.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyStreamingFileUpload.java
@@ -134,16 +134,11 @@ public class NettyStreamingFileUpload implements StreamingFileUpload {
      * @return The temporal file
      */
     protected File createTemp(String location) {
-        File tempFile;
         try {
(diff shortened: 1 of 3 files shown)


jameskleeh:
I think this method should be deprecated in 3.4 and removed in 4.0

yawkat:
@jameskleeh `createTemp` or `transferTo(String)`? We can remove `createTemp` right now, it's an internal class.

jameskleeh:
`transferTo(String)`

yawkat:
yea it seems redundant, ive deprecated it

graemerocher (approved):

PR #6969, opened 2022-02-28
Fix various sonar issues

yawkat:
Mostly just cosmetic issues.


graemerocher (approved):

graemerocher:
Merging this for now, if there are any objections it can be rolled back

PR #6979, opened 2022-03-02
Address sonar static analysis issues

graemerocher:


graemerocher:
I can't increase the coverage because it relates to deprecated code that is not used or build time only code

graemerocher:
@sdelamo can this be merged?

sdelamo (approved):

PR #6980, opened 2022-03-02
Don't create stereotype for `@Nonnull(when=MAYBE)`

### LINKED ISSUE #6795: Unexpected results for `Argument.isNonNull` when mapper for `io.micronaut.core.annotation.Nullable` is present
### Issue description

In 3.3.0-M1, a property marked as `@io.micronaut.core.annotation.Nullable` will be marked as `io.micronaut.core.annotation.AnnotatedElement#isNonNull` if an annotation mapper from `io.micronaut.core.annotation.Nullable` is present on the processor classpath.

The setup is as follows:

- Author.java:
```java

@Introspected
class Author {
    private @Nullable
    String nickName;

    public String getNickName() {
        return nickName;
    }

    public void setNickName(String nickName) {
        this.nickName = nickName;
    }
}
```
- NullableAnnotationMapper.java:

```java
import io.micronaut.core.annotation.AnnotationValue;
import io.micronaut.inject.annotation.TypedAnnotationMapper;
import io.micronaut.inject.visitor.VisitorContext;

import java.util.Collections;
import java.util.List;

public class NullableAnnotationMapper implements TypedAnnotationMapper<Nullable> {
    @Override
    public Class<Nullable> annotationType() {
        return Nullable.class;
    }

    @Override
    public List<AnnotationValue<?>> map(AnnotationValue<Nullable> annotation, VisitorContext visitorContext) {
        return Collections.singletonList(
                AnnotationValue.builder("io.micronaut.core.annotation.Nullable").build()
        );
    }
}
```
- AuthorSpec.groovy:
```groovy
class AuthorSpec extends Specification {
    def 'test'() {
        expect:
        !BeanIntrospector.SHARED.findIntrospection(Author.class).get().getProperty("nickName").get().isNonNull()
    }
}
```

The reason for this bug is as follows:

1. `io.micronaut.core.annotation.Nullable` has the annotation `@javax.annotation.Nonnull(when = When.MAYBE)`. This is correct under the semantics of the `when` value.
2. When the property is visited for introspection, the existence of the annotation mapper will make this line not return null: https://github.com/micronaut-projects/micronaut-core/blob/f52c3bd75aa7ba2c128f76f93ad0f91609fedc4d/inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java#L808 – this is the only reason the mapper is required to trigger this bug, it has no other involvement from what I can tell.
3. Further down in that method, the `io.micronaut.core.annotation.Nullable` has its stereotypes visited, which includes `javax.annotation.Nonnull`, which is correct (it is annotated with `javax.annotation.Nonnull` after all)
4. The stereotype is written to the introspection. 
5. `AnnotatedElement.isNonNull` returns true because the stereotype is present: https://github.com/micronaut-projects/micronaut-core/blob/ff78dd4acdf35f2e127c86421ed846556296578f/core/src/main/java/io/micronaut/core/annotation/AnnotatedElement.java#L45-L47

To give some idea of when the different factors of this bug appeared:

1. The `@Nonnull(when = When.MAYBE)` has been there from the start, probably from `javax.annotation.Nullable`, which has a similar annotation.
2. I don't know why stereotypes are only applied this way when a transformer is present. Sounds like a separate bug.
3. n/a
4. n/a
5. #6716 changed `AnnotatedElement.isNonNull` to use the stereotype, not the annotation directly. This is what makes this a regression in 3.3.0-M1.

I can think of multiple ways this could be fixed:

- Could simply remove `@Nonnull(when = When.MAYBE)` from `io.micronaut.core.annotation.Nullable`. This may have some impact on findbugs, but it's probably not a big issue and it would be an easy fix for now.
- During stereotype copying, we could exclude `@Nonnull` when it is marked with `When.MAYBE`. This seems reasonable, but I'm not sure if the copying step is a good place for it.
- `AnnotatedElement.isNonNull` could check for `When.MAYBE`, but I'm not sure how, it doesn't seem to be a thing the stereotype API allows.
- We could revert #6716 for now.

yawkat:
`javax.annotation.Nullable` is annotated with `@Nonnull(when = When.UNKNOWN)`. This annotation needs to be skipped during stereotype copying, or the property in a class introspection will incorrectly return true for `AnnotatedElement.isNonNull`.

As discussed in #6795 this is not usually an issue because `Nullable` isn't usually visited for stereotype copying, but it turns out this can also happen with a `AnnotationMapper` that returns a `Nullable` annotation (see test case).

Fixes #6795

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/annotation/mapping/AnnotationMappingSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/annotation/mapping/AnnotationMappingSpec.groovy
@@ -84,4 +84,30 @@ class Key {
             idWithMapped.getDeclaredAnnotationNames() == idWithoutMapped.getDeclaredAnnotationNames()
     }
 
+    void "test @NonNull stereotype from @Nullable"() {
+        when:
+        def pojo = buildBeanIntrospection('test.Pojo', '''
+package test;
+
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.annotation.mapping.NonNullStereotyped;
+
+@Introspected
+public final class Pojo {
+    @NonNullStereotyped
+    private final String surname;
+
+    public Pojo(String surname) {
+        this.surname = surname;
+    }
+
+    public String getSurname() {
+        return this.surname;
+    }
+}
+''')
+        then:
+        !pojo.getProperty("surname").get().isNonNull()
+    }
+
 }
--- a/inject-java/src/test/groovy/io/micronaut/annotation/mapping/NonNullProducingMapper.java
+++ b/inject-java/src/test/groovy/io/micronaut/annotation/mapping/NonNullProducingMapper.java
@@ -1,0 +1,21 @@
+package io.micronaut.annotation.mapping;
+
+import io.micronaut.core.annotation.AnnotationValue;
+import io.micronaut.inject.annotation.TypedAnnotationMapper;
+import io.micronaut.inject.visitor.VisitorContext;
+
+import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.List;
+
+public class NonNullProducingMapper implements TypedAnnotationMapper<NonNullStereotyped> {
+    @Override
+    public List<AnnotationValue<?>> map(AnnotationValue<NonNullStereotyped> annotation, VisitorContext visitorContext) {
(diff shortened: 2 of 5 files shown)


graemerocher (approved):

yawkat:
@jameskleeh still broken with using the class name instead of the class literal

PR #6981, opened 2022-03-02
Issue 6881 bean shutdown order

### LINKED ISSUE #6881: ExecutableMethodProcessors have no order in relation to the beans they intercept with regards to shutdown
### Expected Behavior

`ExecutableMethodProcessor<Scheduled>` should be shutdown before beans with `@Scheduled` methods

### Actual Behaviour

The shutdown order is not deterministic and thus the scheduled methods can be invoked after the bean has been shutdown causing errors

### Version

3.3.0

mattmoss:
Fixes #6881

### DIFF
--- a/inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
+++ b/inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
@@ -140,6 +140,8 @@ public class AbstractInitializableBeanDefinition<T> extends AbstractBeanContextC
     @Nullable
     private Collection<Class<?>> requiredComponents;
     @Nullable
+    private Collection<Class<?>> runtimeRequiredComponents;
+    @Nullable
     private Argument<?>[] requiredParametrizedArguments;
 
     @SuppressWarnings("ParameterNumber")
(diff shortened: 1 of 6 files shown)


inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
```diff
@@ -388,47 +390,61 @@ public boolean isSingleton() {
...
+        } else {
+            Set<Class<?>> allRequiredComponents = new HashSet<>(this.requiredComponents);
+            allRequiredComponents.addAll(runtimeRequiredComponents);
+            return Collections.unmodifiableSet(allRequiredComponents);
+        }
+    }
+
+    @Override
+    public void addRequiredComponent(Class<?> clazz) {
+        if (runtimeRequiredComponents == null) {
```
jameskleeh: With this change the `.getRequiredComponents` method will always be creating a new set. This method could check  if the `requiredComponents` field is not null and instead just add it there directly. The `getRequiredComponents` method would always then take that list and return a new unmodifiable set.

jameskleeh (approved):

graemerocher:
Tests failing?

PR #6992, opened 2022-03-04
bug: Don't strip newlines proxying ServerSentEvents

### LINKED ISSUE #6985: ProxyHttpClient.proxy() cannot proxy Server-Sent Event stream from backend.
### Expected Behavior

SSEClient connecting to an endpoint that calls  a backend endpoint using ProxyHttpClient.proxy() should raise events same as calling the endpoint directly.

### Actual Behaviour

SSEClient connecting to an endpoint that calls  a backend endpoint using ProxyHttpClient.proxy() does not raise any event except close.

### Steps To Reproduce

Full code to reproduce the problem https://github.com/rajakumar-sg/sse-reverse-proxy
Run tests to understand the issue.

### Environment Information

MacOS

### Example Application

https://github.com/rajakumar-sg/sse-reverse-proxy

### Version

3.3.4

timyates:
When the HttpClient.proxy is used to proxy an SSE stream, we were stripping
delimiters from the data, which caused consumers of this that were sending;

```
Accept: text/event-stream
```

To get a single stripped event at the end of the stream, instead of the events
arriving individually as expected.

This change passes that we are proxying down to the `HttpClientInitializer`
so we can stop stripping delimiters if this is the case.

Fixes: #6985

Huge thanks to @yawkat for helping triage this

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2022 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -898,6 +898,7 @@ public class DefaultHttpClient implements
                     requestKey.getPort(),
                     false,
                     false,
+                    false,
                     null
             ) {
                 @Override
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

yawkat:
we could probably get by with bypassing `addEventStreamHandlerIfNecessary` entirely when proxying

timyates:
@yawkat nice spot!  Raised https://github.com/micronaut-projects/micronaut-core/pull/6999

PR #6989, opened 2022-03-04
Bump micronaut-jaxrs to 3.2.0

micronaut-build:

PR #7009, opened 2022-03-08
test: ReactiveSequenceSpec ignore if jdk17 and CI

sdelamo:
The test is really flaky. 

see: https://github.com/micronaut-projects/micronaut-core/pull/6989/checks?check_run_id=5417864613

or https://github.com/micronaut-projects/micronaut-core/pull/6906/checks?check_run_id=5413571969


timyates:
I can't see how that is only getting a single book 🤔

I worry this may actually be an issue?

sdelamo:
I think we should revisit it. But for now, I think it is best to ignore it. It is making every build fail.

timyates (approved):
👍

sdelamo:
I've created an [issue](https://github.com/micronaut-projects/micronaut-core/issues/7010) so that we don't forget.

PR #7014, opened 2022-03-08
Fix micronaut-test bom inclusion

timyates:
After the update to 3.1.0 in #7008, we no longer need the aliases for the
managed dependencies


timyates:
Hmmm, no...this is wrong... 🤔

melix (approved):

timyates:
@melix I get failures here and locally...  I can't see why...  It should be importing from the bom shouldn't it?

```
    testImplementation(libs.managed.micronaut.test.junit5) {
```

is failing in http-client

melix:
No, if you remove them from the catalog, there won't be accessors anymore. You'd get accessors if we actually used the catalog that we generate in the same build, which is obviously a chicken and egg problem. It's probably fine to keep them in the catalog then.

timyates:
I'll close this then 😀

PR #7038, opened 2022-03-10
Support getting overridden and hidden methods for Java. Corrected case with duplicate interfaces.

dstepanov:
Only implemented for Java, not sure if it's even possible for Groovy.


graemerocher (approved):

sdelamo (approved):

sdelamo:
not sure if we want to address the two code smells reported by sonar

graemerocher:
not worth delaying the change

PR #7058, opened 2022-03-13
micronaut-data issue 1378

rawilder:
With regards to this issue: https://github.com/micronaut-projects/micronaut-data/issues/1378

In cases where a field matches a getter exactly, a property should not be extracted but rather just passed through.

kotlin example:
```kotlin
class TestEntity(val isDeleted: Boolean)
```
turns into jvm byte code roughly representing this java code:
```java
class TestEntity {
    private Boolean isDeleted;
    public Boolean isDeleted() { return isDeleted; }
    // this part bums be out as it makes it seem difficult to correctly identify in the setter path
    // hence why I ignored it in this PR
    public void setDeleted(deleted: Boolean) {
        isDeleted = deleted;
    }
}
```

### DIFF
--- a/core/src/main/java/io/micronaut/core/naming/NameUtils.java
+++ b/core/src/main/java/io/micronaut/core/naming/NameUtils.java
@@ -38,7 +38,7 @@ public class NameUtils {
     private static final Pattern SERVICE_ID_REGEX = Pattern.compile("[\\p{javaLowerCase}\\d-]+");
     private static final String PREFIX_GET = "get";
     private static final String PREFIX_SET = "set";
-    private static final String PREFIX_IS = "is";
+    public static final String PREFIX_IS = "is";
     private static final Pattern ENVIRONMENT_VAR_SEQUENCE = Pattern.compile("^[\\p{Lu}_{0-9}]+");
     private static final Pattern KEBAB_CASE_SEQUENCE = Pattern.compile("^(([a-z0-9])+([-.:])?)*([a-z0-9])+$");
     private static final Pattern KEBAB_REPLACEMENTS = Pattern.compile("[_ ]");
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
@@ -3763,15 +3763,12 @@ class MyConfig {
         then:
         BeanIntrospection beanIntrospection = beanIntrospector.getIntrospection(TestMatchingGetterClass)
         beanIntrospection != null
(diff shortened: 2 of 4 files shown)


graemerocher:
Needs a test

rawilder:
test added

dstepanov:
This doesn't look correct to me, I think the special case should be regarding `is` prefix for boolean fields. In your test case, the property name should be defined by its getter/setter and not by the private field name.

Please target the next major version, 3.4 which is going to be released soon so maybe 3.5.

rawilder:
@dstepanov Property seems to be a pretty overloaded term so I had trouble actually nailing down what should be the value. The best I could tell in this context is that it's the combination of field and getter, so they should match in some capacity.

I'm fine with changing it to be a special case with `is` only, since the other test cases I wrote (getName backed by getName) seem like they really won't ever happen.


edit: also there is no 3.5.x branch. I'm happy to change it once available. For now I'll open it against 3.4.x

core/src/main/java/io/micronaut/core/naming/NameUtils.java
```diff
@@ -38,7 +38,7 @@
     private static final Pattern SERVICE_ID_REGEX = Pattern.compile("[\\p{javaLowerCase}\\d-]+");
     private static final String PREFIX_GET = "get";
     private static final String PREFIX_SET = "set";
-    private static final String PREFIX_IS = "is";
+    public static final String PREFIX_IS = "is";
```
rawilder, line 41: public may have been too aggressive here, I can put a static field in the inject lib instead if you deem this unnecessary

dstepanov:
```
    public Boolean isDeleted() {
        return isDeleted;
    }

    public void setDeleted(Boolean isDeleted) {
        this.isDeleted = isDeleted;
    }
```
I would expect for this case the property name is `deleted` and not `isDeleted` in the same fashion as `getName/setName` results in `name` property.

graemerocher:
yeah we need to be careful that this doesn't cause a regression in the Java side because of Kotlin oddities

rawilder:
I mean this is definitely a kotlin oddity which is partially so annoying. Interested to hear y'all's thoughts. My understanding of property is it's an ambiguous term that generally refers to a combination of class field, getter, and setter. A lot of the use cases I have seen seem to ignore the setter entirely and just deduce "property" from getters strictly. 

How do y'all think we should approach this. I would consider this a bug, a property is being misidentified in the context of other parts of the framework. Namely micronaut data and openapi.

Data is producing bad HQL statements and openapi is producing incorrect REST documentation which leads to 400 bad requests when a consumer reads the swagger docs.

dstepanov:
I have an alternative PR https://github.com/micronaut-projects/micronaut-core/pull/7094

PR #7072, opened 2022-03-15
Referencing bean properties in annotations

GavrilovSV:
This PR develops the ideas described in #5350 and partly implemented in #6173
Since the original issue #5350 was not only about using bean properties in `@Requires` annotation but mostly about allowing to reference bean properties in any annotations where this can be considered useful, this PR does the following:
- Provides a universal mechanism for referencing bean properties in annotations 
- Adds support for this feature in `@Scheduled` and `@Client ` annotations (these are places where out team would like to use this feature)
- Slightly refactors the usage of bean properties in `@Requires` to make it based on the new mechanism (I've been working on this while waiting for #6173 to be merged)

The basic idea of this PR is to introduce a concept of `AnnotationPropertyReference` which is built as part of annotation metadata at compilation time and contains a method reference to bean property getter, which can be supplied with bean of specified type to obtain bean property value at runtime without using reflection

Would be grateful for review and feedback

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -1265,19 +1265,6 @@ public class AopProxyWriter extends AbstractClassFileWriter implements ProxyingB
     }
 
     @Override
-    public void visitAnnotationMemberPropertyInjectionPoint(TypedElement annotationMemberBeanType,
-                                                            String annotationMemberProperty,
-                                                            String requiredValue,
-                                                            String notEqualsValue) {
-        deferredInjectionPoints.add(() ->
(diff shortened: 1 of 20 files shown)


graemerocher:
I am not sure about this as it is quite invasive and complicates the APIs since you have to have a annotation member for each config property.

We have planned to implement expression language support within annotation values where by you should be able to do something like:

```java
@Scheduled(cron="#{ctx[example.ScheduledConfig].cron}")
```

The expression could be evaluated against the context. In general compilation time equivalent of SpEL (Spring Expression Language) is something that we have been wanting to do a long time and would resolve use cases this.

GavrilovSV:
@graemerocher 
It feels like some of the work I've done as part of this PR can be used to add the support of syntax from your example:
```java
@Scheduled(cron="#{ctx[example.ScheduledConfig].cron}")
```
I could try to implement it, if you don't mind

graemerocher:
Think this one is stale

PR #7073, opened 2022-03-15
Allow query field elements from superclasses and interfaces

dstepanov:


graemerocher (approved):

inject-java/src/test/groovy/io/micronaut/visitors/ClassElementSpec.groovy
```diff
@@ -1330,6 +1312,7 @@ abstract class SuperClassWithFields extends SuperSuperClassWithFields implements
                 "instanceField1": ["SuperClassWithFields"],
                 "instanceField2": ["SuperClassWithFields"],
                 "instanceField3": ["InheritedFields"],
+                "interfaceField": ["SuperInterfaceWithFields", "SuperSuperInterfaceWithFields"]
```
sdelamo, line 1315: @dstepanov This change changes the behaviour for 

`List<FieldElement> fields = classElement.getEnclosedElements(ElementQuery.ALL_FIELDS)`

not just when you use `includeHiddenElements` 

is that expected?

dstepanov: Yes

PR #7092, opened 2022-03-17
Correct bean property type

### LINKED ISSUE #1362: Entity implementation detail compromises metadata regardless Data model supposed to only care about public properties
### Expected Behavior

Type of fields should not affect Data module mapping. Internal private typing concerns Micronaut regardless of public api for pojo is consistent

### Actual Behavior

Private field type is considered by this module and breaks design.
image

#### ~~EDIT~~
~~BTW, same setup within `@Embeddable` works just fine!~~

### Steps To Reproduce

See repro

### Environment Information

Data 3.3.0-RC1
core 3.3.4

### Example Application

https://github.com/soberich/micronaut-data-1362

### Version

3.3.4

dstepanov:
The type of the bean property should be the getter/setter/constructor type, not the private field type.

Not sure what would be the best way @yawkat 

Fixes https://github.com/micronaut-projects/micronaut-data/issues/1362

### DIFF
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
@@ -4099,7 +4099,7 @@ public class Test {
 
     void "verify the type of the bean property should be the getter/setter/constructor type, not the private field type."() {
         when:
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

yawkat (approved):

sdelamo (approved):

soberich:
still there
https://github.com/micronaut-projects/micronaut-data/issues/1362#issuecomment-1080085919

PR #7093, opened 2022-03-17
Add Micronaut AOT version to the BOM as a managed property

alvarosanchez:


melix (approved):
I guess this is for the Maven parent POM?

alvarosanchez:
This is to produce

```xml
<properties>
  <micronaut.aot.version>1.0.0-M7</micronaut.aot.version>
   ...
</properties>
```

In the `micronaut-bom`.

It could instead be added to the `micronaut-parent`, but that has a _stringly-typed_ Groovy DSL that is impossible to automate wrt version updates.

I don't have a strong opinion. @graemerocher thoughts?

PR #7094, opened 2022-03-18
Correct handling of properties starting with 'is' for Kotlin

dstepanov:
It looks like Kotlin will generate `isProp` getter only when the property is prefixed `is` otherwise `Boolean` property will have `getProp` getter.

Fixes https://github.com/micronaut-projects/micronaut-data/issues/1378
Fixes https://github.com/micronaut-projects/micronaut-data/issues/942#issuecomment-1071112210
Fixes https://github.com/micronaut-projects/micronaut-openapi/issues/616

### DIFF
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
@@ -4165,6 +4165,25 @@ class AImpl implements A {
         property.isReadWrite()
     }
 
+    void "test targeting abstract class with @Introspected(classes = ) with getter matching field name"() {
+        ClassLoader classLoader = buildClassLoader("test.Test", """
+package test;
+import io.micronaut.core.annotation.Introspected;
+@Introspected(classes = {io.micronaut.inject.visitor.beans.TestMatchingGetterClass.class})
+class MyConfig {
+}
+""")
+
+        when:
+            BeanIntrospector beanIntrospector = BeanIntrospector.forClassLoader(classLoader)
+
+        then:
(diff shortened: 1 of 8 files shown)


inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
```diff
@@ -4165,6 +4165,25 @@ class AImpl implements A {
...
+""")
+
+        when:
+            BeanIntrospector beanIntrospector = BeanIntrospector.forClassLoader(classLoader)
+
+        then:
+            BeanIntrospection beanIntrospection = beanIntrospector.getIntrospection(TestMatchingGetterClass)
+            beanIntrospection != null
+            beanIntrospection.getBeanProperties().size() == 3
+            beanIntrospection.getPropertyNames() == ["deleted", "updated", "name"]
```
rawilder: For Java, would this cause the same issue with openapi and data in the case someone named their backing field "isDeleted"?

dstepanov: For Java, you need to write getter/setter manually so it depends on how you're going to name them.

graemerocher (approved):

yawkat (approved):

sdelamo (approved):

inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
rawilder: I guess that's fair, the implementer can just change the name when they see the bug and it's still just "Java code" compared to Kotlin's byte code modification annotation. Thanks for looking into this!

PR #7097, opened 2022-03-18
Add support for reflectively invoking executable methods for imported beans 

graemerocher:
When importing beans from other already compiled libraries it is only possible to invoke methods without reflection when they are public. This falls back to use reflection for non-public methods on imported beans.


dstepanov (approved):

inject/src/main/java/io/micronaut/inject/writer/DispatchWriter.java
```diff
@@ -45,6 +46,7 @@
  * @author Denis Stepanov
  * @since 3.1
  */
+@Internal
```
sdelamo, line 49: Are we ok making this class @Internal a minor release?

sdelamo (approved):

inject/src/main/java/io/micronaut/inject/writer/DispatchWriter.java
graemerocher: It should have been internal in the first place. Everything under `io.micronaut.inject.writer` is regarded internal anyway

PR #7112, opened 2022-03-21
Disable InputStreamBodySpec2

yawkat:
James didn't manage to fix it, and I don't plan to spend time looking at it in the near future either, so just disable it for now.


graemerocher (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/stream/InputStreamBodySpec2.groovy
```diff
@@ -42,6 +43,7 @@ class InputStreamBodySpec2 extends Specification {
 
     @Retry
     @Issue('https://github.com/micronaut-projects/micronaut-core/issues/6100')
+    @Ignore // this test is super flakey and I'm not sure why
```
sdelamo: You can put the reason to ignore in the `@Ignore::value`

sdelamo (approved):

PR #7120, opened 2022-03-21
Update Micronaut Parent POM

alvarosanchez:
Refreshes Maven plugin versions. It also changes the GraalVM Native Plugin definition
by not manually passing anymore the application JAR as a classpath argument, since this
is now handled in the Native Plugin (see https://github.com/graalvm/native-build-tools/pull/226)

This PR requires Native Plugin version 0.9.11 released before merging.


melix (approved):
I see that you use `0.9.11`, which is correct since it's technically a breaking change. Just making sure this doesn't get merged until that version is out :)

alvarosanchez:
I wonder how could the build pass since I used a non-existent version? Weren't we checking that?

melix:
We're checking the BOM, not the parent POM afaik.

alvarosanchez:
The version is changed in the catalog file, not the parent POM.

melix:
The catalog file isn't checked: we check the generated BOM.

alvarosanchez:
Ok, understood. Thanks!

alvarosanchez:
@sdelamo this is ready to be merged into 3.4.x

PR #7133, opened 2022-03-22
Fix BeanProvider's find method

### LINKED ISSUE #7132: BeanProvider's find method ignores passed qualifier
### Expected Behavior

A qualifier passed to BeanProvider's `find` method should participate in bean filtering process.

### Actual Behaviour

It is not. 

### Steps To Reproduce

`qualifier` argument is just ignored:
https://github.com/micronaut-projects/micronaut-core/blob/cd6065195aff610c70865c2f9b194070dc01cbb1/inject/src/main/java/io/micronaut/inject/provider/BeanProviderDefinition.java#L72-L74

### Environment Information

### Example Application

_No response_

### Version

3.4.0

Paullo612:
Fixes #7132

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/provider/BeanProviderSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/provider/BeanProviderSpec.groovy
@@ -17,13 +17,12 @@ package io.micronaut.inject.provider
 
 import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
 import io.micronaut.context.ApplicationContext
-import io.micronaut.context.BeanContext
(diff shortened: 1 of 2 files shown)


Paullo612:
Looks like `get(Qualifier<Object> qualifier)` is also affected. It ignores Qualifier specified on provider injection point:
https://github.com/micronaut-projects/micronaut-core/blob/cd6065195aff610c70865c2f9b194070dc01cbb1/inject/src/main/java/io/micronaut/inject/provider/BeanProviderDefinition.java#L82-L84

 I could also fix that in this merge request, but wouldn't it be a breaking change?

May I force push, or it would be better to do another commit?

By the way, something went wrong with SonarCube:
image

PR #7145, opened 2022-03-25
Do not add `Named` to factory bean fields

dstepanov:
I have spent a long time but can only reproduce it on ODI project where the qualifier is resolved as something strange: `@Named("Named")` and then doesn't work when called `getBean(Argument, Qualifier)`.
 
It's a bit confusing as to how this possible can work, added `Named` cannot be recognized as `Qualifier` to adjust `annotationsByStereotype` etc


graemerocher:
Fix is probably correct, the intention is that if you have `@Named` without a value then it should calculate the value of named from the name of the field. Is there a test that can be added for this?

dstepanov:
As I wrote before I’m not able to detect added annotation in core

graemerocher (approved):

sdelamo (approved):

PR #7146, opened 2022-03-25
Update jackson

yawkat:


yawkat:
@melix can you take a look at how to do this? 2.13.2.1 was only released for the jackson-databind artifact, but the build is failing with missing 2.13.2.1 bom. not sure where that comes from or how to fix it.

yawkat:
Apparently we're not the only ones with this problem: https://github.com/FasterXML/jackson-databind/issues/3428

melix:
Right, so it's wrong metadata published on the Jackson side. We can fix it locally via a component metadata rule, but it wouldn't be propagated to our users.

Alternatively, we can add a constraint which would fix this problem but should be removed once a new version of Jackson which fixes the problem is released:

```gradle
dependencies {
    constraints {
        api('com.fasterxml.jackson:jackson-bom') {
            version {
                strictly '[2.13.2, 2.13.3['
                prefer '2.13.2'
                reject '2.13.2.1'
            }
            because 'Jackson Databind references non existent BOM'
        }
    }
}
```

yawkat:
ok i will wait a few days and if tatu fixes it on jacksons side we can update to 2.13.2.2, else use the workaround

melix:
Note however that if we publish this and that 2.13.2.1 BOM is finally published, then we would never get it, which is probably not acceptable. So maybe wait for an official answer from the Jackson folks.

melix:
Issue seems to be fixed now.

yawkat:
@sdelamo please review and merge this before releasing 3.4.1.

sdelamo:
@yawkat thanks for being on top of this.

PR #7165, opened 2022-03-30
Improve how factory method/fields annotation are resolved

dstepanov:
Main points:
- When a factory method/field declares a scope or a qualifier/s non of scopes or qualifiers from the actual bean type are copied to the final annotation metadata
- Fixed cloning of `DefaultAnnotationMetadata/MutableAnnotationMetadata` before the value collections weren't cloned
- Changes in `createFactoryBeanMethodWriterFor` making the problem in the previous PR with `@Named` reproducible
- Changed docs to have the field factory section follow the method factory section

The changes regarding not coping qualifiers from the factory class to the producers might not be 100% correct, the main problem that I'm trying to solve is in the CDI no-qualifier beans are mapped to have `@Default` qualifier and with the current code in the core, we end up with multiple qualifiers like `@Xyz + @Default` which is not correct.

The solution might be for Micronaut to add some kind of limited support for `@Default` in the same way as we support `@Any`.


graemerocher:
Conceptually sounds fine to me. There are test failures in the change.

graemerocher (approved):

Issue #5491, opened 2021-05-24
HttpRequests are mutable in filters


johnowl:
### Task List

- [X] Steps to reproduce provided
- [X] Stacktrace (if present) provided
- [X] Example that reproduces the problem uploaded to Github
- [X] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create two filters that changes the request
2. In the third filter you can read the updated request

### Expected Behaviour

HttpRequests should be immutable, the only property mutable is the attributes field.

### Actual Behaviour

HttpRequests are mutable in a specific scenario.

### Environment Information

- **Operating System**: Darwin
- **Micronaut Version:** 2.5.4
- **JDK Version:** 19.6.0

### Example Application

https://github.com/johnowl/micronaut-api-gateway/

yawkat:
Partial fix in #7174 (similar to #5487). I'm not sure we can or should support the full use case of replacing the request in the filter, though.

graemerocher:
Triage: classified as an HTTP filter API enhancement. PR #7174 (commit fe0bcce83c) fixed an important part of request replacement in filter processing, but it does not make HttpRequest immutable. Suggested direction: keep the incoming request snapshot immutable from a filter’s point of view and make replacement explicit in the filter result/pipeline, with a compatibility adapter for existing MutableHttpRequest use. Add tests for mutation after a filter has passed the request onward, replacement of URI/headers/body, and concurrent observation so the intended ownership and visibility semantics are unambiguous.

PR #7174, opened 2022-04-01
Properly proceed in filter processing when request is replaced

yawkat:
When a filter calls chain.proceed with a new request, that request should be used for further processing.
Replacing the request is still not fully working even with this change, so #5491 is not fully fixed. RoutingInBoundHandler still does not consider the new request, because most of the request processing happens before the filter is applied. In the test case of this patch, this manifests as the response text still being "initial" even though the request has been replaced.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterReplaceRequestSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterReplaceRequestSpec.groovy
@@ -34,8 +34,7 @@ class FilterReplaceRequestSpec extends Specification {
         server.stop()
         client.stop()
     }
-
-
+    
     @Filter(Filter.MATCH_ALL_PATTERN)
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

sdelamo:
> When a filter calls chain.proceed with a new request, that request should be used for further processing.
> Replacing the request is still not fully working even with this change, so #5491 is not fully fixed. RoutingInBoundHandler still does not consider the new request, because most of the request processing happens before the filter is applied. In the test case of this patch, this manifests as the response text still being "initial" even though the request has been replaced.

I assume that in the test scenario what should happen is that the application responds _filter2_. I have created such a test and annotated with `@PendingFeature`. 

https://github.com/micronaut-projects/micronaut-core/pull/7174/commits/3e72e9f22cb75488a5a762ac938b9342b3e88d6d

sdelamo:
should we point this to 3.4.x or 3.5.x?

sdelamo (approved):

yawkat:
This is a bug, it should be 3.4.x

PR #7180, opened 2022-04-05
fix: String[] is convertible into String with comma delimiter

vsapronov:
Without this delimiter micronaut declarative client treats `Array<String>` query params in a buggy way. More specifically: it just concats all strings into single string without delimiter. This change fixes it by adding comma delimiter which is acceptable strategy for passing arrays in query params.

### DIFF
--- a/core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java
+++ b/core/src/main/java/io/micronaut/core/convert/DefaultConversionService.java
@@ -334,7 +334,7 @@ public class DefaultConversionService implements ConversionService<DefaultConver
                 return Optional.empty();
             }
 
-            StringJoiner joiner = new StringJoiner("");
+            StringJoiner joiner = new StringJoiner(",");
             for (String string : object) {
                 joiner.add(string);
             }


sdelamo:
this will be a breaking change.

vsapronov:
@sdelamo Are you sure that it would break any existing functionality? I mean: is this code path even used anywhere?

Also may be I'm totally wrong with my assumptions. I'm trying to use `Array<String>` as a query param. I have a suspicion that either its'n really supported or that I'm doing something totally wrong.

I'm basically trying to pass `@QueryValue(value = "myvar") myVar: Array<String>` into declarative client. Did I configure it correctly?
When I run such client with `arrayOf("the str1", "the str2")` passed into `myvar` on my server side I see that these values just were glued together and the only one string value was passed `"the str1the str2"` - there's no delimiter.

I also tried to add my own `StringArrayConverter: TypeConverter<Array<String>, String>` to override the default converter. However micronaut doesn't seem to like `Array<String>` type as `S` type argument and just not adding this bean into collection of type converters.

The only thing that is "working" for me is adding my own `StringArrayConverter` through `TypeConverterRegistrar`. It worked but seems to be a bit undocumented and because of that potentially far from recommended way (?).

So, are array query params supported/tested? What I'm doing wrong?

PR #7190, opened 2022-04-06
Add grpc back in as a managed version

timyates:
For 3.5.0, we added a BOM to GRPC and used it here.  The grpc version is required in the starter
module however, and removing that caused it to break.

I tried and failed to fix it here https://github.com/micronaut-projects/micronaut-starter/pull/1162

This change adds back in the GRPC bom, which was previously removed in

https://github.com/micronaut-projects/micronaut-core/pull/7162

And updates the grpc and protobuf versions to align with the current versions in the grpc
module v3.2.0

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -41,6 +41,7 @@ wiremock = "2.27.2"
 managed-commons-dbcp = "2.9.0"
 managed-dekorate = "1.0.3"
 managed-elasticsearch = "7.16.3"
+managed-grpc = "1.45.0"
 managed-ignite = "2.8.1"
 managed-junit5 = "5.7.2"
 managed-kotlin = "1.6.10"
@@ -129,7 +130,7 @@ managed-neo4j = "3.5.29"
 managed-neo4j-java-driver = "4.2.7"
 managed-netty = "4.1.73.Final"
 managed-picocli = "4.6.1"
-managed-protobuf = "3.17.2"
+managed-protobuf = "3.19.4"
 managed-reactive-pg-client = "0.11.4"
 managed-reactive-streams = "1.0.3"
 # This should be kept aligned with https://github.com/micronaut-projects/micronaut-reactor/blob/master/gradle.properties from the BOM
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -181,6 +182,7 @@ boms-micronaut-r2dbc = { module = "io.micronaut.r2dbc:micronaut-r2dbc-bom", vers
 boms-micronaut-flyway = { module = "io.micronaut.flyway:micronaut-flyway-bom", version.ref = "managed-micronaut-flyway" }
 
 boms-groovy = { module = "org.codehaus.groovy:groovy-bom", version.ref = "managed-groovy" }
+boms-grpc = { module = "io.grpc:grpc-bom", version.ref = "managed-grpc" }
```
melix: I think you don't need to reintroduce the `boms-grpc`, only the `managed-grpc` version, right?

timyates: You're absolutely right, tested a snapshot and `micronaut-versions.properties` got correctly populated with `grpc.version=1.45.0` in starter 👍  thanks!

I'll get all this one day... 🤣

timyates: Removed in https://github.com/micronaut-projects/micronaut-core/pull/7190/commits/bae5250cffb8d110110bb79dfb11074636580516

sdelamo:
> The grpc version is required in the starter module 

Why? 

I have a bad feeling that we have to add the `managed-grpc` here. Are not we creating BOMs to avoid having to keep track of those versions in core?

sdelamo (commented):
> The grpc version is required in the starter module 

Why? 

I have a bad feeling that we have to add the `managed-grpc` here. Are not we creating BOMs to avoid having to keep track of those versions in core?

melix:
That's because the build scripts have the protobuf configuration which requires to set versions which need to be properties in the BOM:

```
protobuf {
    protoc { artifact = "com.google.protobuf:protoc:@VersionInfo.getBomVersion("protobuf")" }
    plugins {
        grpc { artifact = "io.grpc:protoc-gen-grpc-java:@VersionInfo.getBomVersion("grpc")" }
    ...
```
```

timyates:
I wonder if we could fix starter by digging through all the boms and pulling out all the versions here https://github.com/micronaut-projects/micronaut-starter/blob/3.4.x/buildSrc/src/main/groovy/io/micronaut/internal/starter/tasks/WriteMicronautVersionInfoTask.java#L66

Currently we just did through the main micronaut bom

melix:
This would work, yes, granted that there's no overlap of properties.

timyates:
I have described the issue here https://github.com/micronaut-projects/micronaut-starter/issues/1173 and proposed 2 solutions...

Other (better) solutions welcome! 😀

alvarosanchez (approved):

timyates:
Closed in preference to https://github.com/micronaut-projects/micronaut-core/pull/7202

Issue #6579, opened 2021-11-28
GraalVM error with Micronaut 3: Multiple possible bean candidates found: io.micronaut.aop.InterceptorRegistry


tobiasschaefer:
### Expected Behavior

Building and executing the native image of the provided and minimal application should succeed. Actually, it was successful on Micronaut 2.5.x but fails since Micronaut 3.0.0

The bug is triggered by injecting a configuration interface:
```
@ConfigurationProperties("myconfig")
public interface Config {

    String getBaseUrl();
}
```

into a class which is annotated with `@Context`: 

```
@Context
public class Bean {

    private final Config config;

    public Bean(Config config) {
        this.config = config;
    }
}
```
and building and executing the native image based on a generated native image configuration.

The error message is the same as in #6406 but they don't seem to be related.

### Actual Behaviour

 `build/native/nativeCompile/context` fails with 

```
Message: Multiple possible bean candidates found: [io.micronaut.aop.InterceptorRegistry, io.micronaut.aop.InterceptorRegistry]
Path Taken: new Bean([Config config])
        at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:2364)
        at io.micronaut.context.DefaultBeanContext.createAndRegisterSingletonInternal(DefaultBeanContext.java:3282)
        at io.micronaut.context.DefaultBeanContext.createAndRegisterSingleton(DefaultBeanContext.java:3268)
        at io.micronaut.context.DefaultBeanContext.getBeanForDefinition(DefaultBeanContext.java:2821)
        at io.micronaut.context.DefaultBeanContext.getBeanInternal(DefaultBeanContext.java:2783)
        at io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:1639)
        at io.micronaut.context.AbstractInitializableBeanDefinition.resolveBean(AbstractInitializableBeanDefinition.java:1933)
        at io.micronaut.context.AbstractInitializableBeanDefinition.getBeanForConstructorArgument(AbstractInitializableBeanDefinition.java:1189)
        at context.$Bean$Definition.build(Unknown Source)
        at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:2337)
        at io.micronaut.context.DefaultBeanContext.createAndRegisterSingletonInternal(DefaultBeanContext.java:3282)
[...]

PR #7195, opened 2022-04-08
Allow fully incremental compilation

graemerocher:
This is a significant change that alters Micronaut so that instead of a single registry of type names contains in `META-INF/services` entries a directory is created with file names of each type within `META-INF/micronaut` instead.

The benefit of this change is that we can now fully incrementally compile and avoid completely re-running the annotation processors on each change to an annotated type. We also simplify our annotation processing pipeline with the downstream service generating processor no longer necessary.

The change has implications to AOT and other areas but I don't consider it breaking as Micronaut will continue to work as previously with the change.

This change also addresses issues with Graal since we abuse `META-INF/services` by storing types that cannot actually be loaded with the standard Java service loader (Micronaut's bean references can reference types that not on the classpath).

This should also address improve the situation with #6579


buildSrc/src/main/groovy/io.micronaut.build.internal.sanity-checks.gradle
```diff
@@ -13,5 +13,6 @@ def verifyServiceCount = tasks.register("verifyServiceCount", VerifyServiceCount
 }
 
 tasks.named("compileJava") {
-    finalizedBy(verifyServiceCount)
+//    TODO: restore this?
+//    finalizedBy(verifyServiceCount)
```
melix: I think this check can be removed now. It never failed since we switched to forked compilation.

jameskleeh:
Looks OK to me, but it begs the question whether Micronaut 4 is a good time to break binary compatibility. There is a bit of cruft with the old bean definition style, using javax.inject in annotation metadata, and many other places.

jameskleeh (approved):

test-suite/src/functionalTest/groovy/io/micronaut/context/ApplicationContextConfigurerSpec.groovy
```diff
@@ -67,9 +67,6 @@ class Application {
         hasServiceFileFor(ApplicationContextConfigurer) {
             withImplementations 'demo.app.Application$Configurer'
         }
-        hasServiceFileFor(BeanDefinitionReference) {
```
melix, line 70: Maybe add a new fixture to test this.

melix (approved):
Changes LGTM!

graemerocher:
@jameskleeh just wanted to note this change doesn't break binary compatibility. Old modules still load fine following this change.

graemerocher:
Note, I can't really improve the coverage as the code is in build-time / GraalVM classes which is impossible to cover

graemerocher:
hold off merging the graal visitor is still a problem

core/src/main/java/io/micronaut/core/graal/GraalReflectionConfigurer.java
```diff
@@ -0,0 +1,184 @@
...
+                }
+                if (accessType.contains(TypeHint.AccessType.ALL_PUBLIC_FIELDS)) {
+                    final Field[] fields = t.getFields();
+                    for (Field field : fields) {
+                        if (Modifier.isPublic(field.getModifiers())) {
+                            context.register(field);
+                        }
+                    }
+                }
+                if (accessType.contains(TypeHint.AccessType.ALL_DECLARED_CONSTRUCTORS)) {
```
sdelamo: there is an `ALL_DECLARED_CONSTRUCTOR` block below.

Is this correct?

If it is, I would collapse both if statements.

core/src/main/java/io/micronaut/core/annotation/ReflectionConfig.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * An annotation that models directly the GraalVM reflect-config.json format.
+ *
+ * @since 3.5.0
+ * @author graemerocher
+ * @see io.micronaut.core.graal.GraalReflectionConfigurer
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(ReflectionConfig.ReflectionConfigList.class)
+public @interface ReflectionConfig {
```
sdelamo, line 31: Please, document it in [graalServices.adoc](https://github.com/micronaut-projects/micronaut-core/blob/3.4.x/src/main/docs/guide/languageSupport/graal/graalServices.adoc)

sdelamo (changes requested):

sdelamo (approved):

PR #7196, opened 2022-04-09
Bean context improvements and refactoring

dstepanov:
Multiple improvements for CDI implementation:

- Delegate AbstractInitializableBeanDefinition bean context interaction thru BeanResolutionContext, allowing to inject of custom implementations and aggregate dependable beans
- Refactor of DefaultBeanContext to have main getBean methods return the bean definition which CDI impl. might need to determine the scope of the bean. New methods to get bean registration by the bean definition.

Big refactoring of bean context:
- Extract everything related to the singleton into one class to reduce the size of `DefaultBeanContext`
- Changed how singletons are stored, instead of multiple entries of argument-qualifier keys now there is one representation of `BeanDefinition -> BeanRegistration` plus the fast lookup by argument-qualifier
- Removed duplicate logic
- Added correct `stop` context cleanup, which also spotted a few tests that were using the context after it's been closed
- Extracted some methods to make code more readable

Sonar fixes to introduce generics and simplify methods

### DIFF
--- a/inject/src/main/java/io/micronaut/context/BeanDisposingRegistration.java
+++ b/inject/src/main/java/io/micronaut/context/BeanDisposingRegistration.java
@@ -16,6 +16,7 @@
 package io.micronaut.context;
 
 import io.micronaut.context.exceptions.BeanDestructionException;
+import io.micronaut.core.annotation.Internal;
 import io.micronaut.inject.BeanDefinition;
 import io.micronaut.inject.DisposableBeanDefinition;
 
@@ -26,8 +27,10 @@ import java.util.ListIterator;
  * The disposing bean registration.
  *
  * @param <BT> The bean type
+ * @author Denis Stepanov
  * @since 3.5.0
  */
+@Internal
 final class BeanDisposingRegistration<BT> extends BeanRegistration<BT> {
     private final BeanContext beanContext;
     private final List<BeanRegistration<?>> dependents;
--- a/inject/src/main/java/io/micronaut/context/BeanDisposingRegistration.java
+++ b/inject/src/main/java/io/micronaut/context/BeanDisposingRegistration.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context;
+
+import io.micronaut.context.exceptions.BeanDestructionException;
+import io.micronaut.inject.BeanDefinition;
+import io.micronaut.inject.DisposableBeanDefinition;
+
+import java.util.List;
+import java.util.ListIterator;
+
+/**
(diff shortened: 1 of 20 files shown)


dstepanov:
I did some benchmarking and I see around 5% better startup time

dstepanov:
Forcing every cache to clean revealed a problem in the Micronaut test:

```
        spec.addCleanupSpecInterceptor(invocation -> {
            afterTestClass(buildContext(invocation, null)); // run some tx stuff
            afterClass(invocation); // context close

            invocation.proceed(); // actual method call
            singletonMocks.clear();
        });
   ```

I think the method call should be before the context close and maybe even before the possible transaction rollback.

File: inject/src/main/java/io/micronaut/context/BeanDisposingRegistration.java [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.ListIterator;
+
+/**
+ * The disposing bean registration.
+ *
+ * @param <BT> The bean type
+ * @since 3.5.0
+ */
+final class BeanDisposingRegistration<BT> extends BeanRegistration<BT> {
```
graemerocher, line 34: annotate with `@Internal` and add `@author` tag

inject/src/main/java/io/micronaut/context/SingletonScope.java
```diff
@@ -0,0 +1,431 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Stream;
+
+/**
+ * The singleton scope implementation.
+ *
+ * @author Denis Stepanov
+ * @since 3.5.0
+ */
+final class SingletonScope {
```
graemerocher, line 40: annotate with `@Internal`

graemerocher (changes requested):
Looks good just a few minor changes. Perhaps try to cleanup as many of the code smells or if you cannot supress them with `@SuppressWarning("java:s24234")` or whatever the sonar number is

dstepanov:
I don't like the idea of polluting the code with `@SuppressWarning`s just to satisfy Sonar :-/

graemerocher (approved):

alvarosanchez:
> I don't like the idea of polluting the code with `@SuppressWarning`s just to satisfy Sonar :-/

Note they can be bulk closed from the Sonar UI.

test-suite-jakarta-inject-bean-import/src/test/java/io/micronaut/jakartainject/tck/beanimport/BeanImportTck.java
```diff
@@ -12,8 +12,8 @@
 public class BeanImportTck {
 
     public static Test suite() {
-        try (BeanContext beanContext = BeanContext.run()) {
-            return Tck.testsFor(beanContext.getBean(Car.class), false, true);
-        }
+        BeanContext beanContext = BeanContext.run();
```
sdelamo, line 19: I think we should close the `BeanContext`

https://github.com/micronaut-projects/micronaut-core/pull/7233

inject/src/main/java/io/micronaut/context/NoInjectionBeanDefinition.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Manual injection bean definition.
+ *
+ * @param <T> The bean type
+ * @since 3.5.0
+ */
+final class NoInjectionBeanDefinition<T> implements BeanDefinition<T>, BeanDefinitionReference<T> {
```
sdelamo, line 50: should this be `@Internal`?

dstepanov: added

sdelamo (approved):

PR #7202, opened 2022-04-11
Allow fetching version props from included boms

timyates:
We require the grpc.version from the grpc bom.  This is no longer managed by the core bom.

This change allows setting a list of projects to fetch the bom form, and these properties are
added to those extracted from the core bom

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/build/internal/WriteMicronautVersionInfoTask.java
+++ b/buildSrc/src/main/groovy/io/micronaut/build/internal/WriteMicronautVersionInfoTask.java
@@ -1,0 +1,102 @@
+package io.micronaut.build.internal;
+
+import groovy.lang.Closure;
+import groovy.xml.XmlSlurper;
+import groovy.xml.slurpersupport.GPathResult;
+import groovy.xml.slurpersupport.NodeChild;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.ResolvedArtifact;
+import org.gradle.api.artifacts.ResolvedDependency;
+import org.gradle.api.artifacts.result.ArtifactResolutionResult;
+import org.gradle.api.artifacts.result.ComponentArtifactsResult;
+import org.gradle.api.artifacts.result.ResolvedArtifactResult;
+import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.provider.ListProperty;
+import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.OutputDirectory;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.maven.MavenModule;
+import org.gradle.maven.MavenPomArtifact;
+import org.xml.sax.SAXException;
+
+import javax.xml.parsers.ParserConfigurationException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.TreeMap;
+
+@CacheableTask
+public abstract class WriteMicronautVersionInfoTask extends DefaultTask {
+
+    public static final String MICRONAUT_VERSIONS_PROPERTIES_FILE_NAME = "micronaut-versions.properties";
+
+    @Input
+    public abstract Property<String> getVersion();
+
+    @Input
+    public abstract ListProperty<String> getExtraBomProperties();
+
+    @OutputDirectory
+    public abstract DirectoryProperty getOutputDirectory();
+
+    public WriteMicronautVersionInfoTask() {
+        getOutputs().doNotCacheIf("snapshot version", spec -> getVersion().get().endsWith("SNAPSHOT"));
+    }
+
+    @TaskAction
+    public void writeVersionInfo() throws IOException {
+        Map<String, String> props = new TreeMap<>();
+        for (Dependency dependency : getProject().getConfigurations().getByName("bomVersions").getAllDependencies()) {
+            getLogger().lifecycle("Scanning {}:{}:{}", dependency.getGroup(), dependency.getName(), dependency.getVersion());
+            Map<String, String> bomProperties = bomProperties(dependency.getGroup(), dependency.getName(), dependency.getVersion());
+            for (Map.Entry<String, String> entry : bomProperties.entrySet()) {
+                if (entry.getKey().startsWith("micronaut.")) {
+                    getLogger().lifecycle("Skipping {} from {}", entry.getKey(), dependency);
+                } else {
+                    if (props.containsKey(entry.getKey())) {
+                        getLogger().warn("Property {} from {} already exists ({}). Replacing with {}", entry.getKey(), dependency, props.get(entry.getKey()), entry.getValue());
+                    }
+                    props.put(entry.getKey(), entry.getValue());
+                }
+            }
+        }
+        try (OutputStream out = new FileOutputStream(getOutputDirectory().file(MICRONAUT_VERSIONS_PROPERTIES_FILE_NAME).get().getAsFile())) {
+            for (Map.Entry<String, String> entry : props.entrySet()) {
+                String line = entry.getKey() + "=" + entry.getValue() + "\n";
+                out.write(line.getBytes(StandardCharsets.ISO_8859_1));
+            }
+        }
+    }
+
+    private Map<String, String> bomProperties(String groupId, String artifactId, String version) {
+        ArtifactResolutionResult result = getProject().getDependencies().createArtifactResolutionQuery()
+                .forModule(groupId, artifactId, version)
+                .withArtifacts(MavenModule.class, MavenPomArtifact.class)
+                .execute();
+        Map<String, String> props = new TreeMap<>();
+        for (ComponentArtifactsResult component : result.getResolvedComponents()) {
+            component.getArtifacts(MavenPomArtifact.class).forEach(artifact -> {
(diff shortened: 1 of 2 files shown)


parent/build.gradle
```diff
@@ -1,3 +1,22 @@
+import io.micronaut.build.internal.WriteMicronautVersionInfoTask
+
+apply plugin:'base'
```
timyates, line 3: Added for `clean`

timyates:
There may be a cleaner way of doing this...  It's on the boundary of "reactive" and "declarative" gradle though, so it's kinda tricky

alvarosanchez:
Would you please share a diff of the parent POM before and after this PR?

timyates:
```diff
❯ diff parent-3.5.0.pre-change.pom parent-3.5.0.post-change.pom
54a55,56
>     <grpc.version>1.45.0</grpc.version>
>     <protobuf.version>3.19.4</protobuf.version>
```

The whole properties section is (on this branch)

```xml
  <properties>
    <jdk.version>1.8</jdk.version>
    <release.version>8</release.version>
    <maven.compiler.source>${jdk.version}</maven.compiler.source>
    <maven.compiler.target>${jdk.version}</maven.compiler.target>
    <maven.compiler.parameters>true</maven.compiler.parameters>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
    <exec.executable>java</exec.executable>
    <micronaut.version>3.5.0-SNAPSHOT</micronaut.version>
    <micronaut-maven-plugin.version>3.2.2</micronaut-maven-plugin.version>
    <azure-functions-maven-plugin.version>1.5.0</azure-functions-maven-plugin.version>
    <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
    <function-maven-plugin.version>0.9.8</function-maven-plugin.version>
    <jib-maven-plugin.version>3.1.4</jib-maven-plugin.version>
    <maven-compiler-plugin.version>3.10.1</maven-compiler-plugin.version>
    <maven-failsafe-plugin.version>2.22.3</maven-failsafe-plugin.version>
    <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
    <maven-shade-plugin.version>3.2.4</maven-shade-plugin.version>
    <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
    <protoc-jar-maven-plugin.version>3.11.4</protoc-jar-maven-plugin.version>
    <grpc.version>1.45.0</grpc.version>
    <protobuf.version>3.19.4</protobuf.version>
  </properties>
```

timyates:
Discovered via `./gradlew pAPTBR`

timyates:
We specify `protobuf.version` in our core `libs.version.toml`

https://github.com/micronaut-projects/micronaut-core/blob/bfdf020866f0b3e05dd2088f3bc34cd17ebbf708/gradle/libs.versions.toml#L132

Which is used to pull in the protobuf bom

https://github.com/micronaut-projects/micronaut-core/blob/bfdf020866f0b3e05dd2088f3bc34cd17ebbf708/gradle/libs.versions.toml#L191

I'm not sure if this is required, or will cause problems if removed 🤔

alvarosanchez (approved):

sdelamo (approved):

sdelamo:
Let's merge this once @melix reviews it.

buildSrc/src/main/groovy/io/micronaut/build/internal/WriteMicronautVersionInfoTask.java
```diff
@@ -0,0 +1,102 @@
...
+    public abstract Property<String> getVersion();
+
+    @Input
+    public abstract ListProperty<String> getExtraBomProperties();
+
+    @OutputDirectory
+    public abstract DirectoryProperty getOutputDirectory();
+
+    public WriteMicronautVersionInfoTask() {
+        getOutputs().doNotCacheIf("snapshot version", spec -> getVersion().get().endsWith("SNAPSHOT"));
```
melix: Why do we need this? This hints at a missing task input.

parent/build.gradle
```diff
@@ -32,6 +51,10 @@ ext.extraPomInfo = {
         'maven-shade-plugin.version'('3.2.4')
         'maven-surefire-plugin.version'('2.22.2') // Override actual Maven surefire and failsafe version (2.12) to get native support for executing tests on the JUnit Platform (JUnit 5)
         'protoc-jar-maven-plugin.version'('3.11.4')
+        // inject the properties from the required boms
+        directory.get().file(WriteMicronautVersionInfoTask.MICRONAUT_VERSIONS_PROPERTIES_FILE_NAME).asFile.text.readLines()*.split('=', 2).each {
```
melix: Can't you move `def micronautVersionInfo = tasks.register(...)` up and use `micronautVersionInfo.flatMap { it.outputDirectory }` here?

melix (approved):

buildSrc/src/main/groovy/io/micronaut/build/internal/WriteMicronautVersionInfoTask.java
timyates: I brought it across from the task in starter... I will have a think

parent/build.gradle
timyates: Done 👍

buildSrc/src/main/groovy/io/micronaut/build/internal/WriteMicronautVersionInfoTask.java
timyates: @melix This seems to work, and feels much nicer

https://github.com/micronaut-projects/micronaut-core/pull/7202/commits/e595a35095cc72f673cdcdf6a44ce613a293e1a5

PTAL, I may be committing Gradle heresy 😉

timyates: And then cleaned up the old input in a second commit 🙄. I got too excited

melix: I think you should use a `Configuration` directly, not a `Property<Configuration>`, because that's a `FileCollection` already. As I understood this isn't configuration cache compatible, but we're far from there I think (cc @eskatos)

timyates: If I try that, I get

```
> Plugin validation failed with 1 problem:
    - Error: Type 'io.micronaut.build.internal.WriteMicronautVersionInfoTask' property 'configuration' has @Input annotation used on property of type 'Configuration'.

      Reason: A property of type 'Configuration' annotated with @Input cannot determine how to interpret the file.

      Possible solutions:
        1. Annotate with @InputFile for regular files.
        2. Annotate with @InputDirectory for directories.
        3. If you want to track the path, return File.absolutePath as a String and keep @Input.
```

Which I believe is Gradle telling me I suck 😉

melix: This is because you need `@InputFiles`

timyates: Gotcha 👍

Hopefully this does it 🤞 https://github.com/micronaut-projects/micronaut-core/pull/7202/commits/0cc5b8f58b48dadb0b23b624a7a3e0eae17ded84

Safe travels!

eskatos: I'm late to the party but here are some thoughts.

Yep, `Configuration` is a lazy type already so no need to wrap it in a `Property`.

You're right Cedric, using a `Configuration` input is not compatible with configuration caching. `Configuration` objects are way too heavyweight.

IIUC the inputs of the task are the _declared_ dependencies, correct?

timyates: Hiya 👋😄

Yeah, I need to parse the BOM dependencies' POM to extract out the properties

So i was stick the boms as dependencies in a custom configuration

https://github.com/micronaut-projects/micronaut-core/pull/7202/files#diff-03d6884a03b4390e72895a091b6c223d2d143290ddc4118f5a9a5354753524aeR15

(There's only 1 at the moment, but it may be more later)

melix: The inputs are actually neither the artifacts (jars) nor the _declared dependencies_, but metadata files. We can _simplify_ to declared dependencies probably, as long as the repositories we resolve from are assumed constant.

timyates: Do you have a pointer to what you mean?

timyates: Do you mean as we do in starter?  https://github.com/micronaut-projects/micronaut-starter/pull/1176/files

We should probably align how we do this so it looks like one person wrote both 😔

melix: > Do you have a pointer to what you mean?

Sorry for not being clear enough. What I meant is that _technically speaking_, the inputs of this task are the metadata files (pom files, and possibly their parent POMs). This is what should be declared as an input. In practice there's no easy way to declare this. So we _could_ simplify by saying that the inputs of the task are the declared dependencies, and let the task resolve the pom files. Doing so assumes that we would always download the pom files from the same location (so that we can guarantee that they are always the same pom files for a given dependency declaration). If they are not, then the repositories are _also_ an input.

There's no specific example for this, what we do now is probably enough: we're going to resolve the files, which is a bit too much but no big deal.

PR #7211, opened 2022-04-12
fix: handle @ConfigurationProperties nested in @EachProperty properly

### LINKED ISSUE #7210: Handling of @ConfigurationProperties nested in interface annotated with @EachProperty is broken
### Expected Behavior

Nested configuration should be retrievable from parent configuration when parent configuration is an interface annotated with `@EachProperty`.

### Actual Behaviour

`io.micronaut.context.exceptions.NonUniqueBeanException` is thrown on nested configuration retrieval attempt.

### Steps To Reproduce

Consider following configuration:
```java
import io.micronaut.context.annotation.ConfigurationProperties;
import io.micronaut.context.annotation.EachProperty;

@EachProperty(value = "foo.bar", list = true)
public interface InterfaceAggregatorConfig {

    AddressConfig getAddress();

    @ConfigurationProperties("address")
    interface AddressConfig {

        String getHost();

        int getPort();
    }
}
```

with the following configuration properties:
```yaml
foo:
  bar:
    - address:
        host: host1
        port: 9999
    - address:
        host: host2
        port: 8888
```
`NonUniqueBeanException` is thown on `InterfaceAggregatorConfig#getAddress()` call if there are multiple items in the `bar` list.

### Environment Information

### Example Application

_No response_

### Version

3.4.1

Paullo612:
Fixes #7210

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/context/env/ConfigurationIntroductionAdvice.java
+++ b/context/src/main/java/io/micronaut/runtime/context/env/ConfigurationIntroductionAdvice.java
@@ -30,6 +30,7 @@ import io.micronaut.core.naming.Named;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.ReturnType;
 import io.micronaut.core.value.PropertyNotFoundException;
+import io.micronaut.inject.qualifiers.Qualifiers;
 
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

sdelamo (approved):
Thanks for the contribution

PR #7222, opened 2022-04-15
Warn when bean created event listeners would never be fired 

### LINKED ISSUE #7013: BeanCreatedEventListener created beans should warn if they have a BeanCreatedEventListener
### Feature description

If a BeanCreatedEventListener is registered for a type, but that bean itself creates other beans, then those created beans will not fire any BeanCreatedEventListener that is listening for them.

The current fix is to [create the beans lazily](https://github.com/micronaut-projects/micronaut-kubernetes/pull/437), but this is a hard issue to spot and fix, and a warning should be emitted if this is detected.

jameskleeh:
Because the bean was created too early and log a warning

Closes https://github.com/micronaut-projects/micronaut-core/issues/7013


jameskleeh:
This implementation is flawed because it ignored the dependency tree beyond the first layer. Working on another implementation

jameskleeh:
The cost to startup time is (# of BCELs * size of all BCEL dependencies) type comparisons

graemerocher:
Could this be altered to target 3.5.x? I am conscious of this change playing well with @dstepanov's refactoring work

sdelamo:
@jameskleeh could you merge 3.5.x against this branch? Thanks!

jameskleeh:
@sdelamo Done

dstepanov (approved):

graemerocher (approved):

timyates:
@dstepanov Can you check this, I had to pull the BeanCreatedListener initialization out into a different method to the others, as it performs scanning

I tried to come up with a fancy way of reducing the duplication, but it ended up a bit contrived when I tried

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -3888,6 +3915,10 @@ public <T> BeanRegistration<T> getInFlightBean(BeanIdentifier beanIdentifier) {
         }
     }
 
+    private static final class CollectionHolder<T> {
```
dstepanov, line 3921: Is it needed?

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1770,13 +1770,70 @@ protected Iterable<BeanConfiguration> resolveBeanConfigurations() {
...
+            }
+            LOG.warn("The bean created event listener {} will not be executed because one or more other bean created event listeners inject {}:\n" +
+                    "{}\n" +
+                    "Change at least one point in the path to be lazy initialized by injecting a provider to avoid this issue", listener.getBeanType().getName(), listensTo.getType().getName(), String.join("\n", paths));
+        }
+    }
+
+    @NonNull
+    private Map<Class<?>, List<BeanCreatedEventListener<?>>> loadCreatedListeners() {
+        final Collection<BeanDefinition<BeanCreatedEventListener>> beanDefinitions = getBeanDefinitions(BeanCreatedEventListener.class);
```
dstepanov, line 1799: Is it possible there are no listeners? In that case, I wouldn't create all the collections. Personally, I would move some of the collection initialization just before the block they are going to be used, not having everything at the top.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
timyates: This is already there in 3.5.x

https://github.com/micronaut-projects/micronaut-core/blob/fe1e8a692c48aacb72b1d16c546b6ec04d50ec20/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java#L3861-L3863

🤔  I assume yes

dstepanov: Diff got me confused

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
timyates: Would changing it to:

```java
    private Map<Class<?>, List<BeanCreatedEventListener<?>>> loadCreatedListeners() {
        final Collection<BeanDefinition<BeanCreatedEventListener>> beanDefinitions = getBeanDefinitions(BeanCreatedEventListener.class);
        if (beanDefinitions.isEmpty()) {
            return Collections.emptyMap();
        }
```

suffice?

dstepanov: yeah

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
timyates: You and me both 😆

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
timyates: Done in https://github.com/micronaut-projects/micronaut-core/pull/7222/commits/20fcf654316b029fc81d4a6f933732817df1e775

timyates:
@sdelamo All Green 🐍💚🦖

Well, apart from Sonar which is unhappy that not everything has a generic type.  I tried, but the sweater started to unravel after I pulled too much string

dstepanov (approved):

PR #7228, opened 2022-04-18
doc: @Blocking

### LINKED ISSUE #7183: document @Blocking
### Issue description

@Blocking is almost undocumented. We have this mention: 

 > Micronaut’s version of Transactional is meta-annotated with @Blocking, ensuring that all methods annotated with it use the I/O thread pool when executing within the HTTP server

https://docs.micronaut.io/snapshot/api/io/micronaut/core/annotation/Blocking.html

It seems the logic is here: 

[DefaultExecutorSelector](https://github.com/micronaut-projects/micronaut-core/blob/3.4.x/context/src/main/java/io/micronaut/scheduling/executor/DefaultExecutorSelector.java)

wetted:
Closes #7183

### DIFF
--- a/src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
+++ b/src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
@@ -4,7 +4,7 @@ If you set `micronaut.server.thread-selection` to `AUTO`, The Micronaut Framewor
 
 NOTE: `@Blocking` only works if you are using `AUTO` thread selection. Micronaut Framework defaults to `MANUAL` thread selection since Micronaut 2.0. We recommend the usage of ann:scheduling.annotation.ExecuteOn[] annotation to execute the blocking operations on a different thread pool. `@ExecutesOn` works for both `MANUAL` and `AUTO` thread selection.
 
-There are some places where the Micronaut framework uses ann:core.annotation.Blocking[] already.
+There are some places where the Micronaut framework uses ann:core.annotation.Blocking[] internally:
 
 |===
 |Blocking Type|Description
--- a/src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
+++ b/src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
@@ -1,0 +1,21 @@
+You can use the ann:core.annotation.Blocking[] annotation to mark methods as blocking.
+
+If you set `micronaut.server.thread-selection` to `AUTO`, The Micronaut Framework offloads the execution of methods annotated with `@Blocking` to the IO thread pool (See: api:io.micronaut.scheduling.TaskExecutors[]).
+
+NOTE: `@Blocking` only works if you are using `AUTO` thread selection. Micronaut Framework defaults to `MANUAL` thread selection since Micronaut 2.0. We recommend the usage of ann:scheduling.annotation.ExecuteOn[] annotation to execute the blocking operations on a different thread. `@ExecutesOn` works for both `MANUAL` or `AUTO` thread selection.
+
+There are some places where the Micronaut framework uses ann:core.annotation.Blocking[] already.
+
+|===
+|Blocking Type|Description
+
+|link:{micronautapi}http/client/BlockingHttpClient.html[BlockingHttpClient]
+| Intended for testing, and provides blocking versions for a subset of api:http.client.HttpClient[] operations.
+|link:{micronautapi}core/io/IOUtils.html[IOUtils]
+| Reads the contents of a `BufferedReader` in a blocking manner, and returns that as a `String`.
+|link:{micronautapi}context/env/BootstrapPropertySourceLocator.html[BootstrapPropertySourceLocator]
+| Resolves either remote or local api:context.env.PropertySource[] instances for the current `Environment`.
+
+|===
+
+TIP: https://micronaut-projects.github.io/micronaut-data/latest/guide/[Micronaut Data] also utilizes `@Blocking` for some transaction operations, CRUD interceptors, and repositories.
--- a/src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
+++ b/src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
@@ -18,4 +18,4 @@ There are some places where the Micronaut framework uses ann:core.annotation.Blo
 
 |===
 
-TIP: https://micronaut-projects.github.io/micronaut-data/latest/guide/[Micronaut Data] also utilizes `@Blocking` for some transaction operations, CRUD interceptors, and repositories.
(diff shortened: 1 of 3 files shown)


jameskleeh:
I don't think we should document @Blocking as it relates to controller routes because `@ExecuteOn` should be used instead. That use case should be deprecated and removed in a future version imo. The same goes for thread-selection: AUTO

wetted:
> I don't think we should document @Blocking as it relates to controller routes because `@ExecuteOn` should be used instead. That use case should be deprecated and removed in a future version imo. The same goes for thread-selection: AUTO

I’m OK with that either way. I already have a PR that removes the other [obsolete] mention of it in core docs.

If we want to deprecate its use and not mention in in the user guide, should we meta-annotate it `@Internal` for now?

sdelamo:
I think we should document it. I don't think we should not document a feature just because we plan to remove it.

sdelamo:
@wetted I have changed the copy. Let me know if you are ok with the changes.

wetted:
@sdelamo yes, thanks for improving/correcting/clarifying. Looks good now.

sdelamo:
@graemerocher please review it when you can.

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
```diff
@@ -0,0 +1,21 @@
+You can use the ann:core.annotation.Blocking[] annotation to mark methods as blocking.
+
+If you set `micronaut.server.thread-selection` to `AUTO`, The Micronaut Framework offloads the execution of methods annotated with `@Blocking` to the IO thread pool (See: api:io.micronaut.scheduling.TaskExecutors[]).
+
+NOTE: `@Blocking` only works if you are using `AUTO` thread selection. Micronaut Framework defaults to `MANUAL` thread selection since Micronaut 2.0. We recommend the usage of ann:scheduling.annotation.ExecuteOn[] annotation to execute the blocking operations on a different thread. `@ExecutesOn` works for both `MANUAL` or `AUTO` thread selection.
```
alvarosanchez:
```suggestion
NOTE: `@Blocking` only works if you are using `AUTO` thread selection. Micronaut defaults to `MANUAL` thread selection since Micronaut 2.0. We recommend the usage of ann:scheduling.annotation.ExecuteOn[] annotation to execute the blocking operations on a different thread pool. `@ExecutesOn` works for both `MANUAL` and `AUTO` thread selection.
```

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+There are some places where the Micronaut framework uses ann:core.annotation.Blocking[] already.
```
alvarosanchez:
```suggestion
There are some places where the Micronaut framework uses ann:core.annotation.Blocking[] internally:
```

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+|===
+|Blocking Type|Description
+
+|link:{micronautapi}http/client/BlockingHttpClient.html[BlockingHttpClient]
+| Intended for testing, and provides blocking versions for a subset of api:http.client.HttpClient[] operations.
```
alvarosanchez:
```suggestion
| Intended for testing, provides blocking versions for a subset of api:http.client.HttpClient[] operations.
```

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
```diff
@@ -0,0 +1,21 @@
...
+|link:{micronautapi}core/io/IOUtils.html[IOUtils]
+| Reads the contents of a `BufferedReader` in a blocking manner, and returns that as a `String`.
+|link:{micronautapi}context/env/BootstrapPropertySourceLocator.html[BootstrapPropertySourceLocator]
+| Resolves either remote or local api:context.env.PropertySource[] instances for the current `Environment`.
+
+|===
+
+TIP: https://micronaut-projects.github.io/micronaut-data/latest/guide/[Micronaut Data] also utilizes `@Blocking` for some transaction operations, CRUD interceptors, and repositories.
```
alvarosanchez:
```suggestion
TIP: https://micronaut-projects.github.io/micronaut-data/latest/guide/[Micronaut Data] also utilizes `@Blocking` internally for some transaction operations, CRUD interceptors, and repositories.
```

alvarosanchez (changes requested):

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
sdelamo: Micronaut should not be used as a noun to match the brand guidelines. Hence, the `Micronaut Framework`

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
sdelamo: why do you want to add internally? It is public-facing API. E.g. CrudRepository.

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
sdelamo: why do you want to add internally? It is public facing API.

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
sdelamo: I have incorporated most of your suggestions here: https://github.com/micronaut-projects/micronaut-core/pull/7228/commits/e33ddf10c1321a439bae20b0d7f40ad6ebef9878

alvarosanchez: There are gazillion places where simply "Micronaut" is used. Enforcing "Micronaut Framework" in every occurrence in the documentation is quite inappropriate, to say the least.

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
alvarosanchez: What I mean is that it's implementation details. This piece of documentation is for educating users about how certain things are _internally_ implemented. It is a public API, yes, but still it's implementation details that won't affect them nor something they need to handle specifically.

src/main/docs/guide/httpServer/serverConfiguration/threadPools/atBlocking.adoc
alvarosanchez: See above.

alvarosanchez (approved):

PR #7255, opened 2022-04-24
fix: apply injection point qualifier on BeanProvider's get call

### LINKED ISSUE #7254: BeanProvider's get by qualifier method ignores injection point qualifier
### Expected Behavior

A qualifier specified on BeanProvider's injection point should participate in bean filtering process when get by qualifier method is called.

### Actual Behaviour

It does not.

### Steps To Reproduce

This is the same as #7132, but in reverse direction.
https://github.com/micronaut-projects/micronaut-core/blob/cd6065195aff610c70865c2f9b194070dc01cbb1/inject/src/main/java/io/micronaut/inject/provider/BeanProviderDefinition.java#L82-L84
Qualifier specified on BeanProvider's injection point is just ignored.

I think that `get`'s method behavior should match `find`'s. 

### Environment Information

### Example Application

_No response_

### Version

3.4.2

Paullo612:
Fixes #7254

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/provider/BeanProviderSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/provider/BeanProviderSpec.groovy
@@ -231,30 +231,93 @@ class BeanOneTwo implements BeanNumber { }
 @jakarta.inject.Singleton
 class Test {
     public BeanProvider<BeanNumber> provider;
-    Test(@OneQualifier BeanProvider<BeanNumber> provider) {
+    public BeanProvider<BeanNumber> providerTwo;
+    Test(@OneQualifier BeanProvider<BeanNumber> provider, @TwoQualifier BeanProvider<BeanNumber> providerTwo) {
         this.provider = provider;
+        this.providerTwo = providerTwo;
     }
 }
 ''')
-        when:
+        when: 'retrieve test bean'
         def bean = getBean(context, 'test.Test')
 
-        then:
+        then: 'providerTwo successfully finds BeanOneTwo because of injection point qualifier'
         bean.provider.isPresent()
+        bean.providerTwo.find(null).get().class.name == 'test.BeanOneTwo'
 
-        when:
+        when: 'attempt to find bean through provider with null qualifier as argument'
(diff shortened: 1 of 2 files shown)


inject-java/src/test/groovy/io/micronaut/inject/provider/BeanProviderSpec.groovy
```diff
@@ -231,30 +231,91 @@ class BeanOneTwo implements BeanNumber { }
...
+        this.providerTwo = providerTwo;
     }
 }
 ''')
-        when:
+        when: 'retrieve test bean'
         def bean = getBean(context, 'test.Test')
 
-        then:
-        bean.provider.isPresent()
```
dstepanov, line 243: Why did you change this? Preferably do not modify existing checks without explanation

dstepanov (changes requested):

inject-java/src/test/groovy/io/micronaut/inject/provider/BeanProviderSpec.groovy
Paullo612: This is the test for #7133. So, I've improved `get` method test a little bit, and added the same test for `find` method. Changes in this PR touches both `get` and `find` code paths.

I can move changes to `find` method tests to separate commit if you prefer, or rollback them completely.

dstepanov: Just return that `bean.provider.isPresent()` check

Paullo612: Ok, may I force force push? Or I should do a separate commit?

dstepanov: You can force push

Paullo612: Done. Also added the same check to `get` method test.

dstepanov (approved):

sdelamo (approved):

PR #7272, opened 2022-04-27
Refactor DefaultHttpClient

yawkat:
This PR should have almost no functional changes. The main goal is reducing code duplication between different code paths in the HTTP client (streaming, proxying, normal exchange) in order to make implementation of new features easier.

The diff is quite big, but I tried to make many small commits that should be mostly mechanical (extracting a method, removing some duplication etc), so it may be easier to review the commits individually.

Targeted to 3.5.x because it's quite a big diff and fixes no bugs.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -2127,7 +2127,7 @@ public class DefaultHttpClient implements
         );
         HttpRequest nettyRequest = requestWriter.getNettyRequest();
         ChannelPipeline pipeline = channel.pipeline();
-        pipeline.addLast(ChannelPipelineCustomizer.HANDLER_MICRONAUT_HTTP_RESPONSE_FULL, new SimpleChannelInboundHandlerInstrumented<FullHttpResponse>() {
+        pipeline.addLast(ChannelPipelineCustomizer.HANDLER_MICRONAUT_HTTP_RESPONSE_FULL, new SimpleChannelInboundHandlerInstrumented<FullHttpResponse>(combineFactories()) {
             final AtomicBoolean received = new AtomicBoolean(false);
 
             @Override
@@ -2190,7 +2190,7 @@ public class DefaultHttpClient implements
                 }
             }
         });
-        pipeline.addLast(ChannelPipelineCustomizer.HANDLER_MICRONAUT_HTTP_RESPONSE_STREAM, new SimpleChannelInboundHandlerInstrumented<StreamedHttpResponse>() {
+        pipeline.addLast(ChannelPipelineCustomizer.HANDLER_MICRONAUT_HTTP_RESPONSE_STREAM, new SimpleChannelInboundHandlerInstrumented<StreamedHttpResponse>(combineFactories()) {
 
             final AtomicBoolean received = new AtomicBoolean(false);
 
@@ -2386,7 +2386,7 @@ public class DefaultHttpClient implements
             FluxSink<io.micronaut.http.HttpResponse<O>> emitter,
             Argument<O> bodyType, Argument<E> errorType) {
         ChannelPipeline pipeline = channel.pipeline();
-        final SimpleChannelInboundHandler<FullHttpResponse> newHandler = new SimpleChannelInboundHandlerInstrumented<FullHttpResponse>(false) {
+        final SimpleChannelInboundHandler<FullHttpResponse> newHandler = new SimpleChannelInboundHandlerInstrumented<FullHttpResponse>(combineFactories(), false) {
 
             AtomicBoolean complete = new AtomicBoolean(false);
             boolean keepAlive = true;
@@ -3087,7 +3087,7 @@ public class DefaultHttpClient implements
     private <V, C extends Future<V>> Future<V> addInstrumentedListener(
             Future<V> channelFuture, GenericFutureListener<C> listener
     ) {
-        InvocationInstrumenter instrumenter = combineFactories(invocationInstrumenterFactories);
+        InvocationInstrumenter instrumenter = combineFactories();
 
         return channelFuture.addListener(f -> {
             try (Instrumentation ignored = instrumenter.newInstrumentation()) {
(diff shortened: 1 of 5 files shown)


http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -3660,4 +3081,373 @@ private static class CurrentEvent {
...
+            );
+            // this onComplete call disables further parsing by HttpClientResponseException
+            errorResponse.onComplete();
+            return new HttpClientResponseException(
+                    "Error decoding HTTP error response body: " + t.getMessage(),
+                    t,
+                    errorResponse,
+                    null
+            );
+        }
```
timyates, line 3328: There's no coverage of this new method (there may not have been any previously though) 😉

timyates (approved):
LGTM 👍 

I am no Netty expert yet though

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: Good catch. Tried to hit it in a test case, but there's no easy way – `makeErrorFromRequestBody` doesn't throw, because the body decode call it does also handles any exceptions properly. I'll leave this check as-is anyway in case there's an edge case I missed where it does get triggered, and as defensive coding

http-client/src/main/java/io/micronaut/http/client/netty/ForwardingSubscriber.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.client.netty;
+
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import reactor.core.publisher.FluxSink;
+
+final class ForwardingSubscriber<T> implements Subscriber<T> {
```
graemerocher, line 31: Add `@Internal` and some javadoc on what this does

http-client/src/main/java/io/micronaut/http/client/netty/NettyFuturePublisher.java
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.http.client.netty;
+
+import io.netty.util.concurrent.Future;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+final class NettyFuturePublisher<T> implements Publisher<T> {
```
graemerocher, line 30: Add `@Internal` and some javadoc on what this does

http-client/src/main/java/io/micronaut/http/client/netty/NettyFuturePublisher.java
```diff
@@ -0,0 +1,60 @@
...
+                        }
+                    });
+                }
+            }
+
+            @Override
+            public void cancel() {
+                if (forwardCancel) {
+                    future.cancel(true);
+                }
```
graemerocher, line 67: Does this need to handle the case if the future is already completed?

http-client/src/main/java/io/micronaut/http/client/netty/NettyPromiseSubscriber.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.client.netty;
+
+import io.netty.util.concurrent.Promise;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+
+final class NettyPromiseSubscriber<T> implements Subscriber<T> {
```
graemerocher, line 29: Add `@Internal` and some javadoc describing what this does

http-client/src/main/java/io/micronaut/http/client/netty/NettyPromiseSubscriber.java
```diff
@@ -0,0 +1,49 @@
...
+    private final Promise<? super T> promise;
+    private T value;
+
+    public NettyPromiseSubscriber(Promise<? super T> promise) {
```
graemerocher: Does this need to be public?

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/NettyFuturePublisher.java
yawkat: No, cancel will just do nothing in that case

graemerocher (approved):

asantare:
hi guys, did you test the new changes with the native graalvm by chance?

In my aws lambda (with micronaut 3.6.0) the http call to `exchange` never returns: in the `AbstractMicronautLambdaRuntime#startRuntimeApiEventLoop`:
```java
 final BlockingHttpClient blockingHttpClient = endpointClient.toBlocking();
 try {
   while (loopUntil.test(runtimeApiURL)) {
      final HttpResponse<RequestType> response = blockingHttpClient.exchange(
                          HttpRequest.GET(AwsLambdaRuntimeApi.NEXT_INVOCATION_URI).header(USER_AGENT, USER_AGENT_VALUE), Argument.of(requestType));

      // Never gets here

      final RequestType request = response.body();
          
```

actually, I have a second lambda which works, so must be something else, I'll try to have a deeper look

asantare:
it happens in combination with `io.micronaut.redis:micronaut-redis-lettuce` (`io.lettuce:lettuce-core`), so probably is something in `netty` itself?
I'll try with a non-native project

yawkat:
@asantare yes we test with native as well, though the coverage may not be 100%. But I wouldn't suspect such a bug to be caused by this PR anyway.

If you can make a working example please report it as a separate issue and I can take a look.

asantare:
@yawkat thanks, indeed, it seems related to lettuce/netty/reactor (big ? from where it comes from).

Unfortunately I can't go any futher as it requires better tooling for graalvm which are enterprise  only (and the native debugging tools on Macos M1 are not that advanced I discovered), so I'm proceeding with tentatives.

Here the project I've created: https://github.com/asantare/http-client-test

It has 2 subprojects:
- a local server `local-server` listening to localhost:8000
- a micronaut app with 1 Controller `test-app`, at `http://localhost:8080/test`, the controller will just invoke the local server and return its response.

I'm building the `test-app` locally without docker to speed things up, using just `nativeCompile` and then using the browser to call the endpoint.

If you try to comment out the micronaut-redis and lettuce-core deps and re-build and re-run it will work, or, in alternative if you force the 5.2.2.RELEASE of lettuce it'll work too (it seems it is the last version without the `reflect-config.json`).

yawkat:
i cant reproduce it. but please make a new issue, it is unrelated to this

asantare:
Thank you for checking it

asantare:
it seems to be [a graalvm issue](https://github.com/oracle/graal/issues/4809)

PR #7274, opened 2022-04-27
fix: @Replaces with qualifier doesn't work (fix #7261)

### LINKED ISSUE #7261: @Replaces with qualifier doesn't work
### Expected Behavior

Using `@Replaces(bean = ..., qualifier=...)` works as expected

### Actual Behaviour

The `qualifier` isn't processed by `DefaultBeanContext#filterReplacedBeans`, what could lead to errors like `No bean of type [com.bug.micronaut.example.SomeInterface] exists for the given qualifier`.
`@Named` works fine

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

https://github.com/ivyazmitinov/replace-qualifier-bug-demo

### Version

3.4.2

ivyazmitinov:
fix #7261

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -2477,54 +2477,78 @@ public class DefaultBeanContext implements InitializableBeanContext {
             Class replacedBeanType = replacesAnnotation.classValue().orElse(getCanonicalBeanType(replacingCandidate));
             final Optional<String> named = replacesAnnotation.stringValue(NAMED_MEMBER);
             final Optional<AnnotationClassValue<?>> qualifier = replacesAnnotation.annotationClassValue(QUALIFIER_MEMBER);
-            Optional<Class<?>> factory = replacesAnnotation.classValue("factory");
-            if (LOG.isDebugEnabled()) {
-                if (factory.isPresent()) {
-                    LOG.debug("Bean [{}] replaces existing bean of type [{}] in factory type [{}]",
-                        replacingCandidate.getBeanType(), replacedBeanType, factory.get());
+
+            if (named.isPresent() && qualifier.isPresent()) {
+                throw new ConfigurationException("Both \"named\" and \"qualifier\" should not be present: " + replacesAnnotation);
+            }
+
+            if (named.isPresent()) {
+                final String name = named.get();
+                if (qualifiedByNamed(definitionToBeReplaced, replacedBeanType, name)) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Bean [{}] replaces existing bean of type [{}] qualified by name [{}]",
+                            replacingCandidate.getBeanType(), definitionToBeReplaced.getBeanType(), name);
+                    }
+                    return true;
                 } else {
-                    LOG.debug("Bean [{}] replaces existing bean of type [{}]",
-                        replacingCandidate.getBeanType(), replacedBeanType);
+                    return false;
                 }
             }
+
+            if (qualifier.isPresent()) {
+                final AnnotationClassValue<?> qualifierClassValue = qualifier.get();
+                if (qualifiedByQualifier(definitionToBeReplaced, replacedBeanType, qualifierClassValue)) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Bean [{}] replaces existing bean of type [{}] qualified by qualifier [{}]",
+                            replacingCandidate.getBeanType(), definitionToBeReplaced.getBeanType(), qualifierClassValue);
+                    }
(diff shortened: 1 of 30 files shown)


ivyazmitinov:
@dstepanov, Fixed

ivyazmitinov:
I've also got a bit of the refactoring fever and

- Changed `Replaces#qualifier` type to the `Class<? extends Annotation> instead of an array. It required a bit of a workaround, but it is a lesser evil, IMO
- Added exception when both `named` and `qualifier` present 

Both changes are meant to streamline the API

ivyazmitinov:
@graemerocher, I am a bit confused regarding tests: there is a `io.micronaut.docs.replaces.ReplacesTest`, which I've updated with the new use-cases, but there are also a bunch of `io.micronaut.inject.qualifiers.replaces.ReplacesSpec`s. Should I add tests to them as well, or what's best to do?

graemerocher:
the first one is only for the documentation, the last one is where new and existing tests should be updated yes

ivyazmitinov:
@graemerocher Done :)

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2319,74 +2322,93 @@ private <T> void filterReplacedBeans(BeanResolutionContext resolutionContext, Co
...
-                                    return true;
-                                }
-                            }
-                        } else {
-                            if (comparisonFunction.apply(beanType)) {
-                                return true;
-                            }
-                        }
-                    }
+            if (named.isPresent() || qualifier.isPresent()) {
```
dstepanov: I would prefer not to have so many `if else` statements, when a block returns the value you don't need to have `else`. Also, I would extract the values only when they are needed:
```
final Optional<String> named = ...
LOG ...
if (named...

Optional<Class<?>> factory
LOG...
if (factory.isPresent() 
etc
```

inject/src/main/java/io/micronaut/context/annotation/Replaces.java
```diff
@@ -54,12 +56,19 @@
      *
      * @return The qualifier
      */
-    Class<? extends Annotation>[] qualifier() default {};
+    Class<? extends Annotation> qualifier() default ReplacesQualifierDefaultValue.class;
```
dstepanov: You can use simple `Annotation.class`

dstepanov (changes requested):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2319,74 +2322,93 @@ private <T> void filterReplacedBeans(BeanResolutionContext resolutionContext, Co
...
+                if (named.isPresent() && qualifier.isPresent()) {
```
dstepanov: Here you check if optional is present and then you continue with optional in your code. Please first check if `named` qualifies then if `qualifier` qualifies etc. Preferably reduce the number of lamdas.

ivyazmitinov:
@dstepanov, Fixed all the issues. 
Also, improved the logging a bit: added additional logs when bean is replaced by any qualifier, and now it writes the log only when a bean was actually replaced.

ivyazmitinov:
@dstepanov, kind reminder :)

dstepanov:
@ivyazmitinov Thanks it looks good! One more thing I noticed is that you can replace the stream with a simple loop in this code:

```java
return replacementTypes.stream()
            .filter(isNotTheSameDefinition(definitionToBeReplaced))
            .filter(isNotProxy(definitionToBeReplaced))
            .anyMatch(checkIfReplaces(definitionToBeReplaced, annotationMetadata));
```
We try to use streams as minimal as possible in the core.

Also, now we have the Sonar which unfortunately doesn't run for contributions but it likes to complain about the code style:
```java
if (...) {
 return ...
} else { // not needed else
 return or throw
}
````
It would be nice if you can remove those elses.

ivyazmitinov:
@dstepanov 
Thanks, will do :) 
And I intentionally refactored the original loop into the stream with fancy predicates, wan't aware about "no streams" policy :) 
What is the issue with them, if you don't mind me asking?

dstepanov:
We try to not have them in hot paths like a startup (JVM take some time to optimize them) or request processing to reduce allocations.

ivyazmitinov:
@dstepanov, should be ok now

dstepanov (approved):

ivyazmitinov:
@sdelamo, could you, by any chance, take a look at the PR?

sdelamo:
@ivyazmitinov can you merge 3.5.x into the branch and address the test failures?

ivyazmitinov:
@sdelamo, done. GraalVM CE CI seems to be unrelated to my changes

Issue #7266, opened 2022-04-26
Beans get duplicated with kapt and keepJavacAnnotationProcessors = true


korri123:
### Expected Behavior

Java and Kotlin code should work in harmony, where only one bean (and one path mapping) per Java class gets generated. This bug happened when adding Kotlin code to an existing Micronaut Java app containing Java Micronaut Controllers.

`keepJavacAnnotationProcessors` in kapt is enabled so Lombok in Java still compiles.
If `keepJavacAnnotationProcessors = true` is emitted from `build.gradle` then everything works as expected.

### Actual Behaviour

Duplicate beans and endpoints get instantiated so endpoint returns Bad Request as Micronaut doesn't know which endpoint to choose. Internally in Micronaut the classloader returns resources both from `build/classes/java` and `build/tmp/kapt3/classes`, which Micronaut creates beans from and therefore duplicate controllers.

### Steps To Reproduce

1. Clone https://github.com/korri123/MicronautErrorTest
2. Run the test file named `src/test/kotlin/com/example/DemoTest.kt` which should succeed as it asserts the error message happens.

### Environment Information

- Ubuntu 20.4
- JDK 17

### Example Application

https://github.com/korri123/MicronautErrorTest

### Version

3.4.2

ojebari-ma:
Can you please confirm that this is still an issue in the latest version of Micronaut?

korri123:
It's been 3 years so I'm not sure, and tbh, I forgot about this ticket. There is a reproducer repo attached which could be updated to latest version to be tested.

graemerocher:
Triage: an existing PR, https://github.com/micronaut-projects/micronaut-core/pull/7276, directly addresses this KAPT/Javac duplicate-bean and duplicate-route problem by filtering duplicate generated candidates. Keeping the issue open pending that PR; the regression should cover mixed Java/Kotlin compilation with `keepJavacAnnotationProcessors = true` and ensure only one definition/path is retained.

PR #7276, opened 2022-04-27
#7266 Filter out duplicate beans to fix Kapt issue

korri123:
Addresses https://github.com/micronaut-projects/micronaut-core/issues/7266


graemerocher:
I think this may already be addressed, but differently in the 3.5.x branch

korri123:
Okay awesome, hopefully that can be verified so it doesn't get forgotten in-case it wasn't fixed :smile:

PR #7277, opened 2022-04-28
Bump micronaut-liquibase to 5.3.0

micronaut-build:


sdelamo:
@timyates can you remove the `managed-liquibase` from here. Liquibase has a bom and he exposed the managed version https://github.com/micronaut-projects/micronaut-liquibase/blob/master/gradle/libs.versions.toml#L2

PR #7278, opened 2022-04-28
Intercept suspend functions correctly

bene1618:
When a Kotlin suspend method is intercepted, the associated Continuation should be intercepted. Without this interception, some of the Kotlin coroutine mechanisms cease to work, for instance thread management.

For example:
```
@MicronautTest
class ClientDispatcherTest(
    private val testClient: TestClient
) : StringSpec() {

    init {
        "Execute client call and note the thread" {
            println("Thread before: ${Thread.currentThread().name}")
            testClient.test()
            println("Thread after: ${Thread.currentThread().name}")
        }
    }

    @Client("/test")
    interface TestClient {
        @Get
        suspend fun test(): String
    }

    @Controller("/test")
    class TestController {
        @Get
        fun test(): String = "Test"
    }
}
```

Expected behavior would be to switch back to the original thread:
```
Thread before: pool-3-thread-1
Thread after: pool-3-thread-1
```

Actual output (before the fix):
```
Thread before: pool-3-thread-1
Thread after: default-nioEventLoopGroup-1-2
```

Thus, we see that the code after the suspend call is executed in a Netty event loop thread. This is especially harmful if we have few processor cores (say, one) available, because in this case the Netty event loop thread pool size is only two. This can easily lead to the two threads blocking each other, for example by executing another client call after the first client call.

Also Kotlin tracing relies on intercepting the continuation, and probably there are even more issues.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/internal/intercepted/KotlinInterceptedMethod.java
+++ b/aop/src/main/java/io/micronaut/aop/internal/intercepted/KotlinInterceptedMethod.java
@@ -19,6 +19,7 @@ import io.micronaut.aop.Interceptor;
 import io.micronaut.aop.MethodInvocationContext;
 import io.micronaut.aop.util.CompletableFutureContinuation;
 import io.micronaut.aop.util.DelegatingContextContinuation;
+import io.micronaut.aop.util.KotlinInterceptedMethodHelper;
 import io.micronaut.core.annotation.Experimental;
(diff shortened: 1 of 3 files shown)


dstepanov:
please target 3.5

dstepanov:
Can you also test that the Thread after is the same?

dstepanov (approved):

PR #7283, opened 2022-04-29
Spaces in URI paths and various parsing issues

hollingsworthd:
Work in progress to address micronaut-projects/micronaut-core#7288


sdelamo:
Consider if using [UriBuilder](https://docs.micronaut.io/latest/api/io/micronaut/http/uri/UriBuilder.html) can simplify this PR.

hollingsworthd:
When URIs have no path but they do have query params or a fragment, we do not add a slash after the authority (host+port). So we can generate URIs like `https://www.example.com#frag` or `https://www.example.com?q=val`. From my reading the most compatible URIs would be `https://www.example.com/#frag` and `https://www.example.com/?q=val`.

This might break some apps because conceivably some of our users have relied on the current URI format but I think this has not been part of our contract, and it violates various URI specs.

Issue #7257, opened 2022-04-25
Error finalizing type visitor: URI has an authority component


sgammon:
### Expected Behavior

I was upgrading Micronaut to the latest version (`3.4.2` at time of writing), and added `micronaut-graal` where it was not present before. I'm using Bazel. I was expecting the Micronaut processors to pick it up and use it as normal.

### Actual Behaviour

I get an exception during the build:

```
ERROR: /Users/sam/Workspace/elide/java/elide/runtime/jvm/BUILD.bazel:68:18: Compiling Java headers java/elide/runtime/jvm/libAssetManager-hjar.jar (1 source file) and running annotation processors (TypeElementVisitorProcessor, AggregatingTypeElementVisitorProcessor, PackageConfigurationInjectProcessor, BeanDefinitionInjectProcessor, ServiceDescriptionProcessor) failed: (Exit 1): sandbox-exec failed: error executing command 
  (cd /private/var/tmp/_bazel_sam/e3e455f40daf5c614f7b80b355398a63/sandbox/darwin-sandbox/528/execroot/elide && \
  exec env - \
    LC_CTYPE=en_US.UTF-8 \
    PATH=/bin:/usr/bin:/usr/local/bin \
    TMPDIR=/var/folders/pd/mk9dc77s7cx3j21n6pdt7v200000gn/T/ \
[...]

PR #7298, opened 2022-05-02
Use Path for reflect.json access in GraalTypeElementVisitor

yawkat:
Transforming to File is unnecessary here, and may be leading to #7257

### DIFF
--- a/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
+++ b/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
@@ -36,11 +36,11 @@ import io.micronaut.inject.visitor.VisitorContext;
 import io.micronaut.inject.writer.GeneratedFile;
 import jakarta.inject.Inject;
 
-import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.Writer;
+import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -302,19 +302,19 @@ public final void finish(VisitorContext visitorContext) {
...
-        File userReflectJsonFile = projectDir
-                .map(projectPath -> Paths.get(projectPath.toString(), BASE_REFLECT_JSON).toFile())
+        Path userReflectJsonFile = projectDir
+                .map(projectPath -> projectPath.resolve(BASE_REFLECT_JSON))
                 .orElse(null);
 
-        if (userReflectJsonFile != null && userReflectJsonFile.exists()) {
-            try {
-                json = MAPPER.readValue(userReflectJsonFile, new TypeReference<List<Map>>() {
+        if (userReflectJsonFile != null && Files.exists(userReflectJsonFile)) {
```
timyates, line 313: Do we have anything that tests this?  Any other projects?  I don't think this has changed the coverage, as it probably wasn't tested before.

timyates (approved):
Not sure how this fixes the original issue (as the commit they made to Turbine seems a much more likely candidate), but I tend to prefer Path over File, so lgtm 👍

yawkat:
@timyates if I understand correctly, some sort of VFS is involved, which would break if using File. but this is pure speculation, it just seems like the most likely place where our code is sensitive to an invalid URI.

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
yawkat: i dont know a way to build a test that would make using path only here behave any differently, unfortunately.

graemerocher (commented):
One note about this fix is this code has been completely re-written in 3.5.x so we may have merge challenges when merging up.

yawkat:
Oh right, it generates a Feature now doesnt it? in that case it's easier to just wait for that change

Issue #1112, opened 2019-01-15
Unix socket support


hc-codersatlas:
Hi, for docker, testing and other reasons (e.g. deploying behind a reverse proxy) can unix domain socket support be added? Netty should have support for this. Couldn't find it's available in the documentation.

Thanks!

PR #7300, opened 2022-05-02
More powerful listener config for HTTP server, domain socket support

yawkat:
This commit adds a configuration structure to configure individual netty HTTP listeners with individual host, port, and SSL settings. The existing configuration options (eg dual-stack, port, ...) remain (and are not deprecated, since they're simpler), and are mapped to the new system.

This patch also adds support for binding to domain sockets for the epoll and kqueue backends. Domain socket support (like the embedded channel introduced in 1bc6e6aaea4e922df4028d668a971e5c358597d5) is not 100% supported, because methods like HttpRequest.getRemoteAddress make no sense with such channels.

Adds #1112


sdelamo:
@yawkat is it possible to address some of the Sonar Smells?

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -545,6 +545,22 @@ public void setPcapLoggingPathPattern(String pcapLoggingPathPattern) {
...
+        return listeners;
+    }
+
+    /**
+     * Set the explicit netty listener configurations, or {@code null} if they should be implicit.
+     * @param listeners The listeners
+     */
+    public void setListeners(Map<String, NettyListenerConfiguration> listeners) {
+        this.listeners = listeners;
+    }
```
graemerocher: I'm just wondering will this work in native? I think it might be better if `NettyListenerConfiguration` was an inner class annotated with `@EachProperty("listeners")`?

I think these being POJOs will trigger to code path to use MapToObjectConverter which we are trying to get away from

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
yawkat: i dont understand, what does making it an inner class change?

graemerocher: As it is written now, Jackson will be used to convert from a `Map` to an `Object` (see attached debug session screenshot):

Screenshot 2022-05-06 at 14 56 19

This would require reflection, unless the type is annotated with `@Introspected` then perhaps not. If you write it as a inner class you benefit from several advantages:

1. No reflection config or `@Introspected` is needed
2. The IDE will present code completion to the user when they write listener configurations since we will generate the necessary configuration metadata

yawkat: @graemerocher like this?

yawkat: Hmm the change added a spurious test failure, seems like EachProperty does not preserve order?

graemerocher: hmm the functionality is dependent on order of YAML?

yawkat: mostly for old api, getPort returns the first listener

graemerocher: I am not sure this is good. Some config forms (like java properties) we support don't support ordering at all and we have never required ordering in any config we have had before.

Can you rewrite the test to not depend on order?

yawkat: ok will do. it means theres no way to influence what getPort will return if you use the new listener config, but i guess it's not a big deal

graemerocher:
@yawkat can you resolve the conflicts?

graemerocher (approved):

graemerocher:
@yawkat one thing, there doesn't appear to be documentation for this new configuration method. Could you add some before release?

yawkat:
added docs, how do i test that they display properly?

graemerocher:
You can run `./gradlew publishGuide` or `./gradlew pG` for short and open the generated HTML

graemerocher:
if you could fix the code smells that would be good as well

sdelamo (approved):

Issue #7288, opened 2022-03-01
HTTP Client - `UriSyntaxException: Illegal character in path` when URI contains spaces


cheezy64:
### Expected Behavior

When building a URI, spaces in path parameters should be escaped or encoded.  E.g.
`http://localhost:34218/api/system/this has a space in it`
should be converted to
`http://localhost:34218/api/system/this%20has%20a%20space%20in%20it`

### Actual Behaviour

Spaces are not escaped, and we get `UriSyntaxException: Illegal character in path` when building the URI.

It looks like the we should add an `encode` here:
https://github.com/micronaut-projects/micronaut-core/blob/7e74302d743940ec145f002066fdccfe5a126bac/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java#L331

### Steps To Reproduce

Relevant Dependencies
```kotlin
Micronaut 3.2.3
Micronaut JAX-RS 3.1.0
```

Call that reproduces the issue
```kotlin
WebTarget target = httpClient.target("http://localhost:34218/api/system/this%20has%20a%20space%20in%20it");
```

Here is a stack trace:
```
io.micronaut.http.exceptions.UriSyntaxException: Illegal character in path at index 39: http://localhost:34218/api/system/this has a space in it
	at io.micronaut.http.uri.DefaultUriBuilder.build(DefaultUriBuilder.java:264)
	at io.micronaut.jaxrs.runtime.core.JaxRsUriBuilder.build(JaxRsUriBuilder.java:228)
	at org.glassfish.jersey.client.JerseyWebTarget.getUri(JerseyWebTarget.java:109)
	at org.glassfish.jersey.client.JerseyWebTarget.request(JerseyWebTarget.java:191)
	at org.glassfish.jersey.client.JerseyWebTarget.request(JerseyWebTarget.java:36)
```

### Environment Information

Windows 10
Java 17

### Version

3.2.3

yawkat:
i think this is actually a jax-rs issue

sdelamo:
If I understand @cheezy64 the issue is the expect path params to be Url encoded: 

```java
String path = "this has a space in it";
assertEquals("/this%20has%20a%20space%20in%20it",
                UriBuilder.of("").path(path).build().toString());
``` 

There is an easy workaround which is to encode the the value before passing it to path. 

```java
String path = "this has a space in it";
assertEquals("/this%20has%20a%20space%20in%20it",
[...]

PR #7302, opened 2022-05-02
fix: attempts to encode paths

sdelamo:
Fixes: https://github.com/micronaut-projects/micronaut-core/issues/7288

Currently this works: 

```java
    @Test ✅
    void pathVariableWithSpacesAndPathWithExpand() {
        String path = "this has a space in it";
        String expected = "/this%20has%20a%20space%20in%20it";
        assertEquals(expected,
                UriBuilder.of("").path("{pathwithspaces}")
                    .expand(CollectionUtils.mapOf("pathwithspaces", path)).toString());
    }
```

this fails: 

```java
@Test ❌
    void pathVariableWithSpacesAndPath() {
        String path = "this has a space in it";
        String expected = "/this%20has%20a%20space%20in%20it";
        assertEquals(expected, UriBuilder.of("")
                .path(path)
                .build().toString());
    }
```

@hollingsworthd has a [work in progress with bigger refactor of UriBuilder](https://github.com/micronaut-projects/micronaut-core/pull/7283) to support the handling of fragments and strings with parts of them as templates and parts of them not as templates. I don't think it is worth it to go that rabbit whole for now and I think this PR will fix the most common issue which users use `path` and they expected it to be Url encoded but it is not unless they use a template.

### DIFF
--- a/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
+++ b/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
@@ -331,13 +331,11 @@ class DefaultUriBuilder implements UriBuilder {
             String pathStr = path.toString();
             if (isTemplate(pathStr, values)) {
                 pathStr = UriTemplate.of(pathStr).expand(values);
-            } else {
+            } else if (pathStr.chars().filter(ch -> ch == '/').count() <= 1) {
                 final String templateVariable = "pathString";
-                if (pathStr.chars().filter(ch -> ch == '/').count() <= 1) {
-                    pathStr = (pathStr.charAt(0) == '/') ?
-                            UriTemplate.of("/{" + templateVariable + "}").expand(Collections.singletonMap(templateVariable, pathStr.substring(1))) :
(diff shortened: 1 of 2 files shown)


yawkat:
This change is not backwards-compatible. This test worked before and now fails:

```groovy
    void "path encodes old"() {
        given:
        String path = "this%20has%20a%20space%20in%20it"
        String expected = "/this%20has%20a%20space%20in%20it";

        expect:
        expected == UriBuilder.of("")
                .path(path)
                .build()
                .toString()
    }
```

imo this is a pretty significant change and can't go into 3.x

sdelamo:
> This change is not backwards-compatible. This test worked before and now fails:
> 
> ```groovy
>     void "path encodes old"() {
>         given:
>         String path = "this%20has%20a%20space%20in%20it"
>         String expected = "/this%20has%20a%20space%20in%20it";
> 
>         expect:
>         expected == UriBuilder.of("")
>                 .path(path)
>                 .build()
>                 .toString()
>     }
> ```
> 
> imo this is a pretty significant change and can't go into 3.x

Agree. Let's move it to 4. The users have a workaround.

PR #7305, opened 2022-05-03
Fix proxy with qualifiers and refactor qualifiers code

dstepanov:
The refactoring fixes check if the qualifier is `Named` in some examples:
```
Qualifier<T> declaredQualifier = beanDefinition.getDeclaredQualifier();
boolean propagateQualifier = beanDefinition.isProxy() && declaredQualifier instanceof Named;
```
Now the code is more readable and understandable than before.


inject/src/main/java/io/micronaut/inject/qualifiers/Utils.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Qualifier utils.
+ *
+ * @author Denis Stepanov
+ * @since 3.5.0
+ */
+final class Utils {
```
graemerocher: Can you name this something more specific like `QualifierUtils`

graemerocher: Also annotate with `@Internal`

graemerocher (changes requested):

graemerocher (approved):

sdelamo (approved):

PR #7308, opened 2022-05-04
DefaultBeanContext#getBeansOfType should return programmatically registered singletons

### LINKED ISSUE #7309: DefaultBeanContext#getBeansOfType doesn't return singleton of same type but with different qualifier
### Expected Behavior

Expect `DefaultBeanContext#getBeansOfType` to return multiple beans if multiple beans were registered programatically as singleton using a different qualifier.

### Actual Behaviour

Only one bean is returned event though multiple beans were registered

### Steps To Reproduce

In PR #7308 one test has been modified to register multiple singletons. 
The test fails without the change made in the `DefaultBeanContext`.

### Environment Information

Tested on mac os `12.3` with java `11.0.14`

### Example Application

_No response_

### Version

3.4.2

akorompai:
Fixes #7309


dstepanov (approved):

.github/release-drafter.yml
```diff
@@ -1,21 +0,0 @@
-name-template: Micronaut $NEXT_PATCH_VERSION
```
sdelamo: why are you deleting this file in this PR? this is unrelated to Fixes https://github.com/micronaut-projects/micronaut-core/issues/7309 ?

sdelamo (changes requested):

.github/release-drafter.yml
sdelamo: I restored it here: https://github.com/micronaut-projects/micronaut-core/pull/7308/commits/8a3964715a856258df475b85e3587670e3ee4b16

sdelamo:
@akorompai thanks for the contribution!

PR #7310, opened 2022-05-05
Bump micronaut-cache to 3.4.0

micronaut-build:


timyates:
@melix Not sure why micronaut-build removed the caffeine version I added it back in with https://github.com/micronaut-projects/micronaut-core/pull/7310/commits/300713984c02f86dfd352c944dd9575afd9a3fc0

I just released micronaut-cache 3.4.0 which included a version catalog, and caffeine as a managed version.  Is this the same as the grpc issue?  Can you think of the solution?  My brain is having a slow morning 😞

melix:
What do you mean that `micronaut-build` removed caffeine?

timyates:
The original commit https://github.com/micronaut-projects/micronaut-core/pull/7310/commits/305c907370fccb13e7c54e7931d09ca2a0abd5e7 set `caffeine` to `""`

sdelamo:
same issue here: https://github.com/micronaut-projects/micronaut-core/pull/7277

sdelamo:
I am not sure if caffeine is used in core if it is not needed, we should remove it. since it is defined in the cache bom: https://github.com/micronaut-projects/micronaut-cache/blob/master/gradle/libs.versions.toml#L3

sdelamo:
I think it is removing the caffeine ☕️version because of https://github.com/micronaut-projects/micronaut-cache/blob/master/gradle.properties#L18

melix:
Oh I see, that's probably a bug in the dependency upgrade workflow/task. However we should get rid of this in the template and `micronaut-build` (the project, not the bot), since we moved to renovatebot.

PR #7342, opened 2022-05-10
Correct how proxies and its target beans are destroyed

dstepanov:
This corrects how proxies are destroyed. For target bean proxies we need to ignore the actual proxy and just destroy the target. I had to introduce the same interfaces from AOP for the Inject module to be able to recognize which bean is an introspected proxy and added a new method to check if the target is initialized for the cases when it's lazy.

I had to add a new method to find the bean registration by bean definition in `CustomScope` to properly implement destroying the scoped instance.

This PR also changes to allow `PostCreate` and `PreDestroy` to be on the same method to have the same behavior as CDI.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/Intercepted.java
+++ b/aop/src/main/java/io/micronaut/aop/Intercepted.java
@@ -15,11 +15,13 @@
  */
 package io.micronaut.aop;
 
+import io.micronaut.inject.proxy.InterceptedBean;
+
 /**
  * An interface implemented by generated proxy classes.
  *
  * @author Graeme Rocher
  * @since 1.0
  */
-public interface Intercepted extends io.micronaut.inject.proxy.Intercepted {
+public interface Intercepted extends InterceptedBean {
 }
--- a/aop/src/main/java/io/micronaut/aop/InterceptedProxy.java
+++ b/aop/src/main/java/io/micronaut/aop/InterceptedProxy.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.aop;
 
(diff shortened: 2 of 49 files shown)


aop/src/main/java/io/micronaut/aop/InterceptedProxy.java
```diff
@@ -25,13 +25,23 @@
...
+    @Override
     T interceptedTarget();
 
+    /**
+     * Check if the proxy has the target cached before calling {@link #interceptedTarget()}.
+     *
+     * @return true if the target is cached
+     */
+    @Override
+    default boolean hasCachedInterceptedTarget() {
```
graemerocher, line 46: add `@since`

aop/src/main/java/io/micronaut/aop/Intercepted.java
```diff
@@ -21,5 +21,5 @@
  * @author Graeme Rocher
  * @since 1.0
  */
-public interface Intercepted {
+public interface Intercepted extends io.micronaut.inject.proxy.Intercepted {
```
graemerocher: Perhaps name the new interface something like `InterceptedBean` to avoid confusion.

graemerocher (approved):
Looks good, just two minor suggestions

graemerocher:
Can you look at the binary incompatibilities? It is probably the new defaults methods. You can just add these to the accepted-api-changes.json

sdelamo (approved):

sdelamo:
@dstepanov could you address some of the code smells reported by sonar? Some of them are just swapping the expectation and value in the Junit tests.

graemerocher:
there are still some api changes that need to be accepted

PR #7349, opened 2022-05-11
Don't destroy dependent custom scope beans

dstepanov:
This reverts some of the changes done before. The custom scope bean shouldn't be removed when the parent-bean is discarded.


dstepanov:
Added some refactoring to reduce duplicities, you can review commits

graemerocher (approved):

inject-java/src/test/groovy/io/micronaut/inject/lifecycle/proxytargetbeanwithpredestroy/CustomScopeScope.java
```diff
@@ -68,7 +67,6 @@ public <T> Optional<BeanRegistration<T>> findBeanRegistration(BeanDefinition<T>
         return findByBeanDefinition(beanDefinition);
     }
 
-    @NotNull
```
sdelamo, line 71: you could replace it with our `@Nullable` if you want.

sdelamo (approved):

sdelamo (commented):
There is unused `private final DefaultBeanContext beanContext` in `@SingletonScope`. is that expected? Should the `SingletonScope` constructor be changed os that it does not take `beanContext`?

dstepanov:
Removed it

PR #7354, opened 2022-05-12
Fix metrics CORS problems

n0tl3ss:
This pull requests addresses the https://github.com/micronaut-projects/micronaut-micrometer/issues/397. There is an issue when the CORS OPTIONS requests are fired the requests attributes were not populated and the metrics were dumping the raw value of the path instead of templated one. 

This pull requests adds attributes into OPTIONS request if there is anyRoute that can be matched with different HTTP verb. There is also a test that covers that scenario.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/OptionsRequestAttributesSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/OptionsRequestAttributesSpec.groovy
@@ -1,0 +1,63 @@
+package io.micronaut.http.server.netty
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpAttributes
+import io.micronaut.http.HttpMethod
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpStatus
+import io.micronaut.http.MutableHttpResponse
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Filter
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.client.HttpClient
+import io.micronaut.http.client.exceptions.HttpClientResponseException
+import io.micronaut.http.filter.HttpServerFilter
+import io.micronaut.http.filter.ServerFilterChain
+import io.micronaut.runtime.server.EmbeddedServer
+import jakarta.inject.Singleton
+import org.reactivestreams.Publisher
+import org.spockframework.util.Assert
+import spock.lang.Specification
+
+
+class OptionsRequestAttributesSpec extends Specification {
+
+    def 'test OPTIONS rquests attributes'() {
+        EmbeddedServer server = ApplicationContext.run(EmbeddedServer, ['spec.name': 'OptionsRequestAttributesSpec'])
+        def ctx = server.applicationContext
+        HttpClient client = ctx.createBean(HttpClient, server.getURL())
+
+        when:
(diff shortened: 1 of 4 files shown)


yawkat:
doing this in the cors filter seems like a hack, especially since non-cors options requests can exist too and cause the same issue. I'd much rather have a fix in RoutingInboundHandler

yawkat (approved):

graemerocher (approved):

yawkat:
@timyates points out that this test is still failing: https://github.com/Robin479/micronaut-bugs/blob/main/http-metrics/src/test/java/de/ooch/micronaut/bugs/RunawayClientMetricsTest.java

Looking at https://github.com/micronaut-projects/micronaut-micrometer/issues/397 , it seems like the client test should be a separate issue and can be addressed in a separate PR (though I'm not 100% sure we can fix it in a sane way)

timyates (approved):

graemerocher:
@yawkat I discussed with @n0tl3ss that this can only really be addressed in the client through new APIs that allow passing a template. He will write a proposal for a future feature to solve this use case.

yawkat:
why is static analysis failing? there's no sonar bot info

timyates:
It's an external branch so doesn't get the keys...

One sec, I'll push a "local" branch

timyates:
@yawkat Sonar says https://github.com/micronaut-projects/micronaut-core/pull/7354/files#diff-9af0f6dd8d3c36389a0ff250f900fd204afbe901b57f4e2d23028ecdad048c56R457 should be parameterized

https://github.com/micronaut-projects/micronaut-core/pull/7361#issuecomment-1125964922

n0tl3ss:
> @yawkat Sonar says https://github.com/micronaut-projects/micronaut-core/pull/7354/files#diff-9af0f6dd8d3c36389a0ff250f900fd204afbe901b57f4e2d23028ecdad048c56R457 should be parameterized
> 
> [#7361 (comment)](https://github.com/micronaut-projects/micronaut-core/pull/7361#issuecomment-1125964922)

is changing `HttpRequest` to `HttpRequest<?>` sufficient enough?

yawkat:
yes

n0tl3ss:
> yes

Just added it :) Thanks guys

timyates:
Cool, re-running in the local branch 👍

timyates:
image

💚

yawkat:
Thanks for the help @timyates !

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/OptionsRequestAttributesSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+import io.micronaut.runtime.server.EmbeddedServer
+import jakarta.inject.Singleton
+import org.reactivestreams.Publisher
+import org.spockframework.util.Assert
+import spock.lang.Specification
+
+
+class OptionsRequestAttributesSpec extends Specification {
+
+    def 'test OPTIONS rquests attributes'() {
```
donbeave, line 26: Typo in a word `rquests`

olvesh:
Sorry, it doesn't look like this is completely fixed:
Running version 3.7.0 of Micronaut I see this when getting a 401
image

 

Trying a GET against a controller not configured for GET (and a auth filter before)
Could this be alleviated by some change at our end (not really experienced in Micronaut specifics I must say)

n0tl3ss:
Please try to use `micronaut-micrometer` version v4.8.2. It should be solved. Thanks :)

olvesh:
Thanks, that did the trick!

PR #7364, opened 2022-05-13
Bump micronaut-sql to 4.4.0

micronaut-build:


gradle/libs.versions.toml
```diff
@@ -7,7 +7,7 @@ blaze = "1.6.4"
 caffeine = "2.9.3"
 compile-testing = "0.19"
 geb = "3.4.1"
-hibernate = "5.5.9.Final"
+hibernate = ""
```
sdelamo: Should this be "5.6.8.Final" as:
 https://github.com/micronaut-projects/micronaut-sql/blob/master/gradle/libs.versions.toml#L4 ?

If Micronaut SQL is publishing a BOM ? Why do we need to ahve the hibernate version here?

gradle/libs.versions.toml
```diff
@@ -37,7 +37,7 @@ wiremock = "2.27.2"
 # Versions which start with managed- are managed by Micronaut in the sense
 # that they will appear in the Micronaut BOM as <properties>
 #
-managed-commons-dbcp = "2.9.0"
+managed-commons-dbcp = ""
```
sdelamo: Shoudl this be left to 2.9.0 or an we remove it https://github.com/micronaut-projects/micronaut-sql/blob/master/gradle/libs.versions.toml#L7

sdelamo (changes requested):

sdelamo:
@timyates can you look into this?

sdelamo (approved):

PR #7367, opened 2022-05-13
CORS preflight request should return 404 when route doesn't exists

n0tl3ss:
With this change the CORS preflight requests will return 200 only if there is an existing route and the HTTP method that can handle provided method that is specified inside the preflight request.

Added:
- AVAILABLE_HTTP_METHODS attribute on the request object which is populated only if http request method is `OPTIONS ` and contains all HTTP METHODS on provided route.
- new test that covers scenario

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
@@ -135,6 +135,8 @@ import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static io.micronaut.http.HttpAttributes.AVAILABLE_HTTP_METHODS;
+
 /**
  * Internal implementation of the {@link io.netty.channel.ChannelInboundHandler} for Micronaut.
  *
@@ -344,6 +346,7 @@ class RoutingInBoundHandler extends SimpleChannelInboundHandler<io.micronaut.htt
                     .collect(Collectors.toList());
             if (!anyUriRoutes.isEmpty()) {
                 setRouteAttributes(request, anyUriRoutes.get(0));
+                request.setAttribute(AVAILABLE_HTTP_METHODS, anyUriRoutes.stream().map(UriRouteMatch::getHttpMethod).collect(Collectors.toList()));
             }
         }
 
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -26,6 +26,8 @@ import io.micronaut.runtime.server.EmbeddedServer
 import io.micronaut.web.router.RouteMatch
 import io.micronaut.web.router.Router
 import org.apache.http.client.utils.URIBuilder
+import spock.lang.AutoCleanup
+import spock.lang.Shared
 import spock.lang.Specification
 
 import java.util.stream.Collectors
@@ -34,6 +36,7 @@ import static io.micronaut.http.HttpHeaders.*
 
 class CorsFilterSpec extends Specification {
 
+    @Shared @AutoCleanup
     EmbeddedServer embeddedServer = ApplicationContext.run(EmbeddedServer)
 
     CorsFilter buildCorsHandler(HttpServerConfiguration.CorsConfiguration config) {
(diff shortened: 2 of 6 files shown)


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
```diff
@@ -15,16 +15,27 @@
...
+import org.apache.http.client.utils.URIBuilder
 import spock.lang.Specification
 
+import java.util.stream.Collectors
+
 import static io.micronaut.http.HttpHeaders.*
 
 class CorsFilterSpec extends Specification {
 
+    EmbeddedServer embeddedServer = ApplicationContext.run(EmbeddedServer)
```
yawkat, line 40: construct this in the test methods, and properly shut it down after the test is done

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -150,7 +151,7 @@ protected Optional<MutableHttpResponse<?>> handleRequest(HttpRequest request) {
                     }
                 }
 
-                if (preflight) {
+                if (preflight && RouteMatchUtils.findRouteMatch(request).isPresent()) {
```
yawkat: i don't think this is enough, what if this is a preflight for a POST request for example, but there's no GET controller. this will yield a 404 even though the actual cors request would succeed

yawkat (commented):
do we need this change to be compliant? imo this opens a whole can of worms wrt checking from preflight which requests will succeed and which won't

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
n0tl3ss: I think that you are right. I added new test case that covers that situation and I improved the solution. Currently from the test case you can see that it will return 405 Method Not Allowed if route exists but provided HTTP method doesn't. Do you think that we should return some other status code in that case?

yawkat (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -344,6 +346,7 @@ protected void channelRead0(ChannelHandlerContext ctx, io.micronaut.http.HttpReq
                     .collect(Collectors.toList());
             if (!anyUriRoutes.isEmpty()) {
                 setRouteAttributes(request, anyUriRoutes.get(0));
+                request.setAttribute(AVAILABLE_HTTP_METHODS, anyUriRoutes.stream().map(UriRouteMatch::getHttpMethod).collect(Collectors.toList()));
             }
```
kevin-wise, line 350: This does not consider routes from filters, meaning if you have a filter which handles a route that nothing else does, CORS no longer works with it

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -142,15 +144,17 @@ protected Optional<MutableHttpResponse<?>> handleRequest(HttpRequest request) {
...
-                    HttpMethod methodToMatch = preflight ? headers.getFirst(ACCESS_CONTROL_REQUEST_METHOD, CONVERSION_CONTEXT_HTTP_METHOD).orElse(requestMethod) : requestMethod;
                     if (allowedMethods.stream().noneMatch(method -> method.equals(methodToMatch))) {
                         return Optional.of(HttpResponse.status(HttpStatus.FORBIDDEN));
                     }
                 }
 
-                if (preflight) {
+                Optional<? extends ArrayList<HttpMethod>> availableHttpMethods = (Optional<? extends ArrayList<HttpMethod>>) request.getAttribute(AVAILABLE_HTTP_METHODS, new ArrayList<HttpMethod>().getClass());
+
+                if (preflight && availableHttpMethods.isPresent() && availableHttpMethods.get().stream().anyMatch(method -> method.equals(methodToMatch))) {
```
kevin-wise, line 157: for a purely filter-served route, this returns false when it should be true

kevin-wise (commented):
Seeing a regression in this PR

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
kevin-wise: reported as https://github.com/micronaut-projects/micronaut-core/issues/7934

PR #7393, opened 2022-05-16
fix java.lang.InstantiationError with abstract beans

graemerocher:
This fixes the failures happening at https://gitlab.com/micronaut-projects/micronaut-graal-tests/-/jobs/2459394445


dstepanov (approved):

dstepanov:
I think you should also exclude abstract beans in `BeanDefinitionInjectProcessor#isDeclaredBean`

graemerocher:
not sure, since that would prevent them generating bean metadata which is a breaking change

sdelamo (approved):

Issue #7158, opened 2022-03-28
Allow changing Host header of proxied HTTP request


yawkat:
### Feature description

Also see spring http client.

jameskleeh:
If you remove the host header manually it will be recalculated, however some may want to store the original one in X-Forwarded-For or one of the other strategies

yawkat:
@jameskleeh right now, for proxied requests, we always recalculate the `Host`: https://github.com/micronaut-projects/micronaut-core/blob/29d395f6b74ee7355a146c70aca219d6bfce3704/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java#L3007

graemerocher:
Triage: this remains a valid HTTP-client proxy enhancement. The current proxy path recalculates `Host`, while callers may need to preserve or intentionally set the upstream host independently from the original request host. Suggested solution: define an explicit proxy-host/forwarded-host policy, make it configurable without silently overriding a caller-provided value, and add tests for direct, proxied, and forwarded-header requests.

Issue #3759, opened 2020-07-20
ProxyHttpClient implementation in DefaultHttpClient does not rewrite Host header


jacobmims:
### Task List

- [x] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [x] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create a `ProxyHttpClient`
2. Call the `proxy` method on a URI with a `host()` value that is different from the originating service.
3. Note the `Host` header still references the originating service URI's `host()` value.
    - Known issues with Cloud Providers:
       - GCP AppEngine will not accept requests with a `Host` header different from the service being called.

### Expected Behaviour

The `proxy` method should rewrite the `Host` header to the `host()` value of the URI being called by `proxy()`.

### Actual Behaviour

The `Host` header of the request made by the `proxy` method is set to the `host()` value the originating service's URI.

### Environment Information

- **Operating System**: MacOS 10.15.5
- **Micronaut Version:** 2.0.0
- **JDK Version:** 11

Relevant documentation: https://docs.micronaut.io/latest/guide/httpClient.html#proxyClient

jacobmims:
**Update**: I was able to workaround the issue by removing the header when constructing the request. This should probably be standard behavior for the proxy method, instead of having to do it when mutating the request.

```java
private final String host = "some-gcp-project-name.appspot.com";
 
@Override
  protected Publisher<MutableHttpResponse<?>> doFilterOnce(
      HttpRequest<?> request, ServerFilterChain chain) {
    HttpRequest<?> proxyRequest = request.mutate()
        .uri(b -> b.scheme(HttpRequest.SCHEME_HTTPS).port(443).host(host))
        .headers(h -> h.remove("Host"));

    return Publishers.map(client.proxy(proxyRequest), response -> response);
  }
```

miguelaferreira:
If it's worth anything towards validation, I've just faced this same issue while building a web proxy using the ProxyHttpClient.

cpriisholm:
Can confirm that I've seen load balancers and http server return various errors (503, 421...) unless the above workaround is applied.

PR #7396, opened 2022-05-16
Add option to keep request Host header in ProxyHttpClient

yawkat:
Without this patch, the Host header was always cleared and then recomputed from the request URI. This PR adds an overload with an additional boolean flag that if set to true, does not clear the Host header.

Normally I'm not a big fan of such magic flags from a design pov, but I don't see a better solution.

Adds #7158

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/ProxyRequestOptions.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/ProxyRequestOptions.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


timyates:
FYI: We changed to be "on-spec" (and rewrite the header) here https://github.com/micronaut-projects/micronaut-core/pull/6739 after a user raised an issue here https://github.com/micronaut-projects/micronaut-core/issues/3759

yawkat:
Yes I think the rewrite is usually the correct behavior

graemerocher:
Can we turn the argument into a `ProxyOptions` type with a builder style? Adding a single flag now means that if we ever need another one if the future it will create a mess of the API

http-client-core/src/main/java/io/micronaut/http/client/ProxyRequestOptions.java
```diff
@@ -0,0 +1,108 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.http.client;
+
+import java.util.Objects;
+
+/**
+ * Further options for {@link ProxyHttpClient} when handling proxy requests.
+ */
+public final class ProxyRequestOptions {
```
graemerocher, line 23: Add `@since` and author tags

graemerocher (changes requested):

timyates (approved):
👍 

Can't think of a way to increase the coverage apart from maybe checking equals() works

But that's probably chasing a statistic, and not adding much value

graemerocher (approved):

PR #7411, opened 2022-05-18
build: Switch to using versions from boms, instead of custom versions

timyates:
These versions are used in tests, but can be provided by declaring a platform
dependency on the upstream bom that also pulls them in.

The reasoning for doing this is that renovate-bot should spam us less with
dependency updates as these will be handled by the modules themselves.

All the tests still passed without junit4 or powermock, so I removed these entirely.

This also includes a couple of test fixes where static counters were not being reset in case of a retry

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,6 @@
 [versions]
 asm = "9.2"
 awaitility = "4.0.3"
-aws-java-sdk-lambda = "1.12.220"
 bcpkix = "1.69"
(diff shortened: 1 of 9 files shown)


dstepanov:
Can we reference module’s dependencies using the gradle catalog and remove those entries completely? Mysql can be added using some mn_sql catalog

sdelamo (approved):

sdelamo:
@dstepanov I merged this, and we can tackle further improvements in other PRs.

PR #7484, opened 2022-05-23
Regression, do not check if bean is a singleton for a custom scope

dstepanov:
The problem occurs for the pre 3.5 compiled beans that got the singleton scope inherited from the factory.


graemerocher:
Is there a way to test this?

dstepanov:
Nope

graemerocher (approved):

PR #7486, opened 2022-05-23
Fix resolving `@Parameter`s in factories

dstepanov:


dstepanov:
The main problem is that every qualifier used to be of interface `Named` and `@Primary` somehow resolved to `Named` and then back to `@Primary`

graemerocher (approved):

sdelamo (approved):

PR #7498, opened 2022-05-25
New API for context propagation

dstepanov:
Initial version of the propagation API.

Main points:
- The propagated context is always immutable and consist of different propagated elements similar to Kotlin Coroutines context
- Propagated element can implement `ThreadPropagatedContextElement` to setup/restore thread locals
- The propagated state is never captured but the context needs to be extended and pushed down the downstream
- The capturing API is not needed anymore, this should eliminate a lot of overhead from the reactive code
- The context is automatically propagated to the Kotlin coroutines
- In the Reactor context should be propagated as whole with utility method to extract/extend it (The Reactor context needs to be modified manually)
- The implementation is using try-resources to propagate the contest without extra overhead, this might change in the future to support scoped local, but at this moment, I think this is the best solution.

I want to merge this first PR to have the orther work mergable.

Next PRs will remove the existing instrumentation propagation, add docs etc.

There are already examples how it should be used:
- Tracing https://github.com/micronaut-projects/micronaut-tracing/pull/281
- Micronaut Data https://github.com/micronaut-projects/micronaut-data/pull/2189 (This is a big PR because of the changes in the TX management and removal of forked Spring TX code)


context-propagation/src/main/java/io/micronaut/context/propagation/instrument/reactor/ReactorInstrumentation.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * On scheduler hook for the thread to be aware of {@link PropagatedContext}.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Requires(classes = {Flux.class, Schedulers.Factory.class})
+@Context
+@Internal
+class ReactorInstrumentation {
```
graemerocher: make `final`

context-propagation/src/main/java/io/micronaut/context/propagation/instrument/rxjava2/RxJava2Instrumentation.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * On scheduler hook for the thread to be aware of {@link PropagatedContext}.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Requires(classes = {RxJavaPlugins.class})
+@Context
+@Internal
+class RxJava2Instrumentation {
```
graemerocher: make `final`

context-propagation/src/main/java/io/micronaut/context/propagation/instrument/rxjava3/RxJava3Instrumentation.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * On scheduler hook for the thread to be aware of {@link PropagatedContext}.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Requires(classes = {RxJavaPlugins.class})
+@Context
+@Internal
+class RxJava3Instrumentation {
```
graemerocher: make `final`

core-reactive/src/main/java/io/micronaut/core/async/propagation/ReactivePropagation.java
```diff
@@ -0,0 +1,121 @@
...
+import reactor.core.CoreSubscriber;
+import reactor.util.context.Context;
+
+/**
+ * Reactive propagation of {@link PropagatedContext}.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public class ReactivePropagation {
```
graemerocher: annotate with `@Experimental`

core-reactive/src/main/kotlin/io/micronaut/core/async/propagation/KotlinCoroutinePropagation.kt
```diff
@@ -0,0 +1,48 @@
...
+import io.micronaut.core.propagation.PropagatedContext
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Kotlin Coroutine propagation utils.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+class KotlinCoroutinePropagation private constructor() {
```
graemerocher, line 29: annotate with `@Internal`

core/src/main/java/io/micronaut/core/propagation/PropagatedContext.java
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * The propagation across different threads using the context which is immutable and can be extended/reduced by different elements.
+ * Each element can be a simple data structure carrying its state across the threads, or it can implement {@link ThreadPropagatedContextElement}
+ * for use-cases when thread-local values needs to be updated.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Experimental
+public interface PropagatedContext {
```
graemerocher, line 36: add nullability annotations to all arguments

http/src/main/java/io/micronaut/http/context/ServerHttpRequestContext.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Optional;
+
+/**
+ * Http request propagation.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Experimental
+public final class ServerHttpRequestContext implements PropagatedContextElement {
```
graemerocher: could be a record

runtime/src/test/groovy/io/micronaut/runtime/executor/ExecutorServiceConfigSpec.groovy
```diff
@@ -33,6 +34,8 @@ import java.util.concurrent.ThreadPoolExecutor
  * @author Graeme Rocher
  * @since 1.0
  */
+@Ignore("Instrumented Executor services break this test")
```
graemerocher: need to fix this test

graemerocher (changes requested):

graemerocher (approved):

core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
```diff
@@ -0,0 +1,212 @@
...
+    private static final PropagatedContextImpl EMPTY = new PropagatedContextImpl(Collections.emptyList());
+
+    private final List<PropagatedContextElement> elements;
+    private final boolean containsThreadElements;
+
+    private PropagatedContextImpl(List<PropagatedContextElement> elements) {
+        this.elements = elements;
+        boolean containsThreadElements = false;
+        for (PropagatedContextElement element : elements) {
+            if (element instanceof ThreadPropagatedContextElement) {
```
yawkat, line 62: this check is probably slow

dstepanov: What do you suggest?

core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @Override
+    public List<PropagatedContextElement> getAllElements() {
+        return elements;
+    }
+
+    @Override
+    public Scope propagate() {
+        PropagatedContextImpl prevCtx = THREAD_CONTEXT.get();
```
yawkat, line 172: maybe instead of try-finally, this should be a function that accepts a lambda, so that maybe we can eventually replace the ThreadLocal with a ScopedValue

core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
dstepanov: It think it would be better to use the array here for sure

yawkat (approved):

core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
yawkat: i think it would be best to store a flag alongside each PropagatedContextElement whether the particular element is a ThreadPropagatedContextElement

yawkat: and have separate factory methods for PropagatedContext for the different elements. i.e. differentiate using data (the flag), not polymorphism, for best perf

yawkat (approved):

PR #7536, opened 2022-05-31
Fix incremental compilation of configuration properties metadata

### LINKED ISSUE #7534: Support of Gradle incremental compilation when having more that 1 @ConfigurationProperties annotated classes
### Feature description

When having multiple `@ConfigurationProperties` annotated classes, as per an example app: https://github.com/stepanv/micronaut-full-rebuild-sample-app/tree/main/src/main/java/com/example 

with `micronaut.processing.incremental=true` enabled, the incremental compilation gets broken
```
Task ':compileJava' is not up-to-date because:
  Input property 'stableSources' file /Users/stepan.vavra/Repos/other/micronaut-test-slow-compilation/src/main/java/com/example/MyService.java has changed.
Created classpath snapshot for incremental compilation in 0.008 secs.

Full recompilation is required because the generated resource 'META-INF/spring-configuration-metadata.json in CLASS_OUTPUT' must have exactly one originating element, but had 2. Analysis took 0.013 secs.

```

It would be great if Micronaut was able to support incremental compilation even in such case.

As a workaround one can extract those annotated classes to a different gradle module where incremental compilation is not that needed (thx @graemerocher for that hint).

graemerocher:
Fixes #7534


melix (approved):

inject-java/src/main/java/io/micronaut/annotation/processing/ConfigurationMetadataProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.ServiceLoader;
+import java.util.Set;
+
+/**
+ * Handles Configuration metadata generation.
+ *
+ * @author graemerocher
+ * @since 3.5.1
+ */
+public class ConfigurationMetadataProcessor extends AbstractInjectAnnotationProcessor {
```
sdelamo, line 39: should we annotate this class with `@Internal`?

sdelamo (approved):

inject-java/src/main/java/io/micronaut/annotation/processing/ConfigurationMetadataProcessor.java
sdelamo: we consider `inject-java` to be internal.

PR #7557, opened 2022-06-07
Fix inability to apply AOP advice to BeanElementBuilder created beans

graemerocher:
For CDI we need to be able import beans that have AOP advice applied to them, turns out that doesn't work with the current `BeanElementBuilder` API. This PR fixes that.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/internal/intercepted/InterceptedMethodUtil.java
+++ b/aop/src/main/java/io/micronaut/aop/internal/intercepted/InterceptedMethodUtil.java
@@ -24,11 +24,15 @@ import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.AnnotationUtil;
 import io.micronaut.core.annotation.AnnotationValue;
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.type.ReturnType;
 
 import java.util.List;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Future;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 /**
  * The {@link InterceptedMethod} utils class.
@@ -99,20 +103,10 @@ public final class InterceptedMethodUtil {
      * @param annotationMetadata The annotation metadata
      * @return True if it does
      */
-    public static boolean hasAroundStereotype(AnnotationMetadata annotationMetadata) {
-        if (annotationMetadata == null) {
-            return false;
-        }
-        if (annotationMetadata.hasStereotype(Around.class)) {
-            return true;
-        } else if (annotationMetadata.hasStereotype(AnnotationUtil.ANN_INTERCEPTOR_BINDINGS)) {
-            return annotationMetadata.getAnnotationValuesByType(InterceptorBinding.class)
-                    .stream().anyMatch(av ->
-                            av.enumValue("kind", InterceptorKind.class).orElse(InterceptorKind.AROUND) == InterceptorKind.AROUND
-                    );
-        }
-
-        return false;
+    public static boolean hasAroundStereotype(@Nullable AnnotationMetadata annotationMetadata) {
+        return hasAround(annotationMetadata,
+                annMetadata -> annMetadata.hasStereotype(Around.class),
+                annMetdata -> annMetdata.getAnnotationValuesByType(InterceptorBinding.class));
     }
 
     /**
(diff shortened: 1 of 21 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -185,6 +185,8 @@ public <BT extends BeanType<Object>> Stream<BT> reduce(Class<Object> beanType, S
 
     final Map<BeanIdentifier, BeanRegistration<?>> singlesInCreation = new ConcurrentHashMap<>(5);
 
+    Set<Map.Entry<Class<?>, List<BeanInitializedEventListener>>> beanInitializedEventListeners;
```
dstepanov, line 188: Is it needed?

dstepanov (approved):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
graemerocher: checkstyle violation

aop/src/main/java/io/micronaut/aop/internal/intercepted/InterceptedMethodUtil.java
```diff
@@ -92,4 +93,46 @@ public static io.micronaut.core.annotation.AnnotationValue<?>[] resolveIntercept
         }
         return AnnotationUtil.ZERO_ANNOTATION_VALUES;
     }
+
+    /**
+     * Does the given metadata have AOP advice declared.
+     * @param annotationMetadata The annotation metadata
+     * @return True if it does
+     */
+    public static boolean hasAroundStereotype(AnnotationMetadata annotationMetadata) {
```
sdelamo: annotate parameter as `@Nullable`

aop/src/main/java/io/micronaut/aop/internal/intercepted/InterceptedMethodUtil.java
```diff
@@ -92,4 +93,46 @@ public static io.micronaut.core.annotation.AnnotationValue<?>[] resolveIntercept
...
+
+        return false;
+    }
+
+    /**
+     * Does the given metadata have declared AOP advice.
+     * @param annotationMetadata The annotation metadata
+     * @return True if it does
+     */
+    public static boolean hasDeclaredAroundAdvice(AnnotationMetadata annotationMetadata) {
```
sdelamo: annotate parameter as `@Nullable`

aop/src/main/java/io/micronaut/aop/internal/intercepted/InterceptedMethodUtil.java
```diff
@@ -92,4 +93,46 @@ public static io.micronaut.core.annotation.AnnotationValue<?>[] resolveIntercept
...
+     * @return True if it does
+     */
+    public static boolean hasAroundStereotype(AnnotationMetadata annotationMetadata) {
+        if (annotationMetadata == null) {
+            return false;
+        }
+        if (annotationMetadata.hasStereotype(Around.class)) {
+            return true;
+        } else if (annotationMetadata.hasStereotype(AnnotationUtil.ANN_INTERCEPTOR_BINDINGS)) {
+            return annotationMetadata.getAnnotationValuesByType(InterceptorBinding.class)
```
sdelamo: If understand this logic correctly, the following code:

```java
return annotationMetadata.getAnnotationValuesByType(InterceptorBinding.class)
    .stream().anyMatch(av ->
        av.enumValue("kind", InterceptorKind.class).orElse(InterceptorKind.AROUND) == InterceptorKind.AROUND
);
```

if `annotationMetadata.getAnnotationValuesByType(InterceptorBinding.class)` returns a list of two items. 

- if the first item in the list returns for `av.enumValue("kind", InterceptorKind.class)` =  empty optional
- and the second item returns for `av.enumValue("kind", InterceptorKind.class)` == Optional(`INTRODUCTION`)

The above code will return `true`. Is that intended?

Should the above snippet be replaced by?

```java
Optional<InterceptorKind> interceptorKind = parseInterceptorKind(annotationMetadata.getAnnotationValuesByType(InterceptorBinding.class));
            if (!interceptorKind.isPresent()) {
                return true;
            }
            return interceptorKind.map(it -> it == InterceptorKind.AROUND).orElse(false);
...
}

    @NonNull
    private static Optional<InterceptorKind> parseInterceptorKind(@NonNull List<AnnotationValue<InterceptorBinding>> annotationValueList) {
        return annotationValueList
                .stream()
                .map(av -> av.enumValue("kind", InterceptorKind.class))
                .filter(Optional::isPresent) //TODO JDK 9 Min. Use .flatMap(Optional::stream())
                .map(Optional::get)
                .findFirst();
    }
```

graemerocher: I didn't change this code as part of this PR, all I did was move to a common utility class. I don't think it makes sense to make changes in this area. But yes the default `InterceptorBinding` value if nothing is specified in `AROUND` so the code is correct and we shouldn't replace or change anything with this code.

sdelamo (approved):

PR #7577, opened 2022-06-14
Changed Random to SecureRandom.

JoeBeeContrast:
If developers used the random property feature for cyrptographic purposes, there is a potential that that crypto could be broken. This PR modifies the PropertySourcePropertyResolver class to change Random to SecureRandom


sdelamo:
> graphic purposes, there is a potential that that crypto could be broken. This PR modifies the PropertySourcePropertyResolver class to change Random to SecureRandom

can you elaborate?

Moreover, you will need to sign the CLA please.

JoeBeeContrast:
Hi 
CLA is signed.
So the problem with using java.util.Random is the values the Random generates are not random as it uses the current system time as the initial seed. So an attacker is able to guess what values were generated. If the "random" property was used for any cryptographic or other security sensitive purpose, an attacker knowing what that initial value was would potentially be able to break that crypto or bypass whatever security control was using that random value.

java.security.SecureRandom provides non deterministic random values which are suitable for use in crypto.

JoeBeeContrast:
@sdelamo can you rerun the build. It seems to have failed due to a auth issue.
_Execution failed for task ':sonarqube'.
> Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator_

Rather than a failure due to my code change.

JoeBeeContrast:
Hi @sdelamo  or anyone on the team. The above checks failed due to a build issue not related t the PR. can someone please retrigger the above build?

JoeBeeContrast:
Fixed the issue. It was just an unused Random import.

sdelamo (approved):

sdelamo:
@JoeBeeContrast thanks for the contribution.

PR #7520, opened 2022-05-28
Added saving information about the annotations on enum

altro3:
For this https://github.com/micronaut-projects/micronaut-openapi/issues/471

PR #7580, opened 2022-06-16
Added saving information about the annotations on enum

altro3:
For this https://github.com/micronaut-projects/micronaut-openapi/issues/471


altro3:
This is a copy of the pull request - https://github.com/micronaut-projects/micronaut-core/pull/7520 which was closed by mistake

dstepanov:
This is causing class cast exception because `EnumConstantElement` is now returned from: `classElement.getEnclosedElements(ElementQuery.ALL_FIELDS.includeEnumConstants().includeHiddenElements())`
which only supported to return `FieldElement`. I think we can have `EnumConstantElement` extends `FieldElement`.

PR #7584, opened 2022-06-16
Improve support for runtime bean registration

graemerocher:
Currently it isn't possible to register a bean at runtime that is lazily created by the container nor is it possible to register beans in different scopes. This is needed for improvements to Spring bean import. I have submitted against 3.5.x because if it doesn't go into that branch then we have a chicken and egg problem with the 3.6 release in terms of getting improvements into that release.


inject/src/main/java/io/micronaut/context/BeanDefinitionReferenceRegistry.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2017-2020 original authors
```
burtbeckwith: 2022

burtbeckwith (approved):

dstepanov (approved):

dstepanov:
I wonder if it still makes sense to have the separate concept of bean references and bean definitions. I feel like we should just stick with BeanDefinitions in v4 and reduce the number of extra classes generated.

graemerocher:
Perhaps not, we can revisit that in 4.0

sdelamo (approved):

graemerocher:
hold off merging have a few changes to add

sdelamo:
I am not sure about this going into 3.5.x. I think we should target 3.6.x. Once 3.6.0 is out we can release a next minor of Micronaut Spring referencing core 3.6.0. thoughts @graemerocher ?

sdelamo:
> hold off merging have a few changes to add

I have set the PR to draft.

sdelamo (approved):
Looks good to me. I have created a PR with small changes 

https://github.com/micronaut-projects/micronaut-core/pull/7606

PR #7589, opened 2022-06-17
Document ContextConifigurer

### LINKED ISSUE #7471: Document @ContextConfigurer
### Issue description

We currently recommend people to configure default environment such `dev` with: 

```java
public class Application {
    @ContextConfigurer
    public static class DefaultEnvironmentConfigurer implements ApplicationContextConfigurer {
        @Override
        public void configure(@NonNull ApplicationContextBuilder builder) {
            builder.defaultEnvironments(defaultEnvironment);
        }
    }
        Micronaut.run(Application.class, args);
    }
}
``` 

but the docs still reflect: 

```
    Micronaut.build(args)
             .mainClass(Application.class)
             .defaultEnvironments("dev")
             .start();
}
```

We should document it and explain why it is the preferred way (e.g. visibility to AOT).

timyates:
Fixes #7471

### DIFF
--- a/src/main/docs/guide/config/environments.adoc
+++ b/src/main/docs/guide/config/environments.adoc
@@ -49,19 +49,33 @@ snippet::io.micronaut.docs.context.env.DefaultEnvironmentSpec[tags="disableEnvDe
 
 == Default Environment
 
-Micronaut supports the concept of one or many default environments. A default environment is one that is only applied if no other environments are explicitly specified or deduced. Environments can be explicitly specified either through the application context builder `Micronaut.build().environments(...)`, through the `micronaut.environments` system property, or the `MICRONAUT_ENVIRONMENTS` environment variable. Environments can be deduced to automatically apply the environment appropriate for cloud deployments. If an environment is found through any of the above means, the default environment will *not* be applied.
+Micronaut supports the concept of one or many default environments.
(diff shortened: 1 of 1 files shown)


src/main/docs/guide/config/environments.adoc
```diff
@@ -49,20 +49,34 @@ snippet::io.micronaut.docs.context.env.DefaultEnvironmentSpec[tags="disableEnvDe
...
+Micronaut supports the concept of one or many default environments.
+A default environment is one that is only applied if no other environments are explicitly specified or deduced.
+Environments can be explicitly specified either through the application context builder `Micronaut.build().environments(...)`, through the `micronaut.environments` system property, or the `MICRONAUT_ENVIRONMENTS` environment variable.
+Environments can be deduced to automatically apply the environment appropriate for cloud deployments.
+If an environment is found through any of the above means, the default environment will *not* be applied.
```
timyates, line 56: This was me moving sentences on to separate lines in the asciidoc.  It makes no difference to the generated documentation, but I find it much easier to read

melix (approved):
We should probably add a section which explains that any customization to the application context must be done via the context configurer so that they are visible to AOT.

sdelamo (approved):

PR #7590, opened 2022-06-17
fix: Fix typo

engarik:
Just a small typo fix in docs. V6 engine should have 6 cylinders.


sdelamo:
Thanks for the PR @engarik I have created a PR which includes your fix and it updates also Kotlin and Groovy examples. Moreover, it adds tests. 

https://github.com/micronaut-projects/micronaut-core/pull/7600

Issue #7421, opened 2022-05-19
Add ServiceId information to HttpClientException and HttpClientResponseException to be able to know which service is causing exception


segocago:
### Feature description

Currently, HttpClientExceptions does not have any indication of which HttpClient actually caused the exception. As having many HttpClients is very common, these exception that does not tell which client caused the exception is not very useful. We see the exceptions in the logs but can't tell which HttpClient and which service that HttpClient is calling caused the issue.

Add serviceId information to HttpClientException and HttpClientResponseException that are being thrown by declarative clients so that we can know which service is causing the exception.

DefaultHttpClient that is built for declarative http clients already have the serviceId information but it is not passing this id to thrown exceptions.

![image](https://user-images.githubusercontent.com/22752426/169372107-08950432-e3fe-446e-a0b3-eb2f03cd936c.png)

Adding this id to exceptions should not require much work, i could provide a PR if you can point me in the right direction of how it should be added.

PR #7609, opened 2022-06-23
Save service id alongside http client exceptions

yawkat:
This patch adds a serviceId field to HttpClientException, and sets that field with information from the client where possible.

The service ID is prepended to the message of the exception (does not apply when there is a special JsonError message in the body as in HttpResponseException), and it's accessible with a getter.

Resolves https://github.com/micronaut-projects/micronaut-core/issues/7421

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
@@ -24,6 +24,8 @@ import io.micronaut.http.exceptions.HttpException;
  * @since 1.0
  */
 public class HttpClientException extends HttpException {
+    private String serviceId;
+    private boolean serviceIdLocked;
 
     /**
      * @param message The message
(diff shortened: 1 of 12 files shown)


timyates (approved):

http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
```diff
@@ -50,5 +52,26 @@ public HttpClientException(String message, Throwable cause, boolean shared) {
         if (!shared) {
             throw new IllegalArgumentException("shared must be true");
         }
+        serviceIdLocked = true;
+    }
+
+    public final String getServiceId() {
+        return serviceId;
+    }
```
graemerocher, line 68: Needs javadoc with `@since` and nullable annotations. Perhaps should in fact return `Optional<String>`

http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
```diff
@@ -50,5 +52,26 @@ public HttpClientException(String message, Throwable cause, boolean shared) {
...
+
+    public final void setServiceId(String serviceId) {
+        if (serviceIdLocked) {
+            throw new IllegalStateException("Service ID already set");
+        }
+        this.serviceId = serviceId;
+        serviceIdLocked = true;
+    }
```
graemerocher, line 84: Needs javadoc with `@since` and `@throws` with an explanation it can only be set once

graemerocher: Might even be worth considering making this `@Internal`

graemerocher (changes requested):

sdelamo (approved):

http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
sdelamo: @yawkat 

Can you annotate the return type with `@Nullabe` 

```java
 * @since 3.6.0
 */
@Nullable
public final String getServiceId() {
        return serviceId;
    }
```

or 

```java
 * @since 3.6.0
 */
@NonNull
public final Optional<String> getServiceId() {
        return Optional.ofNullable(serviceId);
    }

```

PR #7610, opened 2022-06-23
Clarify that `endpoints.all.path` configuration property needs to end with a '/' (it fails otherwise).

### LINKED ISSUE #7605: health endpoint status DOWN does not return 503 when app has context and custom endpoint path
### Expected Behavior

When health check fails (status DOWN) the http status code should be 503

### Actual Behaviour

It appears that defining micronaut.server.context-path and endpoints.all.path is not catered for in filter HealthResultFilter. As such the fiter does not fire and when status is DOWN health returns http status 200

### Steps To Reproduce

Start sample application 
./gradlew run

health will always return DOWN as per https://github.com/rbhambhra/micronaut-health-endpoint/blob/master/src/main/java/com/example/MyHealthIndicator.java#L18

Get health to see http status 200 
curl -v http://localhost:8080/mycontext/admin/health

With workaround https://github.com/rbhambhra/micronaut-health-endpoint/blob/master/src/main/java/com/example/MyHealthResultFilter.java#L8 we get 503

MICRONAUT_ENVIRONMENTS=myfilter ./gradlew run

### Environment Information

Windows 11
openjdk 17.0.2 2022-01-18

### Example Application

https://github.com/rbhambhra/micronaut-health-endpoint

### Version

3.5.2

wetted:
closes #7605

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
@@ -96,7 +96,7 @@ public class EndpointDefaultConfiguration {
     }
 
     /**
-     * The endpoints base path. It must include a trailing '/'. Default value ({@value #DEFAULT_ENDPOINT_BASE_PATH}).
+     * The endpoints base path. It must include a leading and trailing '/'. Default value ({@value #DEFAULT_ENDPOINT_BASE_PATH}).
      * @param path The path
      */
     public void setPath(String path) {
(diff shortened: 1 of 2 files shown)


sdelamo (commented):
I am not sure this is correct: 

I wrote a sample app. see: 

https://github.com/micronaut-projects/micronaut-guides/pull/1011

`/endpoints` seems to be working for me (see `HealthPathTest`test). 

Am I missing something?

sdelamo (approved):

PR #7625, opened 2022-06-30
New server channel customization API

yawkat:
The changes in #6842 planned for 4.0 will break compatibility of `ChannelPipelineCustomizer` users that modify the server pipeline. This patch introduces a new `ServerNettyCustomizer` API that is more generic and, importantly, allows for more granularity when waiting for the stages of pipeline construction.

Also included in this patch is an extensive test suite that tests the logbook customizer example from the documentation in all scenarios we support (raw http1, tls alpn, and h2c support).


yawkat:
draft PR while I confirm compatibility with 4.0

yawkat:
Confirmed compatibility with #6842, pending https://github.com/netty/netty/pull/12546

test-suite-groovy/src/test/groovy/io/micronaut/docs/netty/LogbookPipelineCustomizer.groovy
```diff
@@ -30,6 +29,7 @@ class LogbookPipelineCustomizer
         ChannelPipelineCustomizer customizer = event.bean
 
         if (customizer.serverChannel) { // <2>
+            /*
             customizer.doOnConnect( { ChannelPipeline pipeline ->
                 pipeline.addAfter(
                         ChannelPipelineCustomizer.HANDLER_HTTP_SERVER_CODEC,
```
timyates, line 35: Do we want to leave these commented sections? I assume they'll appear in the docs?

timyates (approved):
Looks good as far as I can tell 😳

Do we want to mention this change in `https://github.com/micronaut-projects/micronaut-core/blob/3.5.x/src/main/docs/guide/appendix/breaks.adoc` for v4?

I assume the old way was broken? 🤔

yawkat:
it should definitely go in the 4.0 breaking changelog, but the old pipeline customizer still works as before for 3.x

test-suite-groovy/src/test/groovy/io/micronaut/docs/netty/LogbookPipelineCustomizer.groovy
yawkat: 🤷

graemerocher:
Can we convert this to draft since it is not really supposed to go into 3.6.x?

yawkat:
@graemerocher this is supposed to go into 3.6, if there's still time. it just adds the new api so people can migrate before 4.0

http-server-netty/src/main/java/io/micronaut/http/server/netty/ServerNettyCustomizer.java
```diff
@@ -0,0 +1,117 @@
...
+ * narrower scope are created from the root customizers through
+ * {@link #specializeForChannel specialization} when new netty channels are created by the
+ * server, e.g. when a client connects. These specialized customizers will then receive
+ * notifications when a certain step in the pipeline setup is reached, so they can do their own
+ * customizations on the channel and the channel pipeline they received in the specialization step.
+ *
+ * @since 3.6.0
+ * @author yawkat
+ */
+public interface ServerNettyCustomizer {
```
graemerocher: Rename to `NettyServerCustomizer` which is more consistent with the rest of the code base IMO

http-server-netty/src/main/java/io/micronaut/http/server/netty/ServerNettyCustomizer.java
```diff
@@ -0,0 +1,117 @@
...
+     * listener channel at all. For HTTP/2, each HTTP stream may get its own channel that is
+     * specialized from the overall connection channel. And finally, HTTP/3 support has to use
+     * datagram channels instead of the socket-based ones.
+     *
+     * @param channel The new channel to specialize for.
+     * @param role The role (or scope) of the channel.
+     * @return The new customizer, or {@code this} if no specialization needs to take place.
+     */
+    @NonNull
+    default ServerNettyCustomizer specializeForChannel(@NonNull Channel channel, @NonNull ChannelRole role) {
```
graemerocher: does it make sense to future this API by adding a context parameter that allows looking up the channel and role? This allows adding other lookup methods to the context object without introducing a new method signature

http-server-netty/src/main/java/io/micronaut/http/server/netty/CompositeServerNettyCustomizer.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        if (specialized == null) {
+            return this;
+        } else {
+            return new CompositeServerNettyCustomizer(specialized);
+        }
+    }
+
+    private void forEach(Consumer<ServerNettyCustomizer> consumer) {
+        for (ServerNettyCustomizer member : members) {
```
graemerocher: what about ordering? Is ordering or the ability to order customisers important?

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/ServerNettyCustomizer.java
yawkat: i think in that case i would just add a new specialize method, such as `default ServerNettyCustomizer specializeForX() { return this; }`

http-server-netty/src/main/java/io/micronaut/http/server/netty/CompositeServerNettyCustomizer.java
yawkat: Hmm it depends on what people actually do with these customizers. If they add new handlers at the same position, order may matter. How would you implement this, an optional `int order` parameter in the registry?

graemerocher: Probably just by using `OrderUtil` to sort the list then the developer could implement the `Ordered` interface in the customiser they register

graemerocher (approved):

PR #7633, opened 2022-07-03
feat: Decouple Micronaut banner from version banner

engarik:
Decouple Micronaut default banner and line that shows version of framework.

Usecase: 
I want to have my custom banner AND I still want to have line that shows the version of framework (usefull for searching when k8s pods restart in logs).

Adds more flexibility to customization of banner.

I'm not sure, whether this functionality should be covered with tests, because I didn't find any tests for MicronautBanner class.
Also, maybe it's not the best decision to implement Banner interface.
Maybe naming is not the best, I'm not sure either)

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/Micronaut.java
+++ b/context/src/main/java/io/micronaut/runtime/Micronaut.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.runtime;
 
+import io.micronaut.context.banner.MicronautVersionBanner;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.context.ApplicationContext;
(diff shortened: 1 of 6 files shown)


sdelamo (commented):
We already supports custom banners.  This PR feels too complicated just to add the ability to print a custom banner and a line with the version number. 

I would prefer to move out of `inject/src/main/java/io/micronaut/context/banner/MicronautBanner.java` the logic :

```java
String version = VersionUtils.getMicronautVersion();
         version = (version != null) ? " (v" + version + ")" : "";
         out.println(MICRONAUT + version + "\n");```
```

sdelamo:
@engarik I have created a smaller PR which I fulfils your request 

> Decouple Micronaut default banner and line that shows version of framework.

: https://github.com/micronaut-projects/micronaut-core/pull/7745

engarik:
Yes, I think it's much better, thank you!

engarik:
I will close this PR, because similar PR was already merged. Thanks a lot!

PR #7640, opened 2022-07-05
Fix Test Resources property in the BOM

alvarosanchez:
Currently is being generated as `<micronaut.test.resources.version>`.

This changes it to be `<micronaut.test-resources.version>`

### DIFF
--- a/bom/build.gradle
+++ b/bom/build.gradle
@@ -18,6 +18,7 @@ micronautBom {
             'paho.v5': 'pahov5',
             'graal.sdk': 'graalSdk',
             'neo4j.java.driver': 'neo4j.bolt',
+            'micronaut.test.resources': 'micronaut.test-resources'
     ]
     propertyName = 'core'
     suppressions {


melix:
I am personally :-1: on adding more exceptions and prefer relying on our conventions. I would actually like to _get rid_ of the exceptions for Micronaut 4, so that everything is conventional. The property name is derived from the project name. The name is `micronaut-test-resources`, so we have `micronaut.test.resources.version`.

Similarly, we have a project named `micronaut-hibernate-validator` so we have a `micronaut.hibernate.validator.version` property in the BOM already. I find this easier to reason about.

Now if you prefer to rename the property it means we have to live with exceptions forever.

Adding to the discussion, we _could_ decide to rename `micronaut-test-resources` to `micronaut-testresources`, it's a bit involved and would add a bit of noise to users if they browse the repository, since we have released quite a few milestones already, but we could do this too.

In any case, don't consider my :-1: as a blocker if you feel strongly about this.

alvarosanchez:
My arguments for this change:

1. It's `test-resources` (not `test.resources`) everywhere:
	* https://github.com/micronaut-projects/micronaut-gradle-plugin/search?q=%22test-resources%22
	* https://github.com/micronaut-projects/micronaut-test-resources/
2. Out of the 57 `micronaut.*` properties in the BOM, 55 follow the pattern `micronaut.SOMETHING.version`. `micronaut.hibernate.validator.version` and `micronaut.tracing.legacy.version` are the exceptions, not the convention.
3. There is already a `micronaut.test.version` property. Nesting `test.resources` may led someone to think it's a property related to `micronaut-test`, whilst it really belong to `micronaut-test-resources`.

I also agree that naming the project `micronaut-testresources` would have removed this issue, but it's already late for that and worthless.

Having said this, I don't have a strong opinion either, so will leave this here for others to provide feedback.

melix:
> My arguments for this change:
> 
>     1. It's `test-resources` (not `test.resources`) everywhere:
>        

But it's completely orthogonal. The convention is about the mapping between project names and property names. It is very similar to what is done in Gradle version catalogs where aliases may contain `-`, but generated accessors use `.` (because obviously, `-` would be the "minus" sign in Java/Kotlin/Groovy so it wouldn't be legal to use that sign).

When I'm talking about the convention, I'm talking about what we can automatically generate using `micronaut-build`. The previous implementation was giving a mix of this and that, which wasn't great and led to having to declare exceptions.

>     2. Out of the 57 `micronaut.*` properties in the BOM, 55 follow the pattern `micronaut.SOMETHING.version`. `micronaut.hibernate.validator.version` and `micronaut.tracing.legacy.version` are the exceptions, not the convention.
> 

Well we probably disagree on what `SOMETHING` is. I would argue that `test.resources` or `hibernate.validator` _is a_ something :) More seriously, what I don't like about `micronaut.test-resources.version` is that you have, as a user, to think about whether to use `-` or `.` depending on where you are in the property name.

>     3. There is already a `micronaut.test.version` property. Nesting `test.resources` may led someone to think it's a property related to `micronaut-test`, whilst it really belong to `micronaut-test-resources`.

Agreed but there's no much harm. It's actually about testing (fun fact: initially `test-resources` lived under the `io.micronaut.test` group).

> 
> 
> I also agree that naming the project `micronaut-testresources` would have removed this issue, but it's already late for that and worthless.

Yeah it's a pain to change now. That said I'm not convinced that one is preferable to the other.

> 
> Having said this, I don't have a strong opinion either, so will leave this here for others to provide feedback.

alvarosanchez:
Ok, I tried but...

image

🤣

melix:
I'm sorry if it felt like I want to win the argument, this isn't the case: I'm explaining my rationale and why we came to this "exclusion" list, which was initially designed for backwards compatibility. I am absolutely fine if you prefer merging this: I would disagree, but it won't change my life :D

PR #7647, opened 2022-07-07
Use the version of projectreactor from micronaut-reactor

### LINKED ISSUE #7638: Remove manage-reactor as it is in the BOM
### Issue description

Once https://github.com/micronaut-projects/micronaut-reactor/pull/201 is merged and 2.3.0 is out, we should add the BOM to the dependency catalog, and remove `managed-reactor`

Check the TOML keeps the same names with `./gradlew pAPTBR`

timyates:
Instead of having to keep it aligned.

@melix is this right?  Project reactor no longer appears in the bom-pom

### micronaut-bom Pom changes

```diff
82d81
<     <reactor.version>3.4.18</reactor.version>
119c118
<     <micronaut.reactor.version>2.2.2</micronaut.reactor.version>
---
>     <micronaut.reactor.version>2.3.0</micronaut.reactor.version>
637,641d635
<         <groupId>io.projectreactor</groupId>
<         <artifactId>reactor-core</artifactId>
<         <version>${reactor.version}</version>
<       </dependency>
<       <dependency>
```

### micronaut-bom toml changes

```diff
78c78
< micronaut-reactor = "2.2.2"
---
> micronaut-reactor = "2.3.0"
101d100
< reactor = "3.4.18"
553c552,553
< reactor = {group = "io.projectreactor", name = "reactor-core", version.ref = "reactor" }
---
> reactor = {group = "io.projectreactor", name = "reactor-core", version = "" }
> reactor-core = {group = "io.projectreactor", name = "reactor-core", version = "" }
```

I suspect not, so creating this as a draft

Closes #7638

### DIFF
--- a/function-client/build.gradle
+++ b/function-client/build.gradle
@@ -7,5 +7,6 @@ dependencies {
     api project(":function")
     api project(":http-client")
 
-    implementation libs.managed.reactor
+    implementation(platform(libs.boms.micronaut.reactor))
+    implementation libs.project.reactor
 }
--- a/function-web/build.gradle
+++ b/function-web/build.gradle
@@ -8,7 +8,8 @@ dependencies {
     api project(":router")
     api project(":http-server")
 
-    testImplementation libs.managed.reactor
+    testImplementation(platform(libs.boms.micronaut.reactor))
+    testImplementation libs.project.reactor
 
     testImplementation project(":http-client")
     testImplementation project(":inject")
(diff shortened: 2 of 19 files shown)


melix:
Well both the BOM and catalog have breaking changes, so I guess not. This could be done for 4.0.

timyates:
Moved it to 4.0.0 👍

sdelamo (approved):

dstepanov:
IMHO reactor is so essential to the core, I wouldn't move the version away. We would need to bump Micronaut-Reactor all the time which I don't think is needed.

sdelamo:
I agree with @dstepanov I think it is better to keep core decoupled from any module.

PR #7663, opened 2022-07-11
Downgrade GraalVM SDK to 22.0.0.2

sdelamo:
To avoid having to do in modules : 

https://github.com/micronaut-projects/micronaut-sql/blob/master/buildSrc/src/main/groovy/io.micronaut.build.internal.sql-module.gradle#L5-L9

and: https://github.com/micronaut-projects/micronaut-kafka/pull/564/files

and: https://github.com/micronaut-projects/micronaut-core/pull/7663


dstepanov (approved):

dstepanov:
I would add a comment so we remember not to bump it

graemerocher (approved):

sdelamo:
> I would add a comment so we remember not to bump it

Added a comment.

PR #7673, opened 2022-07-12
Convert `TimeConverterRegistrar` to be loaded using the service loader

dstepanov:
I see errors in logs:
`Invalid @Scheduled definition for method: void everyFiveMinutes$test_suite_kotlin() - Reason: Invalid fixed rate definition: 5m`
 that might indicate that scheduled logic was triggered before the time register. Anyway, I think it's better to register it with the newly added way of the service loader.

### DIFF
--- a/config/accepted-api-changes.json
+++ b/config/accepted-api-changes.json
@@ -189,7 +189,7 @@
     "member": "Method io.micronaut.core.type.TypeInformation.asType()",
     "reason": "Added a default method"
   },
-  {  
+  {
     "type": "io.micronaut.core.annotation.AnnotationMetadata",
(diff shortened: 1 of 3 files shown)


graemerocher:
This removes the requires which doesn't load them on android, don't know if that is a problem anymore

dstepanov:
I checked, and it looks like Android already supports them https://developer.android.com/reference/java/time/LocalDate

graemerocher (approved):

sdelamo:
> I checked, and it looks like Android already supports them https://developer.android.com/reference/java/time/LocalDate

It was added only in API level 26.

PR #7683, opened 2022-07-13
Duplicate Injectable Container Type definition

mtalaga:
Removed duplicate injectable container type definition from the table.


graemerocher:
it isn't a duplicate, they are in different packages?

mtalaga:
Oh, you're right. I must have missed that 🤦‍♂️

PR #7689, opened 2022-07-14
build: update to Micronaut Build 5.3.11

sdelamo:

PR #7690, opened 2022-07-14
Temporarially stop running japiCmp

timyates:
Finding the baseline version is causing issues as we believe we are getting rate limited

Turn it off for now as we need to get a snapshot released because the lack of it is causing
errors downstream.

Related:

https://github.com/micronaut-projects/micronaut-build/pull/366
https://github.com/micronaut-projects/micronaut-build/pull/367


dstepanov:
Maybe we can disable it just for Java 8

timyates:
@dstepanov It's failing across the board... See here for 17 https://github.com/micronaut-projects/micronaut-core/runs/7338478999?check_suite_focus=true

sdelamo:
I have reverted already in this PR so that we don't forget to add it back https://github.com/micronaut-projects/micronaut-core/pull/7689

PR #7728, opened 2022-07-22
Clean up request even when using a customizable invoker

### LINKED ISSUE #7704: HttpRequestTerminated. Event is not fired when Controller is sending Stream
### Expected Behavior

https://docs.micronaut.io/3.5.1/api/io/micronaut/http/context/event/HttpRequestTerminatedEvent.html

Event to be fired once request os finalized

### Actual Behaviour

For structured content like json or when. binary content is loaded in memory, event is fired.

But if we use input stream, event is not getting fired.

### Steps To Reproduce

See reproducer.

### Environment Information

JDK 11

### Example Application

https://github.com/sumitaneja/MicronautIssues/tree/main/Issue7704

### Version

3.4.1

yawkat:
When the response body is a `NettyCustomizableResponseTypeHandlerInvoker` (e.g. from an InputStream), `RoutingInBoundHandler` would never clean up the corresponding request.

This patch makes the `NettyCustomizableResponseType.write` and related classes return the `ChannelFuture` for the last response write operation they trigger. `RoutingInBoundHandler` then listens to that future and cleans up the associated request when the write completes.

Fixes #7704

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
@@ -76,6 +76,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufHolder;
 import io.netty.buffer.ByteBufOutputStream;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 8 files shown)


timyates (approved):
I assume we're allowed to change the public API in a minor release as it's marked `@Internal`?

PR #7751, opened 2022-07-27
Fix request scope leak

### LINKED ISSUE #7748: Request Scope Object Memory Leak starting Micronaut 3.5.0
### Expected Behavior

Objects annotated with `@RequestScope` should not left in heap when the request is completed. `

### Actual Behaviour

`RequestScope` Objects can not be collected from heap when all the traffic is handled by the server (even after full gc)
Screen Shot 2022-07-26 at 18 39 09
 
Screen Shot 2022-07-26 at 18 39 16

It looks like starting Micronaut `3.5` has some changes in creating request scoped objects. For request scoped objects `DefaultBeanContext.createRegistration` is called with `dependent:true` in https://github.com/micronaut-projects/micronaut-core/blob/20e29a27638e2c67e5a070b9ebd618c94fe5ed5b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java#L2973
(changed: https://github.com/micronaut-projects/micronaut-core/commit/b29b1d492e6a878b11f0b9ac91d764cc6d169305#diff-ce7d6e5b454c8f48a3806d448c8943d1ffb409efbfaa7ddc20da97cd77752959R2867) This is causing all the request scoped objects of the same type refers to each other in their `beanResolutionContext.dependentBeans` list, while on Micronaut `3.4.4` it was an empty list.

### Steps To Reproduce

Checkout linked github repo. Just send some traffic to `GET http://localhost:8080/`
Get a heap dump, and search for classes `PageContext` or `io.micronaut.http.server.util.locale.HttpLocalizedMessageSource`
Depending on the number of requests you produce the instance count differs.

### Environment Information

Max OS 12.0.1

### Example Application

https://github.com/orkunozbek-gopuff/request-scope-leak

### Version

3.5.0

dstepanov:
Fixes #7748

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/RequestCertificateSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/RequestCertificateSpec.groovy
@@ -4,7 +4,6 @@ import io.micronaut.http.HttpRequest
 import io.micronaut.http.HttpStatus
 import io.micronaut.http.annotation.Controller
 import io.micronaut.http.annotation.Get
-import io.micronaut.http.server.netty.AbstractMicronautSpec
 import reactor.core.publisher.Flux
 
 import java.security.cert.X509Certificate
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

sdelamo (approved):

raksikkim-gopuff:
@sdelamo , Would you please sharer when this fix can be released? Also, is it doable to include the fix for a lower version of micronaut like 3.5.x too?

sdelamo:
@raksikkim-gopuff 

> @sdelamo , Would you please sharer when this fix can be released? Also, is it doable to include the fix for a lower version of micronaut like 3.5.x too?

3.6.0 was released August 4th.

PR #7773, opened 2022-07-29
 New client channel customization API

yawkat:
Analogous to #7625, this patch introduces a *client* channel customization API. This API has the same advantages as the server equivalent, and is also required if we ever want better HTTP2 connection support for the client (though this is not on the roadmap yet).

 ChannelPipelineListener is also deprecated now that a replacement is available for client and server.

 This patch can go into 3.6.0 or 3.7.0, it's not really urgent. However it resolves the "half-deprecation" of ChannelPipelineListener and makes the docs clearer. Feel free to retarget once a 3.7 branch is available.


timyates (approved):
Do we have any tests for the old deprecated method?

It's deprecated, but it should still be tested so it works?

I couldn't find any usage of `ChannelPipelineListener` 🤔

graemerocher (approved):

yawkat:
@timyates it used to be tested in the test-suite. i deleted those tests, maybe i shouldnt have, but it was fairly broken anyway

PR #7775, opened 2022-07-29
Fix ClientFilterSpec

yawkat:
It doesn't like the shared context for some reason.


melix (approved):

yawkat:
@sdelamo can we still merge this? it's really minor.

sdelamo:
Does it fails in 3.5.x? If yes, we should target 3.5.x as base.

yawkat:
i dont know, it hasnt always failed

PR #7797, opened 2022-07-31
Update contributor development setup instructions

jeremyg484:
There were several outdated and redundant instructions in the README and CONTRIBUTING files, including references to the archived Micronaut Profiles repository and incorrect procedures for building and locally installing the CLI. Also, the image link in the Checkstyle section was broken, and there were some minor grammatical issues.

A pending concern I have is that the updated instructions do not address the caveats of the CLI being in a separate project from micronaut-core. In particular, following the instructions for building and installing the CLI will not pull in a locally built version of micronaut-core. An elegant and "idiomatic Gradle" way to incorporate such local builds would be a good addition to these instructions, but I've yet to find a nice way to do it that isn't a bit of a hack.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,8 @@
 
 ## Finding Issues to Work on
 
-If you are interested in contributing to Micronaut and are looking for issues to work on, take a look at the issues tagged with [help wanted](https://github.com/micronaut-projects/micronaut-core/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).
+If you are interested in contributing to Micronaut and are looking for issues to work on, take a look at the issues
+tagged with [help wanted](https://github.com/micronaut-projects/micronaut-core/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).
 
 ## JDK Setup
 
@@ -15,12 +16,12 @@ Micronaut can be imported into IntelliJ IDEA by opening the `build.gradle` file.
 ## Docker Setup
 
 Micronaut tests currently require docker to be installed.
- 
+
 ## Running Tests
 
-To run the tests use `./gradlew check`. 
+To run the tests use `./gradlew check`.
 
-[Geb](http://gebish.org) functional tests are ignored unless you specify the geb environment via system property. 
+[Geb](http://gebish.org) functional tests are ignored unless you specify the geb environment via system property.
 
 To run with Chrome `./gradlew -Dgeb.env=chrome check`.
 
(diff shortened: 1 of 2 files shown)


sdelamo:
> There were several outdated and redundant instructions in the README and CONTRIBUTING files, including references to the archived Micronaut Profiles repository and incorrect procedures for building and locally installing the CLI. Also, the image link in the Checkstyle section was broken, and there were some minor grammatical issues.

> A pending concern I have is that the updated instructions do not address the caveats of the CLI being in a separate project from micronaut-core. In particular, following the instructions for building and installing the CLI will not pull in a locally built version of micronaut-core. An elegant and "idiomatic Gradle" way to incorporate such local builds would be a good addition to these instructions, but I've yet to find a nice way to do it that isn't a bit of a hack.

sdelamo:
Please, note that we have done changes to `CONTRIBUTING.md` already in 3.6.x https://github.com/micronaut-projects/micronaut-core/commit/b3eb359bfde92cc1ff9754f03d2c87606479faf1

sdelamo:
I have incorporated you changes to README.md here: https://github.com/micronaut-projects/micronaut-core/pull/7800

sdelamo:
> A pending concern I have is that the updated instructions do not address the caveats of the CLI being in a separate project from micronaut-core. In particular, following the instructions for building and installing the CLI will not pull in a locally built version of micronaut-core. An elegant and "idiomatic Gradle" way to incorporate such local builds would be a good addition to these instructions, but I've yet to find a nice way to do it that isn't a bit of a hack.

Usually, you don't need to build Micronaut CLI  to work with Micronaut Core. `micronaut-starter` `CONTRIBUTING.md` contains more up-to-date instructions. 

https://github.com/micronaut-projects/micronaut-starter/blob/3.5.x/CONTRIBUTING.md

jeremyg484:
> Please, note that we have done changes to `CONTRIBUTING.md` already in 3.6.x [b3eb359](https://github.com/micronaut-projects/micronaut-core/commit/b3eb359bfde92cc1ff9754f03d2c87606479faf1)

Great! Those updates are definitely less confusing, and I see they make more explicit the decoupling between micronaut-core and the CLI.
 
> Usually, you don't need to build Micronaut CLI to work with Micronaut Core. `micronaut-starter` `CONTRIBUTING.md` contains more up-to-date instructions.
> 
> https://github.com/micronaut-projects/micronaut-starter/blob/3.5.x/CONTRIBUTING.md

Yes, I was able to see that fairly quickly once I had gotten both projects cloned and set up, but the prior instructions made it seem otherwise.

The image link in the Checkstyle section is still broken in the 3.6.x branch. It would be good to fix that and then to copy all of the updates in that section to the other modules since they seem to all have inherited the same text. Since that section is likely to continue to be the same in all of the projects, perhaps it would be better to break it out into its own single document to which all of the projects could link?

PR #7809, opened 2022-08-02
Allow client config of WebSocket per message deflate

### LINKED ISSUE #7711: DefaultHttpClient no way to turn off compression
### Issue description

Hi,

I'm trying to handle 100s of simultaneously websocket connections sending binary content.
During CPU profiling I'm observing a lot of  CPU usage on `deflate` calls that harm performance of my service in dockerized environment.
Since data are being sent on the same Kubernetes cluster, network bandwidth is bigger than available CPUs and in such case burning CPU cycles on compression and decompression is a huge bottleneck.

What I've found during debugging is that the following line sets the compression level with a default value `6`: https://github.com/micronaut-projects/micronaut-core/blob/3.5.x/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java#L952 - it uses a default instance of `WebSocketClientCompressionHandler` which in turn sets default value for compression (and for other parameters, too).

This has been incorporated in this PR: #6487

I don't see any way to turn off this behaviour or set compression level to 0 (no compression) for http client. Is there any way to enforce no compression? or whether an enhancement request required to let it be configurable (in a similar way as it is for server: `micronaut.server.netty.compression-level`)?

jeremyg484:
Add a configuration property to HttpClientConfiguration that allows the enabling or disabling of per message deflate for WebSocket clients. Defaults to true in order to preserve the existing behavior.

This potentially resolves #7711

Other related WebSocket configuration bits potentially worth supporting are captured in #9162


rjaros87:
Any updates?

sdelamo:
@jeremyg484 can you merge 3.9.x into this PR and change the base to 3.9.x

jeremyg484:
This is actually not quite ready for review yet, I've realized.

I'm going to make one further enhancement here to split the WebSocket configuration into its own namespace, as I think that will be needed to make things cleaner when adding additional configuration options.

jeremyg484:
@sdelamo This is now ready for review again. I moved the config into its own `.ws.compression` namespace to better accommodate adding further options.

sdelamo:
@jeremyg484 can you cherrypick this change with base 4.3.x?

PR #7815, opened 2022-08-02
Close the NettyCompletedFileUpload release tracker when getInputStream is called

yawkat:
getInputStream delegates control of the release of the buffer to the stream it returns. discard is never called. This patch releases the resource tracker on NettyCompletedFileUpload so that it doesn't erroneously report a leak in that case.

Testing is a bit messy because it's another resource leak, but the test case *should* trigger the leak detection sometimes.


graemerocher (approved):

sdelamo (approved):

sdelamo:
I have added it to the Project. @yawkat please use the projects.

PR #7824, opened 2022-08-03
Bump micronaut-test to 3.5.0

micronaut-build:


sdelamo:
@timyates  any idea why this fails?

timyates:
Ok, this is freaking weird...

It uses Kotlin 1.6.21 for everything but the testRuntimeClasspath, where it switches to 1.7.10

[There's a build scan here](https://ge.micronaut.io/s/rl732d6fektrq/dependencies?configurationFilter=WyJ0ZXN0Q29tcGlsZUNsYXNzcGF0aCIsInRlc3RSdW50aW1lQ2xhc3NwYXRoIl0&dependencies=kotlin-st&expandAll&projectFilter=WyI6dGVzdC1zdWl0ZS1rb3RsaW4iXQ)

I don't think Kotest4 works with Kotlin 1.7.x

@melix Can you see why the 1.6.21 platform constraint seems to work for everything but the runtime classpath?  This feels like a hack

melix:
The platform will _not_ downgrade your dependencies. The tests use Kotlin 1.7: https://ge.micronaut.io/s/rl732d6fektrq/dependencies?configurationFilter=WyJ0ZXN0Q29tcGlsZUNsYXNzcGF0aCIsInRlc3RSdW50aW1lQ2xhc3NwYXRoIl0&dependencies=kotlin-st&expandAll&focusedDependency=WzI2LDEyLDExNTYsWzAsMSxbMTRdXV0&focusedDependencyView=versions&projectFilter=WyI6dGVzdC1zdWl0ZS1rb3RsaW4iXQ

So the dependency is upgraded to 1.7, and which brings the Kotlin 1.7 BOM, which in turn triggers all Kotlin dependencies to be bumped to 1.7.

timyates:
So is the solution to downgrade the kotlin version for kotest4 in micronaut-test?

melix:
The responsible library seems to be `io.micronaut.test:micronaut-test-kotest:3.5.0` which is wants `org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.10`. See https://github.com/micronaut-projects/micronaut-test/blob/master/gradle/libs.versions.toml#L13

I'm not sure which one is correct, but having core which wants 1.6 and libraries 1.7 seems weird. So for tests, you can fix it using the rule you wrote, that's fine.

timyates:
> The responsible library seems to be io.micronaut.test:micronaut-test-kotest:3.5.0 which is wants org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.10

Yeah, I think this is wrong, as I don't think Kotest4 works with Kotlin 1.7 🤔

timyates:
Wish I knew kotest 😆

timyates:
Also, in my tests, I can't get kotest to use a different kotlin to kotest5, so I'll leave it as is

PR #7829, opened 2022-08-03
bug: Should continue when predestroy fails with an execption

timyates:
[skip ci]

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -1142,7 +1142,11 @@ public class DefaultBeanContext implements InitializableBeanContext {
         beanToDestroy = triggerPreDestroyListeners(definition, beanToDestroy);
 
         if (definition instanceof DisposableBeanDefinition) {
-            ((DisposableBeanDefinition<T>) definition).dispose(this, beanToDestroy);
+            try {
+                ((DisposableBeanDefinition<T>) definition).dispose(this, beanToDestroy);
+            } catch (Exception e) {
+                LOG.info("Error disposing bean [{}]... Continuing...", beanToDestroy, e);
+            }
         }
         if (beanToDestroy instanceof LifeCycle) {
             try {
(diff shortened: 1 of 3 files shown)


dstepanov (approved):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1142,7 +1142,11 @@ private <T> void destroyBean(@NonNull BeanRegistration<T> registration, boolean
         beanToDestroy = triggerPreDestroyListeners(definition, beanToDestroy);
 
         if (definition instanceof DisposableBeanDefinition) {
-            ((DisposableBeanDefinition<T>) definition).dispose(this, beanToDestroy);
+            try {
+                ((DisposableBeanDefinition<T>) definition).dispose(this, beanToDestroy);
+            } catch (Exception e) {
+                LOG.info("Error disposing bean [{}]... Continuing...", beanToDestroy, e);
```
graemerocher: Should be `warn` and I don't think you can use parameterization like this to include the stack trace. Has to be normal string concatenation. Also include `e.getMessage()` in the error

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
dstepanov: I think it will work, the last parameter can be the exception cause

sdelamo (approved):
@timyates please address @graemerocher request

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
timyates: @dstepanov is right 😄 

Changed it to:

```
                if (LOG.isWarnEnabled()) {
                    LOG.warn("Error disposing bean [" + beanToDestroy.toString() + "]... Continuing...", e);
                }
```

timyates:
@sdelamo done (hopefully)

Issue #7861, opened 2022-08-08
Expired Connect TTL should prevent connections from being used


dhofftgt:
### Expected Behavior

When micronaut.http.client.connect-ttl is set for a http client and a connection expires, it should prevent expired connection from being used. 

### Actual Behaviour

The expired connection is allowed to be checked out of the pool, and after using it, only then it is checked to see if it is expired. 

This somewhat defeats the purpose of a connection ttl, as it should be to prevent stale connections from being used in the first place.  

We have some sparsely used backends that have old half open connections lingering in the pool. When these are checked out, they are used and we get connection reset by peer errors. 

### Steps To Reproduce

PR with failing test. https://github.com/micronaut-projects/micronaut-core/pull/7862

### Version

3.6.0

PR #7862, opened 2022-08-08
Issue #7861 - do expiry check on checkout of the connection

dhofftgt:
I created an issue: https://github.com/micronaut-projects/micronaut-core/issues/7861


graemerocher:
Could you sign the CLA? Thanks

http-client/src/main/java/io/micronaut/http/client/netty/ConnectTTLHandler.java
```diff
@@ -67,12 +68,21 @@ public void handlerRemoved(ChannelHandlerContext ctx) {
...
+            ctx.channel().attr(CHANNEL_EXPIRED).set(true);
         }
     }
+
+    /**
+     * Indicates whether the channels connection ttl has expired
+     * @param channel The channel to check
+     * @return true if the channels ttl has expired
+     */
+    public static Boolean isChannelExpired(Channel channel) {
```
yawkat:
```suggestion
    public static boolean isChannelExpired(Channel channel) {
```

yawkat (commented):
is the channel released at all now?

dhofftgt:
> is the channel released at all now?

The next time it is checked out of the pool It will fail the health check and be closed. Prior it would check it out, try to use it before closing it instead of returning it to the pool.

yawkat:
But if it is marked as expired prior to the release, it will not re-enter the pool at all, and then it will not be closed?

dhofftgt:
> But if it is marked as expired prior to the release, it will not re-enter the pool at all, and then it will not be closed?

Yeah. if it's marked expired while it is being used, it will be closed and not returned to the pool. However if it is idle in the pool, it won't be closed until it is checked out, used, and checked for expiration prior to being returned. We see connections get severed in the pool after being idle for a while. It'd be better for this scenario to check if the connection is expired before using it.

dhofftgt:
> Could you sign the CLA? Thanks

I am working on this. I need to have my company sign a Corporate CLA before I can sign the CLA.

yawkat:
> Yeah. if it's marked expired while it is being used, it will be closed and not returned to the pool

Where is it closed? You removed the code that closed it in channelReleased, and I don't see any code in SimpleChannelPool that would close it.

http-client/src/main/java/io/micronaut/http/client/netty/ConnectTTLHandler.java
```diff
@@ -30,7 +31,7 @@
  */
 public class ConnectTTLHandler extends ChannelDuplexHandler {
 
-    public static final AttributeKey<Boolean> RELEASE_CHANNEL = AttributeKey.newInstance("release_channel");
+    public static final AttributeKey<Boolean> CHANNEL_EXPIRED = AttributeKey.newInstance("channel_expired");
```
yawkat: this class is not `@Internal` so we can't replace this attribute, despite this being a more appropriate name. I'm not sure why japicmp didn't trigger though.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -406,6 +406,7 @@ public DefaultHttpClient(@Nullable LoadBalancer loadBalancer,
 
         Optional<Duration> connectTtl = configuration.getConnectTtl();
         this.connectionTimeAliveMillis = connectTtl.map(duration -> !duration.isNegative() ? duration.toMillis() : null).orElse(null);
+        final ChannelHealthChecker channelHealthChecker = channel -> channel.eventLoop().newSucceededFuture(channel.isActive() && !ConnectTTLHandler.isChannelExpired(channel));
```
dhofftgt, line 409: @yawkat Here is where the expiration check is added. Expired channels won't be considered healthy, and then they are closed in SimpleChannelPool 'notifyHealthCheck' as a result of failing the health check.

yawkat: i understand the idea, but `notifyHealthCheck` is only called from `doHealthCheck`, which is only called during acquire. However if the health check fails in `doHealthCheckOnRelease` because the channel has expired, then it won't be returned to the pool at all, and `notifyHealthCheck` will never be called.

yawkat: ideally the channel closure would be asserted by tests, but it might not be feasible to write a test for this atm, maybe once there is EmbeddedChannel support for the client.

dhofftgt: oh. I see. releaseAndOfferIfHealthy does not close the channel if it is unhealthy.

dhofftgt: There are tests in ConnectionTTLSpec. I updated that earlier to cover the scenario where the connection expires while in the pool. I have checked and can get a test to fail for the returning to pool case. I will push some changes shortly.

yawkat: ah awesome, didn't see those tests. I'll look at your updates tomorrow.

http-client/src/main/java/io/micronaut/http/client/netty/ConnectTTLHandler.java
```diff
@@ -67,12 +68,21 @@ public void handlerRemoved(ChannelHandlerContext ctx) {
     }
 
     /**
-     * Will set RELEASE_CHANNEL as true for the channel attribute when connect-ttl is reached.
+     * Will set CHANNEL_EXPIRED as true for the channel attribute when connect-ttl is reached.
```
yawkat:
```suggestion
     * Will set RELEASE_CHANNEL as true for the channel attribute when connect-ttl is reached.
```

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -2641,12 +2643,9 @@ public void channelReleased(Channel ch) {
                     }
                 }
 
-                if (connectionTimeAliveMillis != null) {
-                    boolean shouldCloseOnRelease = Boolean.TRUE.equals(ch.attr(ConnectTTLHandler.RELEASE_CHANNEL).get());
```
yawkat:
```suggestion
```

yawkat (approved):

timyates (approved):
Agree with @yawkat's suggestions too 👍

dhofftgt:
@yawkat @graemerocher @timyates Thanks for the reviews and feedback on this. I have the CLA signed and just squashed and rebased the commit. Is there anything else you'd need from me to get this merged?

yawkat:
No this is enough, thanks for the contribution! The change will be in 3.6.1. Two things for next time:

- please add a short description of the change as the PR description. I've added that for this merge in the commit.
- You do not need to squash/rebase, we can do that in the github UI.

PR #7874, opened 2022-08-13
Correcting KeyStore spelling in the documenation

anuragdeshpande:
Noticed that the KeyStore declaration in SSL config for application server, needs to be `key-store` to be picked up by the application server properly. 

Updating documentation to avoid confusion


sdelamo:
@anuragdeshpande thanks

sdelamo (approved):

PR #7881, opened 2022-08-15
feat: Add CracContextProvider

sdelamo:
This allows users to provide a Context without using `Core::getGlobalContext`

### DIFF
--- a/crac/src/main/java/io/micronaut/crac/support/GlobalCracContextProvider.java
+++ b/crac/src/main/java/io/micronaut/crac/support/GlobalCracContextProvider.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


crac/src/main/java/io/micronaut/crac/support/GlobalCracContextProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import org.crac.Context;
+import org.crac.Core;
+import org.crac.Resource;
+
+/**
+ * Gets a Global Context through {@link Core#getGlobalContext()}.
+ * @author Sergio del Amo
+ * @since 3.7.0
+ */
+@Requires(classes = {Resource.class, Context.class})
```
timyates: Do we need this?  There is a requires on the package-info.java

timyates (approved):
One question, but that's probably me missing something

crac/src/main/java/io/micronaut/crac/support/GlobalCracContextProvider.java
sdelamo: we don't. if it is present in the `package-info.java` that is enough.

sdelamo: removed here https://github.com/micronaut-projects/micronaut-core/pull/7881/commits/899002da918a6c2454bc0210ddf1a080358b2a3d

PR #7884, opened 2022-08-15
refactor: extract Facade to its own class

sdelamo:


crac/src/main/java/io/micronaut/crac/support/CracFacadeResource.java
```diff
@@ -0,0 +1,49 @@
...
+        orderedCracResource.beforeCheckpoint(createCracContext(context));
+    }
+
+    @Override
+    public void afterRestore(Context<? extends Resource> context) throws Exception {
+        orderedCracResource.afterRestore(createCracContext(context));
+    }
+
+    private CracContext createCracContext(Context<? extends Resource> context) {
+        return new DefaultCracContext(() -> (Context<Resource>) context);
```
sdelamo, line 47: It passes the `context` passed  to original methods: `org.crac.Resource::beforeCheckpoint` and `org.crac.Resource::afterRestore`. 

I think the cast to `Context<Resource>` is correct. Not 100% sure.

crac/src/main/java/io/micronaut/crac/support/CracFacadeResource.java
```diff
@@ -0,0 +1,49 @@
...
+import io.micronaut.core.annotation.Internal;
+import org.crac.Context;
+import org.crac.Resource;
+
+/**
+ * The bridge between a Micronaut CRaC Resource and the CRaC api.
+ * @author Tim Yates
+ * @since 3.7.0
+ */
+@Internal
```
timyates, line 27: Does this also need `@Experimental`?

crac/src/main/java/io/micronaut/crac/support/CracFacadeResource.java
timyates: I spent a long time in this kind of generic hell on Thurs/Friday...  I think this is ok though 👍

timyates (approved):

crac/src/main/java/io/micronaut/crac/support/CracFacadeResource.java
sdelamo: As it is `@Internal`, I think it is ok. But will annotate everything `@Experimental` nonetheless.

Issue #7786, opened 2022-07-30
thrown io.micronaut.http.codec.CodecException does not invoke exception mappers


mrwilby:
### Expected Behavior

In 1.x, when a CodecException was thrown, exception mappers were invoked. This is the expected behavior.

### Actual Behaviour

In 3.5.3, when CodecException is thrown, the exception mappers are never invoked, and instead io.micronaut.http.server.RouteExecutor#createDefaultErrorResponse is invoked.

### Steps To Reproduce

```
@Controller
public class ExceptionMapperAppController {
    @Post(uri = "/exceptions/codec", produces = MediaType.APPLICATION_JSON)
    public ExceptionMapperAppResponse throwCodecException(@Body ExceptionMapperAppRequest request) {

        ExceptionMapperAppResponse r = new ExceptionMapperAppResponse(request.getValue());
        return r;
    }
}

@Valid
@Validated
@Introspected
public class ExceptionMapperAppResponse {

    private final String value;

    @JsonCreator
    public ExceptionMapperAppResponse(@JsonProperty("value") String value) {
        this.value = value;
    }

    @JsonProperty("value")
    public String getValue() {

        if (this.value.equalsIgnoreCase("error")) {
            throw new RuntimeException("Triggering JSON exception");
        }

        return this.value;
    }
}

@Client(value = "/", errorType = PlatformErrorMessage.class)
@Requires(env = "exception-mapper-test")
public interface ExceptionMapperAppClient {
    @Post("/exceptions/codec")
    @Consumes(MediaType.APPLICATION_JSON)
    ExceptionMapperAppResponse throwCodecException(@Body ExceptionMapperAppRequest request);
}

@MicronautTest(environments = "exception-mapper-test")
public class ExceptionMapperAppTest {

    @Test
    public void testCodecException() {

        try {
            ExceptionMapperAppRequest request = new ExceptionMapperAppRequest("error");
            this.client.throwCodecException(request);
            Assertions.fail();
        }
        catch (HttpClientResponseException e) {
            Assertions.assertEquals(HttpStatus.BAD_REQUEST, e.getStatus());
            checkForPlatformErrorMessage(InternalHttpResponseStatus.BAD_REQUEST, e, "Invalid request parameters");
        }
        catch (Throwable t) {
            Assertions.fail();
        }
    }
```

### Version

3.5.3

mrwilby:
Call stack when CodecException is thrown:

```
[...]

PR #7936, opened 2022-09-02
Fix buffer leak on response encode error

yawkat:
Also includes a `@PendingFeature` test case for #7786

### DIFF
--- a/json-core/src/main/java/io/micronaut/json/codec/MapperMediaTypeCodec.java
+++ b/json-core/src/main/java/io/micronaut/json/codec/MapperMediaTypeCodec.java
@@ -19,6 +19,7 @@ import io.micronaut.context.BeanProvider;
 import io.micronaut.core.annotation.Experimental;
 import io.micronaut.core.io.buffer.ByteBuffer;
 import io.micronaut.core.io.buffer.ByteBufferFactory;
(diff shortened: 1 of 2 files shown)


timyates (approved):

json-core/src/main/java/io/micronaut/json/codec/MapperMediaTypeCodec.java
```diff
@@ -264,8 +265,15 @@ public <T, B> ByteBuffer<B> encode(Argument<T> type, T object, ByteBufferFactory
...
         }
         ByteBuffer<B> buffer = allocator.buffer();
-        OutputStream outputStream = buffer.toOutputStream();
-        encode(type, object, outputStream);
-        return buffer;
+        try {
+            OutputStream outputStream = buffer.toOutputStream();
+            encode(type, object, outputStream);
+            return buffer;
+        } catch (Throwable t) {
```
sdelamo, line 272: Sonar recommends catching `Exception` instead of `Throwable`. Should we do that?

yawkat: no, this is supposed to work kind of like a finally block, it should always release

Issue #7948, opened 2022-09-05
Regression in ConversionService for concurrent application contexts


yawkat:
### Issue description

#7635 changed the `ConversionService.DEFAULT` to be reset when an application context is closed. This breaks non-standard conversions (e.g. `NettyConverters`) when there are concurrent application contexts:

```groovy
    def "test convert bytebuf after context reset"() {
        given:
        ApplicationContext ctx1 = ApplicationContext.run()
        ApplicationContext ctx2 = ApplicationContext.run()
        ByteBuf buf = Unpooled.wrappedBuffer("foo".bytes)

        expect:
        // works as expected
        ctx1.getBean(ConversionService).convert(buf, String).get() == 'foo'
        ctx2.getBean(ConversionService).convert(buf, String).get() == 'foo'

        when:
        ctx2.stop()

        then:
        // this fails, the converter has been removed by ctx2.stop()
        ctx1.getBean(ConversionService).convert(buf, String).get() == 'foo'

        cleanup:
        buf.release()
        ctx1.close()
    }
```

@dstepanov mentioned that we should probably fix this in 4.0, but we may need to do this earlier. In particular, this issue lead to a test failure in the aws module: https://github.com/micronaut-projects/micronaut-aws/pull/1452 – the application context for one test was closed late, leading to the conversion service for another test being improperly reset.

I'm not sure of the path forward here. We could revert #7635, but I'm not sure if this regression is worse than the bug that was fixed in that PR. We could also move forward with @dstepanov suggestion of a "wrapping" conversion service for each context, which is the proper way to do this, but I'm not sure if uncoupling the context conversion service would lead to other issues that need a major release.

dstepanov:
Which particular converter from `NettyConverters` is used? 
We might want to implement that solution I mentioned in some backward compatible way, with the idea that `ConversionService.DEFAULT` is not being used there.

yawkat:
this test case uses compositeByteBufCharSequenceTypeConverter

dstepanov:
Those don't rely on the bean context and can be extracted to an additional `TypeConverterRegistrar` registered using the service loader. I would expect those to survive the purge.

yawkat:
that seems like a good workaround for 3.6, I will make a PR. thanks!

musketyr:
What is the actual workaround here? I've tried to inject `Environment` instead of `ConversionService` but I still got the following error:

```
[...]

PR #7952, opened 2022-09-06
Move some netty converters to SPI

yawkat:
Move some converters from NettyConverters into a new converter registrar that does not require further dependencies from the application context. This allows the registrar to be loaded into the default conversion service with no app context present.

This is a workaround for #7948, and fixes https://github.com/micronaut-projects/micronaut-aws/pull/1452 .


dstepanov:
Some of the converters that are using `conversionService` are using convertors from the same class. Maybe we can make them static fields and use them directly or use `ConversionServiceDEFAULT`? The first option removes an option to override them, but that might be fine.

yawkat:
sounds sensible, ive moved the http data converters to the spi class.

dstepanov (approved):

sdelamo (approved):

PR #7963, opened 2022-09-08
Allow static executable methods, support executing private fields + methods

dstepanov:


graemerocher (approved):
Looks good, if you can write some tests for the error conditions (like when executable is declared without reflective access) that would be good

dstepanov:
The tests are already there, that’s the old behavior. The new one allows them to be invokes with the reflection.

PR #7994, opened 2022-09-13
HTTP client refactor

yawkat:
Note: making the PR against 3.7.x, but this can go into 3.8.x once there is a branch for that, if the review isn't done in time for 3.7.0.

This is a first PR for my HTTP client refactor. The goal of these changes is to move the connection / channel management code out of DefaultHttpClient, with minimal functional changes. Request/response mapping (basically anything that happens downstream of HTTP parsing) remains in DefaultHttpClient.

This encapsulation will allow me to make functional changes to pooling (pooling for streaming requests, http2 single-connection pooling) with lower risk of breaking changes. I am making this PR so that my changes don't diverge too much.

To make review doable, I've split up these changes into many individual commits. Each commit contains changes that work standalone, and most commits are mechanical (e.g. moving a method from DefaultHttpClient to ConnectionManager) so should be easy to review. There are a few more interesting changes mixed in, such as removal of closeChannelAsync which turned out to do nothing.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
@@ -9,6 +9,7 @@ import io.micronaut.http.HttpVersion;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.client.HttpClientConfiguration;
 import io.micronaut.http.client.exceptions.HttpClientException;
+import io.micronaut.http.client.netty.ssl.NettyClientSslBuilder;
 import io.micronaut.http.netty.channel.ChannelPipelineCustomizer;
 import io.micronaut.http.netty.channel.ChannelPipelineListener;
 import io.micronaut.http.netty.channel.NettyThreadFactory;
@@ -139,24 +140,31 @@ final class ConnectionManager {
         HttpVersion httpVersion,
         InvocationInstrumenter instrumenter,
         ChannelFactory<? extends Channel> socketChannelFactory,
-        @Nullable Long readTimeoutMillis,
-        @Nullable Long connectionTimeAliveMillis,
-        SslContext sslContext,
+        NettyClientSslBuilder nettyClientSslBuilder,
         NettyClientCustomizer clientCustomizer,
         Collection<ChannelPipelineListener> pipelineListeners,
         String informationalServiceId) {
 
+        if (httpVersion == null) {
+            httpVersion = configuration.getHttpVersion();
+        }
+
         this.log = log;
         this.httpVersion = httpVersion;
         this.threadFactory = threadFactory;
-        this.sslContext = sslContext;
         this.configuration = configuration;
         this.instrumenter = instrumenter;
-        this.readTimeoutMillis = readTimeoutMillis;
-        this.connectionTimeAliveMillis = connectionTimeAliveMillis;
         this.clientCustomizer = clientCustomizer;
         this.pipelineListeners = pipelineListeners;
         this.informationalServiceId = informationalServiceId;
+
+        this.connectionTimeAliveMillis = configuration.getConnectTtl()
+            .map(duration -> !duration.isNegative() ? duration.toMillis() : null)
+            .orElse(null);
+        this.readTimeoutMillis = configuration.getReadTimeout()
+            .map(duration -> !duration.isNegative() ? duration.toMillis() : null)
+            .orElse(null);
+        this.sslContext = nettyClientSslBuilder.build(configuration.getSslConfiguration(), httpVersion).orElse(null);
 
         if (eventLoopGroup != null) {
             group = eventLoopGroup;
--- a/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
@@ -6,15 +6,19 @@ import io.micronaut.http.HttpVersion;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.client.HttpClientConfiguration;
 import io.micronaut.http.client.exceptions.HttpClientException;
+import io.micronaut.http.netty.channel.ChannelPipelineCustomizer;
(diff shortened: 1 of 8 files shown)


http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -0,0 +1,1263 @@
...
+import java.time.Duration;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+final class ConnectionManager {
```
graemerocher, line 130: add javadoc and annotate with `@Internal`

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -0,0 +1,1263 @@
...
+                                emitter.error(customizeException(new HttpClientException("Connect error:" + cause.getMessage(), cause)));
+                            }
+                        });
+                }
+            } catch (HttpClientException e) {
+                emitter.error(e);
+                return;
+            }
+
+            // todo: on emitter dispose/cancel, close channel
```
timyates, line 551: Is this TODO still to be done?

timyates (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1475,229 +1170,6 @@ protected Object getLoadBalancerDiscriminator() {
...
-     *
-     * @param request         The request
-     * @param uri             The URI to connect to
-     * @param sslCtx          The SslContext instance
-     * @param isStream        Is the connection a stream connection
-     * @param contextConsumer The logic to run once the channel is configured correctly
-     * @return A ChannelFuture
-     * @throws HttpClientException If the URI is invalid
-     */
-    protected ChannelFuture doConnect(
```
graemerocher, line 1496: Whilst this is an internal class and this change is probably fine we will need to review whether downstream subclasses use any of these deleted protected methods

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -0,0 +1,1263 @@
...
+            sslCtx = null;
+        }
+        return sslCtx;
+    }
+
+    private PoolHandle mockPoolHandle(Channel channel) {
+        return new PoolHandle(null, channel);
+    }
+
+    Mono<PoolHandle> connectForExchange(DefaultHttpClient.RequestKey requestKey, boolean multipart, boolean acceptEvents) {
```
graemerocher, line 448: it would be good if you could javadoc each `connect*` method if only so other maintainers can understand the differences in behaviour in each method

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
yawkat: I removed code from there that appeared to close the channel, but actually did nothing. I think the channel should be closed in this case, but I think that change would be too risky atm.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: What downstream users are there of these methods (not just this one, there's lots of removed protected methods)? I think the reactive client stuff used to override some methods, but that's not a thing anymore.

graemerocher: The ribbon module from the Netflix module used to subclass these but that is deprecated and archived so maybe this is not a concern anymore

graemerocher (approved):

timyates:
This is causing the native CLI tool to fail to build in starter

https://github.com/micronaut-projects/micronaut-starter/actions/runs/3066455261/jobs/4951646242

The trace for "PlatformDependent" is

```
io.netty.util.internal.PlatformDependent the class was requested to be initialized at run time (from 'META-INF/native-image/io.netty/common/native-image.properties' in 'file:///Users/tim/Code/GitHub/micronaut-projects/micronaut-starter/starter-cli/build/libs/micronaut-cli-3.7.0-SNAPSHOT-all.jar' with 'io.netty.util.AbstractReferenceCounted' and from 'META-INF/native-image/io.netty/codec-http2/native-image.properties' in 'file:///Users/tim/Code/GitHub/micronaut-projects/micronaut-starter/starter-cli/build/libs/micronaut-cli-3.7.0-SNAPSHOT-all.jar' with 'io.netty'). io.micronaut.http.client.netty.ConnectionManager caused initialization of this class with the following trace:
	at io.netty.util.internal.PlatformDependent.<clinit>(PlatformDependent.java:79)
	at io.netty.util.ConstantPool.<init>(ConstantPool.java:34)
	at io.netty.util.AttributeKey$1.<init>(AttributeKey.java:27)
	at io.netty.util.AttributeKey.<clinit>(AttributeKey.java:27)
	at io.micronaut.http.client.netty.ConnectionManager.<clinit>(ConnectionManager.java:132)
```

So it looks like `AttributeKey` triggers a series of classes which are flagged by I think netty itself as being run-time

Do we need to wrap all the AttributeKeys in here in Suppliers to make them lazily created?

https://github.com/micronaut-projects/micronaut-core/blob/69cd05e7fd05dd5dca39bf94455b352436695390/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java#L131

Or something else?

yawkat:
@timyates ugh. why did it work when those fields were in DefaultHttpClient? is there a special case to make DefaultHttpClient init at runtime?

timyates:
Not sure...  I wonder if ConnectionManager is getting created at build time, wheras the client was at run time?

I'll create an issue

PR #8004, opened 2022-09-15
Register a binder for parts that returns unsatisfied

### LINKED ISSUE #7939: OutOfMemory during huge file upload, when one of the nullable form data fields is empty
### Expected Behavior

Parsing of the missing parameters should not load other parameters (at least big files) into memory

### Actual Behaviour

If some of the nullabel parameters is missing in the request, it treats as an Unresolved reference. To resolve this reference it tries to read body into memory, but in my case this body contains a file of 5Gb

### Steps To Reproduce

@yawkat @graemerocher I found another issue related to this problem.

I have a controller with definition like this
image

If I'm uploading files and set all the fields it behaves normal, but if I'm not sending field "tags", the next problem happened:
1. AbstractRouteMatch treats it as an unresolved argument
image
2. And reads body with extremely huge "file" into memory
image
3. OOM
image

### Environment Information

Darwin MacBook-Pro-Andrey.local 21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:25 PDT 2022; root:xnu-8020.140.41~1/RELEASE_X86_64 x86_64

openjdk 17.0.3 2022-04-19
OpenJDK Runtime Environment Temurin-17.0.3+7 (build 17.0.3+7)
OpenJDK 64-Bit Server VM Temurin-17.0.3+7 (build 17.0.3+7, mixed mode, sharing)

### Example Application

_No response_

### Version

3.6.0

jameskleeh:
Fixes #7939 

This prevents @Part annotated arguments from reading query values, the body, etc if they are not sent. The only way I can test the original scenario is by uploading a file to disk that would cause an OOM when read. I don't know if CI can support that much space. Let me know if you want me to create the test.

### DIFF
--- a/http/src/main/java/io/micronaut/http/bind/DefaultRequestBinderRegistry.java
+++ b/http/src/main/java/io/micronaut/http/bind/DefaultRequestBinderRegistry.java
@@ -262,6 +262,9 @@ public class DefaultRequestBinderRegistry implements RequestBinderRegistry {
         RequestBeanAnnotationBinder<Object> requestBeanAnnotationBinder = new RequestBeanAnnotationBinder<>(this, conversionService);
         byAnnotation.put(requestBeanAnnotationBinder.getAnnotationType(), requestBeanAnnotationBinder);
 
+        PartAnnotationBinder<Object> partAnnotationBinder = new PartAnnotationBinder<>(conversionService);
+        byAnnotation.put(partAnnotationBinder.getAnnotationType(), partAnnotationBinder);
+
         if (KOTLIN_COROUTINES_SUPPORTED) {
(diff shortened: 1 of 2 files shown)


sdelamo:
> Fixes #7939
> 
> This prevents @part annotated arguments from reading query values, the body, etc if they are not sent. The only way I can test the original scenario is by uploading a file to disk that would cause an OOM when read. I don't know if CI can support that much space. Let me know if you want me to create the test.

Please, add the test. We can annotate with `@IgnoreIf` if we cannot run it in CI.

yawkat (approved):

yawkat:
i think this change should not go into a patch release

sdelamo:
@jameskleeh can you submit a test in a separate PR?

PR #8010, opened 2022-09-15
Update to Groovy 4.0

timyates:
This PR switches the Groovy version in core to 4.0.5

- [x] The internal plugin pulls in an old version of Groovy, so I added a dependency resolution block.  This can be removed once the internal plugin is updated for Micronaut 4.0.0
- [x] Geb doesn't support Groovy 4.x.  To re-enable the Geb tests, I have split these (one of them) out into a new module `test-groovy-geb` which runs with Groovy 3 on the classpath
- [x] Groovy no longer creates `isXXX` getters for `Boolean` fields (it does for `boolean`) -- this is [discussed over here](https://issues.apache.org/jira/browse/GROOVY-10762) on the Groovy Jira, and a fix has been added to only check `isXXX` for primitive booleans
- [x] Groovy can no longer grab a self-signed certificate using the old sun classes.  I have fixed this by adding bouncycastle onto the test classpath, so it uses that for the test
- [x] Groovy 4.x can no longer use the static methods on an interface declared in the class.  ie:  If an interface declares `static void foo()`, and `Concrete implements Interface` then Groovy must call `Interface.foo()`, not `Concrete.foo()`.  [This is an expected change -- I asked in the Groovy JIRA](https://issues.apache.org/jira/browse/GROOVY-10764)
- [x] `inject-kotlin-test/.../KotlinCompilerTest` fails with Groovy 4 as the `Note` for beans being created are picked up as a catastrophic failure by the test harness.  I added a check to ignore Notes, as this is only under test when Groovy is creating a kotlin build with Kapt processing


timyates:
I think one of the bugs I had to workaround is https://issues.apache.org/jira/browse/GROOVY-10060

inject-groovy/src/test/groovy/io/micronaut/inject/requires/RequiresBeanPropertiesSpec.groovy
```diff
@@ -111,6 +116,11 @@ import jakarta.inject.Singleton
 class Config
 {
     Boolean boolProperty
+
+    // TODO: Why do we need this?  Groovy bug?
```
timyates: Raised https://issues.apache.org/jira/browse/GROOVY-10762 to see if it's a bug

timyates: It's expected and not a bug, fixed in https://github.com/micronaut-projects/micronaut-core/pull/8010/commits/fd8d88542740b94e192593203564274ca77f6874

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
```diff
@@ -46,6 +46,15 @@ tasks.withType(Jar).configureEach {
     preserveFileTimestamps = false
 }
 
+configurations.all {
+    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+        if (details.requested.group == 'org.codehaus.groovy') {
```
melix, line 51: So while this will work to build and execute this module, this won't be propagated to consumers.

inject-kotlin-test/build.gradle
```diff
@@ -43,3 +43,20 @@ tasks.named("compileGroovy") {
     dependsOn tasks.getByPath('compileKotlin')
     classpath += files(compileKotlin.destinationDir)
 }
+
+tasks.named("test", Test) {
+    if(JavaVersion.current().majorVersion.toInteger() >= 17) {
+        jvmArgs(
+                '--add-opens', 'jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED',
```
melix, line 50: This whole block is a bit scary!

test-suite-geb/build.gradle
```diff
@@ -0,0 +1,50 @@
...
+    testImplementation "cglib:cglib-nodep:3.3.0"
+    testImplementation "org.objenesis:objenesis:3.1"
+    testImplementation libs.managed.netty.codec.http
+    testImplementation(project(":inject-groovy")) {
+        exclude module: 'groovy'
+    }
+
+    testImplementation libs.geb.spock
+    testImplementation libs.spock.for.geb
+    testImplementation "org.codehaus.groovy:groovy-test:3.0.12"
```
melix: Is this intentional that it uses an old version?

test-suite/build.gradle
```diff
@@ -22,8 +22,8 @@ micronautBuild {
     }
 }
 
-apply from: "${rootProject.projectDir}/gradle/geb.gradle"
-apply from: "${rootProject.projectDir}/gradle/webdriverbinaries.gradle"
+//apply from: "${rootProject.projectDir}/gradle/geb.gradle"
```
melix: You probably want to remove those comments

validation/build.gradle
```diff
@@ -15,7 +15,9 @@ dependencies {
     api project(":core-reactive")
     api libs.managed.validation
 
-    compileOnly libs.managed.gorm
+    compileOnly(libs.managed.gorm) {
+        exclude(module: 'groovy')
```
melix, line 19: mmm, we should figure out why an exclude is needed here

inject-kotlin-test/build.gradle
timyates: I agree, but without it we get

```
java.lang.IllegalAccessError: class org.jetbrains.kotlin.kapt3.base.KaptContext (in unnamed module @0x7a4f0f29) cannot access class com.sun.tools.javac.util.Context (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.util to unnamed module @0x7a4f0f29
```

All the way down until you end up with the above list.  I *think* it's because we're effectively hand-cranking Kapt from Groovy, but I have no proof...

test-suite-geb/build.gradle
timyates: Yes, `test-suite-geb` runs Geb on Groovy 3 (as there's no Geb that works with Groovy 4 -- [yet](https://github.com/geb/issues/issues/668))

test-suite/build.gradle
timyates: 👍

validation/build.gradle
timyates: `libs.managed.gorm` is `org.grails:grails-datastore-core:7.3.2` at time of writing which has a [compile dependency on Groovy 3](https://search.maven.org/artifact/org.grails/grails-datastore-core/7.3.2/jar)

```
    <dependency>
      <groupId>org.codehaus.groovy</groupId>
      <artifactId>groovy</artifactId>
      <version>3.0.11</version>
      <scope>compile</scope>
    </dependency>
```

This is just a compile only dependency, but we may be breaking gorm integration until grals release a Groovy 4 version of their libraries that we consume

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
timyates: I think that's ok...  It's only stopping the micronaut-build internal plugin pulling in the codehaus dependency here

Once micronaut-build is updated, I believe we can get rid of this

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/HttpResponseSpec.groovy
```diff
@@ -234,9 +234,9 @@ class HttpResponseSpec extends AbstractMicronautSpec {
 
     void "test keep alive connection header is not set by default for > 499 response"() {
         when:
-        EmbeddedServer server = applicationContext.run(EmbeddedServer, [(SPEC_NAME_PROPERTY):getClass().simpleName])
+        EmbeddedServer server = ApplicationContext.run(EmbeddedServer, ['micronaut.server.dateHeader': false, (SPEC_NAME_PROPERTY):getClass().simpleName])
```
sdelamo: this should be `micronaut.sever.date-header`. Lower kebab case.

inject-groovy-test/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
         exclude module:'groovy-all'
     }
     api project(":context")
+    api libs.jetbrains.annotations
```
sdelamo, line 14: why do we need this?

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java
```diff
@@ -906,7 +906,7 @@ public Optional<MethodElement> getReadMethod() {
                     private String getGetterName(String propertyName, ClassElement type) {
                         return NameUtils.getterNameFor(
                                 propertyName,
-                                type.equals(PrimitiveElement.BOOLEAN) || type.getName().equals(Boolean.class.getName())
+                                type.equals(PrimitiveElement.BOOLEAN)
```
sdelamo, line 909: Is this because what you mentioned yesterday about Groovy 4 no longer generating the `isField()` method for `Boolean` fields?

inject-java-test/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
     if (!JavaVersion.current().isJava9Compatible()) {
         api files(org.gradle.internal.jvm.Jvm.current().toolsJar)
     }
+    api libs.jetbrains.annotations
```
sdelamo, line 20: why do we need this?

sdelamo (changes requested):
Should we remove GORM information from the documentation? 

Maybe add it later once there is a version of GORM compatible with Groovy 4. 

Or add an Asciidoc callout?

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java
timyates: Yes

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/HttpResponseSpec.groovy
timyates: Done

sdelamo (approved):

inject-groovy-test/build.gradle
sdelamo: To support code highlighting: 

`@Language('java') String someCode`

inject-java-test/build.gradle
sdelamo: To support code highlighting: 

`@Language('java') String someCode`

melix (approved):

PR #8016, opened 2022-09-16
add an option to completely disable default property source loading

graemerocher:
This adds the ability to disable loading any default property sources if the user wants to control completely what property sources are included

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
@@ -89,6 +89,7 @@ public class DefaultApplicationContextBuilder implements ApplicationContextBuild
     }
 
     @Override
+    @NonNull 
     public ApplicationContextBuilder enableDefaultPropertySources(boolean areEnabled) {
         this.enableDefaultPropertySources = areEnabled;
         return this;
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
@@ -66,6 +66,7 @@ public class DefaultApplicationContextBuilder implements ApplicationContextBuild
     private ClassPathResourceLoader classPathResourceLoader;
     private boolean allowEmptyProviders = false;
     private Boolean bootstrapEnvironment = null;
+    private boolean enableDefaultPropertySources = true;
 
     /**
      * Default constructor.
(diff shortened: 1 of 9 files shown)


dstepanov:
I think the alternative to this should be setting `overrideConfigLocations` to nothing and disabling system and env property sources.

inject/src/main/java/io/micronaut/context/DefaultApplicationContextBuilder.java
```diff
@@ -87,6 +88,17 @@ public boolean isAllowEmptyProviders() {
         return allowEmptyProviders;
     }
 
+    @Override
+    public ApplicationContextBuilder enableDefaultPropertySources(boolean areEnabled) {
```
sdelamo, line 93:
```suggestion
    @NonNull 
    public ApplicationContextBuilder enableDefaultPropertySources(boolean areEnabled) {
```

sdelamo (changes requested):
Can you add documentation?

sdelamo:
What is the goal of this PR? Faster startup times when you don't want to lose time reading property sources on start-up? Should we add a member to the `@MicronautTest`annotation ? `@MicronautTest(enableDefaultPropertySources = false)`

graemerocher:
It isn't about startup time. Some teams do not want to load configuration from any other location other than the ones they specify manually via `builder.propertySources(..)` for security reasons. In addition for Micronaut for Spring and the starter we don't actually need Micronaut to load the property sources because these are already loaded by Spring so I think it will be valuable for that integration as well

graemerocher:
> Can you add documentation?

The documentation is the javadoc? Don't think this needs specific documentation in the user guide.

graemerocher:
On second though, I think docs explaining the behaviour would be good so I added some

graemerocher:
> I think the alternative to this should be setting `overrideConfigLocations` to nothing and disabling system and env property sources.

Think that is more work than needed, this adds a simply way to achieve that

graemerocher:
> What is the goal of this PR? Faster startup times when you don't want to lose time reading property sources on start-up? Should we add a member to the `@MicronautTest`annotation ? `@MicronautTest(enableDefaultPropertySources = false)`

we could but that will require a downstream change in MicronautTest that would need to be done at a later date, but like I said the goal is not faster startup.

sdelamo (approved):

sdelamo:
@dstepanov do you have any concerns merging this?

dstepanov (approved):

PR #8028, opened 2022-09-18
Fix incorrect getPropertiesFromGettersAndSetters groovy visitor for interfaces

altro3:
From micronaut openapi: found problem with standard endpoins instrospection. When visitor try to get bean properties for Interface `HealthResult` - it can't find any properties. Problem in GroovyClassElement.java:
изображение
method isAcceptible return false, because method abstract.

With java all works correct.


dstepanov:
Please add a test

altro3:
@dstepanov Hi! Yes, I tried, but I need help. I couldn't write a test with an interface so that getPropertiesFromGettersAndSetters was called :-(. I didn't find such examples in the existing tests.

altro3:
@dstepanov could you help?

graemerocher:
You can write a test that subclasses `AbstractBeanDefinitionSpec` and calls `buildClassElement`

altro3:
@graemerocher thanks! Done

graemerocher (approved):

sdelamo (approved):

PR #8061, opened 2022-09-21
Make SnakeYAML and javax.annotation optional

### LINKED ISSUE #7042: Remove the direct dependency on SnakeYAML
### Feature description

SnakeYAML should be an optional feature and a YAML feature added to launch that makes it easy to remove it from the dependency graph.

graemerocher:
This PR makes SnakeYAML an optional dependency as well as `javax.annotation` since we have moved to `jakarta.annotation`

Fixes #7042


graemerocher:
test failures are unrelated

sdelamo (changes requested):
I think it would be good to add to [Application Configuration](https://docs.micronaut.io/snapshot/guide/#config) section an asciidoc `NOTE`. 

Something, like 

> To use `application.yml` you need to add a dependency to [SnakeYAML](https://bitbucket.org/snakeyaml/snakeyaml/src/master/)

donbeave (approved):

donbeave:
Do I still need to add a snakeyaml dependency if I use the Micronaut Gradle AOT plugin with enabled convertYamlToJava = true?

sdelamo:
> Do I still need to add a snakeyaml dependency if I use the Micronaut Gradle AOT plugin with enabled convertYamlToJava = true?

With micronaut 3.x you will still have the dependency even if you use convertYamlToJava = true. It is unused but it is on your classpath. 

For Micronaut 4, you could have it as compileOnly, right @melix ?

melix:
It would only make sense to have it `compileOnly` **iff** the only binaries being used are the AOT optimized ones **and** that the conversion of yaml to Java is enabled. In practice, this will never be the case, because AOT is a _deployment_ optimization, you don't run AOT for day-to-day development, even when calling `gradle run` (because the analysis would be too expensive). Similarly, you may not want to use AOT in all cases, so I would recommend to keep the dependency as is.

You could do also `compileOnly` and `developmentOnly`, if you deploy the AOT optimized with `convertYamlToJava = true`

sdelamo (approved):

PR #8068, opened 2022-09-21
Disable Cloud environment deduction by default. Fixes #7758

### LINKED ISSUE #7758: Disable environment deduction by default in Micronaut 4.0
### Issue description

Micronaut currently by default attempts to deduce the cloud environment it is running, whilst convenient this is comes at a performance cost and also is potentially not secure since the logic to deduce the environment has to probe remote endpoints.

This functionality should be disabled by default and enabled on a case by case basis by the developer.

graemerocher:
* Introduces a new method ` deduceCloudEnvironment` to `ApplicationContextBuilder` that allows re-enabling of this functionality
* Retains environment deduction only for the test environment


dstepanov:
Maybe we can have the env detectors being contributed by the cloud support modules, so aws, azure, oracle cloud will contribute own detectors but the pure core will not have it?

sdelamo (changes requested):
This PR should update as well: 

https://docs.micronaut.io/snapshot/guide/#environments

It currently talks about environment deduction being on and how to disable it. We should update to tell the oposite. 

moreover, in addition to the mention in breaks we need a section in the documentation talking specifically about cloud environment deduction.

sdelamo:
> Maybe we can have the env detectors being contributed by the cloud support modules, so aws, azure, oracle cloud will contribute own detectors but the pure core will not have it?

In think we could do that in addition to this PR. We should move that into a separate issue.

graemerocher:
> Maybe we can have the env detectors being contributed by the cloud support modules, so aws, azure, oracle cloud will contribute own detectors but the pure core will not have it?

good idea

dstepanov (approved):

sdelamo:
> This PR should update as well:
> 
> https://docs.micronaut.io/snapshot/guide/#environments
> 
> It currently talks about environment deduction being on and how to disable it. We should update to tell the oposite.
> 
> moreover, in addition to the mention in breaks we need a section in the documentation talking specifically about cloud environment deduction.

@graemerocher can you do this doc update?

graemerocher:
> > This PR should update as well:
> > https://docs.micronaut.io/snapshot/guide/#environments
> > It currently talks about environment deduction being on and how to disable it. We should update to tell the oposite.
> > moreover, in addition to the mention in breaks we need a section in the documentation talking specifically about cloud environment deduction.
> 
> @graemerocher can you do this doc update?

@sdelamo I have added additional docs. There is already a section in the documentation that talks about cloud environment deduction. I have updated this section.

sdelamo (approved):

PR #8070, opened 2022-09-22
Bump Micronaut Test Resources version

melix:


graemerocher:
Shouldn't this target 3.7.x?

melix:
I was pretty conservative by targetting 3.6.x but we can target 3.7.

sdelamo:
> I was pretty conservative by targetting 3.6.x but we can target 3.7.

Yes, it should target 3.7.x. We don't bump minor version of modules in patch releases.

sdelamo (approved):
I have changed the base and resolved the merge.

PR #8074, opened 2022-09-22
Bump managed kotlin version to 1.7.10

wetted:
satisfies https://github.com/micronaut-projects/micronaut-kotlin/issues/414


sdelamo (changes requested):
Misses docs

timyates:
We see lots of 1.6.21 in the dependency list

https://ge.micronaut.io/s/bjjijqfihhmsu/dependencies?dependencies=1.6.21&expandAll

But I believe these are due to the POM for [org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.6.4](https://search.maven.org/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-jdk8/1.6.4/jar) having a dependency on it

I think none of these are being pulled by Gradle, but when they are it's picking the correct one

https://ge.micronaut.io/s/bjjijqfihhmsu/dependencies?dependencies=1.6.21&expandAll&focusedDependency=WzQxLDEsNDI0OCxbNiwwLFsyNl1dXQ

@melix is my understanding correct?

buildSrc/src/main/groovy/io/micronaut/build/internal/ext/DefaultMicronautCoreExtension.java
```diff
@@ -60,6 +60,11 @@ public void usesMicronautTestKotest() {
         addTestImplementationDependency("kotest");
     }
 
+    @Override
```
sdelamo: can we get rid of kotest4 ?

buildSrc/src/main/groovy/io/micronaut/build/internal/ext/MicronautCoreExtension.java
```diff
@@ -28,4 +28,6 @@ public interface MicronautCoreExtension {
 
     void usesMicronautTestKotest();
 
+    void usesMicronautTestKotest5();
```
sdelamo: can we just use kotest5 instaed of mantaining both?

sdelamo (changes requested):

buildSrc/src/main/groovy/io/micronaut/build/internal/ext/DefaultMicronautCoreExtension.java
timyates: Gone

buildSrc/src/main/groovy/io/micronaut/build/internal/ext/MicronautCoreExtension.java
timyates: Gone

sdelamo: why create a new method instead of keeping old one?

timyates: Fixed in https://github.com/micronaut-projects/micronaut-core/pull/8074/commits/624810d8a859c728f53593a814bd9a3a5247d0e2

sdelamo (approved):

sdelamo:
ping @melix

melix (approved):

Issue #7988, opened 2022-09-12
Remove deprecated code in 4.0.x branch for Micronaut Core


sdelamo:
### Issue description

Remove code annotated with `@Deprecated` and a mention to what's new telling users that deprecated code has been removed.

PR #8089, opened 2022-09-27
Remove deprecated code in Micronaut Core for 4.0.x (partial)

wetted:
This PR removes deprecations for `micronaut-core` but is mostly the low-hanging fruit. There are still numerous instances of `@Deprecation` elements, but which are still used elsewhere in src and test code. Some of those remaining deprecations are in `inject` and `http` code, and require a bit of refactoring at a minimum.

This is a partial solution for #7988


context/src/main/java/io/micronaut/runtime/server/watch/event/FileWatchRestartListener.java
```diff
@@ -53,15 +52,6 @@ public FileWatchRestartListener(EmbeddedApplication<?> embeddedApplication) {
         this.embeddedApplication = embeddedApplication;
     }
 
-    /**
-     * @param embeddedServer The embedded server
-     * @deprecated Use {@link #FileWatchRestartListener(EmbeddedApplication)} instead.
-     */
-    @Deprecated
```
sdelamo, line 60: the `@Inject` in the other constructor can be be removed?

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaPropertyElement.java
```diff
@@ -71,30 +69,31 @@ class JavaPropertyElement extends AbstractJavaElement implements PropertyElement
...
+//     * @param declaringElement   The declaring element
+//     * @param rootElement        The element
+//     * @param annotationMetadata The annotation metadata
+//     * @param name               The name
+//     * @param type               The type
+//     * @param readOnly           Whether it is read only
+//     * @param visitorContext     The java visitor context
+//     * @deprecated Use {@link #JavaPropertyElement(ClassElement, Element, AnnotationMetadata, String, ClassElement, boolean, JavaVisitorContext)} instead
+//     */
+//    @SuppressWarnings("DeprecatedIsStillUsed") // used by openapi processor
```
sdelamo: We should not leave it commented it out. Does the OpenAPI processor use it? If the OpenAPI processor uses it, revert this change so we can merge this Pr.

yawkat (approved):
agree with @sdelamo on the commented code

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaPropertyElement.java
wetted: That was unintentional. I meant to delete it.

yawkat (approved):

sdelamo (approved):

PR #8100, opened 2022-09-30
New HTTP client connection pooling

yawkat:
This PR adds a new connection pooling infrastructure to the HTTP client. The biggest changes are in ConnectionManager, which contains network setup code, and in the new PoolResizer class, which handles pool dynamics (it calls into ConnectionManager.Pool to assign requests, establish new connections, etc, but does no network setup by itself).

Changes in no particular order:

- The `http-version` config setting has been replaced by two settings. `plaintext-mode` applies to `http` URLs and can be set to either `HTTP_1` or `H2C`. `alpn-modes` applies to `https` URLs, and determines the protocols that should be supported in protocol negotiation (it's a list, with the supported values `h2` and `http/1.1`). The old http-version setting remains for compatibility, and should remain for 4.0. It is also used in some test cases.
- connection pooling now applies to all connections (except websockets), not just exchange calls, and it is enabled by default.
- HTTP2 connections now use a channel-per-stream model (like #6842), and allow concurrent requests on one connection.
- Connection pool configuration is more fine-grained, with separate settings for HTTP1 and HTTP2.
- Pooled connections are now tracked using netty's resource leak detection, to avoid "dangling" connections in the pool (note: resource leak detection is still only tested widely for the tests in the http-server-netty module atm).
- One minor fix in the http server that came up in the test suite because connection pooling is on by default now.

There are still some TODOs in this PR. Some relate to pending netty improvements (https://github.com/netty/netty/issues/12827 , https://github.com/netty/netty/pull/12830), but work fine for the moment. Some are future improvements that could be made, but that I want to do in separate PRs.


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/HttpResponseSpec.groovy
```diff
@@ -232,6 +229,7 @@ class HttpResponseSpec extends AbstractMicronautSpec {
         server.close()
     }
 
+    @Ignore // why wouldn't keep-alive be set?
     void "test keep alive connection header is not set by default for > 499 response"() {
```
timyates: For my own education, are we getting 'keep-alive' now as the underlying netty server has changed to return 'keep-alive' in the error response? Or are we getting it now because or the client changes and the fact we're now using a pool by default?

I've done some looking around and it seems historically it made some sense to 'close' on errors for a server serving html to people (users are likely to just go elsewhere), but I don't think there's an issue here.

I was just interested in what caused this change in behaviour.

If we stick with keep-alive, and it's being set by the server, I guess we can remove `micronaut.server.netty.keepAliveOnServerError`?  Or do we need config to switch back to use the old close header?

timyates (approved):
Been through locally, and no issues jump out 👍 

I need to improve my netty knowledge though.

Just one question around keep-alive

timyates (changes requested):
I guess this needs updating too?

https://docs.micronaut.io/latest/guide/index.html#_using_http_client_connection_pooling

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/HttpResponseSpec.groovy
yawkat: > For my own education, are we getting 'keep-alive' now as the underlying netty server has changed to return 'keep-alive' in the error response? Or are we getting it now because or the client changes and the fact we're now using a pool by default?

Yea, it's the latter. The server never forced `close`, but because before this pr the client requested using `Connection: close`, so the server replied with that as well. Now that the pool is enabled, the server is perfectly happy to stay `keep-alive`. I guess this test case just didn't test for the feature properly.

> If we stick with keep-alive, and it's being set by the server, I guess we can remove micronaut.server.netty.keepAliveOnServerError? Or do we need config to switch back to use the old close header?

I didn't know about that setting, I will take a look at why it doesn't work as intended.

timyates: > I didn't know about that setting, I will take a look at why it doesn't work as intended.

I only saw it when I was trying the Ignored test locally 🕵️‍♀️

yawkat: the logic was a bit weird, ive fixed it and enabled the setting by default.

yawkat:
Not sure how to replace that section, since pooling doesn't need to be manually enabled anymore. It could be a section that basically says "pooling exists, good for you", but this wouldn't really fit into the guide?

timyates:
> Not sure how to replace that section, since pooling doesn't need to be manually enabled anymore. It could be a section that basically says "pooling exists, good for you", but this wouldn't really fit into the guide?

Maybe a section around configuring the pool which shows the defaults, and how to turn it off if you hate pools?

yawkat:
that would make sense. that's a task for future me :)

yawkat:
@timyates WDYT

http-client/src/main/java/io/micronaut/http/client/netty/PoolResizer.java
```diff
@@ -221,6 +222,10 @@ final void onConnectionInactive2(ResizerConnection connection) {
     }
 
     final void addPendingRequest(Sinks.One<ConnectionManager.PoolHandle> sink) {
+        if (pendingRequests.size() >= connectionPoolConfiguration.getMaxPendingAcquires()) {
+            sink.tryEmitError(new HttpClientException("Cannot acquire connection, exceeded max pending acquires configuration"));
+            return;
+        }
```
timyates, line 228: Do we have a config test for maxPendingAcquires and `max-concurrent-http1-connections`?

timyates (approved):

http-client/src/main/java/io/micronaut/http/client/netty/PoolResizer.java
yawkat: ive added tests for both. i dont like the pure config tests like in DefaultHttpClientConfigurationSpec

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -637,6 +657,56 @@ public Proxy resolveProxy(boolean isSsl, String host, int port) {
...
 
+    /**
+     * The connection mode to use for <i>plaintext</i> (http as opposed to https) connections.
+     * <br>
+     * <b>Note: If {@link #httpVersion} is set, this setting is ignored!</b>
+     *
+     * @return The plaintext connection mode.
+     * @since 4.0.0
+     */
+    public HttpVersionSelection.PlaintextMode getPlaintextMode() {
```
graemerocher, line 669: add nullability annotations

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -637,6 +657,56 @@ public Proxy resolveProxy(boolean isSsl, String host, int port) {
...
+
+    /**
+     * The connection mode to use for <i>plaintext</i> (http as opposed to https) connections.
+     * <br>
+     * <b>Note: If {@link #httpVersion} is set, this setting is ignored!</b>
+     *
+     * @param plaintextMode The plaintext connection mode.
+     * @since 4.0.0
+     */
+    public void setPlaintextMode(HttpVersionSelection.PlaintextMode plaintextMode) {
```
graemerocher: add nullability annotations

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -637,6 +657,56 @@ public Proxy resolveProxy(boolean isSsl, String host, int port) {
...
+    /**
+     * The protocols to support for TLS ALPN. If HTTP 2 is included, this will also restrict the
+     * TLS cipher suites to those supported by the HTTP 2 standard.
+     * <br>
+     * <b>Note: If {@link #httpVersion} is set, this setting is ignored!</b>
+     *
+     * @return The supported ALPN protocols.
+     * @since 4.0.0
+     */
+    public List<String> getAlpnModes() {
```
graemerocher, line 695: add nullability annotations

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -637,6 +657,56 @@ public Proxy resolveProxy(boolean isSsl, String host, int port) {
...
+    /**
+     * The protocols to support for TLS ALPN. If HTTP 2 is included, this will also restrict the
+     * TLS cipher suites to those supported by the HTTP 2 standard.
+     * <br>
+     * <b>Note: If {@link #httpVersion} is set, this setting is ignored!</b>
+     *
+     * @param alpnModes The supported ALPN protocols.
+     * @since 4.0.0
+     */
+    public void setAlpnModes(List<String> alpnModes) {
```
graemerocher: add nullability annotations

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -685,24 +753,6 @@ public void setEnabled(boolean enabled) {
...
-
-        /**
-         * Sets the maximum number of connections. Defaults to no maximum.
-         *
-         * @param maxConnections The count
-         */
-        public void setMaxConnections(int maxConnections) {
-            this.maxConnections = maxConnections;
-        }
-
```
graemerocher, line 705: can we deprecate instead of remove?

http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java
```diff
@@ -0,0 +1,195 @@
...
+        new String[]{ALPN_HTTP_1, ALPN_HTTP_2},
+        true
+    );
+
+    private final PlaintextMode plaintextMode;
+    private final boolean alpn;
+    private final String[] alpnSupportedProtocols;
+    private final boolean http2CipherSuites;
+
+    private HttpVersionSelection(PlaintextMode plaintextMode, boolean alpn, String[] alpnSupportedProtocols, boolean http2CipherSuites) {
```
graemerocher: nullability annotations on the arguments

http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * Get the {@link HttpVersionSelection} that matches Micronaut HTTP client 3.x behavior for the
+     * given version setting.
+     *
+     * @param httpVersion The HTTP version as configured for Micronaut HTTP client 3.x
+     * @return The version selection
+     */
+    public static HttpVersionSelection forLegacyVersion(HttpVersion httpVersion) {
```
graemerocher: add nullability annotations

dstepanov:
Would it be possible to refactor `DefaultHttpClient` to be created using the builder? I just had to add one additional parameter to the constructor and its pain.

yawkat:
it would definitely be an improvement

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
             this.acquireTimeout = acquireTimeout;
         }
+
+        /**
+         * The maximum number of <i>pending</i> (new) connections before they are assigned to a
+         * pool.
+         *
+         * @return The maximum number of pending connections
+         */
+        public int getMaxPendingConnections() {
```
graemerocher, line 801: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            return maxPendingConnections;
+        }
+
+        /**
+         * The maximum number of <i>pending</i> (new) connections before they are assigned to a
+         * pool.
+         *
+         * @param maxPendingConnections The maximum number of pending connections
+         */
+        public void setMaxPendingConnections(int maxPendingConnections) {
```
graemerocher, line 812: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            this.maxPendingConnections = maxPendingConnections;
+        }
+
+        /**
+         * The maximum number of requests (streams) that can run concurrently on one HTTP2
+         * connection.
+         *
+         * @return The maximum concurrent request count
+         */
+        public int getMaxConcurrentRequestsPerHttp2Connection() {
```
graemerocher, line 823: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            return maxConcurrentRequestsPerHttp2Connection;
+        }
+
+        /**
+         * The maximum number of requests (streams) that can run concurrently on one HTTP2
+         * connection.
+         *
+         * @param maxConcurrentRequestsPerHttp2Connection The maximum concurrent request count
+         */
+        public void setMaxConcurrentRequestsPerHttp2Connection(int maxConcurrentRequestsPerHttp2Connection) {
```
graemerocher, line 834: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            this.maxConcurrentRequestsPerHttp2Connection = maxConcurrentRequestsPerHttp2Connection;
+        }
+
+        /**
+         * The maximum number of concurrent HTTP1 connections in the pool.
+         *
+         * @return The maximum concurrent connection count
+         */
+        public int getMaxConcurrentHttp1Connections() {
```
graemerocher, line 844: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            return maxConcurrentHttp1Connections;
+        }
+
+        /**
+         * The maximum number of concurrent HTTP1 connections in the pool.
+         *
+         * @param maxConcurrentHttp1Connections The maximum concurrent connection count
+         */
+        public void setMaxConcurrentHttp1Connections(int maxConcurrentHttp1Connections) {
```
graemerocher, line 854: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            this.maxConcurrentHttp1Connections = maxConcurrentHttp1Connections;
+        }
+
+        /**
+         * The maximum number of concurrent HTTP2 connections in the pool.
+         *
+         * @return The maximum concurrent connection count
+         */
+        public int getMaxConcurrentHttp2Connections() {
```
graemerocher, line 864: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -738,5 +788,82 @@ public Optional<Duration> getAcquireTimeout() {
...
+            return maxConcurrentHttp2Connections;
+        }
+
+        /**
+         * The maximum number of concurrent HTTP2 connections in the pool.
+         *
+         * @param maxConcurrentHttp2Connections The maximum concurrent connection count
+         */
+        public void setMaxConcurrentHttp2Connections(int maxConcurrentHttp2Connections) {
```
graemerocher, line 874: add `@since 4.0.0`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientRegistry.java
```diff
@@ -47,9 +47,24 @@
...
+    /**
+     * Return the client for the client ID and path.
+     *
+     * @param httpVersion The HTTP version
+     * @param clientId    The client ID
+     * @param path        The path (Optional)
+     * @return The client
+     */
+    @NonNull
+    T getClient(HttpVersionSelection httpVersion, @NonNull String clientId, @Nullable String path);
```
graemerocher: add `@since 4.0.0`, specify nullability for `httpVersion`

http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.Arrays;
+
+/**
+ * This class collects information about HTTP client protocol version settings, such as the
+ * {@link PlaintextMode} and the ALPN configuration.
+ *
+ * @author Jonas Konrad
+ * @since 4.0
+ */
+public final class HttpVersionSelection {
```
graemerocher, line 34: perhaps this can be a record since we are in the Java 17 world

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
```diff
@@ -80,6 +81,34 @@
...
+    /**
+     * The protocols to support for TLS ALPN. If HTTP 2 is included, this will also restrict the
+     * TLS cipher suites to those supported by the HTTP 2 standard.
+     * <br>
+     * <b>Note: If {@link #httpVersion} is set, this setting is ignored!</b>
+     *
+     * @return The supported ALPN protocols.
+     * @since 4.0.0
+     */
+    String[] alpnModes() default {
```
graemerocher, line 113: specify nullability

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -302,6 +229,42 @@ private static NioEventLoopGroup createEventLoopGroup(HttpClientConfiguration co
         return group;
     }
 
+    /**
+     * For testing.
+     *
+     * @return Connected channels in all pools
+     */
+    @SuppressWarnings("unused")
+    List<Channel> getChannels() {
```
graemerocher, line 240: specify nullability and add `@since 4.0.0`

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -302,6 +229,42 @@ private static NioEventLoopGroup createEventLoopGroup(HttpClientConfiguration co
...
+        List<Channel> channels = new ArrayList<>();
```
graemerocher, line 241: size the list based on `pools.values()` size?

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -302,6 +229,42 @@ private static NioEventLoopGroup createEventLoopGroup(HttpClientConfiguration co
...
+        return channels;
+    }
+
+    /**
+     * For testing.
+     *
+     * @return Number of running requests
+     */
+    @SuppressWarnings("unused")
+    int liveRequestCount() {
```
graemerocher, line 255: add `@since 4.0.0`

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -903,423 +474,758 @@ private <E extends HttpClientException> E customizeException(E exc) {
...
+            }
+        });
+    }
 
-        private final HttpClientInitializer initializer;
+    /**
+     * Initializer for TLS channels. After ALPN we will proceed either with
+     * {@link #initHttp1(Channel)} or {@link #initHttp2(Pool, Channel, NettyClientCustomizer)}.
+     */
+    private class AdaptiveAlpnChannelInitializer extends ChannelInitializer<Channel> {
```
graemerocher: can this be `final`?

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -903,423 +474,758 @@ private <E extends HttpClientException> E customizeException(E exc) {
...
-            if (hostString == null) {
-                hostString = remote.getAddress().getHostAddress();
-            }
-            upgradeRequest.headers().set(HttpHeaderNames.HOST, hostString + ':' + remote.getPort());
-            ctx.writeAndFlush(upgradeRequest);
+    /**
+     * Initializer for H2C connections. Will proceed with
+     * {@link #initHttp2(Pool, Channel, NettyClientCustomizer)} when the upgrade is done.
+     */
+    private class Http2UpgradeInitializer extends ChannelInitializer<Channel> {
```
graemerocher: can this be `final`?

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -903,423 +474,758 @@ private <E extends HttpClientException> E customizeException(E exc) {
...
+                );
+            } else {
+                switch (httpVersion.getPlaintextMode()) {
+                    case HTTP_1:
+                        initializer = new ChannelInitializer<Channel>() {
+                            @Override
+                            protected void initChannel(@NonNull Channel ch) throws Exception {
+                                configureProxy(ch.pipeline(), false, requestKey.getHost(), requestKey.getPort());
+                                initHttp1(ch);
+                                ch.pipeline().addLast("activity-listener", new ChannelInboundHandlerAdapter() {
```
graemerocher: can this handler name be specified as a constant that describes that it is used for?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1016,8 +994,39 @@ private <I> Flux<MutableHttpResponse<Object>> connectAndStream(
...
+                    super.channelRead(ctx, msg);
+
+                    if (msg instanceof LastHttpContent) {
+                        if (ignoreOneLast) {
+                            ignoreOneLast = false;
+                        } else {
+                            ctx.pipeline()
+                                .remove(this)
+                                .remove(ChannelPipelineCustomizer.HANDLER_HTTP_STREAM);
+                            poolHandle.release();
```
graemerocher: are there other circumstances where this handler needs to call `release()`? What about handler removed events?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1526,16 +1546,32 @@ private <R extends HttpResponse<?>> Flux<R> handleStreamHttpError(
...
-            FluxSink emitter,
+            FluxSink<? super MutableHttpResponse<?>> emitter,
             ConnectionManager.PoolHandle poolHandle,
             boolean secure) throws HttpPostRequestEncoder.ErrorDataEncoderException {
-        NettyRequestWriter requestWriter = prepareRequest(
-                finalRequest,
-                finalRequest.getUri(),
-                emitter
+        if (!(finalRequest instanceof MutableHttpRequest)) {
+            throw new IllegalArgumentException("A MutableHttpRequest is required");
```
graemerocher: can we re-construct a mutable http request if the request is not one?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1949,23 +1953,21 @@ private class NettyRequestWriter {
          * @param channelPool The channel pool
          * @param emitter     The emitter
          */
-        protected void write(Channel channel, boolean isSecure, FluxSink<?> emitter) {
-            final ChannelPipeline pipeline = channel.pipeline();
-            if (connectionManager.httpVersion == io.micronaut.http.HttpVersion.HTTP_2_0) {
+        protected void write(ConnectionManager.PoolHandle poolHandle, boolean isSecure, FluxSink<?> emitter) {
+            if (poolHandle.http2) {
+                // todo: move to CM
```
graemerocher: can you describe more what this TODO is here for?

http-client/src/main/java/io/micronaut/http/client/netty/HttpLineBasedFrameDecoder.java
```diff
@@ -0,0 +1,101 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.LastHttpContent;
+
+/**
+ * Variant of {@link LineBasedFrameDecoder} that accepts
+ * {@link io.netty.handler.codec.http.HttpContent} data. Note: this handler removes itself when the
+ * response has been consumed.
+ */
+@Internal
+final class HttpLineBasedFrameDecoder extends LineBasedFrameDecoder {
```
graemerocher, line 40: add `@since 4.0.0`

http-client/src/main/java/io/micronaut/http/client/netty/HttpLineBasedFrameDecoder.java
```diff
@@ -0,0 +1,101 @@
...
+    static final String NAME = ChannelPipelineCustomizer.HANDLER_MICRONAUT_SSE_EVENT_STREAM;
+
+    private boolean ignoreOneLast = false;
+
+    public HttpLineBasedFrameDecoder(int maxLength, boolean stripDelimiter, boolean failFast) {
```
graemerocher: does this need to be `public`?

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientCustomizer.java
```diff
@@ -79,6 +79,10 @@ enum ChannelRole {
          * {@link io.netty.channel.socket.SocketChannel}, representing an HTTP connection.
          */
         CONNECTION,
+        /**
+         * The channel is a HTTP2 stream channel.
+         */
+        HTTP2_STREAM,
```
graemerocher, line 87: add `@since 4.0.0`

src/main/docs/guide/httpClient/lowLevelHttpClient/clientConfiguration.adoc
```diff
@@ -75,9 +75,17 @@ Alternatively, if you don't use service discovery you can use the `configuration
 ReactorHttpClient httpClient;
 ----
 
-=== Using HTTP Client Connection Pooling
+=== Connection Pooling and HTTP 2
```
graemerocher: Replace `HTTP 2` with `HTTP/2`

src/main/docs/guide/httpClient/lowLevelHttpClient/clientConfiguration.adoc
```diff
@@ -75,9 +75,17 @@ Alternatively, if you don't use service discovery you can use the `configuration
...
 
-A client that handles a significant number of requests will benefit from enabling HTTP client connection pooling. The following configuration enables pooling for the `foo` client:
+Connections using normal HTTP (without TLS/SSL) use HTTP 1.1. This can be configured using the `plaintext-mode` configuration option.
+
+Secure connections (i.e. HTTP**S**, with TLS/SSL) use a feature called "Application Layer Protocol Negotiation" (ALPN) that is part of TLS to select the HTTP version. If the server supports HTTP 2, the Micronaut HTTP Client will use that capability by default, but if it doesn't, HTTP 1.1 is still supported. This is configured using the `alpn-modes` option, which is a list of supported ALPN protocol IDs (`"h2"` and `"http/1.1"`).
+
+NOTE: The HTTP 2 standard forbids the use of certain less secure TLS cipher suites for HTTP 2 connections. When the HTTP client supports HTTP 2 (which is the default), it will not support those cipher suites. Removing `"h2"` from `alpn-modes` will enable support for all cipher suites.
```
graemerocher: Replace HTTP 2 with `HTTP/2`

graemerocher (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
yawkat: sure but why, since it's going into 4.0 anyway?

http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java
yawkat: i don't want to expose the constructor. e.g. if alpn is false, alpnSupportedProtocols makes no sense, so the constructor is a bit confusing api.

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
yawkat: there are multiple channels per pool so that's not possible. but it's not performance critical anyway, the method is only used in some tests (the timeout specs)

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: moved to handlerRemoved. there's leak detection for pool handles btw, so if there are any scenarios i missed, they will eventually show up :)

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: i want to move this code to ConnectionManager, because DefaultHttpClient shouldn't care about the underlying protocol version and scheme

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: i went through the code and changed it to use MutableHttpRequest throughout. at the edges (where the user passes in the possibly-immutable request) i added a new wrapper.

yawkat:
@graemerocher can you look at this some time this week so that we can merge it before my vacation, @timyates wants to do some work moving the http modules out of micronaut-core, and that is blocked by this PR

graemerocher (approved):

yawkat:
thanks!

PR #8121, opened 2022-10-05
Rewrite of Bean Definition processing to use it for both Java and Groovy

dstepanov:
## Bean Processing

The new processing is shared between Java and Groovy and uses our AST Elements API. It is properties aware for languages like Groovy and Kotlin where there can be synthetic setter/getter methods. 
I have separated different implementations like factories and configurations to clarify the code.
It's a bit problematic to use AOP writer in the Inject project because AOP depends on Inject, so we cannot easily add a dependency for AOP. This needs to be addressed in the work of splitting the Inject module.

New processing will allow us to recursively call the processing on the factory beans, allowing more complicated factory bean scenarios.

## Modification of AST model

`PropertyElement` can return a read or write element, and it can be a method or a field. It also defines the access kind.
I have removed the direct use of `AnnotationMetadata` and replaced it with `ElementAnnotationMetadataFactory`. Now, it is only one place where annotation metadata is created. The implementation is shared between Java and Groovy. There are some new copy-constructor methods for the cases with the bean definition writer where you need to override only one value. 
Both Java and Groovy methods `getEnclosedElements` is rewritten to have the same `overridden` and hidden rules.

I have tried to thread `AnnotationMetadata` as immutable as possible, there are new methods for coping and merging the annotation hierarchy.

## Property beans

I have combined the code processing the bean's properties into one. We were resolving the properties in the configuration property beans, the configuration builder processing, and the bean introspection.
A new method, `ClassElement#getBeanProperties(BeanPropertiesConfiguration)` can find properties based on a custom configuration like access kind, visibility, prefixes etc. 
There is a new method, `ClassElement#getNativeBeanProperties()` that will return only native properties with synthetic methods.
The default method `ClassElement#getBeanProperties()` will return the properties resolved by the configuration provided by the newly introduced `@BeanProperties` annotation. This new annotation is mapped to by `@ConfigurationReader`,  `@ConfigurationBuilder`, and in the future PR by `@Introspected`.

## Groovy

There are a lot of fixes to Groovy processing and AST model, methods list isresolved in the same way as in Java, that did change some tests. It also changed how the configuration properties path is resolved for the inner classes. Now it's done in the same way as in Java, but that might be a breaking change for some.

## `@ConfigurationReader` and `@Endpoint`

In Micronaut 3 there was some undocumented and unexpected behavior, where `value` of the `@ConfigurationReader` would act as a base prefix and it would be combined with `prefix`. The was used mostly in `@Endpoint`. I have introduced a new attribute in `@ConfigurationReader` `basePrefix` and made `value` and `prefix` as one value as they should be. This might be a breaking change that needs to be documented.

## Validation

I have introduced a new `@RequiresValidation` marker annotation and a visitor visiting classes and adding this annotation to a class, method, or field if the element requires it. The visitor is in the Inject module but should be moved to the Validation module, which also extracts some of the `javax.validation` use from the core processing.
I have also moved some of the visitors that I think should be in core, one is about having `@Nullable` on primitives, and the other is about printing a warning when `@Internal` API is used.

## `@NextMajorVersion`
I have introduced a new annotation for marking a code that is supposed to be modified in the next major version. This is an alternative the leaving a comment: `In Micronaut 2 do that and this`.

I have also disabled the binary compatibility check.


aop/src/main/java/io/micronaut/aop/writer/AopHelperImpl.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * AOP helper to connect Inject module with AOP.
+ */
+@Internal
+@NextMajorVersion("Correct project dependency so this hack is not needed")
+public class AopHelperImpl implements AopHelper {
```
graemerocher: add `author` and `since` tags. Declare `final`

aop/src/main/java/io/micronaut/aop/writer/AopHelperImpl.java
```diff
@@ -0,0 +1,285 @@
...
+        });
+
+        aopProxyWriter.visitAroundMethod(interfaceToAdapt, targetMethod);
+
+        return aopProxyWriter;
+    }
+
+    @Override
+    public boolean visitIntrospectedMethod(BeanDefinitionVisitor visitor, ClassElement typeElement, MethodElement methodElement) {
+        AopProxyWriter aopProxyWriter = (AopProxyWriter) visitor;
```
graemerocher, line 176: since this is immediately casted maybe it makes sense to accept `AopProxyWriter`

aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
```diff
@@ -1304,16 +1271,13 @@ public void visitAnnotationMemberPropertyInjectionPoint(TypedElement annotationM
 
     @Override
     public void visitFieldValue(
-            TypedElement declaringType,
-            FieldElement fieldType,
-            boolean requiresReflection,
-            boolean isOptional) {
+        TypedElement declaringType,
+        FieldElement fieldType,
+        boolean isOptional, boolean requiresReflection) {
```
graemerocher: is this change needed? It only seems to swap the order of the arguments

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1595,4 +1626,21 @@ default boolean isEmpty() {
         return this == AnnotationMetadata.EMPTY_METADATA;
     }
 
+    /**
+     * Makes a copy of the annotation.
+     * @return the copy
+     * @since 4.0.0
+     */
+    AnnotationMetadata copyAnnotationMetadata();
```
graemerocher: Add `@NonNull` and maybe make this a default method that just returns `this`

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1595,4 +1626,21 @@ default boolean isEmpty() {
...
+
+    /**
+     * Unwraps possible delegate or provider.
+     * @return unwrapped
+     * @since 4.0.0
+     */
+    @Override
+    default AnnotationMetadata unwrapAnnotationMetadata() {
```
graemerocher: better name would be `getTargetAnnotationMetadata`

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1595,4 +1626,21 @@ default boolean isEmpty() {
...
+    /**
+     * Makes a copy of the annotation.
+     * @return the copy
+     * @since 4.0.0
+     */
+    AnnotationMetadata copyAnnotationMetadata();
+
+    /**
+     * Unwraps possible delegate or provider.
+     * @return unwrapped
```
graemerocher, line 1643: clarify in the javadoc that this returns either `this` or the result of the `AnnotationMetadataDelegate` or `AnnotationMetadataProvider` methods

core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
```diff
@@ -1595,4 +1626,21 @@ default boolean isEmpty() {
...
+     * @since 4.0.0
```
graemerocher, line 1646: add `@see` declarations for `AnnotationMetadataDelegate` and `AnnotationMetadataProvider` interfaces

core/src/main/java/io/micronaut/core/annotation/Introspected.java
```diff
@@ -51,6 +51,9 @@
 @Inherited
 public @interface Introspected {
 
+     Introspected.AccessKind[] DEFAULT_ACCESS_KIND = {Introspected.AccessKind.METHOD};
+     Introspected.Visibility[] DEFAULT_VISIBILITY = {Introspected.Visibility.DEFAULT};
```
graemerocher: add javadoc to describe what these mean

core/src/main/java/io/micronaut/core/util/ArrayUtils.java
```diff
@@ -279,6 +285,25 @@ public static Object toPrimitiveArray(final Object[] wrapperArray) {
...
     }
 
+    /**
+     * Reverse the order of items in array.
+     *
+     * @param input The array
+     * @param <T>   The array type
+     * @since 4.0.0
+     */
+    public static <T> void reverse(T[] input) {
```
graemerocher, line 295: document that the passed array is mutated

gradle.properties
```diff
@@ -55,5 +55,5 @@ graalVersion=22.0.0.2
 micronautSecurityVersion=3.8.0
 
 org.gradle.caching=true
-org.gradle.parallel=true
 org.gradle.jvmargs=-Xmx1g
+org.gradle.parallel=true
```
graemerocher: unnecessary change

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/AbstractGroovyElement.java
```diff
@@ -63,26 +68,72 @@ public abstract class AbstractGroovyElement implements AnnotationMetadataDelegat
...
+                elementAnnotationMetadata = elementAnnotationMetadataFactory.build(this, presetAnnotationMetadata);
+            }
+        }
+        return elementAnnotationMetadata;
+    }
+
+    /**
+     * @return the copy
+     */
+    protected abstract AbstractGroovyElement copyThis();
```
graemerocher: add `@NonNull`

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/AbstractGroovyElement.java
```diff
@@ -63,26 +68,72 @@ public abstract class AbstractGroovyElement implements AnnotationMetadataDelegat
...
+
+    /**
+     * @param element the values to be copied to
+     */
+    protected void copyValues(AbstractGroovyElement element) {
```
graemerocher: add `@NonNull` to parameter

graemerocher: Rename to something that better describes what this method does. It seems to copy the annotation metadata so maybe `copyAnnotationMetadata`

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/AbstractGroovyElement.java
```diff
@@ -63,26 +68,72 @@ public abstract class AbstractGroovyElement implements AnnotationMetadataDelegat
...
+        element.presetAnnotationMetadata = presetAnnotationMetadata;
+    }
+
+    protected final AbstractGroovyElement makeCopy() {
```
graemerocher: add `javadoc` and `@NonNull` on return type

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/AbstractGroovyElement.java
```diff
@@ -347,19 +358,15 @@ public Optional<String> getDocumentation() {
...
-            annotationMetadata = AstAnnotationUtils.getAnnotationMetadata(sourceUnit, compilationUnit, type);
-        } else {
-            annotationMetadata = AnnotationMetadata.EMPTY_METADATA;
+            return elementAnnotationMetadataFactory;
         }
-        return annotationMetadata;
+        return emptyAnnotationsForNativeType(nativeType);
+    }
+
+    protected final ElementAnnotationMetadataFactory emptyAnnotationsForNativeType(Object nativeType) {
```
graemerocher: add javadoc and `@NonNull` to return type and argument

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java
```diff
@@ -128,20 +177,68 @@ public GroovyClassElement(GroovyVisitorContext visitorContext, ClassNode classNo
...
     public boolean isTypeVariable() {
         return isTypeVar;
     }
 
     @Override
     public <T extends Element> List<T> getEnclosedElements(@NonNull ElementQuery<T> query) {
+        return getEnclosedElements(query, false);
+    }
+
+    public final <T extends Element> List<T> getEnclosedElements(@NonNull ElementQuery<T> query, boolean isSource) {
```
graemerocher: does this need to be public? If so javadoc needed

inject-groovy/src/test/groovy/io/micronaut/inject/configuration/GroovyConfigurationMetadataBuilderSpec.groovy
```diff
@@ -1,94 +0,0 @@
...
-
-import groovy.json.JsonSlurper
-import io.micronaut.ast.transform.test.AbstractBeanDefinitionSpec
-import io.micronaut.ast.groovy.config.GroovyConfigurationMetadataBuilder
-import org.codehaus.groovy.ast.ClassNode
-/**
- * @author graemerocher
- * @since 1.0
- */
-class GroovyConfigurationMetadataBuilderSpec extends AbstractBeanDefinitionSpec {
```
graemerocher, line 26: why was this deleted?

inject/src/main/java/io/micronaut/context/annotation/BeanProperties.java
```diff
@@ -0,0 +1,134 @@
...
+@Inherited
+public @interface BeanProperties {
+
+    String ACCESS_KIND = "accessKind";
+    String VISIBILITY = "visibility";
+    String INCLUDES = "includes";
+    String EXCLUDES = "excludes";
+    String ALLOW_WRITE_WITH_ZERO_ARGS = "allowWriteWithZeroArgs";
+    String ALLOW_WRITE_WITH_MULTIPLE_ARGS = "allowWriteWithMultipleArgs";
+    String EXCLUDED_ANNOTATIONS = "excludedAnnotations";
```
graemerocher: prefix these with `MEMBER_` and add javadoc for each constant

inject/src/main/java/io/micronaut/inject/ProcessingException.java
```diff
@@ -0,0 +1,48 @@
...
+
+import io.micronaut.inject.ast.Element;
+
+/**
+ * The exception can be used to stop the processing and display an error associated to the element.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public final class ProcessingException extends RuntimeException {
```
graemerocher: Move to `io.micronaut.inject.ast` or `io.micronaut.inject.processing`

inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
```diff
@@ -2440,18 +2314,47 @@ <T1 extends Annotation> AnnotationMetadata removeAnnotationIf(
         return annotationMetadata;
     }
 
+    /**
+     * The caching entry.
+     *
+     * @author Denis Stepanov
+     * @since 4.0.0
+     */
+    public interface CacheEntry extends AnnotationMetadataDelegate {
```
graemerocher: Rename to `CachedAnnotationMetadata`

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataHierarchy.java
```diff
@@ -955,4 +1028,48 @@ public Optional<String> findRepeatableAnnotation(String annotation) {
...
+
+    /**
+     * Merges the hierarchy into one {@link MutableAnnotationMetadata}.
+     *
+     * @return merged metadata
+     * @since 4.0.0
+     */
+    public MutableAnnotationMetadata merge() {
+        MutableAnnotationMetadata newAnnotationMetadata = new MutableAnnotationMetadata();
+        for (AnnotationMetadata annotationMetadata : hierarchy) {
```
graemerocher, line 1040: I think you would need to loop in reverse order because the first entry in the hierarchy is the highest priority. Can you add a test for this?

inject/src/main/java/io/micronaut/inject/annotation/RequiresValidation.java
```diff
@@ -0,0 +1,38 @@
...
+ * Internal method marks a type, method or a field for validation.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Documented
+@Retention(SOURCE)
+@Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD})
+@Internal
+public @interface RequiresValidation {
```
graemerocher: Move this annotation to another package so if you `*` import `inject.annotation` which includes a lot of public API it doesn't appear

inject/src/main/java/io/micronaut/inject/ast/BeanPropertiesConfiguration.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The bean properties configuration.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public final class BeanPropertiesConfiguration {
```
graemerocher: this would be a type that would be nice to model as Java 17 record

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -172,8 +209,117 @@ default boolean isProxy() {
...
+        return staticCreators.stream().filter(Element::isPublic).findFirst();
+    }
+
+    /**
+     * Find accessible constructors.
+     *
+     * @return accessible constructors
+     * @since 4.0.0
+     */
+    default List<ConstructorElement> findConstructors() {
```
graemerocher: Rename to `getAccessibleConstructors`

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -172,8 +209,117 @@ default boolean isProxy() {
...
+        }
+        return constructors.stream()
+            .filter(Element::isPublic)
+            .<MethodElement>map(c -> c)
+            .findFirst();
+
+    }
+
+    /**
+     * Find static creator.
```
graemerocher: Clarify in the javadoc what a static creator is, as it is unclear for the user of this API

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -172,8 +209,117 @@ default boolean isProxy() {
...
+        if (withArgs.size() == 1) {
+            return Optional.of(withArgs.get(0));
+        } else {
+            staticCreators = withArgs;
+        }
+        return staticCreators.stream().filter(Element::isPublic).findFirst();
+    }
+
+    /**
+     * Find default static creator.
```
graemerocher: clarify in the javadoc what a default static creator is as it is unclear and the different in behaviour from `findStaticCreator`

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -172,8 +209,117 @@ default boolean isProxy() {
...
+     */
+    default List<ConstructorElement> findConstructors() {
+        return getEnclosedElements(ElementQuery.CONSTRUCTORS)
+            .stream()
+            .filter(ctor -> !ctor.isPrivate())
+            .collect(Collectors.toList());
+    }
+
+    /**
+     * Find static creators.
```
graemerocher: clarify in the javadoc what a `static` creator is and add references to the other related methods

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -172,8 +209,117 @@ default boolean isProxy() {
...
+     *
+     * @return static creators
+     * @since 4.0.0
+     */
+    default List<MethodElement> findStaticCreators() {
```
graemerocher: add `@NonNull`

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -236,6 +382,27 @@ default List<PropertyElement> getBeanProperties() {
         return Collections.emptyList();
     }
 
+    /**
+     * Returns the native bean properties (synthetic getters and setters) for this class element..
+     *
+     * @return The bean properties for this class element
+     * @since 4.0.0
+     */
+    default List<PropertyElement> getNativeBeanProperties() {
```
graemerocher: Rename to `getSyntheticBeanProperties` since native might be confused with `getNativeType` which returns the underlying type from the underlying AST (Java or Groovy)

Also add `@NonNull`

inject/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -236,6 +382,27 @@ default List<PropertyElement> getBeanProperties() {
...
+    }
+
+    /**
+     * Returns the bean properties (getters and setters) for this class element based on custom configuration.
+     *
+     * @param configuration The configuration
+     * @return The bean properties for this class element
+     * @since 4.0.0
+     */
+    default List<PropertyElement> getBeanProperties(BeanPropertiesConfiguration configuration) {
```
graemerocher: Add `@NonNull` to the return type and the argument. 

For consistency rename `BeanPropertiesQuery` as we already have `ElementQuery` so having this named with `Configuration` confuses matters

inject/src/main/java/io/micronaut/inject/ast/Element.java
```diff
@@ -59,6 +54,14 @@ default boolean isPackagePrivate() {
         return false;
     }
 
+    /**
+     * @return True if the element is synthetic.
+     * @since 4.0.0
+     */
+    default boolean isSynthetic() {
```
graemerocher, line 62: Clarify what a synthetic element is in the javadoc

inject/src/main/java/io/micronaut/inject/ast/ElementAnnotationMetadata.java
```diff
@@ -13,7 +13,23 @@
...
  */
-package io.micronaut.validation.executable;
+public interface ElementAnnotationMetadata extends MutableAnnotatedElement<AnnotationMetadata> {
+
+    /**
+     * Gets the annotation metadata for the element.
+     *
+     * @return the annotation metadata
+     */
+    AnnotationMetadata get();
```
graemerocher: Rename to `getAnnotationMetadata` and extend from `AnnotationMetadataProvider` would be better. Or could we not create an interface like `MutableAnnotationMetadataProvider` which would be more consistent with the rest of the API?

inject/src/main/java/io/micronaut/inject/ast/ElementAnnotationMetadataFactory.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @param element            The element
+     * @param annotationMetadata The preloaded annotation
+     * @return the element's metadata
+     */
+    @NonNull
+    ElementAnnotationMetadata build(@NonNull Element element, @NonNull AnnotationMetadata annotationMetadata);
+
+    default ElementAnnotationMetadataFactory readOnly() {
+        throw new IllegalStateException("Unsupported operation!");
```
graemerocher: given this is a new interface does this need to be default?

inject/src/main/java/io/micronaut/inject/ast/ElementAnnotationMetadataFactory.java
```diff
@@ -0,0 +1,81 @@
...
+     * This method will avoid fetching default annotation metadata from cache.
+     *
+     * @param element            The element
+     * @param annotationMetadata The preloaded annotation
+     * @return the element's metadata
+     */
+    @NonNull
+    ElementAnnotationMetadata build(@NonNull Element element, @NonNull AnnotationMetadata annotationMetadata);
+
+    default ElementAnnotationMetadataFactory readOnly() {
```
graemerocher: add javadoc to describe what this does and `@NonNull` to the return type

inject/src/main/java/io/micronaut/inject/ast/ElementAnnotationMetadataFactory.java
```diff
@@ -0,0 +1,81 @@
...
+        throw new IllegalStateException("Unsupported operation!");
+    }
+
+    default ElementAnnotationMetadataFactory overrideForNativeType(Object nativeType,
```
graemerocher, line 68: add javadoc to describe what this method does

inject/src/main/java/io/micronaut/inject/ast/ElementFactory.java
```diff
@@ -31,135 +30,115 @@
  * @since 2.3.0
  */
 public interface ElementFactory<E, C extends E, M extends E, F extends E> {
```
graemerocher, line 32: A bit concerned about the breaking changes to this interface. Can we retain the original methods and add new methods? The old methods and create a factory that just delegates to the passed metadata.

inject/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
```diff
@@ -0,0 +1,384 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * The AST bean properties utils.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public final class AstBeanPropertiesUtils {
```
graemerocher, line 50: add `@Internal`

inject/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
```diff
@@ -0,0 +1,384 @@
...
+            return true;
+        }
+        return !excludes.isEmpty() && excludes.contains(propertyName);
+    }
+
+    /**
+     * Internal holder class for getters and setters.
+     */
+    @SuppressWarnings("VisibilityModifier")
+    public static class BeanPropertyData {
```
graemerocher: declare `final`

graemerocher: does this need to be public?

inject/src/main/java/io/micronaut/inject/configuration/ConfigurationUtils.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An util class to calculate configuration paths.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public final class ConfigurationUtils {
```
graemerocher, line 35: add `@Internal`

inject/src/main/java/io/micronaut/inject/mappers/ConfigurationBuilderToBeanPropertiesMapper.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Map values of {@link ConfigurationBuilder} to {@link BeanProperties}.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public final class ConfigurationBuilderToBeanPropertiesMapper implements TypedAnnotationMapper<ConfigurationBuilder> {
```
graemerocher, line 35: add `@Internal`

inject/src/main/java/io/micronaut/inject/mappers/ConfigurationPropertiesToBeanPropertiesMapper.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Map values of {@link ConfigurationReader} to {@link BeanProperties}.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public final class ConfigurationPropertiesToBeanPropertiesMapper implements TypedAnnotationMapper<ConfigurationReader> {
```
graemerocher, line 35: add `@Internal`

inject/src/main/java/io/micronaut/inject/processing/AbstractBeanDefinitionBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    @Override
+    public final Collection<BeanDefinitionVisitor> build() {
+        buildInternal();
+        return beanDefinitionWriters;
+    }
+
+    public abstract void buildInternal();
```
graemerocher: add javadoc

inject/src/main/java/io/micronaut/inject/processing/AbstractBeanDefinitionBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * Abstract shared functionality of the builder.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+abstract class AbstractBeanDefinitionBuilder implements BeanDefinitionBuilder {
```
graemerocher: add `@Internal`

inject/src/main/java/io/micronaut/inject/processing/AbstractBeanDefinitionBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+    protected final AopHelper aopHelper;
+
+    protected AbstractBeanDefinitionBuilder(ClassElement classElement, VisitorContext visitorContext) {
+        this.classElement = classElement;
+        this.visitorContext = visitorContext;
+        checkPackage(classElement);
+        try {
+            aopHelper = (AopHelper) ClassUtils.forName("io.micronaut.aop.writer.AopHelperImpl", getClass().getClassLoader()).get().newInstance();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
graemerocher: add more descriptive error message

inject/src/main/java/io/micronaut/inject/processing/AopHelper.java
```diff
@@ -0,0 +1,50 @@
...
+import io.micronaut.inject.visitor.VisitorContext;
+import io.micronaut.inject.writer.BeanDefinitionVisitor;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * AOP helper to connect Inject module with AOP.
+ */
+@NextMajorVersion("Correct dependency graph")
+public interface AopHelper {
```
graemerocher, line 34: add `@Internal` and javadoc methods

inject/src/main/java/io/micronaut/inject/processing/ConfigurationReaderBeanDefinitionBuilder.java
```diff
@@ -0,0 +1,312 @@
...
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Configuration reader bean builder.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+final class ConfigurationReaderBeanDefinitionBuilder extends DeclaredBeanDefinitionBuilder {
```
graemerocher: add `@Internal`

inject/src/main/java/io/micronaut/inject/processing/FactoryBeanDefinitionBuilder.java
```diff
@@ -0,0 +1,300 @@
...
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Factory bean builder.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+final class FactoryBeanDefinitionBuilder extends DeclaredBeanDefinitionBuilder {
```
graemerocher: add `@Internal`

inject/src/main/java/io/micronaut/inject/processing/IntroductionInterfaceBeanDefinitionBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Introduction interface proxy builder.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+final class IntroductionInterfaceBeanDefinitionBuilder extends AbstractBeanDefinitionBuilder {
```
graemerocher: add `@Internal`

inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
```diff
@@ -50,10 +52,18 @@ public interface VisitorContext extends MutableConvertibleValues<Object>, ClassW
...
      */
-    @NonNull ElementFactory<?, ?, ?, ?> getElementFactory();
+    @NonNull
+    ElementFactory<?, ?, ?, ?> getElementFactory();
+
+    @NonNull
+    ElementAnnotationMetadataFactory getElementAnnotationMetadataFactory();
+
+    @NonNull
+    AbstractAnnotationMetadataBuilder<?, ?> getAnnotationMetadataBuilder();
```
graemerocher, line 80: can we not make this public API?

inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
```diff
@@ -50,10 +52,18 @@ public interface VisitorContext extends MutableConvertibleValues<Object>, ClassW
...
+     *
      * @return The element factory
      * @since 2.3.0
      */
-    @NonNull ElementFactory<?, ?, ?, ?> getElementFactory();
+    @NonNull
+    ElementFactory<?, ?, ?, ?> getElementFactory();
+
+    @NonNull
+    ElementAnnotationMetadataFactory getElementAnnotationMetadataFactory();
```
graemerocher, line 70: needs javadoc

inject/src/main/java/io/micronaut/inject/writer/AbstractBeanDefinitionBuilder.java
```diff
@@ -118,17 +119,21 @@ public abstract class AbstractBeanDefinitionBuilder implements BeanElementBuilde
...
-     * @param metadataBuilder the metadata builder
-     * @param visitorContext the visitor context
+     *
+     * @param originatingElement               The originating element
+     * @param beanType                         The bean type
+     * @param metadataBuilder                  the metadata builder
+     * @param visitorContext                   the visitor context
+     * @param elementAnnotationMetadataFactory The element annotation metadata factory
      */
     protected AbstractBeanDefinitionBuilder(
```
graemerocher, line 129: We should rename this type to `AbstractBeanElementBuilder` otherwise we have two classes with the same name

inject/src/main/java/io/micronaut/inject/processing/BeanDefinitionBuilder.java
```diff
@@ -15,25 +15,21 @@
...
  *
- * @author graemerocher
- * @since 1.0
- * @deprecated Use {@link io.micronaut.core.annotation.AnnotationUtil} instead
+ * @author Denis Stepanov
+ * @since 4.0.0
  */
-@Deprecated
-public interface ProcessedTypes {
+public interface BeanDefinitionBuilder {
```
graemerocher: Rename to `BeanDefinitionCreator`

inject/src/main/java/io/micronaut/inject/processing/BeanDefinitionBuilderFactory.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * Bean definition builder factory.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public abstract class BeanDefinitionBuilderFactory {
```
graemerocher: Don't think we need another class here, just move these static methods to `BeanDefinitionBuilder` (or better named `BeanDefinitionCreator`

graemerocher (changes requested):
Good progress. Did an initial review so you can start addressing feedback.

aop/src/main/java/io/micronaut/aop/writer/AopHelperImpl.java
dstepanov: Done but this class should be removed in the `inject` module splitting.

aop/src/main/java/io/micronaut/aop/writer/AopHelperImpl.java
dstepanov: The purpose of this class is that `AopProxyWriter` is not on the classpath of the inject module.

aop/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
dstepanov: Probably unintended

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/AbstractGroovyElement.java
dstepanov: I clarified the use of this method in the Javadoc

inject/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
dstepanov: Yes, it's used to build the property instance.

inject/src/main/java/io/micronaut/inject/processing/AopHelper.java
dstepanov: This class should be removed soon, I don't think it's necessary to document the self-explanatory methods.

inject/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
dstepanov: I need it for one use-case that I can rewrite in the next PR. I have added `@Internal` for now.

inject/src/main/java/io/micronaut/inject/processing/BeanDefinitionBuilderFactory.java
dstepanov: We cannot have non-public methods in the interface, and I don't want to expose those helper methods.

inject/src/main/java/io/micronaut/inject/ast/ElementAnnotationMetadata.java
dstepanov: Not sure I understand

inject/src/main/java/io/micronaut/inject/ast/ElementFactory.java
dstepanov: I had them before as deprecated, but we want to remove everything deprecated in the v4. So I deleted them.

inject/src/main/java/io/micronaut/inject/ast/BeanPropertiesConfiguration.java
dstepanov: Actually, this is intended to be mutable

inject-groovy/src/test/groovy/io/micronaut/inject/configuration/GroovyConfigurationMetadataBuilderSpec.groovy
dstepanov: I have added a new test. It looks like the metadata for the Groovy properties wouldn't be generated because the writer is only present for Java annotation processing.

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataHierarchy.java
dstepanov: It works this way because the values aren't overridden when added. Test added

inject/src/main/java/io/micronaut/inject/ast/PackageElement.java
```diff
@@ -38,7 +39,16 @@ public interface PackageElement extends Element {
...
         Objects.requireNonNull(name, "Name cannot be null");
-
         return new SimplePackageElement(name);
     }
+
+    /**
+     * Is unnamed package?
+     *
+     * @return true if unnamed
+     * @since 3.7.0
```
graemerocher: should be 4.0.0?

graemerocher:
Could you fix the Sonar bugs as well?

dstepanov:
I would leave it as it is, one bug is with checking `Optional#get` should disappear after the inject module is split and the other one is actually to support for equals to work for elements when they are anonymous subclasses.

graemerocher (approved):

PR #8123, opened 2022-10-05
Support http status 103 and 425

### LINKED ISSUE #8120: HttpStatus 425 is UNORDERED_COLLECTION instead of TOO_EARLY
### Issue description

Hi everyone, HttpStatus.value(425) is translated to HttpStatus.UNORDERED_COLLECTION, but shouldn't this be TOO_EARLY as described here: https://www.rfc-editor.org/rfc/rfc8470

SplotyCode:
This PR relates to #8120. There are also quite a few http status codes that got renamed. Even a few common ones like 413. These renames were mostly for consistency reasons and dont give much more information so i believe we should stay with the old ones for backward compatibility right?

This pull request fixes #8120
Signed-off-by: david.scandurra <david.scandurra@check24.de>

### DIFF
--- a/http/src/main/java/io/micronaut/http/HttpStatus.java
+++ b/http/src/main/java/io/micronaut/http/HttpStatus.java
@@ -27,6 +27,7 @@ public enum HttpStatus implements CharSequence {
     CONTINUE(100, "Continue"),
     SWITCHING_PROTOCOLS(101, "Switching Protocols"),
     PROCESSING(102, "Processing"),
+    EARLY_HINTS(103, "Early Hints"),
     OK(200, "Ok"),
     CREATED(201, "Created"),
     ACCEPTED(202, "Accepted"),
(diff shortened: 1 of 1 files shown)


SplotyCode:
The Static Analysis Ci fail does not seemed to be related to the changes

graemerocher:
Thanks could you deprecate instead of remove the previous enum value?

yawkat:
this will conflict with #5621 if it goes into 4.0. could merge it into 3.8 though

yawkat (approved):

yawkat:
since #5621 was rejected, we can merge this.

timyates:
@yawkat We still need to [address this I believe](https://github.com/micronaut-projects/micronaut-core/pull/8123#issuecomment-1269047858):

> Thanks could you deprecate instead of remove the previous enum value?

yawkat:
oh right. but i think it's fine for 4.0, probably nobody uses it anyway. should deprecate it for 3.x though

timyates:
> oh right. but i think it's fine for 4.0, probably nobody uses it anyway. should deprecate it for 3.x though

Agreed 👍

timyates (approved):

PR #8116, opened 2022-10-04
Add a test module to verify simple native execution

timyates:
We detect issues with native compilation quite far down the release chain.

This causes problems, and delays when releasing a new version of Micronaut, especially minor or major releases.

This commit adds a new module test-suite-graal which runs nativeTest against a simple micronaut application.

This should pass once https://github.com/micronaut-projects/micronaut-core/pull/8125 and https://github.com/micronaut-projects/micronaut-core/pull/8126 are merged into 4.0.x and back into here

PR #8126, opened 2022-10-06
Caffeine usage in :session fails with Graal

timyates:
When we try to build a native image (since the unshading of caffeine), we get failures as caffeine uses a lot of reflection.

This PR adds the minimum graal configuration for nativeTest to not simply fail in https://github.com/micronaut-projects/micronaut-core/pull/8116

### DIFF
--- a/core/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core/reflect-config.json
+++ b/core/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core/reflect-config.json
@@ -1,7 +1,0 @@
-[ {
-  "name" : "io.micronaut.caffeine.cache.PSW",
-  "allDeclaredConstructors" : true
-}, {
-  "name" : "io.micronaut.caffeine.cache.SSLA",
-  "allDeclaredConstructors" : true
-} ]
--- a/session/build.gradle
+++ b/session/build.gradle
@@ -9,6 +9,7 @@ dependencies {
 
     compileOnly project(":http-server")
     compileOnly project(":http-server-netty")
+    compileOnly libs.managed.graal
 
     implementation libs.managed.reactor
     implementation libs.caffeine
--- a/session/src/main/java/io/micronaut/session/CaffeineSessionFeature.java
+++ b/session/src/main/java/io/micronaut/session/CaffeineSessionFeature.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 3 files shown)


graemerocher:
I wonder if it makes sense to add a dependency on `micronaut-cache-caffeine` and put the config here https://github.com/micronaut-projects/micronaut-cache/tree/master/cache-caffeine

timyates:
@graemerocher I suspect getting all of the possibilities for the entirety of Caffeine is going to be problematic

Caffeine contains a lot of auto-generated cache handlers/methodologies, which are [loaded via reflection for speed](https://github.com/quarkusio/quarkus/issues/10420#issuecomment-653407502).  An index was added for that linked issue, but it is going to be removed in v3 of Caffeine, as I don't believe Quarkus use it any more.

Not sure how we can catch them all, or what we would need to do to find all the required fields that are accessed 🤔

ben-manes:
Unfortunately I think it would be worse if we didn't use reflection because then the AOT compiler would not dead code eliminate the unnecessary classes. Since the implementations are chosen by evaluating a builder's configuration, that is unlikely to be computable as multiple conditions have to be evaluated. When Quarkus asked us to provide a jandex index to include all at runtime this resulted in a 14mb penalty.

The generated classes are an attempt to minimize the memory overhead of per-entry fields that might not be needed, like expiration timestamps. In pre-AOT Java an unloaded class is free for the runtime and is a tiny amount of jar bloat. In AOT world it still makes sense to minimize the memory costs, but it is unclear how to hint which configurations are needed. Ideally users wouldn't know about our private implementation classes in their configuration and the runtimes would handle things gracefully.

I still do not understand their [suggested approach](https://github.com/quarkusio/quarkus/issues/10420#issuecomment-654483795) for how to allow an AOT compiler to discover the needed classes as a "warm up" phase. I'd be open to improvements to make this better.

timyates:
@ben-manes I saw all your efforts to get a general solution for them.

And thanks for jumping in here to explain the issues 😎 

What we have here seems to be enough for our session caching, and we will discuss what we should do for our caching module which likely has more cache usage -- not yet tested if that's the case though. This may simply be documentation, pulling the common caches as in Quarkus, or a combination of the two

On a personal note: Caffeine is awesome and really appreciated! 🏅

alvarosanchez:
I agree with @graemerocher that this belongs to `micronaut-cache-caffeine`. Even if the configuration isn't complete, it's better than noting.

Regarding the actual change, `@AutomaticFeature` is deprecated, see: https://github.com/oracle/graal/blob/master/sdk/src/com.oracle.svm.core.annotate/src/com/oracle/svm/core/annotate/AutomaticFeature.java

Have you tried with running with the agent and then taking the generated config for only the Caffeine classes?

timyates:
@alvarosanchez So we should switch back to using json/text file based Graal config?

Currently, `session` only has a dependency on the caffeine library itself https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/session/build.gradle#L14

Are you suggesting to add the Graal config to `micronaut-cache-caffeine` and then pull that module in to core?  Won't that give us a tight binding between the two modules?  (as micronaut-cache depends on core).  Can do if agreed, I just worry we're going to get ourselves in a knot?

timyates:
For Micronaut 4.0.0 we are moving `:session` to `micronaut-session` where we will use the same functionality for handling this as will be used in micronaut-cache-caffeine

Closing this

PR #8130, opened 2022-10-07
Add ctor overload to HttpStatusException to allow a cause to be provided

carl-mastrangelo:
The main use case it to be able to catch another kind of exception and provide a specific status code.  For example, without this change, this is how causes are set:

```java
try {
  customer = customerTable.lookupById(request.getId());
} catch (IllegalArgumentException e) {
  var failure = new HttpStatusException(HttpStatus.BAD_REQUEST, "bad customer id");
  failure.initCause(e);
  throw failure;
}
```

### DIFF
--- a/http/src/main/java/io/micronaut/http/exceptions/HttpException.java
+++ b/http/src/main/java/io/micronaut/http/exceptions/HttpException.java
@@ -37,6 +37,13 @@ public abstract class HttpException extends RuntimeException {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


carl-mastrangelo:
I'm not sure I can sign the CLA.  (I am at a small startup where we don't have the legal expertise to say if I can sign or not).    Would it be possible for an exception to be made?   Or a reimplementation?

PR #8138, opened 2022-10-09
Added lost slf4j and logback artifacts. Added log4j bom dependency to final micronaut bom.

altro3:
Added lost slf4j and logback artifacts.
Added log4j bom dependency to final micronaut bom.

At the moment, it is not possible to fully use the micronaut bom file for the reason that sometimes additional slf4j artifacts are required (for example, the slf4j-nop artifact is needed in the micronaut-openapi library, and sometimes the jcl-over-slf4j or jul-to- artifacts are required in the project slf4j), and sometimes additional artifacts for log4j are required (for example, in micronaut-elasticsearch, it is more convenient for me to use logback and slf4j, this requires adding the log4j-to-slf4j artifact). And now, in order for everything to work, I have to duplicate the versions of these libraries in my gradle file, which is not very convenient.

In my opinion, you need to use the logic of the spring-boot team - they add all the artifacts of the family to the final bom file, even if they do not use all of them.

Actually, I don't understand why all the slf4j, logback and log4j artifacts were not added, but the same jackson-bom was added.

I think this change will make life easier not only for me :-)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -22,7 +22,6 @@ jsr107 = "1.1.1"
 javax-el = "3.0.1-b12"
 javax-el-impl = "2.2.1-b05"
 logbook-netty = "2.14.0"
-log4j = "2.19.0"
 selenium = "3.141.59"
 smallrye = "5.5.0"
 systemlambda = "1.2.1"
@@ -134,6 +133,7 @@ managed-reactor = "3.4.23"
 managed-rxjava1 = "1.3.8"
 managed-rxjava1-interop = "0.13.7"
 managed-slf4j = "1.7.36"
+managed-log4j = "2.19.0"
 managed-spock = "2.0-groovy-3.0"
 managed-spotbugs = "4.7.1"
 managed-spring = "5.3.23"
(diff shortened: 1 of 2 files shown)


dstepanov:
Looks like BOM check failed for some reason

melix:
BOM verification fails because it now references non existent modules:

```
BOM io.micronaut:micronaut-bom:3.7.2-SNAPSHOT (via io.micronaut:micronaut-bom:3.7.2-SNAPSHOT) declares a non-resolvable dependency: ch.qos.logback:logback-classic-blackbox:1.2.11
BOM io.micronaut:micronaut-bom:3.7.2-SNAPSHOT (via io.micronaut:micronaut-bom:3.7.2-SNAPSHOT) declares a non-resolvable dependency: ch.qos.logback:logback-core-blackbox:1.2.11
BOM io.micronaut:micronaut-bom:3.7.2-SNAPSHOT (via io.micronaut:micronaut-bom:3.7.2-SNAPSHOT) declares a non-resolvable dependency: org.slf4j:osgi-over-slf4:1.7.36
```

So this has to be fixed. That said, while I understand the request, I am not convinced that this is how we should fix this. We should avoid the Micronaut BOM to include many dependencies, and prefer including other BOMs when possible. Logback doesn't provide a BOM, so we would have the responsibility to figure out all possible modules and their versions. As the error above shows, this is no easy task.

dstepanov:
Personally I hit the same problem where some of the slf4j libs aren’t available in our BOM.

altro3:
@dstepanov @melix 
Sorry, I don't quite understand how it can be fixed :-(.

I would like to use micronaut as a single development platform and I do not suggest adding all possible dependencies, but only taking a single approach to generating a bom file. Those. if the library has several modules, then add all of them to the final bom file, if one of the modules is present in the final bom file.

I gave examples when I needed to use one of the slf4j modules in micronaut-openapi. The same problems exist not only with loggers, but, for example, with grpc - at the moment we have to duplicate the grpc versions in the micronaut-grpc and micronaut-tracing modules, because not all grpc modules are in the final bom file.

And the absence of a bom file for logback and slf4js is rather an exception. That's how it happened historically.

dstepanov:
The dependencies you added don't exist see `osgi-over-slf4` missing `j`.

altro3:
@dstepanov fixed, the same result

dstepanov:
Did you run `:bom:checkBom` locally?

dstepanov:
It doesn't look like those dependencies exist https://repo1.maven.org/maven2/ch/qos/logback/

altro3:
@dstepanov hm... Yes, this is my fault: some of these artifacts (blackbox) appeared in version 1.3-1.4 ...

altro3:
@dstepanov Turns out it was my mistake... Well, thanks for the help! Now everything works correctly

dstepanov (approved):

melix (approved):

graemerocher (approved):

PR #8148, opened 2022-10-12
Allow for Bean Introspection to use different accessor for read and write. 

dstepanov:
Refactoring the introspection visitor to use the same bean properties logic as the rest of Micronaut. Added support for it to use different types of accessors. Now getter and setter don't need to share the same accessor style.

`@BeanProperties` has to be with retention CLASS instead of SOURCE, because Javac would not return source annotations, added the same exclude for CLASS retention as for SOURCE, even so, in the `DefaultAnnotationMetadata` the field is still called source.

PR is targeting my other branch.


inject-groovy/src/test/groovy/io/micronaut/inject/annotation/SourceRetentionSpec.groovy
```diff
@@ -1,28 +0,0 @@
-package io.micronaut.inject.annotation
-
-import io.micronaut.ast.transform.test.AbstractBeanDefinitionSpec
-import io.micronaut.inject.BeanDefinition
-
-import java.lang.annotation.Native
-
-class SourceRetentionSpec extends AbstractBeanDefinitionSpec {
-    void "test source retention annotations are not retained"() {
```
graemerocher, line 9: why was this deleted?

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
```diff
@@ -487,7 +487,14 @@ private static void instantiateInternal(
         // 4th argument: all annotations
         pushCreateAnnotationData(owningType, declaringClassWriter, generatorAdapter, annotationMetadata.allAnnotations, defaultsStorage, loadTypeMethods, annotationMetadata.getSourceRetentionAnnotations());
         // 5th argument: annotations by stereotype
-        pushStringMapOf(generatorAdapter, annotationMetadata.annotationsByStereotype, false, Collections.emptyList(), list -> pushListOfString(generatorAdapter, list));
+        Map<String, List<String>> annotationsByStereotype = annotationMetadata.annotationsByStereotype;
```
graemerocher, line 490: should this create a copy before writing? Otherwise you are mutating internal state?

dstepanov: I make a copy a few lines below

inject-groovy/src/test/groovy/io/micronaut/inject/annotation/SourceRetentionSpec.groovy
dstepanov: It’s renamed to RetentionSpec git diff a bit glitchy

graemerocher (approved):

PR #8153, opened 2022-10-13
Initial tentative support for Loom

melix:


dstepanov:
@graemerocher Can we just use `BeanElementBuilder` and add a factory with `Requires`?

PR #8154, opened 2022-10-13
Use v6.0.1 of the internal build tools

timyates:
The plugin pulls in io.micronaut:micronaut-inject-groovy which can't be resolved versionless.

So I added a resolution to switch this with the :inject-groovy module, which I think is more sensible here

### DIFF
--- a/inject-java-test/build.gradle
+++ b/inject-java-test/build.gradle
@@ -29,6 +29,14 @@ dependencies {
     testImplementation libs.smallrye
 }
 
+tasks.withType(JavaCompile).configureEach {
+    options.compilerArgs += [
+            '--add-exports', 'jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED'
+    ]
+    // https://github.com/gradle/gradle/issues/18824
+    sourceCompatibility = 17
+}
+
 tasks.named("sourcesJar") {
     from "$projectDir/src/main/groovy"
 }
--- a/inject-java-test/build.gradle
+++ b/inject-java-test/build.gradle
@@ -29,14 +29,6 @@ dependencies {
     testImplementation libs.smallrye
 }
 
-tasks.withType(JavaCompile).configureEach {
-    options.compilerArgs += [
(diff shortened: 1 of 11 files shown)


inject-java-test/build.gradle
```diff
@@ -29,6 +29,14 @@ dependencies {
     testImplementation libs.smallrye
 }
 
+tasks.withType(JavaCompile).configureEach {
+    options.compilerArgs += [
+            '--add-exports', 'jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED'
```
melix: :x: you don't need this. The reason it happens is because of this code: https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/inject-java-test/src/main/java/io/micronaut/annotation/processing/test/JavaParser.java#L340-L347

But as the comment states this is only required for Java <15 and that we're building with 17, this whole block can be removed.

melix (changes requested):

inject-java-test/build.gradle
timyates: But then `inject-java-test:compileJava` fails with

https://ge.micronaut.io/s/k3o44k23ckrfq/console-log?task=:inject-java-test:compileJava

melix: I mean that you can remove the code in `JavaParser`, it's redundant since we won't run on Java<15.

timyates: Ahhhhhh...  Sorry about that! ☺️

Done in https://github.com/micronaut-projects/micronaut-core/pull/8154/commits/9fb60e2adccf024338487a34259456b88e1f1e68

melix (approved):

sdelamo:
@timyates any idea why this is failing?

melix:
It is failing because it defaults to Java 17 now, so it cannot run on 8 or 11.

timyates:
Lets see if https://github.com/micronaut-projects/micronaut-core/pull/8154/commits/df89c3594b0f5d6371743d44289602f9dae33fe8 fixes it

timyates:
@sdelamo All green 👍

sdelamo (approved):

PR #8156, opened 2022-10-13
Refactor `ConversionService` to remove mutable operations from `ConversionService.SHARED`

### LINKED ISSUE #7657: Do not allow registering shared state type convertor into the shared instance `ConversionService.DEFAULT`
### Issue description

There are some issues that are causing it https://github.com/micronaut-projects/micronaut-core/pull/7635

Major changes are needed:
- Remove methods that add converters to `ConversionService` and introduce `MutableConversionService`
- Allow contributing converters to `ConversionService.DEFAULT` only by service loader entries
- Register bean context converters to a managed instance of `ConversionService` that is delegating to `ConversionService.DEFAULT` just like Micronaut Data is doing https://github.com/micronaut-projects/micronaut-data/blob/a9c1f96ec45f3b4bf8dd250e8342e59e3b99d5b4/data-runtime/src/main/java/io/micronaut/data/runtime/convert/DataConversionServiceImpl.java

dstepanov:
Fixes #7657

Hopefully replaced all the possible usages of `ConversionService.SHARED`.

Some use cases cannot work after this change:
The detached request body is being converted:
```
        when:
        def request = HttpRequest.POST('/', JsonObject.createObjectNode([:]))

        then:
        request.getBody(String).get() != '{}'
```

Not sure if it's something that is even make sense, the fix now it to use this:
```
        when:
        request.setConversionService(conversionService)

        then:
        request.getBody(String).get() == '{}'
```

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
+++ b/aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
@@ -18,6 +18,7 @@ package io.micronaut.aop;
 import io.micronaut.aop.internal.intercepted.InterceptedMethodUtil;
 import io.micronaut.context.exceptions.ConfigurationException;
 import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.convert.ConversionService;
 import io.micronaut.core.type.Argument;
 import org.reactivestreams.Publisher;
 
@@ -40,7 +41,18 @@ public interface InterceptedMethod {
      * @return The {@link InterceptedMethod}
      */
     static InterceptedMethod of(MethodInvocationContext<?, ?> context) {
-        return InterceptedMethodUtil.of(context);
+        return of(context, ConversionService.SHARED);
+    }
+
+    /**
+     * Creates a new instance of intercept method supporting intercepting different reactive invocations.
+     *
+     * @param context           The {@link MethodInvocationContext}
+     * @param conversionService The conversion service
+     * @return The {@link InterceptedMethod}
+     */
+    static InterceptedMethod of(MethodInvocationContext<?, ?> context, ConversionService conversionService) {
+        return InterceptedMethodUtil.of(context, conversionService);
     }
 
     /**
--- a/core/src/main/java/io/micronaut/core/convert/ConversionService.java
+++ b/core/src/main/java/io/micronaut/core/convert/ConversionService.java
@@ -15,49 +15,24 @@
  */
 package io.micronaut.core.convert;
 
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.convert.exceptions.ConversionErrorException;
 import io.micronaut.core.type.Argument;
-import io.micronaut.core.annotation.Nullable;
+
 import java.util.Optional;
-import java.util.function.Function;
 
 /**
  * A service for allowing conversion from one type to another.
  *
- * @param <Impl> The type
  * @author Graeme Rocher
  * @since 1.0
  */
-public interface ConversionService<Impl extends ConversionService> {
+public interface ConversionService {
 
     /**
      * The default shared conversion service.
      */
-    ConversionService<?> SHARED = new DefaultConversionService();
-
-    /**
-     * Adds a type converter.
-     *
-     * @param sourceType    The source type
-     * @param targetType    The target type
-     * @param typeConverter The type converter
-     * @param <S>           The source generic type
-     * @param <T>           The target generic type
-     * @return This conversion service
-     */
-    <S, T> Impl addConverter(Class<S> sourceType, Class<T> targetType, Function<S, T> typeConverter);
-
-    /**
-     * Adds a type converter.
-     *
-     * @param sourceType    The source type
-     * @param targetType    The target type
-     * @param typeConverter The type converter
-     * @param <S>           The source generic type
-     * @param <T>           The target generic type
-     * @return This conversion service
-     */
-    <S, T> Impl addConverter(Class<S> sourceType, Class<T> targetType, TypeConverter<S, T> typeConverter);
+    ConversionService SHARED = new DefaultMutableConversionService();
 
     /**
      * Attempts to convert the given object to the given target type. If conversion fails or is not possible an empty {@link Optional} is returned.
--- a/core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
+++ b/core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2017-2020 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.convert;
+
+import java.util.function.Function;
+
+/**
+ * A version of {@link ConversionService} that supports adding new converters.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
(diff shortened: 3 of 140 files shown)


yawkat:
fyi this will conflict with #8100, I made a similar ConversionService change in DefaultHttpClient there

graemerocher:
@dstepanov can you resolve the conflicts

core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.function.Function;
+
+/**
+ * A version of {@link ConversionService} that supports adding new converters.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public interface MutableConversionService extends ConversionService {
```
graemerocher, line 28: add nullability annotations to methods

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
dstepanov: Done

dstepanov:
I think we can remove concurrent maps in the default implementation. I have measured it can get us 50% more throughput. The idea would be only to allow adding a new converter via `TypeConverterRegistrar`.

aop/src/main/java/io/micronaut/aop/InterceptedMethod.java
```diff
@@ -40,7 +41,18 @@ public interface InterceptedMethod {
...
+    }
+
+    /**
+     * Creates a new instance of intercept method supporting intercepting different reactive invocations.
+     *
+     * @param context           The {@link MethodInvocationContext}
+     * @param conversionService The conversion service
+     * @return The {@link InterceptedMethod}
+     */
+    static InterceptedMethod of(MethodInvocationContext<?, ?> context, ConversionService conversionService) {
```
graemerocher, line 54: add `@since`

graemerocher: add nullability annotations

core/src/main/java/io/micronaut/core/convert/ConversionService.java
```diff
@@ -15,49 +15,24 @@
...
-     *
-     * @param sourceType    The source type
-     * @param targetType    The target type
-     * @param typeConverter The type converter
-     * @param <S>           The source generic type
-     * @param <T>           The target generic type
-     * @return This conversion service
-     */
-    <S, T> Impl addConverter(Class<S> sourceType, Class<T> targetType, TypeConverter<S, T> typeConverter);
+    ConversionService SHARED = new DefaultMutableConversionService();
```
graemerocher, line 35: Should we deprecate this field?

core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Adds a type converter.
+     *
+     * @param sourceType    The source type
+     * @param targetType    The target type
+     * @param typeConverter The type converter
+     * @param <S>           The source generic type
+     * @param <T>           The target generic type
+     */
+    <S, T> void addConverter(@NonNull Class<S> sourceType, @NonNull Class<T> targetType, @NonNull Function<S, T> typeConverter);
```
graemerocher, line 50: Return `MutableConversionService` to make the API fluent

http/src/main/java/io/micronaut/http/converters/SharedHttpConvertersRegistrar.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Optional;
+
+/**
+ * Converter registrar for HTTP classes.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Internal
+public class SharedHttpConvertersRegistrar implements TypeConverterRegistrar {
```
graemerocher, line 39: make `final`. Does this need to be public?

inject/src/main/java/io/micronaut/context/ApplicationContext.java
```diff
@@ -59,11 +58,6 @@
  */
 public interface ApplicationContext extends BeanContext, PropertyResolver, PropertyPlaceholderResolver {
 
-    /**
-     * @return The default conversion service
-     */
-    @NonNull ConversionService<?> getConversionService();
```
graemerocher, line 65: Why was this deleted?

core/src/main/java/io/micronaut/core/convert/ConversionService.java
```diff
@@ -15,49 +15,24 @@
...
 
 /**
  * A service for allowing conversion from one type to another.
  *
- * @param <Impl> The type
  * @author Graeme Rocher
  * @since 1.0
  */
-public interface ConversionService<Impl extends ConversionService> {
+public interface ConversionService {
```
graemerocher, line 30: I'm concerned about the removal of the generic type argument here as it likely to break a lot of end user code. Is this a necessary change? What is the justification?

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/convert/ConversionService.java
dstepanov: No, you can still access the convertors registered using the service loader.

http/src/main/java/io/micronaut/http/converters/SharedHttpConvertersRegistrar.java
dstepanov: I think it needs to be public for the service loader.

inject/src/main/java/io/micronaut/context/ApplicationContext.java
dstepanov: It's now defined in `ConversionServiceProvider`, which the context extends.

core/src/main/java/io/micronaut/core/convert/ConversionService.java
dstepanov: It was only used for the mutable `add` methods, now those methods are gone so the generic parameter is unused.

core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
dstepanov: Do you want to have it generic like it was before?

core/src/main/java/io/micronaut/core/convert/ConversionService.java
graemerocher: It does mean though that any place a user injected `ConversionService<?>` will break which might be a significant breaking change. Thoughts @sdelamo ?

core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
graemerocher: Makes sense

dstepanov: I'm looking at it again, and I don't like to have it in this way. There are zero usages of chaining in the current codebase. The registration API defines this signature: `void register(ConversionService<?> conversionService)` which prevents to do the chaining. IMHO the generic parameter only makes the code ugly, and the chaining is not very important in this case. I would prefer not to have it. WDYT @micronaut-projects/core-developers ?

core/src/main/java/io/micronaut/core/convert/ConversionService.java
dstepanov: Users might need to modify their code anyway if relying on the shared conversion service.

core/src/main/java/io/micronaut/core/convert/MutableConversionService.java
graemerocher: It is probably fine without it but the generic argument remains a potentially large source of breaking changes

core/src/main/java/io/micronaut/core/convert/ConversionService.java
alvarosanchez: Our docs use `ConversionService<?>` in examples. Still I never liked that generic type argument, so my +1 to remove it if it's not needed

graemerocher (approved):

PR #8158, opened 2022-10-13
Fix use-case when `@Introspected` is added using `anotate`

dstepanov:
Fixes one of the issues in Serde

### DIFF
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/AnnotatedIntrospectedSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/AnnotatedIntrospectedSpec.groovy
@@ -1,0 +1,63 @@
+package io.micronaut.inject.visitor.beans
+
+import io.micronaut.annotation.processing.TypeElementVisitorProcessor
+import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
+import io.micronaut.annotation.processing.test.JavaParser
+import io.micronaut.core.beans.BeanIntrospection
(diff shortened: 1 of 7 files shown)


timyates:
We're seeing another issue in cache that I think is related

In the config for this test https://github.com/micronaut-projects/micronaut-cache/blob/1d5f83408e69a6c9717267b91d2ef8dc01b4aa39/cache-caffeine/src/test/groovy/io/micronaut/cache/SyncCacheSpec.groovy#L190 `maximumSize` is not being picked up, so the test is running with an Unbounded cache instead of a Bounded one as expected...

I've tested it with a snapshot from this commit, and it's not fixed...  Is there something else required here to fix this?

dstepanov:
Not sure how is that related. Is it some configuration bean?

timyates:
Yes, it's a field of `CacheConfiguration` https://github.com/micronaut-projects/micronaut-cache/blob/master/cache-core/src/main/java/io/micronaut/cache/CacheConfiguration.java#L56

Which is extended in Caffeine by https://github.com/micronaut-projects/micronaut-cache/blob/master/cache-caffeine/src/main/java/io/micronaut/cache/caffeine/configuration/CaffeineCacheConfiguration.java

And before a couple of days ago, was set correctly, but now is no longer set

It _feels_ like the same issue that we were seeing in Serde, with certain fields no longer being seen in the beans

graemerocher (approved):

dstepanov:
@timyates Need to investigate this usecase, it’s a bit strange that there is not configuration annotation

PR #8162, opened 2022-10-14
Fix STREAM_PIPELINE_ATTRIBUTE

yawkat:
Apparently there was a race condition with AttributeKey.newInstance being called multiple times


timyates (approved):
Sorry about that...  I thought AtomicReference would have handed races 🤔

PR #8174, opened 2022-10-17
Match interceptors by all the bindings it defines

dstepanov:
The PR changes the Micronaut behavior to be the same as CDI; the interceptor requires the intercepted point to have the same set of interceptor bindings as the interceptor.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/DefaultInterceptorRegistry.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/DefaultInterceptorRegistry.java
@@ -15,11 +15,6 @@
  */
 package io.micronaut.aop.chain;
 
-import java.lang.annotation.Annotation;
-import java.util.Collection;
-import java.util.List;
-import java.util.stream.Stream;
-
 import io.micronaut.aop.Adapter;
 import io.micronaut.aop.ConstructorInterceptor;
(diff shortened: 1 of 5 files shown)


graemerocher:
which issue is the PR addressing?

dstepanov:
I have updated the PR description.

graemerocher (approved):

dstepanov:
@micronaut-projects/core-developers Anyone has any objections to this change?

melix:
Can you illustrate the behavior changes, like before/after?

dstepanov:
In this example: 
```java
@Fast
@Deadly
@MissileAnn
@Singleton
class FastDeadlyInterceptor implements MethodInterceptor<Object, Object> {
}

@Fast
@Deadly
@MissileAnn
void myInterceptedMethod1() {}

@Fast
void myInterceptedMethod2() {}
```

In Micronaut 3 `myInterceptedMethod1` and `myInterceptedMethod2` would be intercepted by `FastDeadlyInterceptor`. After this change, only `myInterceptedMethod1` would be intercepted because it has all the annotations required by the interceptor, just like it's in CDI. So previous behavior was: do match if any annotation is present; now it's do match if all the annotations are present at the intercepted point.

dstepanov:
Added breaking docs

PR #8180, opened 2022-10-17
Virtual thread support

yawkat:
This PR introduces basic virtual thread support to micronaut-context. Virtual threads are available with `--enable-preview` on Java 19. The changes in this PR:

- Introduce a new `ExecutorType.THREAD_PER_TASK`. This uses the new `newThreadPerTaskExecutor`, which launches a new (hopefully virtual) thread for each submitted task.
- Introduce a new setting `ExecutorConfiguration.virtual` that changes the `ThreadFactory` to create virtual threads.
- Change the `TaskExecutors.IO` thread pool to use a virtual thread-per-task executor if virtual threads are available.

I have done preliminary benchmarking of virtual threads using our existing `ExecuteOn` machinery. Unfortunately, there is still overhead associated with dispatching controller methods on virtual threads. For this reason, I'm not proposing making virtual thread dispatch the default. I will have to investigate further whether there are optimization opportunities (removing thread locals?) that could help here.

Replacing the `IO` executor with a virtual thread pool seems like a good path forward. Existing users of `IO` will benefit from virtual threads without moving to a new executor. We also have a few internal uses of the IO executor.

Changing the IO executor to a virtual thread pool could be incompatible with the old executor, however. Virtual threads run on a mostly-fixed-size thread pool, and if many of the virtual threads are pinned (e.g. from a synchronized block), the carrier threads might be exhausted. The old IO pool was a cached pool, and would never exhaust its threads.

On implementation: 19-specific code is packaged into the new `LoomSupport` class, and accessed through static final MethodHandles. My understanding is that static final MHs carry no call overhead. I have yet to test it with graal however.

@melix suggested an approach based on multi-release jars (first commits of #8153). On Java 19, the full LoomSupport class would load, and on previous Java versions, it would be replaced by a stub class that does nothing. However this solution did not work out: Using the preview APIs apparently sets a flag in the class file that prevents it from being loaded on a non-preview Java 19 VM (there's a LinkageError). Fixing this would require guarding all accesses to the class (not just calls into new APIs) against such linkage errors, which is annoying (but probably possible).

@dstepanov proposed an alternative approach using bytecode generation (1265e9f25c8125c4c03a35f3abeae29eeaca39c4). It creates a new Executor bean with a new name (not `IO`), directly calling the `newVirtualThreadPerTaskExecutor` method, using existing machinery in micronaut-inject. My initial concern with this was that the executor cannot be configured, e.g. with the thread name. However most of the `ExecutorConfiguration` options (eg thread pool sizes) make no sense for virtual threads anyway, so this might not be a deal-breaker. Unfortunately, I could not get this approach to inject the executor properly. I'm also not sure of the failure mode on non-preview JDK 19: The bean is gated to only apply on JDK 19+, but the call to `newVirtualThreadPerTaskExecutor` will fail if `--enable-preview` is not set.

I think the ideal solution would be something like the `LoomSupport` class, but calling the API methods directly (with bytecode generation to avoid the failure when `--enable-preview` is missing). It's not as elegant as @dstepanov's approach that uses existing codegen infrastructure, but it won't be a big problem with ASM.

### DIFF
--- a/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
+++ b/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2003-2021 the original author or authors.
+ * Copyright 2017-2021 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     https://www.apache.org/licenses/LICENSE-2.0
+ * https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
+++ b/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
@@ -26,15 +26,18 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 
+/**
+ * @since 4.0.0
+ */
 @Internal
 public final class LoomSupport {
-    private static final boolean supported;
+    private static final boolean SUPPORTED;
     private static Throwable failure;
 
-    private static final MethodHandle MH_newThreadPerTaskExecutor;
-    private static final MethodHandle MH_ofVirtual;
-    private static final MethodHandle MH_name;
-    private static final MethodHandle MH_factory;
+    private static final MethodHandle MH_NEW_THREAD_PER_TASK_EXECUTOR;
+    private static final MethodHandle MH_OF_VIRTUAL;
+    private static final MethodHandle MH_NAME;
+    private static final MethodHandle MH_FACTORY;
 
     static {
         boolean sup;
@@ -67,17 +70,18 @@ public final class LoomSupport {
             failure = e;
         }
 
-        supported = sup;
-        MH_newThreadPerTaskExecutor = newThreadPerTaskExecutor;
-        MH_ofVirtual = ofVirtual;
-        MH_name = name;
-        MH_factory = factory;
+        SUPPORTED = sup;
+        MH_NEW_THREAD_PER_TASK_EXECUTOR = newThreadPerTaskExecutor;
+        MH_OF_VIRTUAL = ofVirtual;
+        MH_NAME = name;
+        MH_FACTORY = factory;
     }
 
-    private LoomSupport() {}
+    private LoomSupport() {
+    }
 
     public static boolean isSupported() {
-        return supported;
+        return SUPPORTED;
     }
 
     public static void checkSupported() {
@@ -89,7 +93,7 @@ public final class LoomSupport {
     public static ExecutorService newThreadPerTaskExecutor(ThreadFactory threadFactory) {
         checkSupported();
         try {
-            return (ExecutorService) MH_newThreadPerTaskExecutor.invokeExact(threadFactory);
+            return (ExecutorService) MH_NEW_THREAD_PER_TASK_EXECUTOR.invokeExact(threadFactory);
         } catch (Throwable e) {
             throw new RuntimeException(e);
         }
@@ -98,9 +102,9 @@ public final class LoomSupport {
     public static ThreadFactory newVirtualThreadFactory(String namePrefix) {
         checkSupported();
         try {
-            Object builder = MH_ofVirtual.invoke();
-            builder = MH_name.invoke(builder, namePrefix, 1L);
-            return (ThreadFactory) MH_factory.invoke(builder);
+            Object builder = MH_OF_VIRTUAL.invoke();
+            builder = MH_NAME.invoke(builder, namePrefix, 1L);
+            return (ThreadFactory) MH_FACTORY.invoke(builder);
         } catch (Throwable e) {
             throw new RuntimeException(e);
         }
--- a/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
+++ b/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
@@ -15,6 +15,8 @@
  */
 package io.micronaut.scheduling;
 
+import io.micronaut.context.condition.Condition;
+import io.micronaut.context.condition.ConditionContext;
 import io.micronaut.core.annotation.Internal;
 
 import java.lang.invoke.MethodHandle;
@@ -101,6 +103,23 @@ public final class LoomSupport {
             return (ThreadFactory) MH_factory.invoke(builder);
         } catch (Throwable e) {
             throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Condition that only matches if virtual threads are supported on this platform.
+     */
+    @Internal
+    public static class LoomCondition implements Condition {
+        @SuppressWarnings("rawtypes")
+        @Override
+        public boolean matches(ConditionContext context) {
+            if (isSupported()) {
+                return true;
+            } else {
+                context.fail("Virtual threads support not available: " + failure.getMessage());
+                return false;
+            }
         }
     }
 }
--- a/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
+++ b/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2003-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


context/src/main/java/io/micronaut/scheduling/executor/ExecutorFactory.java
```diff
@@ -57,7 +60,13 @@ public ExecutorFactory(BeanLocator beanLocator, ThreadFactory threadFactory) {
      */
     @EachBean(ExecutorConfiguration.class)
     protected ThreadFactory eventLoopGroupThreadFactory(ExecutorConfiguration configuration) {
-        return configuration.getName() == null ? threadFactory : new NamedThreadFactory(configuration.getName() + "-executor");
+        if (configuration.isVirtual()) {
```
melix, line 64: Can `configuration.getName()` return `null` in case of virtual?

context/src/main/java/io/micronaut/scheduling/executor/ExecutorType.java
```diff
@@ -42,5 +42,10 @@ public enum ExecutorType {
     /**
      * @see java.util.concurrent.Executors#newWorkStealingPool()
      */
-    WORK_STEALING
+    WORK_STEALING,
+
+    /**
+     * @see java.util.concurrent.Executors#newVirtualThreadPerTaskExecutor()
+     */
+    THREAD_PER_TASK
```
melix, line 50: I wonder if we should use another name, like `VIRTUAL_THREAD_PER_TASK` since technically speaking we could implement `THREAD_PER_TASK` with platform threads.

context/src/main/java/io/micronaut/scheduling/executor/ExecutorFactory.java
yawkat: yea, not sure what to do then, since we can't use the default threadFactory

context/src/main/java/io/micronaut/scheduling/executor/ExecutorType.java
yawkat: it just uses newThreadPerTaskExecutor, so if you don't also configure `virtual: true`, it does use platform threads. so the name is accurate.

context/src/main/java/io/micronaut/scheduling/executor/ExecutorConfiguration.java
```diff
@@ -75,6 +75,8 @@ default String getName() {
      */
     @Min(1L) Integer getCorePoolSize();
 
+    boolean isVirtual();
```
graemerocher, line 81: needs javadoc

context/src/main/java/io/micronaut/scheduling/LoomSupport.java
```diff
@@ -0,0 +1,106 @@
...
+public final class LoomSupport {
+    private static final boolean supported;
+    private static Throwable failure;
+
+    private static final MethodHandle MH_newThreadPerTaskExecutor;
+    private static final MethodHandle MH_ofVirtual;
+    private static final MethodHandle MH_name;
+    private static final MethodHandle MH_factory;
+
+    static {
```
graemerocher, line 42: needs testing on native and possibly registration of reflective metadata

graemerocher (commented):
Looks like a good start, but we do need to analyze the reduction of use of thread locals and general overhead of spinning up virtual threads.

yawkat:
benchmark of a very simple controller that does no io. 'directIO' has an additional `@ExecuteOn(TaskExecutors.IO)`.

enable-preview off (directIO uses normal cached thread pool):
```
Benchmark                      (uri)  Mode  Cnt    Score   Error  Units
JmhBenchmark.test    /urandom/direct  avgt   10  119.468 ± 0.475  us/op
JmhBenchmark.test  /urandom/directIO  avgt   10  168.815 ± 3.249  us/op
```

enable-preview on (directIO uses virtual-thread-per-task pool):
```
Benchmark                      (uri)  Mode  Cnt    Score   Error  Units
JmhBenchmark.test    /urandom/direct  avgt   10  120.734 ± 3.519  us/op
JmhBenchmark.test  /urandom/directIO  avgt   10  182.292 ± 4.063  us/op
```

dstepanov:
I guess the reflection solution is fine too. For Graal we wound need a reflection config.

Do we want to eliminate the existing IO pool? Maybe we could introduce new `TaskExecutors.BLOCKING` and delegate IO or virtual threads there. WDYT?

I would be interested to see benchmarks of a real word use-case: Controller invoking HTTP client with a different setup: Reactive, ExecuteOn IO, ExecuteOn VirtualThreads

melix:
I agree a benchmark without IO doesn't make much sense except for measuring the overhead, if any. Virtual threads should shine when blocking I/O is involved.

yawkat:
Unfortunately something that actually blocks is harder to benchmark :) I tried a simple read from `/dev/urandom` at first, but that is so fast that the results are similar to no IO at all. I've also tried a sleep(1000) in a controller, but the JMH benchmark shows no difference (unsurprising since it's sequential), and a simple gatling test was wayy too noisy. I think the benefit will start appearing once I benchmark the *memory consumption* with parallel requests.

yawkat:
> Do we want to eliminate the existing IO pool? Maybe we could introduce new TaskExecutors.BLOCKING and delegate IO or virtual threads there. WDYT?

This sounds like a good alternative. I'm gonna try this, and move some of the existing uses of TaskExecutors.IO in micronaut-core to the new BLOCKING pool.

dstepanov:
> Unfortunately something that actually blocks is harder to benchmark :) I tried a simple read from `/dev/urandom` at first, but that is so fast that the results are similar to no IO at all. I've also tried a sleep(1000) in a controller, but the JMH benchmark shows no difference (unsurprising since it's sequential), and a simple gatling test was wayy too noisy. I think the benefit will start appearing once I benchmark the _memory consumption_ with parallel requests.

My understanding is that blocking is what we want, devs want to avoid reactive. Maybe you can take the techempower benchmark and run it?

yawkat:
i dont think the TE benchmarks are hugely parallel either

yawkat:
new jmh (sequential) results with the new separate 'virtual' executor: 

```
Benchmark                      (uri)  Mode  Cnt    Score   Error  Units
JmhBenchmark.test    /urandom/direct  avgt   10  117.031 ± 1.198  us/op
JmhBenchmark.test  /urandom/directIO  avgt   10  174.397 ± 1.762  us/op
JmhBenchmark.test   /urandom/directV  avgt   10  180.453 ± 1.169  us/op
```

yawkat:
ive tested with graalvm for java 19, it works fine as-is, no reflection config required. Looks like graal recognizes the method handles properly. It runs methods marked with `@ExecuteOn(VIRTUAL)` on virtual threads as expected.

yawkat:
ive done some more profiling with async-profiler, and the difference between cached pool and virtual threads seems to be down to fork join overhead. I've also looked at thread locals, and they don't seem to be an issue. We can't fully get rid of them anyway, logback uses some internally.

i can also see why @dstepanov wants to get rid of reactive code paths, now :smile:

yawkat:
the performance difference starts to show up with many parallel connections all calling a controller that sleeps for 1ms:

```
Benchmark                              (uri)  Mode  Cnt    Score   Error  Units
ParallelBenchmark.test           /sleep/none  avgt   10   55.109 ± 0.346  us/op
ParallelBenchmark.test         /sleep/io?t=1  avgt   10  101.974 ± 2.690  us/op
ParallelBenchmark.test          /sleep/v?t=1  avgt   10   83.941 ± 0.559  us/op
ParallelBenchmark.test         /sleep/io?t=0  avgt   10   86.504 ± 0.790  us/op
ParallelBenchmark.test          /sleep/v?t=0  avgt   10   78.652 ± 0.512  us/op
```

`/sleep/none` does no sleeping and runs on the event loop, `/sleep/v` runs on a virtual thread, `/sleep/io` runs on the io pool. `t=` is the sleep time in ms.

yawkat:
ive checked out the netty thread locals that quarkus has special handling for here: https://github.com/quarkusio/quarkus/blob/main/extensions/netty-loom-adaptor/deployment/src/main/java/io/quarkus/netty/loom/adaptor/NettyLoomAdaptorProcessor.java

The main concern is PoolThreadCache, which is a thread-local pool for buffer allocation. If such a pool were to be created for every virtual thread, this would cause large allocation pressure, most of it wasted. What quarkus does here is change the pool to be local to the *carrier thread* instead of the *virtual thread*. This requires both bytecode transformation of netty (to change the code), and reflection into the JDK (the carrier thread is not accessible otherwise). However since some time this year, netty already defaults to disable an option called `useCacheForAllThreads`, so it doesn't actually use this pool cache for threads that aren't event loop threads.

I'm not sure why quarkus implemented this processor. My guess is that it is either older than the netty change for useCacheForAllThreads, or that they used it to improve allocation performance when they used the event loop threads as carrier threads (they have since moved on from that approach). We don't need to follow their example here.

Netty also has a thread-local class (https://github.com/netty/netty/blob/cf9dc4099045f5ad13983f02c86191c06bcd2da4/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java) it uses to cache various data structures, e.g. ArrayLists, to further reduce allocations. At some point it may be worth looking at turning off this caching for virtual threads, since they are comparatively shorter-lived, but I don't think this is worth a patch until profiling shows it to be a problem.

graemerocher (approved):

dstepanov (approved):

PR #8182, opened 2022-10-18
Fix NPE in Groovy processing

dstepanov:


inject-groovy/src/main/groovy/io/micronaut/ast/groovy/utils/AstMessageUtils.groovy
```diff
@@ -39,8 +40,8 @@ class AstMessageUtils {
      * @param node The AST node
      * @param message The message
      */
-    static void warning(final SourceUnit sourceUnit, final ASTNode node, final String message) {
-        final String sample = sourceUnit.getSample(node.getLineNumber(), node.getColumnNumber(), new Janitor())
+    static void warning(SourceUnit sourceUnit, @Nullable ASTNode node, String message) {
+        final String sample = node ? null : sourceUnit.getSample(node.getLineNumber(), node.getColumnNumber(), new Janitor())
```
sdelamo: should not be reversed?

```suggestion
        final String sample = node ? sourceUnit.getSample(node.getLineNumber(), node.getColumnNumber() : null, new Janitor())
```

sdelamo (changes requested):
When was this NPE happening? Is there a test for this?

PR #8197, opened 2022-10-20
Reuse enclosed elements querying for Java and Groovy. Added a way to query methods and fields together.

dstepanov:


inject/src/main/java/io/micronaut/inject/ast/utils/EnclosedElementsQuery.java
```diff
@@ -0,0 +1,292 @@
...
+                    }
+                    if (!typePredicate.test(ce)) {
+                        return false;
+                    }
+                }
+            }
+// TODO: FIX only injected
+//                if (result.isOnlyInjected() && !element.hasDeclaredAnnotation(AnnotationUtil.INJECT)) {
+//                    return false;
+//                }
```
graemerocher: should this TODO be addressed before merging? Seems like this would be a breaking change otherwise?

dstepanov: Actually that’s is not needed anymore, this was applying some extra rules which are now by default

graemerocher: perhaps delete the code in that case?

dstepanov: Deleted

dstepanov: Undeleted it :D found some logic that was added in the query method.

timyates:
Awesome @dstepanov, micronaut-test:test-spock now passes with this PR 🎉

graemerocher (approved):

dstepanov:
@graemerocher Can you merge it, please?

PR #8147, opened 2022-10-12
Support custom HTTP status codes

yawkat:
This patch adds support for custom status codes to various HTTP server and client classes.

- On HttpResponse, the code()/reason() getters and the corresponding setter become the "primary" accessors for the HTTP status. Usages of status() and getStatus() are replaced where possible.
- HttpStatus instances are only computed lazily when the user calls status(), which will produce an error for unknown status codes. If the method is not called, there is no error.
- Various HttpResponse implementations are changed to not store HttpStatus.

Fixes #4791

PR #8199, opened 2022-10-21
Deprecate HttpStatus.UNORDERED_COLLECTION

yawkat:
For #8123

### DIFF
--- a/http/src/main/java/io/micronaut/http/HttpStatus.java
+++ b/http/src/main/java/io/micronaut/http/HttpStatus.java
@@ -70,10 +70,11 @@ public enum HttpStatus implements CharSequence {
     LOCKED(423, "Locked"),
     FAILED_DEPENDENCY(424, "Failed Dependency"),
     /**
-     * @deprecated Will be replaced by {@code TOO_EARLY} in 4.0
+     * @deprecated Will be replaced by {@link #TOO_EARLY} in 4.0
      */
     @Deprecated
     UNORDERED_COLLECTION(425, "Unordered Collection"),
+    TOO_EARLY(425, "Too Early"),
     UPGRADE_REQUIRED(426, "Upgrade Required"),
     PRECONDITION_REQUIRED(428, "Precondition Required"),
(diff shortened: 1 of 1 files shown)


timyates (commented):
Should we add

```
TOO_EARLY(425, "Too Early"),
```

To give users a migration path?

timyates:
Is adding an enum binary compatible? 🤔 🤯

yawkat:
we dont have precedent for multiple enum values with the same code, not sure. adding an enum value is binary compatible.

i was going to suggest users use numeric status codes, but apparently #8147 is 4.0 only

yawkat:
though i guess people *can* still use numeric status codes here, since 425 is not invalid in either 3.x or 4.0, it's just represented by different enum values

timyates:
Trying to see what JapiCmp says about

```diff
diff --git a/http/src/main/java/io/micronaut/http/HttpStatus.java b/http/src/main/java/io/micronaut/http/HttpStatus.java
index 81ec802227..b5901afb12 100644
--- a/http/src/main/java/io/micronaut/http/HttpStatus.java
+++ b/http/src/main/java/io/micronaut/http/HttpStatus.java
@@ -74,6 +74,7 @@ public enum HttpStatus implements CharSequence {
      */
     @Deprecated
     UNORDERED_COLLECTION(425, "Unordered Collection"),
+    TOO_EARLY(425, "Too Early"),
     UPGRADE_REQUIRED(426, "Upgrade Required"),
     PRECONDITION_REQUIRED(428, "Precondition Required"),
     TOO_MANY_REQUESTS(429, "Too Many Requests"),
```

But it's currently not running 😿

timyates:
@yawkat japicmp says the above diff change is fine (once you work around Gradle bug above locally)...  Can you think of an edge case where it won't work?

yawkat:
No, unless you're relying on an exhaustive switch, or on the ordinals of other status codes

timyates (approved):

graemerocher (approved):

PR #8203, opened 2022-10-21
change missed Java 11 ref in github workflow files

wetted:


timyates (changes requested):
That's not a version number, that's removing the first 11 chars from GITHUB_REF

altro3:
@timyates Not sure you are right. Look to micronaut tracing commit by @sdelamo :

изображение

Or, maybe it is mistake took and need to revert 11 in this place

timyates:
I believe that's a mistake

timyates:
As an explanation, bash and zsh allow the following 3 forms of variable expansion (there are more, but this change is one of these)

```
# just the variable
echo ${VAR}
# variable with offset
echo ${VAR:3}
# variable with offset and length
echo ${VAR:3:2}
```

So, if VAR was set to "micronaut" then the above would print

```
micronaut
ronaut
ro
```

I believe the line that changed here strips the first 11 chars off GITHUB_REF, which just leaves you with the branch name

PR #8208, opened 2022-10-22
Let SemanticVersion.isAtLeastMajorMinor be able to compare two different major versions.

donbeave:
I'm playing around with Micronaut 4.0.0-SNAPSHOT version and found that the AOT plugin failed with such exception:

```
Caused by: java.lang.RuntimeException: This version of the AOT optimizer requires at least Micronaut 3.3 but found 4.0.0-SNAPSHOT
	at io.micronaut.aot.MicronautAotOptimizer.assertMinimalMicronautVersion(MicronautAotOptimizer.java:281)
```

It was because the current implementation of `isAtLeastMajorMinor` method will return false for comparing 4.0.0 and 3.3.0 versions.

```java
// returns: false
SemanticVersion.isAtLeastMajorMinor("4.0.0", 3, 3)
```

### DIFF
--- a/core/src/main/java/io/micronaut/core/version/SemanticVersion.java
+++ b/core/src/main/java/io/micronaut/core/version/SemanticVersion.java
@@ -121,6 +121,12 @@ public class SemanticVersion implements Comparable<SemanticVersion> {
     }
 
     private static boolean isAtLeastMajorMinorImpl(SemanticVersion version, int majorVersion, int minorVersion) {
-        return version != null && version.major >= majorVersion && version.minor >= minorVersion;
+        if (version != null) {
+            if (version.major == majorVersion) {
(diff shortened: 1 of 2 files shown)


sdelamo (approved):

core/src/main/java/io/micronaut/core/version/SemanticVersion.java
```diff
@@ -121,6 +121,12 @@ public static boolean isAtLeast(String version, String requiredVersion) {
     }
 
     private static boolean isAtLeastMajorMinorImpl(SemanticVersion version, int majorVersion, int minorVersion) {
```
sdelamo, line 123: I think we should add a javadoc or a better method name.

PR #8216, opened 2022-10-24
Correct record constructor search algorithm. Fixes #8187

### LINKED ISSUE #8187: Introspections choose wrong constructor with records that have a constructor initializer
### Expected Behavior

The primary record constructor should always be the creator in the absence of a creator annotation.

### Actual Behaviour

The following code:

```java
@MappedEntity
record Pet(
        @GeneratedValue
        @Id
        Long id,
        String name,
        int age,
        @Relation(Relation.Kind.MANY_TO_ONE)
        Person owner,
        PetType type) implements SentientBeing  {

    public Pet {
        if (age <= 0) {
            throw new IllegalArgumentException("Age must be a positive number");
        }
    }

    public Pet(String name, int age, Person owner) {
        this(null, name, age, owner, PetType.DOG);
    }


    enum PetType {
        DOG, CAT
    }
}

```

Fails with:

```
Caused by: java.lang.UnsupportedOperationException: Cannot mutate property [id] that is not mutable via a setter method or constructor argument for type: java19.demo.Pet
	at java19.demo.$Pet$Introspection.dispatchOne(Unknown Source)
	at io.micronaut.inject.beans.AbstractInitializableBeanIntrospection$BeanPropertyImpl.withValueUnsafe(AbstractInitializableBeanIntrospection.java:476)
	at io.micronaut.inject.beans.AbstractInitializableBeanIntrospection$BeanPropertyImpl.withValue(AbstractInitializableBeanIntrospection.java:462)
```

Unless you annotate the constructor initializer with `@Creator`:

```java
    @Creator
    public Pet {
        if (age <= 0) {
            throw new IllegalArgumentException("Age must be a positive number");
        }
    }
```

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.7.2

graemerocher:
Modifies `ModelUtils.concreteConstructorFor` to do a comparison of the record components vs the constructor parameters to ensure the correct constructor is picked for records.

Fixes #8187

### DIFF
--- a/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
+++ b/inject-java-test/src/test/groovy/io/micronaut/inject/visitor/beans/BeanIntrospectionSpec.groovy
@@ -62,7 +62,7 @@ class Test<T extends CharSequence> {
     public void setArray(T[] array) {
         this.array = array;
     }
-    
+
     @Executable
     T[] myMethod() {
         return array;
@@ -90,7 +90,7 @@ class Test {
     public String getOne() {
         this.invoked = true;
         return one;
-    } 
+    }
 }
 ''');
 
@@ -158,7 +158,7 @@ class Test {
     public String getOne() {
         this.invoked = true;
         return one;
-    } 
+    }
 }
 ''');
 
@@ -227,7 +227,7 @@ class Test {
     String three; // package protected
     protected String four; // not included since protected
     private String five; // not included since private
-    
+
     Test(int two) {
         this.two = two;
     }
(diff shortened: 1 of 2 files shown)


dstepanov (approved):

dstepanov:
I think this issue also occurs for ordinary beans, the current behavior is to take the first constructor. 
We might want to apply the same rules in Micronaut 4 for ordinary beans as well.

graemerocher:
I don't know how you would apply the same behaviour there are no record components in ordinary beans

dstepanov:
In Micronaut 4 we can check if the class is annotated with `@BeanProperties` then we can read the properties and find the constructor in the same fashion.

graemerocher:
yes also in Micronaut 4 we can remove the reflection calls which are not needed there

PR #8217, opened 2022-10-24
HTTP Netty server executing route refactoring

dstepanov:
This PR introduces a new wrapper object, `Flow` that can represent different invocation flows: imperative, reactive, async. The flow supports a basic set of methods that we use for reactive interaction: `map`, `flatMap` etc. The benefit is that in some cases when reactive code is not needed, the flow is implemented as a simple object wrapper, which allows us to eliminate any reactive interaction of the imperative flow. The idea is similar to how Micronaut Data can share some code between reactive and imperative invocation.

This PR also refactors some interactions between `RoutingInBoundHandler` and `RouteExecutor`. I have extracted the part from `RoutingInBoundHandler` that doesn't need to touch Netty, which will allow us to share the same code for the servlet processing implementation.

There is a new test that validates there are no Reactor classes involved in the imperative invocations. Interestingly is that even with filters added, there is no extra stack record of the reactor; that's because the code eliminated the overuse of the `Mono#deferred`. In Microanut 3.7 all the tests of `InvocationStackSpec` fail.

More improvements need to be made:
- `RequestBodyReader` needs to be refactored to eliminate the route modification during the processing
- Kotlin coroutines are always processed using the reactive code instead of `CompletableFuture`


graemerocher:
can we name it something else? There is already `java.util.concurrent.Flow` and this might cause confusion

core-reactive/build.gradle
```diff
@@ -5,4 +5,5 @@ plugins {
 dependencies {
     api project(":core")
     api libs.managed.reactive.streams
+    implementation libs.managed.reactor
```
graemerocher: `core-reactive` shouldn't depend on a concrete impl

core-reactive/src/main/java/io/micronaut/core/async/flow/ReactiveFlowImpl.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.function.Supplier;
+
+/**
+ * The reactive flow implementation.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Internal
+final class ReactiveFlowImpl implements ReactiveFlow<Object> {
```
graemerocher: should be `ReactorFlow` or something since this is for reactor?

core/src/main/java/io/micronaut/core/flow/CompletableFutureFlowImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public Flow<Object> onErrorResume(Function<? super Throwable, ? extends Flow<?>> fallback) {
+        stage = stage.exceptionallyCompose(throwable -> (CompletionStage<Object>) fallback.apply(throwable).toCompletableFuture());
+        return this;
+    }
+
+    @Override
+    public Flow<Object> putInContext(String key, Object value) {
+        return this;
```
graemerocher: probably needs an imlementation?

dstepanov: Yeah I was thinking to have some context associated object

core-reactive/build.gradle
dstepanov: I was thinking to put the reactor related code in some specific place

dstepanov:
> can we name it something else? There is already `java.util.concurrent.Flow` and this might cause confusion

Any suggestions?

dstepanov:
> > can we name it something else? There is already `java.util.concurrent.Flow` and this might cause confusion
> 
> Any suggestions?

Maybe just Flw?

core-reactive/build.gradle
graemerocher: Seems like the reactor impl should be `micronaut-reactor`

graemerocher:
Maybe `ExecutionFlow`?

core-reactive/src/main/java/io/micronaut/core/async/flow/ReactiveFlowImpl.java
dstepanov: Renamed

core-reactive/build.gradle
dstepanov: I moved it to `runtime` for now, we have all the reactor stuff there.

core/src/main/java/io/micronaut/core/flow/CompletableFutureFlowImpl.java
dstepanov: I will leave it like this for now because I don't have a test for it. In the next PR, I want to improve the second point I mentioned in the description, the context might be needed there.

core/src/main/java/io/micronaut/core/execution/ImperativeExecutionFlowImpl.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.function.Supplier;
+
+/**
+ * The imperative execution flow implementation.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Internal
+final class ImperativeExecutionFlowImpl implements ImperativeExecutionFlow<Object> {
```
graemerocher, line 36: If the interface was modified this could be a record

dstepanov: I made all the implementations mutable, so we don't create additional instances.

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * The completable future execution flow implementation.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Internal
+final class CompletableFutureExecutionFlowImpl implements CompletableFutureExecutionFlow<Object> {
+
+    private CompletableFuture<Object> stage;
```
graemerocher, line 35: declare `final`

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public CompletableFutureExecutionFlowImpl(CompletableFuture<Object> stage) {
```
graemerocher: does this need to be public?

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public ExecutionFlow<Object> onErrorResume(Function<? super Throwable, ? extends ExecutionFlow<?>> fallback) {
+        stage = stage.exceptionallyCompose(throwable -> (CompletionStage<Object>) fallback.apply(throwable).toCompletableFuture());
+        return this;
+    }
+
+    @Override
+    public ExecutionFlow<Object> putInContext(String key, Object value) {
+        return this;
```
graemerocher, line 72: an implementation before merging?

core/src/main/java/io/micronaut/core/execution/ImperativeExecutionFlow.java
```diff
@@ -0,0 +1,51 @@
...
+     * @return The exception if present
+     */
+    @Nullable
+    Throwable getError();
+
+    /**
+     * @return The context if present
+     */
+    @Nullable
+    Map<String, Object> getContext();
```
graemerocher, line 50: does it make sense to return an empty map instead of `null`?

core/src/main/java/io/micronaut/core/execution/ImperativeExecutionFlowImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Nullable
+    private Object value;
+    @Nullable
+    private Throwable error;
+    @Nullable
+    private Map<String, Object> context;
```
graemerocher, line 43: can these be `final`?

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
dstepanov: It is mutable

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
dstepanov: Currently there is no use case for it, I would prefer to implement it in the future PRs if it’s needed

core/src/main/java/io/micronaut/core/execution/ImperativeExecutionFlowImpl.java
dstepanov: The implementation does mutates them instead of creating a new instance of the flow

graemerocher:
Can you address the sonar bug? https://sonarcloud.io/project/issues?id=micronaut-projects_micronaut-core&pullRequest=8217&resolved=false&types=BUG

core/src/main/java/io/micronaut/core/execution/ImperativeExecutionFlow.java
dstepanov: done

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
dstepanov: made them all private

graemerocher (approved):

dstepanov:
I had to disable one test. There is a missing null check in the Micronaut Tracing. @graemerocher Please check and merge if it's ok

Issue #7275, opened 2022-04-27
CircuitBreaker should not throw the original exception when the circuit is open


tom59:
### Feature description

The default behaviour for the micronaut CircuitBreaker is to throw the original exception when the circuit is open.
https://github.com/micronaut-projects/micronaut-core/blob/v3.4.2/runtime/src/main/java/io/micronaut/retry/intercept/CircuitBreakerRetry.java#L96
This is very confusing and doesn't really make much sense in a production application.
Why isn't CircuitOpenException thrown for all cases?

For example, if I have a client which is configured to retry once:
```
@Client
public interface SummaryClient {
    @Get("/summary")
    @CircuitBreaker(delay = "15s", attempts = "1", reset = "300s")
    protected abstract Single<Payload> getSummary()
}
```
I would expect the following in the logs:
```
io.micronaut.http.client.exceptions.ReadTimeoutException // First Attempt. Outbound timed out
io.micronaut.http.client.exceptions.ReadTimeoutException //Second Attempt. Outbound timed out
io.micronaut.retry.exception.CircuitOpenException // Circuit breaker is now open. No outbound calls were made
io.micronaut.retry.exception.CircuitOpenException // Circuit breaker is now open. No outbound calls were made
```
While we are getting this:
```
io.micronaut.http.client.exceptions.ReadTimeoutException // First Attempt. Outbound timed out
io.micronaut.http.client.exceptions.ReadTimeoutException //Second Attempt. Outbound timed out
io.micronaut.http.client.exceptions.ReadTimeoutException // Circuit breaker is now open. No outbound calls were made
io.micronaut.http.client.exceptions.ReadTimeoutException // Circuit breaker is now open. No outbound calls were made
```

I have a workaround to create a Fallback class for the client but that does seem to be really a hack.
```
@Fallback
public class FallbackSummaryClient extends SummaryClient {
    @Override
    protected Single<Summary> getSummary() {
        return Single.error(new CircuitOpenException("Circuit breaker is open."));
    }
}
```

jameskleeh:
Can you describe why it doesn't make sense? What problems is it causing? How would a CircuitOpenException help you?

graemerocher:
It is also unclear to me why this is a problem. Seems to be working exactly as designed

tom59:
It is working as designed in the code, that's why I put it as an enhancement.
[...]

PR #8222, opened 2022-10-25
wrap original exception when the circuit breaker is opened

aprietop:
Addresses #7275


graemerocher:
could you target the 3.8.x branch as this is now an enhancement

aprietop:
> could you target the 3.8.x branch as this is now an enhancement
Now targeting 3.8.x @graemerocher

graemerocher:
Thanks for the contribution!

PR #8235, opened 2022-10-26
Split the runtime module into multiple modules

graemerocher:
This PR splits the `micronaut-runtime` module into multiple modules.  The new modules are:

* `micronaut-retry` - The retry logic implementation is now optional. In order to support this by default declarative http clients are no longer automatically `@Recoverable` (Fixes #3719). This is a better arrangement as not everyone needs or uses fallbacks and recovery
* `micronaut-discovery` - The service discovery implementation is split into a separate module that is no longer required. The HTTP client module still depends on it though and would be nice to remove this coupling.

Following this change `micronaut-runtime` is just a meta module that depends on other modules. Other Micronaut modules should update and depend specifically on an individual module such not to pull in more than is necessary.

The motivation for this is to reduce the footprint of a new app since many features are not used (heartbeat, health checks, service discovery, retry etc.) unless you actually need them.

In addition the `micronaut-http-server-netty` module no longer depends on `micronaut-discovery`, `micronaut-health` and `micronaut-websocket`. If users want any of those features then they can add them individually to their application. Micronaut Launch will need to be updated with new feature for WebSocket support. It is probably not necessary to include new features for discovery and health as these will be transitively pulled by the Management module.

Also fixes #8099 by introducing a new interface to resolve the `JsonMapper`

Note this PR also moves some functionality into other modules:

https://github.com/micronaut-projects/micronaut-kotlin/pull/440
https://github.com/micronaut-projects/micronaut-reactor/pull/237

This change also switches Micronaut's integration to GraalVM to be largely runtime initialized including logging. Only Micronaut metadata is build time initialized. This change improves startup time and reduces the size of the images we build.

PR #8238, opened 2022-10-26
Ensure Logback is refreshed before application starts

### LINKED ISSUE #8139: [GRAALVM] Logback not fetching Environment Variable
### Expected Behavior

We are using [Logstash Logback Encoder](https://github.com/logfellow/logstash-logback-encoder) and we are trying to inject Field Values as Environment Variables.

e.g.:
```
<pattern>
                    <omitEmptyFields>true</omitEmptyFields>
                    <pattern>
                        {
                            "CorrelationId": "%mdc{trace_id}",
                            "Resource": {
                                "application": "demo",
                                "logger": "%logger",
                                "version":"${VERSION}",
                                "environment":"${ENVIRONMENT}",
                            }
                        }
                    </pattern>
                </pattern>
```

### Actual Behaviour

Variables are not properly filled. Result:

```
{
    "Timestamp": "2022-10-10T10:01:16.565Z",
    "SeverityText": "INFO",
    "Resource": {
        "application": "demo",
        "logger": "io.micronaut.function.aws.proxy.AbstractLambdaContainerHandler",
        "version": "VERSION_IS_UNDEFINED"
        "environment": "ENVIRONMENT_IS_UNDEFINED"
    },
    "Body": {
        "message": "Starting Lambda Container Handler",
        "thread": "main"
    }
}

```

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

https://github.com/driverpt/micronaut-otel-issue

### Version

3.7.1

driverpt:
Attempts to fix #8139

### DIFF
--- a/runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
+++ b/runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
@@ -42,7 +42,7 @@ import org.slf4j.LoggerFactory;
 @Internal
 public final class LogbackLoggingSystem implements LoggingSystem {
 
-    private static String DEFAULT_LOGBACK_LOCATION = "logback.xml";
+    private static final String DEFAULT_LOGBACK_LOCATION = "logback.xml";
 
     @Property(name = "logger.config")
     private Optional<String> logbackXmlLocation;
--- a/runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
+++ b/runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
@@ -15,12 +15,19 @@
  */
 package io.micronaut.logging.impl;
 
+import java.net.URL;
+import java.util.Optional;
+
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.joran.JoranConfigurator;
+import ch.qos.logback.core.joran.spi.JoranException;
+import io.micronaut.context.annotation.Property;
 import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.logging.LogLevel;
 import io.micronaut.logging.LoggingSystem;
+import io.micronaut.logging.LoggingSystemException;
 import jakarta.inject.Singleton;
 import org.slf4j.LoggerFactory;
 
@@ -35,6 +42,11 @@ import org.slf4j.LoggerFactory;
 @Internal
 public final class LogbackLoggingSystem implements LoggingSystem {
 
+    private static String DEFAULT_LOGBACK_LOCATION = "logback.xml";
+
+    @Property(name = "logger.config")
+    private Optional<String> logbackXmlLocation;
+
     @Override
     public void setLogLevel(String name, LogLevel level) {
         getLoggerContext().getLogger(name).setLevel(toLevel(level));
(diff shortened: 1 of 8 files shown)


graemerocher:
I believe this will be fixed by https://github.com/micronaut-projects/micronaut-core/pull/8235 which moves logback to runtime intiailization

driverpt:
@graemerocher , with this modification you can keep build time initialization, it just resets all internal logback caches when Micronaut Inits.

The issue is that Logback is initialized when JVM Starts. The `LoggerContext#reset` only resets it's internal caches, so we can take advantage of Logback being previously initialized at build time. Won't improve much in terms of performance, but will add something

graemerocher:
This seems safe enough to target 3.8.x branch, can you do that?

driverpt:
Done. Can you unblock the CI? I suspect that the test will fail

graemerocher:
not sure why it is blocked 🤔

graemerocher:
@n0tl3ss this might help with the logging issue for AWS/OCI

driverpt:
Is there any ETA on this?

graemerocher:
waiting for feedback from @n0tl3ss

driverpt:
@n0tl3ss , can you help out here? This is currently the only blocker we have to go live with Micronaut

graemerocher:
@driverpt from the looks of the change it seems like you could wire this into your existing application without a new version of Micronaut necessarily blocking you?

n0tl3ss:
Hi, @driverpt. I created copy of this pull request: https://github.com/micronaut-projects/micronaut-core/pull/8263.

driverpt:
@n0tl3ss , fixed checkstyle issue

n0tl3ss (approved):

n0tl3ss:
> @n0tl3ss this might help with the logging issue for AWS/OCI

I tried but without success

graemerocher:
please fix the broken tests

driverpt:
I will, but at the end of the day

driverpt:
@graemerocher @n0tl3ss , GitHub Workflow needs a maintainers approval to run

runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
```diff
@@ -35,11 +42,31 @@
...
 
+    @Override
+    public void refresh() {
+        getLoggerContext().reset();
+        JoranConfigurator joranConfigurator = new JoranConfigurator();
+        joranConfigurator.setContext(getLoggerContext());
+        String logbackXml = logbackXmlLocation.orElse(DEFAULT_LOGBACK_LOCATION);
+
+        try {
+            URL resource = getClass().getClassLoader().getResource(logbackXml);
```
driverpt: @graemerocher  @n0tl3ss , please review this part. I need your opinion on this one

n0tl3ss: I think something like
```
    @Override
    public void refresh() {
        getLoggerContext().reset();
        try {
            new ContextInitializer(getLoggerContext()).autoConfig();
        } catch (JoranException e) {
            throw new LoggingSystemException("Error while refreshing Logback", e);
        }
    }
```

 would do a trick.

n0tl3ss (changes requested):
There is `io.micronaut.management.endpoint.loggers.impl.LogbackLoggingSystem` class that might replace `io.micronaut.logging.impl.LogbackLoggingSystem`. Should it also override refresh() method?

driverpt:
> There is `io.micronaut.management.endpoint.loggers.impl.LogbackLoggingSystem` class that might replace `io.micronaut.logging.impl.LogbackLoggingSystem`. Should it also override refresh() method?

I'm not a fan of copy + paste the same code in 2 classes. What do you suggest?

driverpt:
@n0tl3ss , were you able to get this "Fix" running smoothly?

I've been testing this and it seems like after I call `loggerContext.reset()` that it goes completely silent.

driverpt:
@n0tl3ss , can you please test if the logging works in AWS/OCI ?

Thanks

driverpt:
UPDATE: This seems to work in AWS Environment. But the downside is that we will need to add a `reflect-config.json` for all Logback components that we are using.

E.g.: In my case, the `logstash-logback-encoder`. But seems to be working with the simple Console Appender

driverpt:
@graemerocher @n0tl3ss , Checkstyle is failing due to NettyServer.java issue

graemerocher:
Thanks for the contribution!

PR #8261, opened 2022-10-31
Fail when the request is with the incorrect type

dstepanov:
I want to align the Netty HTTP server behavior with the Servlet project.

This:
```
String requestGeneric(HttpRequest<Foo> request) {
 request.getBody()...
}
```
Should fail on `request.getBody()` and return `BAD_REQUEST` when the body is something incompatible like an array, current behavior is an empty body.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -3,6 +3,7 @@ plugins {
 }
 
 repositories {
+    mavenCentral()
     gradlePluginPortal()
 }
 
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/JsonBodyBindingSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/JsonBodyBindingSpec.groovy
@@ -273,20 +273,22 @@ class JsonBodyBindingSpec extends AbstractMicronautSpec {
         )).blockFirst()
 
         then:
-        HttpClientResponseException ex = thrown()
-        ex.response.code() == HttpStatus.BAD_REQUEST.code
-        ex.response.getBody(Map).get()._embedded.errors[0].message.contains("Required argument [HttpRequest request] not specified")
+        response.code() == HttpStatus.OK.code
+        response.body() == 'not found'
     }
 
     void "test request generic type conversion error"() {
         when:
         String json = '[1,2,3]'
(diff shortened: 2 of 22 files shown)


buildSrc/build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
 }
 
 repositories {
+    mavenCentral()
```
dstepanov, line 6: Something happened with my local maven index, and this is needed now.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/JsonBodyBindingSpec.groovy
```diff
@@ -273,20 +273,22 @@ class JsonBodyBindingSpec extends AbstractMicronautSpec {
...
+        Flux.from(rxClient.exchange(
                 HttpRequest.POST('/json/request-generic', json), String
         )).blockFirst()
 
         then:
-        response.body() == "not found"
+        def e = thrown(HttpClientResponseException)
+        def response = e.response
+        response.getStatus() == HttpStatus.BAD_REQUEST
+        response.body().toString().contains("no int/Int-argument constructor/factory method to deserialize from Number value")
```
graemerocher, line 291: Where is this error message defined?

dstepanov: Cannot find anything, it might be Jackson as the bean is not introspected

graemerocher (approved):

http/src/main/java/io/micronaut/http/HttpMessageWrapper.java
```diff
@@ -69,4 +70,9 @@ public <T> Optional<T> getBody(Class<T> type) {
     public <T> Optional<T> getBody(Argument<T> type) {
         return delegate.getBody(type);
     }
+
+    @Override
+    public <T> Optional<T> getBody(ArgumentConversionContext<T> conversionContext) {
```
sdelamo, line 75: missing `@NonNull` annotation in return type and in the parameter `conversionContext`

sdelamo (approved):

graemerocher (changes requested):
Can you address the test failures

dstepanov:
@graemerocher Finnaly passing

PR #8265, opened 2022-10-31
Mark HateoasErrorResponseProcessorReplacement as Secondary

timyates:
Currently in a Groovy project, if you pull in problem-json, you end up with two ErrorResponseProcessors.

One for Problem-json and the other for Hateoas.

Marking the replacement as Secondary (as the HateoasErrorResponseProcessor was) fixes this


sdelamo (approved):

sdelamo:
@timyates can we have a test?

timyates:
@sdelamo test added...  Checked it fails if I comment out `@Secondary`

PR #8274, opened 2022-11-03
Introduce `@Vetoed` to exclude particular classes, methods and fields from processing

dstepanov:
There is a use case in Micronaut Groovy function where methods annotated `@Internal` are expected to be excluded, it would be better to use this new `@Vetoed` for this case, and this is also needed for the CDI implementation.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
@@ -2184,15 +2184,11 @@ public abstract class AbstractAnnotationMetadataBuilder<T, A> {
             } else {
                 return defaultMetadata;
             }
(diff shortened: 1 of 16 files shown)


core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
```diff
@@ -2204,6 +2200,9 @@ public <A2 extends Annotation> AnnotationMetadata annotate(
      * @since 3.0.0
      */
     public AnnotationMetadata removeAnnotation(AnnotationMetadata annotationMetadata, String annotationType) {
+        if (annotationMetadata.isEmpty()) {
```
dstepanov, line 2203: This is unrelated but noticed it when trying it with ODI

graemerocher (approved):

PR #8275, opened 2022-11-03
Use micronaut-cache 4.0.0-SNAPSHOT

timyates:
As this is the version that works for Micronaut 4.0.0


timyates:
Already in the BOM project 👍

PR #8281, opened 2022-11-03
Remove the session module, and add micronaut-session

### LINKED ISSUE #8152: Extract session to its own repository
### Issue description

see: https://github.com/micronaut-projects/micronaut-core/tree/4.0.x/session

timyates:


timyates:
Can you point me to the Upgrade wiki/doc?  As this will mean anything that is using a session will need the dependency changing

timyates:
@graemerocher Do we want to keep `session` here with an `api` dependency on `micronaut-session`?  That's what was done with `runtime`...however the session module is outside of micronaut-core so I'm not sure if that's something we want to do?

src/main/docs/guide/httpServer/sessions.adoc
```diff
@@ -99,33 +99,33 @@ snippet::io.micronaut.docs.session.ShoppingControllerSpec[tags="add", indent=0]
 
 == Using @SessionValue
 
-Rather than explicitly injecting the api:session.Session[] into a controller method, you can instead use ann:session.annotation.SessionValue[]. For example:
+Rather than explicitly injecting the https://micronaut-projects.github.io/micronaut-session/snapshot/api/io/micronaut/session/Session.html[Session] into a controller method, you can instead use https://micronaut-projects.github.io/micronaut-session/snapshot/api/io/micronaut/session/annotation/SessionValue.html[@SessionValue]. For example:
```
sdelamo: define `https://micronaut-projects.github.io/micronaut-session/snapshot/api` in https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/gradle.properties and uses it in the asciidoc.

sdelamo (changes requested):
You have to document this breaking change in https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/src/main/docs/guide/appendix/breaks.adoc

sdelamo:
> Can you point me to the Upgrade wiki/doc? As this will mean anything that is using a session will need the dependency changing

breaking changes are documented always in `breaks.adoc`

timyates:
Done and added to `breaks.adoc`

Fixed a couple of incorrect dependency macro usages there as well

sdelamo (approved):

graemerocher (approved):

sdelamo (approved):

PR #8283, opened 2022-11-04
Correct how function method is recognized

dstepanov:
This fixes some of the problems with Micronaut Groovy function support


function-web/src/main/java/io/micronaut/function/web/AnnotatedFunctionRouteBuilder.java
```diff
@@ -169,13 +174,8 @@ public void process(BeanDefinition<?> beanDefinition, ExecutableMethod<?, ?> met
                         LOG.debug("Created Route to Function: {}", route);
                     }
 
-                    if (consumes != null) {
```
dstepanov, line 172: `consumes` and `produces` are never null, I have removed all the null checks and added length check

graemerocher (approved):

Issue #8291, opened 2022-11-05
Can't get DeclaredAnnotation with generics and rxjava (micronaut-openapi)


altro3:
### Expected Behavior

Test code:
```java
package test;

import io.micronaut.core.annotation.Introspected;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;
import io.reactivex.rxjava3.core.Flowable;

@Controller("/pets")
interface TestController<T extends Pet> {

    @Get("/flowable")
    Flowable<T> flowable();
}

@Introspected(accessKind = Introspected.AccessKind.FIELD)
class Pet {
    private PetType type;

    public void setType(PetType t) {
        type = t;
    }

    public PetType getType() {
        return type;
    }
}

enum PetType {
    DOG, CAT
}
```

Can't get annotation for class `Pet`. See PR #8292 with test. This test works fine with micronaut 3.7.3

### Version

4.0.0-SNAPSHOT

altro3:
@dstepanov Hi! Well, I was wrong in the last issue - I found another bug in 4.0.0

PR #8292, opened 2022-11-05
Test for #8291

altro3:


dstepanov:
In this case, the testing visitor collects the elements and will fail at the assertion because the compiler has finished processing.
Adding:
```
        element.getReturnType().getFirstTypeArgument().orElse(element.getReturnType()).getTargetAnnotationMetadata();
```
In `AllElementsVisitor#visitMethod` make the test pass.

altro3:
@dstepanov Yes, it works after I added this line. But is it bug? Right? And need to merge these changes?

But is but, if I understand correctly, this does not solve the problem in general. Let's look at a more complex version:
```java
@Controller("/pets")
interface TestController<T extends Pet, R extends Pet2> {

    @Get("/flowable")
    Flowable<T> flowable();
}

@Introspected(accessKind = Introspected.AccessKind.FIELD)
class Pet {
    private PetType type;

    public void setType(PetType t) {
        type = t;
    }

    public PetType getType() {
        return type;
    }
}

@Introspected(accessKind = Introspected.AccessKind.FIELD)
class Pet2 {
    private PetType type;

    public void setType(PetType t) {
        type = t;
    }

    public PetType getType() {
        return type;
    }
}

enum PetType {
    DOG, CAT
}
```
And write test liek this:
```groovy
        AllElementsVisitor.VISITED_CLASS_ELEMENTS.size() == 1
        def method = AllElementsVisitor.VISITED_METHOD_ELEMENTS[0]
        def type = method.returnType

        def types = type.getTypeArguments()
```
Now, the `type.getTypeArguments()` method will only return the first argument type - `T - test.Pet`, but what about second argument type?

dstepanov:
It's not a bug. I suspect that before, we would resolve everything, but now it's lazy. And Java compiler cannot be used to process its elements after it's done, so the trick is to preload everything in the visitor if you want to validate it later.

Maybe we can try to add some testing visitor that would preload everything that it will find.

altro3:
Hmm... maybe I should do that. Because I can now state the fact that micronaut-openapi does not work with the current version of micronaut. But I can't figure out what I can do to make it work again.

Here I brought one of the tests that stopped working, but, as I understand it, there is no option now to make this test work in openapi, because This is a feature of the new version of Micronaut. Right?

dstepanov:
What is your use case in Open API? This problem should only affect the code that is testing the elements API.

altro3:
@dstepanov 

Look to my test classes structure:
```java
@Controller("/pets")
interface PetOperations<T extends Pet> {

    /**
     * List the pets
     *
     * @return a list of pet names
     */
    @Get("/flowable")
    Flowable<T> flowable();
}

@Schema(name="MyPet", description="Pet description", requiredProperties={"type"})
class Pet {
    private PetType type;

    public void setType(PetType t) {
        type = t;
    }

    public PetType getType() {
        return type;
    }
}

enum PetType {
    DOG, CAT
}
```

Problem with `micronaut 4.0.0` - I can't read `@Schema` annotation for class Pet. But I can do it with `micronaut 3.7.3`.

To try test: checkout this branch: https://github.com/altro3/micronaut-openapi/tree/jdk17 And run test `io.micronaut.openapi.visitor.OpenApiPojoControllerSpec#"test for generics type attributes schema"`. Next you can set debug checkpoint on class AbstractOpenApiVisitor line 786. And you will see this:
изображение
As you see:
1. type == "io.reactivex.rxjava3.core.Flowable" - this is right
2. type.getFirstTypeArgument().orElse(null) == "test.Pet" - this is right too
3. type.getFirstTypeArgument().orElse(null).getAnnotationNames() == emptyList - this is bad. Must be list with annotation `@Schema`

dstepanov:
So it looks like:
- `element.getGenericReturnType().getFirstTypeArgument().get()` will return `JavaGenericPlaceholderElement` without type annotations
- `element.getReturnType().getFirstTypeArgument().get()` will return `JavaClassElement` with the annotations

@graemerocher @yawkat Is it expected?

@altro3 Can you please check what it would return in Micronaut 3?

altro3:
@dstepanov Yes, sure. The same test on master branch:
изображение

As you see, difference only in `type.getFirstTypeArgument().orElse(null).getAnnotationNames()` - returns set with one element `io.swagger.v3.oas.annotations.media.Schema`

dstepanov:
Can you check in the debugger a few frames before in `addResponseContent` what types `element.getReturnType()` vs  `element.getGenericReturnType()` returns?

altro3:
@dstepanov from debugger (call `element.getGenericReturnType()`)

Micronaut 4:
изображение

Micronaut 3.7:
изображение

dstepanov:
Thanks, I should have written that I'm interested i the `.getFirstTypeArgument().get()` value

altro3:
@dstepanov  `element.getGenericReturnType().getFirstTypeArgument().get()` where element is a MethodElement `flowable`

Micronaut 4.0:

изображение

Micronaut 3.7:

изображение

yawkat:
@dstepanov sounds like this jdk bug i found a while back: https://github.com/yawkat/issue-reproducer-type-use-bytecode

yawkat:
@dstepanov imo the annotations on the class should not be included when looking at the specialized generic placeholder, but i dont know if we have code where we depend on it behaving either way...

altro3:
Fixed in #8331

dstepanov:
@yawkat Now there are included but I guess we can add an option to read just the generic argument annotation

dstepanov:
@yawkat Now there are included but I guess we can add an option to read just the generic argument annotation

PR #8299, opened 2022-11-07
Rework BOM as core BOM

melix:
This PR removes the traditional BOM and replaces it with a `micronaut-core-bom` artifact. This core BOM is like any other Micronaut project BOM, and only refers to libraries used by core.

This is an effort to extract the "main" BOM into its own project: https://github.com/micronaut-projects/micronaut-bom

As such, this is a major breaking change, as any project depending on the BOM is likely to fail compilation, but it's a pre-requisite for using the new BOM.

Some implementation comments:

- there are dependencies which could be imported from the new BOM, but are _duplicated_ in the `libs.versions.toml` file, because `core` _cannot_ depend on the BOM (otherwise we'd have a circular dependency between 2 projects)
- the new BOM project is _not_ ready yet
- the internal build plugin has an option for projects to depend on the BOM. Once this is merged, the build will fail, because there will no longer be a BOM published at `io.micronaut:micronaut-bom`. Therefore, the internal build plugins will need to be updated so that _by default_, using the BOM means depending on the _core BOM_ (they shouldn't depend on the "main" BOM)
- the user plugins will also need to be updated, so that they use the new GAV coordinates for the new BOM (once it's ready), or we'll have to publish it at both the new and old coordinates


graemerocher:
We should probably wait for the new BOM project and Gradle plugin PRs and then merge everything at once when it is all ready otherwise there will be downstream disruption for modules probably

core-bom/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+    maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/" }
+}
+
+micronautBom {
+    extraExcludedProjects = [
+            "benchmarks",
+            "inject-test-utils",
+            "inject-java-test",
+            "inject-kotlin-test",
+            "inject-groovy-test"
```
graemerocher: These were actually in the previous BOM

graemerocher (changes requested):

core-bom/build.gradle
melix: They were indeed, but I thought those were internal so a mistake. Isn't it the case?

graemerocher: there are modules that are using the BOM and referring to these without specifying a version. I think it would break. Could you fix the conflicts?

melix:
Duh, I will fix the conflicts later. It's super hard to keep track of all the changes while making changes to the new bom in parallel :(

core-bom/build.gradle
```diff
@@ -0,0 +1,29 @@
+plugins {
+    id 'io.micronaut.build.internal.bom'
+}
+
+group projectGroupId
+version projectVersion
+
+repositories {
+    mavenCentral()
+    maven { url "https://s01.oss.sonatype.org/content/repositories/snapshots/" }
```
graemerocher: can we move this to dependencyManagement block?

melix: Actually I think that's redundant now

graemerocher (approved):

melix:
Comments from the team meeting:

- PR should be merged (no sooner than Friday though)
- then the PR on `micronaut-bom` should be merged (it can be merged today but it would produce an invalid BOM since it depends on this one)
- BOM will _not_ be published at new coordinates
- parent project should also be moved to a the `micronaut-bom` project and published at the same GAV coordinates as today

For projects which import the Micronaut BOM today in order to use the `mn` catalog, this will have to be changed because:
    - there will no longer be a catalog published at `io.micronaut:micronaut-bom`
    - if they do so, they will need to use the `core` catalog : `io.micronaut:micronaut-core-bom`
    - if they need dependencies from another Micronaut project, they will need to import a catalog from that specific project, not the new platform BOM (otherwise we reintroduce a cyclic dependency)

I will work on conveniences in `micronaut-build` to import the catalogs on the new coordinates.

PR #8302, opened 2022-11-08
Property type: select the type with the most generic argument annotations

### LINKED ISSUE #8301: Type parameter annotations lost when building a ClassElement
## Problem

When a property is defined, annotations can be added to the property or getter (as it says in the [validating data classes](https://docs.micronaut.io/latest/guide/#_validating_data_classes) section of documentation). This works for annotations on properties like `@Valid Pojo`. However, the annotation may be lost if it is on a type argument, like `List<@Valid Pojo>` unless it is defined on a setter or setter is not present.

## Expected behavior

In case annotation is defined on a property or its type argument, it should be retained. It seems like current implementation checks if property, getter or setter have any annotations and chooses the one that has among them. However, the implementation doesn't check if type arguments have annotations. Therefore, if a property has type argument annotations, but a setter is present, the annotations are lost (which might be because the type of setter is used instead).

## Example

Given the following code:
```java
@Introspected
class SaladWithSetter {
    List<@Valid Ingredient> ingredients;

    public List<Ingredient> getIngredients() {
        return ingredients;
    }

    public void setIngredients(List<Ingredient> ingredients) {
        this.ingredients = ingredients;
    }

    @Introspected
    public record Ingredient(
        @NotBlank String name
    ) {}
}
```

Building the classElement and checking if the property type argument has the required annotation returns false:
```java
ClassElement classElement = buildClassElement(code);
PropertyElement propertyElement = classElement.getBeanProperties().iterator().next();
ClassElement propertyTypeArgument = propertyElement.getType().getTypeArguments().get("E");
propertyTypeArgument.getAnnotationMetadata().hasStereotype("javax.validation.Valid"); // --> false
```

* Note: if the setter is removed, the annotation is found. If the annotation is added to the setter, the annotation is found, too.
* Note: if we define the annotation on a getter, but not on the property, the annotation is lost, too.

### Recreating

See [this ClassElementAnnotationsRetaining test spec](https://github.com/andriy-dmytruk/micronaut-core/blob/type-argument-annotations-lost/inject-java/src/test/groovy/io/micronaut/ClassElementAnnotationsRetaining.groovy) for tests. See [gradle enterprise run](https://ge.micronaut.io/s/kfh6wbot643uk/tests/overview) for details on which tests fail and succeed.

## Possible solutions

* Choose the property/getter/setter that has type annotations if only one has. Otherwise, create an order in which these are chosen and document it (e.g. getter -> setter -> property).
* Document the fact that type argument annotations must be on setter if one present, otherwise on property. If other annotations are present, then type argument annotations must be in the same place.

@graemerocher @dstepanov

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/8301

@andriy-dmytruk I think it might be useful for you to have read and write type exposed from the property element instead of one. WDYT? The fix is to select the type with the most type annotations, which I'm not sure is the best for the real word scenario.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
@@ -150,9 +150,15 @@ public final class AstBeanPropertiesUtils {
                 // and it has more type arguments annotations - use it as the property type
                 if (value.field != null
                     && value.field.getType().equals(value.type)
-                    && hasGenericTypeAnnotations(value.field.getType())
-                    && !hasGenericTypeAnnotations(value.type.getType())) {
+                    && countGenericTypeAnnotations(value.field.getType()) > countGenericTypeAnnotations(value.type.getType())) {
                     value.type = value.field.getGenericType();
+                }
+                // In a case when the getter's type is the same as the selected property type,
+                // and it has more type arguments annotations - use it as the property type
+                if (value.getter != null
+                    && value.getter.getGenericReturnType().equals(value.type)
+                    && countGenericTypeAnnotations(value.getter.getGenericReturnType()) > countGenericTypeAnnotations(value.type.getType())) {
+                    value.type = value.getter.getGenericReturnType();
                 }
                 if (value.readAccessKind != null || value.writeAccessKind != null) {
                     value.isExcluded = shouldExclude(includes, excludes, propertyName)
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

andriy-dmytruk:
Hi @dstepanov, thanks a lot for the quick fix for the issue. I think this solution makes sense to me.

I am not sure if it is better to have 3 types exposed for the field, getter and setter. If type parameter annotations will be used somewhere else in micronaut, it would make sense to have a consistent behavior for retaining annotations in this case. Also, I wouldn't want to look through the 3 types during runtime.

I think that counting should make sense, as user should either define all the annotations in one place or repeat everywhere. In my opinion, it doesn't make sense to support defining some annotations in a getter and then others in a setter. 

I have one minor question. I didn't see the code looking at annotations on the property itself. Is that handled somewhere else? Also, it would be best if the search for annotations would go into deeper type parameters if some are present, so something like:
```java
    private static int countGenericTypeAnnotations(ClassElement cl) {
        return cl.getAnnotationMetadata().getAnnotationNames().size() +
            cl.getTypeArguments().values().stream()
                .filter(t -> !t.isGenericPlaceholder()) // I think this will avoid infinite loop in some cases
                .mapToInt(AstBeanPropertiesUtils::countGenericTypeAnnotations)
                .sum();
    }
```
Do you think we can change it?

PR #8305, opened 2022-11-08
Implement support for injection of maps

graemerocher:
This PR adds support for injecting of maps. It attempts to be backwards compatible by first looking up a bean that implements map and then falling back to map resolution. 

An error occurs if multiple beans exist that do not define a qualifier.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -71,6 +71,7 @@ import io.micronaut.core.convert.value.MutableConvertibleValues;
 import io.micronaut.core.io.ResourceLoader;
 import io.micronaut.core.io.scan.ClassPathResourceLoader;
 import io.micronaut.core.io.service.SoftServiceLoader;
+import io.micronaut.core.naming.NameResolver;
 import io.micronaut.core.naming.Named;
 import io.micronaut.core.order.OrderUtil;
 import io.micronaut.core.order.Ordered;
@@ -926,6 +927,11 @@ public class DefaultBeanContext implements InitializableBeanContext {
         return streamOfType(null, beanType, qualifier);
     }
 
+    @Override
+    public <V> Map<String, V> mapOfType(Argument<V> beanType, Qualifier<V> qualifier) {
+        return mapOfType(null, beanType, qualifier);
+    }
+
     /**
      * Obtains a stream of beans of the given type and qualifier.
      *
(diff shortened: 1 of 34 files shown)


dstepanov:
So if you’re going to inject a map of something that is not named it will fail? What if only a few of impl are named?

graemerocher:
It works for annotation based qualifiers as well. In that case the name is the qualifier name

graemerocher:
If you have multiple unamed beans then they would be treated as multiple primaries in which case it would error. This is similar to a `NonUniqueBeanException`. Actually maybe we should throw that instead of `DependencyInjectionException`

dstepanov:
Would it behave correctly when one of the beans is primary?

graemerocher:
yes in that case the key will be `Primary`

dstepanov:
What would happen if you inject JDBC datasources as String/Datasource map? I think the default is the primary one.

graemerocher:
good catch, I added a test for this and implemented some changes https://github.com/micronaut-projects/micronaut-core/pull/8305/commits/c6c7f11d24c316e3f7a21ec641c9352c3235e21a

sdelamo (changes requested):
In addition to the docs you wrote, add a paragraph in the What's new section telling it is now possible to inject a map whose key is the qualifier's name.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -934,6 +940,58 @@ protected <T> Stream<T> streamOfType(BeanResolutionContext resolutionContext, Cl
...
+     * Obtains a map of beans of the given type and qualifier.
+     * @param resolutionContext  The resolution context
+     * @param beanType           The bean type
+     * @param qualifier          The qualifier
+     * @param <K>                The key type
+     * @param <V>                The bean type
+     * @return A map of beans, never {@code null}.
+     * @since 4.0.0
+     */
+    protected <V> @NonNull Map<String, V> mapOfType(@Nullable BeanResolutionContext resolutionContext, @NonNull Argument<V> beanType, @Nullable Qualifier<V> qualifier) {
```
dstepanov, line 958: So If I call this with `mapOfType(MyInterface.class, Qualifiers.ANY)` and:
```
@Singleton
class A implements MyInterface {
}
@Singleton
class B implements MyInterface {
}
@Singleton
class C implements MyInterface {
}
```
I will get only one item in the map with `primary` key?

dstepanov: Ok I see it should return `occurs for duplicate keys` because the stream collet wouldn't allow a conflict. It's still strange that this would be OK when only one implementation is defined but will fail with multiple. IMHO correct would be to fail even with one implementation that doesn't specify some criteria like not being named. 

`Multiple beans were found missing a qualifier resulting in duplicate keys` What kind of qualifier is it missing? 

It's confusing to me that in the Micronaut codebase there are two understandings of what qualifier means:
1. Bean context uses it as a filter
2. Bean definition uses is as a set of `@Qualifier`s defined

graemerocher:
@dstepanov I pushed a change that allows maps of beans without qualifiers by using the simple name of the class decaptilized as the bean name by default if no qualifier is present. This appears to be what Spring does as well.

This also matches up with how we qualify since if no qualifier is present we use the simple name of the type. See https://github.com/micronaut-projects/micronaut-core/blob/686b472bc4801887a115c5d06cb015301241875b/inject/src/main/java/io/micronaut/inject/qualifiers/NameQualifier.java#L89

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -986,10 +987,19 @@ protected <T> Stream<T> streamOfType(BeanResolutionContext resolutionContext, Cl
     @NonNull
     private static String resolveKey(BeanRegistration<?> reg) {
         BeanDefinition<?> definition = reg.beanDefinition;
+        BeanIdentifier identifier = reg.identifier;
         if (definition instanceof NameResolver resolver) {
-            return resolver.resolveName().orElse(reg.identifier.getName());
+            return resolver.resolveName().orElse(identifier.getName());
+        } else {
```
dstepanov, line 997: You might want to avoid redundant `else` here and below.

dstepanov (approved):

sdelamo (approved):

PR #8306, opened 2022-11-08
Improve support for configuration property nesting

graemerocher:
Currently injecting nested lists and maps of records or interfaces doesn't work and by default tries to bind the value. This resolves that by using dependency injection for configuration property entries that are beans by default.

Also improves the documentation generation to use a javadoc parser to parse only the relevant parts of the javadoc


dstepanov (approved):

sdelamo (approved):

core-processor/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     api project(":aop")
     api libs.asm.tree
     api libs.bundles.asm
+    implementation 'com.github.javaparser:javaparser-symbol-solver-core:3.24.7'
```
graemerocher: this should  be moved to the version catalog after @melix PR is merged

graemerocher:
Test failures are unrelated

core-processor/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     api project(":aop")
     api libs.asm.tree
     api libs.bundles.asm
+    api 'com.github.javaparser:javaparser-symbol-solver-core:3.24.7'
```
sdelamo, line 10: we are ok adding this extra dependency because core-processor will be a `compileOnly` dependency right?

sdelamo (approved):

core-processor/build.gradle
graemerocher: correct

dstepanov: I think it should be in libs.versions.toml

graemerocher: it should be but as per my comment I didn't want to conflict with @melix's PR so left it like that so we can fix it later

PR #8307, opened 2022-11-08
Add support for Jakarta nulllable annotations. 

### LINKED ISSUE #8264: Use Nullable annotation from Jakarta EE to check the nullability of an element
### Context
As far as I could find, the nullability of an element is determined with the method [AnnotatedElement#isDeclareNullable](https://github.com/micronaut-projects/micronaut-core/blob/85ffc39fa7950dbce5ab5fb9eb0d430ac9ef8c5e/core/src/main/java/io/micronaut/core/annotation/AnnotatedElement.java#L38) (I reached this method after an exception while deserializing an object with a nullable property).

The check is based on the presence of the stereotype `javax.annotation.Nullable` from JSR305 (defined in [AnnotationUtil#NULLABLE](https://github.com/micronaut-projects/micronaut-core/blob/85ffc39fa7950dbce5ab5fb9eb0d430ac9ef8c5e/core/src/main/java/io/micronaut/core/annotation/AnnotationUtil.java#L104)).

### Request
Since Jakarate EE is already used in a Micronaut project (through `@Singleton` for instance), it seems to me that would be natural to include also a test on the presence of the stereotype "jakarta.annotation.Nullable" from Jakarta EE when checking the nullability of an element.

perococco:
Add a `NamedAnnotationTransformer` to transform `Nullable` annotation from Jakarta.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/internal/JakartaNonnullTransformer.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/internal/JakartaNonnullTransformer.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


graemerocher:
Needs also the version for `NonNull` like https://github.com/micronaut-projects/micronaut-core/blob/85ffc39fa7950dbce5ab5fb9eb0d430ac9ef8c5e/core-processor/src/main/java/io/micronaut/inject/annotation/internal/KotlinNotNullMapper.java

graemerocher:
Thanks for the contribution!

PR #8316, opened 2022-11-09
Improve the error message for missing beans when EachBean is used

graemerocher:
When a bean annotated with `@EachProperty` is not found due to missing configuration the error message is not clear as to why the bean is missing. This PR improves the error handling to provide extended diagnosis of configuration errors.


sdelamo (approved):
I think this PR is going to save some developer time. 👏

Could you add something similar to what you wrote in the PR description in the  `What's new in Micronaut 4 section` of the docs.

dstepanov:
`One or more configuration entries under the prefix` This is a bit confusing to me why one or more? IMHO better statement would be something like `No configuration entries found under the prefix:`

graemerocher:
@dstepanov good call, I improved the error message.

dstepanov (approved):

PR #8319, opened 2022-11-10
Correct configuration builder and recognition of the property's field

dstepanov:


graemerocher:
tests failing?

dstepanov:
Fixed tests

PR #8326, opened 2022-11-11
Improve error messages for disabled beans

graemerocher:
When a bean is disabled a `NoSuchBeanException` is thrown however the error doesn't provide diagnosis on why the bean was not found. This PR tracks disabled beans and configuration packages and propagates the failures in the error message so you don't have to enable trace logging with `io.micronaut.context.condition` at the cost of slightly increased memory usage to store these errors. 

This additional storage is probably not a big deal however since Micronaut AOT can optimize away the disabled beans during compilation for production deployment.

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
+++ b/inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
@@ -31,6 +31,7 @@ import java.util.List;
  * An interface for a {@link BeanType} that allows qualifiers.
  *
  * @param <T> The bean type
+ * @since 4.0.0
  */
 public interface QualifiedBeanType<T> extends BeanType<T>, AnnotationMetadataDelegate {
 
--- a/inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
+++ b/inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


dstepanov (approved):

inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+
+/**
+ * An interface for a {@link BeanType} that allows qualifiers.
+ *
+ * @param <T> The bean type
+ */
+public interface QualifiedBeanType<T> extends BeanType<T>, AnnotationMetadataDelegate {
```
dstepanov, line 36: Missing since

inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
```diff
@@ -0,0 +1,82 @@
...
+        List<AnnotationValue<Annotation>> annotations = annotationMetadata.getAnnotationValuesByStereotype(AnnotationUtil.QUALIFIER);
+        if (!annotations.isEmpty()) {
+            if (annotations.size() == 1) {
+                final AnnotationValue<Annotation> annotationValue = annotations.iterator().next();
+                if (annotationValue.getAnnotationName().equals(Qualifier.PRIMARY)) {
+                    // primary is the same as null
+                    return null;
+                }
+                return (Qualifier<T>) Qualifiers.byAnnotation(annotationMetadata, annotationValue);
+            } else {
```
dstepanov, line 59: Sonar is going to complain about redundant `else`s.

dstepanov (changes requested):

inject/src/main/java/io/micronaut/inject/QualifiedBeanType.java
graemerocher: already added to the method to ignore this, this is the same code just moved from the `BeanDefinition` interface to here I haven't changed anything in this code.

dstepanov (approved):

sdelamo (approved):

PR #8336, opened 2022-11-14
Mask headers that match know credentials

timyates:
When trace logging is enabled, all headers are logged.

We should avoid this, as it may be captured by logging or build software.

We could make this configurable, if we think it's worth the extra effort.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -184,7 +184,6 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static io.micronaut.scheduling.instrument.InvocationInstrumenter.NOOP;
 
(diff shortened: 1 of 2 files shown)


yawkat:
this could be one regex, it might be faster

timyates:
@yawkat Good idea, done in https://github.com/micronaut-projects/micronaut-core/pull/8336/commits/bfb8960301be7c50d4073a5caf5bddd224177776

yawkat (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1767,17 +1772,27 @@ private void traceChunk(ByteBuf content) {
...
                 }
             } else if (!all.isEmpty()) {
-                log.trace("{}: {}", name, all.get(0));
+                String maskedValue = isMasked ? mask(all.get(0)) : all.get(0);
+                log.trace("{}: {}", name, maskedValue);
             }
         }
     }
 
+    private String mask(String value) {
```
sdelamo:
```suggestion
    @NonNull
    private String mask(@Nullable String value) {
```

sdelamo (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1767,17 +1772,27 @@ private void traceChunk(ByteBuf content) {
...
+        if (value == null) {
+            return null;
+        }
+        return "*MASKED* [" + value.length() + " chars]";
```
alvarosanchez: Subtle thing: I would not return the length. The fewer clues for attackers, the better :)

alvarosanchez (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
timyates: I couldn't decide...  Then my pet project, I couldn't tell if the value was being injected into an auth header...

Showing the length let me see it was `Bearer ....some long token...` and not `Bearer {github.token}`

I agree though, as this feels convenient, and convenient things are usually security holes

I'll push a commit to remove it

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
timyates: Done...  They're both nullable https://github.com/micronaut-projects/micronaut-core/pull/8336/commits/92b9c7208b11c1cef4388d66ff9fe7d1d72a5364

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
timyates: Done https://github.com/micronaut-projects/micronaut-core/pull/8336/commits/92b9c7208b11c1cef4388d66ff9fe7d1d72a5364

Issue #8342, opened 2022-11-15
Micronaut's ServerRequestContext may get corrupted


stepanv:
Hi team,

we stumbled on an intermittent bug that `ServerRequestContext` may get corrupted. By corrupted I mean that it is still available but not in the expected state (e.g., the object I store there is later found missing).
 
It happens only under given conditions (see also a reproducer bellow):
1. `@Controller` has a method returning a `Mono<HttpResponse<Flux<?>>>`
2. A thread switch (e.g., due to an async call) occurs during the execution of the enclosing `Mono`
3. A thread switch (e.g., due to an async call) occurs in the execution of the inner `Flux`
4. A higher number of parallel request is being served

Why do we do the 2 layer reactive types controller?
1. We need the enclosing `Mono` to reactively decide (because it depends on a result from another backend) whether we want to return `200` and continue processing the request OR to return `4xy` error.
2. When we return `200` we still need to reactively return a series of objects, some get returned sooner other later (e.g., after another backend call)
3. After all, it seems to be supported by Micronaut as it works under a lower load :) 

### Steps to Reproduce

Following `@MicronautTest` annotated test demonstrates the issue:
```
import io.micronaut.http.HttpRequest;
import io.micronaut.http.HttpResponse;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Get;
import io.micronaut.http.client.HttpClient;
import io.micronaut.http.client.StreamingHttpClient;
import io.micronaut.http.client.annotation.Client;
import io.micronaut.runtime.http.scope.RequestScope;
import io.micronaut.test.extensions.junit5.annotation.MicronautTest;
import jakarta.inject.Inject;
import lombok.Data;
import lombok.NonNull;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.ValueSource;
import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;
import reactor.core.scheduler.Schedulers;
import reactor.test.StepVerifier;

import java.util.concurrent.Executors;

@MicronautTest
class DisappearingContextTest {

    @RequestScope
    @Data
    static class ContextBean {
        private volatile Boolean flag;
    }

    @Controller("/reproducer")
    static class EntryController {

        @Inject
        @Client("/control")
        HttpClient client;

        @Inject
        private ContextBean contextBean;

        @Get("/mono-http-response-flux")
[...]

PR #8343, opened 2022-11-16
A reproducer for issue #8342 - Corrupted ServerRequestContext when multi-threading

stepanv:
…


graemerocher:
Thanks will take a look, although in general I will say that `ServerRequestContext` can't be used for reactive flows since it uses thread local which is incompatible with reactive

stepanv:
@graemerocher I can change it to use the `@RequestScope`. 
I thought it was using `ServerRequestContext` underneath which is using ThreadLocal.

Looking in the guide https://docs.micronaut.io/latest/guide/#requestResponse , now I see it says to use the `Mono` / `Flux` context and do 
```
HttpRequest<?> request = ctx.get(ServerRequestContext.KEY);
```
That's is a surprise to me :) thx

yawkat:
yea this test case with `@RequestScope` has the same problem as `ServerRequestContext`. We can't guarantee that the context will always propagate

stepanv:
@yawkat , I thought so :) 
Not sure what you guys think, but I see following options:

1. remove `@RequestScope` and `ServerRequestContext.currentRequest()` from the Micronaut framework
2. somehow detect these concepts are used at a wrong place and fail (during compilation or runtime)
3. make it work everywhere
4. document it and at least warn users to not use it in bold and red with lots of exclamation marks, include such warning even in javadoc
5. do nothing

I understand (5) is the easiest :) but that would be sad :(

cc @graemerocher

PR #8350, opened 2022-11-16
Groovy properties shouldn't become executable methods

graemerocher:
Micronaut 4 is treating Groovy properties as executable methods when the executable annotation is on the type level.

The Kafka module is failing on upgrade to 4.0.x because of this bug.


inject-groovy/src/test/groovy/io/micronaut/aop/introduction/with_around/IntroductionWithAroundOnConcreteClassSpec.groovy
```diff
@@ -78,8 +78,8 @@ class Test {}
         def proxyTargetBeanDefinition = applicationContext.getProxyTargetBeanDefinition(clazz, null)
         def beanDefinition = applicationContext.getBeanDefinition(clazz, null)
         then:
-        proxyTargetBeanDefinition.getExecutableMethods().size() == 5
```
dstepanov, line 81: Reverting to values before the refactoring

dstepanov (approved):

Issue #8198, opened 2022-10-20
Automatically configure a separate event loop for the client if only 1 CPU is detected


graemerocher:
### Feature description

A common deployment scenario is to deploy Micronaut to k8s with pods that have only a single CPU.

In this scenario if you have any blocking HTTP client calls in your application it deadlocks the event loop because the event loop size is calculated from the number of available CPUs.

To resolve this the user has to re-configure a separate event loop for the client to run operations blocking client operations on a separate event loop. This is non-obvious and users keep hitting this problem and wondering why they are getting read time out exceptions.

We should consider altering this for Micronaut 4.x to instead detect the CPU configuration and automatically configure a separate event loop for the HTTP client so users don't keep hitting this problem on K8s

graemerocher:
thoughts @yawkat @dstepanov @sdelamo ?

dstepanov:
It doesn't matter how many threads you have, the blocking client can block them all. IMHO I would switch back to `auto` thread selection. The use-case where we have imperative controller methods that need to be run on the event loop is, in my opinion, far less than the opposite.

graemerocher:
Perhaps defaulting to IO makes more sense

yawkat:
we could implement a simple detection for this and log a warning (or even error, if we can be sure the event loop we're waiting on is the same that we're blocking).

graemerocher:
Fail fast would definitely be better than the current behavior

sdelamo:
@graemerocher can we close this issue since #8354 is merged?

PR #8354, opened 2022-11-17
Detect deadlocks when using BlockingHttpClient on the event loop

yawkat:
Keep track of the thread that called BlockingHttpClient.exchange, and throw an exception when the client attempts to use the same thread for the connection. Before this patch, the blocking call would time out, now it will fail with a specific exception.

The event loop group usually has many threads, and often a blocking call will wait on *another* thread of that event loop group by chance. Such uses currently only lead to sporadic read timeouts at the moment. This implementation is conservative: It will not fail when the same event loop group is used, it will only fail with the exact same thread. So the sporadic timeouts will be replaced by sporadic deadlock exceptions.

Relates to #8198

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
@@ -30,16 +30,7 @@ import io.netty.channel.EventLoop;
  * @since 4.0.0
  */
 @Internal
-final class BlockHint {
-    private final Thread blockedThread;
-    @Nullable
-    private final BlockHint next;
-
-    private BlockHint(Thread blockedThread, @Nullable BlockHint next) {
-        this.blockedThread = blockedThread;
-        this.next = next;
-    }
-
+record BlockHint(Thread blockedThread, @Nullable BlockHint next) {
     public static BlockHint willBlockThisThread() {
         return new BlockHint(Thread.currentThread(), null);
     }
--- a/http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Information about what threads are blocked waiting for a request to complete. This is used to
+ * detect deadlocks when the user does a {@link io.micronaut.http.client.BlockingHttpClient} on the
+ * event loop.
+ *
+ * @author Jonas Konrad
+ * @since 4.0.0
+ */
+@Internal
+final class BlockHint {
```
graemerocher: can be a record

graemerocher (approved):

http-client/src/main/java/io/micronaut/http/client/netty/PoolResizer.java
```diff
@@ -35,7 +35,7 @@
  * {@link io.micronaut.http.client.HttpClientConfiguration.ConnectionPoolConfiguration}.
  * <p>
  * This class consists of various mutator methods (e.g. {@link #addPendingRequest}) that
- * may be called concurrently and in a reentrant fashion (e.g. inside {@link #openNewConnection()}).
+ * may be called concurrently and in a reentrant fashion (e.g. inside {@link #openNewConnection)}).
```
timyates:
```suggestion
 * may be called concurrently and in a reentrant fashion (e.g. inside {@link #openNewConnection}).
```

http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    @NonNull
+    static HttpClientException createException() {
+        return new HttpClientException(
+            "Failed to perform blocking request on the event loop because request execution " +
+                "would be dispatched on the same event loop. This would lead to a deadlock. " +
+                "Either configure the HTTP client to use a different event loop, or use the " +
+                "reactive HTTP client.");
```
graemerocher: add a link to the documentation section on how to do this

graemerocher (changes requested):

graemerocher:
this is a great improvement btw, well done!

We are really improving the error messages for 4.x which users will love

timyates (approved):
Do you think we need a test with 2 threads in the event-loop pool to check the exception isn't thrown? 🤔

graemerocher:
> Do you think we need a test with 2 threads in the event-loop pool to check the exception isn't thrown? 🤔

I think all of our other tests already do that

yawkat:
with 2 threads, the exception can but doesn't have to happen. it will happen if the second connection is scheduled on the same thread, but it won't if it's scheduled on the other threads.

graemerocher (approved):

PR #8360, opened 2022-11-18
build: bump up slf4j to 2.0.4 and logback to 1.4.4

graemerocher:


melix (approved):
I guess we'll need to double check what it means for native builds, as we need config for both logback and slf4j.

graemerocher:
@timyates has the smoke test for GraalVM been merged into 4.0.x?

timyates:
> @timyates has the smoke test for GraalVM been merged into 4.0.x?

It's still languishing on 3.8.x, we should merge up... ~~Should I do it via a PR?~~ Jonas is looking in to the conflicts 🤩

timyates:
`test-suite-graal` will be on 4.0.x once https://github.com/micronaut-projects/micronaut-core/pull/8363 passes and is accepted

PR #8361, opened 2022-11-18
Upgrade logback to 1.4.4

graemerocher:


melix (approved):
This is going to require a change in Micronaut AOT to support logback 1.4. I already have a branch for this.

graemerocher:
can't go into 3.8.x, seems it doesn't support Java 8

PR #8367, opened 2022-11-18
Still emit response if return type is Void

### LINKED ISSUE #8366: Missing client metrics for web metrics for request without response body
### Expected Behavior

I have a reactive REST service that has a POST endpoint with and empty body (it returns `Mono<Void>`). This is forwarded to an external REST service using a declarative `@Client`. I have enabled web metrics in my application.yml using 

```
  metrics:
    binders:
      web:
        enabled: true
```

I expect to see metrics for both server (http_server_requests_\*) and client (http_client_requests_\*).

### Actual Behaviour

The request is measured on the "server side", but the forwarding client part is not measured. This is only the case for requests that return no body, *and* if the HTTP status is 200. For other statuses, the request is measured.

I have debugged this down to [io.micronaut.http.client.netty.DefaultHttpClient](https://github.com/micronaut-projects/micronaut-core/blob/79c56c65b40ca361904f2a723041379b15c7f26f/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java#L1544) where an empty body is deliberately prevented from emitting.

I find this inconsistent at least (why is 404 measured, but 200 not) but actually consider it a bug.

### Steps To Reproduce

I expect this to be reproducible in any environment using a reactive declarative client to issue a request returning a `Mono<Void>` and enabling web metrics as shown above.

Use the repo below to reproduce. Issue a request to http://localhost:7890/broken
Then query http://localhost:7890/metrics/http.client.requests and you will get a 404.

If you send a request to http://localhost:7890/works and then http://localhost:7890/metrics/http.client.requests and you will get a result, but only for the /drive endpoint.

### Environment Information

Mac OS 13.0.1
JDK 17

### Example Application

https://github.com/jakobmagiera/webmetricsbug

### Version

3.6.4

yawkat:
Before this patch, HttpClient would remove the actual response from the response publisher if the expected body type is Void. I don't see a reason to do this, and it leads to HTTP filters not seeing the response as they should. 

However there is an outstanding bug that this uncovered: If there is a declarative client method that is supposed to return Publisher<X>, it instead always returns a Publisher<HttpResponse<X>> (the type returned from HttpClient.exchange). The code here does not convert properly when the `returnType` is `Publisher<Void>` and the result is another publisher: https://github.com/micronaut-projects/micronaut-core/blob/f81eceaba3ac8c550e56288779c686d6f50ea286/aop/src/main/java/io/micronaut/aop/internal/intercepted/PublisherInterceptedMethod.java#L112-L118 – this causes a class cast exception in `HttpGetSpec.test simple get with Publisher<Void> return`.

@dstepanov do you have any idea how to fix this?

Fixes #8366

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java
@@ -152,9 +152,15 @@ public interface HttpClient extends Closeable, LifeCycle<HttpClient> {
      * @param <E>      The error type
      * @return A {@link Publisher} that emits a result of the given type
      */
+    @SuppressWarnings("unchecked")
     default <I, O, E> Publisher<O> retrieve(@NonNull HttpRequest<I> request, @NonNull Argument<O> bodyType, @NonNull Argument<E> errorType) {
         // note: this default impl isn't used by us anymore, it's overridden by DefaultHttpClient
-        return Flux.from(exchange(request, bodyType, errorType)).map(response -> {
+        Flux<HttpResponse<O>> exchange = Flux.from(exchange(request, bodyType, errorType));
+        if (bodyType.getType() == void.class) {
+            // exchange() returns a HttpResponse<Void>, we can't map the Void body properly, so just drop it and complete
+            return (Publisher<O>) exchange.ignoreElements();
+        }
+        return exchange.map(response -> {
             if (bodyType.getType() == HttpStatus.class) {
                 return (O) response.getStatus();
             } else {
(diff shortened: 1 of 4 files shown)


timyates (approved):
Added it to the 3.8.0 project, and kicked off the Java CI again

The Graal CI is failing as there's no Graal for java 11?

yawkat:
not sure it'll be ready for 3.8.0

dstepanov:
What do you mean by removing of the response? Do you mean thr body of the response?

timyates:
> not sure it'll be ready for 3.8.0

Cool, added it to 3.9.0 as well 😎

yawkat:
Nevermind, I've solved it differently.

Basically, HttpClientIntroductionAdvice has a special path for Void, `Publisher<Void>`, etc...: It calls .exchange (which returns a `Publisher<HttpResponse<O>>`) instead of .retrieve (which returns a `Publisher<O>`). Before this change, this worked fine, because for void bodyType, exchange would always return an empty publisher anyway, so the exact type of the publisher did not matter. Now that exchange correctly returns a publisher with a single `HttpResponse<Void>`, there was a cast exception because that HttpResponse can't be cast to Void and the conversion (`HttpResponse<O>` to O) did not run for the `Publisher<Void>` case.

I've updated HttpClientIntroductionAdvice to use .retrieve for Void as well, and I've altered retrieve to work with Void bodyType (a body that is `null` is not received well by the code).

PR #8369, opened 2022-11-18
Exclude maven copy of aop for testRuntime

timyates:
Previously, due to laziness I'd excluded it it from every configuration, including publishing 😳

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
@@ -16,7 +16,7 @@ repositories {
 }
 
 configurations {
-    all {
+    testRuntimeClasspath {
         // Use AOP from here, not from maven
         exclude group: 'io.micronaut', module: 'micronaut-aop'
     }


timyates:
We need this exclusion, or else we end up with `Multiple possible bean candidates found: [InterceptorRegistry, InterceptorRegistry]` as we have the version from maven central, and the local project version

melix:
Why do we get a version from Maven Central?

timyates:
I will investigate... I think it must come from micronaut-test 4.0.0

timyates:
@melix So, looking at this scan (which I got by removing the `configurations` block in this PR, then running `./gradlew check --dry-run`):

https://ge.micronaut.io/s/idjwomtioxkww/dependencies?dependencies=micronaut-aop&expandAll&focusedDependency=WzQzLDE0LDE1NzEsWzcsMSxbMTBdXV0&focusedDependencyView=versions

`micronaut-test-core` pulls in `micronaut-context` which pulls in `micronaut-aop`

We then end up with 2 different aop versions on the classpath (the project here in core, and this transitive version), and so we get the error

```
  Message: Multiple possible bean candidates found: [InterceptorRegistry, InterceptorRegistry]
```

Unless we exclude the transitive version.

I believe micronaut-test is being injected into the modules in that scan by the internal module plugin

PR #8385, opened 2022-11-21
Warn when bean definition class is excluded because of the error

dstepanov:


inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -247,6 +250,10 @@ public final boolean process(Set<? extends TypeElement> annotations, RoundEnviro
         processing round.
         */
         if (processingOver) {
+            for (Map.Entry<String, Element> e : postponed.entrySet()) {
+                javaVisitorContext.warn("Bean definition [" + e.getKey() + "] skipped from processing because of the error element reported by the compiler", e.getValue());
```
graemerocher: Reword to:

```
javaVisitorContext.warn("Bean definition generation [" + e.getKey() + "] skipped from processing because of prior error. This error is normally due to missing classes on the classpath. Verify the compilation classpath is correct to resolve the problem.", e.getValue());
```

graemerocher (changes requested):

PR #8388, opened 2022-11-21
Add logback as a managed dependency

timyates:
We need logback in the modules I believe https://github.com/micronaut-projects/micronaut-crac/pull/49/files


melix (changes requested):
LGTM, however this raises the following points:

- should be renamed to `logback-classic` to conform to what we have in the new BOM and align with what we do for other dependencies
- `logback-classic` should be removed from `micronaut-bom` now that it is instead managed by `core`
- should we do the same for `slf4j`?
- if we add managed dependencies like this in core, then should core add other logback/slf4j modules even if not used directly?

gradle.properties
```diff
@@ -1,4 +1,4 @@
-projectVersion=4.0.0-SNAPSHOT
+projectVersion=9.0.0-SNAPSHOT
```
melix: I think this is not intentional, right? :)

timyates: Whoopsie...  I was living in the future... 😊

timyates:
I agree with the naming questions.

logback is now `logback-classic`, and `slf4j` is now `slf4j-api` to follow our existing pattern

timyates:
For:

> logback-classic should be removed from micronaut-bom now that it is instead managed by core

Agreed

> should we do the same for slf4j?

Probably

> if we add managed dependencies like this in core, then should core add other logback/slf4j modules even if not used directly?

Not sure....

melix (approved):
Let's see what Graeme thinks about this too

graemerocher:
I thought dependencies like this were supposed to go into the new BOM project?

melix:
In theory they should. However, the problem is that all modules will probably need logback, and since they cannot depend on the new platform, they would have to duplicate the version, which was causing a conflict as @timyates found.

So either we move logback to the core BOM, or we live with the duplicating everywhere. Not sure what's best.

timyates:
There's no way we can use the platform BOM in the modules? 🤔

melix:
You can't, because the platform needs to reference the modules, which then would then need to reference the BOM. There's no way we'd be able to release a BOM otherwise.

You _could_ depend on a previous version of the platform, but that would be incorrect. Technically, we have a platform, which should aggregate a number of other platforms.

graemerocher:
@timyates can you resolve the conflicts

timyates:
@graemerocher done 👍

melix:
Didn't we say we'd create `micronaut-logging` yesterday?

timyates:
I raised https://github.com/micronaut-projects/micronaut-core/issues/8414 for discussion

PR #8391, opened 2022-11-21
Remove the validation module

andriy-dmytruk:
Removes the validation module from micronaut core in favor of a separate [micronaut validation project](https://github.com/micronaut-projects/micronaut-validation).

Creates a dependence on micronaut validation 3.7.0 for tests in certain modules. These will probably later need to be replaced to 4.0.0.


graemerocher:
@dstepanov can this be closed?

PR #8399, opened 2022-11-22
Bump GraalVM Native Build Tools to 0.9.18

alvarosanchez:
@melix I suppose this is not being picked up by Renovate because all it's doing is generating a `<maven.native.plugin.version>0.9.18</maven.native.plugin.version>` property in the BOM, which is then referenced by `micronaut-parent` (`<version>${maven.native.plugin.version}</version>`).

Is there a way we can improve this so that these versions are picked up by Renovate?


melix:
I think for it to be picked by Renovate, you'll have to add an entry in the `libraries` section with the module coordinates.

alvarosanchez:
Will that have any other consequence for Gradle users?

melix:
No it won't. It would add an entry to the BOM because it's a _managed_ dependency, but that's it.

sdelamo (approved):

PR #8417, opened 2022-11-23
Switch to using the jakarta validation api

timyates:
Replacing the javax.validation one

This currently just renames the packages, and pulls in the dependency


timyates:
Closed as duplicate (been done elsewhere)

PR #8420, opened 2022-11-24
Support records for `@RequestBean`

### LINKED ISSUE #8295: Java records cannot be used for request beans with @PathVariable
### Expected Behavior

I expect to be able to use Java records with `@RequestBean` and `@PathVariable` for API endpoints in the same way that I can declare a POJO.

### Actual Behaviour

Instead, I get the following compilation error:
```
java: The route declares a uri variable named [id], but no corresponding method argument is present
```

### Steps To Reproduce

1. Clone the provided repository
2. Run the Micronaut application
3. See the following compilation error: `java: The route declares a uri variable named [id], but no corresponding method argument is present`

### Environment Information

- Operating System: macOS Ventura 13.0
- JDK (I presume this would still occur on non-ea builds):  openjdk 20-ea 2023-03-21
OpenJDK Runtime Environment (build 20-ea+21-1545)
OpenJDK 64-Bit Server VM (build 20-ea+21-1545, mixed mode, sharing)

### Example Application

https://github.com/lbkulinski/record_bug_example

### Version

3.7.3

yawkat:
@dstepanov the MissingParameterRule patch is a bit weird, the properties returned by getBeanProperties don't include annotations. Should they?

Fixes #8295

### DIFF
--- a/http-validation/src/main/java/io/micronaut/validation/routes/rules/MissingParameterRule.java
+++ b/http-validation/src/main/java/io/micronaut/validation/routes/rules/MissingParameterRule.java
@@ -15,11 +15,13 @@
  */
 package io.micronaut.validation.routes.rules;
 
+import io.micronaut.core.annotation.AnnotatedElement;
 import io.micronaut.core.bind.annotation.Bindable;
+import io.micronaut.core.naming.Named;
 import io.micronaut.http.uri.UriMatchTemplate;
+import io.micronaut.inject.ast.ClassElement;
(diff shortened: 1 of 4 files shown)


dstepanov (approved):

dstepanov:
I’m not familiar with the behavioral before v4

yawkat:
@dstepanov it's the same with 4.0, AstBeanPropertiesUtils only considers field and method

PR #8421, opened 2022-11-24
8342 corrupted server request context when multithreading - fixed using reactor context

### LINKED ISSUE #8342: Micronaut's ServerRequestContext may get corrupted

stepanv:
This actually fixes #8342 for the given reproducer.


graemerocher:
Appears to cause failures

Issue #8179, opened 2022-10-17
Allow to define HTTP filters with custom methods similar to controller definition


dstepanov:
### Feature description

It would be nice to remove the requirement to always write reactive HTTP filters in Micronaut 4.

The requirements would be the ability to write a filter method using imperative, reactive (reactive-streams), Java 15 Flow, Kotlin coroutines, and completable future, just like it's possible to define a controller method.

The proposal to introduce two annotations:
1. `@RequestFilter` - method can accept a `HttpRequest` (possible mutable depending on Server/Client), and it can mutate its state or return a different instance. The method can return reactive type.

Alternative to writing:
```java
    @Override
    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
        // Mutate request
        return chain.proceed(request);
    }

    @Override
    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
        return Mono.from(fetcher).flatMapMany(result -> {
            // Mutate request in a reactive way
            return chain.proceed(request);
        });
    }
```

2. `@ResponseFilter`- method can accept an originating request, a `HttpResponse` (possible mutable depending on Server/Client), and it can mutate its state or return a different instance. The method can return reactive type.

Alternative to writing:
```java
    @Override
    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
        return Mono.from(chain.proceed(request)).map(response -> {
                // Mutate response
            return response;
        });
    }

    @Override
    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
        return Mono.from(chain.proceed(request)).flatMap(response -> {
            return Mono.from(fetcher).map(val -> {
                // Mutate response in a reactive way
                return response;
            });
        });
    }
```

Examples:

```java
@Filter(Filter.MATCH_ALL_PATTERN)
public class MyFilter {

    MyExternalService service = null;

    @Order(Ordered.LOWEST_PRECEDENCE) // Log after all filters modify the request
    @RequestFilter
    public void loggingRequestFilter(HttpRequest<?> request) {
        System.out.println("REQUEST: " + request);
    }

    @RequestFilter
    public void loggingOnlyGets(HttpRequest<?> request) {
        System.out.println("REQUEST: " + request);
    }

    @RequestFilter
[...]

PR #8422, opened 2022-11-25
Support annotation-based HTTP filter declarations

yawkat:
This PR replaces the filter processing logic with a central `FilterRunner` class, and adds support for annotation-based filter methods (similar to controllers).

Changes:

- Replace most uses of `HttpFilter` with a new sealed `GenericHttpFilter` interface. `GenericHttpFilter`s are opaque, they are only processed by the `FilterRunner`. A `GenericHttpFilter` can be a legacy filter, a new annotation-based filter method, or one of a few special internal types.
- Implement new annotation-based filter parsing in static methods in `FilterRunner`. It scans & validates filter arguments, assigns binders, validates the return value, etc, and finally stuffs the filter metadata into a record that extends `GenericHttpFilter`. This code is also used by a processor that validates filter methods at compile time.
- Implement new filter logic in `FilterRunner`. The filter runner has a coroutine-inspired approach and tries to process filters sequentially, avoiding deep reactive call stacks. Reactive code is avoided altogether where possible.


core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
```diff
@@ -80,6 +81,20 @@ public void onComplete(BiConsumer<? super Object, Throwable> fn) {
         });
     }
 
+    @Nullable
+    @Override
+    public ImperativeExecutionFlow<Object> asDone() {
```
dstepanov: Maybe it can be named something like tryComplete

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -1,64 +1,316 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 @Internal
 public class FilterRunner {
+    private static final Logger LOG = LoggerFactory.getLogger(FilterRunner.class);
+
     private final List<InternalFilter> filters;
 
+    private HttpRequest<?> request;
```
dstepanov: I think both request and response would need to be flows

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -1,64 +1,316 @@
...
+    private HttpResponse<?> response;
+    private Throwable failure;
+    private SuspensionPoint<HttpResponse<?>> responseSuspensionPoint = new SuspensionPoint<>(-1, null);
+    private int index;
+    private boolean responseNeedsProcessing = false;
+
+    private static final Comparator<InternalFilter> SORT = Comparator.comparingInt(f -> {
+        if (f instanceof InternalFilter.Before<?> before) {
```
dstepanov: I would extract the order method

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -1,64 +1,316 @@
...
+        ExecutionFlow<HttpResponse<?>> resultFlow = CompletableFutureExecutionFlow.just(responseSuspensionPoint);
+        workRequest();
+        return resultFlow;
+    }
+
+    @SuppressWarnings("StatementWithEmptyBody")
+    private void workRequest() {
+        while (true) {
+            InternalFilter filter = filters.get(index++);
+            if (filter instanceof InternalFilter.Before<?> before) {
```
dstepanov: Maybe this can be implemented using some visitor pattern to avoid instanceof

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -1,64 +1,316 @@
...
+            }
+        }
+    }
+
+    private <T> boolean invokeBefore(InternalFilter.Before<T> before) {
         // todo: better param binding, handle return value, errors
-        before.method().invoke(before.bean(), request);
-        return ExecutionFlow.just(request);
+        try {
+            before.method().invoke(before.bean(), request);
```
dstepanov: It would be interesting to see how is it going to look with return values.

http/src/main/java/io/micronaut/http/filter/InternalFilter.java
```diff
@@ -2,35 +2,41 @@
...
 import io.micronaut.http.MutableHttpResponse;
-import io.micronaut.inject.ExecutableMethod;
 import org.reactivestreams.Publisher;
 
-import java.util.function.Supplier;
+import java.util.function.Function;
 
 @Internal
 public sealed interface InternalFilter {
     record Before<T>(
```
dstepanov: Maybe you can call them Request/Response filters

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -1,64 +1,316 @@
...
+                if (!invokeAfter(after)) {
+                    // suspend
+                    return;
+                }
+            }
+        }
+    }
+
+    private <T> boolean invokeBefore(InternalFilter.Before<T> before) {
         // todo: better param binding, handle return value, errors
```
dstepanov: I think Before/After Request/Response filters can be implemented as an internal interface that returns ExecutionFlow Request/Response and will have all the logic of calling the method. It would be nice to have some binding parameters registry like we have for the controllers.

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
```diff
@@ -111,6 +112,13 @@ public void onComplete() {
         });
     }
 
+    @Nullable
+    @Override
+    public ImperativeExecutionFlow<Object> asDone() {
```
dstepanov: I wonder if we can also detect resolved Mono/Flux

dstepanov (changes requested):
Look good so far 👍 @yawkat

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: ive added support

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
yawkat: i don't think so, we can't subscribe without the potential for side effects, and i don't see any api to check specifically whether the mono is side-effect-free

yawkat:
I've added the feature I talked about in #8179, which allows "blocking calls" to `chain.proceed`:

```groovy
        @RequestFilter
        @ExecuteOn(TaskExecutors.BLOCKING)
        void request(HttpRequest<?> request, FilterContinuation<HttpResponse<?>> continuation) {
            events.add("request " + request.uri)
            def response = continuation.proceed(request)
            events.add("response " + request.uri + " " + response.status)
        }
```

this will be great with virtual threads.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: the fields are updated when the flows of each filter complete

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: i placed all that logic in FilterRunner because it allows for lots of flexibility, such as the FilterContinuation parameter I added today, while at the same time allowing very fast paths (not even lambda allocations) for simple filters

yawkat:
my current plan on what part of the filter infrastructure to expose is: `InternalFilter` (name tbd) will remain non-`@Internal` but opaque. Users can pass around filters how they want, but they will not have access to details, and can't call filters directly.

The idea is that users (mostly micronaut-servlet) get their filters from the `RouteBuilder`s, sort them using `FilterRunner.sort`, add their `InternalFilter.Terminal` (the only filter type that is exposed), and run the filters using `FilterRunner.run`. That way, we have flexibility for future changes in the structure of filters, different subtypes, etc.

http/src/main/java/io/micronaut/http/filter/InternalFilter.java
yawkat: This is a name collision with the new annotations... but more descriptive. since theyre internal only they dont have to be all that descriptive though

yawkat:
~~Small behavior change: For legacy filters, `chain.proceed()` will now run the downstream filters immediately, instead of waiting for a subscribe to the publisher returned by proceed(). This makes the code a little simpler, and I don't think it's a big deal.~~

nevermind, I need to implement the old behavior anyway for context propagation.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: actually it looks like it's possible with Fuseable.ScalarCallable

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
yawkat: ive looked in reactor and found one approach, i dont think it gets better unfortunately.

http/src/main/java/io/micronaut/http/annotation/RequestFilter.java
```diff
@@ -14,12 +20,42 @@
...
+ *
+ * The return value may be:
+ *
+ * <ul>
+ *     <li>{@code void} to immediately continue execution</li>
+ *     <li>An updated {@link HttpRequest}</li>
+ *     <li>A {@link HttpResponse} to skip execution of the request</li>
+ *     <li>A {@link Publisher} (or other reactive type) that produces any of these return types, to
+ *     delay further execution</li>
+ * </ul>
```
graemerocher, line 62: we need compilation time checking / failures for these restrictions

http/src/main/java/io/micronaut/http/annotation/ResponseFilter.java
```diff
@@ -14,12 +20,41 @@
...
+ * </ul>
+ *
+ * The return value may be:
+ *
+ * <ul>
+ *     <li>{@code void} to immediately continue execution</li>
+ *     <li>An updated {@link HttpResponse}</li>
+ *     <li>A {@link Publisher} (or other reactive type) that produces any of these return types, to
+ *     delay further execution</li>
+ * </ul>
```
graemerocher, line 61: We need compilation time checking for these restrictions

core/src/main/java/io/micronaut/core/execution/ExecutionFlow.java
```diff
@@ -143,12 +143,22 @@ static <T> ExecutionFlow<T> async(@NonNull Executor executor, @NonNull Supplier<
...
 
+    /**
+     * Create an {@link ImperativeExecutionFlow} from this execution flow, if possible. The flow
+     * will have its result immediately available.
+     *
+     * @return The imperative flow, or {@code null} if this flow is not complete or does not
+     * support this operation
+     */
+    @Nullable
+    ImperativeExecutionFlow<T> asComplete();
```
dstepanov: I would prefix it with `try` like `tryComplete`

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1240,39 +1244,41 @@ private <I, R extends io.micronaut.http.HttpResponse<?>> Publisher<R> applyFilte
...
-            } else {
-                try {
-                    responsePublisher = (Publisher<R>) filters.get(0).doFilter(request, filterChain);
-                } catch (Throwable t) {
-                    responsePublisher = Flux.error(t);
+        FilterRunner.sortReverse(filters);
+        filters.add(new GenericHttpFilter.TerminalReactive(responsePublisher));
+
+        FilterRunner runner = new FilterRunner(filters);
+        Mono<R> responseMono = Mono.deferContextual(ctx -> {
```
dstepanov, line 1273: We should avoid extra storing and propagating the Reactor context, it should be done automatically like before. There can be cases where we might want to detect if the route is already reactive and don't use a completable stage but some reactive one to preserve the context.

dstepanov: It would be nice to avoid anything Reactor/Reactive specific, like deferred... Essentially we might want to extract all the reactive code at some point.
If it's needed, we might need to introduce `ExecutionFlow#defered` and `ExecutionFlow#async` which will resolve to reactive/completablefuture based if the current flow is already reactive or not.

yawkat: since filters dont have to use reactive anymore, the only way to propagate the context between filters that do, is to store it in the filter runner

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlowImpl.java
```diff
@@ -80,6 +81,20 @@ public void onComplete(BiConsumer<? super Object, Throwable> fn) {
         });
     }
 
+    @Nullable
+    @Override
+    public ImperativeExecutionFlow<Object> asComplete() {
+        if (stage.isDone()) {
```
dstepanov, line 109: There is some kind of concept of the context, but it's not implemented for the completable future as there was no use-case for it. But the idea is to have some `ObjectWithContext` which can hold the context and pass it to the reactive if needed.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: i saw that you tried to do context propagation in ExecutionFlow, but imo the model for context propagation is limited. we should not have to wait for the caller to "subscribe" to the ExecutionFlow on the off chance that the caller will write context. The possibility to run the context-dependent code immediately should be left open (and in fact it's what's happening in FilterRunner right now).

This requires passing in the context *before* FilterRunner.run is called, not in some contextWrite that is done after

dstepanov: It is limited and might be something to be removed, but I don't see a valid use case where the Reactor context needs to be detached from the actual reactive flow and attached somewhere else.

yawkat: eg in the HttpClient

- user contextWrites to the publisher returned by client.exchange
- first there is a blocking filter, like this:

```java
@RequestFilter
@ExecuteOn(IO)
void filter(FilterContinuation<HttpResponse<?>> continuation, HttpRequest<?> request) {
    // ...
    continuation.proceed(request);
    // ...
}
```

- Then there is an old-style filter that reads the reactive context

Because there is no reactive publisher for the blocking filter, the only way to get context through it is to detach the context before the filter and reattach it afterwards.

dstepanov: Ehh not sure if I like this blocking stuff... In this case, the context should be somehow propagated in `FilterContinuation`, if it's even possible, I don't think Reactor/Reactive mix well with blocking operations.

http/src/main/java/io/micronaut/http/annotation/ResponseFilter.java
yawkat: @graemerocher we don't have a processor for the http module atm, should i create one for this?

yawkat: ive implemented a validator for this, but it would need a new module because if it went into core-processor there would be a circular dependency. im gonna stash this for now since graeme is on vacation

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: i just think that filters should be allowed not to use reactive. especially if we want first-class loom support in the long run. by putting the context in a field that is updated throughout the filter runner, the reactive context simply "skips over" filters that dont use it.

yawkat:
ok open questions on this pr:

- where best to do compile time checking of filter methods? i have a nice implementation that uses FilterRunner, but this introduces a circular dependency between :core-processor and :router.
- im not 100% happy with the `FilterContinuation` API. initially i had it take the new request as a required parameter, similar to FilterChain. ive changed it to be optional, using a separate method. but this still does not seem right to me, ideally the continuation does not care about the request at all, and there would be some independent way to replace the request. but i cant think of a good way. in doubt, keep it as-is (it's `@Experimental`), and look for a better way in the future.

graemerocher:
logical place would be in the same place https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-validation/src/main/java/io/micronaut/validation/routes/RouteValidationVisitor.java is

dstepanov:
We need more tests with request/response filters returning all kinds of variations reactive, optional, nullable responses of request/response (null means continue) + Kotlin coroutines tests. Plus variations with old reactive filters.
Docs would need to be added and previous filter examples updated.

yawkat:
oh i did not see the http-validation module, thanks :D that seems like a good place.

i will look at tests and docs. maybe tests should be in the tck, eventually.

dstepanov:
I don't think we need `FilterContinuation`, you would be able to implement same before/after with simple request/response filters by storing values in the request attributes to track the same request, I know you want to have old-style servlet filter API but IMHO we would be better without it, other frameworks don't support it either.

yawkat:
@dstepanov you mean simply dropping support for replacing the httprequest in filters (except legacy filters)? this could make a lot of other code more sensible too in the long run, but not sure if there is significant use of this feature...

dstepanov:
You would be able to replace the request in the new request filters just by returning it as non-null value.

yawkat:
oh i understand now, you mean simply not supporting "around"-style filters.

i think they are still a good feature to have, especially in the long term. with a full loom-style synchronous programming model, it will be nice to be able to use normal language concepts like try-with-resources in filters. sure, you can always manually introduce a "cut" in the filter and split it up into request/response, and this approach is basically required with reactive filters, but it doesn't have to be this way forever.

dstepanov:
Is it ready for review?

yawkat:
@dstepanov i think it's still missing some docs, but im not sure, i forgot during vacation :D I have to check the status once i have finished looking at notifications.

dstepanov:
I think more tests are needed to or maybe I don't see them :)

yawkat:
oh thats right, i was working on the tck tests...

yawkat:
ok, today the build passed :) This PR is ready now, I added PR description, docs, and (TCK) tests. The remaining sonar warnings are from the TCK.

dstepanov:
Please add tests for `CompletableFuture<HttpRequest/Response>`, `Optional<HttpRequest/Response>`, `@Nullable HttpRequest/Response` return types. I don't see any response filter tests with different return types. We should also support Kotlin suspend filters.

http/src/main/java/io/micronaut/http/filter/FilterContinuation.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+@Experimental
+public interface FilterContinuation<R> {
+    /**
+     * Update the request for downstream processing.
+     *
+     * @param request The new request
+     * @return This continuation, for call chaining
+     */
+    FilterContinuation<R> request(HttpRequest<?> request);
```
graemerocher: nullability annotations

http/src/main/java/io/micronaut/http/filter/FilterContinuation.java
```diff
@@ -0,0 +1,61 @@
...
+     * may block in the netty HTTP server should use
+     * {@link io.micronaut.scheduling.annotation.ExecuteOn} to avoid running on the event loop.</b>
+     * <br>
+     * If {@link R} is a reactive type, this method will return immediately. Downstream processing
+     * will happen when the reactive stream is subscribed to, and the reactive stream will produce
+     * the downstream result when available.
+     *
+     * @return The downstream result, or reactive stream wrapper thereof
+     */
+    R proceed();
```
graemerocher, line 63: nullability annotations

http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
```diff
@@ -1,29 +0,0 @@
...
-import io.micronaut.core.order.Ordered;
-
-/**
- * Describes a bean that contains an order to define the
- * order of a client or server filter.
- *
- * @author James Kleeh
- * @since 1.0
- */
-public interface FilterOrderProvider extends Ordered {
```
graemerocher, line 27: did this exist in the 3.x branch? Probably need to deprecate it there if so

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+ * an instance variable ({@link #index}), and the loop can sometimes <i>suspend</i>, e.g. when a
+ * filter returns a reactive flow instead of an immediate value. When suspension happens, the loop
+ * exits early, and {@link #workRequest()} (or {@link #workResponse()}) will be called again on
+ * unsuspend (e.g. when the reactive flow completes).
+ *
+ * @since 4.0.0
+ * @author Jonas Konrad
+ */
+@Internal
+public class FilterRunner {
```
graemerocher, line 77: can this be made final?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+
+    public static void sort(List<GenericHttpFilter> filters) {
+        checkOrdered(filters);
+        OrderUtil.sort(filters);
+    }
+
+    public static void sortReverse(List<GenericHttpFilter> filters) {
+        checkOrdered(filters);
+        OrderUtil.reverseSort(filters);
+    }
```
graemerocher: javadoc/nulllability annotations for these public methods

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+    private HttpRequest<?> request;
+    private HttpResponse<?> response;
+    private Throwable failure;
+    private SuspensionPoint responseSuspensionPoint = new SuspensionPoint(-1, null);
+    private int index;
+    private boolean responseNeedsProcessing = false;
+
+    private Context reactorContext = Context.empty();
+
+    public FilterRunner(List<GenericHttpFilter> filters) {
```
graemerocher: javadoc

test-suite-groovy/src/test/groovy/io/micronaut/docs/client/filter/GoogleAuthFilter.groovy
```diff
@@ -1,40 +1,34 @@
...
-                )))
+    @RequestFilter
+    void filter(MutableHttpRequest<?> request) {
+        String authURI = encodeURI(request)
+        String token = authClientProvider.get().toBlocking().retrieve(HttpRequest.GET(authURI).header( // <2>
+                "Metadata-Flavor", "Google"
+        ))
 
-        return token.flatMap(t -> chain.proceed(request.bearerAuth(t)))
+        request.bearerAuth(token)
```
graemerocher, line 34: this changes the example to encourage blocking, how would you write this filter without blocking?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+
+    @Internal
+    public static <T> FilterMethod<T> prepareFilterMethod(T bean, ExecutableMethod<T, ?> method, boolean isResponseFilter, FilterOrder order) throws IllegalArgumentException {
+        return prepareFilterMethod(bean, method, method.getArguments(), method.getReturnType().asArgument(), isResponseFilter, order);
+    }
+
+    @Internal
+    public static <T> FilterMethod<T> prepareFilterMethod(T bean, ExecutableMethod<T, ?> method, Argument<?>[] arguments, Argument<?> returnType, boolean isResponseFilter, FilterOrder order) throws IllegalArgumentException {
+        FilterArgBinder[] fulfilled = new FilterArgBinder[arguments.length];
+        Predicate<FilterRunner> filterConditionAlwaysTrue = runner -> true;
```
graemerocher: can this be a constant?

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+            context.fail("Filter method must be declared on a @ServerFilter or @ClientFilter bean", element);
+            return;
+        }
+
+        try {
+            Argument<?>[] args = toArguments(Arrays.stream(element.getParameters()).map(ParameterElement::getType).toList(), context);
+            Argument<?> ret = toArgument(element.getReturnType(), context);
+
+            if (requestFilterAnnotation != null) {
+                FilterRunner.prepareFilterMethod(null, null, args, ret, false, null);
```
graemerocher: it is not clear to me what this call is doing and why it is in a source code processor. You seem to be converting the source code elements to runtime types? And then what?

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+            if (responseFilterAnnotation != null) {
+                FilterRunner.prepareFilterMethod(null, null, args, ret, true, null);
+            }
+        } catch (IllegalArgumentException e) {
+            context.fail(e.getMessage(), element);
+        }
+        context.warn("hi", element);
```
graemerocher: is this a typo?

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+            Argument<?> ret = toArgument(element.getReturnType(), context);
+
+            if (requestFilterAnnotation != null) {
+                FilterRunner.prepareFilterMethod(null, null, args, ret, false, null);
+            }
+            if (responseFilterAnnotation != null) {
+                FilterRunner.prepareFilterMethod(null, null, args, ret, true, null);
+            }
+        } catch (IllegalArgumentException e) {
+            context.fail(e.getMessage(), element);
```
graemerocher, line 94: clearer error message here

graemerocher (changes requested):

yawkat:
@dstepanov ive added a bunch of return type tests, including null and optional. i will do coroutine support in a follow-up pr i think, this is already big enough

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: it has two protected methods processResponse and processFailure, they are overridden by the server so that exceptions are transformed into 500 responses, and so that status routes are called. this is necessary for compatibility

test-suite-groovy/src/test/groovy/io/micronaut/docs/client/filter/GoogleAuthFilter.groovy
yawkat: you can still use reactive (just return a publisher), but imo it shouldn't be the default in the guide (given virtual threads, and to avoid making the code more confusing)

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
yawkat: haha printf debugging

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
yawkat: it will throw an exception when the filter is invalid. i added a comment. it's a bit weird to mix compile time and run time types like this, and it's right on the edge of "too smart to be understandable" imo. But i think it makes the code genuinely simpler and reduces duplication.

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
yawkat: error message comes from prepareFilterMethod and already includes detail

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ClientRequestFilterTest.java
```diff
@@ -0,0 +1,558 @@
...
+            return HttpRequest.GET("/request-filter/replace-request-2");
+        }
+
+        @RequestFilter("/request-filter/replace-mutable-request")
+        public MutableHttpRequest<Object> requestFilterReplaceMutableRequest() {
+            return HttpRequest.GET("/request-filter/replace-mutable-request-2");
+        }
+
+        @RequestFilter("/request-filter/replace-request-null")
+        public HttpRequest<Object> requestFilterReplaceRequestNull() {
```
dstepanov, line 391: This should't be allowed without `@Nullable` annotation

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ResponseFilterTest.java
```diff
@@ -0,0 +1,342 @@
...
+                    .body("responseFilterReplaceCompletionResponse foo")
+                    .build());
+            })
+            .run();
+    }
+
+    @ServerFilter
+    @Singleton
+    @Requires(property = "spec.name", value = SPEC_NAME)
+    public static class MyServerFilter {
```
dstepanov, line 255: Please have Optional / Nullable response tests

dstepanov: Do we also have scenarious when some response types not allowed?

dstepanov (changes requested):

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ResponseFilterTest.java
yawkat: you mean like forbidden body types? i dont think so, the body is treated as opaque

graemerocher: if so we should fail compilation

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ClientRequestFilterTest.java
yawkat: more annoying to implement :see_no_evil: but ok

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ResponseFilterTest.java
dstepanov: Yes, we should fail if somebody uses not allowed body response

http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
yawkat: yes but it was unused. im not sure what it was supposed to do in the first place

yawkat: #8624

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ResponseFilterTest.java
yawkat: @dstepanov i dont understand what you mean, you can return whatever body you want from a filter, i dont think there are any constraints

dstepanov: You can only return:
Request filters:
- void, request or response in optional/futures/reactive...

Response filters:
- void, response in optional/futures/reactive...

yawkat: ohhh i misunderstood, i thought you meant like `HttpResponse<XYZ>`. other return types already lead to a compile error, i will add a test

yawkat: oh there is a test, it's in FilterVisitorSpec

graemerocher (approved):

src/main/docs/guide/httpServer/filters.adoc
```diff
@@ -6,17 +6,17 @@ Filters support the following use cases:
...
 
-For a server application, the api:http.filter.HttpServerFilter[] interface `doFilter` method can be implemented.
+There are two ways to implement a filter. There is the "old" (pre Micronaut HTTP 4.0) approach using api:http.filter.HttpServerFilter[] (or api:http.filter.HttpClientFilter[]), and the "new" way using declarative filter methods. This guide only describes the latter.
 
-The `doFilter` method accepts the api:http.HttpRequest[] and an instance of api:http.filter.ServerFilterChain[].
+A filter method must be declared in a bean annotated with api:http.annotation.ServerFilter[], or api:http.annotation.ClientFilter[] if it should instead intercept requests made by the HTTP client. Each filter method must also be annotated with api:http.annotation.RequestFilter[], to run before the request is processed, or api:http.annotation.ResponseFilter[], to run after the request has completed to process the response.
 
-The `ServerFilterChain` interface contains a resolved chain of filters where the final entry in the chain is the matched route. The api:http.filter.ServerFilterChain.proceed(io.micronaut.http.HttpRequest)[] method resumes processing of the request.
```
dstepanov, line 13: It would be nice to have a table with possible return types and explain their meaning.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+        while (true) {
+            if (!workRequestFilter(filters.get(index++))) {
+                return;
+            }
+        }
+    }
+
+    private boolean workRequestFilter(GenericHttpFilter filter) {
+        Executor executeOn;
+        if (filter instanceof GenericHttpFilter.Async async) {
```
dstepanov, line 241: Would it make sense to use a visitor instead of multiple instanceof?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+    }
+
+    private <T> boolean invokeBefore(FilterMethod<T> before, @Nullable Executor completeOn) {
+        FilterContinuationImpl<?> passedOnContinuation = null;
+        try {
+            if (before.filterCondition != null && !before.filterCondition.test(this)) {
+                return true;
+            }
+            Object[] args = bindArgs(before);
+            if (before.passedOnContinuationArgIndex != -1) {
```
dstepanov: `before` can handle the logic returning the `passedOnContinuation` as nullable

dstepanov: I would put this method into `FilterMethod`, it looks messy just to use its fields.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+    }
+
+    @Internal
+    public static <T> FilterMethod<T> prepareFilterMethod(T bean, ExecutableMethod<T, ?> method, Argument<?>[] arguments, Argument<?> returnType, boolean isResponseFilter, FilterOrder order) throws IllegalArgumentException {
+        FilterArgBinder[] fulfilled = new FilterArgBinder[arguments.length];
+        Predicate<FilterRunner> filterConditionAlwaysTrue = runner -> true;
+        Predicate<FilterRunner> filterCondition = filterConditionAlwaysTrue;
+        boolean skipOnError = isResponseFilter;
+
+        int passedOnContinuationArgIndex = -1;
```
dstepanov: It would be nice to use `ArgumentBinderRegistry`

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+        );
+    }
+
+    private static boolean isReactive(Argument<?> continuationReturnType) {
+        // Argument.isReactive doesn't work in http-validation, this is a workaround
+        return continuationReturnType.isReactive() || continuationReturnType.getType() == Publisher.class;
+    }
+
+    private static FilterReturnHandler prepareReturnHandler(Argument<?> type, boolean isResponseFilter, boolean hasContinuation) throws IllegalArgumentException {
+        if (type.isOptional()) {
```
dstepanov, line 498: Most of the lambdas here can be turn into static constants

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+
+                    // cancel request pipeline, move immediately to response handling
+                    r.response = (HttpResponse<?>) o;
+                    r.failure = null;
+                    r.responseNeedsProcessing = true;
+                    return true;
+                };
+            }
+        }
+        if (isReactive(type)) {
```
dstepanov, line 543: It would be nice to have those values precalculated in some shared definition of the filter

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+
+        ExecutionFlow<HttpResponse<?>> resultFlow = CompletableFutureExecutionFlow.just(responseSuspensionPoint);
+        workRequest();
+        //noinspection unchecked,rawtypes
+        return (ExecutionFlow) resultFlow;
+    }
+
+    private void workRequest() {
+        while (true) {
+            if (!workRequestFilter(filters.get(index++))) {
```
dstepanov: Why don't you split filters into request/filter? Having one collection is confusing to understand.

dstepanov: All the occurrences of `index` are very confusing, sometimes the index goes up sometimes it goes down. Can this be rewritten using an iterator? Or at least pass the index as a downstream parameter, so it's not modified from multiple places.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: Why not a constant?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+    @Internal
+    public static <T> FilterMethod<T> prepareFilterMethod(T bean, ExecutableMethod<T, ?> method, Argument<?>[] arguments, Argument<?> returnType, boolean isResponseFilter, FilterOrder order) throws IllegalArgumentException {
+        FilterArgBinder[] fulfilled = new FilterArgBinder[arguments.length];
+        Predicate<FilterRunner> filterConditionAlwaysTrue = runner -> true;
+        Predicate<FilterRunner> filterCondition = filterConditionAlwaysTrue;
+        boolean skipOnError = isResponseFilter;
+
+        int passedOnContinuationArgIndex = -1;
+        for (int i = 0; i < arguments.length; i++) {
+            Argument<?> argument = arguments[i];
```
dstepanov, line 411: Extract a method to fill the value to avoid duplicating `fulfilled[i] = `

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+            }
+        }
+        if (isReactive(type)) {
+            var next = prepareReturnHandler(type.getWrappedType(), isResponseFilter, hasContinuation);
+            return new DelayedFilterReturnHandler(isResponseFilter, next) {
+                @Override
+                protected ExecutionFlow<?> toFlow(FilterRunner runner, Object o) {
+                    //noinspection unchecked
+                    return ReactiveExecutionFlow.fromPublisher(
+                        Mono.from(Publishers.convertPublisher(o, Publisher.class))
```
dstepanov: Pass `ConversionService`

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -0,0 +1,966 @@
...
+        private boolean addedListener = false;
+
+        ReactiveContinuationImpl(int filterIndex, @Nullable SuspensionPoint next, Class<R> reactiveType) {
+            super(filterIndex, next);
+            this.reactiveType = reactiveType;
+        }
+
+        @Override
+        public R proceed() {
+            return Publishers.convertPublisher(this, reactiveType);
```
dstepanov: Pass `ConversionService`

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov, line 77: It feels to me that some of the code can be extracted into small methods to explain what it does. There is so much work with fields, so it's hard to read.

dstepanov: It looks a bit suspicious to me that some of the methods can be executed on a different thread (If I understand correctly) `executeOn.execute``terminalFlow.onComplete` with the class fields being modified, if this is a case, I would have written this as one flow carrying some data structure along the stream.

dstepanov (changes requested):

dstepanov:
Do we have tests on which thread is used in request/response filters with various combinations like sync, async, reactive? Different combinations like, what thread should be if the previous filter is running on some custom thread? It would be nice to have it defined by tests, at least for future reference.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: i think it would have the same performance (or worse? it goes from a simple instance check to a multimorphic call site), and i tried to keep call stacks small so that stack traces are easier to read

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: done, had to refactor some (they captured the nullable variable) but i moved them to static fields in FilterReturnHandler

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: sorry changed locally but didnt push

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: maybe, but with the limited types supported atm it makes more sense to keep this simple. however i kept the option open, FilterRunner is internal and only used by micronaut-core, so we have an applicationcontext if necessary

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: it can't be extracted because some parameter types also change the other local variables, for example having a Throwable parameter type changes the filterCondition

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: it counts up for the request filters, until it needs to run response filters (at the end of the pipeline, when it hits a filter that returns a response, or when there is an error), and then it counts down again.

i will try to refactor the code using a passing style instead to see if it's easier to understand

yawkat: ok i genuinely think it's worse in passing style. the suspension points become messy. let me try to write better documentation for how it's currently done instead.

yawkat: ive added a looong comment explaining the coroutine style the FilterRunner uses. LMK what you think.

src/main/docs/guide/httpServer/filters.adoc
yawkat: ive linked the javadoc of RequestFilter and ResponseFilter. they have details on the allowed return types, but as a list, not a table. is that enough?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: Most of the time, there is only one thread that modifies the class fields. e.g. when there is an ExecuteOn, it suspends the work on the original thread, and resumes on the new thread.

There are some exceptions to this, e.g. with a continuation (or legacy filter), the user could return from the filter (or complete the publisher returned by the filter) before the downstream `proceed` is done. These cases are handled by the concurrency structures in the SuspensionPoint subclasses

yawkat:
@dstepanov it is intentionally not specified what thread the filters run on by default. There are some tests for ExecuteOn (the async tests in FilterRunnerSpec). But if there is no such annotation, filters just run on the same thread as the upstream filters. If there are no upstream filters with ExecuteOn, that's the event loop, if there are upstream filters with ExecuteOn, that's their threads.

The reasoning here is that filters should not do blocking operations unless they are annotated with ExecuteOn, and if they are non-blocking anyway, it doesn't matter what thread they run on. We could change this in the future though, for example by running filters on a virtual thread by default if the overhead ever becomes negligible.

dstepanov:
We don't need to specify it officially, but it's good to have it tested so we know that some small refactoring is not going to change it unexpectedly. We had multiple cases like that with previous filters.

yawkat:
ok i think it will be easy to add a test in FilterRunnerSpec.

dstepanov:
Thanks for the explanation. I will look at it again today.

sdelamo (changes requested):
Misses docs

dstepanov (approved):

sdelamo (changes requested):
I don't think we should delete the old way of doing filter documentation. Currently documentation looks broken to me: 

CleanShot 2023-02-16 at 16 51 19@2x

yawkat:
ive closed the tables now, please check again.

why should we keep the old filter docs? the new filters can do everything the old filters can.

sdelamo:
> ive closed the tables now, please check again.
> 
> why should we keep the old filter docs? the new filters can do everything the old filters can.

because, we still support the old filters. We should not hide documentation from users. Specially, if we had already written the docs.

graemerocher:
@sdelamo I guess it depends if we regard the old approach to writing filters as deprecated. If we do then it shouldn't be present in the documentation.

yawkat:
i see no reason to use them anymore with the new replacement, but they also cost us basically nothing to support, so they can stick around for 4.x. but yes i removed the docs because i dont want to encourage their use anymore

gradle/libs.versions.toml
```diff
@@ -44,7 +44,7 @@ micronaut-session = "1.0.0-SNAPSHOT"
 micronaut-sql = "4.7.2"
 micronaut-test = "4.0.0-SNAPSHOT"
 micronaut-serde = "2.0.0-SNAPSHOT"
-micronaut-tracing = "4.5.0"
+micronaut-tracing = "5.0.0-SNAPSHOT"
```
sdelamo, line 47: do we need this in core?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1243,45 +1246,46 @@ private <I, R extends io.micronaut.http.HttpResponse<?>> Publisher<R> applyFilte
...
+            mutRequest.setAttribute(HttpAttributes.SERVICE_ID, informationalServiceId);
+        }
 
-            OrderUtil.reverseSort(filters);
-            Publisher<R> finalResponsePublisher = responsePublisher;
-            filters.add((req, chain) -> finalResponsePublisher);
+        List<GenericHttpFilter> filters =
+                filterResolver.resolveFilters(request, clientFilterEntries);
+        if (parentRequest != null) {
+            // todo: migrate to new filter
```
sdelamo, line 1263: is this pending?

router/src/main/java/io/micronaut/web/router/ServerFilterRouteBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            @Override
+            protected void addFilter(Supplier<GenericHttpFilter> factory, AnnotationMetadata methodAnnotations, FilterMetadata metadata) {
+                applyMetadata(ServerFilterRouteBuilder.this.addFilter(factory, methodAnnotations), metadata);
+            }
+
+            private void applyMetadata(FilterRoute route, FilterMetadata metadata) {
+                route.patternStyle(metadata.patternStyle());
+                // todo: handle missing patterns
```
sdelamo: is this pending?

test-suite-groovy/src/test/groovy/io/micronaut/docs/client/filter/GoogleAuthFilter.groovy
```diff
@@ -1,40 +1,34 @@
...
     }
 
-    @Override
-    Publisher<? extends HttpResponse<?>> doFilter(MutableHttpRequest<?> request,
-                                                  ClientFilterChain chain) {
-        Flux<String> token = Mono.fromCallable(() -> encodeURI(request))
-                .flatMap(authURI -> authClientProvider.get().retrieve(GET(authURI).header( // <2>
-                        "Metadata-Flavor", "Google"
-                )))
+    @RequestFilter
```
sdelamo, line 26: should not this be annotated with `@ExecutesOn(BLOCKING)`?

if you have a controller which uses a declarative http client which returns a reactive type this filter will block the event loop.

src/main/docs/guide/httpClient/clientFilter.adoc
```diff
@@ -1,6 +1,6 @@
 Often, you need to include the same HTTP headers or URL parameters in a set of requests against a third-party API or when calling another Microservice.
 
-To simplify this, you can define api:http.filter.HttpClientFilter[] classes that are applied to all matching HTTP client requests.
+To simplify this, you can define api:http.annotation.ClientFilter[] classes that are applied to all matching HTTP client requests. The details of how filters worked are described in the <<filter, server filter documentation>>. Here we will only show some client filters.
```
sdelamo:
```suggestion
To simplify this, you can define api:http.annotation.ClientFilter[] classes that are applied to all matching HTTP client requests. The details of how filters worked are described in the <<filters, server filter documentation>>. Here we will only show some client filters.
```

sdelamo (changes requested):

sdelamo:
@graemerocher 

> @sdelamo I guess it depends if we regard the old approach to writing filters as deprecated. If we do then it shouldn't be present in the documentation.

@yawkat 

> i see no reason to use them anymore with the new replacement, but they also cost us basically nothing to support, so they can stick around for 4.x. but yes i removed the docs because i dont want to encourage their use anymore

I feel strongly that we should not delete documentation of deprecated APIs. 

Moreover, `HttpServerFilter` API is not deprecated. 
We should only delete API documentation when we delete the corresponding APIs.

I see the benefits of Filter methods in many scenarios. 
However, The HttpServerFilter API is easier to grasp in certain use cases than the filter methods APIs. 

For example, when the filter logic depends on a reactive flow. E.g. in the SecurityFilter.

It is clear what you have to do as a user. You implement `HttpServerFilter` in a class annotated with `@Filter` and the IDE forces you to implement a method you can obtain a valid return type for that method doing `chain.proceed(request);`. Both `chain` and `request` are parameters of the functional interface  `HttpServerFilter`. 

With filter methods,  you have to create a method annotated with `@RequestFilter`, and then you have to know that you have to use a parameter of type `FilterContinuation` in that method. Moreover, you have to use reactive type as the type of the `FilterContinuation` `Publisher<HttpResponse<?>>`. 

The latter is a more complex API to understand than `HttpServerFilter`. I am not sure we should deprecate `HttpServerFilter` until we see how the community embraces filter methods and we remove every usage of HttpServerFitler and HttpClientFilter in the other modules. 

I have created a PR that [adds the HttpServerFilter documentation back](https://github.com/micronaut-projects/micronaut-core/pull/8812).

sdelamo:
@yawkat Can you respond to: 

https://github.com/micronaut-projects/micronaut-core/pull/8422#discussion_r1113819282
https://github.com/micronaut-projects/micronaut-core/pull/8422#discussion_r1113840473
https://github.com/micronaut-projects/micronaut-core/pull/8422#discussion_r1113860348
https://github.com/micronaut-projects/micronaut-core/pull/8422#discussion_r1113863392

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: yes to make this pr smaller

gradle/libs.versions.toml
yawkat: idk why i changed this so im gonna revert it

sdelamo (approved):

gradle/libs.versions.toml
yawkat: it was an NPE fix for kotlin apparently, i think https://github.com/micronaut-projects/micronaut-tracing/commit/d27a5dfef858c21494ee71c5088ecb72744c06f7

PR #8424, opened 2022-11-25
Support arbitrary nesting of EachProperty/ConfigurationProperties beans

graemerocher:
This change adds a new ConfigurationPath interface and implementation that is used to keep track of the state of configuration nesting and to allow nesting of @EachProperty declarations. For nested @EachProperty declarations a composite qualifier is formed from all the names in the path. 

For example given a pattern foo.*.bar.*.baz and the property foo.1st.bar.2nd.baz the resulting qualifier will be "1st-2nd".

Note that checkstyle is failing because we need to upgrade to a version of Checkstyle that suppports Java 17 constructs.


dstepanov:
Still failing?

graemerocher:
only checkstyle, see my note:

> Note that checkstyle is failing because we need to upgrade to a version of Checkstyle that suppports Java 17 constructs.

core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -121,6 +122,30 @@ default boolean isOptional() {
         return isAssignable(Optional.class) || isAssignable(OptionalLong.class) || isAssignable(OptionalDouble.class) || isAssignable(OptionalInt.class);
     }
 
+    /**
+     * Checks whether the bean type is a container type.
+     * @return Whether the type is a container type like {@link Iterable}.
+     * @since 4.0.0
+     */
+    default boolean isContainerType() {
```
dstepanov, line 130: I don't think these bean context injection helper methods should belong to the `ClassElement`

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -220,6 +225,49 @@ public Collection<String> getPropertyEntries(@NonNull String name) {
...
+            Map<String, Object> entries = resolveEntriesForKey(
+                pathPattern, false, null);
+
+            if (entries != null) {
+                boolean endsWithWildCard = pathPattern.endsWith(".*");
+                String resolvedPattern = pathPattern
+                    .replace("[*]", "\\[([\\w\\d-]+?)\\]")
+                    .replace(".*.", "\\.([\\w\\d-]+?)\\.");
+                if (endsWithWildCard) {
+                    resolvedPattern = resolvedPattern.replace("*.", "\\S*");
```
dstepanov: Did you want to write `.*` instead of `*.`?

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
```diff
@@ -51,7 +51,9 @@
 
     /**
      * Attribute used to store a dynamic bean name.
+     * @deprecated Previous internal usage, now no longer used.
      */
+    @Deprecated(forRemoval = true, since = "4.0.0")
     String NAMED_ATTRIBUTE = Named.class.getName();
```
dstepanov: I would just remove the deprecations.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2283,7 +2332,9 @@ private <T> Set<BeanDefinition<T>> collectBeanCandidates(
      * @param filterProxied     Whether to filter proxied.
      * @param <T>               The bean type
      * @return The candidates
+     * @deprecated No longer invoked by the framework. Use {@link #collectIterableBeans(BeanResolutionContext, BeanDefinition, Set)} instead.
      */
+    @Deprecated(forRemoval = true, since = "4.0.0")
     protected <T> Collection<BeanDefinition<T>> transformIterables(BeanResolutionContext resolutionContext, Collection<BeanDefinition<T>> candidates, boolean filterProxied) {
```
dstepanov: I would just remove the deprecations.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2193,12 +2193,37 @@ protected <T> Collection<BeanDefinition<T>> findBeanCandidates(@Nullable BeanRes
...
+     * @param <T>               The bean generic type
+     * @param resolutionContext The current resolution context
+     * @param beanType          The bean type
+     * @param filterProxied     Whether to filter out bean proxy targets
+     * @param collectIterables  Whether iterables should be collected
+     * @param predicate         The predicate to filter candidates
+     * @return The candidates
+     */
+    @NonNull
+    protected <T> Collection<BeanDefinition<T>> findBeanCandidates(@Nullable BeanResolutionContext resolutionContext,
```
dstepanov, line 2222: I think we should refactor out all the caching and logic around `BeanDefinition` resolution into a separate class and use it as a field to reduce the complexity of the class, `DefaultApplicationContext` would have a custom impl. adding transform logic.

inject/src/main/java/io/micronaut/inject/qualifiers/Qualifiers.java
```diff
@@ -128,6 +130,17 @@ public static <T> Qualifier<T> byName(String name) {
...
     }
 
+    /**
+     * Qualify by a prefix. Applies starting with logic to the name of the bean..
+     *
+     * @param prefix The name
+     * @param <T>  The component type
+     * @return The qualifier
+     */
+    public static <T> Qualifier<T> byNamePrefix(String prefix) {
```
dstepanov, line 141: Please put `@since`

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -376,223 +369,154 @@ private static String resolveEachPropertyMissingBeanMessage(Qualifier<?> qualifi
...
-                }
-            }
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("Finalized bean definitions candidates: {}", candidates);
-                for (BeanDefinition<?> definition : transformedCandidates) {
-                    LOG.debug("  {} {} {}", definition.getBeanType(), definition.getDeclaredQualifier(), definition);
-                }
+    protected <T> void collectIterableBeans(BeanResolutionContext resolutionContext, BeanDefinition<T> iterableBean, Set<BeanDefinition<T>> targetSet) {
+        try(BeanResolutionContext rc = newResolutionContext(iterableBean, resolutionContext)) {
+            if (iterableBean.hasDeclaredStereotype(EachProperty.class)) {
```
dstepanov, line 406: You can add `if (candidate.isIterable())`, which should be a compile-time evaluated value.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -3381,28 +3441,18 @@ private <T> Optional<BeanDefinition<T>> pickOneBean(
         return Optional.ofNullable(definition);
     }
 
+    @SuppressWarnings("java:S1871")
     private <T> void filterProxiedTypes(
-        Collection<BeanDefinition<T>> candidates,
-        boolean filterDelegates) {
+        Collection<BeanDefinition<T>> candidates) {
         int count = candidates.size();
         Set<Class<?>> proxiedTypes = new HashSet<>(count);
```
dstepanov, line 3433: `new HashSet<>(count)` Looks incorrect as it will be resized before it will reach the count capacity. There are some options on how to do it in Java 19 https://bugs.openjdk.org/browse/JDK-8186958 Maybe we can have similar internal util methods.

inject/src/main/java/io/micronaut/context/env/DefaultConfigurationPath.java
```diff
@@ -0,0 +1,481 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Implementation of {@link ConfigurationPath}.
+ *
+ * @since 4.0.0
+ */
+final class DefaultConfigurationPath implements ConfigurationPath {
+    private final LinkedList<ConfigurationSegment> list = new LinkedList<>();
```
dstepanov, line 38: It might be better to use ArrayDeque, which should be more performant. We might want to also use it for `AbstractBeanResolutionContext.DefaultPath`

dstepanov (changes requested):

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
```diff
@@ -166,6 +157,13 @@ default boolean isIterable() {
         return hasDeclaredStereotype(EachProperty.class) || hasDeclaredStereotype(EachBean.class);
     }
 
+    /**
+     * @return Is the type configuration properties.
+     */
+    default boolean isConfigurationProperties() {
+        return isIterable() || hasDeclaredStereotype(ConfigurationReader.class);
```
dstepanov, line 157: This might not be correct: `EachBean` has nothing to do with configuration properties.

dstepanov (changes requested):

dstepanov:
Maybe you can also include tests for error case messages.

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
graemerocher: True but the existing logic in `BeanDefinitionWriter` which computes this value for `BeanDefinitionReference` has the same logic. See https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java#L4093-L4095

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
graemerocher: I didn't change this, however I have another PR coming after this one that removes this logic

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
graemerocher: That currently includes beans annotated with `@EachBean` and I need to differentiate

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
graemerocher: Agreed, but that can be done in a separate PR

core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
graemerocher: Where do you suggest?

dstepanov: Maybe some injec util class?

inject/src/main/java/io/micronaut/context/env/DefaultConfigurationPath.java
graemerocher: There is no `subList` method in ArrayDeque

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
dstepanov: I feel like we would rely on this check by mistake

graemerocher: I agree but it has been that way for some time 🤔

core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
graemerocher: I moved `isProvider` out since there was only one use just to a private method. The `isContainerType` method doesn't seem to have to do with DI as it relates to collections so I left it

dstepanov (approved):

PR #8429, opened 2022-11-27
build: Bump micronaut-jackson-xml to 3.2.0

micronaut-build:


sdelamo:
Micronaut Jackson XML 3.8 contains the BOM. Could you continue this PR to leverage the BOM in core.

PR #8155, opened 2022-10-13
Fix HttpClient.start

yawkat:
This fixes the tests in micronaut-reactor (https://github.com/micronaut-projects/micronaut-reactor/pull/230), and copies the relevant test to this repo as well.

PR #8434, opened 2022-11-28
Backport #8155

yawkat:
Fix HttpClient.start (#8155)
This fixes the tests in micronaut-reactor, and copies the relevant test to this repo as well.


timyates (approved):

sdelamo:
the failure is unrelated to the Pr is graalvm workflow with 11 which is no longer valid. If the other pass, I will merge it.

timyates:
I pushed a PR to exclude Graal-11-dev (but keep Graal-11-latest), so it should go green

When we get some action runners

PR #8444, opened 2022-11-29
proxy target rework

graemerocher:
- Overhaul the support for configuration property nesting
- fix tests
- fix tests
- fix error message / use cached environment
- fix tests
- javadoc
- fix test
- optimize
- simplify
- simplify
- simplify
- temporarily disable checkstyle until fixed with Java 17
- Rework proxy target resolution
- delete unused annotation


inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinitionReference.java
```diff
@@ -67,6 +68,26 @@ public AbstractInitializableBeanDefinitionReference(String beanTypeName, String
...
+     * @param isPrimary                 Is primary bean?
+     * @param isContextScope            Is context scope?
+     * @param isConditional             Is conditional? = No @Requires
+     * @param isContainerType           Is container type?
+     * @param isSingleton               Is singleton?
+     * @param isConfigurationProperties Is configuration properties?
+     * @param hasExposedTypes           Has exposed types?
+     * @param requiresMethodProcessing  Is requires method processing?
+     */
+    protected AbstractInitializableBeanDefinitionReference(String beanTypeName, String beanDefinitionTypeName, AnnotationMetadata annotationMetadata,
```
dstepanov, line 87: Can you introduce a new object to represent the internal state which would be generated and passed via the constructor? Something like:

```java
class Data {
//...
     private final boolean isConfigurationProperties;
    private final boolean hasExposedTypes;
    private final boolean requiresMethodProcessing;
    private final boolean isProxiedBean;
//...
}

class AbstractInitializableBeanDefinitionReference {

    private final String beanTypeName;
    private final String beanDefinitionTypeName;
    private final AnnotationMetadata annotationMetadata;
    private final Data data;

    protected AbstractInitializableBeanDefinitionReference(Data data) {
          this.data = data;
    }

    public boolean isProxiedBean() {
        return data.isProxiedBean;
    }
}
```

I'm planing to add something for CDI, and adding/modifying the constructor IMHO is a pain.

PR #8451, opened 2022-12-01
Fix bug in introspections when properties are of mixed types

### LINKED ISSUE #8450: Regression in Micronaut 4.0 when reading property values
### Expected Behavior

Getters/setters with mixed types broken.

### Actual Behaviour

Should work

### Steps To Reproduce

Introspection with:

```java
package example;
import com.fasterxml.jackson.annotation.JsonSetter;
import java.util.Optional;
import io.micronaut.serde.annotation.Serdeable;
@Serdeable
class Test {
    @Nullable
    private String foo;
    public Optional<String> getFoo() {
        return Optional.ofNullable(foo);
    }
    @JsonSetter("foo")
    public void setFoo(@Nullable String foo) {
        this.foo = foo;
    }
}
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.0.x

graemerocher:
Tests and fixes https://github.com/micronaut-projects/micronaut-core/issues/8450


dstepanov (approved):

graemerocher:
hmm tests fail

PR #8459, opened 2022-12-02
Fix bug with custom conditions and EachBean definitions

graemerocher:
Conditions were not being executed for `@EachBean` definitions without an upstream `@EachProperty`, this fixes that.


timyates:
@graemerocher was this to fix one of my reported issues?  If so, I'll functionally test it against that module

graemerocher:
yes should fix the tests for security module

timyates:
Tested this PR with the security module, and got these errors:

https://ge.micronaut.io/s/sihwp6ncotlcs/tests/overview?outcome=failed

Which are the same errors Dean was seeing in CI (in https://github.com/micronaut-projects/micronaut-security/pull/1160)

https://ge.micronaut.io/s/vhfk4p5lpjagy/tests/overview?outcome=failed

graemerocher:
I merged this in the meantime as those are different errors

PR #8462, opened 2022-12-06
Adds support for KSP

graemerocher:
Adds support for Kotlin Symbol Processing (KSP)


core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
```diff
@@ -829,7 +829,7 @@ private void writeInstantiateMethod(ClassWriter classWriter, MethodElement const
 
     private void invokeBeanConstructor(GeneratorAdapter writer, MethodElement constructor, BiConsumer<GeneratorAdapter, MethodElement> argumentsPusher) {
         boolean isConstructor = constructor instanceof ConstructorElement;
-        boolean isCompanion = constructor != defaultConstructor && constructor.getDeclaringType().getSimpleName().endsWith("$Companion");
+        boolean isCompanion = constructor.getDeclaringType().getSimpleName().endsWith("$Companion");
```
yawkat, line 832: A companion class does not have to be called `$Companion`, you can declare it with a custom name [like this](https://javap.yawk.at/#GKiuNf):

```kotlin
class Main() {
    companion object Foo {
        val s = "foo"
    }
}
```

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -624,6 +624,9 @@ public AnnotationClassValue<?>[] annotationClassValues(@NonNull String member) {
         if (o instanceof AnnotationClassValue<?>[] annotationClassValues) {
             return annotationClassValues;
         }
+        if (o instanceof Collection<?> collection) {
```
yawkat: why allow both arrays and collections?

inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/AbstractKotlinCompilerSpec.groovy
```diff
@@ -58,40 +57,105 @@ class AbstractKotlinCompilerSpec extends Specification {
...
+    protected static String reconstructTypeSignature(ClassElement classElement, boolean typeVarsAsDeclarations = false) {
+        if (classElement.isArray()) {
+            return "Array<" + reconstructTypeSignature(classElement.fromArray()) + ">"
+        } else if (classElement.isGenericPlaceholder()) {
+            def freeVar = (GenericPlaceholderElement) classElement
+            def name = freeVar.variableName
+            if (typeVarsAsDeclarations) {
+                def bounds = freeVar.bounds
+                if (reconstructTypeSignature(bounds[0]) != 'Object') {
+                    name += bounds.stream().map(AbstractKotlinCompilerSpec::reconstructTypeSignature).collect(Collectors.joining(" & ", " out ", ""))
```
yawkat, line 138: this is wrong, but it looks like the only test that uses it is `@PendingFeature`

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/annotation/KotlinAnnotationMetadataBuilder.kt
```diff
@@ -0,0 +1,564 @@
...
+        }
+        return map
+    }
+
+    override fun readAnnotationDefaultValues(
+        annotationName: String,
+        annotationType: KSAnnotated
+    ): MutableMap<out KSDeclaration, *> {
+        // issue getting default values for an annotation here
+        // TODO: awful hack due to https://github.com/google/ksp/issues/642 and not being able to access annotation defaults for a type
```
yawkat, line 331: https://github.com/google/ksp/issues/642 seems to refer to default values of places that are *not* annotations?

yawkat: oh i think i understand, it's possible to access the value at an annotation use site, but not the default value at the annotation declaration site?

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/extensions.kt
```diff
@@ -0,0 +1,149 @@
...
+        this.modifiers
+    } catch (e: IllegalStateException) {
+        // KSP bug: IllegalStateException: unhandled visibility: invisible_fake
+        setOf(Modifier.INTERNAL)
+    }
+    return when {
+        modifierSet.contains(Modifier.PUBLIC) -> Visibility.PUBLIC
+        modifierSet.contains(Modifier.PRIVATE) -> Visibility.PRIVATE
+        modifierSet.contains(Modifier.PROTECTED) ||
+                modifierSet.contains(Modifier.OVERRIDE) -> Visibility.PROTECTED
```
yawkat, line 90: seems like for OVERRIDE it should instead copy the visibility of the parent? but maybe this is fine because we don't care as long as visibility is > private

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
```diff
@@ -0,0 +1,323 @@
...
+
+    @OptIn(KspExperimental::class)
+    override fun getModifiers(): MutableSet<ElementModifier> {
+        val dec = declaration.unwrap()
+        if (dec is KSDeclaration) {
+            val javaModifiers = visitorContext.resolver.effectiveJavaModifiers(dec)
+            return javaModifiers.mapNotNull {
+                when (it) {
+                    Modifier.FINAL -> ElementModifier.FINAL
+                    Modifier.PRIVATE, Modifier.INTERNAL -> ElementModifier.PRIVATE
```
yawkat: private is not a good mapping for internal

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
```diff
@@ -0,0 +1,857 @@
...
+        classDeclaration.qualifiedName!!.asString()
+    }
+    private val internalName : String by lazy {
+        classDeclaration.getBinaryName(visitorContext.resolver, visitorContext)
+    }
+
+    private var overrideBoundGenericTypes: MutableList<out ClassElement>? = null
+    private var resolvedTypeArguments : MutableMap<String, ClassElement>? = null
+
+    private val nt : KSAnnotated =  if (annotationInfo is KSTypeArgument) annotationInfo else KSClassReference(classDeclaration)
```
yawkat, line 104: `override val nativeType` should work instead

yawkat (commented):
not done reviewing yet

core-processor/src/main/java/io/micronaut/inject/annotation/internal/KotlinDeprecatedTransformer.java
```diff
@@ -0,0 +1,43 @@
...
+import java.lang.annotation.Annotation;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Allows treating the Kotlin deprecated annotation as the Java one.
+ *
+ * @since 4.0.0
+ */
+public final class KotlinDeprecatedTransformer implements NamedAnnotationTransformer {
```
dstepanov, line 31: Why do we need to remap the deprecated annotation?

core-processor/src/main/java/io/micronaut/inject/ast/utils/EnclosedElementsQuery.java
```diff
@@ -51,7 +53,7 @@
 @Internal
 public abstract class EnclosedElementsQuery<C, N> {
 
-    private final Map<N, io.micronaut.inject.ast.Element> elementsCache = new HashMap<>();
+    private final Map<N, Element> elementsCache = new ConcurrentLinkedHashMap.Builder<N, Element>().maximumWeightedCapacity(200).build();
```
dstepanov, line 56: Is it accessed in parallel?

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -4662,6 +4702,14 @@ public boolean isPostConstruct() {
         public boolean isPreDestroy() {
             return preDestroy;
         }
+
```
dstepanov: I don't find any usages if `isProperty` and `getProperty`

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1506,8 +1536,14 @@ public void visitPreDestroyMethod(TypedElement declaringType,
     public void visitMethodInjectionPoint(TypedElement declaringType,
                                           MethodElement methodElement,
                                           boolean requiresReflection,
-                                          VisitorContext visitorContext) {
-        MethodVisitData methodVisitData = new MethodVisitData(declaringType, methodElement, requiresReflection, methodElement.getAnnotationMetadata());
+                                          PropertyElement propertyElement, VisitorContext visitorContext) {
```
dstepanov: Please check if PropertyElement is needed here

inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/AbstractKotlinCompilerSpec.groovy
graemerocher: yeah actually I was wondering if you could help out get this test passing, my understanding of Kotlin generics was not enough to succeed. Doesn't have to be part of this PR

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/annotation/KotlinAnnotationMetadataBuilder.kt
graemerocher: yes you can get them from an annotation, but not from an annotation type (the `ClassDeclaration`)

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -324,17 +331,17 @@ protected boolean visitPropertyWriteElement(BeanDefinitionVisitor visitor,
...
     protected boolean visitMethod(BeanDefinitionVisitor visitor, MethodElement methodElement) {
-        if (visitInjectAndLifecycleMethod(visitor, methodElement)) {
+        if (visitInjectAndLifecycleMethod(visitor, null, methodElement)) {
             return true;
         }
         return visitAopAndExecutableMethod(visitor, methodElement);
     }
 
     @NextMajorVersion("Require @ReflectiveAccess for private methods in Micronaut 4")
-    private boolean visitInjectAndLifecycleMethod(BeanDefinitionVisitor visitor, MethodElement methodElement) {
```
dstepanov: Why do you need to work with the property instead of just method?

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -358,27 +365,36 @@ private boolean visitInjectAndLifecycleMethod(BeanDefinitionVisitor visitor, Met
...
             return true;
         }
-        if (!methodElement.isStatic() && isInjectPointMethod(methodElement)) {
+        if (!methodElement.isStatic() && isInjectPointMethod(propertyElement, methodElement)) {
             staticMethodCheck(methodElement);
             // TODO: Require @ReflectiveAccess for private methods in Micronaut 4
-            visitMethodInjectionPoint(visitor, methodElement);
+            if (propertyElement != null) {
+                // use the properties metadata for property elements
+                FieldElement fieldElement = propertyElement.getField().orElse(null);
```
dstepanov: Defining proper annotation metadata should be part of the property implementation

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -358,27 +365,36 @@ private boolean visitInjectAndLifecycleMethod(BeanDefinitionVisitor visitor, Met
...
+                if (fieldElement != null) {
```
dstepanov: Why do we need to work with field here? It should be handled by the field visiting

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -457,11 +473,22 @@ protected void applyConfigurationInjectionIfNecessary(BeanDefinitionVisitor visi
...
+     * @param memberElement   The method
      * @return true if it is
      */
-    protected boolean isInjectPointMethod(MemberElement memberElement) {
-        return memberElement.hasDeclaredStereotype(AnnotationUtil.INJECT);
+    protected boolean isInjectPointMethod(PropertyElement propertyElement, MemberElement memberElement) {
+        return memberElement.hasDeclaredStereotype(AnnotationUtil.INJECT) || isInjectableProperty(propertyElement);
+    }
+
+    private static boolean isInjectableProperty(@Nullable PropertyElement propertyElement) {
```
dstepanov: Why do we need to have custom property checks? Write method should have the same annotations.

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -324,17 +331,17 @@ protected boolean visitPropertyWriteElement(BeanDefinitionVisitor visitor,
...
         }
         return visitAopAndExecutableMethod(visitor, methodElement);
     }
 
     @NextMajorVersion("Require @ReflectiveAccess for private methods in Micronaut 4")
-    private boolean visitInjectAndLifecycleMethod(BeanDefinitionVisitor visitor, MethodElement methodElement) {
+    private boolean visitInjectAndLifecycleMethod(BeanDefinitionVisitor visitor, @Nullable PropertyElement propertyElement, MethodElement methodElement) {
         // All the cases above are using executable methods
         boolean claimed = false;
-        if (methodElement.hasDeclaredAnnotation(AnnotationUtil.POST_CONSTRUCT)) {
```
dstepanov: Here also property's annotation metadata should be properly propagated to the method so you don't need to use property at all

inject/src/main/java/io/micronaut/inject/annotation/MutableAnnotationMetadata.java
```diff
@@ -50,6 +50,15 @@ public class MutableAnnotationMetadata extends DefaultAnnotationMetadata {
     public MutableAnnotationMetadata() {
     }
 
+    MutableAnnotationMetadata(DefaultAnnotationMetadata annotationMetadata) {
```
dstepanov: You should use existing `of` method that can properly handle defaults and repeatable propagation

dstepanov (changes requested):

core-processor/src/main/java/io/micronaut/inject/annotation/internal/KotlinDeprecatedTransformer.java
graemerocher: because we have some logic that excludes configuration properties based on the annotation https://github.com/micronaut-projects/micronaut-core/blob/3cac50ae10d6ab84a64e6840fa53ea128bd833c8/core-processor/src/main/java/io/micronaut/inject/processing/ConfigurationReaderBeanElementCreator.java#L287

core-processor/src/main/java/io/micronaut/inject/ast/utils/EnclosedElementsQuery.java
graemerocher: No can probably be a normal map, although I used this one since it sets a limit to the size this map can grow

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
graemerocher: not sure what you are referring to here

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
graemerocher: does this refer to the correct line of code? *confused*

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
graemerocher: yeah I think it is always named `$Companion` in the byte code though

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
graemerocher: agreed

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
yawkat: it's not, see the link

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
graemerocher: Good question, these changes were part of the original PR from James. Will check if needed.

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
graemerocher: hmm good point, though this code is not from this PR all I did was delete `constructor != defaultConstructor `

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
graemerocher: I tried that but ran into conflicts with the Java getter name

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
graemerocher: looks like KSP uses collections for annotation values that are arrays in the AST

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
```diff
@@ -0,0 +1,857 @@
...
+            if (kotlinName != null) {
+                ksType = visitorContext.resolver.getKotlinClassByName(kotlinName)?.asStarProjectedType()
+                if (ksType != null) {
+                    if (kotlinType.starProjection().isAssignableFrom(ksType)) {
+                        return true
+                    }
+                }
+            }
+        }
+        return ksType?.isAssignableFrom(kotlinType) ?: false
```
yawkat: move into the if

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
yawkat: maybe we can change it in the ksp processor to arrays, better than having to support collections in the whole annotationvalue api

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinElementFactory.kt
```diff
@@ -0,0 +1,229 @@
...
+        var element = primitiveArrays[qualifiedName]
+        if (hasNoAnnotations && element != null) {
+            return element
+        }
+        if (qualifiedName == "kotlin.Array") {
+            val component = type.arguments[0].type!!.resolve()
+            val componentElement = newClassElement(component, elementAnnotationMetadataFactory, false)
+            return componentElement.toArray()
+        } else if (declaration is KSTypeParameter) {
+            KotlinGenericPlaceholderElement(declaration, elementAnnotationMetadataFactory, visitorContext)
```
yawkat:
```suggestion
            return KotlinGenericPlaceholderElement(declaration, elementAnnotationMetadataFactory, visitorContext)
```

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinFieldElement.kt
```diff
@@ -0,0 +1,95 @@
...
+    override fun isReflectionRequired(): Boolean {
+        return true // all Kotlin fields are private
+    }
+
+    override fun isReflectionRequired(callingType: ClassElement?): Boolean {
+        return true // all Kotlin fields are private
+    }
+
+    override fun isPublic(): Boolean {
+        return false // all Kotlin fields are private
```
yawkat: not true with `@JvmField`, see https://javap.yawk.at/#xulZhX/procyon

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinMethodElement.kt
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+
+    override fun overrides(overridden: MethodElement): Boolean {
+        val nativeType = kspNode()
+        val overriddenNativeType = overridden.kspNode()
+        if (nativeType == overriddenNativeType) {
+            return false
+        } else if (nativeType is KSFunctionDeclaration) {
+            return overriddenNativeType == nativeType.findOverridee()
```
yawkat, line 251: this only checks the direct parent, it doesnt work if youre overriding an overridden method

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinMethodElement.kt
```diff
@@ -0,0 +1,379 @@
...
+            }
+        } + ")"
+    }
+
+    override fun withParameters(vararg newParameters: ParameterElement): MethodElement {
+        return KotlinMethodElement(declaration, name, owningType, annotationMetadataFactory, visitorContext, returnType, newParameters.toList(), abstract, public, private, protected, internal)
+    }
+
+    override fun getThrownTypes(): Array<ClassElement> {
+        return emptyArray() // Kotlin doesn't support throws declarations
```
yawkat: there is `@Throws`

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+        val elements: MutableList<MemberElement> = ArrayList(3)
+        setter.ifPresent { elements.add(it) }
+        getter.ifPresent { elements.add(it) }
+        field.ifPresent { elements.add(it) }
+
+
+        // The instance AnnotationMetadata of each element can change after a modification
+        // Set annotation metadata as actual elements so the changes are reflected
+        // The instance AnnotationMetadata of each element can change after a modification
+        // Set annotation metadata as actual elements so the changes are reflected
```
yawkat: comment is double

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+        val propertyAnnotationMetadata: AnnotationMetadata
+        propertyAnnotationMetadata = if (elements.size == 1) {
+            elements.iterator().next()
+        } else {
+            AnnotationMetadataHierarchy(
+                true,
+                *elements.stream().map { e: MemberElement ->
```
yawkat:
```suggestion
                *elements.map { e: MemberElement ->
```

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+                    if (e is MethodElement) {
+                        return@map object : AnnotationMetadataDelegate {
+                            override fun getAnnotationMetadata(): AnnotationMetadata {
+                                // Exclude type metadata
+                                return e.getAnnotationMetadata().declaredMetadata
+                            }
+                        }
+                    }
+                    e
+                }.toList().toTypedArray()
```
yawkat:
```suggestion
                }.toTypedArray()
```

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+                elementAnnotationMetadataFactory: ElementAnnotationMetadataFactory,
+                visitorContext: KotlinVisitorContext,
+                excluded : Boolean = false) : super(getter, elementAnnotationMetadataFactory, visitorContext) {
+        this.name = name
+        this.type = type
+        this.exc = excluded
+        this.classElement = classElement
+        this.setter = Optional.ofNullable(setter)
+            .map { method ->
+                return@map visitorContext.elementFactory.newMethodElement(classElement, method, elementAnnotationMetadataFactory)
```
yawkat:
```suggestion
                visitorContext.elementFactory.newMethodElement(classElement, method, elementAnnotationMetadataFactory)
```

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+        this.field = Optional.empty()
+        val elements: MutableList<MemberElement> = ArrayList(3)
+        this.setter.ifPresent { elements.add(it) }
+        this.getter.ifPresent { elements.add(it) }
+        field.ifPresent { elements.add(it) }
+
+        // The instance AnnotationMetadata of each element can change after a modification
+        // Set annotation metadata as actual elements so the changes are reflected
+        // The instance AnnotationMetadata of each element can change after a modification
+        // Set annotation metadata as actual elements so the changes are reflected
```
yawkat: double comment

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+        val elements: MutableList<MemberElement> = ArrayList(3)
+        this.setter.ifPresent { elements.add(it) }
+        this.getter.ifPresent { elements.add(it) }
+        this.field.ifPresent { elements.add(it) }
+        this.exc = excluded
+
+        // The instance AnnotationMetadata of each element can change after a modification
+        // Set annotation metadata as actual elements so the changes are reflected
+        // The instance AnnotationMetadata of each element can change after a modification
+        // Set annotation metadata as actual elements so the changes are reflected
```
yawkat: double comment

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+                                // Exclude type metadata
+                                return e.getAnnotationMetadata().declaredMetadata
+                            }
+                        }
+                    }
+                    e
+                }.toList().toTypedArray()
+            )
+        }
+        this.annotationMetadata = object : MutableAnnotationMetadataDelegate<Any?> {
```
yawkat, line 341: these all do the same thing, maybe make an inner class?

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
```diff
@@ -0,0 +1,531 @@
...
+    override fun overrides(overridden: PropertyElement?): Boolean {
+        if (overridden == null) {
+            return false
+        } else {
+            val nativeType = kspNode()
+            val overriddenNativeType = overridden.kspNode()
+            if (nativeType == overriddenNativeType) {
+                return false
+            } else if (nativeType is KSPropertyDeclaration) {
+                return overriddenNativeType == nativeType.findOverridee()
```
yawkat, line 431: only works for one level of override

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
graemerocher: Will investigate

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
graemerocher: doesn't seem to work

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
graemerocher: Actually worked after removing stream()

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
graemerocher: my Kotlin-foo is failing me, if you know how to extract an inner class here let me know

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinMethodElement.kt
graemerocher: not sure how to fix this or if it is an issue

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinPropertyElement.kt
graemerocher: not sure how to fix this or if it is an issue

dstepanov (approved):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/beans/BeanDefinitionProcessor.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    override fun finish() {
+        try {
+            val outputVisitor = KotlinOutputVisitor(environment)
+            val processed = HashSet<String>()
+            for (beanDefinitionCreator in beanDefinitionMap.values) {
+                for (writer in beanDefinitionCreator.build()) {
+                    if (processed.add(writer.beanDefinitionName)) {
+                        processBeanDefinitions(writer, outputVisitor, processed)
+                    }                 }
```
yawkat:
```suggestion
                    }
                }
```

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
```diff
@@ -0,0 +1,328 @@
...
+        val dec = declaration.unwrap()
+        if (dec is KSDeclaration) {
+            val javaModifiers = visitorContext.resolver.effectiveJavaModifiers(dec)
+            return javaModifiers.mapNotNull {
+                when (it) {
+                    Modifier.ABSTRACT -> ElementModifier.ABSTRACT
+                    Modifier.FINAL -> ElementModifier.FINAL
+                    Modifier.PRIVATE -> ElementModifier.PRIVATE
+                    Modifier.PROTECTED -> ElementModifier.PROTECTED
+                    Modifier.PUBLIC -> ElementModifier.PUBLIC
```
yawkat: internal should also be public i think. but it depends on how we use getModifiers.

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
```diff
@@ -0,0 +1,871 @@
...
+            visitorContext,
+            value.isExcluded
+        )
+    }
+
+    override fun getSimpleName(): String {
+        var parentDeclaration = classDeclaration.parentDeclaration
+        return if (parentDeclaration == null) {
+            val simpleName = classDeclaration.simpleName.asString()
+            if (simpleName == "Any") "Object" else simpleName
```
yawkat: better check qualified name, people can still name their types Any

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
graemerocher: seems to be public only for the set of files currently being compiled, not sure there is a way to distinguish that

yawkat: yea, it has no java equivalent, but it's public in bytecode.

graemerocher: ok public it is then

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
```diff
@@ -0,0 +1,873 @@
...
+            elementAnnotationMetadataFactory,
+            visitorContext,
+            value.isExcluded
+        )
+    }
+
+    override fun getSimpleName(): String {
+        var parentDeclaration = classDeclaration.parentDeclaration
+        return if (parentDeclaration == null) {
+            if (classDeclaration.qualifiedName?.asString() == Any::class.java.name)
```
yawkat: is there a test for this? i think `Any::class.java.name` might just be `java.lang.Object`, while `classDeclaration.qualifiedName` might be `kotlin.Any` or `java.lang.Object`, not sure

yawkat (approved):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
yawkat: ah the test caught it :laughing:

graemerocher:
@jameskleeh going to merge this for now, if you get the chance to provide any feedback will address in subsequent PRs

PR #8463, opened 2022-12-06
Route processing refactor

yawkat:
This is a refactor of route execution, excluding filters (those will follow in #8422). The goals are:

- Moving away from reactive APIs where possible.
- Streamlining route execution so that it will be possible to have pre-route filters, and filters that read the request body.

This initial patch factors most of the routing code out of `RoutingInboundHandler` (the response writing code remains), and moves away from reactive for HttpContentProcessor and the route parameter fulfillment code (now called BaseRouteCompleter and FormRouteCompleter).

The goal of this PR is to get to a point where I am confident that it will be easy to delay filter execution until the body is available. Ideally the full request lifecycle, from before route resolution, through filter execution, to route execution, will be contained in the RouteRunner class.

edit, **For reviewers:** Much of the RouteExecutor code was moved to RequestLifecycle. Much of the RoutingInboundHandler code was moved to NettyRequestLifecycle and the two completer classes. Additionally, there is a new MicronautHttpData class that replaces netty's HttpData implementation, and HttpContentProcessor was refactored to not be reactive, which helps avoid concurrency considerations.


yawkat:
jackson-xml pr to fix up content processor: https://github.com/micronaut-projects/micronaut-jackson-xml/pull/235

yawkat:
servlet pr to fix up RouteExecutor dependency: https://github.com/micronaut-projects/micronaut-servlet/pull/363

yawkat:
This PR is now ready. The request routing code is now concentrated in RequestLifecycle. This should make it much easier to implement additional filtering behavior, such as reading the request body.

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyStreamingFileUpload.java
```diff
@@ -213,4 +207,24 @@ private void handleError(Throwable t) {
                         })
         ).flux();
     }
+
+    /**
+     * Factory for instances of {@link NettyStreamingFileUpload}. Wraps the fixed requirements that
+     * don't depend on request.
+     */
+    @Internal
+    public static final class Factory {
```
graemerocher: can be a record

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyCompletedFileUpload.java
```diff
@@ -85,21 +84,10 @@ public NettyCompletedFileUpload(FileUpload fileUpload, boolean controlRelease) {
...
-        } else {
-            File file = fileUpload.getFile();
-            if (file == null) {
-                throw new IOException("The input stream has already been released");
-            }
-            closeTracker();
-            return new NettyFileUploadInputStream(fileUpload, controlRelease);
-        }
+        InputStream s = ((MicronautHttpData<?>) fileUpload).toStream();
+        fileUpload.release(); // it's retained by toStream, and released by InputStream.close
```
graemerocher: does this need a `try/finally`?

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
```diff
@@ -0,0 +1,781 @@
...
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public void cleanAllHttpDatas() {
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    private class StreamImpl extends InputStream {
```
graemerocher: can be final

http-server-netty/src/main/java/io/micronaut/http/server/netty/AbstractBufferingHttpContentProcessor.java
```diff
@@ -1,113 +0,0 @@
...
-
-/**
- * Abtract implementation of the {@link HttpContentProcessor} interface that deals with limiting file upload sizes.
- *
- * @param <T> The type
- * @author Graeme Rocher
- * @since 1.0
- */
-@Internal
-public abstract class AbstractBufferingHttpContentProcessor<T> extends SingleThreadedBufferingProcessor<ByteBufHolder, T> implements HttpContentProcessor<T> {
```
graemerocher, line 37: This is used in Jackson XML, we will need to address that https://github.com/micronaut-projects/micronaut-jackson-xml/blob/1e6f2726de8bcf88cbb200ca7535e3cfa494b901/jackson-xml/src/main/java/io/micronaut/xml/jackson/server/XmlContentProcessor.java

http-server-netty/src/main/java/io/micronaut/http/server/netty/BaseRouteCompleter.java
```diff
@@ -0,0 +1,101 @@
...
+    public BaseRouteCompleter(NettyHttpRequest<?> request, RouteMatch<?> routeMatch) {
+        this.request = request;
+        this.routeMatch = routeMatch;
+    }
+
+    final void add(Object message) throws Throwable {
+        try {
+            if (request.destroyed) {
+                // we don't want this message anymore
+                return;
```
graemerocher, line 58: does it still need to be released?

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormRouteCompleter.java
```diff
@@ -0,0 +1,276 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
+
+final class FormRouteCompleter extends BaseRouteCompleter {
```
graemerocher, line 55: needs javadoc and annotate with `@Internal`

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
```diff
@@ -15,18 +15,57 @@
...
+ * Processors are stateful. They can receive repeated calls to {@link #add} with more data,
+ * followed by a call to {@link #complete} to finish up. Both of these methods accept a
+ * {@link Collection} {@code out} parameter that is populated with the processed items.
  *
- * @param <T> The type
  * @author Graeme Rocher
  * @since 1.0
  */
-public interface HttpContentProcessor<T> extends Publishers.MicronautPublisher<T>, Subscriber<ByteBufHolder>, Toggleable {
+public interface HttpContentProcessor extends Toggleable {
```
graemerocher, line 36: add `@Internal`

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * Extension of {@link HttpContentProcessorAsReactiveSubscriber} that is also a {@link Publisher}.
+ * This class is a bit more complicated because it has to deal with downstream demand, possibly
+ * concurrently.
+ *
+ * @since 4.0.0
+ * @author Jonas Konrad
+ */
+@Internal
+public class HttpContentProcessorAsReactiveProcessor extends HttpContentProcessorAsReactiveSubscriber implements Processor<ByteBufHolder, Object> {
```
graemerocher: can this be `final`?

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveSubscriber.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        done = true;
+        notifyContentAvailable(outBuffer);
+    }
+
+    final void requestContent() {
+        upstream.request(1);
+    }
+
+    abstract void notifyContentAvailable(Collection<Object> out);
```
graemerocher: needs javadoc

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
```diff
@@ -0,0 +1,781 @@
...
+     */
+    public Chunk pollChunk() {
+        if (pollIndex >= chunks.size()) {
+            return null;
+        }
+        Chunk chunk = chunks.get(pollIndex++);
+        if (pollIndex == chunks.size() && !completed) {
+            chunks.add(new Chunk(size));
+        }
+        chunk.retain();
```
graemerocher, line 148: where is this released?

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
```diff
@@ -0,0 +1,781 @@
...
+            } // if tryLock failed, the user already requested the chunk, we can't move it anymore.
+        }
+    }
+
+    private Path newTempFile() throws IOException {
+        Optional<File> location = factory.multipartConfiguration.getLocation();
+        if (location.isPresent()) {
+            return Files.createTempFile(location.get().toPath(), "FUp_", ".tmp");
+        } else {
+            return Files.createTempFile("FUp_", ".tmp");
```
graemerocher, line 270: Just checking we are not going to be hit with another CVE for creating temp files, this uses the secure version so I assume we are ok

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyPartData.java
```diff
@@ -93,6 +84,7 @@ public byte[] getBytes() throws IOException {
     public ByteBuffer getByteBuffer() throws IOException {
         ByteBuf byteBuf = getByteBuf();
         try {
+            // todo: should return a copy
```
graemerocher: can you address this todo?

graemerocher (changes requested):

graemerocher:
have you tested these changes underload for leaks / concurrency issues?

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
yawkat: sonar complained about this, but it's the same thing netty does, and openjdk 17 seems to create temp files with rwx------ permissions, so it should be fine

yawkat:
@graemerocher ive tested it under load, but the "risky" concurrency code is most likely to break in weird reactive setups with form data, and i can't guess when those may appear

http-server-netty/src/main/java/io/micronaut/http/server/netty/AbstractBufferingHttpContentProcessor.java
yawkat: https://github.com/micronaut-projects/micronaut-jackson-xml/pull/235

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
yawkat: it's used in the jackson xml module, it should not be internal

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
yawkat: in Chunk.claim usually, i will add a comment

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
yawkat: yes

yawkat:
i just noticed that this also requires changes in the aws module, i'm working on those now

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyPartData.java
```diff
@@ -91,12 +82,8 @@ public byte[] getBytes() throws IOException {
      */
     @Override
     public ByteBuffer getByteBuffer() throws IOException {
-        ByteBuf byteBuf = getByteBuf();
-        try {
-            return byteBuf.nioBuffer();
-        } finally {
-            byteBuf.release();
-        }
```
timyates, line 99: Do we still need to release this here?  Or is the releasing handled elsewhere?

Is it possible to have a test around multi-part file uploads to check we're not leaking?

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConvertersSpi.java
```diff
@@ -162,7 +162,9 @@ private TypeConverter<Attribute, CompletedPart> attributeToCompletedPartConverte
                     return Optional.empty();
                 }
 
-                return Optional.of(new NettyCompletedAttribute(object));
+                // converter does not claim the input object, so we need to retain it here. it's
+                // released by NettyCompletedAttribute.get*
+                return Optional.of(new NettyCompletedAttribute(object.retain()));
```
timyates, line 169: Why is this `retain` now required, when previously it wasn't? I can't see a change to NettyCompletedAttribute to handle this?

timyates (approved):
A couple of questions

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConvertersSpi.java
yawkat: i think it was a bug that never presented itself, probably because conversion was happening early enough. the new behavior is correct, because it is consistent with other converters. e.g. httpDataToStringConverter does not change the refCnt of the input. but NettyCompletedAttribute wants to release the object, so we need to retain it here

We have code that only releases if refCnt>0, so this kind of bug can be masked easily

but i just saw, the same applies to the fileupload code above, so i will add a retain there too.

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/NettyPartData.java
yawkat: it's released by getBytes

i think we have some multipart tests in http-server-netty (though some are also in the test suites). all tests in http-server-netty have leak detection

yawkat:
it's checkstyle every time

yawkat:
https://github.com/micronaut-projects/micronaut-aws/pull/1527

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConvertersSpi.java
yawkat: haha i think changing it in the fileupload code triggered leak detection. so i guess i have to look at this more after all.

yawkat: oh it's because NettyCompletedFileUpload constructor does a retain, but NettyCompletedAttribute does not. brilliant.

yawkat:
In the AWS module, I've replaced most of the custom request processing code with the existing code in RouteExecutor. They are mostly the same, but there are a few differences.

The biggest change is about uses of `request.getUri().toString()`. For netty requests, this is *usually* identical to the path of the request. However with the aws proxy module, it includes the request host, so any route processing that treats the uri as a path breaks.

I've adjusted a few of these use sites related to route selection to use `request.getPath()` in 4a040afefed934b36f77e696e46543754cf3fb57 and 686de0f0c1f7537f02528de840a218f2ec0b7bea.

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormRouteCompleter.java
```diff
@@ -0,0 +1,285 @@
...
+            request(1);
+            return;
+        }
+
+        Argument<?> argument = requiredInput.get();
+        Supplier<Object> value;
+        boolean isPublisher = Publishers.isConvertibleToPublisher(argument.getType());
+        boolean chunkedProcessing = false;
+
+        if (isPublisher) {
```
dstepanov, line 139: I was investigating this code before, and I think it's a bit strange that we would fill the publisher argument only when the data start to arrive. The correct would be to do some kind of analysis before we read the body and execute the route method immediately without waiting for the data.

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormRouteCompleter.java
```diff
@@ -0,0 +1,285 @@
...
+            }
+            Class<?> typeVariableType = typeVariable.getType();
+
+            Sinks.Many<Object> namedSubject = subjectsByDataName.computeIfAbsent(name, key -> makeDownstreamUnicastProcessor());
+
+            chunkedProcessing = PartData.class.equals(typeVariableType) ||
+                Publishers.isConvertibleToPublisher(typeVariableType) ||
+                ClassUtils.isJavaLangType(typeVariableType);
+
+            if (Publishers.isConvertibleToPublisher(typeVariableType)) {
```
dstepanov, line 161: Do we support cases of `SomePublisher<SomePublisher<Something>> argument`?

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
```diff
@@ -15,18 +15,57 @@
...
+    /**
+     * Process more data.
+     *
+     * @param data The input data
+     * @param out The collection to add output items to
+     */
+    void add(ByteBufHolder data, Collection<Object> out) throws Throwable;
```
dstepanov, line 43: Maybe it would be better to represent `Collection<Object>` as a custom object:
```java
HttpContent hc = processor.create();
processor.add(data, hc);
processor.complete(hc);
```

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * Extension of {@link HttpContentProcessorAsReactiveSubscriber} that is also a {@link Publisher}.
+ * This class is a bit more complicated because it has to deal with downstream demand, possibly
+ * concurrently.
+ *
+ * @since 4.0.0
+ * @author Jonas Konrad
+ */
+@Internal
+public final class HttpContentProcessorAsReactiveProcessor extends HttpContentProcessorAsReactiveSubscriber implements Processor<ByteBufHolder, Object> {
```
dstepanov: Are you sure this cannot be written with Reactor's sinks? It's pretty robust and can handle complicated scenarios like buffering etc.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -414,7 +414,7 @@ public RouteMatch<?> getMatchedRoute() {
     @Internal
     public void addContent(ByteBufHolder httpContent) {
         httpContent.touch();
-        if (httpContent instanceof AbstractHttpData || httpContent instanceof MixedAttribute) {
+        if (httpContent instanceof AbstractHttpData || httpContent instanceof MixedAttribute || httpContent instanceof MicronautHttpData<?>) {
```
dstepanov: With recent talks about the casttype performance problems we might want to review places with multiple `instanceof SomeinterfaceX || instanceof SomeinterfaceY` https://github.com/netty/netty/pull/12806

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
```diff
@@ -0,0 +1,283 @@
...
+        if (!HttpMethod.permitsRequestBody(request().getMethod())) {
+            return false;
+        }
+        if (!(nettyRequest.getNativeRequest() instanceof StreamedHttpRequest)) {
+            // Illegal state: The request body is required, so at this point we must have a StreamedHttpRequest
+            return false;
+        }
+        if (routeMatch instanceof MethodBasedRouteMatch<?, ?> methodBasedRouteMatch) {
+            if (Arrays.stream(methodBasedRouteMatch.getArguments()).anyMatch(argument -> MultipartBody.class.equals(argument.getType()))) {
+                // MultipartBody will subscribe to the request body in MultipartBodyArgumentBinder
```
dstepanov, line 171: Maybe we can refactor to have one place where we can read the body `HttpContentProcessor` now there is one in `MultipartBodyArgumentBinder`.

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final Object END = new Object();
```
dstepanov: I think the processors can be refactored to the following code:
```java
public static Publisher<ByteBufHolder> asPublisher(HttpContentProcessor<HttpContent> processor, NettyHttpRequest<?> request) {
        Processor<HttpContent, ByteBufHolder> customProcessor = new Processor<HttpContent, ByteBufHolder>() {

            final Sinks.Many<ByteBufHolder> sink = Sinks.many().unicast().onBackpressureBuffer();
            final Flux<ByteBufHolder> flux = sink.asFlux();
            Subscription subscription;

            @Override
            public void subscribe(Subscriber<? super ByteBufHolder> s) {
                flux.subscribe(s);
                flux.doOnRequest(demand -> subscription.request(demand));
            }

            @Override
            public void onSubscribe(Subscription s) {
                subscription = s;
            }

            @Override
            public void onNext(HttpContent httpContent) {
                // process and publish
            }

            @Override
            public void onError(Throwable t) {
                sink.emitError(t, Sinks.EmitFailureHandler.FAIL_FAST);
            }

            @Override
            public void onComplete() {
                sink.emitComplete( Sinks.EmitFailureHandler.FAIL_FAST);
            }
        };

        StreamedHttpMessage streamed = (StreamedHttpMessage) request.getNativeRequest();
        streamed.subscribe(customProcessor);
        return customProcessor;
    }
```

That will eliminate custom backpressure buffer implementation. WDYT @graemerocher?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
```diff
@@ -0,0 +1,283 @@
...
+    NettyRequestLifecycle(RoutingInBoundHandler rib, ChannelHandlerContext ctx, NettyHttpRequest<?> request) {
+        super(rib.routeExecutor, request);
+        this.rib = rib;
+        this.ctx = ctx;
+        this.nettyRequest = request;
+
+        multipartEnabled(rib.multipartEnabled);
+    }
+
+    void handleNormal() {
```
dstepanov, line 81: What is the point of having `method0` style?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
```diff
@@ -0,0 +1,283 @@
...
+            .filter(argument -> argument.getAnnotationMetadata().hasAnnotation(Body.class));
+        if (bodyArgument.isPresent() && !routeMatch.isSatisfied(bodyArgument.get().getName())) {
+            // Body argument in the method
+            return true;
+        }
+        // Might be some body parts
+        return !routeMatch.isExecutable();
+    }
+
+    private static class StreamingDataSubscriber implements Subscriber<ByteBufHolder> {
```
dstepanov, line 189: I think this can also be replaced with Reactor's many sink with backpressure buffer.

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/MultipartBodyArgumentBinder.java
```diff
@@ -117,19 +118,21 @@ protected void doOnNext(Object message) {
                         if (LOG.isTraceEnabled()) {
                             LOG.trace("Server received streaming message for argument [{}]: {}", context.getArgument(), message);
                         }
-                        if (message instanceof ByteBufHolder && ((ByteBufHolder) message).content() instanceof EmptyByteBuf) {
+                        if (message instanceof HttpData data) {
```
dstepanov: We might want to review this casttype cases too.

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
```diff
@@ -275,4 +257,41 @@ public void channelInactive(@NonNull ChannelHandlerContext ctx) throws Exception
...
+        WebsocketRequestLifecycle(RouteExecutor routeExecutor, HttpRequest<?> request, @Nullable RouteMatch<?> route) {
+            super(routeExecutor, request);
+            this.route = route;
+        }
+
+        ExecutionFlow<MutableHttpResponse<?>> handle() {
+            MutableHttpResponse<?> proceed = HttpResponse.ok();
+
+            if (route != null) {
+                request().setAttribute(HttpAttributes.ROUTE_MATCH, route);
```
dstepanov, line 276: We might want to have some `InternalHttpRequest` with `setRouteMatch` and `setRouteInfo` to treat these values as fields and avoid modifying the attributes map, with backward compatibility for `setAttribute/getAttribute`

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -269,63 +234,15 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
             ctx.writeAndFlush(new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.INTERNAL_SERVER_ERROR));
             return;
         }
-        routeExecutor.filterPublisher(new AtomicReference<>(nettyHttpRequest), () -> routeExecutor.onError(cause, nettyHttpRequest))
-            .onComplete((response, throwable) -> writeResponse(ctx, nettyHttpRequest, response, throwable));
+        new NettyRequestLifecycle(this, ctx, nettyHttpRequest).handleException(cause);
```
dstepanov, line 230: IMHO the design where `RoutingInBoundHandler` uses `NettyRequestLifecycle` and then it's using `RoutingInBoundHandler` is not the best one.

dstepanov (changes requested):

dstepanov:
I think we need to have more changes done if we want to support reading the request body in the pre-route filters. Probably introducing some internal structure that can represent the request body in different stages (fetched, reactive fetching) and different body contents (JSON, Form etc). That would likely eliminate `receivedContent` and `receivedData` from the `NettyHttpRequest`.

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormRouteCompleter.java
yawkat: we can't tell if a parameter of that name is present before the data comes in, so if we give a publisher immediately, we have to send the error to the publisher instead. this would also bypass eg error routes.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: agreed, it grew mostly out of how i did the refactoring (incrementally move code out of RIB).

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
yawkat: it won't work unfortunately, `.doOnRequest(demand -> subscription.request(demand));` is incorrect. since a processor may yield multiple items, downstream demand does not have to equal upstream demand

Sinks.Many does not expose api to check whether upstream demand is there, maybe could use a custom queue. but then concurrency becomes an issue since we cant use the reactor internal concurrency structure.

this could literally be replaced with a flatMap though, but it would also need completion/error handling, so it would have a bunch of nested publishers.

yawkat: this is very similar to SpreadProcessor in serde btw

yawkat:
@dstepanov yes more changes are needed. but i think they can be limited to RequestLifecycle and NettyRequestLifecycle now. All previous users of RouteExecutor (netty, netty websocket, jetty, aws) now use RL, and e.g. filter processing is encapsulated (no more calls to filterPublisher), so it should be possible to limit changes to RL.

However this PR is too big already. It will be in follow-up PRs.

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
yawkat: maybe we can just modify NettyHttpRequest to do special case these attributes with fields

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
dstepanov: It should work correctly, the sink is not going to push more items it’s just going buffer them.

yawkat: it will *work* but it leads to more upstream demand than necessary, which will lead to unnecessary buffering

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
yawkat: thats overkill imo. the api is the same as netty's ByteToMessageDecoder and it seems to work fine there

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
yawkat: no point, fixed.

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormRouteCompleter.java
dstepanov: I think we can because we can detect it fits somewhere. IMHO the correct behavior would be to publish the error in the publisher if the part is not detected or something, assuming it will be used to produce a response.

yawkat: it would be inconsistent with other parameter fulfillment, but maybe thats fine.

i don't want to touch it in this pr though.

dstepanov:
Going to approve it as nothing is blocking it. I feel like classes like `HttpContentProcessorAsReactiveProcessor` `FormRouteCompleter` `StreamingDataSubscriber` are a bit complex and hard to understand. Maybe we can revisit it after we have all the changes that are needed for Microanut 4.

dstepanov (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
dstepanov: Yeah, good point. But do we respect the requesting from the Netty content publisher?

yawkat:
im still testing some changes for your other comments, so please don't merge yet :)

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
yawkat: yea it translates to netty read calls which generates backpressure

yawkat: ive transformed it to flux only: 

```java
        StreamedHttpMessage streamed = (StreamedHttpMessage) request.getNativeRequest();
        return Flux.concat(Flux.from(streamed)
            .doOnError(e -> {
                try {
                    processor.cancel();
                } catch (Throwable ex) {
                    e.addSuppressed(ex);
                }
            })
            .concatMap(c -> {
                try {
                    List<T> out = new ArrayList<>(1);
                    processor.add(c, (List) out);
                    return Flux.fromIterable(out);
                } catch (Throwable e) {
                    return Flux.error(e);
                }
            }), Flux.defer(() -> {
            try {
                List<T> out = new ArrayList<>(1);
                processor.complete((List) out);
                return Flux.fromIterable(out);
            } catch (Throwable ex) {
                return Flux.error(ex);
            }
        }));
```

This is a lot simpler, but it needs 3 nested fluxes. WDYT?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: i moved one method (findFile) to NRL. The other call to the RIB is `writeResponse`. it is nice to have a separation between the request and response handling logic. maybe it doesnt make sense though. i am keeping it in RIB for now, if only to not blow up this pr further.

there are also various fields on RIB that are accessed, most come from the app context (eg the io executor, or the route executor). these could be passed in individually into the constructor. however that is not worth it as long as there is a link to RIB anyway for response writing.

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketUpgradeHandler.java
yawkat: out of scope for this pr

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
dstepanov: Maybe you can use Flux.using? I’m a bit confused why do we need to call complete…

Anyway it might be a bit to reactive but might save us from potential bugs, we would need to profile some basic cases.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: we can probably get rid of the AbstractHttpData and MixedAttribute checks, since all httpdata extends MicronautHttpData. i will try that.

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessorAsReactiveProcessor.java
yawkat: complete is used e.g. by the xml processor, it accumulates all data and only fires it downstream on the call to complete.

i dont think `using` will help.

maybe the perf hit is not so bad. this processor mostly exists for places where reactive is unavoidable, e.g. when the user has a publisher parameter anyway. in the long term we should strive to not use reactive when the user doesnt want reactive anyway, so this wouldnt be hit.

im going to switch to the flatmap approach for now, we can always dig up the old code if it becomes perf critical. i dont want to microbenchmark them against each other right now.

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/MultipartBodyArgumentBinder.java
yawkat: the processor used here only returns HttpData anyway, so ive removed the checks.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
yawkat: i don't think it's necessary, this one is simple enough.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
yawkat: agreed, but out of scope for this pr

http-server-netty/src/main/java/io/micronaut/http/server/netty/FormRouteCompleter.java
yawkat: yes, when there are multiple parts of the same name. there are tests in the test suite.

yawkat:
@graemerocher just finished it, all comments from @dstepanov addressed

graemerocher (approved):

dstepanov (approved):

PR #8469, opened 2022-12-08
Add tests and refine behaviour of registerSingleton

graemerocher:
There are a number of inconsistencies with `registerSingleton`. Currently the following problems exist:

* Calling `registerSingleton` when the bean type and qualifier match an existing bean type and qualifier doesn't override the bean (it does in Micronaut 3.x but probably shouldn't)
* Calling `registerSingleton` twice consecutively overrides the previously defined bean if the type and qualifier match
* If the registered type and the type of the instance differ (say `Foo` as the bean type but `FooImpl` for the implementation) then `getBeansOfType(Foo.class)` correctly contains the bean and `getBeansOfType(FooImpl.class)` correctly doesn't contain the bean. However the bean can be incorrectly located by `findBean(FooImpl.class)`

This PR tries to address these issues in the following ways:

* Calling `registerSingleton` always adds a new bean and never overrides
* If you want to use `registerSingleton` for bean replacement then you must call `replaces(TypeToReplace)` using the `RuntimeBeanDefinition` API thus formalizing the way to replace a bean.
* Beans that are registered with a particular type (`Foo`) and a particular impl (`FooImpl`) cannot be located by either `getBeansOfType(FooImpl)` or any bean lookup methods like `getBean`, `findBean` etc.
* Also improves handling of runtime beans with generics

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -730,7 +730,7 @@ public class DefaultBeanContext implements InitializableBeanContext {
         BeanDefinition<T> beanDefinition;
         if (inject && running.get()) {
             // Bean cannot be injected before the start of the context
-            beanDefinition = findBeanDefinition(type, qualifier).orElse(null);
+            beanDefinition = findConcreteCandidate(null, Argument.of(type), qualifier, false).orElse(null);
             if (beanDefinition == null) {
                 // Purge cache miss
                 purgeCacheForBeanInstance(singleton);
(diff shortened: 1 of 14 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1689,6 +1675,25 @@ private <B> void purgeCacheForBeanType(Class<B> beanType) {
         containsBeanCache.entrySet().removeIf(entry -> entry.getKey().beanType.isAssignableFrom(beanType));
     }
 
+    /**
+     * The definition to remove.
+     * @param definition The definitino to remove
```
sdelamo:
```suggestion
     * @param definition The definition to remove
```

sdelamo (approved):

graemerocher:
Merged for now since it is blocking some other PRs, @dstepanov if you spot anything when you get back let us know

dstepanov:
@graemerocher The important change here is that the factory method exposes only the return type now, so injecting the implementation wouldn't work. I think we need to add a test and describe the case in the breaking changes.

PR #8471, opened 2022-12-08
refactor: Extract http headers utils

sdelamo:
This PR extracts the HTTP Header logging utility with masking to its own util class. Thus, allowing us, or the framework users, to use it in other places.

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/netty/DefaultClientHeaderMaskTest.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/netty/DefaultClientHeaderMaskTest.groovy
@@ -2,7 +2,13 @@ package io.micronaut.http.client.netty
 
 import ch.qos.logback.classic.Level
 import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
 import io.micronaut.http.client.HttpClient
+import io.micronaut.runtime.server.EmbeddedServer
+import jakarta.inject.Singleton
 import org.slf4j.Logger
 import ch.qos.logback.classic.spi.ILoggingEvent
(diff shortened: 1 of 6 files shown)


yawkat (approved):

http-client/src/test/groovy/io/micronaut/http/client/netty/DefaultClientHeaderMaskTest.groovy
```diff
@@ -1,42 +1,29 @@
 package io.micronaut.http.client.netty
```
timyates, line 2: Do we need this test anymore?

We used to test that the actual client masked logging, but now this just tests the utils class

Which we also test in [http/src/test/groovy/io/micronaut/http/util/HttpHeadersUtilSpec.groovy](https://github.com/micronaut-projects/micronaut-core/pull/8471/files#diff-75a2657025e80f8e9f2d653092d04a309f987455dc0bf67bc91b005d1babfb1e)

sdelamo: do you prefer if I delete it?

timyates: I'm wondering if we should put it back how it was to check the real client is calling the masking utils?

sdelamo: how about https://github.com/micronaut-projects/micronaut-core/pull/8471/commits/dc42add597516c21f7548d6f60f35f84cc93375c ?

timyates: How about https://github.com/micronaut-projects/micronaut-core/pull/8471/commits/86b52a10ffe1b52e0c471f06918f4739b309cef6 ?

To avoid the Stub and the private method?

timyates (approved):

PR #8473, opened 2022-12-09
test: refactor CorsFilterSpec as a black box

sdelamo:
This refactors the `CorsFilterSpec` to test it as a black box instead of testing internal methods such as `handleRequest`. I will send a separate PR to refactor `CorsFilter`.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -16,19 +16,28 @@
 package io.micronaut.http.server.netty.cors
 
 import io.micronaut.context.ApplicationContext
+import io.micronaut.core.annotation.Nullable
+import io.micronaut.core.async.publisher.Publishers
+import io.micronaut.core.util.StringUtils
 import io.micronaut.http.*
 import io.micronaut.http.annotation.Controller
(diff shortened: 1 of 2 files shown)


timyates (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
```diff
@@ -112,198 +133,251 @@ class CorsFilterSpec extends Specification {
...
     }
 
     void "test preflight handleRequest with disallowed header"() {
         given:
-        HttpRequest request = Mock(HttpRequest)
-        HttpHeaders headers = Mock(HttpHeaders)
-        request.getHeaders() >> headers
-        def config = new HttpServerConfiguration.CorsConfiguration()
+        String origin = 'http://www.foo.com'
+        HttpHeaders headers = Stub(HttpHeaders) {
```
yawkat, line 195: do we need a mock here, or is there an implementation we can use?

yawkat (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
sdelamo: In other places, we use `MockHttpHeaders` but I think in this test the Stub is ok.

PR #8482, opened 2022-12-13
Update upgrade documentation

anuragdeshpande:
The gradle plugin version does not match the required version for the project to be configured properly. Correcting the version to the one that is shown in the public launch.micronaut website. A screenshot is attached for reference

image


sdelamo:
@anuragdeshpande Instead of listing a version number, I have linked to the latest github release and the plugin docs.

PR #8485, opened 2022-12-14
Allow context annotations at build time

timyates:
Otherwise we get failures with Primary when compiled under Graal


timyates:
~~@graemerocher Should this go in~~

~~https://github.com/micronaut-projects/micronaut-core/blob/79c56c65b40ca361904f2a723041379b15c7f26f/context/src/main/resources/META-INF/native-image/io.micronaut/micronaut-context/native-image.properties~~

~~Instead?~~

Sorry, on second glance, this is the right place

PR #8486, opened 2022-12-14
Overload DELETE method for URI object

### LINKED ISSUE #8483: Add HttpRequest::DELETE(URI) method
[HttpRequest has  ` static <T> MutableHttpRequest<T> DELETE(String uri)`](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http/src/main/java/io/micronaut/http/HttpRequest.java#L387). 

It would be great to have `static <T> MutableHttpRequest<T> DELETE(URI uri)` as well. 

We have methods taking a `URI` for other HTTP verbs. e.g. https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http/src/main/java/io/micronaut/http/HttpRequest.java#L209

wetted:
closes #8483

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/HttpDeleteSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/HttpDeleteSpec.groovy
@@ -45,9 +45,17 @@ class HttpDeleteSpec extends Specification {
     @Inject
     MyDeleteClient myDeleteClient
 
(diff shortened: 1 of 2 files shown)


wetted:
SonarCloud fails because it doesn't like a method named in uppercase, i.e. DELETE() but we're already matching names/case of HTTP verbs.

sdelamo (approved):

PR #8489, opened 2022-12-15
Backport fixes from 4.0.x and fix ConvertibleValuesDeserializer

graemerocher:
- Don't include write only properties in serialization (#8457)
- Add test for unwrap problems


graemerocher:
@yawkat could you look into why https://github.com/micronaut-projects/micronaut-core/pull/8489/files#diff-0f0ea9c486ca7618687c9a9e147505ba5bb09d770cec2f7673f460e8031d8f95R70 is failing

PR #8490, opened 2022-12-15
Don't apply AOP advice to Groovy synthetic getters/setters

graemerocher:
Found while working on the KSP module. AOP advice is currently being applied to Groovy properties when AOP advice is on the type.


timyates (approved):

graemerocher:
ok seems I was wrong and this is applied by default to synthetics..

PR #8484, opened 2022-12-14
fix configuration inheritance path calculation

graemerocher:
Fixes #8480

PR #8494, opened 2022-12-16
fix: TextStreamCodec requires bean of type ByteBufferFactory

sdelamo:
Otherwise, you get: 

```
Error starting Micronaut container: Bean definition [io.micronaut.http.server.codec.TextStreamCodec] could not be loaded: Failed to inject value for parameter [byteBufferFactory] of class: io.micronaut.http.server.codec.TextStreamCodec?
```

Already in 4.0.x via: https://github.com/micronaut-projects/micronaut-core/pull/8484


graemerocher:
I had already applied this here https://github.com/micronaut-projects/micronaut-core/commit/eddefb836a4eea3809d01790149807e8833a34bc#diff-bb7c25ca689fadb537c650bd76b524cc1ab30c3461ae733ee6c01bf0599be5f4R58

PR #8498, opened 2022-12-19
Fix delegating to intercepted method in Kotlin

dstepanov:
Corrected some casting because in some cases Kotlin will generate a method with primitive type instead of a generic-object one.

### DIFF
--- a/test-suite-kotlin/src/test/kotlin/io/micronaut/respository/CoroutineCrudRepository.kt
+++ b/test-suite-kotlin/src/test/kotlin/io/micronaut/respository/CoroutineCrudRepository.kt
@@ -1,0 +1,2 @@
+package io.micronaut.respository
+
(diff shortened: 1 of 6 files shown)


radovanradic (approved):

graemerocher (approved):

test-suite-kotlin/src/test/kotlin/io/micronaut/respository/CoroutineCrudRepository.kt
```diff
@@ -0,0 +1,2 @@
+package io.micronaut.respository
```
sdelamo: is this file intentionally empty?

dstepanov: Ah yeah, deleted. Please merge after.

yawkat (approved):

dstepanov:
Please merge

PR #8499, opened 2022-12-19
feat: Add HTTP Server TCK

sdelamo:
Adds a new module `http-server-tck`

This module contains multiple tests for a Micronaut HTTP Server.  Those tests were ported from the AWS Module.

It has an API `ServerUnderTest` and `ServerUnderTestProvider` which allows registering a module, implementing the TCK easily via a service loader.

It contains a test module `test-suite-http-server-tck-netty` which runs the TCK against `micronaut-http-server-netty`.

- [Draft PR for using the TCK in Servlet](https://github.com/micronaut-projects/micronaut-servlet/pull/369)
- [Draft PR for using the TCK in Azure HTTP Functions](https://github.com/micronaut-projects/micronaut-azure/pull/412)
- [Draft PR for using the TCK in AWS Lambda](https://github.com/micronaut-projects/micronaut-aws/pull/1538)


yawkat (commented):
the whole test harness seems a bit convoluted. couldnt these be normal junit tests that access the server implementation using a rule or something?

sdelamo:
> the whole test harness seems a bit convoluted. couldnt these be normal junit tests that access the server implementation using a rule or something?

Not sure what you are proposing. Please,  create a PR with this one as the base with your proposal

See linked PRs, specially in azure and lambda to see `ServerUnderTest` implementations for non `EmbeddedServer` solutions.

yawkat:
i dont mean the ServerUnderTest necessarily, i just want the tests to not be default methods in interfaces. e.g. by configuring junit to run the tests from the main tck module in the netty module.

@melix what do you think, is there a better way to achieve this?

dstepanov:
Why can't we have ordinary Spock tests as we have before? It should have been enough to copy existing Netty tests, and different implementations would provide a server implementation.

dstepanov:
I feel like you created some testing framework when it's not needed at all.

melix:
It should be possible to have tests exposed as a transitive dependency. By default Gradle will only add the tests from `src/test` on test classpath, but we can setup things differently, so that a test suite is added. The thing is how you get the thing under test injected into those tests. We can probably have several options here (like a static setter, or a Java service loader).

yawkat:
@melix getting the right server implementation injected is already solved using ServiceLoader in this PR. i think that approach is fine.

melix:
Right so it should only be a matter of:

- having a test suite exposed as a consumable dependency
- having that test suite injected as a dependency on the test classpath of the projects under test

I think we already do something like this in `micronaut-data`.

dstepanov:
There are a few ways:
- Create abstract tests like Micronaut Data and extend them in the actual implementation, something like `NettyFormSpec extends AbstractFormSpec`
- Share tests with a hack just like in TechEmpower project https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Java/micronaut/buildSrc/src/main/groovy/io.micronaut.benchmark.module.gradle

yawkat:
extending each test is not workable here, since we will have tck implementations in different repositories. it's too easy to miss a test.

sdelamo:
The requisite is that we need to publish a library with tests. And use that library in different repositories (servlet, azure, azure).

I think it is safer to publish a standard java library with interfaces and consume it s proposed in this PR than complicating the build. 

To run the TCK means just creating a class which implements an interface and a `ServerUnderTestProvider` implementation. 

See https://github.com/micronaut-projects/micronaut-aws/pull/1538/files#diff-8f69b875a6148983a3f6bbbdb3d29a4883825e685afaeea7a35dd9f3b3d20c72

Moreover, if you are not compliant with a test you can Disable the test see above example. 

Honestly, I think it is a good solution it makes trivial to implement it in other modules. See linked PRs.

About creating a testing framework. 

The test is in this PR are based on the tests that were in the AWS Module. I had to rewrite them so that they are HTTP server framework agnostic. I wrote helper classes (TestScenario, HTTPResponseAssertion) to remove a lot of duplication and to handle things such as bean pollution. We had a lot of duplication and tests difficult to read and these classes allow you to write HTTP Server tests as: 

```
TestScenario.builder()
.specName("baba")
.request(....)
.assertion()
.run()
```

But using `TestScenario` is optional. internally uses the same concepts we use through the framework. HTTP Request, exchanges, assertions of status, body, headers.

sdelamo:
The AWS PR is a good example of consuming the library: 

https://github.com/micronaut-projects/micronaut-aws/pull/1538/files#

Right now, it has `http-server-tck` duplicated as well. But once this gets published in core, I will remove `http-server-tck` in that PR.

yawkat:
i think it is better to have to do small modifications to the build in dependent modules, if it means we can use normal testing patterns: classes instead of interfaces, automatic discovery instead of registering each test class, maybe even spock, and maybe even being able to click the green arrow in intellij to run a test...

i see that in the aws pr, you disabled an individual test, which is indeed more convenient with your approach. however you could also do this with a normal exclusion rule.

imo you should work with @melix to find a good approach to keep the build as clean as possible but still allow normal testing patterns. (the techempower approach seems hacky, i dont know)

sdelamo:
> and maybe even being able to click the green arrow in intellij to run a test...

If you want to run a single test in your tck. You create a class and `class AWSHelloTest implement HelloWorldTest` and click the green button. That works fine or even rerun only the failed tests in the TCK.

sdelamo:
> , maybe even spock,

I like Spock. However, I think it is easier to maintain and to publish a library based on JUnit than in Spock.

melix:
I agree with @yawkat that we should publish a test suite. It shouldn't really be complicated to do it, and it can use Spock too. Consuming shouldn't be complicated either. Especially if the consumers are only Micronaut projects.

graemerocher:
Would like to point out that there is one other advantage of using JUnit/Java in that we can run these TCK tests with native image.

sdelamo:
>  if it means we can use normal testing patterns: classes instead of interfaces, automatic discovery instead of registering each test class, maybe even spock, and maybe even being able to click the green arrow in intellij to run a test...

[I have changed the PR to use `@Suite`](https://github.com/micronaut-projects/micronaut-core/pull/8499/commits/9dd45deac8f0e0a6fc0c633a5a80a476fe83b920). Let me know if this addresses your concerns.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyArgumentTest.java
```diff
@@ -0,0 +1,65 @@
...
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
+import io.micronaut.http.annotation.Produces;
+import io.micronaut.http.server.tck.AssertionUtils;
+import io.micronaut.http.server.tck.HttpResponseAssertion;
+import io.micronaut.http.server.tck.TestScenario;
+import org.junit.jupiter.api.Test;
+import java.io.IOException;
+
+@SuppressWarnings({
```
yawkat, line 32: maybe we can turn of sonar/checkstyle for these modules? we don't use them for spock tests either, they're just gonna be a pita...

yawkat (approved):
thanks!

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyTest.java
```diff
@@ -0,0 +1,173 @@
...
+    "checkstyle:MissingJavadocType",
+    "checkstyle:DesignForExtension"
+})
+public class BodyTest {
+    public static final String SPEC_NAME = "BodyTest";
+
+    @Test
+    void testCustomBodyPOJO() throws IOException {
+        TestScenario.builder()
+            .specName("BodyTest")
```
dstepanov: Use constant

http-server-tck/src/main/java/io/micronaut/http/server/tck/AssertionUtils.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Utility class used to perform assertions.
+ * @author Sergio del Amo
+ * @since 3.8.0
+ */
+@SuppressWarnings({
+    "java:S5960", // We're allowed assertions, as these are used in tests only
+})
+@Experimental
+public final class AssertionUtils {
```
dstepanov, line 44: This module should be internal

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyArgumentTest.java
dstepanov: Agree

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/FilterErrorTest.java
```diff
@@ -0,0 +1,351 @@
...
+    "checkstyle:MissingJavadocType",
+    "checkstyle:DesignForExtension"
+})
+public class FilterErrorTest {
+    public static final String SPEC_NAME = "FilterErrorTest";
+
+    @Test
+    void testFilterThrowingExceptionHandledByExceptionHandlerThrowingException() throws IOException {
+        TestScenario.builder()
+            .specName("FilterErrorSpec3")
```
dstepanov: Why some of the spec names are constants and some not?

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/FiltersTest.java
```diff
@@ -0,0 +1,153 @@
...
+public class FiltersTest {
+    public static final String SPEC_NAME = "FiltersTest";
+
+    @Test
+    void testFiltersAreRunCorrectly() throws IOException {
+        Map<String, Object> configuration = CollectionUtils.mapOf(
+            "micronaut.server.cors.enabled", StringUtils.TRUE
+        );
+        try (ServerUnderTest server = ServerUnderTestProviderUtils.getServerUnderTestProvider().getServer(SPEC_NAME, configuration)) {
+            assertTrue(true);
```
dstepanov: Is it needed?

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
```diff
@@ -0,0 +1,287 @@
...
+import java.util.Collections;
+import java.util.Map;
+
+
+@SuppressWarnings({
+    "java:S5960", // We're allowed assertions, as these are used in tests only
+    "checkstyle:MissingJavadocType",
+    "checkstyle:DesignForExtension"
+})
+public class MiscTest {
```
dstepanov, line 47: I would separate tests into packages named after it's test scenarios: .filters, .reactive, .form etc

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
```diff
@@ -0,0 +1,287 @@
...
+    public static final String SPEC_NAME = "MiscTest";
+
+    /**
+     *
+     * @see <a href="https://github.com/micronaut-projects/micronaut-aws/issues/868">micronaut-aws #868</a>
+     */
+    @Test
+    void testSelectedRouteReflectsAcceptHeader() throws IOException {
+        TestScenario.builder()
```
dstepanov: Looking at this class there are a lot of repeatable code:
```
TestScenario.builder()
            .specName(SPEC_NAME)
```
...
```
run();
```
Maybe it we can have a helper method to look like:
```java
assert(HttpRequest.GET("/bar/ok").header(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON), (server, request) -> AssertionUtils.assertDoesNotThrow(server, request, HttpResponseAssertion.builder()
                .status(HttpStatus.OK)
                .body("{\"status\":\"ok\"}")
                .build())
```

test-suite-http-server-tck-netty/src/test/resources/logback.xml
```diff
@@ -0,0 +1,11 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="info">
+        <appender-ref ref="STDOUT" />
+    </root>
+    <logger name="io.micronaut.http.client" level="TRACE"/>
```
dstepanov: Might cause a lot of noise in logs

dstepanov (changes requested):

http-server-tck/src/main/java/io/micronaut/http/server/tck/AssertionUtils.java
sdelamo: No, other modules may want to use it.

test-suite-http-server-tck-netty/src/test/resources/logback.xml
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/8499/commits/28462ad10ad23ef902662c84ebd29b87ea70c3e7

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/FiltersTest.java
sdelamo: removed

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/FilterErrorTest.java
sdelamo: extracted constants

http-server-tck/src/main/java/io/micronaut/http/server/tck/AssertionUtils.java
dstepanov: Yes, but not outside of Micronaut org

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/8499/commits/1f50932832c6ce88376603f23676584046e618e4

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
sdelamo: I agree we could improve how the TCK is structured. We can do this after this his been merged. The tests were ported from AWS, where there was no clear structure.

dstepanov (approved):

PR #8502, opened 2022-12-20
build:  bump native-maven-plugin to 0.9.19

sdelamo:
see: https://github.com/graalvm/native-build-tools/releases/tag/0.9.19


alvarosanchez:
@melix despite #8399 this is still not being picked up by Renovate. I checked the [Dependency Dashboard](https://github.com/micronaut-projects/micronaut-core/issues/7293) but it's not rate-limited.

I noticed #8370, but that's something else.

melix:
@alvarosanchez it is the same thing: our `libs.versions.toml` only contains a _version_ for the plugin. There are no GAV coordinates associated with it, so there's no way renovate can figure out it has to upgrade it. If we want automatic upgrade we must add a dummy entry in the `[libraries]` section with the coordinates of that library.

alvarosanchez:
I did add an entry: https://github.com/micronaut-projects/micronaut-core/pull/8399/files

melix:
Right, but it's not on the main branch yet: https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/gradle/libs.versions.toml

alvarosanchez:
Aaaaaah ok, I keep forgetting that 4.0.x is the default branch.

Then we are (will be?) good to go.

alvarosanchez (approved):

PR #8521, opened 2022-12-22
Use `ConversionService` from bean context for `InterceptedMethod`

dstepanov:
+ Some improvements for bean definition writer to use `BeanContext#getConversionService` to bypass extracting it from the context

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
@@ -2109,7 +2109,7 @@ public class BeanDefinitionWriter extends AbstractClassFileWriter implements Bea
             writer.loadArg(INSTANTIATE_METHOD_BEAN_CONTEXT_PARAM);
             return true;
         }
-        if (visitorContext.getClassElement(ConversionService.class).orElseThrow().isAssignable(type)) {
+        if (visitorContext.getClassElement(ConversionService.class).orElseThrow().equals(type)) {
             // We only want to assign to exact `ConversionService` classes not to classes extending `ConversionService`
             writer.loadArg(INSTANTIATE_METHOD_BEAN_CONTEXT_PARAM);
             writer.invokeInterface(TYPE_BEAN_CONTEXT, METHOD_BEAN_CONTEXT_GET_CONVERSION_SERVICE);
--- a/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
@@ -50,6 +50,8 @@ import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.beans.BeanConstructor;
+import io.micronaut.core.convert.ConversionService;
+import io.micronaut.core.convert.ConversionServiceProvider;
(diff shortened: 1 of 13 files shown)


core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -2087,6 +2099,32 @@ private int pushGetValueForPathCall(GeneratorAdapter injectMethodVisitor, ClassE
...
+        // Used in instantiate and inject methods
+        if (type.isAssignable(BeanResolutionContext.class)) {
+            writer.loadArg(INSTANTIATE_METHOD_BEAN_RESOLUTION_CONTEXT_PARAM);
+            return true;
+        }
+        if (type.isAssignable(BeanContext.class)) {
+            writer.loadArg(INSTANTIATE_METHOD_BEAN_CONTEXT_PARAM);
+            return true;
+        }
+        if (visitorContext.getClassElement(ConversionService.class).orElseThrow().isAssignable(type)) {
```
yawkat: `.isAssignable(type)` would also be `true` for `Object`

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -2087,6 +2099,32 @@ private int pushGetValueForPathCall(GeneratorAdapter injectMethodVisitor, ClassE
         return optionalInstanceIndex;
     }
 
+    private boolean pushValueBypassingBeanContext(GeneratorAdapter writer, ClassElement type) {
```
yawkat, line 2102: Is the purpose of this shortcut just optimization, or does it have a behavioral difference vs `context.getBean(ConversionService.class)`? If the purpose of this is behavioral, it seems clearer to just do `beanContext.getConversionService()` where it's necessary, instead of changing the codegen

yawkat (approved):

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
dstepanov: Well, we have only one `ConversionService` (resolved exactly by this type) so if you want to inject it, there is no point in going thru the bean context resolution.

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
dstepanov: Hmm right

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
yawkat: so it's just an optimization? in that case i think it's fine.

my concern was that if this is a behavior difference (e.g. getBean doesnt work early in initialization or something), then this code would have to be expanded for every type where a similar case comes up.

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
dstepanov: Fixed

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
dstepanov: Just optimization

sdelamo (approved):

PR #8530, opened 2022-12-23
Fix `@EachBean` with `@Replaces` and a missing class

dstepanov:
Fixes problem in Micronaut Data


graemerocher (approved):

dstepanov:
Unrelated failure

PR #8553, opened 2022-12-30
Support Range requests when returning SystemFiles

### LINKED ISSUE #7911: Support for HTTP 206 Partial Content for serving SystemFile
### Feature description

Current `NettySystemFileCustomizableResponseType` implementation does not support partial content. 
Its support may be useful for building something like a video streaming service.
I made my own implementation of `NettyFileCustomizableResponseType`  based on copy-paste from `NettySystemFileCustomizableResponseType`
https://gist.github.com/Ksnz/b2a223262514fe77218d481aa7ed6d96 
And now endpoint wrapped into it like
```
    @Get(value = "{id}", produces = MediaType.ALL)
    public ByteRangeSupportNettySystemFileCustomizableResponseType fetchPreview(@PathVariable UUID id)
    {
        ...
        return new ByteRangeSupportNettySystemFileCustomizableResponseType(new SystemFile(file));
    }
```
supports requests with byte-range.

My implementation isn't perfect because I just need to solve a problem and have no clue how to properly deal with invalid range exceptions.
So it requires some refactoring to be used in framework.

kgeri:
This is to add [Range](https://httpwg.org/specs/rfc9110.html#field.range) support for serving `java.io.File`s and `SystemFile`s, which is important when streaming videos or other large media (where eg. Chrome might decide to send multiple requests and jump around in the file in order to retrieve metadata used for seeking).

I don't seem to have permissions to link to an issue, but this should resolve #7911

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
@@ -17,8 +17,12 @@ package io.micronaut.http.server.netty.types.files;
 
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.util.SupplierUtil;
+import io.micronaut.http.HttpHeaders;
+import io.micronaut.http.HttpMethod;
 import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpStatus;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.http.netty.NettyMutableHttpResponse;
@@ -51,6 +55,8 @@ import java.io.RandomAccessFile;
 import java.util.Optional;
 import java.util.function.Supplier;
 
+import static io.micronaut.http.HttpHeaders.CONTENT_RANGE;
+
 /**
  * Writes a {@link File} to the Netty context.
  *
@@ -62,6 +68,7 @@ import java.util.function.Supplier;
 public class NettySystemFileCustomizableResponseType extends SystemFile implements NettyFileCustomizableResponseType {
 
     private static final int LENGTH_8K = 8192;
+    private static final String UNIT_BYTES = "bytes";
     private static final Logger LOG = LoggerFactory.getLogger(NettySystemFileCustomizableResponseType.class);
 
     protected Optional<FileCustomizableResponseType> delegate = Optional.empty();
@@ -99,7 +106,6 @@ public class NettySystemFileCustomizableResponseType extends SystemFile implemen
      */
     @Override
     public void process(MutableHttpResponse response) {
-        response.header(io.micronaut.http.HttpHeaders.CONTENT_LENGTH, String.valueOf(getLength()));
         delegate.ifPresent(type -> type.process(response));
     }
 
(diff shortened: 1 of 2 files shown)


sdelamo:
@kgeri can you target this PR with base `3.9.x` ?

http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
```diff
@@ -108,7 +114,33 @@ public ChannelFuture write(HttpRequest<?> request, MutableHttpResponse<?> respon
...
+            if (rangeHeader != null
+                && request.getMethod() == HttpMethod.GET // A server MUST ignore a Range header field received with a request method that is unrecognized or for which range handling is not defined.
+                && rangeHeader.startsWith(UNIT_BYTES) // An origin server MUST ignore a Range header field that contains a range unit it does not understand.
+                && response.status() == HttpStatus.OK // The Range header field is evaluated after evaluating the precondition header fields defined in Section 13.1, and only if the result in absence of the Range header field would be a 200 (OK) response.
+            ) {
+                IntRange range = parseRangeHeader(rangeHeader, fileLength);
+                if (range != null // A server that supports range requests MAY ignore or reject a Range header field that contains an invalid ranges-specifier (Section 14.1.1)
+                    && range.firstPos < range.lastPos // A server that supports range requests MAY ignore a Range header field when the selected representation has no content (i.e., the selected representation's data is of zero length).
```
yawkat, line 133: validate the range against actual file bounds

http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
```diff
@@ -108,7 +114,33 @@ public ChannelFuture write(HttpRequest<?> request, MutableHttpResponse<?> respon
...
+                ) {
+                    position = range.firstPos;
+                    contentLength = range.lastPos + 1 - range.firstPos;
+                    response.status(HttpStatus.PARTIAL_CONTENT);
+                    response.header(CONTENT_RANGE, String.format("%s %d-%d/%d", UNIT_BYTES, range.firstPos, range.lastPos, fileLength));
```
yawkat, line 140: only respond with user-controlled data if file range is valid (see above)

yawkat (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/types/files/NettySystemFileCustomizableResponseType.java
kgeri: Good point, thanks! Added test cases for outside-of-file range requests, and also a couple for non-numeric range requests, and fixed the code.

yawkat (approved):

kgeri:
@graemerocher: Static analysis breaks with `Error: eckstyle] [ERROR] /home/runner/work/micronaut-core/micronaut-core/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java:753:37: Variable access definition in wrong order. [DeclarationOrder]`. I can reproduce (and even fix) it locally, however that file was not touched in this PR, that line was last modified 5 months ago, so I'm a bit puzzled (maybe Checkstyle configs changed along the way?).

I see #6842 on 4.0.x [removes the `private` from that line](https://github.com/micronaut-projects/micronaut-core/pull/6842/files#diff-ae5ec3499c014b9efa0ec7b4b72092983d393bdb159a15f14f2cb1a8376a62c8L766), which I confirmed makes Checkstyle happy - let me know if I should do the same in this PR.

yawkat:
@kgeri i think checkstyle allows one violation per module. you don't need to fix the checkstyle error if it's not in your code.

the "static analysis" build step is failing because of sonarqube, not because of checkstyle, and it's only failing because it can't authenticate with sonar because you aren't in the micronaut-projects org. you don't need to take any action.

graemerocher:
Thanks for the contribution!

icode:
How to fix connection close exception and error writing final response: java.lang.IllegalStateException: cannot send more responses than requests?

PR #8559, opened 2023-01-02
HTTP/3 support

yawkat:
This PR adds HTTP/3 support for the server and client based on the experimental https://github.com/netty/netty-incubator-codec-http3 . The code is very similar to the existing HTTP/2 stack. The major differences relate to UDP vs TCP.

The HTTP/3 server is activated through a special listener protocol family `QUIC` (maybe this should be called `UDP`?). The client is activated by setting the `alpn-modes` to `[h3]` (switching to udp based on this property is a bit weird, but ALPN still exists with HTTP/3. eventually, falling back on tcp would be nice).

Server push is currently not supported. The test case for it in particular (netty-based) would need to be rewritten for UDP.

There are still a bunch of workarounds for netty bugs in this code. This is why this PR is a draft. The upstream PRs are:

- https://github.com/netty/netty-incubator-codec-http3/pull/211
- https://github.com/netty/netty-incubator-codec-http3/pull/212
- https://github.com/netty/netty-incubator-codec-http3/pull/213
- https://github.com/netty/netty-incubator-codec-http3/pull/214


graemerocher:
Can you address the sonar bugs https://sonarcloud.io/project/issues?id=micronaut-projects_micronaut-core&pullRequest=8559&resolved=false&types=BUG

yawkat:
they are false positives

graemerocher:
in that case you can suppress them by adding `@SupressWarnings("java:s3890")` where is `s3890` is whatever the sonar code is for the problem

yawkat:
@graemerocher ive disabled them in the sonar web interface, is that enough?

graemerocher (approved):

timyates (commented):
Could you update [the jdkClient docs](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/src/main/docs/guide/httpClient/httpClientImplementations/jdkHttpClient.adoc) to add Http3 to the list of unsupported things

And I guess we should [add a check here](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java#L142) to fail if Http3 is selected?

timyates (approved):

PR #8565, opened 2023-01-03
add isWithin(..) check to each property handling

graemerocher:


melix (approved):
Confirmed to fix the reported problem, thank you!

PR #8576, opened 2023-01-05
init at build time before instantiating reference

graemerocher:
Fix error:

```
 io.micronaut.data.model.$DefaultPageSerializer$Definition$Reference was unintentionally initialized at build time. To see why io.micronaut.data.model.$DefaultPageSerializer$Definition$Reference got initialized use --trace-class-initialization=io.micronaut.data.model.$DefaultPageSerializer$Definition$Reference
io.micronaut.data.runtime.criteria.$RuntimeCriteriaBuilder$Definition$Reference was unintentionally initialized at build time. 
```


melix (approved):
Confirmed to fix the problem :tada: 

https://ge.micronaut.io/s/oikdghx2ezfmc

Side note: do I understand properly that this code will initialize at build time bean references which may not be used by the application?

graemerocher:
Good point, not sure what implications this has for image sizes 🤔

PR #8580, opened 2023-01-05
Add support for annotation-based CORS configuration (8558)

### LINKED ISSUE #8558: Support annotation based CORs Configuration
### Feature description

Support CORs configuration via annotation. For example, it will enable Micronaut developers only to allow CORS for a few routes in their applications. Thus, having more secure applications. 

It will also ease people migrating from Spring Framework to the Micronaut framework. Thus, find a similar replacement to [Spring's CrossOrigin](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/bind/annotation/CrossOrigin.java)

I think we will need the following steps: 

Create an annotation. E.g. `@CrossOrigin` whose members map [`CorsOriginConfiguration's`](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-server/src/main/java/io/micronaut/http/server/cors/CorsOriginConfiguration.java) fields. 

Enable `Corsfilter` by default by removing this line: 

https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-server/src/main/java/io/micronaut/http/server/cors/package-info.java#L23

Create a method that returns an  [`@NonNull  Optional<CorsOriginConfiguration> getCorsOriginConfigurationForRequest(@NonNull HttpRequest<?> request)`](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-server/src/main/java/io/micronaut/http/server/cors/CorsOriginConfiguration.java).  We can [obtain the `RouteMatch` of a request](https://github.com/micronaut-projects/micronaut-security/blob/master/security/src/main/java/io/micronaut/security/filters/SecurityFilter.java) and [get the annotation contents if the method has the `@CrossOrigin` annotation](https://github.com/micronaut-projects/micronaut-security/blob/master/security/src/main/java/io/micronaut/security/rules/SecuredAnnotationRule.java#L65-L68).

The [`CorsFilter::getConfiguration`](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java#L212-L218) should first look for configuration coming from an annotation in the route match.

wetted:
closes #8558


sdelamo:
I merged https://github.com/micronaut-projects/micronaut-core/pull/8582 into this PR. 

I added a test which tests the scenario: 

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java

@wetted  The test only tests the `value` member of the `CrossOrigin` annotation. Can you add tests for the other members of the `CrossOrigin` annotation?

wetted:
@Sergio I'm not sure why the same three tests in each of `test-suite` for each language are failing. They weren't the other day. Investigating. But if you can look at my changes for exposedHeaders I either fixed an existing bug or I understand how that part should work.

test-suite-groovy/src/test/groovy/io/micronaut/docs/http/server/cors/CorsController.groovy
```diff
@@ -0,0 +1,26 @@
...
+import io.micronaut.http.MediaType
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.Produces
+import io.micronaut.http.server.cors.CrossOrigin
+// end::imports[]
+
+// tag::controller[]
+@Controller("/hello")
+class CorsController {
```
sdelamo, line 16: annotate with `@Requires(property="spec.name"` to avoid bean pollution.

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/http/server/cors/CorsController.kt
```diff
@@ -0,0 +1,26 @@
...
+// tag::imports[]
+import io.micronaut.http.MediaType
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.Produces
+import io.micronaut.http.server.cors.CrossOrigin
+// end::imports[]
+
+// tag::controller[]
+@Controller("/hello")
```
sdelamo, line 14: annotate with `@Requires(property="spec.name"` to avoid bean pollution.

test-suite/src/test/java/io/micronaut/docs/http/server/cors/CorsController.java
```diff
@@ -0,0 +1,26 @@
...
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Get;
+import io.micronaut.http.annotation.Produces;
+import io.micronaut.http.server.cors.CrossOrigin;
+// end::imports[]
+
+// tag::controller[]
+@Controller("/hello")
+public class CorsController {
```
sdelamo, line 15: annotate with `@Requires(property="spec.name"` to avoid bean pollution.

sdelamo (changes requested):
Currently, you add only controllers to `test-suite`, `test-suite-groovy`, `test-suite-kotlinp. You have to add the tests which invoke the controllers in `test-suite`, `test-suite-groovy`, `test-suite-kotlin.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
```diff
@@ -0,0 +1,278 @@
...
+                .status(HttpStatus.FORBIDDEN)
+                .assertResponse(this::assertCorsHeadersNotPresent)
+                .build()));
+    }
+
+    @Test
+    void exposedHeadersHappyPath() throws IOException {
+        asserts(SPECNAME,
+            preflight(UriBuilder.of("/exposedheaders").path("bar"), "https://foo.com", HttpMethod.GET)
+                .header(HttpHeaders.ACCESS_CONTROL_EXPOSE_HEADERS, HttpHeaders.AUTHORIZATION + "," + HttpHeaders.CONTENT_TYPE),
```
sdelamo: preflight requests don't specify a `Access-Control-Expose-Headers` header. 
see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
```diff
@@ -0,0 +1,278 @@
...
+                    assertTrue(response.getHeaders().names().contains(HttpHeaders.ACCESS_CONTROL_EXPOSE_HEADERS));
+                })
+                .build()));
+    }
+
+    @Test
+    void exposedHeadersFailure() throws IOException {
+        asserts(SPECNAME,
+            preflight(UriBuilder.of("/exposedheaders").path("bar"), "https://foo.com", HttpMethod.GET)
+                .header(HttpHeaders.ACCESS_CONTROL_EXPOSE_HEADERS, "foo"),
```
sdelamo, line 203: preflight requests don't specify a Access-Control-Expose-Headers header.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

sdelamo (changes requested):

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
```diff
@@ -0,0 +1,278 @@
...
+        asserts(SPECNAME,
+            preflight(UriBuilder.of("/foo").path("bar"), "https://bar.com", HttpMethod.GET),
+            (server, request) -> AssertionUtils.assertThrows(server, request, HttpResponseAssertion.builder()
+                .status(HttpStatus.METHOD_NOT_ALLOWED)
+                .assertResponse(this::assertCorsHeadersNotPresent)
+                .build()));
+    }
+
+    @Test
+    void corsSimpleRequestMethods() {
```
sdelamo, line 83: The way the docs read. It seems that still you have to enable  CORS via configuration  to use `@CrossOrigin`.

sdelamo (changes requested):

src/main/docs/guide/httpServer/serverConfiguration/cors.adoc
```diff
@@ -158,3 +158,32 @@ micronaut:
     cors:
       single-header: true
 ----
+
```
sdelamo: The way the docs read. It seems that still you have to enable  CORS via configuration  to use `@CrossOrigin`.

sdelamo (changes requested):

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/http/server/cors/CorsController.kt
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/717a22d209991591569b64857243d5c7b7a6ec41

test-suite/src/test/java/io/micronaut/docs/http/server/cors/CorsController.java
sdelamo: resolved via resolved via https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/717a22d209991591569b64857243d5c7b7a6ec41

test-suite-groovy/src/test/groovy/io/micronaut/docs/http/server/cors/CorsController.groovy
sdelamo: resolved via resolved via https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/717a22d209991591569b64857243d5c7b7a6ec41

sdelamo:
> @sergio I'm not sure why the same three tests in each of `test-suite` for each language are failing. They weren't the other day. Investigating. 

because of bean pollution:. See:  https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/717a22d209991591569b64857243d5c7b7a6ec41

> But if you can look at my changes for exposedHeaders I either fixed an existing bug or I understand how that part should work.

exposeHeaders is not something you will specify in a preflight request. https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/3da48c130d9b44bfa32e7fab1cdbbb6fd94b5ccf

src/main/docs/guide/httpServer/serverConfiguration/cors.adoc
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/687cd1135ab1db819d8e933a685a7b99002ca54c

sdelamo:
> Currently, you add only controllers to `test-suite`, `test-suite-groovy`, `test-suite-kotlinp. You have to add the tests which invoke the controllers in `test-suite`, `test-suite-groovy`, `test-suite-kotlin.

I have added the tests: 

https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/64811567b9c839418ba3de99559e200d62086d12

Having classes in `test-suite` modules without tests is almost the same as hardcoding the code in the documentation.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/3da48c130d9b44bfa32e7fab1cdbbb6fd94b5ccf

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
sdelamo: resolved via resolved via https://github.com/micronaut-projects/micronaut-core/pull/8580/commits/3da48c130d9b44bfa32e7fab1cdbbb6fd94b5ccf

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
```diff
@@ -0,0 +1,286 @@
...
+            value = "https://foo.com"
+        )
+        @Produces(MediaType.TEXT_PLAIN)
+        @Get("/foo")
+        String foo() {
+            return "foo";
+        }
+    }
+
+    // TODO: tests for CrossOrigin.allowCredentials, CrossOrigin.maxAge
```
sdelamo: Please, add tests and remove this TODO

sdelamo (changes requested):

sdelamo (approved):

http-server/src/main/java/io/micronaut/http/server/cors/CrossOrigin.java
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @return supported HTTP request methods
+     */
+    HttpMethod[] allowedMethods() default {};
+
+    /**
+     *
+     * @return whether the browser should send credentials
+     */
+    String allowCredentials() default "true";
```
yawkat: using String for this is weird. Maybe it could be an enum.

Also the default seems to actually be `""`?

sdelamo: @yawkat we currently default to true via configuration

https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/http-server/src/main/java/io/micronaut/http/server/cors/CorsOriginConfiguration.java#L47

I think the default annotation behaviour should be match the configuraiton default. 

Maybe, for 4 we should change the default but I think both should be aligned.

sdelamo (approved):

sdelamo:
@yawkat can you approve/request changes ?

src/main/docs/guide/httpServer/serverConfiguration/cors/annotationBasedCors.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+As an alternative, <<corsConfiguration, CORS configuration>> can be applied in a more fine-grained manner to specific routes using the
+link:{api}/io/micronaut/http/server/cors/CrossOrigin.html[@CrossOrigin] annotation. This is applied to `@Controller` to apply the CORS configuration to all endpoints in the controller. Alternatively, the annotation can be applied to specific endpoints on a controller, for even more fine-grained control.
+
+The `@CrossOrigin` annotation maps with a one-to-one correspondence to application-wide link:{api}/io/micronaut/http/server/cors/CorsOriginConfiguration.html[CorsOriginConfiguration] configuration properties. To specify just an allowed origin, use `@CrossOrigin("https://foo.com")`. To specify additional configuration details, use  a combination of annotation attributes the same as you would for specifying global `CorsOriginConfiguration` properties.
+
+[source,java]
+----
+@CrossOrigin(
+	allowedOrigins = { "http://foo.com" ,"^http(|s):\\/\\/www\\.google\\.com$" },
```
yawkat: I think using regex here is problematic. Like this example, it would match http://foo.com and http://fooxcom. Not an issue when there's no subdomain, but if there is, it allows an attacker to hijack the domain: When http://www.foo.com is configured for cors, an attacker may register wwwxfoo.com, and get full cors access. This means it's easy with regex to accidentally make an insecure configuration.

(yes I know this is already the current behavior)

sdelamo: Would you be ok if `allowedOrigins` defaulted not to use regex unless you enable it.  And we will default to be disabled? `allowedOriginsRegex` boolean

yawkat: yes that works

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
```diff
@@ -0,0 +1,437 @@
...
+                (server, request) -> AssertionUtils.assertThrows(server, request, HttpResponseAssertion.builder()
+                    .status(HttpStatus.FORBIDDEN)
+                    .assertResponse(CorsUtils::assertCorsHeadersNotPresent)
+                    .build()))
+        );
+    }
+
+
+    @Test
+    void allowedHeadersRegexHappyPath() throws IOException {
```
wetted: @sdelamo I'm not getting the test to work. Can you see what I did wrong?

src/main/docs/guide/httpServer/serverConfiguration/cors/annotationBasedCors.adoc
graemerocher: Doesn't it make more sense for `allowedOriginsRegex` to be a string and you should set that

yawkat: agreed

sdelamo:
@graemerocher & @yawkat, I would like to move this PR forward. Can you please check my comments here?

https://github.com/micronaut-projects/micronaut-core/pull/8580#discussion_r1127448522

Let me know by reviewing and approving the PR review if you are ok with using the boolean for the regex. Or please, suggest any necessary changes to be able to merge this.

graemerocher:
See my comment here https://github.com/micronaut-projects/micronaut-core/pull/8580#discussion_r1128320337

It doesn't make sense to have a boolean, just make `allowOriginRegex` a string that sets the regex

sdelamo:
> See my comment here [#8580 (comment)](https://github.com/micronaut-projects/micronaut-core/pull/8580#discussion_r1128320337)

Have you seen my comments to that thread?

sdelamo:
> See my comment here [#8580 (comment)](https://github.com/micronaut-projects/micronaut-core/pull/8580#discussion_r1128320337)
> 
> It doesn't make sense to have a boolean, just make `allowOriginRegex` a string that sets the regex

@graemerocher  can you please review again?

yawkat (approved):

graemerocher (approved):

PR #8582, opened 2023-01-09
fix: 403 for cors simple requests & CorsFilter off

sdelamo:
Related to: see: https://github.com/micronaut-projects/micronaut-core/security/advisories/GHSA-583g-g682-crxf#advisory-comment-77446

I created a test which reproduces a scenario for a developer running an application in localhost without cors enabled .

These are the steps to reproduce.

Clone: https://github.com/sdelamo/simple-request-attacks

inside that repository there is a sample app.

```
$ cd micronaut-stop
```

Run the app with `./gradlew run`, open your browser, and visit [https://sdelamo.github.io/simple-request-attacks/stopEndpoint.html](https://sdelamo.github.io/simple-request-attacks/stopEndpoint.html) in Chrome.

The application stops. It should return 503 instead.

Safari has protections built-in, and you will not be able to reproduce it with Safari. Use Chrome.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/cors/package-info.java
+++ b/http-server/src/main/java/io/micronaut/http/server/cors/package-info.java
@@ -19,10 +19,4 @@
  * @author Graeme Rocher
  * @since 1.0
  */
-@Configuration
-@Requires(property = "micronaut.server.cors.enabled", value = StringUtils.TRUE)
 package io.micronaut.http.server.cors;
-
-import io.micronaut.context.annotation.Configuration;
-import io.micronaut.context.annotation.Requires;
-import io.micronaut.core.util.StringUtils;
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterEnabledSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterEnabledSpec.groovy
@@ -12,8 +12,8 @@ class CorsFilterEnabledSpec extends Specification {
     @Shared
     ApplicationContext applicationContext = ApplicationContext.run()
 
-    void "CorsFilter is not enabled by default"() {
+    void "CorsFilter is enabled by default"() {
         expect:
-        !applicationContext.containsBean(CorsFilter)
+        applicationContext.containsBean(CorsFilter)
     }
 }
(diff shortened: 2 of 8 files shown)


timyates (approved):
Do we need a test to ensure cors can be turned off?

sdelamo:
> Do we need a test to ensure cors can be turned off?

what do you mean by `turning off cors`?

http-server/src/main/java/io/micronaut/http/server/cors/package-info.java
```diff
@@ -19,10 +19,4 @@
  * @author Graeme Rocher
  * @since 1.0
  */
-@Configuration
-@Requires(property = "micronaut.server.cors.enabled", value = StringUtils.TRUE)
```
dstepanov, line 23: This means that everything in the package is not available in the bean context. The previous behavior, according to the code is that the property should be true.

dstepanov: I would keep it or add the required check to the filters.
If the filter is not needed, it shouldn't be included. Even a no-op publisher-based filter brings reactive overhead.

sdelamo: > > Do we need a test to ensure cors can be turned off?
> 
> what do you mean by `turning off cors`?

I added an extra test to be clear it behaves as before if enabled is not true 

https://github.com/micronaut-projects/micronaut-core/pull/8582/commits/e17092f6a8a25736ef0eaaafbd311febfc2cce79

sdelamo: > This means that everything in the package is not available in the bean context

Yes, please note there are only two beans in the package  - `CorsFilter` which I need to enable to do this security check and `CorsOriginConverter`. 


> I would keep it or add the required check to the filters.
If the filter is not needed, it shouldn't be included. Even a no-op publisher-based filter brings reactive overhead.

Please, note having the filter enabled is necessary to do this security check.

I am not sure what you are proposing.

graemerocher (approved):

http-server/src/main/java/io/micronaut/http/server/cors/package-info.java
sdelamo: I talked to @dstepanov We are going to merge it as it is in the PR.

PR #8586, opened 2023-01-09
Performance improvements + Java style improvements

dstepanov:
Some performance improvements:
- MethodLookup is now instantiating the services, it should be faster and avoid reflection caches
- Detected some typecast problems in a basic HelloWorld
- Removed reflection uses to find the repeatable annotation container
- `BeanDefinitionProducer` to cache some of the `enabled` checks and the definition instance
- Introduced `CollectionUtils.newSet/../Map` helper methods to create set/map collection with an expected element count
- Reduced some of the streams and lambdas spotted in the profiled (might be a bit of controversial)
- Supplier memorize methods rewritten to use double-check locking, inspired by Guava

Styling in some classes to use the variable defined in the `instanceof`, eliminated some of the embeddings by using fast return.

### DIFF
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -23,10 +23,6 @@ dependencies {
 jmh {
     includes = ['io.micronaut.http.server.StartupBenchmark']
     duplicateClassesStrategy = DuplicatesStrategy.WARN
-//    warmupIterations = 2
-//    iterations = 3
-//    fork = 1
-//    jvmArgs = ["-agentpath:/Applications/YourKit-Java-Profiler-2018.04.app/Contents/Resources/bin/mac/libyjpagent.jnilib"]
 }
 
 tasks.named("processJmhResources") {
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -7,7 +7,8 @@ dependencies {
     annotationProcessor project(":inject-java")
     jmhAnnotationProcessor project(":inject-java")
     jmhAnnotationProcessor libs.bundles.asm
-    
+    jmhAnnotationProcessor libs.jmh.generator.annprocess
+
     annotationProcessor project(":validation")
     compileOnly project(":validation")
     api project(":inject")
@@ -17,15 +18,14 @@ dependencies {
     api project(":router")
     api project(":runtime")
 
-    jmh libs.jmh
-    jmh libs.jmh.generator.annprocess
+    jmh libs.jmh.core
 }
 jmh {
     includes = ['io.micronaut.http.server.StartupBenchmark']
     duplicateClassesStrategy = DuplicatesStrategy.WARN
-    warmupIterations = 2
-    iterations = 3
-    fork = 1
+//    warmupIterations = 2
+//    iterations = 3
+//    fork = 1
 //    jvmArgs = ["-agentpath:/Applications/YourKit-Java-Profiler-2018.04.app/Contents/Resources/bin/mac/libyjpagent.jnilib"]
 }
 
--- a/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
+++ b/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
@@ -59,13 +59,13 @@ public class NettyByteBufferFactory implements ByteBufferFactory<ByteBufAllocato
     }
 
     /**
-     * @param allocatorSupplier The {@link ByteBufAllocator}
+     * @param allocator The {@link ByteBufAllocator}
      */
-    public NettyByteBufferFactory(ByteBufAllocator allocatorSupplier) {
+    public NettyByteBufferFactory(ByteBufAllocator allocator) {
         this.allocatorSupplier = new Supplier<ByteBufAllocator>() {
             @Override
             public ByteBufAllocator get() {
-                return allocatorSupplier;
+                return allocator;
             }
         };
     }
--- a/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
+++ b/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
@@ -26,6 +26,8 @@ import io.netty.buffer.Unpooled;
 import jakarta.annotation.PostConstruct;
 import jakarta.inject.Singleton;
 
+import java.util.function.Supplier;
+
 /**
  * A {@link ByteBufferFactory} implementation for Netty.
  *
@@ -42,20 +44,30 @@ public class NettyByteBufferFactory implements ByteBufferFactory<ByteBufAllocato
      */
     public static final NettyByteBufferFactory DEFAULT = new NettyByteBufferFactory();
 
-    private final ByteBufAllocator allocator;
+    private final Supplier<ByteBufAllocator> allocatorSupplier;
 
     /**
      * Default constructor.
      */
     public NettyByteBufferFactory() {
-        this.allocator = ByteBufAllocator.DEFAULT;
+        this.allocatorSupplier = new Supplier<ByteBufAllocator>() {
+            @Override
+            public ByteBufAllocator get() {
+                return ByteBufAllocator.DEFAULT;
+            }
+        };
     }
 
     /**
-     * @param allocator The {@link ByteBufAllocator}
+     * @param allocatorSupplier The {@link ByteBufAllocator}
      */
-    public NettyByteBufferFactory(ByteBufAllocator allocator) {
-        this.allocator = allocator;
+    public NettyByteBufferFactory(ByteBufAllocator allocatorSupplier) {
+        this.allocatorSupplier = new Supplier<ByteBufAllocator>() {
+            @Override
+            public ByteBufAllocator get() {
+                return allocatorSupplier;
+            }
+        };
     }
 
     @PostConstruct
@@ -71,22 +83,22 @@ public class NettyByteBufferFactory implements ByteBufferFactory<ByteBufAllocato
 
     @Override
     public ByteBufAllocator getNativeAllocator() {
-        return allocator;
+        return allocatorSupplier.get();
     }
 
     @Override
     public ByteBuffer<ByteBuf> buffer() {
-        return new NettyByteBuffer(allocator.buffer());
+        return new NettyByteBuffer(allocatorSupplier.get().buffer());
     }
 
     @Override
     public ByteBuffer<ByteBuf> buffer(int initialCapacity) {
-        return new NettyByteBuffer(allocator.buffer(initialCapacity));
+        return new NettyByteBuffer(allocatorSupplier.get().buffer(initialCapacity));
     }
 
     @Override
     public ByteBuffer<ByteBuf> buffer(int initialCapacity, int maxCapacity) {
-        return new NettyByteBuffer(allocator.buffer(initialCapacity, maxCapacity));
+        return new NettyByteBuffer(allocatorSupplier.get().buffer(initialCapacity, maxCapacity));
     }
 
     @Override
--- a/context/src/main/java/io/micronaut/runtime/context/env/ConfigurationIntroductionAdvice.java
+++ b/context/src/main/java/io/micronaut/runtime/context/env/ConfigurationIntroductionAdvice.java
@@ -106,10 +106,11 @@ public class ConfigurationIntroductionAdvice implements MethodInterceptor<Object
         );
 
         if (defaultValue != null) {
-            return value.orElseGet(() -> environment.convertRequired(
-                defaultValue,
-                argument
-            ));
+            Object result = value.orElse(null);
+            if (result == null) {
+                return environment.convertRequired(defaultValue, argument);
+            }
+            return result;
         } else if (rt.isOptional()) {
             return value.orElse(Optional.empty());
         } else if (context.isNullable()) {
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
@@ -133,7 +133,6 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
     /**
      * @param annotationName The annotation name
      */
-    @SuppressWarnings("unchecked")
     @UsedByGeneratedCode
     @Internal
     public AnnotationValue(String annotationName) {
@@ -216,8 +215,8 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
      * @param member The member
      * @return The properties as a immutable map.
      */
-    public @NonNull
-    Map<String, String> getProperties(@NonNull String member) {
+    @NonNull
+    public Map<String, String> getProperties(@NonNull String member) {
         return getProperties(member, "name");
     }
 
@@ -231,20 +230,21 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
      */
     public Map<String, String> getProperties(@NonNull String member, String keyMember) {
         ArgumentUtils.requireNonNull("keyMember", keyMember);
-        if (StringUtils.isNotEmpty(member)) {
-            List<AnnotationValue<Annotation>> values = getAnnotations(member);
-            if (CollectionUtils.isNotEmpty(values)) {
-                Map<String, String> props = new LinkedHashMap<>(values.size());
-                for (AnnotationValue<Annotation> av : values) {
-                    String name = av.stringValue(keyMember).orElse(null);
-                    if (StringUtils.isNotEmpty(name)) {
-                        av.stringValue(AnnotationMetadata.VALUE_MEMBER, valueMapper).ifPresent(v -> props.put(name, v));
-                    }
-                }
-                return Collections.unmodifiableMap(props);
+        if (StringUtils.isEmpty(member)) {
+            return Collections.emptyMap();
+        }
+        List<AnnotationValue<Annotation>> values = getAnnotations(member);
+        if (CollectionUtils.isEmpty(values)) {
+            return Collections.emptyMap();
+        }
+        Map<String, String> props = CollectionUtils.newLinkedHashMap(values.size());
+        for (AnnotationValue<Annotation> av : values) {
+            String name = av.stringValue(keyMember).orElse(null);
+            if (StringUtils.isNotEmpty(name)) {
+                av.stringValue(AnnotationMetadata.VALUE_MEMBER, valueMapper).ifPresent(v -> props.put(name, v));
             }
         }
-        return Collections.emptyMap();
+        return Collections.unmodifiableMap(props);
     }
 
     /**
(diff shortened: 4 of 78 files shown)


core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -231,20 +230,21 @@ Map<String, String> getProperties(@NonNull String member) {
...
-            if (CollectionUtils.isNotEmpty(values)) {
-                Map<String, String> props = new LinkedHashMap<>(values.size());
-                for (AnnotationValue<Annotation> av : values) {
-                    String name = av.stringValue(keyMember).orElse(null);
-                    if (StringUtils.isNotEmpty(name)) {
-                        av.stringValue(AnnotationMetadata.VALUE_MEMBER, valueMapper).ifPresent(v -> props.put(name, v));
-                    }
-                }
-                return Collections.unmodifiableMap(props);
+        if (StringUtils.isEmpty(member)) {
```
dstepanov, line 233: Just style changes, no logic

core/src/main/java/io/micronaut/core/type/RuntimeTypeInformation.java
```diff
@@ -39,29 +40,23 @@
...
 
     /**
      * Returns whether the annotation metadata specifies the type as single.
+     * @param type The return type
      * @param annotationMetadata The annotation metadata provider
      * @return True if does
      */
-    static boolean isSpecifiedSingle(AnnotationMetadataProvider annotationMetadata) {
-        for (TypeInformationProvider provider : TYPE_INFORMATION_PROVIDERS) {
+    static boolean isSpecifiedSingle(Class<?> type, AnnotationMetadataProvider annotationMetadata) {
```
dstepanov, line 55: Added an extra check to avoid reactive check when the type is a basic Java type.

core/src/main/java/io/micronaut/core/util/StreamUtils.java
```diff
@@ -189,7 +191,10 @@ public static <T, A extends Collection<T>> Collector<T, A, Collection<T>> toImmu
     /**
      * @param <T> The type
      * @return An immutable collection
+     * @deprecated use Stream#toList
      */
+    @Deprecated(forRemoval = true)
+    @NextMajorVersion("Remove after Micronaut 4 milestone 1")
     public static <T> Collector<T, Collection<T>, Collection<T>> toImmutableCollection() {
```
dstepanov, line 198: We should use `toList(`

gradle/libs.versions.toml
```diff
@@ -175,8 +175,8 @@ jcache = { module = "javax.cache:cache-api", version.ref = "jcache" }
 
 jetty-alpn-openjdk8-client = { module = "org.eclipse.jetty:jetty-alpn-openjdk8-client", version.ref = "jetty" }
 
-jmh = { module = "org.openjdk.jmh:jmh-core", version.ref = "jmh" }
-jmh-generator-annprocess = { module = "org.openjdk.jmh:jmh-core", version.ref = "jmh" }
+jmh-core = { module = "org.openjdk.jmh:jmh-core", version.ref = "jmh" }
+jmh-generator-annprocess = { module = "org.openjdk.jmh:jmh-generator-annprocess", version.ref = "jmh" }
```
dstepanov, line 179: I think that's correct dependency now

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBinderRegistrar.java
```diff
@@ -37,7 +37,7 @@
  * @author graemerocher
  * @since 2.0.0
  */
-@Singleton
+@Prototype
 @Internal
 class NettyBinderRegistrar implements BeanCreatedEventListener<RequestBinderRegistry> {
```
dstepanov, line 42: No need to be singleton as it's only loaded once

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConverters.java
```diff
@@ -47,13 +47,13 @@
  * @author graemerocher
  * @since 1.0
  */
-@Singleton
+@Prototype
 @Internal
 public class NettyConverters implements TypeConverterRegistrar {
```
dstepanov, line 52: All type converters should be prototypes as they are only needed at the startup

http/src/main/java/io/micronaut/http/converters/HttpConverterRegistrar.java
```diff
@@ -32,17 +33,17 @@
...
-    private final ResourceResolver resourceResolver;
+    private final Provider<ResourceResolver> resourceResolver;
 
     /**
      * Default constructor.
      *
      * @param resourceResolver The resource resolver
      */
-    protected HttpConverterRegistrar(ResourceResolver resourceResolver) {
+    protected HttpConverterRegistrar(Provider<ResourceResolver> resourceResolver) {
```
dstepanov, line 46: Avoid fetching it at the startup

inject-java/src/test/groovy/io/micronaut/inject/factory/nullreturn/NullReturnFactorySpec.groovy
```diff
@@ -83,22 +80,20 @@ class Test2 {}
...
         expect: "1 is created for D"
         beanContext.getBeansOfType(D).size() == 1
         beanContext.getBean(D, Qualifiers.byName("one"))
-        factory.bCalls == 6
-        factory.cCalls == 4
         factory.dCalls == 2 //2 C beans
 
         and: "1 is created for D2"
-        beanContext.getBeansOfType(D2).size() == 1
+        beanContext.getBeansOfType(D2).size() == 1 // D two is disabled
```
dstepanov, line 90: Changes in `AbstractInitializableBeanDefinition` where `argument.isDeclaredNullabl` is added brought those changes, but I think it's the correct behaviour

inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
```diff
@@ -2130,6 +2141,9 @@ private <K> K resolveBean(
             boolean isNotInnerConfiguration = !precalculatedInfo.isConfigurationProperties || !isInnerConfiguration(argument);
             ConfigurationPath previousPath = isNotInnerConfiguration ? resolutionContext.setConfigurationPath(null) : null;
             try {
+                if (argument.isDeclaredNullable()) {
```
dstepanov, line 2145: This change requires the nullable factory test change

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
```diff
@@ -267,7 +267,7 @@ default <R> Stream<ExecutableMethod<T, R>> findPossibleMethods(String name) {
     /**
      * @return The {@link ExecutableMethod} instances for this definition
      */
-    default Collection<ExecutableMethod<T, ?>> getExecutableMethods() {
+    default Collection<ExecutableMethod<T, Object>> getExecutableMethods() {
```
dstepanov: This signature makes more sense and allows us to aggregate methods without a compiler error

graemerocher:
One failure

graemerocher:
if you could address the sonar bugs as well https://sonarcloud.io/project/issues?id=micronaut-projects_micronaut-core&pullRequest=8586&resolved=false&types=BUG

dstepanov:
@graemerocher PR adds more forgotten caches cleanup on the bean context stop. It detected a use case where Netty embedded server test would stop and start the context and because the environment is also reset, in this case, some of the beans didn't load because of the missing properties provided in the test. The solution is to copy the env during stop-start of the embedded context if it also stops, start the bean context.

graemerocher (approved):

benchmarks/build.gradle
```diff
@@ -17,15 +18,14 @@ dependencies {
...
-    jmh libs.jmh.generator.annprocess
+    jmh libs.jmh.core
 }
 jmh {
     includes = ['io.micronaut.http.server.StartupBenchmark']
     duplicateClassesStrategy = DuplicatesStrategy.WARN
-    warmupIterations = 2
-    iterations = 3
-    fork = 1
+//    warmupIterations = 2
```
sdelamo: should we just remove it?

context/src/main/java/io/micronaut/runtime/context/env/ConfigurationIntroductionAdvice.java
```diff
@@ -106,10 +106,11 @@ private Object resolveProperty(MethodInvocationContext<Object, Object> context,
         );
 
         if (defaultValue != null) {
-            return value.orElseGet(() -> environment.convertRequired(
```
sdelamo, line 109: is there are reason why we prefer the `orElse` null and the if statement vs the `orElseGet`?

core/src/main/java/io/micronaut/core/io/file/DefaultFileSystemResourceLoader.java
```diff
@@ -66,7 +66,7 @@ public DefaultFileSystemResourceLoader(String path) {
      * @param path The path
      */
     public DefaultFileSystemResourceLoader(Path path) {
-        this.baseDir = SupplierUtil.memoizedNonEmpty(() -> {
+        this.baseDir = SupplierUtil.memoized(() -> {
```
sdelamo, line 69: is this change intentional ?

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -432,7 +436,7 @@ static <T> Argument<T> of(
         if (ArrayUtils.isEmpty(typeParameters)) {
             return of(type);
         }
-        return new DefaultArgument<>(type, NameUtils.decapitalize(type.getSimpleName()), AnnotationMetadata.EMPTY_METADATA, typeParameters);
+        return new DefaultArgument<>(type, null, AnnotationMetadata.EMPTY_METADATA, typeParameters);
```
sdelamo, line 439: is this intentional ?

core/src/main/java/io/micronaut/core/type/DefaultArgument.java
```diff
@@ -214,10 +228,13 @@ public Class<T> getType() {
     @Override
     @NonNull
     public String getName() {
-        if (name == null) {
-            return getType().getSimpleName();
+        if (name != null) {
+            return name;
+        }
+        if (namePrecalculated == null) {
+            namePrecalculated = NameUtils.decapitalize(type.getSimpleName());
```
sdelamo, line 235: Before if name was null we were returning : 


```java
return getType().getSimpleName();
```

now we return 

```java
NameUtils.decapitalize(type.getSimpleName());
```

is this intentional?

core/src/main/java/io/micronaut/core/util/StreamUtils.java
```diff
@@ -189,7 +191,10 @@ public static <T, A extends Collection<T>> Collector<T, A, Collection<T>> toImmu
     /**
      * @param <T> The type
      * @return An immutable collection
+     * @deprecated use Stream#toList
      */
+    @Deprecated(forRemoval = true)
```
sdelamo, line 196: I think we should deprecate this in MN 3.9.x branch.

graal/src/main/java/io/micronaut/graal/reflect/GraalReflectionMetadataWriter.java
```diff
@@ -40,7 +40,7 @@ final class GraalReflectionMetadataWriter extends AbstractAnnotationMetadataWrit
 
     public GraalReflectionMetadataWriter(ClassElement originatingElement,
                                          AnnotationMetadata annotationMetadata) {
-        super(resolveName(originatingElement), originatingElement, annotationMetadata, false);
+        super(resolveName(originatingElement), originatingElement, annotationMetadata, true);
```
sdelamo, line 43: its this change to use `true` as the constructor parameter intentional?

http-netty/src/main/java/io/micronaut/http/netty/channel/converters/KQueueChannelOptionFactory.java
```diff
@@ -36,7 +36,7 @@
  * @author croudet
  */
 @Internal
-@Singleton
+@Prototype
```
sdelamo, line 39: is the scope change intentional?

sdelamo (commented):
Can you describe how are you measuring the performance improvements before and after this change?

context/src/main/java/io/micronaut/runtime/context/env/ConfigurationIntroductionAdvice.java
dstepanov: It was a visible point in the profiler, lamba is created and possibly invoked every time

core/src/main/java/io/micronaut/core/io/file/DefaultFileSystemResourceLoader.java
dstepanov: I noticed it’s not needed to be nonempty as it doesn’t return a null or an optional

core/src/main/java/io/micronaut/core/type/Argument.java
dstepanov: Yes, I have added new tests which were failing because that’s the only point providing decapitalized argument name

core/src/main/java/io/micronaut/core/type/DefaultArgument.java
dstepanov: We need to have one way to provide the default argument name. Not really sure what should be the default

graal/src/main/java/io/micronaut/graal/reflect/GraalReflectionMetadataWriter.java
dstepanov: Yes, the field is always true. The constructor is deprecated

gradle/libs.versions.toml
dstepanov: I converted all type registrars to be prototypes, now I see it also implements another interface, maybe it needs to be split

http-netty/src/main/java/io/micronaut/http/netty/channel/converters/KQueueChannelOptionFactory.java
dstepanov: I converted all type registrars to be prototypes, now I see it also implements another interface, maybe it needs to be split

buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
```diff
@@ -42,20 +44,30 @@ public class NettyByteBufferFactory implements ByteBufferFactory<ByteBufAllocato
...
+        };
     }
 
     /**
-     * @param allocator The {@link ByteBufAllocator}
+     * @param allocatorSupplier The {@link ByteBufAllocator}
      */
-    public NettyByteBufferFactory(ByteBufAllocator allocator) {
-        this.allocator = allocator;
+    public NettyByteBufferFactory(ByteBufAllocator allocatorSupplier) {
```
yawkat: param name is wrong

buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
```diff
@@ -42,20 +44,30 @@ public class NettyByteBufferFactory implements ByteBufferFactory<ByteBufAllocato
...
 
-    private final ByteBufAllocator allocator;
+    private final Supplier<ByteBufAllocator> allocatorSupplier;
 
     /**
      * Default constructor.
      */
     public NettyByteBufferFactory() {
-        this.allocator = ByteBufAllocator.DEFAULT;
+        this.allocatorSupplier = new Supplier<ByteBufAllocator>() {
```
yawkat, line 53: what's the point of this change? delaying initialization of ByteBufAllocator?

core/src/main/java/io/micronaut/core/convert/TypeConverter.java
```diff
@@ -72,6 +72,12 @@ default Optional<T> convert(S object, Class<T> targetType) {
      * @return The converter instance
      */
     static <ST, TT> TypeConverter<ST, TT> of(Class<ST> sourceType, Class<TT> targetType, Function<ST, TT> converter) {
-        return (object, targetType1, context) -> Optional.ofNullable(converter.apply(object));
+        // Keep the anonymous class instead of Lambda to reduce the Lambda invocation overhead during the startup
```
yawkat, line 75: i wonder if we could use retrolambda to do these changes automatically. but maybe it's overkill.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -4103,48 +4090,102 @@ public <T> BeanRegistration<T> getInFlightBean(BeanIdentifier beanIdentifier) {
...
 
-        @Override
-        protected void onNewSegment(Segment<?, ?> segment) {
-            Argument<?> argument = segment.getArgument();
-            if (argument.isContainerType()) {
-                argument = argument.getFirstTypeVariable().orElse(null);
-                if (argument == null) {
-                    return;
+        public boolean isDisabled() {
+            return reference == null || referenceEnabled != null && !referenceEnabled || definitionEnabled != null && !definitionEnabled;
```
yawkat: add parentheses, this is unreadable

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -4103,48 +4090,102 @@ public <T> BeanRegistration<T> getInFlightBean(BeanIdentifier beanIdentifier) {
...
-                            .add(getHierarchy());
+            return reference;
+        }
+
+        public <T> BeanDefinition<T> getDefinition(BeanContext beanContext) {
+            if (definitionEnabled != null && !definitionEnabled) {
+                throw new IllegalStateException("The definition is not enabled");
+            }
+            try {
+                if (definition == null) {
```
yawkat, line 4177: this is racey with isDefinitionEnabled

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
yawkat: `?` is correct though? it would be better to change downstream users of `getExecutableMethods` to use proper covariance

tbh maybe we could just drop the return type parameter from ExecutableMethod, i dont think we use it productively anyway

yawkat (approved):

buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyByteBufferFactory.java
dstepanov: Yes, postpone it to a bit later

dstepanov:
I removed the changes to `getExecutableMethods`. I'll add them in another PR. Removing the return generic type might be a solution.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
dstepanov: We don't really care, it can be loaded multiple times like it's now

PR #8598, opened 2023-01-11
Add logger.config to loggingConfiguration.adoc

adamkobor:
A new configuration property was introduced in https://github.com/micronaut-projects/micronaut-core/pull/8238 but it's not documented.


sdelamo:
thanks @adamkobor I have moved it into its own section.

sdelamo (approved):

PR #8608, opened 2023-01-14
Load ApplicationContextConfigurer with passed classloader instead of the class one

Goldmensch:
Currently ApplicationContextConfigurer are loaded with the class loader of the current class, but this seems not like the expected behavior, when passing an own ClassLoader. This pr corrects this behavior.
Hopefully this will not break anything.

Context: I'm currently writing an integration between Micronaut and the bukkit minecraft api (papermc) and they use an own class loader which I have to pass to Micronaut (via Micronaut.run(ClassLoader). This works fine, except for loading the `ApplicationContextConfigurer`.


melix (commented):
Would be nice to add a test for this behavior.

melix:
Thanks for the reminder @sdelamo . I think this works, but I need to double check with Micronaut AOT.

melix (approved):
All good, I also checked that it works property with Micronaut AOT!

PR #8609, opened 2023-01-15
Links to bootstrap configuration in documentation

rogervinas:
Working on my [Micronaut PoC with Hashicorp Vault](https://dev.to/rogervinas/top-5-server-side-frameworks-for-kotlin-in-2022-micronaut-2p5b) it take me a few time to figure out that I needed to enable bootstrap.

If you strictly follow [8.1.3 HashiCorp Vault Support](https://docs.micronaut.io/latest/guide/#distributedConfigurationVault) instructions you may expect it to work only adding `bootstrap.yaml`, which is not the case.

Adding a hint in all "Distributed Config" documentation hoping it helps other developers to make sure [bootstrap](https://docs.micronaut.io/latest/guide/#bootstrap) is enabled, for example adding `io.micronaut.discovery:micronaut-discovery-client` dependency.


graemerocher:
Thanks for the contribution!

PR #8613, opened 2023-01-16
Update docs to the new configuration macro

melix:
This commit reworks the YAML documentation samples to use the new `[configuration]` macro which generates a list of samples in multiple config languages instead.

It also fixes a couple of issues which were discovered as part of doing so.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,7 +14,7 @@ pluginManagement {
 }
 
 plugins {
-    id 'io.micronaut.build.shared.settings' version '6.1.1'
+    id 'io.micronaut.build.shared.settings' version '6.2.0'
 }
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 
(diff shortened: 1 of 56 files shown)


graemerocher (approved):

sdelamo (changes requested):
It is not working for me: 

I run ./gradlew docs -x htmlSanityCheck

I seve my local docs via apache in http://localhost/micronaut-core and this is what I see: 


CleanShot 2023-01-16 at 19 01 42@2x

Clicking on the different languages does nothing. 

I tried both in Chrome and Safari. Does not work in either.

melix:
> Clicking on the different languages does nothing.

> I tried both in Chrome and Safari. Does not work in either.

Those require changes to the js and css files. Can you check by disabling caches (e.g by refreshing the page with the developer console open?).

melix:
@sdelamo Can you double check :point_up: and if it still doesn't work after refreshing caches, build a zip of your `build/docs` directory?

graemerocher:
@sdelamo merging this because it works fine for me, must be a browser caching issue on your end

PR #8624, opened 2023-01-18
Deprecate FilterOrderProvider

yawkat:
For #8422

### DIFF
--- a/http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
+++ b/http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
@@ -23,7 +23,9 @@ import io.micronaut.core.order.Ordered;
  *
  * @author James Kleeh
  * @since 1.0
+ * @deprecated Deprecated without replacement. This class is unused and will be removed in 4.0.0.
  */
+@Deprecated
 public interface FilterOrderProvider extends Ordered {
 
 }


http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
```diff
@@ -23,7 +23,9 @@
  *
  * @author James Kleeh
  * @since 1.0
+ * @deprecated Deprecated without replacement. This class is unused and will be removed in 4.0.0.
  */
+@Deprecated
```
sdelamo, line 28:
```suggestion
@Deprecated( forRemoval = true)
```

http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
```diff
@@ -23,7 +23,9 @@
  *
  * @author James Kleeh
  * @since 1.0
+ * @deprecated Deprecated without replacement. This class is unused and will be removed in 4.0.0.
```
sdelamo, line 26: Can you clarify in the javadoc what users should use instead?

sdelamo (changes requested):

http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
yawkat: the pr is to 3.9 so we can't use forRemoval i dont think?

http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
yawkat: there is no replacement, i dont know what this class was for and how it was supposed to be used. it was never referenced anywhere

http/src/main/java/io/micronaut/http/filter/FilterOrderProvider.java
sdelamo: true, that is Java 9 only

sdelamo (approved):

PR #8631, opened 2023-01-19
Fix NPE in isCandidateBean

### LINKED ISSUE #8627: Intermittent NPE in isCandidateBean
### Expected Behavior

Should not NPE

### Actual Behaviour

NPEs with:

```
java.lang.NullPointerException: Cannot invoke "io.micronaut.inject.BeanDefinitionReference.isCandidateBean(io.micronaut.core.type.Argument)" because "this.reference" is null
      at app//io.micronaut.context.DefaultBeanContext$BeanDefinitionProducer.isReferenceCandidateBean(DefaultBeanContext.java:4187)
      at app//io.micronaut.context.DefaultBeanContext.collectBeanCandidates(DefaultBeanContext.java:2183)
      at app//io.micronaut.context.DefaultBeanContext.findBeanCandidates(DefaultBeanContext.java:2161)
      a
```

Broken in this commit https://github.com/micronaut-projects/micronaut-core/commit/6b8fa8a693a2c311663319734bea80c2ec9ab846

### Steps To Reproduce

Run tests in Micronaut Serialization or core

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.0.x

dstepanov:
Fixes #8627

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -4117,11 +4117,13 @@ public class DefaultBeanContext implements InitializableBeanContext {
         }
 
         public boolean isReferenceEnabled(DefaultBeanContext context, @Nullable BeanResolutionContext resolutionContext) {
-            if (reference == null) {
+            BeanDefinitionReference<?> ref = reference;
+            // The reference needs to be assigned to a new variable as it can change between checks
+            if (ref == null) {
                 return false;
             }
             if (referenceEnabled == null) {
-                if (reference.isEnabled(context, resolutionContext)) {
+                if (ref.isEnabled(context, resolutionContext)) {
                     referenceEnabled = true;
                 } else {
                     referenceEnabled = false;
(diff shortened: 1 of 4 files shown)


yawkat (approved):

sdelamo (commented):
can we add a test?

graemerocher (approved):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -4163,31 +4169,40 @@ public boolean isDefinitionEnabled(DefaultBeanContext context, @Nullable BeanRes
         }
 
         public <T> BeanDefinitionReference<T> getReference() {
-            if (reference == null || referenceEnabled == null || !referenceEnabled) {
+            // The reference needs to be assigned to a new variable as it can change between checks
+            Boolean refEnabled = referenceEnabled;
+            if (reference == null || refEnabled == null || !refEnabled) {
```
sdelamo, line 4174: should we also do   

`BeanDefinitionReference<?> ref = reference;`?

dstepanov: It's only problematic when you read it twice: first null check then compare value, the second check might cause the NPE.

PR #8632, opened 2023-01-19
Rewrite `AbstractAnnotationMetadataBuilder` bugfixes and tests

### LINKED ISSUE #8626: Annotation inheritance broken with introspections in 4.0.x branch
### Expected Behavior

Annotation inheritance should work for introspections.

### Actual Behaviour

The `@Serdable` annotation and friends are not inherited causing this test to fail

https://github.com/micronaut-projects/micronaut-serialization/blob/c63f7583158dbf594457cc33a47b101d6c7d95df/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/object/ObjectSerdeSpec.groovy#L202

It fails because `annotationMetadata.hasStereotype(Serdeable.Deserializable.class)` fails because the stereotypes are empty in the annotation metadata:

https://github.com/micronaut-projects/micronaut-serialization/blob/c63f7583158dbf594457cc33a47b101d6c7d95df/serde-support/src/main/java/io/micronaut/serde/support/DefaultSerdeIntrospections.java#L203

They should look like:

Screenshot 2023-01-19 at 11 29 50

This broke in this commit: 

https://github.com/micronaut-projects/micronaut-core/commit/6b8fa8a693a2c311663319734bea80c2ec9ab846

### Steps To Reproduce

Run https://github.com/micronaut-projects/micronaut-serialization/blob/c63f7583158dbf594457cc33a47b101d6c7d95df/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/object/ObjectSerdeSpec.groovy#L202

from the `master` branch against 4.0.x snapshot

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.0.x

dstepanov:
This PR:
- Rewrite of `AbstractAnnotationMetadataBuilder` to eliminate duplicities, improve how `AliasFor` is working, allowing for transformation to be run on the transformed values
- Corrected in what format annotation values are retrieved - int array will always produce `int[]` even for an empty value. And added various tests for it.
- Adding a new annotation correctly adds its stereotypes too
- Fixes for Groovy annotation value processing. Now Java and Groovy implementation can return the same results. Except for a few Groovy compiler bugs
- `AnnotationValue` is now always created with the defaults value for various cases like mapping etc
- Annotation defaults map is changed to `Map<CharSequence, Object>` from `Map<String, Object>` just like the annotation values, to improve code quality
- Source annotations also include default values like ordinary `RUNTIME` retention.
- Repeatable annotations are correctly processed when added/modified + multiple tests for it

Fixes #8626


graemerocher (changes requested):
Micronaut Serialization is still failing

inject-java/src/test/groovy/io/micronaut/annotation/mapping/ReplacesRepeatableAnnotationSpec.groovy
```diff
@@ -18,22 +16,21 @@ class ReplacesRepeatableAnnotationSpec extends AbstractTypeElementSpec {
             def definition = buildBeanDefinition('addann.ReplaceAnnotationsTo', '''
 package addann;
 
-import io.micronaut.context.annotation.Requires;
 import io.micronaut.context.annotation.Bean;
 
-@Requires(property = "foo")
-@Requires(property = "bar")
+@io.micronaut.annotation.mapping.MyRequires(property = "foo")
+@io.micronaut.annotation.mapping.MyRequires(property = "bar")
```
graemerocher, line 22: can you add a new test instead of replacing the existing one and changing it?

inject-java/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
```diff
@@ -45,7 +45,7 @@ class MyBean  {
             beanDefinition.getInjectedMethods()[0].name == 'setMyValue'
             def metadata = beanDefinition.getInjectedMethods()[0].getAnnotationMetadata()
             metadata.hasAnnotation(Property)
-            metadata.getValue(Property, "name", String).get() == 'endpoints.my-value'
+            metadata.getValue(Property, "name", String).get() == 'simple.my-value'
```
graemerocher, line 48: why was this changed?

inject-java/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
```diff
@@ -99,7 +99,7 @@ class MyBean  {
             beanDefinition.getInjectedMethods()[0].name == 'setMyValue'
             def metadata = beanDefinition.getInjectedMethods()[0].getAnnotationMetadata()
             metadata.hasAnnotation(Property)
-            metadata.getValue(Property, "name", String).get() == 'endpoints.my-value'
+            metadata.getValue(Property, "name", String).get() == 'endpoints.simple.my-value'
```
graemerocher, line 102: why was this changed?

core/src/main/java/io/micronaut/core/annotation/AnnotationUtil.java
```diff
@@ -45,11 +45,12 @@ public class AnnotationUtil {
             Inherited.class.getName(),
             SuppressWarnings.class.getName(),
             Override.class.getName(),
-            Repeatable.class.getName(),
+//            Repeatable.class.getName(),
```
graemerocher: is this intentional?

graemerocher (changes requested):
Looks like a refactoring that simplifies the code a great deal, but it is hard to say for sure if there are breaking changes in here since the changes are so significant. Have you tried this out against modules like OpenAPI, Data and Serialization that exercise this code significantly? I am worried about more regressions, we need to get 4.0.x stable.

inject-java/src/test/groovy/io/micronaut/annotation/mapping/ReplacesRepeatableAnnotationSpec.groovy
dstepanov: I just modified it to use a custom annotation instead of the requires that has a repeatable container entry in `AnnotationMetadataSupport`. The test was added by me a few PRs ago it's not something that old.

inject-java/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
dstepanov: I think the order of the applied `aliasFor` has changed. Those tests are testing the broken functionality anyway so I think it's fine.

inject-java/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
dstepanov: The same as the comment above.

core/src/main/java/io/micronaut/core/annotation/AnnotationUtil.java
dstepanov: Somehow we have this code in the bean definition writer:
```
            final String repeatableName = visitorContext
                    .getClassElement(annotationName)
                    .flatMap(ce -> ce.stringValue(Repeatable.class)).orElse(null);
```
Not sure how it was working before.

graemerocher:
Tested Micronaut Serialization with the latest changes:

Screenshot 2023-01-23 at 13 16 32

inject-java/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
graemerocher: what broken functionality?

dstepanov: `TestEndpoint` is mapping to the `value` which used to work as a base prefix, that would be multiple times modified

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -195,6 +195,15 @@ public List<AnnotationValue<?>> getStereotypes() {
         return stereotypes;
     }
 
+    /**
+     * The default values.
+     * @return The default of the annotation or null if not specified.
+     */
+    @Nullable
+    public Map<CharSequence, Object> getDefaultValues() {
```
graemerocher, line 204: add `@since`. Does this need to be nullable? Can we instead return emptyMap() as a default?

core/src/main/java/io/micronaut/core/annotation/AnnotationValueBuilder.java
```diff
@@ -115,9 +122,12 @@ public AnnotationValueBuilder<T> stereotype(AnnotationValue<?> annotation) {
      * @return This builder
      */
     @NonNull
-    public AnnotationValueBuilder<T> defaultValues(Map<String, Object> defaultValues) {
+    public AnnotationValueBuilder<T> defaultValues(Map<? extends CharSequence, Object> defaultValues) {
         if (defaultValues != null) {
-            this.defaultValues.putAll(defaultValues);
+            if (this.defaultValues == null) {
+                this.defaultValues = new LinkedHashMap<>();
```
graemerocher: can we size the map the same as the passed default values?

core/src/main/java/io/micronaut/core/annotation/AnnotationValueBuilder.java
```diff
@@ -103,6 +107,9 @@ public AnnotationValue<T> build() {
     @NonNull
     public AnnotationValueBuilder<T> stereotype(AnnotationValue<?> annotation) {
         if (annotation != null) {
+            if (stereotypes == null) {
+                stereotypes = new ArrayList<>();
```
graemerocher: Perhaps limit the array size since there are generally no more than say 10 stereotypes

core/src/main/java/io/micronaut/core/reflect/ReflectionUtils.java
```diff
@@ -295,6 +295,26 @@ public static Field getRequiredField(Class<?> type, String name) {
...
+     * Finds field's value or return an empty if exception occurs or if the value is null.
+     *
+     * @param fieldOwnerClass The field owner class
+     * @param fieldName       The field name
+     * @param instance        The instance
+     * @return An {@link Optional} contains the value or empty of the value is null or an error occurred
+     * @since 4.0.0
+     */
+    @Internal
+    public static Optional<Object> getFieldValue(Class<?> fieldOwnerClass, String fieldName, Object instance) {
```
graemerocher: nullability annotations on the method arguments

core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
```diff
@@ -514,60 +505,59 @@ AnnotatedElementValidator getElementValidator() {
...
-                T member = entry.getKey();
-                OptionalValues<?> aliasForValues = getAnnotationValues(originatingElement, member, AliasFor.class);
-                Object annotationValue = entry.getValue();
-                Optional<?> aliasMember = aliasForValues.get("member");
-                Optional<?> aliasAnnotation = aliasForValues.get("annotation");
-                Optional<?> aliasAnnotationName = aliasForValues.get("annotationName");
+            return new AnnotationValue<>(annotationTypeName, Collections.emptyMap());
+        }
+
+        Map<CharSequence, Object> resolvedValues = new LinkedHashMap<>();
```
graemerocher: can we size this map the same as the annotationValues?

graemerocher (approved):
Approved with some minor comments to address

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
dstepanov: No, null state means the default values aren't set and they should be retrieved later

PR #8642, opened 2023-01-20
Fix Cors for alternate local addresses

### LINKED ISSUE #8560: Error 403 forbidden POST login using localhost
### Expected Behavior

The POST /login response with http status 200 and CORS headers included.

Micronaut 3.7.4
``
curl -i -v 'http://localhost:8080/login' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0' -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate, br' -H 'Referer: http://localhost:8090/' -H 'Content-Type: application/x-www-form-urlencoded' -H 'X-Requested-With: XMLHttpRequest' -H 'Authorization: Basic Og==' -H 'Origin: http://127.0.0.1:8090/' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' --data-raw 'grant_type=password&username=admin&password=admin'
``
> Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /login HTTP/1.1
> Host: localhost:8080
> User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0
> Accept: application/json, text/plain, */*
> Accept-Language: en-US,en;q=0.5
> Accept-Encoding: gzip, deflate, br
> Referer: http://localhost:8090/
> Content-Type: application/x-www-form-urlencoded
> X-Requested-With: XMLHttpRequest
> Authorization: Basic Og==
> Origin: http://127.0.0.1:8090/
> Connection: keep-alive
> Sec-Fetch-Dest: empty
> Sec-Fetch-Mode: cors
> Sec-Fetch-Site: same-site
> Content-Length: 49

> Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
>**HTTP/1.1 200 OK**
< date: Mon, 2 Jan 2023 20:57:26 GMT
> date: Mon, 2 Jan 2023 20:57:26 GMT
< Access-Control-Allow-Origin: http://127.0.0.1:8090/
>**Access-Control-Allow-Origin: http://127.0.0.1:8090/**
< Vary: Origin
>Vary: Origin
< Access-Control-Allow-Credentials: true
> **Access-Control-Allow-Credentials: true**
< Content-Type: application/json
>Content-Type: application/json
< content-length: 267
>content-length: 267
< connection: keep-alive
>connection: keep-alive

 
> Connection #0 to host localhost left intact
>{"username":"admin","roles":["ADMIN"],"access_token":"eyJhbGciOiJub25lIn0.eyJzdWIiOiJhZG1pbiIsIm5iZiI6MTY3MjY5MzA0Niwicm9sZXMiOlsiQURNSU4iXSwiaXNzIjoiY29ycy1sb2NhbGhvc3QtY29uZiIsImV4cCI6MTY3MjY5NjY0NiwiaWF0IjoxNjcyNjkzMDQ2fQ.","token_type":"Bearer","expires_in":3600}

Micronaut 3.8.0 host different than localhost 
``
curl -i -v 'http://127.0.0.1:8080/login' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0' -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate, br' -H 'Referer: http://localhost:8090/' -H 'Content-Type: application/x-www-form-urlencoded' -H 'X-Requested-With: XMLHttpRequest' -H 'Authorization: Basic Og==' -H 'Origin: http://127.0.0.1:8090/' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' --data-raw 'grant_type=password&username=admin&password=admin'
``
>   Trying 127.0.0.1:8080...
> Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> POST /login HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0
> Accept: application/json, text/plain, */*
> Accept-Language: en-US,en;q=0.5
> Accept-Encoding: gzip, deflate, br
> Referer: http://localhost:8090/
> Content-Type: application/x-www-form-urlencoded
> X-Requested-With: XMLHttpRequest
> Authorization: Basic Og==
> Origin: http://127.0.0.1:8090/
> Connection: keep-alive
> Sec-Fetch-Dest: empty
> Sec-Fetch-Mode: cors
> Sec-Fetch-Site: same-site
> Content-Length: 49
 
> Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
>**HTTP/1.1 200 OK**
< date: Mon, 2 Jan 2023 20:25:31 GMT
>date: Mon, 2 Jan 2023 20:25:31 GMT
< Access-Control-Allow-Origin: http://127.0.0.1:8090/
>**Access-Control-Allow-Origin: http://127.0.0.1:8090/**
< Vary: Origin
>Vary: Origin
< Access-Control-Allow-Credentials: true
>**Access-Control-Allow-Credentials: true**
< Content-Type: application/json
>Content-Type: application/json
< content-length: 267
>content-length: 267
< connection: keep-alive
> connection: keep-alive

### Actual Behaviour

I use swagger to expose the Rest API documentation. The authentication process call the login endpoint in a different server (I use an auth-server running in other server than the business microservices).
Upgrading to micronaut 3.8.0 version the remote login in swagger doesn't work anymore (the login response is 403 forbidden). After check configuration and differents test using CURL I realize that the CORS headers isn't included in the response. Checking the source code I found some validations when the host is localhost that explain the 403 status code but for example the ip 127.0.0.1 isn't included. Is that correct?
In order to solve this problem I have to include the following configuration to use swagger in local environment

``
micronaut:
  server:
    cors:
      enabled: true
      configurations:
        web:
          allowedOrigins:
            - ^http(|s):\/\/*.*
``

In previous version the mentioned configuration wasn't need to include and CURL and swagger works perfectly.
In the Example Application can reproduce the behavior mentioned using the following curl

``
curl -i -v 'http://localhost:8080/login' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0' -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate, br' -H 'Referer: http://localhost:8090/' -H 'Content-Type: application/x-www-form-urlencoded' -H 'X-Requested-With: XMLHttpRequest' -H 'Authorization: Basic Og==' -H 'Origin: http://127.0.0.1:8090/' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' --data-raw 'grant_type=password&username=admin&password=admin'
``

>   Trying 127.0.0.1:8080...
> Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /login HTTP/1.1
> Host: localhost:8080
> User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0
> Accept: application/json, text/plain, */*
> Accept-Language: en-US,en;q=0.5
> Accept-Encoding: gzip, deflate, br
> Referer: http://localhost:8090/
> Content-Type: application/x-www-form-urlencoded
> X-Requested-With: XMLHttpRequest
> Authorization: Basic Og==
> Origin: http://127.0.0.1:8090/
> Connection: keep-alive
> Sec-Fetch-Dest: empty
> Sec-Fetch-Mode: cors
> Sec-Fetch-Site: same-site
> Content-Length: 49
> 
> Mark bundle as not supporting multiuse
< HTTP/1.1 403 Forbidden
>**HTTP/1.1 403 Forbidden**
< connection: keep-alive
>connection: keep-alive
< transfer-encoding: chunked
>transfer-encoding: chunked

### Steps To Reproduce

1- Clone the example application repository
2- ./gradlew run
3-  Execute this curl command

* Using localhost
``
curl -i -v 'http://localhost:8080/login' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0' -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate, br' -H 'Referer: http://localhost:8090/' -H 'Content-Type: application/x-www-form-urlencoded' -H 'X-Requested-With: XMLHttpRequest' -H 'Authorization: Basic Og==' -H 'Origin: http://127.0.0.1:8090/' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' --data-raw 'grant_type=password&username=admin&password=admin' 
``

* Using 127.0.0.1
``
curl -i -v 'http://127.0.0.1:8080/login' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0' -H 'Accept: application/json, text/plain, */*' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate, br' -H 'Referer: http://localhost:8090/' -H 'Content-Type: application/x-www-form-urlencoded' -H 'X-Requested-With: XMLHttpRequest' -H 'Authorization: Basic Og==' -H 'Origin: http://127.0.0.1:8090/' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' --data-raw 'grant_type=password&username=admin&password=admin'
``

### Environment Information

Manjaro, Ubuntu
OpenJDK 17, 18
It works fine in Micronaut 3.7.4 and Micronaut 3.7.5

### Example Application

https://github.com/fercapi84/cors-localhost-conf

### Version

3.8.0

timyates:
It was raised in https://github.com/micronaut-projects/micronaut-core/issues/8560#issuecomment-1379453245 that we have not completed the circle with CORS and localhost drive-by protection

This PR adds a check for 127.0.01 in addition to localhost.

Currently it has a couple of missing parts which are up for discussion:

1.  For local connections, It does not take the Origin/Server port into account, which it should do to be spec compliant.  I feel that adding this (whilst simple) could hamper local development of multiple services, or front-ends.
2. If the Origin is `localhost` and the request is 127.0.0.1, we currently allow this (and vice versa).  Again this is not spec compliant (I think), but if we have an origin of either of the two, I'm currently happy with the risk.
3. We do not check all 127.0.0.0/8 addresses (which are local).  To do so would require some level of parsing, and I worry that the cost to each request will outweigh the benefit to people with a 127.0.0.32 address (I've never seen anything other than 127.0.01 -- but I admit I've not been specifically looking)

Please comment with your thoughts on these 3 issues (and any others you think about)

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -127,8 +127,8 @@ class CorsFilterSpec extends Specification {
         ['.*']                              | 'http://www.bar.com'
         ['^http://www\\.(foo|bar)\\.com$']  | 'http://www.bar.com'
         ['^http://www\\.(foo|bar)\\.com$']  | 'http://www.foo.com'
-        ['.*bar$', '.*foo$']                | 'asdfasdf foo'
-        ['.*bar$', '.*foo$']                | 'asdfasdf bar'
+        ['.*bar$', '.*foo$']                | 'http://asdfasdf.foo'
+        ['.*bar$', '.*foo$']                | 'http://asdfasdf.bar'
     }
 
     void "test handleRequest with disallowed method"() {
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -127,6 +127,8 @@ class CorsFilterSpec extends Specification {
         ['.*']                              | 'http://www.bar.com'
         ['^http://www\\.(foo|bar)\\.com$']  | 'http://www.bar.com'
         ['^http://www\\.(foo|bar)\\.com$']  | 'http://www.foo.com'
+        ['.*bar$', '.*foo$']                | 'asdfasdf foo'
+        ['.*bar$', '.*foo$']                | 'asdfasdf bar'
         ['.*bar$', '.*foo$']                | 'http://asdfasdf.foo'
         ['.*bar$', '.*foo$']                | 'http://asdfasdf.bar'
     }
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
@@ -185,7 +185,7 @@ public class CorsSimpleRequestTest {
     }
 
     private RequestSupplier createRequestFor(String host, String origin) {
-        return (server) -> createRequest(server.getPort().map(p -> "http://" + host + ":" + p + "/refresh").orElseThrow(() -> new RuntimeException("Unknown port for " + server)), origin);
+        return server -> createRequest(server.getPort().map(p -> "http://" + host + ":" + p + "/refresh").orElseThrow(() -> new RuntimeException("Unknown port for " + server)), origin);
     }
 
     static void isForbidden(ServerUnderTest server, HttpRequest<?> request) {
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
@@ -30,6 +30,8 @@ import io.micronaut.http.annotation.Status;
 import io.micronaut.http.client.multipart.MultipartBody;
 import io.micronaut.http.server.tck.AssertionUtils;
 import io.micronaut.http.server.tck.HttpResponseAssertion;
+import io.micronaut.http.server.tck.RequestSupplier;
+import io.micronaut.http.server.tck.ServerUnderTest;
 import io.micronaut.runtime.context.scope.refresh.RefreshEvent;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
@@ -69,15 +71,30 @@ public class CorsSimpleRequestTest {
         asserts(SPECNAME,
             Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
             createRequest("https://foo.com"),
-            (server, request) -> {
-                RefreshCounter refreshCounter = server.getApplicationContext().getBean(RefreshCounter.class);
-                assertEquals(0, refreshCounter.getRefreshCount());
-                AssertionUtils.assertThrows(server, request, HttpResponseAssertion.builder()
-                        .status(HttpStatus.FORBIDDEN)
-                        .assertResponse(response -> assertFalse(response.getHeaders().contains("Vary")))
-                    .build());
-                assertEquals(0, refreshCounter.getRefreshCount());
-        });
+            CorsSimpleRequestTest::isForbidden
+        );
+    }
+
+    /**
+     * @see <a href="https://github.com/micronaut-projects/micronaut-core/security/advisories/GHSA-583g-g682-crxf">GHSA-583g-g682-crxf</a>
+     *
+     * A malicious/compromised website can make HTTP requests to 127.0.0.1. Normally, such requests would trigger a CORS preflight check which would prevent the request; however, some requests are "simple" and do not require a preflight check. These endpoints, if enabled and not secured, are vulnerable to being triggered.
+     * Example with Javascript:
+     * <pre>
+     * let url = "http://127.0.0.1:8080/refresh";
+     * let body = new FormData();
+     * body.append("force", "true");
+     * fetch(url, { method: "POST", body });
+     * </pre>
+     * @throws IOException may throw the try for resources
+     */
+    @Test
+    void corsSimpleRequestNotAllowedFor127AndAny() throws IOException {
+        asserts(SPECNAME,
+            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
+            createRequestFor("127.0.0.1", "https://foo.com"),
+            CorsSimpleRequestTest::isForbidden
+        );
     }
 
     /**
@@ -89,14 +106,8 @@ public class CorsSimpleRequestTest {
         asserts(SPECNAME,
             Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
             createRequest("http://localhost:8000"),
-            (server, request) -> {
-                RefreshCounter refreshCounter = server.getApplicationContext().getBean(RefreshCounter.class);
-                assertEquals(0, refreshCounter.getRefreshCount());
-                AssertionUtils.assertDoesNotThrow(server, request, HttpResponseAssertion.builder()
-                    .status(HttpStatus.OK)
-                    .build());
-                assertEquals(1, refreshCounter.getRefreshCount());
-            });
+            CorsSimpleRequestTest::isSuccessful
+        );
     }
 
     /**
@@ -108,20 +119,13 @@ public class CorsSimpleRequestTest {
     void corsSimpleRequestAllowedForLocalhostAnd127Origin() throws IOException {
         asserts(SPECNAME,
             Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
-            (server) -> createRequest(server.getPort().map(p -> "http://localhost:" + p + "/refresh").orElseThrow(() -> new RuntimeException("Unknown port for " + server)), "http://127.0.0.1:8000"),
-            (server, request) -> {
-                RefreshCounter refreshCounter = server.getApplicationContext().getBean(RefreshCounter.class);
-                assertEquals(0, refreshCounter.getRefreshCount());
-                AssertionUtils.assertThrows(server, request, HttpResponseAssertion.builder()
-                    .status(HttpStatus.FORBIDDEN)
-                    .assertResponse(response -> assertFalse(response.getHeaders().contains("Vary")))
-                    .build());
-                assertEquals(0, refreshCounter.getRefreshCount());
-            });
+            createRequestFor("localhost", "http://127.0.0.1:8000"),
+            CorsSimpleRequestTest::isForbidden
+        );
     }
 
     /**
-     * A request to 127.0.0.1 with an origin of localhost should fail despite them both being local.
+     * A request to 127.0.0.1 with an origin of localhost should succeed as they're both local.
      *
      * @throws IOException
      */
@@ -129,16 +133,9 @@ public class CorsSimpleRequestTest {
     void corsSimpleRequestAllowedFor127RequestAndLocalhostOrigin() throws IOException {
         asserts(SPECNAME,
             Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
-            (server) -> createRequest(server.getPort().map(p -> "http://127.0.0.1:" + p + "/refresh").orElseThrow(() -> new RuntimeException("Unknown port for " + server)), "http://localhost:8000"),
-            (server, request) -> {
-                RefreshCounter refreshCounter = server.getApplicationContext().getBean(RefreshCounter.class);
-                assertEquals(0, refreshCounter.getRefreshCount());
-                AssertionUtils.assertThrows(server, request, HttpResponseAssertion.builder()
-                    .status(HttpStatus.FORBIDDEN)
-                    .assertResponse(response -> assertFalse(response.getHeaders().contains("Vary")))
-                    .build());
-                assertEquals(0, refreshCounter.getRefreshCount());
-            });
+            createRequestFor("127.0.0.1", "http://localhost:8000"),
+            CorsSimpleRequestTest::isSuccessful
+        );
     }
 
     /**
@@ -171,6 +168,29 @@ public class CorsSimpleRequestTest {
                     .build());
                 assertEquals(1, refreshCounter.getRefreshCount());
             });
+    }
+
+    private RequestSupplier createRequestFor(String host, String origin) {
+        return (server) -> createRequest(server.getPort().map(p -> "http://" + host + ":" + p + "/refresh").orElseThrow(() -> new RuntimeException("Unknown port for " + server)), origin);
+    }
+
+    static void isForbidden(ServerUnderTest server, HttpRequest<?> request) {
+        RefreshCounter refreshCounter = server.getApplicationContext().getBean(RefreshCounter.class);
+        assertEquals(0, refreshCounter.getRefreshCount());
+        AssertionUtils.assertThrows(server, request, HttpResponseAssertion.builder()
+            .status(HttpStatus.FORBIDDEN)
+            .assertResponse(response -> assertFalse(response.getHeaders().contains("Vary")))
+            .build());
+        assertEquals(0, refreshCounter.getRefreshCount());
+    }
+
+    static void isSuccessful(ServerUnderTest server, HttpRequest<?> request) {
+        RefreshCounter refreshCounter = server.getApplicationContext().getBean(RefreshCounter.class);
+        assertEquals(0, refreshCounter.getRefreshCount());
+        AssertionUtils.assertDoesNotThrow(server, request, HttpResponseAssertion.builder()
+            .status(HttpStatus.OK)
+            .build());
+        assertEquals(1, refreshCounter.getRefreshCount());
     }
 
     static HttpRequest<?> createRequest(String origin) {
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
@@ -111,7 +111,7 @@ public class CorsSimpleRequestTest {
     }
 
     /**
-     * A request to localhost with an origin of 127.0.0.1 should be allowed as they are both local
+     * A request to localhost with an origin of 127.0.0.1 should be allowed as they are both local.
      *
      * @throws IOException
      */
@@ -125,7 +125,7 @@ public class CorsSimpleRequestTest {
     }
 
     /**
-     * Spoof attempt with origin should fail
+     * Spoof attempt with origin should fail.
      *
      * @throws IOException
      */
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
@@ -125,6 +125,20 @@ public class CorsSimpleRequestTest {
     }
 
     /**
+     * Spoof attempt with origin should fail
+     *
+     * @throws IOException
+     */
+    @Test
+    void corsSimpleRequestFailsForLocalhostAndSpoofed127Origin() throws IOException {
+        asserts(SPECNAME,
+            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
+            createRequestFor("localhost", "http://127.0.0.1.hac0r.com:8000"),
+            CorsSimpleRequestTest::isForbidden
+        );
+    }
+
+    /**
      * A request to 127.0.0.1 with an origin of localhost should succeed as they're both local.
      *
      * @throws IOException
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
@@ -100,6 +100,48 @@ public class CorsSimpleRequestTest {
     }
 
     /**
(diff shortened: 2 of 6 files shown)


nbrugger-tgm:
> We do not check all 127.0.0.0/8 addresses (which are local). To do so would require some level of parsing, and I worry that the cost to each request will outweigh the benefit to people with a 127.0.0.32 address (I've never seen anything other than 127.0.01 -- but I admit I've not been specifically looking)
@timyates 

The problem is not people using `127.0.0.32` as their API host. Because 90% use `0.0.0.0` as their api host which includes all of `127.0.0.0/8`. If i would create a local drive by attack i would generate a random adress fitting this IP range. If you check out a random Micronaut API on github and start is `localhost` `127.0.0.1`as well as `127.0.0.32` will probably yield the same result.

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -132,12 +134,17 @@ protected boolean shouldDenyToPreventDriveByLocalhostAttack(@NonNull CorsOriginC
...
         }
         String host = httpHostResolver.resolve(request);
-        return isAny(corsOriginConfiguration.getAllowedOrigins()) && host.startsWith(LOCALHOST);
+        return isAny(corsOriginConfiguration.getAllowedOrigins()) && isLocal(host);
+    }
 
+    private boolean isLocal(@NonNull String hostString) {
+        URI uri = URI.create(hostString);
+        String host = uri.getHost();
+        return "localhost".equals(host) || "127.0.0.1".equals(host);
```
nbrugger-tgm:
```suggestion
        return "localhost".equals(host) || host.startsWith("127");
```
this is the most accurate you get without reading the systems network settings and it matches the whole of `127.0.0.0/8`

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
```diff
@@ -89,14 +106,36 @@ void corsSimpleRequestAllowedForLocalhostAndOriginLocalhost() throws IOException
...
+     * @throws IOException
+     */
+    @Test
+    void corsSimpleRequestAllowedForLocalhostAnd127Origin() throws IOException {
+        asserts(SPECNAME,
+            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
+            createRequestFor("localhost", "http://127.0.0.1:8000"),
+            CorsSimpleRequestTest::isSuccessful
+        );
+    }
```
nbrugger-tgm:
```suggestion
    /**
     * A request to localhost with an origin of 127.0.0.1 should be allowed as they are both local
     *
     * @throws IOException
     */
    @Test
    void corsSimpleRequestAllowedForLocalhostAnd127DefaultOrigin() throws IOException {
        asserts(SPECNAME,
            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
            createRequestFor("localhost", "http://127.0.0.1:8000"),
            CorsSimpleRequestTest::isSuccessful
        );
    }
    
        /**
     * A request to localhost with an origin of 127.0.0.1 should be allowed as they are both local
     *
     * @throws IOException
     */
    @Test
    void corsSimpleRequestAllowedForLocalhostAnd127Origin() throws IOException {
        asserts(SPECNAME,
            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
            createRequestFor("localhost", "http://127.200.200.200:8000"),
            CorsSimpleRequestTest::isSuccessful
        );
    }
```

nbrugger-tgm (changes requested):

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
```diff
@@ -69,15 +71,30 @@ void corsSimpleRequestNotAllowedForLocalhostAndAny() throws IOException {
...
+     * @throws IOException may throw the try for resources
+     */
+    @Test
+    void corsSimpleRequestNotAllowedFor127AndAny() throws IOException {
+        asserts(SPECNAME,
+            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
+            createRequestFor("127.0.0.1", "https://foo.com"),
+            CorsSimpleRequestTest::isForbidden
+        );
     }
```
nbrugger-tgm, line 98:
```suggestion
    @Test
    void corsSimpleRequestNotAllowedFor127AndAny() throws IOException {
        asserts(SPECNAME,
            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
            createRequestFor("127.0.0.1", "https://foo.com"),
            CorsSimpleRequestTest::isForbidden
        );
    }
    @Test
    void corsSimpleRequestNotAllowedTo127Subnet() throws IOException {
        asserts(SPECNAME,
            Collections.singletonMap(PROPERTY_MICRONAUT_SERVER_CORS_ENABLED, StringUtils.TRUE),
            createRequestFor("127.200.200.200", "https://foo.com"),
            CorsSimpleRequestTest::isForbidden
        );
    }
```

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
timyates: What happens when someone builds a website `127.0.0.malicious.com`?

nbrugger-tgm: my bad not thinking about this, it would need to be coupled with something like `host.matches("\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}")`

timyates: Yeah, I didn't want to cause a regular expression match to be executed for every request 🤔

nbrugger-tgm: Using manual splitting and parsing makes it about ~30-50% *slower* from my immature and very vague performance tests.

On my local machine the test takes ~2ms with regex ~4ms with custom parsing, Still adding 2ms to a request seem less than ideal (ofc this 2ms depend on the machine/CPU, JIT and runtime warmup) i dont have a better way in mind atm

The only other thing i could think about is using the `NetworkInterface` API to check and adress for a loopback bound

The test i used
```java
class Main {
		var start = System.currentTimeMillis();
		var rest = Arrays.stream("127.0.0.1".split("\\.",4))
			           .allMatch(s -> {
				           try{
					           Integer.parseInt(s);
					           return true;
				           }catch(Exception e){return false;}
			           });
		var end = System.currentTimeMillis();
		System.out.println("Non regex : "+(end-start)+"ms");
		start = System.currentTimeMillis();
		rest = "127.0.0.1".matches("\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}");
		end = System.currentTimeMillis();
		System.out.println("Regex     : "+(end-start)+"ms");
	}
}
```

timyates: I'll have a think...  Have you ever seen `127.0.0.3` being used in the wild? @sdelamo any thoughts?

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
```diff
@@ -127,8 +127,8 @@ class CorsFilterSpec extends Specification {
         ['.*']                              | 'http://www.bar.com'
         ['^http://www\\.(foo|bar)\\.com$']  | 'http://www.bar.com'
         ['^http://www\\.(foo|bar)\\.com$']  | 'http://www.foo.com'
-        ['.*bar$', '.*foo$']                | 'asdfasdf foo'
-        ['.*bar$', '.*foo$']                | 'asdfasdf bar'
+        ['.*bar$', '.*foo$']                | 'http://asdfasdf.foo'
+        ['.*bar$', '.*foo$']                | 'http://asdfasdf.bar'
```
timyates: This test was failing as the origin was not a valid Origin format of:

```
Origin: null
Origin: <scheme>://<hostname>
Origin: <scheme>://<hostname>:<port>
```

And I now parse it with URI to get the hostname

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
nbrugger-tgm: Nobody uses `127.0.0.3` but it allways redirects to localhost. This means that if a webpage requests `127.200.200.200` this bypasses the check. It isn't conscious for serving but it is an attack surface!

The default serving host `0.0.0.0` also includes `127.0.0.1 -> 127.254.254.254` which are 2^24 vulnerable addresses so yes i could argue that nearly every Micronaut uses `127.0.0.3` unconscious. It might prevents the simplest form of attack but any malicious actor who has any knowledge about networks can bypass it by changing 1 bit in the address.

Maybe i understand something wrong, do you see where i am coming from?

timyates: Ahhh, I see now

> Because 90% use 0.0.0.0 as their api host which includes all of 127.0.0.0/8

Micronaut specifically binds to `localhost`, not 0.0.0.0 (unless the user changes the setting)

https://github.com/micronaut-projects/micronaut-core/blob/1d5016b2917e8c3467677c14e543417553be65bc/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java#L399

So I don't believe 127.0.0.2 would resolve?

sdelamo:
We have to protect against these two scenarios, and we should protect without incurring any performance costs if possible. 

If you run this sample app locally. 

https://github.com/sdelamo/simple-request-attacks

And you visit in a browser that does not protect against localhost-driven attach such as Chrome: 

Either:
 
https://sdelamo.github.io/simple-request-attacks/stopEndpoint.html

Or 
https://sdelamo.github.io/simple-request-attacks/127001.html

The former URL sends a request against `http://localhost:8080/stop`, and the latter sends a request against  `http://127.0.0.1:8080/stop`

The former has been patched since 3.8.1, and the latter will be patched via this PR. 

If you have another attack scenario with a reproducible example, please submit a PR against the attack demo https://sdelamo.github.io/simple-request-attacks, and I will be happy to re-evaluate and patch it. 

But we have to be careful not to incur performance costs.

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
nbrugger-tgm: Thats interesting but on my system `127.0.0.2` resolves on the default project using Arch Linux. Maybe on other systems it works different?
```bash
curl --location --request GET 'https://launch.micronaut.io/create/default/com.example.demo?lang=JAVA&build=GRADLE&test=JUNIT&javaVersion=JDK_17' --output demo.zip
unzip demo.zip
cd demo
./gradlew run
(different terminal) curl 127.0.0.3:8080
(different terminal) curl 127.200.200.203:8080
```
On my machine both curl commands pass with 404 not found json template which means the server is reached

timyates: Can you add a reproducer as described in https://github.com/micronaut-projects/micronaut-core/pull/8642#issuecomment-1400245630

nbrugger-tgm:
> If you have another attack scenario with a reproducible example
@sdelamo 

Thanks for the repo! I showed off the problem i mentioned with my PR : https://github.com/sdelamo/simple-request-attacks/pull/1

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
timyates: I think osx only supports 127.0.0.1, which is why I'm not seeing it

nbrugger-tgm: See : https://github.com/sdelamo/simple-request-attacks/pull/1

timyates:
Ignore me... 😿 

I'm now thinking we need to check the server address rather than the HOST header address to reduce the variables...

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
nbrugger-tgm: I think a simple ubuntu docker container should solve your OSX reproduction problem :)

timyates:
@nbrugger-tgm Right....  I *think* that's it...

Tested it inside a container with

```
curl -H "Origin: https://example.reqbin.com" 127.200.200.203:8080/stop
curl -H "Origin: https://127.hack.com" 127.200.200.203:8080/stop
curl -H "Origin: https://127.0.0.1:10101" 127.200.200.203:8080/stop
```

And only the last request worked...

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
timyates: Can't add this test, as it will not be resolved on anything but Linux :-(

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CorsSimpleRequestTest.java
timyates: We will not support an origin of anything but localhost or 127.0.0.1

sdelamo (approved):

nbrugger-tgm:
Looks good to me, but I think that your implementation using `URL` could be simpler with a `startsWith` check. Anyway since you are feature complete and the problem I pointed out is fixed you have my approval (as if you need it :P)

Also it just came to my mind that you also support websockets so maybe check `ws` and `wss` too?

nbrugger-tgm (approved):

timyates:
Thanks for all the help with this @nbrugger-tgm 😎👍 

I added a page https://sdelamo.github.io/simple-request-attacks/stopChecks.html to hopefully make testing easier

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -132,27 +132,51 @@ protected boolean shouldDenyToPreventDriveByLocalhostAttack(@NonNull CorsOriginC
...
+            || hostString.startsWith("wss://127.");
+    }
+
+    /*
+     * For Origin, we need to be more strict as otherwise an address like 127.malicious.com would be allowed.
+     */
+    private boolean isOriginLocal(@NonNull String hostString) {
+        URI uri = URI.create(hostString);
+        String host = uri.getHost();
+        return "localhost".equals(host) || "127.0.0.1".equals(host);
```
graemerocher: use `SocketUtils.LOCALHOST` constant

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -132,27 +132,51 @@ protected boolean shouldDenyToPreventDriveByLocalhostAttack(@NonNull CorsOriginC
...
+     */
+    private boolean isHostLocal(@NonNull String hostString) {
+        return hostString.startsWith("http://localhost")
+            || hostString.startsWith("https://localhost")
+            || hostString.startsWith("http://127.")
+            || hostString.startsWith("https://127.")
+            || hostString.startsWith("ws://localhost")
+            || hostString.startsWith("wss://localhost")
+            || hostString.startsWith("ws://127.")
+            || hostString.startsWith("wss://127.");
```
graemerocher, line 180: can we improve the performance of this with a check that the first char is `w` or `h`?

graemerocher (changes requested):

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -132,27 +132,51 @@ protected boolean shouldDenyToPreventDriveByLocalhostAttack(@NonNull CorsOriginC
...
+    }
+
+    /*
+     * For Origin, we need to be more strict as otherwise an address like 127.malicious.com would be allowed.
+     */
+    private boolean isOriginLocal(@NonNull String hostString) {
+        URI uri = URI.create(hostString);
```
yawkat: add test for invalid uri as origin

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
timyates: Added a pre-check for a host of length 0 and that the first char must be h or w

yawkat (approved):

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
timyates: Caught the Illegal argument exception, and re-instated the test in CorsFilterSpec that tests this

yawkat (approved):

graemerocher (approved):

PR #8644, opened 2023-01-20
Issue 8620 - deprecate ConversionService.SHARED

### LINKED ISSUE #8620: Deprecate `ConversionService.SHARED` in Micronaut 3.9, for removal in Micronaut 4
### Issue description

Deprecate in Micronaut 3.9, for removal in Micronaut 4

wetted:
closes #8620

### DIFF
--- a/core/src/main/java/io/micronaut/core/convert/ConversionService.java
+++ b/core/src/main/java/io/micronaut/core/convert/ConversionService.java
@@ -33,7 +33,7 @@ public interface ConversionService<Impl extends ConversionService> {
     /**
      * The default shared conversion service.
      *
-     * @deprecated This will be removed in the next version. Use an injected {@link ConversionService} instead.
+     * @deprecated This will be removed in the next major version. Use an injected {@link ConversionService} instead.
      */
     @Deprecated
     ConversionService<?> SHARED = new DefaultConversionService();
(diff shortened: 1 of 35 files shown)


core/src/main/java/io/micronaut/core/convert/ConversionService.java
```diff
@@ -32,7 +32,10 @@
 
     /**
      * The default shared conversion service.
+     *
+     * @deprecated This will be removed in the next version. Use an injected {@link ConversionService} instead.
```
alvarosanchez:
```suggestion
     * @deprecated This will be removed in the next major version. Use an injected {@link ConversionService} instead.
```

alvarosanchez (changes requested):
Don't we have other references in core (tests, samples, docs, etc) that use still SHARED?

wetted:
> Don't we have other references in core (tests, samples, docs, etc) that use still SHARED?

Is that the scope of deprecating it in this Issue/PR (not disputing, just need to understand)? But I agree it needs to be done.

wetted:
@sdelamo  @alvarosanchez There are 111 uses of `ConversionService.SHARED` in the 3.9.x branch. 98 in java files, 10 in groovy files, 3 in kotlin files and none in adoc files.

Nine of them are for the docs examples under `test-suite[groovy|kotlin] and addressed in #8582 

It seems to me we should create a separate PR targing 4.0.0 to address (i.e. remove) these?

alvarosanchez:
I personally think that we should not use deprecated code ourselves, even in 3.9. And having references in documentation is even more concerning

wetted:
This is superseded by #8695. Removing all uses of the `ConversionService.SHARED` is problematic and requires more effort than can be prioritized, but this is abandoned now.

PR #8646, opened 2023-01-20
Handle WebLogic classloader URIs

### LINKED ISSUE #8636: Micronaut war fails to start under WebLogic 12c
### Expected Behavior

A Micronaut war file should deploy successfully to Oracle WebLogic 12c.

### Actual Behaviour

The application fails to start because it cannot find a `ApplicationEventPublisher` bean. This bean cannot cannot be created, because the startup process (via `SoftServiceLoader`) fails to find and instantiate the `ApplicationEventPublisherFactory`.

Please see the README file in the sample repository for full details on the application behavior.

Note that this may or may not be the only issue in getting a Micronaut war working in WebLogic 12c; it's certainly the first issue and affects many beans/Micronaut services.

### Steps To Reproduce

I created a default Micronaut 3.5.4 application with `mn create-app demo.weblogic-war` and only a few minor changes to `build.gradle`. I then built a war file and attempted to deploy to Oracle WebLogic 12c.

Please see the README file in the sample repository for full details to reproduce.

### Environment Information

- WebLogic 12c (running in docker)
- Java 8
- Micronaut 3.5.4

### Example Application

https://github.com/mattmoss/weblogic-war

### Version

3.5.4

mattmoss:
Fixes issue #8636

### DIFF
--- a/core/src/main/java/io/micronaut/core/exceptions/UncheckedIOException.java
+++ b/core/src/main/java/io/micronaut/core/exceptions/UncheckedIOException.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


alvarosanchez (changes requested):
This needs to target 3.8.x branch

sdelamo:
> This needs to target 3.8.x branch

No, we need to target 3.5.x because we need a patch version of the minor version used by Grails 5.
 
I will merge up the change.

sdelamo:
I added a test. Moreover, I [verified locally that with this change, a war file deployed to WebLogic no longer throws the exception. ](https://github.com/mattmoss/weblogic-war)

graemerocher (approved):

mattmoss:
Just to be clear on my change in [dcf1ab4](https://github.com/micronaut-projects/micronaut-core/pull/8646/commits/dcf1ab434029e22d7a6c8233052beed435d678f6)...  `Path.of` is not available in Java 8; that came later.

PR #8659, opened 2023-01-24
Add discovery 3.3.0 BOM to build [skip ci]

timyates:
Once 3.3.0 is released with a BOM, we can add it to the catalog


sdelamo:
discovery client with bom is released

sdelamo (approved):

PR #8674, opened 2023-01-27
Allow programmatic logback config if xml config is absent

### LINKED ISSUE #8650: Logback can configure only from XML since 3.8.0
### Expected Behavior

According to the [documentation of Logback ](https://logback.qos.ch/manual/configuration.html#auto_configuration), Logback can be configured either **programmatically** or with a configuration script expressed in XML or Groovy format.

> 1. Logback will search for any custom [Configurator](https://logback.qos.ch/xref/ch/qos/logback/classic/spi/Configurator.html) providers using [service-provider loading facility](http://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html) (introduced in JDK 1.6). If any such custom provider is found, it takes precedence over logback's own DefaultJoranConfigurator.
> 2. If no custom Configurator was found in the previous step, then [DefaultJoranConfigurator](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/util/DefaultJoranConfigurator.java) will be instantiated and invoked.

### Actual Behaviour

With providing default LoggingSystem "logback.xml" in [LogbackLoggingSystem](https://github.com/micronaut-projects/micronaut-core/blob/3fe132aa1cb8d937a73936d1940dd77181530355/management/src/main/java/io/micronaut/management/endpoint/loggers/impl/LogbackLoggingSystem.java#L57) constructor there is no way to configure logback programmatically.

timyates:
closes #8650


sdelamo (changes requested):
I don't think this closes #8650 but #8679 and #8678 

I have extracted a [Utils class](https://github.com/micronaut-projects/micronaut-core/pull/8674/commits/af90618d7b71f174a48c7ce732fb388d70d683cf) to fix the endpoints issue. 

Can you add a test for the logger endpoint scenarios. I think the tests will be best placed in a module such as `test-suite-logback` instead of in `runtime`?

runtime/src/main/java/io/micronaut/logging/impl/LogbackUtils.java
```diff
@@ -0,0 +1,90 @@
...
+     * If resource is not present searches for a custom {@link Configurator} via a service loader.
+     * If no custom configuration, it uses a {@link BasicConfigurator}.
+     * if resource is present it configures the context with the resource.
+     *
+     * @param context Logger Context
+     * @param resource A resource for example logback.xml
+     */
+    private static void configure(@NonNull LoggerContext context,
+                                 @Nullable URL resource) {
+        try {
```
sdelamo: I don't think the logic is correct. According to [Logback](https://logback.qos.ch/manual/configuration.html#auto_configuration)  docs, we should first check if there is a custom configuration and the the rest.

sdelamo (changes requested):

runtime/src/main/java/io/micronaut/logging/impl/LogbackUtils.java
timyates: Switched the logic to match 👍

runtime/src/main/java/io/micronaut/logging/impl/LogbackUtils.java
```diff
@@ -0,0 +1,96 @@
...
+     * If not present it configures the context with the resource.
+     *
+     * @param context  Logger Context
+     * @param logbackXmlLocation the location of the xml logback config file
+     * @param resource A resource for example logback.xml
+     */
+    private static void configure(
+        @NonNull LoggerContext context,
+        @NonNull String logbackXmlLocation,
+        @Nullable URL resource
```
sdelamo: Could we change this to `Supplier<URL>` to attempt to resolve the URL only if the custom configuration is not found.

sdelamo (approved):

timyates:
Ping @yawkat 🔔

graemerocher (approved):

yawkat:
Sorry I saw this issue my inbox but didn't see that it was relevant to me :sweat_smile: I guess it's been reviewed enough now.

Issue #8635, opened 2023-01-19
Unable to inject bytearray into constructor of imported bean


auke-:
### Expected Behavior

Allow injection of byte arrays into constructors of imported beans.

When an upstream project defines a bean with a byte array as an constructor argument (see https://github.com/auke-/micronaut-bytearray-injection-bug/blob/master/upstream/src/main/java/io/codegen/upstream/UpstreamByteConstructorBean.java):
```java
@Singleton
public class UpstreamByteConstructorBean {

  private final byte[] bytes;

  public UpstreamByteConstructorBean(@Named("some-bytes") byte[] bytes) {
    this.bytes = bytes;
  }

  @Override
  public String toString() {
    return new String(bytes, StandardCharsets.UTF_8);
  }
}
```

And a downstream project imports this bean (see https://github.com/auke-/micronaut-bytearray-injection-bug/blob/master/downstream/src/main/java/io/codegen/downstream/BeanFactory.java):
```java
@Import(classes = {UpstreamByteConstructorBean.class})
@Factory
public class BeanFactory {

  @Bean
  @Singleton
  @Named("some-bytes")
  byte[] getSomeBytes() {
    return "some-bytes".getBytes(StandardCharsets.UTF_8);
  }

}
```

One expects to be able to inject `UpstreamByteConstructorBean` which the byte array equivalent of `some-bytes`.

### Actual Behaviour

Injecting byte arrays in fields and constructors works when the bean is defined in the same project and injecting byte arrays in fields also works for imported beans but constructor injection fails for imported beans. When loading the bean definition a ClassNotFoundException is thrown:

```
io.micronaut.context.exceptions.BeanContextException: Error loading bean [io.codegen.upstream.UpstreamByteConstructorBean]: byte

	at io.micronaut.context.DefaultBeanContext.findBeanCandidates(DefaultBeanContext.java:2096)
	at io.micronaut.context.DefaultApplicationContext.findBeanCandidates(DefaultApplicationContext.java:260)
	at io.micronaut.context.DefaultBeanContext.findConcreteCandidateNoCache(DefaultBeanContext.java:3093)
	at io.micronaut.context.DefaultBeanContext.findConcreteCandidate(DefaultBeanContext.java:3058)
	at io.micronaut.context.DefaultBeanContext.findBeanDefinition(DefaultBeanContext.java:788)
	at io.micronaut.context.DefaultBeanContext.resolveBeanRegistration(DefaultBeanContext.java:2790)
[...]

PR #8686, opened 2023-01-30
fix: properly handle primitive types in bean import

auke-:
Primitive types in constructors of imported beans where handled improperly, an array of primitives in synthesized as a type with the name of the primitive (e.g. byte[] becomes Lbyte; instead of `[B;`).

This is caused by a bug in JavaModelUtils.getTypeReference where the given type element is used instead of the actual class element.


auke-:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/8635

auke-:
@graemerocher do you have time to take a look? It looks like the variable `type` was used where it should be `classElement`

graemerocher:
Thanks I need to put together a test case for this

graemerocher:
Replaced by https://github.com/micronaut-projects/micronaut-core/pull/8778

PR #8688, opened 2023-01-30
Refactor `FilterRunner` to use data flow

dstepanov:
The refactoring implements the filtering flow using `ExecutionFlow`, which eliminates shared states of request/response.
Making the data flow makes the code more understandable and eliminates more errors.
The disadvantage is more lambdas, but that's already the case in the server code.

I have noticed that HTTP client, filters the exceptions instead of the response. Is it correct? It looks like something it would be nice to eliminate in V4 and filter the response.


http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -420,26 +161,72 @@ public final FilterRunner reactorContext(Context reactorContext) {
...
+            return ExecutionFlow.just(context);
+        }
+        if (iterator.hasNext()) {
+            GenericHttpFilter filter = iterator.next();
+            return processRequestFilter(filter, context, suspended)
+                    .onErrorResume(throwable -> {
+                        // Un-suspend possibly awaiting filter and exception filtering scenario of the http client
+                        return filterResponse(context, iterator, throwable, suspended).map(context::withResponse);
+                    })
+                    .flatMap(newContext -> filterRequest0(newContext, iterator, suspended));
```
yawkat: this will run on error too

graemerocher: @dstepanov did you address this feedback?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -33,6 +33,7 @@
 import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.http.reactive.execution.ReactiveExecutionFlow;
 import io.micronaut.inject.ExecutableMethod;
+import org.jetbrains.annotations.NotNull;
```
graemerocher: remove this

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -84,241 +75,13 @@ public class FilterRunner {
...
-    continuation.proceed() first kicks off downstream processing (calls workRequest()) and then
-    waits for the SuspensionPoint to complete. When workResponse walks backwards through the
-    filters, it will see the SuspensionPoint, and notify it (complete the CompletableFuture). When
-    myFilter finishes modifying the response, the return handler will then trigger the remaining
-    response filters (call workResponse()).
-
-    The same SuspensionPoint mechanism is also adapted to reactive filters fairly easily. In that
-    case, proceed() returns a publisher that completes when the downstream filters are done. The
-    filter in turn returns a publisher, and once *that* completes, we run the remaining response
-    filters.
```
graemerocher, line 282: could you replace the way this was deleted with an as detailed explanation of the new implementation?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -734,200 +535,206 @@ private static FilterReturnHandler prepareReturnHandler(Argument<?> type, boolea
...
+            return new FilterReturnHandler() {
+
                 @Override
-                protected ExecutionFlow<?> toFlow(FilterRunner runner, Object o) {
-                    //noinspection unchecked
-                    return ReactiveExecutionFlow.fromPublisher(
-                        Mono.from(Publishers.convertPublisher(runner.conversionService, o, Publisher.class))
-                            .contextWrite(runner.reactorContext));
+                public ExecutionFlow<FilterContext> handle(FilterContext context, Object returnValue, FilterContinuationImpl<?> continuation) throws Throwable {
+                    if (returnValue == null && nullable) {
```
yawkat: handle returnValue == null && !nullable

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -734,200 +535,206 @@ private static FilterReturnHandler prepareReturnHandler(Argument<?> type, boolea
...
-        FilterReturnHandler returnHandler
+    record FilterMethod<T>(FilterOrder order,
+                           T bean,
+                           Executable<T, ?> method,
+                           boolean isResponseFilter,
+                           FilterArgBinder[] argBinders,
+                           @Nullable
+                           Predicate<FilterMethodContext> filterCondition,
+                           Function<FilterContext, FilterContinuationImpl<?>> continuationCreator,
+                           int continuationIndex,
```
yawkat: unused, remove

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -734,200 +535,206 @@ private static FilterReturnHandler prepareReturnHandler(Argument<?> type, boolea
...
+                           int exceptionIndex,
```
yawkat: make boolean

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -451,234 +238,244 @@ private boolean workRequestFilter(GenericHttpFilter filter) {
...
+                                                          FilterOrder order) throws IllegalArgumentException {
         FilterArgBinder[] fulfilled = new FilterArgBinder[arguments.length];
-        Predicate<FilterRunner> filterCondition = FILTER_CONDITION_ALWAYS_TRUE;
+        Predicate<FilterMethodContext> filterCondition = FILTER_CONDITION_ALWAYS_TRUE;
         boolean skipOnError = isResponseFilter;
-
-        int passedOnContinuationArgIndex = -1;
+        Function<FilterContext, FilterContinuationImpl<?>> continuationCreator = null;
+        int continuationArgIndex = -1;
+        int exceptionArgIndex = -1;
```
yawkat: remove this variable

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -451,234 +238,244 @@ private boolean workRequestFilter(GenericHttpFilter filter) {
...
+                                                          boolean isResponseFilter,
+                                                          FilterOrder order) throws IllegalArgumentException {
         FilterArgBinder[] fulfilled = new FilterArgBinder[arguments.length];
-        Predicate<FilterRunner> filterCondition = FILTER_CONDITION_ALWAYS_TRUE;
+        Predicate<FilterMethodContext> filterCondition = FILTER_CONDITION_ALWAYS_TRUE;
         boolean skipOnError = isResponseFilter;
-
-        int passedOnContinuationArgIndex = -1;
+        Function<FilterContext, FilterContinuationImpl<?>> continuationCreator = null;
+        int continuationArgIndex = -1;
```
yawkat: make boolean

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: Moved the method before the error handle to skip ot

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -941,168 +748,256 @@ private DelayedFilterReturnHandler(boolean isResponseFilter, FilterReturnHandler
...
+     * After the proceed is triggered the filter is essentially suspended and the next filter in the chain should be executed.
      *
      * @param <R> Return value of the continuation
      */
-    private abstract class FilterContinuationImpl<R> extends SuspensionPoint implements FilterContinuation<R> {
+    private abstract static class FilterContinuationImpl<R> implements FilterContinuation<R> {
+        /**
+         * This flag checks if the proceed method has been called, to request execute the next filter in chain.
+         */
+        final AtomicBoolean proceedRequested = new AtomicBoolean(false);
```
yawkat: these atomic bools just seem to guard the future .complete, you can instead check the return value of complete

yawkat:
> I have noticed that HTTP client, filters the exceptions instead of the response. Is it correct? It looks like something it would be nice to eliminate in V4 and filter the response.

Yes this is also the current behavior. On the server, every exception has to become a response (usually with 500). This is done between each filter. On the client, exceptions in filters stay exceptions, since we can throw them from the client exchange method directly, so they don't need to become responses.

It's a bit awkward and inconsistent for the filter processing, but imo it makes sense

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -451,234 +238,240 @@ private boolean workRequestFilter(GenericHttpFilter filter) {
...
+                                                          ExecutableMethod<T, ?> method,
+                                                          boolean isResponseFilter,
+                                                          FilterOrder order) throws IllegalArgumentException {
+        return prepareFilterMethod(conversionService, bean, method, method.getArguments(), method.getReturnType().asArgument(), isResponseFilter, order);
     }
 
     @Internal
-    public static <T> FilterMethod<T> prepareFilterMethod(T bean, ExecutableMethod<T, ?> method, boolean isResponseFilter, FilterOrder order) throws IllegalArgumentException {
-        return prepareFilterMethod(bean, method, method.getArguments(), method.getReturnType().asArgument(), isResponseFilter, order);
+    public static <T> FilterMethod<T> validateFilterMethod(Argument<?>[] arguments,
```
yawkat: make void

dstepanov:
I just checked, and the tests are passing. I had to ignore one test with blocking continuation without a custom thread for these changes. With the flow style, that's not possible. The changes fixed one pending feature test.

dstepanov:
> > I have noticed that HTTP client, filters the exceptions instead of the response. Is it correct? It looks like something it would be nice to eliminate in V4 and filter the response.
> 
> Yes this is also the current behavior. On the server, every exception has to become a response (usually with 500). This is done between each filter. On the client, exceptions in filters stay exceptions, since we can throw them from the client exchange method directly, so they don't need to become responses.
> 
> It's a bit awkward and inconsistent for the filter processing, but imo it makes sense

I don't think that's correct. IMHO we should filter the response, and the client API should decide if the produced response should be an exception, or not. In some cases you want to always handle the response and converting it to the exception and back is a bit strange.

yawkat:
> I don't think that's correct. IMHO we should filter the response, and the client API should decide if the produced response should be an exception, or not. In some cases you want to always handle the response and converting it to the exception and back is a bit strange.

We could move the response->exception step to happen after response filters run, that seems sensible, if that's what you mean. But it's incompatible.

dstepanov:
Yeah, but after this PR.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -941,168 +742,244 @@ private DelayedFilterReturnHandler(boolean isResponseFilter, FilterReturnHandler
...
+            filterContext = filterContext.withRequest(Objects.requireNonNull(request, "request"));
             return this;
         }
 
-        final void triggerDownstreamWorkRequest() {
-            if (downstreamGuard.compareAndSet(false, true)) {
-                workRequest();
+        protected final void proceedRequested() {
+            if (!nextFilterProcessing.isDone()) {
+                nextFilterProcessing.complete(filterContext);
```
yawkat, line 834: As I understand it, nextFilterProcessing is completed by the proceed call, which is then checked later (via `nextFilterFlow()`) to *actually* proceed. Why is this necessary? It resolves some conflicts (e.g. if proceed was called but the method also throws an exception, it will simply not perform the proceed), but other than that it just obfuscates the control flow imo, and it's the reason why the one test had to be ignored

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: i think that is still wrong, since now onErrorResume can be called after all the other response filters run, if there is another response filter that fails

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: The `nextFilterProcessing` will make the `nextFilterFlow` emit the value, which will go to the next filter, etc. And according to the previous code it needs to proceed to the next filter only after the subscription.

The test that was ignored was continuing the flow process in the proceed method, that's why it was working.

yawkat: for reactor it has to proceed only after the subscription, because otherwise we can't propagate the context. But for CompletableFuture and synchronous proceed, it can proceed immediately

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: If all the response filters are run, nothing will be run. It will continue the response flow on error, but now it will only trigger error filters.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: Yes, the best would be to have reactive/async implementation of the continuation. And for reactive ones, it might be better to use Reactor sinks.

yawkat (approved):

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
yawkat: nothing will be run because the iterator will be back at the start in onErrorResume, correct? it's a bit confusing but i think i understand

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: Added

PR #8701, opened 2023-02-01
Merge up 3.9.x into 4.0.x

sdelamo:


core/src/main/java/io/micronaut/core/exceptions/UncheckedIOException.java
```diff
@@ -0,0 +1,29 @@
...
+package io.micronaut.core.exceptions;
+
+import java.io.IOException;
+
+/**
+ * Wraps a {@link IOException} in a {@link RuntimeException}.
+ * @author Sergio del Amo
+ * @since 3.5.6
+ */
+public class UncheckedIOException extends RuntimeException {
```
dstepanov: Can we use https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html

dstepanov (approved):

core/src/main/java/io/micronaut/core/exceptions/UncheckedIOException.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/8701/commits/1ab8748fba370bf20c03819d8e2085fbcbed1d93

PR #8704, opened 2023-02-01
Tweak uri to account for WebLogic/Windows URIs

mattmoss:
WebLogic on Windows was not checked earlier. It has some classloader URIs that start with "file:C:/path/to....".  The recent adjustment to URIs wasn't sufficient to cover this, but this new PR should work.

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/IOUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/IOUtils.java
@@ -137,11 +137,9 @@ public class IOUtils {
                 if (!jarUri.startsWith(SCHEME_FILE + COLON)) {
                     // Special case WebLogic classloader
                     // https://github.com/micronaut-projects/micronaut-core/issues/8636
-                    if (jarUri.startsWith("/")) {
-                        jarUri = SCHEME_FILE + COLON + jarUri;
-                    } else {
-                        jarUri = SCHEME_FILE + COLON + "/" + jarUri;
-                    }
+                    jarUri = jarUri.startsWith("/") ?
+                        SCHEME_FILE + COLON + jarUri :
+                        SCHEME_FILE + COLON + "/" + jarUri;
                 }
                 // now, add the !/ at the end again so that loadNestedJarUri can handle it:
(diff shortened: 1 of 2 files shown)


mattmoss:
Fixes #8636  (again)

mattmoss:
Relates to PR: https://github.com/micronaut-projects/micronaut-core/pull/8646

core/src/main/java/io/micronaut/core/io/IOUtils.java
```diff
@@ -140,7 +140,11 @@ static Path resolvePath(@NonNull URI uri,
                 if (!jarUri.startsWith(SCHEME_FILE + COLON)) {
                     // Special case WebLogic classloader
                     // https://github.com/micronaut-projects/micronaut-core/issues/8636
-                    jarUri = SCHEME_FILE + COLON + jarUri;
+					if (jarUri.startsWith("/")) {
```
mattmoss: If we want to narrow down this condition much more (to handle specifically the scheme-specific part starting with "C:" or similar, I can change this condition to check that jarUri[0] is alpha and jarUri[1] is `:`.

core/src/main/java/io/micronaut/core/io/IOUtils.java
```diff
@@ -140,7 +140,11 @@ static Path resolvePath(@NonNull URI uri,
                 if (!jarUri.startsWith(SCHEME_FILE + COLON)) {
                     // Special case WebLogic classloader
                     // https://github.com/micronaut-projects/micronaut-core/issues/8636
-                    jarUri = SCHEME_FILE + COLON + jarUri;
+                    if (jarUri.startsWith("/")) {
```
sdelamo:
```java
jarUri = jarUri.startsWith("/") ?
    SCHEME_FILE + COLON + jarUri : 
    SCHEME_FILE + COLON + "/" + jarUri;
```

maybe?

sdelamo (approved):

yawkat (approved):

sdelamo (approved):

Issue #8706, opened 2023-02-02
Annotation processor don't read field annotations with `@Target(TYPE_USE)`


altro3:
### Expected Behavior

Field annotations should be read correctly

### Actual Behaviour

Now such annotations are ignored

### Steps To Reproduce

Create class like this:
```java
@Introspected
class HelloWorldDto {

    @org.eclipse.jdt.annotation.Nullable
    public String eclipse;

    @org.jspecify.annotations.Nullable
    public String jspecify;
}
```

And check field elements in annotation processsor:
```java
element.hasStereotype("org.jspecify.annotations.Nullable")
|| element.hasStereotype("org.eclipse.jdt.annotation.Nullable")
```
Annotations will not be found

### Environment Information

- Win 11
- JDK17

### Version

3.8.3

dstepanov:
Type annotation aren't supported in Micronaut 3 and will be supported in Micronaut 4

graemerocher:
@dstepanov can this be closed?

altro3:
@dstepanov still doesn't work with micronaut 4.0

dstepanov:
Please create an example

altro3:
@dstepanov just try to run test `"test build OpenAPI with eclipse and jspecify Nullable annotations"` here https://github.com/micronaut-projects/micronaut-openapi/blob/905b1e98efb7d448cebd4fcc8ea1df26960e2353/openapi/src/test/groovy/io/micronaut/openapi/visitor/OpenApiNullableTypesSpec.groovy#L297C10-L297C77

And you will see that nullable is false. I mean, method isNullable() will return false, but we have annotation mappers for these annotations - `SpotbugsNullableTransformer` and `JspecifyNullableTransformer`

dstepanov:
@altro3 It works correctly, but in this case the nullable annotation will come on the field's type, the code needs to be adjusted.

PR #8707, opened 2023-02-02
Added test for #8706

altro3:
Added test for #8706


altro3:
@dstepanov As you remember I accidentally found a bug with Nullable annotations in micronaut-openapi . You asked for a test to reproduce the problem

dstepanov:
I'm planning to fix it for Micronaut 4

dstepanov:
BTW `@Retention(RetentionPolicy.CLASS)` shouldn't appear at runtime

PR #8712, opened 2023-02-03
Add a test for octet stream serialization to the TCK

timyates:
Inspired by https://github.com/micronaut-projects/micronaut-aws/issues/1545

Expanded the BodyAssertion to support non-string response bodies

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
@@ -29,14 +29,19 @@ import io.micronaut.http.annotation.Controller;
 import io.micronaut.http.annotation.Get;
 import io.micronaut.http.annotation.Post;
 import io.micronaut.http.server.tck.AssertionUtils;
+import io.micronaut.http.server.tck.BodyAssertion;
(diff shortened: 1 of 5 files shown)


http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
```diff
@@ -158,6 +160,16 @@ void applicationJsonWithBodyAnnotationAndObjectReturnRenderedAsJson() throws IOE
                 .build()));
     }
 
+    @Test
```
sdelamo: Can you move the test into a separate file?. Having so many tests in a single file is a mistake. I caused that. Having them in individual files allow us to exclude them more granularly.

sdelamo (changes requested):

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MiscTest.java
timyates: Good call 👍  Done in https://github.com/micronaut-projects/micronaut-core/pull/8712/commits/130433a9b866a3ecf8d4359e70910b1fbdeb0121

sdelamo (approved):

yawkat (approved):

PR #8715, opened 2023-02-03
Issues 8714: HttpClientIntroductionAdvice Subscriber may not have complete called by some Publisher implementations

rawilder:
I manually cancel the subscription as is dictated by the reactive streams spec, a subscriber must cancel when it does not need items anymore, and then completes on its own. Added a test for the specific bug I was seeing in the kotlin test suite.

I feel like this will need some fine tuning for edge cases like onComplete being called twice, perhaps by inlining the onComplete functionality and making onComplete a NOP

Unfortunately I was not able to build locally, I kept getting
```
An exception occurred applying plugin request [id: 'io.micronaut.build.internal.dependency-updates']
> Failed to apply plugin class 'io.micronaut.build.catalogs.MicronautVersionCatalogUpdatePlugin'.
   > Cannot add task 'dependencyUpdates' as a task with that name already exists.
```

Motivation for why here instead of kotlinx.coroutines:
https://github.com/Kotlin/kotlinx.coroutines/issues/3608#issuecomment-1416156578
https://github.com/reactive-streams/reactive-streams-jvm/issues/543

### DIFF
--- a/test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
+++ b/test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
@@ -2,28 +2,33 @@ package io.micronaut.http.client
 
 import io.micronaut.http.HttpResponse
 import io.micronaut.http.MutableHttpRequest
+import io.micronaut.http.annotation.Filter
 import io.micronaut.http.filter.ClientFilterChain
 import io.micronaut.http.filter.HttpClientFilter
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.flatMapMerge
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.reactive.asFlow
 import kotlinx.coroutines.reactive.asPublisher
+import kotlinx.coroutines.withContext
 import org.reactivestreams.Publisher
 
+@Filter
 class SuspendClientFilter : HttpClientFilter {
     override fun doFilter(request: MutableHttpRequest<*>, chain: ClientFilterChain): Publisher<out HttpResponse<*>> {
         // if request contains filterCheck, then do flow step, else proceed
-        if ((request.body.get() as String).contains(SuspendClientFilter.filterCheck)) {
-            return flow { emit("testString") }.flatMapMerge {
+        return if ((request.body.orElse(null) as? Map<*, *>)?.containsValue(filterCheck) == true) {
+            flow { emit(getValue()) }.flatMapMerge {
                 chain.proceed(request.header("myTestHeader", it)).asFlow()
             }.asPublisher()
         } else {
-            return chain.proceed(request)
+            chain.proceed(request)
         }
     }
 
     companion object {
         val filterCheck = java.util.UUID.randomUUID().toString()
+        suspend fun getValue() = withContext(Dispatchers.Default) { "testString" }
     }
 }
 
--- a/test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
+++ b/test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
@@ -1,0 +1,34 @@
+package io.micronaut.http.client
+
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.MutableHttpRequest
+import io.micronaut.http.annotation.Filter
+import io.micronaut.http.filter.ClientFilterChain
+import io.micronaut.http.filter.HttpClientFilter
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.flatMapMerge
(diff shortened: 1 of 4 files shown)


rawilder:
I read through the errors and am fairly certain they are not caused by my change so I am marking this ready for review.

yawkat:
The test does not work. It passes even without the change you made to HttpClientIntroductionAdvice. Please take a look at it.

rawilder:
Hmm interesting. I am trying to replicate the behavior and I am also having trouble. My original issue example was 3.8.x so I will try rebasing and see if I replicate it. So far it looks like some intermediary subscriber is requesting 32 instead of the 1 I was seeing with my example project. Will test and see what I find.

rawilder:
That's exactly what it was. I am going to rebase this PR against 3.8.x. I had just assumed that 3.7.x was the culprit as it had the large client rewrite. My mistake!

Failed test:
image

Modification to cause failure:
image

yawkat (approved):

yawkat:
ive changed it slightly to match the MonoNext implementation, and removed the unnecessary atomicref

test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientSpec.kt
```diff
@@ -42,4 +43,14 @@ class SuspendClientSpec {
...
     }
 
+    @Test
+    fun testFlowAsPublisherInFilterStep() {
+        val server = ApplicationContext.run(EmbeddedServer::class.java, mapOf("spec.name" to "SuspendClientSpec"))
+        val ctx = server.applicationContext
+        val response = runBlocking {
+            ctx.getBean(SuspendClient::class.java).call(SuspendClientFilter.filterCheck)
+        }
+        Assertions.assertEquals(response, "{\"newState\":\"${SuspendClientFilter.filterCheck}\"}")
```
sdelamo, line 55: we should close the server at the end of the test with `server.close()`

sdelamo (changes requested):

test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.coroutines.withContext
+import org.reactivestreams.Publisher
+
+@Filter
+class SuspendClientFilter : HttpClientFilter {
+    override fun doFilter(request: MutableHttpRequest<*>, chain: ClientFilterChain): Publisher<out HttpResponse<*>> {
+        // if request contains filterCheck, then do flow step, else proceed
+        return if ((request.body.orElse(null) as? Map<*, *>)?.containsValue(filterCheck) == true) {
+            flow { emit(getValue()) }.flatMapMerge {
+                chain.proceed(request.header("myTestHeader", it)).asFlow()
```
sdelamo: Are we doing anything with the HTTP Header `myTestHeader`. Should we test it?

sdelamo (changes requested):

test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientSpec.kt
sdelamo: I added a [commit to close the server](https://github.com/micronaut-projects/micronaut-core/pull/8715/commits/09b140d38ef3ff5fae5a3bfb130d038a6c8f4395)

test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
rawilder: It is simply a usage of the value from the flow publisher. It is not necessary to test. We could easily emit and using nothing and it would have the same effect I imagine.

yawkat: ok i think it's fine as-is

sdelamo: if we don't verify anything, I would prefer to replace: 

```java
 chain.proceed(request.header("myTestHeader", it)).asFlow()
```

with: 

```java
 chain.proceed(request).asFlow()
```

thoughts?

sdelamo:
```suggestion
                chain.proceed(request).asFlow()
```

sdelamo (approved):

test-suite-kotlin/src/test/kotlin/io/micronaut/http/client/SuspendClientFilter.kt
rawilder: I see it's already merged but as long as FlowAsPublisher is somewhere in the chain it should cause the behavior.

PR #8717, opened 2023-02-06
ci: run japiCmp only once

sdelamo:
This PR runs binary compatibility check only once for JDK 17. This change avoids multiple Github API requests, which may cause a rate-limiting failure.

### DIFF
--- a/.github/workflows/corretto.yml
+++ b/.github/workflows/corretto.yml
@@ -35,4 +35,4 @@ jobs:
         GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
         GRADLE_ENTERPRISE_CACHE_USERNAME: ${{ secrets.GRADLE_ENTERPRISE_CACHE_USERNAME }}
         GRADLE_ENTERPRISE_CACHE_PASSWORD: ${{ secrets.GRADLE_ENTERPRISE_CACHE_PASSWORD }}
-      run: unset HOSTNAME ; LANG=en_US.utf-8 LC_ALL=en_US.utf-8 ./gradlew check --no-daemon --parallel --continue
+      run: unset HOSTNAME ; LANG=en_US.utf-8 LC_ALL=en_US.utf-8 ./gradlew check --no-daemon --parallel --continue -x japiCmp
--- a/.github/workflows/graalvm.yml
+++ b/.github/workflows/graalvm.yml
@@ -50,9 +50,9 @@ jobs:
         run: |
           if ./gradlew tasks --no-daemon --all | grep -w "testNativeImage"
           then
-            ./gradlew check testNativeImage --continue --no-daemon
+            ./gradlew check testNativeImage -x japiCmp --continue --no-daemon
           else
-            ./gradlew check --continue --no-daemon
+            ./gradlew check --continue --no-daemon -x japiCmp
           fi
         env:
           TESTCONTAINERS_RYUK_DISABLED: true
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Build with Gradle
         id: gradle
         run: |
-          ./gradlew check --no-daemon --parallel --continue
+          ./gradlew check --no-daemon --parallel --continue -x japiCmp
         env:
           GH_TOKEN_PUBLIC_REPOS_READONLY: ${{ secrets.GH_TOKEN_PUBLIC_REPOS_READONLY }}
           GH_USERNAME: ${{ secrets.GH_USERNAME }}
@@ -74,8 +74,18 @@ jobs:
           check_name: Java CI / Test Report (${{ matrix.java }})
           report_paths: '**/build/test-results/test/TEST-*.xml'
           check_retries: 'true'
+      - name: Check Binary Compatibility
+        run: |
+          ./gradlew japiCmp
+        env:
+          GH_TOKEN_PUBLIC_REPOS_READONLY: ${{ secrets.GH_TOKEN_PUBLIC_REPOS_READONLY }}
+          GH_USERNAME: ${{ secrets.GH_USERNAME }}
+          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          GRADLE_ENTERPRISE_CACHE_USERNAME: ${{ secrets.GRADLE_ENTERPRISE_CACHE_USERNAME }}
+          GRADLE_ENTERPRISE_CACHE_PASSWORD: ${{ secrets.GRADLE_ENTERPRISE_CACHE_PASSWORD }}
+        if: matrix.java == '17'
       - name: "📜 Upload binary compatibility check results"
-        if: always()
+        if: matrix.java == '17'
         uses: actions/upload-artifact@v3
         with:
           name: binary-compatibility-reports
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -48,7 +48,7 @@ jobs:
           [ -f ./setup.sh ] && ./setup.sh || true
       - name: Analyse with Gradle
         run: |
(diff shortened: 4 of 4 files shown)


melix:
This shouldn't change anything since the subsequent runs of `japicmp` will be cached, as we can see on https://ge.micronaut.io/s/ykyokl262e5pc/timeline?task-path=japi for example.

sdelamo:
> This shouldn't change anything since the subsequent runs of `japicmp` will be cached, as we can see on https://ge.micronaut.io/s/ykyokl262e5pc/timeline?task-path=japi for example.

Unless, I am mistaken. We currently run binary compatiblity checks in every matrix combination for gradle workflow. Every matrix combination for GraalVM workflow, every matrix of sonarqube. 

Is jampCamp Github release API response cached between different matrix combinations and different workflows? 

What is the benefit to run binary compatiblity check over and over again for different matrix combinations? Should not be enough to run it once for Java 17?

melix:
It doesn't matter how many workflows use the task. It does matter, however, which Java version is used. For a single version of Java, say 17, then the first workflow to run the task will populate the cache.

Then, all other workflows will fetch the results from the cache. There shouldn't be any request on GitHub releases API, because we are using an `@Internal` annotation here: https://github.com/micronaut-projects/micronaut-build/blob/3b15147d25d7c401c18a08772dba93e432ca7539/src/main/java/io/micronaut/build/compat/FindBaselineTask.java#L61

Internal properties do not participate in up-to-date checking, and since this method is only used in the method body, it will _not_ be executed. Therefore, the only thing which matters is [the timestamp](https://github.com/micronaut-projects/micronaut-build/blob/3b15147d25d7c401c18a08772dba93e432ca7539/src/main/java/io/micronaut/build/compat/FindBaselineTask.java#L53) which means that we cache the JSON result for an hour.

melix:
Note that in any case, disabling japicmp will have no effect if the "docs" task is called, since the task which performs GitHub API requests is `findbaseLine`. That task is a dependency of both `japicmp` and `docs` tasks.

sdelamo:
> It doesn't matter how many workflows use the task. It does matter, however, which Java version is used. For a single version of Java, say 17, then the first workflow to run the task will populate the cache.

Will the different JDK version matter? 
 `graalvm.yml` workflow uses GraalVM CE while `gradle.yml` workflow uses  `temurin`.  

Will different JDKs and different Java versions of the same JDK miss the cache? 

If the answer to the previous question is yes, this PR should be beneficial to avoid rate limiting. Am I correct?

> Note that in any case, disabling japicmp will have no effect if the "docs" task is called, since the task which performs GitHub API requests is findbaseLine. That task is a dependency of both japicmp and docs tasks.

We run `docs` task only on `gradle.yml` workflow with the condition `if: success() && github.event_name == 'push' && matrix.java == '17'` and on `release.yml` workflow. Thus, I suspect it is less harmful to the rate-limiting issues. 

Apart from slight-lightly complicating the workflows with some task exclusions and an extra step. Is there any downside of this PR that I am missing?

melix:
> Will the different JDK version matter?

AFAIK, the version of the JDK should matter (11, 17, ...) but not the vendor.

> Apart from slight-lightly complicating the workflows with some task exclusions and an extra step. Is there any downside of this PR that I am missing?

Well this is what worries me: it complicates the workflows and further diverges from the templates, for something I doubt will have an impact, since as I explained (and the scans show), the `findbaseLine` task is cached.

sdelamo:
>  further diverges from the templates, 

Please, note that if we agree on doing this change in [core](https://github.com/micronaut-projects/micronaut-core/). I will do a PR to [project-template](https://github.com/micronaut-projects/micronaut-project-template) with the same change. 

> I doubt will have an impact, 

Based on the conversation. Since we run against multiple vendors in this module and every other module (corretto - only on core , graalvm, and temurin), it should have an impact. 

I don't know if it is the complete solution to the rate-limiting issues but it should have an impact.

melix:
> Based on the conversation. Since we run against multiple vendors in this module and every other module (corretto - only on core , graalvm, and temurin), it should have an impact.

No, the vendor should _not_ change anything. The _version_ of Java matters.

melix:
So in case there were doubts that the vendor doesn't matter, I have tried it locally:

First run with GraalVM 17: [japiCmp](https://ge.micronaut.io/s/ufsoecuofevrg/timeline?details=jvjelakpi3fhy&task-path=japicmp)

Then second run with Temurin, the task is cached: [japiCmp](https://ge.micronaut.io/s/mbcmuiijygd3s/timeline?details=jvjelakpi3fhy&task-path=japi) and [findbaseLine](https://ge.micronaut.io/s/mbcmuiijygd3s/timeline?details=osmazn27rfgv6&task-path=findb)

sdelamo:
> > Based on the conversation. Since we run against multiple vendors in this module and every other module (corretto - only on core , graalvm, and temurin), it should have an impact.
> 
> No, the vendor should _not_ change anything. The _version_ of Java matters.

Are we checking binary compatibility four times (one per java 8, 11, 17, 19) and doing four requests when it would suffice to do it only once?

melix:
> Are we checking binary compatibility four times (one per java 8, 11, 17, 19) and doing four requests when it would suffice to do it only once?

Yes, that would have more impact than this change. However, that's what I was saying, the problem is not `japiCmp`, it's `findbaseLine`. And this task is required for `japiCmp`, `checkVersionCatalogCompatibility` and `createDropdown`, which means that if you write something like this:

```gradle
tasks.named("findBaseline") {
    onlyIf {
        JavaVersion.current().majorVersion == "17"
    }
}
```

and that the build runs a task which needs its output, then it will break with an error like this:

```
* What went wrong:
Execution failed for task ':bom:checkVersionCatalogCompatibility'.
> Failed to calculate the value of task ':bom:checkVersionCatalogCompatibility' property 'baseline'.
   > Expected configuration ':bom:detachedConfiguration1' to contain exactly one file, however, it contains no files.

```

sdelamo:
we are missing something. right now we are already rate limited in core: 

```
Limit: 5000
Remaining: 0
Reset: 1675679975
Rate-limiting resets in 00:02:37
Used: 5000
Resource: core
```

melix:
So I think https://github.com/micronaut-projects/micronaut-build/pull/484 can reduce the number of requests, typically because when builds are executed in parallel, we can do the same request 4 times. This means that even if we have cache misses, the PR linked should reduce the number of requests by a factor of the number of concurrent tasks.

However, if we have PRs which are not using the latest build plugins (including our latest patches), then they probably perform too many requests or unauthenticated requests.

sdelamo:
Just to clarify, rate-limit applies per user, not per repository. Thus, every repository github action contributes for use being closer to the rate limit.  

> User-to-server requests are limited to 5,000 requests per hour and per authenticated user.

sdelamo:
we have something which clearly issues  a lot of requests

```
 % java -jar build/libs/github-rate-limiting-0.1-all.jar -u micronaut-build -t 
11:36:51.576 [main] INFO  i.m.context.env.DefaultEnvironment - Established active environments: [cli]
Enter value for --token (Github Token): 
Limit: 5000
Remaining: 0
Reset: 1675679975
Rate-limiting resets in 00:02:37
Used: 5000
Resource: core
sdelamo@ig11 github-rate-limiting % ./gradlew build                                                             


BUILD SUCCESSFUL in 1s
15 actionable tasks: 15 up-to-date
sdelamo@ig11 github-rate-limiting % java -jar build/libs/github-rate-limiting-0.1-all.jar -u micronaut-build -t 
11:59:37.876 [main] INFO  i.m.context.env.DefaultEnvironment - Established active environments: [cli]
Enter value for --token (Github Token): 
Limit: 5000
Remaining: 0
Reset: 1675683582
Rate-limiting resets in 00:40:00
Used: 5004
Resource: core
```

The limit reset and in less than 20 minutes we were again off the limit.

sdelamo:
And we get build failures: https://github.com/micronaut-projects/micronaut-core/actions/runs/4102792663/jobs/7076185289

sdelamo:
We got another rate limiting even with 5.4.5 https://github.com/micronaut-projects/micronaut-core/actions/runs/4103519712/jobs/7077817429

PR #8723, opened 2023-02-06
Correct messaging event publishing

### LINKED ISSUE #8673: Regression in ApplicationEventShutdown
### Issue description

It seems that a regression was recently introduced in Micronaut 4 snapshots. It happens systematically on some projects which use messaging, so it's probably a shutdown hook of some kind which is now broken.

Stack trace shows:

```
io.micronaut.context.exceptions.NoSuchBeanException: No bean of type [io.micronaut.context.event.ApplicationEventPublisher<io.micronaut.runtime.event.ApplicationShutdownEvent>] exists. Make sure the bean is not disabled by bean requirements (enable trace logging for 'io.micronaut.context.condition' to check) and if the bean is enabled then ensure the class is declared a bean and annotation processing is enabled (for Java and Kotlin the 'micronaut-inject-java' dependency should be configured as an annotation processor).
```

https://ge.micronaut.io/s/dqjgvdjpc36ia/tests/:test-resources-hivemq:test/io.micronaut.testresources.hivemq.HiveMQStartedTest?top-execution=1

To reproduce, checkout `micronaut-test-resources` branch `master` and run:

`./gradlew :t-r-hivemq:test`

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/8673

### DIFF
--- a/messaging/src/main/java/io/micronaut/messaging/MessagingApplication.java
+++ b/messaging/src/main/java/io/micronaut/messaging/MessagingApplication.java
@@ -80,13 +80,15 @@ public class MessagingApplication implements EmbeddedApplication<MessagingApplic
     @NonNull
     public MessagingApplication start() {
         ApplicationContext applicationContext = getApplicationContext();
-        if (applicationContext != null && !applicationContext.isRunning()) {
-            try {
-                applicationContext.start();
-                applicationContext.publishEvent(new ApplicationStartupEvent(this));
-            } catch (Throwable e) {
-                throw new ApplicationStartupException("Error starting messaging server: " + e.getMessage(), e);
+        if (applicationContext != null) {
+            if (!applicationContext.isRunning()) {
+                try {
+                    applicationContext.start();
+                } catch (Throwable e) {
+                    throw new ApplicationStartupException("Error starting messaging server: " + e.getMessage(), e);
+                }
             }
+            applicationContext.publishEvent(new ApplicationStartupEvent(this));
         }
         return this;
     }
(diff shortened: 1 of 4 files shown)


messaging/src/main/java/io/micronaut/messaging/MessagingApplication.java
```diff
@@ -80,13 +80,15 @@ public boolean isServer() {
...
-                throw new ApplicationStartupException("Error starting messaging server: " + e.getMessage(), e);
+        if (applicationContext != null) {
+            if (!applicationContext.isRunning()) {
+                try {
+                    applicationContext.start();
+                } catch (Throwable e) {
+                    throw new ApplicationStartupException("Error starting messaging server: " + e.getMessage(), e);
+                }
             }
+            applicationContext.publishEvent(new ApplicationStartupEvent(this));
```
graemerocher, line 91: should this not be within he `if` block?

dstepanov: No, it will not be triggered within the `isRunning` check. Netty has it the same but introduces custom events that extend application events.

graemerocher: seems inconsistent with the `stop` implementation

dstepanov: In what way? It looks fine to me.

graemerocher: if the context is already running then calling `start()` will fire the `ApplicationStartupEvent` event again even if the context is already running which is not the same as the previous behaviour?

dstepanov: `ApplicationStartupEvent` is not fired when `start` is called. If I revert the code, the test will fail because the event is never published.

graemerocher (approved):

PR #8736, opened 2023-02-07
Fix NPE producing anntotion metadata for KSP

graemerocher:
Fixes the failures within KSP found here https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/634#issuecomment-1408855293

Also forks the kotlin compiler testing project because it is not compatible with the latest version of Kotlin/KSP, plus does some cleanup.

- fork compiler testing project
- cleanup
- don't include null values in arrays in metadata for KSP


melix (approved):
LGTM, I will probably take a look at the build scripts in the near future since there are a few things I would like to improve (e.g avoiding duplication of dependencies which exist in micronaut test).

PR #8738, opened 2023-02-08
ci: GraalVM only latest and 17, 19

sdelamo:
Until we support 19 and dev.


graemerocher:
looks like we need to pass `--enable-preview` to native compile

PR #8740, opened 2023-02-08
Groovy annotations and generic improvements

dstepanov:
Unfortunately, Groovy's `ClassNode` is not contextually equal, so I had to implement a custom native type for Groovy.
This supports reading annotations of generic arguments and type annotations. I also added more tests for generic types and fixed some cases.


inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyNativeElement.java
```diff
@@ -0,0 +1,103 @@
...
+import org.codehaus.groovy.ast.MethodNode;
+import org.codehaus.groovy.ast.PackageNode;
+
+/**
+ * Groovy's native element.
+ *
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+public interface GroovyNativeElement {
```
graemerocher: can be sealed

graemerocher (approved):

Issue #8661, opened 2023-01-25
Micronaut Http as Function


driverpt:
### Feature description

Hello,

Is it possible to have something like

```
@FunctionalInterface
public interface MicronautHttpServer {
   HttpResponse<?> serveHttp(HttpRequest<?> request) throws Throwable
}

@FunctionalInterface
public interface MicronautAsyncHttpServer {
   Publisher<HttpResponse<?>> serveHttp(HttpRequest<?> request) throws Throwable
}

```

Similar to what we have in [Go](https://pkg.go.dev/net/http#HandleFunc) ? 

The idea is to develop wrappers on top of Micronaut as a closed box.

graemerocher:
Triage: this is a valid HTTP composition enhancement. A stable function-style adapter could expose the request lifecycle as a synchronous or asynchronous handler without requiring a controller class, while still applying routing, filters, binding, error handling, and response writing consistently. The likely design is an explicit handler/embedded-server bridge rather than bypassing RouteExecutor; define lifecycle ownership and backpressure semantics before adding the API.

PR #8741, opened 2023-02-08
Micronaut Http Server exposed as Functions

driverpt:
Attempts to address #8661


driverpt:
@sdelamo , can you let me know once RouteExecutor is stable ? So i can resume this PR

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -524,6 +524,10 @@ ExecutionFlow<MutableHttpResponse<?>> createResponseForBody(HttpRequest<?> reque
                 }
                 response.body(null);
             }
+            if (response.getContentType().isEmpty()) {
+                MediaType mediaType = resolveDefaultResponseContentType(request, routeInfo);
+                response.contentType(mediaType);
+            }
```
graemerocher, line 530: not sure about this change, why is this necessary?

graemerocher:
Seems to be missing a `build.gradle` definition. What is the purpose of this implementation?

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
driverpt: Hello @graemerocher , according to testing, if this is not set, no `Content-Type` is set on response

driverpt:
@graemerocher , the purpose of this would be to expose Micronaut Http as a Function (similar to what exists in Go). The purpose would be to facilitate the Integration of Micronaut with Function-based Frameworks (AWS Lambda, Google Cloud Function, Azure Function, etc...). I have created an example in [this PR](https://github.com/micronaut-projects/micronaut-aws/pull/1595/files#diff-7b4dacb23e848467cdf4a8f6badd2f4a6e720526670e05f52da8a6180240c94fR52-R64)

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
graemerocher: don't think the core Router should be changed for this, btw there are already implementation of the routing for serverless that are based on the Servlet module like https://github.com/micronaut-projects/micronaut-gcp/tree/master/gcp-function-http

graemerocher:
I understand that but we already have integration with Lambda and GCP that use the Servlet module, not sure of the need for a new one? https://github.com/micronaut-projects/micronaut-gcp/tree/master/gcp-function-http

driverpt:
This is what i'm trying to achieve: https://github.com/awslabs/aws-lambda-go-api-proxy/blob/master/echo/adapterv2.go .

driverpt:
> I understand that but we already have integration with Lambda and GCP that use the Servlet module, not sure of the need for a new one? https://github.com/micronaut-projects/micronaut-gcp/tree/master/gcp-function-http

AFAIK, AWS Lambda doesn't use Servlet

graemerocher:
no it uses aws serverless container underneath. Is there something deficient with this module that means we need 2 ways?

driverpt:
With all due respect to the maintainers, AWS Serverless Container brings more issues than the ones it solves.

AWS has 3 types of payloads (Payload V1, Payload V2 and ALB Payload, the latter is almost equal to V2) for Http Serving Lambdas. 2 of them are similar, but from Payload V1 to V2 there are major differences. IMHO this can be easily solved by abstracting the input to Micronaut and just implement an HttpRequest mapping layer.
And I'm pretty sure that in the future there will be more Payload types.

driverpt:
@graemerocher @sdelamo , just to clarify. Should I cancel this PR as it will not be implemented or can I keep it open?

sdelamo:
I agree with you, AWS Serverless container has served as well but it  is problematic. I think with the server tck we are in a position to get rid of its dependency. We are trying to solve it in the AWS module https://github.com/micronaut-projects/micronaut-aws/pull/1648

driverpt:
I've added some comments about the discoveries I've made. Let me know if I can further help you.

sdelamo:
Closing this for now as I think we have a good support in Lambda for the three kinds of payloads Payload V1, Payload V2 and ALB Payload without the use of  AWS Serverless Container.

PR #8746, opened 2023-02-09
Fix regression where AliasFor is no longer recursive 

graemerocher:
The Micronaut Test module is failing because bean replacements are not working correctly. It seems to be a regression in the way `@AliasFor` is processed since the change here works around the issue.


inject-groovy/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
```diff
@@ -37,7 +37,7 @@ class MyBean  {
         beanDefinition.getInjectedMethods()[0].name == 'setMyValue'
         def metadata = beanDefinition.getInjectedMethods()[0].getAnnotationMetadata()
         metadata.hasAnnotation(Property)
-        metadata.getValue(Property, "name", String).get() == 'endpoints.my-value'
+        metadata.getValue(Property, "name", String).get() == 'simple.my-value'
```
dstepanov, line 40: Both tests are "testing" the broken functionality of `ConfigurationReader` where the value can be aliased multiple times incorrectly, correct use-case is for `TestEndpoint1` to use `base` value.

dstepanov (approved):

PR #8751, opened 2023-02-09
Allow user to explicity turn off the localhost drive-by CORS check

### LINKED ISSUE #8749: Localhost Bound Server Blocks CORS Requests for other origins
### Expected Behavior

1. Server running bound to `127.0.0.1:8080`
    1. CORS disabled (`micronaut.server.cors.enabled` unset or false)
3. Reverse Proxy Application running on same VM but accepting traffic on 0.0.0.0:80
4. CORS related request is sent with a load balancer URL that is eventually forwarded to the Proxy container
5. Proxy container performs successful CORS validation and forwards API call to 127.0.0.1:8080 with the `Origin` header set to the previously mentioned load balancer hostname
6. Micronaut processes requests as normal

### Actual Behaviour

1. Server running bound to `127.0.0.1:8080`
    1. CORS disabled (`micronaut.server.cors.enabled` unset or false)
3. Reverse Proxy Application running on same VM but accepting traffic on 0.0.0.0:80
4. CORS related request is sent with a load balancer URL that is eventually forwarded to the Proxy container
5. Proxy container performs successful CORS validation and forwards API call to 127.0.0.1:8080 with the `Origin` header set to the previously mentioned load balancer hostname
6. Micronaut responds with a 403 with no visible logging (unless TRACE is enabled)
    1. I believe this is related to #8582

### Steps To Reproduce

1. Run a Micronaut web service bound to 127.0.0.1
2. Make an OPTIONS request against the service with origin set to something other than localhost

### Environment Information

- Linux (x86_64) / macOS (ARM)
- JDK 11

### Example Application

_No response_

### Version

3.8.3

timyates:
To disallow drive-by localhost attacks on a server under development, we added an Origin check for all requests even if Cors itself was disabled.

However, under certain circumstances (such as those described in #8749), this is not the required behavior.

This PR adds a new configuration setting `micronaut.server.cors.localhost-pass-through` which defaults to false (for the current behavior).

Setting this to `true` will skip this localhost/origin test and the request will be allowed."

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
+++ b/http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
@@ -664,9 +664,11 @@ public class HttpServerConfiguration implements ServerContextPathProvider {
 
         public static final boolean DEFAULT_ENABLED = false;
         public static final boolean DEFAULT_SINGLE_HEADER = false;
+        public static final boolean DEFAULT_LOCALHOST_PASS_THROUGH = false;
 
         private boolean enabled = DEFAULT_ENABLED;
         private boolean singleHeader = DEFAULT_SINGLE_HEADER;
+        private boolean localhostPassThrough = DEFAULT_LOCALHOST_PASS_THROUGH;
 
         private Map<String, CorsOriginConfiguration> configurations = Collections.emptyMap();
 
@@ -678,6 +680,13 @@ public class HttpServerConfiguration implements ServerContextPathProvider {
         @Override
         public boolean isEnabled() {
             return enabled;
+        }
+
+        /**
+         * @return Whether localhost pass-through is enabled. Defaults to {@value #DEFAULT_LOCALHOST_PASS_THROUGH}.
+         */
+        public boolean isLocalhostPassThrough() {
+            return localhostPassThrough;
         }
 
         /**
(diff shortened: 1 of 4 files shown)


http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -708,6 +717,16 @@ public void setEnabled(boolean enabled) {
             this.enabled = enabled;
         }
 
+        /**
+         * Sets whether localhost pass-through is enabled. Default value {@value #DEFAULT_LOCALHOST_PASS_THROUGH}.
+         * Setting this to true will allow requests to be made to localhost from any origin.
+         *
+         * @param localhostPassThrough True if localhost pass-through is enabled
+         */
+        public void setLocalhostPassThrough(boolean localhostPassThrough) {
```
graemerocher, line 728: add `@since` declarations to the new methods

graemerocher (approved):

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
timyates: 👍

sdelamo (approved):

PR #8761, opened 2023-02-10
refactor: extract ContextPathUtils and HttpClientExceptionUtils

sdelamo:


http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientExceptionUtils.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2017-2022 original authors
```
timyates, line 2: 2023 😜

timyates (approved):

PR #8777, opened 2023-02-14
Mark config that isn't for javanet client as such

timyates:


http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -237,6 +237,8 @@ public void setHttpVersion(HttpVersion httpVersion) {
     }
 
     /**
+     * [not in java.net HttpClient].
```
sdelamo: why is this only for netty?

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -314,6 +318,8 @@ public void setExceptionOnErrorStatus(boolean exceptionOnErrorStatus) {
...
```
sdelamo: why is this only for netty?

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -369,6 +379,8 @@ public void setChannelOptions(Map<String, Object> channelOptions) {
...
```
sdelamo: should not we support this for Java net http Client as well? https://www.baeldung.com/java-httpclient-timeout

sdelamo (changes requested):
I would prefer to say that it is available only in the Netty HTTP Client instead of telling in which clients is not available.

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
timyates: Agreed, fixed in https://github.com/micronaut-projects/micronaut-core/pull/8441/commits/eac7562a4c0cefe86abd7e1961cc2696b67b0894

and 

https://github.com/micronaut-projects/micronaut-core/pull/8777/commits/5db898f2d6d8004e03fd4236bae21b7d6ffa2515

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
timyates: Fixed in https://github.com/micronaut-projects/micronaut-core/pull/8441/commits/1d85bc88b501c2af65fe06616de24ff8f2e85a12 and https://github.com/micronaut-projects/micronaut-core/pull/8777/commits/596807a53f579eaf423bec7da1b7c84875708723

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
timyates: This is for setting the internal Netty log level.

There is nothing I can find in the java HttpClient, other than setting a system property to turn on logging

ie:

https://github.com/micronaut-projects/micronaut-core/blob/a4abbecd37c982157b2335a43f0dde13e0317dc2/http-client-javanet/build.gradle.kts#L22

A lot of HttpClient seems to be hidden away, and only programmable via system properties :-(

sdelamo (approved):

PR #8779, opened 2023-02-15
Add SSL support via SslBuilder implementation

timyates:
This adds support for provided keystores and certificates.

With HttpClient, there's no way of setting the host-check in ssl without using a system property.

This causes issues under test, as the properties are cached internally, so test fail depending on the order they are executed.

For this reason, the tests run with `forkEvery` set to 1, so we get a fresh JVM for each specification.


sdelamo (commented):
Can we use something like @RestoreSystemProperties in Spock but ofr Junit?

timyates:
No, because it seems to cache the values.  Or at least the first time it checks the setting is what is used for the remainder of the tests.

So when the tests run in a certain order and the "hostname checking not required" tests run last, all is good

Otherwise, we get failures.

I believe the issue is what is requested here https://bugs.openjdk.org/browse/JDK-8213309

timyates:
Ok, created `test-suite-http-client-javanet-ssl` and moved the hostname ssl tests over there

test-suite-http-client-javanet-ssl/src/test/groovy/io/micronaut/http/client/javanet/SslStaticCertSpec.groovy
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    void "expect the url to be https"() {
+        expect:
+        embeddedServer.getURL().toString() == "https://${host}:${embeddedServer.port}"
+    }
+
+    void "test send https request"() {
+        when:
+        Flux<HttpResponse<String>> reactiveSequence = Flux.from(client.exchange(
```
sdelamo, line 61: why not just write the test with `client.toBlocking` instead of the `Flux`. Are we testing anything that we need to use a reactive sequence in the test?

sdelamo (changes requested):

test-suite-http-client-javanet-ssl/src/test/groovy/io/micronaut/http/client/javanet/SslStaticCertSpec.groovy
timyates: We have 2 clients effectively, calling `toBlocking` gives you a `JavanetBlockingHttpClient` which has a different `exchange` method to the non-blocking version.

The client TCK should be updated to check both, currently it only tests the blocking one IIRC

timyates: https://github.com/micronaut-projects/micronaut-core/pull/8790

http-client-javanet/src/main/java/io/micronaut/http/client/javanet/AbstractJavanetHttpClient.java
```diff
@@ -152,25 +150,28 @@ protected PasswordAuthentication getPasswordAuthentication() {
...
-            } else {
-                sslContext = SSLContext.getDefault();
-            }
-            return sslContext;
-        } catch (GeneralSecurityException e) {
-            throw new SslConfigurationException(e);
+    private void configureSsl(HttpClient.Builder builder, ClientSslConfiguration clientSslConfiguration) {
+        sslBuilder.build(clientSslConfiguration).ifPresent(builder::sslContext);
+        if (System.getProperty("jdk.internal.httpclient.disableHostnameVerification") != null && log.isWarnEnabled()) {
+            log.warn("The jdk.internal.httpclient.disableHostnameVerification system property is set. This is not recommended for production use.");
```
yawkat: also mention why it's not recommended

yawkat:
Please add the badssl tests from SslSpec

timyates:
@yawkat These ones? https://github.com/micronaut-projects/micronaut-core/blob/javanet-ssl/http-client-javanet/src/test/groovy/io/micronaut/http/client/javanet/SslSpec.groovy

yawkat:
oh good. do you know why the dh tests are passing? that is slightly concerning.

http-client-javanet/src/main/java/io/micronaut/http/client/javanet/AbstractJavanetHttpClient.java
timyates: https://github.com/micronaut-projects/micronaut-core/pull/8779/commits/48fd0a1491f3d4e09c61943e5a1be627369f0724

timyates:
@yawkat no, I'm not sure 😔 I'll see if I can find anything

http-client-javanet/src/main/java/io/micronaut/http/client/javanet/AbstractJavanetHttpClient.java
```diff
@@ -153,7 +153,7 @@ protected PasswordAuthentication getPasswordAuthentication() {
     private void configureSsl(HttpClient.Builder builder, ClientSslConfiguration clientSslConfiguration) {
         sslBuilder.build(clientSslConfiguration).ifPresent(builder::sslContext);
         if (System.getProperty("jdk.internal.httpclient.disableHostnameVerification") != null && log.isWarnEnabled()) {
-            log.warn("The jdk.internal.httpclient.disableHostnameVerification system property is set. This is not recommended for production use.");
+            log.warn("The jdk.internal.httpclient.disableHostnameVerification system property is set. This is not recommended for production use as it prevents proper certificate validation may allow man-in-the-middle attacks.");
```
yawkat: i think this is missing a word

yawkat (approved):

timyates:
I cannot see why the dh-tests are passing...  They also pass in curl in bash...

I've added a pending feature for them so we can see if they get fixed

PR #8781, opened 2023-02-15
Javanet client works with http proxies

timyates:
This should pass, but currently fails


sdelamo:
@timyates is this in draft or ready for review?

timyates:
I was working on it.

It's now ready for review 🎉

___.clicks button.___

PR #8784, opened 2023-02-16
build: Apache groovy 4.0.9

sdelamo:


inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/LoadedVisitor.groovy
```diff
@@ -94,6 +95,7 @@ class LoadedVisitor implements Ordered {
         return true
     }
 
+    @CompileDynamic
```
sdelamo, line 98: without `@CompilieDynamic`

```
 [Static type checking] - Non-static method java.lang.Object#hashCode cannot be called from static context
 @ line 99, column 16.
           return visitor.getClass().hashCode()
```

graemerocher: Seems like a regression in Groovy

graemerocher (approved):

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/LoadedVisitor.groovy
sdelamo: it is present since 4.0.7

sdelamo: I submitted an issue https://issues.apache.org/jira/browse/GROOVY-10939

PR #8801, opened 2023-02-20
Draft what's new guide

graemerocher:


graemerocher:
@sdelamo we should work on draft changes to the what's new guide here

PR #8803, opened 2023-02-20
Buffered JSON parsing

yawkat:
Currently, we use a non-blocking parser to transform input JSON to a JsonNode, then send it downstream, and finally convert it to the desired request type when the route arguments are bound. This patch delays parsing until the conversion to the route type. It consists of these parts:

- A change to JsonContentProcessor. The JsonContentProcessor has to handle certain scenarios (`application/x-json-stream`, or our support for streaming a json array as a Publisher) where the input contains multiple json tokens, Instead of forwarding the input bytes to an async parser, they are first split into individual json values, that are then forwarded or parsed in bulk.
- A new JsonCounter class. This class handles splitting the input into its individual json nodes. It is basically a limited JSON parser that counts braces. It is already very optimized, but it also has a special optimization when the content type is `application/json`, in which case it assumes only a single json value in the input (this is technically a breaking change, but breaks no tests). I also wrote fuzz tests for JsonCounter, I will submit them in a separate PR to https://github.com/micronaut-projects/micronaut-fuzzing .
- A benchmark. It's not run by the build tools, but it is good to have around.
- An option in NettyHttpServerConfiguration to re-enable eager parsing to JsonNode. JsonCounter is still used, so it's not 100% the old behavior, but it should be more compatible in some respects because it doesn't change the conversion logic anymore.
- A new LazyJsonNode class. It contains a ByteBuffer with the actual unparsed bytes. It's a bit complicated because it contains a reference counted buffer that has to be released after conversion, but also sometimes multiple converters are called for the same LazyJsonNode (once to JsonNode, once to a user-defined type). To solve this, when there is a conversion to JsonNode, it keeps that JsonNode around (and releases the buffer). If there are future conversions to a specific type, it will use the JsonNode as the source instead.
- A new JsonSyntaxException class. JsonMappers can throw this to signal a syntax error in the JSON, which will be reported differently than a data binding error.
- Changes to JsonMapper API: The asynchronous parser is deprecated. There is a new readValue method that takes a ByteBuffer. The jackson implementation has an optimization for netty ByteBuffers.
- New converters for LazyJsonNode. Also removed some old superfluous converters for JsonNode.
- A change to RequestLifecycle to show JSON syntax error messages like we did previously, instead of the opaque 400 that we usually send for conversion errors.
- Some test changes to reflect changes in error messages. Because we now parse the input in bulk, in some cases jackson can decorate errors with short snippets of the failing input data. Is this an issue?

There are a few potential incompatibilities with this change:

- Non-standard JSON features when combined with JsonCounter (i.e. when streaming a JSON array, or when using `application/x-json-stream`) will break even when the JSON parser is configured to support them. e.g. if the user configured jackson to ignore comments, that may fail now.
- When the input body is never bound, JSON syntax errors may not be caught.
- JsonCounter only supports UTF-8, by design. This is permitted by the JSON standard, however Jackson also supports UTF-16 and UTF-32. imo this is a net benefit, to avoid potential parser differential vulnerabilities. I also made a jackson feature to match this, though JsonCounter is sufficient now: https://github.com/FasterXML/jackson-core/pull/921
- JsonMapper implementations that do not use JsonSyntaxException yet will have less verbose HTTP errors until they switch to JsonSyntaxException.


yawkat:
oh there is a simple solution: don't scan when the content type is application/json, because for that type, only a single root value is valid anyway.

graemerocher:
It isn't clear to me from your outline if there are benefits to this change.

yawkat:
@graemerocher pretty much just performance. but until it's faster in benchmarks, it stays draft.

yawkat:
and tbh if this turns out to be worth it with the application/json exception, then most of the optimizations in JsonCounter are probably not worth it anymore.

yawkat:
Okay with this change, the results look a lot better:

image

Top is before this change, bottom is after this change.

The benchmark basically parses an array of strings. The green runs have 6 strings of varying length (length=6 for light green, length=1000 for normal green, length=100000 for dark green). The other two runs have the same string length 6, but different array lengths (1000 elements for blue, 100000 elements for pink).

As expected, this patch shows an improvement when there are many JSON tokens (the blue and pink cases). This is also imo the more realistic scenario. At this point, this PR is still slightly worse for extremely long strings, but only maybe 10%, and this is an extreme case (100000 bytes per string). I think this is acceptable to get a >30% speedup for the many-tokens case.

WDYT @graemerocher ?

graemerocher:
Looks good 👍

http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonContentProcessor.java
```diff
@@ -62,11 +61,18 @@ public JsonContentProcessor(
             JsonMapper jsonMapper) {
         super(nettyHttpRequest, configuration);
         this.jsonMapper = jsonMapper;
+
+        if (nettyHttpRequest.getContentType()
+            .map(mediaType -> mediaType.equals(MediaType.APPLICATION_JSON_TYPE))
+            .orElse(false)) {
```
graemerocher: for performance avoiding the lambda here might be useful

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -506,6 +513,24 @@ public void setHttpToHttpsRedirect(boolean httpToHttpsRedirect) {
         this.httpToHttpsRedirect = httpToHttpsRedirect;
     }
 
+    /**
+     * Parse incoming JSON data eagerly, before route binding.
+     *
+     * @return Whether to parse incoming JSON data eagerly before route binding
+     */
+    public boolean isEagerParsing() {
```
graemerocher: Should this be in `JsonConfiguration`? Also clarify what is the default

Add `@since` to both methods

http-server/src/main/java/io/micronaut/http/server/exceptions/JacksonExceptionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+ * Default exception handler for jackson processing errors.
+ *
+ * @author Graeme Rocher
+ * @since 1.0
+ */
+@Produces
+@Singleton
+@Requires(classes = JsonProcessingException.class)
+@Internal
+public class JacksonExceptionHandler implements ExceptionHandler<JsonProcessingException, Object> {
```
graemerocher: make `final` and move the logic into an abstract exceptionhandler that both `JsonExceptionHandler` and this one extend since it seems the `handle` method is duplicated

http-server/src/main/java/io/micronaut/http/server/exceptions/JacksonExceptionHandler.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private final ErrorResponseProcessor<?> responseProcessor;
+
+    /**
+     * Constructor.
+     * @param responseProcessor Error Response Processor
+     */
+    @Inject
```
graemerocher: unnecessary `@Inject` annotation

jackson-core/src/main/java/io/micronaut/jackson/core/parser/JacksonCoreParserFactory.java
```diff
@@ -0,0 +1,63 @@
...
+public class JacksonCoreParserFactory {
+    private static final boolean HAS_NETTY_BUFFER;
+
+    private JacksonCoreParserFactory() {
+    }
+
+    static {
+        boolean hasNettyBuffer;
+        try {
+            Class.forName("io.netty.buffer.ByteBuf", false, null);
```
graemerocher, line 47: is this registered for reflection for GraalVM?

jackson-core/src/main/java/io/micronaut/jackson/core/parser/JacksonCoreParserFactory.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Helper class for implementing
+ * {@link io.micronaut.json.JsonMapper#readValue(ByteBuffer, Argument)} with optimizations for
+ * netty ByteBufs.
+ */
+@Internal
+public class JacksonCoreParserFactory {
```
graemerocher: make `final`

jackson-core/src/main/java/io/micronaut/jackson/core/parser/JacksonCoreParserFactory.java
```diff
@@ -0,0 +1,63 @@
...
+        try {
+            Class.forName("io.netty.buffer.ByteBuf", false, null);
+            hasNettyBuffer = true;
+        } catch (ClassNotFoundException e) {
+            hasNettyBuffer = false;
+        }
+        HAS_NETTY_BUFFER = hasNettyBuffer;
+    }
+
+    public static JsonParser createJsonParser(JsonFactory factory, ByteBuffer<?> buffer) throws IOException {
```
graemerocher, line 63: add javadoc

json-core/src/main/java/io/micronaut/json/JsonSyntaxException.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package io.micronaut.json;
+
+import java.io.IOException;
+
+/**
+ * Exception thrown when there is a syntax error in JSON (e.g. mismatched braces).
+ */
+public final class JsonSyntaxException extends IOException {
```
graemerocher, line 26: add `@since` and `author` tags. Add constructor that takes a message and the exception.

http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonContentProcessor.java
```diff
@@ -78,91 +84,78 @@ public HttpContentProcessor resultType(Argument<?> type) {
...
     }
 
-    @Override
-    public void complete(Collection<Object> out) throws Throwable {
-        if (jacksonProcessor == null) {
-            resultType(null);
+    private void bufferForNextRun(ByteBuf buffer) {
+        if (this.buffer == null) {
+            // number of components should not be too small to avoid unnecessary consolidation
+            this.buffer = buffer.alloc().compositeBuffer(4096);
```
graemerocher: should this buffer size be configurable?

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/server/json/PersonController.kt
```diff
@@ -15,24 +15,20 @@
...
+import io.micronaut.core.async.annotation.SingleResult
 import io.micronaut.http.HttpRequest
 import io.micronaut.http.HttpResponse
 import io.micronaut.http.HttpStatus
-import io.micronaut.http.annotation.Body
-import io.micronaut.http.annotation.Controller
-import io.micronaut.http.annotation.Error
-import io.micronaut.http.annotation.Get
-import io.micronaut.http.annotation.Post
+import io.micronaut.http.annotation.*
```
graemerocher: revert this change that uses wildcard imports

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/json/PersonController.kt
```diff
@@ -15,24 +15,20 @@
...
+import io.micronaut.core.async.annotation.SingleResult
 import io.micronaut.http.HttpRequest
 import io.micronaut.http.HttpResponse
 import io.micronaut.http.HttpStatus
-import io.micronaut.http.annotation.Body
-import io.micronaut.http.annotation.Controller
-import io.micronaut.http.annotation.Error
-import io.micronaut.http.annotation.Get
-import io.micronaut.http.annotation.Post
+import io.micronaut.http.annotation.*
```
graemerocher: revert this change that uses wildcard imports

graemerocher (changes requested):

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
yawkat: i moved it to NettyHttpServerConfiguration

it's more related to the routing pipeline, so it's more appropriate in one of the http config classes.

json-core/src/main/java/io/micronaut/json/JsonSyntaxException.java
yawkat: I intentionally only allowed these two constructors, one for "delegating" to an implementation exception (e.g. JsonParseException), one for a message thrown by micronaut code (i.e. a syntax error detected by JsonCounter). The behavior does not exactly match the super constructors: The JsonSyntaxException(Throwable) constructor copies the message to the JsonSyntaxException, so that the JsonError that is shown to the user contains the proper message.

http-server-netty/src/main/java/io/micronaut/http/server/netty/jackson/JsonContentProcessor.java
yawkat: can't hurt, though i didnt see a disadvantage to large values in testing. i will add a config

graemerocher:
not compiling

PR #8806, opened 2023-02-20
Adding test for headers to verify conformance to HTTP/1.1 4.2 Message Headers specification

### LINKED ISSUE #8687: Add a test to HTTP Server TCK which verifies HTTP Headers are case insensitive
### Issue description

[RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1", Section 4.2, "Message Headers"](http://www.w3.org/Protocols/rfc2616/rfc2616.html):

Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

wetted:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

closes #8687

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/HeadersTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/HeadersTest.java
@@ -30,6 +30,11 @@ import java.io.IOException;
 
 import static io.micronaut.http.server.tck.TestScenario.asserts;
 
+@SuppressWarnings({
+    "java:S5960", // We're allowed assertions, as these are used in tests only
+    "checkstyle:MissingJavadocType",
+    "checkstyle:DesignForExtension"
+})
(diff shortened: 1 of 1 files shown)


wetted:
@yawkat what do you think? Should we trim leading/trailing whitespace and collapse LWS to single SP for header values, as the spec suggests can happen? I think that test fails in Netty though.

Any other tests you think we should have for it?

yawkat:
Yes it should be accepted, and it already should work for netty. The test probably fails because the client doesn't allow it and that's fine. You need to test it differently.

sdelamo (changes requested):
could you add to the test a custom header ?

`HttpRequest.GET("/bar/ok").header("fOO",  "ok")...`


` String getOkAsJson() {`


` String getOkAsJson(@Header("Foo") String bar) {`

sdelamo (approved):

PR #8821, opened 2023-02-23
Cleanup GraalVM metadata and used Metadata Repository

graemerocher:
With the GraalVM Metadata Repository now active we no longer need metadata for Netty and Logback so this PR removes that metadata. In addition this PR removes the use of all GraalVM internal APIs from core and upgrades to the latest version of SVM.

Finally, this PR deprecates `AutomaticFeatureUtils` because in future it should not be necessary to use this class with metadata living the repository.


melix (approved):
Fingers crossed that it passes tests!

graemerocher:
Tests passed 🕺

graemerocher:
Hold off merging, the logback metadata is incomplete in the repository

graemerocher:
ok fixed logging. For Launch we have to remove the line:

```xml
<withJansi>true</withJansi>
```

Since JANSI is not the classpath anyway

.github/workflows/graalvm.yml
```diff
@@ -19,14 +19,8 @@ jobs:
...
     strategy:
       matrix:
-        graalvm: [ 'latest', 'dev' ]
+        graalvm: [ 'latest']
         java: [ '17' ]
-        include:
-          - graalvm: 'dev'
-            java: '19'
-        exclude:
-          - graalvm: 'dev'
```
fniephaus, line 28: JDK17-based dev builds are back online 🙂

sdelamo (approved):
see: https://github.com/micronaut-projects/micronaut-core/pull/8824

PR #8824, opened 2023-02-24
reduce complexity

sdelamo:


graemerocher:
not sure this is any easier to read IMO

PR #8825, opened 2023-02-24
fix fallback for suspended methods

### LINKED ISSUE #7101: AOP - Using Retryable and Recoverable with a suspend function fails
### Expected Behavior

A `suspend` function annotated with both `@Retryable` and `@Recoverable` should retry as many times as configured in the `Retryable` annotation before triggering the `Recoverable` Fallback.

### Actual Behaviour

During execution, a ClassCastException is raised during the RecoveryInterceptor that causes the Fallback to be executed during execution of the declared function.

This causes the calling function to get the 'Fallback' output before the retries are completed.
 
This also leads to the retries happening in the background, despite the fact that the Fallback has already fired.

The below stacktrace is logged:
```kotlin
java.lang.ClassCastException: class kotlin.Unit cannot be cast to class java.util.concurrent.CompletableFuture (kotlin.Unit is in unnamed module of loader 'app'; java.util.concurrent.CompletableFuture is in module java.base of loader 'bootstrap')
	at io.micronaut.retry.intercept.RecoveryInterceptor.lambda$fallbackForFuture$3(RecoveryInterceptor.java:167)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:859)
	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:837)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at io.micronaut.aop.util.CompletableFutureContinuation.resumeWith(CompletableFutureContinuation.kt:51)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

### Steps To Reproduce

- Declare a suspend function and annotate it with both Recoverable and Retryable.
- Implement a sufficient Fallback implementation for this function
- Throw an exception inside the suspend function that should trigger a retry
- The callers invocation of this suspend function is then 'completed' with the Fallback's return value while retries continue in the background.
  - My example repo has the suspend function return `Unit` however the same issue happens when you return a value, I tested with Boolean

### Environment Information

OS - macOS 12.2.1 (x86_64)
JVM - 11.0.11

### Example Application

https://github.com/tmapes/micronaut-suspend-retry-recover

### Version

3.3.4

ondrej-hrstka:
Fixes #7101

### DIFF
--- a/runtime/src/main/java/io/micronaut/retry/intercept/RecoveryInterceptor.java
+++ b/runtime/src/main/java/io/micronaut/retry/intercept/RecoveryInterceptor.java
@@ -78,10 +78,6 @@ public class RecoveryInterceptor implements MethodInterceptor<Object, Object> {
         }
         InterceptedMethod interceptedMethod = InterceptedMethod.of(context);
         try {
-            if (context.isSuspend()) {
-
-            }
-
             switch (interceptedMethod.resultType()) {
                 case PUBLISHER:
                     return interceptedMethod.handleResult(
--- a/runtime/src/main/java/io/micronaut/retry/intercept/RecoveryInterceptor.java
+++ b/runtime/src/main/java/io/micronaut/retry/intercept/RecoveryInterceptor.java
@@ -78,15 +78,25 @@ public class RecoveryInterceptor implements MethodInterceptor<Object, Object> {
         }
         InterceptedMethod interceptedMethod = InterceptedMethod.of(context);
         try {
+            if (context.isSuspend()) {
+
+            }
+
             switch (interceptedMethod.resultType()) {
                 case PUBLISHER:
                     return interceptedMethod.handleResult(
                             fallbackForReactiveType(context, interceptedMethod.interceptResultAsPublisher())
                     );
                 case COMPLETION_STAGE:
-                    return interceptedMethod.handleResult(
+                    if (context.isSuspend()) {
+                        return interceptedMethod.handleResult(
+                            fallbackForSuspend(context, interceptedMethod.interceptResultAsCompletionStage())
+                        );
+                    } else {
+                        return interceptedMethod.handleResult(
                             fallbackForFuture(context, interceptedMethod.interceptResultAsCompletionStage())
-                    );
+                        );
+                    }
                 case SYNCHRONOUS:
                     try {
                         return context.proceed();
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/io/micronaut/retry/intercept/RecoveryInterceptor.java
```diff
@@ -78,15 +78,25 @@ public Object intercept(MethodInvocationContext<Object, Object> context) {
         }
         InterceptedMethod interceptedMethod = InterceptedMethod.of(context);
         try {
+            if (context.isSuspend()) {
+
+            }
+
```
graemerocher: remove this please

graemerocher (changes requested):

runtime/src/main/java/io/micronaut/retry/intercept/RecoveryInterceptor.java
ondrej-hrstka: Removed, thanks.

dstepanov (approved):

dstepanov:
Why does this not work by default? `KotlinInterceptedMethod` should fill the future if the value is not the coroutine:
```
  public CompletableFuture<Object> interceptResultAsCompletionStage() {
        CompletableFutureContinuation completableFutureContinuation = new CompletableFutureContinuation((Continuation<Object>) continuation);
        replaceContinuation.accept(completableFutureContinuation);
        Object result = context.proceed();
        replaceContinuation.accept(continuation);
        if (result != KotlinUtils.COROUTINE_SUSPENDED) {
            completableFutureContinuation.resumeWith(result);
        }
        return completableFutureContinuation.getCompletableFuture();
    }
    ```

ondrej-hrstka:
@dstepanov : The problem is the fallback method. The 'MethodExecutionHandle' is located in 'findFallbackMethod' method. So for that one, there is no instance of 'KotlinInterceptedMethod'

PR #8836, opened 2023-02-25
fix: allow to disable  disable `ServiceReadyHealthIndicator` via configuration

sdelamo:
This PR adds the ability to disable `ServiceReadyHealthIndicator` via configuration.

For example, in Lambda right now health endpoint will be always down.


management/src/main/java/io/micronaut/management/health/indicator/service/ServiceReadyHealthIndicatorConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static io.micronaut.management.health.indicator.service.ServiceReadyHealthIndicatorConfiguration.PREFIX;
+
+/**
+ * @author Sergio del Amo
+ * @since 3.8.6
+ */
+@ConfigurationProperties(PREFIX)
+@Requires(property = PREFIX + ".enabled", notEquals = StringUtils.FALSE)
+public class ServiceReadyHealthIndicatorConfiguration implements Toggleable {
```
graemerocher: a whole new configuration class for one property seems overkill to me. This can be added to the existing `HealthEndpoint` class as a property which is already configurable https://docs.micronaut.io/latest/guide/configurationreference.html#io.micronaut.management.endpoint.health.HealthEndpoint

graemerocher (changes requested):

management/src/main/java/io/micronaut/management/health/indicator/service/ServiceReadyHealthIndicatorConfiguration.java
sdelamo: Please note that we follow the same approach (a configuration class) for discovery. 

https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/management/src/main/java/io/micronaut/management/health/indicator/discovery/DiscoveryClientHealthIndicatorConfiguration.java

graemerocher: which also seems overkill

sdelamo: I could delete `ServiceReadyHealthIndicatorConfiguration` and annotate the indicator with an `@Requires`:

```java
@Requires(property = "endpoints.health.service.enabled", notEquals = StringUtils.FALSE, defaultValue = StringUtils.TRUE)
public class ServiceReadyHealthIndicator implements HealthIndicator {
```

However, this will mean no configuration reference documentation is generated. Users will have no idea how to disable the indicator unless they go to the code.

graemerocher: agreed, which is why it makes more sense these config options are in `HealthEndpoint` where they will show up in the docs

sdelamo: As requested, I have moved the class as a static class to `HealthEndpoint`.   

https://github.com/micronaut-projects/micronaut-core/pull/8836/commits/2f7fc2250b1060138eb1fdfe5143cf3575654959

The amount of code is almost identical. It is just an inner class instead of a separate class. 

Moreover, having a separate configuration class in the same package of the indicator is a nicer design. `HealthPoint` is an aggregator of indicators without knowing anything about indicators or how the indicators are configured, or if they can be toggled.

graemerocher (approved):
Since this sounds like not something we want on Lambda and serverless it may be you want to add:

```
@Requires(notEnv = Environment.FUNCTION)
```

To the indicator as well

sdelamo:
> Since this sounds like not something we want on Lambda and serverless it may be you want to add:
> 
> ```
> @Requires(notEnv = Environment.FUNCTION)
> ```
> 
> To the indicator as well

I think probably we should trigger the startup events as we do for netty in the other server implementations (e.g. lambda).

sdelamo:
@graemerocher You [changed it from `endpoints.health.service.enabled` to `endpoints.health. service-ready-indicator-enabled`]( https://github.com/micronaut-projects/micronaut-core/pull/8836/commits/d269828e33d8e253af424d74e3ed16b15d6c9dc4). 

I explicity used `endpoints.health.service.enabled` to be consistent with other properties names such as: 

- `endpoints.health.disk-space.enabled`
- `endpoints.health.discovery-client.enabled`

Do we want to break consistency to avoid having an extra class?

graemerocher:
yes in 4.0.x we can remove the additional class for the other property. Extra classes for 1 configuration property just add unnecessary bloat.

PR #8841, opened 2023-02-27
Remove more GraalVM build-time init flags

graemerocher:
from Micronaut 4 onwards we should avoid build-time initialization in core and all modules. This change removes some remaining `initialize-at-build-time` flags which don't seem to have any impact on startup or memory performance


graemerocher:
This PR https://github.com/micronaut-projects/micronaut-kotlin/pull/468 requires merging this one first.

yawkat (approved):

PR #8861, opened 2023-03-01
test: HTTP Server TCK LOG body and contains string

### LINKED ISSUE #8860: Improve HTTP Server TCK body assertion feedback
### Issue description

Currently it is hard to diagnose what failed.

sdelamo:
After this change a body contains failure fails with:

```
16:36:27.534 [main] WARN  i.m.http.server.tck.BodyAssertion - The following body does not contains B.
A

[com.gradle.enterprise.testdistribution] Test was part of session 1 and ran on a local executor named 'localhost-executor-1' on host 'localhost'.

expected: <true> but was: <false>
Expected :true
Actual   :false
<Click to see difference>

org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
```

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/BodyAssertion.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/BodyAssertion.java
@@ -57,7 +57,7 @@ public final class BodyAssertion<T> {
      */
     @SuppressWarnings("java:S5960") // Assertion is the whole point of this method
     public void evaluate(T body) {
-        assertTrue(this.evaluator.test(expected, body));
+        assertTrue(this.evaluator.test(expected, body), "Expected [" + expected + "] but was [" + body + "]");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


graemerocher:
I am kind of confused why 

```
    public void evaluate(T body) {
        assertTrue(this.evaluator.test(expected, body));
    }
```

Can't be changed to use `assertEquals` 🤔

sdelamo:
> I am kind of confused why
> 
> ```
>     public void evaluate(T body) {
>         assertTrue(this.evaluator.test(expected, body));
>     }
> ```
> 
> Can't be changed to use `assertEquals` 🤔

we have many test which test if the body contains a substring not the exact body. But then you can pass your own `BodyAssertion`

graemerocher:
ok but then you can still pass a better error message to `assertTrue`?:

```
assertTrue(this.evaluator.test(expected, body), "Expected [" + expected + "] but was [" + body + "]");
```

sdelamo:
> ok but then you can still pass a better error message to `assertTrue`?:
> 
> ```
> assertTrue(this.evaluator.test(expected, body), "Expected [" + expected + "] but was [" + body + "]");
> ```

Great idea.

PR #8879, opened 2023-03-03
Correct processing of the inner classes

dstepanov:
When the inner class element is annotated, Javac wouldn't populate its class type, so we need to extract it.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/TypeElementVisitorProcessor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/TypeElementVisitorProcessor.java
@@ -61,6 +61,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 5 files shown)


inject-java/src/main/java/io/micronaut/annotation/processing/TypeElementVisitorProcessor.java
```diff
@@ -228,18 +229,21 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
...
+                modelUtils.resolveTypeElements(
+                    roundEnv.getElementsAnnotatedWith(annotation)
+                ).filter(notGroovyObject).forEach(elements::add);
             }
 
             // This call to getRootElements() should be removed in Micronaut 4. It should not be possible
             // to process elements without at least one annotation present and this call breaks that assumption.
-            final Set<? extends Element> rootElements = roundEnv.getRootElements();
-            includeElements(elements, rootElements, groovyObjectType);
+            modelUtils.resolveTypeElements(
```
dstepanov, line 244: > // This call to getRootElements() should be removed in Micronaut 4. It should not be possible
> // to process elements without at least one annotation present and this call breaks that assumption.
> 

I left it as it is because some tests will fail if this is removed.

graemerocher (approved):

PR #8884, opened 2023-03-06
Tweak gradle-build-action a bit

Goooler:
- `actions/cache` could be replaced by `gradle/gradle-build-action`.
- Use more `gradle/gradle-build-action` to cache things.
- Update [gradle/gradle-build-action@v2.4.0](https://github.com/gradle/gradle-build-action/releases/tag/v2.4.0).
- Enable [gradle-home-cache-cleanup](https://github.com/gradle/gradle-build-action/blob/main/README.md#removing-unused-files-from-gradle-user-home-before-saving-to-cache).

### DIFF
--- a/.github/workflows/central-sync.yml
+++ b/.github/workflows/central-sync.yml
@@ -26,6 +26,8 @@ jobs:
           java-version: '17'
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2.4.0
+        with:
+          gradle-home-cache-cleanup: true
       - name: Publish to Sonatype OSSRH
         env:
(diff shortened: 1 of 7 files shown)


sdelamo:
@Goooler we use a central repository which we sync to others. If we are going to do this change, it should be done in https://github.com/micronaut-projects/micronaut-project-template 

What do you think we should do these changes? What is the motivation of this PR?

Goooler:
1. We can patch these changes to that template.
2. `actions/cache` is now used for caching Gradle things, which is redundant with `gradle/gradle-build-action` doing, the reduant ones could be removed, and apply more `gradle/gradle-build-action` for CI jobs to accelerate builds.

PR #8898, opened 2023-03-07
Add back the platform for the server-tck tests

timyates:
Not sure where this was lost, but no tests were being executed

### DIFF
--- a/test-suite-http-server-tck-netty/build.gradle
+++ b/test-suite-http-server-tck-netty/build.gradle
@@ -54,7 +54,7 @@ tasks.named("check") { task ->
     }
 }
 
-test {
+tasks.named("test") {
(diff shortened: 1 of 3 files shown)


test-suite-http-server-tck-netty/build.gradle
```diff
@@ -54,6 +54,10 @@ tasks.named("check") { task ->
     }
 }
 
+test {
```
melix: Should use `tasks.named("test")` to avoid configuring the task eagerly.

timyates: 👍

melix (approved):

sdelamo (approved):

PR #8899, opened 2023-03-07
Allow to introspect classes by names

dstepanov:
+ Fix to use the correct bean configuration


inject-java/src/test/groovy/io/micronaut/annotation/AnnotateTypeArgSpec.groovy
```diff
@@ -0,0 +1,135 @@
...
+
+}
+
+''')
+        then:
+            introspection.hasAnnotation(MyAnnotation)
+    }
+
+    @PendingFeature(reason = "Annotation processor doesn't trigger for inner classes with type argument only annotation")
+    void 'test annotating 2'() {
```
dstepanov, line 50: It looks like Javac wouldn't trigger the annotation processing when there is only a type annotation in the inner class.

graemerocher (approved):

PR #8903, opened 2023-03-08
imp: Add STANDARD_HEADERS list

sdelamo:
STANDARD_HEADERS contains every HttpHeader constant

see: https://github.com/micronaut-projects/micronaut-azure/pull/437


http/src/main/java/io/micronaut/http/HttpHeaders.java
```diff
@@ -499,6 +499,103 @@ public interface HttpHeaders extends Headers {
      */
     String X_AUTH_TOKEN = "X-Auth-Token";
 
+    List<String> STANDARD_HEADERS = Arrays.asList(
```
graemerocher: should be immutable

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/HttpHeaders.java
graemerocher: and should have javadoc

sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/8903/commits/998f634d13f9d15a91e9cab4d5e233cfc1525b27 and https://github.com/micronaut-projects/micronaut-core/pull/8903/commits/ec236c48dc22339a9b60a62930b01ffa534de15d

http/src/main/java/io/micronaut/http/HttpHeaders.java
```diff
@@ -499,6 +499,106 @@ public interface HttpHeaders extends Headers {
      */
     String X_AUTH_TOKEN = "X-Auth-Token";
 
+    /**
+     * Unmodifiable List of every header define in {@link HttpHeaders}.
```
graemerocher:
```suggestion
     * Unmodifiable List of every header constant defined in {@link HttpHeaders}.
```

graemerocher (approved):

PR #8906, opened 2023-03-08
remove parallelism from default methods in AnnotationScanner

graemerocher:
It is possible to deadlock the annotation scanner if multiple packages are scanned since `parallelStream()` is used. This removes the use of `parallelStream()` since the downstream implementation is already parallel

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/scan/AnnotationScanner.java
+++ b/core/src/main/java/io/micronaut/core/io/scan/AnnotationScanner.java
@@ -119,7 +119,7 @@ public interface AnnotationScanner {
     default @NonNull Stream<Class<?>> scan(@NonNull Class<? extends Annotation> annotation, @NonNull Collection<String> packages) {
         Objects.requireNonNull(annotation, "Annotation type cannot be null");
         Objects.requireNonNull(packages, "Packages to scan cannot be null");
-        return scan(annotation.getName(), packages.parallelStream());
+        return scan(annotation.getName(), packages.stream());
     }
 
     /**
@@ -134,7 +134,6 @@ public interface AnnotationScanner {
         Objects.requireNonNull(packages, "Packages to scan cannot be null");
 
         return packages
-            .parallel()
             .flatMap(pkg -> scan(annotation, pkg));
     }
 


yawkat:
what do we do that deadlocks the fjp in the first place?

graemerocher:
Hibernate uses `@Parallel` which initializes the `SessionFactory` bean in the `ForkJoinPool` in parallel. During creation of this bean it calls `BeanIntrospector.findIntrospections(..)` which uses `ServiceScanner` to find introspections which runs the tasks also on the `ForkJoinPool`. So if the `ForkJoinPool` has `parallelism` set to 1 (for 1 CPU) then this blocks forever.

yawkat:
i dont understand why that is a problem. i thought the idea of the fjp was that in such a scenario, while the fjp thread is blocking on some tasks to complete, it can work on those same submitted tasks.

graemerocher:
The issue is the `ForkJoinPool` task being executed (to create the Hibernate session factory) is trying to schedule something else (the location of introspections) on the `ForkJoinPool` and that is still busy / not yet complete. 

FWIW parallelism is calculated as:

```
int parallelism = Math.max(1, Runtime.getRuntime().availableProcessors() - 1);
```

yawkat:
yes but my point is that this code:

```
            ForkJoinPool.commonPool().invoke(this);
            for (RecursiveActionValuesCollector<S> task : tasks) {
                task.join();
                task.collect(values);
            }
```

if this is called on a fjp thread, even if that fjp only has the single thread, it should still work fine. because while the fjp appears to be blocking in invoke/join, it should be executing the tasks that were just submitted.

so unless we are waiting on some non-fjp futures, thread starvation shouldnt actually be a problem on fjp (but it obviously is in this case).

graemerocher:
if you have a better solution let me know

yawkat:
is there an issue with a test case somewhere?

dstepanov:
@yawkat What if you try to join inside of the task?

Can we have our own pool for scanning?

@graemerocher I think we already have the option to disable the parallel execution, I would add the check there

yawkat:
@dstepanov yes if you join inside the fjp task, the join will execute other tasks from the same fjp

graemerocher:
hard to reproduce, need a parallel bean with 1 CPU allocations that reads introspections

graemerocher:
@dstepanov couldn't find such an option

yawkat:
@graemerocher it should be SoftServiceLoader.disableFork()

if it's hard to reproduce i wont bother trying to figure it out :D

sdelamo (commented):
Should we have a utils class to encapsulate all this in a single place? 

```java
public static boolean allowFork() {
    return ForkJoinPool.getCommonPoolParallelism() > 2;
}
public static Executor createExecutor() { 
    return allowFork ? ForkJoinPool.commonPool() : command -> new Thread(command).start();) }
```

graemerocher:
the fix has changed since my initial assessment was incorrect

dstepanov (approved):

PR #8909, opened 2023-03-08
ci: remove build scan url step

sdelamo:
To avoid extra github api calls


graemerocher:
shouldn't this be done in the template?

sdelamo:
> shouldn't this be done in the template?

[I did it in the template as well](https://github.com/micronaut-projects/micronaut-project-template/pull/335)

We will not get automatic PRs in 3.9.x branch.

Issue #8085, opened 2022-09-26
Time zone only used for the first fire time in the Task Scheduler


pqab:
### Expected Behavior

nextFireTime from the delaySupplier

```
2022-09-26T00:41:08.249268Z[UTC]
2022-09-26T15:00Z[UTC]
2022-09-27T15:00Z[UTC]
2022-09-28T15:00Z[UTC]
...
```

### Actual Behaviour

nextFireTime from the delaySupplier
`Asia/Tokyo` is the system time zone

```
2022-09-26T00:41:08.249268Z[UTC]
2022-09-26T15:00+09:00[Asia/Tokyo]
2022-09-27T15:00+09:00[Asia/Tokyo]
2022-09-28T15:00+09:00[Asia/Tokyo]
...
```

### Steps To Reproduce

we have a schedule task and expected to be run at `00:00+09:00` every day

```
TaskScheduler taskScheduler;
taskScheduler.scheduler("0 0 15 * * ?", "UTC", () -> { ..... })
```

the time zone seems only used for the first fire time

### Environment Information

JDK11

### Version

3.5.4

jordihereu:
We are facing the same issue as #8900 using Micronaut 3.7.2. Mostly detected after the time change occurred last weekend in UK (UK moved from **GMT** to **BST** last Saturday). Our scheduled task are now running an hour late as Daylight Savings (DLS) are not taken into account in the default timezone.

Changing `ZoneId` to "Europe/London" did not have any effect, and schedule is still using system default (GMT).

PR #8911, opened 2023-03-08
Fix bug preventing scheduled jobs from being ran in the specified time zone

Gradsta:
This change attempts to address #8085. I have confirmed this change works by fiddling with the existing tests locally, but I am not sure how to write an automated test for this. In order to test this properly, I believe the cron expression will need to contain a particular hour which means we'll need to instruct `ZonedDateTime.now()` to return a predetermined value. This is where I got stuck. I attempted to use `Mockito.mockStatic(ZonedDateTime.class)` with the appropriate stubs, but I was not successful. Any guidance is appreciated.


graemerocher:
Thanks for the contribution!

PR #8914, opened 2023-03-09
build: remove dependency to validation dependency from core

sdelamo:
This PR removes validation as a dependency of http-server-tck and http-tck

It updates to 6.3.4

the task gralde task `preRelease` is now successful

We need https://github.com/micronaut-projects/micronaut-validation/issues/63 to enable validation of external class such as a POJO in the test tck without depending on validation.

I have disabled one test because of that.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Build with Gradle
         id: gradle
         run: |
-          ./gradlew check --no-daemon --parallel --continue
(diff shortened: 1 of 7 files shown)


melix (approved):
Can it be a `compileOnly` dependency instead of `implementation`? That way you could keep `annotationProcessor` (won't leak into publication).

sdelamo:
> Can it be a `compileOnly` dependency instead of `implementation`? That way you could keep `annotationProcessor` (won't leak into publication).

I changed to `compileOnly` and seems to solve the issue.

graemerocher (approved):

PR #8915, opened 2023-03-09
[core] Update common files for branch 4.0.x

micronaut-build:
Update common files


sdelamo:
We did not decided to test only with GraalVM JDK. However, this PR does that.

PR #8919, opened 2023-03-10
Ensure that Logging System is always refreshed

### LINKED ISSUE #8918: Ensure Logging System is always refreshed
### Expected Behavior

Logging System to be refreshed everytime application starts and Context is Refreshed

### Actual Behaviour

Only happens if `logger.levels` property is defined

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.8.7

driverpt:
Attempt to fix #8918


timyates:
Thanks for this @driverpt 

I think https://github.com/micronaut-projects/micronaut-core/pull/9859 is a slightly better fix though, as it keeps the bean off the classpath unless `logging.` is defined

PR #8932, opened 2023-03-12
Include stereotypes into the mapped/remapped annotation value, keep `@Inherited`

dstepanov:
This PR changes the way how annotations are processed. Now, we include stereotypes into the `AnnotationValue` and recursively build the tree of the annotations and their stereotypes before mappers/transformers are run, making them more powerful. That allowed me to extract some specific code related to qualifiers and interceptors into its own remappers, which also gained the ability to be executed on all annotations.

Now we also include `@Inherited` stereotype into stereotypes of `AnnotationValue` and check if the annotation is inherited after the transformation, allowing the remappers to make the annotation inherited, something that is needed for JAX-RS and Validation. The change in the processing also returns the behavior of Micronaut 3, allowing remapping non-inherited annotation to an inherited one https://github.com/micronaut-projects/micronaut-serialization/blob/master/serde-jackson/src/test/groovy/io/micronaut/serde/jackson/annotation/JsonIgnoreSpec.groovy#L438

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationUtil.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationUtil.java
@@ -15,12 +15,20 @@
  */
 package io.micronaut.core.annotation;
 
-import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
 
-import java.lang.annotation.*;
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
 import java.lang.reflect.AnnotatedElement;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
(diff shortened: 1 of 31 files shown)


inject-groovy/src/test/groovy/io/micronaut/inject/configproperties/InheritedConfigurationReaderPrefixSpec.groovy
```diff
@@ -37,7 +37,7 @@ class MyBean  {
         beanDefinition.getInjectedMethods()[0].name == 'setMyValue'
         def metadata = beanDefinition.getInjectedMethods()[0].getAnnotationMetadata()
         metadata.hasAnnotation(Property)
-        metadata.getValue(Property, "name", String).get() == 'simple.my-value'
+        metadata.getValue(Property, "name", String).get() == 'endpoints.my-value'
```
dstepanov, line 40: These tests are "testing" broken functionality of configuration path being remapped multiple times by aliases, these tests keep changing the output based on the annotation processing order.

core/src/main/java/io/micronaut/core/annotation/AnnotationUtil.java
```diff
@@ -186,6 +184,16 @@ public Annotation[] getDeclaredAnnotations() {
      */
     public static final String POST_CONSTRUCT = "javax.annotation.PostConstruct";
 
+    /**
+     * The annotation attribute containing all the attributes marked as non binding.
+     */
+    public static final String NON_BINDING_ATTRIBUTE = "$nonBinding";
```
dstepanov, line 190: This attribute is not included in all annotations that are using non binding annotation

graemerocher (approved):

PR #8946, opened 2023-03-15
Add ability to disable log in environments

altro3:
Need it to disable SLF4J initialization when micronaut environments used in micronaut-openapi.

ALso fixed some doclint errors and minor corections for slf4j.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
+++ b/inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
@@ -45,6 +45,11 @@ public abstract class AbstractPropertySourceLoader implements PropertySourceLoad
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractPropertySourceLoader.class);
 
+    /**
+     * If you don't need to initialize SLF4J, set 'false'.
+     */
+    protected boolean logEnabled = true;
+
     @Override
     public int getOrder() {
         return DEFAULT_POSITION;
@@ -95,8 +100,8 @@ public abstract class AbstractPropertySourceLoader implements PropertySourceLoad
     private Map<String, Object> loadProperties(ResourceLoader resourceLoader, String qualifiedName, String fileName) {
         Optional<InputStream> config = readInput(resourceLoader, fileName);
         if (config.isPresent()) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("Found PropertySource for file name: " + fileName);
+            if (logEnabled) {
+                LOG.debug("Found PropertySource for file name: {}", fileName);
             }
             try (InputStream input = config.get()) {
                 return read(qualifiedName, input);
@@ -104,8 +109,8 @@ public abstract class AbstractPropertySourceLoader implements PropertySourceLoad
                 throw new ConfigurationException("I/O exception occurred reading [" + fileName + "]: " + e.getMessage(), e);
             }
         } else {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("No PropertySource found for file name: " + fileName);
+            if (logEnabled) {
+                LOG.debug("No PropertySource found for file name: {}", fileName);
             }
         }
         return Collections.emptyMap();
@@ -151,5 +156,27 @@ public abstract class AbstractPropertySourceLoader implements PropertySourceLoad
                 finalMap.put(prefix + key, value);
             }
         }
+    }
+
+    /**
+     * Return logEnabled value.
+     *
+     * @return is log enabled
+     *
+     * @since 3.9.0
+     */
+    public boolean isLogEnabled() {
+        return logEnabled;
+    }
+
+    /**
+     * Setter for logEnabled.
+     *
+     * @param logEnabled is log enabled
+     *
+     * @since 3.9.0
+     */
+    public void setLogEnabled(boolean logEnabled) {
+        this.logEnabled = logEnabled;
     }
 }
--- a/core/src/main/java/io/micronaut/core/io/scan/ClassPathResourceLoader.java
+++ b/core/src/main/java/io/micronaut/core/io/scan/ClassPathResourceLoader.java
@@ -20,7 +20,7 @@ import io.micronaut.core.io.ResourceLoader;
 import io.micronaut.core.annotation.Nullable;
 
 /**
- * Abstraction to load resources from the the classpath.
+ * Abstraction to load resources from the classpath.
  *
  * @author James Kleeh
  * @author Graeme Rocher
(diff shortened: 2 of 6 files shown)


altro3:
@graemerocher @sdelamo @dstepanov With this feature, I can stop using slf4j-nop in micronaut-openapi

.gitattributes
```diff
@@ -1,18 +1,19 @@
 # Auto detect text files and perform LF normalization
 *        text=auto
 
-*.java   text
-*.html   text
-*.kt     text
-*.kts    text
-*.md     text diff=markdown
+*.java   text eol=lf
```
sdelamo: this changes seem unrelated to this pr.

sdelamo:
> Need it to disable SLF4J initialization when micronaut environments used in micronaut-openapi.

can you elaborate what is the issue? I don't understand.

altro3:
@sdelamo 

I use `DefaultEnvironment` class to load `application.yml` file from annotation processor micronaut-openapi, but this class use slf4j and when you run annotation processing you see this:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

I've added `slf4j-nop` to solve this problem (https://github.com/micronaut-projects/micronaut-openapi/issues/821)

But, many people don't like this solution: https://github.com/micronaut-projects/micronaut-openapi/issues/908
And also this solution very bad for groovy projects: https://github.com/micronaut-projects/micronaut-openapi/issues/942

That's why I need to disable slf4j for `micronaut-openapi` and after that I can remove slf4j-nop from dependecies and don't scarry about slf4j initialization in openapi annotation level processing.

altro3:
@sdelamo @graemerocher What do you think about this simple solution for slf4j problem?

graemerocher (approved):

.gitattributes
sdelamo: why do you need this?

altro3: @sdelamo I can't commit my changes without this, otherwise build will be failed.

Now there is such a problem: if I leave this file as it is, then locally I cannot build the project. If I run SpotlesApply the project is built locally but not built on your СI-СD. My settings fix this problem, but it's time for you to fix the template itself so that I and other Windows users don't have these problems in the future.

sdelamo: can you send a PR to [micronaut-project-template](https://github.com/micronaut-projects/micronaut-project-template/blob/master/.gitattributes) explain why this change is necessary?

sdelamo: > I can't commit my changes without this, otherwise build will be failed.
> 
> Now there is such a problem: if I leave this file as it is, then locally I cannot build the project. If I run SpotlesApply the project is built locally but not built on your СI-СD. My settings fix this problem, but it's time for you to fix the template itself so that I and other Windows users don't have these problems in the future.

Is the issue that current `.gitattributes` is incompatible for Windows users? I don't understand what you say about CI. Currently, CI builds fine with our current `.gitattributes`

altro3: @sdelamo done: https://github.com/micronaut-projects/micronaut-project-template/pull/348

inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
```diff
@@ -152,4 +157,22 @@ protected void processMap(Map<String, Object> finalMap, Map map, String prefix)
             }
         }
     }
+
+    /**
+     * Return logEnabled value.
+     *
+     * @return is log enabled
```
alvarosanchez, line 164:
```suggestion
     * @return is log enabled
     * @since 3.9.0
```

Same in `PropertySourcePropertyResolver`

inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
```diff
@@ -152,4 +157,22 @@ protected void processMap(Map<String, Object> finalMap, Map map, String prefix)
...
+     */
+    public boolean isLogEnabled() {
+        return logEnabled;
+    }
+
+    /**
+     * Setter for logEnabled.
+     *
+     * @param logEnabled is log enabled
```
alvarosanchez, line 175:
```suggestion
     * @param logEnabled is log enabled
     * @since 3.9.0    
```

Same in `PropertySourcePropertyResolver`

inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
```diff
@@ -95,17 +100,17 @@ public int getOrder() {
     private Map<String, Object> loadProperties(ResourceLoader resourceLoader, String qualifiedName, String fileName) {
         Optional<InputStream> config = readInput(resourceLoader, fileName);
         if (config.isPresent()) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("Found PropertySource for file name: " + fileName);
+            if (logEnabled) {
```
alvarosanchez, line 103: Shouldn't this be:

```suggestion
            if (logEnabled && LOG.isDebugEnabled()) {
```

Same in the other places

alvarosanchez (changes requested):

inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
altro3: fixed

inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
altro3: @alvarosanchez And here I would argue: calling `LOG.isDebugEnabled()` and other `isXXXEnabled` methods makes sense if you have some calculations before calling the logging method. If you just make a method call without calculation, like `log.debug("My message {}", myParam);`, it makes no sense for you to check the logging level, because methods debug / info / error etc alredy have it inside. By the way, this is one of the reasons why logging formatting was invented with symbols `{}`. in this place you haven't any calculation for logging and don't need to check log level twice

inject/src/main/java/io/micronaut/context/env/AbstractPropertySourceLoader.java
altro3: fixe

.gitattributes
altro3: @sdelamo reverted this file

alvarosanchez (approved):

sdelamo (approved):

Issue #8323, opened 2022-11-10
Compile time annotation expression language


GavrilovSV:
### Feature description

@graemerocher 
Following the discussion started in #7897, this issue is aimed at working out a vision of annotation expression language similar to SpEL, but with most evaluation stages executed at compilation time. 

By default an expression is annotation member value surrounded with `#{ }`. It would also be useful to introduce special annotation members which values will be treated as expressions without `#{ }`. For example, `@Requires(expr = "10 > 9")`

The following is a set of features it is expected to include, mostly inspired by SpEL with certain adjustments
1.  Declaring literals. These should include 
- strings (delimited by single quotes) 
- numeric types (int, long, float, double, incl. scientific notation and hex) 
- boolean
- `null`

2. Mathematical operators
- `-`, `/`, `*`, `%`,`^` on numeric types
- `+` on both numeric types and strings 

3. Relational operators
- Standart relational operators (`>`, `<`, `>=`, `<=`, `==`, `!=`)
- `instanceof` operator (e.g. `'abc' instanceof T(java.langString)`)
- `matches` for regex mathing (e.g. `'abc' matches '^[a-z]*'`) (including `/.+/` syntax for regex)
- `empty` for checking of an object is null or empty
- Relational operators need to support comparables as well

 4. Logical operators
- `&&`, `||`, `!`. We should probably also allow using aliases here (`and`, `or`, `not`)

 5. Working with type references
 
 Type reference should be provided in format of `T(io.micronaut.CustomType)` (allowing to omit full package for `java.lang` classes) and should be treated in different way depending on context
- as part of `instanceof` operation
- as an argument for method invocation which is treated as io.micronaut.CutsomType.class (e.g. `getBean(T(io.micronaut.CustomType))`
- as a type reference followed by static method invocation (e.g. `T(java.lang.Math).random()`)
 
6. Inlining collections
- lists with `{1, 2, 3}` syntax
- maps with `{ 'first': 1, 'second': 2 }`
  
Nested collections should also be supported

7. Working with collections
- Accessing by index for lists (e.g. `list[1]`)
- Accessing by key for maps (e.g. `map['key']`)
- Collection filtering (e.g. `ages[age > 18]` )
- Collection mapping (e.g. `persons[person.age]` )

8. Accessing object properties
- This should allow chaining property access with `.` (e.g. `object.property.nestedProperty`)
- Safe property access allowing to avoid NPE (`object?.property.?.nestedProperty`)

[...]

PR #8954, opened 2023-03-16
Support for Compilation Time Expressions in Annotations

graemerocher:
From feature request https://github.com/micronaut-projects/micronaut-core/issues/8323 this PR implements the following set of features:

1.  Declaring literals. 

- [x]  strings (delimited by single quotes) 
- [x] numeric types (int, long, float, double, incl. scientific notation and hex) 
- [x] boolean
- [x] `null`

2. Mathematical operators
- [x] `-`, `/`, `*`, `%`,`^` on numeric types
- [x] `+` on both numeric types and strings 

3. Relational operators
- [x] Standart relational operators (`>`, `<`, `>=`, `<=`, `==`, `!=`)
- [x] `instanceof` operator (e.g. `'abc' instanceof T(java.langString)`)
- [x] `matches` for regex mathing (e.g. `'abc' matches '^[a-z]*'`) (including `/.+/` syntax for regex)
- [x] `empty` for checking of an object is null or empty
- [ ]  Relational operators need to support comparables as well

 4. Logical operators
- [x]  `&&`, `||`, `!`. We should probably also allow using aliases here (`and`, `or`, `not`)

 5. Working with type references
 
- [x]  as part of `instanceof` operation
- [x]  as an argument for method invocation which is treated as io.micronaut.CutsomType.class (e.g. `getBean(T(io.micronaut.CustomType))`
- [x] as a type reference followed by static method invocation (e.g. `T(java.lang.Math).random()`)
   
6. Working with collections
- [x] Accessing by index for lists (e.g. `list[1]`)
- [x] Accessing by key for maps (e.g. `map['key']`)
- [ ] Collection filtering (e.g. `ages.filter(age -> age > 18)` )
- [ ] Collection mapping (e.g. `persons.map(p -> p.age)` )

For filtering/mapping probably makes sense to support lambda syntax.

7. Accessing object properties
- [x] This should allow chaining property access with `.` (e.g. `object.property.nestedProperty`)
- [x] Safe property access allowing to avoid NPE (`object?.property.?.nestedProperty`)

8. Methods invocation
- [x] Object method invocation (e.g. `object.name().lenght()`)
- [x] Static method invocation (`T(java.lang.Math).random()`)

9. Ternary operator
- [x] Standard `expr ? 'trueResult' : 'falseResult'`
- [x] Elvis operator (e.g. `age?:18`, note only makes sense if we implement coercion to boolean for numbers, lists, strings etc.)

10. Object construction
- [ ] construction with `new` keyword (this might be useful in `@Value` annotation, e.g. `@Value(#{ new io.micronaut.CustomType() })` 
- [ ] array construcation (`new int[] {1, 2, 3}`) - not sure whether this is a useful one, since inlining list gives similar behavior 
- [ ] lists with `{1, 2, 3}` syntax
- [ ] maps with `{ 'first': 1, 'second': 2 }`

11. Predefined syntax constructs, e.g.
- [ ] Getting beans from context (e.g. `ctx[io.micronaut.CustomBean]`)
- [ ] Accessing properties (e.g. `env['custom.property']` )

See https://github.com/micronaut-projects/micronaut-core/issues/8323


graemerocher:
test failures are unrelated and related to the latest changes to the validation module

sdelamo:
I will review this PR tomorrow.

core-processor/src/main/java/io/micronaut/expressions/context/ExpressionCompilationContextFactory.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Adds evaluated expression context class element to context loader
+     * at compilation time.
+     *
+     * <p>This method should be invoked from the {@link io.micronaut.inject.visitor.TypeElementVisitor#start(VisitorContext)} of a {@link io.micronaut.inject.visitor.TypeElementVisitor}</p>
+     *
+     * @param contextClass context class element
+     * @return This context factory
+     */
+    ExpressionCompilationContextFactory registerContextClass(@NonNull ClassElement contextClass);
```
sdelamo, line 61:
```suggestion
    @NonNull
    ExpressionCompilationContextFactory registerContextClass(@NonNull ClassElement contextClass);
```

src/main/docs/guide/config/evaluatedExpressions.adoc
```diff
@@ -0,0 +1,408 @@
+Since 4.0 Micronaut supports embedding evaluated expressions in annotation values using `#{...}` syntax which
```
sdelamo:
```suggestion
Since 4.0, Micronaut framework supports embedding evaluated expressions in annotation values using `#{...}` syntax which
```

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
```diff
@@ -155,6 +155,29 @@ public boolean isDefault() {
...
     }
 
+    @Override
+    public boolean isVarArgs() {
+        return executableElement.isVarArgs();
+    }
+
+    @Override
+    public boolean overrides(MethodElement overridden) {
+//        if (this.equals(overridden) || isStatic() || overridden.isStatic()) {
```
sdelamo: should we remove this commented out code?

sdelamo (approved):
I have created several improvements PRs: 

- https://github.com/micronaut-projects/micronaut-core/pull/9000
- https://github.com/micronaut-projects/micronaut-core/pull/8999
- https://github.com/micronaut-projects/micronaut-core/pull/8998
- https://github.com/micronaut-projects/micronaut-core/pull/8997
- https://github.com/micronaut-projects/micronaut-core/pull/8996


It misses `package-info.java` in the package and subpackages of `io.micronaut.expressions`. 

What’s new section does not contain anything about Expression. It would be great if you could add a link such as: 

```
 `<<evaluatedExpressions, Evaluated Expressions>>`
 
 ```
 
 
 I think this is an exciting feature. I will create an issue in security to support something like: `@Secured(condition = "#{authentication.attributes.email == 'john@gameothrones.com' }`

graemerocher:
@GavrilovSV I have merged the first cut of this. Thank you for your work implementing the basics of this feature, I think it has great potential.

There are still areas of improvement that I would like to address before GA:

* The performance of bean lookups via the evaluation context. I think we should make context objects effectively singleton and cache them in the evaluation context otherwise I am not sure how well this will perform for hot paths like request processing. Given most of the execution part is marked as `@Experimental` I am fine to address this later.
* A default context that gives access to the `Environment` object to simplify writing conditions that resolve properties
* Research into extending the HTTP annotations to allow conditional routes (if we address the performance concern above)
* Probably we should support `Comparable` interface for logical operators
* Probably we should support collection filtering and mapping. For this I think we should use the lambda syntax alreaded defined in Java and Jakarta EL.

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
graemerocher: I removed this change

graemerocher:
@sdelamo 

> I have created several improvements PRs:
> 
>     * #9000
> 
>     * #8999
> 
>     * #8998
> 
>     * #8997
> 
>     * #8996

closed / merged these.

> 
> 
> It misses `package-info.java` in the package and subpackages of `io.micronaut.expressions`.

This is not public API as it is part of the processor module which is mostly internal API. However we can add this if you feel strongly about it.

Issue #8912, opened 2023-03-09
LocaleArgumentBinder does not respect default locale


larsduelfer:
### Expected Behavior

As stated in the micronaut documentation, a default locale should be injected (via argument binder) into a controller method declaring a property of type java.util.Locale if it is missing in the request. 

### Actual Behaviour

When the request does not contains an Accept-Language header, the default locale is not respected and a bad request is returned in case the locale is a mandatory property of a controller method.

Solution: 

The current implementation of the `LocaleArgumentBinder` uses the `resolve() `method of the LocalResolver interface, not the`resolveOrDefault()` method. This should be replaced to be inline with the documentation.

The following shows the mentioned code using the `resolve()` method instead of the `resolveOrDefault()`

```
@Singleton
public class LocaleArgumentBinder implements TypedRequestArgumentBinder<Locale> {
    ...        
    @Override
    public BindingResult<Locale> bind(ArgumentConversionContext<Locale> context, HttpRequest<?> source) {
        final Optional<Locale> locale = localeResolver.resolve(source);
        if (locale.isPresent()) {
            return () -> locale;
        } else {
            return BindingResult.UNSATISFIED;
        }
    }
}
```

### Version

3.8.7

SantiagoPVazquez:
I would like to take on this issue.

SantiagoPVazquez:
So as I understand, the only change is to replace the resolve() method for resolveOrDefault() inside the "bind" method from LocaleArgumentBinder class? Or should I replace all the resolve() method calls?

larsduelfer:
The first approach should do the job, just replacing the one call from .resolve() to .resolveOrDefault(). @SantiagoPVazquez

SantiagoPVazquez:
Alright, Thank you! I'll post a PR with the changes.

graemerocher:
Triage: this remains a valid HTTP server bug. Current Core still has LocaleArgumentBinder.bind call localeResolver.resolve(source) and return UNSATISFIED when no locale is present, while the documented default behavior requires resolveOrDefault(source). The focused fix is the single call-site change, with a regression test for a missing Accept-Language header and a configured default locale.

PR #8955, opened 2023-03-16
resolve() method changed to resolveOrDefault() in LocalArgumentBinder class

SantiagoPVazquez:
This change fixes issue #8912


jameskleeh:
How would one detect no locale was passed with the request after this change?

PR #8958, opened 2023-03-17
doc: don't use Micronaut as a noun

sdelamo:
According to Trademark policy
https://objectcomputing.com/files/6816/4867/5102/Micronaut_Trademark_Policy.pdf

> § You may not use the word Micronaut as a verb or a noun. It should be used only as an adjective followed by a noun (e.g., “I love using the Micronaut® framework,” not “I love using Micronaut!”).


graemerocher:
I don't think I need to repeat about how stupid I think this is, but ok if you want to merge

dstepanov:
Does the English or the law language has the concept of a context? If you use `Micronaut` on the website of the Micronaut framework, it's most likely about the framework and not something else.

graemerocher (approved):

PR #8959, opened 2023-03-17
doc: move to whats next to the bottom

sdelamo:
- add link to Micronaut Upgrade
- call to action in the top linking to instructions to upgrade for Micronaut 4.


sdelamo:
Maybe, we should move every what's new and breaks documentation to [Micronaut Upgrade](https://micronaut-projects.github.io/micronaut-upgrade/snapshot/guide). That way, we could improve whats new and breaking docs without the need for a release. 

Thoughts?

PR #8967, opened 2023-03-18
Caffeine 3.1.5

altro3:


altro3:
@graemerocher this is duplicate PR for https://github.com/micronaut-projects/micronaut-core/pull/7908, but without conflicts

graemerocher:
I am not sure where Caffeine is used in core anymore I think this can be removed

sdelamo:
> I am not sure where Caffeine is used in core anymore I think this can be removed

we have a test https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/inject-java/src/test/groovy/io/micronaut/inject/factory/named/RepositoryService.java

PR #8968, opened 2023-03-18
Added lost slf4j, logback, log4j versions

altro3:
Lost changes from micronaut 3.8 (see https://github.com/micronaut-projects/micronaut-core/pull/8138)

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -36,7 +36,6 @@ kotlin-coroutines = "1.6.4"
 ktor = "1.6.8"
 managed-logback = "1.4.6"
 logbook-netty = "2.14.0"
-log4j = "2.19.0"
 micronaut-aws = "3.9.2"
 micronaut-groovy = "4.0.0-SNAPSHOT"
 micronaut-session = "1.0.0-SNAPSHOT"
@@ -72,6 +71,7 @@ managed-reactive-streams = "1.0.4"
 # This should be kept aligned with https://github.com/micronaut-projects/micronaut-reactor/blob/master/gradle.properties from the BOM
 managed-reactor = "3.4.24"
 managed-slf4j = "2.0.4"
+managed-log4j = "2.19.0"
 managed-snakeyaml = "2.0"
 managed-java-parser-core = "3.24.9"
 managed-ksp = "1.8.0-1.0.9"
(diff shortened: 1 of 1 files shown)


sdelamo:
I think they should be in https://github.com/micronaut-projects/micronaut-platform not in core

melix:
They should probably neither be in `core`, nor in `platform`. The problem if you put them in `core` is that we're introducing _unused_ dependencies in `core`. Obviously in `platform` then we would introduce a cyclic dependency between platform and core, so this is a no-go. There should probably be a `micronaut-logging` project for this.

gradle/libs.versions.toml
```diff
@@ -207,7 +219,7 @@ kotlinx-coroutines-rx2 = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-rx
 kotlinx-coroutines-slf4j = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-slf4j", version.ref = "kotlin-coroutines" }
 kotlinx-coroutines-reactor = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-reactor", version.ref = "kotlin-coroutines" }
 
-log4j = { module = "org.apache.logging.log4j:log4j-core", version.ref = "log4j" }
+log4j = { module = "org.apache.logging.log4j:log4j-core", version.ref = "managed-log4j" }
```
melix, line 222: :x: This should be managed too, right?

melix (changes requested):

gradle/libs.versions.toml
altro3: @melix Yes, it's necessary. For example, if you want to use slf4j in your project in conjunction with micronaut-elasticsearch - for this you will need the 'org.apache.logging.log4j:log4j-to-slf4j:2.20.0' library and at the moment you will have to duplicate the log4j version in your project because this module is not in the bom file. We already discussed this here: https://github.com/micronaut-projects/micronaut-core/pull/8138 . And now everything works fine with version 3.7.x. But these changes are lost in version 4.x

sdelamo: @altro3 we moved versions outside of core intentionally in 4.0.x. I think `micronaut-logging` is probably the way to go.

altro3: @sdelamo good solution! But where is project micronaut-logging? :-)

sdelamo:
Closed via https://github.com/micronaut-projects/micronaut-core/issues/8986

PR #8971, opened 2023-03-20
tck: static resource test

sdelamo:
see: https://github.com/micronaut-projects/micronaut-aws/issues/1361 
Azure seems to have the same issue as well.


http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/staticresources/StaticResourceTest.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2017-2022 original authors
```
timyates:
```suggestion
 * Copyright 2017-2023 original authors
```

timyates (approved):

PR #8974, opened 2023-03-20
Embedded HTTP benchmark, various small perf improvements

yawkat:
This PR contains a benchmark for the full http stack using EmbeddedChannel. It's almost as close to reality as testing TCP directly, but it's on a single thread so it's a bit easier to work with. It stresses the full HTTP stack including netty handlers, routing, filters, serialization and so on.

There are a few minor optimizations with relatively small diffs. There is also one fairly big change: Introducing a new DelayedExecutionFlow which acts kind of like CompletableFuture, but is faster (it doesn't have to support multiple downstreams for every CF). 

The changes are separated into individual commits, so you can look at each commit one by one if you want. But they don't cross over each other very much anyway.

I'm not done optimizing, but because the DelayedExecutionFlow is a fairly big diff I want to make this PR before it blows up further.

### DIFF
--- a/benchmarks/src/jmh/java/io/micronaut/http/server/stack/SearchController.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/server/stack/SearchController.java
@@ -1,0 +1,39 @@
+package io.micronaut.http.server.stack;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.MutableHttpResponse;
+import io.micronaut.http.annotation.Body;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
+
+import java.util.List;
+
+@Controller("/search")
+@Requires(property = "spec.name", value = "FullHttpStackBenchmark")
+public class SearchController {
+    @Post("find")
+    public HttpResponse<?> find(@Body Input input) {
+        return find(input.haystack, input.needle);
+    }
+
+    private static MutableHttpResponse<?> find(List<String> haystack, String needle) {
+        for (int listIndex = 0; listIndex < haystack.size(); listIndex++) {
+            String s = haystack.get(listIndex);
+            int stringIndex = s.indexOf(needle);
+            if (stringIndex != -1) {
+                return HttpResponse.ok(new Result(listIndex, stringIndex));
+            }
+        }
+        return HttpResponse.notFound();
+    }
+
+    @Introspected
+    record Input(List<String> haystack, String needle) {
+    }
+
+    @Introspected
+    record Result(int listIndex, int stringIndex) {
+    }
+}
--- a/core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
+++ b/core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
@@ -1,0 +1,434 @@
+package io.micronaut.core.execution;
+
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+@SuppressWarnings("rawtypes")
+final class DelayedExecutionFlowImpl<T> implements DelayedExecutionFlow<T> {
+    private static final Logger LOG = LoggerFactory.getLogger(DelayedExecutionFlowImpl.class);
+
+    /**
+     * Object used as a stand-in for a {@code null} completion to distinguish it from the
+     * uncompleted state.
+     */
+    private static final Object NULL = new Object();
+
+    /**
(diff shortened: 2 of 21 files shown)


graemerocher (approved):

benchmarks/src/jmh/java/io/micronaut/http/server/stack/SearchController.java
```diff
@@ -0,0 +1,39 @@
...
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
+
+import java.util.List;
+
+@Controller("/search")
+@Requires(property = "spec.name", value = "FullHttpStackBenchmark")
+public class SearchController {
+    @Post("find")
+    public HttpResponse<?> find(@Body Input input) {
```
dstepanov, line 17: It would be nice to test more variations, like returning a simple string, a completable future reactive response.
Right now the async flow (completable future) is executed using the reactor which can be improved

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
```diff
@@ -0,0 +1,449 @@
...
+        /**
+         * Atomically set the output of this step. If this returns non-null, the caller must call
+         * {@link #work(Step, Object)} with the returned step.
+         *
+         * @param output The output of this step
+         * @return The next step to execute using {@link #work(Step, Object)}, or {@code null} if
+         * the next step will be executed later
+         */
+        @Nullable
+        final Step atomicSetOutput(Object output) {
```
dstepanov, line 195: It would be better to name the methods like they are actually in the flow, `setSomethin` is very confusing

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
```diff
@@ -0,0 +1,449 @@
...
+            Step next = this.next;
+            if (next != null) {
+                return next;
+            }
+            this.output = output;
+            next = this.next;
+            if (next != null) {
+                // another thread completed at the same time! one or both threads hit this sync
+                // block.
+                synchronized (this) {
```
dstepanov, line 209: Should we avoid using `synchronized` for Loom?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -287,7 +287,7 @@ public MutableConvertibleValues<Object> getAttributes() {
             synchronized (this) { // double check
                 attributes = this.attributes;
                 if (attributes == null) {
-                    attributes = new MutableConvertibleValuesMap<>(new HashMap<>(4));
+                    attributes = new MutableConvertibleValuesMap<>(new HashMap<>(8));
```
dstepanov, line 290: It might be a good idea to avoid creating a map, if a known attribute is set, just set it to a field

dstepanov (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: i tried this but it was slower, maybe because it needs to be an if with like 5 branches

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
yawkat: it could be implemented using a CAS but i think this is simpler and it should almost never be hit anyway. for loom it should be fine since it's only blocked for a very short time

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
yawkat: what do you mean like in the flow? the method sets the output field (if step is unset) which is why it's called that, but yea it does a bit more than set

benchmarks/src/jmh/java/io/micronaut/http/server/stack/SearchController.java
yawkat: this benchmark is from micronaut-benchmark, which does not use reactive / async code so that it's a more fair comparison between frameworks. benchmarking reactive code is also worthwhile but this simple controller is basically the best "bare-bones" test of the http stack

dstepanov: Ok, having some other benchmark testing overhead of async/reactive flow would be nice

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
dstepanov: Is just the naming is not very clear to me, I would expect something like process next, but it doesn't matter much

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
dstepanov: Ah, so maybe we can store those attributes also as fields if we use them internally

PR #8993, opened 2023-03-23
More performance improvements

yawkat:
New changes from the last PR start at the merge commit.

Only big change is the introduction of a copy on write map, based on HashMap. It's slightly faster than ConcurrentHashMap for reading. I think it will be useful elsewhere too.


dstepanov (approved):

dstepanov:
We might want to check all of the concurrent instances we use

yawkat:
fwiw ConcurrentHashMap with a concurrency level of 1 (like in the benchmark) is very close in read perf compared to this new map, and it doesn't have the problem of being very inefficient to populate. so it may be a good choice for some other places like the conversion service. aiui, concurrencyLevel of 1 makes reads faster but concurrent writes slower.

graemerocher (approved):

yawkat:
oh i just realized concurrency level is 1 by default anyway :) so no improvement possible in that area.

PR #8996, opened 2023-03-23
test: Test ObjectUtils::coerceToBoolean

sdelamo:


core/src/test/groovy/io/micronaut/core/util/ObjectUtilsSpec.groovy
```diff
@@ -21,4 +22,34 @@ class ObjectUtilsSpec extends Specification {
...
+        true     | "string"
+        false    | ""
+        false    | 0L
+        false    | new BigDecimal("0.0")
+        false    | 0
+        false    | 0.0f
+        true     | 1L
+        true     | new BigDecimal("0.1")
+        true     | 1
+        false    | -1
```
sdelamo, line 43: In Apache Groovy a negative number evaluates to `true`. Not sure if we should change this to return true for negative numbers as well.

graemerocher: yeah I think the logic for numbers is questionable. We should return `true` as well

PR #8997, opened 2023-03-23
use ObjectUtils::coerceToBoolean in Utils classes

sdelamo:
To avoid having the logic of coercing from string, collection and map to boolean duplicated.


graemerocher:
not sure about this change, `coerceToBoolean` is more expensive so it should be used sparely

PR #9000, opened 2023-03-23
test: add missing groovy tests

sdelamo:


test-suite-groovy/src/test/groovy/io/micronaut/docs/expressions/ExampleJob.groovy
```diff
@@ -9,7 +9,7 @@ class ExampleJob {
 
     @Scheduled(
             fixedRate = "1s",
-            condition = "#{!jobControl.paused}") // <1>
+            condition = '#{!jobControl.paused}') // <1>
```
sdelamo, line 12: this is not required but I think it is better to show in the docs strings with single quotes to avoid the user falling into the trap of creating a GString with string interpolation.

PR #9001, opened 2023-03-24
build: remove build breaking unicode filename

nbrugger-tgm:
The file contains non-standard (UTF-8, ASCII) but Unicode characters (in the file name) which bricks the build on my machine with "FileNotFoundException: ??????-0.0.yml"

### DIFF
--- a/http-server-netty/src/test/resources/public/测试-0.0.yml
+++ b/http-server-netty/src/test/resources/public/测试-0.0.yml
@@ -1,4 +1,0 @@
-openapi: 3.0.1
-info:
-  title: 测试
-  version: "0.0"


nbrugger-tgm:
There is a test that verifies the existence of the file (or rather is correct serving by netty).
Is it maybe possible to remove the file from gradle indexing - any gradle command fails for me when the file is there

```
Execution failed for task ':http-server-netty:processTestResources'.
> Cannot access input property 'rootSpec$1' of task ':http-server-netty:processTestResources'. Accessing unreadable inputs or outputs is not supported. Declare the task as untracked by using Task.doNotTrackState(). See https://docs.gradle.org/8.0.2/userguide/incremental_build.html#disable-state-tracking for more details.
   > Failed to create MD5 hash for file '~/micronaut-core/http-server-netty/src/test/resources/public/??????-0.0.yml' as it does not exist.
```

graemerocher:
you can probably skip the task for now `./gradlew test -X http-server-netty:processTestResources`

nbrugger-tgm:
So this pr is dismissed/to close then?
It might be a problem more people run into (maybe) but it seems OS specific or on other environmental factors.

For me, it would be ok to close it if you don't see it as a big issue or no good solution is available.

nbrugger-tgm:
I still think this should be fixed but this PR is too old and it seems to be unwanted

Issue #6395, opened 2021-10-21
Add module-info to micronaut-inject and micronaut-core


msgilligan:
### Feature description

It has been possible to use Micronaut dependency injection as a Java Module System-compatible standalone dependency injection framework by processing the jars with tools like [Badass Jlink Plugin](https://badass-jlink-plugin.beryx.org/releases/latest/) since the 1.x series. As Micronaut and its dependencies have evolved, it is now possible for (at least) `micronaut-inject` and `micronaut-core` to be "fully modularized" by adding `module-info.java` files to their jars.

This can be done while retaining Java 8 compatibility via two different mechanisms:

1. Add the `module-info.class` file directly to an otherwise JDK8-compatible jar
2. Create Multi-Release JARs that completely hide `module-info.class` from JDK8

The current dependency tree looks like:

```
+--- project <root>
+--- io.micronaut:micronaut-inject:3.1.1
|    +--- org.slf4j:slf4j-api:1.7.29 -> 2.0.0-alpha5
|    +--- javax.annotation:javax.annotation-api:1.3.2
|    +--- jakarta.inject:jakarta.inject-api:2.0.0 -> 2.0.1
|    +--- jakarta.annotation:jakarta.annotation-api:2.0.0
|    +--- io.micronaut:micronaut-core:3.1.1
|    |    \--- org.slf4j:slf4j-api:1.7.29 -> 2.0.0-alpha5
|    \--- org.yaml:snakeyaml:1.29
```

Analyzing each 3rd-party dependency:

* `slf4j-api`: Is an automatic module that can be updated by applications to 2.0.0-alpha5 or later to get fully-modular
* `:javax.annotation-api` - is an optional dependency that can be a `requires static`
* `jakarta.inject-api` - has been updated to fully-modular 2.0.1 via 2860a466b16206a83fc5eefad3bee89049e18fd6
* `jakarta.annotation-api` - fully-modular
* `snakeyaml` - is an optional dependency that can be a `requires static`

So it looks like `micronaut-inject` and `micronaut-core` are ready to go fully modular.

It would be great to see these two jars released with `module-info` support in [Milestone 3.2.0](https://github.com/micronaut-projects/micronaut-core/milestone/101)

I'm not 100% sure my analysis is correct, but I'd like to at least start a discussion here.

graemerocher:
[...]

PR #9002, opened 2023-03-24
Modularize micronaut core for JPMS

nbrugger-tgm:
Some internal classes were moved in order to utilize the hiding of internal packages with JPMS


graemerocher:
Don't think `ObjectUtils` is internal

nbrugger-tgm:
> Don't think `ObjectUtils` is internal

I assumed it is since it is annotated with `@Internal`. Should I move it back and remove the annotation, or leave as is?

graemerocher:
It is internal in that it is is not part of the public API as such, however other modules will need to access it from generated byte code for sure. also in general we should avoid breaking changes for the API where possible

nbrugger-tgm:
Reverted the requested changes: all files that were `public` and marked with `@internal` are now in their original state. It might be worth to annotate the regarding classes `BeanMap` and possibly `ObjectUtils` as `UsedByGeneratedCode` if thats the case - thats out of the scope tho.

Is everything else fine?

(the formatting stuff was to please checkstyle because of the changed imports btw)

core/src/main/java/io/micronaut/core/beans/internal/DefaultBeanWrapper.java
```diff
@@ -31,7 +33,7 @@
  * @since 1.1
  */
 @Internal
-final class DefaultBeanWrapper<T> implements BeanWrapper<T> {
+public final class DefaultBeanWrapper<T> implements BeanWrapper<T> {
```
graemerocher: seems weird we have to make these classes public when they were already package private?

dstepanov:
Can you please avoid reformating the files, as it is hard to understand the diff?

core/src/main/java/io/micronaut/core/beans/internal/DefaultBeanWrapper.java
nbrugger-tgm: Yes I don't like it either. It was necessary to move them to a "internal" package. Should I move them back out again?

dstepanov:
We should probably have some test app that is using modules

nbrugger-tgm:
> Can you please avoid reformating the files, as it is hard to understand the diff?

Gradlew check failed without it (locally at least). Any idea how to redo it without the failing checks?

dstepanov:
It can fail for some checkstyle errors you have introduced that need to be fixed

core/src/main/java/io/micronaut/core/beans/internal/DefaultBeanWrapper.java
graemerocher: yeah for me I don't understand why this change is necessary, is it a JPM requirement? If it isn't then they shouldn't moved. They fact they are internal is already expressed by the fact they are package private

nbrugger-tgm:
> We should probably have some test app that is using modules

Thats true indeed, should I create such a module as submodule in core like `examples/jpms-app`?

graemerocher:
there are `test-suite` subprojects that can be used as a model

core/src/main/java/io/micronaut/core/beans/internal/DefaultBeanWrapper.java
nbrugger-tgm: I reverted all non-necessary changes locally and will push them when i finished the "example project test suite"

nbrugger-tgm:
Sorry for asking again but this is my first contribution.
The ticket which I created this PR for was to add jpms to core and inject(-java).
But just modularizing the core is not enough to build a modular Java app. At least micronaut-runtime (Micronaut class) would be needed as a module. So there are 3 options that I would consider
- [ ] just adding core support (current state of PR) and no test suite
- [ ] adding core support and a test-suite that uses a 3rd party plugin to simulate modular support for all other modules such as micronaut-runtime
- [x] implementing jpms for all components that are required to create a sensible test suite

> As a small bonus it would be very helpful to know which parts should be tested. Is it enought if I just create a class that runs a `MicronautApplicationTest` suite (like Jakarta jck)? I will definetly add tests myself that check if permissions and so on (jpms specifics) are correct but from a noob perspective it is hard to tell wich components are affected by changes in core and in which processes it plays a major role in. For me it seems that it provides the runtime details for bean definitions, proxies that are generated by processors, but not being used by the processor directly - could be wrong ofc.

graemerocher:
seems to me we are either module or we are not so option 3?

nbrugger-tgm:
Ok, so I now pushed my work in progress state to get a look from you to verify that what I do is ok before sinking many more hours into it (extensive testing, manual testing, research, optimization, documentation etc.) when you don't even like/accept the foundation.

So while yes it touches 100 files: all of these files are Annotation processors or their add-ons, I believe that no files that would break anything were touched. 

The state of the code is the following
* It compiles 🥳 
* Its internally modular
* Not ready
* not tested

I split the code into 5 commits so that you can easily review without the 100 file distraction.

If you want to know why these files were moved: It is the least intrusive, least invasive (and only non-breaking change) way to do it. A more detailed explanation can be found in the regarding commit. If this is not enough information i have drawn diagrams that show the problems and all solutions that i came up with but i didn't want to do information overload again - if interested feel free to ask :)

Also, test most likely will partially fail since I didn't aim for that (for now!)

TL;DR: please review & read commit messages

graemerocher:
need to rebase and resolve merge conflicts so we can run the tests

aop/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
+module micronaut.aop {
+    requires transitive micronaut.core;
+    requires static micronaut.core.reactive;
+    requires static kotlin.stdlib;
```
graemerocher, line 4: why is Kotlin required? There is no requirement for Kotlin to be on the class path in a non-modular scenario?

context/src/main/java/module-info.java
```diff
@@ -0,0 +1,45 @@
+module micronaut.context {
+    requires jakarta.inject;
+    requires org.apache.logging.log4j.core;
```
graemerocher: again this is no direct dependency or requirement for log4j or logbook to be there, so why is this needed?

core-processor/src/main/java/io/micronaut/inject/beans/visitor/EntityIntrospectedAnnotationMapper.java
```diff
@@ -20,7 +20,7 @@
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Introspected;
 import io.micronaut.core.annotation.ReflectiveAccess;
-import io.micronaut.inject.annotation.NamedAnnotationMapper;
+import io.micronaut.inject.annotation.processor.NamedAnnotationMapper;
```
graemerocher, line 23: we can't change the package of this interface it would be a major breaking change to the framework and downstream modules

inject/build.gradle
```diff
@@ -11,11 +22,12 @@ micronautBuild {
 }
 
 dependencies {
-    compileOnly libs.javax.inject
     api libs.jakarta.inject.api
     api libs.managed.jakarta.annotation.api
     api project(':core')
+    api libs.jetbrains.annotations
```
graemerocher: this is a source level only dependency, it shouldn't be needed

aop/src/main/java/module-info.java
nbrugger-tgm: In a non modular scenario this file is ignored. When you use the jar on the classpath it is just a jar with classes. Module-info.java is only used when a its on the Module-path (eg. when used in modular context)

`requires static` also makes it a compile only dependency which is what we need - there references to kotlin
the dependency is from `build.gradle` -> `compileOnly libs.managed.kotlinx.coroutines.core`

context/src/main/java/module-info.java
nbrugger-tgm: Regarding to the `build.gradle` it is a requirement `compileOnly libs.log4j` but yes it should be static to reflect its compileOnly status

inject/build.gradle
nbrugger-tgm: yeah my bad - fixed jetbrains annotations and added comments to explain it

core-processor/src/main/java/io/micronaut/inject/beans/visitor/EntityIntrospectedAnnotationMapper.java
nbrugger-tgm: The only other solution that i saw was a user facing breaking change in the sense that they need to add a new annotation-processor dependency to make `:context` work.

Yes this is a breaking change for projects such as `micronaut-openapi` or `micronaut-data` but i believe this is the best and most future-proof solution as described in the related commit.

Shouldn't it be possible to adapt all "downstream projects" and ship the new version with the BOM. The only thing that would change as far as i can see is that `micronaut-data:3.x.x` is not compatible with `micronaut 4.x.x`, which i don't see a big problem with. Yes it would be better if all 2.x  & 3.x modules still worked with 4.x core but i would deem this to be acceptable to have an internal breaking change that is not user-facing

nbrugger-tgm: @graemerocher hi, I wanted to ask back if this is a hard "no" criteria or do you see any other solutions to this problem.

to summarize the problem in short words: the component `:context` should be a module. `:context` depends (compile time dependencies are dependencies too) on two modules `:inject` and `:core-processor` which both expose the same package `io.micronaut.inject.annotation`.  2 dependencies exposing the same named package to a JPMS module is not possible.

nbrugger-tgm:
As of a more recent version of micronaut the core contributors fixed the core issue that prevented JPMS.

Therefore this is not needed anymore - at least not in its current form

See https://github.com/micronaut-projects/micronaut-core/issues/6395#issuecomment-2410443457

PR #9006, opened 2023-03-24
add ability to disable streaming http request processing

graemerocher:
Adds a new `micronaut.server.netty.server-type` setting that can be set to `FULL_CONTENT` to disable streaming requests. This will make it easier to implement certain use cases like retrieving the body from a filter and also improves performance.

This PR also optimises route execution further to improvement the overall performance of the framework.


yawkat:
the tests pass now, i had to make a bunch of changes unfortunately to make all the features work. eg i reverted NettyHttpRequest to still use the old content/data approach.

i'm going to do some profiling to see if there are new simple optimizations

core/src/main/java/io/micronaut/core/bind/OneShotBindingResult.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Optional;
+
+/**
+ * Wrapper around a {@link io.micronaut.core.bind.ArgumentBinder.BindingResult} that ensures
+ * {@link #getValue()} is only called once.
+ *
+ * @param <T> The result type
+ */
+@Internal
+public final class OneShotBindingResult<T> implements ArgumentBinder.BindingResult<T> {
```
dstepanov: This shouldn't be needed after changes in my PR

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/ImmediateMultiObjectBody.java
```diff
@@ -0,0 +1,140 @@
...
+    private static CompositeByteBuf coerceToComposite(List<?> objects, ByteBufAllocator alloc) {
+        CompositeByteBuf composite = alloc.compositeBuffer();
+        for (Object object : objects) {
+            composite.addComponent(true, ((ByteBufHolder) object).content());
+        }
+        return composite;
+    }
+
+    public static Map<String, Object> toMap(Charset charset, Iterable<? extends MicronautHttpData<?>> dataList) {
+        Map<String, Object> singleMap = new LinkedHashMap<>();
```
dstepanov: You might want to use CollectionUtils.newLinkedHashMap to create a map of size dataList.size

http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/MultipartBodyArgumentBinder.java
```diff
@@ -76,88 +79,97 @@ public Argument<MultipartBody> argumentType() {
...
+                .orElse(new DefaultHttpContentProcessor(nhr, httpServerConfiguration.get()));
+            MultiObjectBody multiObjectBody;
+            try {
+                multiObjectBody = nhr.rootBody()
+                    .processMulti(processor);
+            } catch (Throwable e) {
+                throw new RuntimeException(e);
+            }
+            //noinspection RedundantCast,unchecked
+            return () -> Optional.of(subscriber -> ((Publisher<HttpData>) multiObjectBody.asPublisher()).subscribe(new CompletionAwareSubscriber<>() {
```
dstepanov: My PR doesn't use a custom subscriber as it isn't needed just a simple filter/map

http/src/main/java/io/micronaut/http/bind/binders/DefaultBodyAnnotationBinder.java
```diff
@@ -68,14 +87,66 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
             //noinspection unchecked
             return BindingResult.EMPTY;
         }
-        Optional<?> body = source.getBody();
-        if (body.isEmpty()) {
+        Object body = source.getBody().orElse(null);
```
dstepanov: How will it work when HttpRequest is received in the controller and the body is read?

core/src/main/java/io/micronaut/core/bind/OneShotBindingResult.java
dstepanov: Please add all since, author to the new files

dstepanov (changes requested):

yawkat:
@dstepanov re changes from #9017. i will merge #9006 and #9017 once #9006 is merged. if i merge before then, i would have to redo the merge since the changes in #9006 will be squashed.

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingMultiObjectBody.java
```diff
@@ -0,0 +1,272 @@
...
+
+        /**
+         * Get the next object.
+         *
+         * @return The next object, or {@code null} if the stream is done
+         */
+        @Nullable
+        public Object take() throws InterruptedException {
+            boolean demanded = false;
+            while (true) {
```
dstepanov, line 177: Why is this looping? I see it as one or nothing

dstepanov: I guess because of the missing subscription

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingMultiObjectBody.java
```diff
@@ -0,0 +1,272 @@
...
+    @Override
+    public void handleForm(FormRouteCompleter formRouteCompleter) {
+        prepareClaim().subscribe(formRouteCompleter);
+        next(formRouteCompleter);
+    }
+
+    /**
+     * A subscriber that allows blocking reads from a publisher. Handles resource cleanup properly.
+     */
+    private static final class PublisherAsBlocking implements Subscriber<Object>, Closeable {
```
dstepanov, line 83: I would implement it as a blocking queue to reduce the locking code

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingMultiObjectBody.java
dstepanov: I would implement this as `boolean consumeBlocking(Consumer c)` to handle the release in the same class

dstepanov (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingMultiObjectBody.java
yawkat: yea it's because of the subscription, and also because await() can have a spurious wakeup.

can't do release in the same method, because PublisherAsStream stores the bytebuf for a time in case the read output buffer is too small

dstepanov: Ok

http/src/main/java/io/micronaut/http/bind/binders/DefaultBodyAnnotationBinder.java
yawkat: same as before. i think this comment relates to the changes ive just reverted.

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingMultiObjectBody.java
yawkat: blockingqueue would not work i think. there would be a race condition with setting the subscription field. checking the subscription field and blocking on the result must happen in one atomic step, i dont think it's possible with blockingqueue.

core/src/main/java/io/micronaut/core/bind/OneShotBindingResult.java
yawkat: oh this became unused anyway apparently

yawkat:
idk what's wrong with sonar coverage, it shows missing coverage for some code that is clearly covered. e.g. the code for writing a non-streaming response is shown as uncovered. but i wrote that code specifically to fix tests...

dstepanov (approved):

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
```diff
@@ -282,7 +282,8 @@ public FullHttpRequest toFullHttpRequest() {
                 req = new DefaultFullHttpRequest(
                         HttpVersion.HTTP_1_1,
                         method,
-                        uriStr
+                        uriStr,
+                        false
```
JLLeitschuh, line 286: Why is header validation disabled here?

Related: https://github.com/micronaut-projects/micronaut-core/security/advisories/GHSA-694p-xrhg-x3wm

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
```diff
@@ -306,7 +307,7 @@ public StreamedHttpRequest toStreamHttpRequest() {
             String uriStr = resolveUriPath();
             io.netty.handler.codec.http.HttpMethod method = getMethod(httpMethodName);
             DefaultStreamedHttpRequest req = new DefaultStreamedHttpRequest(
-                    HttpVersion.HTTP_1_1, method, uriStr, (Publisher<HttpContent>) body);
+                                HttpVersion.HTTP_1_1, method, uriStr, false, (Publisher<HttpContent>) body);
```
JLLeitschuh, line 310: Similar question to the above, why is header validation disabled here?

http-netty/src/main/java/io/micronaut/http/netty/NettyMutableHttpResponse.java
```diff
@@ -287,6 +301,18 @@ public <T> MutableHttpResponse<T> body(@Nullable T body) {
...
     }
 
+    @Override
+    public MutableHttpResponse<B> contentType(MediaType mediaType) {
+        if (mediaType == null) {
+            headers.remove(HttpHeaderNames.CONTENT_TYPE);
+        } else {
+            // optimization for content type validation
+            mediaType.validate(() -> NettyHttpHeaders.validateHeader(HttpHeaderNames.CONTENT_TYPE, mediaType));
+            headers.setUnsafe(HttpHeaderNames.CONTENT_TYPE, mediaType);
```
JLLeitschuh, line 311: Why do this validation here? Why not put this logic in `NettyHttpHeaders` so that there exists no API on `NettyHttpHeaders` that allows for an unsafe use?

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
graemerocher: The headers were already validated in `headers.getNettyHeaders()` so we know it is safe

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
graemerocher: same answer as above

http-netty/src/main/java/io/micronaut/http/netty/NettyMutableHttpResponse.java
graemerocher: these are internal non-public APIs inaccessible to users, this PR was partly to optimise performance, so excessive validation when it is not needed is not great for performance.

PR #9009, opened 2023-03-27
Logback configuration defined via JAVA_TOOL_OPTIONS is ignored

### LINKED ISSUE #8596: logback configuration defined via JAVA_TOOL_OPTIONS is ignored
### Expected Behavior

When the path to the `logback.xml` file is defined in the `JAVA_TOOL_OPTIONS` environment variable (like e.g.: `JAVA_TOOL_OPTIONS=-Dlogback.configurationFile=/path/to/logback-external.xml`) this configuration should be used

### Actual Behaviour

The external logback XML file is used as expected until the end of the bean registration phase. From that point, the framework switches back to the default `logback.xml` file placed under the resources folder.

This behaviour was introduced in PR https://github.com/micronaut-projects/micronaut-core/pull/8238, it's caused by this method: https://github.com/micronaut-projects/micronaut-core/blob/3.8.x/management/src/main/java/io/micronaut/management/endpoint/loggers/impl/LogbackLoggingSystem.java#L135-L149

### Steps To Reproduce

1. define e.g. `logback-external.xml` file either in the resources folder or anywhere else. Make sure that it logs somehow differently than the default `logback.xml`.
2. run the application with `JAVA_TOOL_OPTIONS=-Dlogback.configurationFile=/path/to/logback-external.xml` environment variable
3. configuration defined in `logback-external.xml` is ignored once the application started up

### Environment Information

- Micronaut version: 3.8.0 - everything works fine below that
- Operating system: MacOS
- JVM: Temurin-17.0.5+8

### Example Application

_No response_

### Version

3.8.0

timyates:
Fixes #8596 

Previously, we only checked the classpath for logback configuration files, and when refreshing we ignored the `logback.configurationFile` setting.

This PR checks the filesystem if the config cannot be found on the classpath

It also adds `logback.configurationFile` as an optional property that points to the location of the config for when we refresh the configuration.

`logback.configurationFile` has precedence over the existing `logger.config` property

### DIFF
--- a/runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
+++ b/runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2023 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,8 +27,6 @@ import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import org.slf4j.LoggerFactory;
 
-import java.util.Optional;
-
 /**
  * An implementation of {@link LoggingSystem} that works with logback.
  *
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
```diff
@@ -41,8 +44,24 @@ public final class LogbackLoggingSystem implements LoggingSystem {
...
+        this(
+            Optional.ofNullable(System.getProperty("logback.configurationFile")),
+            Optional.ofNullable(logbackXmlLocation)
+        );
+    }
+
+    @Inject
+    public LogbackLoggingSystem(
+        @Property(name = "logback.configurationFile") Optional<String> logbackExternalConfigLocation,
+        @Property(name = "logger.config") Optional<String> logbackXmlLocation
```
sdelamo: instead of using a `Optional<String>` use a `@Nullable String`

sdelamo (changes requested):

runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
timyates: And change

```
        this.logbackXmlLocation = logbackExternalConfigLocation.orElseGet(() -> logbackXmlLocation.orElse(DEFAULT_LOGBACK_LOCATION));
```

to be a load of `if`s?

graemerocher (approved):

runtime/src/main/java/io/micronaut/logging/impl/LogbackLoggingSystem.java
sdelamo: Yes.

sdelamo (approved):

PR #9010, opened 2023-03-27
Remove managed logging from catalog

### LINKED ISSUE #9003: Use logging versions from micronaut-logging
These should no longer be managed here, we should instead use the versions defined in micronaut-logging

This will be a breaking change for all the modules that currently use mn.logback.classic

timyates:
We have micronaut-logging which publishes a BOM containing the managed logging libraries we want to use.

This PR stops exporting logger versions as managed dependencies.

This is a breaking change upstream, as any modules using mn.logback.classic will now need to change to importing micronaut-logging, and then using mnLogging.logback.classic

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
@@ -89,7 +89,8 @@ dependencies {
         transitive = false
     }
 
-    api libs.managed.slf4j.api
+    api platform(libs.boms.micronaut.logging)
+    api libs.slf4j.api
     compileOnly libs.caffeine
     compileOnly libs.bundles.asm
 
--- a/context/build.gradle
+++ b/context/build.gradle
@@ -11,7 +11,7 @@ dependencies {
     compileOnly project(':core-reactive')
     compileOnly project(':core-processor')
     compileOnly libs.log4j
-    compileOnly libs.managed.logback.classic
+    compileOnly libs.logback.classic
 
     // Support validation annotations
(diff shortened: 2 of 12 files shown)


sdelamo (changes requested):
I think what we should do is keep a version of logback in this repository to be used in the test classpath but micronaut-core should not depend on micronaut logging to avoid any cycles. 

logback related code in the main classpath should be moved to micronaut-logging

timyates:
@sdelamo @graemerocher @alvarosanchez @melix 

So there are some questions here.

1. Should we be moving certain (as yet undecided) logging code into micronaut-logging?
    - micronaut-logging is currently just a pom/catalog with the versions of logging we are using
2. micronaut-logging should now be added as a dependency by the Gradle and Maven build tools?
3. Are we ok keeping a version of logback in core (local to core) as suggested by Sergio?  Currently, there is no dependency cycle, as micronaut-logging is just a BOM, but if we move code in 1, I guess we should break the cycle here...

graemerocher:
1. To me it seems that we should move anything related to logback that exists in core to this module but perhaps it needs discussion
2. would have to be added explicitly by users I imagine
3. thing this would be fine

timyates:
@melix for when you get back, I guess we could we add mnLogging as a default import to micronaut-build?  Then we can switch mn.logback.classic to mnLogging.logback.classic in all our other modules

melix:
> I guess we could we add mnLogging as a default import to micronaut-build?

We can do this, yes

> Then we can switch mn.logback.classic to mnLogging.logback.classic in all our other modules

You can, already, right? By importing the catalog manually as any other micronaut catalog?

graemerocher:
Duplicate

PR #9017, opened 2023-03-28
Refactor and optimize routes

dstepanov:
PR improves routes processing:
- Separated route builder object and the object used at the runtime. That removed `HttpAttributes.ROUTE` attribute
- Multiple improvements to the request argument binders allow to postpone binding or have a binding result that is pending


core/src/main/java/io/micronaut/core/bind/ArgumentBinderRegistry.java
```diff
@@ -31,20 +31,19 @@
...
+     * Locate an {@link ArgumentBinder} for the given argument.
      *
      * @param argument The argument
-     * @param source   The source
      * @param <T>      The argument type
      * @return An {@link Optional} of {@link ArgumentBinder}
+     * @since 4.0.0
      */
-    <T> Optional<ArgumentBinder<T, S>> findArgumentBinder(Argument<T> argument, S source);
+    <T> Optional<ArgumentBinder<T, S>> findArgumentBinder(Argument<T> argument);
```
graemerocher, line 75: keep this method and add a new default one  that passes `null` to minimize the breakage

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyRequestArgumentBinder.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Optional;
+
+/**
+ * A version of {@link RequestArgumentBinder} that requires {@link NettyHttpRequest}.
+ * @param <T> A type
+ * @author Denis Stepanov
+ * @since 4.0.0
+ */
+@Experimental
+public interface NettyRequestArgumentBinder<T> extends RequestArgumentBinder<T> {
```
graemerocher: make `sealed`

http/src/main/java/io/micronaut/http/bind/binders/BodyRequestArgumentBinder.java
```diff
@@ -0,0 +1,89 @@
...
+                if (ClassUtils.isJavaLangType(argumentType.getType())) {
+                    return Optional::empty;
+                } else {
+                    return () -> source.getBody(argumentType);
+                }
+            }
+            return result;
+        } else {
+            if (source.getBody().isPresent()) {
+                Optional<String> text = source.getBody(String.class);
```
graemerocher: use `stringValue`

http/src/main/java/io/micronaut/http/bind/binders/BodyRequestArgumentBinder.java
```diff
@@ -0,0 +1,89 @@
...
+
+        Argument<T> argument = context.getArgument();
+        Argument<T> argumentType;
+        if (argument.getType() == Optional.class) {
+            argumentType = (Argument<T>) argument.getFirstTypeVariable().orElse(argument);
+        } else {
+            argumentType = argument;
+        }
+
+        String inputName = argument.getAnnotationMetadata().stringValue(Bindable.NAME).orElse(argument.getName());
```
graemerocher: Perhaps we can add a new interface:

```java
interace IntializableArgumentBindiner<T> extends ArgumentBinder<T> {
    void initialize(Argument<?> argument);
}
```

And then this logic can be moved to the part that sets up the argument binder instead of executing each time we bind.

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
```diff
@@ -134,7 +129,7 @@ private Chunk lastChunk() {
      * @return The chunk, or {@code null} if this data is {@link #isCompleted() completed} and all
      * chunks have been polled.
      */
-    public Chunk pollChunk() {
+    public synchronized Chunk pollChunk() {
```
yawkat: these syncs should not be necessary. These calls should not happen concurrently

http/src/main/java/io/micronaut/http/bind/binders/BodyRequestArgumentBinder.java
dstepanov: The binders are singletons shared with different routes, we might want to have something like `IntializableArgumentBindiner specific(..)`

http/src/main/java/io/micronaut/http/bind/binders/BodyRequestArgumentBinder.java
dstepanov: Not sure what do you mean

core/src/main/java/io/micronaut/core/bind/ArgumentBinderRegistry.java
dstepanov: Reverted and made them deprecated

http/src/main/java/io/micronaut/http/bind/binders/BodyRequestArgumentBinder.java
dstepanov: I will add this in the next PR

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
dstepanov: Reverted

yawkat:
i like the approach to form handling, but i think it needs some work wrt duplicate field handling and such, so i'm working on that as part of the merge with #9006

yawkat:
merged into #9006

PR #9018, opened 2023-03-29
Improve support for logback 1.3+

timyates:
Logback 1.3.x+ changed the binary format for Configurators, and introduced a default one.

The issue is that when we call .configure on one, the JVM crashes as there is now an unexpected return value.

I cannot find a way of detecting and logging this issue (without some sort of reflection)

This change checks to see if the configurator we've detected is the default one added to logback in 1.3.x here

https://github.com/qos-ch/logback/blob/5ac98f440dabec45d8ab9b3519b2aa308f05793b/logback-classic/src/main/java/ch/qos/logback/classic/util/DefaultJoranConfigurator.java

And if it is, we ignore it.

This will not fix it for people using a Custom Configurator that they compile under Logback 1.3+, they will still see the crash with no warning.

But people upgrading to 1.3+ should see their application just run, and if they change their logback.xml emit debug with

```
<configuration debug="true">
```

They will see

```
12:55:21,222 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Skipping ch.qos.logback.classic.util.DefaultJoranConfigurator as it's assumed to be from an unsupported version of Logback
```

### DIFF
--- a/runtime/src/main/java/io/micronaut/logging/impl/LogbackUtils.java
+++ b/runtime/src/main/java/io/micronaut/logging/impl/LogbackUtils.java
@@ -39,6 +39,8 @@ import java.util.function.Supplier;
  */
 public final class LogbackUtils {
 
+    private static final String DEFAULT_LOGBACK_13_PROGRAMMATIC_CONFIGURATOR = "ch.qos.logback.classic.util.DefaultJoranConfigurator";
+
     private LogbackUtils() {
     }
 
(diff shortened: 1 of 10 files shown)


timyates:
To test this, if you undo the change here

https://github.com/micronaut-projects/micronaut-core/pull/9018/files#diff-48c626371a6ef34faaa5773e2a97b0168e7c916d6532f65956b5e736fd54a644R93

Then the new `test-suite-logback-14` should fail with initialization exceptions

graemerocher (approved):

sdelamo (approved):

PR #9022, opened 2023-03-29
Support JDK 20 in annotation processors

dstepanov:
Do we need to do this every Java release?


dstepanov:
I think this will require Java 20 build.

PR #9032, opened 2023-03-30
Use `Map.of` instead of custom implementation

dstepanov:
It looks like Java's `Map.of` is more performant than our `ImmutableSortedStringsArrayMap`, I think it's better to use the native implementation.

We still have to keep the previous implementation till everything is recompiled and all nulls in map values are eliminated.

There are some stats:
Before:
AnnotationValueBenchmark.benchMarkGetValue  thrpt    5  4686109.127 ± 176737.942  ops/s
After:
AnnotationValueBenchmark.benchMarkGetValue  thrpt    5  4999775.733 ± 276019.217  ops/s

Before:
FullHttpStackBenchmark.test  MICRONAUT  avgt   30  20530.775 ± 252.348  ns/op
After:
FullHttpStackBenchmark.test  MICRONAUT  avgt   30  19438.006 ± 94.318  ns/op

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
@@ -15,18 +15,17 @@
  */
 package io.micronaut.inject.annotation;
 
+import io.micronaut.context.expressions.AbstractEvaluatedExpression;
 import io.micronaut.core.annotation.AnnotationClassValue;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.AnnotationMetadataDelegate;
-import io.micronaut.core.annotation.AnnotationUtil;
-import io.micronaut.core.expressions.EvaluatedExpressionReference;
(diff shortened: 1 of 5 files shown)


core-processor/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
```diff
@@ -698,13 +659,56 @@ private static void pushValue(Type declaringType, ClassVisitor declaringClassWri
...
 
             methodVisitor.invokeConstructor(type, CONSTRUCTOR_CONTEXT_EVALUATED_EXPRESSION);
-
         } else {
-            methodVisitor.visitInsn(ACONST_NULL);
+            throw new IllegalStateException("Unsupported Map value:  " + value + " " + value.getClass().getName());
+        }
+    }
+
+    public static boolean isSupportedMapValue(Object value) {
```
yawkat, line 672: i think this should just be a giant ||

yawkat (approved):

graemerocher (approved):

PR #9033, opened 2023-03-31
ci: fold release-notes & sonar in gradle workflow

sdelamo:


graemerocher (approved):

sdelamo:
I kept `temurin` in `gradle.yml` workflow since we need to test 11 8 and 17. I kept the graalvm.yml to test Graal with 17

PR #9034, opened 2023-03-31
Make evaluation context objects effectively singleton

graemerocher:
To ensure that the performance of evaluating expressions we should specify that any context objects are effectively singleton and cache them in the evaluation context. This is eliminates questions about how/where these objects are destroyed (singletons are always destroyed by the bean context).


sdelamo (approved):

GavrilovSV:
The `ExpressionEvaluationContext` implements `AutoCloseable` and is currently used in `AbstractEvaluatedExpression` in the following way 
```
try (evaluationContext) {
    return doEvaluate(evaluationContext);
}
```
which means that `resolutionContext` inside `DefaultExpressionEvaluationContext` will be cleared and reset to `null` after each evaluation. So if it is reused and caches beans, it needs to be closed somewhere else

PR #9039, opened 2023-03-31
Avoid the hierarchy annotaiton metadata for @Introspected(classes=)

andriy-dmytruk:
Stop adding all the annotations from class that uses `@Introspected(classes=)` to the classes that it imports.

For example, given
```java
class MyClass {}

@MyAnn
@Introspected(classes=MyClass.class)
class MyImportClass {}
```
`@MyAnn` will be present on the `MyClass` introspection even though it wasn't explicitly defined on it. Additionally, the `@Introspected(classes=MyClass.class)` will be present, too. This can be a problem when user wants to import multiple classes like `@Introspected(classes={MyCalss1.class, ..., MyClass1000.class})` creating huge introspections for all the classes.

I couldn't find any documentation for this behavior, so I think it makes sense to remove it. I also don't know of any implementations that could depend on it. But if needed, additional annotations during import can be added by creating a type element visitor.


andriy-dmytruk:
@graemerocher Can we add this to `4.0.x` as it shouldn't matter if it is a breaking change?

yawkat (approved):

PR #9061, opened 2023-04-04
Idea(8855)/service health indicator

alexsimpson440dev:
Discussion Reference: https://github.com/micronaut-projects/micronaut-core/discussions/8855 

**What this does:**
- Adds additional configuration to endpoints for service-http-client.enabled (default is false)
- Provides a health indicator for each service http client (needs health-check set to true)

**Why:**
- Allows us to use the health endpoint, if configured, to validate application health
- Ties into the current health check on service http clients to determine health

**Example use:**
Configuration:
image

image

Healthy Response:
image

Unhealthy Service:
image


http-client-core/build.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
     implementation libs.managed.reactor
 
     api project(":runtime")
+    api project(":management")
```
graemerocher: this management dependency can't be required here

graemerocher (changes requested):
Please target the 3.9.x branch as this is an enhancement

alexsimpson440dev:
@graemerocher When you day management cannot be a dep. do you want me to move that indicator under management then? From my understanding to keep it under http-client-core it would need that dep as either api or implementation?

http-client-core/build.gradle
alexsimpson440dev: @graemerocher When you say management cannot be required, do you want me to move that indicator under management then? From my understanding to keep it under http-client-core it would need that dep as either api or implementation?

graemerocher: needs to be `compileOnly`

graemerocher: or moved to management module

management/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
         exclude module:'micronaut-inject'
         exclude module:'micronaut-bom'
     }
+    implementation project(":http-client-core")
```
graemerocher: should be `compileOnly`

management/src/main/java/io/micronaut/management/health/indicator/client/ServiceHttpClientHealthIndicatorFactory.java
```diff
@@ -0,0 +1,86 @@
...
+ * Returns {@link HealthStatus#DOWN} if there are no available URLs in the load balancer.</p>
+ *
+ * @author Alexander Simpson
+ * @since 3.9
+ */
+@Factory
+@EachBean(ServiceHttpClientConfiguration.class)
+@Requires(beans = HealthEndpoint.class)
+@Requires(property = HealthEndpoint.PREFIX + ".service-http-client.enabled", defaultValue = StringUtils.FALSE, notEquals = StringUtils.FALSE)
+public class ServiceHttpClientHealthIndicatorFactory implements HealthIndicator {
```
graemerocher: add `@Requires(classes= ServiceHttpClientConfiguration.class)`

graemerocher (changes requested):

alexsimpson440dev:
@graemerocher any other changes needed in this PR? I added the required annotation.

management/src/main/java/io/micronaut/management/health/indicator/client/ServiceHttpClientHealthIndicatorFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Map;
+
+/**
+ * <p>A {@link io.micronaut.management.health.indicator.HealthIndicator} used to display available load balancer URLs.
+ * Returns {@link HealthStatus#DOWN} if there are no available URLs in the load balancer.</p>
+ *
+ * @author Alexander Simpson
+ * @since 3.9
+ */
+@Factory
```
graemerocher: Remove this annotation, this is not a factory as it produces no beans from methods defined in the type

graemerocher (changes requested):

management/src/main/java/io/micronaut/management/health/indicator/client/ServiceHttpClientHealthIndicatorFactory.java
alexsimpson440dev: done

graemerocher (approved):

PR #9063, opened 2023-04-05
Performance optimizations: cache executor, optimize HttpMethod enum

graemerocher:


http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
```diff
@@ -104,7 +104,8 @@ private boolean isCollectionOrArray(Class<?> clazz) {
     }
 
     @Override
-    public List<String> getAll(CharSequence name) {
+    public List<String>
```
yawkat: revert

yawkat (approved):

dstepanov (approved):

PR #9069, opened 2023-04-05
Microoptimizations

yawkat:
Reverted the executeOn change because it likely duplicates #9063

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
@@ -20,14 +20,16 @@ import io.micronaut.core.convert.ArgumentConversionContext;
 import io.micronaut.core.convert.ConversionService;
 import io.micronaut.core.type.MutableHeaders;
 import io.micronaut.http.HttpHeaderValues;
-import io.micronaut.http.HttpHeaders;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.MutableHttpHeaders;
+import io.micronaut.http.util.HttpHeadersUtil;
 import io.netty.handler.codec.http.DefaultHttpHeaders;
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValidationUtil;
+import jakarta.annotation.Nullable;
 
 import java.net.URI;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDateTime;
@@ -38,6 +40,7 @@ import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Set;
@@ -269,7 +272,7 @@ public class NettyHttpHeaders implements MutableHttpHeaders {
     @Override
     public Optional<MediaType> contentType() {
         // optimization to avoid ConversionService
-        String str = get(HttpHeaders.CONTENT_TYPE);
+        String str = get(HttpHeaderNames.CONTENT_TYPE);
         if (str != null) {
             try {
                 return Optional.of(MediaType.of(str));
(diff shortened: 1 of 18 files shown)


http/src/main/java/io/micronaut/http/util/HttpHeadersUtil.java
```diff
@@ -92,4 +93,28 @@ private static String mask(@Nullable String value) {
...
+     * Split an accept-x header and get the first component. If the header is {@code *}, return
+     * null.
+     *
+     * @param text The input header
+     * @return The first part of the header, or {@code null} if the header is {@code *}
+     * @since 4.0.0
+     */
+    @Internal
+    @Nullable
+    public static String splitAcceptHeader(@NonNull String text) {
```
sdelamo, line 107: it would be good to have a test for this method

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
```diff
@@ -292,8 +295,48 @@ public OptionalLong contentLength() {
...
+    @Override
+    public List<MediaType> accept() {
+        // use HttpHeaderNames instead of HttpHeaders
+        return MediaType.orderedOf(getAll(HttpHeaderNames.ACCEPT));
+    }
+
+    @Nullable
+    @Override
+    public Charset acceptCharset() {
+        String text = get(HttpHeaderNames.ACCEPT_CHARSET);
```
sdelamo, line 307: I think this method would be better in `HttpHeaderUtils` with signature 

```java
@Nullable
public static Charset acceptCharset(@Nullable String acceptHttpHeaderValue)
````

That way we could use it in non netty implementations

File: http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java [hunk 3]
```diff
@@ -292,8 +295,48 @@ public OptionalLong contentLength() {
...
+            } catch (Exception ignored) {
+            }
+        }
+        // default to UTF-8
+        return StandardCharsets.UTF_8;
+    }
+
+    @Nullable
+    @Override
+    public Locale acceptLanguage() {
```
sdelamo, line 324: I think this method would be better in `HttpHeaderUtils` with signature 

```java
@Nullable
public static Locale acceLanguage(@Nullable String acceptHttpHeaderValue)
````

That way we could use it in non netty implementations

File: http/src/main/java/io/micronaut/http/HttpHeaders.java [hunk 4]
```diff
@@ -722,4 +767,15 @@ default Optional<String> getAuthorization() {
...
     }
+
+    /**
+     * Equivalent to {@code contains(SERVER)}, optimized for internal use only.
+     *
+     * @return {@code contains(SERVER)}
+     * @since 4.0.0
+     */
+    @Internal
+    default boolean containsServer() {
```
graemerocher: rather than include an internal method into the public API which is not ideal just cast down to NettyHttpRequest and invoke the logic directly

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
```diff
@@ -292,8 +295,48 @@ public OptionalLong contentLength() {
...
 
+    @Override
+    public List<MediaType> accept() {
+        // use HttpHeaderNames instead of HttpHeaders
+        return MediaType.orderedOf(getAll(HttpHeaderNames.ACCEPT));
+    }
+
+    @Nullable
+    @Override
+    public Charset acceptCharset() {
```
graemerocher, line 306: is there some performance related motivation for not returning optional here?

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
graemerocher, line 324: Same question is there some performance related reason for not just returning optional?

inject/src/main/java/io/micronaut/context/event/ApplicationEventPublisher.java
```diff
@@ -63,4 +64,14 @@ static <K> ApplicationEventPublisher<K> noOp() {
...
     }
 
+    /**
+     * Check whether this publisher is empty (i.e. has no listeners). If this method returns
+     * {@code true}, {@link #publishEvent(Object)} does not need to be called.
+     *
+     * @return {@code true} iff there are no subscribers
+     * @since 4.0.0
+     */
+    default boolean empty() {
```
graemerocher: rename to `isEmpty()`

graemerocher (changes requested):

dstepanov (approved):

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
yawkat: it uses netty-specific apis (HttpHeaderNames) for performance

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
yawkat: it uses netty-specific apis (HttpHeaderNames) for performance

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
yawkat: i saw optional overhead in the profile so i avoided it for new methods. it should be eliminated if the method is inlined but i don't want to rely on it

http/src/main/java/io/micronaut/http/HttpHeaders.java
yawkat: it's called from RouteExecutor which is in http-server so it can't access netty

graemerocher: code should be refactored then to allow subclass RouteExecutor for the Netty implementation. There shouldn't be internal methods like this in public interfaces. If we absolutely must keep it, then it should be renamed `containsServerHeader`

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
graemerocher: weird because looking at the implementation and where these methods are called they are wrapped in optionals straight a way, or am I missing something?

http/src/main/java/io/micronaut/http/HttpHeaders.java
yawkat: i think i'll just revert it then, it's a really tiny perf difference

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
yawkat: the default implementation uses optional because i just copied the old code. the netty one does not.

the callers still use optional atm. i didn't really try to remove optional, i just avoided it as part of the api for future proofing.

graemerocher (approved):

PR #9073, opened 2023-04-06
Expose ConnectionManager for use in Oracle Cloud Netty client

yawkat:
This PR exposes some ConnectionManager APIs in the netty http client so that I can use them to implement connection pooling in oraclecloud-httpclient-netty. Unfortunately OCI is still peculiar about how requests are processed so the client cannot be implemented with the normal Micronaut HTTP client API, but ConnectionManager offers a good enough API to make use of the basic HTTP stack (SSL, HTTP2/HTTP3...) and connection pooling.

ConnectionManager did change a bit in 4.0.x, and connection pooling for oraclecloud-httpclient-netty isn't super urgent, which is why this PR is targeted to 4.0.x.

In local testing these changes in micronaut-core are sufficient to implement the pooling client. I need to do more thorough testing though.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -456,6 +456,15 @@ public class DefaultHttpClient implements
         return log;
     }
 
+    /**
+     * Access to the connection manager, for micronaut-oracle-cloud.
+     *
+     * @return The connection manager of this client
+     */
+    public ConnectionManager connectionManager() {
+        return connectionManager;
+    }
+
     @Override
     public HttpClient start() {
         if (!isRunning()) {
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -456,6 +456,15 @@ public Logger getLog() {
         return log;
     }
 
+    /**
+     * Access to the connection manager, for micronaut-oracle-cloud.
+     *
+     * @return The connection manager of this client
+     */
+    public ConnectionManager connectionManager() {
```
sdelamo, line 464: Annotate with `@NonNull` the return type?

sdelamo (approved):

PR #9074, opened 2023-04-06
Disable leak detection by default

yawkat:


dstepanov (approved):

graemerocher (approved):

sdelamo (changes requested):
@yawkat can you update the docs to clearly state that it is disabled by default and how to enable it. 

Currently,  it says: 

> Netty enables simplistic sampling resource leak detection which reports there is a leak or not, at the cost of small overhead. You can disable it or enable more advanced detection by setting property netty.resource-leak-detector-level to one of: SIMPLE (default), DISABLED, PARANOID or ADVANCED.

yawkat:
@sdelamo done

PR #9076, opened 2023-04-06
More performance optimizations

dstepanov:


graemerocher (approved):

http/src/main/java/io/micronaut/http/HttpResponse.java
```diff
@@ -428,4 +428,25 @@ default Cookies getCookies() {
...
         throw new UnsupportedOperationException("Operation not supported on a " + this.getClass() + " response.");
     }
+
+    /**
+     * Returns a mutable response based on this response.
+     * @return the mutable response
+     * @since 4.0.0
+     */
+    default MutableHttpResponse<?> toMutableResponse() {
+        if (this instanceof MutableHttpResponse<?> mutableHttpResponse) {
```
yawkat, line 438: just override it in MutableHttpResponse instead

yawkat: oh i see you did that, remove this check then.

router/src/main/java/io/micronaut/web/router/DefaultMethodBasedRouteInfo.java
```diff
@@ -54,6 +54,7 @@
     private final Optional<Argument<?>> optionalFullBodyArgument;
 
     private RequestArgumentBinder<Object>[] argumentBinders;
+    private boolean needsBody;
```
yawkat: final

yawkat:
```suggestion
    private final boolean needsBody;
```

router/src/main/java/io/micronaut/web/router/DefaultMethodBasedRouteInfo.java
```diff
@@ -160,4 +171,9 @@ public Optional<Argument<?>> getFullBodyArgument() {
     public String[] getArgumentNames() {
         return argumentNames;
     }
+
+    @Override
+    public boolean needsRequestBody() {
+        return needsBody || super.needsRequestBody();
```
yawkat, line 177: i think you could fold super.needsRequestBody into needsBody too, call it in the constructor. probably doesn't matter though.

yawkat (approved):

PR #9085, opened 2023-04-11
use Milestone versions

sdelamo:


sdelamo:
@timyates can you check why this fails?

timyates:
Failing scan here (obfuscated in the action) https://ge.micronaut.io/s/cizejqpdmkbeq

timyates:
The management test was failing as micronaut-sql was adding a constraint on a newer version of H2 than we had in core (so the version string was different)

I've fixed the version string in the test, and aligned the version here to match that in micronaut-sql

timyates:
For the KSP error, we need M3 of the validation library with https://github.com/micronaut-projects/micronaut-validation/commit/eea7bd05b8761fccf13f5307f8b9d512a0e0011e

PR #9090, opened 2023-04-11
io_uring support

yawkat:
draft for the moment, i think this is broken because of https://github.com/netty/netty-incubator-transport-io_uring/pull/202#issuecomment-1503146258


dstepanov (approved):

dstepanov:
Maybe we should have some test suite project that tests it if run on linux

PR #9094, opened 2023-04-11
Support for conditional routing

graemerocher:
Adds a new `@RequestCondition` annotation which can make the route execution logic conditional.

### DIFF
--- a/http/src/main/java/io/micronaut/http/expression/RequestConditionContext.java
+++ b/http/src/main/java/io/micronaut/http/expression/RequestConditionContext.java
@@ -15,6 +15,8 @@
  */
 package io.micronaut.http.expression;
 
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.context.ServerRequestContext;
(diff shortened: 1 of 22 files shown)


sdelamo (commented):
I think this could potentially replace our API Version logic. https://docs.micronaut.io/latest/guide/#apiVersioning

Maybe we could replace `@Version("1")` with an annotation mapper to `condition = "#{request.headers.getFirst('X-API-Version')?.equals('1')}"`

graemerocher:
could be challenging to replace that since it works on the client and allows pluggable strategies

http/src/main/java/io/micronaut/http/expression/RequestConditionContext.java
```diff
@@ -0,0 +1,47 @@
...
+ * An expression evaluation context for use with HTTP annotations and the {@code condition} member.
+ *
+ * <p>This context allows access to the current request via a {@code request} object which is an instance of {@link HttpRequest}.</p>
+ *
+ * @see HttpRequest
+ * @author graemerocher
+ * @since 4.0.0
+ */
+@Singleton
+public final class RequestConditionContext {
```
yawkat, line 36: make constructor internal

yawkat (approved):

sdelamo (approved):

PR #9095, opened 2023-04-11
Bump micronaut-data to 3.10.0

micronaut-build:


sdelamo:
@radovanradic @dstepanov we should have changed `githubCoreBranch` to 3.9.x in Micronaut Data so that a minor of data goes to the next minor of Micronaut

sdelamo (approved):

PR #9105, opened 2023-04-14
Add targetPackage property to the @Introspected annotation

andriy-dmytruk:
The property can be used in the following manner:
```
@Introspected(introspectionPackage="my.package.introspections")
class SomeClass {}
```
This can be used when `SomeClass` was built externally but the class was processed by an annotation processor as sources to create introspection metadata (and avoid "signer information does not match signer information of other classes in the same package" error). It can be also used in combination with the `classes` property of the annotation.


core/src/main/java/io/micronaut/core/annotation/Introspected.java
```diff
@@ -156,6 +156,13 @@
      */
     String withPrefix() default "with";
 
+    /**
+     * @return The package to write introspections to. By default, uses the class package.
+     * @since 3.9.0
+     */
+    @Experimental
+    String introspectionPackage() default "";
```
graemerocher: rename to `targetPackage`

inject/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
```diff
@@ -180,6 +180,8 @@ private void processIntrospected(ClassElement element, VisitorContext context, A
         Introspected.AccessKind[] accessKinds = introspected.enumValues("accessKind", Introspected.AccessKind.class);
         Introspected.Visibility[] visibilities =
                 introspected.enumValues("visibility", Introspected.Visibility.class);
+        final String introspectionPackage = introspected.stringValue("introspectionPackage").orElse(element.getPackageName());
```
graemerocher: If this package names differ we need to change the default visibility to only public members (see `Visibility.PUBLIC`) since package private and protected members will no longer be visible from different packages. In the 4.0.x branch we support reflection as a fall back but not here.

graemerocher (changes requested):

sdelamo:
@andriy-dmytruk should I move this PR to ready for review?

graemerocher (approved):

PR #9115, opened 2023-04-15
Fix the issue where ssl. protocol cannot be reset

ai-houzi:
I fixed the issue where the SSL protocol option could not be reset


sdelamo:
@ai-houzi we would need for you to sign the CLA. I have added the fix with a test in this PR https://github.com/micronaut-projects/micronaut-core/pull/9118

ai-houzi:
@sdelamo Thank you for your reminder. If I had the opportunity to submit a pull request here, this would not have happened

PR #9122, opened 2023-04-16
Custom Messager for annotation processor

### LINKED ISSUE #9121: Need to implement custom Messager for annotation processor
### Feature description

Look to this problem: https://github.com/micronaut-projects/micronaut-openapi/issues/869

Now all micronaut annotation messages from context printed as errors.

So, about this problem: I looked at the code again - everything was done correctly from our side. The annotation processor writes informative messages with Diagnostic.Kind.NOTE. Look at screenshots
изображение
Next class io.micronaut.inject.visitor.VisitorContext:
изображение
Next JavaVisitorContext:
изображение
Next JavacMessager:
изображение
Next method `printMessage(Diagnostic.Kind kind, CharSequence msg, Element e, AnnotationMirror a, AnnotationValue v)`:
изображение
We run `log.note()` method.
As you see, we use standard JDK JavacMessager which print logs by `com.sun.tools.javac.util.Log` class.

The problem in this block in `Log` class:
изображение
We print message with type `WriterKind.NOTICE` and as you see writer for `NOTICE` is `System.err`. 

This is either a bug inside the `com.sun` library, or it's meant to be. In any case, to change this, it is tedious to create a custom messenger that will write messages with the correct logging level.

That's why all messages from openapi now printed with error mark.

Link to issue: https://github.com/micronaut-projects/micronaut-openapi/issues/869

altro3:
Fixed #9121


graemerocher:
we can't use internal non-public Java APIs in the processors. This would need to be reported as a JDK issue.

altro3:
@graemerocher But this is just a sample. I can implement the messenger implementation without using JDK internals because the `javax.annotation.processing.Messager` interface is public. What do you think?

graemerocher:
sure if only public APIs can be used

altro3:
Oh, I tried, but understand, can't do it without access to context, but this is internal api :-(

PR #9128, opened 2023-04-19
Replace RoutingInboundHandler and streams handlers by a single handler

yawkat:
This PR replaces the HttpStreamsServerHandler, HandlerPublisher, HttpRequestDecoder and HttpResponseDecoder by a new PipeliningServerHandler. PipeliningServerHandler directly receives netty HttpMessages, combines them into a FullHttpRequest or StreamedHttpRequest, and passes them on to a handler. It also takes care of pipelining, so that the responses for different requests are sent in the same order those requests came in, which fixes some old bugs in RoutingInboundHandler.

PipeliningServerHandler optimizes the read and write operations. Concurrent reads are handled safely, removing the need for the FlowControlHandler even in the streaming case. Flushes are delayed until readComplete, so that multiple requests in the same TCP packet can be responded to in a single TCP packet as well.

The biggest performance win comes from message aggregation, however. While PipeliningServerHandler can handle a FullHttpRequest from FULL_CONTENT, it can also do some basic aggregation on its own. Small requests commonly have the HttpRequest and HttpContent come in in the same read call. PipeliningServerHandler will "hold back" the HttpRequest until readComplete to merge it with the HttpContent if available, into a FullHttpRequest. This gives a performance benefit downstream, it's actually ~600ns faster in FullHttpStackBenchmark by default than 4.0.x with FULL_CONTENT.

The big advantage over FULL_CONTENT is that this aggregation is always enabled and is not a compatibility issue. Large requests will still safely fall back to StreamedHttpRequest. FULL_CONTENT is still useful for customizers that want to observe the full HTTP message, however, and may give some perf benefit for medium-sized request bodies.

The reason this PR is so big is that now, many small requests are processed using the FullHttpRequest APIs. There are many subtle differences in this code path. For example, form items are parsed immediately and then passed to FormRouteCompleter, where previously parsing and processing would be interleaved (the reactive code path). This exposed some bugs because it moves some `if (refCnt > 0) release` calls to before the processing.


http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -216,42 +183,51 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
...
 
     @Override
-    protected void channelRead0(ChannelHandlerContext ctx, io.micronaut.http.HttpRequest<?> httpRequest) {
-        new NettyRequestLifecycle(this, ctx, (NettyHttpRequest<?>) httpRequest).handleNormal();
+    public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRequest request, PipeliningServerHandler.OutboundAccess outboundAccess) {
+        NettyHttpRequest<Object> mnRequest;
+        try {
+            mnRequest = new NettyHttpRequest<>(request, ctx, conversionService, serverConfiguration);
+        } catch (IllegalArgumentException e) {
+            NettyHttpRequest<Object> errorRequest = new NettyHttpRequest<>(
```
dstepanov, line 204: What is this scenario?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -216,42 +183,51 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
...
+                new DefaultFullHttpRequest(request.protocolVersion(), request.method(), "/", Unpooled.EMPTY_BUFFER),
+                ctx,
+                conversionService,
+                serverConfiguration
+            );
+            new NettyRequestLifecycle(this, outboundAccess, errorRequest).handleException(e.getCause() == null ? e : e.getCause());
+            if (request instanceof StreamedHttpRequest streamed) {
```
dstepanov, line 211: I saw this code before. Maybe we can have something like `.discard()` and every request type will have it's own logic there

http-client/src/test/groovy/io/micronaut/http/client/StreamRequestSpec.groovy
```diff
@@ -88,8 +88,8 @@ class StreamRequestSpec extends Specification {
         )).contentType(MediaType.TEXT_PLAIN_TYPE), List)).blockFirst()
 
         then:
-        result.body().size() == 5
-        result.body() == ["Number 0", "Number 1", "Number 2", "Number 3", "Number 4"]
+        // no guarantee that the strings aren't merged, so we have to allow for it
+        result.body().join("") == "Number 0Number 1Number 2Number 3Number 4"
```
graemerocher, line 92: can you explain this change?

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpRequestBuilder.java
```diff
@@ -15,85 +15,118 @@
...
-        Objects.requireNonNull(request, "The request cannot be null");
-        while (request instanceof HttpRequestWrapper) {
-            request = ((HttpRequestWrapper<?>) request).getDelegate();
-        }
-        if (request instanceof NettyHttpRequestBuilder) {
-            return ((NettyHttpRequestBuilder) request).toHttpRequest();
+        return asBuilder(request).toHttpRequestWithoutBody();
+    }
+
+    static NettyHttpRequestBuilder asBuilder(@NonNull io.micronaut.http.HttpRequest<?> request) {
```
graemerocher, line 119: javadoc

dstepanov (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: parse error of the URI

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: unfortunately FullHttpRequest is a netty type. we could make StreamedHttpRequest implement ReferenceCounted but i dont think i want to open that can of worms

http-client/src/test/groovy/io/micronaut/http/client/StreamRequestSpec.groovy
yawkat: because it's TEXT_PLAIN mode, the client in this test simply sends the strings after one another, so the actual http stream is "Number 0Number 1Number 2Number 3Number 4". Each part is a different http chunk i think (often also a different tcp packet, but not always, which is why this test still worked sometimes).

With the change in this PR, the framework now can merge the different http chunks if they happen in the same TCP packet, which doesn't change the "HTTP stream", but changes how it's received by the controller (a single string vs a string for each part). 

it's not a problem if the user uses a format like json streams, where there is an actual delimiter between items

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: i will add a comment here and in other places

graemerocher:
Looks like the last test run resulted in a failure

PR #9129, opened 2023-04-19
Fix disable log in environments

altro3:
This PR is continues #8946. Sorry, missed that slf4j initialized in constructor of `DefaultEnvironment`. Need add ability to set `logEnabled` flag in constructor.

Also fixed checkstyle errors.


altro3:
@graemerocher @dstepanov  could you look?

altro3:
@graemerocher @dstepanov Could you include this fix to release 3.9.1?

PR #9132, opened 2023-04-20
Add a TCK test for String based Exception handler

timyates:
This fails in 4.0.0 as the response is OK


graemerocher:
in 4.x we need to document this breaking change

timyates:
I'll do that as part of the merge-up and fix 👍

sdelamo (approved):

PR #9137, opened 2023-04-21
Use StringIntMap for propertyIndexOf

yawkat:
This patch replaces the propertyIndexOf method generated for each introspection with an optimized String->int map. For cases where propertyIndexOf may be called for many different types, this avoids a megamorphic call site (benchmark about 5x as fast). On the other hand, for cases where propertyIndexOf is only called with one or two different introspections at the same call site, this is slightly slower (benchmark about 0.65x as fast).

imo this tradeoff is acceptable, because the speedup for megamorphic case far exceeds the slowdown for the monomorphic case. Monomorphic call sites should also be less "hot" in practical code, so performance of propertyIndexOf matters less. Finally, the absolute performance difference is quite small for the monomorphic case, in the benchmark it's 70ns so roughly 1.5ns per item, while for the megamorphic case it's roughly 30ns per item.

Benchmark results, before this change:
```
Benchmark                    (itemCount)  (typeCount)  Mode  Cnt    Score   Error  Units
PropertyIndexBenchmark.test           50            1  avgt    5  115.205 ± 1.016  ns/op
PropertyIndexBenchmark.test           50            2  avgt    5  118.745 ± 0.106  ns/op
PropertyIndexBenchmark.test           50            3  avgt    5  832.091 ± 8.271  ns/op
```

After this change:
```
Benchmark                    (itemCount)  (typeCount)  Mode  Cnt    Score   Error  Units
PropertyIndexBenchmark.test           50            1  avgt    5  183.315 ± 0.435  ns/op
PropertyIndexBenchmark.test           50            2  avgt    5  162.453 ± 0.881  ns/op
PropertyIndexBenchmark.test           50            3  avgt    5  162.690 ± 1.004  ns/op
```

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
+++ b/inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
@@ -31,6 +31,7 @@ import io.micronaut.core.reflect.exception.InstantiationException;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.ReturnType;
 import io.micronaut.core.util.ArgumentUtils;
+import io.micronaut.core.util.StringIntMap;
 import io.micronaut.inject.ExecutableMethod;
 import io.micronaut.inject.annotation.EvaluatedAnnotationMetadata;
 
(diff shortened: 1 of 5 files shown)


inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
```diff
@@ -64,6 +65,7 @@
     private final Argument<?>[] constructorArguments;
     private final List<BeanProperty<B, Object>> beanProperties;
```
dstepanov: It might be a good idea to convert this two collections to arrays

dstepanov (approved):

graemerocher (approved):

inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
dstepanov: Added

PR #9141, opened 2023-04-21
Correct body binder

dstepanov:


graemerocher:
is this still needed?

PR #9149, opened 2023-04-23
Add postfix to created dummy file to solve conflicts between VisitorContexts.

altro3:


altro3:
@graemerocher @dstepanov Could you include this fix to release 3.9.1?

PR #9150, opened 2023-04-24
Update jackson to 2.15.0

yawkat:


yawkat:
@melix can you take a look at the failure? i removed the managed-jackson-databind version and use managed-jackson again, and I think that's the issue. But it should be safe.

melix:
The failure is expected. If a user for some reason uses `mn.versions.jackson.databind` in their build script, then this change would break it. I would recommend to keep both versions since it's a 3.x fix and we're not supposed to break things. It's an ok change for `master` though.

yawkat:
ok thanks

sdelamo (approved):

yawkat:
@sdelamo can you merge this up to 4.x soonish, i want to make some optimizations using 2.15 changes

PR #9165, opened 2023-04-27
Do not avoid processing enum methods

andriy-dmytruk:
`TypeElementVisitors` currently skips visiting enum methods and I am not sure that this behavior is correct.

Enum methods may need visiting, just like bean methods, especially if annotated with `@Creator` and therefore present in the introspection.

For example, having this enum,
```java
@Serdeable
public enum TestStateEnum implements BmcEnum {
    Active,
    Inactive,
    Deleted,
    UnknownEnumValue;

    ....

    @JsonCreator
    public static TestStateEnum create(String value) {
        return valueMap.getOrDefault(value, UnknownEnumValue);
    }
}
```
micronaut serde will deserialize the class using the `create` method. If I want to make its parameter `@Nullable` with a type visitor, I could achieve this in the `visitClass` method with
```java
element.getMethods().stream()
      .filter(m -> m.getName().equals(OCI_SDK_ENUM_CREATOR_NAME))
      .findAny()
      .ifPresent(e -> e.getParameters()[0].annotate(Nullable.class));
```
but it seems more logical for the method to be processed in the `visitMethod` method.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/TypeElementVisitorProcessor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/TypeElementVisitorProcessor.java
@@ -511,9 +511,6 @@ public class TypeElementVisitorProcessor extends AbstractInjectAnnotationProcess
                 executableElement,
                 javaVisitorContext.getElementAnnotationMetadataFactory()
             );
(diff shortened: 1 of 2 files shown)


dstepanov:
Please add a test

dstepanov (approved):

Issue #9161, opened 2023-04-27
micronaut.server.context-path not displaying in terminal


haaah:
### Expected Behavior

```
micronaut:
  server:
    context-path: /document
    port: 8082

```
The expected display url within the terminal should be 
`http://localhost:8082/document`

### Actual Behaviour

The terminal shows only
`10:47:18.696 [main] INFO  io.micronaut.runtime.Micronaut - Startup completed in 2513ms. Server Running: http://localhost:8082`

The context path is working as intended when we access http://localhost:8082/document, this bug affect only the terminal display path

### Environment Information

Mac BigSur
Azul Zulu 18.0.1

### Version

3.9.0

aditya-32:
hi i believe this issue is fixed can wen please close this  , as it is still open?
@graemerocher

PR #9177, opened 2023-04-28
Fixed the startup url print

filip-filipovic:
Fixed the print of the url at the startup of the application to include the context path. Issue [#9161](https://github.com/micronaut-projects/micronaut-core/issues/9161)

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/server/EmbeddedServer.java
+++ b/context/src/main/java/io/micronaut/runtime/server/EmbeddedServer.java
@@ -50,6 +50,11 @@ public interface EmbeddedServer extends EmbeddedApplication<EmbeddedServer> {
     URL getURL();
 
     /**
+     * @return The full URL to the server with the context path
+     */
+    URL getURLWithContextPath();
+
+    /**
      * @return The full URI to the server
      */
     URI getURI();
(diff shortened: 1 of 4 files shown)


context/src/main/java/io/micronaut/runtime/server/EmbeddedServer.java
```diff
@@ -49,6 +49,11 @@ public interface EmbeddedServer extends EmbeddedApplication<EmbeddedServer> {
      */
     URL getURL();
 
+    /**
+     * @return The full URL to the server with the context path
+     */
+    URL getURLWithContextPath();
```
graemerocher: introducing a new abstraction method in a patch release is a breaking change. Rename this method to `getContextURI()`, you should return URI instead of URL and make it a default method that looks like this:

```
default URI getContextURI() {
    return getURI();
}
```

graemerocher: add `@since 3.9.1`

graemerocher (changes requested):
Please add a test for this new functionality.

context/src/main/java/io/micronaut/runtime/server/EmbeddedServer.java
filip-filipovic: Fixed

filip-filipovic: Fixed

filip-filipovic:
Just need to add tests

filip-filipovic:
Added tests

graemerocher (approved):

Issue #8411, opened 2022-11-23
@Retryable annotation should be abble to provide a post failure callback


elielsonms:
### Feature description

I think it would be helpful if either the DefaultRetryInterceptor class sent an Event or the @Retryable annotation provided an attribute with a Runnable when the last attempt of retry fails and the exception is rethrown. I've been struggling to achieve this to be able to send a Notification to my team everytime a method that annotated with a Retry is unscessfull in which indicates that we have to do an manual intervention.

oscar-garzon:
Hi! I'm doing my first contributions to open source , can I take this issue? 

If yes, which option is better to implement: that DefaultRetryInterceptor sent an Event or that @retryable annotation provides a Runnable attribute?

oscar-garzon:
The approach I'm thinking  for this issue is to add a member to the Retryable annotation that its value will be the callback function that's call after the fail attempts. Then, modify the intercept method in DefaultRetryInterceptor to make the call in the required case. 

Is this a good approach?
 I'm having some trouble to fully understand the intercept method, could you provide some insight about it?

graemerocher:
Triage: this is a valid retry observability enhancement. Prefer publishing a retry-exhausted event from DefaultRetryInterceptor over adding a Runnable member to @Retryable, because an annotation callback is difficult to configure, test, and keep non-blocking. The event should include the method/context, final exception, attempt count, and an opt-in listener path so notification work does not run on the intercepted call thread.

PR #9183, opened 2023-04-29
added test to verify Retryable new member: runnable

oscar-garzon:
This Draft PR aims to help issue #8411. It's the test to verify the not yet implemented enhancement to pass a callback function to Retryable when all attempts failed. I'd like to implement the feature but would really appreciate some guidance.

A test is implemented to check runnable, the proposed new member of Retryable.
Part of line 98 of Returnable.java:
```
<? extends Object>
```
is dummy code. I'm not sure which class should be extended.


dstepanov:
This should be implemented as a method callback not a runnable callback, so you would reference the method name.
Check `RecoveryInterceptor`, `findFallbackMethod`, and how it's invoked.

PR #9185, opened 2023-04-30
Reworked ability to disable initializing Slf4J

altro3:
Unfortunately, the previous solution didn't work because the static fields were still initialized regardless of usage.

This solution is much more efficient and works 100% because I tested it several times :). Now, if necessary, the logEnabled flag is set in the constructor and the logger is not static, but is created through the factory only if necessary. Otherwise, a dummy logger is created.

Thus, the code remains as it was and we have the opportunity to turn off the initialization of slf4j in micronaut-openapi.

This solution also corrects the names of loggers when inheriting. Now the child and parent methods use a single logger with the name of the child class (this cannot be done when you have a static logger)

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -35,7 +35,10 @@ import io.micronaut.core.util.StringUtils;
 import io.micronaut.core.value.MapPropertyResolver;
 import io.micronaut.core.value.PropertyResolver;
 import io.micronaut.core.value.ValueException;
+
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.helpers.NOPLogger;
 
 import java.security.SecureRandom;
 import java.util.ArrayList;
@@ -66,8 +69,6 @@ import java.util.stream.Collectors;
  */
 public class PropertySourcePropertyResolver implements PropertyResolver, AutoCloseable {
 
-    private static final Logger LOG = ClassUtils.getLogger(PropertySourcePropertyResolver.class);
-
     private static final EnvironmentProperties CURRENT_ENV = StaticOptimizations.get(EnvironmentProperties.class)
             .orElseGet(EnvironmentProperties::empty);
     private static final Pattern DOT_PATTERN = Pattern.compile("\\.");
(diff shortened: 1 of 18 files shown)


graemerocher:
The PR includes several breaking API changes which will have to be corrected first plus the tests are failing

altro3:
@graemerocher Need to return methods `isLogEnabled()` and `setLogEnabled(boolean logEnabled)` ?

altro3:
@graemerocher tests fixed. Also fixed several checkstyle errors, doclint errors and fixed 2 tests for windows

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -101,8 +99,12 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
      * Creates a new, initially empty, {@link PropertySourcePropertyResolver} for the given {@link ConversionService}.
      *
      * @param conversionService The {@link ConversionService}
+     * @param logEnabled logEnabled flag to enable or disable logger
      */
-    public PropertySourcePropertyResolver(ConversionService<?> conversionService) {
+    public PropertySourcePropertyResolver(ConversionService<?> conversionService, boolean logEnabled) {
```
graemerocher, line 104: this is a breaking change, need to add a constructor with `PropertySourcePropertyResolver(ConversionService)`

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -909,28 +907,6 @@ public void close() throws Exception {
...
-     * Setter for logEnabled.
-     *
-     * @param logEnabled is log enabled
-     *
-     * @since 3.9.0
-     */
-    public void setLogEnabled(boolean logEnabled) {
-        this.logEnabled = logEnabled;
-    }
-
```
graemerocher, line 933: removing these methods is a breaking change

graemerocher (changes requested):
Also needs test coverage for the case this is trying to fix

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
altro3: Sorry, missed it. Fixed.

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
altro3: Reverted these methods with mark Deprecated

graemerocher (approved):

sdelamo (approved):

PR #9189, opened 2023-05-04
graal: remove static final Logger from  DefaultHttpFactories

sdelamo:
see: https://github.com/micronaut-projects/micronaut-starter/pull/1773

I assume this PR fixes: 

```
 org.slf4j.LoggerFactory was unintentionally initialized at build time. To see why org.slf4j.LoggerFactory got initialized use --trace-class-initialization=org.slf4j.LoggerFactory
io.micronaut.http.DefaultHttpFactories was unintentionally initialized at build time. To see why io.micronaut.http.DefaultHttpFactories got initialized use --trace-class-initialization=io.micronaut.http.DefaultHttpFactories
To see how the classes got initialized, use --trace-class-initialization=org.slf4j.LoggerFactory,io.micronaut.http.DefaultHttpFactories```
```
```
asses that should be initialized at run time got initialized during image building:
 io.micronaut.http.DefaultHttpFactories was unintentionally initialized at build time. io.micronaut.http.HttpResponseFactory caused initialization of this class with the following trace:
        at io.micronaut.http.DefaultHttpFactories.<clinit>(DefaultHttpFactories.java:37)
        at io.micronaut.http.HttpResponseFactory.<clinit>(HttpResponseFactory.java:29)

``

### DIFF
--- a/http/src/main/java/io/micronaut/http/DefaultHttpFactories.java
+++ b/http/src/main/java/io/micronaut/http/DefaultHttpFactories.java
@@ -34,8 +34,6 @@ import java.util.Optional;
 @Internal
 class DefaultHttpFactories {
 
-    private static final Logger LOG = LoggerFactory.getLogger(DefaultHttpFactories.class);
-
     /**
      * Resolves the default request factory.
      *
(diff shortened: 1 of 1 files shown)


sdelamo:
No idea how to test this so that we catch it in core and not downstream.

graemerocher:
This is not the correct fix. What has to be understood is why this type is being initialized in the first place. There must be some initialize at build time flags somewhere

sdelamo:
@graemerocher  this: https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-api-proxy/src/main/resources/META-INF/native-image/io.micronaut.aws/micronaut-function-aws-api-proxy/native-image.properties should we remove that file?

PR #9201, opened 2023-05-09
Support MessageBodyWriter/Reader in the HTTP client

yawkat:
This implements the reader/writer API for the client. There's also some server changes that I missed.

Some changes from MediaTypeCodec to MessageBodyHandler are still missing, such as for form upload conversions.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1358,6 +1358,7 @@ public class DefaultHttpClient implements
                 ByteBuf bodyContent = null;
                 if (hasBody) {
                     Object bodyValue = body.get();
+                    DynamicWriter dynamicWriter = new DynamicWriter(handlerRegistry, List.of(requestContentType));
                     if (Publishers.isConvertibleToPublisher(bodyValue)) {
                         boolean isSingle = Publishers.isSingle(bodyValue.getClass());
 
@@ -1365,9 +1366,8 @@ public class DefaultHttpClient implements
                                 new IllegalArgumentException("Unconvertible reactive type: " + bodyValue)
                         );
 
-                        DynamicWriter dynamicWriter = new DynamicWriter(handlerRegistry, List.of(requestContentType));
                         Flux<HttpContent> requestBodyPublisher = Flux.from(publisher).map(o -> {
-                            ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, o, request.getHeaders(), NettyByteBufferFactory.DEFAULT);
+                            ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, o, request.getHeaders(), byteBufferFactory);
                             return new DefaultHttpContent(((ByteBuf) buffer.asNativeBuffer()));
                         });
 
@@ -1383,15 +1383,8 @@ public class DefaultHttpClient implements
                     } else if (bodyValue instanceof CharSequence) {
                         bodyContent = charSequenceToByteBuf((CharSequence) bodyValue, requestContentType);
                     } else if (mediaTypeCodecRegistry != null) {
-                        Optional<MediaTypeCodec> registeredCodec = mediaTypeCodecRegistry.findCodec(requestContentType);
-                        bodyContent = registeredCodec.map(codec -> {
-                                    if (bodyType != null && bodyType.isInstance(bodyValue)) {
-                                        return codec.encode((Argument<Object>) bodyType, bodyValue, byteBufferFactory).asNativeBuffer();
-                                    } else {
-                                        return codec.encode(bodyValue, byteBufferFactory).asNativeBuffer();
-                                    }
-                                })
-                                .orElse(null);
+                        ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, bodyValue, request.getHeaders(), byteBufferFactory);
+                        bodyContent = (ByteBuf) buffer.asNativeBuffer();
                     }
                     if (bodyContent == null) {
                         bodyContent = conversionService.convert(bodyValue, ByteBuf.class).orElseThrow(() ->
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -49,7 +49,7 @@ import io.micronaut.http.bind.DefaultRequestBinderRegistry;
 import io.micronaut.http.bind.RequestBinderRegistry;
 import io.micronaut.http.body.ChunkedMessageBodyReader;
 import io.micronaut.http.body.ContextlessMessageBodyHandlerRegistry;
-import io.micronaut.http.body.DynamicWriter;
+import io.micronaut.http.body.DynamicMessageBodyWriter;
 import io.micronaut.http.body.MessageBodyHandlerRegistry;
 import io.micronaut.http.body.MessageBodyReader;
 import io.micronaut.http.client.BlockingHttpClient;
@@ -90,7 +90,7 @@ import io.micronaut.http.multipart.MultipartException;
 import io.micronaut.http.netty.NettyHttpHeaders;
 import io.micronaut.http.netty.NettyHttpRequestBuilder;
 import io.micronaut.http.netty.NettyHttpResponseBuilder;
-import io.micronaut.http.netty.body.ByteBufRawHandler;
+import io.micronaut.http.netty.body.ByteBufRawMessageBodyHandler;
 import io.micronaut.http.netty.body.NettyJsonHandler;
 import io.micronaut.http.netty.body.NettyJsonStreamHandler;
 import io.micronaut.http.netty.channel.ChannelPipelineCustomizer;
@@ -517,11 +517,22 @@ public class DefaultHttpClient implements
         }
     }
 
+    /**
+     * Get the handler registry for this client.
+     *
+     * @return The handler registry
+     */
+    @NonNull
     public final MessageBodyHandlerRegistry getHandlerRegistry() {
         return handlerRegistry;
     }
 
-    public final void setHandlerRegistry(MessageBodyHandlerRegistry handlerRegistry) {
+    /**
+     * Set the handler registry for this client.
+     *
+     * @param handlerRegistry The handler registry
+     */
+    public final void setHandlerRegistry(@NonNull MessageBodyHandlerRegistry handlerRegistry) {
         this.handlerRegistry = handlerRegistry;
     }
 
@@ -1328,7 +1339,7 @@ public class DefaultHttpClient implements
                 ByteBuf bodyContent = null;
                 if (hasBody) {
                     Object bodyValue = body.get();
-                    DynamicWriter dynamicWriter = new DynamicWriter(handlerRegistry, List.of(requestContentType));
+                    DynamicMessageBodyWriter dynamicWriter = new DynamicMessageBodyWriter(handlerRegistry, List.of(requestContentType));
                     if (Publishers.isConvertibleToPublisher(bodyValue)) {
                         boolean isSingle = Publishers.isSingle(bodyValue.getClass());
 
@@ -1791,7 +1802,7 @@ public class DefaultHttpClient implements
     }
 
     private static MessageBodyHandlerRegistry createDefaultMessageBodyHandlerRegistry() {
-        ContextlessMessageBodyHandlerRegistry registry = new ContextlessMessageBodyHandlerRegistry(new ApplicationConfiguration(), NettyByteBufferFactory.DEFAULT, new ByteBufRawHandler());
+        ContextlessMessageBodyHandlerRegistry registry = new ContextlessMessageBodyHandlerRegistry(new ApplicationConfiguration(), NettyByteBufferFactory.DEFAULT, new ByteBufRawMessageBodyHandler());
         JsonMapper mapper = JsonMapper.createDefault();
         registry.add(MediaType.APPLICATION_JSON_TYPE, new NettyJsonHandler<>(mapper));
         registry.add(MediaType.APPLICATION_JSON_STREAM_TYPE, new NettyJsonStreamHandler<>(mapper));
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -47,9 +47,11 @@ import io.micronaut.http.MutableHttpRequest;
 import io.micronaut.http.MutableHttpResponse;
 import io.micronaut.http.bind.DefaultRequestBinderRegistry;
 import io.micronaut.http.bind.RequestBinderRegistry;
+import io.micronaut.http.body.ChunkedMessageBodyReader;
 import io.micronaut.http.body.ContextlessMessageBodyHandlerRegistry;
 import io.micronaut.http.body.DynamicWriter;
 import io.micronaut.http.body.MessageBodyHandlerRegistry;
+import io.micronaut.http.body.MessageBodyReader;
 import io.micronaut.http.client.BlockingHttpClient;
 import io.micronaut.http.client.DefaultHttpClientConfiguration;
 import io.micronaut.http.client.HttpClient;
@@ -75,7 +77,6 @@ import io.micronaut.http.client.netty.ssl.NettyClientSslBuilder;
 import io.micronaut.http.client.netty.websocket.NettyWebSocketClientHandler;
 import io.micronaut.http.client.sse.SseClient;
 import io.micronaut.http.codec.CodecException;
-import io.micronaut.http.codec.MediaTypeCodec;
 import io.micronaut.http.codec.MediaTypeCodecRegistry;
 import io.micronaut.http.context.ContextPathUtils;
 import io.micronaut.http.context.ServerRequestContext;
@@ -106,8 +107,6 @@ import io.micronaut.http.util.HttpHeadersUtil;
 import io.micronaut.json.JsonMapper;
 import io.micronaut.json.codec.JsonMediaTypeCodec;
 import io.micronaut.json.codec.JsonStreamMediaTypeCodec;
-import io.micronaut.json.codec.MapperMediaTypeCodec;
-import io.micronaut.json.tree.JsonNode;
 import io.micronaut.runtime.ApplicationConfiguration;
 import io.micronaut.scheduling.instrument.InvocationInstrumenter;
 import io.micronaut.scheduling.instrument.InvocationInstrumenterFactory;
@@ -118,7 +117,6 @@ import io.micronaut.websocket.context.WebSocketBean;
 import io.micronaut.websocket.context.WebSocketBeanRegistry;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.EmptyByteBuf;
 import io.netty.buffer.Unpooled;
@@ -167,7 +165,6 @@ import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.DefaultThreadFactory;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
-import org.reactivestreams.Processor;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -725,22 +722,12 @@ public class DefaultHttpClient implements
     @Override
     public <I, B> Publisher<Event<B>> eventStream(@NonNull io.micronaut.http.HttpRequest<I> request, @NonNull Argument<B> eventType, @NonNull Argument<?> errorType) {
         setupConversionService(request);
+        MessageBodyReader<B> reader = handlerRegistry.findReader(eventType, List.of(MediaType.APPLICATION_JSON_TYPE)).orElseThrow(() -> new CodecException("JSON codec not present"));
         return Flux.from(eventStreamOrError(request, errorType)).map(byteBufferEvent -> {
             ByteBuffer<?> data = byteBufferEvent.getData();
-            Optional<MediaTypeCodec> registeredCodec;
 
-            if (mediaTypeCodecRegistry != null) {
-                registeredCodec = mediaTypeCodecRegistry.findCodec(MediaType.APPLICATION_JSON_TYPE);
-            } else {
-                registeredCodec = Optional.empty();
-            }
-
-            if (registeredCodec.isPresent()) {
-                B decoded = registeredCodec.get().decode(eventType, data);
-                return Event.of(byteBufferEvent, decoded);
-            } else {
-                throw new CodecException("JSON codec not present");
-            }
+            B decoded = reader.read(eventType, MediaType.APPLICATION_JSON_TYPE, request.getHeaders(), data);
+            return Event.of(byteBufferEvent, decoded);
         });
     }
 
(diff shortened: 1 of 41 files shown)


http-client/src/test/groovy/io/micronaut/http/client/ServerRedirectSpec.groovy
```diff
@@ -326,7 +332,7 @@ class ServerRedirectSpec extends Specification {
         @Produces([MediaType.TEXT_EVENT_STREAM, MediaType.APPLICATION_JSON_STREAM])
         HttpResponse<Flux<?>> title(final HttpHeaders headers) {
             if (headers.accept().contains(MediaType.TEXT_EVENT_STREAM_TYPE)) {
-                return HttpResponse.ok(Flux.just("The Stand")).contentType(MediaType.TEXT_EVENT_STREAM)
+                return HttpResponse.ok(Flux.just(Event.of("The Stand"))).contentType(MediaType.TEXT_EVENT_STREAM)
```
dstepanov, line 335: Is it required now?

http-client/src/test/groovy/io/micronaut/http/client/filter/MutateRequestClientFilterSpec.groovy
```diff
@@ -55,7 +55,7 @@ class MutateRequestClientFilterSpec extends Specification {
 
     void "test mutate stream request URI"() {
         expect:
-        myClient.stream().blockFirst() == "xxxxxxxxxxx"
+        myClient.stream().collectList().block().join("") == '["xxxxxxxxxxx"]'
```
dstepanov, line 58: Why is this change?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -307,30 +298,20 @@ private void encodeHttpResponse(
...
-                NettyBodyWriter<Object> closure = wrap(messageBodyWriter);
-                handleMissingConnectionHeader(response, nettyRequest, outboundAccess);
-                if (closure.isBlocking()) {
-                    MediaType finalResponseMediaType = responseMediaType;
-                    getIoExecutor().execute(() -> writeNettyMessageBody(nettyRequest.getChannelHandlerContext(), nettyRequest, (MutableHttpResponse<Object>) response, responseBodyType, finalResponseMediaType, body, closure, outboundAccess));
-                } else {
-                    writeNettyMessageBody(nettyRequest.getChannelHandlerContext(), nettyRequest, (MutableHttpResponse<Object>) response, responseBodyType, responseMediaType, body, closure, outboundAccess);
-                }
+            Argument<Object> actualResponseType;
+            if (messageBodyWriter == null || !responseBodyType.isInstance(body) || !messageBodyWriter.isWriteable(responseBodyType, responseMediaType)) {
```
dstepanov, line 302: Maybe `messageBodyWriter` can be reused in `RouteInfo`?

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -17,205 +17,101 @@
...
-                        writer = beanLocator.getBean(exactMatch.iterator().next());
+    protected <T> MessageBodyWriter<T> findWriterImpl(Argument<T> type, List<MediaType> mediaTypes) {
+        Collection<BeanDefinition<MessageBodyWriter>> beanDefinitions = beanLocator.getBeanDefinitions(Argument.of(MessageBodyWriter.class, type), new MediaTypeQualifier<>(type, mediaTypes, Produces.class));
+        if (beanDefinitions.size() == 1) {
+            return beanLocator.getBean(beanDefinitions.iterator().next());
+        } else {
+            List<BeanDefinition<MessageBodyWriter>> exactMatch = beanDefinitions.stream()
+                .filter(d -> {
+                    List<Argument<?>> typeArguments = d.getTypeArguments(MessageBodyWriter.class);
+                    if (typeArguments.isEmpty()) {
```
dstepanov, line 100: @graemerocher Do we have some qualifier from `Qualifiers.` that can do this match? I think the correct way would be `Qualifiers.byQualifiers(new MediaTypeQualifier(...), byTypeArgumentMatching(type))`

http/src/main/java/io/micronaut/http/body/DynamicWriter.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public MessageBodyWriter<Object> createSpecific(Argument<Object> type) {
+        return registry.findWriter(type, mediaTypes).orElse(this);
+    }
+
+    public MessageBodyWriter<Object> find(Argument<Object> type, MediaType mediaType, Object object) {
+        Optional<MessageBodyWriter<Object>> specific = registry.findWriter(type, List.of(mediaType));
+        if (specific.isPresent() && !(specific.get() instanceof DynamicWriter)) {
```
dstepanov: If the writer is shared maybe we can store `lastFound` and `lastArgument` if it matches, return it, avoiding the writer.

http/src/main/java/io/micronaut/http/body/RawMessageBodyHandlerRegistry.java
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    protected abstract <T> MessageBodyReader<T> findReaderImpl(Argument<T> type, List<MediaType> mediaTypes);
+
+    @SuppressWarnings({"unchecked"})
+    @Override
+    public <T> Optional<MessageBodyReader<T>> findReader(Argument<T> type, List<MediaType> mediaTypes) {
+        HandlerKey<T> key = new HandlerKey<>(type, mediaTypes);
+        MessageBodyReader<?> messageBodyReader = readers.get(key);
+        if (messageBodyReader == null) {
```
dstepanov, line 92: Would it make sense to have two maps:
- 1. All readers/writers contributed at the startup - simple hashmap - fast, the most common scenario
- 2. Precalculated readers/writers - requires a concurrent map

The minus is that we would need to access two maps in the worse scenario.
If the readers/writers are cached in the route info, this is unnecessary.

http/src/main/java/io/micronaut/http/body/RawMessageBodyHandlerRegistry.java
```diff
@@ -0,0 +1,352 @@
...
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            HandlerKey<?> that = (HandlerKey<?>) o;
+            return type.equalsType(that.type) && mediaTypes.equals(that.mediaTypes);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(type.typeHashCode(), mediaTypes);
```
dstepanov: You can use `io.micronaut.core.util.ObjectUtils` that avoids new array allocation. You might want to store the hash code.

http/src/main/java/io/micronaut/http/body/TextPlainHandler.java
```diff
@@ -44,26 +44,26 @@
 @Consumes(MediaType.TEXT_PLAIN)
 @Singleton
 @Experimental
-public final class TextPlainHandler implements MessageBodyHandler<String> {
+public final class TextPlainHandler implements MessageBodyHandler<CharSequence> {
```
dstepanov, line 47: We might want to have an additional `MessageBodyHandler<String>` so the handler is found as a direct match.

dstepanov (changes requested):

dstepanov:
Can you please add comments in all modified tests if it's a breaking change or a test improvement?

http-client/src/test/groovy/io/micronaut/http/client/ServerRedirectSpec.groovy
yawkat: String is a raw type, the new implementation thinks it's already serialized. It's more consistent with other implementations (e.g. json) so I think it's fine.

http-client/src/test/groovy/io/micronaut/http/client/filter/MutateRequestClientFilterSpec.groovy
yawkat: bc String is a raw type, it doesn't deserialize it from array anymore. the controller already added the quotes manually so the test case was a bit weird in the first place

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: it should be in most cases. dynamic writer is used only as fallback, eg when theres an error where the route writer does not match the type of the actual response body.

http/src/main/java/io/micronaut/http/body/RawMessageBodyHandlerRegistry.java
yawkat: the problem is that the matching is kind of fuzzy. for example the json mapper matches for any body type, and is cached separately for each type. so adding it for the base type (Object) is basically pointless.

The map is kind of like the cache in ConversionService.

http-client/src/test/groovy/io/micronaut/http/client/HttpGetSpec.groovy
```diff
@@ -311,7 +311,7 @@ class HttpGetSpec extends Specification {
         String body = res.getBody(String).orElse(null)
 
         then:
-        body == null
```
yawkat, line 314: behavior change, but not a breaking change. body is now available forever.

yawkat: i gave up on lifecycle management, but it's not worse than before, because the body was stored as byte[] before anyway :)

http-client/src/test/groovy/io/micronaut/http/client/JsonStreamSpec.groovy
```diff
@@ -209,7 +209,6 @@ class JsonStreamSpec  extends Specification {
         Publisher<Book> list();
 
         @Post(uri = "/count", processes = MediaType.APPLICATION_JSON_STREAM)
-        @SingleResult
         Publisher<LibraryStats> count(@Body Flux<Book> theBooks)
```
yawkat, line 212: breaking change, APPLICATION_JSON_STREAM doesn't make sense for parsing to single result. it gives an error now.

http-client/src/test/groovy/io/micronaut/http/client/StreamRequestSpec.groovy
```diff
@@ -273,7 +274,7 @@ class StreamRequestSpec extends Specification {
 
         @Post("/pojo-flowable-error")
         Publisher<Book> pojoReactiveSequenceError(@Header MediaType contentType, @Body Publisher<Book> books) {
-            return Flux.from(books).flatMap({ Book book ->
```
yawkat, line 276: no change, concatMap is better but it makes no difference here

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/context/ServerRequestContextSpec.groovy
```diff
@@ -109,30 +113,30 @@ class ServerRequestContextSpec extends Specification {
         @Get("/method")
         String method() {
             def request = ServerRequestContext.currentRequest().orElseThrow { -> new RuntimeException("no request") }
-            request.uri
```
yawkat, line 112: technically a breaking change, but this was a test bug. the methods declare return type String but actually returned URI.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/threading/ThreadSelectionSpec.groovy
```diff
@@ -163,7 +163,7 @@ class ThreadSelectionSpec extends Specification {
         String alterflowable()
 
         @Post(uri = "/alterflowablePost", produces = MediaType.TEXT_PLAIN)
-        String alterflowablePost(String body)
```
yawkat, line 166: breaking change

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ResponseStatusTest.java
```diff
@@ -48,7 +48,7 @@ public class ResponseStatusTest {
     @Test
     void testConstraintViolationCauses400() throws IOException {
         asserts(SPEC_NAME,
-            HttpRequest.POST("/response-status/constraint-violation", Collections.emptyMap()).header(HttpHeaders.CONTENT_TYPE, MediaType.TEXT_PLAIN),
```
yawkat, line 51: breaking change. the text/plain codec cannot serialize a map. previously it would fall back to json i think, now it will fail. the body is unused in the test anyway.

test-suite-groovy/src/test/groovy/io/micronaut/docs/server/consumes/ConsumesControllerSpec.groovy
```diff
@@ -76,7 +77,7 @@ class ConsumesControllerSpec extends Specification {
                         .contentType(MediaType.TEXT_PLAIN))
 
         then:
-        noExceptionThrown()
```
yawkat, line 79: breaking change, can't write arbitrary type (Book) as text/plain anymore

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
graemerocher: yes you can have a composite qualifier, this could work

test-suite/src/test/groovy/io/micronaut/docs/http/client/proxy/ProxyRequestSpec.groovy
```diff
@@ -102,7 +102,7 @@ class ProxyRequestSpec extends Specification {
         response.body().text == 'Hello John XXX'
 
         when:"A POST request is sent with an invalid content type"
-        response = client.exchange(HttpRequest.POST("/proxy/post/json", new Message(text: "John")).contentType(MediaType.TEXT_PLAIN), Message).blockFirst()
```
yawkat, line 105: breaking change, can't write arbitrary type (Book) as text/plain anymore

test-suite/src/test/java/io/micronaut/docs/function/client/aws/LocalFunctionInvokeSpec.java
```diff
@@ -80,10 +79,11 @@ interface MathClient {
 
         //tag::functionRnd[]
         @Named("round")
-        int rnd(float value);
```
yawkat, line 83: this change used to be necessary, don't think it is anymore. float and other primitive types can be written to text/plain still, for compat reasons.

test-suite/src/test/groovy/io/micronaut/docs/http/client/proxy/ProxyRequestSpec.groovy
graemerocher: can you explain why?

http-client/src/test/groovy/io/micronaut/http/client/filter/MutateRequestClientFilterSpec.groovy
yawkat: to explain more:

The controller returns a Flux<String>, composed of a single string "xxx" including the quotes. The http stack does not serialize the string as json because it's a raw type (serializing it would yield double quoting "\"xxx\""). However it does add the brackets because it's not byte[] or ByteBuffer (yes it's weird, but this is the most compatible behavior). This hasn't changed.

On the client side, it used to split up the input and parse the string fully as json. This has changed, it treats it as a raw type because it's String, so it just yields the unparsed body now.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -498,12 +510,21 @@ public MediaTypeCodecRegistry getMediaTypeCodecRegistry() {
...
     }
 
+    public final MessageBodyHandlerRegistry getHandlerRegistry() {
+        return handlerRegistry;
+    }
+
+    public final void setHandlerRegistry(MessageBodyHandlerRegistry handlerRegistry) {
+        this.handlerRegistry = handlerRegistry;
+    }
+
```
graemerocher: javadoc / nullability annotations

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
```diff
@@ -156,65 +155,23 @@ public <T> Optional<T> getBody(Class<T> type) {
         return getBody(Argument.of(type));
     }
 
-    /**
-     * @return The Netty native response object
-     */
-    public FullHttpResponse getNativeResponse() {
-        return nettyHttpResponse;
-    }
```
graemerocher, line 164: Why this change?

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
```diff
@@ -264,20 +217,9 @@ private <T> Optional convertByteBuf(ByteBuf content, Argument<T> type) {
...
-                    return Optional.of(bodyBytes);
-                } else {
-                    Optional<MediaTypeCodec> foundCodec = mediaTypeCodecRegistry.findCodec(contentType.get());
-                    if (foundCodec.isPresent()) {
-                        MediaTypeCodec codec = foundCodec.get();
-                        return Optional.of(codec.decode(type, bodyBytes));
-                    }
-                }
+            Optional<MessageBodyReader<T>> reader = handlerRegistry.findReader(type, List.of(contentType.get()));
+            if (reader.isPresent() && reader.get().isReadable(type, contentType.get())) {
```
graemerocher: split into 2 `if` blocks to avoid the multiple calls to `get` also a extract a variable for `contentType` to avoid multiple calls

http-netty/src/main/java/io/micronaut/http/netty/body/ByteBufRawHandler.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @since 4.0.0
+ * @author Jonas Konrad
+ */
+@Internal
+@Singleton
+@Experimental
+@BootstrapContextCompatible
+@Bean(typed = RawMessageBodyHandler.class)
+public class ByteBufRawHandler implements RawMessageBodyHandler<ByteBuf> {
```
graemerocher: can be `final`. Rename to `ByteBufRawMessageBodyHandler`

http-netty/src/main/java/io/micronaut/http/netty/body/NettyJsonStreamHandler.java
```diff
@@ -33,23 +36,29 @@
...
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.List;
 
 @Internal
 @Singleton
 @Produces(MediaType.APPLICATION_JSON_STREAM)
 @Consumes(MediaType.APPLICATION_JSON_STREAM)
-class NettyJsonStreamHandler<T> implements MessageBodyReader<T>, ChunkedMessageBodyReader<T> {
+public class NettyJsonStreamHandler<T> implements MessageBodyHandler<T>, ChunkedMessageBodyReader<T>, CustomizableNettyJsonHandler {
```
graemerocher: can be `final`?

http/src/main/java/io/micronaut/http/body/ContextlessMessageBodyHandlerRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package io.micronaut.http.body;
+
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.io.buffer.ByteBufferFactory;
+import io.micronaut.core.type.Argument;
+import io.micronaut.http.MediaType;
+import io.micronaut.runtime.ApplicationConfiguration;
+import org.jetbrains.annotations.Nullable;
```
graemerocher: use Micronaut nullable

http/src/main/java/io/micronaut/http/body/ContextlessMessageBodyHandlerRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * {@link MessageBodyHandlerRegistry} implementation that does not need an application context.
+ *
+ * @author Jonas Konrad
+ * @since 4.0.0
+ */
+@Internal
+@Experimental
+public final class ContextlessMessageBodyHandlerRegistry extends RawMessageBodyHandlerRegistry {
```
graemerocher, line 39: Since this is only used by the client can it be moved to the client module and made package private?

http/src/main/java/io/micronaut/http/body/ContextlessMessageBodyHandlerRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+        super(Stream.concat(Stream.of(new RawStringHandler(applicationConfiguration), new RawByteArrayHandler(), new RawByteBufferHandler(byteBufferFactory)), Stream.of(otherRawHandlers)).toList());
+    }
+
+    /**
+     * Add a {@link MessageBodyHandler} for the given media type.
+     *
+     * @param mediaType The media type the handler applies to
+     * @param handler   The handler
+     */
+    public void add(MediaType mediaType, MessageBodyHandler<?> handler) {
```
graemerocher: nullability annotations for the arguments

http/src/main/java/io/micronaut/http/body/DynamicWriter.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Placeholder {@link MessageBodyWriter} implementation that decides which writer to use based on
+ * the dynamic (runtime) type of the body. Used as fallback where the type is not known statically.
+ *
+ * @author Jonas Konrad
+ * @since 4.0.0
+ */
+@Internal
+public final class DynamicWriter implements MessageBodyWriter<Object> {
```
graemerocher: Rename to `DynamicMessageBodyWriter`

test-suite/src/test/groovy/io/micronaut/docs/http/client/proxy/ProxyRequestSpec.groovy
yawkat: TextPlainCodec used to just accept any object and call toString on it. TextPlainHandler only works with String, so it gives an error that there is no handler for text/plain for this type. i think it makes more sense.

http/src/main/java/io/micronaut/http/body/RawMessageBodyHandler.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Supported types of this raw body handler. Exact match is used for reading. For writing, the
+     * match is covariant. For example, if this returns {@code [String, CharSequence]}, then this
+     * raw handler will be used for reading types declared as exactly {@code String} or
+     * {@code CharSequence}, and will additionally be used for writing (but not reading) subtypes
+     * (e.g. {@code StringBuilder}).
+     *
+     * @return The supported types
+     */
+    Collection<? extends Class<?>> getTypes();
```
graemerocher, line 44: nullability annotations for return type

http/src/main/java/io/micronaut/http/body/TextStreamBodyHandler.java
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Handler for SSE events.
+ *
+ * @param <T> The type to write, may be {@link Event}
+ */
+@Internal
+@Singleton
+@Produces(MediaType.TEXT_EVENT_STREAM)
+@Consumes(MediaType.TEXT_EVENT_STREAM)
+public class TextStreamBodyHandler<T> implements MessageBodyWriter<T> {
```
graemerocher: since it is only a writer rename to `TextStreamBodyWriter` and make `final`

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
```diff
@@ -54,6 +54,10 @@ public JsonMessageHandler(JsonMapper jsonMapper) {
         this.jsonMapper = jsonMapper;
     }
 
+    public JsonMapper getJsonMapper() {
```
graemerocher, line 64: javadoc / nullability annotations

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
yawkat: it's unused

http/src/main/java/io/micronaut/http/body/ContextlessMessageBodyHandlerRegistry.java
yawkat: RawMessageBodyHandlerRegistry is package-private so moving ContextlessMessageBodyHandlerRegistry would need to make RawMessageBodyHandlerRegistry public

http-netty/src/main/java/io/micronaut/http/netty/body/ByteBufRawMessageBodyHandler.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @since 4.0.0
+ * @author Jonas Konrad
+ */
+@Internal
+@Singleton
+@Experimental
+@BootstrapContextCompatible
+@Bean(typed = RawMessageBodyHandler.class)
+public final class ByteBufRawMessageBodyHandler implements RawMessageBodyHandler<ByteBuf> {
```
graemerocher, line 55: does this need to be public?

http/src/main/java/io/micronaut/http/body/TextStreamBodyWriter.java
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Handler for SSE events.
+ *
+ * @param <T> The type to write, may be {@link Event}
+ */
+@Internal
+@Singleton
+@Produces(MediaType.TEXT_EVENT_STREAM)
+@Consumes(MediaType.TEXT_EVENT_STREAM)
+public final class TextStreamBodyWriter<T> implements MessageBodyWriter<T> {
```
graemerocher: does this need to be public?

graemerocher (changes requested):
Can you update `breaks.adoc` and document the breaking changes in this PR

http-netty/src/main/java/io/micronaut/http/netty/body/ByteBufRawMessageBodyHandler.java
yawkat: yes, it's used by the contextless http client

http/src/main/java/io/micronaut/http/body/DynamicWriter.java
yawkat: DynamicWriter is not shared. It used to be (the registry used to return it as fallback) but it didn't turn out to work so it's used directly now.

http/src/main/java/io/micronaut/http/body/RawMessageBodyHandlerRegistry.java
yawkat: will use ObjectUtils. storing the hash shouldn't matter, the map caches it anyway.

http/src/main/java/io/micronaut/http/body/TextPlainHandler.java
yawkat: don't think it matters since it's cached

graemerocher (approved):

dstepanov (approved):

PR #9217, opened 2023-05-11
Fix not supported error with type use annotations and expression

graemerocher:
TYPE_USE annotations can't be used with expression at the moment because they fail to locate the originating class name, not sure how to fix this so for the moment improve the error and document that it is not supported.


dstepanov (approved):
Why do we need the class annotation at the runtime?

graemerocher:
expressions generate classes and generated class needs to live in the same package as the originating class, but it seems with type use annotations there is no way to retrieve the originating type? Maybe we can fix this, I don't know.

PR #9218, opened 2023-05-11
Implement SSL config reloading for the netty http client

### LINKED ISSUE #7873: SSL Certs Reload with Refresh event not working in v3.6.0?
Outside of official docs, I am unable to find any references online for this config. So responding to the merge where the feature was added. 

@graemerocher Thank you so much for adding this. It would really help us out in our application. I am on micronaut v3.6.0, gradle 7.3, java 11 project. However, I don't seem to be able to get this config to work. 

Intent: I get new .p12 files everytime the certs are renewed. Without restarting the service, I would like to refresh the keymanagement system in the micronaut server to use the new .p12 file (public key value changes from the previous one in use). 

I have tried several versions and even tried passing in the full file path. It still seems to be using the file that was included in the original build. I am attaching screenshots of my config, attempts and logs. Am I missing anything? Could you please help? 

Original SSL config in application
image

Logging on refresh event
image

Trying full ssl.* reload
image

Confirming refresh event in logs
image

Application returning old value in property as response. Also tried new requests in a new browser window, different browser as well
image

Trying just path refresh for SSL. Event was successfully recorded in logs, but still not using new p12 file.
image

_Originally posted by @anuragdeshpande in https://github.com/micronaut-projects/micronaut-core/issues/6112#issuecomment-1213625246_

yawkat:
NettyHttpServer also apparently didn't reload when changing the micronaut.ssl configuration, so I fixed that too.

Likely fixes #7873


graemerocher:
can you resolve the conflicts?

graemerocher (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java
```diff
@@ -535,6 +545,21 @@ private static MapperMediaTypeCodec getJsonCodec(BeanContext beanContext) {
         return beanContext.getBean(MapperMediaTypeCodec.class, Qualifiers.byName(MapperMediaTypeCodec.REGULAR_JSON_MEDIA_TYPE_CODEC_NAME));
     }
 
+    @Override
+    public Set<String> getObservedConfigurationPrefixes() {
+        return Set.of(DefaultHttpClientConfiguration.PREFIX, ServiceHttpClientConfiguration.PREFIX, DefaultSslConfiguration.PREFIX);
```
timyates:
```suggestion
        return Set.of(DefaultHttpClientConfiguration.PREFIX, ServiceHttpClientConfiguration.PREFIX, SslConfiguration.PREFIX);
```

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
```diff
@@ -692,7 +693,7 @@ public void doOnConnect(@NonNull ChannelPipelineListener listener) {
 
     @Override
     public Set<String> getObservedConfigurationPrefixes() {
-        return Collections.singleton(HttpServerConfiguration.PREFIX);
+        return Set.of(HttpServerConfiguration.PREFIX, DefaultSslConfiguration.PREFIX);
```
timyates:
```suggestion
        return Set.of(HttpServerConfiguration.PREFIX, SslConfiguration.PREFIX);
```

timyates (approved):
2 suggestions to appease Sonar a bit

PR #9238, opened 2023-05-16
Fix for Kotlin test failures after validation upgrade

graemerocher:


graemerocher:
@dstepanov the failure seems to be a regression or change in behaviour in the validation module with KSP. This type is no longer validated:

```
@ConfigurationProperties("my.engine") // <1>
interface EngineConfig {

    @get:Bindable(defaultValue = "Ford") // <2>
    @get:NotBlank // <3>
    val manufacturer: String

    @get:Min(1L)
    val cylinders: Int

    @get:NotNull
    val crankShaft: CrankShaft // <4>

    @ConfigurationProperties("crank-shaft")
    interface CrankShaft { // <5>
        val rodLength: Double? // <6>
    }
}
```

It broke when upgrading to validation M6

dstepanov (approved):

PR #9241, opened 2023-05-16
Replace `javax.inject.*`, `javax.annotations.Nullable`, `javax.annotations.Nonnull` with Jakarta annotations

dstepanov:


core/src/main/java/io/micronaut/core/annotation/AnnotatedElement.java
```diff
@@ -29,27 +29,27 @@ public interface AnnotatedElement extends AnnotationMetadataProvider, Named {
      * @return Whether the element is nullable.
      */
     default boolean isDeclaredNullable() {
-        return getAnnotationMetadata().hasDeclaredStereotype(AnnotationUtil.NULLABLE);
+        return getAnnotationMetadata().hasDeclaredStereotype(AnnotationUtil.NULLABLE) || getAnnotationMetadata().hasDeclaredStereotype(AnnotationUtil.JAVAX_NULLABLE);
```
graemerocher, line 32: why is this needed? The annotation should be mapped

graemerocher (approved):

sdelamo (approved):

core/src/main/java/io/micronaut/core/annotation/AnnotatedElement.java
dstepanov: It was mapped to the javax one so everything compiled before would have the old annotation. We need to recompile the platform

PR #9245, opened 2023-05-17
adds Controller Constraint Handler test

sdelamo:


yawkat (approved):

graemerocher (approved):

sdelamo:
@graemerocher @yawkat currently we don't support such a POJO: 

```
@Introspected
    static class CredentialsWithNonNull {
        @NotBlank
        @Email
        @NonNull
        private final String username;
```

As a user I think it is natural to user a nullability annotation (`@NonNull`) which matches your POJOs constraints. 

We support 

```
@Introspected
    static class CredentialsWithNonNull {
        @NotBlank
        @Email
        private final String username;
```

and: 

```
@Introspected
    static class CredentialsWithNonNull {
        @NotBlank
        @Email
         @Nullable
        private final String username;
```

thoughts?

PR #9252, opened 2023-05-17
Workaround for recognizing "javax.inject.Named"

dstepanov:
This can be removed after the platform is recompiled with the latest Core.


dstepanov:
@sdelamo This is needed for the next milestone

PR #9253, opened 2023-05-17
Workaround for recognizing "javax.inject.Qualifier"

dstepanov:
The same problem as with javax named


dstepanov:
This changes make Tracing pass. The `javax` checks should be removed after the plaform is recompiled with the next milestone.

graemerocher:
Can you create draft PRs to revert these changes so it is not forgotten before GA

dstepanov:
After it's merged I will create one

sdelamo (approved):

PR #9287, opened 2023-05-21
Regression Writable in Filter

sdelamo:
Adds a test which reproduces the way the Micronaut Views Filter writes a writable as the body in the response within a filter.


sdelamo:
This test passes in [Micronaut Framework  3](https://github.com/micronaut-projects/micronaut-core/pull/9286) but it fails currently in 4. @yawkat @graemerocher can you help with this?

This is currently breaks the views module.

yawkat:
this one is caused because the type of the response is not statically known (controller has different body type than filter). handler matching is invariant so it cant find the proper handler.

i will fix it by making WritableBodyWriter a raw message handler, which has covariant matching, but it's ugly

graemerocher (approved):

PR #9296, opened 2023-05-22
Move TCK tests for Client Filters to client TCK

timyates:
These test the client, not the server so I think it's better to have them in the Client TCK


timyates:
I think the failing test is caused by the 3.10.x mergeup

https://github.com/micronaut-projects/micronaut-core/actions/workflows/graalvm-latest.yml?query=branch%3A4.0.x

sdelamo (approved):

PR #9299, opened 2023-05-22
Fix ControllerConstraintHandlerTest for JDK client

timyates:


timyates:
The TCK exclusions will be fixed in another PR (incoming as soon as I work out the issue) 😉

sdelamo (changes requested):
Tim can you merge master against this PR and update the HTTP Client documentation to reflect the things which are not supported. 

Currently we say: 

```
[Proxy request](https://docs.micronaut.io/4.0.0-M3/guide/#proxyClient) support (we do support [HTTP Proxies](https://docs.micronaut.io/4.0.0-M3/guide/#proxy)).

Client Filters.

Streaming support.

Multipart requests.

H2C and HTTP/3.
```

Are any of those supported?

timyates:
> Are any of those supported?

I don't believe so

sdelamo (approved):

Issue #9275, opened 2023-05-20
HTTP client changes HTTP method to GET when handling 308 Permanent Redirect


loicgreffier:
### Expected Behavior

When a `PUT` request returns a 308, the Micronaut HTTP client follows the redirect, but it modifies the HTTP method by `GET`.

Here an exemple where a resource returns 308 on `http://`:

```
17:17:24.234 [default-nioEventLoopGroup-3-6] DEBUG i.m.h.client.netty.DefaultHttpClient - Sending HTTP PUT to http://kafka-connect/connector-plugins/io.confluent.kafka.connect.datagen.DatagenConnector/config/validate
17:17:24.268 [default-nioEventLoopGroup-3-6] DEBUG i.m.h.client.netty.DefaultHttpClient - Received response 308 from http://kafka-connect/connector-plugins/io.confluent.kafka.connect.datagen.DatagenConnector/config/validate
17:17:24.298 [default-nioEventLoopGroup-3-8] DEBUG i.m.h.client.netty.DefaultHttpClient - Sending HTTP GET to https://kafka-connect/connector-plugins/io.confluent.kafka.connect.datagen.DatagenConnector/config/validate
17:17:24.374 [default-nioEventLoopGroup-3-8] DEBUG i.m.h.client.netty.DefaultHttpClient - Received response 405 from https://kafka-connect/connector-plugins/io.confluent.kafka.connect.datagen.DatagenConnector/config/validate
```

The HTTP client follow the `Location` header but changes the HTTP method.

I was expecting a redirection with the original HTTP method: `PUT`.

According to [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308), the body and the HTTP method should not be altered.

### Actual Behaviour

The Micronaut HTTP client follows redirects using GET for 308.

It looks like [this issue](https://github.com/micronaut-projects/micronaut-core/pull/6466), but for HTTP 308 instead of 307.

### Version

3.8.1

loicgreffier:
Linking the following PR #9304

PR #9304, opened 2023-05-23
Preserves original HTTP method while following permanent redirection

loicgreffier:
Attempt to address #9275

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -2049,7 +2049,7 @@ public class DefaultHttpClient implements
                 String location = headers1.get(HttpHeaderNames.LOCATION);
 
                 MutableHttpRequest<Object> redirectRequest;
-                if (code == 307) {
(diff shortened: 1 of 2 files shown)


yawkat:
reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections , i think it should be changed so that only 303 forces GET

loicgreffier:
Well as specified, for `301` and `302`, the other HTTP methods "**may or may not be changed to GET**".

I believe preserving redirections with `GET` for HTTP codes different from 307 and 308 is justified and will avoid some breaking changes.

yawkat:
yes, and I think we should only use GET for 303. but either way is acceptable.

loicgreffier:
@yawkat CI is failing due to:

```
http-client/src/test/groovy/io/micronaut/http/client/StreamRequestSpec.groovy | Expected exception of type 'java.lang.RuntimeException', but no exception was thrown
```

Not sure it is related to this PR, can you confirm ?

yawkat:
it's unrelated

yawkat (approved):

graemerocher:
Thanks for the contribution!

PR #9307, opened 2023-05-25
TCK Plain text for Boolean response

sdelamo:
The tests passes for Http Client JDK both manual and declarative. 

for Netty HTP Client , it fails for declarative passes for manual. 

This bug was first spotted in servlet. 

Conversation: 

@yawkat 
> can't deserialize boolean from text/plain

@graemerocher 
> why?

@yawkat
because it has no specified structure. we could parse it as json, but that could be wrong

@graemerocher 

> shouldn't it go through the normal conversion infrastructure for text/plain? We already have a String -> Boolean converter

@yawkat 
> this is what TextPlainCodec did, toString for serialization and conversion for deserialization. but going through conversion kind of defeats the point of the message handlers.

> On the write side, there is this workaround: https://github.com/micronaut-projects/micronaut-core/blob/37874c634202233f35b7c9376a5edfd5d49861f2/http/src/main/java/io/micronaut/http/body/DynamicMessageBodyWriter.java#L64-L69 – it's restricted to java.lang types, and just calls toString. but this does not work on the read side since there is no real inverse of toString, only conversion. it's in breaks.adoc here: https://github.com/micronaut-projects/micronaut-core/blob/37874c634202233f35b7c9376a5edfd5d49861f2/src/main/docs/guide/appendix/breaks.adoc#L186


sdelamo:
I have added the test and disable it for netty based http client for now

yawkat:
this is the intended behavior though? why add a test for something that's not supposed to work

yawkat:
you can reenable this behavior for netty using `micronaut.http.legacy-text-conversion`

sdelamo:
Please, note that it works for the HTTP Client based on Java HTTP Client.

yawkat:
yes and it shouldn't when the jdk client starts using MessageBodyHandler

sdelamo:
@yawkat do you have strong feelings about not merging this PR?

yawkat:
not particularly, i just think it's testing for behavior we don't want. though we can just remove it in the future.

sdelamo:
> not particularly, i just think it's testing for behavior we don't want. though we can just remove it in the future.

I am not sure I agree we don't want that behaviour. 

This looks like something users may want to do: 

```
  @Get
        Boolean index() {
            return Boolean.TRUE;
        }
```

Is the behaviour we want to encourage that users return a String instead:

```
  @Get
        String index() {
            return Boolean.TRUE.toString();
        }
```

yawkat:
Yes, they should manually call toString. Using toString for serialization is arbitrary, and is a mess for everything but the most simple types, especially for deserialization.

PR #9310, opened 2023-05-25
Only convert config placeholders when necessary

yawkat:


yawkat:
@graemerocher this breaks the build because it breaks caching for placeholders. Basically:

https://github.com/micronaut-projects/micronaut-core/blob/94f458ac04b0180a989d0cf28984d5e3b5df592f/http-client/src/test/groovy/io/micronaut/http/client/filter/StaticClientFilterSpec.groovy#L24-L28

The resolution of micronaut.server.port used to request the value of test-port as String. With this patch, it requests the value as Object instead (`.getValue(Object.class)`). Object is not cachable however, so it gets a new uncached random port, which causes the test failure.

I'm not really sure how to fix it. I tried making Object cachable. That did not resolve the issue, because the caching for Object and String happens separately. I think this logic is a bit weird in the first place, e.g. if you were to request test-port manually as int, it will return a new random port. But I don't want to open the can of worms of rewriting this.

https://github.com/micronaut-projects/micronaut-core/blob/75b767e393c68d661ce95e2331475b6ff66c4e7d/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java#L315

Maybe we just need to revert this PR.

yawkat:
It seems to me like caching is a bit "misused" here. For the caching purposes (converting the same value to the same type every time) it makes sense to cache this way. But for preventing regeneration of the random value, the value should be cached regardless of requested type. Maybe there should be a two-level cache, one depending on the requested type, one depending on the actual unconverted type of the value.

PR #9311, opened 2023-05-25
Add config option to restore legacy text/plain writing behavior

yawkat:


graemerocher (approved):

sdelamo (approved):
Can you document it, also in a section outside `breaks.adoc`. The `@Requires` will not appear in [ConfigurationReference](https://docs.micronaut.io/latest/guide/configurationreference.html) and we will eventually move breaks.adoc to https://micronaut-projects.github.io/micronaut-upgrade/snapshot/ ?

yawkat:
How do i add it to the config reference?

And I don't think it needs to be documented outside breaks, it's never a necessary option.

sdelamo:
> How do i add it to the config reference?

Our build automatically adds classes annotated with `@ConfigurationProperties` to the config reference. It uses the field name and the Javadoc in the setter/getter. 

> And I don't think it needs to be documented outside breaks, it's never a necessary option.

We should not have settings documented only in `breaks.adoc`. We clean up the "breaking changes" section between releases.

PR #9312, opened 2023-05-25
refreshAndDiff() should handle array values and property placeholders

graemerocher:
Currently byte[] values (such as secrets) if they are re-read they are not correctly diffed

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
+++ b/inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
@@ -816,6 +816,23 @@ public class DefaultEnvironment extends PropertySourcePropertyResolver implement
                 }
             }
         }
+        if (!changes.isEmpty()) {
+            Map<String, Object> placeholdersAltered = new LinkedHashMap<>();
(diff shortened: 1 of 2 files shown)


inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -833,13 +833,35 @@ private void diffMap(Map<String, Object> map, Map<String, Object> newMap, Map<St
...
-                } else if (hasNew && hasOld && !newValue.equals(oldValue)) {
+                } else if (hasNew && hasOld && hasChanged(newValue, oldValue)) {
                     changes.put(key, oldValue);
                 }
             }
         }
     }
 
+    @SuppressWarnings({"java:S1871", "java:S3776" })
+    private static boolean hasChanged(Object newValue, Object oldValue) {
```
yawkat, line 861: Objects.deepEquals does this

graemerocher: good point

yawkat (approved):

PR #9313, opened 2023-05-26
Fix property caching

yawkat:
This fixes the build broken by #9310. Basically:

https://github.com/micronaut-projects/micronaut-core/blob/94f458ac04b0180a989d0cf28984d5e3b5df592f/http-client/src/test/groovy/io/micronaut/http/client/filter/StaticClientFilterSpec.groovy#L24-L28

The resolution of micronaut.server.port used to request the value of test-port as String. With this patch, it requests the value as Object instead (`.getValue(Object.class)`). The caching in PropertySourcePropertyResolver depends on the requested type however, so the Object request (micronaut.server.port) and the String request (micronaut.http.services.a.url) are cached separately and have different computed ports. (Additionally Object request type is not cached at all, but this is not the core issue.)

This patch adds another cache *before* conversion that is independent of the requested type. This cache ensures that the value is not recomputed for different request types.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -84,14 +84,22 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
     protected Logger log;
 
     private final Map<String, Boolean> containsCache = new ConcurrentHashMap<>(20);
-    private final Map<String, Object> resolvedValueCache = new ConcurrentHashMap<>(20);
+    /**
+     * Cache for values <i>before</i> conversion. This avoids recomputing placeholders, which keeps
+     * random values (e.g. {@code ${random.port}} stable).
+     */
+    private final Map<String, Object> placeholderResolutionCache = new ConcurrentHashMap<>(20);
+    /**
+     * Cache for values <i>after</i> conversion.
+     */
+    private final Map<ConversionCacheKey, Object> resolvedValueCache = new ConcurrentHashMap<>(20);
     private final EnvironmentProperties environmentProperties = EnvironmentProperties.fork(CURRENT_ENV);
 
     /**
      * Creates a new, initially empty, {@link PropertySourcePropertyResolver} for the given {@link ConversionService}.
      *
      * @param conversionService The {@link ConversionService}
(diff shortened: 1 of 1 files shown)


inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -373,7 +388,11 @@ public <T> Optional<T> getProperty(@NonNull String name, @NonNull ArgumentConver
 
                     if (value != null) {
                         Optional<T> converted;
-                        value = resolvePlaceHoldersIfNecessary(value);
+                        if (entries != null) {
+                            // iff entries is null, the value is from placeholderResolutionCache and doesn't need this step
```
altro3, line 392: iff - typo

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -84,14 +84,22 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
...
+     */
+    private final Map<ConversionCacheKey, Object> resolvedValueCache = new ConcurrentHashMap<>(20);
     private final EnvironmentProperties environmentProperties = EnvironmentProperties.fork(CURRENT_ENV);
 
     /**
      * Creates a new, initially empty, {@link PropertySourcePropertyResolver} for the given {@link ConversionService}.
      *
      * @param conversionService The {@link ConversionService}
-     * @param logEnabled logEnabled flag to enable or disable logger
+     * @param logEnabled        logEnabled flag to enable or disable logger
```
altro3: That's my mistake `logEnabled` word twice. Could you fix it?

yawkat:
```suggestion
     * @param logEnabled        flag to enable or disable logger
```

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
yawkat: it's short for 'if and only if'

altro3: Sorry, didn't know :-)

PR #9318, opened 2023-05-26
Add test for `@Body` not being required in 3.9.x

timyates:
This should be removed in 4.0.0 as the annotation is required

Backed it up behind https://github.com/micronaut-projects/micronaut-core/pull/9317 as it requires the tests to be enabled again.


sdelamo (changes requested):
Add a second test in this class which tests it works with the @Body annotation

sdelamo (approved):

PR #9319, opened 2023-05-26
ErrorHandlerStringTest fix

n0tl3ss:
Test should be passing because string "hello" is not `MediaType.APPLICATION_JSON` type. It should be  `MediaType.TEXT_PLAIN`

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerStringTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerStringTest.java
@@ -44,13 +44,13 @@ public class ErrorHandlerStringTest {
     @Test
     void testErrorHandlerWithStringReturn() throws IOException {
         asserts(SPEC_NAME,
-            HttpRequest.GET("/exception/my").header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON),
+            HttpRequest.GET("/exception/my"),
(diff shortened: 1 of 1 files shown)


timyates:
Is this another breaking change of behaviour that needs documenting here https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/src/main/docs/guide/appendix/breaks.adoc#exceptionhandler-with-pojo-response-type-no-longer-results-in-an-error-response

?

sdelamo:
@n0tl3ss I think the test was verifying the that default content type for a exception handler returning string is application/json 

Why did you change the test?

n0tl3ss:
I think we should return correct content type of response. String "hello" is not json it makes more sense to be text/plain. I edited test with @Produce annotation but maybe we want to return correct media type by default.

sdelamo:
> I think we should return correct content type of response. String "hello" is not json it makes more sense to be text/plain. I edited test with @produce annotation but maybe we want to return correct media type by default.

we could change `hello` with `{"message": "hello"}"` but I don't think we should change the test and add annotations. The goal of the test is to verifies what is the default content type for a return String when no `@Produces` annotations  are present.

graemerocher:
the default if no produces is specified is `application/json` and as @n0tl3ss said "hello" is not valid JSON

sdelamo:
@n0tl3ss I have changed the string to be a valid JSON and removed the @Produces annotation. Let me know if you are ok.

PR #9320, opened 2023-05-26
Handle errors in publisher responses in RoutingInBoundHandler

yawkat:
This moves the handling logic from PipeliningServerHandler to RoutingInBoundHandler, because the latter can actually write the body properly.

This should fix https://github.com/micronaut-projects/micronaut-starter/pull/1764 .

Gradle is broken locally for me, so I'm not sure if the tests pass :)

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/body/NettyWriteContext.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/body/NettyWriteContext.java
@@ -18,11 +18,12 @@ package io.micronaut.http.netty.body;
 import io.micronaut.core.annotation.Experimental;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NonNull;
-import io.micronaut.http.netty.stream.StreamedHttpResponse;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpChunkedInput;
+import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpResponse;
+import org.reactivestreams.Publisher;
 
 import java.io.RandomAccessFile;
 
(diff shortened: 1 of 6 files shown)


graemerocher:
looks like Gradle is not just broken locally 🤔

yawkat:
@melix can you take a look at this? CI crashes, and locally the gradle build just hangs

melix:
The problem can be reproduced by running `./gradlew :h-s-netty:test --tests io.micronaut.http.server.netty.binding.ByteBufferSpec` locally. The build doesn't hang anymore, but throws read timeouts and kills the daemon.

melix:
@yawkat there's probably something happening in the native stack of Netty which kills the communication between the Gradle daemon and the test process. However, you can run this single test: `./gradlew :http-server-netty:test --tests "io.micronaut.http.server.netty.binding.ByteBufferSpec.test reading the body with a publisher of bytebuffers"`

and see that it spits a lot of exceptions:

```
Caused by: io.netty.util.IllegalReferenceCountException: refCnt: 0, increment: 1
	at io.netty.util.internal.ReferenceCountUpdater.retain0(ReferenceCountUpdater.java:133)
	at io.netty.util.internal.ReferenceCountUpdater.retain(ReferenceCountUpdater.java:120)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.retain(AbstractReferenceCountedByteBuf.java:81)
	at io.netty.handler.codec.http.DefaultHttpContent.retain(DefaultHttpContent.java:68)
	at io.netty.handler.codec.http.DefaultHttpContent.retain(DefaultHttpContent.java:25)
	at io.netty.util.ReferenceCountUtil.retain(ReferenceCountUtil.java:41)
	at io.netty.handler.codec.http.HttpContentEncoder.encode(HttpContentEncoder.java:216)
	at io.micronaut.http.server.netty.SmartHttpContentCompressor.encode(SmartHttpContentCompressor.java:65)
	at io.micronaut.http.server.netty.SmartHttpContentCompressor.encode(SmartHttpContentCompressor.java:33)
	at io.netty.handler.codec.MessageToMessageCodec$1.encode(MessageToMessageCodec.java:67)
	at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:90)
	... 47 common frames omitted
```

I got the stack trace from running the test in the IDE.

yawkat:
@melix thanks! that gives me a test to work on.

However the gradle issue is likely not caused by netty natives, the test does not use any of them. I think it is an issue caused by a *lot* of log output produced by the test. I've seen gradle behave similarly in other cases where I accidentally left a lot of logging enabled.

yawkat:
i had to change ErrorResponseSpec slightly. previously it would put the message in the HTTP status line. I've decided not to replicate that behavior anymore because the status line doesn't work well for carrying information, e.g. the message is dropped entirely with http/2.

PR #9347, opened 2023-05-29
Improve MessageBodyHandler selection

andriy-dmytruk:
* Use the message body writer and reader order in selection
* For writers, only select those that can consume a particular type

### DIFF
--- a/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
+++ b/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
@@ -96,7 +96,7 @@ public final class DefaultMessageBodyHandlerRegistry extends RawMessageBodyHandl
             } else {
                 // Pick the highest priority
                 return beanDefinitions.stream()
-                    .max(OrderUtil.COMPARATOR)
+                    .max(OrderUtil.REVERSE_COMPARATOR)
                     .map(beanLocator::getBean)
                     .orElse(null);
             }
@@ -156,7 +156,7 @@ public final class DefaultMessageBodyHandlerRegistry extends RawMessageBodyHandl
             } else {
                 // Pick the highest priority
                 return beanDefinitions.stream()
-                    .max(OrderUtil.COMPARATOR)
+                    .max(OrderUtil.REVERSE_COMPARATOR)
                     .map(beanLocator::getBean)
                     .orElse(null);
             }
--- a/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
+++ b/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
@@ -156,8 +156,8 @@ public final class DefaultMessageBodyHandlerRegistry extends RawMessageBodyHandl
             } else {
                 // Pick the highest priority
                 return beanDefinitions.stream()
-                    .map(beanLocator::getBean)
                     .max(OrderUtil.COMPARATOR)
+                    .map(beanLocator::getBean)
                     .orElse(null);
             }
         }
--- a/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
+++ b/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
@@ -23,6 +23,7 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.order.OrderUtil;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.GenericPlaceholder;
+import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.http.MediaType;
 import io.micronaut.http.annotation.Consumes;
 import io.micronaut.http.annotation.Produces;
@@ -95,8 +96,8 @@ public final class DefaultMessageBodyHandlerRegistry extends RawMessageBodyHandl
             } else {
                 // Pick the highest priority
                 return beanDefinitions.stream()
-                    .map(beanLocator::getBean)
                     .max(OrderUtil.COMPARATOR)
+                    .map(beanLocator::getBean)
                     .orElse(null);
             }
         }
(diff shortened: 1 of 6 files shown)


http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -132,8 +133,10 @@ private List<MediaType> resolveMediaTypes(List<MediaType> mediaTypes) {
     @Override
     protected <T> MessageBodyWriter<T> findWriterImpl(Argument<T> type, List<MediaType> mediaTypes) {
         Collection<BeanDefinition<MessageBodyWriter>> beanDefinitions = beanLocator.getBeanDefinitions(
-            Argument.of(MessageBodyWriter.class, type),
-            newMediaTypeQualifier(type, mediaTypes, Produces.class)
+            // Do not put the type here since we are looking for writers that can process the type
+            //      but beanLocator will provide types that can be injected into the searched type
+            Argument.of(MessageBodyWriter.class),
```
yawkat, line 139: i dont think this change is correct

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -150,10 +153,10 @@ protected <T> MessageBodyWriter<T> findWriterImpl(Argument<T> type, List<MediaTy
             if (exactMatch.size() == 1) {
                 return beanLocator.getBean(exactMatch.iterator().next());
             } else {
-                // pick highest priority
-                return OrderUtil.sort(beanDefinitions.stream())
-                    .findFirst()
+                // Pick the highest priority
+                return beanDefinitions.stream()
```
yawkat, line 158: is this a functional change? it seems like this flips the order

andriy-dmytruk: It picks the highest priority as the comment suggests it should. I don't know what the correct behavior should be, so it can be flipped if needed.

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -166,6 +169,16 @@ protected <T> MessageBodyWriter<T> findWriterImpl(Argument<T> type, List<MediaTy
         @Override
         public <B extends BeanType<T>> Stream<B> reduce(Class<T> beanType, Stream<B> candidates) {
             return candidates.filter(c -> {
+                // Only for writer, verify that the writer can consume the required type
+                if (type.getType() == MessageBodyWriter.class && c instanceof BeanDefinition<?> definition) {
+                    List<Argument<?>> consumedType = definition.getTypeArguments(type.getType());
+                    Argument<?> requiredType = type.getTypeParameters()[0];
```
graemerocher: need to handle the case if `getTypeParmeters()` returns 0

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -166,6 +169,16 @@ protected <T> MessageBodyWriter<T> findWriterImpl(Argument<T> type, List<MediaTy
...
+                    if (consumedType.isEmpty() ||
+                        !(consumedType.get(0).isTypeVariable() || consumedType.get(0).isAssignableFrom(requiredType.getType()))
```
graemerocher: extra a variable for the `consumedType.get(0)` call

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
yawkat: it should comply with Ordered.HIGHEST_PRECEDENCE and LOWEST_PRECEDENCE, so it should be flipped

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
yawkat: are you moving this logic to reduce down below? what is the actual change here? I dont understand what youre fixing and why it's fixed

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
yawkat: it's never empty, it's only constructed in the registry

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -95,8 +96,8 @@ protected <T> MessageBodyReader<T> findReaderImpl(Argument<T> type, List<MediaTy
             } else {
                 // Pick the highest priority
                 return beanDefinitions.stream()
-                    .map(beanLocator::getBean)
                     .max(OrderUtil.COMPARATOR)
+                    .map(beanLocator::getBean)
```
andriy-dmytruk, line 100: `BeanDefinition` does not implement an ordered interface, so this might be not working correctly.

yawkat: i guess it should be both? people may implement Ordered, or use @Order. if sorting is done before getBean the latter will work, if it's done after the former.

graemerocher: it doesn't need to. This ends up calling `getOrder(Object o)` which is implemented like:

```
    if (o instanceof Ordered) {
            return getOrder((Ordered) o);
        } else if (o instanceof AnnotationMetadata) {
            return getOrder(((AnnotationMetadata) o));
        }
    
```

The annotation metadata will be used.

andriy-dmytruk: So that is how it is done... looks efficient.
Can we stop extending the `Ordered` interface in message writers then, since it is not used?

graemerocher: pushed the changes to do this

graemerocher (approved):

andriy-dmytruk:
Looks good. Thanks, Graeme!

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
andriy-dmytruk: The change is the difference between reader and writer: 
* A `MessageBodyReader<T>` would have `T` as a return type therefore for a type `P` we would be looking for any `MessageBodyReader<T>` such that `T extends P`. 
* A `MessageBodyWriter<T>` would have `T` as a method parameter therefore for a type `P` we would be looking for any `MessageBodyWriter<T>` such that `P extends T`.

The main difference is how the `isAssignableFrom` method is called.

yawkat: so basically, the change is to make writer matching covariant?

andriy-dmytruk: Yes, and prevent the selection of a body writer for contravariant types, which would throw an error when calling the method.

PR #9349, opened 2023-05-29
Introduce `MutablePropagationContext` for HTTP filters

dstepanov:
I will add the documentation as part of context propagation documentation.

### DIFF
--- a/http/src/main/java/io/micronaut/http/filter/FilterRunner.java
+++ b/http/src/main/java/io/micronaut/http/filter/FilterRunner.java
@@ -25,6 +25,7 @@ import io.micronaut.core.execution.ExecutionFlow;
 import io.micronaut.core.execution.ImperativeExecutionFlow;
 import io.micronaut.core.order.OrderUtil;
 import io.micronaut.core.order.Ordered;
+import io.micronaut.core.propagation.MutablePropagatedContext;
 import io.micronaut.core.propagation.PropagatedContext;
 import io.micronaut.core.type.Argument;
 import io.micronaut.core.type.Executable;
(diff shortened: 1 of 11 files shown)


http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -425,6 +433,8 @@ public static <T> FilterMethod<T> prepareFilterMethod(ConversionService conversi
                 } else {
                     throw new IllegalArgumentException("Unsupported continuation type: " + continuationReturnType);
                 }
+            } else if (argument.getType() == MutablePropagatedContext.class) {
```
yawkat, line 436: maybe could support normal PropagatedContext as param type too?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -580,7 +592,12 @@ private ExecutionFlow<FilterContext> filter(FilterContext filterContext,
                 } else {
                     returnValue = method.invoke(bean, args);
                 }
-                return returnHandler.handle(filterContext, returnValue, methodContext.continuation);
+                ExecutionFlow<FilterContext> executionFlow = returnHandler.handle(filterContext, returnValue, methodContext.continuation);
+                PropagatedContext mutatedPropagatedContext = methodContext.mutablePropagatedContext.getContext();
```
yawkat, line 596: it would be nice if the return handler could do this, only if the filter actually defines a mutable context param. but probably overkill for this pr.

yawkat (approved):

graemerocher (approved):

PR #9353, opened 2023-05-30
allow access to the body from an annotation driven filter

graemerocher:
* allows using regular HTTP binding annotations in filters
* allows binding certain body types and fails otherwise

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/body/ImmediateByteBody.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/body/ImmediateByteBody.java
@@ -34,10 +34,7 @@ import io.netty.handler.codec.http.DefaultLastHttpContent;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.LastHttpContent;
 
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -48,14 +45,12 @@ import java.util.List;
  * @author Jonas Konrad
  */
 @Internal
-public final class ImmediateByteBody extends ManagedBody<ByteBuf> implements ByteBody, ByteBuffer<ByteBuf> {
+public final class ImmediateByteBody extends ManagedBody<ByteBuf> implements ByteBody {
     private final boolean empty;
-    private final ByteBuffer<ByteBuf> byteBuffer;
 
     ImmediateByteBody(ByteBuf buf) {
         super(buf);
         this.empty = !buf.isReadable();
-        this.byteBuffer = NettyByteBufferFactory.DEFAULT.wrap(buf);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/body/ImmediateByteBody.java
```diff
@@ -45,12 +48,14 @@
  * @author Jonas Konrad
  */
 @Internal
-public final class ImmediateByteBody extends ManagedBody<ByteBuf> implements ByteBody {
+public final class ImmediateByteBody extends ManagedBody<ByteBuf> implements ByteBody, ByteBuffer<ByteBuf> {
```
yawkat: just access the ByteBuf value directly instead and wrap it in NettyByteBuffer, no need to delegate this much

graemerocher: then I would need to couple the filter logic to netty which I can't do the way the current code is structured

yawkat: i also think that conversion to ByteBuffer should fail if the body is already claimed, it should not be an issue in filters and will prevent race conditions

graemerocher: not sure what you mean here, maybe you want to send a PR with these improvements

http/src/main/java/io/micronaut/http/filter/BaseFilterProcessor.java
```diff
@@ -52,10 +67,50 @@
...
+            public <T> Optional<ArgumentBinder<T, HttpRequest<?>>> findArgumentBinder(Argument<T> argument) {
+                Class<? extends Annotation> annotation = argument.getAnnotationMetadata().getAnnotationTypeByStereotype(Bindable.class).orElse(null);
+                if (annotation != null && PERMITTED_BINDING_ANNOTATIONS.contains(annotation.getName())) {
+                    if (annotation == Body.class) {
+                        return Optional.of((context, source) -> {
+                            if (source instanceof FullHttpRequest<?> fullHttpRequest && fullHttpRequest.isFull()) {
+                                ByteBuffer<?> contents = fullHttpRequest.contents();
+                                if (contents != null) {
+                                    Argument<T> t = context.getArgument();
+                                    if (t.isAssignableFrom(ByteBuffer.class)) {
```
yawkat, line 95: could also go through RawBodyReader for this

yawkat: advantage would be ByteBuf param support, but maybe that's too difficult to be worth it

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/ImmediateByteBody.java
yawkat: sure

yawkat: i think i can make it work without FULL_CONTENT too, i will do it after lunch.

dstepanov (approved):

dstepanov:
Needs some documentation. Maybe we can have a table of all possible things that can be injected.

yawkat:
@graemerocher i didnt fix all test failures btw

yawkat (approved):

PR #9357, opened 2023-05-30
Move filters to new API

yawkat:


yawkat:
@graemerocher please wait for tests to pass before merging :smile: I only ran related tests locally.

PR #9373, opened 2023-06-01
Improve scheduled exception handling

graemerocher:
a timing issue can occur when the context is being shutdown. If a scheduled job runs and the context is shutdown and available beans cleared then the bean is no longer available. The best thing to do here is just ignore the failure.


sdelamo (commented):
CI fails: https://ge.micronaut.io/s/qlisckfow54lg/tests/:runtime:test/io.micronaut.scheduling.RetrySchedulingSpec/test%20scheduled%20annotation%20with%20retry?top-execution=1

sdelamo (approved):

PR #9387, opened 2023-06-05
Extract Netty specific RequestFilter TCK test

timyates:
This test requires FullBodyRequest support which doesn't exist in Servlet.

Introduced here https://github.com/micronaut-projects/micronaut-core/pull/9353

To avoid having to exclude all of the Request Filter tests from the Servlet TCK, we have extracted this single test into it's own specification.

See https://github.com/micronaut-projects/micronaut-servlet/pull/470 for discussion.

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/FullBodyRequestFilterTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/FullBodyRequestFilterTest.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.server.tck.tests.filter;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.HttpStatus;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.Body;
+import io.micronaut.http.annotation.Controller;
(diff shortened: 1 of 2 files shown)


sdelamo:
what does it mean to be a `Full HTTP request`?

yawkat:
that the full body is immediately available for reading

yawkat (approved):

timyates:
I guess the more correct fix is to implement the required parts of https://github.com/micronaut-projects/micronaut-core/pull/9353 in Servlet, so the full TCK passes there too...

yawkat:
the test doesn't even work all the time in netty, it's best effort. if you used a larger body it would fail.

(unless you enable FULL_CONTENT mode, which the test doesn't)

graemerocher:
don't think we should do this, this interface should be simply to implement in servlet

graemerocher (approved):

graemerocher:
extracting the test is probably fine tho

yawkat:
@graemerocher i dont see a sensible way to implement it though

graemerocher:
servlet exposes `getInputStream()` on the request. So you would have a `ByteBuffer` implementation that backs onto the this `InputStream` implementation and resets the input stream to the start position after the bytes are read (which is how any servlet filter that reads the body would have to be implemented)

graemerocher:
actually I think resetting the input stream will probably not work, it would need to be implemented as a wrapper, and cache the bytes a bit like https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/util/ContentCachingRequestWrapper.java

yawkat:
it would essentially require something like full_content though. we can't risk blocking in the body annotation binder, so we can't block in FullHttpRequest.contents

graemerocher:
a servlet container is not like Netty and always has the full content available from the request

yawkat:
not always, the stream is still lazy

graemerocher:
the way you toggle that off and on is by enabling/disabling async request processing. This would be the equivalent of `full_content`

yawkat:
the body is not necessarily fully present no matter whether youre using async mode or not in the servlet.

graemerocher:
how so? It is available via `getInputStream()` is it not? Reading the body is a servlet filter is technically possible as demonstrated by the range of other implementations  that exist out there including the one referenced above.

yawkat:
but reading the stream may block while the netty implementation never will. theres no api to know whether it will block or not

graemerocher:
if you disable asynchronous request processing you already assume the code that is going to be executed is blocking

timyates:
I got this together as a POC

Not sure it's right...  PTAL: https://github.com/micronaut-projects/micronaut-servlet/pull/471

timyates:
Closing this now, as it's superseded by https://github.com/micronaut-projects/micronaut-servlet/pull/471

PR #9405, opened 2023-06-07
Wait for request body if necessary in filter

yawkat:
Use FullHttpRequest.bufferContents to wait for the request body if it is needed for a filter method but isn't immediately available.

Also changed the bufferContents contract a bit so that this will fail more gracefully in servlet.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
@@ -1,0 +1,61 @@
+package io.micronaut.http.server.netty.filters
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.core.annotation.Order
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.MediaType
+import io.micronaut.http.annotation.Body
+import io.micronaut.http.annotation.RequestFilter
+import io.micronaut.http.annotation.ServerFilter
+import io.micronaut.http.client.HttpClient
+import io.micronaut.runtime.server.EmbeddedServer
+import jakarta.inject.Singleton
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Sinks
+import spock.lang.Specification
+
+class FilterBodySpec extends Specification {
+    def "delayed body"() {
+        given:
+        def ctx = ApplicationContext.run(["spec.name":"FilterBodySpec"])
+        def server = ctx.getBean(EmbeddedServer)
+        server.start()
+        def client = ctx.createBean(HttpClient, server.URI)
+        def filter = ctx.getBean(MyFilter)
+
+        when:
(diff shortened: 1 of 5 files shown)


yawkat:
@dstepanov I think it would be really useful to have an ArgumentBinder that returns an ExecutionFlow<BindingResult<T>> at some point, it would save us a lot of trouble

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
```diff
@@ -0,0 +1,61 @@
...
+import io.micronaut.http.annotation.ServerFilter
+import io.micronaut.http.client.HttpClient
+import io.micronaut.runtime.server.EmbeddedServer
+import jakarta.inject.Singleton
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Sinks
+import spock.lang.Specification
+
+class FilterBodySpec extends Specification {
+    def "delayed body"() {
```
graemerocher: can you add another test that tests the case where a controller method is reached after a request filter processes the delayed body

graemerocher (changes requested):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
yawkat: ok

graemerocher (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+        @RequestFilter("/**")
+        @Order(-1)
+        void signal() {
+            secondBodyPartSink.emitValue("bar", Sinks.EmitFailureHandler.FAIL_FAST)
+        }
+
+        @RequestFilter("/delayed-body")
+        @Order(0)
+        HttpResponse<String> await(@Body String body) {
```
dstepanov, line 66: Why would you have two filters with the same order?

dstepanov (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
yawkat: one for testing with the controller and one for without

dstepanov:
I think we should add an option to wait for the body to the `@RequestFilter` for use cases when you want to inject simple `HttpRequest` and read the body.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterBodySpec.groovy
dstepanov: Yeah I didn't notice the different path

graemerocher:
yes perhaps if the request specifies a concrete type like `HttpRequest<Foo>`

PR #9406, opened 2023-06-07
By default, add JSON additional Types

sdelamo:
We have several modules, e.g. problem+json, which require a JSON additional type.

I am not sure that the JsonMediaTypeCodec additional types are used. It seems the key is the @Produces and @Consumes annotations' values in NettyJsonHandler.

This PR creates a constant in the JsonMediaTypeCodec with the additional types which I think is the correct way of defining those built in additional types and it adds a test which verifies the NettyJsonHandler produces and consumes annotation match that constant.


yawkat:
JsonMessageHandler should get the new types too

sdelamo:
> JsonMessageHandler should get the new types too

@yawkat done via https://github.com/micronaut-projects/micronaut-core/pull/9406/commits/b941de5aae317b3444b220874da853d51bcc3c0f

yawkat (approved):

PR #9410, opened 2023-06-07
Add breaking change note about the removal of `OncePerRequestHttpServerFilter`

### LINKED ISSUE #9384: Document OncePerRequestHttpServerFilter removal
### Issue description

We removed OncePerRequestHttpServerFilter in #9344, I think we should add a note in breaks.doc 

This is what our javadoc said: 

> * @deprecated All filters are executed once per request starting in Micronaut 3.0. Directly implement {@link HttpServerFilter} instead of extending this class and replace any usages of `micronaut.once` attributes with a custom attribute name.

wetted:
closes #9384


sdelamo (changes requested):
we have to tell the users what to use instead

dstepanov:
There is no need for this filter `server filters are executed once per request`

sdelamo (approved):

sdelamo:
@jeremyg484 if possible this should be incorporated to OpenRewrite

jeremyg484:
> @jeremyg484 if possible this should be incorporated to OpenRewrite

We've already got a recipe for it - https://github.com/openrewrite/rewrite-micronaut/blob/main/src/main/java/org/openrewrite/java/micronaut/OncePerRequestHttpServerFilterToHttpServerFilter.java

I'll add it to the list of recipes that get applied in the 3-to-4 migration.

PR #9413, opened 2023-06-08
Fix SSL context leak

### LINKED ISSUE #9409: Buffer leaks with HTTPS in 4.0.x branch
### Expected Behavior

No buffer leaks

### Actual Behaviour

```
18:02:59.493 [Test worker] ERROR io.netty.util.ResourceLeakDetector - LEAK: ReferenceCountedOpenSslContext.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
  Recent access records:
  Created at:
        Hint: expired
        io.netty.handler.ssl.ReferenceCountedOpenSslContext.<init>(ReferenceCountedOpenSslContext.java:262)
        io.netty.handler.ssl.ReferenceCountedOpenSslServerContext.<init>(ReferenceCountedOpenSslServerContext.java:74)
        io.netty.handler.ssl.ReferenceCountedOpenSslServerContext.<init>(ReferenceCountedOpenSslServerContext.java:62)
        io.netty.handler.ssl.SslContext.newServerContextInternal(SslContext.java:480)
        io.netty.handler.ssl.SslContextBuilder.build(SslContextBuilder.java:610)
        io.micronaut.http.server.netty.ssl.AbstractServerSslBuilder.build(AbstractServerSslBuilder.java:88)
        io.micronaut.http.server.netty.ssl.AbstractServerSslBuilder.build(AbstractServerSslBuilder.java:76)
        io.micronaut.http.server.netty.ssl.AbstractServerSslBuilder.build(AbstractServerSslBuilder.java:69)
        io.micronaut.http.server.netty.HttpPipelineBuilder.<init>(HttpPipelineBuilder.java:132)
```

### Steps To Reproduce

run `./gradlew http-server-netty:test`

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.0.x

yawkat:
This patch releases the SSL context when needed. Should fix #9409

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
@@ -206,6 +206,7 @@ public class ConnectionManager {
     }
 
     final void refresh() {
+        SslContext oldSslContext = sslContext;
         if (configuration.getSslConfiguration().isEnabled()) {
             sslContext = nettyClientSslBuilder.build(configuration.getSslConfiguration(), httpVersion);
         } else {
@@ -220,6 +221,7 @@ public class ConnectionManager {
         for (Pool pool : pools.values()) {
             pool.forEachConnection(c -> ((Pool.ConnectionHolder) c).windDownConnection());
         }
+        ReferenceCountUtil.release(oldSslContext);
     }
 
     /**
@@ -364,6 +366,8 @@ public class ConnectionManager {
                 Thread.currentThread().interrupt();
             }
         }
+        ReferenceCountUtil.release(sslContext);
+        sslContext = null;
     }
 
     /**
--- a/http-client/build.gradle
+++ b/http-client/build.gradle
@@ -1,3 +1,5 @@
+import org.apache.tools.ant.taskdefs.condition.Os
+
 plugins {
     id "io.micronaut.build.internal.convention-library"
 }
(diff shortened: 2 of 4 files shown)


http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -220,6 +221,7 @@ final void refresh() {
         for (Pool pool : pools.values()) {
             pool.forEachConnection(c -> ((Pool.ConnectionHolder) c).windDownConnection());
         }
+        ReferenceCountUtil.release(oldSslContext);
```
graemerocher, line 224: should this be in a `try/finally`?

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -364,6 +366,8 @@ public final void shutdown() {
                 Thread.currentThread().interrupt();
             }
         }
+        ReferenceCountUtil.release(sslContext);
```
graemerocher, line 369: same here `finally` block?

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
yawkat: i think it's safer this way, we can't really recover from error in this method and if it does fail, it's better not to release the old context in case a connection may make use of the old context still

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
yawkat: again can't really recover from failure and we don't want to release if there may be connections left to clean

PR #9416, opened 2023-06-08
Fix out-of-order write in HttpStreamsHandler

yawkat:
HttpStreamsHandler delays writing the first HttpContent to when the request has been fully written. The second HttpContent was not delayed however, so when writing the request took time (apparently with TLS?), the second HttpContent was written before the first.

This patch delays all further writes (normal, complete and error) until the first content has been written.

Should fix https://github.com/micronaut-projects/micronaut-tracing/issues/316


graemerocher:
in retrospect can you target 3.9.x the new will manually back port to avoid the build issues

timyates (approved):

graemerocher:
now we need a back port PR to 3.8.x that also fixes the build

yawkat:
@graemerocher the streams-order branch is still based on 3.8, so it can pr fine to 3.8.x. but i dont know how to fix the build.

graemerocher:
it is a matter of upgrading to the latest patch releases of the build plugins used by that branch

PR #9417, opened 2023-06-08
Escape header values correctly for access log

n0tl3ss:
- This PR correctly escaping header value for access log.
- Current pattern now makes more sense:
```
public static final String COMBINED_LOG_FORMAT = "%h %l %u %t \"%r\" %s %b \"%{Referer}i\" \"%{User-Agent}i\"";
```
and it will put Referer and  User-Agent between double quotes without single one.

Earlier behaviour:
```
localhost - - [08/Jun/2023:16:34:40 +0200] "GET /tet12313 HTTP/1.1" 401 - "'Referer-Header-Value'" "'User-Agent-Value'"
```

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/AbstractHttpMessageLogElement.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/AbstractHttpMessageLogElement.java
@@ -29,6 +29,15 @@ import java.util.Set;
  * @since 2.0
  */
 abstract class AbstractHttpMessageLogElement implements LogElement {
+    private static final Set<Character> CHARACTERS_TO_ESCAPE = new HashSet<>(){{
+        add('\b');
+        add('\n');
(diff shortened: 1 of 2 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/AbstractHttpMessageLogElement.java
```diff
@@ -28,6 +29,15 @@
...
  */
 abstract class AbstractHttpMessageLogElement implements LogElement {
+    private static final Set<Character> CHARACTERS_TO_ESCAPE = new HashSet<>(){{
+        add('\b');
+        add('\n');
+        add('\t');
+        add('\r');
+        add('\\');
+        add('"');
+    }};
```
graemerocher: Use `Set.of`

graemerocher (changes requested):

graemerocher (approved):

PR #9420, opened 2023-06-08
fixes #7290 resolve also raw map dataSourceProperties...

### LINKED ISSUE #7290: StringConvention.RAW breaks resolution of Map property with camel cased name
### Expected Behavior

When parsing a property of type `Map`, we use `StringConvention.RAW` when we don't want the keys under the `Map` to be converted to kebab case. 

ie. for properties defined thusly:

```
foo:
  config:
    barBlah:
      a.b.c-d: value1
      e.f.g-h: value2
      e.fooBarBlah: value3
    fooBar: something
```

We use `StringConvention.RAW` for the corresponding java class definition. eg.

```
@ConfigurationProperties("foo.config")
public class FooConfig {
  private final String fooBar;
  private final Map<String, String> barBlah;

  @ConfigurationInject
  public FooConfig(String fooBar,
    @MapFormat(transformation = MapFormat.MapTransformation.FLAT, keyFormat = StringConvention.RAW) Map<String, String> barBlah) {
    this.fooBar = fooBar;
    this.barBlah = barBlah;
  }

  public String getFooBar() {
    return fooBar;
  }

  public Map<String, String> getBarBlah() {
    return barBlah;
  }
}
```

The expectation is that the the properties are parsed correctly with `FooConfig.barBlah` getting all the map keys and values from the properties verbatim.

### Actual Behaviour

In such a case the parsing fails completely for the `FooConfig.barBlah` map, which ends up being assigned an empty valued map. If instead we remove the `StringConvention.RAW` specification, the parsing works correctly, however each of the map keys for `FooConfig.barBlah` end up being converted to kebab-case. ie. it gets the values:

```
{a.b.c-d=value1, e.f.g-h=value2, e.foo-bar-blah=value3}
```

The only way to get this to work with `StringConvention.RAW` is to rename `barBlah` in `application.yml` to `bar-blah` which allows things to work as intended.

### Steps To Reproduce

1. Have a property of type `Map` with a camel-cased name in the `application.yml`
2. Have keys under this following arbitrary casing rules
3. Try to wire this up to a java Map using `MapFormat.FLAT` and `StringConvention.RAW`.

### Environment Information

- MacOS Big Sur 11.6.2
- OpenJDK 11

### Example Application

https://github.com/hhajela/micronaut-test

### Version

3.4.2

heli80:
and not just data-source-properties

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -500,11 +500,13 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
         }
         String prefix = name + '.';
         entries.entrySet().stream()
-            .filter(map -> map.getKey().startsWith(prefix))
+            // NOTE: entries may have raw form while the prefix is normalized, so we also need to compare with a normalized key;
+            //  removing a matching prefix must then be done via counting dots since substring() won't work always
+            .filter(map -> map.getKey().startsWith(prefix) || normalizeKey(map.getKey()).startsWith(prefix))
             .forEach(entry -> {
                 Object value = entry.getValue();
                 if (value != null) {
-                    String key = entry.getKey().substring(prefix.length());
+                    String key = removePrefixByCountingDots(entry.getKey(), prefix);
                     key = keyConvention != null ? keyConvention.format(key) : key;
                     properties.put(key, resolvePlaceHoldersIfNecessary(value.toString()));
                 }
@@ -563,8 +565,10 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
                 continue;
             }
 
-            if (key.startsWith(prefix)) {
-                String subMapKey = key.substring(prefix.length());
+            // NOTE: entries may have raw form while the prefix is normalized, so we also need to compare with a normalized key;
+            //  removing a matching prefix must then be done via counting dots since substring() won't work always
+            if (key.startsWith(prefix) || normalizeKey(key).startsWith(prefix)) {
+                String subMapKey = removePrefixByCountingDots(key, prefix);
 
                 Object value = resolvePlaceHoldersIfNecessary(entry.getValue());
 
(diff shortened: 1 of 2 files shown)


graemerocher:
what is the use case?

heli80:
Ah sorry, thought the explanation from #7290 was already good enough.

Well in my example, I had the issue that timeout values defined via datasource properties in _application.yml_ did actually not work:
```
datasources:
  default:
    ...
    dataSourceProperties:
      options: -cstatement_timeout=240000
      socketTimeout: 270
```
Through experimentation I found out that they only work if kebap-case style is used:
```
    data-source-properties:
      options: -cstatement_timeout=240000
      socketTimeout: 270
```
That was a surprise, since my general believe was that camel-case and kebap-case can be used interchangeably in _application.yml_.

As written in #7290, the problem stems from the fact that in _jdbc-hikari_ _DatasourceConfiguration.java_ this annotation is used to retrieve the properties in raw form:
```
@Override
public void setDataSourceProperties(@MapFormat(transformation = MapFormat.MapTransformation.FLAT, keyFormat = StringConvention.RAW) Map<String, ?> dsProperties) {
    super.getDataSourceProperties().putAll(dsProperties);
}
```
But that only ever returns the resulting map, when kebap case is used in _application.yml_.

Datasource properties are typically not essential and you only find out that they actually don't work when things go south :disappointed:

sdelamo:
Configuration is by design in kebab case (lowercase and hyphen-separated)

heli80:
But isn't it unintuitive that as soon as you want to retrieve the content of a map with `StringConvention.RAW`, you **MUST** specify the maps' key only in kebab-case (or otherwise you just get an empty map).

That is opposite to all other `StringConventions`, since their content is searched from the default (normalized) catalog. And therefore you can give the maps' key in kebap-case OR camel-case.

I'm thinking about the _principle of least surprise_ here :-)

heli80:
Hi @graemerocher, @sdelamo,
any final thoughts if this works as intended (although surprising)?
If yes, than I would close this PR.
Thx!

sdelamo:
Yes, supplying configuration in kebab case (lowercase and hyphen-separated) is intended. I am not sure it is worth complexity to support extra casing.

Issue #5315, opened 2021-04-26
Worse performance for `application/json` publisher response


dstepanov:
If a reactive publisher is defined as a controller response performance degrades because deserialization needs to be done for every item and additional processing.

The solution is to convert the publisher to a single value publisher if streaming is not required.

For the future version, we should consider streaming the `application-json` response only if the method is annotated with something like `@Streaming`.

dstepanov:
Another example of people using response of Flux and getting bad performance https://dev.to/nix_united/can-micronaut-replace-spring-boot-lets-take-a-look-at-an-example-3nna

graemerocher:
yeah I agree we need to change this

graemerocher:
Triaged as Feature/enhancement. Current ResponseLifecycle still detects Publisher responses, serializes each item with concatMap, and concatenates JSON, so the reported per-element work is a performance/design concern rather than a malformed-response bug. Suggested fix: make non-streaming publisher responses explicit (for example, collapse single-value publishers to the normal body path) and provide an opt-in streaming marker such as @Streaming, while preserving JSON array framing and backpressure. Add a benchmark/test for large Flux/Mono responses and document the trade-off. Keep under next-major consideration.

PR #9425, opened 2023-06-09
Make Publishers.just(Throwable) behave like a Mono

### LINKED ISSUE #9418: Regression in HttpStatusException response from proxied, validated publishers
There seems to be a regression in the response body from Proxied validated endpoints that return Publishers.

I have created a reproducer over here to show the issue: https://github.com/bloidonia/micronaut-publisher-error

But this was first seen in the micronaut-starter in https://github.com/micronaut-projects/micronaut-starter/pull/1764

With Micronaut 4.0.0-M3, an HttpStatusException thrown inside an endpoint returning a raw type (String, etc) results in a correctly wrapped Hateoas response.

However, the same exception thrown in an endpoint returning a Publisher does not generate the same response.

### Steps to Reproduce

Check out https://github.com/bloidonia/micronaut-publisher-error

The setup in this repo is for 3.9.3

There is a controller which has 2 methods.  One returns a `Publisher<String>` and the other returns a raw `String` response.  Both simply throw an HttpStatusExcepton.

Run the test [MicronautPublisherErrorSpec](https://github.com/bloidonia/micronaut-publisher-error/blob/main/src/test/groovy/micronaut/publisher/error/MicronautPublisherErrorSpec.groovy) and it will pass.

Both client methods return a Json response with the encoded error.

#### Try with 4.0.0-M3

Edit [libs.versions.toml](https://github.com/bloidonia/micronaut-publisher-error/blob/main/gradle/libs.versions.toml) to comment out the 3.9.3 dependencies (ie: change it to)

```toml
[versions]
#micronaut = "3.9.3"
#application-plugin = "3.7.9"
micronaut = "4.0.0-M3"
application-plugin = "4.0.0-M4"
```

And edit the imports in the [the specification](https://github.com/bloidonia/micronaut-publisher-error/blob/main/src/test/groovy/micronaut/publisher/error/MicronautPublisherErrorSpec.groovy#L16-L17) to import `jakarta.validation.constraints.NotBlank` instead of `javax.validation.constraints.NotBlank`.

Then run the test again.

The publisher test fails.

### Expected Behaviour

Both responses should result in the same content.

### Actual Behaviour

Publisher based endpoints fail to encode the response as json

### Environment Information

- **Operating System**: OSX
- **Micronaut Version:** 3.9.3/4.0.0-M3
- **JDK Version:** 17

### Example Application

- https://github.com/bloidonia/micronaut-publisher-error

yawkat:
Register JustThrowPublisher as a single publisher, so that it gets treated like a Mono.

JustThrowPublisher is used in PublisherInterceptedMethod.handleException, which in turn is called by micronaut-validation when a Publisher-returning method throws an exception (as in the test case).

Treating JustThrowPublisher as a Mono makes RouteExecutor wait for its completion instead of just using it as the body of a streaming response. This leads to better error handling, including hitting the ExceptionHandler.

Fixes #9418

### DIFF
--- a/core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
+++ b/core-reactive/src/main/java/io/micronaut/core/async/publisher/Publishers.java
@@ -99,6 +99,7 @@ public class Publishers {
         return List.of(
             "io.micronaut.core.async.publisher.CompletableFuturePublisher",
             "io.micronaut.core.async.publisher.Publishers$JustPublisher",
+            "io.micronaut.core.async.publisher.Publishers$JustThrowPublisher",
             "io.reactivex.Single",
             "reactor.core.publisher.Mono",
             "io.reactivex.Maybe",
--- a/core-reactive/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core-reactive/reflect-config.json
+++ b/core-reactive/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core-reactive/reflect-config.json
@@ -1,4 +1,12 @@
 [ {
   "name" : "io.micronaut.core.async.publisher.Publishers",
   "allDeclaredConstructors" : true
-}]
+},
+  {
+    "name" : "io.micronaut.core.async.publisher.Publishers$JustPublisher",
+    "allDeclaredConstructors" : true
+  },
+  {
+    "name" : "io.micronaut.core.async.publisher.Publishers$JustThrowPublisher",
+    "allDeclaredConstructors" : true
+  }]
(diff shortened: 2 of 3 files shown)


dstepanov (approved):

dstepanov:
@yawkat @graemerocher Can we make this change for v4 https://github.com/micronaut-projects/micronaut-core/issues/5315 - Disabling streaming by default

graemerocher:
I think it would be safe to gate the current behaviour behind a configuration flag and have streaming disabled by default

graemerocher:
@dstepanov could you look at doing a PR for that?

yawkat:
ive neglected the whole publisher-based response stack a bit in the changes of the last months.

PR #9436, opened 2023-06-12
Added ability to describe contentType to Part annotation

altro3:
It would be nice to extend the Part annotation and allow the type content to be specified if required. I can't yet suggest how this can be handled at the controller level, but I can say for sure that this will help the micronaut-openapi generator, because now there is no way to correctly describe ,multipart form data without using swagger annotations


graemerocher:
until it is fully implemented don't think we can include it

PR #9451, opened 2023-06-15
Check for optional type in response adaptor

### LINKED ISSUE #9223: JDK HTTP client probably doesn't handle raw strings properly
### Issue description

This is a follow up to https://github.com/micronaut-projects/micronaut-test-resources/pull/255#issuecomment-1545396551

Using a server which returns a response like this:

```
HTTP/1.1 200 OK
Content-Type: application/json
connection: keep-alive
content-length: 5
date: Fri, 12 May 2023 07:26:03 GMT

world
```

and a declarative client using the `micronaut-http-client-jdk`:

```java
@Client
interface SomeClient {
   ...
   Optional<String> getString()
}
```

Client fails with:

> io.micronaut.json.JsonSyntaxException: Unrecognized token 'world': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')

This error goes away when switching from the JDK client to the Netty client. Quoting @yawkat : 

> Yes, types like String (and byte[] and ByteBuffer) are treated specially and don't go through the json encoding/decoding process. It's possible that the jdk client is missing this special handling.

### LINKED ISSUE #9300: JDK Client doesn't handle responses like netty
I had to exclude `io.micronaut.http.server.tck.tests.FilterErrorTest` from https://github.com/micronaut-projects/micronaut-core/pull/9299 as the JDK client fails when the response is

```
HttpResponse.badRequest("from filter exception handler");
```

Somehow, the Netty client knows how to parse this as text/plain, even though the content type is application/json

timyates:
And fallback if json decoding fails

This fixes the StreamTest and FilterErrorTest TCK tests that were failing in #9299

Closes #9223
Closes #9300

### DIFF
--- a/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/HttpResponseAdapter.java
+++ b/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/HttpResponseAdapter.java
@@ -111,6 +111,7 @@ public class HttpResponseAdapter<O> implements HttpResponse<O> {
             if (CharSequence.class.isAssignableFrom(finalArgument.getType())) {
                 Charset charset = contentType.getCharset().orElse(StandardCharsets.UTF_8);
                 var converted = Optional.of(new String(bytes, charset));
+                // If the requested type is an Optional, then we need to wrap the result again
                 return isOptional ? Optional.of(converted) : converted;
             } else if (finalArgument.getType() == byte[].class) {
                 var converted = Optional.of(bytes);
--- a/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/HttpResponseAdapter.java
+++ b/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/HttpResponseAdapter.java
@@ -114,6 +114,7 @@ public class HttpResponseAdapter<O> implements HttpResponse<O> {
                 return isOptional ? Optional.of(converted) : converted;
             } else if (finalArgument.getType() == byte[].class) {
                 var converted = Optional.of(bytes);
+                // If the requested type is an Optional, then we need to wrap the result again
                 return isOptional ? Optional.of(converted) : converted;
             } else {
                 Optional<MediaTypeCodec> foundCodec = mediaTypeCodecRegistry.findCodec(contentType);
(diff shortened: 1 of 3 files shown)


yawkat:
netty certainly does not do this, why is this necessary?

timyates:
@yawkat This was driven by this test

https://github.com/micronaut-projects/micronaut-core/pull/9451/files#diff-de814b22f33b8aacf434b0ffe71b59b3c1f0f74cc5458763e4b36b79810907e8

I wonder if this test goes too far 🤔 Maybe the Client `string` call shouldn't be optional 🤔

timyates:
But even ignoring those tests, and concentrating on the `Optional` one (inspired by @melix's issue https://github.com/micronaut-projects/micronaut-core/issues/9223), we end up making a call to 

```
conversionService.convert(bytes, ConversionContext.of(type))
```

Where bytes is `Hello World` and type is `Optional`

I cannot see how this shoudl then be handled (as the converter says it's UNCONVERTIBLE)

yawkat:
replicate this: https://github.com/micronaut-projects/micronaut-core/blob/37874c634202233f35b7c9376a5edfd5d49861f2/http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java#L170-L171

timyates:
Like this?

https://github.com/micronaut-projects/micronaut-core/pull/9451/commits/202b6e5676dbd5abb1bf26e8dc2937c8c9d397ae

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/HttpResponseAdapter.java
```diff
@@ -104,59 +104,39 @@ public <T> Optional<T> getBody(Argument<T> type) {
...
+        final Argument finalArgument = isOptional ? type.getFirstTypeVariable().orElse(type) : type;
+
+        if (mediaTypeCodecRegistry != null && contentType != null) {
+            if (CharSequence.class.isAssignableFrom(finalArgument.getType())) {
                 Charset charset = contentType.getCharset().orElse(StandardCharsets.UTF_8);
-                return maybeWrap(Optional.of(new String(bytes, charset)), isOptional(type));
-            } else if (isByteArray(type)) {
-                return maybeWrap(Optional.of(bytes), isOptional(type));
+                var converted = Optional.of(new String(bytes, charset));
+                return isOptional ? Optional.of(converted) : converted;
```
yawkat, line 115: the double optional is a bit weird (but correct), maybe add a comment

yawkat (approved):

PR #9456, opened 2023-06-16
Fix all-opens support in ksp

### LINKED ISSUE #9426: Cannot apply AOP advice to final class
### Expected Behavior

I tried to upgrade my example project to 4.0, the example used R2dbc/Postgres, Kotlin Coroutines.

When compiling it, it throws exception, even I add the following configuration in Gradle build script explicitly.

```kotlin 
allOpen{
  annotation("~.Controller")
}
```

### Actual Behaviour

```bash
e: [ksp] D:/hantsylabs/micronaut-sandbox/data-r2dbc-kotlin-co/src/main/kotlin/com/example/PostController.kt:18: Cannot apply AOP advice to final class. Class must be made non-final to support proxying: com.example.PostController
e: Error occurred in KSP, check log for detail
```

### Steps To Reproduce

_No response_

### Environment Information

Windows 10 Pro, Java 17

### Example Application

https://github.com/hantsy/micronaut-sandbox/tree/master/data-r2dbc-kotlin-co

### Version

4.0.0-M3

yawkat:
ksp does not currently see when classes or methods are opened by the all-opens plugin. This patch assumes these elements are open if the annotations that all-opens considers by default are present.

There's no test in this PR, I have one in micronaut-starter. However from the tests that I had to disable in FinalModifierSpec, I'm pretty sure this will work.

Fixes #9426

### DIFF
--- a/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
+++ b/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
@@ -40,6 +40,7 @@ import io.micronaut.core.annotation.AnnotationMetadata
 import io.micronaut.inject.ast.ClassElement
 import io.micronaut.inject.ast.Element
 import io.micronaut.inject.ast.ElementModifier
+import io.micronaut.inject.ast.MemberElement
 import io.micronaut.inject.ast.PrimitiveElement
 import io.micronaut.inject.ast.WildcardElement
 import io.micronaut.inject.ast.annotation.AbstractAnnotationElement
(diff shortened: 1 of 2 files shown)


dstepanov:
> the annotations that all-opens considers by default are present

Which annotations are those? Does it recognize our annotations?

yawkat:
yea https://github.com/JetBrains/kotlin/blob/8c2f1acc7ecf1d3b37243dc4f7b748caa4e2aefa/plugins/allopen/allopen.common/src/org/jetbrains/kotlin/allopen/AllOpenPluginNames.kt#L22

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
```diff
@@ -94,7 +113,18 @@ internal abstract class AbstractKotlinElement<T : KotlinNativeElement>(
...
+        if (annotatedInfo.modifiers.contains(Modifier.FINAL)) {
+            true
+        } else if (annotatedInfo.isOpen()) {
+            false
+        } else {
+            // ksp does not see when all-open opens up these classes
+            // https://github.com/micronaut-projects/micronaut-core/issues/9426
+            // this logic is similar to what all-opens does.
+            val cl = annotatedInfo as? KSClassDeclaration ?: annotatedInfo.parentDeclaration as? KSClassDeclaration
+            cl == null || !visitorContext.elementFactory.newClassElement(cl)
```
dstepanov: Why do you need to create a new element? You should be able to access the annotation from the superclass `AbstractAnnotationElement`

sdelamo (approved):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
yawkat: it's not always the right annotations. e.g. when the ClassElement is loaded from a field, it will contain the annotations of the field too.

dstepanov: I don’t understand what are you talking about. If the class is annotated with one of those annotations it should not indicate final class. I don’t see anything that prevents to use ordinary class’ annotation metadata

yawkat: for example `test final modifier on factory with AOP advice doesn't compile` fails if i use `this` instead of a new class element, because the method is annotated with `@Mutating` so the return type ClassElement inherits that annotation from the method even though the class itself does not have the annotation.

yawkat: to be clear in that test, with `this`, the processor incorrectly thinks the all-open plugin would open up `MyBean.`

dstepanov: I will check

dstepanov: Corrected

dstepanov (approved):

graemerocher (approved):

PR #9459, opened 2023-06-16
Introduce `UnsafeBeanInstantiationIntrospection`

dstepanov:


graemerocher (approved):

core/src/main/java/io/micronaut/core/beans/AbstractBeanIntrospection.java
```diff
@@ -1,356 +0,0 @@
...
-/**
- * Abstract implementation of the {@link BeanIntrospection} interface. This class is subclasses at compilation time by generated byte code and should not be used directly.
- *
- * @param <T> The generic type
- * @author graemerocher
- * @since 1.1
- */
-@UsedByGeneratedCode
-@Internal
-public abstract class AbstractBeanIntrospection<T> implements BeanIntrospection<T> {
```
sdelamo, line 43: Is this file not needed any more?

dstepanov: Not needed anymore

graemerocher: no the introspections have used `AbstractInitializableBeanIntrospection` for a while now

PR #9469, opened 2023-06-19
Update corsAllowedMethods.adoc

Omasiri:
It looks like the documentation means 'do' not "don't" for this case.


wetted:
No, the documentation is correct. When CORS is enabled any method is allowed by default. `allowed-methods` is used to constrain in a more fine-grained manner.

PR #9470, opened 2023-06-19
Update corsAllowedHeaders.adoc

Omasiri:
To allow request header for a given request, add the 'allowed-headers' attribute.

### DIFF
--- a/src/main/docs/guide/httpServer/serverConfiguration/cors/corsAllowedHeaders.adoc
+++ b/src/main/docs/guide/httpServer/serverConfiguration/cors/corsAllowedHeaders.adoc
@@ -1,4 +1,4 @@
-To allow any request header for a given configuration, don't include the `allowed-headers` key in your configuration.
+To allow any request header for a given configuration, include the `allowed-headers` key in your configuration.
 
 For multiple allowed headers, set the `allowed-headers` key of the configuration to a list of strings.
 


sdelamo:
> To allow request header for a given request, add the 'allowed-headers' attribute.

I think the docs were correct. Are they not?

wetted:
@sdelamo 

Offhand, I believe this change is correct. CORS safe-listed request headers are always allowed. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

Our own tests assert `allowed-headers` is required otherwise. At least it is for the `@CrossOrigin` annotation, and that should be consistent. https://github.com/micronaut-projects/micronaut-core/blob/25b89a8a5bcbaea5d7dc1dfcc4258431da0dc66c/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java#LL148C1-L176C1

wetted:
We are not consistent. Configuration-based version defaults `allowed-headers` to "*" while annotation-based defaults `allowed-headers` to "" (empty string). We should make it consistent, but which is correct? @yawkat ?

yawkat:
We should default to no allowed headers.

sdelamo:
@wetted can you do the change so that both configuration and annotation default to  no allowed headers. Update docs (see this PR) and document it as a breaking change.

wetted:
The micronaut team agreed we should make the behavior of the annotation-based CORS match the configuration based version, which means when CORS is turned on then all headers and all origins are enabled by default, and the options for `allowed-headers` and `allowed-origins` are used to constrain those in a more fine-grained way. 

That means the docs are correct as is. I'll create a separate issue to make sure that the annotation-based CORS has the same behavior as configuration.

PR #9472, opened 2023-06-20
Update to netty 4.1.94.Final

yawkat:
Apparently includes a fix for CVE-2023-34462. Details aren't public yet, though the changelog mentions SniHandler which we don't use, so we are probably not affected.


sdelamo (approved):

sdelamo:
graal failed due to` Terminating due to java.lang.OutOfMemoryError: Java heap space`

PR #9478, opened 2023-06-21
Deprecate some unused classes

yawkat:
Deprecate and make internal some unused classes. I will remove them in 4.1, this patch prepares for that without breaking compat.


dstepanov:
Why not remove them now?

sdelamo (approved):

sdelamo:
I agree with @dstepanov, let's remove them now if they are unused.

yawkat:
i dont want to break anything in the RC if they turn out to be used somewhere. there's no rush.

PR #9479, opened 2023-06-21
Method binding with no @Body TCK tests, updated docs and docs examples.

### LINKED ISSUE #9388: Documentation needs to be updated now that @Body is required
The Variable resolution documentation says: 

> If the request can have a body and no @Body is defined then try to parse the body (either JSON or form data) and bind the method arguments from the body.

However, now `@Body` is required.

wetted:
Add http-tck tests for json and form data body and controller with no `@Body` annotation.
Update binding docs, and add docs examples for java, groovy, kotlin.

closes #9388

### DIFF
--- a/test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/binding/PointController.kt
+++ b/test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/binding/PointController.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.docs.server.binding
+
+import io.micronaut.context.annotation.Requires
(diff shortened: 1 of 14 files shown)


sdelamo (changes requested):
Can you use [Kotlin Data Classes](https://kotlinlang.org/docs/data-classes.html) for the `Point` class in Kotlin ?

yawkat (approved):

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/binding/PointController.kt
```diff
@@ -0,0 +1,44 @@
...
+import io.micronaut.http.annotation.Status
+
+@Requires(property = "spec.name", value = "PointControllerTest")
+// tag::class[]
+@Controller("/point")
+class PointController {
+
+    @Post(uri = "/no-body-json")
+    @Status(HttpStatus.CREATED)
+    fun noBodyJson(x: Int, y: Int): Point { // (1)
```
sdelamo: I think you could change:

```
 fun noBodyJson(x: Int, y: Int): Point { // (1)
        return Point(x, y)
    }
```

to

```
 fun noBodyJson(x: Int, y: Int) = Point(x,y)
```

which is more Kotlin idiomatic.

sdelamo: same in the other method and in the other controller

sdelamo (changes requested):

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/binding/PointController.kt
wetted: Cool. Thanks for the suggestion. I should probably spend some more time in the Kotlin Primer. :)

sdelamo (approved):

PR #9483, opened 2023-06-22
Replace HttpStreamsHandler for netty HTTP client

### LINKED ISSUE #9652: HttpClient jsonStream: clients might not get/know about a server exception
### Issue description

The test `StreamRequestSpec#"test json stream post request with POJOs flowable error"` seems to be non-deterministic, I managed to get an error after running the whole file few times.

The gist of the test is to expect an exception when the server receive an error (on Book 3), but the client doesn't fail (i.e. the test fails) if the server closes the connection (`PipelineServerHandler$StreamingOutboundHandler#onError`) exactly when a valid json buffer has been transferred.

To consistently reproduce this behaviour the channel content can be flushed before closing in `PipelineServerHandler$StreamingOutboundHandler#onError` method (see image below, btw is this a change that makes sense to commit?)

With this change the client always gets valid json pieces, all Books before number 3 (Books 0, 1, 2) and continues happily without knowing about the error server side on Book 3, so the test fails because no exceptions are thrown.

I don't know how it was in `3.9.x` but it seems to be something missing in the jsonStream protocol or in the client implementation itself, or in the server? just closing the channel without notifying any error.

image

yawkat:
Just like #9128 did for the server, this PR replaces the HttpStreamsHandler and HandlerPublisher for the client. Instead, incoming HttpRequest and HttpContent objects are handled directly and transformed to the needed types.

This allows full removal of the streams handler. Some of the functionality is reimplemented in ReactiveClientReader and ReactiveClientWriter, but these are much simpler classes and only need to be used for streaming requests / responses.

This implementation is not quite as optimized as PipeliningServerHandler, there's no read coalescing for example, and it uses multiple handlers where a more complex implementation could do with one.


graemerocher:
it would be good to re-enable binary compatible checks and rebase this PR so we get a clearer idea of what the binary incompatible changes are

yawkat:
with #9478 there should be none, it's only internal classes now.

graemerocher:
@yawkat please rebase

yawkat:
I've added a change to fix the brittle test reported in #9652.

http-client-tck/src/main/java/io/micronaut/http/client/tck/tests/ContinueTest.java
```diff
@@ -0,0 +1,125 @@
...
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED),
+            (server, request) -> AssertionUtils.assertDoesNotThrow(server, request,
+                HttpResponseAssertion.builder()
+                    .status(HttpStatus.OK)
+                    .body(BodyAssertion.builder().body("bar").equals())
+                    .build())
+        );
+    }
+
+    @Controller("/continue")
```
sdelamo, line 112: Please, annotate the controller with `@Requires(property = spec.name` to avoid bean pollution between tests

graemerocher (approved):

http-netty/src/test/groovy/io/micronaut/http/netty/reactive/HandlerPublisherSpec.groovy
```diff
@@ -1,88 +0,0 @@
-package io.micronaut.http.netty.reactive
-
-import io.netty.channel.ChannelDuplexHandler
-import io.netty.channel.ChannelHandlerContext
-import io.netty.channel.embedded.EmbeddedChannel
-import org.reactivestreams.Subscriber
-import org.reactivestreams.Subscription
-import spock.lang.Specification
-
-class HandlerPublisherSpec extends Specification {
```
sdelamo, line 10: do we test this somewhere else?

http-netty/src/test/groovy/io/micronaut/http/netty/stream/HttpStreamsClientHandlerSpec.groovy
```diff
@@ -1,66 +0,0 @@
...
-import io.netty.handler.codec.http.LastHttpContent
-import reactor.core.publisher.Flux
-import spock.lang.Issue
-import spock.lang.Specification
-
-import java.nio.charset.StandardCharsets
-
-class HttpStreamsClientHandlerSpec extends Specification {
-    @Issue('https://github.com/micronaut-projects/micronaut-tracing/issues/316')
-    def 'out of order write'() {
```
sdelamo, line 24: do we test this somewhere else?

sdelamo (changes requested):

http-netty/src/test/groovy/io/micronaut/http/netty/reactive/HandlerPublisherSpec.groovy
yawkat: HandlerPublisher is gone now, the reactive publisher doesn't buffer so it's not susceptible

http-netty/src/test/groovy/io/micronaut/http/netty/stream/HttpStreamsClientHandlerSpec.groovy
yawkat: content isnt delayed anymore so the test case is obsolete

sdelamo (approved):

PR #9491, opened 2023-06-23
Skip not needed AST elements when a name predicate exists

dstepanov:


inject-java-helper/build.gradle
```diff
@@ -0,0 +1,10 @@
+plugins {
+    id "io.micronaut.build.internal.convention-library"
```
graemerocher: We don't want to publish this, should be `id "io.micronaut.build.internal.convention-test-library"`

graemerocher (changes requested):

graemerocher (approved):

PR #9492, opened 2023-06-23
Add a way to extract the value from `JsonNode` and construct it from it

dstepanov:
It would be nice to have them, otherwise it's a bit complicated to do

### DIFF
--- a/json-core/src/main/java/io/micronaut/json/tree/JsonArray.java
+++ b/json-core/src/main/java/io/micronaut/json/tree/JsonArray.java
@@ -36,6 +36,11 @@ public class JsonArray extends JsonContainer {
     }
 
     @Override
+    public Object getValue() {
+        return values.stream().map(JsonNode::getValue).toList();
+    }
+
+    @Override
     public int size() {
         return values.size();
     }
--- a/json-core/src/main/java/io/micronaut/json/tree/JsonBoolean.java
+++ b/json-core/src/main/java/io/micronaut/json/tree/JsonBoolean.java
@@ -37,6 +37,11 @@ final class JsonBoolean extends JsonScalar {
         return value ? TRUE : FALSE;
     }
 
+    @Override
+    public Object getValue() {
+        return value;
+    }
+
     @NonNull
     @Override
     public String coerceStringValue() {
--- a/json-core/src/main/java/io/micronaut/json/tree/JsonNode.java
+++ b/json-core/src/main/java/io/micronaut/json/tree/JsonNode.java
@@ -76,6 +76,7 @@ public abstract class JsonNode {
         }
         throw new IllegalStateException("Unrecognized value: " + value);
     }
+
     /**
      * @return The singleton node representing {@code null}.
      */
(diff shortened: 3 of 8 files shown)


yawkat:
whats the point of this? we already have decodeArbitrary

graemerocher:
I'm wondering that too

dstepanov:
Micronaut Data document support is based on Micronaut Serialization. To convert an entity to a list of attributes I'm using this:
```java
    private <T> JsonNode persistAsJsonNode(T object) {
        try {
            JsonNodeEncoder encoder = JsonNodeEncoder.create();
            Argument<Object> type = (Argument<Object>) Argument.of(object.getClass());
            Serializer.EncoderContext encoderContext = serdeRegistry.newEncoderContext(null);
            final Serializer<Object> serializer = encoderContext.findSerializer(type).createSpecific(encoderContext, type);
            serializer.serialize(encoder, encoderContext, type, object);
            return encoder.getCompletedValue();
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
    }

    private <T> T readFromJsonNode(JsonNode tree, Argument<T> type) {
        try {
            Deserializer.DecoderContext decoderContext = serdeRegistry.newDecoderContext(null);
            final Deserializer<? extends T> deserializer = decoderContext.findDeserializer(type).createSpecific(decoderContext, type);
            return deserializer.deserialize(JsonNodeDecoder.create(tree), decoderContext, type);
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
    }
```

And I'm working with `JsonNode` tree.

yawkat:
You can use JsonMapper.readValueFromTree(tree, Object.class) to go from node to Map

dstepanov:
Is there a way to do it without injecting `JsonMapper`? Imho the methods are still useful, maybe `from` method is a bit controversial

yawkat:
jsonnodedecoder.readArbitrary i guess

yawkat:
JsonNodeDecoder and JsonNodeEncoder are internal in serialization though and i recently changed them with the limits PR

dstepanov:
Looks like `toArbitrary` can be converted to use my changes

dstepanov:
```java
private static Object toArbitrary(JsonNode node) {
        if (node.isNull()) {
            return null;
        } else if (node.isNumber()) {
            return node.getNumberValue();
        } else if (node.isBoolean()) {
            return node.getBooleanValue();
        } else if (node.isString()) {
            return node.getStringValue();
        } else if (node.isArray()) {
            List<Object> transformed = new ArrayList<>();
            for (JsonNode value : node.values()) {
                transformed.add(toArbitrary(value));
            }
            return transformed;
        } else if (node.isObject()) {
            Map<String, Object> transformed = new LinkedHashMap<>();
            for (Map.Entry<String, JsonNode> entry : node.entries()) {
                transformed.put(entry.getKey(), toArbitrary(entry.getValue()));
            }
            return transformed;
        } else {
            throw new AssertionError(node);
        }
    }
```

This can be simply replaced with `.getValue()`

yawkat:
yea i just dont really like exposing this

yawkat:
generally discouraged to first go to one tree type and then to another

dstepanov:
@yawkat You don't like these changes at all?

yawkat:
im not a fan, but i dont think it's a big deal either.

yawkat (approved):

PR #9495, opened 2023-06-24
Support for builders in bean introspections and bean mapping

graemerocher:
## Builders

Adds a new `BeanIntrospection.builder()` method and corresponding annotation members in `@Introspected(builder=..)` to allowing defining a builder for a bean introspection.

The `builder()` method is flexible enough to deal with all the different cases of instantiation (builders, records, javabeans etc.), however consumers may want to optimise direct instantiation so a new method `hasBuilder()` is added to check if an introspection has a backing builder.

This PR will allow us to support `@JsonDeserialize(builder=..)` in Micronaut Serialization.

## Bean Mapping 

Currently if you need to map one type to another you have to add a third party library like MapStruct. This is very typical with DTOs, command objects etc.

Adds a new annotation `@Mapper` that allows using introduction advice to generate mappers from one type to another. Uses the builder functionality underneath the hood and expressions to allow customisation.


context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
```diff
@@ -0,0 +1,305 @@
...
+        Collection<BeanProperty<I, Object>> beanProperties = left.getBeanProperties();
+        BeanMapper.MapStrategy.ConflictStrategy conflictStrategy = mapStrategy.conflictStrategy();
+        if (conflictStrategy == null || conflictStrategy == BeanMapper.MapStrategy.ConflictStrategy.ERROR) {
+
+                for (BeanProperty<I, Object> leftProperty : beanProperties) {
+                    try {
+                        BeanProperty<O, Object> rightProperty =
+                                right.getRequiredProperty(leftProperty.getName(), leftProperty.getType());
+                        if (!rightProperty.isReadOnly()) {
+                            ((UnsafeBeanProperty<O, Object>) rightProperty).setUnsafe(output, leftProperty.get(input));
```
dstepanov: Do we support immutable beans?

dstepanov: Those that can be mutated by creating a new instance like records

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    @Override
+    public <O> O map(Map<String, Object> input, BeanMapper.MapStrategy mapStrategy, BeanIntrospection<O> outputIntrospection) {
+        if (mapStrategy == null) {
+            mapStrategy = BeanMapper.MapStrategy.DEFAULT;
+        }
+        BeanMapper.MapStrategy.ConflictStrategy conflictStrategy = mapStrategy.conflictStrategy();
+        @NonNull Argument<?>[] constructorArguments = outputIntrospection.getConstructorArguments();
+        if (constructorArguments.length == 0) {
```
dstepanov: I think it would be nice to extract this method into `BeanIntrospection`. There are cases where we have a map or list of values and we want to construct the bean no carrying if properties are coming to the constructor or setter/field.

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
```diff
@@ -0,0 +1,305 @@
...
+            O output = outputIntrospection.instantiate();
+            return map(input, output, mapStrategy, outputIntrospection);
+        } else {
+            Objects.requireNonNull(input, "Input cannot be null");
+            // constructor binding
+            Object[] parameters = new Object[constructorArguments.length];
+            for (int i = 0; i < constructorArguments.length; i++) {
+                Argument<?> constructorArgument = constructorArguments[i];
+                String key = constructorArgument.getName();
+                if (input.containsKey(key)) {
```
dstepanov, line 196: I would just use `get` and not null check

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
```diff
@@ -0,0 +1,305 @@
...
+                    mapOrError(output, right, key, value, isNonNull);
+                } else {
+                    mapEntry(output, conflictStrategy, right, key, value, isNonNull);
+                }
+            });
+        }
+        return output;
+    }
+
+    private static <O> void mapEntry(O output, BeanMapper.MapStrategy.ConflictStrategy conflictStrategy, BeanIntrospection<O> right, String key, Object value, boolean isNonNull) {
```
dstepanov: We might want to support `with` like methods as well

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
graemerocher: yes records are supported

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
graemerocher: yes I agree, I actually have in mind a new API because I want to add generic support for the `Builder` pattern as well and if we had `BeanIntrospection.builder() -> BeanBuilder` we could abstract constructor arguments.

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
graemerocher: Agreed

graemerocher:
Converting to draft since I have more ideas here and the feedback from @dstepanov is good

context/src/main/java/io/micronaut/runtime/beans/DefaultIntrospectionBeanMapHandler.java
dstepanov: I think you should use `withValueUnsafe`

graemerocher:
Thing we may want to do:

* Automatically support `groovy.transform.builder.Builder` by mappings annotations
* Automatically support `lombok.Builder` by mappings annotations

core/src/main/java/io/micronaut/core/beans/BeanIntrospection.java
```diff
@@ -243,4 +277,75 @@ default int propertyIndexOf(@NonNull String name) {
...
+     * @since 4.1.0
+     */
+    interface Builder<T> {
+        /**
+         * All the arguments possible for creating the instance.
+         *
+         * @param <A> The generic type
+         * @return The arguments
+         */
+        <A> @NonNull Argument<A>[] getArguments();
```
yawkat:
```suggestion
        @NonNull Argument<?>[] getArguments();
```

core/src/main/java/io/micronaut/core/beans/BeanIntrospection.java
```diff
@@ -243,4 +277,75 @@ default int propertyIndexOf(@NonNull String name) {
...
+
+        /**
+         * Create the instance with the given value for the given property.
+         *
+         * @param name     The name of the argument
+         * @param value    The value
+         * @param <A>      The generic type of the value
+         * @return This builder
+         */
+        <A> @NonNull Builder<T> with(@NonNull String name, @Nullable A value);
```
yawkat:
```suggestion
        @NonNull Builder<T> with(@NonNull String name, @Nullable Object value);
```

inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
```diff
@@ -376,6 +396,257 @@ public String toString() {
...
+                if (constructorLength == params.length) {
+                    return introspection.instantiateInternal(params);
+                } else {
+                    Object[] constructorParams = Arrays.copyOfRange(params, 0, constructorLength);
+                    B bean = introspection.instantiateInternal(constructorParams);
+                    UnsafeBeanProperty<Object, Object>[] writeableProperties = builderData.writeableProperties;
+                    if (writeableProperties != null) {
+                        for (int i = constructorLength; i < builderData.arguments.length; i++) {
+                            UnsafeBeanProperty<Object, Object> property = writeableProperties[i - constructorLength];
+                            property.setUnsafe(bean, params[i]);
```
yawkat: should this really be unsafe? `with` could set an incompatible value

graemerocher: they're already validated in the `with` methods of the `Builder` interface

yawkat: `with(int index, Argument<A> argument, A value)` does not check the type

graemerocher: Good catch

core-processor/src/main/java/io/micronaut/inject/beans/visitor/MapperVisitor.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void visitMethod(MethodElement element, VisitorContext context) {
+        if (element.hasDeclaredAnnotation(Mapper.class)) {
+            if (element.isAbstract()) {
+
+                ClassElement toType = element.getGenericReturnType();
+                @NonNull ParameterElement[] parameters = element.getParameters();
+                if (parameters.length == 1) {
+                    if (toType.isVoid()) {
+                        context.fail("A void return type is not permitted for a mapper", element);
```
dstepanov: You can use `ProcessingException` instead of `context.fail`

inject/src/main/java/io/micronaut/context/annotation/Mapper.java
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * An annotation that can be used on abstract methods that define a return type and exactly a single argument.
+ *
+ * <p>Inspired by similar frameworks like MapStruct but internally uses the {@link io.micronaut.core.beans.BeanIntrospection} model.</p>
+ *
+ * @author Graeme Rocher
+ * @since 4.1.0
+ */
+@Retention(RetentionPolicy.RUNTIME)
```
dstepanov, line 34: Do we need this annotation at runtime?

dstepanov (approved):

context/src/main/java/io/micronaut/runtime/beans/MapperIntroduction.java
```diff
@@ -0,0 +1,360 @@
...
+                }
+
+                Object from = values.get(Mapper.Mapping.MEMBER_FROM);
+                Object condition = values.get(Mapper.Mapping.MEMBER_CONDITION);
+                ArgumentConversionContext<?> finalConversionContext = conversionContext;
+                if (from instanceof EvaluatedExpression evaluatedExpression) {
+                    customMappers.put(to, (expressionEvaluationContext ->
+                        (object, builder) -> {
+                            ExpressionEvaluationContext evaluationContext = (ExpressionEvaluationContext) expressionEvaluationContext;
+                            if (condition instanceof EvaluatedExpression conditionExpression) {
```
yawkat: what else can condition be here? null?

yawkat (approved):

core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
```diff
@@ -192,11 +261,71 @@ private void processElement(boolean metadata,
...
+        AnnotationMetadata builderMetadata) {
+        if (builderMetadata == null) {
+            builderMetadata = AnnotationMetadata.EMPTY_METADATA;
+        }
+        if (!isIntrospected(context, builderType)) {
+            ElementQuery<MethodElement> buildMethodQuery = ElementQuery
+                .ALL_METHODS
+                .onlyAccessible(classToBuild)
+                .onlyInstance()
+                .filter(m -> m.getGenericReturnType().getName().equals(classToBuild.getName()) && !m.hasParameters());
```
andriy-dmytruk: Sometimes the `.build()` method can have parameters, which are required for building.
Like this: https://github.com/oracle/oci-java-sdk/blob/bcb3737a9c598827254209249f24ed4a649206f5/bmc-apigateway/src/main/java/com/oracle/bmc/apigateway/ApiGatewayClient.java#L130-L133
But, I suppose this is an edge case, and not sure if good practice. Also, I think the linked class does not need to be introspected anyway 😄.

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
```diff
@@ -594,6 +608,14 @@ private void writeIntrospectionClass(ClassWriterOutputVisitor classWriterOutputV
...
     }
 
+    private void writeBooleanMethod(ClassWriter classWriter, String methodName, boolean isBuildable) {
+        GeneratorAdapter isBuildableMethod = startPublicMethodZeroArgs(classWriter, boolean.class, methodName);
+        isBuildableMethod.push(isBuildable);
+        isBuildableMethod.returnValue();
+        isBuildableMethod.visitMaxs(2, 1);
+        isBuildableMethod.endMethod();
+    }
+
```
andriy-dmytruk: Rename the argument from `isBuildable`:
```suggestion
    private void writeBooleanMethod(ClassWriter classWriter, String methodName, boolean returnValue) {
        GeneratorAdapter booleanMethod = startPublicMethodZeroArgs(classWriter, boolean.class, methodName);
        booleanMethod.push(returnValue);
        booleanMethod.returnValue();
        booleanMethod.visitMaxs(2, 1);
        booleanMethod.endMethod();
    }

```

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
```diff
@@ -574,14 +575,27 @@ private void writeIntrospectionClass(ClassWriterOutputVisitor classWriterOutputV
...
-        if (constructor != null && ArrayUtils.isNotEmpty(constructor.getParameters())) {
+
+        if (constructor != null) {
+            if (ArrayUtils.isEmpty(constructor.getParameters())) {
+                writeInstantiateMethod(classWriter, constructor, "instantiate");
+            }
             writeInstantiateMethod(classWriter, constructor, "instantiateInternal", Object[].class);
+            writeBooleanMethod(classWriter, METHOD_IS_BUILDABLE, true);
+        } else if (defaultConstructor == null) {
+            writeBooleanMethod(classWriter, METHOD_IS_BUILDABLE, hasBuilder);
```
andriy-dmytruk, line 595: Would be cool if we could also write the builder as a constructor (write the `instantiate` method or another method that would take arguments, but would construct object using the builder). This way serde and other modules that use the API would need less changes. It could also eliminate some overhead that is in the `BeanIntrospection.builder`. But I guess it would be hard and we would need to decide which one to write if there is both builder and constructor.

timyates (approved):

core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
graemerocher: Add support for this, thanks for the feedback

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
graemerocher: so I implemented this because it is probably useful for compatibility however, I believe that we will want to optimise this to use the `builder()` API in serde because otherwise the JSON will need to be buffered // cc @yawkat

inject/src/main/java/io/micronaut/context/annotation/Mapper.java
graemerocher: I believe we do yes

andriy-dmytruk (approved):

src/main/docs/guide/ioc/introspection.adoc
```diff
@@ -116,6 +116,18 @@ snippet::io.micronaut.docs.ioc.beans.Business[tags="class"]
 
 TIP: There can be multiple "creator" methods annotated. If there is one without arguments, it will be the default construction method. The first method with arguments will be used as the primary construction method.
 
+==== Builders
+
+If a type can only be constructed via the builder pattern then you can use the `builder` member of the ann:core.annotation.Introspected[] annotation to generate a dynamic builder. For example given this class:
+
+snippet::io.micronaut.docs.ioc.builders.Person[tags="class"]
```
sdelamo, line 123: it misses Groovy  `Person`.

CleanShot 2023-07-24 at 11 05 57@2x

src/main/docs/guide/ioc/introspection.adoc
```diff
@@ -116,6 +116,18 @@ snippet::io.micronaut.docs.ioc.beans.Business[tags="class"]
 
 TIP: There can be multiple "creator" methods annotated. If there is one without arguments, it will be the default construction method. The first method with arguments will be used as the primary construction method.
 
+==== Builders
+
+If a type can only be constructed via the builder pattern then you can use the `builder` member of the ann:core.annotation.Introspected[] annotation to generate a dynamic builder. For example given this class:
```
sdelamo, line 121: We say: 

> If a type can only be constructed via the builder pattern

However, `Person` in the examples can be constructed without the builder pattern. Since a Java `record`  cannot hide its canonical constructor I think it is better to use a Java POJO in this particular example.

test-suite-groovy/src/test/groovy/io/micronaut/docs/ioc/builders/Person.groovy
```diff
@@ -0,0 +1,31 @@
+package io.micronaut.docs.ioc.builders
+
+import groovy.transform.Canonical
+import io.micronaut.core.annotation.Introspected
+
+
+@Canonical
```
sdelamo: It misses opening and closing `class` asciidoc tag.

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
+Since 4.1.x the ann:context.annotation.Mapper[] annotation can be used on any abstract method to automatically create a mapping between one type and another.
+
+Inspired by similar functionality in libraries like https://mapstruct.org[Map Struct], a `Mapper` uses the <<introspections, Bean Introspection>> and <<evaluatedExpressions, Expressions>> features built into Micronaut that are already reflection free.
```
sdelamo:
```suggestion
Inspired by similar functionality in libraries like https://mapstruct.org[Map Struct], a `Mapper` uses the <<introspections, Bean Introspection>> and <<evaluatedExpressions, Expressions>> features, built into the Micronaut Framework, which are already reflection free.
```

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+For example given the following type:
```
sdelamo:
```suggestion
For example, given the following type:
```

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+snippet::io.micronaut.docs.ioc.mappers.Product[tags="class"]
+
+It is common to want to alter the representation of this type in say HTTP responses. For example consider this response type:
```
sdelamo:
```suggestion
It is common to want to alter this type's representation in HTTP responses. For example, consider this response type:
```

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+snippet::io.micronaut.docs.ioc.mappers.ProductDTO[tags="class"]
+
+Here the `price` property is of a different type and an extra property exists called `distributor`. You could write manual logic to deal the mapping and these differences, or you could define a mapper:
+
+snippet::io.micronaut.docs.ioc.mappers.ProductMappers[tags="class"]
+
+The above example is an interface that will be implemented for you at compilation type. Each abstract method can define either a single ann:context.annotation.Mapper[] annotation or one or many `@Mapping` annotations to define how properties map onto the target type.
```
sdelamo:
```suggestion
The Micronaut compiler implements the interface in the above example for you at compilation type.  Each abstract method can define a single ann:context.annotation.Mapper[@Mapper] annotation or one or many ann:context.annotation.Mapper.Mapping[@Mapping] annotations to define how properties map onto the target type.
```

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+The `from` member can be used to define either a property name on the source type or a <<evaluatedExpressions, expression>> that reads values from the method argument and transforms them in whatever way you choose including invoking other methods of the instance.
```
sdelamo:
```suggestion
The `from` member can be used to define either a property name on the source type or an <<evaluatedExpressions, expression>> that reads values from the method argument and transforms them in whatever way you choose, including invoking other methods of the instance.
```

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+NOTE: A `@Mapping` definition is only needed if you need to apply a transformation for the mapping to be successful, other properties will be automatically mapped and converted.
```
sdelamo:
```suggestion
NOTE: A `@Mapping` definition is only needed if you need to apply a transformation for the mapping to be successful. Other properties will be automatically mapped and converted.
```

src/main/docs/guide/ioc/beanMappers.adoc
```diff
@@ -0,0 +1,29 @@
...
+
+The `ProductMappers` type can then be injected or retrieved from the context and the `toProductDTO` method used to map from the `Product` type to the `ProductDTO` type:
```
sdelamo:
```suggestion
You can retrieve from the context or inject a bean of type `ProductMappers`. Then, you can use the `toProductDTO` method  to map from the Product type to the ProductDTO type:
```

inject/src/main/java/io/micronaut/inject/annotation/EvaluatedAnnotationMetadata.java
```diff
@@ -48,6 +49,13 @@ private EvaluatedAnnotationMetadata(AnnotationMetadata targetMetadata,
         this.evaluationContext = evaluationContext;
     }
 
+    /**
+     * @return The evaluation context.
+     */
```
sdelamo, line 55:
```suggestion
     * @since 4.1.0
     */
```

core/src/main/java/io/micronaut/core/beans/BeanIntrospection.java
```diff
@@ -61,6 +65,33 @@ public interface BeanIntrospection<T> extends AnnotationMetadataDelegate, BeanIn
...
     @NonNull Collection<BeanProperty<T, Object>> getIndexedProperties(@NonNull Class<? extends Annotation> annotationType);
 
+    /**
+     * @return Obtain a builder for this introspection.
+     * @since 4.1.0
+     */
+    @NonNull Builder<T> builder();
+
+    /**
+     * Can the introspection be constructor with {@link #builder()}.
```
sdelamo:
```suggestion
     * Can the introspection be constructed with {@link #builder()}.
```

core/src/main/java/io/micronaut/core/annotation/Introspected.java
```diff
@@ -185,6 +185,49 @@
...
+         * @return The builder method.
+         */
+        String builderMethod() default "";
+
+        /**
+         * The name of the method that builds the instance.
+         *
+         * @return The method name.
+         */
+        String creatorMethod() default "build";
```
sdelamo, line 223: I think we miss a test for `creatorMethod`

core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
```diff
@@ -1,305 +0,0 @@
...
-
-/**
- * Utility methods for implementing GraalVM.
- *
- * @author Álvaro Sánchez-Mariscal
- * @author graemerocher
- * @since 2.0.0
- * @deprecated Use GraalVM's own public API under {@code org.graalvm} or conditional metadata in JSON format instead
- */
-@Deprecated
```
sdelamo, line 40: do we consider this class internal?

sdelamo (changes requested):
Introspection Builder does not work with [Groovy @Builder AST](https://groovy-lang.org/metaprogramming.html#xform-Builder)


I expected this to work: 

```groovy
import groovy.transform.builder.Builder
@CompileStatic
@Introspected(builder = @Introspected.IntrospectionBuilder(builderMethod = "builder"))
@Builder
class Person {
    String name
    int age
}
```

And I expected this to work as well: 


```groovy
import groovy.transform.builder.Builder
@CompileStatic
@Introspected(builder = @Introspected.IntrospectionBuilder(builderMethod = Person.PersonBuilder))
@Builder
class Person {
    String name
    int age
}
``

test-suite-groovy/src/test/groovy/io/micronaut/docs/ioc/builders/Person.groovy
sdelamo: this PR adds the missing tags: https://github.com/micronaut-projects/micronaut-core/pull/9602

src/main/docs/guide/ioc/introspection.adoc
sdelamo: this PR adds the missing tags: https://github.com/micronaut-projects/micronaut-core/pull/9602

src/main/docs/guide/ioc/introspection.adoc
sdelamo: this PR changes the examples to show types which can only be constructed via the builder pattern https://github.com/micronaut-projects/micronaut-core/pull/9602

graemerocher:
I am aware but unfortunately the methods are not there in the Groovy AST possibly because they are added in a later phase. I don't see a way to fix that

sdelamo:
> I am aware but unfortunately the methods are not there in the Groovy AST possibly because they are added in a later phase. I don't see a way to fix that

Ok, I will [document it](https://github.com/micronaut-projects/micronaut-core/pull/9604).

core/src/main/java/io/micronaut/core/graal/AutomaticFeatureUtils.java
graemerocher: it was deprecated

sdelamo: yes, but we are deleting it in a patch version. Do we consider `io.micronaut.core.graal` internal?

sdelamo (commented):
I would like to incorporate:

- [x]  #9603 
- [x] #9617

sdelamo (approved):

PR #9503, opened 2023-06-27
Core Merge up 3.10.x to 4.0.x

sdelamo:


sdelamo:
@yawkat can you help me out with `PoolTimeoutSpec`? The test was introduced here: https://github.com/micronaut-projects/micronaut-core/pull/9449/files but `ConnectionManager` has changed so much that I don't see where to apply the merge.

yawkat:
oh yea that is already fixed, i will fix the test to use the right config properties.

PR #9505, opened 2023-06-27
core: enable binary compatability check

### LINKED ISSUE #9500: Enable binary compatibility check for 4.0.x branch of Micronaut Core
I have not checked but you may need to set a baseline. 

```groovy
micronautBuild {
    binaryCompatibility {
        baselineVerison = “4.0.0-RC3”
        enabled = true 
    }
}
```

wetted:
closes #9500

### DIFF
--- a/core-bom/build.gradle
+++ b/core-bom/build.gradle
@@ -19,28 +19,3 @@ micronautBuild {
         enabled = true
     }
 }
-
-// Workaround for https://github.com/micronaut-projects/micronaut-build/issues/584
-
-def writebaseline = tasks.register("writebaseline", WriteBaseline) {
-   baseline = micronautBuild.binaryCompatibility.baselineVersion
-   baselineFile = layout.buildDirectory.file("baseline.txt")
-}
-
-tasks.named("checkVersionCatalogCompatibility") {
-    baseline.set(writebaseline.flatMap { it.baselineFile })
-}
-
-abstract class WriteBaseline extends DefaultTask {
-   @Input
-   abstract Property<String> getBaseline()
-
-   @OutputFile
-   abstract RegularFileProperty getBaselineFile()
-
-   @TaskAction
-   void execute() {
-      baselineFile.get().asFile.text = baseline.get()
-   }
-}
-
--- a/core-bom/build.gradle
+++ b/core-bom/build.gradle
@@ -15,9 +15,7 @@ micronautBom {
 
 micronautBuild {
     binaryCompatibility {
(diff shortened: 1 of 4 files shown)


wetted:
see https://github.com/micronaut-projects/micronaut-build/issues/584

sdelamo (changes requested):
binary compatibility is still disabled. See https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-library.gradle#L20 or https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/context-propagation/build.gradle#L9

core-bom/build.gradle
```diff
@@ -15,9 +15,33 @@ micronautBom {
 
 micronautBuild {
     binaryCompatibility {
-        def dash = version.indexOf('-')
-        def v = dash > 0 ? version.substring(0, dash) : version
-        def (major, minor, patch) = v.split('[.]').collect { it.toInteger() }
-        enabled = major > 4 || (major == 4 && minor > 0) || (major == 4 && minor == 0 && patch > 0)
+        baselineVersion = "4.0.0-RC3"
```
sdelamo, line 18: doe we need to set the baseline version?

wetted: Apparently, so. If I try to build it resolves 3.9.4 and that causes problems with non-existent modules and `japiCmp` checks. @melix said we will need to remove it after the 4.0.0 release

wetted:
I am having some pain getting this to work. I wonder if we need a new build plugin with the fix for `binaryCompatibility`.  @melix ?

core-bom/build.gradle
wetted: https://github.com/micronaut-projects/micronaut-core/issues/9500#issuecomment-1609603337

melix:
let me check. I have pushed a fix on `micronaut-build` but I'll double check nothing else is needed

melix:
There are more changes needed. In addition it seems that `micronaut-core` supplies the wrong group for the baseline comparator... On it.

sdelamo:
@wetted @melix any idea about the sonar failure which made the build fail?

melix:
Most likely due to the sonar plugin upgrade in the latest release of build plugins. Apparently the plugin expects a String somewhere but gets a GString. Trying to figure out where...

melix:
See https://github.com/micronaut-projects/micronaut-build/pull/587

PR #9508, opened 2023-06-27
Allow setting of cloud deduction with probes via env vars or property

### LINKED ISSUE #9506: Cloud environment detection cannot be turn on again via env. variable in Micronaut Framework 4
### Expected Behavior

Cloud environment detection is off by default in Micronaut framework 4. However, we should give users the ability to turn it on by defining the environment variable `MICRONAUT_ENV_DEDUCTION` to true. 

### Actual Behaviour

I deployed a Micronaut 4 application to Elastic beanstalk.

✅ With Micronaut framework 3 both cloud and ec2 where detected. 
✅ With Micronaut framework 4 none were detected .
❌ With Micronaut framework 4 and `MICRONAUT_ENV_DEDUCTION ` set to true `cloud` and `ec2` where detected. 

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.0.0-M4

timyates:
In https://github.com/micronaut-projects/micronaut-core/pull/8068 we disabled probing the OS/network to try and discover the cloud we were running on.

This PR adds an environment variable and property to allow it to be enabled.

Closes #9506

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
+++ b/inject/src/main/java/io/micronaut/context/env/CachedEnvironment.java
@@ -23,6 +23,8 @@ import io.micronaut.core.optim.StaticOptimizations;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.function.UnaryOperator;
 
 /**
  * A "cached environment" is a performance optimization aimed at minimizing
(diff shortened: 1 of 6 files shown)


timyates:
I couldn't find a way of testing this (apart from manually)...

In the test build script, we set the cloud provider to `other` and the methods that are invoked are both private an static to the DefaultEnvironment class...

sdelamo:
> I couldn't find a way of testing this (apart from manually)...
> 
> In the test build script, we set the cloud provider to `other` and the methods that are invoked are both private an static to the DefaultEnvironment class...

I wrote a unit test https://github.com/micronaut-projects/micronaut-core/pull/9508/commits/2e8ad83273c4958318298629d95ee8073d667bc4  for the `shouldDeduceCloudEnvironment` method.

sdelamo (approved):
I mentioned the system property and environment variable in the breaking docs as well.

graemerocher (approved):

timyates:
@sdelamo removed the unused import and repetition in the test...

PR #9509, opened 2023-06-27
Change annotation-based CORS to match configuration-based defaults

### LINKED ISSUE #9504: Configuration- and annotation- based  CORS should have same behavior.
### Issue description

Configuration-based CORS allows all headers and all origins when turned on, unless constrained in a more fine-grained manner using the `allowed-headers` and `allowed-origins` options. Annotation-based CORS should behave exactly the same way.

see #9470

wetted:
Change CrossOrigin annotation-based CORS to match configuration-based CORS defaults so that enabling CORS allows all origins by default. This is a breaking change from Micronaut 3.9.x

closes #9504

### DIFF
--- a/src/main/docs/guide/appendix/breaks.adoc
+++ b/src/main/docs/guide/appendix/breaks.adoc
@@ -209,5 +209,5 @@ Since Micronaut 3.0 the `OncePerRequestHttpServerFilter` class was deprecated an
 
 ==== CORS support with the `@CrossOrigin` annotation
 
-In previous and current versions of Micronaut Framework, configuration-based CORS defaults to allow all origins when CORS is enabled and `allowed-origins` is not set otherwise. However, in previous versions of Micronaut Framework using the `@CrossOrigin` annotation defaults to no origins allowed, requiring the `allowedOrigins` attribute to be set. In Micronaut 4, `@CrossOrigin` is changed to match the behavior of configuration-based CORS. In other words, enabling CORS with `@CrossOrigin` without specifying any attributes means all origins are allowed.
+Micronaut Framework 4 changes `@CrossOrigin`  behavior to match configuration-based CORS behavior. A method annotated with `@CrossOrigin` allows any origin if you don't specify any value for the `allowedOrigins` and `allowedOriginsRegex` members.
 
--- a/src/main/docs/guide/appendix/breaks.adoc
+++ b/src/main/docs/guide/appendix/breaks.adoc
@@ -206,3 +206,8 @@ old behavior.
 ==== `OncePerRequestHttpServerFilter` removed
 
 Since Micronaut 3.0 the `OncePerRequestHttpServerFilter` class was deprecated and marked for removal. This class is now removed. Implement link:{api}/io/micronaut/http/filter/HttpServerFilter.html[HttpServerFilter] instead, and replace any usages of `micronaut.once` attributes with a custom attribute name.
+
+==== CORS support with the `@CrossOrigin` annotation
+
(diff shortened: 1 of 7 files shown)


src/main/docs/guide/httpServer/serverConfiguration/cors/corsAllowedOrigins.adoc
```diff
@@ -18,4 +18,4 @@ micronaut:
             - http://foo.com
 ----
 
-WARNING: Use the `allowed-origins-regex` configuration judiciously. You may accidentally make an insecure configuration which could be targeted by an attacker registering domains targeting the regular expression.
+WARNING: Use the `allowed-origins-regex` configuration judiciously. You may accidentally make an insecure configuration which could be targeted by an attacker registering domains targeting the regular expression. Note that if `allowed-origins-regex` is set then `allowed-origins` is ignored.
```
sdelamo: is that so when you use cors via configuration?

src/main/docs/guide/appendix/breaks.adoc
```diff
@@ -206,3 +206,8 @@ old behavior.
 ==== `OncePerRequestHttpServerFilter` removed
 
 Since Micronaut 3.0 the `OncePerRequestHttpServerFilter` class was deprecated and marked for removal. This class is now removed. Implement link:{api}/io/micronaut/http/filter/HttpServerFilter.html[HttpServerFilter] instead, and replace any usages of `micronaut.once` attributes with a custom attribute name.
+
+==== CORS support with the `@CrossOrigin` annotation
+
+In previous and current versions of Micronaut Framework, configuration-based CORS defaults to allow all origins when CORS is enabled and `allowed-origins` is not set otherwise. However, in previous versions of Micronaut Framework using the `@CrossOrigin` annotation defaults to no origins allowed, requiring the `allowedOrigins` attribute to be set. In Micronaut 4, `@CrossOrigin` is changed to match the behavior of configuration-based CORS. In other words, enabling CORS with `@CrossOrigin` without specifying any attributes means all origins are allowed.
```
sdelamo:
```suggestion
Micronaut Framework 4 changes `@CrossOrigin`  behavior to match configuration-based CORS behavior. A method annotated with `@CrossOrigin` allows any origin if you don't specify any value for the `allowedOrigins` and `allowedOriginsRegex` members.
```

sdelamo (changes requested):

yawkat:
i dont think we should allow all headers by default

sdelamo:
> i dont think we should allow all headers by default

what default do you think we should use?

yawkat:
no allowed headers

sdelamo:
@yawkat can you review the contents of this PR? I think it would be important to have configuration and annotation based CORs aligned for  MN$4. 

I have moved  the suggestion of changing the default of allowed headers to https://github.com/micronaut-projects/micronaut-core/issues/9518

yawkat (approved):

src/main/docs/guide/httpServer/serverConfiguration/cors/corsAllowedOrigins.adoc
wetted: I don't think configuration prevents setting both. I'm checking to see what happens. I'll write a test.

wetted: OK, it looks like with configuration, you use something like

```
micronaut.server.cors.configurations.foo.allowed-origins-regex = ^http(|s):\\/\\/foo\\.com$
micronaut.server.cors.configurations.foo.allowed-origins = https://bar.com
```

then both `http://foo.com` and `https://bar.com` are allowed.

I stand corrected and will change annotation-based to match.

sdelamo (approved):

PR #9530, opened 2023-06-30
Add BinaryTypeConfiguration

timyates:
This is used by the AWS and Azure modules.  Originally we moved this duplication to Servlet-core:

https://github.com/micronaut-projects/micronaut-servlet/pull/504

But it was then decided that core was a better home for it.

It specified defaults, and allows users to ignore these defaults, and add their own mime types.

### DIFF
--- a/http/src/main/java/io/micronaut/http/binary/BinaryTypeConfiguration.java
+++ b/http/src/main/java/io/micronaut/http/binary/BinaryTypeConfiguration.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.binary;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.util.ArgumentUtils;
+import io.micronaut.http.MediaType;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Allows checking for MediaTypes that should be considered binary.
+ * This is used for example to determine whether to mime encode an AWS lambda response message (where the response body is a String).
+ *
+ * @author Tim Yates
+ * @since 4.0.0
+ */
+@ConfigurationProperties(BinaryTypeConfiguration.PREFIX)
+public class BinaryTypeConfiguration {
+
+    static final String PREFIX = "micronaut.http.binary-types";
+
+    private static final Set<String> DEFAULT_BINARY_TYPES = Set.of(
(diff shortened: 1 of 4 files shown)


graemerocher:
but it isn't integrated with the Netty server and client?

timyates:
@graemerocher we ended up with the same thing in Aws and Azure. First, we moved it to servlet-core here https://github.com/micronaut-projects/micronaut-servlet/pull/504 but then it was decided that it should move to core.

But it's not used in netty (server or client)

Finding a home for things is hard ☹️

http/src/main/java/io/micronaut/http/binary/BinaryTypeConfiguration.java
```diff
@@ -0,0 +1,104 @@
...
+     * If this is false then calls to {@link #isMediaTypeBinary(String)} will only check the additional types, and ignore the defaults.
+     *
+     * @return Whether to use the default binary types
+     */
+    public boolean isUseDefaults() {
+        return useDefaultBinaryTypes;
+    }
+
+    /**
+     * Sets whether to use the default binary types.
```
sdelamo: Tell the user in the javadocs which shows in the Configuration reference documentation which are the default binary types.

sdelamo (changes requested):
add missing package-info.java file

sdelamo:
@graemerocher and @timyates I think this belongs in `micronaut-http` it is related with media types. It make sense, we place it in the module where `MediaType.java` is.

graemerocher:
I don't think this should be merged if it is not integrated with Netty, it will just be confusing

sdelamo:
> I don't think this should be merged if it is not integrated with Netty, it will just be confusing

Do you think we should add it `servlet-core` or duplicate it across the serverless modules? For me, `micronaut-http` should serve every http implementation not just netty.

graemerocher:
if it is specific to server less probably it makes sense in the `function` module. If you add it to `servlet-core` then the same question applies, will it be applied to Jetty/Tomcat/Undertow?

sdelamo:
@timyates Please, check if every usage of BinaryTypeConfiguratino is an a module which current depends on `io.micronaut:micronaut-function`

timyates:
@sdelamo 

### Azure

Azure does as it has an [api dependency on azure-function](https://github.com/micronaut-projects/micronaut-azure/blob/845bfb619b3c9ed668fc675a2a60fa6e3f4e5a82/azure-function-http/build.gradle.kts#L9) which has an [api dependency on micronaut-function](https://github.com/micronaut-projects/micronaut-azure/blob/845bfb619b3c9ed668fc675a2a60fa6e3f4e5a82/azure-function/build.gradle.kts#L7)

### AWS

And AWS [does](https://github.com/micronaut-projects/micronaut-aws/blob/36dbb044080b51b35cc2692aeb3943d4029b4caf/function-aws-api-proxy/build.gradle.kts#L7) the [same](https://github.com/micronaut-projects/micronaut-aws/blob/36dbb044080b51b35cc2692aeb3943d4029b4caf/function-aws/build.gradle.kts#L6) 

I'll move this into micronaut-function 👍

http/src/main/java/io/micronaut/http/binary/BinaryTypeConfiguration.java
timyates: Done in https://github.com/micronaut-projects/micronaut-core/pull/9530/commits/06e7c9f1950a219e0ff59023b84b23467562b186

sdelamo (approved):

sdelamo:
@graemerocher can you review this ?

sdelamo:
ping @graemerocher

graemerocher (approved):

sdelamo:
@timyates can you create prs in the different modules to use this API.

PR #9534, opened 2023-07-03
Possible bug in Factory and EachBean

timyates:
More likely to be me making a mistake, but we have the following pattern in the micronaut-security-keys-jwks guide, which is failing (as does this) with io.micronaut.context.exceptions.NonUniqueBeanException

Shouldn't this be handled by the EachBean annotation

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/factory/FactoryEachBeanNamedSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/factory/FactoryEachBeanNamedSpec.groovy
@@ -1,0 +1,67 @@
+package io.micronaut.inject.factory
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.BeanContext
+import io.micronaut.context.annotation.EachBean
+import io.micronaut.context.annotation.Factory
+import io.micronaut.context.annotation.Requires
+import io.micronaut.runtime.context.scope.Refreshable
+import jakarta.inject.Named
+import spock.lang.Specification
+
+class FactoryEachBeanNamedSpec extends Specification {
+
+    static final SPEC_NAME = "FactoryEachBeanNamedSpec"
+
+    void "check the factory"() {
+        given:
+        BeanContext beanContext = ApplicationContext.run(["spec.name": SPEC_NAME])
+
+        expect:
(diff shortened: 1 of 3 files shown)


dstepanov:
@graemerocher Not sure if that is the correct fix. Essentially if the candidate is not qualified, we cannot properly find it when constructing another each-bean. I have to add a by-type qualifier to fix it.
The other solution would be to propagate the candite bean definition to the bean delegate, but anyway, that would require `@Parameter.`

@timyates The solution here is to add some qualifier to the bean named one

Also, the problem is not releated to `@Refreshable` as it also fails without it

graemerocher (approved):

dstepanov:
I’m not really sure if this should be “fixed”. 
Maybe we can just require a qualifier for each bean candidates

sdelamo:
> I’m not really sure if this should be “fixed”.
> Maybe we can just require a qualifier for each bean candidates

I thought we required a name qualifier for each bean candidates.

timyates:
> I thought we required a name qualifier for each bean candidates.

Is this new for 4.0.0?  As previously the `micronaut-security-keys-jwks` guide worked?

sdelamo:
> >> I’m not really sure if this should be “fixed”.
> >> Maybe we can just require a qualifier for each bean candidates
 
> > I thought we required a name qualifier for each bean candidates.

> Is this new for 4.0.0? As previously the `micronaut-security-keys-jwks` guide worked?

I think it is a bug. We did not require it for 3. But often, each bean was using a each property which had a name qualifier. I think we should support not having beans without a name qualifier being used in `@EachBean`

dstepanov:
FYI This was working because we had a bug https://github.com/micronaut-projects/micronaut-core/pull/9399

sdelamo:
@dstepanov is your opinion that we should close this PR and document that `@EachBean` requires beans to have a name qualifier?

dstepanov:
Yes, we can try to fix it later

sdelamo:
I don't think this is the correct fix. the test in this PR fails in 3.x as well.

sdelamo:
I have created a PR which I think supperseeds this PR and we can close this one: 

https://github.com/micronaut-projects/micronaut-core/pull/9548

PR #9535, opened 2023-07-04
fix: txt/plain possible for BigDecimal

sdelamo:
`isJavaBasicType` does `isJavaLangType` plus it supports `UUID`, `BigDecimal`, `BigInteger`, `URL`, `URI`, `TimeZone`, `Charset`, `Locale`, `Duration`, `Date`, `LocalDate`, `Instant` and `ZonedDateTime`.


graemerocher (approved):

graemerocher (changes requested):
Seems to have broken something

timyates:
From the logs, it looks like the MDC context tests failed to start, and got themselves into a death-spiral of logging

Re-ran it and it all passed 👍

PR #9536, opened 2023-07-04
use binary name to store annotation names in metadata for KSP

graemerocher:
KSP is currently using the canonical name to store annotation names in annotation metadata. This breaks for inner annotations since the binary name is the expected format produced by other languages. This PR changes the KSP implementation to use the binary name instead.

### DIFF
--- a/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/annotation/KotlinAnnotationMetadataBuilder.kt
+++ b/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/annotation/KotlinAnnotationMetadataBuilder.kt
@@ -50,13 +50,9 @@ internal class KotlinAnnotationMetadataBuilder(private val symbolProcessorEnviro
         private fun getTypeForAnnotation(annotationMirror: KSAnnotation, visitorContext: KotlinVisitorContext): KSClassDeclaration {
             return annotationMirror.annotationType.resolve().declaration.getClassDeclaration(visitorContext)
         }
-        fun getAnnotationTypeName(annotationMirror: KSAnnotation, visitorContext: KotlinVisitorContext): String {
+        fun getAnnotationTypeName(resolver: Resolver, annotationMirror: KSAnnotation, visitorContext: KotlinVisitorContext): String {
             val type = getTypeForAnnotation(annotationMirror, visitorContext)
-            return if (type.qualifiedName != null) {
-                type.qualifiedName!!.asString()
-            } else {
-                annotationMirror.shortName.asString()
-            }
+            return type.getBinaryName(resolver, visitorContext)
         }
     }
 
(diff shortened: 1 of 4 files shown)


timyates:
So I tested this with the micronaut-microstream-persistence guide

It gets further than before (can serialize things) 🎉  but this test fails:

https://github.com/timyates/micronaut-microstream-persistence-gradle-kotlin/blob/main/src/test/kotlin/example/micronaut/FruitValidationControllerTest.kt

As with Java, it serializes the object in the request to `{}` (as both values are empty), but then the server throws an exception:

```
11:19:32.499 [default-nioEventLoopGroup-1-3] ERROR i.m.http.server.RouteExecutor - Unexpected error occurred: Parameter specified as non-null is null: method example.micronaut.FruitCommand.<init>, parameter name
java.lang.NullPointerException: Parameter specified as non-null is null: method example.micronaut.FruitCommand.<init>, parameter name
```

I guess this is because we are using Kotlin to check the nullability instead of the validator, and the order of things has changed somewhere...

The more I think about it, the more I think we should change the code in the guide to allow a nullable `name` and check it with validation...

You can run the above repo after running `./gradlew pTML` on this branch in core so you get a SNAPSHOT of this in your local maven repo

graemerocher:
this error seems a problem with the guide. If name is allowed to be null it should be `val name?: String` and not `val name: String`

timyates:
It's not allowed to be null or empty. Previously in 3.x this was a bad request, but now it seems to be an internal server error

timyates:
Checked with the micronaut-cache guide (the original instigator of this issue) and it works with this PR applied 🎉

timyates (approved):

timyates:
> It's not allowed to be null or empty. Previously in 3.x this was a bad request, but now it seems to be an internal server error

It also passes with Kapt 🤔 Could there be a similar issue in validation?

graemerocher:
likely a separate issue with a separate root cause, will investigate

graemerocher:
the FruitValidationControllerTest test passes locally for me, not sure how to reproduce

timyates:
> the FruitValidationControllerTest test passes locally for me, not sure how to reproduce

Using the [above repo](https://github.com/timyates/micronaut-microstream-persistence-gradle-kotlin/blob/main/src/test/kotlin/example/micronaut/FruitValidationControllerTest.kt)?

I get: https://scans.gradle.com/s/xruxjq6yckeqw/tests/:test/example.micronaut.FruitValidationControllerTest/fruitIsValidated()?top-execution=1

I wonder what we have different...

timyates:
Checked, and I [seem to be using the correct inject-kotlin](https://scans.gradle.com/s/xruxjq6yckeqw/dependencies?focusedDependency=WzAsMTMsNCxbMCwxMyxbNF1dXQ&toggled=W1swXSxbMCwxN10sWzAsMTVdLFswLDEyXSxbMCwxM10sWzAsMTMsWzRdXV0) 🤔

PR #9538, opened 2023-07-05
Fix KSP nullability handling

graemerocher:
With Kotlin it is not enough to just add `Nullable` to elements marked as nullable, we also have to store the fact that elements not marked as such as `NonNull`. This PR stores `NonNull` in the annotation metadata for Kotlin elements that cannot be null.


timyates (approved):
Ran the microstream-persistence guide with this and confirmed that the validation worked correctly with Ksp (as it did with Kapt) 👍

PR #9552, opened 2023-07-10
imp: MethodRoute stringValues behave as getValues

sdelamo:


inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataHierarchy.java
```diff
@@ -322,15 +322,18 @@ public String[] stringValues(@NonNull Class<? extends Annotation> annotation, @N
...
-        for (int i = 1; i < hierarchy.length; i++) {
-            AnnotationMetadata annotationMetadata = hierarchy[i];
-            final String[] moreValues = annotationMetadata.stringValues(annotation, member);
-            if (ArrayUtils.isNotEmpty(moreValues)) {
-                values = ArrayUtils.concat(values, moreValues);
+        return firstNotEmptyStringValuesInHierarchy(annotation, member);
+    }
+
+    @NonNull
+    private String[] firstNotEmptyStringValuesInHierarchy(@NonNull String annotation, @NonNull String member) {
```
dstepanov, line 329: I think the previous code is more correct, all the values are aggregated.

graemerocher:
It is too late to make a change this big IMO

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataHierarchy.java
sdelamo: Please note that with: 

```
     @Controller("/example")
    @RolesAllowed("isAuthenticated()")
    static class ExampleController {

        @Get("/admin")
        @RolesAllowed({"ROLE_ADMIN", "ROLE_X"})
        void  admin(HttpRequest<?> request) {
        }
```

old `stringValues` returns  `"ROLE_ADMIN", "ROLE_X", "isAuthenticated()"` while getValues returns `"ROLE_ADMIN", "ROLE_X"`. The latter looks correct to me.

graemerocher: There may be some behaviour inconsistency here, but this change has the potential to break a lot of things. I don't think we should be merging this 2 days before GA

sdelamo: I am ok not merging it. Do you want me to target 5.0.x as the PR base?

dstepanov: @sdelamo You might need to modify the code that reads the values to only take method annotation metadata

PR #9553, opened 2023-07-10
KSP: Visit synthetic properties in factories with advice

graemerocher:
In order for advice to apply to Groovy/Kotlin synthetic properties we need to visit these in the implementation of `@Factory`. This replicates the same logic as for introduction advice.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataWriter.java
@@ -328,6 +328,10 @@ public class AnnotationMetadataWriter extends AbstractClassFileWriter {
             Map<String, GeneratorAdapter> loadTypeMethods,
             AnnotationMetadata annotationMetadata) {
         annotationMetadata = annotationMetadata.getTargetAnnotationMetadata();
+        if (annotationMetadata instanceof AnnotationMetadataHierarchy annotationMetadataHierarchy) {
+            // Synthetic property getters / setters can consist of field + (setter / getter) annotation hierarchy
+            annotationMetadata = MutableAnnotationMetadata.of(annotationMetadataHierarchy);
+        }
         if (annotationMetadata.isEmpty()) {
             generatorAdapter.getStatic(Type.getType(AnnotationMetadata.class), "EMPTY_METADATA", Type.getType(AnnotationMetadata.class));
         } else if (annotationMetadata instanceof MutableAnnotationMetadata mutableAnnotationMetadata) {
--- a/core-processor/src/main/java/io/micronaut/inject/processing/FactoryBeanElementCreator.java
+++ b/core-processor/src/main/java/io/micronaut/inject/processing/FactoryBeanElementCreator.java
@@ -244,6 +244,21 @@ final class FactoryBeanElementCreator extends DeclaredBeanElementCreator {
                 .filter(m -> m.isPublic() && !m.isFinal() && !m.isStatic()).toList();
             methodElements
                 .forEach(methodElement -> visitAroundMethod(aopProxyWriter, methodElement.getDeclaringType(), methodElement));
+            List<PropertyElement> syntheticBeanProperties = producedType.getSyntheticBeanProperties();
+            for (PropertyElement syntheticBeanProperty : syntheticBeanProperties) {
+                syntheticBeanProperty.getReadMethod().ifPresent(m -> {
+                        if (!m.isFinal()) {
+                            visitAroundMethod(aopProxyWriter, m.getDeclaringType(), m);
+                        }
+                    }
+                );
+                syntheticBeanProperty.getWriteMethod().ifPresent(m -> {
+                        if (!m.isFinal()) {
+                            visitAroundMethod(aopProxyWriter, m.getDeclaringType(), m);
+                        }
+                    }
+                );
+            }
 
         } else if (producedAnnotationMetadata.hasStereotype(Executable.class)) {
             if (producedType.isArray()) {
--- a/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinPropertyAccessorMethodElement.kt
+++ b/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinPropertyAccessorMethodElement.kt
@@ -45,7 +45,7 @@ internal abstract class AbstractKotlinPropertyAccessorMethodElement<T : KotlinNa
 
     override fun isSynthetic() = true
 
-    override fun isFinal() = true
+    override fun isFinal() = !accessor.receiver.isOpen()
 
     override fun isAbstract(): Boolean = accessor.receiver.isAbstract()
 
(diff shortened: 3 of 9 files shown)


timyates:
Tested it against the [guide for configuration](https://scans.gradle.com/s/6oa3veavk726s/console-log?task=:kspKotlin), and the [rabbit guide book project](https://scans.gradle.com/s/o7d4s6rachdgs/console-log?task=:kspKotlin), both failed with ksp errors 🤔

jeffscottbrown:
Is this a change that will likely result in an RC2 prior to GA?

Thank you for any input!

dstepanov:
I did fix some of the problems, but still we have failures. Some of the tests don't expect synthetic methods. We might need to add more options to the methods query. Right now we have duplicate methods provided from the properties and the query methods. We might want to have `ALL_METHODS.includePropertyMethods` or `.includeSynthetic`

dstepanov:
We should validate that the logic is consistent with the Groovy implementation.

jeffscottbrown:
btw... my question above regarding RC2 isn't expressing a preference in favor of RC2 or expressing concern about doing an RC2 or anything like that. I am only asking for planning purposes.  

Thank you!

sdelamo:
> Is this a change that will likely result in an RC2 prior to GA?

No.

timyates:
Still seeing mockito failures with this in the rabbitmq guide

```
org.mockito.exceptions.misusing.NotAMockException: 
Argument passed to verify() is of type $BookControllerTest$AnalyticsClient0$Definition$Intercepted and is not a mock!
Make sure you place the parenthesis correctly!
See the examples of correct verifications:
    verify(mock).someMethod();
    verify(mock, times(10)).someMethod();
    verify(mock, atLeastOnce()).someMethod();
	at app//example.micronaut.BookControllerTest.testMessageIsPublishedToRabbitMQWhenBookFound(BookControllerTest.kt:31)
	at java.base@17.0.7/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base@17.0.7/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base@17.0.7/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base@17.0.7/java.lang.reflect.Method.invoke(Method.java:568)
```

It works with kapt 🤔

graemerocher:
@dstepanov changed the fix to use the existing synthetic properties stuff, should work for Groovy too

graemerocher:
@timyates can you retest the current changes

timyates:
Will try asap...  Groovy compilation seems to be failing here with `Unknown annotation metadata`?

timyates:
So, I believe this was the originator of this PR? https://github.com/micronaut-projects/micronaut-serialization/issues/513

Running the `micronaut-configuration` guide with this branch published to maven local, all the tests pass!

https://scans.gradle.com/s/xz4jioxbmnt46/tests/overview

We still have an issue testing the rabbit guide with this branch [published to mavenLocal](https://scans.gradle.com/s/ibwjpiiyfk64m/dependencies?focusedDependency=WzAsMTEsMixbMCwxMSxbMl1dXQ&toggled=W1swXSxbMCwxMV1d), I still see the [issues with mockito](https://scans.gradle.com/s/ibwjpiiyfk64m/tests/overview?outcome=failed)

jeffscottbrown:
> No.

Thank you for the info @sdelamo.

timyates:
Testing against the k8s guide where we have

```
    @get:Get("/orders")
    val orders: List<Order>
```

Think we just get the mockito error...

Is it the case that when a Client is constructed from

```
@Client("orders") // <1>
interface OrdersClient {

    @Post("/orders")
    fun createOrder(@Body order: Order?): Order
}
```

the `createOrder` method is not `open`? 🤔

dstepanov (approved):

timyates (approved):
Any ideas on https://github.com/micronaut-projects/micronaut-core/pull/9553#issuecomment-1630849539 ?

dstepanov:
@timyates What exatly is the problem there?

timyates:
@dstepanov When we have a Client defined as so:

```
@Client("orders") // <1>
interface OrdersClient {

    @Post("/orders")
    fun createOrder(@Body order: Order?): Order
}
```

Does the generated class have an `open` createOrder function? And is it open itself?

Trying to think why

```
`when`(ordersClient.createOrder(any())).thenReturn(order)
```

Fails with 

```
Misplaced or misused argument matcher detected here:

-> at example.micronaut.GatewayControllerTest.createOrder(GatewayControllerTest.kt:134)

You cannot use argument matchers outside of verification or stubbing.
Examples of correct usage of argument matchers:
    when(mock.get(anyInt())).thenReturn(null);
    doThrow(new RuntimeException()).when(mock).someVoidMethod(any());
    verify(mock).someMethod(contains("foo"))

This message may appear after an NullPointerException if the last matcher is returning an object 
like any() but the stubbed method signature expect a primitive argument, in this case,
use primitive alternatives.
    when(mock.get(any())); // bad use, will raise NPE
    when(mock.get(anyInt())); // correct usage use

Also, this error might show up because you use argument matchers with methods that cannot be mocked.
Following methods *cannot* be stubbed/verified: final/private/equals()/hashCode().
Mocking methods declared on non-public parent classes is not supported.

Path Taken: GatewayControllerTest.setOrdersClient([OrdersClient <set-?>])
```

dstepanov:
We don't generate anything different for KSP. If it works for Java it should work for Kotlin. Do you use https://github.com/mockito/mockito-kotlin ?

timyates:
> We don't generate anything different for KSP. If it works for Java it should work for Kotlin. Do you use https://github.com/mockito/mockito-kotlin ?

Yeah tried with that and no difference... We use the `all open` plugin with kapt but it doesn't work with ksp, and I remember @yawkat did something to work around this limitation. I guess I wonder if this workaround also works for things generated from interfaces it abstract classes

Or maybe it's something totally different...

dstepanov:
Can you try mock a method without a body?

timyates:
It's specifically when we make use of `any()` in the matcher...  This works

```
        `when`(usersClient.getById(user.id)).thenReturn(user)
```

But this shows the above error

```
        `when`(ordersClient.createOrder(any())).thenReturn(order)
```

I don't know enough Mockito (or Kotlin it seems) to know why it's failing with KSP

Not sure if this answers

>  Can you try mock a method without a body?

If not, what do you mean?

dstepanov:
Right, so I guess the open plugin somehow doesn't work so it cannot create a body mock. Maybe try `any(Order.class)`

timyates:
> Right, so I guess the open plugin somehow doesn't work so it cannot create a body mock. Maybe try `any(Order.class)`

Yeah, I'd tried that

```
        `when`(ordersClient.createOrder(any(Order::class.java))).thenReturn(order)
```

Same error

dstepanov:
Try to open the class file of `Order` and see if it's final

timyates:
@dstepanov It is, but if I change it from a `data class` to an `open class`, it makes no difference :-(

PR #9562, opened 2023-07-13
Wait for Netty shutdown when calling stopServerOnly

timyates:
When running with CRaC, we encountered an error with Micronaut 4 running on Google Cloud Run.

Micronaut 3 apps worked as expected, but when a Micronaut 4 application was deployed to GCP it would fail to start with the error message:

```
free(): invalid pointer
```

Whilst trying to work out what was broken, I discovered that sleeping after calling `stopServerOnly` caused the deployment to pass.

I believe something has changed in Micronaut 4 that is either taking longer to close the ELG, or has some native code in it that is calling free in glibc on shutdown

The issue is that CRaC is snapshotting these objects in the process of shutting down, so on startup they continue and attempt to free memory that they no longer have ownership of.

I have no idea why the snapshotted image worked in a local docker installation, and only seemed to fail on GCP Cloud Run. I suspect the error was there in local docker, but was simply ignored by running it as --privileged

This PR fixes this by gathering the Netty promises when we shut down, and if we are calling `stopServerOnly` (ie: CRaC) then `await()` all the Futures in this list in turn before returning from `stopInternal`.

See: https://github.com/micronaut-projects/micronaut-crac/issues/156

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyEmbeddedServer.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyEmbeddedServer.java
@@ -60,6 +60,7 @@ public interface NettyEmbeddedServer
     /**
      * Stops the Netty instance, but keeps the ApplicationContext running.
      * This for CRaC checkpointing purposes.
+     * This method will only return after waiting for netty to stop.
      *
      * @return The stopped NettyEmbeddedServer
      */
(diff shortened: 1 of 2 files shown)


graemerocher:
if this has a performance impact on server shutdown I am opposed to merging this just to support CraC. It should be written in a way that for CraC it performs the slower shutdown process

timyates:
@graemerocher I don't believe it does (well, not a significant one).  The only thing that calls `stopServerOnly` is CRaC AFAIK, and it is only this that calls `stopInternal` with the `stopServerOnly` boolean set to `true`.

So for the general case, the only cost should be the creation of an array list, and adding up to 3 Futures to it 🤔

yawkat (approved):

graemerocher (approved):

sdelamo (approved):

PR #9596, opened 2023-07-23
Minor fixes

altro3:
1. Javadoc lint fixes
2. Checkstyle fixes
3. Some typos fixes
4. Fixed some tests for different system locale
5. Fixed minimal java release for inject-test JavaParser
6. Added '--add-opens java.base/java.lang=ALL-UNNAMED' for correct work SystemLambda and '--add-opens jdk.compiler/com.sun.tools.javac.lang=ALL-UNNAMED' for kotlin in tests
7. Fixed DefaultFileSystemResourceLoader normalize method for windows


inject/src/main/java/io/micronaut/inject/ParametrizedProvider.java
```diff
@@ -1,47 +0,0 @@
...
-import java.util.Map;
-
-/**
- * An extended version of the {@link Provider} interface for {@link ParametrizedBeanFactory}.
- *
- * @param <T> The type
- * @author graemerocher
- * @since 1.0
- */
-public interface ParametrizedProvider<T> extends Provider<T> {
```
graemerocher: this is a breaking change we can't remove it even if it is not used until next major

graemerocher (changes requested):

inject/src/main/java/io/micronaut/inject/ParametrizedProvider.java
altro3: @graemerocher And how to be? Will this class be present in all releases up to version 5.0 now?

As far as I understand, you forgot to remove it when removed the `ParametrizedBeanFactory` class.

altro3: Anyway I rolled back this change

altro3:
@sdelamo @dstepanov Hi! Could you review this changes?

context/src/main/java/io/micronaut/scheduling/executor/ExecutorType.java
```diff
@@ -45,7 +45,7 @@ public enum ExecutorType {
     WORK_STEALING,
 
     /**
-     * @see java.util.concurrent.Executors#newThreadPerTaskExecutor()
+     * @see io.micronaut.scheduling.LoomSupport#newThreadPerTaskExecutor(java.util.concurrent.ThreadFactory)
```
sdelamo: @yawkat is this change in the javadoc correct?

core/src/main/java/io/micronaut/core/io/file/DefaultFileSystemResourceLoader.java
```diff
@@ -129,6 +129,9 @@ private static String normalize(String path) {
         }
         if (path.startsWith("file:")) {
             path = path.substring(5);
+            if (path.startsWith("//")) {
```
sdelamo, line 132: Why this change? can you add a test?

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/MultiObjectBody.java
```diff
@@ -36,7 +36,7 @@
 public sealed interface MultiObjectBody extends HttpBody permits ImmediateMultiObjectBody, ImmediateSingleObjectBody, StreamingMultiObjectBody {
     /**
      * Coerce this value to an {@link InputStream}. This implements
-     * {@link io.micronaut.http.server.netty.binders.InputStreamBodyBinder}. Requires the objects
+     * io.micronaut.http.server.netty.binders.NettyInputStreamBodyBinder. Requires the objects
```
sdelamo: why remove the `@link`?

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -1071,7 +1071,7 @@ public static class Parent extends EventLoopConfig {
     }
 
     /**
-     * Allows configuration of properties for the {@link io.micronaut.http.server.netty.types.files.FileTypeHandler}.
+     * Allows configuration of properties for the {@link io.micronaut.http.server.netty.body.AbstractFileBodyWriter}.
```
sdelamo: @yawkat is this change of javadoc correct?

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/AbstractKotlinCompilation.kt
```diff
@@ -293,7 +293,6 @@ abstract class AbstractKotlinCompilation<A : CommonCompilerArguments> internal c
 
 internal fun convertKotlinExitCode(code: ExitCode) = when(code) {
     ExitCode.OK -> KotlinCompilation.ExitCode.OK
-    ExitCode.OOM_ERROR,
```
sdelamo: why this removal?

inject/build.gradle
```diff
@@ -30,8 +30,11 @@ dependencies {
 
 tasks.withType(Test).configureEach {
     if (JavaVersion.current().majorVersion.toInteger() >= 17) {
-        logger.warn("Opening java.util, so SystemLambda can work")
-        jvmArgs += ['--add-opens', 'java.base/java.util=ALL-UNNAMED']
+        logger.warn("Opening java.util and java.lang, so SystemLambda can work")
+        jvmArgs += [
+                '--add-opens', 'java.base/java.util=ALL-UNNAMED',
+                '--add-opens', 'java.base/java.lang=ALL-UNNAMED',
```
sdelamo: why is this necessary?

jackson-core/src/main/java/io/micronaut/jackson/core/tree/TreeGenerator.java
```diff
@@ -167,7 +167,7 @@ public JsonNode getCompletedValue() {
     }
 
     @Override
-    public void writeStartArray() throws IOException {
+    public void writeStartArray() {
```
sdelamo, line 170: why removing the `throws`. Does not throw IOException anymore?

jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
```diff
@@ -104,6 +97,13 @@ private JacksonDatabindMapper(JacksonDatabindMapper from, Argument<?> type) {
         this.specializedWriter = from.createWriter(type);
     }
 
+    private static ObjectMapper createDefaultMapper() {
```
sdelamo: why this change? does it just move the method or I am missing something?

runtime/build.gradle
```diff
@@ -50,7 +50,10 @@ spotless {
 
 tasks.withType(Test).configureEach {
     if (JavaVersion.current().majorVersion.toInteger() >= 17) {
-        logger.warn("Opening java.util, so SystemLambda can work")
-        jvmArgs += ['--add-opens', 'java.base/java.util=ALL-UNNAMED']
+        logger.warn("Opening java.util and java.lang, so SystemLambda can work")
+        jvmArgs += [
+                '--add-opens', 'java.base/java.util=ALL-UNNAMED',
+                '--add-opens', 'java.base/java.lang=ALL-UNNAMED',
```
sdelamo: why is this needed ?

core/src/main/java/io/micronaut/core/io/file/DefaultFileSystemResourceLoader.java
altro3: Problem with pathes in windows.... Path like this `//D:/Wok/....` is wrong

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/AbstractKotlinCompilation.kt
altro3: duplicate OOM_ERROR:
изображение

inject/build.gradle
altro3: without it tests doesn't work on Windows and I see exception with SystemLambda

jackson-core/src/main/java/io/micronaut/jackson/core/tree/TreeGenerator.java
altro3: You don't need to write exceptions from interface if your implementation don't throw them

jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
altro3: checkstyle order error

context/src/main/java/io/micronaut/scheduling/executor/ExecutorType.java
yawkat: no. the old one is correct. i assume it triggers a lint failure because it points to a method that is not available on java 17, but that's fine imo.

runtime/build.gradle
altro3: without it tests doesn't work on Windows and I see exception with SystemLambda

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/MultiObjectBody.java
altro3: Hm... this is maybe a byg in IDEA. I see now this picture:
изображение
But I also don't understand what is wrong...

Reverted

context/src/main/java/io/micronaut/scheduling/executor/ExecutorType.java
altro3: @yawkat hm... yes, it's true. But why it's Ok if we use java 17 and java 17 doesn't have this method in JDK ?

yawkat: this setting does not work on java 17 anyway. only on jdks where this method is available does it work.

altro3: I wonder if there are such JDKs? ok i will revert this change

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
altro3: @yawkat Hi! Could you answer? I think, this is last unanswered question

yawkat: yes

aop/src/main/kotlin/io/micronaut/aop/util/KotlinInterceptedMethodHelper.kt
```diff
@@ -35,8 +35,8 @@ internal object KotlinInterceptedMethodHelper {
     suspend fun handleResult(result: CompletionStage<*>, isUnitValueType: Boolean): Any? = suspendCoroutine { continuation ->
         result.whenComplete { value: Any?, throwable: Throwable? ->
             if (throwable == null) {
-                val result = Result.success(value ?: if (isUnitValueType) Unit else null)
-                continuation.resumeWith(result)
+                val res = Result.success(value ?: if (isUnitValueType) Unit else null)
```
chalkedgoose, line 38: This change seems a bit superfluous is there a reason for renaming the symbol from 
result to res? I mean you could make a case for a more explicit name but in this case 
res does not feel like an improvement over result.

edit: if these changes are related to linting or checkstyle in any way just ignore this 
comment and the other I made below.

test-suite-helper/src/main/java/io/micronaut/testsuitehelper/TestGeneratingAnnotationProcessor.java
```diff
@@ -37,7 +37,7 @@
 @SupportedSourceVersion(RELEASE_17)
 public class TestGeneratingAnnotationProcessor extends AbstractProcessor {
 
-    private boolean executed = false;
+    private boolean executed;
```
chalkedgoose, line 40: There is no concrete difference between these two lines I don't want 
to be overly nit picky but what are the reasons for this change?

aop/src/main/kotlin/io/micronaut/aop/util/KotlinInterceptedMethodHelper.kt
altro3: @chalkedgoose The problem is that you use two variables with the same name - a method parameter and a local variable in the lambda. This makes your code less clear. It's not obvious at what point you're using the parameter, but at what variable. Idea inspection highlighted me this problem, I didn't see it myself. Although, it's strange why sonar didn't catch the smell code before I did

test-suite-helper/src/main/java/io/micronaut/testsuitehelper/TestGeneratingAnnotationProcessor.java
altro3: @chalkedgoose It makes no sense to assign default values to primitives. You don't write an assignment next to every attribute of the class, so why should you assign a default value here? Anyway, it's just a smell code correction.

yawkat: i disagree with the linter, i think in this case `= false` is more clear, it shows the value is initially false and not e.g. set in a constructor. Though it's not a big deal.

altro3: @yawkat Well, then we need to decide: either we rely on the knowledge of java, or we are against abbreviations and write always default values, so that the code is "cleaner and understandable". That is, you need to correct in the other direction, classes, like:
```java
class MyClass {

    private String f1;
    private float f2;
    private Double f3;
    private int f4;
...
}
```
It should be described as follows:
```java
class MyClass {

    private String f1 = null;
    private float f2 = 0F;
    private Double f3 = null;
    private int f4 = 0;
...
}
```
It's getting really weird right now. Everyone knows that the default boolean value is false, and in most cases the default value is not written, and it is written in this place. So what's the difference between this place and the others?

yawkat: the difference is entirely subjective. in some places it's more clear to have an explicit initializer, in others it is not.

altro3: I will just say that everyone decides for themselves what is more convenient for them. I still attribute this change to the code style. If you think this is not a necessary change, no problem, I don't insist ;-)

inject/build.gradle
sdelamo: merged in https://github.com/micronaut-projects/micronaut-core/pull/9807

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/AbstractKotlinCompilation.kt
sdelamo: Merged in https://github.com/micronaut-projects/micronaut-core/pull/9854

jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/9856

runtime/build.gradle
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/9807/files

aop/src/main/kotlin/io/micronaut/aop/util/KotlinInterceptedMethodHelper.kt
sdelamo: I think it is a good change. Having different variable names is more clear.

sdelamo:
Closing this as I think we have incorporated almost every suggestion. @altro3 please reopen separate prs if you think we need them.

PR #9598, opened 2023-07-24
Allow fallback body conversion

### LINKED ISSUE #9583: Request body not available in @Error method
### Expected Behavior

Main controller:

```
@Controller( "/path" )
public class Controller
{
	@Post
	@Consumes( MediaType.APPLICATION_JSON )
	public HttpResponse< Void > create( @Body( ) Request request )
	{
		<...>
	}
}
```

Exception handler:

```
@Controller( "/Exception" )
public class HttpExceptionController
{
	@Error( global = true )
	public HttpResponse< String > GenericException( HttpRequest request, Exception e )
	{
		Optional<String> body = request.getBody( String.class );
	}
}
```

If request body is malformed (`e instanceOf ConversionErrorException`), I expect the body to be present in `GenericException` (this is the case with Micronaut 3.9.4)

### Actual Behaviour

Body is not present

### Steps To Reproduce

Netty server
Send `{}` body to application

### Environment Information

JDK 17

### Example Application

_No response_

### Version

4.0.0

yawkat:
When ImmediateByteBody.processSingle fails, e.g. because of a json syntax error, the body is not "claimed". This patch makes sure the buffer position is also unchanged. This allows binding the body in an error controller to a different type, e.g. `@Body String rawBody`.

It would also be simple to allow access to this body through request.getBody, simply by implementing support for ImmediateByteBody in that method, completely fixing #9583. However my concern with this is that it would also allow this access through getBody e.g. in filters, where the body is sometimes but not always available immediately. Requiring the use of the argument binder allows that binder to handle the streaming case properly by buffering the request body. This ensures the same code will work for both immediate and streaming requests, and won't just fail in production when the streaming case is hit.

Fixes #9583

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerTest.java
@@ -17,6 +17,7 @@ package io.micronaut.http.server.tck.tests;
 
 import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Introspected;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.HttpHeaders;
@@ -31,7 +32,6 @@ import io.micronaut.http.annotation.Get;
 import io.micronaut.http.annotation.Post;
 import io.micronaut.http.annotation.Produces;
 import io.micronaut.http.annotation.Status;
-import io.micronaut.http.codec.CodecException;
 import io.micronaut.http.hateoas.JsonError;
 import io.micronaut.http.hateoas.Link;
 import io.micronaut.http.server.exceptions.ExceptionHandler;
@@ -41,10 +41,10 @@ import io.micronaut.http.tck.ServerUnderTest;
 import io.micronaut.http.tck.ServerUnderTestProviderUtils;
 import io.micronaut.json.JsonMapper;
 import jakarta.inject.Singleton;
-import org.junit.jupiter.api.Test;
-
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.Min;
+import org.junit.jupiter.api.Test;
+
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


graemerocher (approved):

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerTest.java
```diff
@@ -309,17 +331,6 @@ String globalControllerHandlerForStatus() {
 
     }
 
-    @Singleton
```
sdelamo, line 312: why delete this? Would not be better to write a separate test?

sdelamo (changes requested):
If I understand correctly, the body will still not be accessible with `request.getBody(String.class)` in global Exception Handler which is what the user was trying to do.  But after this change it is possible to add `@Body String body`as a method parameter to access the body. 

Can we please document this behaviour in the [error handling part of the documentation(https://docs.micronaut.io/latest/guide/#errorHandling)]?

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerTest.java
yawkat: it did nothing

graemerocher (approved):

sdelamo (commented):
will binding `@Body String body` always work?

yawkat:
no, it will only work when the body is not converted yet

sdelamo:
> no, it will only work when the body is not converted yet

so @Body String body works only on conversion exceptions?

yawkat:
correct

src/main/docs/guide/httpServer/errorHandling/localErrorHandling.adoc
```diff
@@ -11,3 +11,8 @@ snippet::io.micronaut.docs.server.json.PersonController[tags="statusError", inde
 <1> The api:http.annotation.Error[] declares which api:http.HttpStatus[] error code to handle (in this case 404)
 <2> A api:http.hateoas.JsonError[] instance is returned for all 404 responses
 <3> An api:http.HttpStatus#NOT_FOUND[] response is returned
+
+Similar to other controller methods, error handlers can use <<binding, request binding annotations>> on parameters e.g.
+to access header values. However, binding the request body comes with additional restrictions depending on the HTTP
+server implementation used. If the body has already been bound to a parameter of the original controller method, it may
+not be possible to bind the body to a different type, as the original bytes may already have been discarded.
```
sdelamo, line 18: Can you add that for conversion error you can bind the body as a method parameter with `@Body String body`?

sdelamo (changes requested):

src/main/docs/guide/httpServer/errorHandling/localErrorHandling.adoc
yawkat: i dont think we can guarantee this for e.g. servlet. and it may not be the case anymore if we move the netty parsing code to loom at some point

yawkat: or to be more specific: this works right now in netty, because we basically buffer the full input bytes, and then transform them to the object as one, so we still have the bytes if the transform fails. in servlet, you transform an input stream, and dont keep the bytes around. this may also be a desirable approach for netty if it can be achieved without context switching (which will need some loom work).

sdelamo: If I am understanding correctly. In Micronaut Framework 4, we no longer support accessing the body in an exception handler. At least, not in a reliable way. is that correct?

yawkat: it was not reliable before either, i dont think accessing the body as string would work if there was no conversion error in 3.x either.

sdelamo (approved):

PR #9603, opened 2023-07-24
Add @Mapper example to docs

sdelamo:


inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
```diff
@@ -586,8 +586,11 @@ private static Argument<?>[] toArguments(BeanProperty<?, ?>[] writeablePropertie
         private void init(Argument<?>[] arguments) {
             for (int i = 0; i < arguments.length; i++) {
                 Argument<?> argument = arguments[i];
-                argumentIndex.put(argument.getName(), i);
-                defaultValues[i] = argument.getAnnotationMetadata().getValue(Bindable.class, "defaultValue", argument).orElse(null);
+                String name = argument.getName();
+                if (!argumentIndex.containsKey(name)) {
```
sdelamo: I had to add this check because for the Kotlin Data class


```
@Introspected
data class ContactEntity(@Nullable var id: Long? = null, val firstName: String, val lastName: String)
```

similar to a Micronaut Data Mapped Entity. 

We generate 4 arguments. with `id` duplicated. I assume one for the constructor parameter and one for the setter since we use `var` for id.

graemerocher: seems like a bug

PR #9617, opened 2023-07-26
Introspection builder with a class not originally annotated with introspected

sdelamo:
This test tries to verify this works: 

```

@Introspected(classes = Book.class, builder = @Introspected.IntrospectionBuilder(builderClass = Book.Builder.class))
public class BookIntrospected {
```


sdelamo:
@graemerocher this test fails do you know if I should write the test differently or is this not supported?

PR #9599, opened 2023-07-24
doc: Split introspection documentation

sdelamo:

PR #9623, opened 2023-07-26
Fix introspection docs

sdelamo:
We did not correctly remove from core `introspection.adoc` the documentation that we moved into sections here: #9599


src/main/docs/guide/ioc/introspection/introspectExistingAnnotations.adoc
```diff
@@ -1,7 +1,5 @@
-
 If there is an existing annotation that you wish to introspect by default you can write an api:inject.annotation.AnnotationMapper[].
 
 An example of this is https://github.com/micronaut-projects/micronaut-core/blob/master/inject/src/main/java/io/micronaut/inject/beans/visitor/EntityIntrospectedAnnotationMapper.java[EntityIntrospectedAnnotationMapper] which ensures all beans annotated with `javax.persistence.Entity` are introspectable by default.
```
wetted:
```suggestion
An example of this is https://github.com/micronaut-projects/micronaut-core/blob/master/core-processor/src/main/java/io/micronaut/inject/beans/visitor/EntityIntrospectedAnnotationMapper.java[EntityIntrospectedAnnotationMapper] which ensures all beans annotated with `javax.persistence.Entity` are introspectable by default.
```

wetted (changes requested):

wetted (approved):

Issue #6246, opened 2021-09-29
UriBuilder/UriTemplate.PATTERN_FULL_URI: Query Parameters Parsed as Part of Path


DMouse10462:
### Expected Behavior

When using `UriBuilder` to add query parameters to a URI that already has query parameters, the RegEx within `UriTemplate.PATTERN_FULL_URI` should recognize the query parameters as query parameters.

For example, `UriBuilder.of("https://google.com/search?q=TESTING")` should interpret a `path` of `"/search"` and a `queryParams`  with the key/value pair of `[q: "TESTING"]`.

### Actual Behaviour

`DefaultUriBuilder(CharSequence uri)`, calling the PATTERN_FULL_URI matcher, returns the query portion of the URI as part of the path. 

For example, `UriBuilder.of("https://google.com/search?q=TESTING")` interprets a `path` of "/search?q=TESTING" and an empty map for `queryParams`. 

### Steps To Reproduce

1. Build the following:

```java
package example;

import io.micronaut.http.uri.UriBuilder;

public class Application {

    public static void main(String[] args) {
        String testUri = "https://google.com/search?q=Thisisatest";

        UriBuilder builder = UriBuilder.of(testUri);

        builder.queryParam("bad_param", "ThisIsInvalid");

        System.out.println(builder.toString());
    }
}
```
2. (Optional) Break at line 12 (`builder.queryParam(<snip>`) and observe empty map at `builder.queryParams`

3. Observe two question marks in results:
```
https://google.com/search?q=Thisisatest?bad_param=ThisIsInvalid
```

### Environment Information

Operating System: Windows 10 Version 20H2, Build 19042 (x86_64)
Java Version: 1.8

### Version

3.0.2

ashishkujoy:
I would like to raise PR for this.

ashwini-desai:
Since the above PR is still in draft, I would like to attempt this.

Crow-EH:
Hi,

Is there any news on this issue ?

I just got it today, and something like changing `STRING_PATTERN_PATH = "([^#]*)"` to `STRING_PATTERN_PATH = "([^?#]*)"` seems to do the trick in this particular case, but it might break something else as @ashwini-desai said in https://github.com/micronaut-projects/micronaut-core/pull/6320.

sdelamo:
I think this should be fixed via https://github.com/micronaut-projects/micronaut-core/pull/9635

PR #9635, opened 2023-07-28
Issue 6246 - UriBuilder/UriTemplate.PATTERN_FULL_URI: Query Parameters Parsed as Part of Path - simpler approach

asantare:
- separated patterns of UriBuilder from UriTemplate
- added the '?' in the STRING_PATTERN_PATH for the UriBuilder

### DIFF
--- a/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
+++ b/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
@@ -31,9 +31,7 @@ import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.regex.Matcher;
-
-import static io.micronaut.http.uri.UriTemplate.PATTERN_FULL_PATH;
-import static io.micronaut.http.uri.UriTemplate.PATTERN_FULL_URI;
+import java.util.regex.Pattern;
 
 /**
  * Default implementation of the {@link UriBuilder} interface.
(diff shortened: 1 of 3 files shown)


asantare:
fixes #6246 

Respect to the PR 6320 this one decouples the `UriTemplate` regexp patterns from the `UriBuilder` regexp patterns as the scopes are indeed different.

Then a single change is required in the UriBuilder to fix the query parsing.

sdelamo (approved):

asantare:
the errors from the Java CI build on `io.micronaut.http.client.StreamRequestSpec` (if I got the report correctly) doesn't seem related to this change, they seem to occur also without it (at least in my local)

asantare:
some findings out of the scope of this issue but on the failing test:

regarding the test `StreamRequestSpec#"test json stream post request with POJOs flowable error"` there are different errors thrown in `3.9.x` than `4.0.x`.
In `4.0.x` the test is not deterministic (I managed to get failures when running the entire spec file many times).

In `3.9.x` I get a `com.fasterxml.jackson.core.io.JsonEOFException: Unexpected end-of-input` while in `4.0.x` I get a `reactor.core.Exceptions$ReactiveException: java.nio.channels.ClosedChannelException` (probably caused by the channel closing in `PipeliningServerHandler.StreamingOutboundHandler#onError`).

These exceptions are thrown and the test passes as it expects a `RuntimeException`, but when the test fails in `4.0.x`, i.e. no exception is thrown, the `.collectList().block()` succeed (also if I see the warn log from `PipeliningServerHandler.StreamingOutboundHandler#onError`).

I don't know what it is the expected behaviour/wanted exception here but the non-deterministic doesn't seem good.

asantare:
last comment on the failing test I swear :)

if we flush the channel before closing it (in `PipelineServerHandler$StreamingOutboundHandler`, see also image below) the client in the test always gets the processed Books (0, 1 and 2) and continues happily without knowing about the error server side on Book 3.
I don't know how it was in `3.9.x` but it seems to be something missing in the jsonStream protocol or in the client implementation itself (or in the server? just closing the channel without notifying any error).



Here the change flushing the data before closing the channel in `PipelineServerHandler$StreamingOutboundHandler`:

image

yawkat (approved):

yawkat:
@asantare can you make a separate issue about the StreamingOutboundHandler thing? that code really should not be hit.

asantare:
> @asantare can you make a separate issue about the StreamingOutboundHandler thing? that code really should not be hit.

here I tried to summarized it
https://github.com/micronaut-projects/micronaut-core/issues/9652

PR #9660, opened 2023-08-02
Implement workaround for Eclipse JDT compiler

graemerocher:
The Eclipse JDT compiler still has issues with Micronaut's annotation processors and fails for some unknown reason throwing an AbortCompilation exception. See https://github.com/redhat-developer/vscode-java/issues/3223#issuecomment-1661641760 

This implements a workaround for the bug in the Eclipse JDT Compiler.


dstepanov (approved):

dstepanov:
Is there a way we can include a test?

graemerocher:
I don't know how to test this, we would need to embed JDT somehow

PR #9673, opened 2023-08-03
don't publish to latest docs unless in the default branch

sdelamo:
see: https://docs.github.com/en/actions/learn-github-actions/expressions#startswith

see: https://github.com/micronaut-projects/github-pages-deploy-action/blob/master/entrypoint.sh#L106-L117

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,7 +91,7 @@ jobs:
         if: success()
         uses: micronaut-projects/github-pages-deploy-action@master
         env:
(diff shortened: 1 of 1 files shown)


alvarosanchez (approved):

sdelamo:
@alvarosanchez sorry I changed the original contents of this PR to now publish only to latest if the release is running in the default branch. Can you review it again?

I tried this in a personal repo: 

Non default branch

https://github.com/sdelamo/github-actions-tests/actions/runs/5752560676

default branch: 

https://github.com/sdelamo/github-actions-tests/actions/runs/5752507393

I wonder if this will work with tags.

alvarosanchez (approved):

Issue #9670, opened 2023-08-03
KSP: context.visitMetaInfFile(...) generate file in wrong place


altro3:
### Expected Behavior

From micronaut-openapi: whe you try to generate openapi files, files must be created in build directory in META-INF subdir, but that's what I see:
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/18f020c6-bf16-4f48-8386-d8897cd29b35)

and from file manager:
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/5d3de2f5-e5eb-4470-bfc6-9cf250adf728)

As you see, files created in kotlin daemon directory, not in build directory

### Version

4.0.2

altro3:
@dstepanov could you look?

PR #9684, opened 2023-08-05
Some fixes kasp for windows

altro3:


dstepanov:
What does it fix?

altro3:
in micronaut openapi some pathes can conains `/` symbol, windows normally understand both symbols in pathes - `\` and `/`, but when you ask FileSystem  separator java will return you `\`. 

Because of it final path can contain both symbols. Your solution works only in all places will be one of these symbols - `\` or `/`. My solution allow you use symbols `\` and `/` together. This case often happens when part of the path can be set by the user.

Second minor fix (not for windows), just more optimized code - don't need to create arrayList for empty array

altro3:
@dstepanov Hi! What do you think about this fix?

dstepanov:
Does it fix https://github.com/micronaut-projects/micronaut-core/issues/9670 ?

altro3:
@dstepanov no, this fix for windows: when you set part of path with `/`, but system separator is `\`. For example, you have part of path `META-INF/swagger`, but system separator is `\` (windows) - in this case you can't correct split `META-INF/swagger`. My solution fixed it

dstepanov:
Ok, thanks! Are you going to the other meta-inf issue?

altro3:
@dstepanov I have tried. But I still don't understand why the directory is not correct. I see that other micronaut files are created correctly, but in my case, open api files for some reason try to be created in the daemon directory. I tried to debug the code to figure out what's wrong - but I still can't figure it out :-(

PR #9692, opened 2023-08-07
Fix buffer leak when proxying multipart requests

### LINKED ISSUE #9677: ResourceLeakDetector reports missing ByteBuf.release() call before garbage collection frequently
### Expected Behavior

The _ResourceLeakDetector_ should not report any missing _release()_ calls and the service should not run out of memory.

### Actual Behaviour

The application is a proxy taking POST requests containing JSON, manipulate the string, sending the request to the destination server and just returns the answer to the client.
The service answers about 20k requests per hour and runs out of memory after a day or one and a half.

I built the service using several versions of micronaut 3.9.x and 4.0.x.  

The `ResourceLeakDetector` writes the following message to the logs:
`LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:403)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
	io.netty.buffer.UnsafeByteBufUtil.copy(UnsafeByteBufUtil.java:436)
	io.netty.buffer.PooledUnsafeDirectByteBuf.copy(PooledUnsafeDirectByteBuf.java:216)
...`

### Steps To Reproduce

Unfortunately I didn't make it to reproduce the error locally. Starting the service on my Mac didn't produce the leak messages neither. Running load tests didn't cause the local instance to run out of memory.

So the hints I can offer you to get to the bottom of this behavior are the sample data in the repository:
- different leak messages
- heap dumps of two different service instances ran out of memory
- one hour of flight recording
- the system properties produced by the `jcmd <pid> VM.system_properties_ command

### Environment Information

Operating System	Linux 5.0.0-1035-azure
VM Version	OpenJDK 64-Bit Server VM version 17.0.1+12-39 (Java version 17.0.1+12-39)
OS Architecture	amd64
VM Arguments	-XX:+UseZGC -Xmx256M -Xms256M -Dcom.sun.management.jmxremote.port=7501 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.password.file=/opt/apps/proxy-service/jmxremote.password -Dlog4j.configurationFile=/opt/apps/proxy-service/log4j2.xml -Dmicronaut.config.files=/opt/apps/proxy-service/proxy-service.properties

### Example Application

https://github.com/hunsricker/proxyRunningOutOfMemory/tree/master

### Version

3.9.3, 3.9.4, 4.0.0, 4.0.1

yawkat:
Two bugs here:
- The test client would close the connection after the LastHttpContent but before onComplete. This would lead to responseWritten not being called, causing a leak.
- The leak exposed a superfluous release call that happened only sometimes. When a FormRouteCompleter.Claimant is wrapped in a Mono, the Mono cancels the Claimant immediately on onNext. This caused the data to be released early.

Fixes #9677

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/ProxyStressTest.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/ProxyStressTest.groovy
@@ -1,0 +1,67 @@
+package io.micronaut.http.server.netty
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.MediaType
+import io.micronaut.http.MutableHttpResponse
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Part
+import io.micronaut.http.annotation.Post
+import io.micronaut.http.client.HttpClient
+import io.micronaut.http.client.ProxyHttpClient
+import io.micronaut.http.client.multipart.MultipartBody
+import io.micronaut.runtime.server.EmbeddedServer
+import jakarta.inject.Inject
+import org.reactivestreams.Publisher
+import spock.lang.Specification
+
+class ProxyStressTest extends Specification {
+    def 'proxy multipart stress test'() {
+        given:
+        def ctx = ApplicationContext.run(['spec.name': 'ProxyStressTest'])
+        def server = ctx.getBean(EmbeddedServer)
+        server.start()
+        def client = ctx.createBean(HttpClient, server.URI).toBlocking()
+
+        when:
+        for (var i = 0; i < 1000; i++) {
+            client.retrieve(HttpRequest.POST("/proxy", MultipartBody.builder()
+                    .addPart("data", "data")
+                    .addPart("testEventCode", "testEventCode"))
+                    .contentType(MediaType.MULTIPART_FORM_DATA_TYPE))
+        }
+        then:
+        true
+
+        cleanup:
+        ctx.close()
+    }
+
+    @Controller
+    @Requires(property = "spec.name", value = "ProxyStressTest")
+    static class MyController {
+        @Inject
+        ProxyHttpClient proxyHttpClient
+        @Inject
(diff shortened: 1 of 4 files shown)


timyates:
Added to project and added label 😉

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/ProxyStressTest.groovy
timyates, line 1: This test passes both with and without the changes above.

And for both with and without, it writes a lot of `BUG` logs such as

```
10:51:27.138 [default-nioEventLoopGroup-1-6] WARN  i.m.h.client.netty.DefaultHttpClient - BUG - Handler not removed: ChannelHandlerContext(chunk-writer, [id: 0x0336ad04, L:/127.0.0.1:64349 - R:localhost/127.0.0.1:64336])
```

Is this right?

timyates (approved):
Question I'm not sure about

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/ProxyStressTest.groovy
yawkat: it passes w/o the fix because the PipelinedServerHandler exposed the bug. (also i dont remember but it might just be a logged error)

The client error is separate, it's not fatal and i didnt want to fix it because it conflicts with changes in 4.1. i need to check if it's still an issue then

sdelamo: Sorry I don't understand. Is the code added to discard necessary?

yawkat: yes it is necessary but wasn't hit before the other change in this PR

PR #9700, opened 2023-08-11
Support Client Filters Pass in the JDK client

timyates:
Enables Http Client Filters for the JDK client.

### DIFF
--- a/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
+++ b/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
@@ -388,11 +388,11 @@ abstract class AbstractJdkHttpClient {
             .onErrorMap(IOException.class, e -> new HttpClientException("Error sending request: " + e.getMessage(), e))
             .onErrorMap(InterruptedException.class, e -> new HttpClientException("Error sending request: " + e.getMessage(), e))
             .handle((netResponse, sink) -> {
+                if (log.isDebugEnabled()) {
+                    log.debug("Client {} Received HTTP Response: {} {}", clientId, netResponse.statusCode(), netResponse.uri());
+                }
                 boolean errorStatus = netResponse.statusCode() >= 400;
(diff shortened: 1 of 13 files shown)


timyates:
Fixed up the logging as per https://github.com/micronaut-projects/micronaut-core/pull/9710 (so we don't lose it in the merge)

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
```diff
@@ -312,4 +343,64 @@ private <I> Mono<URI> resolveURI(io.micronaut.http.HttpRequest<I> request) {
...
+        io.micronaut.http.HttpRequest<?> request,
+        Argument<O> bodyType
+    ) {
+        return Flux.defer(() -> mapToHttpRequest(request, bodyType)) // defered so any client filter changes are used
+            .map(httpRequest -> {
+                if (log.isDebugEnabled()) {
+                    log.debug("Client {} Sending HTTP Request: {}", clientId, httpRequest);
+                }
+                if (log.isTraceEnabled()) {
+                    httpRequest.headers().map().forEach((k, v) -> log.trace("Client {} Sending HTTP Request Header: {}={}", clientId, k, v));
```
sdelamo: let's use `HttpHeaderUtils` to take into account logging `HEADER_MASK_PATTERNS`

sdelamo: I created a PR https://github.com/micronaut-projects/micronaut-core/pull/9715

sdelamo (changes requested):

yawkat (approved):

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
timyates: 👍

sdelamo (approved):

PR #9710, opened 2023-08-14
fix: JDK HTTP Client error logging on successful response

### LINKED ISSUE #9702: JDK HTTP Client error logging on successful response
## Expected Behavior
HTTP client doesn't error log on successful response

## Actual Behaviour

Started seeing a new error log when using the new JDK http client. 

`11:06:32.974 [ForkJoinPool.commonPool-worker-4] ERROR i.m.h.c.jdk.DefaultJdkHttpClient - Client conversation Received HTTP Response: 200 <Server URL>`

seems to be better suited as a debug log here:
https://github.com/micronaut-projects/micronaut-core/blob/4b4a8ccdebe3bb36843857ae5e5e5965b471dcdc/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/DefaultJdkHttpClient.java#L152

## Steps To Reproduce
Execute a http call to any http server which returns a 200 response for the request.

## Environment Information
JDK 17
Micronaut 4.0.0
Micronaut JDK HTTP Client

## Version
4.0.0

timyates:
Changed to do that same as with the blocking client call.

### DIFF
--- a/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/DefaultJdkHttpClient.java
+++ b/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/DefaultJdkHttpClient.java
@@ -149,9 +149,11 @@ public class DefaultJdkHttpClient extends AbstractJdkHttpClient implements JdkHt
             })
             .flatMap(Mono::fromCompletionStage)
             .map(netResponse -> {
-                log.error("Client {} Received HTTP Response: {} {}", clientId, netResponse.statusCode(), netResponse.uri());
                 boolean errorStatus = netResponse.statusCode() >= 400;
                 if (errorStatus && configuration.isExceptionOnErrorStatus()) {
+                    if (log.isErrorEnabled()) {
+                        log.error("Client {} Received HTTP Response: {} {}", clientId, netResponse.statusCode(), netResponse.uri());
+                    }
                     throw HttpClientExceptionUtils.populateServiceId(new HttpClientResponseException(HttpStatus.valueOf(netResponse.statusCode()).getReason(),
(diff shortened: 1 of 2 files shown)


yawkat:
why should this log? isExceptionOnErrorStatus i would expect the user to handle the exception

timyates:
@yawkat You're right... Switched it to a debug log (if enabled)

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/DefaultJdkHttpClient.java
```diff
@@ -149,7 +149,9 @@ public <I, O, E> Publisher<HttpResponse<O>> exchange(@NonNull HttpRequest<I> req
             })
             .flatMap(Mono::fromCompletionStage)
             .map(netResponse -> {
-                log.error("Client {} Received HTTP Response: {} {}", clientId, netResponse.statusCode(), netResponse.uri());
+                if (log.isDebugEnabled()) {
+                    log.error("Client {} Received HTTP Response: {} {}", clientId, netResponse.statusCode(), netResponse.uri());
```
yawkat: debug

yawkat:
```suggestion
                    log.debug("Client {} Received HTTP Response: {} {}", clientId, netResponse.statusCode(), netResponse.uri());
```

timyates: 🫣

yawkat (approved):

sdelamo (approved):

PR #9711, opened 2023-08-14
Fix KSP properties interception

### LINKED ISSUE #9579: @RequestScope no longer works in micronaut 4.0
### Expected Behavior

@RequestScope continues to work like in micronaut 3.9. The following two features no longer work:
- Default value assignments in Kotlin class constructors of @RequestScope classes
- @RequestScope objects should be freshly initialized for each http request.

### Actual Behaviour

- Default value assignments in Kotlin class constructors are ignored. The value is assigned null. If the type is defined as non-nullable the request fails.
- @RequestScope object behaves like a Singleton and retains state of previous requests.

### Steps To Reproduce

- Assign a default value to a @RequestScope object, e.g:
```
@RequestScope
class RequestScopeClass(
    // this no longer works in micronaut 4.0, default value is ignored and as a result type can't be non-null
    var text: String = "defaultText",
)
```

- Assign a request scoped value and check in a subsequent request if the assigned value was reset.

The example application contains tests for both cases.

### Environment Information

- Linux
- JDK 17

### Example Application

https://github.com/StefanSrf/micronaut-4-0-request-scope-bug

### Version

4.0.0

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/9579

### DIFF
--- a/test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/request_scope/DemoController.kt
+++ b/test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/request_scope/DemoController.kt
@@ -1,0 +1,29 @@
+package io.micronaut.docs.server.request_scope
+
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.QueryValue
+
+@Controller
+class DemoController(
+    private val requestScopeClass: RequestScopeClass
+) {
+
+    @Get("/testEndpoint")
+    fun testEndpoint(
+        @QueryValue("text") text: String?
+    ): HttpResponse<DemoObject> {
+        if (text != null) {
+            requestScopeClass.text = text
+        }
+        requestScopeClass.list.add("listEntry")
+        val list = requestScopeClass.list
+        return HttpResponse.ok(
+            DemoObject(
+                text = requestScopeClass.text ?: "defaultText",
+                list = list
+            )
+        )
+    }
+}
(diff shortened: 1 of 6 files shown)


sdelamo (commented):
There is a test failure https://ge.micronaut.io/s/ykb6cjysjyamc/tests/task/:inject-kotlin:test/details/io.micronaut.kotlin.processing.aop.compile.RequestBeanScope/test%20request%20scope%20bean%20with%20properties?top-execution=1

yawkat (approved):
amazing that removing code was all that was necessary to fix the bug

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/request_scope/TestControllerTest.kt
```diff
@@ -0,0 +1,37 @@
+package io.micronaut.docs.server.request_scope
+
+import io.micronaut.http.client.HttpClient
+import io.micronaut.http.client.annotation.Client
+import io.micronaut.test.extensions.junit5.annotation.MicronautTest
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.RepeatedTest
+import org.junit.jupiter.api.Test
+
+@MicronautTest
```
sdelamo, line 12: Annotate with `@Property(name = "spec.name", value = "TestControllerTest")`

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/docs/server/request_scope/DemoController.kt
```diff
@@ -0,0 +1,29 @@
+package io.micronaut.docs.server.request_scope
+
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.QueryValue
+
+@Controller
```
sdelamo, line 10: Annotate with `@Requires(property = "spec.name", value = "TestControllerTest")`

sdelamo (changes requested):
Please use `spec.name` to avoid bean pollution

sdelamo (approved):

PR #9720, opened 2023-08-15
Fix context propagation in Filters

### LINKED ISSUE #9719: Propagated context happening before security filter phase is cleared when Micronaut Security is enabled
### Expected Behavior

I do not know if this is intended behaviour or a bug. I believe it is a bug and therefore I created it as a bug type. 

We use [context propagation ](https://docs.micronaut.io/latest/guide/#contextPropagation) to propagate some custom tracing information. However when we enable Micronaut Security the context is removed if it is applied before the Micronaut Security Filter phase. Is this intended to prevent some kind of attack or is this a bug? 
I would expect the tracing context to remain around the security concerns. 

### Actual Behaviour

Enabling Micronaut Security remove the propagated context that was added before. 

### Steps To Reproduce

See linked repository for a failing test but in short for some immediate context
```
@Filter(MATCH_ALL_PATTERN)
class CustomTracingFilter : HttpServerFilter {

    override fun getOrder(): Int {
        return ServerFilterPhase.TRACING.after()
        // Replace order to after security to make the test pass
//        return ServerFilterPhase.SECURITY.after()
    }

    override fun doFilter(request: HttpRequest<*>, chain: ServerFilterChain): Publisher<MutableHttpResponse<*>> {
        val trace = request.headers.get("X-Trace")
        PropagatedContext.get().plus(MyPropagationContext(trace)).propagate().use {
            return chain.proceed(request)
        }
    }
}
```
Either by placing the filter after SECURITY or by setting `micronaut.security.enabled=false` will "fix" the test. 

### Environment Information

_No response_

### Example Application

https://github.com/olamelin/micronaut-security-tracing

### Version

4.0.3

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/9719

### DIFF
--- a/http/src/main/java/io/micronaut/http/filter/FilterRunner.java
+++ b/http/src/main/java/io/micronaut/http/filter/FilterRunner.java
@@ -47,7 +47,6 @@ import io.micronaut.http.bind.RequestBinderRegistry;
 import io.micronaut.http.reactive.execution.ReactiveExecutionFlow;
 import io.micronaut.inject.ExecutableMethod;
 import org.reactivestreams.Publisher;
-import reactor.core.publisher.Mono;
 
 import java.util.List;
 import java.util.ListIterator;
@@ -554,11 +553,10 @@ public class FilterRunner {
                     context.propagatedContext,
                     Publishers.convertPublisher(conversionService, returnValue, Publisher.class)
                 );
-                Mono<?> mono = Mono.from(publisher);
                 if (continuation instanceof ResultAwareContinuation resultAwareContinuation) {
-                    return resultAwareContinuation.processResult(mono);
+                    return resultAwareContinuation.processResult(publisher);
                 }
-                return ReactiveExecutionFlow.fromPublisher(mono).flatMap(v -> {
+                return ReactiveExecutionFlow.fromPublisher(publisher).flatMap(v -> {
                     try {
                         return next.handle(context, v, continuation);
                     } catch (Throwable e) {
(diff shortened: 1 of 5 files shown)


sdelamo:
@dstepanov ci failed. Can you check it?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -552,8 +551,8 @@ private static FilterReturnHandler prepareReturnHandler(ConversionService conver
                     return next.handle(context, null, continuation);
                 }
 
-                Mono<?> publisher = Mono.from(Publishers.convertPublisher(conversionService, returnValue, Publisher.class));
-
+                Mono<?> monoPublisher = Mono.from(Publishers.convertPublisher(conversionService, returnValue, Publisher.class));
+                Publisher<?> publisher = ReactivePropagation.propagate(context.propagatedContext, monoPublisher);
```
yawkat: I think the Mono.from should happen after the propagate call

yawkat (approved):

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
sdelamo: @dstepanov can you address this?

dstepanov: @yawkat I don't think we need that mono call anyway

yawkat: yea im not sure. ReactivePropagation.propagate has some logic that branches on Flux vs Mono, which after this PR might hit the different branch. But I doubt there's any impact.

sdelamo (approved):

PR #9722, opened 2023-08-15
Another HTTP filters refactoring

dstepanov:
After https://github.com/micronaut-projects/micronaut-core/pull/9720 I found it's a bit complicated to maintain that code, refactoring to split the main filter runner into multiple classes. Avoid if-instanceof, converted most of the code to private, separate async execution.

Targeting 4.2

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
@@ -26,6 +26,7 @@ import io.micronaut.core.order.OrderUtil;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
+import io.micronaut.core.util.Toggleable;
 import io.micronaut.http.HttpMethod;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.annotation.ClientFilter;
@@ -94,7 +95,7 @@ public class DefaultHttpClientFilterResolver extends BaseFilterProcessor<ClientF
         List<GenericHttpFilter> filterList = new ArrayList<>(filterEntries.size());
         for (FilterEntry filterEntry : filterEntries) {
             final GenericHttpFilter filter = filterEntry.getFilter();
-            if (filter instanceof GenericHttpFilter.AroundLegacy al && !al.isEnabled()) {
+            if (filter instanceof Toggleable toggleable && !toggleable.isEnabled()) {
                 continue;
             }
             if (matchesFilterEntry(method, requestPath, filterEntry)) {
@@ -167,7 +168,7 @@ public class DefaultHttpClientFilterResolver extends BaseFilterProcessor<ClientF
         FilterPatternStyle patternStyle = annotationMetadata.enumValue(Filter.class,
             "patternStyle", FilterPatternStyle.class).orElse(FilterPatternStyle.ANT);
         return new ClientFilterEntry(
-            new GenericHttpFilter.AroundLegacy(httpClientFilter, new FilterOrder.Dynamic(OrderUtil.getOrder(annotationMetadata))),
+            GenericHttpFilter.createLegacyFilter(httpClientFilter, new FilterOrder.Dynamic(OrderUtil.getOrder(annotationMetadata))),
             annotationMetadata,
             methodsForFilter(annotationMetadata),
             patternStyle,
--- a/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
@@ -26,7 +26,6 @@ import io.micronaut.core.order.OrderUtil;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.core.util.StringUtils;
-import io.micronaut.core.util.Toggleable;
 import io.micronaut.http.HttpMethod;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.annotation.ClientFilter;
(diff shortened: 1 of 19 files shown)


http-client-core/src/main/java/io/micronaut/http/client/filter/DefaultHttpClientFilterResolver.java
```diff
@@ -94,7 +95,7 @@ public List<GenericHttpFilter> resolveFilters(HttpRequest<?> request, List<Filte
         List<GenericHttpFilter> filterList = new ArrayList<>(filterEntries.size());
         for (FilterEntry filterEntry : filterEntries) {
             final GenericHttpFilter filter = filterEntry.getFilter();
-            if (filter instanceof GenericHttpFilter.AroundLegacy al && !al.isEnabled()) {
+            if (filter instanceof Toggleable toggleable && !toggleable.isEnabled()) {
```
yawkat: this should remain as-is, using the interface will have worse performance

http/src/main/java/io/micronaut/http/filter/AroundLegacyFilter.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public boolean isFiltersResponse() {
+        return false;
+    }
+
+    @Override
+    public ExecutionFlow<FilterContext> processRequestFilter(FilterContext context,
+                                                             Function<FilterContext, ExecutionFlow<FilterContext>> downstream) {
```
yawkat, line 57: maybe this should be a dedicated interface

http/src/main/java/io/micronaut/http/filter/FilterContext.java
```diff
@@ -0,0 +1,73 @@
...
+        Objects.requireNonNull(request);
+        return new FilterContext(request, response, propagatedContext);
+    }
+
+    public FilterContext withResponse(@NonNull HttpResponse<?> response) {
+        if (this.response == response) {
+            return this;
+        }
+        Objects.requireNonNull(response);
+        // New response should remove the failure
```
yawkat: does it actually do this still? the exception is not passed via the filtercontext

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -76,25 +48,29 @@
...
 
     /**
      * Create a new filter runner, to be used only once.
      *
-     * @param filters           The filters to run
+     * @param filters The filters to run
      */
     public FilterRunner(List<GenericHttpFilter> filters) {
-        this.filters = Objects.requireNonNull(filters, "filters");
+        this.filters = Objects.requireNonNull(filters, "filters").stream().map(f -> {
```
yawkat: replace with an unchecked cast for performance. This stream can never fail because GenericHttpFilter is sealed.

http/src/main/java/io/micronaut/http/filter/GenericHttpFilter.java
```diff
@@ -15,141 +15,71 @@
...
-        @Override
-        public boolean isFiltersException() {
-            return actual.isFiltersException();
-        }
-
-        @Override
-        public int getOrder() {
-            return ((Ordered) actual).getOrder();
-        }
+    static GenericHttpFilter createLegacyFilter(HttpFilter bean, FilterOrder order) {
```
yawkat, line 63: these factory methods can be `@Internal`

router/src/main/java/io/micronaut/web/router/DefaultFilterRoute.java
```diff
@@ -134,7 +135,7 @@ public Optional<GenericHttpFilter> match(HttpMethod method, URI uri) {
         for (String pattern : patterns) {
             if (matcher.matches(pattern, uriStr)) {
                 GenericHttpFilter filter = getFilter();
-                if (filter instanceof GenericHttpFilter.AroundLegacy al && !al.isEnabled()) {
+                if (filter instanceof Toggleable toggleable && !toggleable.isEnabled()) {
```
yawkat: same here

router/src/test/groovy/io/micronaut/web/router/DefaultFilterRouteSpec.groovy
```diff
@@ -32,7 +33,7 @@ class DefaultFilterRouteSpec extends Specification {
 
     void "test filter route matching with no methods specified"() {
         given:
-        def filter = new GenericHttpFilter.AroundLegacy(new HttpFilter() {
+        def filter = new AroundLegacyFilter(new HttpFilter() {
```
yawkat: use factory method

router/src/test/groovy/io/micronaut/web/router/DefaultFilterRouteSpec.groovy
```diff
@@ -55,7 +56,7 @@ class DefaultFilterRouteSpec extends Specification {
 
     void "test filter route matching with methods specified"() {
         given:
-        def filter = new GenericHttpFilter.AroundLegacy(new HttpFilter() {
+        def filter = new AroundLegacyFilter(new HttpFilter() {
```
yawkat: use factory method

router/src/test/groovy/io/micronaut/web/router/DefaultFilterRouteSpec.groovy
```diff
@@ -78,7 +79,7 @@ class DefaultFilterRouteSpec extends Specification {
 
     void "test filter route matching with regex pattern style specified"() {
         given:
-        def filter = new GenericHttpFilter.AroundLegacy(new HttpFilter() {
+        def filter = new AroundLegacyFilter(new HttpFilter() {
```
yawkat: use factory method

yawkat (changes requested):

http/src/main/java/io/micronaut/http/filter/AroundLegacyFilter.java
dstepanov: Gonna leave it for now

yawkat (approved):

PR #9762, opened 2023-08-22
MDCRxJava2Spec is flaky

### LINKED ISSUE #9733: MDCRxJava2Spec is flaky
### Issue description

see: https://ge.micronaut.io/scans/tests?search.timeZoneId=Europe/Madrid&tests.container=io.micronaut.context.propagation.MDCRxJava2Spec

guillermocalvo:
Refactor `MDCRxJava2Spec`

- Log warning messages if the context cannot be found.
- Server filter will create an empty context if it's not there yet.
- Return error messages instead of throwing exceptions, because those that can't be serialized and mask the actual root cause.
- `@Retry` the test 3 times.

### DIFF
--- a/context-propagation/src/test/groovy/io/micronaut/context/propagation/MDCRxJava2Spec.groovy
+++ b/context-propagation/src/test/groovy/io/micronaut/context/propagation/MDCRxJava2Spec.groovy
@@ -58,9 +58,9 @@ class MDCRxJava2Spec extends Specification {
             'mdc.rxjava2test.enabled': true
     ])
 
-    @Shared
-    @AutoCleanup
-    HttpClient client = HttpClient.create(embeddedServer.URL)
+    HttpClient getClient() {
+        return embeddedServer.applicationContext.getBean(HttpClient)
+    }
 
     void "test MDC propagates"() {
         expect:
@@ -68,7 +68,7 @@ class MDCRxJava2Spec extends Specification {
                 .flatMap {
                     String tracingId = UUID.randomUUID()
                     HttpRequest<Object> request = HttpRequest
-                            .POST("/mdc/enter", new SomeBody())
+                            .POST(embeddedServer.URL.toString() + "/mdc/enter", new SomeBody())
                             .header("X-TrackingId", tracingId)
                     return Mono.from(client.retrieve(request)).map(response -> {
                         Tuples.of(tracingId, response)
--- a/context-propagation/src/test/groovy/io/micronaut/context/propagation/MDCRxJava2Spec.groovy
+++ b/context-propagation/src/test/groovy/io/micronaut/context/propagation/MDCRxJava2Spec.groovy
@@ -37,6 +37,7 @@ import reactor.core.publisher.Mono
 import reactor.util.function.Tuple2
 import reactor.util.function.Tuples
 import spock.lang.AutoCleanup
+import spock.lang.Retry
 import spock.lang.Shared
 import spock.lang.Specification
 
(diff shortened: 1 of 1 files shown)


dstepanov:
What is the main problem?

To me, it looks like the HTTP client timeout is not being applied.

guillermocalvo:
@dstepanov 

> What is the main problem?
> 
> To me, it looks like the HTTP client timeout is not being applied.

Not 100% sure, because I could not reproduce the problem locally. But yeah, looking at the failed executions, they seem to be mostly timing out. This will hopefully be mitigated via `@Retry`.

Apart from that, there were some HTTP client exceptions that were masking the root cause. So I made some changes to make controllers return an error message instead of throwing exceptions. These error scenarios will be detected by `assert expected == actual`.

I added some more logging messages to help us understand what's going on when/if the test ever fails again.

Now I'll be re-running the CI jobs a few times, to make sure that it stops randomly failing.

dstepanov:
I did add http timeout conf but might be applied for some reason

sdelamo (approved):

sdelamo:
ping @dstepanov

dstepanov:
I would prefer to analyze why is there a timeout, Maybe we can get 1h timeout or something.

guillermocalvo:
@dstepanov 

> I would prefer to analyze why is there a timeout

The HTTP client used by the test was being created manually, therefore `micronaut.http.client.read-timeout` was being ignored. In order to get a fully configured HTTP client, we need to retrieve it from the running application context (see d5f2f6088686b48c31b1799fd2807fe9238b8309).

sdelamo:
> @dstepanov
> 
> > I would prefer to analyze why is there a timeout
> 
> The HTTP client used by the test was being created manually, therefore `micronaut.http.client.read-timeout` was being ignored. In order to get a fully configured HTTP client, we need to retrieve it from the running application context (see [d5f2f60](https://github.com/micronaut-projects/micronaut-core/commit/d5f2f6088686b48c31b1799fd2807fe9238b8309)).

we can supply configuration while creating a HTTP Client manually  https://github.com/micronaut-projects/micronaut-core/blob/4.1.x/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java#L263

dstepanov:
Ah didn’t realize that. Can we merge it without the retry and see

guillermocalvo:
@sdelamo 

> we can supply configuration while creating a HTTP Client manually https://github.com/micronaut-projects/micronaut-core/blob/4.1.x/http-client-core/src/main/java/io/micronaut/http/client/HttpClient.java#L263

Yeah I know. Retrieving it from the context was the simplest way to make all the HTTP clients use the same config.

guillermocalvo:
@dstepanov 

> Can we merge it without the retry and see

Ok, there you go: 333f676065b3d78fa1456f60b77a67d007abd468

context-propagation/src/test/groovy/io/micronaut/context/propagation/MDCRxJava2Spec.groovy
```diff
@@ -37,6 +37,7 @@ import reactor.core.publisher.Mono
 import reactor.util.function.Tuple2
 import reactor.util.function.Tuples
 import spock.lang.AutoCleanup
+import spock.lang.Retry
```
guillermocalvo:
```suggestion
```

context-propagation/src/test/groovy/io/micronaut/context/propagation/MDCRxJava2Spec.groovy
```diff
@@ -46,6 +47,7 @@ import java.util.concurrent.TimeUnit
 import static io.micronaut.http.annotation.Filter.MATCH_ALL_PATTERN
 import static io.micronaut.scheduling.TaskExecutors.IO
 
+@Retry
```
guillermocalvo:
```suggestion
```

PR #9766, opened 2023-08-23
Performance improvement and cleanup of `PropertySourcePropertyResolver`

dstepanov:
Added hashcode/equals for `ConversionCacheKey` to avoid invokeDynamic overhead
Improvements: style, fast returns, generics, primitive types, collection of size creation

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -41,7 +41,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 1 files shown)


inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -869,5 +859,22 @@ protected enum PropertyCatalog {
...
+            if (o == null) {
+                return false;
+            }
+            ConversionCacheKey that = (ConversionCacheKey) o;
+            return Objects.equals(name, that.name) && Objects.equals(requiredType, that.requiredType);
+        }
+
+        @Override
+        public int hashCode() {
+            return name.hashCode();
```
yawkat, line 877: why only the name?

yawkat (approved):
im going to trust you on the style changes, this diff is awful to read :laughing:

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
dstepanov: I assume that in most cases, the property is represented by one type

graemerocher (approved):

PR #9767, opened 2023-08-23
Route without at body

sdelamo:


sdelamo:
I tried to build a route without an `@Body` parameter and do `request.getBody` within the method. see new test in this PR. However, `request.getBody(` returns null. 


@yawkat @dstepanov do you know what could be the culprit?

timyates:
I thought the `@Body` annotation was required as of Micronaut 4?

sdelamo:
> I thought the `@Body` annotation was required as of Micronaut 4?

but please note the failing test has  no parameter other than `HttpRequest<?>`

timyates:
@yawkat this seems to be because the body is a `ImmediateByteBody` which isn't handled here

https://github.com/micronaut-projects/micronaut-core/blob/51650efcb97939ef02bf726bdcb28266b7ca20d1/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java#L437-L452

Any ideas?

The form post content is encoded in this buffer correctly it seems 🤔 

image

yawkat:
ok i think what's happening is that FormRouteCompleter is marking the route as ready for execution before all data is processed. This is imo the correct behavior. There is nothing in the route to indicate we should wait for all fields, so we don't.

This could easily be changed in this test case because it's an ImmediateByteBody, but I don't want that because it would mean a behavioral difference to StreamingByteBody which should be avoided.

The workaround of making a `@Body Map` is the correct approach imo.

timyates:
@sdelamo Are you ok with closing this as wontfix?

timyates:
> @sdelamo Are you ok with closing this as wontfix?

ping!

yawkat:
he can always reopen it :)

PR #9770, opened 2023-08-23
Conversion service improvements

dstepanov:
- Introduced internal not-concurrent map
- Added common source type implementation converter variations
- Added missing converters to avoid runtime creation
- Created a way to convert an object from a specific type, allowing to bypass looking for the implementation converter
- Added a common preset class hierarchy, avoiding reflection call

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -16,6 +16,7 @@
 package io.micronaut.runtime.converters.time;
 
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NextMajorVersion;
 import io.micronaut.core.annotation.TypeHint;
 import io.micronaut.core.convert.ConversionContext;
 import io.micronaut.core.convert.MutableConversionService;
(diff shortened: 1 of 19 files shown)


graemerocher (approved):

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -81,6 +82,7 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
     private static final Pattern DURATION_MATCHER = Pattern.compile("^(-?\\d+)([unsmhd])(s?)$");
     private static final int MILLIS = 3;
 
+    @NextMajorVersion("Consider deletion of LocalDate and LocalDateTime converters")
```
yawkat, line 85: whats the reason?

core/src/main/java/io/micronaut/core/convert/DefaultMutableConversionService.java
```diff
@@ -94,9 +105,112 @@ public class DefaultMutableConversionService implements MutableConversionService
...
+
+    /**
+     * The additional converters which are variations of different converters.
+     */
     private final Map<ConvertiblePair, TypeConverter> converterCache = new ConcurrentHashMap<>();
 
+    /**
+     * The mutable conversion service which is adding new converters to the internal collection which is not synchronized.
+     */
+    private final MutableConversionService internalMutableConversionService = new MutableConversionService() {
```
yawkat, line 156: would be nice to separate out the methods into a new interface. if we could ensure a single implementation of ConversionService only, that would help native image optimize it. but not in this pr.

yawkat (approved):

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
dstepanov: There is a forgotten TODO, I have added the annotation so it will be easier to find

PR #9771, opened 2023-08-23
Implement enum constant in kotlin element factory

altro3:

PR #9774, opened 2023-08-24
Implement enum constant in kotlin element factory

altro3:


altro3:
@dstepanov Hi! Could you help with this change? Without it we can't use micronaut-openapi with KSP for enums. But I did something wrong and now I have this exception in test:

изображение

Need help!

altro3:
@dstepanov Continue #9771 here

dstepanov:
Please add test

altro3:
@dstepanov done, but this test works fine.... When I try to use this snapshot inside test project, i have this error:
изображение

dstepanov:
I will check

altro3:
@dstepanov I figured out what the problem was and fixed it. Now everything works correctly, checked with micronaut-openapi :-)

dstepanov:
Thanks!

PR #9778, opened 2023-08-25
Improve bean qualifiers

dstepanov:
Introducing new methods that allow to skip the stream processing

### DIFF
--- a/inject/src/main/java/io/micronaut/context/Qualifier.java
+++ b/inject/src/main/java/io/micronaut/context/Qualifier.java
@@ -18,6 +18,7 @@ package io.micronaut.context;
 import io.micronaut.context.annotation.Primary;
 import io.micronaut.inject.BeanType;
 
+import java.util.Collection;
 import java.util.Optional;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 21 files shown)


inject/src/main/java/io/micronaut/context/Qualifier.java
```diff
@@ -65,4 +66,41 @@ default boolean contains(Qualifier<T> qualifier) {
...
+
+    /**
+     * Check if the candidate qualifies.
+     *
+     * @param beanType  The bean type
+     * @param candidate The candidate
+     * @return true if matches
+     * @since 4.2.0
+     */
+    default boolean isQualifies(Class<T> beanType, BeanType<T> candidate) {
```
yawkat: maybe doesQualify? @graemerocher naming suggestions please :D

yawkat (approved):

inject/src/main/java/io/micronaut/context/Qualifier.java
graemerocher: either `doesQualifiy` is `isQualified`

dstepanov: renamed

graemerocher (approved):

PR #9779, opened 2023-08-25
Changing Jakarta provide to `BeanProvider` as it's lazy

dstepanov:


yawkat (approved):

http/src/main/java/io/micronaut/http/converters/HttpConverterRegistrar.java
```diff
@@ -36,14 +36,14 @@
...
-    private final Provider<ResourceResolver> resourceResolver;
+    private final BeanProvider<ResourceResolver> resourceResolver;
 
     /**
      * Default constructor.
      *
      * @param resourceResolver The resource resolver
      */
-    protected HttpConverterRegistrar(Provider<ResourceResolver> resourceResolver) {
+    protected HttpConverterRegistrar(BeanProvider<ResourceResolver> resourceResolver) {
```
graemerocher, line 49: is this internal? If not need to keep the constructor, deprecate the old one and add `@Inject` to this one

dstepanov: Corrected. Those classes should be internal

http/src/main/java/io/micronaut/http/converters/HttpConverterRegistrar.java
```diff
@@ -36,16 +38,31 @@
...
      */
-    protected HttpConverterRegistrar(Provider<ResourceResolver> resourceResolver) {
+    @Inject
+    protected HttpConverterRegistrar(BeanProvider<ResourceResolver> resourceResolver) {
         this.resourceResolver = resourceResolver;
     }
+    /**
+     * The constructor.
+     *
+     * @param resourceResolver The resource resolver
```
sdelamo, line 56: it misses @deprecated javadoc

sdelamo (changes requested):

graemerocher (approved):

PR #9784, opened 2023-08-28
Fix processing order of introduced method in Java

dstepanov:
Originally wanted to do this for 4.0 but didn't manage. Now the Java type element visitor is using the AST to process the hierarchy, in the same way as Groovy and Kotlin.

+ Groovy visitor improvement

Fixes https://github.com/micronaut-projects/micronaut-data/issues/2465

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/websocket/QueryParamClientWebSocket.java
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/websocket/QueryParamClientWebSocket.java
@@ -1,6 +1,7 @@
 package io.micronaut.http.server.netty.websocket;
 
 import io.micronaut.http.HttpRequest;
+import io.micronaut.http.annotation.QueryValue;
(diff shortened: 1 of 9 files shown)


graemerocher (approved):

dstepanov:
Looks like this fixed some validation:
```
:http-server-netty:compileTestGroovy FAILED
/home/runner/work/micronaut-core/micronaut-core/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/websocket/QueryParamClientWebSocket.java:17: error: Parameter to @OnOpen must either be a URI variable, a WebSocketSession , the HttpRequest, or annotated with an HTTP binding annotation (such as @Header)
    public void onOpen(String dinner, WebSocketSession session, HttpRequest request) { // <3>
                              ^
Note: Creating bean classes for 37 type elements
1 error
startup failed:
Compilation failed; see the compiler error output for details.

1 error
```

Should we just remove the parameter? @yawkat

yawkat:
add QueryParam maybe

PR #9789, opened 2023-08-28
if multiple JsonBodyWritter picks the one which is writable

sdelamo:
currently, first will be picked even if `isWritable` returns false

related to: https://github.com/micronaut-projects/micronaut-views/pull/590


yawkat (approved):

yawkat:
ive refactored the nested stream a bit to make it more readable (imo)

sdelamo:
> ive refactored the nested stream a bit to make it more readable (imo)

it looks good to me. However, I have added a commit to protect against an empty list: 

https://github.com/micronaut-projects/micronaut-core/pull/9789/commits/bf9cfb996675178349b874d069f34866391d6b8e

yawkat:
i did too :D

graemerocher (approved):

PR #9793, opened 2023-08-29
Update kotlin to 1.9.10

sdelamo:


sdelamo:
@timyates see if you can make it pass.

timyates:
Starting point https://ge.micronaut.io/s/ffg3lcvxu5vvy

```
:inject-kotlin-test:compileKotlin FAILED
:test-suite-kotlin:kaptTestKotlin FAILED
:context-propagation:compileTestKotlin FAILED
```

timyates:
Ok, one failure

https://ge.micronaut.io/s/a4hp6gehhurja/tests/overview?outcome=FAILED

Hopefully fixed by https://github.com/micronaut-projects/micronaut-core/pull/9793/commits/040041ab49f94b107b780f8301bf78c08b5065ae

timyates:
@melix Any idea why we're seeing

```
The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build. 	
This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.	
Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.	
If the parent project does not need the plugin, add 'apply false' to the plugin line.	
See: https://docs.gradle.org/current/userguide/plugins.html#sec:subprojects_plugins_dsl	
The Kotlin plugin was loaded in the following projects: ':aop', ':inject-kotlin'
```

https://ge.micronaut.io/s/x33xdbydcmbke/console-log?page=1#L24

What it says doesn't seem to be the case 🤔 

## Edit

Is this the fix? https://github.com/micronaut-projects/micronaut-core/pull/9793/commits/6b28a0fa2adad2b7c02d7032eadbbebf03172bd7

dstepanov:
You need to update gradle.properties

timyates:
> You need to update gradle.properties

Which bit? I've already bumped the kotlin version 🤔

dstepanov:
The bit that says kotlinVersion

timyates:
@dstepanov I already had done in https://github.com/micronaut-projects/micronaut-core/pull/9793/commits/f33adbf0ddd38d6ce04dc3a814d2f86d5294cd2e

And now I'm seeing if it works with it removed in https://github.com/micronaut-projects/micronaut-core/pull/9793/commits/6b28a0fa2adad2b7c02d7032eadbbebf03172bd7

dstepanov:
Sorry I didn't see it in diff

timyates:
Going to investigate the deprecations

https://ge.micronaut.io/s/xwtd572yhxyby/console-log?page=1#L149

### later...

So it looks like the ComponentRegistrar [deprecation/removal has been delayed](https://youtrack.jetbrains.com/issue/KT-52665/Deprecate-ComponentRegistrar#focus=Comments-27-7999959.0-0) so I guess we can ignore for now.

test-suite-kotlin/src/test/kotlin/io/micronaut/core/beans/KotlinBeanIntrospectionSpec.kt
timyates, line 1: This test no longer throws an exception. I am not sure why...maybe the KAPT update?

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/AbstractKotlinCompilation.kt
timyates, line 1: This label was duplicated, I removed the general one and left the more specific

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/AbstractKotlinCompilation.kt
```diff
@@ -293,7 +293,6 @@ abstract class AbstractKotlinCompilation<A : CommonCompilerArguments> internal c
 
 internal fun convertKotlinExitCode(code: ExitCode) = when(code) {
     ExitCode.OK -> KotlinCompilation.ExitCode.OK
-    ExitCode.OOM_ERROR,
```
sdelamo, line 296: why are we removing this?

test-suite-kotlin/src/test/kotlin/io/micronaut/core/beans/KotlinBeanIntrospectionSpec.kt
dstepanov: I have tried and the exception is `io.micronaut.core.reflect.exception.InstantiationException: Argument count [2] doesn't match required argument count: 0`
Maybe the constructor order is different.
I see the same test without the exception in KSP so it should be fine

dstepanov (approved):

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/AbstractKotlinCompilation.kt
timyates: As per [the file comment](https://github.com/micronaut-projects/micronaut-core/pull/9793#discussion_r1310094295), we had duplicate labels in the when (we had OOM_ERROR twice)

I removed this first one, so it will throw an OOM error instead of a generic one

sdelamo:
ping @melix

sdelamo:
I would like to merge this PR into 4.2.x

 @melix or @graemerocher would you like to review it?

melix (approved):

PR #9809, opened 2023-08-31
DefaultApplicationContext should close an Environment that it created

Spikhalskiy:
This PR fixes an issue with `DefaultApplicationContext` not closing an environment it created.
This addresses the problem that was attempted to be fixed by a [stale PR](https://github.com/micronaut-projects/micronaut-test/issues/615), but does it according to the advice in the PR review: https://github.com/micronaut-projects/micronaut-test/pull/582#issuecomment-1132715214

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
@@ -69,6 +69,11 @@ public class DefaultApplicationContext extends DefaultBeanContext implements App
     private final ClassPathResourceLoader resourceLoader;
     private final ApplicationContextConfiguration configuration;
     private Environment environment;
+    /**
+     * True if the {@link #environment} was created by this context,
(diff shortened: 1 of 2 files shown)


dstepanov:
Please add a test

Spikhalskiy:
@dstepanov added the test. Checked that it fails without the change and passes with the change.

inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -200,7 +209,9 @@ protected void registerConversionService() {
     public synchronized @NonNull
     ApplicationContext stop() {
         ApplicationContext stop = (ApplicationContext) super.stop();
-        environment = null;
```
Spikhalskiy, line 203: I removed setting the environment to null here. Otherwise, if someone calls context.stop() and context.getEnvironment, a new environment is getting created, which is not right.

Spikhalskiy:
@dstepanov this is ready for you!

dstepanov (approved):

PR #9816, opened 2023-09-01
Add a system property to prohibit dynamic service loading

yawkat:
This property can be used to confirm that all services are optimized by micronaut-aot and none are loaded dynamically.

### DIFF
--- a/core/src/main/java/io/micronaut/core/optim/AotHelper.java
+++ b/core/src/main/java/io/micronaut/core/optim/AotHelper.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/micronaut/core/optim/AotHelper.java
```diff
@@ -0,0 +1,48 @@
...
+import io.micronaut.core.annotation.Internal;
+
+/**
+ * AOT-specific helper.
+ *
+ * @since 4.2.0
+ * @author Jonas Konrad
+ */
+@Internal
+public final class AotHelper {
```
graemerocher: move it to the same package as SoftServiceLoader and make package private

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/optim/AotHelper.java
yawkat: my intention was to add more similar flags and collect them here but maybe that's overkill. ive moved the class.

dstepanov:
I think @melix mentioned that the application context configurer cannot be optimized by AOP right now.

melix:
That's correct. There's a chicken and egg problem I haven't figured out yet.

sdelamo:
should this PR be in draft? What is the status?

Issue #7552, opened 2022-06-06
Property endpoints.all.path is concat with micronaut.server.context-path


weesftw:
### Expected Behavior

It would be attractive to keep them on separate paths, as they are different contexts. Or have some variable that allows you to make this distinction because they are totally different routes.

Inside the processors package in the `management` module, in class [AbstractEndpointRouteBuilder](https://github.com/micronaut-projects/micronaut-core/blob/c23b826872bd4afa72d3267d1229b399e0d5e204/management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java#L166) there is the `resolveUriByRouteId()` method returning `uriNamingStrategy.resolveUri()` which does the proper concatenation. If it makes sense for the Micronaut context, would allowing a configuration that changes this behavior for greater flexibility be interesting? I'm glad in conttribute.

### Actual Behaviour

Management endpoints such as `/health` would be available at the URI `localhost:9090/foo/fff/health` instead of `localhost:9090/fff/health`.

There is an [issue](https://github.com/micronaut-projects/micronaut-core/issues/2090) that has been closed similar to this issue, but with no apparent conclusion.

### Steps To Reproduce

Set this properties in application.yml:

```
micronaut.server.context-path=/foo
endpoints.all.path=/fff
```

### Environment Information

Ubuntu 22.04
Java 11

### Version

3.5.x

agugan:
I am willing to pick this up. @graemerocher

vexxuh:
I don't see a resolution to this issue. I have not worked on this project in a while so I don't mind grabbing this if @agugan has not worked on a resolution!

graemerocher:
feel free

ginaesps:
Hello! Is this issue already solved, or could I take it?

graemerocher:
Sure go for it

ginaesps:
I'm new in here and have read documentation. Where would you recommend me to start so I can get a better grasp of the issue?
And, is there a Slack/Discord or some other kind or community where I can ask for help?

graemerocher:
there is a gitter/matrix community https://micronaut.io/2022/04/13/how-to-access-the-micronaut-gitter-community-from-the-element-matrix-client/

bpstelios10:
👀

bpstelios10:
hi @graemerocher. Regarding this issue, do we really want it? we had a discussion with @yawkat and we feel it is debatable.
[...]

PR #9820, opened 2023-09-02
#7552 Allow adding a custom context for path management

viniciusxyz:
#7552 The idea of ​​this MR is to create the endpoints.all.context.path property that can be used to separate the management context-path from the server.context-path

With this change, the behavior of the paths is as follows:

**config:**

micronaut.server.context-path=/foo
endpoints.all.context-path=/fff
endpoints.all.path=/eee

**Now:**

Default app -> http://localhost:8080/foo
Management -> http://localhost:8080/fff/eee

**Before**

Default app -> http://localhost:8080/foo
Management -> http://localhost:8080/foo/eee

**config:**

micronaut.server.context-path=/foo
endpoints.all.path=/eee

**Now:**

Default app -> http://localhost:8080/foo
Management -> http://localhost:8080/foo/eee

**Before**

Default app -> http://localhost:8080/foo
Management -> http://localhost:8080/foo/eee

With the addition of this property there should be no change in behavior in relation to the current properties.

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
@@ -41,6 +41,11 @@ public class EndpointDefaultConfiguration {
     public static final String PATH = "endpoints.all.path";
 
     /**
+     * The context for endpoints settings.
+     */
+    public static final String CONTEXT_PATH = "endpoints.all.context.path";
+
+    /**
      * The path for endpoints settings.
      */
     public static final String PORT = "endpoints.all.port";
@@ -50,10 +55,16 @@ public class EndpointDefaultConfiguration {
      */
     public static final String DEFAULT_ENDPOINT_BASE_PATH = "/";
 
+    /**
+     * The default context path.
+     */
+    public static final String DEFAULT_ENDPOINT_CONTEXT_PATH = null;
+
     private Boolean enabled;
     private Boolean sensitive;
     private Integer port;
     private String path = DEFAULT_ENDPOINT_BASE_PATH;
+    private String contextPath = DEFAULT_ENDPOINT_CONTEXT_PATH;
 
     /**
      *
@@ -61,6 +72,14 @@ public class EndpointDefaultConfiguration {
      */
     public String getPath() {
         return path;
+    }
+
+    /**
+     *
+     * @return endpoints Context Path (defaults is null)
+     */
+    public @Nullable String getContextPath() {
+        return contextPath;
     }
 
     /**
@@ -102,6 +121,16 @@ public class EndpointDefaultConfiguration {
     public void setPath(String path) {
         if (StringUtils.isNotEmpty(path)) {
             this.path = path;
+        }
+    }
+
+    /**
+     * The endpoints context path. Default value is null.
+     * @param contextPath The Context Path
+     */
+    public void setContextPath(String contextPath) {
+        if (StringUtils.isNotEmpty(contextPath)) {
+            this.contextPath = contextPath;
         }
     }
 
--- a/management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
@@ -168,7 +168,10 @@ abstract class AbstractEndpointRouteBuilder extends DefaultRouteBuilder implemen
         if (path.charAt(0) == '/') {
             path = path.substring(1);
         }
-        return uriNamingStrategy.resolveUri(path);
+        var completePath = endpointDefaultConfiguration.getContextPath() == null
+            ? uriNamingStrategy.resolveUri(path)
+            : NameUtils.hyphenate(StringUtils.prependUri(endpointDefaultConfiguration.getContextPath(), path));
+        return completePath.charAt(0) == '/' ? completePath : "/".concat(completePath);
     }
 
     /**
--- a/http/src/main/java/io/micronaut/http/annotation/ServerFilter.java
+++ b/http/src/main/java/io/micronaut/http/annotation/ServerFilter.java
@@ -71,4 +71,9 @@ public @interface ServerFilter {
      * @return The methods to match. Defaults to all
      */
     HttpMethod[] methods() default {};
+
+    /**
+     * @return Whether the contextPath should be concatenated into the filter pattern
+     */
+    boolean appendContextPath() default true;
 }
--- a/http/src/main/java/io/micronaut/http/filter/BaseFilterProcessor.java
+++ b/http/src/main/java/io/micronaut/http/filter/BaseFilterProcessor.java
@@ -168,7 +168,7 @@ public abstract class BaseFilterProcessor<A extends Annotation> implements Execu
                 .toList();
         }
 
-        if (patterns != null) {
+        if (patterns != null && (beanLevel.appendContextPath == null || beanLevel.appendContextPath)) {
             patterns = prependContextPath(patterns);
         }
 
@@ -189,7 +189,8 @@ public abstract class BaseFilterProcessor<A extends Annotation> implements Execu
             order,
             methodLevel.executeOn == null ? beanLevel.executeOn : methodLevel.executeOn,
             beanLevel.serviceId, // only present on bean level
-            beanLevel.excludeServiceId // only present on bean level
+            beanLevel.excludeServiceId, // only present on bean level
+            beanLevel.appendContextPath // Define if contextPath is appended
         );
     }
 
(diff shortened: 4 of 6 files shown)


sdelamo (changes requested):
This will be a breaking change.

graemerocher:
not sure this makes sense to be honest. There should be a separate context path setting for endpoints in which case it wouldn't be breaking

graemerocher:
To make this non-breaking IMO we should support `endpoints.all.context-path`

viniciusxyz:
> To make this non-breaking IMO we should support `endpoints.all.context-path`

I will try to update MR over the weekend by adding this property.

viniciusxyz:
> > To make this non-breaking IMO we should support `endpoints.all.context-path`
> 
> I will try to update MR over the weekend by adding this property.

I apologize, but due to Black Friday I had to work on the weekend and was unable to make the necessary changes, I will try to make it as soon as possible

viniciusxyz:
@sdelamo I made the changes based on the comments made and instead of changing the behavior of **endpoints.all.path** the idea is to add the **endpoints.all.context-path** property, I am finishing the implementation, but I came across a difficulty that I still haven't been able to overcome, for some reason when adding the new property the **EndpointsFilter** filter is never executed, I'm investigating this and as soon as possible I'll add the documentation and remove the draft PR.

If you have any idea why a filter with the /** pattern is not filtering a request after adding a path, the tip would be very welcome.

management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
```diff
@@ -50,10 +55,16 @@ public class EndpointDefaultConfiguration {
...
+    /**
+     * The default context path.
+     */
+    public static final String DEFAULT_ENDPOINT_CONTEXT_PATH = "";
+
     private Boolean enabled;
     private Boolean sensitive;
     private Integer port;
     private String path = DEFAULT_ENDPOINT_BASE_PATH;
+    private String contextPath = DEFAULT_ENDPOINT_CONTEXT_PATH;
```
sdelamo, line 67: Can you please this default to `null`

management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
```diff
@@ -63,6 +74,14 @@ public String getPath() {
         return path;
     }
 
+    /**
+     *
+     * @return endpoints Context Path (defaults to: {@value #DEFAULT_ENDPOINT_CONTEXT_PATH})
+     */
+    public String getContextPath() {
```
sdelamo: annotate the return type `@Nullable`

management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
```diff
@@ -105,6 +124,16 @@ public void setPath(String path) {
         }
     }
 
+    /**
+     * The endpoints context path. It must include a leading and trailing '/'. Default value ({@value #DEFAULT_ENDPOINT_CONTEXT_PATH}).
+     * @param contextPath The Context Path
+     */
+    public void setContextPath(String contextPath) {
```
sdelamo, line 131: since contextPath must start with `/` and end with `/`, validate it here and throw a `ConfigurationException` if invalid.

management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
```diff
@@ -168,7 +168,10 @@ String resolveUriByRouteId(String id) {
         if (path.charAt(0) == '/') {
             path = path.substring(1);
         }
-        return uriNamingStrategy.resolveUri(path);
+        var completePath = endpointDefaultConfiguration.getContextPath().isBlank()
```
sdelamo:
```suggestion
        var completePath = endpointDefaultConfiguration.getContextPath() == null
```

sdelamo (changes requested):
Can you add a controller to the test to verify the controller is still accessible in the expected path?

Moreover, can you modify your test to test an endpoint which does not return 401. The Unauthorized makes the test confusing.

sdelamo:
can you merge 4.3.x against your branch. I changed the base to be 4.3.x

management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
viniciusxyz: Done

management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
viniciusxyz: Done

management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
viniciusxyz: Done

management/src/main/java/io/micronaut/management/endpoint/EndpointDefaultConfiguration.java
viniciusxyz: In fact, the comment is wrong, there is no need for it, I even believe that the comment about the context path for standard endpoints is also wrong.

viniciusxyz:
> Can you add a controller to the test to verify the controller is still accessible in the expected path?
> 
> Moreover, can you modify your test to test an endpoint which does not return 401. The Unauthorized makes the test confusing.

As the test refers to the context path of the management endpoints and they are configured as sensitive by default, I have to guarantee that even with the change in context there is no change in the security model that the framework adds today, which is why I am following the applied testing model in the [EndpointsBasePathSpec](https://github.com/micronaut-projects/micronaut-core/blob/4.3.x/management/src/test/groovy/io/micronaut/management/endpoint/EndpointsBasePathSpec.groovy#L63) test which has a very similar objective, I still haven't been able to define why the security tests aren't passing, it's as if there was some trigger to activate the [EndpointsFilter](https://github.com/micronaut-projects/micronaut-core/blob/4.3.x/management/src/main/java/io/micronaut/management/endpoint/EndpointsFilter.java) that I can't see

viniciusxyz:
> can you merge 4.3.x against your branch. I changed the base to be 4.3.x

I will do it tomorrow, I tried today, but there were a series of conflicts.

viniciusxyz:
@sdelamo There seems to be a rule somewhere that when **micronaut.server.context-path** is defined the **EndpointsFilter** is only called when executing a path containing that context, in my unit test **" test routes with a server context path and endpoints context path"** when I define **micronaut.server.context-path** as **/** or remove the property, the test passes because I see it with a break point **EndpointsFilter** doFilter method is triggered in this case, for any other value I pass in **micronaut.server.context-path** the **EndpointsFilter**'s doFilter is never fired and therefore the endpoint is insecure . Can you tell me if there really is any rule like that? I'm confused because a filter that has the pattern `/**` should activate doFilter for any call. Or could I have misunderstood and really in Micronaut the Filter with pattern `/**` will always consider it as `{contextPath}/**` ?

viniciusxyz:
@sdelamo The problem currently is that when defining **micronaut.server.context-path** and **endpoints.all.context-path** as in [this test](https://github.com/micronaut-projects/micronaut-core/pull/9820/files#diff-a7e03f43d7a65b07d3577352e534d84a6decf667a08c5939173bee737fabb0c4R64), **EndpointsFilter -> doFilter** is never called, this occurs because of this code snippet [ServerFilterRouteBuilder.java](https://github.com/micronaut-projects/micronaut-core/blob/4.2.x/router/src/main/java/io/micronaut/web/router/ServerFilterRouteBuilder.java#L68).

As a mental model, let’s think about the following:

Let's assume we have an application configured with Prometheus and the following properties:

- micronaut.server.context-path: /any
- endpoints.all.context-path: /test
- endpoints.prometheus.sensitive: true

1. A new enhancement of [AbstractEndpointRouteBuilder.java](https://github.com/micronaut-projects/micronaut-core/pull/9820/files#diff-295073bce4e6713906436635c42659f1c94c8fa58b80b719ea3534ede779e43f) creates the path as **/test/prometheus** ignoring the context path as it is what is desired.

2. A class [ServerFilterRouteBuilder](https://github.com/micronaut-projects/micronaut-core/blob/4.2.x/router/src/main/java/io/micronaut/web/router/ServerFilterRouteBuilder.java#L68) changes the default pattern `/**` of **EndpointsFilter** to ` /any/**`

This causes calls to the prometheus point, even if they are defined as sensitive true, to still be displayed without security. This is due to the filter's behavior. Is there any filter that I can use that does not have this behavior of append the contextPath?

viniciusxyz:
@sdelamo Since I couldn't get the filter to ignore the ContextPath in the filters, I added a new property to the ServerFilter to establish whether the ContextPath should be added or not. With this and adding the option to not put ContextPath in the security filter now the solution works and all tests pass.

If you have any comments about the solution or need me to modify or test something, please let me know.

altro3:
Any news about this improvement?

viniciusxyz:
> Any news about this improvement?

I'm still waiting for approval :)

altro3:
@graemerocher @sdelamo ping

graemerocher:
it is scheduled for 4.5

graemerocher:
Merged with https://github.com/micronaut-projects/micronaut-core/pull/10870 

Thanks for the contribution

PR #9838, opened 2023-09-07
Improve default annotation values

dstepanov:
Added Core annotations default values to prevent repeating them in bean definitions.
Fixed a bug with missing defaults for nested annotations.
Changed for `AnnotationValues` to have default values fetched lazily, preventing additional calls to the concurrent map during the class init.

We might want to remove additional calls to fill `Map<String, Class<? extends Annotation>> ANNOTATION_TYPES` and make that resolution at the runtime. In don't think `AnnotationMetadata` `Optional<Class<? extends Annotation>> getAnnotationType(@NonNull String name)` is very much used.

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationDefaultValuesProvider.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationDefaultValuesProvider.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.annotation;
+
+import java.util.Map;
+
+/**
+ * The annotation default values provider.
+ *
+ * @author Denis Stepanov
+ * @since 4.3.0
+ */
+public interface AnnotationDefaultValuesProvider {
+
+    /**
+     * @param annotationName The annotation name
+     * @return The annotation default values
+     */
+    @NonNull
+    Map<CharSequence, Object> provide(String annotationName);
+
+}
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
@@ -68,11 +68,14 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
     private final ConvertibleValues<Object> convertibleValues;
     private final Map<CharSequence, Object> values;
     @Nullable
-    private final Map<CharSequence, Object> defaultValues;
+    private Map<CharSequence, Object> defaultValues;
+    @Nullable
     private final Function<Object, Object> valueMapper;
     private final RetentionPolicy retentionPolicy;
     @Nullable
     private final List<AnnotationValue<?>> stereotypes;
+    @Nullable
+    private final AnnotationDefaultValuesProvider defaultValuesProvider;
 
     /**
      * @param annotationName The annotation name
@@ -105,7 +108,6 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
         this(annotationName, values, null, retentionPolicy, stereotypes);
     }
 
-
     /**
      * @param annotationName The annotation name
      * @param values         The values
@@ -115,6 +117,18 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
     @Internal
     public AnnotationValue(String annotationName, Map<CharSequence, Object> values, Map<CharSequence, Object> defaultValues) {
         this(annotationName, values, defaultValues, RetentionPolicy.RUNTIME, null);
+    }
+
+    /**
+     * @param annotationName        The annotation name
+     * @param values                The values
+     * @param defaultValuesProvider The default values provider
+     * @since 4.3.0
+     */
+    @UsedByGeneratedCode
+    @Internal
+    public AnnotationValue(String annotationName, Map<CharSequence, Object> values, AnnotationDefaultValuesProvider defaultValuesProvider) {
+        this(annotationName, values, null, RetentionPolicy.RUNTIME, null, defaultValuesProvider);
     }
 
     /**
@@ -137,7 +151,29 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
      * @param stereotypes     The stereotypes of the annotation
      */
     @Internal
-    public AnnotationValue(String annotationName, Map<CharSequence, Object> values, Map<CharSequence, Object> defaultValues, RetentionPolicy retentionPolicy, List<AnnotationValue<?>> stereotypes) {
+    public AnnotationValue(String annotationName,
+                           Map<CharSequence, Object> values,
+                           Map<CharSequence, Object> defaultValues,
+                           RetentionPolicy retentionPolicy,
+                           List<AnnotationValue<?>> stereotypes) {
+        this(annotationName, values, defaultValues, retentionPolicy, stereotypes, null);
+    }
+
+    /**
+     * @param annotationName        The annotation name
+     * @param values                The values
+     * @param defaultValues         The default values
+     * @param retentionPolicy       The retention policy
+     * @param stereotypes           The stereotypes of the annotation
+     * @param defaultValuesProvider The defaults values provider
+     */
+    @Internal
+    public AnnotationValue(String annotationName,
+                           Map<CharSequence, Object> values,
+                           Map<CharSequence, Object> defaultValues,
+                           RetentionPolicy retentionPolicy,
+                           List<AnnotationValue<?>> stereotypes,
+                           AnnotationDefaultValuesProvider defaultValuesProvider) {
         this.annotationName = annotationName;
         this.convertibleValues = newConvertibleValues(values);
         this.values = values;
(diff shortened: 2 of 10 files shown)


core/src/main/java/io/micronaut/core/annotation/AnnotationDefaultValuesProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * @since 4.3.0
+ */
+public interface AnnotationDefaultValuesProvider {
+
+    /**
+     * @param annotationName The annotation name
+     * @return The annotation default values
+     */
+    @NonNull
+    Map<CharSequence, Object> provide(String annotationName);
```
yawkat, line 33: Why CharSequence instead of String?

CS doesn't have a well-defined eq/hc

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -233,6 +272,9 @@ public List<AnnotationValue<?>> getStereotypes() {
      */
     @Nullable
     public Map<CharSequence, Object> getDefaultValues() {
+        if (defaultValues == null && defaultValuesProvider != null) {
```
yawkat, line 275: needs to be made thread safe

core/src/main/java/io/micronaut/core/annotation/AnnotationDefaultValuesProvider.java
dstepanov: Not really sure, we have it everywere

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
dstepanov: It should be fine to resolve it multiple times

yawkat: yes but defaultValues needs to use safe publication at least. see e.g. kotlin SafePublicationLazyImpl

core/src/main/java/io/micronaut/core/annotation/AnnotationDefaultValuesProvider.java
alvarosanchez: `CharSequence` is parent of both Java's `String` and Groovy's `GString`.

yawkat: that means we need it for the processor i guess, but maybe we can move to String on the runtime side

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
dstepanov: What is the unsafe thing that can happen?

yawkat: You can see a partially-initialized HashMap in other threads that access defaultValues.

I think an easy resolution without volatile would be to wrap the map in a record. That takes advantage of the safe publication of final fields. As long as the map is read through the final field of the record, there is a guarantee that it is initialized by the time it is read.

This scenario is described in the "Publication" section of JCIP (16.2 in the edition of the book that I have)

dstepanov: You would never see that because the concurrent map backs the provider. Unfortunately, I cannot use the class directly.

yawkat: i dont think that's correct. while the concurrentmap inside the AnnotationDefaultValuesProvider does ensure that the writer thread sees the correct values (HB edge), there is no HB edge to the thread that reads the `defaultValues` field, so the reader thread could see the hashmap in an incomplete state.

yawkat: this would be my solution:

```patch
Index: core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java b/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java	(revision 94ee5488e98ea9db33c32bd79531b9b44c663efa)
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java	(date 1694423879590)
@@ -68,7 +68,7 @@
     private final ConvertibleValues<Object> convertibleValues;
     private final Map<CharSequence, Object> values;
     @Nullable
-    private Map<CharSequence, Object> defaultValues;
+    private MapHolder defaultValues;
     @Nullable
     private final Function<Object, Object> valueMapper;
     private final RetentionPolicy retentionPolicy;
@@ -177,7 +177,7 @@
         this.annotationName = annotationName;
         this.convertibleValues = newConvertibleValues(values);
         this.values = values;
-        this.defaultValues = defaultValues;
+        this.defaultValues = new MapHolder(defaultValues);
         this.valueMapper = null;
         this.retentionPolicy = retentionPolicy != null ? retentionPolicy : RetentionPolicy.RUNTIME;
         this.stereotypes = stereotypes;
@@ -222,7 +222,7 @@
                            ConvertibleValues<Object> convertibleValues,
                            Function<Object, Object> valueMapper) {
         this.annotationName = target.annotationName;
-        this.defaultValues = defaultValues;
+        this.defaultValues = new MapHolder(defaultValues);
         this.values = target.values;
         this.convertibleValues = convertibleValues;
         this.valueMapper = valueMapper;
@@ -272,10 +272,11 @@
      */
     @Nullable
     public Map<CharSequence, Object> getDefaultValues() {
-        if (defaultValues == null && defaultValuesProvider != null) {
-            defaultValues = defaultValuesProvider.provide(annotationName);
+        MapHolder mh = defaultValues;
+        if (mh == null && defaultValuesProvider != null) {
+            this.defaultValues = mh = new MapHolder(defaultValuesProvider.provide(annotationName));
         }
-        return defaultValues;
+        return mh == null ? null : mh.map();
     }
 
     /**
@@ -1702,4 +1703,7 @@
             }
         }
     }
+
+    private record MapHolder(Map<CharSequence, Object> map) {
+    }
 }

```

dstepanov: I don't understand how it can be in an incomplete state. This operation internally should only assign the pointer to the already initialized immutable map.

yawkat: it would be fine if defaultValuesProvider always returns an immutable map. but looking at the code it can return e.g. a LinkedHashMap.

dstepanov: Where do you see `LinkedHashMap`?

yawkat: getCachedAnnotationDefaults does ANNOTATION_DEFAULTS.put(annotationName, new LinkedHashMap<>(defaultValues));

dstepanov: That's done only at the annotation processing phase and it's never concurrent

yawkat: ok, and i assume registerAnnotationDefaults is only called with Map.of? then it's fine. but i think it's really brittle.

dstepanov: Yeah, that's generated by `AnnotationMetadataWriter`.
Honestly, I don't think that's a problem in the first place. All of the maps should be initialized before entering and exiting the concurrent map. IMHO the problem you describe can only occur with the concurrent field access.

yawkat: that does not matter. even if the map is initialized before it is inserted into the concurrent map, there is no HB edge to the reader thread, so it may still see the map as uninitialized.

dstepanov: I have no idea how it can see it uninitialized. Based on the timeline, it should be already initialized and in the main memory when returned from the concurrent map, but I'm not an expert on the JVM internals.

The values should be immutable so it should be fine right?

dstepanov:
@graemerocher Pls also check

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
yawkat: You can't really reason about the JMM by what might be happening in hardware. The JMM works by restricting the executions with HB order and similar constraints: https://shipilev.net/blog/2014/jmm-pragmatics/#_java_memory_model

In this sample, we have three actors. The **C**reator thread, which puts the Map into the ConcurrentMap. The **W**riter thread, which reads from the ConcurrentMap and writes the Map to the field. And the **R**eader thread that reads from the field.

Now, the first thing to consider is the happens-before order. C's write to the ConcurrentMap happens-before W's read from the ConcurrentMap (by CM spec). This means that W will see the map correctly, with all items as expected.

However, there is no second HB edge between W and R. When W writes the field and R reads the reference to the map, all bets are off. We can be sure that R will see a Map (if it was null it would make its own value, and JMM still guarantees some things here), but there is no guarantee of what the Map will contain, whether it is halfway initialized, or anything like that.

There is an additional rule wrt final fields that saves us here though. If we access data through a final field (and the object with the final field has been properly initialized, i.e. the constructor doesn't leak), then the JMM states that this data will be properly initialized. This is explained here: https://shipilev.net/blog/2014/jmm-pragmatics/#_what_do_we_have_5

Formally, Map.of introduces a 'Freeze' action because it has final fields. When we can see the map returned by Map.of, and dereference it (i.e. call .get on it), we introduce an ordering guarantee that this data is available. This only works because Map.of is truly immutable, as in it uses final fields. I wrote about this a few years back in the Concurrency section: https://javachannel.org/posts/immutability-in-java/

---

Now to talk about how this could go wrong in practice. One scenario I can think of uses caches only. R might observe the halfway-initialized map e.g. because a GC thread runs on the same core at some point. Then, when R loads the field, it sees the reference to the Map. But when it reads from that Map, there's no barrier that forces R to reload the data, so it might not see the fully initialized version.

But again, trying to determine from the hardware what could or couldn't happen doesn't work very well. There are simply too many factors–caching, compiler optimizations, CPU pipelining. Instead, the proper way to think about this is to consider what guarantees the JMM gives to the code, and whether those guarantees are enough to ensure everything works as expected.

graemerocher (approved):

PR #9858, opened 2023-09-12
test: GraalVM logging

sdelamo:
I have extracted the logging GraalVM test-suite to its own PR


sdelamo:
@msupic thanks for this test. I added the same test in starter so that we capture the issues caused by modifications to micronaut-logging https://github.com/micronaut-projects/micronaut-starter/pull/2044 before releasing

msupic (approved):

PR #9859, opened 2023-09-12
Fix logging with logger.config and no levels

### LINKED ISSUE #8918: Ensure Logging System is always refreshed

timyates:
When we configure logger.config as per the docs:

https://docs.micronaut.io/latest/guide/#loggingConfiguration

It is not used unless we also configure logger.levels.xxx

This PR fixes that by relaxing the property required for PropertiesLoggingLevelsConfigurer so the logging system gets initialized if this config is defined

Closes #8918

### DIFF
--- a/context/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java
+++ b/context/src/main/java/io/micronaut/logging/PropertiesLoggingLevelsConfigurer.java
@@ -45,11 +45,12 @@ import java.util.Map;
 @Context
 @Requires(beans = LoggingSystem.class)
 @Requires(beans = Environment.class)
-@Requires(property = PropertiesLoggingLevelsConfigurer.LOGGER_LEVELS_PROPERTY_PREFIX)
+@Requires(property = PropertiesLoggingLevelsConfigurer.LOGGER_PROPERTY_PREFIX)
 @Internal
 final class PropertiesLoggingLevelsConfigurer implements ApplicationEventListener<RefreshEvent> {
 
-    static final String LOGGER_LEVELS_PROPERTY_PREFIX = "logger.levels";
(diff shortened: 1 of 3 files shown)


msupic (approved):

sdelamo (commented):
the issue here https://github.com/micronaut-projects/micronaut-core/pull/9859/files was that the requires was setting `logger.levels` and you changed it to `logger`. And if you set `@Requires(property = "logger")`If you define `logger.config` that fullfils the `@Requires`?

sdelamo (approved):

timyates:
Yeah, so previously setting `logger.config` didn't enable this bean, so the loggers were not initialized, so the config was ignored, but setting `logger.levels` enabled this bean so it was used.

This change means the bean is enabled on any config starting with `logger`, so `logger.config` triggers its creation

sdelamo:
@timyates there are CI failures: 

https://ge.micronaut.io/s/crjdaekz67jm4/tests/task/:management:test/details/io.micronaut.management.endpoint.beans.BeansEndpointSpec/test%20beans%20endpoint?top-execution=1

sdelamo (changes requested):
@timyates there are CI failures: https://ge.micronaut.io/s/crjdaekz67jm4/tests/task/:management:test/details/io.micronaut.management.endpoint.beans.BeansEndpointSpec/test%20beans%20endpoint?top-execution=1

timyates:
On it

sdelamo (approved):

PR #9872, opened 2023-09-14
Implement JsonAnyGetter for BeanIntrospectionModule

### LINKED ISSUE #9870: Classes annotated with @Serdeable are not honoring @JsonAnyGetter when using jackson databind
### Expected Behavior

_No response_

### Actual Behaviour

_No response_

### Steps To Reproduce

crossposting here, because i think it is not graphql related, see the example test at https://github.com/micronaut-projects/micronaut-graphql/issues/440#issuecomment-1718315776 

### Environment Information

OSX

### Example Application

_No response_

### Version

4.1.1

yawkat:
Normally this would be a feature I skip out on, because we can't support every jackson feature in BeanIntrospectionModule. However in this case, the BIM actually duplicated the AnyGetter as a normal property also, so a fix was necessary anyway, and implementing the full logic is not that difficult and a test case existed already.

A similar problem exists for JsonValue, but because there's no existing test case for it, I don't want to implement it fully. That means JsonValue will still not work with native image, likely. But JsonAnyGetter should.

Fixes #9870

### DIFF
--- a/jackson-databind/src/test/groovy/io/micronaut/jackson/modules/BeanIntrospectionModuleSpec.groovy
+++ b/jackson-databind/src/test/groovy/io/micronaut/jackson/modules/BeanIntrospectionModuleSpec.groovy
@@ -354,6 +354,7 @@ class BeanIntrospectionModuleSpec extends Specification {
     void "test that introspected serialization works with @JsonAnyGetter"() {
         given:
         ApplicationContext ctx = ApplicationContext.run()
+        ctx.getBean(BeanIntrospectionModule).ignoreReflectiveProperties = ignoreReflectiveProperties
         JsonMapper objectMapper = ctx.getBean(JsonMapper)
 
         when:
@@ -373,6 +374,9 @@ class BeanIntrospectionModuleSpec extends Specification {
 
         cleanup:
         ctx.close()
+
+        where:
+        ignoreReflectiveProperties << [true, false]
     }
 
     void "test that introspected serialization of errors works"() {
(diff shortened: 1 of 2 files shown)


yawkat:
@sdelamo i added this to the 4.1.2 project because that's the latest project, but it's not that urgent if you don't want to make another core release

jackson-databind/src/test/groovy/io/micronaut/jackson/modules/BeanIntrospectionModuleSpec.groovy
```diff
@@ -354,6 +354,7 @@ class BeanIntrospectionModuleSpec extends Specification {
     void "test that introspected serialization works with @JsonAnyGetter"() {
```
timyates, line 354: I'm confused.  This test used to pass...  Wasn't it testing that `@JsonAnyGetter` worked as expected?

yawkat: There are two branches in the BIM. One branch "enhances" the properties discovered by jackson using reflection by replacing the property getters/setters with non-reflective BeanIntrospection access. The second branch runs when jackson discovered no properties using reflection, which is typically the case with native image (but can also happen when there are simply no normal properties). It replaces the properties from jackson entirely.

The first branch worked fine and was covered by this test already, the second branch is fixed by this PR.

timyates (approved):

sdelamo:
> It means JsonValue will still not work with native image 

Is that documented?

sdelamo (approved):

PR #9893, opened 2023-09-19
fix health monitor task

codewithbear:
I discovered that HealthMonitoringTask is currently broken. While the periodic task is scheduled according to config, the logic fails to update the CurrentHealthStatus bean.

See example in the added test HealthMonitorTaskSpec.groovy

I found out that the subscriber to reactiveSequence is not getting triggered. It looks like in the  `onSubscribe` we have to explicitly call `request` method so that HealthIndicators would emit HealthResults

If we fix the above problem with use of `request` we run into another problem. `onComplete` in the subscriber reports that the status is UP, even though some checks might be DOWN, and the web endpoints report DOWN. So it looks like incorrect behaviour.

This PR proposes a rewrite in a more clear way and fixing the above problems:
- first, collect all publishers in a List
- use Flux.collectList to subscribe to all publishers and collect all results in one list
- Then we can operate on the final result list in a simple way to check if any status is down and report that. Only if all statuses are up then we report UP

### DIFF
--- a/management/src/main/java/io/micronaut/management/health/monitor/HealthMonitorTask.java
+++ b/management/src/main/java/io/micronaut/management/health/monitor/HealthMonitorTask.java
@@ -27,8 +27,6 @@ import io.micronaut.scheduling.annotation.Scheduled;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import org.reactivestreams.Publisher;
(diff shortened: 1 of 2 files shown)


wetted (approved):
This apparently fixes issues I am having getting a related test to pass on #9895

PR #9898, opened 2023-09-20
Add "functional" benchmark

melix:
This commit introduces a new test project which is a JMH benchmark testing an application. It makes use of the Micronaut Gradle plugins (that is to say builds an application in a similar way as for a user application) and applies the AOT plugin.

There are 2 JMH jars being created. The standard JMH jar is using the regular application classpath. Another jar is created with the optimized jar.

To run the benchmark with the AOT optimized application, run:

`./gradlew :test-suite-benchmarks:optimizedJmhJar`

Then:

`./test-suite-benchmarks/build/libs/test-suite-benchmarks-jmh-all-<version>.jar`

### DIFF
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'io.micronaut.build.internal.convention-base'
-    id "me.champeau.jmh" version "0.7.1"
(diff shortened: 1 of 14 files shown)


melix:
Closed in favor of #9908 9908

Issue #9874, opened 2023-09-14
@Scheduled registered more than once when multiple annotations present in the same class


garbolim:
### Expected Behavior

@Scheduled method should only be registered once

### Actual Behaviour

@Scheduled method is registered to the TaskScheduler more than once, resulting in the method being run twice at the same time by different thread.

### Steps To Reproduce

Run the example application provided.
We can see that the "run()" method is being called twice every 5s, by different scheduler thread. 

### Environment Information

Java 17

### Example Application

https://github.com/garbolim/demo-micronaut/tree/master

### Version

4.1.1

PR #9901, opened 2023-09-21
Fix for double scheduling intercepted methods

codewithbear:
I managed to reproduce locally the issue described here
https://github.com/micronaut-projects/micronaut-core/issues/9874

Added a spec as well in this PR that reproduces the issue.

It looks like combining @Scheduled annotation with an aop annotation would result in both the original bean and the proxy bean registered for scheduled task and execute simultaneously.

Proposing a potential fix in the ScheduledMethodProcessor, closest to the problem. If the bean is not in context, then do not schedule it.

Another possible approach could be not do the annotation processing for beans that are being proxied, given that this logic would happen on the proxy bean as well. However I'm not sure what are the implications of this.


context/src/main/java/io/micronaut/scheduling/processor/ScheduledMethodProcessor.java
```diff
@@ -94,6 +94,13 @@ public void process(BeanDefinition<?> beanDefinition, ExecutableMethod<?, ?> met
             return;
         }
 
+        boolean containsInContext = beanContext.getBeansOfType(beanDefinition.getBeanType())
+            .stream()
+            .anyMatch(b -> b.getClass().equals(beanDefinition.getBeanType()));
+        if (!containsInContext) {
```
graemerocher: not sure this is the correct fix. You can use `beanContext.containsBean(..)`.. you will need to include the qualifier

codewithbear: Thanks for the feedback! I changed this code to use a qualifier that checks for an exact type match. Please take another look

context/src/main/java/io/micronaut/scheduling/processor/ScheduledMethodProcessor.java
```diff
@@ -93,6 +94,10 @@ public void process(BeanDefinition<?> beanDefinition, ExecutableMethod<?, ?> met
         if (!(beanContext instanceof ApplicationContext)) {
             return;
         }
+        Class<Object> beanType = (Class<Object>) beanDefinition.getBeanType();
+        if (!beanContext.containsBean(beanType, ExactTypeQualifier.INSTANCE)) {
```
graemerocher: this should be `beanDefinition.getDeclaredQualifier()` and probably `beanDefinition.asArgument()` for the type

codewithbear: I tried this approach and it does not work unfortunately. For the beanDefinition of the original bean `beanContext.getBean` returns the intercepted bean and `beanContext.containsBean()` is true. So both beans end up being scheduled.

What issue do you see with the current version?

context/src/main/java/io/micronaut/scheduling/processor/ScheduledMethodProcessor.java
sdelamo: Thanks @codewithbear when you want another review click the 

CleanShot 2023-09-22 at 09 31 27@2x


that will trigger another notification to the reviewer.

codewithbear: Is there a way to retrigger tests? I failed for some reason and passes locally, so I'm wondering if it was a flake

graemerocher:
I am unconvinced the fix is correct, I have pushed another fix that disables the processing of proxied beans which are not enabled. Let's see if this is better.

graemerocher:
Tests pass except ExactTypeQualifier can be deleted https://github.com/micronaut-projects/micronaut-core/pull/9901/files#diff-630804e28921320a1d3af0c2e768107a7b2428ded20264856bcc9b59bbc13534R15 can be deleted

graemerocher (approved):

dstepanov (approved):

PR #9906, opened 2023-09-21
Issue 9905 websphere deploy war

### LINKED ISSUE #9905: Micronaut war fails to start under IBM WebSphere 9
### Expected Behavior

A Micronaut war file should deploy and run successfully to IBM WebSphere 9.

NOTE: This issue is for record. I have a patch incoming shortly.

### Actual Behaviour

The application fails to start because it cannot find the `ApplicationEventPublisher` bean. This bean cannot be created because the startup process fails to find and instantiate the `ApplicationEventPublisherFactory`.

This occurs due to unexpected quirks in the IBM WebSphere-provided class loaders. `IOUtils` and `ServiceScanner` need to be adjusted to this special case.

### Steps To Reproduce

Consult the README in the linked example application.

Build a war, load to a WebSphere server, and attempt to run.

### Environment Information

- Docker image: `ibmcom/websphere-traditional:9.0.5.16-ubi8-amd64`
- Java 8
- Micronaut 3.8.10

### Example Application

https://github.com/mattmoss/micronaut3-websphere9-sample

### Version

3.8.10

mattmoss:
Fixes #9905


mattmoss:
I have checked on my local machine/docker configuration that the provided patch fixes a Micronaut app as well as a Grails 5.x application.

If you want to try this yourself, let me know and I can provide instructions on running WebSphere in Docker.

sdelamo:
> I have checked on my local machine/docker configuration that the provided patch fixes a Micronaut app as well as a Grails 5.x application.
> 
> If you want to try this yourself, let me know and I can provide instructions on running WebSphere in Docker.

provide instructions please

core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
```diff
@@ -119,12 +119,28 @@ private Set<String> computeStandardServiceTypeNames(URL url) {
...
+    private boolean isWebSphereClassLoader() {
+        return classLoader.getClass().getName().startsWith("com.ibm.ws.classloader");
+    }
+
+    private String buildResourceSearchPath() {
+        String path = "META-INF/micronaut/" + serviceName;
+
+        if (isWebSphereClassLoader()) {
+            // Special case WebSphere classloader
+            // https://github.com/micronaut-projects/micronaut-core/issues/....
```
sdelamo: can you add a correct link?

sdelamo (changes requested):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1751,6 +1751,7 @@ protected List<BeanDefinitionReference> resolveBeanDefinitionReferences() {
             final SoftServiceLoader<BeanDefinitionReference> definitions = SoftServiceLoader.load(BeanDefinitionReference.class, classLoader);
             beanDefinitionReferences = new ArrayList<>(300);
             definitions.collectAll(beanDefinitionReferences, BeanDefinitionReference::isPresent);
+            beanDefinitionReferences.removeIf(Objects::isNull);
```
yawkat: seems like this should happen in SoftServiceLoader instead

mattmoss:
> provide instructions please

[WebSphere-Process.pdf](https://github.com/micronaut-projects/micronaut-core/files/12701691/WebSphere-Process.pdf)

Attached instructions were sent to me and worked for me, with a couple of adjustments annotated.

Some of the docker commands I have here (easier to copy/paste from here, perhaps, than from the pdf)
```
# Create the container
docker run --name test -h test -v $(pwd)/PASSWORD:/tmp/PASSWORD -p 9043:9043 -p 9443:9443 -p 9080:9080 -d ibmcom/websphere-traditional:latest

# To login to the WebSphere console, use username 'wsadmin'
# Get the password with this (changes daily?)
docker exec -it test cat /tmp/passwordupdated

# Copy war file it to container
docker cp ./build/libs/sample-0.1.war test:/opt/IBM/WebSphere/AppServer/temp

# After copy, look in PDF for how to install and start the war file.
```

core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
mattmoss: Done: [79fbcc8](https://github.com/micronaut-projects/micronaut-core/pull/9906/commits/79fbcc8360288ee3d7487311ba5e09a332ada159)

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
mattmoss: I'm not exactly sure where the `null` came from... not directly from the `SoftServiceLoader`, but from one of the `ServiceCollector`/`RecursiveActionValuesCollector`s. But at the moment, I'm not able to reproduce the issue that lead me to make this change... certainly not in local dev.  I'll need to check again on WebSphere and see if it's only there.

mattmoss: I was unable to reproduce the behavior that prompted this change. And I was still able to deploy and run both a Micronaut and Grails application using the other changes from this patch, so I have reverted this change.

mattmoss:
As a point of info, 3.8.x is needed to support Grails 5.3.x, but these changes should be merged forward as well.

yawkat (approved):

sdelamo (approved):

PR #9922, opened 2023-09-26
avoid unnecessary calls to ConcurrentLinkedDeque.size() 

s15r:
because it's a non-constant-time operation


yawkat (approved):

sdelamo:
@yawkat can we add a test?

yawkat:
no, there's no behavioral difference

PR #9923, opened 2023-09-26
fix builders with private or protected constructors

graemerocher:
Fixes https://github.com/micronaut-projects/micronaut-serialization/pull/573


yawkat (approved):
This is a bit hacky :(

PR #9925, opened 2023-09-27
Micronaut 4 Groovy projects fail Intellij compilation

### LINKED ISSUE #9915: Micronaut 4 Groovy projects fail Intellij compilation
### Expected Behavior

All information on https://youtrack.jetbrains.com/issue/IDEA-331174 as I initially believed this to be an Intellij-specific issue.  

Example project included in this ticket:
[groovy-compile-bug.zip](https://github.com/micronaut-projects/micronaut-core/files/12708921/groovy-compile-bug.zip)

### Actual Behaviour

Compilation fails when building Micronaut 4 groovy projects using Intellij's compiler, that have annotations that are defined in Groovy files.  

### Steps To Reproduce

All information on https://youtrack.jetbrains.com/issue/IDEA-331174 as I initially believed this to be an Intellij-specific issue.  

They state that the problem is caused by the library `io.micronaut:micronaut-inject-groovy` starting from version 4.0.0, so raising it here for further investigation. 

### Environment Information

Operating System: MacOS Ventura, Version 13.4
JDK Version: Coretto 17

### Example Application

see attached 

### Version

4.0.3

timyates:
Something in this PR has broken compilation with Groovyc and inject-groovy.

https://github.com/micronaut-projects/micronaut-core/pull/9232

when d273fdae71a75974a97528a552f665bd7f5e1f3c is reverted then this test passes.

I am not sure why

### DIFF
--- a/inject-groovy/src/functionalTest/groovy/io/micronaut/ast/groovy/GroovyCompilationSpec.groovy
+++ b/inject-groovy/src/functionalTest/groovy/io/micronaut/ast/groovy/GroovyCompilationSpec.groovy
@@ -1,0 +1,34 @@
+package io.micronaut.ast.groovy
+
+import io.micronaut.fixtures.context.MicronautApplicationTest
+import org.codehaus.groovy.tools.FileSystemCompiler
+import spock.lang.Issue
+
+class GroovyCompilationSpec extends MicronautApplicationTest {
+
+    @Issue("https://github.com/micronaut-projects/micronaut-core/issues/9915")
+    def "compilation with the Groovy compiler works as expected"() {
+        given:
+        groovySourceFile("Annotation.groovy", """
+import java.lang.annotation.ElementType
+import java.lang.annotation.Target
+
+@Target(ElementType.FIELD)
+@interface Annotation {
+}
+""")
+
+        groovySourceFile("Usage.groovy", """
+// Usage.groovy
+class Usage {
+    @Annotation
+    String field
+}
+""")
+        when:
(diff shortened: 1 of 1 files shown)


timyates:
From @paulk-asert 

> It probably isn't the usual case of two places using ClassHelper.make when a cached version is needed but rather one being a primary classnode (since we are compiling it)

dstepanov:
Your test is failing becase the class is not found.
Please try your use case with `resolveName` changed to `findClassNode`. Otherwise this needs to be addressed with the Groovy team. All this is added so we can properly get the default annotation values (GroovyMappedStereotypesSpec).

dstepanov:
So I looked a bit more, and I see a few issues:
- @timyates The test should have `FileSystemCompiler.commandLineCompile(testDirectory.resolve("src/main/java/Usage.groovy").toString(), testDirectory.resolve("src/main/java/Annotation.groovy").toString())`
- It looks like our code have added that resolution, because it's not possible to retrive the annotation default values from the ordinary `ClassNode` but it works with `DecompiledClassNode`
- The problem is in `ClassNodeResolver#tryAsScript` which calls `CompilationUnit#addSource`:
```
public SourceUnit addSource(final SourceUnit source) {
        String name = source.getName();
        source.setClassLoader(getClassLoader());
        for (SourceUnit su : queuedSources) {
            if (name.equals(su.getName())) return su;
        }
        queuedSources.add(source);
        return source;
    }
```
That one is not checking if the source is already in `sources` field.

PR #9934, opened 2023-09-29
fix: alias propagation / configuration builder  (#9919)

### LINKED ISSUE #9919: ConfigurationBuilder doesn't work on builder with getters and setters and empty prefixes
### Expected Behavior

The builder has getters and setters with the same name (empty prefix) and the getter is declared before the setter.

The setter should be called because it has one argument and returns the builder. 

When debugging, I have noticed that the `allowWriteWithZeroArgs` of the `BeanProperties` seems to be set to `true` despite the fact the `ConfigurationBuilder` sets it to false (default value of `allowZeroArgs`).

### Actual Behaviour

The getter is called.

### Steps To Reproduce

Add the test below to `ConfigurationPropertiesBuilderSpec`.

```
    void "test configuration builder with both getters and setters and no prefixes"() {
        given:
        ApplicationContext ctx = buildContext('''
package cpbtest11

import io.micronaut.context.annotation.*
import io.micronaut.core.annotation.AccessorsStyle

@ConfigurationProperties("pool")
final class PoolConfig {

    @ConfigurationBuilder(prefixes = [""])
    public ConnectionPool.Builder builder = DefaultConnectionPool.builder()

}

interface ConnectionPool {

    interface Builder {
        Integer maxConcurrency()
        Builder maxConcurrency(Integer maxConcurrency)
        ConnectionPool build()
    }

    int getMaxConcurrency()
}

class DefaultConnectionPool implements ConnectionPool {
    private final int maxConcurrency

    DefaultConnectionPool(int maxConcurrency) {
        this.maxConcurrency = maxConcurrency
    }

    static ConnectionPool.Builder builder() {
        return new DefaultBuilder()
    }

    @Override
    int getMaxConcurrency() {
        return maxConcurrency
    }

    private static class DefaultBuilder implements ConnectionPool.Builder {

        private int maxConcurrency

        private DefaultBuilder() {
        }

        @Override
        ConnectionPool.Builder maxConcurrency(Integer maxConcurrency) {
            this.maxConcurrency = maxConcurrency
            return this
        }

        @Override
        Integer maxConcurrency() {
            return maxConcurrency
        }

        ConnectionPool build() {
            return new DefaultConnectionPool(maxConcurrency)
        }
    }
}
''')
        ctx.getEnvironment().addPropertySource(PropertySource.of(["pool.max-concurrency": 123]))

        when:
        Class<?> testProps = ctx.classLoader.loadClass("cpbtest11.PoolConfig")
        def testPropBean = ctx.getBean(testProps)

        then:
        noExceptionThrown()
        testPropBean.builder.build().getMaxConcurrency() == 123
    }
```

### Environment Information

- OpenJDK Runtime Environment (build 17.0.8.1+1-Ubuntu-0ubuntu122.04)

### Example Application

_No response_

### Version

4.1.x

rorueda:
This is an attempt to fix #9919.

**Aliases Propagation**
I had more time to dig into the problem and what I found is that the introduced aliases are added at the end of the annotation stereotypes and because of this they are the last to be processed and they do not override the annotations processed before. 

The fix consists in prepend the aliases so they are prioritized - no need to say that I'm not sure this has no unintended consequences as I'm not familiar with the framework internals, but the tests all passed.

I also added one new test and had to change some that didn't seem right to me (they seem to have been changed in the past to match what I consider to be incorrect behavior).

**Configuration Builder**
I changed the default to do not allow zero arguments as this behavior is very explicit in the documentation (although it might be a breaking change for those relying on the current behavior). 

I also added/corrected some tests.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
@@ -1017,7 +1017,7 @@ public abstract class AbstractAnnotationMetadataBuilder<T, A> {
         if (!introducedAliasForAnnotations.isEmpty()) {
             newAnn = newAnn.withAnnotationValue(
                     newAnn.getAnnotationValue().mutate()
-                            .stereotypes(
+                            .stereotypesPrepend(
                                     introducedAliasForAnnotations.stream()
                                             .flatMap(a -> processAnnotation(context, a))
                                             .<AnnotationValue<?>>map(ProcessedAnnotation::getAnnotationValue)
--- a/core-processor/src/main/java/io/micronaut/inject/ast/PropertyElementQuery.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/PropertyElementQuery.java
@@ -16,8 +16,6 @@
 package io.micronaut.inject.ast;
 
 import io.micronaut.context.annotation.BeanProperties;
-import io.micronaut.context.annotation.ConfigurationBuilder;
-import io.micronaut.context.annotation.ConfigurationProperties;
 import io.micronaut.core.annotation.AccessorsStyle;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.AnnotationValue;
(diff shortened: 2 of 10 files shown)


dstepanov:
Thanks for the PR! I will investigate if it's the correct solution. Please sign the CLA with the commit email.

dstepanov:
I have a new PR with a different solution https://github.com/micronaut-projects/micronaut-core/pull/9942

PR #9968, opened 2023-10-11
Add missing Groovy source file implementation

dstepanov:


melix (approved):
Have you checked the behavior? Won't this generate source files in the classes directory, so alongside the .class files?

graemerocher:
probably we need a test, and also just realised there are no tests for the other languages

graemerocher:
also what happens KAPT vs KSP

Issue #9969, opened 2023-10-12
Micronaut fails for LocalDateTime method parameter


rlconst:
### Expected Behavior

You can pass LocalDateTime as parameter
```
String localDateTime(@NotNull @QueryValue LocalDateTime time) {
    ...
}
```

### Actual Behaviour

```
ERROR i.m.r.intercept.RecoveryInterceptor - Type [io.micronaut.http.client.convert.DateTimeConversionSpec$TimeClient$Intercepted] executed with error: Unsupported field: OffsetSeconds
java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: OffsetSeconds
	at java.base/java.time.LocalDate.get0(LocalDate.java:709)
	at java.base/java.time.LocalDate.getLong(LocalDate.java:688)
	at java.base/java.time.LocalDateTime.getLong(LocalDateTime.java:718)
	at java.base/java.time.format.DateTimePrintContext.getValue(DateTimePrintContext.java:308)
	at java.base/java.time.format.DateTimeFormatterBuilder$OffsetIdPrinterParser.format(DateTimeFormatterBuilder.java:3628)
	at java.base/java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(DateTimeFormatterBuilder.java:2402)
	at java.base/java.time.format.DateTimeFormatter.formatTo(DateTimeFormatter.java:1849)
	at java.base/java.time.format.DateTimeFormatter.format(DateTimeFormatter.java:1823)
	at io.micronaut.runtime.converters.time.TimeConverterRegistrar.lambda$register$3(TimeConverterRegistrar.java:141)
	at io.micronaut.core.convert.DefaultConversionService.convert(DefaultConversionService.java:155)
	at io.micronaut.context.env.DefaultEnvironment.convert(DefaultEnvironment.java:311)
	at io.micronaut.core.convert.ConversionService.convert(ConversionService.java:120)
	at io.micronaut.http.client.bind.binders.QueryValueClientArgumentRequestBinder.bind(QueryValueClientArgumentRequestBinder.java:119)
```

### Version

3.10.1

rlconst:
I'm on 3.10.x and I did a simple fix. 4.x logic is significantly different. What is the best aproach: backport or keep it as is?

sdelamo:
This does not happen in Micronaut Framework 4, does it?

sdelamo:
[...]

PR #9970, opened 2023-10-12
Fix @QueryValue of LocalDateTime type for 3.10.x

rlconst:
Fix for #9969 with test case

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -24,7 +24,19 @@ import io.micronaut.core.convert.TypeConverterRegistrar;
 import io.micronaut.core.convert.format.Format;
 import io.micronaut.core.util.StringUtils;
 
-import java.time.*;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.MonthDay;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.time.Period;
+import java.time.Year;
+import java.time.YearMonth;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.TemporalAccessor;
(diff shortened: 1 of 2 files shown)


rlconst:
@sdelamo I tried to backport code from 4.x but it's really painfull because of Java 8 requirement

rlconst:
Removed irrelevan code formatting

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -162,6 +162,27 @@ public void register(ConversionService<?> conversionService) {
...
+        /*
+         * There's a not-a-bug <a href="https://bugs.openjdk.org/browse/JDK-8069324">JDK-8069324</a>
+         * which unexpectedly fail because offset is mandatory for RFC_1123
+         * To work around this we use special converter for LocalDateTime treating it as UTC timezone
+         */
+        final TypeConverter<LocalDateTime, CharSequence> localDateTimeConverter = (object, targetType, context) -> {
+            try {
+                DateTimeFormatter formatter = resolveFormatter(context);
+                // Treat missing zone as UTC
+                return Optional.of(formatter.format(object.atOffset(ZoneOffset.UTC)));
```
dstepanov: Why not `ZoneOffset.systemDefault`?

yawkat: im against any implicit time zone at all, be it UTC or systemDefault. Why does this converter need to support RFC 1123? RFC 1123 makes no sense for LDT, we should not support it just like the JDK doesnt support it.

sdelamo:
I added a test to the TCK. However, I don't reproduce the issue https://github.com/micronaut-projects/micronaut-core/pull/10027

context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
rlconst: @dstepanov because systemDefault offset on client and server could be different

rlconst: @yawkat I tried to minimize changes without changing the current semantic, especialy it was properly handled in 4.x

rlconst:
@sdelamo what the difference between TCK and DateTimeConversionSpec? Because if I rollback changes in converter, DateTimeConversionSpec reproducibly fails

yawkat:
ok, ive looked at the code some more. the reason why it works on 4 but not 3 is that we changed away from using RFC 1123 datetime by default, to ISO 8601 timestamps. 

I am against fixing this by adding a timezone out of thin air. However I think an alternative solution is possible and I will update this PR.

yawkat (approved):

sdelamo:
do the changes done by @yawkat in this PR satisfy your needs @rlconst ?

rlconst:
@sdelamo sure. @yawkat thank you

sdelamo (approved):

PR #9978, opened 2023-10-13
Add Java 21

timyates:
- Exclude Kapt test-suite-kotlin under Java 21
- Switch to use Ksp for context-propagation tests
- Accept differences to generated classes caused by 21 (which we don't publish)

PR #9985, opened 2023-10-16
Core Test java 21

dstepanov:


dstepanov:
Looks like 21 build took all the tasks from cache...

timyates:
It took them from the cached java 17 ones

https://ge.micronaut.io/s/xnwym4gcburfo/timeline?details=uvi2dadwsnupq&page=2

That doesn't seem right...

@melix Java version change should invalidate the cache shouldn't it?

timyates:
@dstepanov I think you also need to set the jvm version to 17 (or something less than 21) for kotlin compilation

dstepanov:
How?

timyates:
I cannot currently see why this works without having to do something like https://github.com/micronaut-projects/micronaut-crac/blob/master/test-suite-kotlin-ksp/build.gradle.kts#L18-L25

timyates:
Ahhh, we already set jvmTarget to 17 it seems in multiple places...

timyates:
All tasks in this PR and https://github.com/micronaut-projects/micronaut-core/pull/9978 are being run through Java 17 via the Gradle toolchain support 😞

timyates:
@dstepanov [pushed a change to actually use the JDK version](https://github.com/micronaut-projects/micronaut-core/pull/9985/commits/bd55a2b30c613f6ff09ff07d6e8c4417cd60c5d6) that gradle is being executed by

I think `:context-propagation:kaptKotlin` is failing

core-reactive/build.gradle
```diff
@@ -14,6 +14,11 @@ dependencies {
 tasks.named("compileKotlin") {
     kotlinOptions {
         jvmTarget = 17
-        languageVersion = "1.7"
+    }
+}
+
+tasks.named("compileTestKotlin") {
```
melix, line 20: This should be done in a common plugin rather than copy/pasting into every build script

melix (changes requested):

melix:
> All tasks in this PR and https://github.com/micronaut-projects/micronaut-core/pull/9978 are being run through Java 17 via the Gradle toolchain support 😞

We should really not mix the VM used to run the build with the one used to build/run the tests. That's because ideally we shouldn't have to rebuild everything _just_ to test that it works on 21: there should be 2 tasks which test different JDK versions. Since we're not there and that it would require many changes to our scripts, then it means that you have to update the test tasks to use the toolchain which matches the VM that the build is running on.

timyates:
> it means that you have to update the test tasks to use the toolchain which matches the VM that the build is running on.

Do you have a link showing how this is done?

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
```diff
@@ -8,12 +8,17 @@ plugins {
 micronautBuild {
     enableBom = false
     enableProcessing = false
+    javaVersion = JavaVersion.current().majorVersion.toInteger()
```
melix, line 11: To be clear, this is a risky change. As potentially, this means that we could have a release workflow which accidentally uses Java 21 and release a version with the wrong version.

The build should compile on 17, not 21. What you want is running _tests_ on 21, I think.

timyates:
@dstepanov can we close this as superseded by https://github.com/micronaut-projects/micronaut-core/pull/9989 ?

Thanks for all the help 👍

PR #9999, opened 2023-10-20
Add eq/hc for HttpVersionSelection

### LINKED ISSUE #9994: Missing equals/hashCode in HttpVersionSelection class generates a memory leak
### Expected Behavior

no memory leak

### Actual Behaviour

`io.micronaut.http.client.netty.DefaultHttpClient` instances are leaking and are retained in `unbalancedClients` map in `io.micronaut.http.client.netty.DefaultNettyHttpClientRegistry`

Screenshot 2023-10-17 at 15 13 36

### Steps To Reproduce

In our application we have defined a http client like this:

```
@Client(
    id = "rc-api",
    path = "/api/unstable/remote_config/products/product1",
    alpnModes = [HttpVersionSelection.ALPN_HTTP_1],
)
interface RcApiClient {
}
```
We have noticed after a certain time, that the Java Heap was filling up to the maximum.
By analysis of the heap dump we noticed a large amount of `DefaultHttpClient` instances stored into the unbalancedClients map in `DefaultNettyHttpClientRegistry`.

Insertion into the map is done by [this](https://github.com/micronaut-projects/micronaut-core/blob/426e891f3258ddadaa513c78ec10df2c9bfa1a2c/http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java#L328):
```
private DefaultHttpClient getClient(ClientKey key, BeanContext beanContext, AnnotationMetadata annotationMetadata) {
        return unbalancedClients.computeIfAbsent(key, clientKey -> {
```
and the key of this `unbalancedClients` map is `io.micronaut.http.client.netty.DefaultNettyHttpClientRegistry.ClientKey`

[ClientKey](https://github.com/micronaut-projects/micronaut-core/blob/426e891f3258ddadaa513c78ec10df2c9bfa1a2c/http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java#L561) class is made from the following fields:
```
        final HttpVersionSelection httpVersion;
        final String clientId;
        final List<String> filterAnnotations;
        final String path;
        final Class<?> configurationClass;
        final JsonFeatures jsonFeatures;
```

and [`equals` and `hashCode`](https://github.com/micronaut-projects/micronaut-core/blob/426e891f3258ddadaa513c78ec10df2c9bfa1a2c/http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java#L585):
```
       @Override
        public boolean equals(Object o) {
            if (this == o) {
                return true;
            }
            if (o == null || getClass() != o.getClass()) {
                return false;
            }
            ClientKey clientKey = (ClientKey) o;
            return httpVersion == clientKey.httpVersion &&
                    Objects.equals(clientId, clientKey.clientId) &&
                    Objects.equals(filterAnnotations, clientKey.filterAnnotations) &&
                    Objects.equals(path, clientKey.path) &&
                    Objects.equals(configurationClass, clientKey.configurationClass) &&
                    Objects.equals(jsonFeatures, clientKey.jsonFeatures);
        }

        @Override
        public int hashCode() {
            return Objects.hash(httpVersion, clientId, filterAnnotations, path, configurationClass, jsonFeatures);
        }
```

The problem is that `httpVersion` field when not null is using the type [`io.micronaut.http.client.HttpVersionSelection`](https://github.com/micronaut-projects/micronaut-core/blob/426e891f3258ddadaa513c78ec10df2c9bfa1a2c/http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java#L34) that does not provide any `equals` and `hashCode` methods.
So in the case of the ClientKey::hashCode it means that we are using an identityHashCode and using @Client annotation with `alpn` will create a new instance of `HttpVersionSelection` (through `DefaultNettyHttpClientRegistry#getClientKey` -> `HttpVersionSelection#forClientAnnotation`.
Then every http client have a different hashCode that will spread across the `unbalancedClients` map.
Moreover the `ClientKey::equals` use reference equality (`==`) for the `httpVersion` field which will be always false.

It ends up creating a new http client for every request and stored into the map.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.1.5

yawkat:
Fixes #9994

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpVersionSelection.java
@@ -23,6 +23,7 @@ import io.micronaut.http.HttpVersion;
 import io.micronaut.http.client.annotation.Client;
 
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * This class collects information about HTTP client protocol version settings, such as the
@@ -194,6 +195,25 @@ public final class HttpVersionSelection {
     @Internal
     public boolean isHttp3() {
         return http3;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        HttpVersionSelection that = (HttpVersionSelection) o;
+        return alpn == that.alpn && http2CipherSuites == that.http2CipherSuites && http3 == that.http3 && plaintextMode == that.plaintextMode && Arrays.equals(alpnSupportedProtocols, that.alpnSupportedProtocols);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(plaintextMode, alpn, http2CipherSuites, http3);
+        result = 31 * result + Arrays.hashCode(alpnSupportedProtocols);
+        return result;
     }
 
     /**
--- a/http-client-core/src/test/groovy/io/micronaut/http/client/HttpVersionSelectionSpec.groovy
+++ b/http-client-core/src/test/groovy/io/micronaut/http/client/HttpVersionSelectionSpec.groovy
@@ -1,0 +1,24 @@
+package io.micronaut.http.client
+
+
+import io.micronaut.core.annotation.Introspected
+import io.micronaut.core.beans.BeanIntrospector
+import io.micronaut.http.client.annotation.Client
+import spock.lang.Specification
+
+class HttpVersionSelectionSpec extends Specification {
+    def 'annotation equals'() {
+        given:
+        def introspection = BeanIntrospector.SHARED.findIntrospection(TestClass).get()
+        def s1 = HttpVersionSelection.forClientAnnotation(introspection)
+        def s2 = HttpVersionSelection.forClientAnnotation(introspection)
+
+        expect:
+        s1 == s2
+    }
+
+    @Client(alpnModes = ["h2"], plaintextMode = HttpVersionSelection.PlaintextMode.HTTP_1)
+    @Introspected
+    static interface TestClass {
+    }
+}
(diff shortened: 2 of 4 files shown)


graemerocher (changes requested):
please add a test

yawkat:
hmm not sure how to test the original bug since it's a perf issue. but i guess i can unit the class.

graemerocher (approved):

jpbempel:
I think there is still an issue in [`io.micronaut.http.client.netty.DefaultNettyHttpClientRegistry.ClientKey::equals`](https://github.com/micronaut-projects/micronaut-core/blob/d14081567aff5f35c9a767b64c89eb87c34770fa/http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java#L593) code related to `httpVersion`

```
            return httpVersion == clientKey.httpVersion &&
```

should use equals because could be different instances

graemerocher (changes requested):
Good catch @yawkat there is still an issue

altro3:
@yawkat you forgot this:
```
Moreover the ClientKey::equals use reference equality (==) for the httpVersion field which will be always false.
```

yawkat:
thanks @jpbempel

yawkat:
that one is even less testable unfortunately :(

jpbempel:
> that one is even less testable unfortunately :(

My suggestion would be to test the call `io.micronaut.http.client.netty.DefaultNettyHttpClientRegistry#getClient(io.micronaut.core.annotation.AnnotationMetadata)` twice and make sure the 2 `DefaultHttpClient` instances are the same.
This will call `getClientKey` generating a `ClientKey` instance with the `HttpVersionSelection` instance generated by `forClientAnnotation::forClientAnnotation`

yawkat:
good idea!

graemerocher (approved):

PR #10001, opened 2023-10-20
Support GraalVM JDK 21+ --strict-image-heap flag

graemerocher:
In the next version of GraalVM `--strict-image-heap` will be the default which means that any objects stored in the image heap that are not primitive needs to be initialised at build time.

Currently the only objects Micronaut initialises are build time is the annotation metadata. These can potentially have instantiated conditions (types that implement `io.micronaut.context.condition.Condition`) inside the annotation metadata. This is because `@Requires(condition=..)` member is annotated with `@InstantiatedMember` so that these instances are instantiated and stored in the metadata.

This change alters our GraalVM feature so that these condition objects are marked for initialisation at build time and fixes GraalVM compilation when `--strict-image-heap` is specified. We already had restrictions in place that conditions cannot have constructor arguments. So another restriction that they can't have weird static initialisers that do IO/start threads etc. is not going to be an issue.

A cursory search through existing conditions shows they none of them have any static initialisers so IMO this is a safe change to make in preparation for the next release of GraalVM.

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/service/ServiceLoaderInitialization.java
+++ b/core/src/main/java/io/micronaut/core/io/service/ServiceLoaderInitialization.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.core.io.service;
 
+import io.micronaut.core.annotation.AnnotationClassValue;
 import io.micronaut.core.annotation.AnnotationValue;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.beans.BeanInfo;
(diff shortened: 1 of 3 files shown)


fniephaus:
Maybe add `--strict-image-heap` to the argument list used for testing so that we see it's working? The latest patch release of GraalVM should give you a meaningful error if that is not the case.

graemerocher:
the dev builds will do that will they not?

sdelamo (changes requested):
Can you target 4.2.x as base?

graemerocher:
IMO It makes sense it goes into 4.1.x as well because it is backwards compatible with old GraalVM

sdelamo (approved):

fniephaus:
> the dev builds will do that will they not?

Yes, that is correct and probably the best solution. Thanks!

PR #10010, opened 2023-10-23
Implement zealous block detection for HttpClient

yawkat:
This new check throws an exception if *any* event loop thread would be blocked by a call to the netty BlockingHttpClient. This is usually a bug, as a blocked event loop can lose data.

However there are legitimate cases where this is fine, so there's also a config option to disable this detection. But all the cases in the micronaut-core test suite where this made a test fail were unintentional and potentially buggy, so most users should leave the setting as-is.

The error message provides two code remediation options. It also mentions that there's a config property for this, but doesn't explain which, because I want to discourage an "easy fix" that will leave the bug in place.


http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -733,6 +735,14 @@ public void setAlpnModes(@NonNull List<String> alpnModes) {
         this.alpnModes = Objects.requireNonNull(alpnModes, "alpnModes");
     }
 
+    public boolean isAllowBlockEventLoop() {
+        return allowBlockEventLoop;
+    }
+
+    public void setAllowBlockEventLoop(boolean allowBlockEventLoop) {
+        this.allowBlockEventLoop = allowBlockEventLoop;
```
graemerocher, line 767: add javadoc

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -545,6 +546,16 @@ public void close() {
...
             public <I, O, E> io.micronaut.http.HttpResponse<O> exchange(io.micronaut.http.HttpRequest<I> request, Argument<O> bodyType, Argument<E> errorType) {
+                if (!configuration.isAllowBlockEventLoop() && Thread.currentThread() instanceof FastThreadLocalThread) {
+                    throw new HttpClientException(
+                        "You are trying to run a BlockingHttpClient operation on a netty event " +
+                            "loop thread. This is a common cause for bugs: Event loops should " +
+                            "never be blocked. You can either mark your controller as " +
+                            "@ExecuteOn(TaskExecutors.BLOCKING), or use the reactive HTTP client " +
+                            "to resolve this bug. There is also a configuration option to " +
+                            "disable this check if you are certain a blocking operation is fine " +
+                            "here.");
```
graemerocher: Use a java multiline string since we are Java 17 now

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: it's a single line though

graemerocher: https://openjdk.org/jeps/368#New-escape-sequences

yawkat: oh cool thanks

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -161,6 +161,8 @@ public abstract class HttpClientConfiguration {
 
     private LogLevel logLevel;
 
+    private boolean allowBlockEventLoop = false;
```
sdelamo: Use a a constant 

```java
public boolean DEFAULT_ALLOW_BLOCK_EVENT_LOOP = false;
private boolean allowBlockEventLoop = DEFAULT_ALLOW_BLOCK_EVENT_LOOP;
````
And use the constant in the javadoc of the setter with `Default Value {@value #DEFAULT_ALLOW_BLOCK_EVENT_LOOP}`. The javadoc of the setter should show in the configuration reference documentation.

File: http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java [hunk 4]
```diff
@@ -545,6 +546,16 @@ public void close() {
 
             @Override
             public <I, O, E> io.micronaut.http.HttpResponse<O> exchange(io.micronaut.http.HttpRequest<I> request, Argument<O> bodyType, Argument<E> errorType) {
+                if (!configuration.isAllowBlockEventLoop() && Thread.currentThread() instanceof FastThreadLocalThread) {
+                    throw new HttpClientException("""
+                        You are trying to run a BlockingHttpClient operation on a netty event \
+                        loop thread. This is a common cause for bugs: Event loops should \
+                        never be blocked. You can either mark your controller as \
+                        @ExecuteOn(TaskExecutors.BLOCKING), or use the reactive HTTP client \
+                        to resolve this bug. There is also a configuration option to \
```
sdelamo, line 555:
```suggestion
                        to resolve this bug. There is also a configuration option micronaut.http.client.allow-block-event-loop=true to \
```

>  It also mentions that there's a config property for this, but doesn't explain which, because I want to discourage an "easy fix" that will leave the bug in place.

I don't agree with this. I think we should say which is the config property.

sdelamo (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: people will just enable the setting and move on without fixing their bug

sdelamo: I understand you are trying to encourage correct behavior. 

However, we should not have secret properties they have to seek. If a setting exists, we have to document it. Let's say it is discouraged, and we can warn users about user it. But, we cannot expect our users to clone micronaut core and dive into our codebase to understand what is available.

alvarosanchez: Why not linking to a place in the docs where we explain why this is bad, and also document the property in a warning box?

graemerocher (approved):

timyates (approved):

sdelamo (changes requested):
@yawkat could solve the merge conflicts?

sdelamo:
> @yawkat could solve the merge conflicts?

solved the merge.

PR #10011, opened 2023-10-23
feat: dispatch OPTIONS requests

sdelamo:
It adds `micronaut.server.dispatch-options-requests` configuration setting which by default is `false`. If set to `true`, an OPTION request, unless it is a CORS Preflight request, returns 204 and the Allow HTTP Header populated. 

Spring Boot 3.1. by default sets[`spring.mvc.dispatch-options-request`](https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.web.spring.mvc.dispatch-options-request) to `true` and it behaves the same way as we will in this PR. Spring Boot returns 200, I returned 204 which I think it is more correct.


sdelamo:
@yawkat @graemerocher @dstepanov let me know if you agree with this PR and I will add more docs.

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @RequestFilter
+    @Nullable
+    @Internal
+    public final HttpResponse<?> filterRequest(HttpRequest<?> request) {
+        if (request.getMethod() != HttpMethod.OPTIONS) {
+            return null; // proceed
+        }
+        if (CorsUtil.isPreflightRequest(request)) {
```
dstepanov, line 65: It might be better to run the CORS filter first instead of this check.

dstepanov (changes requested):

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
sdelamo: I have added a commit to ensure the [CorsFilter happens before the Optionsfilter](https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/79ea59684ddcbdb6b13f50a665b1cf68fa4f4738). However, I think it is more correct to leave this if check in place.

dstepanov (approved):

yawkat:
this essentially hijacks all our options processing? that seems wrong

yawkat:
also 200 is better here, cant actually say what content length will be returned by GET

sdelamo:
> this essentially hijacks all our options processing? that seems wrong

what "Option processing" are we currently doing?

sdelamo:
> also 200 is better here, cant actually say what content length will be returned by GET

if the response has no content as in this PR, I think a 204 is more correct.

yawkat:
i mixed up HEAD and OPTIONS, both my points don't apply to OPTIONS.

yawkat (approved):

dstepanov:
@sdelamo Jonas is right, what if somebody defines the OPTIONS methods manually? How do we handle the same scenario with HEAD?

sdelamo:
> @sdelamo Jonas is right; what if somebody defines the OPTIONS methods manually? How do we handle the same scenario with HEAD?

Do you mean if someone annotates a controller method with  `@Options`? We don't have any test of that in our codebase. The `@Options` annotation only appears in the docs  ;-)

I can modify the filter and check for a route match, and if there is a route match, proceed. I assume a route match will happen when you have a @Controller method annotated with `@Options`. Will that fulfill your concerns? @dstepanov

dstepanov:
Sound good, please test the HEAD use case as well

sdelamo:
> Sound good, please test the HEAD use case as well

Please, note this filter proceeds if the request methods is other than option

graemerocher (approved):

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * This Filter intercepts HTTP OPTIONS requests which are not CORS Preflight requests.
+ * It responds with a NO_CONTENT(204) response, and it populates the Allow HTTP Header with the supported HTTP methods for the request URI.
+ * @author Sergio del Amo
+ * @since 4.2.0
+ */
+@Requires(property = "micronaut.server.dispatch-options-requests", value = StringUtils.TRUE, defaultValue = StringUtils.FALSE)
+@ServerFilter(MATCH_ALL_PATTERN)
+public class OptionsFilter implements Ordered {
```
graemerocher: make `final`

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
```diff
@@ -0,0 +1,74 @@
...
+import static io.micronaut.http.annotation.Filter.MATCH_ALL_PATTERN;
+import static io.micronaut.http.server.cors.CorsFilter.CORS_FILTER_ORDER;
+
+/**
+ * This Filter intercepts HTTP OPTIONS requests which are not CORS Preflight requests.
+ * It responds with a NO_CONTENT(204) response, and it populates the Allow HTTP Header with the supported HTTP methods for the request URI.
+ * @author Sergio del Amo
+ * @since 4.2.0
+ */
+@Requires(property = "micronaut.server.dispatch-options-requests", value = StringUtils.TRUE, defaultValue = StringUtils.FALSE)
```
graemerocher: use a constant for this name

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
graemerocher: should this be annotated with `@Internal`?

http-server/src/main/java/io/micronaut/http/server/cors/CorsUtil.java
```diff
@@ -29,13 +29,13 @@
  * @author Graeme Rocher
  * @since 1.0
  */
-class CorsUtil {
+public class CorsUtil {
```
graemerocher: if we make this public annotate with `@Internal`

graemerocher (changes requested):

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -339,6 +348,14 @@ public boolean isHttpToHttpsRedirect() {
         return httpToHttpsRedirect;
     }
 
+    /**
+     *
+     * @return Whether OPTIONS requests should be dispatched.
+     */
+    public boolean isDispatchOptionsRequests() {
```
graemerocher, line 356: add `@since`

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -506,6 +523,14 @@ public void setHttpToHttpsRedirect(boolean httpToHttpsRedirect) {
         this.httpToHttpsRedirect = httpToHttpsRedirect;
     }
 
+    /**
+     *
+     * @param dispatchOptionsRequests  Set to true to dispatch OPTIONS requests. Default value ({@value #DEFAULT_DISPATCH_OPTIONS_REQUESTS}
+     */
+    public void setDispatchOptionsRequests(boolean dispatchOptionsRequests) {
```
graemerocher, line 532: add `@since`

graemerocher (changes requested):

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/417f6514b174d85170bc9b68239642e7d23c6865

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/417f6514b174d85170bc9b68239642e7d23c6865

http-server/src/main/java/io/micronaut/http/server/cors/CorsUtil.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/28b330007dbf003e6db33dfee2667faa5528ed59

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/9ae496b11aaa50848c2ff3b86b7830a7eefca6e8

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/d05b56acd9ebfff3f4085b7c794425beda071f0c

sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/d05b56acd9ebfff3f4085b7c794425beda071f0c

sdelamo:
> > @sdelamo Jonas is right; what if somebody defines the OPTIONS methods manually? How do we handle the same scenario with HEAD?
> 
> Do you mean if someone annotates a controller method with `@Options`? We don't have any test of that in our codebase. The `@Options` annotation only appears in the docs ;-)
> 
> I can modify the filter and check for a route match, and if there is a route match, proceed. I assume a route match will happen when you have a @controller method annotated with `@Options`. Will that fulfill your concerns? @dstepanov

@dstepanov I added it here https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/2f54ebee560e2da08bbbf7e2b1212c4b7a3ec1c9

dstepanov (approved):

graemerocher (approved):

src/main/docs/guide/httpServer/routing.adoc
```diff
@@ -117,6 +117,19 @@ The previous example uses the link:{api}/io/micronaut/http/annotation/Get.html[@
...
 
+== @Options
+
+<<cors, CORS>> support handles OPTIONS preflight requests. However, if you want to dispatch OPTIONS requests without an Origin HTTP Header, you can enable it via:
+
+[configuration]
+----
+micronaut:
+  server:
+    dispatch-options-requests: true
```
graemerocher, line 128: this should be under `CorsConfigration`

graemerocher (changes requested):

src/main/docs/guide/httpServer/routing.adoc
dstepanov: That's doesn't need to related to CORS, Spring MVC has it the sam way

sdelamo: This is not related to CORS or necessary for CORS.

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
```diff
@@ -0,0 +1,93 @@
...
+        if (request.getMethod() != HttpMethod.OPTIONS) {
+            return null; // proceed
+        }
+        if (CorsUtil.isPreflightRequest(request)) {
+            return null; // proceed
+        }
+        if (hasOptionsRouteMatch(request)) {
+            return null; // proceed
+        }
+        MutableHttpResponse<?> mutableHttpResponse = HttpResponse.status(HttpStatus.NO_CONTENT);
```
jeremyg484: While I agree that a `204` response code seems at first to be more semantically correct, I think it is more widely compatible and safer to return a `200` with `Content-Length` of 0. 

The HTTP spec does not explicitly state the status code to use for a general successful OPTIONS request, but web browsers have historically exhibited some strange behavior when handling a `204` in response to CORS pre-flight OPTIONS requests, and the general HTTP 1.1 semantics spec (RFC 7231) *used* to say that a `Content-Length` header of 0 must be set on a successful  OPTIONS response, and the `Content-Length` header is not allowed on a `204` response.  Thus I think there are a number of existing clients that somewhat expect the `200`. 

The most recent relevant spec (https://www.rfc-editor.org/rfc/rfc9110) explicitly removed the `Content-Length` header requirement, but the description of `200` makes explicit mention of OPTIONS requests, whereas the `204` description does not.

sdelamo: Please note this filter does not handle CORS preflight requests. This PR is unrelated to CORS.

jeremyg484: @sdelamo Yes, I understand that. I was using CORS as a specific example of known strange behavior of clients, but I can see how it confused matters.

The HTTP spec definitions I was referring to have nothing to do with CORS either. 

Even the most recently updated version of the specs are (and have historically been) ambiguous on this point, but I think the superfluous inclusion of the `Content-Length` header requirement in [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.7) led some clients to implement OPTIONS requests in a way that expects a 200 response.

Re-reading the most recent version ([RFC 9110 - 15.3.5](https://www.rfc-editor.org/rfc/rfc9110#status.204)) again, I notice the wording of the spec as regards the 204 status is really targeted more towards requests that result in some kind of actual action on the resource in question, such as a PUT or DELETE.

Whereas the description of the [200 status code](https://www.rfc-editor.org/rfc/rfc9110#status.200) explicitly states that a 200 response indicates that the "intended meaning of the content" in response to an OPTIONS request is "communication options for the target resource" which is very much what is being supplied by this filter.

sdelamo: [I have changed to use 200 instead of 204](https://github.com/micronaut-projects/micronaut-core/pull/10011/commits/c05064292883cf6a7b928ddcd65ebe17cb4479e0)

jeremyg484 (approved):

dstepanov (approved):

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -506,6 +524,15 @@ public void setHttpToHttpsRedirect(boolean httpToHttpsRedirect) {
         this.httpToHttpsRedirect = httpToHttpsRedirect;
     }
 
+    /**
+     *
```
yawkat:
```suggestion
     * Set to true to dispatch OPTIONS requests. Default value ({@value #DEFAULT_DISPATCH_OPTIONS_REQUESTS}.
```

yawkat (approved):

graemerocher (approved):

PR #10013, opened 2023-10-23
Change default connection pool idle timeout from 0 to null

yawkat:
This was an oversight, but I think 0 should be handled the same way so not sure this PR changes anything in practice.


graemerocher (approved):

sdelamo (changes requested):
Can you add a test which verifies the default value?

dstepanov (approved):

graemerocher (changes requested):
Add a test

yawkat:
you want me to test that a field is null?

graemerocher (approved):

PR #10021, opened 2023-10-25
Fix logged error when using form fields with blocking filter

### LINKED ISSUE #9917: Warning when using multipart request combined with new FilterRequest
### Expected Behavior

```
@ServerFilter("/**")
class SomeFilter {
	@RequestFilter
	@ExecuteOn(TaskExecutors.BLOCKING)
	fun filterRequest(request: HttpRequest<*>) {

	}
}

@Controller
class Controller {
	@Post("/upload", consumes = [MediaType.MULTIPART_FORM_DATA])
	fun upload(file: ByteArray) : String {
		return "OK"
	}
}

fun main(args: Array<String>) {
	run(*args)
}
```

Trying to submit file should result into no warnings.

Removing @ExecuteOn(TaskExecutors.BLOCKING) fixes the issue (i.e if no blocking filter is present - it works as expected)

### Actual Behaviour

The warning is being logged:

```
[WARN] A task raised an exception. Task: io.micronaut.http.server.netty.FormRouteCompleter$Claimant$$Lambda$1761/0x00000008028ddf10@4fc820d9 java.lang.NullPointerException: Cannot invoke “org.reactivestreams.Subscription.request(long)” because “this.this$0.upstreamSubscription” is null at io.micronaut.http.server.netty.FormRouteCompleter$Claimant.request(FormRouteCompleter.java:255) at io.micronaut.http.server.netty.FormRouteCompleter$Claimant.lambda$request$0(FormRouteCompleter.java:243) at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174) at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:167) at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:470) at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:566) at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) at java.base/java.lang.Thread.run(Thread.java:833)
```

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

https://github.com/schikin/micronaut-filter-warning-demo

### Version

4.x

yawkat:
The blocking filter would cause onSubscribe to be delayed, leading to a logged exception when trying to request from upstream. This patch simply skips the request call in that case. The request will be done in onSubscribe later.

Also make sure the subscriber methods are only called on the event loop to avoid concurrency issues.

Fixes #9917

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterFormSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterFormSpec.groovy
@@ -1,0 +1,55 @@
+package io.micronaut.http.server.netty.filters
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.MediaType
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Post
+import io.micronaut.http.annotation.RequestFilter
+import io.micronaut.http.annotation.ServerFilter
+import io.micronaut.http.client.HttpClient
+import io.micronaut.http.client.multipart.MultipartBody
+import io.micronaut.runtime.server.EmbeddedServer
+import io.micronaut.scheduling.TaskExecutors
+import io.micronaut.scheduling.annotation.ExecuteOn
+import spock.lang.Issue
+import spock.lang.Specification
+
+class FilterFormSpec extends Specification {
+    @Issue('https://github.com/micronaut-projects/micronaut-core/issues/9917')
+    def "exception in form parameter"() {
+        given:
+        def ctx = ApplicationContext.run(["spec.name": "FilterFormSpec"])
+        def server = ctx.getBean(EmbeddedServer)
+        server.start()
+        def client = ctx.createBean(HttpClient, server.URI).toBlocking()
+
+        when:
+        def resp = client.exchange(HttpRequest.POST("/form-error", MultipartBody.builder().addPart("file", "foo").build()).contentType(MediaType.MULTIPART_FORM_DATA_TYPE), String)
+        then:
+        resp.body() == "Ok 3"
+
+        cleanup:
+        client.close()
+        server.close()
+        ctx.close()
+    }
+
+    @Controller
+    @Requires(property = "spec.name", value = "FilterFormSpec")
+    static class Ctrl {
+        @Post(value = "/form-error", consumes = MediaType.MULTIPART_FORM_DATA, produces = MediaType.TEXT_PLAIN)
+        def upload(byte[] file) {
+            return "Ok $file.length"
+        }
+    }
+
+    @ServerFilter("/form-error")
+    static class Fltr {
+        @RequestFilter
+        @ExecuteOn(TaskExecutors.BLOCKING)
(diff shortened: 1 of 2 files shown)


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterFormSpec.groovy
```diff
@@ -0,0 +1,55 @@
...
+import io.micronaut.scheduling.TaskExecutors
+import io.micronaut.scheduling.annotation.ExecuteOn
+import spock.lang.Issue
+import spock.lang.Specification
+
+class FilterFormSpec extends Specification {
+    @Issue('https://github.com/micronaut-projects/micronaut-core/issues/9917')
+    def "exception in form parameter"() {
+        given:
+        def ctx = ApplicationContext.run(["spec.name": "FilterFormSpec"])
```
sdelamo, line 23: Can you replace: 

```groovy
     def ctx = ApplicationContext.run(["spec.name": "FilterFormSpec"])
        def server = ctx.getBean(EmbeddedServer)
        server.start()
```

with: 

```groovy
        EmbeddedServer server = ApplicationContext.run(EmbeddedServer, ["spec.name": "FilterFormSpec"])        
```

sdelamo (changes requested):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/filters/FilterFormSpec.groovy
yawkat: i need the context to create the client though

graemerocher: you can get the context from `server.getApplicationContext()`

yawkat: feels like the wrong way around of doing things to me

sdelamo: We do ` EmbeddedServer server = ApplicationContext.run(EmbeddedServer,` a lot of times in our codebase.

yawkat: yea and i dislike it every time :)

yawkat: we do both approaches a lot.

sdelamo: It is hard for me to understand why you don't want to replace three lines with one line. Do you as you want.

sdelamo (approved):

graemerocher (approved):

PR #10025, opened 2023-10-25
Tck: Tests for error handlers with flux controller methods

jeremyg484:
After investigating https://github.com/micronaut-projects/micronaut-reactor/issues/238, I've got some questions to improve my understanding of how `@Error` handlers work in conjunction with `@Controller` methods that return a `Flux`, and try to figure out if there are any actual issues we need to address.

The test added by this PR exercises the existing behavior in various combinations of a `Flux` returning controller method in conjunction with an `@Error` handler method. It demonstrates the following:

A) A `Flux` returning controller method that is annotated with `@SingleResult` results in the `@Error` handler getting invoked as expected if the Flux signals an error - i.e., `return Flux.error(new MyException())`

B) A `Flux` returning controller method that throws an exception results in the `@Error` handler getting invoked regardless of whether or not the method is annotated with `@SingleResult`.

C) A `Flux` returning controller method that is not annotated with `@SingleResult` results in the `@Error` handler not getting invoked if the Flux signals an error as in scenario A.

D) A `Flux` returning controller method that is not annotated with `@SingleResult` that has an error signaled after the body has already begun to be written in the chunked response causes the connection to be closed and a warning logged without invoking the `@Error` handler as expected.

My questions:

1) I would guess that the behavior of scenario C is intentional because it is a chunked response and it is possible that the headers have already been written. Is that guess correct?

2) If so, would it be possible for us to treat a Flux that immediately signals an error (without any data events) differently? Could we invoke the error handler at that point, or is it too difficult because status and headers have already been written? 

3) Is the behavior in the B scenario also intentional? It seems inconsistent to me that throwing an exception results in different behavior than signaling an error.

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerFluxTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerFluxTest.java
@@ -8,7 +8,7 @@ import io.micronaut.http.HttpStatus;
 import io.micronaut.http.annotation.Controller;
 import io.micronaut.http.annotation.Error;
 import io.micronaut.http.annotation.Get;
-import io.micronaut.http.client.exceptions.ResponseClosedException;
+import io.micronaut.http.client.exceptions.HttpClientException;
 import io.micronaut.http.tck.AssertionUtils;
 import io.micronaut.http.tck.HttpResponseAssertion;
 import org.junit.jupiter.api.Assertions;
@@ -65,7 +65,7 @@ public class ErrorHandlerFluxTest {
             HttpRequest.GET("/errors/flux-chunked-delayed-error"),
             (server, request) -> {
                 Executable e = () -> server.exchange(request);
-                Assertions.assertThrows(ResponseClosedException.class, e);
+                Assertions.assertThrows(HttpClientException.class, e);
             });
     }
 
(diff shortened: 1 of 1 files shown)


sdelamo:
@jeremyg484 did you forget to push the code?

jeremyg484:
> @jeremyg484 did you forget to push the code?

@sdelamo Fixed.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/ErrorHandlerFluxTest.java
```diff
@@ -0,0 +1,149 @@
...
+                } else {
+                    sink.next(data);
+                }
+            });
+        }
+
+        @Error(global = true)
+        public HttpResponse<String> handleMyTestException(HttpRequest<?> request, MyTestException exception) {
+            var error = "Your request is erroneous: " + exception.getMessage();
+            return HttpResponse.<String>status(HttpStatus.BAD_REQUEST, "Bad request")
```
sdelamo: Respond a weird status code .e.g I am a teapot so we know in the test assertion we are being handled by the test error handler.

sdelamo (changes requested):

sdelamo (commented):
The current assertions describe the current behavior, do they match your expected behavior? If not write comments in the test describing why they are not behaving as expected.

jeremyg484:
In separate discussion with @yawkat, it was confirmed that my assumptions in questions (1) and (3) are correct and the existing behavior is as intended. Re: question (2), he also confirmed that it might be possible, but difficult, to change the implementation such that the error handler still gets invoked in the chunked response case if the error is signaled immediately before any data. 

I think it would be good to do this if we can, as it would be more consistent with the behavior of immediately thrown exceptions and less surprising to the user. @sdelamo I have added a comment to the `testErrorHandlerWithFluxChunkedSignaledImmediateError` method to point out this desired change. I will investigate the implementation further to see how it might be done.

The test has also been updated to have the error handler produce a more unusual status code (418) to better ensure the assertions are correctly identifying that the handler was invoked.

sdelamo (approved):

PR #10036, opened 2023-10-27
Deprecations fixes and minor improvements in gradle scripts

altro3:


buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
```diff
@@ -14,7 +14,7 @@ repositories {
     mavenCentral()
 }
 
-configurations.all {
+configurations.configureEach {
```
melix: This is actually not required. Configurations cannot be configured lazily. Actually using `configureEach` on the configuration container can lead to hard to debug errors, so I would simply revert these changes.

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-quality.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 tasks.named("htmlSanityCheck") {
-    sourceDir = new File("${rootProject.buildDir}/docs/guide/")
+    sourceDir = file("${rootProject.layout.buildDirectory.asFile.get()}/docs/guide/")
```
melix: :x: This is a mix of `file` and `layout`.
```suggestion
        sourceDir = rootProject.layout.buildDirectory.dir("docs/guide").get().asFile
```

```

File: core/build.gradle [hunk 3]
```diff
@@ -31,7 +31,7 @@ spotless {
 }
 
 def versionInfo = tasks.register("micronautVersionInfo", WriteProperties) {
-    outputFile = "${buildDir}/resources/version/micronaut-version.properties"
+    destinationFile = file("${layout.buildDirectory.asFile.get()}/resources/version/micronaut-version.properties")
```
melix: :x: `destinationFile` is a proper `RegularFileProperty`

```suggestion
    destinationFile = layout.buildDirectory.file("resources/version/micronaut-version.properties")
```

test-suite-http-server-tck-netty/build.gradle
```diff
@@ -48,7 +48,7 @@ graalvmNative {
         enabled = true
     }
     binaries {
-        all {
+        configureEach {
```
melix: Same here, `configureEach` doesn't make sense since it's a model element, so it will register tasks in any case. This will never be lazy.

melix (changes requested):
Looks almost good. I would revert the `all` changes which do not make sense and fix the `layout` things.

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
altro3: @melix I didi it, because Idea every time tell me, that is configureEach better than all:

изображение

melix: Don't always trust your IDE I guess ;)

altro3:
I reverted `all` and fixed layout

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-quality.gradle
altro3: fixed

core/build.gradle
altro3: fixed

test-suite-http-server-tck-netty/build.gradle
altro3: reverted

melix (approved):

PR #10058, opened 2023-11-01
Error handling for streaming reactive controllers

### LINKED ISSUE #238: Global @Error doesn't work when Controller return is Flux
### Expected Behavior

When an endpoint does return a `Flux<Something>` if an error occurs and there's a global @Error handler defined it should catch the exception.

### Actual Behaviour

Exception is not captured and a 500 error is returned as the response of the HTTP request

### Steps To Reproduce

Having the following Global @Error 
```
@Error(global = true)
public HttpResponse<JsonError> entityNotFoundHandler(HttpRequest<?> request, EntityNotFoundException exception) {
    JsonError error = new JsonError("Entity Not Found: " + exception.getMessage());
    return HttpResponse.<JsonError>status(HttpStatus.NOT_FOUND, "Not found")
            .body(error);
}
```

And the following Endpoint inside a Controller
```
@Get
Flux<Entity> findAllEntities() {
    return service.findAll(); -> This line throws EntityNotFoundException
}
```
The code in the global @Error above is not being executed and the Exception is propagated to the HTTP response which ends up in a 500 status error instead of the 404 specified in the Error handling.

Note that the following works perfect (when Mono is the return type)
```
@Get
Mono<Entity> findById() {
    return service.findById(); -> This line throws EntityNotFoundException;
}
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

3.6.0

jeremyg484:
Error handling for streaming reactive controller methods, such as those that use a `Flux` as the return type, is 
improved such that immediate error signals in the stream that occur before any data has been written will be routed 
appropriately to any user-supplied `@Error` handler methods.

A new test specification is added to cover this scenario, and the TCK's `ErrorHandlerFluxTest` is improved to cover 
the enhanced error handling.

This resolves https://github.com/micronaut-projects/micronaut-reactor/issues/238

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -757,7 +757,7 @@ public final class RouteExecutor {
                 }
                 bufferWritten = true;
                 bufferedFirstValue = null;
-                subscriber.request(Long.MAX_VALUE);
+                bodySink.onRequest(subscriber::request);
             }
         }
 
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -62,10 +62,13 @@ import io.micronaut.web.router.UriRouteMatch;
 import io.micronaut.web.router.exceptions.UnsatisfiedRouteException;
 import jakarta.inject.Singleton;
 import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.CorePublisher;
+import reactor.core.publisher.BaseSubscriber;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
@@ -79,6 +82,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
@@ -559,7 +563,7 @@ public final class RouteExecutor {
                             response = httpResponse.toMutableResponse();
                             final Argument<?> bodyArgument = routeInfo.getReturnType().getFirstTypeVariable().orElse(Argument.OBJECT_ARGUMENT);
                             if (bodyArgument.isAsyncOrReactive()) {
-                                return processPublisherBody(propagatedContext, request, response, routeInfo);
+                                return Mono.from(processPublisherBody(propagatedContext, request, response, routeInfo));
                             }
                         } else {
                             response = forStatus(routeInfo, null);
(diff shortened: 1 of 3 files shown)


http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -667,11 +676,126 @@ private Mono<MutableHttpResponse<?>> processPublisherBody(PropagatedContext prop
...
+                    bodySink.complete();
+                } else {
+                    bodySink.next(bufferedFirstValue);
+                    if (isComplete) {
+                        bodySink.complete();
+                    }
+                }
+                bufferWritten = true;
+                bufferedFirstValue = null;
+                subscriber.request(Long.MAX_VALUE);
```
yawkat: we can't do that, backpressure must be maintained

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -667,11 +676,126 @@ private Mono<MutableHttpResponse<?>> processPublisherBody(PropagatedContext prop
...
+        private boolean isComplete;
+
+        private boolean isEmpty;
+
+        private StreamingDataProcessor(Flux<Object> bodyPublisher) {
+            this.bodyPublisher = bodyPublisher;
+        }
+
+        @SuppressWarnings("unchecked")
+        private CorePublisher<MutableHttpResponse<?>> chunkedResponsePublisher(PropagatedContext propagatedContext,
```
yawkat, line 719: this needs to be thread safe

yawkat:
The general idea is correct but implementing a custom reactive publisher is complicated wrt concurrency, backpressure, nesting etc.. We regularly have bugs like #10017 even with simpler code than this. This PR needs a better reactive processor (without FluxSink etc bc of backpressure).

Even better would be if you could come up with a solution using the built-in reactor operators. I'm not sure if it's possible, but if this PR used only normal operators, it'd be much easier to be sure it works for all the edge cases.

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
jeremyg484: Yes, this was a mistake. I've replace this line with an onRequest callback on the sink that will still respect demand.

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
jeremyg484: @yawkat Maybe I'm missing something...how is this method not thread-safe?

jeremyg484:
> The general idea is correct but implementing a custom reactive publisher is complicated wrt concurrency, backpressure, nesting etc.. We regularly have bugs like #10017 even with simpler code than this. This PR needs a better reactive processor (without FluxSink etc bc of backpressure).
> 
> Even better would be if you could come up with a solution using the built-in reactor operators. I'm not sure if it's possible, but if this PR used only normal operators, it'd be much easier to be sure it works for all the edge cases.

@yawkat FluxSink supports backpressure and I have made a quick fix to take advantage of that. 

I agree in that I would prefer a simpler solution. I attempted several other approaches purely using the built-in operators before settling on this...they all failed to support this use case in a way that did not result in problems such as the first item in the stream needing to generated twice, etc.

Now that I have a working solution and understand the necessary flow completely, I can revisit it one more time to see if there is a way to reproduce this flow purely using the built-in operators, but I too am skeptical that it is possible.

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
yawkat: for example the second accept can be called concurrently with any of the methods of StreamSubscriber

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
yawkat: it can still lose requests when the subscription happens after the downstream request

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
jeremyg484: Oh, you mean the publisher itself. That is a good point. I don't believe that will currently happen just because of the way the publisher is being consumed by the rest of the pipeline, but that is definitely too unsafe an assumption.

jeremyg484:
Thinking through this some more, I agree that I need to try harder to simplify this implementation further. I've got some further ideas to try...I'm going to convert this to a draft until I have a chance to explore them.

PR #10059, opened 2023-11-02
Update Kotlin to 1.9.20

timyates:


timyates:
I *think* this is the right branch for the 4.2.0 micronaut release 🤔

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/KotlinCompilation.kt
```diff
@@ -57,7 +57,11 @@ class KotlinCompilation : AbstractKotlinCompilation<K2JVMCompilerArguments>() {
 	/** Generate metadata for Java 1.8 reflection on method parameters */
 	var javaParameters: Boolean = false
 
-	/** Use the IR backend */
+    /**
+     * Use the IR backend
+     *
+     * @deprecated Removed in 1.9.20
+     */
```
sdelamo: is not a test project? can we not just remove it?

sdelamo (changes requested):

inject-kotlin-test/src/main/kotlin/io/micronaut/annotation/processing/test/support/KotlinCompilation.kt
timyates: You're right...gone

sdelamo (approved):

graemerocher (approved):

PR #10067, opened 2023-11-06
More efficient discard when request body is ignored

yawkat:
If the response is written before the request body is consumed, RIB would subscribe to the body to release the incoming data. This was not necessary, as PipeliningServerHandler already has more efficient discard logic triggered by closeIfNoSubscriber. This made no functional difference, but it's faster with this PR.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/RequestCancelSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/RequestCancelSpec.groovy
@@ -1,0 +1,93 @@
+package io.micronaut.http.server.netty
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.RequestFilter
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/RequestCancelSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+import io.netty.handler.codec.http.HttpMethod
+import io.netty.handler.codec.http.HttpObjectAggregator
+import io.netty.handler.codec.http.HttpResponseStatus
+import io.netty.handler.codec.http.HttpVersion
+import spock.lang.Specification
+
+import java.nio.charset.StandardCharsets
+
+class RequestCancelSpec extends Specification {
+    def 'filter returning early response'() {
```
timyates, line 27: I assume this just checks the filter works, and not the change? 👍

timyates (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/RequestCancelSpec.groovy
yawkat: yes this is just to get some coverage of the deleted code. this test hits the branch that was deleted.

PR #10072, opened 2023-11-06
systemd socket activation support

yawkat:
This PR adds support for binding to a systemd socket. It's a fairly minor internal change, but it could be useful.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
@@ -1476,8 +1476,6 @@ public class NettyHttpServerConfiguration extends HttpServerConfiguration {
         public boolean isBind() {
             return bind;
         }
-
-
         /**
          * Whether the server should bind to the socket. {@code true} by default. If set to
          * {@code false}, the socket must already be bound and listening.
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
@@ -1324,6 +1324,8 @@ public class NettyHttpServerConfiguration extends HttpServerConfiguration {
         private int port;
         private String path;
         private boolean exposeDefaultRoutes = true;
+        private Integer fd = null;
+        private boolean bind = true;
 
         /**
          * Create a TCP listener configuration.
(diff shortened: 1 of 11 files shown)


src/main/docs/guide/httpServer/serverConfiguration/listener.adoc
```diff
@@ -43,3 +43,48 @@ micronaut:
...
+  netty:
+    event-loops:
+      # use epoll
+      parent:
+        prefer-native-transport: true
+      default:
+        prefer-native-transport: true
+    listeners:
+      systemd:
+        fd: 3 # systemd passes the server socket as fd 3
```
timyates, line 63: Is this always `3`?

Or is this dependant on when systemd sets up the socket via the below config?

Are you on a Mac?  If so, how did you test this? ;-)

yawkat: it's always 3. and no, i'm on linux.

yawkat: https://github.com/systemd/systemd/blob/main/src/systemd/sd-daemon.h#L56

graemerocher: can we not default the value to 3 if it is always 3?

yawkat: i think it could be 0 if you were using inetd for example. the 3 is systemd-specific, but the fd config option is more generic.

yawkat: ive thought about detecting it from the env variables that systemd sets. but it would be a bit more complicated of a PR, there's no way to test it, it might have negative perf impact, and it is pretty niche in the first place.

this pr is basically the minimum to make it work with systemd with config options only, with no adverse impact since it's a niche feature.

timyates: > https://github.com/systemd/systemd/blob/main/src/systemd/sd-daemon.h#L56

If I'm reading that right, it starts at 3, but if you have multiple services, they'll be on `3 + n`?

yawkat: right. i dont really see a use case for micronaut there though, so i didnt add it to the docs.

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -1445,6 +1447,47 @@ public void setExposeDefaultRoutes(boolean exposeDefaultRoutes) {
...
+         * Whether the server should bind to the socket. {@code true} by default. If set to
+         * {@code false}, the socket must already be bound and listening.
+         *
+         * @return Whether to bind to the socket
+         */
+        public boolean isBind() {
+            return bind;
+        }
+
+
```
timyates:
```suggestion

```

timyates (approved):
👍 

I can't test it though, as I don't have Epoll :-(

xism4 (approved):

graemerocher (approved):

PR #10076, opened 2023-11-07
Cleanup Gradle JVM compatibility checks

dstepanov:
Not sure about this one https://github.com/micronaut-projects/micronaut-core/blob/4.2.x/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle#L40


graemerocher (approved):

melix (approved):
For the test task you can probably unwrap the jvmArg call, so that tests are executed with `--enable-preview`.

sdelamo (approved):

PR #10079, opened 2023-11-07
Fix goaway test

yawkat:
Seems the test was just broken. The internal error goaway frame made netty close the channel, which may send some bogus TLS and doesn't actually hit the code this test was supposed to test.

Please try on mac.


timyates (approved):
image

PR #10084, opened 2023-11-08
Reduce test pollution with spec.name and requires in http-client tests

### LINKED ISSUE #9844: PoolTimeoutSpec is flaky
### Issue description

https://ge.micronaut.io/scans/tests?search.timeZoneId=Europe/Madrid&tests.container=io.micronaut.http.client.PoolTimeoutSpec&tests.test=cancelled%20pool%20acquire%20should%20not%20leak%20connection

timyates:
In an attempt to reduce (or remove) the flakiness in PoolTimeoutSpec, I've been through the http-client tests looking for places where we were not using spec.name and Requires.

I've tried to introduce some consistency on how the tests are run, with an embedded server created at field level (shared if multiple tests) and annotated with `@AutoCleanup` to ensure it's closed (instead of relying on each spec method to create and close the application)

Hopefully closes #9844

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/aop/RequestAttributeSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/aop/RequestAttributeSpec.groovy
@@ -1,22 +1,7 @@
-/*
- * Copyright 2017-2019 original authors
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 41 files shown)


http-client/src/test/groovy/io/micronaut/http/client/aop/RequestAttributeSpec.groovy
```diff
@@ -37,28 +22,44 @@ class RequestAttributeSpec extends Specification {
...
         given:
-        StoryClient storyClient = context.getBean(StoryClient)
         Story story = storyClient.get()
 
         expect:
         story.storyId == "ABC123"
         story.title == "The Hungry Caterpillar"
     }
 
+    @Issue('https://github.com/micronaut-projects/micronaut-core/issues/6717')
```
yawkat, line 44: this method should stay next to the classes it uses

yawkat (approved):
i'm not a big fan of moving the context to class level, i've had issues with interference between test methods before.

sdelamo (approved):

PR #10093, opened 2023-11-09
Docs: Update reference field name in explaination

dnyaneshwar2023:
Field name in explaination is confusing since above code snippet does not have `my.url`. This statement can be improved by mentioning the first example of using `my.url` but I have updated this according to current example given above.


sdelamo (approved):
Thanks for your contribution

PR #10110, opened 2023-11-13
test: javax/jakarta.persistence.Transient remains in Introspected

sdelamo:


sdelamo:
this fails in Micronaut Data after updating to Micronaut Core 4.2.0 https://github.com/micronaut-projects/micronaut-data/pull/2629

dstepanov:
So do we need to fix it or not?

sdelamo:
I think there is an issue because some tests started failing in Micronaut Data but I am not able to reproduce it in core. At latest these tests verify current core behaviour.

dstepanov (approved):

PR #10118, opened 2023-11-14
Remove warning about Logbook bug

msdousti:
The official documentation mentions a **major bug** for Logbook:

image

It refers to [this issue](https://github.com/zalando/logbook/issues/1216) on the Logbook project, which has been fixed by [this PR](https://github.com/zalando/logbook/pull/1538), and integrated into [Logbook 3.1.0](https://github.com/zalando/logbook/compare/3.0.0...3.1.0) on May 19, 2023.

As such, I'd like to suggest removing this waning from the documentation. Thanks!


sdelamo (approved):
I have updated logbook which we use in the test classpath to 3.6.0

sdelamo:
Test `def 'h2c do not upgrade` fails after upgrading to the latest logbook. @yawkat could you check it out?

yawkat:
ive merged 4.2.x and now the logbook test suite tests fail

msdousti:
The `h2c do not upgrade` test is failing because of these lines:

https://github.com/micronaut-projects/micronaut-core/blob/0530ff704695dedee4d614f2e6cb2e055a6128dc/test-suite/src/test/groovy/io/micronaut/docs/netty/LogbookNettyServerCustomizerSpec.groovy#L338-L339

As explained in the description of the PR, with new Logbook version (3.1 onwards), this bug is resolved, so the "second" `bar` should now be visible in the logs. I'll adapt the test.

For some other reason, another test `tls alpn http 2` started to fail after merging `4.2.x` onto this branch. I bumbed Logbook version on `4.2.x` itself and the test passed. ~~A bit strange...~~

**PS**
The test `tls alpn http 2` failed because of the `@Requires` added in 4fefb40. For the moment, "fixed" by removing it. Please see 6920568857a90b872e7c201f5c83497dd6970ada.

test-suite/src/test/groovy/io/micronaut/docs/netty/LogbookNettyServerCustomizerSpec.groovy
```diff
@@ -480,7 +479,6 @@ class LogbookNettyServerCustomizerSpec extends Specification {
         }
     }
 
-    @Requires(property = 'spec.name', value = 'LogbookNettyServerCustomizerSpec')
```
yawkat: this should stay

yawkat (changes requested):

test-suite/src/test/groovy/io/micronaut/docs/netty/LogbookNettyServerCustomizerSpec.groovy
yawkat: let me check why `tls alpn http 2` is fixed by this

yawkat:
@msdousti can you sign the CLA?

msdousti:
@yawkat Signed.

yawkat:
Thanks!

PR #10104, opened 2023-11-10
test. ThreadSelection accept event loop in the handler in blocking and auto scenarios

sdelamo:
The test is really flaky:

https://ge.micronaut.io/scans/tests?search.timeZoneId=Europe%2FMadrid&tests.container=io.micronaut.http.server.netty.threading.ThreadSelectionSpec

I am not sure why but it seems sometime the handler uses the event loop. I don't know if it is a bug that the sometimes run in the event loop or if we just need to adjust the test as in this pull request. .

PR #10121, opened 2023-11-15
Use bean context classloader for Jackson

dstepanov:
Compiled tests fail for Jackson subtypes


jackson-databind/src/main/java/io/micronaut/jackson/ObjectMapperFactory.java
```diff
@@ -143,6 +147,10 @@ public ObjectMapper objectMapper(@Nullable JacksonConfiguration jacksonConfigura
             objectMapper.findAndRegisterModules();
         }
         objectMapper.registerModules(jacksonModules);
+        if (beanContext != null) {
```
yawkat, line 150: can this be null?

yawkat (approved):

jackson-databind/src/main/java/io/micronaut/jackson/ObjectMapperFactory.java
dstepanov: There are some cases where `ObjectMapperFactory` is used without the bean context

dstepanov:
@yawkat Any idea why the tests are failing?

yawkat:
looks like flakey tests, incl #10104

graemerocher (approved):

PR #10126, opened 2023-11-16
Update OptionsFilter to not emit duplicate headers

jeremyg484:
OptionsFilter is updated to act upon the response and ensure that the `Allow` header is only written one time.

The filter was previously not working as expected with any of the Servlet implementations (including cloud 
functions). The problem being that an initial response has already been generated in the `onRouteMiss` method 
of `RequestLifecycle` and the `Allow` header has already been set using the same route lookup logic that was 
implemented in this filter. The result was that there would be multiple `Allow` headers emitted, containing mostly 
duplicate values.

The approach I've taken is to turn this filter into a `ResponseFilter` that relies on the fact that the allowed 
methods have already been set in the `RequestLifecycle` if there is not any explicit `Options` route implemented by 
the user.

Ultimately this simplifies things a bit, as we no longer need to look up the allowed routes here in the filter. 

Additionally, this ensures that the `Allow` header will only be emitted once, with a comma separated list of 
acceptable methods. The implementation was previously writing out multiple `Allow` headers each with a single method 
value. While this is technically allowed by the HTTP spec, I think single multi-valued header format is 
cleaner and less surprising.

This should clear the way further for being able to successfully update the Micronaut Core dependency to 4.2.x in Servlet, GCP, AWS, etc.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
+++ b/http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
@@ -20,20 +20,29 @@ import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.order.Ordered;
 import io.micronaut.core.util.StringUtils;
-import io.micronaut.http.*;
-import io.micronaut.http.annotation.RequestFilter;
+import io.micronaut.http.HttpAttributes;
+import io.micronaut.http.HttpHeaders;
+import io.micronaut.http.HttpMethod;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.HttpStatus;
+import io.micronaut.http.MutableHttpResponse;
+import io.micronaut.http.annotation.ResponseFilter;
 import io.micronaut.http.annotation.ServerFilter;
 import io.micronaut.http.server.cors.CorsUtil;
-import io.micronaut.web.router.Router;
-import io.micronaut.web.router.UriRouteMatch;
 import io.micronaut.web.router.RouteMatch;
+import io.micronaut.web.router.UriRouteMatch;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 import static io.micronaut.http.annotation.Filter.MATCH_ALL_PATTERN;
 import static io.micronaut.http.server.cors.CorsFilter.CORS_FILTER_ORDER;
 
 /**
  * This Filter intercepts HTTP OPTIONS requests which are not CORS Preflight requests.
- * It responds with a NO_CONTENT(204) response, and it populates the Allow HTTP Header with the supported HTTP methods for the request URI.
+ * It responds with an OK(200) response, and it populates the Allow HTTP Header with the supported HTTP methods for the request URI.
  * @author Sergio del Amo
  * @since 4.2.0
  */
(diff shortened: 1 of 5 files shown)


sdelamo (approved):

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
```diff
@@ -68,13 +67,15 @@ public HttpResponse<?> filterRequest(HttpRequest<?> request) {
...
         }
-        MutableHttpResponse<?> mutableHttpResponse = HttpResponse.status(HttpStatus.OK);
-        router.findAny(request.getUri().toString(), request)
-            .map(UriRouteMatch::getHttpMethod)
-            .map(HttpMethod::toString)
-            .forEach(allow -> mutableHttpResponse.header(HttpHeaders.ALLOW, allow));
-        mutableHttpResponse.header(HttpHeaders.ALLOW, HttpMethod.OPTIONS.toString());
-        return mutableHttpResponse;
+        if (HttpStatus.METHOD_NOT_ALLOWED.equals(response.getStatus())) {
+            List<String> allowedMethods = response.getHeaders().get(HttpHeaders.ALLOW, String[].class)
```
sdelamo, line 71: don't we have a getAll method to simplify this

jeremyg484: IMO, `getAll` makes it more complicated because it does not do any parsing of the values, it just returns the raw value strings. So in this case, the header has already been written in `RequestLifecycle` as something like `Allow: GET,POST,HEAD`. `getAll` then returns a list with a single `String` value of `"GET,POST,HEAD"`. Thus to produce this same result we would need conditional logic to check whether that `String` is empty to determine whether to add a comma before appending `OPTIONS` to it.

wetted:
I had to disable the OptionsFilterTest on for the servlet TCK on https://github.com/micronaut-projects/micronaut-servlet/pull/588. That should be re-enabled once this is merged.

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
```diff
@@ -68,13 +67,15 @@ public HttpResponse<?> filterRequest(HttpRequest<?> request) {
...
             return null; // proceed
         }
-        MutableHttpResponse<?> mutableHttpResponse = HttpResponse.status(HttpStatus.OK);
-        router.findAny(request.getUri().toString(), request)
-            .map(UriRouteMatch::getHttpMethod)
-            .map(HttpMethod::toString)
-            .forEach(allow -> mutableHttpResponse.header(HttpHeaders.ALLOW, allow));
-        mutableHttpResponse.header(HttpHeaders.ALLOW, HttpMethod.OPTIONS.toString());
-        return mutableHttpResponse;
+        if (HttpStatus.METHOD_NOT_ALLOWED.equals(response.getStatus())) {
```
yawkat, line 70: at this point might as well move the entire logic into RequestLifecycle or wherever METHOD_NOT_ALLOWED is thrown...

yawkat (approved):

http-server/src/main/java/io/micronaut/http/server/OptionsFilter.java
jeremyg484: @yawkat Yes, I had that same thought. It happens in [RequestLifecycle#onRouteMiss](https://github.com/micronaut-projects/micronaut-core/blob/622cb9698249db85837194396d07bf35147ba9fa/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java#L305).

My main concern, and reason for not doing that yet, is that `onRouteMiss` executes before `runWithFilters` and the original intention of this filter to avoid conflicting with the CORS filter. 

That said, it occurs to me now that the `onRouteMiss` logic must be getting executed anyway (and the `Allow` header getting populated) even in the case of CORS being enabled. Thus in a Servlet environment, we are likely sending the `Allow` header even in response to a CORS pre-flight request (alongside all the other headers that CORS adds) because of the same issue of not being able to create a new ServletResponse in the middle of request processing.

PR #10131, opened 2023-11-17
Add fast compiled route matcher

yawkat:
Note this is two commits that are mostly independent, but the second commit has the vast majority of the changes.

---

This PR adds an alternate request processing path that mostly bypasses RoutingInBoundHandler. The goal is to speed up processing of simple routes significantly.

The approach taken here is to accumulate all routes, prepare those that can be executed quickly (i.e. simple argument / return binding, exact path match, no difficult filters), and then compile the routing tree into an optimized MatchPlan. In principle this approach could be extended to all routes, we "just" need to adapt other framework features (e.g. path parameters) to be included in the compiled MatchPlan. This is my long-term vision for the future of routing in the Micronaut HTTP stack.

The results are very promising so far. FullHttpStackBenchmark goes from ~7µs to ~5µs, a big improvement. And there aren't even very many microoptimizations yet.


graemerocher:
Converted to draft because I think we will need to iterate a bit on this and @dstepanov and me are away next weeks. Will try do an initial review Monday

http-netty/src/main/java/io/micronaut/http/netty/body/ShortCircuitNettyBodyWriter.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param requestHeaders  The request headers
+     * @param status          The response status
+     * @param responseHeaders The response headers
+     * @param object          The object to write
+     * @param nettyContext    The netty context
+     * @throws CodecException If an error occurs decoding
+     */
+    void writeTo(
+        HttpHeaders requestHeaders,
```
graemerocher: does this mean we need to parse the headers? Or could this be the raw Netty headers

http-netty/src/main/java/io/micronaut/http/netty/body/ShortCircuitNettyBodyWriter.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * {@link NettyBodyWriter} extension that can do without a {@link HttpRequest}.
+ *
+ * @param <T> The body type
+ * @since 4.3.0
+ * @author Jonas Konrad
+ */
+@Internal
+@Experimental
+public interface ShortCircuitNettyBodyWriter<T> extends NettyBodyWriter<T> {
```
graemerocher: could this be sealed for the moment?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -144,6 +167,14 @@ public final class RoutingInBoundHandler implements RequestHandler {
         this.multipartEnabled = isMultiPartEnabled.isEmpty() || isMultiPartEnabled.get();
         this.routeExecutor = embeddedServerContext.getRouteExecutor();
         this.conversionService = conversionService;
+
+        if (serverConfiguration.isFastRouting()) {
```
graemerocher: rename to `isOptimizedRouting()`

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -221,6 +263,124 @@ public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRe
...
+            return and.rules().stream()
+                .map(RoutingInBoundHandler::findFixedContentType)
+                .filter(Objects::nonNull)
+                .findFirst().orElse(null);
+        } else {
+            return null;
+        }
+    }
+
+    private ExecutionLeaf<RequestHandler> shortCircuitHandler(MatchRule rule, UriRouteInfo<?, ?> routeInfo) {
```
graemerocher: this is a long method I think we need to make this logic extensible and split this up, can the `MathRule` interface be extended somehow such that it can handle different return types, method parameters etc.?

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -221,6 +263,124 @@ public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRe
...
+            rawWriter = raw;
+            scWriter = null;
+        } else {
+            return ExecutionLeaf.indeterminate();
+        }
+        return new ExecutionLeaf.Route<>(new RequestHandler() {
+            @Override
+            public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRequest request, ByteBody body, PipeliningServerHandler.OutboundAccess outboundAccess) {
+                try {
+                    NettyHttpHeaders requestHeaders = new NettyHttpHeaders(request.headers(), conversionService);
```
graemerocher: like I said do we need this headers object?

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
```diff
@@ -159,4 +163,38 @@ Optional<T> transform(NettyHttpRequest<?> nhr, ArgumentConversionContext<T> cont
             .convert(conversionService, context)
             .map(o -> (T) o.claimForExternal());
     }
+
+    @Override
+    public Optional<Prepared> prepare(Argument<T> argument, MatchRule.ContentType fixedContentType) {
```
graemerocher, line 176: since this is going to be an internal non-public API do we need the overhead of `Optional`? Can we just make the return value nullable and return null?

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -728,6 +737,30 @@ public void setJsonBufferMaxComponents(int jsonBufferMaxComponents) {
...
 
+    /**
+     * Whether fast routing should be enabled for routes that are simple enough to be supported.
+     * During fast routing, not all framework features may be available, such as the thread-local
+     * request context or some error handling. Default {@value DEFAULT_FAST_ROUTING}.
+     *
+     * @return {@code true} if fast routing should be enabled
+     * @since 4.3.0
+     */
+    public boolean isFastRouting() {
```
graemerocher: rename if `optimizedRouting`

http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/DiscriminatorStage.java
```diff
@@ -0,0 +1,144 @@
...
+enum DiscriminatorStage {
+    PATH {
+        @Override
+        <R> MatchPlan<R> planDiscriminate(Map<MatchRule, MatchPlan<R>> nextPlans) {
+            // simplify already removed any overlaps between patterns and exact matches. only the exact matches are left
+            Map<String, MatchPlan<R>> byString = nextPlans.entrySet().stream()
+                .filter(e -> e.getKey() instanceof MatchRule.PathMatchExact)
+                .collect(Collectors.toMap(e -> ((MatchRule.PathMatchExact) e.getKey()).path(), Map.Entry::getValue));
+            return request -> {
+                // this replicates AbstractNettyHttpRequest.getPath but not exactly :(
```
graemerocher: please clarify

http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/DiscriminatorStage.java
```diff
@@ -0,0 +1,144 @@
...
+            };
+        }
+    },
+    CONTENT_TYPE {
+        @Override
+        <R> MatchPlan<R> planDiscriminate(Map<MatchRule, MatchPlan<R>> nextPlans) {
+            Map<String, MatchPlan<R>> byContentType = coerceRules(MatchRule.ContentType.class, nextPlans)
+                .entrySet().stream().collect(Collectors.toMap(e -> {
+                    MediaType expectedType = e.getKey().expectedType();
+                    return expectedType == null ? null : expectedType.getName();
```
graemerocher: a null key?

http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/ShortCircuitArgumentBinder.java
```diff
@@ -0,0 +1,62 @@
...
+    interface Prepared {
+        /**
+         * Bind the parameter.
+         *
+         * @param nettyRequest The netty request
+         * @param mnHeaders    The request headers (micronaut-http class)
+         * @param body         The request body
+         * @return The bound argument
+         */
+        Object bind(@NonNull HttpRequest nettyRequest, HttpHeaders mnHeaders, @NonNull ImmediateByteBody body);
```
graemerocher: since this is a new interface maybe `httpHeaders`  can be the netty headers object

graemerocher (changes requested):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/shortcircuit/FastRoutingSpec.groovy
```diff
@@ -0,0 +1,55 @@
...
+        where:
+        fastRouting << [false, true]
+    }
+
+    @Controller
+    @Requires(property = "spec.name", value = "FastRoutingSpec")
+    static class MyController {
+        @Get("/simple")
+        String simple() {
+            return "foo: " + ServerRequestContext.currentRequest().isPresent()
```
graemerocher, line 44: instead of completely disabling `ServerRequestContext.currentRequest()` perhaps we can make it lazy. We bind the netty request to the propagation context and then if the Micronaut request is resolved with `currentRequest()` we initialise it.

dstepanov:
If I understand correctly, we have to duplicate matching and method invocation logic to do this. The already complicated code gets more complex, and it's not worse IMHO. I feel it's going to be another option like `HttpServerType`, and is never going to be enabled. We should try to optimize `RoutingInBoundHandler` if it's that problematic. I see that we don't cache filters, which might be a good idea to add.

When we discussed this task, I assumed we wanted to optimize the Netty pipeline as it is the most visible in the profiling. (Note: I don't fully understand Netty)
My idea was to rewrite this code to be more dynamic based on the route we resolve:
```java
private void insertMicronautHandlers(boolean zeroCopySupported) {
            channel.attr(STREAM_PIPELINE_ATTRIBUTE.get()).set(this);
            if (sslHandler != null) {
                channel.attr(CERTIFICATE_SUPPLIER_ATTRIBUTE.get()).set(sslHandler.findPeerCert());
            }

            SmartHttpContentCompressor contentCompressor = new SmartHttpContentCompressor(embeddedServices.getHttpCompressionStrategy());
            if (zeroCopySupported) {
                channel.attr(PipeliningServerHandler.ZERO_COPY_PREDICATE.get()).set(contentCompressor);
            }
            pipeline.addLast(ChannelPipelineCustomizer.HANDLER_HTTP_COMPRESSOR, contentCompressor);
            pipeline.addLast(ChannelPipelineCustomizer.HANDLER_HTTP_DECOMPRESSOR, new HttpContentDecompressor());

            Optional<NettyServerWebSocketUpgradeHandler> webSocketUpgradeHandler = embeddedServices.getWebSocketUpgradeHandler(server);
            if (webSocketUpgradeHandler.isPresent()) {
                pipeline.addLast(NettyServerWebSocketUpgradeHandler.COMPRESSION_HANDLER, new WebSocketServerCompressionHandler());
            }
            if (server.getServerConfiguration().getServerType() != NettyHttpServerConfiguration.HttpServerType.STREAMED) {
                pipeline.addLast(ChannelPipelineCustomizer.HANDLER_HTTP_AGGREGATOR,
                    new HttpObjectAggregator(
                        (int) server.getServerConfiguration().getMaxRequestSize(),
                        server.getServerConfiguration().isCloseOnExpectationFailed()
                    )
                );
            }
            pipeline.addLast(ChannelPipelineCustomizer.HANDLER_HTTP_CHUNK, new ChunkedWriteHandler()); // todo: move to PipeliningServerHandler

            RequestHandler requestHandler = routingInBoundHandler;
            if (webSocketUpgradeHandler.isPresent()) {
                webSocketUpgradeHandler.get().setNext(routingInBoundHandler);
                requestHandler = webSocketUpgradeHandler.get();
            }
            if (server.getServerConfiguration().isDualProtocol() && server.getServerConfiguration().isHttpToHttpsRedirect() && sslHandler == null) {
                requestHandler = new HttpToHttpsRedirectHandler(routingInBoundHandler.conversionService, server.getServerConfiguration(), sslConfiguration, hostResolver);
            }
            pipeline.addLast(ChannelPipelineCustomizer.HANDLER_MICRONAUT_INBOUND, new PipeliningServerHandler(requestHandler));
        }
 ```
 We add those steps after we resolve the route. So the idea:
1. Add a step as much in the front as possible where we can get the request
2. Convert the request into our abstraction, probably as lazy as possible. I don't think a basic hello-world with disabled CORS should touch headers
3. Find a route
4. Based on the route, add pipeline steps. GET method should not add content decompression etc. Websocket added only for WS routes. I assume we can add the steps needed for the response (chunks, decompression) after we know what kind it is and if it does exist. Looking at `HttpContentDecompressor` and `HttpContentCompressor`, those can be skipped for some content types.

**Comment by dstepanov** [comment]
I guess you can comment-out those steps and see if that is worse the change.

File: http-netty/src/main/java/io/micronaut/http/netty/body/ShortCircuitNettyBodyWriter.java (hunk 1 above)
Comment (yawkat) [reply http-netty/src/main/java/io/micronaut/http/netty/body/ShortCircuitNettyBodyWriter.java]: `new NettyHttpHeaders(headers, conversionService)` has no logic so this is not a big concern. i tried adding a new header abstraction first to abstract over netty and micronaut http header classes but i saw NettyHttpHeaders is so light that it's pointless.

File: http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java (hunk 6 above)
Comment (yawkat) [reply http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java:176]: either is fine but this is not a hot path. this is called once per route, so i think it's better to have a nicer api here than to overoptimize this.

File: http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/DiscriminatorStage.java (hunk 9 above)
Comment (yawkat) [reply http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/DiscriminatorStage.java]: For MatchRule.ContentType, `null` means no content type header. it's convenient because `request.headers().get(HttpHeaderNames.CONTENT_TYPE)` returns `null` when the content type is missing, which will then match `MatchRule.ContentType(null)`

File: http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/ShortCircuitArgumentBinder.java (hunk 10 above)
Comment (yawkat) [reply http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/ShortCircuitArgumentBinder.java]: see above

File: http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java (hunk 5 above)
Comment (yawkat) [reply http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java]: see above

File: http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/DiscriminatorStage.java (hunk 8 above)
Comment (yawkat) [reply http-server-netty/src/main/java/io/micronaut/http/server/netty/shortcircuit/DiscriminatorStage.java]: AbstractNettyHttpRequest.getPath for some reason uses the charset from the content-type of the request, if set. i do not want the charset parsing logic anymore. i dont think anyone uses it and it adds overhead to every request.

**Comment by yawkat** [comment]
@dstepanov

This PR addresses RoutingInBoundHandler, RequestLifecycle and so on. It is true that routing and execution logic is duplicated, but I do not see a way around this. The current routing *API* (not just the implementation) is sub-optimal, and the only solution I could see is to flip it around (accumulate the routes in a builder and use essentially a compiler to build them into a final optimized routing tree). Yes it is similar to HttpServerType, and people probably won't turn it on, but don't forget that HttpServerType also got us the PipeliningServerHandler which now gives the benefit of HttpServerType.FULL_CONTENT for most requests (those that don't exceed a certain body size) for *everyone* even without FULL_CONTENT enabled. Similarly, in my opinion this new approach can be extended (maybe with breaking some compat, so 5.x) to extend to most or all routes by default.

---

The netty pipeline is a separate issue. It is not as easy as you say to simply add the handlers dynamically. Many of the handlers need to see the full request lifecycle so cannot be added halfway into the request dynamically. Additionally, HTTP pipelining is a huge issue when it comes to modifying the handler pipeline. There can be many requests in flight at the same time in the same pipeline and it's hard to avoid bugs when modifying it dynamically.

However we already *have* compressed some parts of the pipeline with the introduction of PipeliningServerHandler. The job that was done before by I think three separate handlers is now done by one. In my opinion, this is the correct approach to remove the other handlers also: The chunked write handler and the compression handlers should go into PipeliningServerHandler where they can be loaded dynamically per request, and where HTTP pipelining is already handled properly. PipeliningServerHandler with its InboundHandler and OutboundHandler internal abstractions is already set up to do exactly this, without adding overhead for uncompressed requests, it's just not implemented yet. But such an optimization cannot be a substitute for improved routing.

File: http-server-netty/src/test/groovy/io/micronaut/http/server/netty/shortcircuit/FastRoutingSpec.groovy (hunk 11 above)
Comment (yawkat) [reply http-server-netty/src/test/groovy/io/micronaut/http/server/netty/shortcircuit/FastRoutingSpec.groovy:44]: Yes that is a good idea, it would also make things like fallback for error routes easier.

File: http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java (hunk 4 above)
Comment (yawkat) [reply http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java]: Not sure what you mean by extending MatchRule. MatchRule's only purpose is to match the request, essentially what Router.find does at the moment. The one thing Router.find does that MatchRule can't is to return e.g. path parameters, but imo this is a good thing because the stateless nature enables NettyShortCircuitRouterBuilder to do logical operations (eg DNF transformation) without restrictions. It is imo better to keep MatchRule simple and add further steps (eg path param extraction) later only if they are necessary. 

wrt extensibility, yes it would be nice. What I did for this method is go through RequestLifecycle and exclude all the weird cases (e.g. HttpStatus method return type). When you remove those exclusions, the prepare step already becomes a lot simpler:

```java
MethodExecutionHandle<?, ?> executionHandle = routeInfo.getTargetMethod();
        boolean unwrapResponse = HttpResponse.class.isAssignableFrom(executionHandle.getReturnType().getType());
        MatchRule.ContentType fixedContentType = findFixedContentType(rule);
        MediaType responseMediaType;
        if (fixedContentType != null) {
            responseMediaType = fixedContentType.expectedType();
        } else {
            List<MediaType> produces = routeInfo.getProduces();
            if (!produces.isEmpty()) {
                responseMediaType = produces.get(0);
            } else {
                responseMediaType = MediaType.APPLICATION_JSON_TYPE;
            }
        }
        RequestArgumentBinder<Object>[] argumentBinders = routeInfo.resolveArgumentBinders(requestArgumentSatisfier.getBinderRegistry());
        ShortCircuitArgumentBinder.Prepared[] shortCircuitBinders = new ShortCircuitArgumentBinder.Prepared[argumentBinders.length];
        for (int i = 0; i < argumentBinders.length; i++) {
            Optional<ShortCircuitArgumentBinder.Prepared> prep = scb.prepare(executionHandle.getArguments()[i], fixedContentType);
            shortCircuitBinders[i] = prep.get();
        }
        MessageBodyWriter<Object> messageBodyWriter = (MessageBodyWriter<Object>) routeInfo.getMessageBodyWriter();
```

It's essentially just content type determination for the MessageBodyWriter, and the prep stage for the argument binders. Similarly, the actual accept implementation is very simple (call arg binders, invoke method, invoke body writer), it's just a bit blown up by the need to handle HttpResponse return types for FullHttpStackBenchmark.

So I think that the current extension points (MessageBodyWriter, argument binders) are already fine conceptually, they just need some refinement. For example maybe there should just be a MessageBodyWriter for HttpStatus, and the special handling for HttpStatus could be removed. And there should also be a MessageBodyWriter for HttpResponse though that is a bit harder.

Until we have some improvements there I don't think we should add new abstractions to simplify this method.

http-netty/src/main/java/io/micronaut/http/netty/body/ShortCircuitNettyBodyWriter.java
yawkat: sure

graemerocher:
I think this is an interesting line of investigation and we should continue it, it could be that it becomes the default if we can make it pass the HTTP TCK.

I think compatibility is possible if we make certain things lazy (ServerRequestContext). I agree we should probably not have a configuration option for it as no one will enable it, and the goal should be full compatibility somehow. 

Also I think we should investigate optimising route argument binding so that it can use the raw netty headers and provide an optimised API without all the `BindingResult/Optional` wrapping which will reduce allocations.

Looking at something like Vertx, they have separate routing for normal/templates routes for regex routes as well. So I think the only way to achieve that level of speed is to have static route lookups without the current route matching we have now.

yawkat:
I think a config option is unavoidable for 4.x. Though they are a bit reduced from 3.x with the MessageBodyHandler introduction, we *still* have many cases where certain decisions cannot be done statically. For example, RouteExecutor does an `body instanceof HttpStatus` every time on controller return type instead of determining this from the route statically. So if you have a controller `Object get() {return HttpStatus.NOT_FOUND;}` we can't know statically that this needs special handling instead of going through the standard json body writer.

I want to get rid of these cases and make all response handling decisions statically (this example decision is made statically in this PR), but I don't think this is viable for 4.x without an explicit config option.

graemerocher:
You do know statically that it is a dynamic route since the return type is `Object` and therefore has to fallback to the slower original behaviour

yawkat:
yea i guess we can be more conservative with Object return type and skip a lot of the logic.

dstepanov:
I will take a look in more detail next week, but I would prefer optimizing the current routing/invocation or reusing it instead of duplicating the implementation.

graemerocher:
Agree that if it is possible to optimise the existing routing that is preferable to duplicate routing

yawkat:
fyi @dstepanov when i'm on vacation, feel free to look at the routing side but please leave the netty pipeline alone, i have changes already lined up for that.

dstepanov:
I'm working on optimizing filtering/routing. It looks like this change detected the overhead of it.

yawkat:
i added filter support, it added maybe 100ns static overhead + 100ns per filter. cors filter is still skipped to avoid those 100ns.

dstepanov:
Can you extract the propagation improvements into a separate PR?

yawkat:
sure

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -212,22 +239,169 @@ public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRe
...
+    private ExecutionLeaf<PreparedHandler> shortCircuitHandler(MatchRule rule, UriRouteInfo<?, ?> routeInfo) {
+        if (routeInfo.isWebSocketRoute()) {
+            return ExecutionLeaf.indeterminate();
+        }
+        List<GenericHttpFilter> fixedFilters = routeExecutor.getRouter().getFixedFilters().orElse(null);
+        if (fixedFilters == null) {
+            return ExecutionLeaf.indeterminate();
+        }
+        // CorsFilter is handled specially here. It's always present, so we can't bail, but it only does anything when the Origin header is set, which is checked in accept().
+        fixedFilters = fixedFilters.stream().filter(f -> !FilterRunner.isCorsFilter(f, CorsFilter.class)).toList();
```
dstepanov, line 300: Do we still need this? Filters can now be filtered by request, eliminating the CORS filter.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -212,22 +239,169 @@ public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRe
...
+        };
+        String serverHeader = serverConfiguration.getServerHeader().orElse(null);
+        boolean dateHeader = serverConfiguration.isDateHeader();
+        return new ExecutionLeaf.Route<>((nhr, outboundAccess) -> {
+            try {
+                new FilterRunner(finalFixedFilters, exec).run(nhr, PropagatedContext.empty()).onComplete((response, err) -> {
+                    if (err != null) {
+                        RoutingInBoundHandler.this.handleUnboundError(err);
+                    } else {
+                        HttpHeaders responseHeaders = ((NettyHttpHeaders) response.getHeaders()).getNettyHeaders();
```
dstepanov: Use `contentType()`

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
yawkat: i will profile

altro3:
Is it dead improvement?

yawkat:
@altro3 we've decided that for now the marginal perf benefit does not outweigh the maintenance effort of duplicating the routing code paths. this decision may change in the future however

PR #10136, opened 2023-11-20
Guide: Added instructions for setup NetBeans for Micronaut

lkishalmi:
Well, there is not too much setup needed to work with Micronaut in NetBeans. Though I think it would be nice to mention it.


sdelamo:
> Well, there is not too much setup needed to work with Micronaut in NetBeans. Though I think it would be nice to mention it.

would you be willing to contribute a guide. Similar to [what we have for IntelliJ IDEA](https://guides.micronaut.io/latest/micronaut-intellij-idea-ide-setup.html).

lkishalmi:
I think I can do that. Even if there is not too much to add there, maybe the new Micronaut project wizard and using Gradle continuous mode to work on Micronaut projects really effectively.

Otherwise it is full NetBeans Experience: Works out of the box.

sdelamo (approved):

PR #10138, opened 2023-11-20
Replace chunked write API

yawkat:
This PR replaces the writeChunked and writeFile APIs with a new writeStream API that takes an InputStream. This removes the need for the ChunkedWriteHandler.

Chunked writes were used for two purposes: Sending file regions and sending InputStreams. This has always complicated the HTTP pipeline somewhat as the pipeline had to deal with not just HttpContent objects but also ChunkedInput and FileRegion objects.

This PR replaces the machinery for InputStream writing with a more straight-forward solution that reads the data on the IO thread and then sends it down the channel.

Additionally, the file-specific APIs based on RandomAccessFile are removed. The body writer now just creates an InputStream for the file region in question and sends that. This removes support for zero-copy transfers, however that is a niche feature anyway because it doesn't work with TLS or HTTP/2. If someone wants a performant HTTP server, HTTP/2 takes priority over zero-copy so it makes little sense.

This PR may have small conflicts with #10131 as that PR changed the PipeliningServerHandler body handling a little bit. Otherwise this PR should have no visible impact on users.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/body/SystemFileBodyWriter.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/body/SystemFileBodyWriter.java
@@ -33,15 +33,23 @@ import io.micronaut.http.netty.body.NettyBodyWriter;
 import io.micronaut.http.netty.body.NettyWriteContext;
 import io.micronaut.http.server.netty.configuration.NettyHttpServerConfiguration;
 import io.micronaut.http.server.types.files.SystemFile;
+import io.micronaut.scheduling.TaskExecutors;
 import io.netty.handler.codec.http.DefaultHttpResponse;
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValues;
+import jakarta.inject.Named;
 import jakarta.inject.Singleton;
+import org.jetbrains.annotations.NotNull;
 
+import java.io.EOFException;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.RandomAccessFile;
+import java.util.concurrent.ExecutorService;
 
 import static io.micronaut.http.HttpHeaders.CONTENT_RANGE;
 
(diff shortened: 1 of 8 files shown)


timyates:
> This removes support for zero-copy transfers, however that is a niche feature anyway because it doesn't work with TLS or HTTP/2. If someone wants a performant HTTP server, HTTP/2 takes priority over zero-copy so it makes little sense.

Does the techempower test take advantage of zero copy transfers? 🤔

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/SystemFileBodyWriter.java
```diff
@@ -175,4 +174,90 @@ private static class IntRange {
         }
     }
 
+    private static class RafInputStream extends InputStream {
```
timyates: Is this class used?

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/SystemFileBodyWriter.java
```diff
@@ -113,34 +124,22 @@ public void writeTo(HttpRequest<?> request, MutableHttpResponse<SystemFile> resp
 
                 // Write the request data
                 final DefaultHttpResponse finalResponse = new DefaultHttpResponse(nettyResponse.getNettyHttpVersion(), nettyResponse.getNettyHttpStatus(), nettyResponse.getNettyHeaders());
-                writeFile(systemFile, nettyContext, position, contentLength, finalResponse);
+
+                File file = systemFile.getFile();
+                InputStream is;
+                try {
+                    is = new FileInputStream(file);
```
timyates, line 130: Who closes this?

timyates (approved):
I *think* there's an unused class, and I'm not sure who closes FileInputStreams

But that's probably a gap in my knowledge, rather than an issue here

I always feel out of my depth with these reviews 😉

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/SystemFileBodyWriter.java
yawkat: nope, will remove

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/SystemFileBodyWriter.java
yawkat: it is passed on to PipeliningServerHandler where BlockingOutboundHandler.work closes it using try-with-resources

yawkat:
@timyates it does not, it doesn't serve file system files afaik.

timyates:
@dstepanov Do you know if any of the techempower tests serve files from the filesystem and might be affected by the removal of zero-copy transfers?

yawkat:
denis is on vacation.

ive looked, and they don't use the file system.

dstepanov (approved):

PR #10142, opened 2023-11-21
Fix request backpressure

yawkat:
PipeliningServerHandler was supposed to implement backpressure, but it turns out that auto read was still enabled and that the implementation didn't really work. This means that it would keep reading even if that means buffering data when the downstream can't keep up.

This PR disables auto read and fixes the read implementation in PipeliningServerHandler. In principle there should be no change to users, this just means that instead of buffering any incoming data internally, backpressure is now applied to the client.

This PR is based on #10138 but is separate for easier review. It also has conflicts with #10131. I will retarget this pr for 4.3.x when #10138 is merged.


timyates (commented):
`StreamPressureSpec` seems to hang when I run it locally

yawkat:
@timyates huh. only in this pr or also in #10138? i dont see how this pr would affect StreamPressureSpec

timyates:
@yawkat don't know... I only tested here as none of the CI actions run on 2nd level PRs

dstepanov (approved):

PR #10149, opened 2023-11-22
[bug] EL fails when using primitive method arguments in expression

timyates:
The first test in this PR with `Integer` as the argument type works as expected:

```
                @Executable
                @Requires(value = "#{ #second + 'abc' }")
                void test(String first, Integer second) {
                }
```

However, when we use a primitive for the second argument (as in the second test):

```
                @Executable
                @Requires(value = "#{ #second + 'abc' }")
                void test(String first, int second) {
                }
```

We get a `VerifyError`

```
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    test/$Expr$Expr4.doEvaluate(Lio/micronaut/core/expressions/ExpressionEvaluationContext;)Ljava/lang/Object; @17: invokevirtual
  Reason:
    Type 'I' (current frame, stack[1]) is not assignable to integer
  Current Frame:
    bci: @17
    flags: { }
    locals: { 'test/$Expr$Expr4', 'io/micronaut/core/expressions/ExpressionEvaluationContext' }
    stack: { 'java/lang/StringBuilder', 'I' }
  Bytecode:
    0000000: bb00 0d59 b700 102b 04b9 0016 0200 c000
    0000010: 18b6 001c 121e b600 21b6 0025 b0       
```

Whilst this can be worked around in Java (by using the wrapper class), in Kotlin we need to introduce our own container class, as it also fails with `Int` as seen in the cache test here https://github.com/micronaut-projects/micronaut-cache/blob/master/test-suite-caffeine-kotlin/src/test/kotlin/io/micronaut/cache/ConditionalService.kt#L14

### DIFF
--- a/core-processor/src/main/java/io/micronaut/expressions/parser/ast/access/ContextMethodParameterAccess.java
+++ b/core-processor/src/main/java/io/micronaut/expressions/parser/ast/access/ContextMethodParameterAccess.java
@@ -58,7 +58,11 @@ final class ContextMethodParameterAccess extends ExpressionNode {
         // invoke getArgument method
         mv.invokeInterface(EVALUATION_CONTEXT_TYPE, GET_ARGUMENT_METHOD);
         if (nodeType != null) {
-            mv.checkCast(nodeType);
+            if (TypeDescriptors.isPrimitive(nodeType)) {
+                mv.unbox(nodeType);
+            } else {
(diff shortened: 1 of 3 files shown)


timyates:
😎

sdelamo (approved):

PR #10155, opened 2023-11-23
Implement decompression in PipeliningServerHandler

yawkat:
This patch implements the logic of HttpContentDecompressor in PipeliningServerHandler. This allows us to shrink the pipeline a little. The perf impact for uncompressed requests should basically be zero.

This builds on the changes in #10142.


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
```diff
@@ -328,18 +335,81 @@ void read(Object message) {
...
+
+            boolean full = request instanceof FullHttpRequest;
+            if (full && decompressionChannel == null) {
+                requestHandler.accept(ctx, request, outboundAccess);
             } else if (!hasBody(request)) {
                 inboundHandler = droppingInboundHandler;
                 if (message instanceof HttpContent) {
                     inboundHandler.read(message);
                 }
+                if (decompressionChannel != null) {
```
dstepanov, line 376: You can avoid creating a decompressor for this case

dstepanov (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/PipeliningServerHandlerSpec.groovy
```diff
@@ -354,6 +361,161 @@ class PipeliningServerHandlerSpec extends Specification {
         ch.checkException()
     }
 
+    def 'decompression parts to full'(ChannelHandler compressor, CharSequence contentEncoding) {
```
dstepanov, line 364: Wouldn't it be better to write these tests using ordinary controller-client tests? That would be more understandable for people like me who are not that familiar with the netty and it might be better if you do some refactoring/Netty upgrade.

dstepanov (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/PipeliningServerHandlerSpec.groovy
yawkat: yea i can add some normal controller tests

yawkat: added one though it still uses netty for the client compression part

graemerocher (approved):

PR #10161, opened 2023-11-26
refactor: Apply SLF4J best practices

timtebeek:
Hi all, continuation here of earlier work on
 - https://github.com/micronaut-projects/micronaut-starter/pull/2002

I've run these three recipes together, to hopefully improve the logging output, with some small fixes thrown in too:
- [Parameterize SLF4J's logging statements](https://github.com/recipes/java/logging/slf4j/parameterizedlogging)
- [SLF4J logging statements should begin with constants](https://github.com/recipes/java/logging/slf4j/slf4jlogshouldbeconstant)
- [Enhances logging of exceptions by including the full stack trace in addition to the exception message](https://github.com/recipes/java/logging/slf4j/completeexceptionlogging)

When looking at other projects these fixes could be applied a few more times:
image

Use this link to re-run the recipe: https://app.moderne.io/recipes/builder/NBXKac4Ye?organizationId=TWljcm9uYXV0

### DIFF
--- a/discovery-core/src/main/java/io/micronaut/discovery/cloud/digitalocean/DigitalOceanMetadataResolver.java
+++ b/discovery-core/src/main/java/io/micronaut/discovery/cloud/digitalocean/DigitalOceanMetadataResolver.java
@@ -142,7 +142,7 @@ public class DigitalOceanMetadataResolver implements ComputeInstanceMetadataReso
             }
         } catch (IOException ioe) {
             if (LOG.isErrorEnabled()) {
-                LOG.error("Error connecting to{}reading instance metadata", configuration.getUrl(), ioe);
+                LOG.error("Error connecting to {} reading instance metadata", configuration.getUrl(), ioe);
             }
         }
 
--- a/discovery-core/src/main/java/io/micronaut/discovery/cloud/digitalocean/DigitalOceanMetadataResolver.java
+++ b/discovery-core/src/main/java/io/micronaut/discovery/cloud/digitalocean/DigitalOceanMetadataResolver.java
@@ -138,11 +138,11 @@ public class DigitalOceanMetadataResolver implements ComputeInstanceMetadataReso
             }
         } catch (MalformedURLException mue) {
             if (LOG.isErrorEnabled()) {
-                LOG.error("Digital Ocean metadataUrl value is invalid!: " + configuration.getUrl(), mue);
+                LOG.error("Digital Ocean metadataUrl value is invalid!: {}", configuration.getUrl(), mue);
             }
         } catch (IOException ioe) {
             if (LOG.isErrorEnabled()) {
-                LOG.error("Error connecting to" + configuration.getUrl() + "reading instance metadata", ioe);
+                LOG.error("Error connecting to{}reading instance metadata", configuration.getUrl(), ioe);
             }
         }
 
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -271,7 +271,7 @@ public final class RouteExecutor {
             logIgnoredException(cause);
         } else {
             if (LOG.isErrorEnabled()) {
-                LOG.error("Unexpected error occurred: " + cause.getMessage(), cause);
+                LOG.error("Unexpected error occurred: {}", cause.getMessage(), cause);
             }
         }
     }
@@ -283,7 +283,7 @@ public final class RouteExecutor {
 
     static void logIgnoredException(Throwable cause) {
         if (LOG.isDebugEnabled()) {
-            LOG.debug("Swallowed an IOException caused by client connectivity: " + cause.getMessage(), cause);
+            LOG.debug("Swallowed an IOException caused by client connectivity: {}", cause.getMessage(), cause);
         }
     }
 
@@ -337,7 +337,7 @@ public final class RouteExecutor {
 
         if (errorRoute != null) {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute);
+                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute, cause);
             }
             requestArgumentSatisfier.fulfillArgumentRequirementsBeforeFilters(errorRoute, httpRequest);
         }
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -337,7 +337,7 @@ public final class RouteExecutor {
 
         if (errorRoute != null) {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute, cause);
+                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute);
             }
             requestArgumentSatisfier.fulfillArgumentRequirementsBeforeFilters(errorRoute, httpRequest);
         }
--- a/http-server/src/main/java/io/micronaut/http/server/exceptions/UnsatisfiedRouteHandler.java
+++ b/http-server/src/main/java/io/micronaut/http/server/exceptions/UnsatisfiedRouteHandler.java
@@ -55,7 +55,7 @@ public class UnsatisfiedRouteHandler implements ExceptionHandler<UnsatisfiedRout
     @Override
     public HttpResponse handle(HttpRequest request, UnsatisfiedRouteException exception) {
         if (LOG.isTraceEnabled()) {
-            LOG.trace("{} (Bad Request): {}", request, exception.getMessage());
+            LOG.trace("{} (Bad Request): {}", request, exception.getMessage(), exception);
         }
         return responseProcessor.processResponse(ErrorContext.builder(request)
                 .cause(exception)
--- a/http-server/src/main/java/io/micronaut/http/server/exceptions/UnsatisfiedRouteHandler.java
+++ b/http-server/src/main/java/io/micronaut/http/server/exceptions/UnsatisfiedRouteHandler.java
@@ -55,7 +55,7 @@ public class UnsatisfiedRouteHandler implements ExceptionHandler<UnsatisfiedRout
     @Override
     public HttpResponse handle(HttpRequest request, UnsatisfiedRouteException exception) {
         if (LOG.isTraceEnabled()) {
-            LOG.trace("{} (Bad Request): {}", request, exception.getMessage(), exception);
+            LOG.trace("{} (Bad Request): {}", request, exception.getMessage());
         }
         return responseProcessor.processResponse(ErrorContext.builder(request)
                 .cause(exception)
--- a/inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
+++ b/inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
@@ -911,7 +911,7 @@ public abstract class AbstractInitializableBeanDefinition<T> extends AbstractBea
         MethodReference methodRef = methodInjection[methodIndex];
         Argument<?>[] methodArgumentTypes = methodRef.arguments == null ? Argument.ZERO_ARGUMENTS : methodRef.arguments;
         if (ClassUtils.REFLECTION_LOGGER.isDebugEnabled()) {
-            ClassUtils.REFLECTION_LOGGER.debug("Bean of type [" + getBeanType() + "] uses reflection to inject method: '" + methodRef.methodName + "'");
+            ClassUtils.REFLECTION_LOGGER.debug("Bean of type [{}] uses reflection to inject method: '{}'", getBeanType(), methodRef.methodName);
         }
         try {
             Method method = ReflectionUtils.getMethod(
@@ -945,7 +945,7 @@ public abstract class AbstractInitializableBeanDefinition<T> extends AbstractBea
         FieldReference fieldRef = fieldInjection[index];
         try {
             if (ClassUtils.REFLECTION_LOGGER.isDebugEnabled()) {
-                ClassUtils.REFLECTION_LOGGER.debug("Bean of type [" + getBeanType() + "] uses reflection to inject field: '" + fieldRef.argument.getName() + "'");
+                ClassUtils.REFLECTION_LOGGER.debug("Bean of type [{}] uses reflection to inject field: '{}'", getBeanType(), fieldRef.argument.getName());
             }
             Field field = ReflectionUtils.getRequiredField(fieldRef.declaringType, fieldRef.argument.getName());
             field.setAccessible(true);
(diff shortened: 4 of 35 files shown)


discovery-core/src/main/java/io/micronaut/discovery/cloud/digitalocean/DigitalOceanMetadataResolver.java
```diff
@@ -138,11 +138,11 @@ public Optional<ComputeInstanceMetadata> resolve(Environment environment) {
             }
         } catch (MalformedURLException mue) {
             if (LOG.isErrorEnabled()) {
-                LOG.error("Digital Ocean metadataUrl value is invalid!: " + configuration.getUrl(), mue);
+                LOG.error("Digital Ocean metadataUrl value is invalid!: {}", configuration.getUrl(), mue);
             }
         } catch (IOException ioe) {
             if (LOG.isErrorEnabled()) {
-                LOG.error("Error connecting to" + configuration.getUrl() + "reading instance metadata", ioe);
+                LOG.error("Error connecting to{}reading instance metadata", configuration.getUrl(), ioe);
```
timtebeek: In this case the recipe run brought to light an issue with how the message is composed.
```suggestion
                LOG.error("Error connecting to {} reading instance metadata", configuration.getUrl(), ioe);
```

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -337,7 +337,7 @@ RouteMatch<?> findErrorRoute(Throwable cause,
 
         if (errorRoute != null) {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute);
+                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute, cause);
```
timtebeek: Here the recipe had added the `cause`, but I'm thinking that's maybe not desired here given that we're handling the error route.
```suggestion
                LOG.debug("Found matching exception handler for exception [{}]: {}", cause.getMessage(), errorRoute);
```

http-server/src/main/java/io/micronaut/http/server/exceptions/UnsatisfiedRouteHandler.java
```diff
@@ -55,7 +55,7 @@ public UnsatisfiedRouteHandler(ErrorResponseProcessor<?> responseProcessor) {
     @Override
     public HttpResponse handle(HttpRequest request, UnsatisfiedRouteException exception) {
         if (LOG.isTraceEnabled()) {
-            LOG.trace("{} (Bad Request): {}", request, exception.getMessage());
+            LOG.trace("{} (Bad Request): {}", request, exception.getMessage(), exception);
```
timtebeek: For `trace` level it might make sense to keep this added exception stacktrace, but I'll leave that up to the review. Pointing it out such that it's not missed.

inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
```diff
@@ -2170,7 +2170,7 @@ private <K> K resolveBean(
             }
         } catch (DisabledBeanException e) {
             if (ConditionLog.LOG.isDebugEnabled()) {
-                ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", argument.getTypeName(), e.getMessage());
+                ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", argument.getTypeName(), e.getMessage(), e);
```
timtebeek: Since we're rethrowing we likely don't want the stacktrace logged here.
```suggestion
                ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", argument.getTypeName(), e.getMessage());
```

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1797,7 +1797,7 @@ public <T> Optional<T> findBean(@Nullable BeanResolutionContext resolutionContex
             }
         } catch (DisabledBeanException e) {
             if (AbstractBeanContextConditional.ConditionLog.LOG.isDebugEnabled()) {
-                AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", beanType.getSimpleName(), e.getMessage());
+                AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", beanType.getSimpleName(), e.getMessage(), e);
```
timtebeek: Stacktrace unlikely to add value here.
```suggestion
                AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", beanType.getSimpleName(), e.getMessage());
```

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1974,7 +1974,7 @@ protected void initializeContext(
                     initializeEagerBean(eagerInitDefinition);
                 } catch (DisabledBeanException e) {
                     if (AbstractBeanContextConditional.ConditionLog.LOG.isDebugEnabled()) {
-                        AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", eagerInitDefinition.getBeanType().getSimpleName(), e.getMessage());
+                        AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", eagerInitDefinition.getBeanType().getSimpleName(), e.getMessage(), e);
```
timtebeek: Stacktrace unlikely to add value here.
```suggestion
                        AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", eagerInitDefinition.getBeanType().getSimpleName(), e.getMessage());
```

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2448,7 +2448,7 @@ protected void processParallelBeans(List<BeanDefinitionProducer> parallelBeans)
                             loadEagerBeans(producer, parallelDefinitions);
                         } catch (Throwable e) {
                             BeanDefinitionReference<Object> beanDefinitionReference = producer.getReference();
-                            LOG.error("Parallel Bean definition [" + beanDefinitionReference.getName() + MSG_COULD_NOT_BE_LOADED + e.getMessage(), e);
+                            LOG.error("Parallel Bean definition [{}{}{}", beanDefinitionReference.getName(), MSG_COULD_NOT_BE_LOADED, e.getMessage(), e);
```
timtebeek: Should the log message here contain a closing bracket?
```suggestion
                            LOG.error("Parallel Bean definition [{}{}{}]", beanDefinitionReference.getName(), MSG_COULD_NOT_BE_LOADED, e.getMessage(), e);
```

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2462,7 +2462,7 @@ protected void processParallelBeans(List<BeanDefinitionProducer> parallelBeans)
                         try {
                             initializeEagerBean(beanDefinition);
                         } catch (Throwable e) {
-                            LOG.error("Parallel Bean definition [" + beanDefinition.getName() + MSG_COULD_NOT_BE_LOADED + e.getMessage(), e);
+                            LOG.error("Parallel Bean definition [{}{}{}", beanDefinition.getName(), MSG_COULD_NOT_BE_LOADED, e.getMessage(), e);
```
timtebeek: Should the log message here contain a closing bracket?
```suggestion
                            LOG.error("Parallel Bean definition [{}{}{}]", beanDefinition.getName(), MSG_COULD_NOT_BE_LOADED, e.getMessage(), e);
```

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -3509,7 +3509,7 @@ private <T> void addCandidateToList(@Nullable BeanResolutionContext resolutionCo
             }
         } catch (DisabledBeanException e) {
             if (AbstractBeanContextConditional.ConditionLog.LOG.isDebugEnabled()) {
-                AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", beanType.getTypeName(), e.getMessage());
+                AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", beanType.getTypeName(), e.getMessage(), e);
```
timtebeek: Stacktrace unlikely to add value here.
```suggestion
                AbstractBeanContextConditional.ConditionLog.LOG.debug("Bean of type [{}] disabled for reason: {}", beanType.getTypeName(), e.getMessage());
```

retry/src/main/java/io/micronaut/retry/intercept/CircuitBreakerRetry.java
```diff
@@ -93,7 +93,7 @@ public void close(@Nullable Throwable exception) {
     public void open() {
         if (currentState() == CircuitState.OPEN && lastError != null) {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Rethrowing existing exception for Open Circuit [{}]: {}", method, lastError.getMessage());
+                LOG.debug("Rethrowing existing exception for Open Circuit [{}]: {}", method, lastError.getMessage(), lastError);
```
timtebeek: Rethrow again likely does not require a stacktrace at this point.
```suggestion
                LOG.debug("Rethrowing existing exception for Open Circuit [{}]: {}", method, lastError.getMessage());
```

retry/src/main/java/io/micronaut/retry/intercept/CircuitBreakerRetry.java
```diff
@@ -188,7 +188,7 @@ private CircuitState openCircuit(Throwable cause) {
             throw new IllegalArgumentException("Exception cause cannot be null");
         }
         if (LOG.isDebugEnabled()) {
-            LOG.debug("Opening Circuit Breaker [{}] due to error: {}", method, cause.getMessage());
+            LOG.debug("Opening Circuit Breaker [{}] due to error: {}", method, cause.getMessage(), cause);
```
timtebeek: In retry stacktraces add noise.
```suggestion
            LOG.debug("Opening Circuit Breaker [{}] due to error: {}", method, cause.getMessage());
```

retry/src/main/java/io/micronaut/retry/intercept/DefaultRetryInterceptor.java
```diff
@@ -244,11 +244,11 @@ private Object retrySync(MethodInvocationContext<Object, Object> context, Mutabl
                             try {
                                 eventPublisher.publishEvent(new RetryEvent(context, retryState, e));
                             } catch (Exception e1) {
-                                LOG.error("Error occurred publishing RetryEvent: " + e1.getMessage(), e1);
+                                LOG.error("Error occurred publishing RetryEvent: {}", e1.getMessage(), e1);
                             }
                         }
                         if (LOG.isDebugEnabled()) {
-                            LOG.debug("Retrying execution for method [{}] after delay of {}ms for exception: {}", context, delayMillis, e.getMessage());
+                            LOG.debug("Retrying execution for method [{}] after delay of {}ms for exception: {}", context, delayMillis, e.getMessage(), e);
```
timtebeek: In retry stacktraces add noise.
```suggestion
                            LOG.debug("Retrying execution for method [{}] after delay of {}ms for exception: {}", context, delayMillis, e.getMessage());
```

http-server/src/main/java/io/micronaut/http/server/exceptions/UnsatisfiedRouteHandler.java
timtebeek:
```suggestion
            LOG.trace("{} (Bad Request): {}", request, exception.getMessage());
```

sdelamo (approved):

altro3:
It is interesting that this change is not entirely correct: it makes sense to use placeholders only if you do not have a preliminary check for isDebugEnabled / isInfoEnabled / ... . In general, they were invented precisely for this purpose, to simplify the code, but at the same time not to carry out unnecessary calculations. That is, when you call a logger with some calculations, and then it turns out that logging is not happening due to another level of logging, but you have already spent time calculating the parameters.

In general, what I want to say is that the current changes will result in logging execution being slower than before, even if only slightly.

But since the slowdown is really insignificant, they can be done.

An example of such changes:
изображение

timtebeek:
> It is interesting that this change is not entirely correct: it makes sense to use placeholders only if you do not have a preliminary check for isDebugEnabled / isInfoEnabled / ... . In general, they were invented precisely for this purpose, to simplify the code, but at the same time not to carry out unnecessary calculations. That is, when you call a logger with some calculations, and then it turns out that logging is not happening due to another level of logging, but you have already spent time calculating the parameters.
> 
> In general, what I want to say is that the current changes will result in logging execution being slower than before, even if only slightly.

Interesting perspective, thanks! So if I understand you correctly you're saying not to use parameterized logging when contained within a logging guard block? As then the arguments are guaranteed to be used if executed, and String concatenation is faster than the log string formatting.

Scrolling through the pull request changes again it seems the project is pretty consistent in using guard blocks. An alternative point can also be made that those guard blocks are perhaps not as needed when using parameterized logging, and could clean up the code a bit. Not saying we should do this, but it's an interesting thought.

Looking at for instance the following snippet, in the before situation the guard block prevented unnecessary string concatenation, whereas when using parameterized logging the `if ` could also be removed, such that you only interact with `LOG` once. I have no measurements on performance there, but the code reads cleaner without the guarding conditional.
```diff
         if (LOG.isDebugEnabled()) {
-            LOG.debug("Swallowed an IOException caused by client connectivity: " + cause.getMessage(), cause);
+            LOG.debug("Swallowed an IOException caused by client connectivity: {}", cause.getMessage(), cause);
         }
```

Curious to hear your thoughts; I'd now be inclined to write a recipe to clear out unnecessary guard blocks, even if it's not applied here. :)

altro3:
Yes, that's right, your example is correct. Of course, string concatenation is faster than string parsing and replacing placeholders.

Yes, if you switched to placeholders, then there is no need to check the logging level before doing so. The check will still occur in the Slf4J implementation. for example, in the ch.qos.logback.classic.Logger class. But even here everything is not entirely clear: in the logger, the level is checked not immediately, but after several quick operations.

Therefore, the following conclusions can be drawn:

1. The most effective method of logging in terms of speed (But at the same time, the most inconvenient method, which existed BEFORE the creation of Slf4J and was proposed in Log4J):
```java
if (LOG.isDebugEnabled()) {
    LOG.debug("Swallowed an IOException caused by client connectivity: " + cause.getMessage(), cause);
}
```
2. The most convenient way, but with several additional internal operations:
```java
LOG.debug("Swallowed an IOException caused by client connectivity: {}", cause.getMessage(), cause);
```

As always, we have a standard conflict in programming: convenience or performance :-)

timtebeek:
Agreed! And then this PR somewhat counter-intuitively uses both guards and parameterized logging, with the performance penalty you've indicated. So then I'll leave up to all of you if you still want to merge this for the readability improvements and minor fixes, or revisit at a later point in time.

As perhaps known these code changes were mostly automated (apart from the fixes and reverts), so no harm in rejecting a PR if you're not happy with it. Merely trying to help, and learn & improve in the process. :)

altro3:
The guys will decide that :). There is nothing wrong with these changes. There is little performance loss. I would just then remove the checks at the logging level, because... they will happen inside the method anyway. Those. in the case of logging, it seems to me that it is quite possible to sacrifice performance for the sake of convenience. But it's not up to me to decide ;-)

In any case, I think it’s worth adding to these changes and removing checks at the logging level in cases where there are no calculations

sdelamo:
what do you think? @dstepanov  I was about to merge this.

dstepanov (approved):

Issue #10080, opened 2023-08-03
ContextualConnection$Intercepted doesn't implement all methods


marcinfigiel:
### Expected Behavior

When trying to insert a row into the PostgreSQL DB via JDBI SqlObject method annotated with `@GetGeneratedKeys` the generated ID should be returned, as it was on Micronaut 3.x.

### Actual Behaviour

The following error is thrown:
`
Unexpected error occurred: Receiver class io.micronaut.data.connection.jdbc.advice.ContextualConnection$Intercepted does not define or inherit an implementation of the resolved method 'abstract java.sql.PreparedStatement prepareStatement(java.lang.String, int)' of interface java.sql.Connection.
`
from `org.jdbi.v3.core.statement.DefaultStatementBuilder.create(DefaultStatementBuilder.java:53)`

### Environment Information

- Operating system: pop!_OS 22.04 LTS (kernel 6.2.6)
- JDK version: OpenJDK Runtime Environment Temurin-17.0.8+7 (build 17.0.8+7)

### Example Application

https://github.com/marcinfigiel/demo-jdbc-error

### Version

4.0.2

dstepanov:
Can you extend your test to use test containers / test resources? Thanks

oddaolse:
I can just add that I face the same issue.

frodeo:
@graemerocher @dstepanov Any idea when a fix for this will be available? This is preventing us from upgrading to Micronaut 4 now.

marcinfigiel:
@frodeo There is a workaround for this issue - it works fine if you explicitly specify the PK columns in the `value` of the `@GetGeneratedKeys` annotation, e.g. `@GetGeneratedKeys("id")`

@dstepanov I'm sorry, I haven't had time lately. I will try to do it, but cannot give you any ETA at the moment :( 
@oddaolse if you can spare some time I will gladly accept a PR :)

frodeo:
We are using JDBC and java.sql.Statement.getGeneratedKeys(), and we are still getting the same error. There we are not able to specify the PK (there is no getGeneratedKeys("id")), but we have specified the PK in the schema. Is there any workaround for plain JDBC?

radovanradic:
> Can you extend your test to use test containers / test resources? Thanks

Here is an example with test resources reproducing the error
[mn-demo-jdbi-error.zip](https://github.com/micronaut-projects/micronaut-core/files/13472471/mn-demo-jdbi-error.zip)

PR #10163, opened 2023-11-27
ContextualConnection intercepted missing method repro

radovanradic:
This is the dumb and simple repro for https://github.com/micronaut-projects/micronaut-core/issues/10080


inject-java-test/src/test/groovy/io/micronaut/aop/MyInterface.java
```diff
@@ -0,0 +1,16 @@
...
+interface MyInterface {
+
+    String process(String str);
+
+    // This method will not be added to the generated intercepted class for some reason.
+    String process(String str, int intParam);
+
+    // This one will
+    String process2(String str, int intParam);
+
```
radovanradic: Ok, created repro without `Connection` as this reproduces with any interface. Seems like method `process(String, int)` clashes with `process(String, int[])` and the one with array gets created and one with just int gets skipped for some reason.

dstepanov: How is it that your test is passing?

radovanradic: Because I wrote it maybe the inverse way, wanting to prove that AbstractError is thrown (this means that interface does not implement expected method)
```
when:"Intercepted MyInterface does not have method implemented"
        myInterface.process("abc", 1)
        then:
        AbstractMethodError ex = thrown()
        ex
```
this should not be thrown if method is generated properly in the interceptor

radovanradic: Basically the same happening with Connection `PreparedStatement prepareStatement(String sql, int autoGeneratedKeys)`

PR #10167, opened 2023-11-28
Refactor filtering to improve performance

dstepanov:
- Filter runner refactored to enable imperative methods invocation in a loop without multiple lambdas
- Added more unwrapping flow checks for filters and route execution
- Removed a need to have terminal filters
- Changes modified some of the generics and signatures. Filters should work with non-mutable request/response with wildcard body type


graemerocher (approved):

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
```diff
@@ -371,10 +372,18 @@ protected <I, R extends io.micronaut.http.HttpResponse<?>> Publisher<R> applyFil
...
+            try {
+                try (PropagatedContext.Scope ignore = propagatedContext.propagate()) {
+                    return ReactiveExecutionFlow.fromPublisher((Publisher<HttpResponse<?>>) responsePublisher);
+                }
+            } catch (Throwable e) {
+                return ExecutionFlow.error(e);
+            }
+        });
+        return (Publisher<R>) Mono.from(ReactiveExecutionFlow.fromFlow(runner.run(request)).toPublisher())
+            .map(x -> x);
```
yawkat: what's this map for

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
```diff
@@ -65,29 +66,18 @@
...
     private final RouteExecutor routeExecutor;
-    private HttpRequest<?> request;
     private boolean multipartEnabled = true;
 
     /**
      * @param routeExecutor The route executor to use for route resolution
-     * @param request       The request to process
      */
-    protected RequestLifecycle(RouteExecutor routeExecutor, HttpRequest<?> request) {
+    protected RequestLifecycle(RouteExecutor routeExecutor) {
```
yawkat, line 79: i see it's passed around everywhere now, but what's the point of this change?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -113,11 +115,12 @@ protected ExecutionFlow<? extends HttpResponse<?>> processResponse(HttpRequest<?
      * @param request           The current request
      * @param failure           The failure
      * @param propagatedContext The propagated context
-     * @return A flow that will be passed on to the next filter
+     * @return A flow that will be passed on to the next filter, or null if exception is not remapped
```
yawkat, line 118: do we need this change, it's overridden for the server anyway?

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -127,67 +130,152 @@ protected ExecutionFlow<? extends HttpResponse<?>> processFailure(HttpRequest<?>
...
-                    return processFailure(context.request(), throwable, context.propagatedContext()).map(context::withResponse)
-                        .onErrorResume(throwable2 -> {
-                            // Exception filtering scenario of the http client
-                            return filterResponse(context, iterator, throwable2).map(context::withResponse);
-                        });
+            if (!filter.isFiltersRequest()) {
+                continue;
+            }
+            // At-least one request filter
+            ExecutionFlow<FilterContext> flow;
```
yawkat, line 164: maybe at some point we should just go back to the coroutine approach from #8422...

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
dstepanov: reverted

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
dstepanov: With this change, it's clear how the request is processed, because it can be modified it's better to pass it with the flow and avoid mistakes.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: The server is overriding it, but the client just lets it fail, to avoid unwrapping the flow, I just changed it to be nullable.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
dstepanov: I don't remember how it is different with coroutines.

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
yawkat: brings back the param list inflation RequestLifecycle was made to avoid though...

dstepanov: I feel the code flow is more transparent and more accessible to understand this way.

yawkat: fair enough

yawkat (approved):

PR #10172, opened 2023-11-29
Allow to disable filters based on the request

dstepanov:
Based on https://github.com/micronaut-projects/micronaut-core/pull/10167

PR #10177, opened 2023-11-30
Optimize routing

dstepanov:
Based on https://github.com/micronaut-projects/micronaut-core/pull/10172

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
@@ -179,7 +179,7 @@ public class RequestLifecycle {
     }
 
     /**
-     * Handle an error in this request. Also runs filters for the error handling.
+     * Handle an error in this request. It also runs filters for the error handling.
      *
      * @param request   The request
      * @param throwable The error
--- a/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
@@ -285,21 +285,25 @@ public class RequestLifecycle {
      * @return Execution flow that completes after the all the filters and the downstream flow
      */
     protected final ExecutionFlow<HttpResponse<?>> runWithFilters(HttpRequest<?> request, BiFunction<HttpRequest<?>, PropagatedContext, ExecutionFlow<HttpResponse<?>>> responseProvider) {
-        List<GenericHttpFilter> httpFilters = routeExecutor.router.findFilters(request);
-        FilterRunner filterRunner = new FilterRunner(httpFilters, responseProvider) {
-            @Override
-            protected ExecutionFlow<HttpResponse<?>> processResponse(HttpRequest<?> request, HttpResponse<?> response, PropagatedContext propagatedContext) {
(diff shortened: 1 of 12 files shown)


http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
```diff
@@ -285,21 +285,25 @@ private ExecutionFlow<HttpResponse<?>> onErrorNoFilter(HttpRequest<?> request, T
...
-            }
-        };
-        return filterRunner.run(request);
+                @Override
+                protected ExecutionFlow<HttpResponse<?>> processFailure(HttpRequest<?> request, Throwable failure, PropagatedContext propagatedContext) {
+                    return onErrorNoFilter(request, failure, propagatedContext);
+                }
+            };
+            return filterRunner.run(request);
+        } catch (Throwable e) {
```
yawkat, line 304: what can fail here? getUri?

router/src/main/java/io/micronaut/web/router/DefaultRouteInfo.java
```diff
@@ -141,6 +143,8 @@ public DefaultRouteInfo(AnnotationMetadata annotationMetadata,
             this.consumesMediaTypes = consumesMediaTypes;
             consumesMediaTypesContainsAll = this.consumesMediaTypes.contains(MediaType.ALL_TYPE);
         }
+        this.imperative = isVoid || !suspended && !reactive && !async && !returnType.getType().equals(Object.class)
+            && (returnType.getType().getPackageName().startsWith("java.") || BeanIntrospector.SHARED.findIntrospection(returnType.getType()).isPresent());
```
yawkat, line 147: this is really broad, what return types are we excluding here?

yawkat (approved):

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
dstepanov: Yes

router/src/main/java/io/micronaut/web/router/DefaultRouteInfo.java
dstepanov: Well right now we always do instance-of-reactive if not detected as reactive/async before, we can definitely say it's imperative if the type is not reactive/async and part of java package + introspected classes as well

yawkat: ah i see, this is basically the check from isConvertibleToPublisher.

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
yawkat: it's really ugly that that can fail all over the place now

dstepanov: It will now fail during the filter check and propagate the errors. After the URI change, it will fail before the route execution and propagate the error in filters.

yawkat: well if the filters dont need the route it can now fail in a controller too. it can fail anywhere

dstepanov: Only if you disable the validation the other PR

PR #10178, opened 2023-11-30
Allow to skip parsing the request's url into `URI`

dstepanov:
Based on https://github.com/micronaut-projects/micronaut-core/pull/10177

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
@@ -80,19 +80,6 @@ public class NettyClientHttpRequest<B> implements MutableHttpRequest<B>, NettyHt
     private ConversionService conversionService = ConversionService.SHARED;
 
     /**
-     * This constructor is actually required for the case of non-standard http methods.
-     *
-     * @param httpMethod     The http method. CUSTOM value is used for non-standard
-     * @param uri            The uri
-     * @param httpMethodName Method name. Is the same as httpMethod.name() value for standard http methods.
-     */
-    NettyClientHttpRequest(HttpMethod httpMethod, URI uri, String httpMethodName) {
-        this.httpMethod = httpMethod;
-        this.uri = uri;
-        this.httpMethodName = httpMethodName;
-    }
-
-    /**
      * @param httpMethod The Http method
      * @param uri        The URI
      */
@@ -104,11 +91,13 @@ public class NettyClientHttpRequest<B> implements MutableHttpRequest<B>, NettyHt
      * This constructor is actually required for the case of non-standard http methods.
      *
      * @param httpMethod     The http method. CUSTOM value is used for non-standard
-     * @param uri            The uri
+     * @param url            The uri
      * @param httpMethodName Method name. Is the same as httpMethod.name() value for standard http methods.
      */
-    NettyClientHttpRequest(HttpMethod httpMethod, String uri, String httpMethodName) {
-        this(httpMethod, URI.create(uri), httpMethodName);
+    NettyClientHttpRequest(HttpMethod httpMethod, String url, String httpMethodName) {
+        this.httpMethod = httpMethod;
+        this.uri = URI.create(url);
+        this.httpMethodName = httpMethodName;
     }
 
     @Override
--- a/http-netty/src/main/java/io/micronaut/http/netty/AbstractNettyHttpRequest.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/AbstractNettyHttpRequest.java
@@ -22,7 +22,6 @@ import io.micronaut.core.convert.ConversionService;
 import io.micronaut.http.HttpMethod;
 import io.micronaut.http.HttpParameters;
 import io.micronaut.http.HttpRequest;
-import io.micronaut.http.MediaType;
 import io.micronaut.http.netty.stream.DefaultStreamedHttpRequest;
 import io.micronaut.http.netty.stream.StreamedHttpRequest;
 import io.netty.handler.codec.http.DefaultFullHttpRequest;
@@ -34,9 +33,6 @@ import io.netty.util.DefaultAttributeMap;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
-import java.util.Collection;
-import java.util.Locale;
-import java.util.Optional;
 
 /**
  * Abstract implementation of {@link HttpRequest} for Netty.
@@ -51,15 +47,13 @@ public abstract class AbstractNettyHttpRequest<B> extends DefaultAttributeMap im
     protected final io.netty.handler.codec.http.HttpRequest nettyRequest;
     protected final ConversionService conversionService;
     protected final HttpMethod httpMethod;
-    protected final URI uri;
+    protected final String url;
     protected final String httpMethodName;
 
+    private URI uri;
     private NettyHttpParameters httpParameters;
-    private Optional<MediaType> mediaType;
     private Charset charset;
-    private Optional<Locale> locale;
     private String path;
-    private Collection<MediaType> accept;
 
     /**
      * @param nettyRequest      The Http netty request
@@ -68,22 +62,7 @@ public abstract class AbstractNettyHttpRequest<B> extends DefaultAttributeMap im
     public AbstractNettyHttpRequest(io.netty.handler.codec.http.HttpRequest nettyRequest, ConversionService conversionService) {
         this.nettyRequest = nettyRequest;
         this.conversionService = conversionService;
-        URI fullUri = URI.create(nettyRequest.uri());
-        if (fullUri.getAuthority() != null || fullUri.getScheme() != null) {
-            // https://example.com/foo -> /foo
-            try {
-                fullUri = new URI(
-                        null, // scheme
-                        null, // authority
-                        fullUri.getPath(),
-                        fullUri.getQuery(),
-                        fullUri.getFragment()
-                );
-            } catch (URISyntaxException e) {
-                throw new IllegalArgumentException(e);
-            }
-        }
-        this.uri = fullUri;
+        this.url = nettyRequest.uri();
         this.httpMethodName = nettyRequest.method().name();
         this.httpMethod = HttpMethod.parse(httpMethodName);
     }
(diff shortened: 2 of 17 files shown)


graemerocher (approved):

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
```diff
@@ -74,24 +74,12 @@ public class NettyClientHttpRequest<B> implements MutableHttpRequest<B>, NettyHt
     private final HttpMethod httpMethod;
     private final String httpMethodName;
     private final Map<String, String> cookies = new LinkedHashMap<>(1);
+    private final String url;
```
yawkat: please call it something else. maybe unvalidatedUri.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -200,34 +197,6 @@ public NettyHttpRequest(io.netty.handler.codec.http.HttpRequest nettyRequest,
         this.channelHandlerContext = ctx;
         this.headers = new NettyHttpHeaders(nettyRequest.headers(), conversionService);
         this.body = ByteBody.of(nettyRequest);
-        this.contentLength = headers.contentLength().orElse(-1);
```
yawkat, line 199: are these three actually worthwhile? i would guess they are always accessed anyway.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
```diff
@@ -59,8 +59,6 @@ final class NettyRequestLifecycle extends RequestLifecycle {
         super(rib.routeExecutor);
         this.rib = rib;
         this.outboundAccess = outboundAccess;
-
-        multipartEnabled(rib.multipartEnabled);
```
yawkat, line 63: what is this replaced by

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/MalformedUriDisabledValidationSpec.groovy
```diff
@@ -0,0 +1,86 @@
...
+
+    @Shared @AutoCleanup EmbeddedServer embeddedServer = ApplicationContext.run(EmbeddedServer, [
+            'spec.name': 'MalformedUriDisabledValidationSpec',
+            'micronaut.server.validate-url': 'false'
+    ])
+    @Shared @AutoCleanup HttpClient client = embeddedServer.applicationContext.createBean(HttpClient, embeddedServer.getURL())
+
+    void "test malformed URI exceptions"() {
+        when:
+        def result = new URL("$embeddedServer.URL/malformed/[]").text
```
yawkat, line 37: why is this URL and not URI?

http/src/main/java/io/micronaut/http/HttpRequest.java
```diff
@@ -68,6 +68,16 @@ public interface HttpRequest<B> extends HttpMessage<B> {
      */
     @NonNull URI getUri();
 
+    /**
+     * Returns the supplied url, the value might not be validated.
+     * @return The full request url
+     * @since 4.3.0
+     */
+    @NonNull
+    default String getUrl() {
```
yawkat: different name

yawkat (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
dstepanov: Configuration is extracted from the route executor in `RequestLifecycle`.

yawkat: did you test this with servlet? i think this check was only present in RoutingInBoundHandler which is why i only added it to NettyRequestLifecycle and not the RequestLifecycle

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
dstepanov: After thinking about it more, I will move the caching to headers and avoid nullifying it. Keep it this way for now. Routing should avoid checking the headers in cases when it will always match for any content-type/accepts.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/MalformedUriDisabledValidationSpec.groovy
dstepanov: What do you mean? I just copied the existing test with disabled validation.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: i think always matching is rare, it's not the default

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyRequestLifecycle.java
dstepanov: The check is coming from the shared config so I expect it's valid in the servlet

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/MalformedUriDisabledValidationSpec.groovy
yawkat: ah ok. then the old test is at fault :D

dstepanov:
@yawkat I'm trying to remove the new `getUrl` method to avoid creating an invalidated method, but if I revert `HateoasErrorResponseProcessor` I get a memory leak detected in `FuzzyInputSpec`. Can you please check? I think we have a leak when the error is thrown from `ErrorResponseProcessor`

yawkat:
ok i'll do my best, still on the train though so it might take a while :)

dstepanov:
Hopefully, it's not a Czech train and Wi-Fi :D

yawkat:
worse. german train wifi

dstepanov:
I added a check to prevent an exception from propagating and removed all `getUrl` methods.

http/src/main/java/io/micronaut/http/HttpRequest.java
dstepanov: Removed

yawkat:
the leak happens because RequestHandler.accept throws an exception, which it shouldnt. but PipeliningServerHandler needs to handle that better. so i will make a note to improve that

http-netty/src/main/java/io/micronaut/http/netty/AbstractNettyHttpRequest.java
```diff
@@ -46,14 +46,16 @@
...
  */
 @Internal
+@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 public abstract class AbstractNettyHttpRequest<B> extends DefaultAttributeMap implements HttpRequest<B>, NettyHttpRequestBuilder {
 
     protected final io.netty.handler.codec.http.HttpRequest nettyRequest;
     protected final ConversionService conversionService;
     protected final HttpMethod httpMethod;
-    protected final URI uri;
+    protected final String url;
```
yawkat: pls still rename the field though

yawkat:
what did the force push change, just rebase? i dont want to review everything again

dstepanov:
Had to rewrite it a bit, moved some of the cached fields into the headers implementation.

http-netty/src/main/java/io/micronaut/http/netty/AbstractNettyHttpRequest.java
yawkat: @dstepanov

dstepanov: Done

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
```diff
@@ -141,16 +156,34 @@ public Optional<String> findFirst(CharSequence name) {
...
     @Override
     public MutableHeaders set(CharSequence header, CharSequence value) {
         validateHeader(header, value);
         nettyHeaders.set(header, value);
+        onModify(header);
         return this;
     }
 
+    private void onModify(CharSequence header) {
+        if (contentType != null && header.equals(HttpHeaders.CONTENT_TYPE)) {
```
yawkat: can't use equals here, it's CharSequence. Is this called on the hot path? maybe it's fine to just unset all the caches and be done with it

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -721,7 +668,7 @@ private static ByteBuffer<ByteBuf> toByteBuffer(ImmediateByteBody immediateByteB
      */
     private final class NettyMutableHttpRequest implements MutableHttpRequest<T>, NettyHttpRequestBuilder {
 
-        private URI uri = NettyHttpRequest.this.uri;
+        private URI uri;
```
yawkat, line 671: unused?

dstepanov: This is to override

yawkat: i see

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
dstepanov: Removed, looks like the the modifications are mostly for the response anyway

yawkat (approved):

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
```diff
@@ -91,48 +91,68 @@ protected final void multipartEnabled(boolean multipartEnabled) {
...
+            if (routeMatch.getRouteInfo().isWebSocketRoute()) {
+                return onStatusError(
+                    request,
+                    HttpResponse.status(HttpStatus.BAD_REQUEST),
+                    "Not a WebSocket request");
+            }
 
-        return runWithFilters(request, (filteredRequest, propagatedContext) -> executeRoute(filteredRequest, propagatedContext, routeMatch));
+            return runWithFilters(request, (filteredRequest, propagatedContext) -> {
+                if (validateUrl) {
```
yawkat, line 144: @dstepanov why does this happen inside the filter execution? i think it should go before. if it's placed here, it's possible a filter will see and work with an invalid path, which might be a security issue.

dstepanov: We can move it somewere else, the filter cannot see the invalid path because it will throw an exception

PR #10194, opened 2023-12-04
Add cache for DateTimeFormatter for `@Format` annotations

altro3:
Minor performance improvement and memory reducer for `TimeConverterRegistrar`. Don't need to recreate custom DateTimeFormatter for every request / response, we can cache it.

This simple test show timings with cache and without cache:
```java
@MicronautTest
@Property(name = "spec.name", value = "DateFormatterTest")
@Property(name = "micronaut.http.client.read-timeout", value = "30s")
class DateFormatterTest {

    private static final Logger log = LoggerFactory.getLogger(DateFormatterTest.class);

    @Inject
    MyGetClient myGetClient;

    @Test
    void test() {

        var date = LocalDate.now();
        var size = 1_000_000;
        var results = new ArrayList<Integer>(size);

        for (var i = 0; i < size; i++) {
            var startTime = System.nanoTime();
            myGetClient.formatDateTime(date);
            results.add((int) (System.nanoTime() - startTime));
        }

        results.sort(Comparator.naturalOrder());

        var sum = 0;

        // remove 5% of the best timings and 5% of the worst timings
        for (var i = size / 20; i < size - (size / 20); i++) {
            sum += (results.get(i) / 1000);
        }

        log.info("!!!!!!!!!!!!!! time = {}", Duration.ofNanos(sum * 1000L));
    }

    @Requires(property = "spec.name", value = "DateFormatterTest")
    @Client("/get")
    interface MyGetClient {

        @Get("/dateTime/{myDate}")
        String formatDateTime(@Format("yyyy-MM-dd") LocalDate myDate);
    }

    @Requires(property = "spec.name", value = "DateFormatterTest")
    @Controller("/get")
    static class GetController {

        @Get("/dateTimeQuery")
        String formatDateTimeQuery(@QueryValue @Format("yyyy-MM-dd") LocalDate myDate) {
            return myDate.toString();
        }
    }
}
```

Test results:
```
Requests count: 100_000
With cache:  PT21.427236S
Without cache: PT21.560763S

Requests count: 1_000_000
With cache: PT3M21.53591S
Without cache: PT3M24.615995S

```


dstepanov:
Please target 4.3

gradle/libs.versions.toml
```diff
@@ -30,14 +30,14 @@ jakarta-el-impl = "5.0.0-M1"
 jcache = "1.1.1"
 junit5 = "5.9.3"
 junit-platform="1.9.3"
-logback = "1.4.13"
```
dstepanov, line 33: Please cleanup updates

dstepanov (changes requested):

gradle/libs.versions.toml
altro3: done

altro3:
@dstepanov continue here: https://github.com/micronaut-projects/micronaut-core/pull/10198

Issue #10145, opened 2023-11-21
KAPT: kaptKotlin Performance regression in Micronaut Core 4.2.0


Spikhalskiy:
### Actual Behaviour

`kaptKotlin` gradle building step with Micronaut Platform 4.2.0 takes 5x the time that it takes with Platform 4.1.6 on the same source code.

CPU Sampling profile of kaptKotlin task for Micronaut 4.1.6: 
[micronaut-4.1.nps.zip](https://github.com/micronaut-projects/micronaut-core/files/13431471/micronaut-4.1.nps.zip)

CPU Sampling profile of kaptKotlin task for Micronaut 4.2.0 on the same sources: 
[micronaut-4.2.nps.zip](https://github.com/micronaut-projects/micronaut-core/files/13431475/micronaut-4.2.nps.zip)

Potential culprit:
![Screenshot 2023-11-21 at 5 00 37 PM](https://github.com/micronaut-projects/micronaut-core/assets/532108/94bb0da7-57ca-4a29-bf81-f6d5ec82bafc)

Change in hot methods:

![Screenshot 2023-11-21 at 5 38 46 PM](https://github.com/micronaut-projects/micronaut-core/assets/532108/58778c4d-4289-483f-a072-9d859590def6)
![Screenshot 2023-11-21 at 5 39 21 PM](https://github.com/micronaut-projects/micronaut-core/assets/532108/a9253a52-454e-43b9-b32b-a49a270aace1)

### Version

Micronaut Platform 4.1.6 -> 4.2.0

Spikhalskiy:
@dstepanov I tried SNAPSHOT core with #10195 and while it may decrease the duration of `kaptKotlin` stage by 10%, it unfortunately doesn't compensate for 5x degradation that comes from the switch 4.1.x -> 4.2.x. The root cause is probably somewhere else. Let me know if you want me to upload updated sampling snapshots.

dstepanov:
I think it was caused by https://github.com/micronaut-projects/micronaut-core/commit/e1a990bfe69672599f4fcc19dd1568d7387729d5 Do you have a lot of interfaces with annotations?

Can you benchmark using the async profiler? https://github.com/async-profiler/async-profiler: You can attach it to the Gradle worker remotely.

Spikhalskiy:
@dstepanov I can't say so. We have some, but it's in low double digits. Nothing to justify a 30 seconds -> 2.5 minutes jump. Mostly it's controllers with data classes.
[...]

PR #10196, opened 2023-12-04
Use native check for overrides of Java methods

dstepanov:
Not sure why was it skipped, I assume this method is more performant than our version.
Might help with https://github.com/micronaut-projects/micronaut-core/issues/10145


yawkat (approved):
maybe it's something related to bridge methods? who knows

dstepanov:
Looks like there is a bug in Javac that prevents it

PR #10211, opened 2023-12-05
Only add content-length if the request type permits a body.

pbthorste:
Currently, `DefaultHttpClient.java` sends a `Content-Length: 0` for GET requests. This can cause issues, for example in AWS ALBs, which will most often refuse the request.

For more info on the AWS ALB issue, see https://stackoverflow.com/questions/77288320/aws-application-load-balancer-rejecting-requests-with-content-length-0

The PR makes sure that the content-length header is not included for requests that do not permit a body, such as GET requests.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1301,7 +1301,7 @@ public class DefaultHttpClient implements
                 Object bodyValue = body.get();
                 if (bodyValue instanceof CharSequence sequence) {
                     ByteBuf byteBuf = charSequenceToByteBuf(sequence, requestContentType);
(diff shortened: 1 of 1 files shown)


sdelamo (changes requested):
Can you use `HttpMethod::permitsRequestBody` instead?

pbthorste:
@sdelamo The method you mention is actually how the boolean `permitsBody` is created, before being sent to the `buildNettyRequest` method, which reads the value and then uses it to decide whether or not to set a body.
(see lines nr. 1488 and 1560 )

I guess the code could be changed to make this clearer. It could be argued that the `buildNettyRequest` method does not really need the permitsBody parameter since it could figure it out itself by reading the request object. But it is `protected` so maybe changing the signature of the method could be risky.

sdelamo:
@timyates can you write a http client tck test which verifies the current behavior of the HTTP Header Content-Length for POST and GET requests?

sdelamo:
@pbthorste we are incorporating your patch here: https://github.com/micronaut-projects/micronaut-core/pull/10218

PR #10212, opened 2023-12-05
Execute async controller methods without Reactor

dstepanov:


graemerocher (approved):

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlow.java
```diff
@@ -38,8 +38,8 @@ public interface CompletableFutureExecutionFlow<T> extends ExecutionFlow<T> {
      * @return a new flow
      */
     @NonNull
-    static <K> ExecutionFlow<K> just(@NonNull CompletableFuture<K> value) {
```
yawkat, line 41: i hope this method isn't used outside core because this is a binary incompatible change. it's probably fine because this is Internal but it might be another servlet situation

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -661,6 +670,47 @@ private CorePublisher<MutableHttpResponse<?>> fromReactiveExecute(PropagatedCont
         return processPublisherBody(propagatedContext, request, response, routeInfo);
     }
 
+    @NonNull
+    private CompletionStage<MutableHttpResponse<?>> fromCompletionStage(@NonNull HttpRequest<?> request,
+                                                                        @NonNull Object body,
+                                                                        @NonNull RouteInfo<?> routeInfo) {
+        CompletionStage<Object> completionStage = (CompletionStage<Object>) body;
```
yawkat, line 677: could be null i think

yawkat (approved):

core/src/main/java/io/micronaut/core/execution/CompletableFutureExecutionFlow.java
dstepanov: Ahh right, will see

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
dstepanov: Nulls are handled before

PR #10223, opened 2023-12-06
More detailed error messages

dstepanov:
Previously new detailed messages introduced in v4 only included one level of information about why the bean was disabled. That doesn't help much for more complicated configuration beans like MongoDB, JDBC etc. Now it should be more detailed.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
@@ -292,65 +292,73 @@ public class DefaultApplicationContext extends DefaultBeanContext implements App
     }
 
     @Override
-    protected <T> NoSuchBeanException newNoSuchBeanException(@Nullable BeanResolutionContext resolutionContext, Argument<T> beanType, Qualifier<T> qualifier, String message) {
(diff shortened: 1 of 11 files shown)


inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -292,65 +292,73 @@ protected void initializeContext(List<BeanDefinitionProducer> contextScopeBeans,
...
+                                                             @Nullable Qualifier<T> qualifier,
+                                                             @Nullable String message) {
         if (message == null) {
-            message = super.resolveDisabledBeanMessage(resolutionContext, beanType, qualifier);
-        }
-
-        if (message == null) {
-            message = resolveIterableBeanMissingMessage(resolutionContext, beanType, qualifier, message);
+            StringBuilder stringBuilder = new StringBuilder();
+            String ls = CachedEnvironment.getProperty("line.separator");
```
yawkat, line 301: can we get someone to run the test suite on windows

yawkat (approved):
cool feature

dstepanov:
Yeah, that was originally implemented by @graemerocher. I have just improved it a bit.

graemerocher (approved):

PR #10224, opened 2023-12-06
Restore `RequestLifecycle` for backwards compatibility with Servet

dstepanov:


yawkat (approved):

timyates:
@dstepanov @yawkat this isn't backward compatible

This is in 4.3.x https://github.com/micronaut-projects/micronaut-core/blob/c4e9e4c9577be5f74f74d653751b5c59142670df/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java#L99-L108

But on 4.2.x it's got a different return type

https://github.com/micronaut-projects/micronaut-core/blob/3220c09e42f49c02869fe1b7652cff52809401de/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java#L97-L116

Suggestions?

PR #10236, opened 2023-12-08
fix: consider current package when generating bean imports

auke-:
The internal bean element implementation didn't implement the methods for the determination of the visibility of the class, which resulted in the use of reflection for the bean definitions generated for the bean import.

By implementing the visibility methods the generated code for the imported bean definitions don't use reflection when the fields are visible.

I've tried to add a test to [inject-java-test/src/test/groovy/io/micronaut/inject/beanimport/BeanImportSpec.groovy](https://github.com/micronaut-projects/micronaut-core/blob/4.2.x/inject-java-test/src/test/groovy/io/micronaut/inject/beanimport/BeanImportSpec.groovy) but i'm unable to determine if reflection is used, this information isn't available in the BeanDefinition:

```java
    void "test bean import for current package"() {
        given:
        ApplicationContext context = buildContext('''
package io.micronaut.inject.beanimport;

import io.micronaut.context.annotation.*;
import jakarta.inject.Named;
import java.nio.charset.StandardCharsets;

@Import(classes=io.micronaut.inject.beanimport.BeanWithFields.class)
class Application {}
''')
        def bean = context.getBeanDefinition(BeanWithFields)

        expect:
        bean.getInjectedFields()[0].getName() == 'publicField' // How to determine that reflection isn't used?
        bean.getInjectedFields()[1].getName() == 'packageField' // How to determine that reflection isn't used?
        bean.getInjectedFields()[2].getName() == 'protectedField' // How to determine that reflection is used?
        bean.getInjectedFields()[3].getName() == 'privateField'  // How to determine that reflection is used?
    }

@Singleton
class BeanWithFields {
    public @Named("public") String publicField;
    @Named("package") String packageField;
    protected @Named("protected") String protectedField;
    private @Named("private") String privateField;
}
```

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/ast/utils/EnclosedElementsQuery.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/utils/EnclosedElementsQuery.java
@@ -137,7 +137,9 @@ public abstract class EnclosedElementsQuery<C, N> {
                 if (element.isPrivate() || element.getName().startsWith("$")) {
                     return false;
                 }
-                if (element instanceof MemberElement memberElement && !memberElement.isAccessible()) {
+                // exclude when not accessible from type
+                if (element instanceof MemberElement memberElement && !result.getOnlyAccessibleFromType()
+                    .filter(fromType -> memberElement.isAccessible(fromType)).isPresent()) {
                     return false;
                 }
             }
--- a/core-processor/src/main/java/io/micronaut/inject/writer/AbstractBeanDefinitionBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/AbstractBeanDefinitionBuilder.java
@@ -923,6 +923,16 @@ public abstract class AbstractBeanDefinitionBuilder implements BeanElementBuilde
             return element.isPublic();
         }
 
+        @Override
+        public boolean isPackagePrivate() {
+            return element.isPackagePrivate();
+        }
+
+        @Override
+        public boolean isPrivate() {
+            return element.isPrivate();
+        }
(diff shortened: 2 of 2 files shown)


graemerocher:
could you add a test?

graemerocher:
Seems like the change broke something, so the logic is not quite correct

auke-:
I think my change exposed a bug in [core-processor/src/main/java/io/micronaut/inject/ast/utils/EnclosedElementsQuery.java](https://github.com/micronaut-projects/micronaut-core/pull/10236/files#diff-1f8112947db2668dc0553034a6d3db32d368d98bfee69e69aa97352a73ddaffd) which didn't filter the package private elements in case the `OnlyAccessibleFromType` was set.

I've noticed that the [4.3.x branch](https://github.com/micronaut-projects/micronaut-core/commits/4.3.x/) contains a major change to [EnclosedElementsQuery.java](https://github.com/micronaut-projects/micronaut-core/commit/c037ea18f03d45997c7838582ceb610a13246a88) and does seem to work with my change to [AbstractBeanDefinitionBuilder](https://github.com/micronaut-projects/micronaut-core/pull/10236/files#diff-a53514cb285129ecf940be5a5aeadb1b016c9f489f6952c9a9bb05d110ff8e01) but also doesn't consider `OnlyAccessibleFromType`.

I would like to add a test, but i'm not sure where to start since the result of the changes are a change in the generated bytecode and i'm unable to assert that the generated bean definition isn't using reflection for the package private fields.

(and i was triggering the wrong shortcuts which closed and re-openend  this merge request, sorry about that)

Issue #10228, opened 2023-12-06
KSP incorrectly interprets fields in the interface


altro3:
### Expected Behavior

Try to analyze `io.micronaut.data.model.Pageable`. Here differences between KSP and java:
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/d5a3b95d-615f-49de-805f-b295fc971c8c)

As you see we have two bugs:

1. Constants `UNPAGED` and `UNSORTED` were interpreted as properties
2. Problem with processing class `Sort.Order`: KSP lost property `direction`,  but interpreted wrong property `ascending`

### Version

4.2.1

altro3:
@dstepanov one more :-)

altro3:
@dstepanov The problem with method `KotlinClassElement.getBeanProperties()`: KSP identificate static final fields as bean property:

![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/084077f8-42eb-4e50-964c-a41df50c9348)

altro3:
About second problem with class `io.micronaut.data.model.Sort$Order`. 

Hm.... KSP identificate method `isAscending` as bean property, but it's a bug., because method `getBeanProperties` must return properties with getters and setters:
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/3e271db6-6c69-470d-8e3c-549eb74f4bcd)
Java and groovy ignore method `isAscending`

This screenshot for process with KSP:
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/ba951a5e-ac41-4a10-9341-7cbfe99419fb)

altro3:
It's unlikely, but perhaps the bug is not in the KSP, but in the Java visitor. You need to decide how to interpret the `isAscending` method - as a property of the bean or not. In the first case, KSP works correctly, but Java does not work correctly. In the second case - vice versa

altro3:
@dstepanov Found third critical bug: KSP can't identificate internal enums (that's why field `direction` was lost):
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/9071fc7a-fdde-4d3b-81d5-75d8f9b8ef14)
Class `io.micronaut.data.model.Sort$Order$Direction` identificated as simple `ClassElement`, but must be `EnumElement`

altro3:
[...]

PR #10245, opened 2023-12-11
KSP: Exclude static properties

dstepanov:
Fixes partially https://github.com/micronaut-projects/micronaut-core/issues/10228

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/visitors/ClassElementSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/visitors/ClassElementSpec.groovy
@@ -3319,6 +3319,77 @@ interface MyInterface {
         methods.size() == 4
     }
 
+    void "test bean properties interfaces"() {
+        def ce = buildClassElement('''
+package test;
+import io.micronaut.context.annotation.Executable;
+import io.micronaut.context.annotation.Prototype;
+import io.micronaut.visitors.data.Pageable;
+import jakarta.inject.Singleton;
+
+class TestNamed {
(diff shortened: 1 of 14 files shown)


graemerocher (approved):

dstepanov:
@altro3 This only fixes the static fields problem. `isAscending` is also working in Java (see test), and I cannot reproduce the  enum. How do you access it? Maybe you can modify the test to reproduce it.

altro3:
@dstepanov About method `isAscending` - yeah, sorry, this is my fault. I agree, java visitor has this property too. Just,  forget about this bug.

About enum `io.micronaut.data.model.Sort$Order$Direction`: There is still a problem, I identified the place where classElement is returned, instead of enumElement.

Problem inside method `KotlinSimplePropertyElement.getGenericType()`. Look to the sreenshot (KSP):
изображение
As you see, method `publicField.getType()` return right `KotlinEnumElement`, but method `publicField.getGenericType()` return `KotlinClassElement`.

Same code with java:
изображение
As you see, both methods return the same `JavaEnumElement`

dstepanov:
Added a fix for enums

altro3:
@dstepanov Forgot to tell you about another one bug: As you see on screenshot

изображение

required properties list is different between java and kotlin. For example, property `number`. That's because of wrong return value for method `element.isNonNull()`

KSP `element.isNonNull()` returns `true`:
изображение

Java `element.isNonNull()` returns `false`:
изображение

Need to fix primitive types processing for KSP or for java. Which value is right in this case?

PR #10246, opened 2023-12-11
Compression support in PipeliningServerHandler

yawkat:
Like #10155


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
```diff
@@ -0,0 +1,188 @@
...
+    @SuppressWarnings("FloatingPointEquality")
+    private Algorithm determineEncoding(List<String> acceptEncoding) {
+        // from HttpContentCompressor, slightly modified
+        float starQ = -1.0f;
+        float brQ = -1.0f;
+        float zstdQ = -1.0f;
+        float snappyQ = -1.0f;
+        float gzipQ = -1.0f;
+        float deflateQ = -1.0f;
+        for (String encoding : acceptEncoding) {
```
dstepanov, line 117: I wonder if this can be refactored as a `record(encoding, q, Algorithm) implements Comparable` and use `stream().sort().first()`

dstepanov (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
yawkat: It's copied straight from netty so i dont want to change it

yawkat: easier to port future changes that way :)

dstepanov: Ok

graemerocher (approved):

PR #10254, opened 2023-12-13
Use FastThreadLocal for context propagation if available

yawkat:
From #10131


dstepanov:
Can we have an abstraction around the thread context and only switch to Netty's implementation when it's available (server/client)?

yawkat:
I don't understand what you mean? If netty is not on the classpath, the SLOW variable will be used. If netty is on the classpath but the current thread is not a netty FastThreadLocalThread, then netty FastThreadLocal has its own handling already that falls back on ThreadLocal.

dstepanov (approved):

dstepanov:
Ah right, I didn't notice `compileOnly`

graemerocher (approved):

Issue #10255, opened 2023-12-13
Declarative HTTP Client should not force "error" logs on exceptions


oliverblaha:
### Issue description

When Micronaut's declarative HTTP client receives any error besides "not found", the generated exception will cause an "error" log message:

https://github.com/micronaut-projects/micronaut-core/blob/v4.2.1/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L384

However, this is not always desirable; in certain situations, this might be expected/part of a workflow.

Example: The application might not have knowledge if a user has permission to access an upstream resource, with the only option being to try it - if there's no access, it would receive "FORBIDDEN". In this case, it would be a normal part of the workflow, therefore it should not be an "ERROR" in the application's logs (making it too easy to miss real errors).

Just disabling logging altogether for that class would somewhat mitigate that, but feels like an ugly workaround.
I could not find a way to circumvent it otherwise - I appreciate any hint.

If there's no current option, could we make this log optional or the log level configurable? If so, I'll be happy to submit a PR.

yawkat:
i think the log message should be removed entirely from error level, it is enough to return it from the completion stage. feel free to make a pr that downgrades it to debug.

PR #10257, opened 2023-12-13
Downgrade HTTP client log message

oliverblaha:
Downgrades the log level of the HTTP client as discussed in #10255.


yawkat (approved):

yawkat:
Thank you!

PR #10261, opened 2023-12-13
imp: add default methods to BeanDefinition to ease checking whether a method is annotated  

sdelamo:
see: https://github.com/micronaut-projects/micronaut-security/pull/1526


graemerocher (changes requested):
I don't agree with exposing yet another public set of static utility methods which will confuse users and the API. Seems like these methods could added as default methods to `BeanRegistration` / `BeanDefinition` interface if you want to simplify the logic bit

sdelamo:
> I don't agree with exposing yet another public set of static utility methods which will confuse users and the API. Seems like these methods could added as default methods to `BeanRegistration` / `BeanDefinition` interface if you want to simplify the logic bit

I have removed the `Utils` classes.

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
```diff
@@ -253,6 +254,30 @@ default <R> Optional<ExecutableMethod<T, R>> findMethod(String name, Class<?>...
...
 
+    /**
+     * Whether an {@link ExecutableMethod} exists which is annotated with the supplied annotation
+     *
+     * @param methodName          The method name
+     * @param argumentTypes The argument types
+     * @return Whether an {@link ExecutableMethod} exists which is annotated with the supplied annotation
+     * @since 4.3.0
+     */
+    default boolean existsMethodWithAnnotation(Class<? extends Annotation> annotationClass, String methodName, Class<?>... argumentTypes) {
```
graemerocher: rename to `hasAnnotatedMethod`

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
```diff
@@ -253,6 +254,30 @@ default <R> Optional<ExecutableMethod<T, R>> findMethod(String name, Class<?>...
...
+
+    /**
+     * Whether an {@link ExecutableMethod} exists which is annotated with {@link Blocking}.
+     *
+     * @param methodName          The method name
+     * @param argumentTypes The argument types
+     * @return Whether an {@link ExecutableMethod} exists which is annotated with with {@link Blocking}.
+     * @since 4.3.0
+     */
+    default boolean existsMethodAnnotatedWithBlocking(String methodName, Class<?>... argumentTypes) {
```
graemerocher: I don't think this method adds any value, and we shouldn't have methods for every annotation type here. Just remove it and you can pass `Blocking.class` when calling the above method

graemerocher (changes requested):

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
graemerocher: Also add nullability annotations to the arguments

dstepanov:
I'm not sure if we should push forward the idea of annotating blocking services and let us do something about it, if we use virtual threads blocking is fine. It's better to have something like `@ExecuteOn`

inject/src/main/java/io/micronaut/inject/BeanDefinition.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/10261/commits/b02ad851014ab4a1a9d15cabb3acba13c3d1f114

graemerocher (approved):

dstepanov:
Do we really want to add `findBeanDefinition(@NonNull T bean)`? That requires iterating all the beans

sdelamo:
> Do we really want to add `findBeanDefinition(@NonNull T bean)`? That requires iterating all the beans

Yes, I think a method to get a bean definition given an instance it is necessary. Is there are more optimal way to do this?

Based on your comment I assume the issue is not this PR but that `BeanDefinitionRegistry::findBeanRegistration` implementation is not optimised for lookups and does iterate all the beans.

dstepanov:
The best way would be to get a bean definition/s and after that get a bean instance.

sdelamo:
> The best way would be to get a bean definition/s and after that get a bean instance.

but you may have a bean instance already.

dstepanov:
Getting a bean definition from an instance should be avoided, maybe refactored the usage of it.

sdelamo:
> Getting a bean definition from an instance should be avoided, maybe refactored the usage of it.

This PR is the usecase: https://github.com/micronaut-projects/micronaut-security/pull/1526

It seems you think we should not have the method `BeanDefinitionRegistry::findBeanRegistration(Object bean)`. Please note this PR does not add this method. It was already there.

sdelamo:
@dstepanov can you approve this PR o should I close it?

dstepanov (approved):

PR #10264, opened 2023-12-14
Allow Refreshable scope under test for Env.FUNCTION and ANDROID

timyates:
It was seen in the Aws module that `@MockLambdaTest` and `@MockBean` did not work as expected.

https://github.com/micronaut-projects/micronaut-aws/issues/1870

This is because:

1. `@MockBean` uses the Refreshable scope
2. `@MockLambdaTest` adds Environment.FUNCTION to the test context
3. The Refresh scope is disabled for the FUNCTION environment

After this PR, RefreshScope is still disabled for Environment.FUNCTION...

...*UNLESS* Environment.TEST is also enabled.

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/context/scope/refresh/RefreshScope.java
+++ b/context/src/main/java/io/micronaut/runtime/context/scope/refresh/RefreshScope.java
@@ -54,7 +54,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
  * @since 1.0
  */
 @Singleton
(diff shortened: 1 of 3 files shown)


timyates:
I *think* this is ok for 4.2.2, but it is a (potentially unexpected) change in functionality, so should we target 4.3.0?

Also, should we have some sort of property flag to restore the previous functionality (and ignore the TEST env) 🤔

graemerocher (approved):

sdelamo (approved):

PR #10272, opened 2023-12-15
KSP: use `overrides` method instead of `findOverridee()`

dstepanov:


graemerocher:
broken?

dstepanov:
I have deleted all injection private-package tests. Kotlin doesn't have a package-private scope.

graemerocher:
What does this fix?

dstepanov:
Nothing was reported, and two tests got un-pending after this change.
I noticed that this method exists while fixing some other KSP stuff, and I assume it's faster than the previous one.

graemerocher (approved):

PR #10275, opened 2023-12-15
Improve PathVariable Docs

### LINKED ISSUE #5094: Improve PathVariable Docs
At [docs.micronaut.io/2.4.0/guide/](https://docs.micronaut.io/2.4.0/guide/index.html) there is only 1 code sample which shows using `@PathVariable` on a method parameter.

```java
import io.micronaut.http.annotation.Controller
import io.micronaut.http.annotation.Get
import io.micronaut.http.annotation.PathVariable

@Controller("/issues") 
class IssuesController {

    @Get("/{number}") 
    String issue(@PathVariable Integer number) { 
        "Issue # " + number + "!" 
    }
}
```

There is a footnote attached to that code sample which reads:

> The method argument can optionally be annotated with PathVariable

There is no explanation of why you might do that, and without reading the footnote one might conclude from the code sample that `@PathVariable` may be required there.

It would be good to add some info to the docs explaining why/when to use `@PathVariable` on method parameters in particular (there is a little info later on using `@PathVariable` on a property, a separate use case for the annotation).

wetted:
closes #5094

### DIFF
--- a/src/main/docs/guide/httpServer/routing.adoc
+++ b/src/main/docs/guide/httpServer/routing.adoc
@@ -19,7 +19,7 @@ You can use api:http.uri.UriTemplate[] to build paths to include in your respons
 
 == URI Path Variables
 
-URI variables can be referenced via method arguments. For example:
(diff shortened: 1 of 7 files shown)


wetted:
more coming...

sdelamo (approved):

sdelamo (approved):
Thanks @wetted this documentation PRs are great.

PR #10281, opened 2023-12-18
Use Javac's `overrides` to determine if the method is overridden

dstepanov:
I investigated why we cannot use it by default, there is some limitation to the reflection invocation of the package-private methods. Added it as a note.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/ast/MemberElement.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/MemberElement.java
@@ -76,17 +76,7 @@ public interface MemberElement extends Element {
      * @since 3.4.0
      */
     default boolean isReflectionRequired(@NonNull ClassElement callingType) {
-        if (isPublic()) {
-            return false;
-        }
-        if (isPackagePrivate() || isProtected()) {
-            // the declaring type might be a super class in which
-            // case if the super class is in a different package then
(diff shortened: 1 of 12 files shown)


graemerocher (approved):

core-processor/src/main/java/io/micronaut/inject/ast/MemberElement.java
```diff
@@ -104,6 +94,61 @@ default boolean isAccessible() {
...
+        if (isPackagePrivate()) {
+            // the declaring type might be a super class in which
+            // case if the super class is in a different package then
+            // the method or field is not visible and hence reflection is required
+            final ClassElement declaringType = getDeclaringType();
+            String packageName = declaringType.getPackageName();
+            if (!packageName.equals(callingType.getPackageName())) {
+                return allowReflection && hasAnnotation(ReflectiveAccess.class);
+            }
+            if (isPackagePrivate()) {
```
Spikhalskiy, line 134: @dstepanov Do we need one more call for `isPackagePrivate()` here? We just checked it in the if above.
The first part of `isPackagePrivate` branch looks like a copy of the `isProtected` branch, so it looks like it could be a refactoring leftover.

dstepanov: Yeah, I tried to refactor and failed :) I will give it one more chance.

PR #10283, opened 2023-12-18
Initialize classes at build time that can deadlock graal compiler

graemerocher:
A class loading deadlock can occur when `ConversionContext` is initialised concurrently since it has static final fields that are sub interfaces which are also initialised causing a loop.

This doesn't manifest on JIT likely because these classes are never initialised concurrently, but a deadlock can occur in the Graal compiler.

A real fix would to be alter the code but unfortunately there is no way to do this in a backwards compatible way. 

A workaround in Graal is to explicitly initialise these classes at build time which happens very early, avoiding the concurrent access that causes the deadlock.


graemerocher:
//cc @christianwimmer

graemerocher:
Unfortunately I don't know how to reproduce or test this change

dstepanov:
Can you make a comment in the file referring to this issue?

dstepanov (approved):

PR #10293, opened 2023-12-21
Javac: Extract type variable name using the element

dstepanov:
Graalvm dev build includes the annotations in the `TypeVariable`'s `toString` representation. It's better to correct its usage for the future.


graemerocher (approved):

n0tl3ss:
Should this be also back-ported to other releases?

PR #10316, opened 2024-01-02
Add test reproducing importing a bean with a primitive parameter

dstepanov:
The problem is probably fixed by https://github.com/micronaut-projects/micronaut-core/commit/afbb6f979cfbef17eec8043a62f1729e0215e19d


graemerocher (approved):

dstepanov:
It looks like it cannot be a pending feature.

PR #10320, opened 2024-01-03
Expression implementation improvements

dstepanov:
- Moved some of the code that is only used in the annotation processor.
- Naming improvements
- Introduced a new context that can be used for retrieving the expression type without providing the ASM writer (that is used in Micronaut Data)

### DIFF
--- a/core-processor/src/main/java/io/micronaut/expressions/EvaluatedExpressionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/expressions/EvaluatedExpressionWriter.java
@@ -19,7 +19,7 @@ import io.micronaut.context.expressions.AbstractEvaluatedExpression;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.expressions.ExpressionEvaluationContext;
 import io.micronaut.expressions.context.ExpressionWithContext;
-import io.micronaut.expressions.parser.CompoundEvaluatedEvaluatedExpressionParser;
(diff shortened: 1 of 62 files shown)


graemerocher:
It would be nice to somehow expose something in the Element API for getting the type of an expression

dstepanov:
I do this in Micronaut Data:
```java
    public static ClassElement extractExpressionType(MatchContext matchContext, AnnotationValue<ParameterExpression> parameterExpression) {
        Object expressionValue = parameterExpression.getValues().get("expression");
        if (expressionValue instanceof String) {
            throw new ProcessingException(matchContext.getMethodElement(), "Expected an expression '#{...}' found a string!");
        }
        EvaluatedExpressionReference ref = (EvaluatedExpressionReference) expressionValue;
        DefaultExpressionCompilationContextFactory factory = new DefaultExpressionCompilationContextFactory(matchContext.getVisitorContext());
        ExpressionCompilationContext compilationContext = factory.buildContextForMethod(ref, matchContext.getMethodElement());
        String expression = (String) ref.annotationValue();
        return new CompoundEvaluatedEvaluatedExpressionParser(expression)
            .parse()
            .resolveClassElement(new ExpressionVisitorContext(compilationContext, matchContext.getVisitorContext(), null));
    }
```

graemerocher:
could it be made into a utility method upstream?

dstepanov:
Added

graemerocher (approved):

PR #10323, opened 2024-01-04
Don't log decompression exception

yawkat:
Decompression errors should be treated as DEBUG to prevent log spam.

I've also copied over some other fuzz tests from #10131. None of them were an issue except one caused by a bug in netty EmbeddedChannel ( https://github.com/netty/netty/pull/13730 ).


timyates (approved):
Not sure the Sonar issues can be addressed (or are even an issue)

graemerocher (approved):

PR #10325, opened 2024-01-04
Publish service ready / stopped in consistent order

dstepanov:


graemerocher (approved):

test-suite/src/test/groovy/io/micronaut/EventListenerSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+import io.micronaut.runtime.server.event.ServerShutdownEvent
+import io.micronaut.runtime.server.event.ServerStartupEvent
+import jakarta.inject.Singleton
+import spock.lang.RepeatUntilFailure
+import spock.lang.Specification
+import spock.util.concurrent.PollingConditions
+
+class EventListenerSpec extends Specification {
+
+    @RepeatUntilFailure(maxAttempts = 100) // Retry the test multiple times to test the listeners order
```
sdelamo: https://spockframework.org/spock/docs/2.3/extensions.html
> @RepeatUntilFailure annotation can be used to repeat a single feature until it fails or until it reaches the maxAttempts value. This is intended to aid in manual debugging of flaky tests and should not be committed to source control.

Should not we remove `@RepeatUntilFailure` and don't commit it so that we don't run this test a hundred times? I assume you run it locally a 100 times and it passed always. Thus, it should not be flaky.

test-suite/src/test/groovy/io/micronaut/EventListenerSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+        conditions.eventually {
+            t.events.size() == 6
+        }
+
+        t.events[0] == StartupEvent
+        t.events[1] == ServerStartupEvent
+        t.events[2] == ServiceReadyEvent
+        t.events[3] == ServerShutdownEvent
+        t.events[4] == ServiceStoppedEvent
```
sdelamo, line 74: should not the order be 

```
StartupEvent
ServerStartupEvent
ServiceReadyEvent
ServiceStoppedEvent
ServerShutdownEvent
ShutdownEvent
```

instead of

```
StartupEvent
ServerStartupEvent
ServiceReadyEvent
ServerShutdownEvent
ServiceStoppedEvent
ShutdownEvent
```
?

test-suite/src/test/groovy/io/micronaut/EventListenerSpec.groovy
dstepanov: We can, but then if you remove the order in the future it will become flaky instead of failing directly on some attempt

test-suite/src/test/groovy/io/micronaut/EventListenerSpec.groovy
dstepanov: The service events are published by the `NettyServiceDiscovery` reacting on the server started /  shutdown

graemerocher: the event order is correct in the PR

test-suite/src/test/groovy/io/micronaut/EventListenerSpec.groovy
sdelamo: I think it is better to become flaky if we break it and then we should fix it than run it 100 times.

sdelamo:
```suggestion
```

test-suite/src/test/groovy/io/micronaut/EventListenerSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+import io.micronaut.core.order.OrderUtil
+import io.micronaut.discovery.event.ServiceReadyEvent
+import io.micronaut.discovery.event.ServiceStoppedEvent
+import io.micronaut.inject.qualifiers.Qualifiers
+import io.micronaut.runtime.event.annotation.EventListener
+import io.micronaut.runtime.server.EmbeddedServer
+import io.micronaut.runtime.server.event.ServerShutdownEvent
+import io.micronaut.runtime.server.event.ServerStartupEvent
+import jakarta.inject.Singleton
+import spock.lang.RepeatUntilFailure
```
sdelamo:
```suggestion
```

sdelamo (changes requested):

sdelamo (approved):

Issue #9984, opened 2023-10-16
Validation with generic annotations doesn't work


altro3:
### Expected Behavior

Controller:

```java
public interface RequestBodyApi {
    @Post("/sendValidatedCollection")
    Mono<Void> sendValidatedCollection(
        @Body @HardNullable List<@NotNull List<@Pattern(regexp = "[a-zA-Z]+") @Size(min = 3) @NotNull String>> requestBody
    );
}
```

Test:

```groovy
    void "test send validated collection"() {
        given:
        HttpRequest<?> request =
                HttpRequest.POST("/sendValidatedCollection", List.of(List.of("a", "hello", "123")))
                        .contentType(MediaType.APPLICATION_JSON_TYPE)

        when:
        client.retrieve(request, Argument.of(String), Argument.of(String))

        then:
        def e = thrown(HttpClientResponseException)

        HttpStatus.BAD_REQUEST == e.status
        e.message.contains("collection[0][0]: size must be between 3 and 2147483647")
    }
```

Response status always OK.

All sources you can see here: https://github.com/micronaut-projects/micronaut-openapi

### Version

4.1.9

dstepanov:
Does it work with a single List?

altro3:
@dstepanov yes, the sam. But I think I found the reason for all these 3 bugs with validation: The problem arises when annotations are written in the interface and are not duplicated in the implementation class.

the current bug is 100% related to this controller design, because if you explicitly duplicate the annotations in the implementation class, then the validation starts working.

Perhaps for the same reason there are problems with Kotlin.

dstepanov:
It’s strange, we remap the annotations to be inherited. Do you have added the validation annotations processor?

altro3:
Of course it’s added, otherwise validation wouldn’t work at all, but it doesn’t work only for generic annotations.

In general, I was right, if I duplicate the annotations in the implementation class, then everything works

Dependencies:
![изображение](https://github.com/micronaut-projects/micronaut-core/assets/48474279/20ce66cd-1e53-4539-beac-d37116f13b8c)

Classes structure:
```java
@Controller
public interface RequestBodyApi {
    @Post("/sendValidatedCollection")
    Mono<Void> sendValidatedCollection(
        @Body @HardNullable List<@NotNull List<@Pattern(regexp = "[a-zA-Z]+") @Size(min = 3) @NotNull String>> requestBody
    );
}
```
Implementation
```java
@Controller
[...]

PR #10342, opened 2024-01-08
AST: Introduce `MethodElement#getOverriddenMethods` to get all the methods overridden by this method

### LINKED ISSUE #10338: AST: Provide a way to get all overridden methods
### Feature description

Currently, we don't support the inheritance of the generic arguments across all inherited methods. I don't think there is an easy way to implement it in Micronaut 4. The only way is to traverse the generic arguments manually https://github.com/micronaut-projects/micronaut-validation/pull/228 but for that, we need a new API that will return those methods and doesn't slow the performance of the annotation processor.

dstepanov:
Reusing some of the logic from `EnclosedElementsQuery` to calculate the overridden methods for the annotation inheritance.

Fixes https://github.com/micronaut-projects/micronaut-core/issues/10338


inject-groovy/src/main/groovy/io/micronaut/ast/groovy/GroovyNativeElementHelper.java
```diff
@@ -0,0 +1,102 @@
...
+    private boolean isAssignable(ClassNode c1, ClassNode c2) {
+        if (c1.equals(c2)) {
+            return true;
+        }
+//        if (subDeclaringClass.isInterface()) {
+//            if (!superDeclaringClass.isInterface()) {
+//                return false;
+//            }
+//            return subDeclaringClass.declaresInterface(superDeclaringClass);
+//        }
```
graemerocher: remove this commented out code

inject-java/src/main/java/io/micronaut/annotation/processing/GenericUtils.java
```diff
@@ -1,345 +0,0 @@
...
-import javax.lang.model.util.Types;
-import java.util.*;
-
-/**
- * Utility methods for dealing with generic type signatures.
- *
- * @author Graeme Rocher
- */
-@Internal
-public class GenericUtils {
```
graemerocher, line 38: are we sure this is not used by other modules?

inject-java/src/main/java/io/micronaut/annotation/processing/JavaNativeElementsHelper.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.List;
+
+/**
+ * The Java native element helper.
+ *
+ * @author Denis Stepanov
+ * @since 4.3.0
+ */
+@Internal
+public final class JavaNativeElementsHelper extends NativeElementsHelper<TypeElement, ExecutableElement> {
```
graemerocher, line 42: does it need to be public?

graemerocher (approved):

inject-java/src/main/java/io/micronaut/annotation/processing/JavaNativeElementsHelper.java
dstepanov: It's used from different packages

dstepanov:
@Spikhalskiy Would you mind testing the performance of this change in your project?

Spikhalskiy:
@dstepanov It cuts kaptKotlin stage from ~20 sec to ~16-17 sec for us, so 15-20% improvement. Great. Thank you so much for all this build performance work. It greatly improves the experience with Micronaut and we are looking forward to 4.3.0 release.

dstepanov:
@Spikhalskiy Thanks! I hope we will not lose much of the perf by implementing generic annotations inheritance that is missing now https://github.com/micronaut-projects/micronaut-validation/pull/228

Spikhalskiy:
@dstepanov feel free to tag me when the fix for #9984 (or a localized fix for micronaut-validation) is merged, I will report how it looks on our end.

Issue #9162, opened 2023-04-27
Expose fine-grained settings of WebSocket compression


jeremyg484:
### Feature description

Capturing as an off-shoot of #7809 

"I believe it would be useful to allow further configuration of the per message deflate extension, potentially in both the client and server websocket implementations.

Things in the spec that are meant to potentially be configurable are:

Context Takeover
Window Bits
Compression Level
Memory Level
I believe these should be exposed in a similar manner, to whatever extent Netty allows them to be configured under the hood.

Some references as to why this additional configuration is useful:

https://mailarchive.ietf.org/arch/msg/hybi/F9t4uPufVEy8KBLuL36cZjCmM_Y/
https://www.igvita.com/2013/11/27/configuring-and-optimizing-websocket-compression/
"

graemerocher:
Triage: this remains a valid WebSocket enhancement distinct from the completed enable/disable setting in Core PR #7809. Current configuration exposes only coarse WebSocket compression controls; evaluate Netty’s supported context-takeover, window-bits, compression-level, and memory-level options for both client and server, with safe defaults and compatibility tests for negotiated extension parameters.

PR #10357, opened 2024-01-10
Allow client config of WebSocket per message deflate

### LINKED ISSUE #7711: DefaultHttpClient no way to turn off compression

jeremyg484:
Add a configuration property to HttpClientConfiguration that allows the enabling or disabling of per message deflate for WebSocket clients. Defaults to true in order to preserve the existing behavior.

This potentially resolves https://github.com/micronaut-projects/micronaut-core/issues/7711

This PR is from a new branch based on 4.3.x and replaces #7809

Other related WebSocket configuration bits potentially worth supporting are captured in https://github.com/micronaut-projects/micronaut-core/issues/9162


jeremyg484:
@sdelamo Updated PR to 4.3.x. #7809 can be closed in favor of this one, assuming we're not trying to back port the feature to 3.x.

timyates:
@jeremyg484 `:http-client-jdk:compileTestGroovy` is failing...

jeremyg484:
> @jeremyg484 `:http-client-jdk:compileTestGroovy` is failing... 

I'll check it out, thanks for the heads up.

http-client/src/test/groovy/io/micronaut/http/client/config/DefaultHttpClientConfigurationSpec.groovy
```diff
@@ -78,6 +78,25 @@ class DefaultHttpClientConfigurationSpec extends Specification {
...
+        compressionConfig[property] == expected
+
+        cleanup:
+        ctx.close()
+
+        where:
+        key               | property         | value   | expected
+        'enabled'         | 'enabled'        | 'false' | false
+    }
+
```
timyates: Shouldn't we test the enabled 'true' setting?

And do we need a test for the default?

timyates (changes requested):

http-client/src/test/groovy/io/micronaut/http/client/config/DefaultHttpClientConfigurationSpec.groovy
jeremyg484: Maybe? :) TBH, I was just following the existing patterns in this spec. IIRC every test in here is testing the config properties being set to something other than the default.

IMO, the more interesting (and effective) tests are in `BinaryWebSocketSpec` - we do test the default/true case there.

timyates: Hehehe, yeah I think this spec isn't the best... Using a `where:` provider for a single case just adds an un-required level of indirection and confusion 😢 

Pushed https://github.com/micronaut-projects/micronaut-core/pull/10357/commits/492b6955b2a32fa762664d750ee20de20aaea8c7 to modernize and simplify things 😎

timyates (approved):

sdelamo (changes requested):
This is a breaking change.

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -302,6 +301,13 @@ public void setSslConfiguration(SslConfiguration sslConfiguration) {
         this.sslConfiguration = sslConfiguration;
     }
 
+    /**
+     * Obtains the WebSocket compression configuration.
+     *
+     * @return The WebSocket compression configuration.
+     */
+    public abstract WebSocketCompressionConfiguration getWebSocketCompressionConfiguration();
```
sdelamo: it misses `@since`. It misses nullability annotation in the return type.

sdelamo (approved):

PR #10358, opened 2024-01-10
Improve how AOP proxy target is generated

dstepanov:
I noticed that we would always call `getProxyTargetMethod` which would look for the bean definition all the time instead of reusing it.
Plus a few byte code improvements


graemerocher:
There are check style warnings

graemerocher (approved):

PR #10359, opened 2024-01-11
Inject `InterceptorRegistry` into an AOP proxy, avoid multiple get bean calls

dstepanov:


aop/src/main/java/io/micronaut/aop/chain/InterceptorChain.java
```diff
@@ -121,36 +121,76 @@ public R proceed() throws RuntimeException {
...
      * @return The filtered array of interceptors
      */
     @SuppressWarnings("WeakerAccess")
     @Internal
     @UsedByGeneratedCode
-    public static Interceptor[] resolveAroundInterceptors(
-            @Nullable BeanContext beanContext,
-            ExecutableMethod<?, ?> method,
-            List<BeanRegistration<Interceptor<?, ?>>> interceptors) {
+    public static <T> Interceptor<T, ?>[] resolveAroundInterceptors(BeanContext beanContext,
```
dstepanov, line 130: No signature changes to the old method, just some generics added to satisfy the compiler.

graemerocher (approved):

Issue #9764, opened 2023-08-22
KSP: Problem with making @Singleton allOpen and ksp


Spikhalskiy:
### Actual Behaviour

We have the following class and the project builds with kapt, allopen and micronaut 4.0.4

```
@Singleton
class SomeRequestValidator @Inject constructor(
    private val someService: SomeService
) : RequestValidator<SomeRequest> {
    ...
}
```
```
allOpen {
    annotations("jakarta.inject.Singleton")
    preset("micronaut")
}
```

If I try to migrate to ksp and use micronaut-core:4.0.6-SNAPSHOT and mictonaut-inject-kotlin:4.0.6-SNAPSHOT, I get

```
> Task :kspKotlin
e: [ksp] Originating element: SomeRequestValidator
e: [ksp] /<path>/SomeRequestValidator.kt:32: Cannot apply AOP advice to final class. Class must be made non-final to support proxying: <package>.SomeRequestValidator
e: Error occurred in KSP, check log for detail
```

Making the class explicitly "open" helps. 

It looks like there is a hardcoded list of classes that are considered "shouldBeOpen": https://github.com/micronaut-projects/micronaut-core/pull/9456/files#diff-8f68af29dd990607ef86587858c4f4e22a496688049ab10f74029a93149206ebR136
It would be great if we can add to it through grade configuration the same way allOpen plugin can be configured.

### Version

Micronaut 4.0.4 / 4.0.6-SNAPSHOT

dstepanov:
Unfortunately, KSP doesn't see the open classes by the `allOpen` plugin.

What is your use case with `SomeRequestValidator`? Is there an intercepted method?

Spikhalskiy:
@dstepanov This specific class is just a `@Singleton` that implements `RequestValidator`. There is no intercepted methods. I think this problem will be present for any Singleton with `@Inject` on its constructor. That’s why we have to add `jakarta.inject.Singleton` to the `allOpen` config section.

yawkat:
AOP advice is usually added when there is an additional AOP annotation at the method level. You would then mark that annotation for the all open plugin, not the entirety of `@Singleton`.

dstepanov:
A simple singleton doesn't need to have AOP advice.
Maybe you are using Jakarta Validation?

yawkat:
note that there is an open issue with method-level annotations for all-open support too: #9727

Spikhalskiy:
>Maybe you are using Jakarta Validation?

[...]

Issue #9750, opened 2023-08-21
KSP: Performance regression after migration to KSP


tobias-bahls:
### Expected Behavior

After migrating from kapt to KSP, we notice a large performance regression: 

Both `kaptGenerateStubs` and  `kaptKotlin` combined take ~1.5 min on my machine. `kspKotlin` takes over 30 mins, at which point I aborted the task.

### Actual Behaviour

I expected the `kspKotlin` step taking significantly less time than `kaptGenerateStubs` and  `kaptKotlin` combined.

### Steps To Reproduce

Unfortunately I don't know what exactly in our codebase triggers this issue. If there are anything more that I can provide apart from the attached VisualVM sampling profile, I'm happy to help.

<img width="2519" alt="Screenshot 2023-08-21 at 11 10 52" src="https://github.com/micronaut-projects/micronaut-core/assets/1154052/57c92478-1246-48c9-8736-05e70786cecd">

[micronaut-ksp-perf-issue.nps.zip](https://github.com/micronaut-projects/micronaut-core/files/12394347/micronaut-ksp-perf-issue.nps.zip)

### Environment Information

- Operating System: macOS
- JVM: OpenJDK Runtime Environment Temurin-17.0.8+7 (build 17.0.8+7)

### Version

4.0.4

dstepanov:
How big is your project?
I don't think we can do anything about it, you better report it here https://github.com/google/ksp

tobias-bahls:
Thanks for looking into this. We have roughly 3k classes and interfaces in our project

Spikhalskiy:
We have the same experience with a small project.
kapt - about 1 minute time for all kapt steps. 
ksp - 30 minutes+. Never finished, currently facing failures that were reported separately, but they happen after 30 minutes. 
We have around 1k classes.
We are open to collecting any telemetry during the build needed to investigate the reason. Currently, unfortunately, I don't see any options to turn on any kind of verbosity on ksp side that may be helpful: https://github.com/google/ksp/issues/1415

dstepanov:
I have merged an improvement, you can try 4.0 SNAPSHOT or wait for next 4.0.X release

Spikhalskiy:
I didn't find big improvements with 4.0.6-SNAPSHOT.
I'm not sure if I plug the snapshot `micronaut-core` and `micronaut-inject-kotlin` into `ksp` scope correctly though.

```
[libs.version.toml]
  [versions]
[...]

PR #10360, opened 2024-01-11
KSP: Allow to specify extra all-open annotations

dstepanov:
Workaround for https://github.com/micronaut-projects/micronaut-core/issues/9764

### DIFF
--- a/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
+++ b/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
@@ -132,12 +132,17 @@ internal abstract class AbstractKotlinElement<T : KotlinNativeElement>(
         false
     }
 
-    private fun shouldBeOpen(annotationMetadata: AnnotationMetadata) = annotationMetadata.declaredMetadata.hasDeclaredStereotype(
-        Around::class.java,
-        Introduction::class.java,
-        InterceptorBinding::class.java,
-        InterceptorBindingDefinitions::class.java
-    )
+    private fun shouldBeOpen(annotationMetadata: AnnotationMetadata): Boolean {
+        if (extraOpenAnnotations != null && annotationMetadata.declaredMetadata.hasDeclaredStereotype(*extraOpenAnnotations)) {
+            return true
+        }
+        return annotationMetadata.declaredMetadata.hasDeclaredStereotype(
+            Around::class.java,
+            Introduction::class.java,
+            InterceptorBinding::class.java,
+            InterceptorBindingDefinitions::class.java
+        )
+    }
 
     override fun isAbstract(): Boolean {
         return if (annotatedInfo is KSModifierListOwner) {
@@ -605,6 +610,8 @@ internal abstract class AbstractKotlinElement<T : KotlinNativeElement>(
     }
 
     companion object {
+        val extraOpenAnnotations = System.getProperty("kotlin.allopen.annotations")?.split(",")?.toTypedArray()
+
         val primitives = mapOf(
             "kotlin.Boolean" to PrimitiveElement.BOOLEAN,
             "kotlin.Char" to PrimitiveElement.CHAR,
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

yawkat (approved):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
```diff
@@ -605,6 +610,8 @@ internal abstract class AbstractKotlinElement<T : KotlinNativeElement>(
     }
 
     companion object {
+        val extraOpenAnnotations = System.getProperty("kotlin.allopen.annotations")?.split(",")?.toTypedArray()
```
Spikhalskiy, line 613: Can we maybe get in as a part of the gradle plugin configuration too?

dstepanov: Probably yes, maybe even have it automatically if the allopen plugin exists @melix, right?
Let's first see that it's working.

sdelamo (changes requested):
Please, document the system property`kotlin.allopen.annotations`

dstepanov:
@sdelamo Let's first check that it is working, and then I can document it. We might need to add it for KAPT, too, if the next version is broken.

sdelamo:
> @sdelamo Let's first check that it is working, and then I can document it. We might need to add it for KAPT, too, if the next version is broken.

Please document it in this pull request. I don't want to have secret system properties in the code. I assume you have tested it, and it works. If, for whatever reason, we have to roll it back, then we can roll back the squashed commit resulting from this PR, including documentation.

dstepanov:
@sdelamo Added

sdelamo (approved):

Spikhalskiy:
@dstepanov It looks like it's not easy to make an environment variable with dots in it. Can we change it to screaming case?

dstepanov:
You can do it this way like in the OpenApi:
```
tasks.withType(JavaCompile) {
    options.fork = true
    options.forkOptions.jvmArgs << '-Dmicronaut.openapi.property.naming.strategy=SNAKE_CASE'

    ...
}
```

dstepanov:
https://micronaut-projects.github.io/micronaut-openapi/latest/guide/#systemPropertyConfiguration

Spikhalskiy:
@dstepanov OpenApi uses https://github.com/micronaut-projects/micronaut-openapi/blob/adc7d7c9554aa1fab8e79657d6a4a60ec92339c4/openapi/src/main/java/io/micronaut/openapi/visitor/ConfigUtils.java#L732 to read the property. I don't think System.getProperty will return you the property supplied with -D.

My results:
1. Snapshot micronaut, kotlin2.0-Beta2 with kapt and kapt.use.k2=true, supplying "-Dkotlin.allopen.annotations=jakarta.inject.Singleton" has no effect, it throws "Cannot apply AOP advice to final class. Class must be made non-final to support proxying" 
2. Snapshot micronaut, Kotlin 1.9.22 with ksp hangs for me on kspKotlin with the latest micronaut snapshot. With or without the property. It doesn't throw an error it used to throw about the final class after 30 minutes.

dstepanov:
It's only supported for KSP for now

Spikhalskiy:
I posted CPU sampling results of my ksp hanging in the KAPT -> KSP performance degradation issue: https://github.com/micronaut-projects/micronaut-core/issues/9750#issuecomment-1892652585
It looks like there is a severe ksp performance degradation between 4.2.x and 4.3.0-SNAPSHOT.

Spikhalskiy:
Update, I gave kspKotlin 40 minutes and it failed with the same error as before
```
Method defines AOP advice but is declared final. Change the method to be non-final in order for AOP advice to be applied.
```

The following is in gradle.properties:
```
kotlin.daemon.jvmargs=-Xmx4096M -Dkotlin.allopen.annotations="jakarta.inject.Singleton"
org.gradle.jvmargs=-Xmx4096M -Dkotlin.allopen.annotations="jakarta.inject.Singleton"
```

PR #10364, opened 2024-01-12
extract netty cookie into its own module

sdelamo:
This allows from serverless modules (e.g. AWS) to remove the dependency from http-netty which has a negative impact in application startart.


yawkat:
the serverless modules depend on netty to do their cookie parsing?

yawkat (approved):

graemerocher (approved):

sdelamo:
> the serverless modules depend on netty to do their cookie parsing?

Yes, we do use it 

https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-api-proxy/src/main/java/io/micronaut/function/aws/proxy/AwsCookies.java#L53-L67

PR #10366, opened 2024-01-12
Correct visibility field in `BeanElementBuilder`

### LINKED ISSUE #10262: Bean import is using reflection for visible fields
### Expected Behavior

When using the [Bean Import](https://docs.micronaut.io/latest/guide/#beanImport) functionality the generated bean definitions use reflection for fields which are visible. This is especially the case for package protected fields when the class annotated with [@Import](https://docs.micronaut.io/latest/api/io/micronaut/context/annotation/Import.html) is in the same package as the bean and thus the generated bean definition is in the same package.

It looks like the issue is caused by the absence of the `isPackagePrivate` and `isPrivate` methods in [AbstractBeanDefinitionBuilder.java](https://github.com/micronaut-projects/micronaut-core/pull/10236/files#diff-a53514cb285129ecf940be5a5aeadb1b016c9f489f6952c9a9bb05d110ff8e01), see https://github.com/micronaut-projects/micronaut-core/pull/10236 for a possible fix.

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.2.1

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/10262

I am not sure how to test this, but I can see the reflection is not used in the imported class definition.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/writer/AbstractBeanDefinitionBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/AbstractBeanDefinitionBuilder.java
@@ -919,6 +919,16 @@ public abstract class AbstractBeanDefinitionBuilder implements BeanElementBuilde
         }
 
         @Override
+        public boolean isPackagePrivate() {
+            return element.isPackagePrivate();
+        }
+
+        @Override
+        public boolean isPrivate() {
+            return element.isPrivate();
+        }
+
+        @Override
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

dstepanov:
It looks like it's not possible because the private-protected field doesn't work for classes in different classloaders

PR #10405, opened 2024-01-23
accept breaking changes

sdelamo:
Most breaking changes accepted by this PR were introduced in:

- [AST: Introduce MethodElement#getOverriddenMethods to get all the methods overridden by this method](https://github.com/micronaut-projects/micronaut-core/pull/10342)
- [Avoid having TypeConverter defined as beans](https://github.com/micronaut-projects/micronaut-core/pull/10315/files)


dstepanov:
Why do we need to track the backward compatibility of the TCK? I think we should try to generate the metadata with the internal annotation to avoid the check.

graemerocher (approved):

sdelamo:
> Why do we need to track the backward compatibility of the TCK? I think we should try to generate the metadata with the internal annotation to avoid the check.

Please, note there are many breaking changes accepted in this PR. The TCK accepted changes are a minority.

dstepanov (approved):

PR #10407, opened 2024-01-24
Fix boxing a primitive value

dstepanov:
Spotbugs detects that the boxing is done using the constructor invocations instead of using `valueOf`


sdelamo (approved):
is there a way to write a test for this?

dstepanov:
No

PR #10441, opened 2024-02-01
Minor HTTP performance improvements

yawkat:
- Use linux tcnative artifact for tests
- Negative compression threshold disables compression entirely
- Add an internal field to NettyHttpRequest for the NettyBodyAnnotationBinder convertible body cache, to avoid the attribute map
- Add a simple single-element cache to JacksonDatabindMapper for the reader/writer lookup
- Bypass ChannelPipeline.get if no logging handler is configured


http-server-netty/build.gradle
```diff
@@ -91,7 +91,7 @@ dependencies {
     }
     testImplementation(libs.managed.netty.tcnative.boringssl.static) {
         artifact {
-            classifier = Os.isArch("aarch64") ? "osx-aarch_64" : "osx-x86_64"
+            classifier = Os.isArch("aarch64") ? "linux-aarch_64" : "linux-x86_64"
```
dstepanov: What about 🍎 users ?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -164,6 +166,12 @@ public class NettyHttpRequest<T> extends AbstractNettyHttpRequest<T> implements
 
     private final BodyConvertor bodyConvertor = newBodyConvertor();
 
+    /**
+     * ONLY for NettyBodyAnnotationBinder use.
+     */
+    @Internal
+    public ArgumentBinder.BindingResult<ConvertibleValues<?>> convertibleBody;
+
```
dstepanov: I was thinking we could have a custom object representing the body, which can accept some-kind of visitor, so we can avoid `instanceof` for cases when the body is known.

dstepanov (approved):

http-server-netty/build.gradle
graemerocher: yeah probably need an OS check here

graemerocher (approved):

PR #10444, opened 2024-02-02
Fix for context propagation bug

### LINKED ISSUE #10443: New context propagation API works incorrectly on reactive subscribers
### Expected Behavior

In the following example code in a controller:

```java
  @Get("/foo")
  public Mono<String> foo() {
    // (1)
    return downstreamCall().map(this::mapResponse);
  }

  protected Mono<String> downstreamCall() { ... }

  private String mapResponse(String body) {
    // (2)
  }
```

Upon execution at code point (2) the `PropagatedContext` should always be the same as it was at code point (1).

### Actual Behaviour

When `mapResponse(...)` is called on a different thread than `foo()` was, then the `PropagatedContext` at code point (2) is non-deterministic, i.e. empty.

### Steps To Reproduce

1. Check out https://github.com/lgathy/micronaut-tracing/tree/bug/subsequent-client-traces
2. Run `./gradlew :micronaut-tracing-opentelemetry-http:test`
3. The newly added test case in `io.micronaut.tracing.opentelemetry.instrument.http.OpenTelemetryHttpSpec` fails

The failure is also demonstrated by unit tests added in my forked repo, see the Example application link.

### Environment Information

- Operating system: MacOS 14.2.1
- JDK version: 17.0.10

### Example Application

https://github.com/lgathy/micronaut-core/commit/8db3f9c02ee864fc27290e51b06d9a75b6101214

### Version

[4.0.0; 4.3.3]

lgathy:
Fixes #10443

Added unit tests then the fix: the calls to the `propagate` method on the `Subscriber` arguments were missing.

I would also see the benefit in defining the visibility of the second `propagate` method stricter, i.e. private or package private, mainly because static code analysis would catch it if it's unused.

### DIFF
--- a/core-reactive/src/test/groovy/io/micronaut/core/async/propagation/ReactivePropagationSpec.groovy
+++ b/core-reactive/src/test/groovy/io/micronaut/core/async/propagation/ReactivePropagationSpec.groovy
@@ -1,0 +1,252 @@
+package io.micronaut.core.async.propagation
+
+import io.micronaut.core.propagation.PropagatedContext
+import io.micronaut.core.propagation.PropagatedContextElement
+import org.reactivestreams.Publisher
+import org.reactivestreams.Subscriber
+import org.reactivestreams.Subscription
+import reactor.core.CorePublisher
+import reactor.core.CoreSubscriber
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
+import spock.lang.Specification
+
+import static io.micronaut.core.async.propagation.ReactorPropagation.addPropagatedContext
+
+class ReactivePropagationSpec extends Specification {
+
+    def "test context propagation on Publisher methods"() {
+        given:
+        PropagatedContext outerContext = PropagatedContext.empty()
+        PropagatedContext innerContext = PropagatedContext.empty().plus(new PropagatedElement())
+        ContextCapturingPublisher<Object> publisher = new ContextCapturingPublisher<>()
+        Subscriber<Object> subscriber = new EmptySubscriber<Object>()
+
+        when:
+        try (def ignore = outerContext.propagate()) {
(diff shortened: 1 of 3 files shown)


core-reactive/src/test/groovy/io/micronaut/core/async/propagation/ReactivePropagationSpec.groovy
```diff
@@ -0,0 +1,252 @@
...
+import reactor.core.CoreSubscriber
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
+import spock.lang.Specification
+
+import static io.micronaut.core.async.propagation.ReactorPropagation.addPropagatedContext
+
+class ReactivePropagationSpec extends Specification {
+
+    def "test context propagation on Publisher methods"() {
```
dstepanov, line 18: Can you pls create a real-world test where is that needed?

dstepanov (changes requested):

core-reactive/src/test/groovy/io/micronaut/core/async/propagation/ReactivePropagationSpec.groovy
dstepanov: Ah I see there is an example in tracing

dstepanov:
We can merge this, but as I wrote in the issue the propagation this way is very limited. I think we should completely remove it in Micronaut 5 and just rely on the Reactor-Micrometer propagation.

core-reactive/src/test/groovy/io/micronaut/core/async/propagation/ReactivePropagationSpec.groovy
lgathy: Please take a look at these tests in tracing: https://github.com/micronaut-projects/micronaut-tracing/pull/498

PR #10445, opened 2024-02-02
Restore thread contexts in reversed order

lgathy:
When multiple `ThreadPropagatedContextElement` are present in a `PropagatedContext`, restoring thread states happen in the same order as they were captured, whereas it should happen in a reversed order. Otherwise, incorrect thread state might be restored, as the added test cases demonstrate this behavior.


dstepanov (approved):

dstepanov:
Thanks!

PR #10448, opened 2024-02-03
small style fix in docs/streaming/HeadlineController.kt

matey-jack:


sdelamo (approved):
Thanks for the contribution @matey-jack

PR #10466, opened 2024-02-07
Inline `SameSiteConverter` to fix Graalvm Dev build

dstepanov:


sdelamo (approved):
What build will this fix? https://github.com/micronaut-projects/micronaut-core/actions/runs/7808528100 ?

dstepanov:
Well, it fixes our error, but there is another Junit related error.

graemerocher (approved):

graemerocher:
it will get us further in terms of passing the dev build ci. Then there is a mother problem with JUnit 5 that we can't address

PR #10469, opened 2024-02-08
Fix parameter expressions for executable methods

dstepanov:


core-processor/src/main/java/io/micronaut/inject/writer/EvaluatedExpressionProcessor.java
```diff
@@ -95,6 +97,11 @@ public void processEvaluatedExpressions(MethodElement methodElement) {
                 return new ExpressionWithContext(expression, evaluationContext);
             })
             .forEach(this::addExpression);
+
+        ClassElement resolvedThis = methodElement.isStatic() || methodElement instanceof ConstructorElement ? null : methodElement.getOwningType();
```
sdelamo, line 101: can you add a tests:

for  `methodElement.isStatic()` evaluates to true
`methodElement instanceof ConstructorElement` evaluates to true

?

sdelamo (approved):

graemerocher (approved):

PR #10471, opened 2024-02-08
testcontainers via test-resources bom in catalog

wetted:
I think this is the correct way to use the test-resources bom since we don't import catalogs in core. 🤷‍♂️


sdelamo (approved):

sdelamo:
ping @melix

melix:
This will work but it will introduce a circular dependency I think (we'd have to update the micronaut module checker to verify). Because core shouldn't depend on other modules. For this use case I don't think we have a good solution.

sdelamo:
Ok, let's close it and maintain this dependency manually.

PR #10490, opened 2024-02-13
Don't generate `BeanDefinitionReference` and `BeanIntrospectionReference`

dstepanov:
I don't see any benefit to generating `BeanDefinitionReference` and `BeanIntrospectionReference`, but there are some benefits now:
- Generating fewer classes should speed up the compilation
- Probably less JVM overhead
- BDs are now initialized at the build time for the native image
- Executable is smaller `:test-suite-http-server-tck-netty:nativeTestCompile` 77->73.7 MB


dstepanov:
Native size.
Before:
Screenshot 2024-02-14 at 15 19 02
After:
Screenshot 2024-02-14 at 15 18 53
Somehow Serde size got smaller even, so the changes are not projected there

dstepanov:
I verified that it works with Micronaut Serialization, and I ran some Micronaut Data tests.

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
```diff
@@ -495,13 +494,18 @@ private void writeIntrospectionClass(ClassWriterOutputVisitor classWriterOutputV
         final Type superType = Type.getType(AbstractInitializableBeanIntrospection.class);
 
         ClassWriter classWriter = new AptClassWriter(ClassWriter.COMPUTE_MAXS | ClassWriter.COMPUTE_FRAMES, visitorContext);
-        classWriter.visit(V17, ACC_SYNTHETIC | ACC_FINAL,
+        classWriter.visit(
+            V17,
+            ACC_SYNTHETIC | ACC_FINAL | ACC_PUBLIC,
```
yawkat, line 513: why does it need to be public now?

yawkat: is it because it's instantiated directly now?

inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
```diff
@@ -177,9 +175,59 @@ protected AbstractInitializableBeanDefinition(
...
-    @Override
-    public final boolean isConfigurationProperties() {
-        return precalculatedInfo.isConfigurationProperties;
+    /**
+     * Represents {@link BeanDefinitionReference#getBeanDefinitionName()} when the class implements {@link BeanDefinitionReference}.
+     *
+     * @return The name of this bean definition
+     */
+    // @Override for BeanDefinitionReference
+    public String getBeanDefinitionName() {
```
yawkat: i think it would be good to introduce another subclass of AbstractInitializableBeanDefinition that implements BeanDefinitionReference and properly overrides these methods. It is both more maintainable, and it may be better for performance.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -4113,6 +4113,9 @@ static final class BeanDefinitionProducer {
 
         BeanDefinitionProducer(@NonNull BeanDefinitionReference reference) {
             this.reference = reference;
+            if (reference instanceof BeanDefinition<?>) {
```
yawkat: you should also add an `instanceof AbstractInitializableBeanDefinition` check, it will be faster and should be hit most of the time.

inject/src/main/java/io/micronaut/context/RequiresCondition.java
```diff
@@ -98,6 +98,21 @@ public boolean matches(ConditionContext context) {
         }
         AnnotationMetadataProvider component = context.getComponent();
         boolean isBeanReference = component instanceof BeanDefinitionReference;
+        boolean both = isBeanReference && component instanceof BeanDefinition<?>;
```
yawkat: same

yawkat (approved):
it's a good change but a bit risky

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
dstepanov: Yes

graemerocher:
What is the binary compatibility story here? Will it break binary compatibility and require rebuilds of all modules?

dstepanov:
It is binary-compatible with the previously compiled references and bean definitions.

dstepanov:
Added abstract classes

graemerocher (approved):

PR #10501, opened 2024-02-16
Do not override with 404 when returning response inside CompletableFuture

### LINKED ISSUE #10419: A CompletableFuture Controller method returning an empty body results in 404
### Expected Behavior

A `@Controller` method with a return signature of `CompletableFuture<MutableHttpResponse<Object>>` can return a response with an empty body such as

```
return CompletableFuture.completedFuture(HttpResponse.ok());
```

### Actual Behaviour

Any CompletableFuture response with an empty body returns a 404 Not Found response.

The logic that results in this is at: https://github.com/micronaut-projects/micronaut-core/blob/443de14b0dbc4fa9449b6726fe25f60b4dacc6c3/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java#L711-L713

Note that this in contrast to doing the same with a `Publisher<MutableHttpResponse<Object>>` signature.

For example, this works as expected:
```
return Mono.just(HttpResponse.ok());
``` 

### Version

4.3.1

yawkat:
Fixes #10419

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyTest.java
@@ -20,11 +20,13 @@ import io.micronaut.core.annotation.Introspected;
 import io.micronaut.core.async.annotation.SingleResult;
 import io.micronaut.http.HttpHeaders;
 import io.micronaut.http.HttpRequest;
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

timyates (approved):
👍 Checked, and Servlet passes too 😻

PR #10516, opened 2024-02-19
Fix special case of NameUtils.underscoreSeparate where the first char is the separator char

### LINKED ISSUE #10140: Name Utils underscoreSeparate method adds extra underscores to the beginning of strings when starting with underscore
### Expected Behavior

underscoreSeparate("_nameOfThing") should result in "_name_of_thing" but actually results in "__name_of_thing"

### Actual Behaviour

String prepended with an underscore result in double underscore prepended strings

### Steps To Reproduce

1. Call underscoreSeparate with a string like "_nameOfThing"
2. Look at result, which is wrong. 

Note I looked for unit tests for this method and there aren't any or I would have commented that you could add one to a test case. The private method beneath this method is the real problem specifically these lines 

```
         private static String separateCamelCase(String name, boolean lowerCase, char separatorChar) {

...
                   if (c != separatorChar) {
                            if (last == separatorChar) {
                                newName.append(separatorChar);
                            }
                            newName.append(c);
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.2.0

wetted:
Fix special case of `NameUtils.underscoreSeparate()` where the first char is the separator char results in a double separator char prefix. 

Also overload `underscoreSeparate()` to support a lowercase result.

fixes #10140

### DIFF
--- a/core/src/main/java/io/micronaut/core/naming/NameUtils.java
+++ b/core/src/main/java/io/micronaut/core/naming/NameUtils.java
@@ -167,7 +167,18 @@ public class NameUtils {
      * @return The underscore separated version
      */
     public static String underscoreSeparate(String camelCase) {
-        return separateCamelCase(camelCase.replace('-', '_'), false, '_');
+        return underscoreSeparate(camelCase, false);
+    }
+
+    /**
+     * Returns the underscore separated version of the given camel case string, optionally with lowercase result.
+     *
+     * @param camelCase The camel case name
+     * @param lowercase true to lowercase the result
+     * @return The underscore separated version
+     */
+    public static String underscoreSeparate(String camelCase, boolean lowercase) {
+        return separateCamelCase(camelCase.replace('-', '_'), lowercase, '_');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/micronaut/core/naming/NameUtils.java
```diff
@@ -555,13 +566,19 @@ public static String decapitalize(String name) {
...
     private static String separateCamelCase(String name, boolean lowerCase, char separatorChar) {
+        StringBuilder newName = new StringBuilder();
         if (!lowerCase) {
-            StringBuilder newName = new StringBuilder();
             boolean first = true;
             char last = '0';
             for (char c : name.toCharArray()) {
                 if (first) {
-                    newName.append(c);
+                    if (c == separatorChar) {
```
sdelamo: Can 

```java
                    if (c == separatorChar) {
                        // special case where first char == separatorChar, don't double it
                        // https://github.com/micronaut-projects/micronaut-core/issues/10140
                        last = separatorChar;
                    } else {
                        newName.append(c);
                    }
```

be simplified as: 

```java
                    if (c != separatorChar) {
                        // special case where first char == separatorChar, don't double it
                        // https://github.com/micronaut-projects/micronaut-core/issues/10140
                        newName.append(c);
                    }
```

sdelamo (approved):

PR #10523, opened 2024-02-21
Fix process EachProperty configurations with abstract class for KSP

### LINKED ISSUE #10505: @EachProperty with abstract class override + KSP does not compile
### Expected Behavior

While using `@EachProperty` on a class extending an abstract class + overriding an abstract value and `KSP` for processing I should get the same behavior as with KAPT processing in the same scenario (working)

### Actual Behaviour

While using KSP, `@EachProperty` coupled with an abstract class + override results in error along the lines of:

```
Task :kspKotlin FAILED
e: [ksp] Originating element: thing
e: [ksp] C:/Users/hroth/git/micronaut-ksp-bug/src/main/kotlin/com/github/hrothwell/Example.kt:12: Prefix is required for com.github.hrothwell.MyAbstractClass
e: java.lang.IllegalStateException: Prefix is required for com.github.hrothwell.MyAbstractClass
```

### Steps To Reproduce

1. create an abstract class with an abstract value
2. create a class annotated with `@EachProperty` that implements said abstract class
3. run `./gradlew clean build`

alternative: In example application linked, uncomment [these lines](https://github.com/hrothwell/micronaut-ksp-bug/blob/main/src/main/kotlin/com/github/hrothwell/Example.kt#L28-L34) and run `./gradlew clean build` 

### Environment Information

- operating system: Mac 14.1.2 + Windows 11 (likely others)

### Example Application

https://github.com/hrothwell/micronaut-ksp-bug

### Version

4.3.2

altro3:
Fixed #10505

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationUtils.java
+++ b/core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationUtils.java
@@ -48,7 +48,7 @@ public final class ConfigurationUtils {
         } else {
             typePath = getRequiredTypePath(owningType);
         }
-        return typePath + "." + propertyName;
+        return typePath + '.' + propertyName;
     }
 
     public static String getRequiredTypePath(ClassElement classElement) {
@@ -57,7 +57,7 @@ public final class ConfigurationUtils {
 
     public static Optional<String> getTypePath(ClassElement classElement) {
         if (!classElement.hasStereotype(ConfigurationReader.class)) {
-            return Optional.empty();
+            return Optional.of(StringUtils.EMPTY_STRING);
         }
         if (classElement.isTrue(ConfigurationReader.class, ConfigurationReader.PREFIX_CALCULATED)) {
             return classElement.stringValue(ConfigurationReader.class, ConfigurationReader.PREFIX);
@@ -75,7 +75,7 @@ public final class ConfigurationUtils {
 
     private static String combinePaths(String p1, String p2) {
         if (StringUtils.isNotEmpty(p1) && StringUtils.isNotEmpty(p2)) {
-            return p1 + "." + p2;
+            return p1 + '.' + p2;
         }
         if (StringUtils.isNotEmpty(p1)) {
             return p1;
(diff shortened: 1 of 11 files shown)


altro3:
@dstepanov Hi! Look to this, pls

altro3:
CI-CD is ok, just some problems with github action:
```
2024-02-21T08:52:17.7500878Z в„№пёЏ - Java CI / Test Report (17) - 15379 tests run, 15279 passed, 100 skipped, 0 failed.
2024-02-21T08:52:17.7516214Z в„№пёЏ - Java CI / Test Report (17) - Creating check (Annotations: 0)
2024-02-21T08:52:18.0176377Z ##[error]вќЊ Failed to create checks using the provided token. (HttpError: Resource not accessible by integration)
2024-02-21T08:52:18.0180259Z ##[warning]вљ пёЏ This usually indicates insufficient permissions. More details: https://github.com/mikepenz/action-junit-report/issues/23
```

dstepanov:
Does it fix something for Java as well? Maybe you can add a test for it too

hrothwell:
Hey, opener of #10505 here. How exactly does this fix the issue? I'm a little bewildered that all it is is changing how strings are concatenated (double quote vs single quote) and using empty strings from utils. My java skills are definitely not the strongest, but here I feel there is some hidden java knowledge I am just not aware of? 😅

dstepanov:
@hrothwell It looks like the same issue that @altro3 is fixing

dstepanov:
Thanks @altro3!

hrothwell:
@dstepanov yes it is the issue @altro3 is fixing, I am wondering _how_ this fixes it. I am unsure of how just updating some string concatenation is fixing the issue. 

I am asking more for my own knowledge and curiosity. 

Main confusion: what is the difference in the end result of `string1 + "." + string2` and `string1 + '.' + string2` in java? Are these not the same?

dstepanov:
Those are just code improvements, the main fix is  `return Optional.of(StringUtils.EMPTY_STRING);`

hrothwell:
~~So the main difference then just being a static empty string vs putting in a new `""` empty string? Values are the same, so I guess the only other thing it could be is that static vs non-static?~~ 

see below

hrothwell:
Ah wait I missed a line, sorry about that. Thanks for pointing that out @dstepanov

PR #10526, opened 2024-02-22
Include serdeable metadata to internal JSON objects

dstepanov:
This fixes one of the examples from https://github.com/micronaut-projects/micronaut-serialization/issues/691
And eliminates the need for a custom deserializer for HealthStatus in Micronaut Serialization


graemerocher (approved):

sdelamo:
I think this PR creates a cycle between Serde and core and that it is a mistake.

PR #10537, opened 2024-02-23
Allow setter/getter with a different type

dstepanov:


dstepanov:
Initially, I also wanted to separate annotations into read/write, based on the accessors, but that doesn't really work well with Jackson. I still kept the API in AST.

graemerocher (approved):

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
```diff
@@ -419,32 +419,95 @@ private void buildStaticInit(ClassWriter classWriter) {
...
             beanPropertyData.name,
-            beanPropertyData.typedElement,
-            beanPropertyData.annotationMetadata,
-            beanPropertyData.typeArguments,
+            beanPropertyData.type,
             defaults,
             loadTypeMethods
         );
+
+        int typeLocal = -1;
```
yawkat, line 435: Could use Supplier.memoized

yawkat: meh probably not more readable though

inject-groovy-test/src/main/groovy/io/micronaut/ast/transform/test/AbstractBeanDefinitionSpec.groovy
```diff
@@ -243,7 +243,7 @@ abstract class AbstractBeanDefinitionSpec extends Specification {
      *
      * @return the introspection if it is correct
      **/
-    protected BeanIntrospection buildBeanIntrospection(String className, @Language("groovy") String cls) {
+    protected BeanIntrospection buildBeanIntrospection(String className, String cls) {
```
yawkat: why this change?

yawkat (approved):
does this add significant size to the generated introspections?

core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
dstepanov: That would be more confusing; the invocation is contextual, and if a variable is not needed, we don't create it.

inject-groovy-test/src/main/groovy/io/micronaut/ast/transform/test/AbstractBeanDefinitionSpec.groovy
dstepanov: Reverted

dstepanov:
Not much, considering we generate an additional argument only if the setter/getter value is different (in most cases, it shouldn't), plus it needs to be enabled.

PR #10545, opened 2024-02-26
Create synthesized annotations cache lazily

dstepanov:


inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadata.java
```diff
@@ -38,41 +38,52 @@
...
      */
     protected AbstractAnnotationMetadata() {
-        annotationMap = new ConcurrentHashMap<>(2);
-        declaredAnnotationMap = new ConcurrentHashMap<>(2);
+    }
+
+    private Map<String, Annotation> getAnnotationMap() {
+        Map<String, Annotation> map = annotationMap;
+        if (map == null) {
+            synchronized (this) { // double check
```
graemerocher, line 55: can we use a reentrant read/write lock?

inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadata.java
```diff
@@ -38,41 +38,52 @@
...
+        return annotationMap;
+    }
+
+    private Map<String, Annotation> getDeclaredAnnotationMap() {
+        Map<String, Annotation> map = declaredAnnotationMap;
+        if (map == null) {
+            synchronized (this) { // double check
+                map = declaredAnnotationMap;
+                if (map == null) {
+                    declaredAnnotationMap = new ConcurrentHashMap<>();
```
graemerocher, line 72: can we size this map?

inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadata.java
```diff
@@ -38,41 +38,52 @@
...
-        declaredAnnotationMap = new ConcurrentHashMap<>(2);
+    }
+
+    private Map<String, Annotation> getAnnotationMap() {
+        Map<String, Annotation> map = annotationMap;
+        if (map == null) {
+            synchronized (this) { // double check
+                map = annotationMap;
+                if (map == null) {
+                    annotationMap = new ConcurrentHashMap<>();
```
graemerocher, line 58: can we size this map?

graemerocher (approved):

inject/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadata.java
dstepanov: We can, considering we don't synthesize annotations anywhere this might be overkill

dstepanov: Creating a lock might be also a bit of overhead

yawkat (approved):

PR #10547, opened 2024-02-26
graal: reflect-config.json for hateoas classes

sdelamo:


dstepanov (approved):

sdelamo:
@timyates can you investigate why this fails?

timyates:
In both the regular and the native test, the recieved body json is

```
{"message":"Not Found","_links":{"self":{"href":"/jsonError","templated":false}},"_embedded":{"errors":[{"message":"Page Not Found"}]}}
```

We then call `Optional<VndError> vndErrorOptional = ex.getResponse().getBody(VndError.class);`

https://github.com/micronaut-projects/micronaut-core/blob/60525302144c5532c66f6c50699830ad2ed04b8c/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/hateoas/JsonErrorTest.java#L55

In the non-graal test, this passes, but in the native test we get back an `Optional.empty()`

Something must be up with deserialization and `VndError`... Any ideas @dstepanov ?

sdelamo:
@timyates @dstepanov I added reflect-config configuration for the `hateoas` package classes.

timyates (approved):

dstepanov (approved):

PR #10550, opened 2024-02-27
Accept inetd-style FDs

yawkat:
In #10072, I implemented support for inheriting a ServerSocketChannel from a fixed file descriptor to support systemd socket activation. This PR extends that in two ways:

- Allow specifying an "accepted FD". That means an established *connection* (SocketChannel), not a listening socket (ServerSocketChannel). This allows for inetd-style socket activation which starts a process for each connection.
- Add support for nio when fd=0 or accepted-fd=0, using System.inheritedChannel().


graemerocher (approved):

timyates (approved):
Is there a way of testing any of this? 🤔

yawkat:
only with linux i believe

PR #10552, opened 2024-02-27
Add support for inheriting method documentation

andriy-dmytruk:
Javadoc is able to inherit method documentation for overridden method: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDFAGJH. Therefore, this behavior makes sense in processing, too. The order of overridden methods had to be changed for this to work correctly.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/utils/NativeElementsHelper.java
+++ b/core-processor/src/main/java/io/micronaut/inject/utils/NativeElementsHelper.java
@@ -201,11 +201,11 @@ public abstract class NativeElementsHelper<C, M> {
         if (excludeClass(classNode)) {
             return;
         }
+        reduce(owner, collectedMethods, getMethods(classNode), cache, false, false);
         C superClass = getSuperClass(classNode);
         if (superClass != null) {
             processClassHierarchy(owner, superClass, cache, collectedMethods, includeAbstract);
         }
-        reduce(owner, collectedMethods, getMethods(classNode), cache, false, false);
         for (C anInterface : getInterfaces(classNode)) {
             processInterfaceHierarchy(owner, anInterface, cache, collectedMethods, includeAbstract);
         }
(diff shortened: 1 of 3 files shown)


graemerocher:
not sure this is wise, it means the annotation processor has to factor in multiple source inputs potentially

core-processor/src/main/java/io/micronaut/inject/utils/NativeElementsHelper.java
```diff
@@ -201,11 +201,11 @@ private void processClassHierarchy(C owner,
         if (excludeClass(classNode)) {
             return;
         }
+        reduce(owner, collectedMethods, getMethods(classNode), cache, false, false);
```
dstepanov, line 204: Why do you need this change? I think it was tested multiple times, and this should work.

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaMethodElement.java
```diff
@@ -86,6 +86,24 @@ public JavaMethodElement(JavaClassElement owningType,
...
     }
 
+    @Override
+    public Optional<String> getDocumentation() {
+        Optional<String> doc = super.getDocumentation();
+        if (doc.isPresent()) {
+            return doc;
+        }
+
+        Collection<ExecutableElement> overriddenMethods = visitorContext.getNativeElementsHelper()
```
dstepanov, line 96: Why not to use `getOverridenMethods` ?

dstepanov (changes requested):

dstepanov:
@graemerocher we already generate annotation metadata aggregated from overriden methods.

graemerocher:
Ok so probably it is fine, but it won't work if the super type is a class file not a source file

graemerocher:
Note this is different from annotations where the annotations are ready from the byte code for compiled code or the source code for uncompiled code so you may get a different result for incremental compilation

andriy-dmytruk:
> Note this is different from annotations where the annotations are ready from the byte code for compiled code or the source code for uncompiled code so you may get a different result for incremental compilation

Hmmm, yes you are right - this seems like it breaks the whole idea. I will try working around this instead.

PR #10579, opened 2024-03-06
Fix out of order writes with async reactive calls

yawkat:
In a few places, I used the pattern where reactive calls would be dispatched to the event loop if they weren't already on the loop. However, if there is a reactive call outside the event loop, and then immediately a reactive call *on* the event loop, the calls can run in the wrong order. The second call on the event loop was not delayed with an execute() call, so it could run before the first call runs.

This fix introduces a central EventLoopSerializer that takes care of this problem. In the above problem scenario, it takes care to also delay the second call even though it was submitted on the event loop.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/ReactiveClientWriter.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/ReactiveClientWriter.java
@@ -34,7 +34,7 @@ import org.reactivestreams.Subscription;
 @Internal
 final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements Subscriber<HttpContent> {
     private final Publisher<HttpContent> source;
-    private EventLoopFlow serializer;
+    private EventLoopFlow flow;
     private ChannelHandlerContext ctx;
     private Subscription subscription;
     private boolean writtenLast;
@@ -45,7 +45,7 @@ final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements
 
     @Override
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-        this.serializer = new EventLoopFlow(ctx.channel().eventLoop());
+        this.flow = new EventLoopFlow(ctx.channel().eventLoop());
         this.ctx = ctx;
         source.subscribe(this);
     }
@@ -68,7 +68,7 @@ final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements
 
     @Override
     public void onSubscribe(Subscription s) {
-        if (serializer.executeNow(() -> onSubscribe0(s))) {
+        if (flow.executeNow(() -> onSubscribe0(s))) {
             onSubscribe0(s);
         }
     }
@@ -86,7 +86,7 @@ final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements
 
     @Override
     public void onNext(HttpContent httpContent) {
-        if (serializer.executeNow(() -> onNext0(httpContent))) {
+        if (flow.executeNow(() -> onNext0(httpContent))) {
             onNext0(httpContent);
         }
     }
@@ -112,7 +112,7 @@ final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements
 
     @Override
     public void onError(Throwable t) {
-        if (serializer.executeNow(() -> onError0(t))) {
+        if (flow.executeNow(() -> onError0(t))) {
             onError0(t);
         }
     }
@@ -124,7 +124,7 @@ final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements
 
     @Override
     public void onComplete() {
-        if (serializer.executeNow(this::onComplete0)) {
+        if (flow.executeNow(this::onComplete0)) {
             onComplete0();
         }
     }
--- a/http-client/src/main/java/io/micronaut/http/client/netty/ReactiveClientWriter.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/ReactiveClientWriter.java
@@ -16,7 +16,7 @@
 package io.micronaut.http.client.netty;
 
 import io.micronaut.core.annotation.Internal;
-import io.micronaut.http.netty.EventLoopSerializer;
+import io.micronaut.http.netty.EventLoopFlow;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.handler.codec.http.HttpContent;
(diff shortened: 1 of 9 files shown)


http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
```diff
@@ -0,0 +1,152 @@
...
+ *     method.</li>
+ * </ul>
+ * Both of these invariants are guaranteed by the reactive spec, but may not apply to other use
+ * cases.
+ *
+ * @since 4.4.0
+ * @author Jonas Konrad
+ */
+@Internal
+public final class EventLoopSerializer {
```
graemerocher: can we rename to `EventLoopFlow`, the name `Serializer` can be confused with serialization

graemerocher: does it need go be public?

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
```diff
@@ -0,0 +1,152 @@
...
+        int generation = submitGeneration++;
+        if (loop.inEventLoop()) {
+            if (runGeneration == generation) {
+                if (STRICT_CHECKING) {
+                    runGeneration = generation + 1;
+                    try {
+                        delayTask.run();
+                    } finally {
+                        if (runGeneration != generation + 1 || submitGeneration != generation + 1) {
+                            throw new AssertionError("Nested call?");
```
graemerocher: throw IllegalStateException with a better message

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
```diff
@@ -0,0 +1,152 @@
...
+                runGeneration = generation + 1;
+                return true;
+            }
+            // another task already submitted, need to delay to stay serialized
+        }
+        loop.execute(new Delayed(delayTask, generation));
+        return false;
+    }
+
+    private final class Delayed implements Runnable {
```
graemerocher: can be a record

graemerocher (changes requested):

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
yawkat: no, because it's a local class (needs access to runGeneration)

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
yawkat: this is only with STRICT_CHECKING which is off by default and i only use it for debugging

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
yawkat: yes it's used in multiple places

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
```diff
@@ -0,0 +1,152 @@
...
+     * manually. Iff this method returns {@code false}, the caller should take no further action as
+     * {@code delayTask} will be run in the future.
+     *
+     * @param delayTask The task to run if it can't be run immediately
+     * @return {@code true} if the caller should instead run the task immediately
+     */
+    public boolean executeNow(Runnable delayTask) {
+        // pick a generation ID for this task.
+        int generation = submitGeneration++;
+        if (loop.inEventLoop()) {
```
dstepanov: Would it be better to implement this with a lock? Immediate execution (or execution in general) is acquiring the lock, if lock is taken the task should be scheduled. IMHO it would be much easier to understand.

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Determine whether the next step can be executed immediately. Iff this method returns
+     * {@code true}, {@code delayTask} will be ignored and the caller should call the target method
+     * manually. Iff this method returns {@code false}, the caller should take no further action as
+     * {@code delayTask} will be run in the future.
+     *
+     * @param delayTask The task to run if it can't be run immediately
+     * @return {@code true} if the caller should instead run the task immediately
+     */
+    public boolean executeNow(Runnable delayTask) {
```
dstepanov: I'm not a fan of this signature. Can we execute the runnable?

timyates (commented):
Need to address the checkstyle issues https://ge.micronaut.io/s/wut53koeap4xq/console-log/task/:http-server-netty:checkstyleMain?anchor=7&page=1

Also Japicmp is complaining about metadata binary compatibility if you run `./gradlew :http-netty:japiCmp`

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
yawkat: these ops often cant run concurrently with other event loop operations which definitely cannot use a lock

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
yawkat: there is a happy path that is only on the event loop where i want to avoid an interface call into the lambda. the goal here is to reduce cost as much as possible for the case where everything is inEventLoop

dstepanov: It looks to me like you can implement this as:

```
        if (reentrantLock.isLocked()) {
            // Schedule
        }
        try {
            reentrantLock.lock();
            // Execute
        } finally {
            reentrantLock.unlock();
        }
```

Or with `tryLock`

yawkat: no it really has to run on the event loop so that it does not conflict with other calls that happen on the loop

http-client/src/main/java/io/micronaut/http/client/netty/ReactiveClientWriter.java
```diff
@@ -34,7 +34,7 @@
 @Internal
 final class ReactiveClientWriter extends ChannelInboundHandlerAdapter implements Subscriber<HttpContent> {
     private final Publisher<HttpContent> source;
-    private EventLoop eventLoop;
+    private EventLoopFlow serializer;
```
graemerocher: rename field

http-netty/src/main/java/io/micronaut/http/netty/EventLoopSerializer.java
yawkat: also note that this pattern would not actually solve the serialization. if one task fails to lock, schedules, then a second task succeeds the lock and runs immediately before the first scheduled task runs, we still have incorrect ordering.

graemerocher (approved):

yawkat:
@timyates i dont see the japicmp issue

timyates:
> @timyates i dont see the japicmp issue

japicmp seems to be being tempremental... 🤔

timyates (approved):

PR #10596, opened 2024-03-11
Http2FrameListener-based HTTP/2 server handler implementation

yawkat:
Before this patch, HTTP/2 was implemented by taking a netty Http2MultiplexHandler, which creates a new netty channel for each HTTP/2 stream, and adding the HTTP/1.1 PipeliningServerHandler to each of those stream channels. This introduced unnecessary overhead due to the channel creation and the unnecessary features from PipeliningServerHandler.

This patch instead implements a Http2FrameListener directly, which is the lowest level netty HTTP/2 abstraction. This avoids the channel creation and simplifies some parts compared to PipeliningServerHandler (e.g. backpressure and no more need for pipelining). The implementation is split in two classes (MultiplexedServerHandler and Http2ServerHandler) in order to accommodate HTTP/3 in the future.

The new server handler has some missing features that I plan to add in future PRs, to keep this PR as minimal as possible:

- (de)compression
- h2c
- server push
- access logger
- http/3
- ping? (see comments)

When these changes are implemented, I plan to enable the new approach by default, likely in 4.4.0 still.

This new approach also removes some possibility for user pipeline customization. The only thing in the netty pipeline now, after the SSL handler, is the HTTP/2 decoder. It directly passes the HTTP/2 frames to the Micronaut Http2ServerHandler, without an intermediate step to HTTP/1.1 (Http2ServerHandler still converts the messages to HTTP/1.1 internally, so that RoutingInboundHandler can remain as-is, but the messages are not sent on any netty pipeline as before). This means e.g. logbook will not work. For this reason I still plan to keep the option of reverting to the old multiplexing approach, at least until 5.0.0. It is fairly low-maintenance anyway.


core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
```diff
@@ -0,0 +1,149 @@
...
+    private Object completion;
+    private long demand;
+    private boolean cancel;
+
+    @Override
+    public void subscribe(Subscriber<? super T> s) {
+        s.onSubscribe(new Subscription() {
+            @Override
+            public void request(long n) {
+                synchronized (DelayedSubscriber.this) {
```
graemerocher: can we use read/write locks for all these synchronised blocks?

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
```diff
@@ -0,0 +1,149 @@
...
+            if (completion != null) {
+                if (completion == COMPLETE) {
+                    downstream.onComplete();
+                } else {
+                    downstream.onError((Throwable) completion);
+                }
+            } else if (demand != 0) {
+                upstream.request(demand);
+            } else {
+                assert cancel;
```
graemerocher, line 144: this requires enabling assertions in the JDK, should this be an illegal statue exception?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyServerCustomizer.java
```diff
@@ -116,6 +116,7 @@ enum ChannelRole {
         /**
          * The channel is a channel representing an individual HTTP2 stream.
          */
+        // todo: deprecate
```
graemerocher, line 119: address this TODO

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
```diff
@@ -0,0 +1,346 @@
...
+
+    }
+
+    @Override
+    public void onSettingsRead(ChannelHandlerContext ctx, Http2Settings settings) throws Http2Exception {
+
+    }
+
+    @Override
+    public void onPingRead(ChannelHandlerContext ctx, long data) throws Http2Exception {
```
graemerocher: should we responding with pong here?

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
```diff
@@ -0,0 +1,346 @@
...
+            return true;
+        });
+    }
+
+    @Override
+    public void onWindowUpdateRead(ChannelHandlerContext ctx, int streamId, int windowSizeIncrement) throws Http2Exception {
+    }
+
+    @Override
+    public void onUnknownFrame(ChannelHandlerContext ctx, byte frameType, int streamId, Http2Flags flags, ByteBuf payload) throws Http2Exception {
```
graemerocher: should this be a warning?

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
```diff
@@ -0,0 +1,346 @@
...
+            Http2ServerHandler.Http2Stream stream = s.getProperty(streamKey);
+            if (stream != null) {
+                stream.onGoAwayRead(e);
+            }
+            return true;
+        });
+    }
+
+    @Override
+    public void onWindowUpdateRead(ChannelHandlerContext ctx, int streamId, int windowSizeIncrement) throws Http2Exception {
```
graemerocher: do we not support window resizing?

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -728,6 +729,39 @@ public void setJsonBufferMaxComponents(int jsonBufferMaxComponents) {
...
 
+    /**
+     * Prior to 4.4.0, the Micronaut HTTP server used a multi-pipeline approach for handling HTTP/2
+     * connections where every request got its own netty pipeline with HTTP/2 to HTTP/1.1
+     * converters on the pipeline. This allowed for using mostly unchanged HTTP/1.1 in the request
+     * handling and any {@link io.micronaut.http.server.netty.NettyServerCustomizer}s.
+     * <p>
+     * As of 4.4.0, this approach was replaced with a more performant HTTP/2-specific
+     * implementation. This means worse compatibility with HTTP/1.1-based code paths and
+     * customizers, however. Setting this option to {@code true} returns to the old behavior.
```
timyates, line 742: Does this mean that unless this is true, Http/1.1 services will notice a slowdown?

Are we going to keep this as `true` for the forseeable future? (I see the field is marked "TODO") 🤔

It's also something to rememeber if there's Http/2 benchmarks out there...

timyates (approved):
👍

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
yawkat: no, http/1.1 code paths are basically unchanged

i plan to make this false by default when compression and maybe server push is implemented

graemerocher (changes requested):

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
yawkat: yes but this sync block is fine for loom because it only guards memory access (no big function calls or IO)

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
yawkat: cancel is always true here because if none of the branches are hit, we would have returned above. intellij complained when i had `if (cancel)` here because it was able to figure out that cancel was always true, so i changed it to an assert.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyServerCustomizer.java
yawkat: can only deprecate when the new impl is actually enabled by default

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
yawkat: im not sure if netty already does it internally. but either way this can go into the next pr since this one is already big enough, i will put it on the list.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
yawkat: it's already handled by netty internally

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
yawkat: no, implementations must ignore unknown frames according to spec.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ServerHandler.java
yawkat: netty does it internally, i added a test.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyServerCustomizer.java
dstepanov: You can put `@NextMajorVersion`

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public void onComplete() {
+        synchronized (this) {
+            completion = COMPLETE;
+        }
+        work();
+    }
+
+    private void work() {
```
dstepanov, line 104: I would add `synchronized` to all the methods, that would avoid potential bugs

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private Subscription upstream;
+    private Subscriber<? super T> downstream;
+    private Object completion;
+    private long demand;
+    private boolean cancel;
+
+    @Override
+    public void subscribe(Subscriber<? super T> s) {
+        s.onSubscribe(new Subscription() {
```
dstepanov: Maybe you can make `DelayedSubscriber` implement `Subscription` and just pass this with synchronized methods.

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
yawkat: work() cannot be synchronized as it calls into downstream/upstream methods. it could lead to a deadlock.

core-reactive/src/main/java/io/micronaut/core/async/publisher/DelayedSubscriber.java
yawkat: done

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -196,6 +196,7 @@ public class NettyHttpServerConfiguration extends HttpServerConfiguration {
     private List<NettyListenerConfiguration> listeners = null;
     private boolean eagerParsing = DEFAULT_EAGER_PARSING;
     private int jsonBufferMaxComponents = DEFAULT_JSON_BUFFER_MAX_COMPONENTS;
+    private boolean legacyMultiplexHandlers = true; // TODO
```
graemerocher, line 199: remove this TODO

graemerocher (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyServerCustomizer.java
yawkat: i want to deprecate it in 4.5 already

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
yawkat: i want to change this in 4.5

PR #10597, opened 2024-03-11
Don't call out to micronaut-core from micronaut-graal

timyates:
When we have a Graal/Kotlin/Maven project we add micronaut-graal to the Kapt annotation processor paths.

When we moved empty arrays to static variables in https://github.com/micronaut-projects/micronaut-core/pull/10539, we added a requirement for micronaut-core to be also on the classpath.

This PR breaks that requirement, so just adding micronaut-graal to the processors is enough.

### DIFF
--- a/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
+++ b/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 original authors
+ * Copyright 2017-2024 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -55,8 +55,6 @@ import java.util.TreeSet;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import static io.micronaut.core.annotation.AnnotationUtil.ZERO_ANNOTATION_VALUES;
-
 /**
  * Generates the GraalVM reflect.json file at compilation time.
  *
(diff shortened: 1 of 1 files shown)


sdelamo (approved):

ggerbaud:
@timyates 
IMO, there is still a link to micronaut-core here : https://github.com/micronaut-projects/micronaut-core/blob/4.3.x/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java#L351

timyates:
@ggerbaud Gah, you're right... and I think [that was the line causing the issue previously with Maven, Kotlin, Kapt and Graal](https://ge.micronaut.io/s/al6cwjqpz5kxs/console-log?page=1#L477)

Something must be grabbing the wrong version of core... What doesn't help is Kapt configuration in Maven has no method of excluding dependencies, and Maven has a bizarre method of dependency conflict resolution (first past the post)

I will investigate...  And thanks for the message!  I'd have missed this, and been red-faced once 4.3.6 went out...

ggerbaud:
@timyates Not sure it helps but I had the issue with a pure Java/Maven project (no Graal, no Kotlin) and IntelliJ.
In my project, maven compilation works, but if I try to run it with IJ, IJ build fails.
See the logs below, can't say why a Graal Visitor is involved with IJ build.
[ij_build_logs.txt](https://github.com/micronaut-projects/micronaut-core/files/14564325/ij_build_logs.txt)

I reproduce it with [this small configuration](https://micronaut.io/launch?type=DEFAULT&name=demo&package=com.example&javaVersion=JDK_21&lang=JAVA&build=MAVEN&test=JUNIT&features=yaml&features=data-jpa&version=4.3.5) (just add an entity and run with IJ).

sdelamo:
@ggerbaud thanks for commenting.

timyates:
Ok, so the issue is that we seem to be ending up with 2 versions of micronaut in the Kapt processor classpath...

Something is pulling in Micronaut 4.3.1... Investigating...

timyates:
I reverted this here ☝️

timyates:
@ggerbaud So I believe the issue is that serde has a transitive dependency on core... and Maven is picking this older version of core to use in some places (annotation processing)...  I believe the way Maven resolved dependency conflicts is to pick the first one it has in it's list of dependencies. Maybe this list is getting resolved in a different order when run via intelliJ, and run from the command line...

In https://github.com/micronaut-projects/micronaut-core/pull/10539 we renamed some variables, and so Maven is unable to see the new names when the old version ends up being resolved...

We have released serialization 2.8.2 and validation 4.4.4 which update their transitive core version, so if you add:

```
    <micronaut.validation.version>4.4.4</micronaut.validation.version>
    <micronaut.serialization.version>2.8.2</micronaut.serialization.version>
```

To your POM, you should see these issues go away...

We don't have this issue with Gradle, as Gradle resolves conflicts by picking the later of the two dependencies

ggerbaud:
@timyates I think there is still some outdated transitive dependencies.

micronaut-platform 4.3.5 -> micronaut-data-processor 4.6.2 -> micronaut-core-bom 4.3.1
micronaut-platform 4.3.5 -> micronaut-openapi 6.5.1 -> micronaut-core-bom 4.3.1

timyates:
Yeah, so until we can get releases out for them to update them you can exclude the core dependency from the relevant `annotationProcessorPaths` (as we do with `micronaut-inject` currently)

We're investigating what we can do for a better solution (as this doesn't work for Kotlin projects, as the Kapt processor has no concept of exclusion) :-(

PR #10601, opened 2024-03-12
Handle parsing of form attributes with no value

### LINKED ISSUE #10446: Parsing of application/x-www-form-urlencoded is broken, can result in Internal Server Error.
### Expected Behavior

Parsing should conform to https://url.spec.whatwg.org/#application/x-www-form-urlencoded.

### Actual Behaviour

Parsing does not support the case when a single sequence is just `key`. Moreover, if this `key` is followed by more sequences an exception is thrown and Internal Server Error is returned.

### Steps To Reproduce

Just run the test.

### Environment Information

Windows.
JDK Corretto 17.0.10.

### Example Application

https://github.com/marek-stanulewicz/micronaut-issues/tree/www-form-urlencoding-parsing

### Version

4.2.4

jeremyg484:
`MicronautHttpData.AttributeImpl` is updated to provide an implementation of the setValue method. This implementation is needed in the corner cases where Netty's `HttpPostStandardRequestDecoder` successfully parses an attribute with no value such as in the body "a&b=2".

Tests are added to more thoroughly test the various forms of x-www-form-urlencoded bodies as specified by https://url.spec.whatwg.org/#application/x-www-form-urlencoded. These tests include scenarios where we know and expect that `HttpPostStandardRequestDecoder` will currently fail to parse an attribute with a name but no value - namely when that attribute is the last one in a given POST body. If `HttpPostStandardRequestDecoder` is patched in the future, then these expected parsing failures can be moved into the success scenario instead.

This partially resolves #10446.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
@@ -513,7 +513,7 @@ public abstract sealed class MicronautHttpData<D extends HttpData> extends Abstr
 
         @Override
         public void setValue(String value) throws IOException {
-            throw new UnsupportedOperationException();
+            setContent(Unpooled.copiedBuffer(value.toCharArray(), factory.characterEncoding));
         }
 
         @Override
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
@@ -513,7 +513,7 @@ public abstract sealed class MicronautHttpData<D extends HttpData> extends Abstr
 
         @Override
         public void setValue(String value) throws IOException {
-            setContent(Unpooled.copiedBuffer(value.toCharArray(), factory.characterEncoding));
+            setContent(Unpooled.copiedBuffer(value, factory.characterEncoding));
         }
 
         @Override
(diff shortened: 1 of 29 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
```diff
@@ -513,7 +513,7 @@ public String getValue() throws IOException {
 
         @Override
         public void setValue(String value) throws IOException {
-            throw new UnsupportedOperationException();
+            setContent(Unpooled.copiedBuffer(value.toCharArray(), factory.characterEncoding));
```
yawkat: no need for the toCharArray

yawkat (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/MicronautHttpData.java
jeremyg484: Ahh, great, thanks. I somehow missed the `CharSequence` variation.

[c5b0e8d](https://github.com/micronaut-projects/micronaut-core/pull/10601/commits/c5b0e8dcc5a481c316c592e6fed66232dcfa8ffb)

sdelamo (commented):
Why are we using a reactive HTTP Client for these tests?

sdelamo (approved):
[I have changed to use the `BlockingHttpClient`](https://github.com/micronaut-projects/micronaut-core/pull/10601/commits/fa39a6ee159319c0cc761746304e1a33b6c3e364)

graemerocher (approved):

jeremyg484:
Note that I have submitted a PR to Netty to fix the remaining cases that don't parse correctly: https://github.com/netty/netty/pull/13904

PR #10609, opened 2024-03-14
Let event loop threads implement reactor NonBlocking

yawkat:
This patch updates the default netty ThreadFactory to have the threads implement reactor NonBlocking. Any reactor blocking operations will then throw when they're done on an event loop thread.


http-netty/src/main/java/io/micronaut/http/netty/channel/NettyThreadFactoryConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+ * @param reactorNonBlocking If {@code true} (the default), netty event loop threads will implement
+ *                           project reactor {@link reactor.core.scheduler.NonBlocking} by default.
+ *                           This will make reactor blocking operations throw an exception.
+ * @since 4.4.0
+ * @author Jonas Konrad
+ */
+@ConfigurationProperties("netty.default.thread-factory")
+@BootstrapContextCompatible
+@Internal
+public record NettyThreadFactoryConfiguration(
```
graemerocher: do we need a whole new configuration properties class for this? Can it not go in `NettyGlobalConfiguration`?

yawkat: sure

yawkat:
this will likely break security. in the meeting i suggested we move the security filter to blocking for now.

graemerocher:
security is not implemented correctly if it is blocking and if it is blocking needs to be declared as such.

graemerocher (approved):

http-netty/src/main/java/io/micronaut/http/netty/channel/NettyThreadFactory.java
```diff
@@ -60,6 +63,28 @@ public class NettyThreadFactory {
...
 
+    private final NettyGlobalConfiguration configuration;
+
+    /**
+     * Create a new netty ThreadFactory factory.
+     *
+     * @param configuration The configuration
+     * @since 4.4.0
+     */
+    public NettyThreadFactory(NettyGlobalConfiguration configuration) {
```
sdelamo, line 76: should not we annotate the new constructor with `@Inject`?

sdelamo (changes requested):
I created several suggestions: 

- https://github.com/micronaut-projects/micronaut-core/pull/10622
- https://github.com/micronaut-projects/micronaut-core/pull/10621
- https://github.com/micronaut-projects/micronaut-core/pull/10620

http-netty/src/main/java/io/micronaut/http/netty/channel/NettyThreadFactory.java
sdelamo: see: https://github.com/micronaut-projects/micronaut-core/pull/10621

sdelamo (approved):

PR #10610, opened 2024-03-14
Manage Kotlin plugins in the catalog

timyates:
With the build plugin 6.7.0 we can add managed plugins to the catalog, and these can be used in downstream modules.

This means we can stop defining the version of Kotlin in most modules, and fight to keep them all updated, and instead use:

```
alias(mn.plugins.kotlin.jvm)
```

or

```
alias(mn.plugins.ksp)
```

### DIFF
--- a/aop/build.gradle
+++ b/aop/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id "io.micronaut.build.internal.convention-core-library"
-    id "org.jetbrains.kotlin.jvm"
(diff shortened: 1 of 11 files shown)


sdelamo (approved):

melix (approved):

graemerocher:
do these not belong in the Kotlin module?

melix:
That's the same problem as with Groovy, this would create a circular dependency.

PR #10611, opened 2024-03-14
Update title and description

sdelamo:
The idea is to convey the documentation is for the core of the framework. 

CleanShot 2024-03-14 at 16 25 47@2x

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 projectVersion=4.3.11-SNAPSHOT
 projectGroupId=io.micronaut
-projectDesc=Natively Cloud Native
-title=Micronaut Framework
+projectDesc=Micronaut core components: IoC, AOP, Configuration
+title=Micronaut Core
 githubSlug=micronaut-projects/micronaut-core
 docsRepository=micronaut-projects/micronaut-docs
 testsdir=inject-groovy/src/test/groovy/io/micronaut/docs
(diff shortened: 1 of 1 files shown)


andriy-dmytruk (approved):

timyates (approved):

gradle.properties
```diff
@@ -1,7 +1,7 @@
 projectVersion=4.3.11-SNAPSHOT
 projectGroupId=io.micronaut
-projectDesc=Natively Cloud Native
-title=Micronaut Framework
+projectDesc=Micronaut core components: IoC, AOP, Configuration
```
jeremyg484: Perhaps it would be better to leave out the ": IoC, AOP, Configuration" part of the description? Somebody completely new seeing that at the top when they're looking for the HTTP server and client docs, for example, might get the impression they are in the wrong place.

I think either cut it, or expand it to something like "IoC, AOP, Configuration and Core APIs"

wetted: How about something like "Core components supporting the Micronaut Framework." That way we don't have to remember to change it if something we list otherwise gets moved out (or in).

jeremyg484: I also think it might be good to briefly explain the distinction between the Framework and Core right away in the Introduction.

alvarosanchez: +1 to @wetted's suggestion

sdelamo:
```suggestion
projectDesc=Core components supporting the Micronaut Framework
```

andriy-dmytruk (approved):

wetted (approved):

jeremyg484 (approved):

PR #10626, opened 2024-03-19
Build HTTP request in declarative client Publisher

jeremyg484:
`HttpClientIntroductionAdvice` is refactored to ensure that binding to a new HTTP request will happen within the reactive context when using a `Publisher` return type in a declarative HTTP client.

This fixes the issue of the same `MutableHttpRequest` instance being re-used on retry attempts when the client is annotated with `@Retryable`, which can lead to an incorrect state when used in conjunction with client filters that modify the request, for example by adding header fields.

This makes the reactive behavior consistent with the blocking behavior.

Tests are enhanced to check that the expected number of new request instances are constructed and bound when using `@Retryable` with the declarative HTTP client.

**Background:**

The prior inconsistency in behavior was found  while searching for the root of a Micronaut Security issue where it was shown that the Authorization header was being added multiple times when using `@Retryable` in conjunction with the declarative http client and the `ClientCredentialsHttpClientFilter`:
https://github.com/micronaut-projects/micronaut-security/issues/1619

The additional header values only get sent when returning a reactive type from the declarative client method. This was traced to the difference in the way `DefaultRetryInterceptor` implements retry for blocking methods vs reactive methods. In the blocking case, the retryable method gets re-invoked, whereas in the reactive case the returned `Publisher` is simply re-subscribed. This results in any state that is captured from outside of the  Publisher lambda(s) being re-used in subsequent retries.

`HttpClientIntroductionAdvice` was building and binding the HTTP request prior to constructing the `Publisher` to be returned in the reactive return type code path, resulting in the same `MutableHttpRequest` instance from the first request being used on all subsequent retry attempts.

**Note For Reviewers**
In order to detangle the request construction and binding to make it able to be done within a Publisher, I broke the prior `intercept` method of `HttpClientIntroductionAdvice` down into a series of smaller methods. This is making the diff on that class a bit of a mess.


http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -21,10 +21,7 @@
 import io.micronaut.aop.MethodInvocationContext;
 import io.micronaut.context.annotation.BootstrapContextCompatible;
 import io.micronaut.context.exceptions.ConfigurationException;
-import io.micronaut.core.annotation.AnnotationMetadata;
-import io.micronaut.core.annotation.AnnotationValue;
-import io.micronaut.core.annotation.Internal;
-import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.annotation.*;
```
yawkat: no star imports

http-client/src/test/groovy/io/micronaut/http/client/retry/HttpClientRetryWithFallbackSpec.groovy
```diff
@@ -49,27 +52,33 @@ class HttpClientRetryWithFallbackSpec extends Specification{
...
 
         when:"The threshold can never be met"
+        countFilter.requests.clear()
         controller.countThreshold = Integer.MAX_VALUE
         controller.count = 0
         countClient.getCount()
 
         then:"The fallback is called"
         countClient.getCount() == 9999
+        countFilter.requests.size() == 12 //TODO - This seems like a bug in the way blocking retry works with fallback, but is the same as the behavior before HttpClientIntroductionAdvice refactoring
```
yawkat, line 75: how much do you expect here?

yawkat (approved):

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
jeremyg484: Thanks. IntelliJ settings got borked again. Fixed [20a991d](https://github.com/micronaut-projects/micronaut-core/pull/10626/commits/20a991d36b1748c95b3af13aed956cafe668ac0c)

http-client/src/test/groovy/io/micronaut/http/client/retry/HttpClientRetryWithFallbackSpec.groovy
jeremyg484: I was thinking it should be 6, the same as it is now in the reactive case below this.

My logic is: 1 initial failed attempt, and 5 failing retries, then it falls back to CountClientFallback, which does not actually initiate an HTTP request.

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -578,4 +691,24 @@ private String appendQuery(String uri, Map<String, List<String>> queryParams) {
...
+
+        static RequestBinderResult withRequest(@NonNull MutableHttpRequest<?> request) {
+            Objects.requireNonNull(request, "Bound HTTP request must not be null");
+            return new RequestBinderResult(request, null, false);
+        }
+
+        static RequestBinderResult withErrorResult(@Nullable Object errorResult) {
+            return new RequestBinderResult(null, errorResult, true);
+        }
+    }
```
timyates:
```suggestion
    private record RequestBinderResult(
        @Nullable MutableHttpRequest<?> request,
        @Nullable Object errorResult,
        boolean isError
    ) {

        static RequestBinderResult withRequest(@NonNull MutableHttpRequest<?> request) {
            Objects.requireNonNull(request, "Bound HTTP request must not be null");
            return new RequestBinderResult(request, null, false);
        }

        static RequestBinderResult withErrorResult(@Nullable Object errorResult) {
            return new RequestBinderResult(null, errorResult, true);
        }
    }
```

Constructor is implied in `record` I think... 🤔

timyates (approved):

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
jeremyg484: Yes, it is. Thanks. I missed that IntelliJ didn't remove it when refactored from a class.

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -171,312 +174,421 @@ public Object intercept(MethodInvocationContext<Object, Object> context) {
...
+            Object definedValue = getValue(argument, context, parameters);
+
+            if (definedValue != null) {
+                final ClientArgumentRequestBinder<Object> binder = (ClientArgumentRequestBinder<Object>) binderRegistry
+                    .findArgumentBinder((Argument<Object>) argument)
+                    .orElse(defaultBinder);
+                ArgumentConversionContext conversionContext = ConversionContext.of(argument);
+                binder.bind(conversionContext, uriContext, definedValue, request);
+                if (conversionContext.hasErrors()) {
+                    return Optional.of(interceptedMethod.handleException(new ConversionErrorException(argument, conversionContext.getLastError().get())));
```
timyates:
```suggestion
                    return conversionContext.getLastError().map(e -> interceptedMethod.handleException(new ConversionErrorException(argument, e)));
```

Does this appease Sonar (and do the same thing?)

timyates (approved):

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
jeremyg484: Yes, I think that's better. Thanks!

http-client/src/test/groovy/io/micronaut/http/client/retry/HttpClientRetryWithFallbackSpec.groovy
sdelamo: @jeremyg484 please, investigate it and fix it in a separate PR.

timyates: @jeremyg484 @sdelamo It's running it twice

Once on line 71 and then again in the `then` block on line 74:

https://github.com/micronaut-projects/micronaut-core/blob/988c60262c92129db4566a5f11d3548b99bd4bed/http-client/src/test/groovy/io/micronaut/http/client/retry/HttpClientRetryWithFallbackSpec.groovy#L71-L75

If you change it to:

```groovy
        result = countClient.getCount()

        then: "The fallback is called"
        result == 9999
        countFilter.requests.size() == 6
```

Then it's 6 as expected

jeremyg484: 🤦‍♂️ @timyates good catch!

timyates: I love a good mystery 🕵️

PR #10628, opened 2024-03-20
Update netty to 4.1.107.Final

yawkat:


sdelamo (changes requested):
why 3.8.x and not 3.10.x? It seems the sonar task fails.

yawkat:
we have an internal customer on 3.8. we probably dont need a release of 3.8.x but i just want to make sure it works

graemerocher (approved):

sdelamo:
I will try to make the build pass.

yawkat:
thanks!

PR #10634, opened 2024-03-21
Remove build time initialisation of user metadata for GraalVM

### LINKED ISSUE #10630: 4.4.x broken in certain cases on native image (with enums)
Seems to be caused by https://github.com/micronaut-projects/micronaut-core/commit/ff72307850ad4b4f7bf932f59f742c474b22852c

### Expected Behavior

Should build image

### Actual Behaviour

```
Fatal error: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: An object of type 'io.micronaut.logging.LogLevel' was found in the image heap. This type, however, is marked for initialization at image run time for the following reason: classes are initialized at run time by default.
```

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

https://github.com/micronaut-projects/micronaut-core/commit/ff72307850ad4b4f7bf932f59f742c474b22852c

graemerocher:
Seems with strict image heap it is not safe to initialise anything that could represent user code. Previously we could get away with initialising our generated classes, but since this reference user classes the viral nature of build time init results in initialising user code. 

This manifested in dev builds failing with an error on the `LogLevel` enum, I attempted to mark this is build time init but there are 2 problems with this:

1. It means any user defined enum will need to be marked as build time init
2. Once fixed I hit another problem with expression language expressions which also need to be build time initialised. These are again user defined and we cannot know what the are.

So unfortunately the only solution I see going forward is to completely remove the build time init or our generated metadata. This change does this and in local benchmarks does not seem to have a major impact on memory or startup, however does increase the image size by 2mb.

Fixes #10630

### DIFF
--- a/.github/workflows/graalvm-dev.yml
+++ b/.github/workflows/graalvm-dev.yml
@@ -4,9 +4,18 @@
 #
 # and edit them there. Note that it will be sync'ed to all the Micronaut repos
 name: GraalVM Dev CI
+#on:
+#  schedule:
+#    - cron: "0 1 * * 1-5" # Mon-Fri at 1am UTC
 on:
-  schedule:
-    - cron: "0 1 * * 1-5" # Mon-Fri at 1am UTC
+  push:
+    branches:
+      - master
+      - '[1-9]+.[0-9]+.x'
+  pull_request:
+    branches:
+      - master
+      - '[1-9]+.[0-9]+.x'
 jobs:
   build_matrix:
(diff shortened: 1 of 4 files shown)


sdelamo (approved):

dstepanov:
We have some issues with enums values when the class is not on the classpath, I was thinking we can replace enums instance values with string or do try-catch aproach like classes.

sdelamo:
@dstepanov what enums where you thinking of?

PR #10635, opened 2024-03-21
Correct expectations in retry with fallback spec

jeremyg484:
The blocking test scenario in the HttpClientRetryWithFallbackSpec is corrected to not mistakenly initiate the HTTP requests twice when verifying that fallback works as expected.

Thanks @timyates for the spot. :)


sdelamo (approved):

timyates (approved):
😎

PR #10659, opened 2024-03-28
remove DefaultCookieDecoder

n0tl3ss:
#10435 remove DefaultCookieDecoder from services because its broken for parsing request cookies

PR #10661, opened 2024-03-28
fix: Ensure Netty Cookie Encoder/Decoder take precedence

### LINKED ISSUE #10435: Cookie Parsing Failure
### Expected Behavior

Parsing multiple cookies on a single Cookie HTTP Header should see all the cookies.

### Actual Behaviour

Cookies are not parsed correctly due to two bugs.
1) Even when the `NettyLaxServerCookieDecoder` is present, it's not used because of the presence of the `META-INF/services/io.micronaut.http.cookie.ServerCookieDecoder` within the micronaut-http library that points to `io.micronaut.http.cookie.DefaultServerCookieDecoder`. This service files should not be present, since all the loading code already has this class as a fallback, so having them present can cause problems since the loading order is indeterminate.

2) While the `NettyLaxServerCookieDecoder` does decode cookies properly, the fallback, `DefaultServerCookieDecoder` does not. It relies on the `java.net.HttpCookie.parse` methods to perform parsing. However, these methods are based on parsing the `Set-Cookie` header, not the `Cookie` header. The format of these headers are quite different. `Set-Cookie` is `Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnly`, while `Cookie` is `Cookie: <cookie-name-1>=<cookie-value-1>; <cookie-name-2>=<cookie-value-2>; <cookie-name-3>=<cookie-value-3>`. This means that if you receive multiple cookies in a `Cookie` header, the `HttpCookie` code treats it as one cookie with multiple parameters, instead of multiple cookies.

### Steps To Reproduce

_No response_

### Environment Information

- JDK 17

### Example Application

_No response_

### Version

4.3.3

sdelamo:
See: https://github.com/micronaut-projects/micronaut-core/issues/10435
Close: https://github.com/micronaut-projects/micronaut-core/pull/10659

This PRs ensure Netty Cookie Encoder/Decoder are loaded before DefaultServerCookieEncoder Decoders

It allows also for users to provider their own implementation which may take precedence.

### DIFF
--- a/http-netty/src/test/groovy/io/micronaut/http/netty/cookies/NettyLaxClientCookieEncoderSpec.groovy
+++ b/http-netty/src/test/groovy/io/micronaut/http/netty/cookies/NettyLaxClientCookieEncoderSpec.groovy
@@ -1,8 +1,9 @@
 package io.micronaut.http.netty.cookies
 
+import io.micronaut.core.order.OrderUtil
 import io.micronaut.http.cookie.ClientCookieEncoder
 import io.micronaut.http.cookie.Cookie
-
+import io.micronaut.http.cookie.DefaultClientCookieEncoder
 import spock.lang.Specification
 
 class NettyLaxClientCookieEncoderSpec extends Specification {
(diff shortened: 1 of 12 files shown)


yawkat:
seems a bit overkill for an interface that is really only supposed to have one way of implementing it

PR #10663, opened 2024-03-28
h2c support for new Http2ServerHandler

yawkat:
This PR implements h2c support for the new non-multiplexed Http2ServerHandler.

Again this can go in 4.4.0 or 4.5.0, I will only turn on the handler by default in 4.5 anyway.


sdelamo (changes requested):
Can you write a test?

yawkat:
it's covered by the existing H2cSpec when the legacy flag is disabled

dstepanov (approved):

PR #10685, opened 2024-04-05
Set Accept header correctly in declarative client

jeremyg484:
`HttpClientIntroductionAdvice` is updated to ensure the Accept header is correctly set in all cases.

Tests are added to verify the header is being set as expected.

**Background:**
I discovered a bug in the recently refactored `HttpClientIntroductionAdvice` while debugging https://github.com/micronaut-projects/micronaut-discovery-client/pull/604.

The Accept header was only being set correctly for the streaming publisher branch of the code, due to a mistake I made in the refactoring, and none of the existing tests were checking for the header. :/

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
@@ -400,6 +400,18 @@ public class HttpClientIntroductionAdvice implements MethodInterceptor<Object, O
         // The original query can be added by getting it from the request.getUri() and appending
         request.uri(URI.create(appendQuery(uri, uriContext.getQueryParameters())));
 
+        final MediaType[] acceptTypes;
+        Collection<MediaType> accept = request.accept();
+        if (accept.isEmpty()) {
+            String[] consumesMediaType = context.stringValues(Consumes.class);
+            if (ArrayUtils.isEmpty(consumesMediaType)) {
+                acceptTypes = DEFAULT_ACCEPT_TYPES;
+            } else {
+                acceptTypes = MediaType.of(consumesMediaType);
+            }
+            request.accept(acceptTypes);
+        }
+
         if (body != null && !request.getContentType().isPresent()) {
             MediaType[] contentTypes = MediaType.of(context.stringValues(Produces.class));
             if (ArrayUtils.isEmpty(contentTypes)) {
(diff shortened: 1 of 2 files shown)


sdelamo (changes requested):
if this is a bug fix set 4.3.x as the base of the PR.

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -400,6 +400,18 @@ private RequestBinderResult bindRequest(MethodInvocationContext<Object, Object>
         // The original query can be added by getting it from the request.getUri() and appending
         request.uri(URI.create(appendQuery(uri, uriContext.getQueryParameters())));
 
+        final MediaType[] acceptTypes;
+        Collection<MediaType> accept = request.accept();
+        if (accept.isEmpty()) {
```
sdelamo, line 405: should we use `CollectionUtils::isEmpty` to protect against a NPE.

http-client/src/test/groovy/io/micronaut/http/client/ClientIntroductionAdviceSpec.groovy
```diff
@@ -132,6 +158,35 @@ class ClientIntroductionAdviceSpec extends Specification {
...
+        @Get(value = "/xml", produces = MediaType.APPLICATION_XML)
+        String xml(HttpRequest<?> request) {
+            if (!request.accept().contains(MediaType.APPLICATION_XML_TYPE)) {
+                throw new IllegalStateException("Accept should be set to XML")
+            }
+            return "<answer>success</answer>"
+        }
+
+        @Get(value = "/xml-annotated", produces = MediaType.APPLICATION_XML)
+        String xmlAnnotated(HttpRequest<?> request) {
```
sdelamo: Delete this method.

http-client/src/test/groovy/io/micronaut/http/client/ClientIntroductionAdviceSpec.groovy
```diff
@@ -194,6 +249,21 @@ class ClientIntroductionAdviceSpec extends Specification {
...
+    @Requires(property = 'spec.name', value = 'ClientIntroductionAdviceSpec')
+    static interface AcceptTypeClient {
+        @Get
+        String index()
+
+        @Get(value = "/xml", consumes = MediaType.APPLICATION_XML)
+        String xml()
+
+        @Consumes(MediaType.APPLICATION_XML)
+        @Get("/xml-annotated")
```
sdelamo:
```suggestion
        @Get("/xml")
```

sdelamo (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
jeremyg484: I don't think we need to in this case - the contract of HttpRequest.accept() is to always return "A list of zero or many MediaType instances"

http-client/src/test/groovy/io/micronaut/http/client/ClientIntroductionAdviceSpec.groovy
jeremyg484: Done in [6c3106b](https://github.com/micronaut-projects/micronaut-core/pull/10685/commits/6c3106b2a3d55b0587eb04188547d10d5980fe63)

http-client/src/test/groovy/io/micronaut/http/client/ClientIntroductionAdviceSpec.groovy
jeremyg484: Done in [6c3106b](https://github.com/micronaut-projects/micronaut-core/pull/10685/commits/6c3106b2a3d55b0587eb04188547d10d5980fe63)

jeremyg484:
> if this is a bug fix set 4.3.x as the base of the PR.

@sdelamo The bug does not exist in 4.3.x. It was unfortunately introduced in the refactoring I did in https://github.com/micronaut-projects/micronaut-core/pull/10626, which we confined to 4.4.x.

yawkat (approved):
I don't think we should send any accept header by default, but if that's the previous behavior, fine

jeremyg484:
> I don't think we should send any accept header by default, but if that's the previous behavior, fine

Yes, I found that a bit surprising at first. The logic for setting the header is unchanged from 4.3.x, it's just moved to a different place.

PR #10704, opened 2024-04-12
KSP: Fix computation of Kotlin Defaults Mask in Data Classes with 32+ parameters.

temofey1989:
Related to the problem in Micronaut Serialization: https://github.com/micronaut-projects/micronaut-core/issues/10730


dstepanov (approved):

sdelamo (approved):

temofey1989:
@sdelamo 
Thanks for the merge!

One question...
Should I cherry-pick this fix to `4.4.x` and `4.5.x` branches? Or there is some automate which propagates this change into these branches?

PR #10723, opened 2024-04-16
bug: Writeable responses not being offloaded since 4.0.0

timyates:
In the documentation for [Writing Response Data](https://docs.micronaut.io/latest/guide/index.html#serverIO) we have a section for Writeables which states:

> When returning a [Writable](https://docs.micronaut.io/latest/api/io/micronaut/core/io/Writable.html), the blocking I/O operation is shifted to the I/O thread pool so the Netty event loop is not blocked.

This is true in Micronaut 3.x, but does not happen in certain cases for 4.x

This PR provides a test which passes when added to the 3.10.x branch, but fails on all the current 4.x.x branches

This was discovered whilst investigating fixing views https://github.com/micronaut-projects/micronaut-views/issues/757 as the views were not being delegated to an IO thread pool despite the current documentation there as well

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
@@ -80,6 +80,8 @@ import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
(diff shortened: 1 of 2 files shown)


timyates:
@yawkat awesome 🥹

@sdelamo should I backport this to 4.3.x once merged?

sdelamo:
> @yawkat awesome 🥹
> 
> @sdelamo should I backport this to 4.3.x once merged?

Maybe we can cherry pick the commit.

sdelamo (approved):

Issue #10737, opened 2024-04-18
MN:4.4.0 - Changed behavior for @JsonGetter in @Introspected-annotated classes


roar-skinderviken:
### Expected Behavior

With the example application and MN:4.4.0, I expect this JSON as output from the controller.
```
{"afterCare":[{"klassifisering":{"regionKode":"01"},"stats":{"SomeField1":0,"SomeField2":0}}]}
```

### Actual Behaviour

This is a simplistic reconstruction of what happened: 
```
{"afterCare":[{"klassifisering":{"regionKode":"01"},"stats":{"SomeField1":0,"SomeField2":0}}]}
```
changed to
```
{"afterCare":[{"klassifisering":{"regionKode":"01"},"stats":{"shouldNotAppearInJson":{"SomeField1":0,"SomeField2":0}}}]}
```
with MN:4.4.0.

### Steps To Reproduce

1. In the example application, switch to MN:4.4.0 in gradle.properties.
2. Start the application with `./gradlew clean run`
3. Open http://localhost:8080/get-aggregation

Observe that JSON has changed to
```
{"afterCare":[{"klassifisering":{"regionKode":"01"},"stats":{"shouldNotAppearInJson":{"SomeField1":0,"SomeField2":0}}}]}
```

Please note that removing `@Introspected` fixes the issue.

### Environment Information

JDK 21
Kotlin 1.9.23
Ubuntu 22.04

### Example Application

https://github.com/roar-skinderviken/MN-4.4-jackson-issue

### Version

4.4.0

timyates:
@dstepanov Any ideas?

dstepanov:
It's probably some KSP stuff. I have added similar test to Micronaut Serialization and fixed some bug https://github.com/micronaut-projects/micronaut-serialization/pull/826

dstepanov:
Looks like the annotation is not retained in the property annotations

roar-skinderviken:
Seems like you're right about KSP @dstepanov. When replacing it with `kotlin("kapt") version "1.9.23"`, returned JSON is as expected.

timyates:
It might be us, as changing the field name in the reproducer to be different to the JsonGetter name seems to work

```diff
diff --git a/gradle.properties b/gradle.properties
index ab4cf7a..c3fd1bb 100644
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,3 @@
-micronautVersion=4.3.8
[...]

PR #10750, opened 2024-04-22
KSP: Correct replacing annotations

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/10737


graemerocher (approved):

timyates (approved):
Tested against the reproducer in https://github.com/micronaut-projects/micronaut-core/issues/10737 and it works!

🦸

PR #10759, opened 2024-04-24
The health check endpoint configuration must be more explicit on regards of sensitive

shadai-rafael:
The health check endpoint immediately returns status code 401 Unathorized if sensitive is set to true.


sdelamo (approved):
Thanks, I added a test to match the documentation statement.

PR #10760, opened 2024-04-24
Update conditionalBeans.adoc to clarify how to depend on a conditional bean

### LINKED ISSUE #10757: Docs: not entirely clear what happens if you inject a bean with an unsatisfied requirement
### Issue description

From reading the docs I understand that I can add requirements to a bean so that it may not be available at runtime. What's not really clear is what happens if I inject a bean that has an unsatisfied requirement. Is the reference set to null? Is there an exception? If I want to say "give me this bean if it's available, but, not if it isn't" what am I supposed to do here? Presumably injecting lists of interfaces means it just won't appear in the list, but what if I don't need an extra interface and list? Maybe use `Optional<BEAN>`? The docs seem to think this is obvious but for people who haven't worked with Spring or similar frameworks, it isn't especially obvious.

mikehearn:
Fixes #10757

### DIFF
--- a/src/main/docs/guide/ioc/conditionalBeans.adoc
+++ b/src/main/docs/guide/ioc/conditionalBeans.adoc
@@ -1,6 +1,6 @@
 At times, you may want a bean to load conditionally based on various potential factors including the classpath, the configuration, the presence of other beans, etc.
 
-The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean.
+The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean. If a bean is used somewhere but it's requirements aren't met then `NoSuchBeanException` will be thrown. That's probably not what you want. If you want a bean to conditionally depend on a bean then mark the injection site as `@Nullable` (this is done automatically if you're using Kotlin and declare the bean with a `?`), or alternatively declare it as `Optional<Bean>`. Finally you can inject a container of interfaces, in which case unloaded beans just won't appear in the collection.
 
 Consider the following example:
 
--- a/src/main/docs/guide/ioc/conditionalBeans.adoc
+++ b/src/main/docs/guide/ioc/conditionalBeans.adoc
@@ -1,6 +1,18 @@
 At times, you may want a bean to load conditionally based on various potential factors including the classpath, the configuration, the presence of other beans, etc.
 
-The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean. If a bean is used somewhere but it's requirements aren't met then `NoSuchBeanException` will be thrown. That's probably not what you want. If you want a bean to conditionally depend on a bean then mark the injection site as `@Nullable` (this is done automatically if you're using Kotlin and declare the bean with a `?`), or alternatively declare it as `BeanProvider<Bean>`. Finally you can inject a container of interfaces, in which case unloaded beans just won't appear in the collection.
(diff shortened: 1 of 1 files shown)


src/main/docs/guide/ioc/conditionalBeans.adoc
```diff
@@ -1,6 +1,6 @@
 At times, you may want a bean to load conditionally based on various potential factors including the classpath, the configuration, the presence of other beans, etc.
 
-The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean.
+The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean. If a bean is used somewhere but it's requirements aren't met then `NoSuchBeanException` will be thrown. That's probably not what you want. If you want a bean to conditionally depend on a bean then mark the injection site as `@Nullable` (this is done automatically if you're using Kotlin and declare the bean with a `?`), or alternatively declare it as `Optional<Bean>`. Finally you can inject a container of interfaces, in which case unloaded beans just won't appear in the collection.
```
graemerocher: `Optional<Bean>` shouldn't really be used although it is supported. Generally `Optional` should only be used for return values. You can use `BeanProvider<Bean>` which provides the same functionality as `Optional`

mikehearn: Done, PTAL

mikehearn:
@graemerocher Done, PTAL.

sdelamo (approved):

src/main/docs/guide/ioc/conditionalBeans.adoc
```diff
@@ -1,6 +1,6 @@
 At times, you may want a bean to load conditionally based on various potential factors including the classpath, the configuration, the presence of other beans, etc.
 
-The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean.
+The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean. If a bean is used somewhere but it's requirements aren't met then `NoSuchBeanException` will be thrown. That's probably not what you want. If you want a bean to conditionally depend on a bean then mark the injection site as `@Nullable` (this is done automatically if you're using Kotlin and declare the bean with a `?`), or alternatively declare it as `BeanProvider<Bean>`. Finally you can inject a container of interfaces, in which case unloaded beans just won't appear in the collection.
```
sdelamo:
```suggestion
The link:{api}/io/micronaut/context/annotation/Requires.html[Requires] annotation provides the ability to define one or many conditions on a bean.

If you declare an injection point, for example, `Bean a`, but no bean of type `Bean` can be instantiated, the framework throws a `NoSuchBeanException`.
The framework cannot inject a bean into an injection point if the bean requirements specified by `@Requires` are not fulfilled.

If you want to declare an injection point as Optional, you have several options:

* Mark the injection point as `@Nullable`. In Kotlin, nullable type names have `?` at the end; you don't need to specify the `@Nullable` annotation.
* Declare it as `BeanProvider<Bean>`. Learn about <<types, Injectable Container Types>>.

You can also inject a collection of beans. For example, you can inject `List<Bean>`. The collection is empty if no bean of type `Bean` fulfills the injection point.

=== @Requires Example
```

sdelamo (changes requested):

src/main/docs/guide/ioc/conditionalBeans.adoc
sdelamo: @mikehearn what do you think about 👆🏻 ?

mikehearn:
@sdelamo OK, accepted.

graemerocher (approved):

PR #10761, opened 2024-04-24
HTTP/2 prior knowledge support

### LINKED ISSUE #6301: Support H2C prior knowledge
### Issue description

Micronaut does not appear to support h2c with prior knowledge. Also see https://curl.se/docs/http2.html and https://github.com/micronaut-projects/micronaut-core/issues/5005#issuecomment-935889902 . Netty does seem to have support, so it should be possible to implement.

yawkat:
When accepting HTTP/2 plaintext connections, also accept prior knowledge HTTP/2. This actually already mostly worked for the new Http2ServerHandler, the changes are only necessary for the old multiplex handler.

Also fixed some test flakiness.

Fixes #6301

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/http2/H2cSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/http2/H2cSpec.groovy
@@ -229,6 +229,72 @@ class H2cSpec extends Specification {
         content.release()
     }
 
+    def 'prior knowledge'() {
+        given:
+        def responseFuture = new CompletableFuture()
+
+        def group = new NioEventLoopGroup(1)
+        def bootstrap = new Bootstrap()
+                .remoteAddress(embeddedServer.host, embeddedServer.port)
+                .group(group)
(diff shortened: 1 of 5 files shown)


graemerocher (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/http2/H2cSpec.groovy
```diff
@@ -229,6 +229,72 @@ class H2cSpec extends Specification {
         content.release()
     }
 
+    def 'prior knowledge'() {
```
timyates, line 232: I don't understand this test, but that's my failing, not yours 😉

timyates (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/http2/H2cSpec.groovy
yawkat: \~~netty magic~~

PR #10769, opened 2024-04-26
Remove unnecessary noise printed during a normal build.

mikehearn:
This change prevents IntelliJ opening the log window when using the "Show log on stderr" option just due to a non-actionable diagnostic being printed. Users don't need to see this and the convention in Java builds is to not print anything unless there's a problem.


alvarosanchez (approved):

mikehearn:
@dstepanov Could you take a look please?

PR #10772, opened 2024-04-27
Support ExecuteOn annotation with ServerWebSocket

### LINKED ISSUE #10758: @ExecuteOn(TaskExecutors.IO) does not work with @ServerWebSocket
### Expected Behavior

I am trying to run my websocket on IO threads, but its still running on default Netty event loop group

### Actual Behaviour

It should run on IO threads

### Steps To Reproduce

you can check the code here
https://github.com/seqeralabs/wave/pull/471

### Environment Information

OS: Mac 14.4.1 (23E224)

### Example Application

https://github.com/seqeralabs/wave/pull/471

### Version

3.10.3

jeremyg484:
NettyServerWebSocketHandler is updated to check for the ExecuteOn
annotation when invoking any of the callback methods on a
ServerWebSocket annotated class, and to use the specified
ExecutorService when invoking the methods.

A test is added to verify the enhanced behavior.

This resolves #10758

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
@@ -68,6 +68,7 @@ import java.security.Principal;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
```diff
@@ -345,8 +356,25 @@ protected Object invokeExecutable(BoundExecutable boundExecutable, MethodExecuti
...
+        return this.executorSelector.select(messageHandler.getExecutableMethod(), threadSelection)
+            .map(
+                executorService -> {
+                    ReturnType<?> returnType = messageHandler.getExecutableMethod().getReturnType();
+                    if (returnType.isReactive()) {
+                        return Mono.from((Publisher<?>) boundExecutable.invoke(messageHandler.getTarget()))
+                            .subscribeOn(Schedulers.fromExecutor(executorService))
+                            .contextWrite(reactorContext -> reactorContext.put(ServerRequestContext.KEY, originatingRequest));
+                    } else {
+                        return executorService.submit(() -> ServerRequestContext.with(originatingRequest,
```
yawkat: what happens if the method throws an exception in this code path. is it dropped or logged?

jeremyg484: @yawkat Good question...it appears to just get dropped right now because there's no error handling logic specific to `Future` in `AbstractNettyWebSocketHandler`, but there is for `Publisher`. I'm thinking the simplest approach may be to have the `ExectueOn` path always return a `Publisher`.

yawkat (approved):

PR #10775, opened 2024-04-29
Define default value for Cookie Max-Age

### LINKED ISSUE #10773: CookieFactory META-INF/services cause loading failures
### Expected Behavior

The design of the CookieFactory should look for an META-INF/services/io.micronaut.http.cookie.CookieFactory resource (and use it for the class name) and if not found, then it should fall back to HttpCookieFactory.

### Actual Behaviour

However, due to the presence of the META-INF/services/io.micronaut.http.cookie.CookieFactory within the micronaut-http library, and this file points to the fallback of HttpCookieFactory, then other overrides are frequently ignored.

So, if the micronaut-http-netty library is present, it has its own META-INF/services/io.micronaut.http.cookie.CookieFactory that points at its own CookieFactory, and it is frequently ignored. Unfortunately, it's even worse, since there's no defined order of ClassLoader.getResources, so depending on fairly interminate factors, sometimes the micronaut-http-netty version will be returned, and sometimes the micronaut-http version will be.

I think the correct response would be to delete the file micronaut-core/http/src/main/resources/META-INF/services/io.micronaut.http.cookie.CookieFactory

This should allow the micronaut-http-netty copy to be used if present, and then fallback will work if it's not.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.3, 4.4

jeremyg484:
Close: #10773 
Close: https://github.com/micronaut-projects/micronaut-security/issues/1682

The `Cookie` interface is updated with a constant value to represent an
undefined Max-Age, and its JavaDocs are updated to explicitly state
that the undefined value should not be encoded.

The `CookieHttpCookieAdapter` implementation is updated to explicitly set
the required undefined max age value in its constructor.

The new constant is consistent with the behavior of the Netty `Cookie`
implementation, and is meant to enforce consistency between it and the
`HttpCookie` based implementation, if only by explicitly stating the
intended contract.

The `CookieFactory` service definition is removed from the `http` module
as `HttpCookieFactory` already gets loaded by default (if no other
service definitions are loaded) via `CookieFactory.INSTANCE`. This
allows other explicit service definitions such as that in `http-netty`
to reliably override the default implementation.

Tests are added to verify the undefined max age value in both
implementations, and to ensure that `NettyServerCookieEncoder` can
correctly encode a `Cookie` created by the default `HttpCookieFactory`.

This should resolved issues occurring in Micronaut Security with session
based authentication such as https://github.com/micronaut-projects/micronaut-security/issues/1682

### DIFF
--- a/http-netty/src/test/groovy/io/micronaut/http/netty/cookies/CookeFactorySpec.groovy
+++ b/http-netty/src/test/groovy/io/micronaut/http/netty/cookies/CookeFactorySpec.groovy
@@ -1,5 +1,6 @@
 package io.micronaut.http.netty.cookies
 
+import io.micronaut.http.cookie.Cookie
 import io.micronaut.http.cookie.CookieFactory
 import spock.lang.Specification
 
@@ -8,5 +9,10 @@ class CookieFactorySpec extends Specification {
     void cookieFactoryResolvedViaSpi() {
         expect:
         CookieFactory.INSTANCE instanceof NettyCookieFactory
+    }
+
+    void "default cookie is a netty cookie with max age undefined"() {
+        expect:
+        Cookie.of("SID", "31d4d96e407aad42").getMaxAge() == Cookie.UNDEFINED_MAX_AGE
     }
 }
--- a/http-netty/src/test/groovy/io/micronaut/http/netty/cookies/NettyServerCookieEncoderSpec.groovy
+++ b/http-netty/src/test/groovy/io/micronaut/http/netty/cookies/NettyServerCookieEncoderSpec.groovy
@@ -1,6 +1,7 @@
 package io.micronaut.http.netty.cookies
 
 import io.micronaut.http.cookie.Cookie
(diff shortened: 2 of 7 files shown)


sdelamo (approved):
I added the same test for the [DefaultServerCookieEncoder](https://github.com/micronaut-projects/micronaut-core/pull/10775/commits/3f6496e5df2206525fe34435dd28478b76955078). 

A difference in encoding is the Netty encoder uses HTTPOnly while the other uses HttpOnly. The latter [seems more correct](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies) but I guess both work. 

@jeremyg484 I also [changed the adapter to set max only if it is -1](https://github.com/micronaut-projects/micronaut-core/pull/10775/commits/8567726ac8f82cc05d0cd3bc94c7f1c620b3cda8).

yawkat (approved):

PR #10777, opened 2024-04-29
fix: KSP processing for package private properties

timyates:
We seem to be processing package private properties when using Ksp.

This is incorrect and they should be skipped.

Package-private isn't really a concept in Kotlin, but the Kotlin classes may be based on Java classes where it is a thing.

I believe when I have a fix, that this will fix https://github.com/micronaut-projects/micronaut-sql/issues/1324

Closes: https://github.com/micronaut-projects/micronaut-sql/issues/1324

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/visitor/BeanIntrospectionSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/visitor/BeanIntrospectionSpec.groovy
@@ -2522,4 +2522,24 @@ class Holder<A extends Animal> {
             animal.isTypeVariable()
     }
 
+    void "test package private field instrospection"() {
+        when:
+        def introspection = buildBeanIntrospection('test.Test', '''
+package test
+
+import io.micronaut.core.annotation.Introspected
+import io.micronaut.inject.visitor.MySuperclass
+
+@Introspected
+class Test extends MySuperclass {
+
+    String name
+}
+''')
+        then:
+        introspection.getProperty("name").orElse(null)
+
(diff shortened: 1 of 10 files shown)


graemerocher:
So if it is package private but in the same package that is not allowed? Does this align with java?

timyates:
The original issue is that the PersonRecord extends a public Jooq class `UpdatableRecordImpl` (which is in a different package)

https://github.com/davidsonnabend/micronaut-illegal-access-error/blob/main/src/main/generated/com/example/jooq/generated/tables/records/PersonRecord.kt

This `UpdatableRecordImpl` eventually extends `AbstractRecord`

https://github.com/jOOQ/jOOQ/blob/d0bf1d180fc2f7cd10f7a49e0974b4594211889e/jOOQ/src/main/java/org/jooq/impl/AbstractRecord.java#L122

and **this** is package private... 🤔 

I thought I added the same class to all 3 languages (groovy, java and kotlin) and kotlin was different...

I will check again to make sure 🤔

timyates:
@graemerocher You're right, I've found a difference between kotlin and java/groovy... working on it

timyates:
@graemerocher In https://github.com/micronaut-projects/micronaut-core/pull/10777/commits/d1d7e9e698772ba0a52aa203473aa0d9e7c9c18f I added a test for all 3 languages to make sure we see the same...

When the superclass is in the same package as the introspected bean, we _do_ see the package private property in all 3

So I *think* this is right 🤔

graemerocher (approved):

sdelamo (approved):

PR #10780, opened 2024-04-30
refactor: use Blocking HTTP Client in HealthEndpointSpec

sdelamo:
- Use StringUtils.FALSE or StringUtils.FALSE
- Uses BlockingHttpClient intead of Reactive HTTP Client
- Remove usage of def

### DIFF
--- a/management/src/test/groovy/io/micronaut/management/endpoint/health/HealthEndpointSpec.groovy
+++ b/management/src/test/groovy/io/micronaut/management/endpoint/health/HealthEndpointSpec.groovy
@@ -19,10 +19,13 @@ import io.micronaut.context.ApplicationContext
 import io.micronaut.context.annotation.Requires
 import io.micronaut.core.convert.ArgumentConversionContext
 import io.micronaut.core.type.Argument
+import io.micronaut.core.util.StringUtils
 import io.micronaut.health.HealthStatus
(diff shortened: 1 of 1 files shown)


management/src/test/groovy/io/micronaut/management/endpoint/health/HealthEndpointSpec.groovy
```diff
@@ -103,7 +106,7 @@ class HealthEndpointSpec extends Specification {
 
     void "test the beans are not available with all disabled"() {
         given:
-        ApplicationContext context = ApplicationContext.run(['endpoints.all.enabled': false])
+        ApplicationContext context = ApplicationContext.run(['endpoints.all.enabled': StringUtils.FALSE])
```
timyates, line 109: Not sure why this is better, but there's 3 other places we can change if we want to be consistent?

Lines 67, 81 and 95

timyates (approved):
What was the motivation for this change?

sdelamo:
> What was the motivation for this change?

I added a test to this [documentation contribution](https://github.com/micronaut-projects/micronaut-core/pull/10759), and I saw the complicated use of Reactor in the test - for example in the exception handling. In my opinion, using the `BlockingHttpClient` simplifies the test.

PR #10781, opened 2024-04-30
New request body API

yawkat:
This PR adds a new *public* API for accessing the bytes of a request in the HTTP server. While it shares some naming with the ByteBody API we used before, it is a very different design. It's designed to be public API, not netty-specific, but still powerful enough to allow for the ByteBody optimizations we had before.

At the moment, the new api is only used for requests and only on the server. I'd like to expand it to the response and to the client if possible. But there are some challenges, so it won't make it into this PR:

- ByteBody *must* be closed. Only NettyHttpRequest has the wiring to make that possible at the moment, the other HttpReq/Resp implementations are much more loose about resource management.
- The server request is the only point where filters are actually executed when the body is in byte form, at the moment. In client messages and in the server response, the body is in object form when filters are executed.

Another missing piece is a non-netty implementation of ByteBody. I am particularly interested in a servlet implementation based on InputStream. I will implement that as a separate PR.

Some pieces of the old netty-only body api remain (the ObjectBody impls) to keep changes down. Also AbstractHttpContentProcessor is finally removed, all fields are folded into FormDataHttpContentProcessor.


src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
+Micronaut HTTP server version 4.5.0 introduced a new, more advanced API to access the bytes of an incoming request.
+When a api:io.micronaut.http.HttpRequest[] implements api:io.micronaut.http.ServerHttpRequest[], the new `byteBody()`
```
timyates:
```suggestion
When an api:io.micronaut.http.HttpRequest[] implements api:io.micronaut.http.ServerHttpRequest[], the new `byteBody()`
```

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+method returns an api:io.micronaut.http.body.ByteBody[] with buffered, reactive and blocking APIs.
```
timyates:
```suggestion
method returns a api:io.micronaut.http.body.ByteBody[] with buffered, reactive and blocking APIs.
```

src/main/docs/guide/httpServer/byteBody.adoc
timyates, line 1: This isn't included in the TOC, so doesn't end up in the docs

http/src/main/java/io/micronaut/http/exceptions/BufferLengthExceededException.java
```diff
@@ -0,0 +1,33 @@
...
+ * @since 4.5.0
+ */
+public class BufferLengthExceededException extends HttpException {
+
+    /**
+     * @param advertisedLength The advertised length
+     * @param receivedLength   The received length
+     */
+    public BufferLengthExceededException(long advertisedLength, long receivedLength) {
+        super("The content length [" + receivedLength + "] exceeds the maximum allowed bufferable length [" + advertisedLength + "]. Note that the maximum buffer size got its own configuration property (micronaut.server.max-request-buffer-size) in 4.5.0 that you may have to configure. Alternatively you can rewrite your controller to stream the request instead of buffering it.");
```
timyates: This is arguably easier to read?

```suggestion
        super("""
            The content length [%d] exceeds the maximum allowed bufferable length [%d]. \
            Note that the maximum buffer size got its own configuration property (micronaut.server.max-request-buffer-size) in 4.5.0 that you may have to configure. \
            Alternatively you can rewrite your controller to stream the request instead of buffering it.""".formatted(receivedLength, advertisedLength));
```

Also, can we have a test for it being thrown?

timyates (changes requested):
Sonar has some concerns with Null pointer exceptions

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentSubscriberFactory.java
```diff
@@ -1,36 +0,0 @@
...
-
-/**
- * Interface that allows plugging into the constructor of the {@link io.netty.handler.codec.http.HttpContent}
- * subscriber.
- *
- * @author Graeme Rocher
- * @see DefaultHttpContentProcessor
- * @since 1.0
- */
-public interface HttpContentSubscriberFactory {
```
graemerocher, line 26: this doesn't seem to be annotated with `@Internal` can you restore and add `@Deprecated` with documentation

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/BufferConsumer.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * This is a netty-specific reactor-like API for streaming bytes. It's a bit better than reactor
+ * because it's more explicit about reference counting semantics, has more fine-grained controls
+ * for cancelling, and has more relaxed concurrency semantics.
+ *
+ * @since 4.5.0
+ * @author Jonas Konrad
+ */
+@Internal
+public interface BufferConsumer {
```
graemerocher, line 31: can this be sealed? Does it need to be public?

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/NettyByteBody.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+@Internal
+public abstract sealed class NettyByteBody implements ByteBody permits ImmediateNettyByteBody, StreamingNettyByteBody {
+    protected static final Logger LOG = LoggerFactory.getLogger(NettyByteBody.class);
+
+    static void failClaim() {
+        throw new IllegalStateException("Request body has already been claimed: Two conflicting sites are trying to access the request body. If this is intentional, the first user must ByteBody#split the body. To find out where the body was claimed, turn on TRACE logging for io.micronaut.http.server.netty.body.NettyByteBody.");
+    }
+
+    protected abstract Flux<ByteBuf> toByteBufPublisher();
```
graemerocher: try reorganise these methods, public at the top, protected at the bottom. Also review whether package private is necessary

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingNettyByteBody.java
```diff
@@ -0,0 +1,541 @@
...
+        upstream.disregardBackpressure();
+        upstream.start();
+        sharedBuffer.subscribe(null, upstream);
+    }
+
+    /**
+     * This class buffers input data and distributes it to multiple {@link StreamingNettyByteBody}
+     * instances.
+     */
+    public static final class SharedBuffer implements BufferConsumer {
```
graemerocher, line 189: can you make this a private class with some kind of factory, we need to try make less API surface area public if possible

graemerocher, line 189: I assume this class is thread safe and not subject to concurrent access?

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingNettyByteBody.java
```diff
@@ -0,0 +1,541 @@
...
+            if (eventLoopFlow.executeNow(() -> subscribe0(subscriber, specificUpstream))) {
+                subscribe0(subscriber, specificUpstream);
+            }
+        }
+
+        private void subscribe0(@Nullable BufferConsumer subscriber, Upstream specificUpstream) {
+            assert !working;
+
+            if (reserved == 0) {
+                throw new IllegalStateException("Need to reserve a spot first");
```
graemerocher, line 309: improve this error message

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/UpstreamBalancer.java
```diff
@@ -0,0 +1,293 @@
...
+ *     {@code n} demand upstream</li>
+ *     <li>{@code l<r}, but {@code l+n>r}: left downstream was lagging behind but now right
+ *     downstream is. Just send {@code r-l=abs(l-r)} upstream</li>
+ * </ul>
+ * <p>
+ * The last two cases can be combined into sending {@code min(n, abs(l-r))} upstream. So we
+ * only need to test for the first case.
+ */
+@Internal
+public final class UpstreamBalancer {
```
graemerocher: does it need to be public?

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/UpstreamBalancer.java
```diff
@@ -0,0 +1,293 @@
...
+        UpstreamBalancer balancer = new UpstreamBalancer(upstream);
+        return new UpstreamPair(balancer.new FastestUpstreamImpl(false), balancer.new FastestUpstreamImpl(true));
+    }
+
+    static UpstreamPair first(BufferConsumer.Upstream upstream) {
+        UpstreamBalancer balancer = new UpstreamBalancer(upstream);
+        return new UpstreamPair(balancer.new PassthroughUpstreamImpl(), balancer.new IgnoringUpstreamImpl());
+    }
+
+    static UpstreamPair balancer(BufferConsumer.Upstream upstream, ByteBody.SplitBackpressureMode mode) {
```
graemerocher, line 103: describe this method and in more detailed documentation the different modes and what they do. Whilst this class is not public API maintainers need to understand this implementation.

Also add complete javadoc to the various factory methods.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
```diff
@@ -306,6 +312,22 @@ private void writeSome() {
...
+    /**
+     * This is a wrapper around {@link ImmediateNettyByteBody#ImmediateNettyByteBody(ByteBuf)}
+     * with an extra body length check.
+     */
+    static CloseableByteBody createImmediateByteBody(EventLoop loop, BodySizeLimits bodySizeLimits, ByteBuf buf) {
+        if (buf.readableBytes() > bodySizeLimits.maxBodySize() || buf.readableBytes() > bodySizeLimits.maxBufferSize()) {
+            BufferConsumer.Upstream upstream = bytesConsumed -> {
+            };
+            StreamingNettyByteBody.SharedBuffer mockBuffer = new StreamingNettyByteBody.SharedBuffer(loop, bodySizeLimits, upstream);
+            mockBuffer.add(buf); // this will trigger the exception
```
graemerocher: document what exception and why it needs to be implemented this way

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
```diff
@@ -490,130 +512,103 @@ void handleUpstreamError(Throwable cause) {
...
      * Handler that exposes incoming content as a {@link Flux}.
      */
-    private final class StreamingInboundHandler extends InboundHandler {
-        private final Queue<HttpContent> queue = Queues.<HttpContent>unbounded().get();
-        private final Sinks.Many<HttpContent> sink = Sinks.many().unicast().onBackpressureBuffer(queue);
-        private long requested = 0;
+    private final class StreamingInboundHandler extends InboundHandler implements BufferConsumer.Upstream {
+        final StreamingNettyByteBody.SharedBuffer dest;
+        final OutboundAccessImpl outboundAccess;
+        long requested = 65536;
```
graemerocher: what is the magic number? Provide documentation here

http-server/src/main/java/io/micronaut/http/server/exceptions/BufferLengthExceededHandler.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Default handler for {@link BufferLengthExceededHandler} errors.
+ *
+ * @author Jonas Konrad
+ * @since 4.5.0
+ */
+@Singleton
+@Produces
+public class BufferLengthExceededHandler implements ExceptionHandler<BufferLengthExceededException, HttpResponse> {
```
graemerocher, line 36: provide documentation on how to customise this

http-server/src/main/java/io/micronaut/http/server/exceptions/BufferLengthExceededHandler.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final ErrorResponseProcessor<?> responseProcessor;
+
+    /**
+     * Constructor.
+     * @param responseProcessor Error Response Processor
+     */
+    @Inject
```
graemerocher: inject is unnecessary here

http/src/main/java/io/micronaut/http/body/CloseableByteBody.java
```diff
@@ -0,0 +1,46 @@
...
+ * that may still be held. No other operations on this body are valid after {@link #close()}, but
+ * multiple calls to {@link #close()} are allowed (though only the first will do anything). If a
+ * primary operation (see {@link ByteBody}) is performed on this body, you can but do not
+ * need to close it anymore. Closing becomes a no-op in that case.
+ *
+ * @author Jonas Konrad
+ * @since 4.5.0
+ */
+@Experimental
+public interface CloseableByteBody extends ByteBody, Closeable {
```
graemerocher, line 34: implement `AutoCloseable` as well

http/src/main/java/io/micronaut/http/body/ImmediateByteBody.java
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * This is an extension of {@link ByteBody} when the entire body is immediately available
+ * (without waiting). It has the same semantics as {@link ByteBody}, but it adds a few other
+ * primary operations for convenience.
+ *
+ * @author Jonas Konrad
+ * @since 4.5.0
+ */
+@Experimental
+public interface ImmediateByteBody extends ByteBody {
```
graemerocher: rename to AvailableByteBody

http/src/main/java/io/micronaut/http/exceptions/BufferLengthExceededException.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package io.micronaut.http.exceptions;
+
+/**
+ * Exception thrown when attempting to buffer more than the configured limit.
+ *
+ * @author Jonas Konrad
+ * @since 4.5.0
+ */
+public class BufferLengthExceededException extends HttpException {
```
graemerocher: can be final

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Get this body as an {@link InputStream}.
+     * <p>This is a primary operation. After this operation, no other primary operation or
+     * {@link #split()} may be done.
+     *
+     * @return The streamed bytes
+     */
+    @NonNull
+    InputStream toInputStream();
```
graemerocher, line 109: what happens if called on the event loop in a blocking manner?

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+
+A normal `ByteBody` has two groups of streaming primary operations. `toInputStream()` gives access to the body
+as a regular `InputStream`. The `toByteArrayPublisher()` and `toByteBufferPublisher()` methods return a reactive stream
+of byte arrays or `ByteBuffer`s.
+
+WARNING: `InputStream` is blocking API, and the netty event loop must never be blocked. If you wish to read from the
+body using an `InputStream`, take care to do so only on another thread, or to annotate your filter with
+`@ExecuteOn(TaskExecutors.BLOCKING)`.
+
+If you need full access to the body, the `buffer()` method returns an `ExecutionFlow` that completes with an
```
graemerocher: is `ExecutionFlow` now public API? Currently it is annotated with `@Internal`? Is it ready to be public API?

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+means that if a filter were to claim the `ServerHttpRequest.byteBody()` directly (e.g. to print it to a log),
+controllers could not access it anymore. The argument binder for the `@Body` argument would throw an exception.
+
+To resolve this exclusivity problem, an `ByteBody` can be _split_ before it is claimed. The split operation
+essentially duplicates the body stream so that the two consumers (logging and argument binding) can process it
+independently. A body can be `split` any number of times, but only before the primary operation.
+
+While `ServerHttpRequest.byteBody()` returns a normal `ByteBody` -- cleanup is done by the HTTP server if the
+body is not consumed--the body returned by `split` is a `CloseableByteBody`. The caller *must* ensure that the
+new instance is closed, otherwise there can be resource and memory leaks, stalled connections, or other issues.
```
graemerocher: we need a documented example of this use case and splitting

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+These two approaches are already the two most important
+api:io.micronaut.http.body.ByteBody.SplitBackpressureMode[]s. The full list of options is as follows:
+
+* `SplitBackpressureMode.SLOWEST` uses the backpressure of the _slowest_ of the two consumers (first example)
+* `SplitBackpressureMode.FASTEST` uses the backpressure of the _fastest_ of the two consumers (second example)
+* `SplitBackpressureMode.ORIGINAL` uses the backpressure of the original consumer (the one `split()` was called on)
+* `SplitBackpressureMode.NEW` uses the backpressure of the new consumer (the one `split()` returns)
+
+The argument-less `split()` method uses `SLOWEST`, but you should pick the mode that is most appropriate for your use
+case.
```
graemerocher: again documentation needs example code

test-suite/src/test/java/io/micronaut/docs/server/body/BodyLogFilter.java
```diff
@@ -0,0 +1,36 @@
...
+    public void logBody(ServerHttpRequest<?> request) { // <2>
+        try (CloseableByteBody ourCopy = // <4>
+                 request.byteBody()
+                     .split(ByteBody.SplitBackpressureMode.SLOWEST) // <3>
+                     .allowDiscard()) { // <5>
+            Flux.from(ourCopy.toByteArrayPublisher()) // <6>
+                .onErrorComplete(ByteBody.BodyDiscardedException.class) // <7>
+                .subscribe(array -> LOG.info("Received body: {}", Base64.getEncoder().encodeToString(array))); // <8>
+        }
+    }
```
graemerocher, line 34: Provide example of how to do this in the blocking case

graemerocher (changes requested):
I would like @dstepanov to review this when he is back

graemerocher:
overall looks like a good change but think it needs some refinement to docs and clarification on blocking use cases.

test-suite/src/test/java/io/micronaut/docs/server/body/BodyLogFilter.java
yawkat: This cant really be implemented in a blocking way, because that would block the filter until the whole body is received, which we dont want for logging

http/src/main/java/io/micronaut/http/exceptions/BufferLengthExceededException.java
yawkat:
```suggestion
public final class BufferLengthExceededException extends HttpException {
```

http/src/main/java/io/micronaut/http/exceptions/BufferLengthExceededException.java
yawkat: there's a test in MaxRequestSizeSpec (the one with startsWith)

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/BufferConsumer.java
yawkat: It's `@Internal`. It needs to be public because theres implementations in the body classes and in the HTTP/1 and HTTP/2 server handlers

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingNettyByteBody.java
yawkat: It's internal

yawkat: i will write the thread safety in a comment

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/StreamingNettyByteBody.java
yawkat: It is not reachable by the user. This is guarded by the failClaim checks in StreamingNettyByteBody. So it's just a sanity check. I will improve the message a bit to state this

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/UpstreamBalancer.java
yawkat: no

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/UpstreamBalancer.java
yawkat: The semantics of the different modes are described in SplitBackpressureMode

http-server/src/main/java/io/micronaut/http/server/exceptions/BufferLengthExceededHandler.java
yawkat: What do you mean by customize?

yawkat: fwiw this class is just copied ContentLengthExceededHandler

http/src/main/java/io/micronaut/http/body/CloseableByteBody.java
yawkat: Closeable extends AutoCloseable so it's not necessary

http/src/main/java/io/micronaut/http/body/ByteBody.java
yawkat: it's not allowed. i will add a check to PublisherAsBlocking

src/main/docs/guide/httpServer/byteBody.adoc
yawkat: oh i forgot about this. @dstepanov 

I talked with Denis about this a few weeks ago and he wasn't against it iirc. But we may want to only make some of the methods public, the API is pretty big atm.

Another alternative is to have a second method that returns CompletionStage, and make the ExecutionFlow one Internal and default. That way we can still have the perf benefit of ExecutionFlow without exposing anything.

src/main/docs/guide/httpServer/byteBody.adoc
yawkat: the BodyLogFilter has splitting

src/main/docs/guide/httpServer/byteBody.adoc
yawkat: oops

graemerocher:
can you address the 6 new bugs identified by sonar

src/main/docs/guide/httpServer/byteBody.adoc
yawkat: ive moved to CompletableFuture

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+[source]
+----
+16:29:30.562 [default-nioEventLoopGroup-1-3] INFO  i.m.docs.server.body.BodyLogFilter - Received body: ...
+16:29:30.584 [default-nioEventLoopGroup-1-3] INFO  i.m.docs.server.body.BodyLogFilter - Received body: ...
+16:29:30.642 [default-nioEventLoopGroup-1-3] INFO  i.m.docs.server.body.BodyLogFilter - Received body: ...
+16:29:30.773 [default-nioEventLoopGroup-1-3] INFO  i.m.d.server.body.BodyLogController - Creating person Person[firstName=..., lastName=...]
+16:29:30.708 [default-nioEventLoopGroup-1-3] INFO  i.m.docs.server.body.BodyLogFilter - Received body: ...
+----
+
+Note that the logging in the above example is asynchronous, so the log statements may be interleaved as shown.
```
graemerocher: is there a way to do `Flux.toList()` to avoid this interleaving?

yawkat: no, i simply do not want to make guarantees about ordering. in fact the sentence and the example log are lying, there is no interleaving atm, but i dont want people to rely on that.

all subscribers to the same body have equal priority and may get parts of the body in any order.

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
+Micronaut HTTP server version 4.5.0 introduced a new, more advanced API to access the bytes of an incoming request.
+When an api:io.micronaut.http.HttpRequest[] implements api:io.micronaut.http.ServerHttpRequest[], the new `byteBody()`
+method returns a api:io.micronaut.http.body.ByteBody[] with buffered, reactive and blocking APIs.
+
+When an HTTP request comes in, it starts out with an unparsed and unclaimed stream of bytes as its `byteBody()`. After
+all request filters have run, typically an argument binder matching the `@Body` parameter of the controller will
+"claim" the `byteBody()` and e.g. parse the JSON. Finally, the body is closed at the end of the request lifecycle,
+discarding any data if it has not been claimed by the argument binder.
+
+=== Primary operations
```
sdelamo: split in multiple files

sdelamo (changes requested):

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+body using an `InputStream`, take care to do so only on another thread, or to annotate your filter with
+`@ExecuteOn(TaskExecutors.BLOCKING)`.
+
+If you need full access to the body, the `buffer()` method returns a `CompletableFuture` that completes with an
+`AvailableByteBody` when the full body has been received. `AvailableByteBody` has a few more convenient
+primary operations: `toByteArray()`, `toByteBuffer()` and `toString(Charset)`.
+
+Buffering is limited to the number of bytes configured in the `micronaut.server.max-request-buffer-size` property.
+
+=== Splitting
```
sdelamo: This should be in a different file to easy navigation and linking to the docs

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+
+To resolve this exclusivity problem, an `ByteBody` can be _split_ before it is claimed. The split operation
+essentially duplicates the body stream so that the two consumers (logging and argument binding) can process it
+independently. A body can be `split` any number of times, but only before the primary operation.
+
+While `ServerHttpRequest.byteBody()` returns a normal `ByteBody` -- cleanup is done by the HTTP server if the
+body is not consumed--the body returned by `split` is a `CloseableByteBody`. The caller *must* ensure that the
+new instance is closed, otherwise there can be resource and memory leaks, stalled connections, or other issues.
+
+==== Backpressure
```
sdelamo: This should be in a different file to simplify linking and navigation to the docs.

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+
+* `SplitBackpressureMode.SLOWEST` uses the backpressure of the _slowest_ of the two consumers (first example)
+* `SplitBackpressureMode.FASTEST` uses the backpressure of the _fastest_ of the two consumers (second example)
+* `SplitBackpressureMode.ORIGINAL` uses the backpressure of the original consumer (the one `split()` was called on)
+* `SplitBackpressureMode.NEW` uses the backpressure of the new consumer (the one `split()` returns)
+
+The argument-less `split()` method uses `SLOWEST`, but you should pick the mode that is most appropriate for your use
+case.
+
+==== Discarding
```
sdelamo: This should be in a different file

src/main/docs/guide/httpServer/byteBody.adoc
```diff
@@ -0,0 +1,134 @@
...
+When there are multiple consumers, discard behavior is dependent on use case. In the above scenario of an unmatched
+request, when there is a filter that is also subscribed to the body data, it may be appropriate to drop the request in
+some cases (e.g. for logging) but may be necessary to still receive all the data in others.
+
+To signal that the upstream may discard the body, you can call `ByteBody.allowDiscard()`. Only if all consumers
+call `allowDiscard()` (or `close()` without a primary operation) may the remaining data actually be discarded. Before
+that, _all_ consumers, even those that called `allowDiscard()`, will still receive all data. For the logging use case,
+you can call `allowDiscard()` and be assured that you will still log the full body if the controller needs it.
+
+=== Example
```
sdelamo: this should be in a different file

sdelamo (changes requested):
You should create 

- `httpServer/byteBody/byteBodyPrimaryOperations.adoc`
- `httpServer/byteBody/byteBodySplitting.adoc`
- `httpServer/byteBody/byteBodyBackpressure.adoc`
- `httpServer/byteBody/byteBodyDiscarding.adoc`
- `httpServer/byteBody/byteBodyExample.adoc`
- `httpServer/byteBody/byteBodyExample.adoc`

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -707,25 +658,25 @@ public long getContentLength() {
 
     @Override
     public boolean isFull() {
-        return byteBody() instanceof ImmediateByteBody;
+        return byteBody() instanceof AvailableNettyByteBody;
```
dstepanov, line 663: Can we deprecate this and the `contents` methods and move them to the body interface?

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,212 @@
...
+ */
+package io.micronaut.http.body;
+
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.execution.CompletableFutureExecutionFlow;
+import io.micronaut.core.execution.ExecutionFlow;
+import io.micronaut.core.io.buffer.ByteBuffer;
+import org.jetbrains.annotations.Contract;
```
dstepanov, line 21: I don't think we use jetbrains annotations

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Create a new, independent {@link ByteBody} that contains the same data as this one.
+     *
+     * @param backpressureMode How to handle backpressure between the old and new body. See
+     *                         {@link SplitBackpressureMode} documentation
+     * @return The newly split body. Must be closed by the caller, unless a terminal operation is
+     * performed on it
+     */
+    @NonNull
+    CloseableByteBody split(@NonNull SplitBackpressureMode backpressureMode);
```
dstepanov, line 71: I don't understand the naming here. Why `split`? Should it be a copy with buffered data? So maybe `buffered` or something else? WDYT @graemerocher

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,212 @@
...
+     * body allow discarding will the body be discarded, otherwise it will still be sent to all
+     * consumers. It is an optional operation.
+     * <p>Discarding may be implemented e.g. by closing the input side of an HTTP/2 stream.
+     * <p>This method must be called before any primary operation.
+     *
+     * @return This body
+     */
+    @Contract("-> this")
+    @NonNull
+    default ByteBody allowDiscard() {
```
dstepanov, line 84: Does it affect the current instance? The name sounds more like a setter. I would name it `discarded` if only the returned instance is affected.

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Get this body as an {@link InputStream}.
+     * <p>This is a primary operation. After this operation, no other primary operation or
+     * {@link #split()} may be done.
+     *
+     * @return The streamed bytes
+     */
+    @NonNull
+    InputStream toInputStream();
```
dstepanov, line 109: Maybe we should allow only using the input stream in lambda, which it will close: `R read(is -> ...)`. And perhaps it would be user-friendly to have API like `R readAndDiscard(is -> ...)` + `R readAndRetain(is -> ...)` to identify directly what should happen to the data. WDYT @graemerocher

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Get this body as a reactive stream of byte arrays.
+     * <p>This is a primary operation. After this operation, no other primary operation or
+     * {@link #split()} may be done.
+     *
+     * @return The streamed bytes
+     */
+    @NonNull
+    Publisher<byte[]> toByteArrayPublisher();
```
dstepanov, line 119: Maybe these two should be moved to some `StreamingByteBody` interface?

http/src/main/java/io/micronaut/http/body/ByteBody.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * <b>Internal</b> variant of {@link #buffer()} that uses the {@link ExecutionFlow} API for
+     * extra efficiency.
+     *
+     * @return A flow that completes when all bytes are available
+     */
+    @NonNull
+    @Internal
+    default ExecutionFlow<? extends CloseableAvailableByteBody> bufferFlow() {
```
dstepanov: Maybe this one can be moved to some `InternalByteBody` to not leak the internal class

dstepanov (changes requested):

dstepanov:
I am not a fan of the naming `ByteBody`. Maybe we can add `HttpBody`, `HttpBody body()` and `RequestHttpBody`/`ResponseHttpBody`. There are a few cases right now that can be improved by this hierarchy. In many cases (at least in the server implementation for the request), we set some anonymous body and a few steps later try to analyze its content, so setting the `HttpBody` with some media type would help eliminate the interface switch. Also, it will help to have the body already analyzed for the reactive cases, etc, avoiding additional checks.
The `ResponseHttpBody` would be the current `ByteBody`.

I will look more on Thursday when I'm back.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: the entire FullHttpRequest is now deprecated.

http/src/main/java/io/micronaut/http/body/ByteBody.java
yawkat: it's compile only so it shouldnt hurt

http/src/main/java/io/micronaut/http/body/ByteBody.java
yawkat: it is not necessarily buffered. it splits the stream so that there can be two consumers

http/src/main/java/io/micronaut/http/body/ByteBody.java
yawkat: it affects the current instance

yawkat:
It is named ByteBody because it specifically deals with bytes. It is independent of HttpRequest.getBody which can also have independent, unserialized beans

http/src/main/java/io/micronaut/http/body/ByteBody.java
graemerocher: those seem complementary and useful but there will be cases where the user needs to be in control when the input stream is closed I think.

yawkat: what would be the difference between discard and retain?

http/src/main/java/io/micronaut/http/body/ByteBody.java
yawkat: It is intentional that ImmediateByteBody extends the streaming case, so that users write their code for the streaming case but we can give optimized implementations where possible.

graemerocher:
@yawkat can you rebase?

yawkat:
ive merged 4.5.x into this, we can squash on PR merge

graemerocher (approved):

PR #10792, opened 2024-05-03
Add API to access SSL session

yawkat:
This is useful for e.g. the session ID.

Because I replaced the getCertificate implementation with this, this is already covered by tests.

### DIFF
--- a/http/src/main/java/io/micronaut/http/HttpRequest.java
+++ b/http/src/main/java/io/micronaut/http/HttpRequest.java
@@ -19,6 +19,8 @@ import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.http.cookie.Cookies;
 
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
 import java.net.InetSocketAddress;
 import java.net.URI;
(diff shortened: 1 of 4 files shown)


http/src/main/java/io/micronaut/http/HttpRequest.java
```diff
@@ -179,7 +181,28 @@ default Optional<Locale> getLocale() {
      */
     @SuppressWarnings("deprecation")
     default Optional<Certificate> getCertificate() {
-        return this.getAttribute(HttpAttributes.X509_CERTIFICATE, Certificate.class);
+        Optional<Certificate> attribute = this.getAttribute(HttpAttributes.X509_CERTIFICATE, Certificate.class);
```
timyates, line 184: When is this ever present?  Noticed that this being present isn't covered by the tests, and just wondered as it's the only usage of this deprecated constant 🤔

timyates (approved):

graemerocher (approved):

http/src/main/java/io/micronaut/http/HttpRequest.java
yawkat: unfortunately HttpAttributes.X509_CERTIFICATE is public api so a user couldve set the attribute themselves

Issue #10745, opened 2024-04-19
http-client throws NoSuchMethodError 'boolean com.fasterxml.jackson.core.io.NumberInput.looksLikeValidNumber(java.lang.String)'


nedelva:
### Expected Behavior

I tried to upgrade the platform from 4.3.8 to 4.4.0. When doing so, a few tests using Jackson serialization threw the `NoSuchMethodError` error.
Looking at the `micronaut-core-bom-4.4.3.pom` I noticed that Jackson Databind and Jackson Core have different versions:

>   <jackson.databind.version>2.17.0</jackson.databind.version>
    <jackson.version>2.16.1</jackson.version>

Since Databind depends on Core, their versions usually go hand in hand.  Any particular reason why this happened?

### Actual Behaviour

Here's the stacktrace:
```
2024-04-19 10:34:30  DEBUG DefaultHttpClient - Sending HTTP GET to http://localhost:4444/etc/etc/some_url_here
2024-04-19 10:34:30  WARN  DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.NoSuchMethodError: 'boolean com.fasterxml.jackson.core.io.NumberInput.looksLikeValidNumber(java.lang.String)'
	at com.fasterxml.jackson.databind.deser.std.NumberDeserializers$BigDecimalDeserializer.deserialize(NumberDeserializers.java:1052) ~[jackson-databind-2.17.0.jar:2.17.0]
	at com.fasterxml.jackson.databind.deser.std.NumberDeserializers$BigDecimalDeserializer.deserialize(NumberDeserializers.java:990) ~[jackson-databind-2.17.0.jar:2.17.0]
	at com.fasterxml.jackson.databind.deser.SettableBeanProperty.deserialize(SettableBeanProperty.java:545) ~[jackson-databind-2.17.0.jar:2.17.0]
	at io.micronaut.jackson.modules.BeanIntrospectionModule$BeanIntrospectionSetter.deserializeAndSet(BeanIntrospectionModule.java:1124) ~[micronaut-jackson-databind-4.4.3.jar:4.4.3]
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:273) ~[jackson-databind-2.17.0.jar:2.17.0]
[...]

PR #10795, opened 2024-05-03
build: Use only one jackson version

sdelamo:
Close: #10745

Use `managed-jackson` version everywhere.

Use `managed-jackson-databaind` should be removed in Micronaut Framework 5.


timyates (approved):
For anyone interested, this split to have 2 managed versions was added here

https://github.com/micronaut-projects/micronaut-core/pull/7146

Because Jackson ended up with different version numbering for core, bom and databind

https://github.com/FasterXML/jackson-databind/issues/3428#issuecomment-1079418162

Going back to a single version however is the correct route to take now there is agreement inside jackson as renotave doesn't know these should be aligned so we keep ending up with different versions...

Issue #10797, opened 2024-05-03
Property Placeholder No Longer Resolves to Null for Nullable Arguments When Environment Variable Does Not Exist


helios2k6:
### Expected Behavior
Property placeholders should resolve to null when they are not specified by an environment variable.

# Summary
Given a Bean Factory, a `@Nullable` argument that retrieves its value from a property via the `@Value` annotation, and this value is expected to be resolved through the environment property source, Micronaut will now throw an exception stating that it could not resolve the placeholder.

### Configuration File
```yaml
some:
  optional:
    value: ${SOME_OPTIONAL_VALUE_PLACEHOLDER}
```

Assume the `SOME_OPTIONAL_VALUE_PLACEHOLDER` environment variable has not been set.

### Factory Class
```kotlin
@Factory
class ExampleBeanFactory {
  @Singleton
  fun exampleBean(
    @Nullable
    @Value("\${some.optional.value}")
    optionalValue: String?,
  ): ExampleBean {
    return ExampleBean(optionalValue)
  }
```

### Actual Behaviour

An exception is thrown stating that the placeholder could not be resolved.

### Steps To Reproduce

1. Use Kotlin
2. Create configuration file that uses a placeholder
3. Use the Environment Property Source
4. Create bean factory
5. Create a function within the bean factory class with the `@Singleton` annotation and a nullable argument that has the `@Nullable` and `@Value` annotations
6. Do not set the placeholder value equal to anything in your environment
7. Observe the exception is thrown

### Environment Information

OS: Pop!_OS 22.04 LTS
Java Version:
```
java --version
openjdk 17.0.8.1 2023-08-24 LTS
OpenJDK Runtime Environment Zulu17.44+53-CA (build 17.0.8.1+1-LTS)
OpenJDK 64-Bit Server VM Zulu17.44+53-CA (build 17.0.8.1+1-LTS, mixed mode, sharing)
```
Micronaut Dependencies and Versions:

gradle.properties
```
micronaut_kotlin_runtime_version=4.3.0
micronaut_version=4.4.6
snake_yaml_version=2.2
kotlin_version=1.9.23
kotlinx_coroutines_core_version=1.8.0
all_open_plugin_version=1.9.23
```

gradle.build
```
plugins {
    id "application"
    id "idea"
    id "org.jetbrains.kotlin.jvm"
    id "org.jetbrains.kotlin.kapt"
[...]

PR #10798, opened 2024-05-03
#10797 Re-add Try-Catch for ConfigurationExceptions To DefaultPropertyPlaceholderResolver

helios2k6:
# Summary
This PR fixes a regression that was introduced in PR #9701, which removed a try-catch statement within an implementation of the `Segment::findValue()` function.

The contract for this interface implies that no exception should be thrown from it, and indeed the default implementation directly catches `ConfigurationException`s and returns `Optional.empty()`. 

The regression manifests itself as an exception when trying to resolve property placeholders that are expected to be nullable, and thus are not present in the environment or any other property resolver. 

Prior behavior simply returned a value of null for these missing and nullable placeholders, which is the expected behavior.

This PR restores this behavior.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/DefaultPropertyPlaceholderResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/DefaultPropertyPlaceholderResolver.java
@@ -394,14 +394,18 @@ public class DefaultPropertyPlaceholderResolver implements PropertyPlaceholderRe
 
         @Override
         public <T> Optional<T> findValue(Class<T> type) {
-            for (String expression: expressions) {
-                Optional<T> optionalValue = resolveOptionalExpression(expression, type);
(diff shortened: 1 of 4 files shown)


graemerocher:
Adding a test case that reproduces this codepath would be good

helios2k6:
I would, but I actually don't know Groovy. Is there any guidance on how to write unit tests for Micronaut?

graemerocher:
You can write the test in java here https://github.com/micronaut-projects/micronaut-core/tree/4.5.x/test-suite/src/test/java/io/micronaut/docs/config/value

helios2k6:
Understood. Writing it now.

helios2k6:
@graemerocher PR + UTs are ready.

graemerocher (approved):

sdelamo (approved):

PR #10805, opened 2024-05-07
Improve ability to introspect bean context prior to start

graemerocher:
The existing `InitializableBeanContext` is problematic because it still creates context scoped beans and runs event listeners. We need a way to configure the context and allow it to be introspected prior to start so this deprecates `InitializableBeanContext` and replaces it with `ConfigurableBeanContext` and `ConfigurableApplicationContext` interfaces.

These feature a `configure()` method that loads the bean definitions and state needed to start the context but doesn't actually start it.

The PR also adds logic to assert the context is running when resolving beans which was missing before and could lead to odd behaviour.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/context/ApplicationContextConfigurerSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/context/ApplicationContextConfigurerSpec.groovy
@@ -1,0 +1,16 @@
+package io.micronaut.inject.context
+
+import io.micronaut.context.ApplicationContext
+import spock.lang.Specification
+
+class ApplicationContextConfigurerSpec
+    extends Specification {
+
+    void 'test context configurer registers bean'() {
+        given:
+        def ctx = ApplicationContext.run()
+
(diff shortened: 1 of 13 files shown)


sdelamo (approved):

yawkat (approved):

melix (approved):
I have checked that these changes to [not break AOT (without changing the code)](https://ge.micronaut.io/s/ectidnm4ihzoe/dependencies?toggled=W1szXSxbMywxXSxbMywxLFsyXV1d) and also [after changing it](https://ge.micronaut.io/s/46wehwexcocou/tests/overview) to call the new `configure()` method.

Both work correctly :+1: 

Note that we were already [using reflection](https://github.com/micronaut-projects/micronaut-aot/blob/bae7f171bafc5228b7ff887ebeadf45740796195/aot-core/src/main/java/io/micronaut/aot/core/context/ApplicationContextAnalyzer.java#L98) to configure the context, so I could probably do something similar to support it right now (and probably remove the code for older Micronaut releases).

PR #10820, opened 2024-05-13
Make behaviour of @DefaultImplementation consistent

graemerocher:
The `@DefaultImplementation` annotation is currently confusing since it doesn't influence bean selection but only the result of `@Replaces`. This PR makes it so that the annotation can also be used to control bean selection which makes it more logical and less confusing.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -3241,16 +3241,14 @@ public class DefaultBeanContext implements InitializableBeanContext {
         }
 
         // try resolve @DefaultImplementation
-        for (BeanDefinition<T> candidate : candidates) {
-            if (candidate.hasStereotype(DefaultImplementation.class)) {
-                String n = candidate.stringValue(DefaultImplementation.class, "name").orElse(null);
-                if (n != null) {
-                    for (BeanDefinition<T> bd : candidates) {
-                        if (bd.getBeanType().getName().equals(n)) {
-                            return candidate;
-                        }
+        BeanDefinition<T> first = candidates.iterator().next();
+        if (first.hasStereotype(DefaultImplementation.class)) {
+            String n = first.stringValue(DefaultImplementation.class, "name").orElse(null);
+            if (n != null) {
+                for (BeanDefinition<T> bd : candidates) {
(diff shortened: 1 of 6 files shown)


sdelamo (changes requested):
I think this change is a good idea but we should document this in the "Default Implementation" section of the docs. We should mention that @Primary still has precedence over @DefaultImplementation

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -3240,6 +3240,20 @@ private <T> BeanDefinition<T> lastChanceResolve(Argument<T> beanType,
             return bean;
         }
 
+        // try resolve @DefaultImplementation
+        for (BeanDefinition<T> candidate : candidates) {
+            if (candidate.hasStereotype(DefaultImplementation.class)) {
```
dstepanov: Checking the first candidate should be enough to know if there is a default impl

dstepanov:
Personally, I don't like the idea of `@DefaultImplementation` defined on the interface, considering this can be implemented with priority.

I think this will not work correctly in a case `class SomeImpl{X,Y, Z} extends SomeBase; abstract class SomeBase implements ISomeWithDefault` and injecting `SomeBase`. The correct way would be to use some annotation processing to map the class name on which the annotation is applied to `@DefaultImplementation`  and check if the inject point is trying to resolve that class.

graemerocher:
right, I added this because the current implementation is confusing and because Guice supports this concept https://github.com/google/guice/blob/2b37f020862e45be58f246a63c2f97fdaff0d433/core/test/com/google/inject/ImplicitBindingTest.java#L54

The example you mention should work fine, we just need to make sure the implementing type matches the type to be injected.

If we don't agree to this change then I can maybe implement is some other way in the Guice bridge, but I would then suggest we deprecate `@DefaultImplementation` all together because the only place it is used is with `@Replaces` and the name and behaviour are very misleading

dstepanov:
I'm OK with this. Does it work to not have any scope defined on the implementation like it's in the JavaDoc of `DefaultImplementation`?

dstepanov (approved):

graemerocher:
yes it does

sdelamo (approved):

PR #10845, opened 2024-05-21
Small performance improvements

dstepanov:


graemerocher (approved):

altro3:
@dstepanov could you revert changes in MediaType? I added them here: https://github.com/micronaut-projects/micronaut-core/pull/10846  so as not to solve problems with conflicts later

core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
```diff
@@ -103,7 +103,8 @@ public static PropagatedContextImpl getOrEmpty() {
 
     @Override
     public PropagatedContextImpl plus(PropagatedContextElement element) {
-        PropagatedContextElement[] newElements = Arrays.copyOf(elements, elements.length + 1);
+        PropagatedContextElement[] newElements = new PropagatedContextElement[elements.length + 1];
+        System.arraycopy(elements, 0, newElements, 0, elements.length);
```
yawkat, line 107: I highly doubt this makes any difference

core/src/main/java/io/micronaut/core/util/StringUtils.java
```diff
@@ -16,6 +16,7 @@
 package io.micronaut.core.util;
 
 import io.micronaut.core.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
```
yawkat: wrong annotation

inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
```diff
@@ -804,7 +804,8 @@ public B build(Object... builderParams) {
                 if (constructorLength == params.length) {
                     return introspection.instantiateInternal(params);
                 } else {
-                    Object[] constructorParams = Arrays.copyOfRange(params, 0, constructorLength);
+                    Object[] constructorParams = new Object[constructorLength];
+                    System.arraycopy(params, 0, constructorParams, 0, constructorLength);
```
yawkat, line 808: same here

yawkat (changes requested):

core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
dstepanov: I see reflection access because of `Array.newInstance(newType.getComponentType(), newLength);`

yawkat: Even after JIT? copyOf is an intrinsic method

dstepanov: Screenshot 2024-05-22 at 14 32 54

yawkat: does not look like reflection

altro3: изображение
```java
Array.newInstance(newType.getComponentType(), newLength);
```
@dstepanov So, this method doesn't use reflection. If you look to the `Array.newInstance` method, you will see this:
изображение
 and
изображение

@yawkat But I agree with @dstepanov , his code is faster than `Array.copyOf`

dstepanov: It doesn't, but it does access some `reflect` packages that might trigger the creation of reflection metadata. Jonas is right; the method is marked for being replaced with a native method, but somehow, JVM has a problem deciding to remove it.

PR #10850, opened 2024-05-23
Interceptor should intercept non-abstract methods for abstract classes

### LINKED ISSUE #2943: `@Transactional` annotation misbehavior
### Expected Behavior

When `@Transactional` is applied at a class level, all non-private methods should be automatically wrapped in a transaction and have a connection available.

When `@Transactional` is applied at a method level (non-private), it should be wrapped in a transaction and have a connection available.

When using `MockBean`, `TransactionalInterceptor` shouldn't be applied (as it expects a `DataSource` to be created and the idea of mocking a repository is not having to have one).

It doesn't matter if it is:
- `@jakarta.transaction.Transactional`
- `@io.micronaut.transaction.annotation.Transactional`

The bug manifests itself with both annotations.

### Actual Behaviour

When `@Transactional` is applied at a class level, all non-private methods are not automatically wrapped in a transaction and/or don't have a connection available. The following exception is thrown: `io.micronaut.data.connection.exceptions.NoConnectionException: Expected an existing connection, but none was found.`

When `@Transactional` is applied at a method level, tests involving those methods fail (see this [Micronaut test bug I created](https://github.com/micronaut-projects/micronaut-test/issues/1025) and [duplicated here in Micronaut-core repo](https://github.com/micronaut-projects/micronaut-core/issues/10813)).

It doesn't matter if it is `@jakarta.transaction.Transactional` or `@io.micronaut.transaction.annotation.Transactional`.

#### Bug test cases

| Annotation / Annotated on | Repository | Repository method | Tests execution result [^2] | Production execution result |
|---------------------------|------------|-------------------|-----------------------------|-----------------------------|
| `@Transactional` [^1]     | Yes        | No                | Pass                        | Fail                        |
| `@Transactional` [^1]     | No         | Yes               | Fail                        | Works well                  |
| `@Transactional` [^1]     | Yes        | Yes               | Fail                        | Works well                  |

[^1]: It doesn't matter whether it's `@jakarta.transaction.Transactional`
or `@io.micronaut.transaction.annotation.Transactional`.

[^2]: Tests mock repositories with `@MockBean(RepositoryClass.class)`.

### Steps To Reproduce

For your convenience, you can use the provided app example. But here are the steps anyway:

- Create an app with Micronaut Data JDBC + H2 driver (for instance).
- Create a Repository. Annotate it with both `@Repository` and `@Transactional` (at class level).
- Inject a `JdbcOperations` object and run a query against H2, which should have been populated beforehand (using Liquibase, for instance). Query can be a simple `SELECT * FROM mytable`.
- Run the app normally (you'll have to have some entrypoint such as a `@EventListener` or any other place where you inject the repository and call that method).

> **Important:** Please notice that to expose the misbehavior, you'll need to run the app in Prod mode (that is, run the `Application.java` class).

Interestingly, another similar error happens during test execution (depending on what you annotate with `@Transactional`). Please take a look at:
- [Micronaut test bug](https://github.com/micronaut-projects/micronaut-test/issues/1025)
- [duplicated bug in Micronaut-core repo](https://github.com/micronaut-projects/micronaut-core/issues/10813)

A workaround I tried (hoping to pass tests and not receiving exceptions at runtime) but without any luck is to annotate both the class and non-private methods with `@Transactional`. 

Thanks.

### Environment Information

- Windows/ Linux (it doesn't really matter).
- JDK 21 (I don't think it matters).
- jUnit 5 (I don't think it matters).
- Mockito (I don't think it matters).

### Example Application

https://github.com/Nahuel92/poc-projects/tree/master/micronaut/Micronaut-test-bug

I added my findings to the README.md file.

> Please set `MICRONAUT_ENVIRONMENTS=prod` before starting the app.

### Version

4.4.2

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-data/issues/2943

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/processing/AbstractBeanElementCreator.java
+++ b/core-processor/src/main/java/io/micronaut/inject/processing/AbstractBeanElementCreator.java
@@ -109,20 +109,20 @@ abstract class AbstractBeanElementCreator implements BeanDefinitionCreator {
         return memberElement.getAnnotationMetadata();
     }
 
-    protected boolean visitIntrospectedMethod(BeanDefinitionVisitor visitor, ClassElement typeElement, MethodElement methodElement) {
+    protected boolean visitIntrospectedMethod(BeanDefinitionVisitor visitor, ClassElement classElement, MethodElement methodElement) {
         AopProxyWriter aopProxyWriter = (AopProxyWriter) visitor;
 
-        final AnnotationMetadata resolvedTypeMetadata = typeElement.getAnnotationMetadata();
+        final AnnotationMetadata resolvedTypeMetadata = classElement.getAnnotationMetadata();
         final boolean resolvedTypeMetadataIsAopProxyType = InterceptedMethodUtil.hasDeclaredAroundAdvice(resolvedTypeMetadata);
 
         if (methodElement.isAbstract()
             || resolvedTypeMetadataIsAopProxyType
             || InterceptedMethodUtil.hasDeclaredAroundAdvice(methodElement.getAnnotationMetadata())) {
-            addToIntroduction(aopProxyWriter, typeElement, methodElement, false);
+            addToIntroduction(aopProxyWriter, classElement, methodElement, false);
             return true;
-        } else if (!methodElement.isAbstract() && methodElement.hasDeclaredStereotype(Executable.class)) {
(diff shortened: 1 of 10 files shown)


graemerocher (approved):

graemerocher (changes requested):
Breaks Kotlin tests

graemerocher (approved):

PR #10858, opened 2024-05-24
test: ignore revoked.badssl.com/

sdelamo:
Not sure if the certificate is revoked anymore. 

CleanShot 2024-05-24 at 13 33 27@2x


dstepanov (approved):

yawkat:
it is still revoked, just the revocation check never worked. #10853

PR #10859, opened 2024-05-24
Add HEAD_BODY attribute

yawkat:
For jax-rs tck


http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -564,6 +564,7 @@ private MutableHttpResponse<?> finaliseResponse(HttpRequest<?> request, RouteInf
                 referenceCounted.release();
             }
             response.body(null);
+            response.attribute(HttpAttributes.HEAD_BODY, o);
```
dstepanov, line 567: Isn’t it a bit strange to add a released body?

dstepanov:
What spec defines this behavior?

graemerocher (approved):

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
yawkat: We only need it in jax-rs for JaxRsResponse which is not ref counted

yawkat:
@dstepanov it's necessary for JaxRsResponseFilter. the body will not actually be sent, but JaxRsResponseFilter needs access to set e.g. headers

dstepanov (approved):

PR #10892, opened 2024-06-07
Fix race condition in task scheduling

yawkat:
Under some circumstances, ExecutableMethodProcessor.process may be called during or even after the StartupEvent is fired. In that case, scheduled methods may be ignored or there may be a ConcurrentModificationException.

This patch moves the schedule logic into its own method, called by a class ScheduleTaskRunnable that guards the schedule logic using an AtomicBoolean so that it is run only once. Inside process(), the task is created, registered in a ConcurrentHashMap and, if the StartupEvent has already been fired, invoked. Inside the StartupEvent listener, all previously registered tasks are invoked.

- If process is called before the started flag is set, the task will definitely be run by the StartupEvent listener.
- If process is called after the started flag is set, the task will definitely be called inside process, and potentially also inside the StartupEvent listener if that is still running.

This ensures that the ScheduleTaskRunnable is invoked at least once.


yawkat:
The diff is unfortunately pretty big, but you can ignore scheduleTask. It just contains the code that was previously in scheduleTasks.

graemerocher (approved):

PR #10902, opened 2024-06-13
EachBean: Allow to inject the origin bean and the registration

dstepanov:
For each bean functionality, we require a qualifier to properly identify the origin bean, but there are use cases where the bean cannot be modified.

Previously, we would allow each bean without a qualifier, but that will fail when you want to inject an origin; this PR is allowing to inject it. 

Another solution for the qualifiers would be to add a qualifier that matches the bean type by the exact class, and maybe it can be reused elsewhere.

This should target 4.6 after it's created

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/qualifiers/EachBeanQualifier.java
+++ b/inject/src/main/java/io/micronaut/inject/qualifiers/EachBeanQualifier.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017-2024 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.inject.qualifiers;
+
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.inject.BeanDefinition;
+import io.micronaut.inject.BeanType;
+
+/**
+ * Qualifies the origin bean definition that was used to create an each bean.
+ *
+ * @param <T> The type
+ * @author Denis Stepanov
+ * @since 4.6
+ */
+@Internal
+final class EachBeanQualifier<T> extends FilteringQualifier<T> {
+
+    private final BeanDefinition<?> beanDefinition;
+
+    EachBeanQualifier(BeanDefinition<?> beanDefinition) {
+        this.beanDefinition = beanDefinition;
+    }
+
+    @Override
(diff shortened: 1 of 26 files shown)


inject/src/main/java/io/micronaut/inject/qualifiers/Qualifiers.java
```diff
@@ -431,6 +432,18 @@ private static <T> Qualifier<T> findCustomByName(@NonNull AnnotationMetadata met
...
 
+    /**
+     * Qualifies the origin bean definition that was used to create an each bean.
+     *
+     * @param beanDefinition The beanDefinition
+     * @param <T>            The component type
+     * @return The qualifier
+     * @since 4.6
+     */
+    public static <T> Qualifier<T> eachBeanOf(BeanDefinition<?> beanDefinition) {
```
graemerocher: don't think we should make this public API since it quite internal specific. Either way I would name it `childOf` or `descendedFrom` something.

inject/src/main/java/io/micronaut/inject/qualifiers/EachBeanQualifier.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        EachBeanQualifier<?> that = (EachBeanQualifier<?>) o;
+
+        return beanDefinition.equals(that.beanDefinition);
+    }
+
+    @Override
+    public String toString() {
+        return "EachBeanQualifier('" + beanDefinition + "')";
```
graemerocher, line 59: what happens with a qualifier like this when I inject `Map<String, Foo>`? Please add tests for maps.

graemerocher (changes requested):

graemerocher:
I created the 4.6.x branch please target that. Also could you describe what issue this is trying to solve?

dstepanov:
I want to add:
```
@EachBean(jakarta.ws.rs.ext.MessageBodyWriter.class)
public final class JaxRsMessageBodyWriter<T> implements MessageBodyWriter<T> {

    private final jakarta.ws.rs.ext.MessageBodyWriter<T> messageBodyWriter;

    JaxRsMessageBodyWriter(jakarta.ws.rs.ext.MessageBodyWriter<T> messageBodyWriter) {
        this.messageBodyWriter = messageBodyWriter;
    }
```
But Jax-Rs "beans" don't have any qualifier that fails when the origin bean is injected into the constructor.

inject/src/main/java/io/micronaut/inject/qualifiers/EachBeanQualifier.java
dstepanov: At this moment, it fails because there is no name. I think in this case we should throw an error saying that `Map` cannot be used on multiple beans that don't have the name defined.

inject/src/main/java/io/micronaut/inject/qualifiers/Qualifiers.java
dstepanov: Made private

graemerocher (approved):

inject/src/main/java/io/micronaut/inject/qualifiers/EachBeanQualifier.java
dstepanov: I changed the message a bit reflecting that it doesn't need to be each bean + tests

PR #10909, opened 2024-06-14
TCK: constraint handler improve

andriy-dmytruk:
This would solve this issue: https://github.com/micronaut-projects/micronaut-oracle-cloud/issues/925 that the oracle cloud HTTP functions are facing with server TCK.


graemerocher:
probably this is small enough to target 4.5.x

graemerocher (approved):

yawkat:
I don't think this change is correct. I believe we have code that relies on the controller body type being set in the request in case of validation error.

@sdelamo do you know more? since you wrote this tck test.

sdelamo:
yes,  the test was verifying that you could access the method annotated with `@Error` the request body with the same type used in the `@Post` methods.

yawkat:
Then we need to revert this

PR #10918, opened 2024-06-18
Type pollution test

yawkat:
Add a test to the benchmarks module that runs certain JMH benchmarks with the redhat type pollution agent ( https://github.com/RedHatPerf/type-pollution-agent ) and triggers a test failure if there is excessive type pollution.

This works by adding a new test task which includes the agent, and then using JFRUnit to access the JFR events emitted by the agents. If there are too many thrash events for a single concrete type, details are logged and the test fails.

Additionally:
- Add a JMH benchmark that mimics the TechEmpower benchmark
- Fix various type pollution issues that were found using the FullHttpStackBenchmark

This PR is still pending the upstream PR to introduce JFR events to the agent ( https://github.com/RedHatPerf/type-pollution-agent/pull/19 ), and actual publishing of the type pollution agent. We will either wait for RH to publish the agent to maven central, or we will publish it ourselves if they don't want to do it (reply pending).


router/src/main/java/io/micronaut/web/router/DefaultRouter.java
```diff
@@ -75,11 +75,12 @@ public class DefaultRouter implements Router, HttpServerFilterResolver<RouteMatc
...
     private final List<FilterRoute> preconditionFilterRoutes = new ArrayList<>();
-    private final Supplier<List<GenericHttpFilter>> alwaysMatchesHttpFilters = SupplierUtil.memoized(() -> {
+    // ArrayList to avoid interface checkcast
+    private final Supplier<ArrayList<GenericHttpFilter>> alwaysMatchesHttpFilters = SupplierUtil.memoized(() -> {
         if (alwaysMatchesFilterRoutes.isEmpty()) {
-            return Collections.emptyList();
+            return new ArrayList<>(0);
         }
-        List<GenericHttpFilter> httpFilters = new ArrayList<>(alwaysMatchesFilterRoutes.size());
+        ArrayList<GenericHttpFilter> httpFilters = new ArrayList<>(alwaysMatchesFilterRoutes.size());
```
dstepanov, line 83: Can you just use `var`?

router/src/main/java/io/micronaut/web/router/DefaultRouter.java
```diff
@@ -75,11 +75,12 @@ public class DefaultRouter implements Router, HttpServerFilterResolver<RouteMatc
     private Set<Integer> ports;
     private final List<FilterRoute> alwaysMatchesFilterRoutes = new ArrayList<>();
     private final List<FilterRoute> preconditionFilterRoutes = new ArrayList<>();
-    private final Supplier<List<GenericHttpFilter>> alwaysMatchesHttpFilters = SupplierUtil.memoized(() -> {
+    // ArrayList to avoid interface checkcast
+    private final Supplier<ArrayList<GenericHttpFilter>> alwaysMatchesHttpFilters = SupplierUtil.memoized(() -> {
         if (alwaysMatchesFilterRoutes.isEmpty()) {
-            return Collections.emptyList();
+            return new ArrayList<>(0);
```
dstepanov, line 81: Please add a comment why this cannot be emptyList or List.of

dstepanov (approved):

melix (approved):

graemerocher (approved):

router/src/main/java/io/micronaut/web/router/DefaultRouter.java
yawkat: I changed the field to be ArrayList to avoid an interface cast, so this lambda has to return ArrayList as well

PR #10921, opened 2024-06-19
Uri template rewrite

dstepanov:
This is an attempt to get rid of the old `UriTemplate`, `UriMatchTemplate`, and `UriTypeMatchTemplate` (not used in Core). The implementation of matching/expanding is very confusing and hard to understand, with all the things being changed in the inherited constructor methods, etc. It does not allow adding any changes that we might need.

The new implementation is split into a parser, an expander, and an alternative URI template matcher. The matcher now matches segments of the path, allowing to match without always using the regexp; this should improve the performance of matching basic `/hello` or `/hello/{world}.` In the future, we can implement routing, which will combine similar segments into one, reducing the complexity for N to something better, considering most routes share the same prefix.

The new matcher passes previous tests and is now used to match the routes. However, there are still some cases where the old template is used: resolving conflicts and the URI builder. Next, I will investigate what needs to be aligned with the JAX-RS implementation.

Unfortunately, the classes are public and cannot be removed or changed, so we have a new implementation until v5.

### DIFF
--- a/http/src/main/java/io/micronaut/http/uri/DefaultUriMatchInfo.java
+++ b/http/src/main/java/io/micronaut/http/uri/DefaultUriMatchInfo.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2017-2024 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


yawkat (approved):

graemerocher (approved):

dstepanov:
@graemerocher Pls merge

PR #10922, opened 2024-06-19
Revert "TCK: constraint handler improve"

yawkat:
Reverts micronaut-projects/micronaut-core#10909


andriy-dmytruk:
Could we add description to the test for the expected bahavior and reason for it if possible, please?

It is not entirely clear how edge cases should work. Are only bean bodies binded?For example, if a publisher is bound for post method, is the publisher also bound in the error handler? Is it sufficient if the HttpRequest stores this data?

PR #10937, opened 2024-07-02
Add config option for DNS resolution modes

yawkat:
Fixes LMP-266

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -811,7 +811,6 @@ public abstract class HttpClientConfiguration {
         this.dnsResolutionMode = dnsResolutionMode;
     }
 
-
     /**
      * Name of a fixed netty AddressResolverGroup to use for this client, or {@code null} to
      * instead use {@link #getDnsResolutionMode()}. This option is specific to the netty client.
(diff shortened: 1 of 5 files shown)


http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -982,4 +1009,23 @@ public void setEnabled(boolean enabled) {
             this.enabled = enabled;
         }
     }
+
+    /**
+     * The DNS resolution mode.
+     */
+    public enum DnsResolutionMode {
```
graemerocher, line 1047: add `@since`  to the new methods and enums

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -352,6 +366,21 @@ private void initBootstrap() {
                 bootstrap.option(ChannelOption.valueOf(channelOption), v);
             }
         }
+
+        bootstrap.resolver(resolverGroup);
```
graemerocher, line 371: should we allow injecting a custom `AddressResolverGroup` using DI so users can override?

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
yawkat: I've added an option

graemerocher (approved):

PR #10940, opened 2024-07-03
Support unix domain sockets for the access log handler

yawkat:
Fixes LMP-319


graemerocher (approved):

graemerocher:
@yawkat can you fix the check style stuff.

PR #10950, opened 2024-07-05
Reintroduce build time init of compilation time metadata

graemerocher:
This PR reintroduces build time initialisation of compilation time metadata that was removed in 4.5.0

The changes allow Micronaut to support `--strict-image-heap` whilst still optimising for build time initialisation.

To achieve this evaluated expressions are now initialised at build time. This is safe because these cannot feature static initialisers. In addition enum introspections now no longer reference the enum values but lazily evaluate these.


core/src/main/java/io/micronaut/core/annotation/BuildTimeInit.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Annotate that stores references to class names that can be build time initialized inside bean definition metadata.
+ *
+ * <p>This annotation is currently designed for internal use only and not for public API.</p>
+ */
+@Experimental
+@UsedByGeneratedCode
+@Internal
+@Retention(RetentionPolicy.RUNTIME)
```
dstepanov, line 31: Make it `SOURCE` we don't need this annotation at runtime

dstepanov (changes requested):

core/src/main/java/io/micronaut/core/annotation/BuildTimeInit.java
graemerocher: unfortunately `SOURCE` annotations cannot be read by native image features so don't know how to do this

dstepanov: Are you sure? Because that's using our own AnnotationMetadata

graemerocher: yes but the runtime metadata

graemerocher: if I change it to `SOURCE` the test fails

dstepanov (approved):

PR #10961, opened 2024-07-09
Support binding RouteMatch in server filter methods

### LINKED ISSUE #10942: Expose AnnotationMetadata of the route match in server filters
### Feature description

I'd like to write a server request filter method that triggers based on controller method annotations, not hard-coded URL patterns. This can be done already using some internal knowledge and request attributes:

https://github.com/micronaut-projects/micronaut-core/blob/4f353fa79288d2ee56314ac2982586216590fffc/http-server/src/main/java/io/micronaut/http/server/cors/CrossOriginUtil.java#L56

but it should be exposed more directly in a documented way. Either by supporting annotations directly in `@ServerFilter` or by allowing the `AnnotationMetadata` to be injected.

yawkat:
This PR adds support for RouteMatch parameters to server filter methods.

Unfortunately, binding was previously fixed for the server and client. This code had no access to RouteMatch (http does not depend on router). I refactored the code to move the arg binding into BaseFilterProcessor. This is then overridden in the server implementation to add RouteMatch binding support.

I also moved ServerHttpRequest binding there while I was at it.

I've tried to keep the move limited. Most of the arg binding code is still in MethodFilter. This avoids having to expose a dozen private classes.

Fixes #10942


router/src/main/java/io/micronaut/web/router/ServerFilterRouteBuilder.java
```diff
@@ -102,6 +109,22 @@ private void applyMetadata(FilterRoute route, FilterMetadata metadata) {
...
+
+            @Override
+            protected @NonNull BaseFilterArgBinder prepareArgument(ExecutableMethod<?, ?> method, boolean isResponseFilter, FilterMethodBindingBuilder builder, Argument<?> argument) {
+                if (!argument.getAnnotationMetadata().hasStereotype(Bindable.class)) {
+                    if (argument.isAssignableFrom(ServerHttpRequest.class)) {
+                        return (FilterArgBinder) ctx -> (ServerHttpRequest<?>) ctx.request();
+                    }
+                    if (argument.isAssignableFrom(RouteMatch.class)) {
+                        if (!argument.isNullable()) {
+                            builder.addFilterCondition(ctx -> ctx.request().getAttribute(HttpAttributes.ROUTE_MATCH).isPresent());
```
graemerocher: shouldn't this always be conditional? I mean why would you want an error from the filter if there is not `RouteMatch`? Also this seems to indicate that `null` will never be passed.

router/src/main/java/io/micronaut/web/router/ServerFilterRouteBuilder.java
```diff
@@ -102,6 +109,22 @@ private void applyMetadata(FilterRoute route, FilterMetadata metadata) {
...
+                        }
+                        return (FilterArgBinder) ctx -> ctx.request().getAttribute(HttpAttributes.ROUTE_MATCH).orElse(null);
```
graemerocher: should probably be `orElseThrow()` and some internal server error or something

http/src/main/java/io/micronaut/http/filter/MethodFilter.java
```diff
@@ -93,144 +92,90 @@ record MethodFilter<T>(FilterOrder order,
...
+                        return ctx.failure();
                     }
-                    fulfilled[i] = ctx -> ctx.continuation;
-                } else if (continuationReturnType.getType().isAssignableFrom(MutableHttpResponse.class)) {
-                    continuationCreator = BlockingContinuationImpl::new;
-                    fulfilled[i] = ctx -> ctx.continuation;
+                    return null;
+                };
+            }
+        } else if (argumentType == FilterContinuation.class) {
```
graemerocher: might be good to cleanup all of this `if/else` into individual binders

dstepanov:
Looks overcomplicated. We can simple remove restriction: `if (annotationMetadata.hasStereotype(Bindable.class))` because it's being ignored in `BaseFilterProcessor` anyway. And add some inherited method in `BaseFilterProcessor`.

yawkat:
@dstepanov it's not possible to skip filter execution for non-routematch requests that way. it'd have to involve some changes to the binding logic that i'd rather avoid

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/RequestFilterTest.java
```diff
@@ -502,6 +516,11 @@ public CompletableFuture<MutableHttpResponse<String>> requestFilterCompletableRe
         public CompletionStage<MutableHttpResponse<String>> requestFilterCompletionResponse() {
             return CompletableFuture.completedStage(HttpResponse.ok("requestFilterCompletionResponse"));
         }
+
+        @RequestFilter("/request-filter/route-match")
+        public void requestFilterRouteMatch(MutableHttpRequest<?> request, RouteMatch<?> routeMatch) {
```
graemerocher, line 548: If we are keeping the nullable case it needs a test

graemerocher (changes requested):

dstepanov:
@yawkat Just create a new interface 

```java

interface FilterPredicate {

boolean test(MutablePropagatedContext mutablePropagatedContext,
        HttpRequest<?> request,
        @Nullable HttpResponse<?> response,
        @Nullable Throwable failure)

}

```

If `ArgumentBinder` implements `FilterPredicate` then convert it to `Predicate<FilterMethodContext>`

router/src/main/java/io/micronaut/web/router/ServerFilterRouteBuilder.java
yawkat: nullable -> no condition -> filter is executed even for static resources (with null as the arg)

not nullable -> condition -> filter is not executed for static resources (orElse is never hit)

yawkat:
got a bit more ugly because i have to check for nullability as well, but it works

dstepanov (approved):

graemerocher:
@yawkat can you add documentation?

yawkat:
@graemerocher it's documented in RequestFilter and ResponseFilter where we also document all the other supported argument types

graemerocher (approved):

PR #10962, opened 2024-07-09
Add a reflection util class to extract generic type and annotations for a class

dstepanov:
Originaly created in the validation module to pass some of the reflection TCK. Moving to Core to allow using it for some edge cases of Jakarta APIs.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
+++ b/inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
@@ -1,0 +1,242 @@
+/*
+ * Copyright 2017-2024 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context;
+
+import io.micronaut.core.annotation.AnnotationMetadata;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.type.Argument;
+import io.micronaut.inject.annotation.AnnotationMetadataHierarchy;
+import io.micronaut.inject.annotation.MutableAnnotationMetadata;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.AnnotatedParameterizedType;
+import java.lang.reflect.AnnotatedType;
+import java.lang.reflect.GenericArrayType;
+import java.lang.reflect.Method;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
+import java.lang.reflect.WildcardType;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The annotation reflection utils.
+ *
+ * @author Denis Stepanov
+ * @since 4.6.0
+ */
+@Internal
+public final class AnnotationReflectionUtils {
+
+    AnnotationReflectionUtils() {}
+
+    /**
+     * Find implementation as an argument with types and annotations.
+     *
+     * @param implementationType The implementation class of the interface
+     * @param implementedType    The implementedType type - interface or an abstract class
+     * @return The argument of the interface with types and annotations
+     * @since 4.6
+     */
+    @Nullable
+    public static Argument<?> findImplementationAsArgument(@NonNull Class<?> implementationType,
+                                                           @NonNull Class<?> implementedType) {
+        return findImplemented(implementationType, implementedType, Map.of());
+    }
+
+    @Nullable
+    private static Argument<?> findResolvedImplementation(Class<?> implementationType,
+                                                          Class<?> implementedType, // Interface or an abstract class
+                                                          AnnotatedType annotatedType,
+                                                          Map<String, AnnotatedType> resolvedVariables) {
+        Argument<?> argument = argumentOf(annotatedType, implementedType, resolvedVariables);
+        if (argument != null) {
+            if (implementationType.getAnnotations().length > 0) {
+                // Append implementation annotations if any
+                argument = Argument.of(
+                    argument.getType(),
+                    new AnnotationMetadataHierarchy(argument.getAnnotationMetadata(), annotationMetadataOf(implementationType)),
+                    argument.getTypeParameters());
+            }
+        }
+        return argument;
+    }
+
+    private static Argument<?> findImplemented(Class<?> implementationType,
+                                               Class<?> implementedType,
+                                               Map<String, AnnotatedType> resolvedVariables) {
+        if (!implementedType.isAssignableFrom(implementationType)) {
+            return null;
+        }
+
+        Class<?> superClass = implementationType.getSuperclass();
+        if (superClass != null && !Object.class.equals(superClass)) {
+            AnnotatedType annotatedSuperclass = implementationType.getAnnotatedSuperclass();
+            Argument<?> resolvedImplementation;
+            if (implementedType.equals(superClass)) {
+                resolvedImplementation = findResolvedImplementation(implementationType, implementedType, annotatedSuperclass, resolvedVariables);
+            } else {
+                resolvedImplementation = findImplemented(superClass, implementedType, resolvedVariables(annotatedSuperclass, superClass));
+            }
+            if (resolvedImplementation != null) {
+                return resolvedImplementation;
+            }
+        }
+        Class<?>[] interfaces = implementationType.getInterfaces();
+        AnnotatedType[] annotatedInterfaces = implementationType.getAnnotatedInterfaces();
+        for (int i = 0; i < annotatedInterfaces.length; i++) {
+            AnnotatedType annotatedInterface = annotatedInterfaces[i];
+            Class<?> classFromType = getClassFromType(annotatedInterface.getType());
+            Argument<?> resolvedImplementation;
+            if (implementedType.equals(classFromType)) {
+                resolvedImplementation = findResolvedImplementation(implementationType, implementedType, annotatedInterface, resolvedVariables);
+            } else {
+                resolvedImplementation = findImplemented(interfaces[i], implementedType, resolvedVariables(annotatedInterface, interfaces[0]));
+            }
+            if (resolvedImplementation != null) {
+                return resolvedImplementation;
+            }
+        }
+        return null;
+    }
+
+    private static Class<?> getClassFromType(Type type) {
+        if (type instanceof Class<?> classType) {
+            return classType;
+        }
+        if (type instanceof ParameterizedType parameterizedType) {
+            return getClassFromType(parameterizedType.getRawType());
+        }
+        if (type instanceof GenericArrayType) {
+            return Object[].class;
+        }
+        if (type instanceof WildcardType wildcardType) {
+            return getClassFromType(wildcardType.getUpperBounds()[0]);
+        }
+        throw new IllegalArgumentException("Unknown type: " + type);
+    }
+
+    private static Map<String, AnnotatedType> resolvedVariables(AnnotatedType annotatedType, Class<?> type) {
+        if (annotatedType instanceof AnnotatedParameterizedType parameterizedType) {
+            AnnotatedType[] actualTypeArguments = parameterizedType.getAnnotatedActualTypeArguments();
+            TypeVariable<? extends Class<?>>[] typeParameters = type.getTypeParameters();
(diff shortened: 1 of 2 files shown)


inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.Map;
+
+/**
+ * The annotation reflection utils.
+ *
+ * @author Denis Stepanov
+ * @since 4.6.0
+ */
+@Internal
+public final class AnnotationReflectionUtils {
```
graemerocher, line 56: can this be combined with the current `ReflectionUtils` class? Seems unnecessary to create a new one

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Find implementation as an argument with types and annotations.
+     *
+     * @param implementationType The implementation class of the interface
+     * @param implementedType    The implementedType type - interface or an abstract class
+     * @return The argument of the interface with types and annotations
+     * @since 4.6
+     */
+    @Nullable
+    public static Argument<?> findImplementationAsArgument(@NonNull Class<?> implementationType,
```
graemerocher: can we rename this `resolveArgumentFromGenerics` the name here is confusing imo

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,242 @@
...
+                if (!method.getDeclaringClass().equals(annotationType)) {
+                    continue;
+                }
+                try {
+                    Object value = method.invoke(annotation);
+                    if (value != null) {
+                        values.put(method.getName(), value);
+                    }
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
```
graemerocher: This should use `ReflectionUtils.invokeMethod`

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,242 @@
...
+    private static AnnotationMetadata annotationMetadataOf(AnnotatedElement annotatedElement) {
+        Annotation[] annotations = annotatedElement.getAnnotations();
+        if (annotations.length == 0) {
+            return AnnotationMetadata.EMPTY_METADATA;
+        }
+        MutableAnnotationMetadata mutableAnnotationMetadata = new MutableAnnotationMetadata();
+        for (Annotation annotation : annotations) {
+            Map<CharSequence, Object> values = new LinkedHashMap<>();
+            Class<? extends Annotation> annotationType = annotation.annotationType();
+            Method[] methods = annotationType.getMethods();
```
graemerocher, line 190: we should add reflection logging for Graal

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: That class doesn't access annotation metadata

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: Is it understandable that it will return `Argument.of(implementedType, ...)` not just a generic type?

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: Changed

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: How?

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+        if (annotatedType instanceof AnnotatedParameterizedType parameterizedType) {
+            AnnotatedType[] actualTypeArguments = parameterizedType.getAnnotatedActualTypeArguments();
+            TypeVariable<? extends Class<?>>[] typeParameters = type.getTypeParameters();
+            if (actualTypeArguments.length != typeParameters.length) {
+                return Map.of();
+            }
+            Map<String, AnnotatedType> resolvedVariables = new LinkedHashMap<>();
+            for (int i = 0; i < actualTypeArguments.length; i++) {
+                AnnotatedType actualTypeArgument = actualTypeArguments[i];
+                if (actualTypeArgument.getType() instanceof TypeVariable<?>) {
```
yawkat: this is very common, e.g. `ArrayList<T> implements List<T>`

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+        if (type instanceof GenericArrayType) {
+            return Object[].class;
+        }
+        if (type instanceof WildcardType wildcardType) {
+            return getClassFromType(wildcardType.getUpperBounds()[0]);
+        }
+        throw new IllegalArgumentException("Unknown type: " + type);
+    }
+
+    private static Map<String, AnnotatedType> resolvedVariables(AnnotatedType annotatedType, Class<?> type) {
```
yawkat: why do you need to pass type here? it should always be the raw type of `annotatedType`, no?

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+
+    private static Class<?> getClassFromType(Type type) {
+        if (type instanceof Class<?> classType) {
+            return classType;
+        }
+        if (type instanceof ParameterizedType parameterizedType) {
+            return getClassFromType(parameterizedType.getRawType());
+        }
+        if (type instanceof GenericArrayType) {
+            return Object[].class;
```
yawkat: should be based on getGenericComponentType

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+            }
+            List<Argument<?>> list = new ArrayList<>();
+            for (int i = 0; i < annotatedActualTypeArguments.length; i++) {
+                TypeVariable<? extends Class<?>> typeVariable = typeParameters[i];
+                Class<?> variableBaseType = null;
+                AnnotatedType resolvedVariable = resolvedVariables.get(typeVariable.getName());
+                if (resolvedVariable == null) {
+                    resolvedVariable = annotatedActualTypeArguments[i];
+                    TypeVariable<? extends Class<?>> typeParameter = typeParameters[i];
+                    variableBaseType = getClassFromType(typeParameter.getBounds()[0]);
```
yawkat: getClassFromType does not support TypeVariable but `typeParameter.getBounds()[0]` can be a TypeVariable

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+            }
+            if (resolvedImplementation != null) {
+                return resolvedImplementation;
+            }
+        }
+        Class<?>[] interfaces = implementationType.getInterfaces();
+        AnnotatedType[] annotatedInterfaces = implementationType.getAnnotatedInterfaces();
+        for (int i = 0; i < annotatedInterfaces.length; i++) {
+            AnnotatedType annotatedInterface = annotatedInterfaces[i];
+            Class<?> classFromType = getClassFromType(annotatedInterface.getType());
```
yawkat: isnt this just `interfaces[i]`

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+            } else {
+                resolvedImplementation = findImplemented(superClass, implementedType, resolvedVariables(annotatedSuperclass, superClass));
+            }
+            if (resolvedImplementation != null) {
+                return resolvedImplementation;
+            }
+        }
+        Class<?>[] interfaces = implementationType.getInterfaces();
+        AnnotatedType[] annotatedInterfaces = implementationType.getAnnotatedInterfaces();
+        for (int i = 0; i < annotatedInterfaces.length; i++) {
```
yawkat: not sure if it's worth it, but maybe you can merge this loop with the superclass type?

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
yawkat: i think this method needs the previous resolved arguments to resolve these type vars

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+     */
+    @Nullable
+    public static Argument<?> findImplementationAsArgument(@NonNull Class<?> implementationType,
+                                                           @NonNull Class<?> implementedType) {
+        return findImplemented(implementationType, implementedType, Map.of());
+    }
+
+    @Nullable
+    private static Argument<?> findResolvedImplementation(Class<?> implementationType,
+                                                          Class<?> implementedType, // Interface or an abstract class
```
yawkat: implementedType is always the raw type of annotatedType

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
```diff
@@ -0,0 +1,239 @@
...
+            return resolvedVariables;
+        }
+        return Map.of();
+    }
+
+    @Nullable
+    private static Argument<?> argumentOf(@NonNull AnnotatedType type,
+                                          @NonNull Class<?> interfaceType,
+                                          @NonNull Map<String, AnnotatedType> resolvedVariables) {
+        if (type.getType() instanceof TypeVariable<?> typeVariable) {
```
yawkat: can this happen? i think type is always a concrete type?

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
graemerocher: log the methods accessed reflectively with `ClassUtils.REFLECTION_LOGGER`

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
graemerocher: Perhaps put it into `ClassUtils` in that case

dstepanov: The same problem as `ReflectionUtils`, `core` doesn't see mutable annotation metadata

graemerocher (approved):

yawkat (changes requested):

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: How else would you get type variable names?

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: It only resolve the generic type of the implemented class not of the superclasses / interfaces. If you are loocking for that we need to geet it from the resolvd variables.

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: Not sure what do you mean

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
dstepanov: We don't really support 100% cases, if something else we can try to fix it

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
yawkat: i would just resolve the type by getting the raw type of `annotatedType`. it's a bit more clear imo. but if you want to keep it for performance, maybe make it clear that one is the raw type of the other

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
yawkat: when you have `interface A<T>`, `interface B<T> extends A<T>`, `class C extends B<String>`, this code will fail to propagate the String all the way to A

inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
yawkat: what I mean is that I believe that argumentOf is always called with one of the types that is directly declared inside an extends or implements expression, so it can never be a type variable (you cant `class A<T> extends T`)

dstepanov:
@yawkat Would you like to refactor it? I don't really understand much of the reflection

yawkat:
i can try tomorrow, meetings all day today

yawkat:
@dstepanov @graemerocher i've pushed a different implementation. I don't think it's much better, but it is a bit more complete

dstepanov:
Looks fine

PR #10971, opened 2024-07-11
Correct parsing locale in headers

dstepanov:
Underscore notation should be supported


yawkat:
I don't think so. Underscore notation violates the spec: https://httpwg.org/specs/rfc9110.html#field.accept-language

dstepanov:
You are right. The problem was in the builder on my side that converts locale.

Issue #10235, opened 2023-12-08
MutableHttpRequest doesn't set multiple cookies properly


j1m-renwick:
### Expected Behavior

MutableHttpRequest (specifically NettyMutableHttpRequest)'s `cookies(Set<Cookie>)` method should result in all cookies being set in the request

### Actual Behaviour

`cookies(Set<Cookie>)` results in only one cookie being set (the last one in the set).  

This appears to be an issue with the NettyMutableHttpRequest `cookie` method, which overrides any existing cookie header with the new cookie input.

### Steps To Reproduce

invoke the following code in an http-enabled project and observe the resulting request
```
@ServerFilter("/**")
class IncomingRequestFilter {

    @RequestFilter
    @ExecuteOn(TaskExecutors.BLOCKING)
    void filterRequest(MutableHttpRequest<?> request) {
        Cookie newCookie2 = Cookie.of("COOKIE_1", "one")
        Cookie newCookie = Cookie.of("COOKIE_2", "two")

        Set<Cookie> cookies = new HashSet<>()
        cookies.addAll(newCookie)
        cookies.addAll(newCookie2)

        request.cookies(cookies)
    }
}
```

### Version

4.1.6

sordell:
Hi, Micronaut noob (not even sure as a non-contributor if I'm supposed to post questions here); apologies in advance. This issue is exactly what I am experiencing using 4.2.4 so I was excited to see this fixed in 4.6.3. But after upgrading I'm seeing the same issue. Specifically, the cookies are set correctly (response.getCookies) but the headers are not. Specifically, both cookies are getting set in one Set-Cookie header (comma separated). According to the spec this is considered bad practice; each cookie should get its own Set-Cookie header and shouldn't be combined. I can't say for sure the fact that Micronaut seems to combine them into one Set-Header is the source of our problem but we are seeing this exact problem; when we look at the cookies in the browser only the second one is set in the cookie cache.

<img width="1049" alt="Image" src="https://github.com/user-attachments/assets/e13fe7c7-0c05-43b6-8a5d-9c74d62fc537" />

<img width="465" alt="Image" src="https://github.com/user-attachments/assets/08fe1a72-0586-45d4-b812-c239638d32b4" />

PR #10972, opened 2024-07-11
`NettyMutableHttpResponse` should support `getCookies`

dstepanov:


yawkat (approved):
Could also fix #10235 while you're at it

PR #10982, opened 2024-07-16
Can't find micronautserviceconfigs when running as jlink-ed application

### LINKED ISSUE #10842: Can't find micronautserviceconfigs when running as jlink-ed application
### Expected Behavior

Even though micronaut is not yet fully modularized, there are some steps to be taken before #6395 can be solved. This is an important one.
Root cause is the io.micronaut.core.io.service.ServiceScanner.findMicronautMetaServiceConfigs. If the classloader is coming from a ModuleLayer, non-specific resources (let's say directories) won't be found.

### Actual Behaviour

It fails with the following exception:
```
Exception in thread "main" io.micronaut.context.exceptions.NoSuchBeanException: No bean of type [io.micronaut.context.event.ApplicationEventPublisher<io.micronaut.context.event.StartupEvent>] exists.
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultBeanContext.newNoSuchBeanException(DefaultBeanContext.java:2768)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultApplicationContext.newNoSuchBeanException(DefaultApplicationContext.java:309)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultBeanContext.resolveBeanRegistration(DefaultBeanContext.java:2737)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:1731)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:837)
        at io.micronaut.di@4.4.9/io.micronaut.context.BeanLocator.getBean(BeanLocator.java:96)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultBeanContext.publishEvent(DefaultBeanContext.java:1817)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultBeanContext.start(DefaultBeanContext.java:357)
        at io.micronaut.di@4.4.9/io.micronaut.context.DefaultApplicationContext.start(DefaultApplicationContext.java:202)
```

### Steps To Reproduce

A new ClasspathResourceLoaderSpec would probably look a bit like this (I'm struggling to get Gradle tests running), just make sure there's a jar available as test-resource with the specified modulename.

    void "test resolving a resource using ModulePath"() {
        given:
        String modulename = "core.foo.bar" 
        ModuleFinder finder = ModuleFinder.of(Path.of("build/resources/test"))
        ModuleLayer parent = ModuleLayer.boot()
        Configuration cf = parent.configuration().resolve(finder, ModuleFinder.of(), Set.of(modulename))
        ClassLoader scl = ClassLoader.getSystemClassLoader()
        ModuleLayer layer = parent.defineModulesWithOneLoader(cf, scl)
        ClassPathResourceLoader loader = new DefaultClassPathResourceLoader(layer.findLoader(modulename), base)

        expect:
        loader.getResource(resource).get().text == "bar.txt"

        where:
        base             | resource
        null             | "foo/bar.txt"
        "foo"            | "bar.txt"
        "/foo"           | "bar.txt"
    }

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.4.9

rfscholte:
These few lines of code should be the base of the fix. And it actually doesn't require a predefined set. Tests are still missing as I'm struggling with the spock support in my IDE. 

This fixes #10842


graemerocher:
You can write the test in JUnit 5 / java if it is easier for you

graemerocher:
could you sign the CLA? Thanks.

core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
```diff
@@ -187,6 +189,20 @@ private void findMicronautMetaServiceConfigs(BiConsumer<URI, String> consumer) t
             uniqueURIs.add(uri);
         }
 
+        if(uniqueURIs.isEmpty()) {
```
rfscholte: Not 100% sure if this emtpy-check is correct. For JLink it is, which was the usecase. In case of combining classpath and modulepath this might need to be revisited, although I wonder if people would do that.

graemerocher (approved):

graemerocher (changes requested):
Is `test.jar` missing from the commit? The test appears to be failing and `"build/resources/test/test.jar"` doesn't exist

graemerocher:
//cc @rfscholte

rfscholte:
that's absolutely possible. Currently enjoying JCrete, while this code is still at home. I'll come back on this next week.

graemerocher:
thank you!

graemerocher:
unfortunately seems to break native image support. Let me investigate what can be done

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #10983, opened 2024-07-16
Improvements to HTTP client timeout handling

yawkat:
- Introduce new config properties to automatically send HTTP/2 pings on idle connections
- Introduce a new request-timeout property for the high-level timeout
- Remove ResettableReadTimeoutHandler now that it's upstream
- Adjust documentation of read-idle-timeout to match actual behavior (does not apply to streaming requests), add note to rename in 5.0
- Add note to event loop executor properties


http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -833,6 +865,69 @@ public void setAddressResolverGroupName(@Nullable String addressResolverGroupNam
...
     }
 
+    /**
+     * For HTTP/2 connections, the interval from the last inbound message to when an automated ping
+     * should be sent. This can be used to keep low-traffic connections alive.
+     *
+     * @return The timeout when to send a ping frame
+     */
+    @Nullable
+    public Duration getHttp2PingIntervalRead() {
```
dstepanov: Would it make sense to extract those into some http2 config? The property would be `http2.pingIntervalRead`

yawkat: i think then we'd have to duplicate it for the different HttpClientConfiguration classes too, im not sure how the config api works there

graemerocher: I agree with @dstepanov this would be better

yawkat: ok i will try it

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -413,12 +416,34 @@ public Optional<Duration> getReadTimeout() {
...
 
     /**
-     * For streaming requests and WebSockets, the {@link #getReadTimeout()} method does not apply instead a configurable
+     * The request timeout for non-streaming requests. This is the maximum time until the response
+     * must be completely received. Defaults to one second more than read-timeout.
+     *
+     * @return The request timeout
+     */
+    @NextMajorVersion("Set a default that isn't just requestTimeout+1 in DefaultHttpClient")
+    public Duration getRequestTimeout() {
```
graemerocher, line 427: add `@since`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -413,12 +416,34 @@ public Optional<Duration> getReadTimeout() {
...
+        return requestTimeout;
+    }
+
+    /**
+     * The request timeout for non-streaming requests. This is the maximum time until the response
+     * must be completely received. Defaults to one second more than read-timeout.
+     *
+     * @param requestTimeout The request timeout
+     */
+    public void setRequestTimeout(Duration requestTimeout) {
```
graemerocher: add nullability annotations to getter/setter

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -833,6 +859,17 @@ public void setAddressResolverGroupName(@Nullable String addressResolverGroupNam
...
     }
 
+    /**
+     * Obtains the HTTP/2 configuration.
+     *
+     * @return The HTTP/2 configuration.
+     * @since 4.6.0
+     */
+    @Nullable
+    public Http2Configuration getHttp2Configuration() {
```
graemerocher: maybe return `Optional<Http2Configuration>`

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -1039,6 +1076,82 @@ public void setEnabled(boolean enabled) {
         }
     }
 
+    public static class Http2Configuration {
```
graemerocher: add javadoc and `@since`

graemerocher: rename to `Http2ClientConfiguration` to avoid confusion

http-client-core/src/main/java/io/micronaut/http/client/ServiceHttpClientConfiguration.java
```diff
@@ -294,6 +317,13 @@ public static class ServiceConnectionPoolConfiguration extends ConnectionPoolCon
     public static class ServiceWebSocketCompressionConfiguration extends WebSocketCompressionConfiguration {
     }
 
+    /**
+     * The service HTTP/2 configuration.
+     */
+    @ConfigurationProperties(WebSocketCompressionConfiguration.PREFIX)
+    public static class ServiceHttp2Configuration extends Http2Configuration {
```
graemerocher: does this need to be `@BootstrapCompatible`?

http-client/src/main/java/io/micronaut/http/client/netty/Http2PingSender.java
```diff
@@ -0,0 +1,46 @@
...
+    public Http2PingSender(long readerIdleTime, long writerIdleTime, long allIdleTime, TimeUnit unit) {
+        super(readerIdleTime, writerIdleTime, allIdleTime, unit);
+    }
+
+    @Override
+    protected void channelIdle(ChannelHandlerContext ctx, IdleStateEvent evt) {
+        resetReadTimeout();
+        resetWriteTimeout();
+
+        ctx.writeAndFlush(new DefaultHttp2PingFrame(ThreadLocalRandom.current().nextLong()), ctx.voidPromise());
```
graemerocher, line 44: do we need to catch any exceptions here if the things go wrong writing the ping?

graemerocher (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
yawkat: I am matching what getWebSocketCompressionConfiguration does here

http-client-core/src/main/java/io/micronaut/http/client/ServiceHttpClientConfiguration.java
yawkat: I don't know. The other subconfigs arent either

http-client/src/main/java/io/micronaut/http/client/netty/Http2PingSender.java
yawkat: no the voidPromise and normal exception handling will take care of it

graemerocher (approved):

PR #10998, opened 2024-07-19
#9607: Endpoint path must support URI path as per rfc3986 appendix-B

### LINKED ISSUE #9607: Allow the use of dashes in management endpoint names
### Feature description

Consider below management endpoint definition:
```
import io.micronaut.management.endpoint.annotation.Endpoint;

@Endpoint("service-metadata")
public class ServiceMetadataController {
... 
}
```

### Current behaviour:
When defining a new management endpoint that contains dashes (e.g. /service-metadata) does not work, since when accessing this endpoint the application returns 404.

### Desired behaviour:
Allow defining management endpoints containing dashes. 
Ideally it should allow using any name for the management endpoint that is a valid URL path. 

The management endpoint id is verified against [this regex](https://github.com/micronaut-projects/micronaut-core/blob/4.0.x/management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java#L52).

pragyamutluru:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/9607
I have referred to DefaultUriBuilder.STRING_PATTERN_PATH to change the regex being checked in AbstractEndpointRouteBuilder.ENDPOINT_ID_PATTERN.

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
@@ -49,7 +49,7 @@ import java.util.regex.Pattern;
 @Internal
 abstract class AbstractEndpointRouteBuilder extends DefaultRouteBuilder implements ExecutableMethodProcessor<Endpoint>, LifeCycle<AbstractEndpointRouteBuilder> {
 
-    private static final Pattern ENDPOINT_ID_PATTERN = Pattern.compile("\\w+");
+    private static final Pattern ENDPOINT_ID_PATTERN = Pattern.compile("([^#?]*)");
 
     private Map<Class<?>, Optional<String>> endpointIds = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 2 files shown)


sdelamo:
@pragyamutluru you need to sign the contributor license agreement

management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
```diff
@@ -49,7 +49,7 @@
 @Internal
 abstract class AbstractEndpointRouteBuilder extends DefaultRouteBuilder implements ExecutableMethodProcessor<Endpoint>, LifeCycle<AbstractEndpointRouteBuilder> {
 
-    private static final Pattern ENDPOINT_ID_PATTERN = Pattern.compile("\\w+");
+    private static final Pattern ENDPOINT_ID_PATTERN = Pattern.compile("([^#?]*)");
```
yawkat: I think this is too wide. 
```suggestion
    private static final Pattern ENDPOINT_ID_PATTERN = Pattern.compile("[\\w-]+");
```

yawkat (changes requested):

management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
graemerocher: Agreed, what pattern do you suggest?

pragyamutluru: Updated code to narrow down the check as per suggestion. Kindly check the latest change.

yawkat (approved):

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #11002, opened 2024-07-22
Improve message body handlers

dstepanov:
- Previous raw handlers didn't allow to define a custom `String` handler of a different media type
- Raw handlers replaced by `Type*` reader/writer that include an argument.
- Extracted raw handlers to a separate writer / reader bean (no need to hide them with `@Bean(typed=`)
- Improved how dynamic body writer works, it will be always included from the registry for object types
- Introduced a separate writer/reader for a String and a JSON type. That is simply threads it a plain type.
- Small corrections


test-suite/src/test/java/io/micronaut/docs/server/consumes/ConsumesControllerSpec.java
```diff
@@ -77,7 +76,7 @@ void testConsumes() {
                 client.toBlocking().exchange(HttpRequest.POST("/consumes/multiple", book)
                 .contentType(MediaType.APPLICATION_JSON)));
 
-        assertThrows(CodecException.class, () ->
+        assertDoesNotThrow(() ->
```
dstepanov: This was failing because toString convertion was allowed only for primitives `ClassUtils.isJavaBasicType(object.getClass())`. I might revert it.

dstepanov: Reverted the change

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -493,6 +493,19 @@ static <T> Argument<T> of(
...
+    /**
+     * Creates a new argument for the type of the given instance.
+     *
+     * @param instance The argument instance
+     * @param <T>      The generic type
+     * @return The argument instance
+     * @since 4.6
+     */
+    @NonNull
+    static <T> Argument<T> of(T instance) {
```
yawkat: I don't think it's a good idea to overload `of` for this, it'll just lead to accidental calls

http-netty/src/test/groovy/io/micronaut/http/netty/body/CustomBodyReaderSpec.groovy
```diff
@@ -85,4 +94,16 @@ class CustomBodyReaderSpec extends Specification {
         }
     }
 
+    @Singleton
+    @Consumes("my/string")
+    static class MyStringTypeReader implements MessageBodyReader<String> {
+
+        @Override
```
yawkat:
```suggestion
```

http/src/main/java/io/micronaut/http/body/DynamicMessageBodyWriter.java
```diff
@@ -49,22 +48,13 @@ public MessageBodyWriter<Object> createSpecific(Argument<Object> type) {
         return registry.findWriter(type, mediaTypes).orElse(this);
     }
 
-    public MessageBodyWriter<Object> find(Argument<Object> type, MediaType mediaType, Object object) {
-        MessageBodyWriter<Object> specific = registry.findWriter(type, List.of(mediaType)).orElse(null);
-        if (specific != null && !(specific instanceof DynamicMessageBodyWriter)) {
-            return specific;
+    public <T> MessageBodyWriter<T> find(Argument<T> type, MediaType mediaType, T object) {
+        if (type.getType() == Object.class) {
+            type = Argument.of(object);
```
yawkat: if you have this logic you don't need to change all the callers. just keep them as OBJECT_ARGUMENT

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1322,8 +1325,8 @@ protected NettyRequestWriter buildNettyRequest(
                             new IllegalArgumentException("Unconvertible reactive type: " + bodyValue)
                         );
 
-                        Flux<HttpContent> requestBodyPublisher = Flux.from(publisher).map(o -> {
-                            ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, o, request.getHeaders(), byteBufferFactory);
+                        Flux<HttpContent> requestBodyPublisher = Flux.from(publisher).map(value -> {
+                            ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.of(value), requestContentType, value, request.getHeaders(), byteBufferFactory);
```
yawkat:
```suggestion
                            ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, value, request.getHeaders(), byteBufferFactory);
```

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1339,7 +1342,7 @@ protected NettyRequestWriter buildNettyRequest(
                     } else if (bodyValue instanceof CharSequence sequence) {
                         bodyContent = charSequenceToByteBuf(sequence, requestContentType);
                     } else {
-                        ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, bodyValue, request.getHeaders(), byteBufferFactory);
+                        ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.of(bodyValue), requestContentType, bodyValue, request.getHeaders(), byteBufferFactory);
```
yawkat:
```suggestion
                        ByteBuffer<?> buffer = dynamicWriter.writeTo(Argument.OBJECT_ARGUMENT, requestContentType, bodyValue, request.getHeaders(), byteBufferFactory);
```

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
```diff
@@ -213,21 +213,22 @@ private <T> Optional<T> convertByteBuf(Argument<T> type) {
             }
             return Optional.empty();
         }
+        // All content operation should call slice to prevent reading the buffer completely
```
yawkat, line 216: Why? Ownership should now be in the response

http-netty/src/main/java/io/micronaut/http/netty/body/NettyByteBufMessageBodyHandler.java
```diff
@@ -74,9 +77,8 @@ public ByteBuf read(Argument<ByteBuf> type, MediaType mediaType, Headers httpHea
 
     @Override
     public void writeTo(Argument<ByteBuf> type, MediaType mediaType, ByteBuf object, MutableHeaders outgoingHeaders, OutputStream outputStream) throws CodecException {
-        try {
-            new ByteBufInputStream(object).transferTo(outputStream);
-            object.release();
+        try (ByteBufInputStream byteBufInputStream = new ByteBufInputStream(object)) {
```
yawkat: Revert. This does not release properly

http-netty/src/main/java/io/micronaut/http/netty/body/NettyTextPlainHandler.java
```diff
@@ -46,17 +46,21 @@
...
 
     @Override
     public void writeTo(HttpRequest<?> request, MutableHttpResponse<CharSequence> outgoingResponse, Argument<CharSequence> type, MediaType mediaType, CharSequence object, NettyWriteContext nettyContext) throws CodecException {
+        writePlain(outgoingResponse, mediaType, object, nettyContext);
+    }
+
+    static void writePlain(MutableHttpResponse<?> outgoingResponse, MediaType mediaType, CharSequence object, NettyWriteContext nettyContext) {
         MutableHttpHeaders headers = outgoingResponse.getHeaders();
-        ByteBuf byteBuf = Unpooled.wrappedBuffer(object.toString().getBytes(MessageBodyWriter.getCharset(headers)));
+        ByteBuf byteBuf = Unpooled.wrappedBuffer(object.toString().getBytes(MessageBodyWriter.getCharset(mediaType, headers)));
```
yawkat:
```suggestion
        ByteBuf byteBuf = Unpooled.copiedBuffer(object.toString(), MessageBodyWriter.getCharset(mediaType, headers));
```

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
dstepanov: One reader would read the buffer completely and prevent others reading it.

core/src/main/java/io/micronaut/core/type/Argument.java
dstepanov: I can rename the method. Otherwise I will just inline

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
yawkat: We should not read the same body twice.

http/src/main/java/io/micronaut/http/body/DynamicMessageBodyWriter.java
dstepanov: But that's the call to the registry not to the dynamic writer. The dynamic writer will skip dynamic writers

http-netty/src/main/java/io/micronaut/http/netty/body/NettyByteBufMessageBodyHandler.java
dstepanov: Done

core/src/main/java/io/micronaut/core/type/Argument.java
yawkat: Either is fine with me

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
dstepanov: Why not? The body is already buffered in the client response

yawkat: I don't want to keep it that way forever

http/src/main/java/io/micronaut/http/body/DynamicMessageBodyWriter.java
yawkat: Then writeTo should also be adjusted. The whole point of DynamicMessageBodyWriter is to work with OBJECT_ARGUMENT

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
dstepanov: Then we need to find a way around it considering `FullNettyClientHttpResponse` is now reading the body with the predefined type in the constructor which will makes the buffer fully read. Different combinations now make interactions very unpleasant, sometimes the body is there sometimes it's not.

We can introduce something like `bufferEntity` in JAX-RS. I don't really want to dig more into it.

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -493,6 +493,19 @@ static <T> Argument<T> of(
...
+    /**
+     * Creates a new argument for the type of the given instance.
+     *
+     * @param instance The argument instance
+     * @param <T>      The generic type
+     * @return The argument instance
+     * @since 4.6
+     */
+    @NonNull
+    static <T> Argument<T> instanceType(T instance) {
```
graemerocher: add a null check to the instance argument and nullability annotation. Rename to `forInstance`

core/src/main/java/io/micronaut/core/type/MutableHeaders.java
```diff
@@ -53,4 +53,19 @@ default MutableHeaders set(CharSequence header, CharSequence value) {
...
+
+    /**
+     * Sets an HTTP header if missing.
+     *
+     * @param header The header
+     * @param value  The value
+     * @return This headers
+     * @since 4.6
+     */
+    default MutableHeaders setIfMissing(CharSequence header, CharSequence value) {
```
graemerocher: this calls `add` so should be named `addIfMissing`. Add nullability annotations to the arguments and null check the `header` argument

http/src/main/java/io/micronaut/http/body/AbstractMessageBodyHandlerRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Base class for {@link MessageBodyHandlerRegistry} that handles caching and exposes the raw
+ * handlers (String, byte[] and such).
+ *
+ * @author Graeme Rocher
+ * @since 4.0.0
+ */
+@Internal
+@Experimental
+abstract class AbstractMessageBodyHandlerRegistry implements MessageBodyHandlerRegistry {
```
graemerocher: can this be `sealed`?

http/src/main/java/io/micronaut/http/body/ByteBufferBodyHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * The body handler for {@link ByteBuffer}.
+ *
+ * @author Denis Stepanov
+ * @since 4.6
+ */
+@Requires(bean = ByteBufferFactory.class)
+@Singleton
+@BootstrapContextCompatible
```
graemerocher, line 46: can we reduce the number of bean definitions with a registrar type pattern?

http/src/main/java/io/micronaut/http/body/MessageBodyHandlerRegistry.java
```diff
@@ -70,4 +71,19 @@ <T> Optional<MessageBodyWriter<T>> findWriter(
         @NonNull Argument<T> type,
         @NonNull List<MediaType> mediaType
     );
+
+    /**
+     * Gets a writer for the type and annotation metadata at declaration point or fails.
```
graemerocher: document what exception is thrown on failure

http/src/main/java/io/micronaut/http/body/MessageBodyWriter.java
```diff
@@ -89,13 +90,54 @@ default boolean isBlocking() {
...
+    }
+
+    /**
+     * Resolve the charset.
+     *
+     * @param mediaType The mediaType
+     * @param headers   The headers
+     * @return The charset
+     */
+    static @NonNull Charset getCharset(@Nullable MediaType mediaType, @NonNull Headers headers) {
```
graemerocher, line 103: should these be moved to utility methods of `HttpHeaders`?

http/src/main/java/io/micronaut/http/body/TypedMessageBodyHandler.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A body handler {@link MessageBodyHandler} with a type argument.
+ *
+ * @param <T> The type
+ * @author Denis Stepanov
+ * @since 4.6
+ */
+@Experimental
+public interface TypedMessageBodyHandler<T> extends MessageBodyHandler<T>, TypedMessageBodyWriter<T>, TypedMessageBodyReader<T> {
```
graemerocher, line 30: should this be `@Internal`?

http/src/main/java/io/micronaut/http/body/TypedMessageBodyHandler.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @return The body type.
+     */
+    Argument<T> getType();
```
graemerocher, line 36: add nullability annotation

http/src/main/java/io/micronaut/http/body/TypedMessageBodyReader.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Denis Stepanov
+ * @since 4.6
+ */
+@Experimental
+public interface TypedMessageBodyReader<T> extends MessageBodyReader<T> {
+
+    /**
+     * @return The body type.
+     */
+    Argument<T> getType();
```
graemerocher, line 37: add nullability annotation

http/src/main/java/io/micronaut/http/body/TypedMessageBodyWriter.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Denis Stepanov
+ * @since 4.6
+ */
+@Experimental
+public interface TypedMessageBodyWriter<T> extends MessageBodyWriter<T> {
+
+    /**
+     * @return The body type.
+     */
+    Argument<T> getType();
```
graemerocher, line 37: add nullability annotation

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
```diff
@@ -137,9 +125,57 @@ private static CodecException decorateWrite(Object object, IOException e) {
...
+        MediaType.APPLICATION_JSON,
+        MediaType.TEXT_JSON,
+        MediaType.APPLICATION_HAL_JSON,
+        MediaType.APPLICATION_JSON_GITHUB,
+        MediaType.APPLICATION_JSON_FEED,
+        MediaType.APPLICATION_JSON_PROBLEM,
+        MediaType.APPLICATION_JSON_PATCH,
+        MediaType.APPLICATION_JSON_MERGE_PATCH,
+        MediaType.APPLICATION_JSON_SCHEMA
+    })
```
graemerocher, line 155: should this be configurable?

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
```diff
@@ -137,9 +125,57 @@ private static CodecException decorateWrite(Object object, IOException e) {
...
```
graemerocher, line 176: should this be configurable?

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/body/ByteBufferBodyHandler.java
dstepanov: Don't think so

http/src/main/java/io/micronaut/http/body/MessageBodyWriter.java
dstepanov: Not sure, I think in v5 we should change both read/writers to use `HttpHeaders` instead of `Headers`

http/src/main/java/io/micronaut/http/body/TypedMessageBodyHandler.java
dstepanov: It might be usefull at some point.

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
dstepanov: Maybe, but for now it's not, I just create this annotation to reduce repeated code

http/src/main/java/io/micronaut/http/body/DynamicMessageBodyWriter.java
dstepanov: Eliminated `DynamicMessageBodyWriter` completely

core/src/main/java/io/micronaut/core/type/MutableHeaders.java
dstepanov: That an alternative to the method above called `set` which will remove the previous value

graemerocher (approved):

http-netty/src/main/java/io/micronaut/http/netty/body/NettyRawJsonStringWriter.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Denis Stepanov
+ * @since 4.6
+ */
+@Singleton
+@Replaces(RawJsonStringHandler.class)
+@JsonMessageHandler.ProducesJson
+@JsonMessageHandler.ConsumesJson
+@Internal
+public final class NettyRawJsonStringWriter implements MessageBodyWriter<CharSequence>, MessageBodyReader<String>, NettyBodyWriter<CharSequence>, ChunkedMessageBodyReader<String> {
```
yawkat: why do we need separate writers per content type for this now? old behavior was that for these "raw types", the body would always be written the same way, regardless of content type

http-client/src/main/java/io/micronaut/http/client/netty/FullNettyClientHttpResponse.java
yawkat: ok let's leave it. i think it will need a bigger redesign eventually.

yawkat (approved):

http-netty/src/main/java/io/micronaut/http/netty/body/NettyRawJsonStringWriter.java
dstepanov: Not sure 100% how did it work before.
Without this the object type JSON writer will pick it up and convert it into the JSON. 
This writer is more correct because we actually want to convert strings as a plain text instead of wrapping it.

yawkat: Hm, you removed RawMessageHandlerRegistry, but didn't replace the lookup logic. I only see it in the ContextlessMessageBodyHandlerRegistry.

The old logic is very simple: If the type has a RawMessageHandler, that handler is used no matter what. This is the behavior from 3.x and it is the behavior we need to keep.

yawkat (changes requested):
see above comment on raw handlers

http-netty/src/main/java/io/micronaut/http/netty/body/NettyRawJsonStringWriter.java
dstepanov: Raw handlers were internal. Why would you disallow custom handlers?

dstepanov: The problem is if you want to handle the type in a different way. I don't think it's a problem if a user defines a custom handler, he chooses to change the behaviour.

yawkat: I don't really care if there's a way to define custom handlers. But for compatibility, we must stay with the current behavior, that the raw message handlers take full priority over the content-type based handlers.

dstepanov: I don't think there is any problem. Body readers / writers are since v4 and if a user chooses to define a custom one it's not our problem. WDYT @graemerocher ?

graemerocher: @yawkat @dstepanov  were any tests altered in this PR that introduce a breaking change in behaviour? @yawkat is the behaviour you describe tested somewhere?

yawkat: I believe it is only tested for json, we regularly return `String` from json controllers and expect it to be sent raw. @dstepanov fixed this for json specifically by introducing the NettyRawJsonStringWriter, but that means that now, a typed raw handler is needed for every content type that also declares a normal handler. So imo the existence NettyRawJsonStringWriter is hiding this issue by fixing it for one media type only.

dstepanov: Typed handlers are only used for client.
Having a raw string writer makes sense only for text types like XML, JSON. Protocol buffers will need to persist the string not just send it as bytes.
So I would say the behviour now is more correct, default convertion to string shouldn't work for every type.

yawkat: Can argue about whether String should be treated as a raw type all day, but not doing so would be a pretty major breaking change.

dstepanov: I can introduce a new writer with a higher order and `@Requires(property = "micronaut.http.legacy-always-string-writer", value = "true")` for people to enable it if needed.

yawkat: why can't we just have the old behavior?

dstepanov: Bacase that is esentially broken for writers that don't use textual formats like protocol buffers.

yawkat: that's how it's always been. If you want to change it, do so in 5.0.

dstepanov: So actually it was working like that anyway, there is no need for an extra string json handler. Removed unneeded handlers and added tests.

yawkat: thanks!

yawkat (approved):

PR #11003, opened 2024-07-22
Allow for `@EachBean` to delegate the generics to the newly created bean definition

dstepanov:


core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -219,6 +219,26 @@ default boolean isAssignableFrom(@NonNull Argument<?> candidateArgument) {
         return true;
     }
 
+    /**
+     * Creates a copy of this argument with a different name.
+     * @param name The new name
+     * @return A new argument
+     * @since 4.6
+     */
+    default Argument<T> withName(@Nullable String name) {
```
graemerocher, line 229: add nullability annotation to return type

core/src/main/java/io/micronaut/core/type/Argument.java
```diff
@@ -219,6 +219,26 @@ default boolean isAssignableFrom(@NonNull Argument<?> candidateArgument) {
...
+        return Argument.of(getType(), name, getAnnotationMetadata(), getTypeParameters());
+    }
+
+    /**
+     * Creates a copy of this argument with a different annotation metadata.
+     * @param annotationMetadata The annotation metadata
+     * @return A new argument
+     * @since 4.6
+     */
+    default Argument<T> withAnnotationMetadata(@NonNull AnnotationMetadata annotationMetadata) {
```
graemerocher, line 240: add nullability annotation to return type

graemerocher (approved):

yawkat (approved):

PR #11004, opened 2024-07-23
Fix idle timeouts for websockets and streaming connections in 3.8.x

yawkat:
Before this patch, the configured timeout would be applied for the read, write *and* combined streams. That means that if data was only sent in one direction on the connection, the timeout would trigger and the connection would close.

This patch changes the timeout to only apply when no data is sent in either direction.


yawkat:
@melix so, what do we do about PRs like this, now that jcenter is broken? Can't build or test

melix:
3.8.x ? wow. I guess it will require a fair amount of changes to fix the build :/

graemerocher:
do we need the PR given the workaround is to set 0 to the timeout?

PR #11006, opened 2024-07-23
Fix returning unit from suspended methods

### LINKED ISSUE #10992: `suspend` controller function returning `Unit` has a body set
### Expected Behavior

If an `@Controller` function is both `suspend`-able and returns `Unit` i would expect no body to be written to the caller.

### Actual Behaviour

After the coroutine is resumed, a `Unit` body is set on the response. This then causes issues later with [serialization](https://github.com/micronaut-projects/micronaut-serialization/issues/890) and sometimes causes an empty body `{}` to be written.

### Steps To Reproduce

- Clone project
- Set JDK/JRE to 21
- Run included tests
```shell
./gradlew test
```

### Environment Information

- macOS (arm)
- Java 21 & 17

### Example Application

https://github.com/tmapes/mn-serde-unit

### Version

4.5.0 platform, 4.5.3 core

yawkat:
The code was not working, it compared a RouteInfo against a RouteMatch. I added a test case.

Fixes #10992


dstepanov (approved):

graemerocher (approved):

graemerocher:
should this target 4.5.x?

yawkat:
how long until 4.6.0?

graemerocher:
we can release 4.6 of core when we want, we should discuss what else should be included

PR #11009, opened 2024-07-24
Fix body handler

dstepanov:


test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/annotation/processing/SuspendMethodSpec.kt
```diff
@@ -37,7 +37,7 @@ class SuspendMethodSpec {
         val res = Flux.from(client
             .retrieve(
                 HttpRequest.GET<Any>("/sync/any"),
-                Any::class.java
+                String::class.java
```
dstepanov, line 40: Not sure why this worked before

yawkat (approved):

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/annotation/processing/SuspendMethodSpec.kt
dstepanov: Fixed

dstepanov:
Had to add `TextPlainObjectBodyReader` for special cases like this

yawkat:
fine

PR #11011, opened 2024-07-24
Allow body readers to read form uploads

dstepanov:


graemerocher (approved):

dstepanov:
Need to fix some tck

http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
```diff
@@ -265,6 +265,7 @@ private ExecutionFlow<HttpResponse<?>> onErrorNoFilter(HttpRequest<?> request, T
 
         RouteMatch<?> errorRoute = routeExecutor.findErrorRoute(cause, declaringType, request);
         if (errorRoute != null) {
+            request.setAttribute(HttpAttributes.ROUTE_INFO, errorRoute);
```
dstepanov: Look like we forgot to change the route info for error an status routes

yawkat (approved):

PR #11017, opened 2024-07-26
Implement `@PreMatching` request filter methods

dstepanov:


dstepanov:
@yawkat How to fix the type polution check?

yawkat:
Wouldn't it be easier to just run the FilterRunner twice, once before once after routing?

yawkat:
The type pollution is caused by this RouteInfo stuff being moved before the conditional filter routes check: https://github.com/micronaut-projects/micronaut-core/blob/634b3b256da680114cd0a39174656ae70980dfc7/router/src/main/java/io/micronaut/web/router/DefaultRouter.java#L588

dstepanov:
> Wouldn't it be easier to just run the FilterRunner twice, once before once after routing?

Probably not, this change also eliminates multiple invocation of filtering for status routes etc.

yawkat:
I don't really like how specific the FilterRunner is to the server now

yawkat:
Is it maybe possible to make the routing decision a request filter and insert it between the pre- and post-match filters? I'm not a fan of the api changes to FilterRunner

dstepanov:
@yawkat Please check the latest version

graemerocher:
Can someone enlighten me what the problem is we are trying to solve here?

dstepanov:
Just to align with JaxRs. You can also have a filter that will change the request to hit a different route.

graemerocher:
ok are we adding an API to achieve that without JAX-RS? If so can we document it?

dstepanov:
It is documented. The is a new PreMatching annotation.

graemerocher:
ok but how does one change the route?

yawkat:
I think you'd have to change the request path.

http/src/main/java/io/micronaut/http/filter/FilterRunner.java
```diff
@@ -163,6 +252,16 @@ public final ExecutionFlow<HttpResponse<?>> run(HttpRequest<?> request,
         return flow.flatMap(context -> filterResponse(context, iterator, null));
     }
 
+    private List<InternalHttpFilter> filterFilters(List<InternalHttpFilter> filters, HttpRequest<?> request) {
+        List<InternalHttpFilter> filtersToRun = new ArrayList<>(filters.size() + 1);
```
yawkat, line 257:
```suggestion
        // 1 free spot for the RouteMatchResolverHttpFilter
        List<InternalHttpFilter> filtersToRun = new ArrayList<>(filters.size() + 1);
```

http-validation/src/main/java/io/micronaut/validation/routes/FilterVisitor.java
```diff
@@ -169,6 +170,9 @@ public void visitMethod(MethodElement element, VisitorContext context) {
                     }
                 }
             }
+            if (continuationCreator != null && element.hasStereotype(PreMatching.class)) {
```
yawkat: Maybe also check that this is a ServerFilter

yawkat (approved):

dstepanov:
@graemerocher The route is resolved after the pre-matching filter so you can simply modify the request

graemerocher:
but the request is not mutable. Does calling `mutate()` work? It would be good to have a test and example of mutating the route

dstepanov:
There are ways to mutate it:
```java
        @Order(1)
        @PreMatching
        @RequestFilter
        MutableHttpRequest<?> preMatchingRequest1(MutableHttpRequest<?> request) {
            return request.uri(new URI(request.getUri().toString().replace("xyz", "HELLO")))
        }

        @Order(2)
        @PreMatching
        @RequestFilter
        void preMatchingRequest2(MutableHttpRequest<?> request) {
            request.uri(new URI(request.getUri().toString().replace("HELLO", "abc")))
        }
  ```

dstepanov:
@yawkat The limitation of using the continuation was in my intial implementation, after the changes it also works in pre-matching

graemerocher (approved):
Ok, understood. Maybe fix the failing test :)

PR #11060, opened 2024-08-07
Fix usage of `@Header(name="..")`

### LINKED ISSUE #11053: AliasFor in annotation properties modify real values
Based on this: https://github.com/micronaut-projects/micronaut-openapi/issues/1676

For example, create HttpClient with `@Header` annotation like this:
```java
@Client("test")
public interface DefaultApi {

    /**
     * {@summary Check if a file is already available}
     *
     * @param xtESTCrc32 Decimal CRC32B representation of the file (required)
     * @param xtESTSha512 Lowercase SHA512hex representation of the file (required)
     * @param xtESTSize Size of the file in bytes (required)
     */
    @Head("/")
    void fileAvailable(
        @Header(name = "X-TEST-crc32") @NotNull String xtESTCrc32,
        @Header(name = "X-TEST-sha512") @NotNull String xtESTSha512,
        @Header(name = "X-TEST-size") @NotNull Long xtESTSize
    );
```
If you send request by this client you will see wrong header names:

изображение

Wron heade names: `xt-estcrc32`, `xt-estsha512`, `xt-estsize`. Must be `x-test-crc32`, `x-test-sha512`, `x-test-size`.

If you use value, instead name for Header annotaions, like this:
```java
@Client("test")
public interface DefaultApi {

    /**
     * {@summary Check if a file is already available}
     *
     * @param xtESTCrc32 Decimal CRC32B representation of the file (required)
     * @param xtESTSha512 Lowercase SHA512hex representation of the file (required)
     * @param xtESTSize Size of the file in bytes (required)
     */
    @Head("/")
    void fileAvailable(
        @Header("X-TEST-crc32") @NotNull String xtESTCrc32,
        @Header("X-TEST-sha512") @NotNull String xtESTSha512,
        @Header("X-TEST-size") @NotNull Long xtESTSize
    );
}
```
All works fine:
изображение

I think, the same value can be with all annotaions tha uses `@AliasFor`

### Version

4.5.4

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/11053


graemerocher (approved):

altro3:
@dstepanov Does this bug only with `@Header`? Is it posibble the same issue with `@PathVariable`?

dstepanov:
Actually I’m not sure that’s a bug, the Javadoc clearly says the name peopert is not supposed to be used inn this case

dstepanov:
That’s why the tests fail

altro3:
Sorry I do not understand. In my opinion, name is an auxiliary property, just for convenience and is an alias for value . Those. name is convenient to use to understand that you are specifying the name of a header or variable in the path. When you write value = - it's not so obvious.

So I don't understand why I can't write @Header(name = "my header") . Actually, I don’t understand why my values ​​​​change to some crooked strings

dstepanov:
It’s changed because the logic is to use the parameter name if the value is not specified. Please read the Javadoc, it explicitly says that the name is only used when applied on the class

altro3:
What can I say, in my opinion this is an extremely bad solution, because it is very confusing for the user. In the case of PathVariable, these two values ​​have the same meaning, but in the case of Header, it turns out that magic happens and is not obvious at all. By the way, it is not written anywhere that the value in the name field will be distorted

altro3:
Well, the decision to put different values ​​in these fields depending on the situation, in my opinion, is very ugly. This annotation is very popular. I can imagine the faces of newcomers to Micronaut when they stumble upon this error, I think they’ll just go back to the good old spring

dstepanov:
I will see what can be done

graemerocher (approved):

PR #11061, opened 2024-08-07
Fix resolution of `@Retryable(value=)` attribute

### LINKED ISSUE #11048: @Retryable value() is not interchangeable with includes()
### Expected Behavior

In the javadocs, the `@Retryable` annotation's `includes()` attribute is marked as an `@AliasFor` the value attribute, so I was expecting the two to be interchangeable.
https://docs.micronaut.io/latest/api/io/micronaut/retry/annotation/Retryable.html#includes()

There's at least one example in the [Retry Advice](https://docs.micronaut.io/latest/guide/#retry) documentation that uses the value attribute, so I don't think this expectation was unreasonable:
> `@Retryable(ServiceUnavailableException.class)`

### Actual Behaviour

The actual behaviour is that `value()` is completely ignored by `AnnotationRetryStateBuilder`:
When building the `DefaultRetryPredicate`, the code specifically only looks for `INCLUDES`.
https://github.com/micronaut-projects/micronaut-core/blob/11c55706d4c651ed51bd6ee952aaf06a14b92ba9/retry/src/main/java/io/micronaut/retry/intercept/AnnotationRetryStateBuilder.java#L84-L90

If I'm understanding the intent correctly, then either it should be the other way around where `value()` has `@AliasFor("includes")`, or `AnnotationRetryStateBuilder` should be looking for "value".

### Steps To Reproduce

A reproducing example test case might have, on the latest 4.6.x commit 9a67968f:
```diff
diff --git a/retry/src/test/groovy/io/micronaut/retry/intercept/SimpleRetrySpec.groovy b/retry/src/test/groovy/io/micronaut/retry/intercept/SimpleRetrySpec.groovy
index 8b33c8981c..d71e5e9b19 100644
--- a/retry/src/test/groovy/io/micronaut/retry/intercept/SimpleRetrySpec.groovy
+++ b/retry/src/test/groovy/io/micronaut/retry/intercept/SimpleRetrySpec.groovy
@@ -249,6 +249,29 @@ class SimpleRetrySpec extends Specification {
         context.stop()
     }

+    void "test retry with value"() {
+        given:
+        ApplicationContext context = ApplicationContext.run()
+        CounterService counterService = context.getBean(CounterService)
+
+        when:
+        counterService.getCountValue(true)
+
+        then: "retry didn't kick in because the exception thrown doesn't match includes"
+        thrown(IllegalStateException)
+        counterService.countIncludes == 1
+
+        when:
+        counterService.getCountValue(false)
+
+        then: "retry kicks in because the exception thrown matches includes"
+        noExceptionThrown()
+        counterService.countIncludes == counterService.countThreshold
+
+        cleanup:
+        context.stop()
+    }
+
     void "test retry with excludes"() {
         given:
         ApplicationContext context = ApplicationContext.run()
@@ -416,6 +439,19 @@ class SimpleRetrySpec extends Specification {
             return countIncludes
         }

+        @Retryable(attempts = '5', delay = '5ms', value = MyCustomException.class)
+        Integer getCountValue(boolean illegalState) {
+            countIncludes++
+            if(countIncludes < countThreshold) {
+                if (illegalState) {
+                    throw new IllegalStateException("Bad count")
+                } else {
+                    throw new MyCustomException()
+                }
+            }
+            return countIncludes
+        }
+
         @Retryable(attempts = '5', delay = '5ms', excludes = MyCustomException.class)
         Integer getCountExcludes(boolean illegalState) {
             countExcludes++
```
(Note that this is mostly just copying the existing test on `includes()`; I didn't really check thoroughly if this is completely appropriate.)

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.5.4, 4.6.x

Kushal-Dalasaniya:
Fixes #11048


yawkat:
@dstepanov Why does `@AliasFor` not work here? Is it because the code uses AnnotationValue directly, not AnnotationMetadata?

graemerocher:
I don't think this is the correct fix. The correct fix is to modify the `@Retryable` annotation and add to the `value` member:

```java
    @AliasFor(member = "includes")
    Class<? extends Throwable>[] value() default {};
```

Kushal-Dalasaniya:
Hi @graemerocher,

Thank you for the suggestion.

I've implemented the recommended changes and pushed the code. Could you please review it and let me know if any further adjustments are necessary? Thanks!

graemerocher:
change looks fine, let's see if the tests pass

yawkat (approved):

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #11065, opened 2024-08-09
Add API to access main context from bootstrap context

yawkat:
This is the least intrusive change I could come up with. The idea is that bootstrap beans can register singletons in the main context, and then normal beans can access those singletons to share data with the bootstrap.


inject/src/main/java/io/micronaut/context/BootstrapContextAccess.java
```diff
@@ -0,0 +1,36 @@
...
+import io.micronaut.core.annotation.Experimental;
+
+/**
+ * Interface for accessing aspects of the main context from the bootstrap context.
+ *
+ * @since 4.6.0
+ * @author Jonas Konrad
+ */
+@Experimental
+public interface BootstrapContextAccess {
```
graemerocher, line 30: also add `@Internal` for now

inject/src/main/java/io/micronaut/context/BootstrapContextAccess.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Get the {@link BeanDefinitionRegistry} of the main context. Note that when the bootstrap
+     * context is active, much of this registry will not be initialized yet, so most beans will be
+     * missing. However, you can register your own beans.
+     *
+     * @return The registry
+     */
+    BeanDefinitionRegistry getMainRegistry();
```
graemerocher, line 39: add nullability annotations

graemerocher (changes requested):

sdelamo (approved):

dstepanov (approved):

PR #11069, opened 2024-08-12
Correct order of body handlers

### LINKED ISSUE #11068: NotFoundSpec.test 404 handling with Flowable fails since #11002
### Issue description

Since PR https://github.com/micronaut-projects/micronaut-core/pull/11002 the following test is flaky.  

https://ge.micronaut.io/scans/tests?search.query=project:micronaut%20&search.timeZoneId=Europe%2FMadrid&tests.container=io.micronaut.http.client.aop.NotFoundSpec&tests.test=test%20404%20handling%20with%20Flowable

The following script passes with this commit https://github.com/micronaut-projects/micronaut-core/commit/bf860b1e34f1775042ea33ecec50e6aed977ba55

starts failing in the next commit https://github.com/micronaut-projects/micronaut-core/commit/f4a7f9dd38b1167f41999d26157c9f9546fefec4 which his the merge of this PR: https://github.com/micronaut-projects/micronaut-core/pull/11002

You can reproduce by running the following script: 

```
#!/bin/bash

EXIT_STATUS=0

for ((i=1; i<=10; i++))
do
  ./gradlew :http-client:test --tests "io.micronaut.http.client.aop.NotFoundSpec.test 404 handling with Flowable" --rerun-tasks || EXIT_STATUS=$?
  if [ $EXIT_STATUS -ne 0 ]; then
    exit $EXIT_STATUS
  fi
done

exit $EXIT_STATUS
```

This failure is causing every PR to fail.

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/11068

### DIFF
--- a/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
+++ b/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
@@ -139,7 +139,21 @@ public final class DefaultMessageBodyHandlerRegistry extends AbstractMessageBody
 
         @Override
         public <K extends BeanType<T>> Collection<K> filter(Class<T> beanType, Collection<K> candidates) {
-            List<K> all = new ArrayList<>(candidates);
+            List<K> all = new ArrayList<>(candidates.size());
+            for (K candidate : candidates) {
+                String[] applicableTypes = candidate.getAnnotationMetadata().stringValues(annotationType);
+                if (applicableTypes.length == 0) {
+                    all.add(candidate);
+                    continue;
+                }
+                for (String mt : applicableTypes) {
+                    if (mediaTypes.contains(new MediaType(mt))) {
+                        all.add(candidate);
+                        break;
+                    }
+                }
+            }
+            // Handlers with a media type defined should have a priority
             all.sort(Comparator.comparing(this::findOrder).reversed());
             return all;
         }
--- a/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
+++ b/http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
@@ -34,6 +34,7 @@ import jakarta.inject.Singleton;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -139,26 +140,39 @@ private MediaTypeQualifier(Argument<?> type,
...
+        private int findOrder(BeanType<?> beanType) {
+            int order = 0;
+            String[] applicableTypes = beanType.getAnnotationMetadata().stringValues(annotationType);
+            int size = mediaTypes.size();
+            for (String mt : applicableTypes) {
+                int index = mediaTypes.indexOf(new MediaType(mt));
+                if (index == -1) {
+                    continue;
+                }
+                int compareValue = size - index; // First value should have the priority
```
yawkat, line 170: since `size` is the same for every handler, i think you can replace size with 0, if you initialize `order` with `MIN_VALUE`. You can also probably get rid of the `-`

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
```diff
@@ -139,26 +140,39 @@ private MediaTypeQualifier(Argument<?> type,
...
-                        matchesMediaType.add(candidate);
+                        all.add(candidate);
+                        break;
                     }
                 }
             }
             // Handlers with a media type defined should have a priority
-            all.addAll(matchesMediaType);
-            all.addAll(matchesAll);
+            all.sort(Comparator.comparing(this::findOrder).reversed());
```
yawkat:
```suggestion
            all.sort(Comparator.comparingInt(this::findOrder).reversed());
```

http/src/main/java/io/micronaut/http/body/DefaultMessageBodyHandlerRegistry.java
dstepanov: I had -1 before and somehow it doesn't work

dstepanov: let's leave it this way

yawkat (approved):

sdelamo (approved):

PR #11070, opened 2024-08-12
Allow to handle all possible controller errors

dstepanov:
Including errors like incorrect HTTP method, content type etc. Right now it's only possible to catch this kind of error using a status route per status.

This is needed to align with JAX-RS where the exception handler can handle all the exceptions.


http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java
```diff
@@ -239,71 +237,14 @@ private ExecutionFlow<HttpResponse<?>> onErrorNoFilter(HttpRequest<?> request, T
         }
         final Throwable cause = t;
 
-        RouteMatch<?> errorRoute = routeExecutor.findErrorRoute(cause, declaringType, request);
+        RouteMatch<?> errorRoute = routeExecutor.findErrorRoute(cause, findDeclaringType(request), request);
         if (errorRoute != null) {
```
dstepanov, line 241: Just extracted two methods

yawkat (approved):

dstepanov:
Looks like one test is failign because not it triggers:
```
        @Error(exception = RuntimeException)
        HttpResponse onError() {
            HttpResponse.badRequest()
        }
```

dstepanov:
@yawkat @graemerocher @sdelamo Should there be a configuration option to enable it?

sdelamo (changes requested):
Can you please ship a default [built-in exception handler](https://docs.micronaut.io/4.5.3/guide/#builtInExceptionHandlers) for `HttpServerException` or `HttpException`?

sdelamo:
> Looks like one test is failign because not it triggers:
> 
> ```
>         @Error(exception = RuntimeException)
>         HttpResponse onError() {
>             HttpResponse.badRequest()
>         }
> ```

I think we could rewrite the test not to capture such a generic exception: https://github.com/micronaut-projects/micronaut-core/pull/11077

PR #11078, opened 2024-08-13
Add a ByteBody-based write API

yawkat:
This PR adds a NettyWriteContext.write(HttpResponse, ByteBody) API that, in the long term, should replace all the other write methods (except the headResponse one). The advantage is that we can get rid of reactive code for streaming, and merge some code paths.

These are the modifications:

- `StreamingNettyByteBody` exposes its `primary` method to allow the HTTP writers direct access to the data using the internal byte streaming API.
- `NettyBodyAdapter` is added to transform generic `ByteBody`s to `NettyByteBody`s.
- `AvailableByteArrayBody` and `InputStreamByteBody` are moved here from micronaut-servlet and made `@Experimental`.
- The `InputStream` based write method is removed from NettyWriteContext and replaced with `InputStreamByteBody`. This gets rid of a bunch of code.


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/body/NettyBodyAdapterSpec.groovy
```diff
@@ -0,0 +1,7 @@
+package io.micronaut.http.server.netty.body
+
+import spock.lang.Specification
+
+class NettyBodyAdapterSpec extends Specification {
```
sdelamo: delete this empty test

sdelamo (changes requested):

http/src/main/java/io/micronaut/http/body/stream/AvailableByteArrayBody.java
```diff
@@ -0,0 +1,100 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.body.stream;
+
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.NonNull;
```
sdelamo, line 19:
```suggestion
import io.micronaut.core.annotation.Internal;
import io.micronaut.core.annotation.NonNull;
```

http/src/main/java/io/micronaut/http/body/stream/AvailableByteArrayBody.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+
+/**
+ * {@link io.micronaut.http.body.AvailableByteBody} implementation based on a simple byte array.
+ *
+ * @author Jonas Konrad
+ * @since 4.6.0
+ */
+@Experimental
```
sdelamo, line 36:
```suggestion
@Internal
@Experimental
```

http/src/test/groovy/io/micronaut/http/body/stream/StreamPairSpec.groovy
```diff
@@ -0,0 +1,149 @@
...
+            return new InputStream() {
+                int i = 0
+
+                @Override
+                int read() throws IOException {
+
+                }
+            }
+        }
+    }
```
sdelamo:
```suggestion
```

http/src/main/java/io/micronaut/http/body/stream/AvailableByteArrayBody.java
sdelamo: Why did you not annotate with `@Internal`? is it because you are going to delete these classes in servlet and use this class form core in Servlet?

sdelamo (approved):

sdelamo:
@yawkat can you check the sonar issues?

http/src/main/java/io/micronaut/http/body/stream/AvailableByteArrayBody.java
yawkat: correct

yawkat: I think it's fine to have this class as API now. ByteBody is already public, and the only new thing exposed here is the static factory methods

dstepanov (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/body/NettyBodyAdapter.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.function.LongUnaryOperator;
+
+/**
+ * Adapter from generic streaming {@link ByteBody} to {@link StreamingNettyByteBody}.
+ *
+ * @author Jonas Konrad
+ * @since 4.6.0
+ */
+@Internal
+public final class NettyBodyAdapter implements BufferConsumer.Upstream, Subscriber<ByteBuf> {
```
dstepanov, line 41: Can some of those classes be package private?

sdelamo: @yawkat 👆🏻

PR #11087, opened 2024-08-14
Issue with Requires condition with missingClasses as string values

radovanradic:


inject-java/src/test/groovy/io/micronaut/inject/requires/RequiresSpec.groovy
```diff
@@ -384,6 +384,40 @@ class MyBean {
         context.close()
     }
 
+    void "test requires missing classes as string class name when class present"() {
+        when:
+        BeanDefinition beanDefinition = buildBeanDefinition('test.MyBean', '''
+package test;
+
+import io.micronaut.context.annotation.*;
+
```
radovanradic, line 393: Seems like there is an issue in micronaut-core 4.6.x when declaring something like this 
`@Requires(missingClasses = "org.hibernate.reactive.provider.ReactiveServiceRegistryBuilder")` because this code in `MatchesConditionUtils` 
```
if (requirement.contains(RequiresCondition.MEMBER_MISSING_CLASSES)) {
            AnnotationClassValue<?>[] classes = requirement.annotationClassValues(RequiresCondition.MEMBER_MISSING_CLASSES);
            if (classes.length > 0) {
                preConditions.add(new MatchesAbsenceOfClassesCondition(classes));
            }
        }
```
expects to get classes in annotation value
```
public AnnotationClassValue<?>[] annotationClassValues(@NonNull String member) {
        if (StringUtils.isEmpty(member)) {
            return AnnotationClassValue.ZERO_ANNOTATION_CLASS_VALUES;
        }
        Object o = values.get(member);
        if (o instanceof AnnotationClassValue<?> annotationClassValue) {
            return new AnnotationClassValue[]{annotationClassValue};
        }
        if (o instanceof AnnotationClassValue<?>[] annotationClassValues) {
            return annotationClassValues;
        }
        return AnnotationClassValue.ZERO_ANNOTATION_CLASS_VALUES;
    }
```
in this case, instanceof returns `AnnotationClassValue.ZERO_ANNOTATION_CLASS_VALUES` since value is string and then fails to match condition, actually condition will not be evaluated.

dstepanov:
The "problem" is actually with:
```
if (requirement.contains(RequiresCondition.MEMBER_MISSING_CLASSES)) {
            AnnotationClassValue<?>[] classes = requirement.annotationClassValues(RequiresCondition.MEMBER_MISSING_CLASSES);
            if (classes.length > 0) {
                preConditions.add(new MatchesAbsenceOfClassesCondition(classes));
            }
        }
```
Previous version would use `stringValues` there.

But I think your fix is also correct, you might want to fix and test the singular method `annotationClassValue` below. Please extract a private method  to create `AnnotationClassValue` from a class name.

radovanradic:
> The "problem" is actually with:
> 
> ```
> if (requirement.contains(RequiresCondition.MEMBER_MISSING_CLASSES)) {
>             AnnotationClassValue<?>[] classes = requirement.annotationClassValues(RequiresCondition.MEMBER_MISSING_CLASSES);
>             if (classes.length > 0) {
>                 preConditions.add(new MatchesAbsenceOfClassesCondition(classes));
>             }
>         }
> ```
> 
> Previous version would use `stringValues` there.
> 
> But I think your fix is also correct, you might want to fix and test the singular method `annotationClassValue` below. Please extract a private method to create `AnnotationClassValue` from a class name.

Added util method, will improve coverage as well.
Singular method to create `AnnotationClassValue` is used for `@Requires(bean = Class)` or `@Requires(condition = Class)` and can't get class names. Only `@Requires` field possibly taking strings is `classNames`: 
```
@AliasFor(member = "missing")
String[] missingClasses() default {};
```

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -1703,4 +1709,18 @@ private static <T extends Enum> Optional<T> convertToEnum(Class<T> enumType, Obj
...
     }
+
+    /**
+     * Creates {@link AnnotationClassValue} from the class name.
+     *
+     * @param className The class name
+     * @return AnnotationClassValue for given class name
+     */
+    private static AnnotationClassValue<?> getAnnotationClassValue(String className) {
+        Optional<Class<?>> theClass = ClassUtils.forName(className, null);
```
radovanradic, line 1726: The difference from previous version is that it was using classloader from `beanContext.getClassLoader()` and this will fallback to current thread classloader.
```
if (classLoader == null) {
                classLoader = Thread.currentThread().getContextClassLoader();
            }
            if (classLoader == null) {
                classLoader = ClassLoader.getSystemClassLoader();
            }
```
Not sure if this could cause other issues.

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -707,6 +707,12 @@ public AnnotationClassValue<?>[] annotationClassValues(@NonNull String member) {
         if (o instanceof AnnotationClassValue<?>[] annotationClassValues) {
             return annotationClassValues;
         }
+        if (o instanceof String className) {
+            return new AnnotationClassValue[] { getAnnotationClassValue(className) };
+        }
+        if (o instanceof String[] classNames) {
+            return Arrays.stream(classNames).map(AnnotationValue::getAnnotationClassValue).toArray(AnnotationClassValue[]::new);
+        }
```
dstepanov, line 715: Please also add the same logic to the method below `public Optional<AnnotationClassValue<?>> annotationClassValue(@NonNull String member) {`

radovanradic: Singular method to create `AnnotationClassValue` is used for `@Requires(bean = Class)` or `@Requires(condition = Class)` and can't get class names. Only `@Requires` field possibly taking strings is `classNames`: 
```
@AliasFor(member = "missing")
String[] missingClasses() default {};
```

I can add the same change but can't test it because only case when `@Requires` takes in string(s) is `missingClasses` and it is covered above.

dstepanov: Please add it and add a test to `AnnotationValueSpec`

radovanradic: Done in https://github.com/micronaut-projects/micronaut-core/pull/11087/commits/b771aca05739eb6255fa84f6320dc80780681751

gradle.properties
```diff
@@ -1,4 +1,4 @@
-projectVersion=4.6.1-SNAPSHOT
+projectVersion=4.6.0
```
dstepanov: Remove this

dstepanov (approved):

gradle.properties
radovanradic: Done

sdelamo (approved):

rsalbergaria:
@sdelamo and @radovanradic, could the issue reported be related to the problem described in [micronaut-projects/micronaut-platform/discussions/1704](https://github.com/micronaut-projects/micronaut-platform/discussions/1704)?

PR #11094, opened 2024-08-16
Introduce ByteBody-based server response type

yawkat:
- [x] Will retarget to 4.7.x when that branch is ready

This PR adds and uses a new `ServerHttpResponse` type that contains the encoded bytes of a server response in form of a `ByteBody`. We need a new type here instead of just `HttpResponse<ByteBody>` because we need to ensure the body is closed properly.

`MessageBodyWriter` gets a new `write` method that returns the `ServerHttpResponse`. It has an expanded argument list from the other writeTo methods (additional HttpRequest and full HttpResponse access) so that it can be used to implement all features that NettyBodyWriter supported. NettyBodyWriter is removed.

`NettyWriteContext` has all of its write methods (except the HEAD response one) replaced with a single `ByteBody`-based method.

RoutingInboundHandler is refactored to return `ExecutionFlow<ServerHttpResponse>` instead of passing around the `NettyWriteContext`. This removes some duplication and helps despaghettify it a little bit.

There are two long-term goals for this API:

- Move more code from RoutingInboundHandler into RequestLifecycle to share with servlet.
- Allow filters to run after encoding so that they can access response bytes.


yawkat:
Converted to draft because I absolutely must not forget to retarget to 4.7.x :laughing: otherwise it's ready

http/src/main/java/io/micronaut/http/body/MessageBodyWriter.java
```diff
@@ -157,6 +163,31 @@ void writeTo(
...
+     * @param request       The request
+     * @param httpResponse  The response
+     * @param type          The response body type
+     * @param mediaType     The media type
+     * @param object        The object to write
+     * @return A {@link ServerHttpResponse} with the response bytes
+     * @throws CodecException If an error occurs encoding
+     */
+    @NonNull
+    default ServerHttpResponse<?> write(
```
dstepanov: This method doesn't look good to me, we have a mutable response and we return one more response. Can you just return ByteBody?

dstepanov (changes requested):

http/src/main/java/io/micronaut/http/body/MessageBodyWriter.java
dstepanov: Considering this method is only used in `RoutingInBoundHandler` you can simply convert the result into some internal record ResponseWithByteBody

yawkat: it's essentially the same pattern as before. NettyBodyWriter implementations sometimes wrote responses that are completely different from the input response too, which made this necessary. Filters can also do this.

it can't be an internal record ResponseWithByteBody because this method is public api. I want to expand it to be used in servlet, and I want filters to be able to access the encoded response as well eventually.

yawkat: Eventually I want to make the whole writer code a filter. Not sure that's possible given performance considerations, but it would be nice design-wise. The writer would simply transform the response to this "encoded response" type and continue passing it down the filter chain, where other filters can be placed before or after the encoder.

dstepanov: I don't like the idea of `ServerHttpResponse` being introduced in the API shared between client/server. Why can we just set the bytebody into the ordinary `HttpResponse` and return it.

yawkat: ServerHttpResponse is necessary because of lifecycle management. ByteBody _must_ be closed so we can't just discard instances like we do with normal HttpResponses. This is not an issue on the request side because we just close NettyHttpRequest when the response is written.

I'm not yet sure how to deal with this on the client. The same mechanism would work on the HttpRequest side there, kind of a ClientHttpRequest type. This would need a new MessageBodyWriter method too. However for the response side it's difficult.

graemerocher: I agree that this method doesn't belong in this interface. Should be a utility method or something `MessageBodyWriter` is not related to the server

yawkat: It can't be a utility method. The implementations need full access to request and response. In fact they already make use of it before this change, see StreamFileBodyWriter which returns a notModified response in some cases.

graemerocher: in that case it should be an internal interface only implemented only by message body writers that support it

yawkat: the same pattern is useful for servlet and the function modules, so it can't be internal. StreamFileBodyWriter for example is now mostly independent of netty and could be shared with servlet.

Maybe a non-internal subinterface would work. This would also allow us to remove access to HttpHeaders from MessageBodyWriter/Reader. That is a thorn in my side anyway since it prevents using MessageBodyReader for form fields. The few implementations that really need headers could then use the new request-/response-based subinterfaces with more power.

WDYT?

graemerocher: A non-internal sub interface is a better arrangement yes

yawkat: PTAL

I'll look at replacing the MessageBodyWriter methods with non-header ones in another PR.

dstepanov: Why can't we check if the body is of type ByteBody and discard it after the response is closed?

yawkat: Because normal responses are not always closed. There's no concept of lifecycle management for responses at all right now

dstepanov: But at least we can close the body no?

yawkat: The problem is that we don't. We have a lot of code that assumes the HttpResponse and its body can just be discarded. If we start putting the ByteBody into the body, it adds risk of the body being transferred into code that isnt prepared for it and that won't close the body properly.

graemerocher (approved):

http/src/main/java/io/micronaut/http/body/MessageBodyWriter.java
dstepanov: IMHO the problem still remains, byteBody can be forgotten if the response is replaced.

Also I think everything server based like `io.micronaut.http.ServerHttpResponse` should not be in http package/module.

dstepanov: I'm not a fan of this small API fixes, we clearly miss some better body API

yawkat: ServerHttpResponse at least gives us a type that is statically Closeable so that IDE can warn when it's created without being closed.

dstepanov: Maybe we can introduce request / response lifecycle hooks to cleanup the body. That can be used to implement the request scope cleanup now implemented using the bean context contributing to worse perf.

yawkat: The request kind of does have lifecycle hooks, in the form of RequestHandler.responseWritten. But this is not viable for responses because anyone could create a response. The request is always created inside the netty handler, but the response is not.

dstepanov (approved):

Issue #7934, opened 2022-09-02
CorsFilter ignores http methods available via filters


kevin-wise:
### Expected Behavior

The CORS filter should allow pre-flight requests for any HTTP method that is handled by a filter.
Alternatively the filter should be able to specify which methods are available for the handled paths.

### Actual Behaviour

The CORS filter only considers paths handled by routes, not paths handled by filters. It gets to the point in CorsFilter.handleRequest where it is considering which HTTP methods are available, and it finds none. The CORS pre-flight is not processed by CorsFilter, resulting in a 401 Forbidden.

An example use case for such a filter is one which uses a ProxyHttpClient to forward all requests under a specific path to another server.

This was working in Micronaut 3.3. Seems to have been broken by https://github.com/micronaut-projects/micronaut-core/pull/7367.  

### Steps To Reproduce

1. Create filter which matches a path that is not handled by any controller.
2. Issue a valid CORS pre-flight request against a path matching the filter.
3. The response will be 401 when it should be 200.

For example, a filter like this:
```
@Filter("/proxy/**")
public class ProxyFilter implements HttpServerFilter {

  private final ProxyHttpClient client;

  @Property(name = "micronaut.http.services.upstream.url")
  private String upstreamUrl;

  public ProxyFilter(ProxyHttpClient client) {
    this.client = client;
  }

  @Override
  public Publisher<MutableHttpResponse<?>> doFilter(
      HttpRequest<?> request, ServerFilterChain chain) {
    if (!isPreflightRequest(request)) {
      return Publishers.map(client.proxy(mutateRequest(request)), response -> response);
    } else {
      return chain.proceed(request);
    }
  }

  @VisibleForTesting
  @SneakyThrows
  MutableHttpRequest<?> mutateRequest(HttpRequest<?> request) {
    URI uri = new URI(upstreamUrl);
    MutableHttpRequest<?> mutableHttpRequest =
        request.mutate().uri(b ->
            b.scheme(uri.getScheme()).host(uri.getHost()).port(uri.getPort()));
    mutableHttpRequest.getHeaders().set(HttpHeaders.HOST, uri.getHost());
    return mutableHttpRequest;
  }

  boolean isPreflightRequest(HttpRequest<?> request) {
    io.micronaut.http.HttpHeaders headers = request.getHeaders();
    Optional<String> origin = headers.getOrigin();
    return origin.isPresent()
        && headers.contains(ACCESS_CONTROL_REQUEST_METHOD)
[...]

PR #11097, opened 2024-08-19
Use @PreMatching for CORS Preflight request

dstepanov:


http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -430,39 +451,37 @@ private void decorateResponseWithHeaders(@NonNull HttpRequest<?> request,
...
 
-        Optional<? extends ArrayList<HttpMethod>> availableHttpMethods = (Optional<? extends ArrayList<HttpMethod>>) request.getAttribute(AVAILABLE_HTTP_METHODS, new ArrayList<HttpMethod>().getClass());
-        if (CorsUtil.isPreflightRequest(request) &&
-            availableHttpMethods.isPresent() &&
-            availableHttpMethods.get().stream().anyMatch(method -> method.equals(methodToMatch))) {
-            if (!hasAllowedHeaders(request, config)) {
-                return Optional.of(HttpStatus.FORBIDDEN);
+        if (hasAllowedHeaders(request, config)) {
+            List<UriRouteMatch<Object, Object>> anyUriRoutes = router.findAny(request);
+            if (anyUriRoutes.isEmpty()) {
```
dstepanov: This is how it worked before, but I think that's not correct. `methodToMatchOptional` is OPTIONS for preflight check and there is no such method in mocked tests

graemerocher:
seems to be failing?

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
yawkat: methodToMatchOptional can be not OPTIONS if there's a Access-Control-Request-Method header

yawkat (approved):

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
dstepanov: Shouldt'n we fail OPTIONS call if there is no route present under that path?

yawkat: No, in that case the preflight should return a good response and the actual cors request should return the 404. This will allow the cors client to see that the route is missing. Otherwise it will just get a permission error

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
```diff
@@ -456,85 +306,34 @@ class CorsFilterSpec extends Specification {
...
 
         then:
-        result.isPresent()
-
-        when:
-        MutableHttpResponse<?> response = result.get()
-
-        then:
-        HttpStatus.OK == response.status()
+        HttpStatus.FORBIDDEN == response.status()
```
dstepanov, line 329: I had to rewite the mocked test. This is not passing anymore, the mock didn't include the proper origin mapping so the request filter was skipped. Please check how it supposted to work @yawkat @sdelamo

http-client-jdk/src/test/groovy/io/micronaut/http/client/jdk/OptionsRequestAttributesSpec.groovy
```diff
@@ -37,6 +37,12 @@ class OptionsRequestAttributesSpec extends Specification {
         HttpClientResponseException e = thrown()
         e.response.status == HttpStatus.METHOD_NOT_ALLOWED
 
+        and:
+        MyFilter myFilter = ctx.getBean(MyFilter)
+        !myFilter.containsRouteInfo
```
sdelamo, line 45: @dstepanov I had to this change as well.

sdelamo (approved):

sdelamo:
@dstepanov I have rewritten the test and the filter slightly. I think it is now correct. @yawkat can you check it out again.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/OptionsRequestAttributesSpec.groovy
```diff
@@ -37,6 +37,12 @@ class OptionsRequestAttributesSpec extends Specification {
         HttpClientResponseException e = thrown()
         e.response.status == HttpStatus.METHOD_NOT_ALLOWED
 
+        and:
+        MyFilter myFilter = ctx.getBean(MyFilter)
+        !myFilter.containsRouteInfo
```
yawkat, line 45: maybe also check that the filter was called at all

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -353,6 +384,26 @@ private Optional<CorsOriginConfiguration> getConfiguration(@NonNull HttpRequest<
             .findFirst();
     }
 
+    @NonNull
+    private Optional<CorsOriginConfiguration> getAnyConfiguration(@NonNull HttpRequest<?> request) {
+        String requestOrigin = request.getOrigin().orElse(null);
+        if (requestOrigin == null) {
+            return Optional.empty();
+        }
+        for (UriRouteMatch<Object, Object> routeMatch : router.findAny(request)) {
```
yawkat, line 393: so, if we have to run the router in order to respond to the cors request, why are we changing the filter to be pre-routing at all?

dstepanov: Otherwise we need to do a hack and store all the routes because it will not match

yawkat (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/OptionsRequestAttributesSpec.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11097/commits/70d91bf05e49896caa6ab0735d430334f7afe051

dstepanov:
LGTM

ThomasIommi:
Hello!

As mentioned in issue #7934, it seems that the router might not recognize paths defined using patterns like those used by `@Filter`s. 
I'm afraid that checking for known paths reintroduce the same issue, causing all requests to a proxy filter to be blocked by the CORS filter with a 403 Forbidden error.

This situation seems to prevent us from implementing solutions suggested in the official documentation, such as the one found here: https://docs.micronaut.io/latest/guide/#proxyClient

Here's a snippet of the code in question:

```java
@Nullable
private boolean validatePreflightRequest(@NonNull HttpRequest<?> request,
                                         @NonNull CorsOriginConfiguration config) {
    ...
    List<HttpMethod> availableHttpMethods = router.findAny(request).stream().map(UriRouteMatch::getHttpMethod).toList();
    if (availableHttpMethods.stream().noneMatch(method -> method.equals(methodToMatch))) {
        return false;
    }    
    ...
    return true;
}
```

PR #11109, opened 2024-08-23
Replace the DefaultHttpClient constructors with a builder

yawkat:
Unfortunately we have a bit of a mess in micronaut-oracle-cloud with dependencies on DefaultHttpClient. oke identity depended on internal classes that broke in 4.6, requiring a change downstream.

This patch replaces the DefaultHttpClient constructors with a builder. The defaults are now unified between constructors, and we can limit exposure to only those settings that micronaut-oracle-cloud actually needs.

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -539,7 +539,9 @@ public class DefaultHttpClient implements
 
     /**
      * @return The {@link MediaTypeCodecRegistry} used by this client
+     * @deprecated Use body handlers instead
      */
+    @Deprecated
     public MediaTypeCodecRegistry getMediaTypeCodecRegistry() {
         return mediaTypeCodecRegistry;
     }
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -477,6 +477,12 @@ public class DefaultHttpClient implements
         );
     }
 
+    /**
+     * Create a new builder for a {@link DefaultHttpClient}.
+     *
+     * @return The builder
+     * @since 4.7.0
+     */
     @NonNull
     public static DefaultHttpClientBuilder builder() {
         return new DefaultHttpClientBuilder();
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -33,7 +33,6 @@ import io.micronaut.core.io.buffer.ReferenceCounted;
 import io.micronaut.core.order.Ordered;
 import io.micronaut.core.propagation.PropagatedContext;
 import io.micronaut.core.type.Argument;
-import io.micronaut.core.util.ArgumentUtils;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.ObjectUtils;
 import io.micronaut.core.util.StringUtils;
@@ -69,7 +68,6 @@ import io.micronaut.http.client.exceptions.NoHostException;
 import io.micronaut.http.client.exceptions.ReadTimeoutException;
 import io.micronaut.http.client.exceptions.ResponseClosedException;
 import io.micronaut.http.client.filter.ClientFilterResolutionContext;
-import io.micronaut.http.client.filter.DefaultHttpClientFilterResolver;
 import io.micronaut.http.client.filters.ClientServerContextFilter;
 import io.micronaut.http.client.multipart.MultipartBody;
 import io.micronaut.http.client.multipart.MultipartDataFactory;
@@ -131,8 +129,6 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.channel.MultithreadEventLoopGroup;
 import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.SocketChannel;
-import io.netty.channel.socket.nio.NioDatagramChannel;
-import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.handler.codec.TooLongFrameException;
 import io.netty.handler.codec.http.DefaultFullHttpRequest;
 import io.netty.handler.codec.http.DefaultFullHttpResponse;
@@ -190,7 +186,6 @@ import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -428,29 +451,35 @@ public DefaultHttpClient(@Nullable URI uri, @NonNull HttpClientConfiguration con
...
-            ConversionService.SHARED);
+        this(
+            builder()
+                .loadBalancer(loadBalancer)
+                .configuration(configuration)
+        );
+    }
+
+    @NonNull
+    public static DefaultHttpClientBuilder builder() {
```
graemerocher, line 487: add javadoc and `@since`

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+ * <p>
+ * If you need to make a method of this builder public, please document the module that uses it.
+ *
+ * @author Jonas Konrad
+ * @since 4.7.0
+ */
+@Internal
+public final class DefaultHttpClientBuilder {
+    @Nullable
+    LoadBalancer loadBalancer = null;
```
graemerocher, line 62: can all these fields be made `private`?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+
+    @NonNull
+    DefaultHttpClientBuilder conversionService(@NonNull ConversionService conversionService) {
+        ArgumentUtils.requireNonNull("conversionService", conversionService);
+        this.conversionService = conversionService;
+        return this;
+    }
+
+    @NonNull
+    DefaultHttpClientBuilder resolverGroup(@Nullable AddressResolverGroup<?> resolverGroup) {
```
graemerocher, line 262: none of these builder methods are `public`, how will it be used from the oracle cloud module?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+ * that use it directly, in particular micronaut-oracle-cloud. This builder acts as API for those
+ * users.
+ * <p>
+ * If you need to make a method of this builder public, please document the module that uses it.
+ *
+ * @author Jonas Konrad
+ * @since 4.7.0
+ */
+@Internal
+public final class DefaultHttpClientBuilder {
```
graemerocher, line 60: complete the javadoc for public methods

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
yawkat: No they are all accessed from DefaultHttpClient

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
yawkat: Only the public ones are needed. That's the point, I want to make sure the visibility modifier shows which ones we can break and which ones we cant

dstepanov: I'm also confused by the non-public methods. Before you can set `conversionService` in the constructor but now it's not possible

http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java
```diff
@@ -425,49 +421,38 @@ public <T> Optional<MessageBodyWriter<T>> findWriter(Argument<T> type, List<Medi
...
             String clientId,
-            String contextPath,
             BeanContext beanContext,
             AnnotationMetadata annotationMetadata) {
 
-        EventLoopGroup eventLoopGroup = resolveEventLoopGroup(configuration, beanContext);
-        ConversionService conversionService = beanContext.getBean(ConversionService.class);
         String addressResolverGroupName = configuration.getAddressResolverGroupName();
-        AddressResolverGroup<?> resolverGroup = addressResolverGroupName == null ? null : beanContext.getBean(AddressResolverGroup.class, Qualifiers.byName(addressResolverGroupName));
-        return new DefaultHttpClient(
```
dstepanov, line 445: Can we create an instance of the builder with scope prototype using the bean context? Maybe a factory or something.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
yawkat: Yes I removed options that weren't used. If you use the constructor, now only ConversionService.SHARED is possible, except for the managed DefaultHttpClient from the context.

If there's code that needs to set the ConversionService but cant use the managed DefaultHttpClient, we can make the setter public. The idea here is that we know which properties are set by outside projects and which ones are only used with default settings.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+     */
+    @NonNull
+    public DefaultHttpClientBuilder nettyClientSslBuilder(@NonNull ClientSslBuilder nettyClientSslBuilder) {
+        ArgumentUtils.requireNonNull("nettyClientSslBuilder", nettyClientSslBuilder);
+        this.nettyClientSslBuilder = nettyClientSslBuilder;
+        return this;
+    }
+
+    @NonNull
+    DefaultHttpClientBuilder codecRegistry(@NonNull MediaTypeCodecRegistry codecRegistry) {
```
dstepanov, line 194: Please already deprecate

dstepanov (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultNettyHttpClientRegistry.java
yawkat: The problem is that from the bean context, clients always have to go to the registry for (a) service id configuration (b) service id pooling. It is not actually possible to use the Builder to create a managed (in-context) client, you always have to get a ready-built client from the registry

dstepanov (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClientBuilder.java
yawkat: Done. In fact this is part of the motivation for the builder. [In the oracle cloud module](https://github.com/micronaut-projects/micronaut-oracle-cloud/blob/8c60ad850afd16e94ac71585332f2507794811d3/oraclecloud-oke-workload-identity/src/main/java/io/micronaut/oraclecloud/oke/workload/identity/MicronautOkeWorkloadIdentityResourcePrincipalsFederationClient.java#L163), there is code that has to construct this registry because the constructor it uses requires it, even though the defaults would be fine. With this change, that code will only have to set the configuration and ssl builder, leaving the other builder properties as default, and we can change or remove them in core without breaking things.

graemerocher (approved):

PR #11117, opened 2024-08-26
Implement pcap logging for netty HTTP client

yawkat:
The server has supported this for a long time. This PR adds the same feature for client logging. The code is very similar, and the test is basically copied.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -16,6 +16,7 @@
 package io.micronaut.http.client;
 
 import io.micronaut.context.env.CachedEnvironment;
+import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.NextMajorVersion;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


graemerocher (approved):

sdelamo (changes requested):
Please, mention in https://docs.micronaut.io/4.6.5/guide/#jdkHttpClient that JDK Client implementation does it does not support pcap logging

yawkat:
@dstepanov PTAL

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -872,6 +875,28 @@ public HttpClientConfiguration.Http2ClientConfiguration getHttp2Configuration()
...
     }
 
+    /**
+     * The path pattern to use for logging outgoing connections to pcap. This is an unsupported option: Behavior may
+     * change, or it may disappear entirely, without notice! Only implemented for netty.
+     *
+     * @return The path pattern, or {@code null} if logging is disabled.
+     */
+    @Internal
+    public String getPcapLoggingPathPattern() {
```
dstepanov, line 885: What's the point of having the accessors internal? People will not see the property configuration as an internal

dstepanov (approved):

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
yawkat: the comment is for config users, the annotation is for programmatic users.

PR #11121, opened 2024-08-27
Deprecate `MediaTypeCodec` and use Message Body Writer / Reader API for clients

dstepanov:
I tested it by disabling all codecs, and the tests passed.

Right now, we still don't use the message body API for functions. We should convert other modules: servlet, Oracle Cloud, Azure, and GCP.

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
+++ b/core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
@@ -42,7 +42,7 @@ public class ByteArrayByteBuffer implements ByteBuffer<byte[]> {
      *
      * @param underlyingBytes the bytes to wrap
      */
-    public ByteArrayByteBuffer(byte[] underlyingBytes) {
+    ByteArrayByteBuffer(byte[] underlyingBytes) {
         this(underlyingBytes, underlyingBytes.length);
     }
 
@@ -54,7 +54,7 @@ public class ByteArrayByteBuffer implements ByteBuffer<byte[]> {
      * @param underlyingBytes the bytes to wrap
      * @param capacity        the capacity of the buffer
      */
-    public ByteArrayByteBuffer(byte[] underlyingBytes, int capacity) {
+    ByteArrayByteBuffer(byte[] underlyingBytes, int capacity) {
         if (capacity < underlyingBytes.length) {
             this.underlyingBytes = Arrays.copyOf(underlyingBytes, capacity);
         } else if (capacity > underlyingBytes.length) {
@@ -97,7 +97,7 @@ public class ByteArrayByteBuffer implements ByteBuffer<byte[]> {
 
     @Override
     public ByteArrayByteBuffer readerIndex(int readPosition) {
-        this.readerIndex = Math.min(readPosition, underlyingBytes.length - 1);
+        this.readerIndex = Math.min(readPosition, underlyingBytes.length);
         return this;
     }
 
@@ -108,7 +108,7 @@ public class ByteArrayByteBuffer implements ByteBuffer<byte[]> {
 
     @Override
     public ByteArrayByteBuffer writerIndex(int position) {
-        this.writerIndex = Math.min(position, underlyingBytes.length - 1);
+        this.writerIndex = Math.min(position, underlyingBytes.length);
         return this;
     }
 
--- a/core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
+++ b/core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
@@ -1,0 +1,236 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.io.buffer;
+
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.Internal;
+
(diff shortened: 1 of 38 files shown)


core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * Construct a new {@link ByteArrayByteBuffer} for the given bytes and capacity.
+     * If capacity is greater than the length of the underlyingBytes, extra bytes will be zeroed out.
+     * If capacity is less than the length of the underlyingBytes, the underlyingBytes will be truncated.
+     *
+     * @param underlyingBytes the bytes to wrap
+     * @param capacity        the capacity of the buffer
+     */
+    public ByteArrayByteBuffer(byte[] underlyingBytes, int capacity) {
```
yawkat: Make it private. People that need this can just call `new ByteArrayByteBuffer(arr).capacity(capacity)` and it will do the same

core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Override
+    public int readerIndex() {
+        return readerIndex;
+    }
+
+    @Override
+    public ByteArrayByteBuffer readerIndex(int readPosition) {
+        this.readerIndex = Math.min(readPosition, underlyingBytes.length - 1);
```
yawkat:
```suggestion
        this.readerIndex = Math.min(readPosition, underlyingBytes.length);
```

core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Override
+    public int writerIndex() {
+        return writerIndex;
+    }
+
+    @Override
+    public ByteArrayByteBuffer writerIndex(int position) {
+        this.writerIndex = Math.min(position, underlyingBytes.length - 1);
```
yawkat:
```suggestion
        this.writerIndex = Math.min(position, underlyingBytes.length);
```

core/src/main/java/io/micronaut/core/io/buffer/ByteArrayByteBuffer.java
dstepanov: Are you sure? I just copied it from Servlet

yawkat: Yes, otherwise you can never set readerIndex so that readableBytes == 0

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/HttpResponseAdapter.java
```diff
@@ -95,53 +102,69 @@ public MutableConvertibleValues<Object> getAttributes() {
...
+            }
+        }
+        if (messageBodyHandlerRegistry != null) {
+            MessageBodyReader<T> reader = messageBodyHandlerRegistry.findReader(type, contentType).orElse(null);
+            if (reader != null) {
+                try {
+                    T value = reader.read(
+                        type,
+                        contentType,
+                        new SimpleHttpHeaders(),
```
yawkat: Use proper headers

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -347,19 +352,28 @@ protected void handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame ms
...
                     } catch (CodecException e) {
                         messageProcessingException(ctx, e);
                         return;
                     }
+                    if (data == null) {
+                        MessageBodyReader<?> reader = messageBodyHandlerRegistry.findReader(bodyArgument, mediaType)
+                            .orElse(null);
+                        if (reader != null) {
+                            ByteBuffer<ByteBuf> byteBuffer = new NettyByteBufferFactory(ctx.alloc()).wrap(msg.content().retain());
+                            data = reader.read((Argument) bodyArgument, mediaType, originatingRequest.getHeaders(), byteBuffer);
```
yawkat: does not make sense to use request headers here

http-netty/src/main/java/io/micronaut/http/netty/websocket/WebSocketMessageEncoder.java
```diff
@@ -69,11 +91,26 @@ public WebSocketFrame encodeMessage(Object message, MediaType mediaType) {
...
+            if (messageBodyHandlerRegistry != null) {
+                Argument<Object> argument = Argument.ofInstance(message);
+                MessageBodyWriter<Object> messageBodyWriter = messageBodyHandlerRegistry.findWriter(argument, theMediaType).orElse(null);
+                if (messageBodyWriter != null) {
+                    io.micronaut.core.io.buffer.ByteBuffer<?> encoded = messageBodyWriter.writeTo(
+                        argument,
+                        theMediaType,
+                        message,
+                        new SimpleHttpHeaders(),
+                        new NettyByteBufferFactory(UnpooledByteBufAllocator.DEFAULT)
```
yawkat: use NettyByteBufferFactory.DEFAULT

http/src/main/java/io/micronaut/http/HttpRequest.java
```diff
@@ -445,4 +445,24 @@ static <T> MutableHttpRequest<T> create(HttpMethod httpMethod, String uri, Strin
...
         return HttpRequestFactory.INSTANCE.create(httpMethod, uri, httpMethodName);
     }
+
+    /**
+     * Returns a mutable request based on this request.
+     * @return the mutable request
+     * @since 4.7
+     */
+    default MutableHttpRequest<B> toMutableRequest() {
+        if (this instanceof MutableHttpRequest<B> mutableHttpRequest) {
```
yawkat, line 455: just override the method in MutableHttpRequest instead of using this instanceof

yawkat (approved):

http/src/main/java/io/micronaut/http/HttpRequest.java
dstepanov: I did also override it

graemerocher (approved):

PR #11123, opened 2024-08-27
Fix issue with root routing + improve tests

### LINKED ISSUE #11122: Wrong route matched when http request to /
### Expected Behavior

With three routes /, /{id} and /{id}/relations and sending a request to / I expect the root route to match. 

### Actual Behaviour

But instead the /{id}/relations route is matched.

### Steps To Reproduce

1. Add a controller with three routes /, /{id} and /{id}/something (maybe the middle one is not needed)
2. Send a http request to /
3. Get 400 response -> GET / (Bad Request): Required argument [String id] not specified

### Environment Information

 - MacOS 14.6.1
 - JDK 21

### Example Application

https://github.com/goeh/micronaut-460-issue

### Version

4.6.0

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/11122

Included `@Requires(property` to all the controllers in Netty Server tests to make debugging easier, added a test to fail if somebody adds a not-property-hidden controller.

### DIFF
--- a/http/src/main/java/io/micronaut/http/uri/UriTemplateMatcher.java
+++ b/http/src/main/java/io/micronaut/http/uri/UriTemplateMatcher.java
@@ -68,7 +68,7 @@ public final class UriTemplateMatcher implements UriMatcher, Comparable<UriTempl
         this.parts = parts;
         List<UriMatchVariable> variables = new ArrayList<>();
         this.segments = provideMatchSegments(parts, variables);
-        this.isRoot = segments.length == 0 || segments[0].type == SegmentType.LITERAL && isRoot(segments[0].value);
+        this.isRoot = segments.length == 0 || segments.length == 1 && segments[0].type == SegmentType.LITERAL && isRoot(segments[0].value);
         this.variables = Collections.unmodifiableList(variables);
     }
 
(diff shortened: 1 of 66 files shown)


graemerocher:
should there be an HTTP TCK test for the root mapping thing?

dstepanov:
Added

graemerocher (approved):

http/src/main/java/io/micronaut/http/uri/UriTemplateMatcher.java
```diff
@@ -68,7 +68,7 @@ private UriTemplateMatcher(String templateString, List<UriTemplateParser.Part> p
         this.parts = parts;
         List<UriMatchVariable> variables = new ArrayList<>();
         this.segments = provideMatchSegments(parts, variables);
-        this.isRoot = segments.length == 0 || segments[0].type == SegmentType.LITERAL && isRoot(segments[0].value);
+        this.isRoot = segments.length == 0 || segments.length == 1 && segments[0].type == SegmentType.LITERAL && isRoot(segments[0].value);
```
yawkat, line 71:
```suggestion
        this.isRoot = segments.length == 0 || (segments.length == 1 && segments[0].type == SegmentType.LITERAL && isRoot(segments[0].value));
```

yawkat (approved):

PR #11129, opened 2024-08-29
Cache Micronaut META-INF services

dstepanov:
I noticed some time ago that we loaded the Miconaut META-INF service multiple times. Now, it should be cached and loaded only once. Also, for Micronaut services, we can avoid previous text file entries. That should improve some startup time.

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
@@ -93,9 +93,10 @@ public final class MicronautMetaServiceLoaderUtils {
                                                               @NonNull String serviceName) throws IOException {
         CacheEntry ce = cacheEntry;
         if (ce == null || ce.classLoader != classLoader) {
-            cacheEntry = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
+            ce = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
+            cacheEntry = ce;
         }
-        return cacheEntry.services.getOrDefault(serviceName, Set.of());
+        return ce.services.getOrDefault(serviceName, Set.of());
     }
 
     /**
--- a/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
@@ -59,8 +59,8 @@ public final class MicronautMetaServiceLoaderUtils {
     private static final MethodHandles.Lookup LOOKUP = MethodHandles.publicLookup();
     private static final MethodType VOID_TYPE = MethodType.methodType(void.class);
 
-    private static ClassLoader CACHED_SERVICES_CLASSLOADER;
-    private static Map<String, Set<String>> CACHED_SERVICES;
+    private static ClassLoader cachedServicesClassloader;
+    private static Map<String, Set<String>> cachedServices;
 
     /**
      * Find all instantiated Micronaut service entries.
@@ -96,11 +96,11 @@ public final class MicronautMetaServiceLoaderUtils {
     @NonNull
     public static Set<String> findMicronautMetaServiceEntries(@NonNull ClassLoader classLoader,
                                                               @NonNull String serviceName) throws IOException {
-        if (CACHED_SERVICES == null || CACHED_SERVICES_CLASSLOADER != classLoader) {
-            CACHED_SERVICES = findAllMicronautMetaServices(classLoader);
-            CACHED_SERVICES_CLASSLOADER = classLoader;
+        if (cachedServices == null || cachedServicesClassloader != classLoader) {
+            cachedServices = findAllMicronautMetaServices(classLoader);
+            cachedServicesClassloader = classLoader;
         }
-        return CACHED_SERVICES.getOrDefault(serviceName, Set.of());
+        return cachedServices.getOrDefault(serviceName, Set.of());
     }
 
     /**
--- a/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
@@ -93,9 +93,9 @@ public final class MicronautMetaServiceLoaderUtils {
                                                               @NonNull String serviceName) throws IOException {
         CacheEntry ce = cacheEntry;
         if (ce == null || ce.classLoader != classLoader) {
-            cacheEntry = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
+            cacheEntry = ce = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
         }
-        return cacheEntry.services.getOrDefault(serviceName, Set.of());
+        return ce.services.getOrDefault(serviceName, Set.of());
     }
 
     /**
--- a/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
@@ -59,7 +59,8 @@ public final class MicronautMetaServiceLoaderUtils {
     private static final MethodHandles.Lookup LOOKUP = MethodHandles.publicLookup();
     private static final MethodType VOID_TYPE = MethodType.methodType(void.class);
 
-    private static Map<String, Set<String>> services;
+    private static ClassLoader CACHED_SERVICES_CLASSLOADER;
+    private static Map<String, Set<String>> CACHED_SERVICES;
 
     /**
      * Find all instantiated Micronaut service entries.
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -0,0 +1,334 @@
...
+     * @param serviceName The service name
+     * @return The entries
+     * @throws IOException
+     */
+    @NonNull
+    public static Set<String> findMicronautMetaServiceEntries(@NonNull ClassLoader classLoader,
+                                                              @NonNull String serviceName) throws IOException {
+        if (services == null) {
+            // Should we cache based on the classloader?
+            services = findAllMicronautMetaServices(classLoader);
```
dstepanov: Should this be synchronized?

graemerocher (approved):

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -0,0 +1,338 @@
...
+    private static final MethodHandles.Lookup LOOKUP = MethodHandles.publicLookup();
+    private static final MethodType VOID_TYPE = MethodType.methodType(void.class);
+
+    private static ClassLoader cachedServicesClassloader;
+    private static Map<String, Set<String>> cachedServices;
+
+    /**
+     * Find all instantiated Micronaut service entries.
+     *
+     * @param constructor  The final collection constructor by size
```
yawkat: do we need this argument? it's only ArrayList anyway

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -0,0 +1,338 @@
...
+     * @param <S> The service type
+     */
+    @SuppressWarnings("java:S1948")
+    private static final class MicronautServiceCollector<S> extends RecursiveActionValuesCollector<S> {
+
+        private final ClassLoader classLoader;
+        private final String serviceName;
+        private final Predicate<S> predicate;
+        private final List<RecursiveActionValuesCollector<S>> tasks = new ArrayList<>();
+        private final AtomicInteger size = new AtomicInteger();
```
yawkat: doesn't need to be atomic

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -0,0 +1,338 @@
...
+            }
+        }
+        return services;
+    }
+
+    private static <S> S instantiate(String className, ClassLoader classLoader) {
+        try {
+            @SuppressWarnings("unchecked") final Class<S> loadedClass =
+                (Class<S>) Class.forName(className, false, classLoader);
+            // MethodHandler should more performant than the basic reflection
```
yawkat, line 189: idk about that given that there's still a lookup involved

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
yawkat: maybe not but definitely needs to be volatile

yawkat: also there's a race condition with cachedServicesClassloader, maybe instead put the cachedServices and cachedServicesClassloader into a record and set them all at once so there's no race condition.

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -0,0 +1,338 @@
...
+
+            private Set<String> definitions;
+
+            @Override
+            public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
+                if (dir.endsWith(MICRONAUT_SERVICES_PATH)) {
+                    return FileVisitResult.CONTINUE;
+                }
+                String serviceName = dir.getFileName().toString();
+                definitions = services.get(serviceName);
```
yawkat, line 132: intellij tells me computeIfAbsent is possible here.

core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
```diff
@@ -162,80 +106,10 @@ private Set<String> computeStandardServiceTypeNames(URL url) {
...
     }
 
-    private boolean isWebSphereClassLoader() {
-        return classLoader.getClass().getName().startsWith("com.ibm.ws.classloader");
-    }
-
-    private String buildResourceSearchPath() {
-        String path = "META-INF/micronaut/" + serviceName;
-
-        if (isWebSphereClassLoader()) {
```
yawkat, line 172: do we still need this websphere special case

dstepanov: That was to look at the service directly now we should process the folder

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
dstepanov: I better to avoid lambda in the init phase

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
dstepanov: It's just a copy form the prev loader

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
dstepanov: Removed

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
dstepanov: Now it' a record

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
dstepanov: Removes

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -96,11 +91,10 @@ public static <S> List<S> findMetaMicronautServiceEntries(@NonNull Function<Inte
     @NonNull
     public static Set<String> findMicronautMetaServiceEntries(@NonNull ClassLoader classLoader,
                                                               @NonNull String serviceName) throws IOException {
-        if (cachedServices == null || cachedServicesClassloader != classLoader) {
-            cachedServices = findAllMicronautMetaServices(classLoader);
-            cachedServicesClassloader = classLoader;
+        if (cacheEntry == null || cacheEntry.classLoader != classLoader) {
+            cacheEntry = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
```
yawkat: Modify this code using local variables to have a single read from the volatile field and (if necessary) a single write. Otherwise there's still race conditions here.

dstepanov: Please check

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -91,7 +91,8 @@ public static <S> List<S> findMetaMicronautServiceEntries(@NonNull ClassLoader c
     @NonNull
     public static Set<String> findMicronautMetaServiceEntries(@NonNull ClassLoader classLoader,
                                                               @NonNull String serviceName) throws IOException {
-        if (cacheEntry == null || cacheEntry.classLoader != classLoader) {
+        CacheEntry ce = cacheEntry;
+        if (ce == null || ce.classLoader != classLoader) {
             cacheEntry = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
         }
         return cacheEntry.services.getOrDefault(serviceName, Set.of());
```
yawkat: also this read to local variable :)

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -0,0 +1,336 @@
...
+     * @throws IOException
+     */
+    @NonNull
+    public static Set<String> findMicronautMetaServiceEntries(@NonNull ClassLoader classLoader,
+                                                              @NonNull String serviceName) throws IOException {
+        CacheEntry ce = cacheEntry;
+        if (ce == null || ce.classLoader != classLoader) {
+            cacheEntry = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
+        }
+        return cacheEntry.services.getOrDefault(serviceName, Set.of());
```
yawkat:
```suggestion
        CacheEntry ce = cacheEntry;
        if (ce == null || ce.classLoader != classLoader) {
            cacheEntry = ce = new CacheEntry(classLoader, findAllMicronautMetaServices(classLoader));
        }
        return ce.services.getOrDefault(serviceName, Set.of());
```

yawkat (approved):

Issue #10771, opened 2024-04-27
Build native app on macOS without error at kotlin.coroutines.intrinsics.CoroutineSingletons


kurb70:
### Expected Behavior

Should build macOS native app without initializing during image building.

### Actual Behaviour

I got an error: Classes that should be initialized at run time got initialized during image building: kotlin.coroutines.intrinsics.CoroutineSingletons …

```
09:40:13: Executing 'nativeBuild'...

Starting Gradle Daemon...
Gradle Daemon started in 686 ms
> Task :generateJte UP-TO-DATE
> Task :compileJava UP-TO-DATE
> Task :compileGroovy NO-SOURCE
> Task :processResources UP-TO-DATE
> Task :classes UP-TO-DATE
> Task :jar UP-TO-DATE
> Task :generateResourcesConfigFile UP-TO-DATE

> Task :nativeCompile
[native-image-plugin] GraalVM Toolchain detection is disabled
[native-image-plugin] GraalVM location read from environment variable: JAVA_HOME
[native-image-plugin] Native Image executable path: /Users/kuba/.sdkman/candidates/java/21.0.2-graalce/lib/svm/bin/native-image
========================================================================================================================
GraalVM Native Image: Generating 'demo' (executable)...
========================================================================================================================
For detailed information and explanations on the build output, visit:
https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md
------------------------------------------------------------------------------------------------------------------------
[1/8] Initializing...                                                                                    (7,9s @ 0,27GB)
 Java version: 21.0.2+13, vendor version: GraalVM CE 21.0.2+13.1
 Graal compiler: optimization level: 2, target machine: armv8-a
 C compiler: cc (apple, arm64, 15.0.0)
 Garbage collector: Serial GC (max heap size: 80% of RAM)
 2 user-specific feature(s):
 - com.oracle.svm.thirdparty.gson.GsonFeature
 - io.micronaut.core.io.service.ServiceLoaderFeature
------------------------------------------------------------------------------------------------------------------------
Build resources:
 - 12,09GB of memory (75,6% of 16,00GB system memory, determined at start)
 - 8 thread(s) (100,0% of 8 available processor(s), determined at start)
[2/8] Performing analysis...  [*]                                                                       (73,3s @ 3,69GB)
   36 844 reachable types   (92,9% of   39 648 total)
   59 769 reachable fields  (69,2% of   86 405 total)
[...]

PR #11135, opened 2024-08-30
Add Kotlin Graalvm reproducer

dstepanov:
For https://github.com/micronaut-projects/micronaut-core/issues/10771


graemerocher:
it passes? Shouldn't this be in `micronaut-kotlin`?

dstepanov:
Yes, it's passed to the referenced issue that works. It's better to have one graal test in core

yawkat (approved):

PR #11142, opened 2024-08-31
Add support lost `java.time` classes:

altro3:
Period
LocalTime
MonthDay
YearMonth
Year
ZoneId

Also added these classes to `BASIC_TYPE_MAP`


context/src/main/java/io/micronaut/runtime/converters/time/IsoFormatter.java
```diff
@@ -0,0 +1,62 @@
...
+import static java.time.temporal.ChronoField.DAY_OF_MONTH;
+import static java.time.temporal.ChronoField.MONTH_OF_YEAR;
+import static java.time.temporal.ChronoField.YEAR;
+
+/**
+ * Just copied DateTimeFormatters form 'java.time' package.
+ *
+ * @since 4.7.0
+ */
+public class IsoFormatter {
```
yawkat: Do these constants have to be public and need to be placed here?

altro3: It's hard to say which is correct. All other formats are public, but for some reason these are private in java.time , although they are also formats according to the ISO standard. So, it's up to you to decide.

yawkat: Please make them private where they are used, we do not need a new IsoFormatter class

altro3: fixed

yawkat (approved):

PR #11158, opened 2024-09-06
Refactor DefaultHttpClient to use ByteBody for the request body

yawkat:
This is an intermediate PR of my refactor, in the interest of keeping the changes reviewable.


http-netty/src/main/java/io/micronaut/http/netty/NettyHttpRequestBuilder.java
```diff
@@ -74,12 +75,25 @@ default HttpRequest toHttpRequest() {
...
     }
 
+    /**
+     * Directly convert this request body to a {@link ByteBody}, if possible. If the body of this
+     * request has been changed, this will return an empty value.
+     *
+     * @return The body
+     */
+    @NonNull
+    default Optional<ByteBody> byteBodyDirect() {
```
dstepanov, line 93: For internal API I would just use nullable type

dstepanov (approved):

http-netty/src/main/java/io/micronaut/http/netty/NettyHttpRequestBuilder.java
yawkat: Fair but I don't want to deal with merge conflicts now :D I will do it in a future PR

yawkat: included in #11182

graemerocher (approved):
Does this solve turning the body into an input stream?

yawkat:
No, not yet. There is no API change. But it does make it easier, because now in principle we just need a conversion ByteBody->InputStream that just does `byteBody.toInputStream()`.

PR #11160, opened 2024-09-06
Store enum values as strings

dstepanov:
I think this was fixed before, but it's still broken for manually added annotations / values


graemerocher:
can this be back ported to 4.6.x?

dstepanov:
I targeted a wrong branch

graemerocher (approved):

graemerocher:
some tests are failing

Issue #11152, opened 2024-09-04
Problem with service loading in native executable


burtbeckwith:
### Expected Behavior

Missing JrtFileSystemProvider shouldn't cause running to fail

### Actual Behaviour

When running a native executable using 4.6.1, we're seeing

```
Exception in thread "main" java.util.ServiceConfigurationError: java.nio.file.spi.FileSystemProvider: Provider jdk.internal.jrtfs.JrtFileSystemProvider not found
	at java.base@21.0.5/java.util.ServiceLoader.fail(ServiceLoader.java:593)
	at java.base@21.0.5/java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(ServiceLoader.java:1219)
	at java.base@21.0.5/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1228)
	at java.base@21.0.5/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1273)
	at java.base@21.0.5/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1309)
	at java.base@21.0.5/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1393)
	at java.base@21.0.5/java.nio.file.FileSystems.newFileSystem(FileSystems.java:349)
	at io.micronaut.core.io.service.ServiceScanner.findMicronautMetaServiceConfigs(ServiceScanner.java:204)
	at io.micronaut.core.io.service.ServiceScanner$DefaultServiceCollector.compute(ServiceScanner.java:257)
	at java.base@21.0.5/java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:194)
	at java.base@21.0.5/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base@21.0.5/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base@21.0.5/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base@21.0.5/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base@21.0.5/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
[...]

PR #11166, opened 2024-09-08
Ignore any errors when getting jrt provider

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/11152

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
+++ b/core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
@@ -18,9 +18,6 @@ package io.micronaut.core.io.service;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.io.IOUtils;
-import java.nio.file.FileSystemNotFoundException;
-import java.nio.file.ProviderNotFoundException;
-import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/micronaut/core/io/service/ServiceScanner.java
```diff
@@ -234,6 +220,31 @@ private void findMicronautMetaServiceConfigs(BiConsumer<URI, String> consumer) t
...
+            URI uri = URI.create("jrt:/");
+            FileSystem fs = FileSystems.getFileSystem(uri);
+            if (fs.isOpen()) {
+                return fs;
+            }
+            fs = FileSystems.newFileSystem(uri, Collections.emptyMap(), classLoader);
+            if (fs.isOpen()) {
+                return fs;
+            }
+        } catch (Exception e) {
```
burtbeckwith: need to catch Throwable here, ServiceLoader.fail() throws ServiceConfigurationError

dstepanov: Will update

dstepanov: BTW I wasn't able to reproduce it. Is there something special in your case?

graemerocher (approved):

PR #11177, opened 2024-09-11
ByteBody-based response handling for the HTTP client

yawkat:
First version of replacing the previous mixed response handling with a single ByteBody-based handler (similar to PipeliningServerHandler on the server). The purpose of this is to (a) unify some code between the streaming and aggregating code paths in DefaultHttpClient, (b) allow ByteBody-based streaming for ProxyHttpClient in the future so that ProxyHttpClient can be implemented for the jdk client and servlet server, (c) eventually create a ByteBody-based HTTP client API (maybe just in ConnectionManager) that allows lower level access to the request and thus can be used to implement the oci sdk http client.


http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+import java.util.List;
+
+/**
+ * Inbound message handler for the HTTP/1.1 client. Also used for HTTP/2 and /3 through message
+ * translators.
+ *
+ * @author Jonas Konrad
+ * @since 4.7.0
+ */
+final class Http1ResponseHandler extends SimpleChannelInboundHandlerInstrumented<HttpObject> implements BufferConsumer.Upstream {
```
graemerocher: annotate `@Internal` as well

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+                    }
+                    break;
+                }
+                this.response = response;
+                state = State.BUFFERED_CONTENT;
+                // if the response is a FullHttpResponse, fall through to BUFFERED_CONTENT handling
+                if (!(msg instanceof HttpContent)) {
+                    break;
+                }
+                // fall-through
```
graemerocher: is there a way we can write this logic without fall through which is a bit of an anti-pattern except if you are a writing a lexer/parser

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+            case BUFFERED_CONTENT:
+                content = (HttpContent) msg;
+                content.touch();
+                if (content.content().isReadable()) {
+                    if (buffered == null) {
+                        buffered = new ArrayList<>();
```
graemerocher: is there a way to size this array list?

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+                }
+                break;
+            case DISCARDING_CONTINUE_CONTENT:
+                ((HttpContent) msg).content().release();
+                if (msg instanceof LastHttpContent) {
+                    state = State.BEFORE_RESPONSE;
+                }
+                break;
+            case AFTER_CONTENT:
+                LOG.warn("Discarding unexpected message {}", msg);
```
graemerocher: wrap in an if statement

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+        if (oldDemand <= 0 && state == State.UNBUFFERED_CONTENT) {
+            streamingContext.read();
+        }
+    }
+
+    private void complete(CloseableByteBody body) {
+        listener.complete(response, body);
+    }
+
+    private enum State {
```
graemerocher: rename to something more descriptive like BufferingState or something

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+        BUFFERED_CONTENT,
+        UNBUFFERED_CONTENT,
+        DISCARDING_CONTENT,
+        AFTER_CONTENT,
+    }
+
+    /**
+     * The response listener.
+     */
+    public interface ResponseListener {
```
graemerocher: does this need to be public? can it be sealed?

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+        /**
+         * Size limits for the request body.
+         *
+         * @return The size limits
+         */
+        default BodySizeLimits sizeLimits() {
```
graemerocher, line 407: add nullability annotations if this is designed for extension and public.

http-client/src/main/java/io/micronaut/http/client/netty/SseSplitter.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Split buffers into lines for SSE parsing.
+ *
+ * @author Jonas Konrad
+ * @since 4.7.0
+ */
+final class SseSplitter {
```
graemerocher, line 36: add @internal

http-client/src/main/java/io/micronaut/http/client/netty/SseSplitter.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Split a single buffer.
+     *
+     * @param buf The buffer
+     * @return The individual split pieces
+     */
+    @NonNull
+    static List<ByteBuf> split(@NonNull ByteBuf buf) {
+        buf.touch();
+        List<ByteBuf> split = new ArrayList<>();
```
graemerocher, line 46: can this be sized?

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1671,48 +1639,66 @@ private String getHostHeader(URI requestURI) {
...
+        // this function acts like a wrapper around HttpPostRequestEncoder. HttpPostRequestEncoder
+        // takes a request + form data and transforms it to a request + bytes. Because we only want
+        // the bytes, we need to copy the data from the netty request back to the original
+        // MutableHttpRequest. This is just the Content-Type header, which sometimes gets an extra
+        // boundary specifier that we need.
+
+        // build the mock netty request (only the content-type matters)
+        HttpRequest nettyRequest = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+        List<AsciiString> relevantHeaders = List.of(HttpHeaderNames.CONTENT_TYPE);
+        for (AsciiString header : relevantHeaders) {
```
dstepanov: What is the point of defining a list of one item and then iterating it twice?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -2085,30 +2131,40 @@ private static class CurrentEvent {
...
             this.responsePromise = responsePromise;
             this.parentRequest = parentRequest;
             this.finalRequest = finalRequest;
         }
 
-        @Override
-        public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-            ctx.read();
+        private static void setRedirectHeaders(@Nullable io.micronaut.http.HttpRequest<?> request, MutableHttpRequest<Object> redirectRequest) {
+            if (request != null) {
```
dstepanov, line 2146: I would prefer check for null and `return` to void more nesting

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: Unfortunately not. If I wanted to optimize this more I'd probably use netty recycleable lists though

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: Http1ResponseHandler is package-private so I think it's NBD

http-client/src/main/java/io/micronaut/http/client/netty/SseSplitter.java
yawkat: Technically yes but we'd have to scan the input buffer twice, I think it's not advantageous to do so

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: i made it package-private too

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+        if (state != State.AFTER_CONTENT) {
+            exceptionCaught(ctx, new ResponseClosedException("Connection closed before response was received"));
+        }
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        switch (state) {
```
dstepanov: Can we use a new style of switch and avoid fall-through?

http-netty/src/main/java/io/micronaut/http/netty/body/AvailableNettyByteBody.java
```diff
@@ -62,6 +64,31 @@ public static ByteBuf toByteBuf(AvailableByteBody body) {
...
+     * @param bodySizeLimits The body size limits to check
+     * @param buf The input buffer
+     * @return The body with the given input buffer, or a {@link StreamingNettyByteBody} with the
+     * appropriate content length error
+     */
+    @NonNull
+    public static CloseableByteBody createChecked(@NonNull EventLoop loop, @NonNull BodySizeLimits bodySizeLimits, @NonNull ByteBuf buf) {
+        // AvailableNettyByteBody does not support exceptions, so if we hit one of the configured
+        // limits, we return a StreamingNettyByteBody instead.
+        if (buf.readableBytes() > bodySizeLimits.maxBodySize() || buf.readableBytes() > bodySizeLimits.maxBufferSize()) {
```
dstepanov: I would extract `buf.readableBytes()`

http-netty/src/main/java/io/micronaut/http/netty/body/AvailableNettyByteBody.java
```diff
@@ -62,6 +64,31 @@ public static ByteBuf toByteBuf(AvailableByteBody body) {
...
+            BufferConsumer.Upstream upstream = bytesConsumed -> {
```
dstepanov, line 83: Can this be a constant?

http-netty/src/main/java/io/micronaut/http/netty/body/AvailableNettyByteBody.java
```diff
@@ -62,6 +64,31 @@ public static ByteBuf toByteBuf(AvailableByteBody body) {
...
+            };
+            StreamingNettyByteBody.SharedBuffer mockBuffer = new StreamingNettyByteBody.SharedBuffer(loop, bodySizeLimits, upstream);
+            mockBuffer.add(buf); // this will trigger the exception for exceeded body or buffer size
+            return new StreamingNettyByteBody(mockBuffer);
+        } else {
+            return new AvailableNettyByteBody(buf);
```
dstepanov, line 89: Personally, I would revert the if and return `AvailableNettyByteBody` and avoid redundant else

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+    protected void channelReadInstrumented(ChannelHandlerContext ctx, HttpObject msg) throws Exception {
+        if (msg.decoderResult().isFailure()) {
+            ReferenceCountUtil.release(msg);
+            exceptionCaught(ctx, msg.decoderResult().cause());
+            return;
+        }
+
+        HttpContent content;
+        switch (state) {
+            case BEFORE_RESPONSE:
```
dstepanov: This switch-enum makes more sense with `state == null` as BEFORE_RESPONSE.
All other cases are content, which means you can already cast it to `HttpContent`

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: It could be refactored, but I think the alternatives are worse. This case is only for a FullHttpResponse, which is a combination of HttpResponse and LastHttpContent, so this code treats it like a HttpResponse (in the BEFORE_RESPONSE branch) and then a HttpContent (in the BUFFERED_CONTENT branch). To remove this fall-through I'd have to use a loop or recursion to handle the same message twice which is imo even more confusing because this is the only case where it's necessary.

In the CONTINUE handling code just above these lines the same thing is actually handled without fall-through: if the message is *not* a FullHttpResponse, the handler enters a DISCARDING_CONTINUE_CONTENT state, which then goes back to BEFORE_RESPONSE once the associated LastHttpContent comes in. If the message *is* a FullHttpResponse, the DISCARDING_CONTINUE_CONTENT code is done immediately. It's just that for CONTINUE, the code is very simple (just release the content) so the duplication is a non-issue.

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
```diff
@@ -0,0 +1,355 @@
...
+                HttpResponse response = (HttpResponse) msg;
+                if (response.status().code() == HttpResponseStatus.CONTINUE.code()) {
+                    listener.continueReceived(ctx);
+                    if (response instanceof LastHttpContent lhc) {
+                        lhc.release();
+                    } else {
+                        assert !(response instanceof HttpContent);
+                        state = State.DISCARDING_CONTINUE_CONTENT;
+                    }
+                    break;
```
dstepanov: This code looks complex because of the different state fields and switch breaks, which make it hard to follow the flow. I would represent the state as an object that can react to read, exception, discard, etc. Each state would contain only the fields it needs.

pseudocode:

if (state == null) {
    assume response
    state = new BUFFERED_CONTENT()
}

state = state.read(...)


void channelReadComplete() {
    state.readComplete()
}

dstepanov (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: Not sure. I don't want to touch this rn, it's from the old implementation (just moved). Maybe this happened as part of the form handling changes

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: Good idea. That will make the code more clear

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: ok

http-netty/src/main/java/io/micronaut/http/netty/body/AvailableNettyByteBody.java
yawkat: In principle yes, but I think it would actually be worse because this branch is rarely hit, and putting the lambda in a constant means it needs to be eagerly initialized

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: that sounds like the PipeliningServerHandler approach. will try

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: refactored

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
yawkat: pushed, it looks quite different now.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: don't want to touch because this is old code.

http-client/src/main/java/io/micronaut/http/client/netty/Http1ResponseHandler.java
dstepanov: Can you make the classes static and set the state only in the handler?

dstepanov: Have the methods in `ReaderState` always return `ReaderState`?

yawkat: No it's not possible. There are some necessary ordering conditions. For example, BufferedContent must update the state before listener.complete is called, because listener.complete may trigger further operations that need to be sent to the new state.

dstepanov: ok

dstepanov: I thought that the handler could not be accessed in parallel, so the method has to finish before a second invocation of any method on a handler or a listener

yawkat: netty allows no multi-threading for a handler but it does allow reentrancy. e.g. if you do ctx.read inside a handler method, channelRead of the same handler may be called inside the ctx.read call. It's less strict than reactor in this regard.

dstepanov: I see

dstepanov (approved):

graemerocher (approved):

yawkat:
i added some additional sanity checks in the state transitions of the decoder.

PR #11182, opened 2024-09-13
HTTP client refactor

yawkat:
This is the continuation of #11158 and #11177.

Now that the previous PRs have migrated the reading/writing to ByteBody, it is time to unify the code paths. This PR is a complex refactor of DefaultHttpClient that merges most of the code for normal exchange requests, streaming requests, and proxy requests into a single `sendRequestWithRedirects` method. In particular, logging, filtering, connection management, redirect handling and error handling are now merged using the ByteBody API.

This PR is a result of various small refactors that all preserve behavior, but the combined change is large enough that it basically looks like a rewrite. Behavior should be *almost* the same, and all tests still pass. There are some minor changes I can think of, but they seem positive:

- For exchange requests, the configured request-timeout is only applied to the "top" request instead of all nested redirect requests as well. The inner timeouts are pointless anyway, since the outer timeout will trigger before the inner timeouts do.
- There was some confusion in the code between the "parent request" (i.e. the request from the ServerRequestContext) and the "original request" (the request that produced a redirect) when it comes to redirect handling, in particular for streamed requests. I think this was simply a bug that was never hit by the test cases because they mostly hit the non-streaming client.
- Client filters should be a bit more powerful now. Before this PR, URI resolution (i.e. when you do `GET("/foo")` it would resolve to `GET("https://example.com/foo")` depending on service ID) would happen before filters, and then would be carried next to the request through the filters. This means that filters can (a) not replace the request because provideResponse would ignore the new request, and (b) they could not mutate the URI of the original request because the new URI would be ignored in favor of the previously resolved URI. In the new implementation, the URI is still resolved before the filters, but it's set on the request, sent through the filters, and then the filtered request is used in the downstream code. So it should be possible to change the URI in a filter, but I've not added a test.

There are still some more changes I want to make to the client, e.g. getting rid of StreamingHttpRequest for the proxy client. I'm also thinking of exposing `sendRawRequest` as a new API (another HttpClient interface). But these will happen in future PRs.


yawkat:
FYI @dstepanov, this PR also migrates the client to use mostly Mono, so this is 90% of the way to ExecutionFlow already. I may do that in a future PR, it just needs a solution to the cancellation feature.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1844,6 +1963,116 @@ private <E extends HttpClientException> E decorate(E exc) {
...
+            result = ReadTimeoutException.TIMEOUT_EXCEPTION;
+        } else if (cause instanceof HttpClientException hce) {
+            result = decorate(hce);
+        } else {
+            result = decorate(new HttpClientException("Error occurred reading HTTP response: " + message, cause));
+        }
+        return result;
+    }
+
+    private static void setRedirectHeaders(@Nullable io.micronaut.http.HttpRequest<?> request, MutableHttpRequest<Object> redirectRequest) {
```
graemerocher, line 2000: can we reduce the number of nested for/if statements by refactoring this method to a make it easier to read

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,182 +1449,359 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
+                } else {
+                    channel.writeAndFlush(nettyRequest, channel.voidPromise());
+                }
+            } else {
+                channel.writeAndFlush(nettyRequest, channel.voidPromise());
+                if (!expectContinue) {
+                    streamWriter.startWriting();
+                }
+            }
+        }).subscribeOn(Schedulers.fromExecutor(poolHandle.channel.eventLoop()));
```
graemerocher, line 1765: does the executor here need to be customisable?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,182 +1449,359 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
-        // return the body bytes
+
+            if (log.isTraceEnabled()) {
+                HttpHeadersUtil.trace(log, nettyRequest.headers().names(), nettyRequest.headers()::getAll);
+                if (byteBuf != null) {
+                    traceBody("Request", byteBuf);
+                }
+            }
+
+            pipeline.addLast(ChannelPipelineCustomizer.HANDLER_MICRONAUT_HTTP_RESPONSE, new Http1ResponseHandler(new Http1ResponseHandler.ResponseListener() {
```
graemerocher, line 1663: can this be pulled into an inner class to make it easier to read

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,182 +1449,359 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
-            },
-            poolHandle.channel.eventLoop()
-        );
-        String newUri = requestURI.getRawPath();
-        if (requestURI.getRawQuery() != null) {
-            newUri += "?" + requestURI.getRawQuery();
+        List<GenericHttpFilter> filters =
+            filterResolver.resolveFilters(request, clientFilterEntries);
+        if (parentRequest != null) {
+            // todo: migrate to new filter
```
graemerocher, line 1498: when will this todo be addressed?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1047,18 +1125,42 @@ private <I> Publisher<HttpResponse<?>> buildStreamExchange(
...
+        return this.sendRequestWithRedirects(
+            parentRequest,
+            null,
+            request.uri(requestURI),
+            (req, resp) -> {
+                ByteBody bb = resp.byteBody();
+                Publisher<HttpContent> body;
+                if (!hasBody(resp)) {
+                    resp.close();
+                    body = Publishers.empty();
```
graemerocher: can probably return `Flux.empty()` here since we are already using reactor

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1047,18 +1125,42 @@ private <I> Publisher<HttpResponse<?>> buildStreamExchange(
...
-                applyFilterToResponsePublisher(parentRequest, request, requestURI, streamResponsePublisher)
+                return readBodyOnError(errorType, Mono.<HttpResponse<?>>just(toStreamingResponse(resp, body))
+                    .flatMap(r -> handleStreamHttpError(r, true)));
+            }
         );
+    }
 
-        return streamResponsePublisher;
+    private <B> MutableHttpResponse<B> toStreamingResponse(NettyClientByteBodyResponse resp, Publisher<HttpContent> content) {
+        DefaultStreamedHttpResponse nettyResponse = new DefaultStreamedHttpResponse(resp.nettyResponse.protocolVersion(), resp.nettyResponse.status(), resp.getHeaders().getNettyHeaders(), content);
```
graemerocher: try and shorten this line be extracting passed arguments to veriables

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1070,27 +1172,32 @@ public Publisher<MutableHttpResponse<?>> proxy(@NonNull io.micronaut.http.HttpRe
...
-                            )
+                    return this.sendRequestWithRedirects(
+                        request,
+                        null,
+                        httpRequest.uri(requestURI),
+                        (req, resp) -> {
+                            Publisher<HttpContent> body;
+                            if (!hasBody(resp)) {
+                                resp.close();
+                                body = Publishers.empty();
```
graemerocher: can probably be `Mono.empty()`

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -627,8 +627,8 @@ public Optional<io.netty.handler.codec.http.HttpRequest> toHttpRequestDirect() {
     }
 
     @Override
-    public @NonNull Optional<ByteBody> byteBodyDirect() {
-        return Optional.of(byteBody());
+    public ByteBody byteBodyDirect() {
```
graemerocher, line 630: is this now `Nullable`?

http-client/src/main/java/io/micronaut/http/client/netty/StreamWriter.java
```diff
@@ -0,0 +1,157 @@
...
+            // discarded
+            return;
+        }
+
+        ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT, ctx.voidPromise());
+        completed = true;
+    }
+
+    @Override
+    public void discard() {
```
graemerocher, line 142: add implementation note here why this method is not implemented

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: possibly, I want to keep it this way in this PR though because it's the same code as before and this PR is already complicated enough

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: it's from the old code (applyFilterToResponsePublisher). it's a bit annoying to port to the new filter api since it's a programmatic filter and the new filter api is only declarative

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: No. It is important that the above `Mono.create` lambda runs on the event loop so that the pipeline setup, write calls etc all happen synchronously

yawkat: I will add a comment

http-client/src/main/java/io/micronaut/http/client/netty/NettyClientByteBodyResponse.java
```diff
@@ -0,0 +1,81 @@
...
+import io.micronaut.http.body.ByteBody;
+import io.micronaut.http.body.CloseableByteBody;
+import io.micronaut.http.netty.NettyHttpHeaders;
+import io.netty.handler.codec.http.HttpResponse;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+@Internal
+final class NettyClientByteBodyResponse implements ByteBodyHttpResponse<Object> {
```
dstepanov, line 40: Please add javadoc

http-client/src/main/java/io/micronaut/http/client/netty/StreamWriter.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.function.Consumer;
+
+/**
+ * This class is used to write from a {@link StreamingNettyByteBody} to a channel with appropriate
+ * backpressure control.
+ *
+ * @author Jonas Konrad
+ * @since 4.7.0
+ */
+final class StreamWriter extends ChannelInboundHandlerAdapter implements BufferConsumer {
```
dstepanov, line 38: Make internal

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: id prefer not to, it's only like 50 lines but it would require 7 constructor parameters from the method

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: yes, it's marked as such on the super method

graemerocher (approved):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -867,16 +859,102 @@ public <I, O> Publisher<O> jsonStream(@NonNull io.micronaut.http.HttpRequest<I>
...
-            .switchMap(uri -> (Publisher) exchangeImpl(uri, parentRequest, toMutableRequest(request), bodyType, errorType, blockHint));
+        Mono<HttpResponse<O>> mono = resolveRequestURI(request)
+            .flatMap(uri -> {
+                MutableHttpRequest<?> mutableRequest = toMutableRequest(request).uri(uri);
+                //noinspection unchecked
+                return sendRequestWithRedirects(
+                    parentRequest,
+                    blockHint,
+                    mutableRequest,
+                    (req, resp) -> Mono.<HttpResponse<O>>from(ReactiveExecutionFlow.fromFlow(InternalByteBody.bufferFlow(resp.byteBody())
```
dstepanov: Can you extract a new method?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,181 +1454,360 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
-        );
-        String newUri = requestURI.getRawPath();
-        if (requestURI.getRawQuery() != null) {
-            newUri += "?" + requestURI.getRawQuery();
+        List<GenericHttpFilter> filters =
+            filterResolver.resolveFilters(request, clientFilterEntries);
+        if (parentRequest != null) {
+            // todo: migrate to new filter
+            filters.add(
+                GenericHttpFilter.createLegacyFilter(new ClientServerContextFilter(parentRequest), new FilterOrder.Fixed(Ordered.HIGHEST_PRECEDENCE))
```
dstepanov, line 1500: Do we still need this?

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1047,18 +1125,47 @@ private <I> Publisher<HttpResponse<?>> buildStreamExchange(
             @NonNull MutableHttpRequest<I> request,
             @NonNull URI requestURI,
             @Nullable Argument<?> errorType) {
+        return this.sendRequestWithRedirects(
+            parentRequest,
+            null,
+            request.uri(requestURI),
+            (req, resp) -> {
```
dstepanov, line 1134: Please extract a new method

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1070,27 +1177,32 @@ public Publisher<MutableHttpResponse<?>> proxy(@NonNull io.micronaut.http.HttpRe
...
-                                    request,
-                                    requestWrapper.get(),
-                                    requestURI,
-                                    proxyResponsePublisher
-                            )
+                    return this.sendRequestWithRedirects(
+                        request,
+                        null,
+                        httpRequest.uri(requestURI),
+                        (req, resp) -> {
```
dstepanov, line 1193: Please extract a new method

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,181 +1454,360 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
-                }
-            },
-            poolHandle.channel.eventLoop()
-        );
-        String newUri = requestURI.getRawPath();
-        if (requestURI.getRawQuery() != null) {
-            newUri += "?" + requestURI.getRawQuery();
+        List<GenericHttpFilter> filters =
+            filterResolver.resolveFilters(request, clientFilterEntries);
+        if (parentRequest != null) {
```
dstepanov, line 1497: Do we always have filters? Maybe we can skip some code if it's emtpy

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,181 +1454,360 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
+    ) {
+        RequestKey requestKey;
+        try {
+            requestKey = new RequestKey(this, request.getUri());
+        } catch (Exception e) {
+            return Mono.error(e);
         }
+        // first: connect
+        return connectionManager.connect(requestKey, blockHint)
+            .flatMap(poolHandle -> {
```
dstepanov, line 1552: Please extract a new methoda

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,181 +1454,360 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
-        return byteBufferFactory.copiedBuffer(
-                bodyValue.toString().getBytes(
-                        requestContentType.getCharset().orElse(defaultCharset)
-                )
-        ).asNativeBuffer();
-    }
+                // send the raw request
+                return sendRawRequest(poolHandle, request, byteBody, c -> handleResponseError(request, c));
+            })
+            .flatMap(byteBodyResponse -> {
```
dstepanov, line 1577: Please extract a new method

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,181 +1454,360 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
 
-    private ByteBuf charSequenceToByteBuf(CharSequence bodyValue, MediaType requestContentType) {
-        return byteBufferFactory.copiedBuffer(
-                bodyValue.toString().getBytes(
-                        requestContentType.getCharset().orElse(defaultCharset)
-                )
-        ).asNativeBuffer();
-    }
+                // send the raw request
+                return sendRawRequest(poolHandle, request, byteBody, c -> handleResponseError(request, c));
```
dstepanov, line 1575: More correct would be to flatMap response of this method instead of the outer one to avoid an extra call for the error path

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -1491,181 +1454,360 @@ private <I> Publisher<URI> resolveURI(io.micronaut.http.HttpRequest<I> request,
...
+            if (byteBody instanceof AvailableNettyByteBody available) {
+                byteBuf = AvailableNettyByteBody.toByteBuf(available);
+                streamWriter = null;
+            } else {
+                streamWriter = new StreamWriter((StreamingNettyByteBody) byteBody, e -> {
+                    poolHandle.taint();
+                    sink.error(e);
+                });
+                pipeline.addLast(streamWriter);
+                byteBuf = null;
```
dstepanov, line 1653: Intellij noticed that byteBuf can be null but the Netty API doesn't allow it to be null

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -867,16 +859,102 @@ public <I, O> Publisher<O> jsonStream(@NonNull io.micronaut.http.HttpRequest<I>
...
+                                ));
+                                return ExecutionFlow.error(clientResponseError);
+                            } finally {
+                                fullHttpResponse.release();
+                            }
+                        })).toPublisher())
+                ).map(r -> (HttpResponse<O>) r);
+            });
+
+        Duration requestTimeout = configuration.getRequestTimeout();
```
dstepanov, line 885: Please extract `applyConfiguration`

dstepanov (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: imo extracting these lambdas makes the code more spaghetti because it's not immediately clear that this code is only called in one place.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: there's usually at least the parent request filter

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: i dont agree, this code is basically imperative

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: it's functionally identical

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: all uses of byteBuf are guarded with checks of either byteBuf or streamWriter

dstepanov:
Some of the code looks way too long, and the code is going beyond my screen. I prefer to extract some of the lamdas to have an appropriate name to make the code more understandable.
Sometimes, there are two different requests, which makes the code confusing.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
yawkat: i don't see why not

yawkat:
ive extracted the response handling for exchange. For the other four cases, here is my reasoning why I don't want to:

- The `buildStreamExchange` and `proxy` methods are thin wrappers around sendRequestWithRedirects, essentially in the form `proxy() { return sendRequestWithRedirects((req, resp) -> { ... actual code ... }) }`. Because they do so little outside the lambda, I think it's more readable to keep the code there, so there's less "indirection" when reading the code. This outweighs the disadvantage of having a large lambda, imo.
- `sendRequestWithRedirectsNoFilter` has two big lambdas, but imo it's still more readable this way. The mono code is essentially acting like an async-await, the flatMap calls wait for the result of the previous lambda and run the next lambda when that result is available. It is essentially imperative, so you can read the method from top to bottom.
- `sendRawRequest` has a giant Http1ResponseHandler.ResponseListener implementation that is not great. However, that anonymous class only has fairly light implementations of the six methods, only a few lines each. However it uses seven variables from the enclosing method, so if this was a separate class, id need 7 fields, 7 constructor args, 7 sets, which makes everything even more unreadable. So while I think the current version isn't great, the alternative seems worse.

dstepanov (approved):

PR #11187, opened 2024-09-15
fix(router): get 'uri' mapping default value if no 'uris' found in http annotation

### LINKED ISSUE #10957: DuplicateRouteException when using two Controller uris
### Expected Behavior

When I create two end-points like this:

```java
@Get(uris = { "/foo", "/bar" }
public String method1(){ ... }

@Get(uris = { "/one", "/two" }
public String method2(){ ... }
```

And call /foo then I expect `method1()` to be executed.

I would also expect if there is an error with duplicate routing that the error log would include the controller class and method names for easier debugging.

### Actual Behaviour

Instead I get DuplicateRouteException thrown from the `Router#findClosest(request)` with the message:

`More than 1 route matched the incoming request. The following routes matched /: GET - /, GET - /`

From this message it is not clear which controller methods are conflicting nor is the error correct. 

Apparently the Router only considers the `uri` parameter, not `uris`.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.2.2

weesftw:
Checks the default value of 'uri' only if 'uris' is empty, and adds the 'uri' value to the 'uris' list if it has any value assigned.

closes #10957

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/HttpUriMappingSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/HttpUriMappingSpec.groovy
@@ -1,0 +1,60 @@
+package io.micronaut.http.client
+
+import io.micronaut.context.annotation.Property
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpStatus
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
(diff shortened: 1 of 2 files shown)


dstepanov (approved):

dstepanov:
@weesftw Thanks!

PR #11210, opened 2024-09-27
feat: nicer error pages for HTML responses

sdelamo:
This PR changes the error processor to respond to a nice HTML error page if the request accepts an HTML Response. [Similar to the error pages of Kamal proxy](https://github.com/basecamp/kamal-proxy/pull/13). The background blue, is the same as the blue used in the Micronaut Website. 

This pull request also adds a `HtmlSanitizer` API, which [Micronaut Security will further enhance](https://github.com/micronaut-projects/micronaut-security/issues/1817).

~~I generated a SVG per status code, using a Shortcuts + Pixelmator script.~~

For JSON responses, the application will respond [`vnd.error`](https://github.com/blongden/vnd.error)as it currently does. 

It also allows multiple body assertions in TCK.

Currently, a not found page is displayed as: 

CleanShot 2024-09-27 at 14 23 07@2x

With the changes in this pull-request, it is displayed as: 

CleanShot 2024-09-27 at 14 22 33@2x

A Bad request (e.g., after a validation) currently is displayed as: 

CleanShot 2024-09-27 at 14 24 00@2x

With the changes in this pull-request, it is displayed as: 

CleanShot 2024-09-27 at 14 26 10@2x


http-server/src/main/java/io/micronaut/http/server/exceptions/response/BodyErrorResponseProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.http.HttpResponse;
+
+/**
+ * Provides an HTTP Response body of an error response.
+ * @author Sergio del Amo
+ * @since 4.7.0
+ *  @param <T> The body type
+ */
+public interface BodyErrorResponseProvider<T> {
```
graemerocher: rename to `ErrorResponseBodyProvider` which is more logical a name

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Default implementation of {@link ErrorResponseProcessor}.
+ * It delegates to {@link JsonBodyErrorResponseProvider} for JSON responses and to {@link HtmlBodyErrorResponseProvider} for HTML responses.
+ *
+ * @author Sergio del Amo
+ * @since 4.7.0
+ */
+@Internal
+@Singleton
+@Secondary
```
graemerocher: use `@Requires(missingBeans= ErrorResponseProcessor.class)` instead if possible

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+    private final HtmlBodyErrorResponseProvider<?> htmlBodyErrorResponseProvider;
+
+    DefaultErrorResponseProcessor(JsonBodyErrorResponseProvider<?> jsonBodyErrorResponseProvider, HtmlBodyErrorResponseProvider<?> htmlBodyErrorResponseProvider) {
+        this.jsonBodyErrorResponseProvider = jsonBodyErrorResponseProvider;
+        this.htmlBodyErrorResponseProvider = htmlBodyErrorResponseProvider;
+    }
+
+    @Override
+    public MutableHttpResponse processResponse(ErrorContext errorContext, MutableHttpResponse response) {
+        if (errorContext.getRequest().getMethod() == HttpMethod.HEAD) {
```
graemerocher: extract the request into a local variable and reuse

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+            return (MutableHttpResponse<JsonError>) response;
+        }
+        final boolean isError = response.status().getCode() >= 400;
+        if (errorContext.getRequest().accept().stream().anyMatch(mediaType -> mediaType.equals(MediaType.TEXT_HTML_TYPE)) && isError) {
```
graemerocher: flip the condition to avoid evaluating the lambda if `isError` returns false

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultJsonBodyErrorResponseProvider.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+
+/**
+ * Default implementation of {@link JsonBodyErrorResponseProvider} which returns a {@link JsonError}.
+ *
+ * @since 4.7.0
+ */
+@Internal
+@Singleton
+class DefaultJsonBodyErrorResponseProvider implements JsonBodyErrorResponseProvider<JsonError> {
```
graemerocher: should be `final` if possible

graemerocher (changes requested):
I couldn't comment directly in the diff for https://github.com/micronaut-projects/micronaut-core/blob/b057c7457fb34429649b0967b1994f3dcc1ec0f0/http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlBodyErrorResponseProvider.java for some reason

But this class should use a concurrent linked hash map to avoid growing too big or attacks where someone specifies infinite locales and thus filling memory

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultJsonBodyErrorResponseProvider.java
sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/4ded5ae25b2f9e0aba7744bd7505211e5ba6883e

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/5fd2b03688f47d57440c13c4d1e8b11dbc5b48cc

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/845e94ab174d24e55800c1af9876e58719e8e2d6

http-server/src/main/java/io/micronaut/http/server/exceptions/response/BodyErrorResponseProvider.java
sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/a36464e806064b6123641db2699fdede29b0d1e9

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
sdelamo: see: https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/0136f34b1ac065aa73ca2e890d7787a602661e7b

sdelamo (commented):
@graemerocher, I have removed the SVG maps and replaced them with CSS, which has a similar effect. I did this to reduce the size of the file. With an SVG for every status code, the size was more than 1Mb. 

You should now be able to see the file in the GitHub Pull-Request review.

http-server-netty/src/test/resources/logback.xml
```diff
@@ -24,4 +24,5 @@
     <!--<logger name="io.micronaut.websocket" level="TRACE"/>-->
 
     <logger name="io.micronaut.core.async.subscriber.CompletionAwareSubscriber" level="DEBUG"/>
+    <logger name="io.micronaut.http.server.exceptions.response" level="TRACE"/>
```
yawkat:
```suggestion
```

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public MutableHttpResponse processResponse(ErrorContext errorContext, MutableHttpResponse response) {
+        HttpRequest<?> request = errorContext.getRequest();
+        if (request.getMethod() == HttpMethod.HEAD) {
+            return (MutableHttpResponse<JsonError>) response;
+        }
+        final boolean isError = response.status().getCode() >= 400;
+        if (isError && request.accept().stream().anyMatch(mediaType -> mediaType.equals(MediaType.TEXT_HTML_TYPE))) {
```
yawkat: i think this should also have a check that the client does not accept application/json or text/json.

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public String body(@NonNull ErrorContext errorContext, @NonNull HttpResponse<?> response) {
+        final HttpStatus status = response.status();
+        Locale locale = localeResolver.resolveOrDefault(errorContext.getRequest());
+        return cache.computeIfAbsent(new LocaleStatus(locale, status), key -> html(locale, status, errorContext));
+    }
+
+    private String html(Locale locale, HttpStatus status, ErrorContext errorContext) {
+        final String errorTitleCode = status.getCode() + ".error.title";
+        final String errorTitle = messageSource.getMessage(errorTitleCode, status.getReason(), locale);
```
yawkat: XSS potential on status.getReason

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,190 @@
...
+
+    DefaultHtmlErrorResponseBodyProvider(MessageSource messageSource,
+                                         LocaleResolver<HttpRequest<?>> localeResolver) {
+        this.messageSource = messageSource;
+        this.localeResolver = localeResolver;
+    }
+
+    @Override
+    public String body(@NonNull ErrorContext errorContext, @NonNull HttpResponse<?> response) {
+        final HttpStatus status = response.status();
```
yawkat: this can error for custom status codes

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
yawkat: well not right now, since HttpStatus is an enum, but potentially if you fix the status error thing from above.

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,190 @@
...
+        final String errorCode = status.getCode() + ".error";
+        String defaultErrorBold = DEFAULT_ERROR_BOLD.get(status);
+        String defaultError = DEFAULT_ERROR.get(status);
+        String errorBold = defaultErrorBold != null ? messageSource.getMessage(errorBoldCode, defaultErrorBold, locale) : messageSource.getMessage(errorBoldCode, locale).orElse(null);
+        String error = defaultError != null ? messageSource.getMessage(errorCode, defaultError, locale) : messageSource.getMessage(errorCode, locale).orElse(null);
+        StringBuilder sb = new StringBuilder();
+
+        for (io.micronaut.http.server.exceptions.response.Error e : errorContext.getErrors()) {
+            if (!e.getMessage().equalsIgnoreCase(status.getReason())) {
+                sb.append(e.getMessage());
```
yawkat: XSS

http-server/src/main/java/io/micronaut/http/server/exceptions/response/ErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,34 @@
...
+package io.micronaut.http.server.exceptions.response;
+
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.http.HttpResponse;
+
+/**
+ * Provides an HTTP Response body of an error response.
+ * @author Sergio del Amo
+ * @since 4.7.0
+ *  @param <T> The body type
```
yawkat:
```suggestion
 * @param <T> The body type
```

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
dstepanov: I think this should be `.matchesType`

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,190 @@
...
+import static io.micronaut.http.HttpStatus.*;
+
+/**
+ * It generates HTML error response page for a given {@link HttpStatus}.
+ * @author Sergio del Amo
+ * @since 4.7.0
+ */
+@Internal
+@Singleton
+final class DefaultHtmlErrorResponseBodyProvider implements HtmlErrorResponseBodyProvider<String> {
```
dstepanov: Maybe all the templates can be a property, so a user can override it?

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultJsonErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,67 @@
...
+        JsonError error;
+        if (!errorContext.hasErrors()) {
+            error = new JsonError(response.reason());
+        } else if (errorContext.getErrors().size() == 1 && !alwaysSerializeErrorsAsList) {
+            Error jsonError = errorContext.getErrors().get(0);
+            error = new JsonError(jsonError.getMessage());
+            jsonError.getPath().ifPresent(error::path);
+        } else {
+            error = new JsonError(response.reason());
+            List<Resource> errors = new ArrayList<>();
```
dstepanov: You can initialize the collection to have the size of `errorContext.getErrors().size`

http-server/src/main/java/io/micronaut/http/server/exceptions/response/ErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public interface ErrorResponseBodyProvider<T> {
+    @NonNull
+    T body(@NonNull ErrorContext errorContext, @NonNull HttpResponse<?> response);
+
+    @NonNull
+    String contentType();
```
dstepanov, line 41: Missing JavaDoc

dstepanov (changes requested):

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
yawkat: Then this error page would also be used for `Accept: */*`

dstepanov: Ah ok

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
sdelamo: I made the class internal. If I want to the class extendable I would make it non internal. For now, I think you can provide your own implementation of `HtmlErrorResponseBodyProvider` if you want to customize it. Or provide localization for other languages.

http-server/src/main/java/io/micronaut/http/server/exceptions/response/HtmlErrorResponseBodyProvider.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A {@link ErrorResponseBodyProvider} for HTML responses.
+ * Responses with content type {@link io.micronaut.http.MediaType#TEXT_HTML}.
+ * @author Sergio del Amo
+ * @since 4.7.0
+ * @param <T> The body type
+ */
+@DefaultImplementation(DefaultHtmlErrorResponseBodyProvider.class)
+public interface HtmlErrorResponseBodyProvider<T> extends ErrorResponseBodyProvider<T> {
```
dstepanov: Maybe T should be a String?

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultErrorResponseProcessor.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/810bc55238d6f26381b0e4a88cb1bee8a3402215

http-server/src/main/java/io/micronaut/http/server/exceptions/response/HtmlErrorResponseBodyProvider.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/6918fc35ca248f1b0e39f46649e8477c07f91553

http-server/src/main/java/io/micronaut/http/server/exceptions/response/ErrorResponseBodyProvider.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/031a604869e88b3b6ee021a053370385e38ab1d6

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultJsonErrorResponseBodyProvider.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/de6f89cbe94e20742b0a61f4938de6c1124e5bbb

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
sdelamo: not sure how Cross Site Scripting (XSS) attack could be done because we are outputting the `Error::getMessage` in the error processor. Are we afraid of leaking internals of the application to a potential attacker? 

Please, note [we are already doing this](https://github.com/micronaut-projects/micronaut-core/blob/4.7.x/http-server/src/main/java/io/micronaut/http/server/exceptions/response/HateoasErrorResponseProcessor.java#L59) for JSON responses.

http/src/main/java/io/micronaut/http/MediaTypeUtils.java
```diff
@@ -0,0 +1,44 @@
...
+ * @since 4.7.0
+ */
+public final class MediaTypeUtils {
+
+    /**
+     *
+     * @param mediaType Media Type
+     * @return Returns true if the media type is {@link MediaType#APPLICATION_JSON_TYPE}, {@link MediaType#TEXT_JSON_TYPE}, {@link MediaType#APPLICATION_HAL_JSON_TYPE}, {@link MediaType#APPLICATION_JSON_GITHUB_TYPE}, {@link MediaType#APPLICATION_JSON_FEED_TYPE}, {@link {@link MediaType#APPLICATION_JSON_PROBLEM_TYPE}, {@link MediaType#APPLICATION_JSON_PATCH_TYPE}, {@link MediaType#APPLICATION_JSON_MERGE_PATCH_TYPE} or {@link MediaType#APPLICATION_JSON_SCHEMA_TYPE}.
+     */
+    public static boolean isJson(@NonNull MediaType mediaType) {
```
dstepanov: Use `mediaType.matchesExtension(MediaType.EXTENSION_JSON)`

dstepanov (changes requested):

http/src/main/java/io/micronaut/http/MediaTypeUtils.java
sdelamo: This list the current implementation: 

```java
public boolean matchesExtension(String matchExtension) {
        return extension.equalsIgnoreCase(ALL_TYPE.extension) || extension.equals(WILDCARD) || extension.equalsIgnoreCase(matchExtension);
    }
```
    
It will match `"*/*" and `*`. We don't want that, do we?

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
sdelamo: I don't find any reference in the documentation about custom status codes? https://docs.micronaut.io/latest/guide/#statusAnnotation

where is that?

sdelamo: I assume is this is the PR for custom HTTP Status https://github.com/micronaut-projects/micronaut-core/pull/8147/files

http/src/main/java/io/micronaut/http/MediaTypeUtils.java
dstepanov: First skip processing for any media type an then do this check

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
sdelamo: Is this what you are suggesting? https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/27a15c35d0938d25fd9fd65231b823cab430839c 

do you suggest to use `HttpResponse::code` instead of `HttpResponse::status`?

yawkat: yep use .code and .reason. Note however that .reason is potentially user-controlled.

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
yawkat: The error message can potentially contain user-controlled data which may be HTML which can be used for XSS

sdelamo: so, you are suggesting that we don't output non standard reasons ?

yawkat: Either that, or sanitize/escape them

http/src/main/java/io/micronaut/http/MediaTypeUtils.java
sdelamo: Honestly, I think the `MediaType::matchesExtension` method is misleading. It should be named, `matches extensionsAndAllAndWildcard`

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
sdelamo: >  user-controlled data 

is not the programmer of the application the one who populates custom status reasons?

yawkat: Yep but we can't rely on them to only do sensible things

sdelamo: I don't see the difference between a programmer outputting a custom status reason or a programmer outputting a custom response in a controller. He could create a security vulnerability in any case. I don't see how a custom status reason can be leverage as an attack vector from an outside user.

sdelamo: We could introduce [an API to sanitize HTML ](https://github.com/micronaut-projects/micronaut-security/issues/1817).

yawkat: To avoid [injection vulnerabilities](https://owasp.org/www-community/Injection_Flaws) (not just XSS), when outputting a structured format, strings must be validated, sanitized or escaped, unless they are from a limited set of strings known not be harmful (e.g. the reason phrases of the HttpStatus enum). This principle applies everywhere in the framework, such as HTTP/1 header generation, JSON generation, form generation etc. 

Where the strings come from is secondary. We cannot take strings supplied by the app programmer and output them unescaped.

http/src/main/java/io/micronaut/http/MediaTypeUtils.java
dstepanov: Can you use existing methods and not introduce a new util class?

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
yawkat: fwiw escaping should be very easy, especially since this is html 5: https://stackoverflow.com/questions/7381974/which-characters-need-to-be-escaped-in-html

sdelamo: > We cannot take strings supplied by the app programmer and output them unescaped.

If a malicious programmer returns a malicious HTML in his controller method, we don't escape that. 

I just don't see how someone could leverage custom status as an attack unless they have complete control of writing application code. At that point, I don't think we can do anything. If he wants to write a malicious output, we could do that in a million ways.  

I will try to create an API for HTML sanitisation. But I don't see how to leverage custom status code as an attack.

yawkat: JSON analogy: If a programmer has a controller like `@Get String foo() { return "\"foo\""; }`, then no, we don't escape the JSON. If they define a controller that returns a structure *containing* a string though, like `@Get MyBean foo() { return new MyBean("\"foo\""); }`, then the JSON contains the string escaped, i.e. `{"field":"\"foo\""}`.

It's a basic principle used to avoid injection vulnerabilities. "Error" is a complex structure containing a string, not a string that is explicitly meant as raw HTML, so it must be escaped.

http/src/main/java/io/micronaut/http/MediaTypeUtils.java
sdelamo: Let me know if this is ok to you https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/6d63118a568bd75f7485640abaca083fafa7a81f

http-server/src/main/java/io/micronaut/http/server/exceptions/response/DefaultHtmlErrorResponseBodyProvider.java
sdelamo: I have sanitized error and status https://github.com/micronaut-projects/micronaut-core/pull/11210/commits/3f95b0cb0d3dababfa22c13606450363037f3fc4

function/src/main/java/io/micronaut/function/executor/FunctionInitializer.java
```diff
@@ -96,7 +95,6 @@ public void close() {
      *
      * @param args The arguments passed to main
      * @param supplier The function that executes this function
-     * @throws IOException If an error occurs
```
sdelamo, line 99: Otherwise, Javadoc complains of a throw javadoc for an exception not being thrown.

dstepanov (approved):

yawkat (approved):

graemerocher (approved):

Issue #11137, opened 2024-08-30
Add StringConvention param to environment.getProperty() so we can disable kebab translation


jdeex:
### Feature description

There is an option in Environment to getProperties() with a StringConvention.RAW to avoid the kebab translation, but this return each element as an independent entry

It would be useful to allow this same behaviour to getProperty() so you can control how the configuration is returned: as kebab or RAW

graemerocher:
Triage: valid Core API enhancement. Add a `StringConvention` overload for `getProperty`, preferably on the conversion-context path and with convenience overloads as appropriate, so RAW lookup bypasses kebab normalization while preserving existing precedence, placeholder resolution, conversion, and default-value behavior. Add tests covering mixed-case/dashed keys, nested paths, typed conversion, and the existing kebab-default behavior.

PR #11216, opened 2024-09-29
Add ability to disable normalization for property keys in EachProperty

altro3:
In my case I faced such a problem: I created a hierarchical structure for configuring several databases, each of which had settings for several tables. I used the `EachProperty` annotation and did not expect that the value injected into the Parameter variable would not be the one I set, but already normalized. Thus, the names of the databases and tables were converted, although I did not need this.

I realized that there is currently no way to disable normalization for such a case. This PR adds the `normalizePropertyKeys` flag to disable automatic key normalization for beans created using EachProperty

### DIFF
--- a/core/src/main/java/io/micronaut/core/value/PropertyCatalog.java
+++ b/core/src/main/java/io/micronaut/core/value/PropertyCatalog.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2017-2024 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.value;
+
+/**
+ * The property catalog to use.
+ *
+ * @since 4.7.0
+ */
+public enum PropertyCatalog {
+    /**
+     * The catalog that contains the raw keys.
+     */
+    RAW,
+    /**
+     * The catalog that contains normalized keys. A key is normalized into
+     * lower case hyphen separated form. For example an environment variable {@code FOO_BAR} would be
+     * normalized to {@code foo.bar}.
+     */
+    NORMALIZED,
+    /**
+     * The catalog that contains normalized keys and also generated keys. A synthetic key can be generated from
+     * an environment variable such as {@code FOO_BAR_BAZ} which will produce the following keys: {@code foo.bar.baz},
+     * {@code foo.bar-baz}, and {@code foo-bar.baz}.
+     */
+    GENERATED
+}
--- a/core/src/main/java/io/micronaut/core/value/PropertyResolver.java
+++ b/core/src/main/java/io/micronaut/core/value/PropertyResolver.java
@@ -90,6 +90,29 @@ public interface PropertyResolver extends ValueResolver<String> {
     }
 
     /**
+     * Returns a collection of properties entries under the given key, but .
+     * For example, if you set {@code PropertyCatalog.RAW} then the following keys:
+     *
+     * <pre>
+     * <code>datasource.MyDs-1.url=localhost
+     * datasource.MyDs-2.url=someother</code>
+     * </pre>
+     *
+     * Calling {@code getPropertyEntries(String)} with a value of {@code datasource} will result in a collection
+     * containing {@code MyDs-1} and {@code MyDs-2} (without normalization).
+     *
+     * @param name The name to resolve
+     * @param propertyCatalog property catalog to use
+     * @return The property entries.
+     *
+     * @since 4.7.0
+     */
+    @NonNull
+    default Collection<String> getPropertyEntries(@NonNull String name, @NonNull PropertyCatalog propertyCatalog) {
+        return Collections.emptySet();
+    }
+
+    /**
      * <p>Resolve the given property for the given name, type and generic type arguments.</p>
      *
      * <p>Implementers can choose to implement more intelligent type conversion by analyzing the typeArgument.</p>
--- a/inject/src/main/java/io/micronaut/context/env/ConfigurationPath.java
+++ b/inject/src/main/java/io/micronaut/context/env/ConfigurationPath.java
@@ -20,6 +20,7 @@ import io.micronaut.context.annotation.ConfigurationReader;
 import io.micronaut.context.annotation.EachProperty;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.value.PropertyCatalog;
 import io.micronaut.core.value.PropertyResolver;
 import io.micronaut.inject.BeanDefinition;
 import io.micronaut.inject.qualifiers.Qualifiers;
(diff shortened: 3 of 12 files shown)


altro3:
@dstepanov @graemerocher look to this, pls

dstepanov:
I think the correct way would be to have the catalog option to the EachProperty annotation and use it during the fetching of the properties

sdelamo:
> In my case I faced such a problem: I created a hierarchical structure for configuring several databases, each of which had settings for several tables. I used the `EachProperty` annotation and did not expect that the value injected into the Parameter variable would not be the one I set, but already normalized. Thus, the names of the databases and tables were converted, although I did not need this.
> 
> I realized that there is currently no way to disable normalization for such a case. This PR adds the `normalizePropertyKeys` flag to disable automatic key normalization for beans created using EachProperty

I don't not think you should used the bean name, the @EachProperty @Parameter variable, as the database name.  You should create a different property in your configuration object. 

```java
@EachProperty("app.tables")
public class TableProps {

    private String beanName;
    private String name;
    private String url;
    private int serverPort;

    public TableProps(@Parameter String beanName) {
        this.name = name;
    }
```

altro3:
Yes, of course you can duplicate the name in a separate field, as you suggest, but when you have more than 30 databases with 2-3 tables, you try to optimize the configs and (most importantly!) avoid duplicating values, because it is very difficult to find an error in such a config later.

In my opinion, normalization is good. BUT! It should not be forced, i.e. there should always be a way to use variable names or keys for the map generated by eachProperty "as is".

Recently there was a request to do something similar, regarding other methods: https://github.com/micronaut-projects/micronaut-core/issues/11137

sdelamo:
> Yes, of course you can duplicate the name in a separate field, as you suggest, but when you have more than 30 databases with 2-3 tables, you try to optimize the configs and (most importantly!) avoid duplicating values, because it is very difficult to find an error in such a config later.
> 
> In my opinion, normalization is good. BUT! It should not be forced, i.e. there should always be a way to use variable names or keys for the map generated by eachProperty "as is".
> 
> Recently there was a request to do something similar, regarding other methods: #11137

For me, the issue is not duplication. The issue is you are using the same value for different things, such as the bean name and database name, which is prone to errors.

altro3:
Everything is correct, one value for two variables. I am not embarrassed that the beans will have names not in the kebab case. The problem is that I cannot set the names of such beans either. That is, I believe that the best solution is to give the user the ability to manage behavior in this situation and not make strict rules for normalization.

It all comes from the fact that in this context, in fact, the name of the bean does not matter to me, because I manage the configuration not by the names of the beans, and the inject occurs as a map. And all I want is just for the micronaut not to modify my configuration, if this is a conscious decision on my part. Therefore, be that as it may, my opinion remains the same: a tool is needed to disable forced normalization.

I like Denis's @dstepanov  solution, I will try to remake this PR in the near future so that it is possible to disable normalization at the EachProperty annotation level

graemerocher:
I don't believe we can include this. Enabling this will break many things that depend on this normalisation. It might work for your isolated case but not generally and will just lead users down the wrong path.

graemerocher:
As an alternative if you must have this change is adding a `skipNormalizationForPrefixes` method that allows to specify a limited set of prefixes that you want to skip this normalisation for. That would be an acceptable change.

altro3:
As I said earlier, I liked Denis's suggestion. I implemented it. All that changed was the ability to specify a catalog in the EachProperty annotation. I think this solution is very good and allows you to disable normalization consciously and only for certain subsets of properties.

@dstepanov @sdelamo @graemerocher Please, review it

core/src/main/java/io/micronaut/core/value/PropertyCatalog.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.value;
+
+/**
+ * The property catalog to use.
+ */
+public enum PropertyCatalog {
```
graemerocher, line 23: add `@since`

core/src/main/java/io/micronaut/core/value/PropertyResolver.java
```diff
@@ -89,6 +89,29 @@ public interface PropertyResolver extends ValueResolver<String> {
...
+     * containing {@code MyDs-1} and {@code MyDs-2} (without normalization).
+     *
+     * @param name The name to resolve
+     * @param propertyCatalog property catalog to use
+     * @return The property entries.
+     *
+     * @since 4.7.0
+     */
+    @NonNull
+    default Collection<String> getPropertyEntries(@NonNull String name, PropertyCatalog propertyCatalog) {
```
graemerocher: add nullability annotation to second parameter

inject/src/main/java/io/micronaut/context/env/ConfigurationPath.java
```diff
@@ -120,6 +121,15 @@ static ConfigurationPath of(BeanDefinition<?>... definitions) {
      */
     int index();
 
+    /**
+     * @return the current property catalog
+     *
+     * @since 4.7.0
+     */
+    default PropertyCatalog propertyCatalog() {
```
graemerocher, line 130: add nullability annotation to return type

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -835,29 +842,6 @@ public void close() throws Exception {
         }
     }
 
-
-    /**
-     * The property catalog to use.
-     */
-    protected enum PropertyCatalog {
```
graemerocher, line 842: this is a breaking change for classes that subclass this class. Deprecate using `forRemoval` instead.

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/value/PropertyResolver.java
altro3: fixed

inject/src/main/java/io/micronaut/context/env/ConfigurationPath.java
altro3: fixed

core/src/main/java/io/micronaut/core/value/PropertyCatalog.java
altro3: fixed

inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
altro3: Why it's breaking change? This enum was internal, not public api. I just moved it to public API.

graemerocher: it doesn't appear to be annotated with `@Internal`

altro3: It is not annotated, but it has a scope of protected and cannot be used from outside.

graemerocher: This code suggests otherwise:

```java
package example;

import io.micronaut.context.env.PropertySourcePropertyResolver;

public class TestResolver
    extends PropertySourcePropertyResolver {
    public TestResolver() {
        super();
        PropertyCatalog[] values = PropertyCatalog.values();
        System.out.println("values = " + values);
    }
}
```

altro3: It turns out that I can't use PropertyCatalog in an annotation and I need to create a second PropertyCatalog enum that will be used specifically in the annotation. And do mapping between them. Right?

graemerocher: or duplicate them. We follow semantic versioning which prohibits breaking changes in a minor release so we cannot merge changes that may break user code.

altro3: Okay, I did everything, now the changes will be invisible

altro3: {D7CF0F89-FD03-45D4-BBE3-EE9B393BFEFF}

graemerocher (approved):

sdelamo (approved):

PR #11225, opened 2024-10-02
Correct routing check

dstepanov:
Just a tiny correct for the route check, which should use the matched collection


sdelamo (commented):
Can you add a test?

dstepanov:
No, the change is only in the performance

sdelamo (approved):

graemerocher (approved):

PR #11226, opened 2024-10-02
Make `MediaType` comparable

dstepanov:
We can avoid creating a lambda to do sorting


sdelamo (approved):

http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -27,6 +27,7 @@
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.core.value.OptionalValues;
 import io.micronaut.http.annotation.Produces;
+import org.jetbrains.annotations.NotNull;
```
graemerocher: use the Micronaut version

graemerocher (changes requested):

graemerocher (approved):

PR #11227, opened 2024-10-02
RawHttpClient

yawkat:
Add a new HTTP client variant for making raw requests. The main purpose of this is to allow implementing the micronaut-oraclecloud HTTP client without direct access to ConnectionManager.

This client is a fairly thin wrapper around #11182. It gives direct access to the request/response body bytes, but still runs filters and follows redirects.

The fact that filters are run and redirects are followed creates some limitations for this new client:

- If a filter replaces the response, it is possible that the client returns a non-ByteBody response. This means that the client has to return `Publisher<? extends HttpResponse<?>>` instead of `Publisher<? extends ByteBodyHttpResponse<?>>`, even though `ByteBodyHttpResponse` is almost always what users of this API want. Nonetheless I've decided to leave handling this up to the user.
- Unfortunately it is *not* possible to implement the other HttpClients using this RawHttpClient. The reason is that filters for other clients see the "bean" form of the body, not the bytes. As a consequence, my medium-term goal of sharing more code between the JDK and netty HTTP clients will have to use an abstract class with an even lower level API than what is introduced in this PR.

Nonetheless, I think it is better to follow redirects and run filters. It would be odd to have a HttpClient that completely ignores the configured follow-redirects behavior, and I think it would also be unexpected for users to have a (non-internal) client that does not run client filters – we already had problems with this in micronaut-oraclecloud.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/RawHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/RawHttpClient.java
@@ -1,0 +1,69 @@
+package io.micronaut.http.client;
+
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.async.annotation.SingleResult;
+import io.micronaut.http.ByteBodyHttpResponse;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.body.ByteBody;
+import io.micronaut.http.body.CloseableByteBody;
+import org.reactivestreams.Publisher;
+
+import java.io.Closeable;
+import java.net.URI;
+
+/**
+ * HTTP client that allows sending "raw" requests with a {@link ByteBody} for the request and
+ * response body.
+ *
+ * @author Jonas Konrad
+ * @since 4.7.0
+ */
+@Experimental
+public interface RawHttpClient extends Closeable {
+    /**
+     * Send a raw request.
+     *
+     * @param request       The request metadata (method, URI, headers). The
+     *                      {@link HttpRequest#getBody() body} of this object is ignored
+     * @param requestBody   The request body bytes. {@code null} is equivalent to an empty body.
+     *                      The ownership of the body immediately transfers to the client, i.e. the
+     *                      client will always call {@link CloseableByteBody#close()} on the body
+     *                      even if there is an error before the request is sent.
+     * @param blockedThread The thread that is blocked waiting for this request. This is used for
+     *                      deadlock detection. Optional parameter.
+     * @return A mono that will contain the response to this request. This response will
+     * <i>usually</i> be a {@link ByteBodyHttpResponse}, unless a filter replaced it.
+     */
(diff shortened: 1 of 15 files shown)


http-client-core/src/main/java/io/micronaut/http/client/RawHttpClient.java
```diff
@@ -0,0 +1,84 @@
...
+     *                      client will always call {@link CloseableByteBody#close()} on the body
+     *                      even if there is an error before the request is sent.
+     * @param blockedThread The thread that is blocked waiting for this request. This is used for
+     *                      deadlock detection. Optional parameter.
+     * @return A mono that will contain the response to this request. This response will
+     * <i>usually</i> be a {@link ByteBodyHttpResponse}, unless a filter replaced it.
+     */
+    @NonNull
+    @SingleResult
+    Publisher<? extends HttpResponse<?>> exchange(@NonNull HttpRequest<?> request, @Nullable CloseableByteBody requestBody, @Nullable Thread blockedThread);
```
dstepanov, line 57: Can we return `CompletableFuture` instead of `Publisher`?

yawkat: Yes it's possible, that was my initial version, but I want to keep it close to HttpClient

yawkat: Hmm also not sure if cancellation works well with CompletableFuture.

http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
```diff
@@ -54,12 +54,6 @@ public static BlockHint combine(@Nullable BlockHint a, @Nullable BlockHint b) {
         }
     }
 
-    void checkIsNotBlocked(EventLoop eventLoop) {
```
sdelamo, line 57: why are we removing this in this PR?

sdelamo (changes requested):
> but still runs filters and follows redirects.

Are there any tests which verify that the raw client runs filters and follow redirects?
If they are not, please add them.

http-client/src/main/java/io/micronaut/http/client/netty/BlockHint.java
yawkat: it was unused

sdelamo (approved):

http-client-core/src/main/java/io/micronaut/http/client/RawHttpClient.java
graemerocher, line 57: why is the signature not `HttpRequest<CloseableByteBody> request`?

graemerocher (approved):

http-client-core/src/main/java/io/micronaut/http/client/RawHttpClient.java
yawkat: Usually it would be `ByteBodyHttpResponse` (mentioned in the javadoc). But a filter can replace the response with anything it likes, including non-ByteBody body types, so I want to force the caller to handle the case where there's no ByteBody

PR #11229, opened 2024-10-02
Fix for service loader

andriy-dmytruk:
Fix for service loader, so that single-threaded `MicronautServiceCollector` implementation matches `ServiceInstanceLoader implementation`.

The `ServiceInstanceLoader` checks if predicate is null:
```java
@Override
        protected void compute() {
            try {
                result = instantiate(className, classLoader);
                if (result != null && predicate != null && !predicate.test(result)) {
                    result = null;
                }
            } catch (Throwable e) {
                throwable = e;
            }
        }
```
so `MicronautServiceCollector` must do that, too. I am getting null poineter error when I set `-Djava.util.concurrent.ForkJoinPool.common.parallelism=1` for Micronaut application.

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
@@ -257,7 +257,7 @@ public final class MicronautMetaServiceLoaderUtils {
                 List<S> collection = new ArrayList<>(serviceEntries.size());
                 for (String serviceEntry : serviceEntries) {
                     S val = instantiate(serviceEntry, classLoader);
-                    if (val != null && predicate.test(val)) {
+                    if (val != null && predicate != null && !predicate.test(val)) {
                         collection.add(val);
                     }
                 }


graemerocher (approved):

dstepanov (approved):

tbashour:
@andriy-dmytruk @graemerocher @dstepanov 

I believe the change is not complete and still has a logical error. 
Another way to change the parallelisim would be to set -XX:ActiveProcessorCount. This option is used in dockerized environments to limit resources used by the application.

In my tests, wenn I set the option a value smaller than 3 (parallelism is active processor count -1), then micronaut application will not start.

I believe the condition to add the services in the class MicronautMetaServiceLoaderUtils (line 260) is not correct:

is:

`if (val != null && predicate != null && !predicate.test(val)) {`

but should be:

`if (val != null && !collection.contains(val) && (predicate == null || predicate.test(val))) {`

graemerocher:
@tbashour could you send a PR?

tbashour:
@graemerocher 
https://github.com/micronaut-projects/micronaut-core/pull/11326

PR #11239, opened 2024-10-07
KSP: Properly update current round resolver in all helper classes

dstepanov:
Not sure how to properly test it


graemerocher:
Seems to be failing compilation

graemerocher (approved):

PR #11256, opened 2024-10-16
Modified getTrustManagerFactory to not load trust store twice

msupic:


yawkat (approved):

yawkat:
Thanks! Need to fix the build though, I think graal 1.2.3 is broken. please merge 4.7.x into this PR to try 1.2.4

PR #11259, opened 2024-10-17
JDK RawHttpClient

yawkat:
This PR adds a RawHttpClient implementation in the JDK client. In principle it's not so complicated, but to actually be useful, it needed a streaming reactive ByteBody implementation, which is complicated to implement.

To implement this, I've refactored the StreamingNettyByteBody implementation into a "generic" implementation that is implemented both for the netty ByteBuf/EventLoop approach, and a more traditional ByteBuffer/lock approach for the jdk client.

- BodySizeLimits, PublisherAsBlocking, UpstreamBalancer, BufferConsumer.Upstream moved from http-netty into http
- BufferConsumer refactored into a base BufferConsumer interface (http) and a ByteBufConsumer extension (http-netty)
- Most StreamingNettyByteBody.SharedBuffer code (http-netty) extracted into an abstract BaseSharedBuffer (http) that is also extended for the JDK client

In particular the SharedBuffer refactor is fairly complicated. While the general code remains unchanged, making it work for different buffer types and concurrency models is still a bit complicated.


http-client-jdk/src/main/java/io/micronaut/http/client/jdk/DefaultJdkHttpClientRegistry.java
```diff
@@ -152,6 +155,28 @@ protected DefaultJdkHttpClient httpClient(
...
+     * @param loadBalancer
+     * @param configuration
+     * @param beanContext
+     * @return A {@literal java.net.http.*} HTTP Client
+     */
+    @Bean
+    @BootstrapContextCompatible
+    @Primary
+    @Order(2) // If both this and the netty client are present, netty is the default.
+    JdkRawHttpClient rawHttpClient(
```
graemerocher: perhaps return only the interface and not the concrete type

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/JdkRawHttpClient.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    protected <O> Publisher<HttpResponse<O>> responsePublisher(@NonNull HttpRequest<?> request, @Nullable Argument<O> bodyType) {
+        return Mono.defer(() -> mapToHttpRequest(request, bodyType)) // defered so any client filter changes are used
+            .map(httpRequest -> {
+                if (log.isDebugEnabled()) {
+                    log.debug("Client {} Sending HTTP Request: {}", clientId, httpRequest);
+                }
+                HttpHeadersUtil.trace(log,
```
graemerocher: can we add a `isTraceEnabled()` check to avoid the allocations of the various lambdas

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/JdkRawHttpClient.java
```diff
@@ -0,0 +1,87 @@
...
+                    () -> httpRequest.headers().map().keySet(),
+                    headerName -> httpRequest.headers().allValues(headerName));
+                return client.sendAsync(httpRequest, responseInfo -> new ReactiveByteBufferByteBody.ByteBodySubscriber(new BodySizeLimits(Long.MAX_VALUE, configuration.getMaxContentLength())));
```
graemerocher: split this onto multiple lines to make it more readable

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/ReactiveByteBufferByteBody.java
```diff
@@ -0,0 +1,508 @@
...
+        }
+        upstream.allowDiscard();
+        return this;
+    }
+
+    interface ByteBufferConsumer extends BufferConsumer {
+        void add(ByteBuffer buffer);
+    }
+
+    private static class AsFlux extends BaseSharedBuffer.AsFlux<ByteBuffer> implements ByteBufferConsumer {
```
graemerocher: declare `final`

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/ReactiveByteBufferByteBody.java
```diff
@@ -0,0 +1,508 @@
...
+                if (last) {
+                    buffer = null;
+                }
+                return snapshot;
+            }
+        }
+
+        @Override
+        protected void addDoNotBuffer() {
+
```
graemerocher: add a comment here why this is not implemented

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/ReactiveByteBufferByteBody.java
```diff
@@ -0,0 +1,508 @@
...
+        @Override
+        public CompletionStage<CloseableByteBody> getBody() {
+            return CompletableFuture.completedFuture(root);
+        }
+
+        @Override
+        public void onSubscribe(Flow.Subscription subscription) {
+            boolean initialDemand;
+            boolean cancelled;
+            synchronized (this) {
```
graemerocher, line 428: can these synchronised blocks be replaced by locks?

http-netty/src/main/java/io/micronaut/http/netty/body/ByteBufConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * This is a netty-specific reactor-like API for streaming bytes. It's a bit better than reactor
+ * because it's more explicit about reference counting semantics, has more fine-grained controls
+ * for cancelling, and has more relaxed concurrency semantics.
+ *
+ * @since 4.5.0
+ * @author Jonas Konrad
+ */
+@Internal
+public interface ByteBufConsumer extends BufferConsumer {
```
graemerocher, line 32: can this be `sealed`?

http-netty/src/main/java/io/micronaut/http/netty/body/ByteBufConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Consume a buffer. Release ownership is transferred to this consumer.
+     *
+     * @param buf The buffer to consume
+     */
+    void add(ByteBuf buf);
```
graemerocher: nullability annotations

http/src/main/java/io/micronaut/http/MutableHttpRequestWrapper.java
```diff
@@ -38,20 +33,20 @@
...
+public class MutableHttpRequestWrapper<B> extends HttpRequestWrapper<B> implements MutableHttpRequest<B> {
     private ConversionService conversionService;
 
     @Nullable
     private B body;
     @Nullable
     private URI uri;
 
-    MutableHttpRequestWrapper(ConversionService conversionService, HttpRequest<B> delegate) {
+    public MutableHttpRequestWrapper(ConversionService conversionService, HttpRequest<B> delegate) {
```
graemerocher: does the constructor need to be public? The static method can be used no?

http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
```diff
@@ -0,0 +1,496 @@
...
+ * streaming {@link io.micronaut.http.body.ByteBody}s.<br>
+ * The subclass handles concurrency (for netty, event loop) and the specific buffer type
+ * (for netty, ByteBuf).
+ *
+ * @param <C> The {@link BufferConsumer} type for the specific buffer type
+ * @param <F> The type returned to {@link #subscribeFull0} subscribers. This is usually just the
+ *           buffer type
+ */
+@Internal
+public abstract class BaseSharedBuffer<C extends BufferConsumer, F> {
```
graemerocher, line 49: can it be sealed?

graemerocher (changes requested):
Please increase the code coverage and address the sonar warnings.

http-client-jdk/src/main/java/io/micronaut/http/client/jdk/ReactiveByteBufferByteBody.java
yawkat: it's not necessary because they only access fields, not call any other code

http-netty/src/main/java/io/micronaut/http/netty/body/ByteBufConsumer.java
yawkat: No, there's implementations in tests for example

http/src/main/java/io/micronaut/http/MutableHttpRequestWrapper.java
yawkat: There's some subclasses that extend it

yawkat: Made it protected

http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
yawkat: No, the subclasses are in other modules

yawkat:
I think the sonar coverage stats are not including the TCK

graemerocher (approved):

dstepanov (approved):

PR #11267, opened 2024-10-22
Fix controller `@Body Object` parameters

### LINKED ISSUE #11266: @Body annotation causes "Request body has already been claimed" error
### Expected Behavior

Ideally the body parameter should be available for use. This was working in micronaut-starter version `4.5.1`.

### Actual Behaviour

In some scenarios, using a @Body parameter in a controller causes a "Request body has already been claimed" error.

### Steps To Reproduce

Create a controller using micronaut-starter `4.6.3` like this:
```java
import io.micronaut.http.HttpRequest;
import io.micronaut.http.HttpResponse;
import io.micronaut.http.annotation.Body;
import io.micronaut.http.annotation.PathVariable;
import io.micronaut.http.annotation.Post;
import io.micronaut.json.tree.JsonNode;
import reactor.core.publisher.Mono;

@Controller
public class PostController {

    @Post("/{+path}")
    public Mono<HttpResponse<JsonNode>> doPost(@PathVariable String path,
                                               @Nullable @Body Object body,
                                               HttpRequest<?> request) {
        return Mono.just(HttpResponse.ok());
    }
}
```

Create this unit test:

```java

import io.micronaut.core.type.Argument;
import io.micronaut.http.HttpMethod;
import io.micronaut.http.HttpRequest;
import io.micronaut.http.HttpResponse;
import io.micronaut.http.HttpStatus;
import io.micronaut.http.client.annotation.Client;
import io.micronaut.reactor.http.client.ReactorHttpClient;
import io.micronaut.test.extensions.junit5.annotation.MicronautTest;
import jakarta.inject.Inject;
import org.junit.jupiter.api.Test;

import java.util.Map;

import static org.junit.jupiter.api.Assertions.*;

@MicronautTest
class PostControllerTest {

    @Inject
    @Client("/")
    ReactorHttpClient client;

    @Test
    void ok_request_with_body() {
        var post = HttpRequest.POST( "/do-post", Map.of("key1", "val1"));

        HttpResponse<Map<String, Object>> result = client.toBlocking().exchange(post, Argument.mapOf(String.class, Object.class));
        assertEquals(HttpStatus.OK, result.getStatus());
    }
}

```

Running this test will give this error:
```
12:42:04.980 [default-nioEventLoopGroup-1-3] ERROR i.m.http.server.RouteExecutor - Unexpected error occurred: Request body has already been claimed: Two conflicting sites are trying to access the request body. If this is intentional, the first user must ByteBody#split the body. To find out where the body was claimed, 
```

### Environment Information

JDK version 21.
OS version: Mac os

### Example Application

_No response_

### Version

4.6.3

yawkat:
There's two bugs here.

Let's consider the following cases:

1. `application/x-www-form-urlencoded`, `@Body Object` param (tested by the old FormDataDiskSpec)
2. `application/json`, `@Body Object` param (prev. no test, tested by a new TCK test)
3. `application/x-weird`, `@Body Object` param (prev. no test, tested by a new netty-specific test)
4. `application/x-weird`, `@Body MyRecord` param (prev. no test, tested by a new netty-specific test)

Prior to 4.6, case 2 has a registered "normal" reader (the JSON reader) that can handle this case fine, the other three cases do not have readers and rely on fallback conversion logic. Forms have some special handling here.

In 4.6, the way raw type readers (e.g. that for String) were resolved changed. This means that for `Object` params, all raw type readers were now eligible, since they can read a subtype of Object (e.g. String). For cases 1, 3, and 4, this is a breaking change, since the reading is now done by a (random) raw type reader, instead of falling back to conversion logic.

This change broke the FormDataDiskSpec (case 1). To fix this, 1d09d95f033fa6270b0b187c28183e63c250ae26 introduced a check that would blanket-ignore the reader for all Object-typed params, falling back to the conversion logic. This fixed cases 1 and 3, but inadvertently broke case 2, since the JSON reader was also ignored.

Case 2 is apparently common enough that there have been multiple reports about it. It also exposed the second bug, a double-claim of the body, which is a fairly simple fix. But it shows that the fallback conversion logic had no test coverage, since this bug should trigger reliably. That is what `BodyConversionSpec` is there to improve.

My proposed fix is to make the raw message readers not apply to Object parameters, and to remove the band-aid fix from 1d09d95f033fa6270b0b187c28183e63c250ae26 that broke case 2.

Fixes #11266

### DIFF
--- a/http/src/main/java/io/micronaut/http/body/TypedMessageBodyReader.java
+++ b/http/src/main/java/io/micronaut/http/body/TypedMessageBodyReader.java
@@ -38,6 +38,6 @@ public interface TypedMessageBodyReader<T> extends MessageBodyReader<T> {
 
     @Override
     default boolean isReadable(Argument<T> type, MediaType mediaType) {
-        return type.isAssignableFrom(getType());
+        return type.isAssignableFrom(getType()) && !type.getType().equals(Object.class);
     }
 }
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -45,7 +45,7 @@ jobs:
           fetch-depth: 0
 
       - name: "🔧 Setup GraalVM CE"
-        uses: graalvm/setup-graalvm@v1.2.3
+        uses: graalvm/setup-graalvm@v1.2.4
         with:
           distribution: 'graalvm'
           java-version: ${{ matrix.java }}
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
@@ -157,10 +157,6 @@ final class NettyBodyAnnotationBinder<T> extends DefaultBodyAnnotationBinder<T> 
         if (mediaType != null && (reader == null || !reader.isReadable(context.getArgument(), mediaType))) {
             reader = bodyHandlerRegistry.findReader(context.getArgument(), List.of(mediaType)).orElse(null);
         }
-        if (reader != null && context.getArgument().getType().equals(Object.class)) {
-            // Prevent random object convertors
-            reader = null;
-        }
         if (reader == null && nhr.isFormOrMultipartData()) {
             FormDataHttpContentProcessor processor = new FormDataHttpContentProcessor(nhr, httpServerConfiguration);
             ByteBuf buf = AvailableNettyByteBody.toByteBuf(imm);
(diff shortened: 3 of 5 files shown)


graemerocher (approved):

sdelamo:
@yawkat can you change to graalvm 1.2.4 in your pr so the the build passes and we can merge it?

http/src/main/java/io/micronaut/http/body/TypedMessageBodyReader.java
```diff
@@ -38,6 +38,6 @@ public interface TypedMessageBodyReader<T> extends MessageBodyReader<T> {
 
     @Override
     default boolean isReadable(Argument<T> type, MediaType mediaType) {
-        return type.isAssignableFrom(getType());
+        return type.isAssignableFrom(getType()) && !type.getType().equals(Object.class);
```
dstepanov, line 41: I don't think this a correct fix, typed reader should only add a type and nothing else

yawkat: I'm open to suggestions. The new test cases cover the issue pretty well.

dstepanov:
Let's keep it this way. We shouldn't allow Object for body in the future

dstepanov (approved):

graemerocher:
@sdelamo @yawkat we need to merge up 4.6.x in 4.7.x

Issue #10889, opened 2024-06-05
Instantiation methods in introspection of abstract class tries to instantiate the abstract class


rorueda:
### Expected Behavior

Introspection for an abstract class doesn't override the instantiation methods trying to instantiate the abstract class.

### Actual Behaviour

For Java and Kotlin, the introspection for abstract classes is overriding the instantiation methods with a call to the abstract class constructor. This results in a `j.l.InstantiationError` instead of a `i.m.c.r.e.InstantiationException`. 

This is important in my use case because only the latter is caught by micronaut serde and then wrapped in a `SerdeException`.

Note that it already works for groovy classes.

### Steps To Reproduce

See the changes in the tests at https://github.com/micronaut-projects/micronaut-core/compare/4.5.x...rorueda:issue-abstract-class-introspection

Note it also has a fix inspired by the groovy implementation (making the `ClassElement` ignore the constructors if the class is abstract).

### Version

core 4.5.x

graemerocher:
The current BeanIntrospectionWriter still generates instantiate methods when an abstract Java/Kotlin class has a constructor, so the issue is not resolved in the current source. Earlier fixes #11268 and #12540 are closed rather than merged; no active PR currently addresses it. The fix should mark abstract/interface introspections non-buildable and omit direct constructor instantiation while retaining valid static @Creator paths, then add Java, Kotlin, and Groovy parity tests asserting the framework's expected InstantiationException instead of a JVM InstantiationError.

PR #11268, opened 2024-10-22
Fix introspection for abstract classes (#10889)

rorueda:
Ignore constructors for abstract classes, so the introspection is written as not buildable and the instantiation methods in the introspection are not overridden (trying to call a constructor for an abstract class).

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
@@ -269,6 +269,10 @@ public interface ClassElement extends TypedElement {
             // only static inner classes can be constructed
             return Optional.empty();
         }
+        if (isAbstract()) {
+            // abstract classes constructors are accessible only from their subtypes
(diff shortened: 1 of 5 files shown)


dstepanov:
Doesn't look like a correct fix. This should be fixed in `IntrospectedTypeElementVisitor`

graemerocher:
yes this fix is not in the correct place, though the tests add value

rorueda:
Thanks for the pointers, @dstepanov.  I'll look into it.

Should the abstract verification below be removed in favor of the fix on `IntrospectedTypeElementVisitor`? I was unsure of the semantic of `getDefaultConstructor`, and because of this, I ended up thinking it was to consider only constructors that can be directly called.

https://github.com/micronaut-projects/micronaut-core/blob/dfe9b0f94e7ce2dd15528893bf2d6dc211392a3d/inject-groovy/src/main/groovy/io/micronaut/ast/groovy/visitor/GroovyClassElement.java#L342-L349

PR #11270, opened 2024-10-23
javadoc: @RequestFilter Publisher<Optional<HttpResponse>>

sdelamo:


http/src/main/java/io/micronaut/http/annotation/RequestFilter.java
```diff
@@ -32,6 +32,7 @@
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+import java.util.Optional;
```
sdelamo:
```suggestion
```

yawkat:
I don't think it is appropriate to add a special comment about `Publisher<Optional<HttpResponse<?>>>`, here. It is merely a special case of the other allowed return value types (void, null, HttpRequest, HttpResponse) wrapped in a `Publisher`, which is already mentioned in the javadoc. However the fact that `Optional` can replace a `@Nullable T` is not mentioned and should be added

sdelamo:
> I don't think it is appropriate to add a special comment about `Publisher<Optional<HttpResponse<?>>>`, here. It is merely a special case of the other allowed return value types (void, null, HttpRequest, HttpResponse) wrapped in a `Publisher`, which is already mentioned in the javadoc. However the fact that `Optional` can replace a `@Nullable T` is not mentioned and should be added

the issue is that if you use `Publisher<HttpResponse<?>>` as the return type. [To proceed with the request neither return null or an empty publisher works](https://github.com/micronaut-projects/micronaut-security/pull/1826#discussion_r1814640674). You have to know the magic return type `Publisher<Optional<HttpResponse<?>>>`. This pull-request documents this to avoid people losing hours trying to understand how to proceed with a request if they need to have a Publisher return type.

yawkat:
`Publisher<HttpResponse<?>>` does not work because the `HttpResponse<?>` is not nullable. `Publisher<@Nullable HttpResponse<?>>` may work, not sure how the null-ness is propagated through Argument.

The reason I would prefer a more general note about `Optional` instead of pointing out the particular `Publisher<Optional<HttpResponse<?>>>` case is that it is also possible to make more complex combination of return types. `Publisher<Optional<HttpRequest<?>>>` is the obvious example, but `Publisher<Publisher<Optional<HttpResponse<?>>>>` is also allowed (though probably useless). We can't point out every combination in the javadoc, though we could point out *useful* combinations in the guide

sdelamo:
I am completely puzzled. 

my test using `Publisher<Optional<HttpResponse<?>>` fails to compile with: 

```
/Users/sdelamo/github/micronaut-projects/micronaut-core/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/RequestFilterPublisherProceedTest.java:60: error: Unsupported filter return type: java.util.Optional
        public Mono<Optional<HttpResponse<?>>> filter(@NonNull HttpRequest<?> request) {
```

However, a filter using such as method signature worked fine in the CSRF PR. https://github.com/micronaut-projects/micronaut-security/pull/1826/commits/aba5267a2e31ba97c5c788c966cc8356aeff7a2c

@yawkat @dstepanov do either of you see what am I missing? I did not see any recent changes in the `RequestFilter`  code to make it work on security but fail in core.

dstepanov:
Maybe the difference is that you had `Publisher` before, but now it's `Mono`? Anyway, using `Optional` in a publisher doesn't make sense.

sdelamo:
> Maybe the difference is that you had `Publisher` before, but now it's `Mono`? 

Fails to compile as well

https://github.com/micronaut-projects/micronaut-core/pull/11270/commits/c7381c1efb671b7863f64be49b9e41139e364087

> ? Anyway, using `Optional` in a publisher doesn't make sense.

Complete strange, I agree. But that [was the only way to use a Publisher with an HTTP response and be able to proceed](https://github.com/micronaut-projects/micronaut-security/pull/1826#discussion_r1814640674).

sdelamo:
Closing this, as I don't get it to compile.

PR #11272, opened 2024-10-23
feat: add FilterBodyParser API

sdelamo:
What do you think about this API as a way to get a form URL encoded as a map in the filter? It only supports netty, though. 

Context: https://github.com/micronaut-projects/micronaut-security/pull/1826#discussion_r1810840306

If you think this is the way to go, I will document it.


http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,76 @@
...
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Named(NettyFormUrlEncodedBodyParserInFilter.NAME_QUALIFIER)
+@Singleton
+public class NettyFormUrlEncodedBodyParserInFilter implements FilterBodyParser<Map<String, Object>> {
+    public static final String NAME_QUALIFIER =  MediaType.APPLICATION_FORM_URLENCODED;
```
sdelamo:
```suggestion
    public static final String NAME_QUALIFIER = MediaType.APPLICATION_FORM_URLENCODED;
```

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,76 @@
...
+        if (request.getContentType().isEmpty()) {
+            return Publishers.empty();
+        }
+        if (!request.getContentType().get().toString().equals(NAME_QUALIFIER)) {
+            return Publishers.empty();
+        }
+        if (request instanceof ServerHttpRequest<?> serverHttpRequest) {
+            return parseBody(serverHttpRequest);
+        }
+        return null;
```
sdelamo:
```suggestion
        return Publishers.empty();
```

yawkat:
It seems like a very generic API for a very specific use case. So I doubt it will be useful to many. But it's also not a big "investment" in terms of maintainability so I'm not against it

sdelamo:
> It seems like a very generic API for a very specific use case. So I doubt it will be useful to many. But it's also not a big "investment" in terms of maintainability so I'm not against it

accessing the request body in the security filters API is a demanded feature in Micronaut Security. see: https://github.com/micronaut-projects/micronaut-security/issues/689

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,80 @@
...
+import jakarta.inject.Singleton;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Named(NettyFormUrlEncodedBodyParserInFilter.NAME_QUALIFIER)
+@Singleton
+public class NettyFormUrlEncodedBodyParserInFilter implements FilterBodyParser<Map<String, Object>> {
```
graemerocher: shouldn't be public and should be `final`

yawkat:
@sdelamo accessing the request body is already possible with ServerHttpRequest. what is new here is the form thing

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,80 @@
...
+            return Mono.fromFuture(closeableByteBody.buffer())
+                    .map(bb -> bb.toString(request.getCharacterEncoding()))
+                    .map(NettyFormUrlEncodedBodyParserInFilter::parse);
+        }
+    }
+
+    private static Map<String, Object> parse(String formUrlEncoded) {
+        QueryStringDecoder decoder = new QueryStringDecoder(formUrlEncoded, false);
+        Map<String, List<String>> parameters = decoder.parameters();
+        Map<String, Object> result = new HashMap<>();
```
graemerocher: should be a linked hash map to retain order

graemerocher: also size the map the same size as the parameters

http/src/main/java/io/micronaut/http/filter/FilterBodyParser.java
```diff
@@ -0,0 +1,37 @@
...
+import reactor.util.annotation.NonNull;
+
+/**
+ * API to parse a body within a filter.
+ * @author Sergio del Amo
+ * @since 4.7.1
+ * @param <T> Body Type
+ */
+@FunctionalInterface
+public interface FilterBodyParser<T> {
```
graemerocher: add `@Experimental`

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
graemerocher: add javadoc to the class

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,80 @@
...
+@Named(NettyFormUrlEncodedBodyParserInFilter.NAME_QUALIFIER)
+@Singleton
+public class NettyFormUrlEncodedBodyParserInFilter implements FilterBodyParser<Map<String, Object>> {
+    public static final String NAME_QUALIFIER = MediaType.APPLICATION_FORM_URLENCODED;
+
+    @Override
+    @NonNull
+    @SingleResult
+    public Publisher<Map<String, Object>> parseBody(@NonNull HttpRequest<?> request) {
+        if (request.getContentType().isEmpty()) {
```
graemerocher: extract `getContentType` into a local variable to avoid the repeated calls

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,80 @@
...
+    private Mono<Map<String, Object>> parseBody(@NonNull ServerHttpRequest<?> request) {
+        try (CloseableByteBody closeableByteBody = request.byteBody().split(ByteBody.SplitBackpressureMode.FASTEST)) {
+            return Mono.fromFuture(closeableByteBody.buffer())
+                    .map(bb -> bb.toString(request.getCharacterEncoding()))
+                    .map(NettyFormUrlEncodedBodyParserInFilter::parse);
+        }
+    }
+
+    private static Map<String, Object> parse(String formUrlEncoded) {
+        QueryStringDecoder decoder = new QueryStringDecoder(formUrlEncoded, false);
```
graemerocher: you should pass the charset to the decoder and I would say a `maxParams` setting to avoid denial of service attacks. This should probably be configurable

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,80 @@
...
+        Map<String, List<String>> parameters = decoder.parameters();
+        Map<String, Object> result = new HashMap<>();
+        for (Map.Entry<String, List<String>> entry : parameters.entrySet()) {
+            if (entry.getValue().size() > 1) {
+                result.put(entry.getKey(), entry.getValue());
```
graemerocher: extract `getKey()` and `getValue()` to avoid repeated calls. Consider using `parameters.forEach(key, value)` which avoids allocating an entry

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/filter/FilterBodyParser.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/72fcd2e13d4d0c69972edb33f89b2cf9a76059e4

sdelamo:
> @sdelamo accessing the request body is already possible with ServerHttpRequest. what is new here is the form thing

Yup, but the current api is not really user-friendly. The currently example we have in the docs shows how to get a byte array. 

```java
 @RequestFilter
    public void logBody(ServerHttpRequest<?> request) { // (2)
        try (CloseableByteBody ourCopy = // (4)
                 request.byteBody()
                     .split(ByteBody.SplitBackpressureMode.SLOWEST) // (3)
                     .allowDiscard()) { // (5)
            Flux.from(ourCopy.toByteArrayPublisher()) // (6)
                .onErrorComplete(ByteBody.BodyDiscardedException.class) // (7)
                .subscribe(array -> LOG.info("Received body: {}", Base64.getEncoder().encodeToString(array))); // (8)
        }
    }
```
However, we ask users to do a lot of low-level operations. In my opinion, users should not need to write `CloseableByteBody ourCopy = request.byteBody()split(ByteBody.SplitBackpressureMode.SLOWEST).allowDiscard())`. That looks quite complicated to me.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: done via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/7314c570898158494ea9aa1197fc3598d21588f1

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/c220b3052c3bd78a8d318974cc317236056c39b0

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/94e715554ef63e35c50d64dbd2578be2b539bf7f

sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/4a97e581ad42ff0121dda7fdd30f183f2dc6ec0e

yawkat:
I'm open to adding new helper classes to simplify common access operations (e.g. `request.split(FASTEST).buffer().thenApply(AvailableByteBody::toByteArray)`), but this API is way too generic for that purpose

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/a866ab9da6e4218ac87ee82cb04252f4b0dcc30f

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/dd6ef4bd50fe415e8a85d777a95b0b6c65f1432b

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/e5f3d3ae62676f15a212a9ec270e0482e5dc0c68

sdelamo:
> I'm open to adding new helper classes to simplify common access operations (e.g. `request.split(FASTEST).buffer().thenApply(AvailableByteBody::toByteArray)`), but this API is way too generic for that purpose

@yawkat @graemerocher can you please review again. 

I have added a `JsonFilterBodyParser` and a `FormUrlEncodedFilterBodyParser` subclasses. These APIs allow users to bind a form-url-encoded payload to map or a JSON payload to a POJO within a filter. 

It is good to have generic APIs outside the `ServerHttpRequest` as we don't support `ServerHttpRequest` outside of Netty. It would be great to add `JsonFilterBodyParser` and a `FormUrlEncodedFilterBodyParser` implementations in servlet-core. Otherwise, if we release the CSRF mitigations APIs, we cannot read a CSRF token submitted in a hidden form field in a non-netty micronaut application. .

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
```diff
@@ -0,0 +1,91 @@
...
+@Singleton
+@Experimental
+final class NettyFormUrlEncodedBodyParserInFilter implements FormUrlEncodedFilterBodyParser {
+    private final Integer maxParams;
+
+    /**
+     *
+     * @param maxParams Max Parameters passed to {@link QueryStringDecoder} to avoid denials of service attack.
+     */
+    NettyFormUrlEncodedBodyParserInFilter(@Value("${micronaut.http.server.filter.body-parser.form-url-encoded.max-params:20}") Integer maxParams) {
```
graemerocher: never use `@Value` and enhance one of the existing `HttpServerConfiguration` class for this.

http-server/src/main/java/io/micronaut/http/server/util/NettyJsonFilterBodyParser.java
```diff
@@ -0,0 +1,78 @@
...
+        try (CloseableByteBody closeableByteBody = request.byteBody().split(ByteBody.SplitBackpressureMode.FASTEST)) {
+            return Mono.fromFuture(closeableByteBody.buffer())
+                    .mapNotNull(bb -> {
+                        try {
+                            return jsonMapper.readValue(bb.toByteArray(), type);
+                        } catch (IOException e) {
+                            if (LOG.isErrorEnabled()) {
+                                LOG.error("could not bind JSON body to {}", type);
+                            }
+                            return null;
```
graemerocher: I don't think we should be swallowing JSON parsing exceptions and logging them as errors. If the request is invalid it should be a bad request

http/src/main/java/io/micronaut/http/filter/bodyparser/FilterBodyParser.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Experimental
+public interface FilterBodyParser<T> {
+    /**
+     * @param request HTTP Request
+     * @param type The type to parse the body into
+     * @return a publisher which emits a single item or an empty publisher if the request body cannot be parsed to the requested type.
+     */
+    @SingleResult
+    Publisher<T> parseBody(@NonNull HttpRequest<?> request, @NonNull Class<T> type);
```
graemerocher: why does this need a `type`? Surely you want `Map<String, Object>` for both JSON and form encoding?

http/src/main/java/io/micronaut/http/filter/bodyparser/JsonFilterBodyParser.java
```diff
@@ -0,0 +1,33 @@
...
+import io.micronaut.http.MediaType;
+
+/**
+ * Implementation of {@link FilterBodyParser} for a request with JSON content type.
+ * @author Sergio del Amo
+ * @since 4.7.1
+ * @param <T> Body type
+ */
+@Experimental
+public interface JsonFilterBodyParser<T> extends FilterBodyParser<T> {
```
graemerocher: not sure I am can on the proliferation of interfaces and classes in this PR. It should be just inject `@Named("application/json") FilterBodyParser` without having to add a bazillion interfaces

http-server/src/main/java/io/micronaut/http/server/util/NettyJsonFilterBodyParser.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * {@link JsonFilterBodyParser} implementation which leverages {@link ServerHttpRequest#byteBody()} API and a  {@link JsonMapper}.
+ * @param <T> Body type
+ */
+@Singleton
+@Requires(bean = JsonMapper.class)
+@Requires(missingBeans = JsonFilterBodyParser.class)
+@Experimental
+final class NettyJsonFilterBodyParser<T> implements JsonFilterBodyParser<T> {
```
graemerocher: there doesn't seem to be anything Netty specific about this code

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/filter/bodyparser/FilterBodyParser.java
sdelamo: do we want the `FilterBodyParser`s to bind to Map?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyFormUrlEncodedBodyParserInFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/aef9cd8800cd3cc649c0250a4d249fe7b5ebc233

http-server/src/main/java/io/micronaut/http/server/util/NettyJsonFilterBodyParser.java
sdelamo: I changed this to https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/7cdfd5db1058d7c09e928e0423726b490d951e8c

```java
.flatMap(bb -> {
                            try {
                                return Mono.just(jsonMapper.readValue(bb.toByteArray(), Argument.mapOf(String.class, Object.class)));
                            } catch (IOException e) {
                                if (LOG.isErrorEnabled()) {
                                    LOG.error("could not bind JSON body to Map");
                                }
                                return Mono.error(e);
                            }
                        });
```

http/src/main/java/io/micronaut/http/filter/bodyparser/FilterBodyParser.java
sdelamo: I have simplified the API to be just: 

```java
@FunctionalInterface
public interface FilterBodyParser {
    @SingleResult
    Publisher<Map<String, Object>> parseBody(@NonNull HttpRequest<?> request);

}
```

http-server/src/main/java/io/micronaut/http/server/util/NettyJsonFilterBodyParser.java
sdelamo: [There is now only a single implementation of `JsonFilterBodyParser`. ](https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/7cdfd5db1058d7c09e928e0423726b490d951e8c)

http/src/main/java/io/micronaut/http/filter/bodyparser/JsonFilterBodyParser.java
sdelamo: [I have simplified the API](https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/7cdfd5db1058d7c09e928e0423726b490d951e8c)

dstepanov:
Can we use completable future instead of Publisher?

yawkat:
It is not correct that ServerHttpRequest is specific to netty. I am categorically against a new body access API that will have to be implemented for each server like this one.

The only netty-specific code in this PR is the use of QueryStringDecoder. Everything else does not need to be server-specific and does not even need an abstraction. And even QueryStringDecoder can be replaced by either the code you wrote for this in security, or by the copy of QueryStringDecoder that we have in micronaut-http.

sdelamo:
> It is not correct that ServerHttpRequest is specific to netty. I am categorically against a new body access API that will have to be implemented for each server like this one.
> The only netty-specific code in this PR is the use of QueryStringDecoder. Everything else does not need to be server-specific and does not even need an abstraction. And even QueryStringDecoder can be replaced by either the code you wrote for this in security, or by the copy of QueryStringDecoder that we have in micronaut-http.

Thanks for the suggestions. I have changed the PR. No downstream implementation will be needed.  @yawkat can you check again?

@dstepanov 
> Can we use completable future instead of Publisher? 

I have changed the API to use `CompleteableFuture` instead of `Publisher`

http-netty/src/main/java/io/micronaut/http/netty/NettyFormUrlEncodedDecoder.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Implementation of {@link FormUrlEncodedDecoder} based on Netty's {@link QueryStringDecoder}.
+ * @author Sergio del Amo
+ * @since 4.7.1
+ */
+@Singleton
+public class NettyFormUrlEncodedDecoder implements FormUrlEncodedDecoder {
```
graemerocher: it is still unclear to me why we need custom code for Netty when we have `QueryStringDecoder` forked from Netty in the `http` module

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/DefaultFilterBodyParser.java
```diff
@@ -0,0 +1,108 @@
...
+ * Implementation of {@link FilterBodyParser} which leverages {@link ServerHttpRequest#byteBody()} API.
+ * For form-url-encoded payloads it uses the {@link FormUrlEncodedDecoder} API.
+ * For JSON payloads it uses a {@link JsonMapper}.
+ * @author Sergio del Amo
+ * @since 4.11.0
+ */
+@Requires(beans = {JsonMapper.class, FormUrlEncodedDecoder.class})
+@Singleton
+@Experimental
+final class DefaultFilterBodyParser implements FilterBodyParser {
```
graemerocher: why is this still part of `http-server-netty`?

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/DefaultFilterBodyParser.java
sdelamo: A mistake. [I moved them to `http-server`](https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/0adaccfcdbdeb74af26e933cd438c38979a2b23b).

http-netty/src/main/java/io/micronaut/http/netty/NettyFormUrlEncodedDecoder.java
sdelamo: [I deleted the netty based implementation ](https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/e3a9611787265b4a75f978d20d3e3977aa03a510).

yawkat (approved):

http/src/main/java/io/micronaut/http/form/FormConfigurationProperties.java
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.form;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+
+@ConfigurationProperties("micronaut.form")
+final class FormConfigurationProperties implements FormConfiguration {
```
graemerocher, line 26: add javadoc to the type and extract the prefix into a constant

http/src/main/java/io/micronaut/http/uri/DefaultFormUrlEncodedDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+    private final FormConfiguration formConfiguration;
+
+    DefaultFormUrlEncodedDecoder(FormConfiguration formConfiguration) {
+        this.formConfiguration = formConfiguration;
+    }
+
+    @Override
+    @NonNull
+    public Map<String, Object> decode(@NonNull String formUrlEncodedString, @NonNull Charset charset) {
+        QueryStringDecoder decoder = new QueryStringDecoder(formUrlEncodedString, charset, false, formConfiguration.getMaxDecodedKeyValueParameters());
```
graemerocher, line 39: separate the parameters onto new lines to make this easier to read

graemerocher, line 39: does a new instance need creating on each invocation of `decode` or can it be assigned to a final field of the type?

http/src/main/java/io/micronaut/http/form/FormConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+ * Configuration related to forms.
+ * @author Sergio del Amo
+ * @since 4.7.1
+ */
+public interface FormConfiguration {
+    /**
+     *
+     * @return default maximum of decoded key value parameters
+     */
+    int getMaxDecodedKeyValueParameters();
```
graemerocher, line 28: specify here in the javadoc what is the default if not specific

http/src/main/java/io/micronaut/http/form/FormConfigurationProperties.java
```diff
@@ -0,0 +1,45 @@
...
+    @SuppressWarnings("WeakerAccess")
+    private static final int DEFAULT_MAX_DECODED_KEY_VALUE_PARAMETERS = 1024;
+
+    private int maxDecodedKeyValueParameters = DEFAULT_MAX_DECODED_KEY_VALUE_PARAMETERS;
+
+    /**
+     *
+     * @return default maximum of decoded key value parameters
+     */
+    public int getMaxDecodedKeyValueParameters() {
```
graemerocher, line 44: add missing `@Overrides` annotation

http-server/src/main/java/io/micronaut/http/server/filter/FilterBodyParser.java
```diff
@@ -0,0 +1,41 @@
...
+ * API to parse a body within a filter.
+ * @author Sergio del Amo
+ * @since 4.7.1
+ */
+@Experimental
+@FunctionalInterface
+@DefaultImplementation(DefaultFilterBodyParser.class)
+public interface FilterBodyParser {
+    /**
+     * @param request HTTP Request
```
graemerocher, line 40: add javadoc for the method and specify that the map is flattened and normalised as per the semantic of the `flatten()` method

http-server/src/main/java/io/micronaut/http/server/filter/FilterBodyParser.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.micronaut.context.annotation.DefaultImplementation;
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.http.HttpRequest;
+import reactor.util.annotation.NonNull;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * API to parse a body within a filter.
```
graemerocher:
```suggestion
 * API to parse a request body within a server filter.
```

http-server/src/main/java/io/micronaut/http/server/filter/FilterBodyParser.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Sergio del Amo
+ * @since 4.7.1
```
graemerocher, line 28:
```suggestion
 * @since 4.7.1
 * @see io.micronaut.http.filter.HttpServerFilter
 * @see io.micronaut.http.annotation.ServerFilter
```

http/src/main/java/io/micronaut/http/form/FormConfigurationProperties.java
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.form;
+
+import io.micronaut.context.annotation.ConfigurationProperties;
+
+@ConfigurationProperties("micronaut.form")
```
graemerocher: rename prefix to `micronaut.http.forms`

http-server/src/main/java/io/micronaut/http/server/filter/DefaultFilterBodyParser.java
```diff
@@ -0,0 +1,107 @@
...
+        this.jsonMapper = jsonMapper;
+    }
+
+    @Override
+    @NonNull
+    @SingleResult
+    public CompletableFuture<Map<String, Object>> parseBody(@NonNull HttpRequest<?> request) {
+        Optional<MediaType> mediaTypeOptional = request.getContentType();
+        if (mediaTypeOptional.isEmpty()) {
+            LOG.debug("Could not parse body into a Map because the request does not have a Content-Type");
```
graemerocher: wrap all the debug statements in if statements

http-server/src/main/java/io/micronaut/http/server/filter/DefaultFilterBodyParser.java
```diff
@@ -0,0 +1,107 @@
...
+        LOG.debug("Could not parse body into a Map because the request's content type is not either application/x-www-form-urlencoded or application/json");
+        return CompletableFuture.completedFuture(Collections.emptyMap());
+    }
+
+    private CompletableFuture<Map<String, Object>> parseJson(@NonNull ServerHttpRequest<?> request) {
+        try (CloseableByteBody closeableByteBody = request.byteBody().split(ByteBody.SplitBackpressureMode.FASTEST)) {
+            return closeableByteBody.buffer()
+                .thenApply(bb -> {
+                    try {
+                        return jsonMapper.readValue(bb.toByteArray(), Argument.mapOf(String.class, Object.class));
```
graemerocher: extract the `mapOf` call into a constant

dstepanov (approved):

graemerocher (changes requested):

http-server/src/main/java/io/micronaut/http/server/filter/DefaultFilterBodyParser.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/569cdd5a7417bb6820e71967301459641dbd9af1

http/src/main/java/io/micronaut/http/form/FormConfigurationProperties.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/7be3b6be79c7d2ddd865db81c15686362bd71a63

http-server/src/main/java/io/micronaut/http/server/filter/DefaultFilterBodyParser.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/235c74d51945f9ee4b9da864cfb5039adb1027cc

http/src/main/java/io/micronaut/http/form/FormConfigurationProperties.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/497d3729bc21acdaae4451a69ce7710c9a67cdd3

http/src/main/java/io/micronaut/http/uri/DefaultFormUrlEncodedDecoder.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/69d762ecab6f0f7a181cd2126c5b6a5237b84e03

http/src/main/java/io/micronaut/http/form/FormConfigurationProperties.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/d52f7c74ecb4436a41acd74349385225d3417aa1

http/src/main/java/io/micronaut/http/form/FormConfiguration.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/243edad24a190e265f08575c7168a993e7e94f35

http/src/main/java/io/micronaut/http/uri/DefaultFormUrlEncodedDecoder.java
sdelamo: Yes, I think you need to instantiate one instance and then invoke `QueryStringDecoder:parameters()` per invocation.

http-server/src/main/java/io/micronaut/http/server/filter/FilterBodyParser.java
sdelamo: I don't think an api should define in its javadoc how the internal implementations work. That should be in the javadoc of the implementations. However, as requested I have added javadoc in https://github.com/micronaut-projects/micronaut-core/pull/11272/commits/86e4fc072741e788188fbfccdcd789e244f9cf87

http-server/src/main/java/io/micronaut/http/server/filter/FilterBodyParser.java
```diff
@@ -0,0 +1,46 @@
...
+ * @see io.micronaut.http.annotation.ServerFilter
+ */
+@Experimental
+@FunctionalInterface
+@DefaultImplementation(DefaultFilterBodyParser.class)
+public interface FilterBodyParser {
+    /**
+     * Attempts to parse the request body into a Map.
+     * The default implementation {@link DefaultFilterBodyParser} uses {@link io.micronaut.http.form.FormUrlEncodedDecoder} for form-url-encoded payloads and a {@link io.micronaut.json.JsonMapper} for JSON payloads.
+     * For form-url-encoded payloads are first decoded into a Map with key {@code String} and value {@code List<String>} is flattened to a Map of key {@code String} and value {@code Object>}. If the value is a list of 1, the first item of that list becomes the value of the resulting map.
```
sdelamo:
```suggestion
     * form-url-encoded payloads are first decoded into a Map with key {@code String} and value {@code List<String>}. Then, the map is flattened to a Map of key {@code String} and value {@code Object>}. If original's map value is a list of 1 item, the item of that list becomes the value of the resulting map.
```

graemerocher (approved):

PR #11278, opened 2024-10-27
Add new `definitionType` parameter for `@Client` annotation

altro3:
This was asked in the Discord, and I was also interested in this question earlier: is it possible to use a single interface to describe the client and the controller? The feign library was invented for Spring (8-9 years ago), which later became part of Spring Cloud. And so, when I saw that Micronaut has built-in support for a single interface, I was happy, but it turned out that the client and controller do not support the single interface mode.

So, for now, the main problem, in my opinion, is the opposite perception of the Consumes and Produces annotations for the client and the controller. I added the `definitionType` parameter, which can be one of two values - `CLIENT` (default) and `SERVER`. `SERVER` value means that you use server (controller) interface to describe http client endpoints.

Perhaps there are some other significant differences between the interface for the controller and the client, but I do not know about them.

Why is this important: in general, it is quite common practice to make a microservice with 2 submodules - API and business logic, so that you can write all the endpoints and DTOs in the API module, and then simply inherit from the interface and thus get a ready-made HTTP client.

@graemerocher @yawkat @sdelamo @dstepanov

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
@@ -17,8 +17,11 @@ package io.micronaut.http.client.annotation;
 
 import io.micronaut.aop.Introduction;
 import io.micronaut.context.annotation.AliasFor;
+import io.micronaut.core.annotation.Experimental;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.http.HttpVersion;
+import io.micronaut.http.annotation.Consumes;
+import io.micronaut.http.annotation.Produces;
 import io.micronaut.http.client.HttpClientConfiguration;
 import io.micronaut.http.client.HttpVersionSelection;
 import io.micronaut.http.hateoas.JsonError;
@@ -54,6 +57,28 @@ public @interface Client {
     @AliasFor(member = "value") // <2>
     String id() default "";
 // end::value[]
+
+    /**
+     * The interface definition type. When set to {@link DefinitionType.SERVER} the {@link Produces} and {@link Consumes}
+     * definition evaluated for each method of the interface will be reversed .
+     *
+     * <p>Whilst not necessarily recommended, there are scenarios like testing where it is useful to share a common interface between client and server and use the
+     * interface to create a new client declarative client. The client typically needs to produce the content type accepted by the server and consume the content type
+     * produced by the server. In this arrangement using the interface directly will not result in the correct behaviour.</p>
+     *
+     * <p>In this scenario you can set {@link DefinitionType} to {@link DefinitionType.SERVER} which will ensure the requests sent by the client use the content type declared by the {@link Consumes}
+     *  annotation of the interface and that responses use the content type declared by the {@link Produces}.</p>
+     *
+     * <p>The default behaviour is to use {@link DefinitionType.CLIENT} where the inverse of the above is true.</p>
+     *
+     * @return The interface definition type
+     * @since 4.8.0
+     * @see Consumes
+     * @see Produces
+     */
+    @Experimental
+    @NonNull
+    DefinitionType definitionType() default DefinitionType.CLIENT;
 
     /**
      * The base URI for the client. Only to be used in
@@ -109,4 +134,39 @@ public @interface Client {
         HttpVersionSelection.ALPN_HTTP_2,
         HttpVersionSelection.ALPN_HTTP_1
     };
+
+    /**
+     * The interface definition type.
+     *
+     * @since 4.8.0
+     */
+    @Experimental
+    enum DefinitionType {
+        /**
+         * Client interface definition type.
+         */
+        CLIENT,
+        /**
+         * Server (controller) interface definition type.
+         */
+        SERVER;
+
+        /**
+         * Returns true, if this definition type is {@link DefinitionType.CLIENT}.
+         *
+         * @return true, if this definition type is {@link DefinitionType.CLIENT}.
+         */
+        public boolean isClient() {
+            return this == CLIENT;
+        }
+
+        /**
+         * Returns true, if this definition type is {@link DefinitionType.SERVER}.
+         *
+         * @return true, if this definition type is {@link DefinitionType.SERVER}.
+         */
+        public boolean isServer() {
+            return this == SERVER;
+        }
+    }
 }
--- a/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
@@ -401,8 +401,10 @@ public class HttpClientIntroductionAdvice implements MethodInterceptor<Object, O
 
         final MediaType[] acceptTypes;
         Collection<MediaType> accept = request.accept();
+        var definitionType = annotationMetadata.enumValue(Client.class, "definitionType", Client.DefinitionType.class)
+            .orElse(Client.DefinitionType.CLIENT);
         if (accept.isEmpty()) {
-            String[] consumesMediaType = context.stringValues(Consumes.class);
+            String[] consumesMediaType = context.stringValues(definitionType.isClient() ? Consumes.class : Produces.class);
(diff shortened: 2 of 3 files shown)


sdelamo:
I personally don't think we should do this. It adds more complexity.

dstepanov:
The use-case of reusing the same interface for a client and a controller is valid. Maybe the property can be called `controllerDefinition` to indicate that the interface is mainly used to define the controller. WDYT @graemerocher

graemerocher:
I would say we add `kind = ClientKind.SERVER` with an enum. Committing to a boolean makes it difficult to evolve and deprecate. We can also mark the enumeration as experimental so we can get more feedback.

altro3:
@graemerocher Perhaps it would be better to give this parameter another name. It is a bit strange to name the `ClientKind` - CLIENT and SERVER.

For example, `definitionType`

graemerocher:
sure

yawkat:
Is there anything that needs to be mirrored except for Consumes/Produces? We could also simply add RequestType and ResponseType annotations that alias to Consumes or Produces depending on use site.

altro3:
@yawkat In my opinion, adding one additional parameter is much easier than adding new annotations.

New annotations are more likely to confuse users - when what should be used. My personal opinion is that adding new annotations is exactly the same complication of logic and understanding. And an additional parameter in an annotation is quite laconic and is perceived as a setting

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
```diff
@@ -109,4 +120,28 @@ String[] alpnModes() default {
...
+         */
+        SERVER;
+
+        public boolean isClient() {
+            return this == CLIENT;
+        }
+
+        public boolean isServer() {
+            return this == SERVER;
+        }
```
graemerocher: these methods need javadoc

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -401,8 +401,10 @@ private RequestBinderResult bindRequest(MethodInvocationContext<Object, Object>
 
         final MediaType[] acceptTypes;
         Collection<MediaType> accept = request.accept();
+        var definitionType = annotationMetadata.enumValue(Client.class, "definitionType", Client.DefinitionType.class)
+            .orElse(Client.DefinitionType.CLIENT);
         if (accept.isEmpty()) {
-            String[] consumesMediaType = context.stringValues(Consumes.class);
+            String[] consumesMediaType = context.stringValues(definitionType.isClient() ? Consumes.class : Produces.class);
             if (ArrayUtils.isEmpty(consumesMediaType)) {
                 acceptTypes = DEFAULT_ACCEPT_TYPES;
```
graemerocher, line 409: it is likely this default handling needs altering based on the definition type

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
```diff
@@ -55,6 +55,17 @@
...
 // end::value[]
 
+    /**
+     * The interface definition type. Use {@code DefinitionType.SERVER} type, if you want
+     * to use controller interface to describe HTTP client. Use {@code DefinitionType.CLIENT} type,
+     * if you want to use custom interface to describe HTTP client.
+     *
+     * @return The interface definition type
+     * @since 4.8.0
+     */
```
graemerocher:
```suggestion
 /**
     * The interface definition type. When set to {@link DefinitionType.SERVER} the {@link Produces} and {@link Consumes}
     * definition evaluated for each method of the interface will be reversed .
     *
     * <p>Whilst not necessarily recommended, there are scenarios like testing where it is useful to share a common interface between client and server and use the
     * interface to create a new client declarative client. The client typically needs to produce the content type accepted by the server and consume the content type
     * produced by the server. In this arrangement using the interface directly will not result in the correct behaviour.</p>
     *
     * <p>In this scenario you can set {@link DefinitionType} to {@link DefinitionType.SERVER} which will ensure the requests sent by the client use the content type declared by the {@link Consumes}
     *  annotation of the interface and that responses use the content type declared by the {@link Produces}.</p>
     *  
     * <p>The default behaviour is to use {@link DefinitionType.CLIENT} where the inverse of the above is true.</p>
     *
     * @return The interface definition type
     * @since 4.8.0
     * @see Consumes
     * @see Produces
     */
```

graemerocher: you will need to add the imports when merging this

graemerocher (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
altro3: I don't understand what you're suggesting?

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
altro3: fixed

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
altro3: fixed

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
graemerocher: perhaps it is ok

yawkat (approved):
I think a new annotation is "simpler" than a parameter on `@Client` that magically switches around the meaning of method annotations at-a-distance. But this PR is a low maintenance commitment so I'm not opposed

yawkat:
BTW @altro3 you do not need to force-push your changes into a single commit each time, we will squash any commits when merging anyway.

altro3:
@yawkat 
Yes, I understand, but I have a habit. If this is a problem, then I will try not to do it next time (although, it is indeed a habit, and I may do it again)

Is this a bad habit?

graemerocher:
it makes it a bit more difficult to review individual changes that address code review

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
```diff
@@ -55,6 +57,27 @@
...
+     *
+     * <p>The default behaviour is to use {@link DefinitionType.CLIENT} where the inverse of the above is true.</p>
+     *
+     * @return The interface definition type
+     * @since 4.8.0
+     * @see Consumes
+     * @see Produces
+     */
+    @NonNull
+    DefinitionType definitionType() default DefinitionType.CLIENT;
```
graemerocher, line 81: can you annotate this with `@Experiemental`

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
```diff
@@ -109,4 +132,38 @@ String[] alpnModes() default {
         HttpVersionSelection.ALPN_HTTP_2,
         HttpVersionSelection.ALPN_HTTP_1
     };
+
+    /**
+     * The interface definition type.
+     *
+     * @since 4.8.0
+     */
+    enum DefinitionType {
```
graemerocher, line 144: can you annotate this with `@Experiemental`

graemerocher (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
altro3: fixed. Sorry, again force pushed. Forgot

http-client-core/src/main/java/io/micronaut/http/client/annotation/Client.java
altro3: fixed

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #11284, opened 2024-10-29
deprecate HttpResponse::getStatus

sdelamo:
This pull-request deprecates `HttpResponse::getStatus` and `HttpClientResponseException::getStatus`. It adds `HttpClientResponseException::code`, and `HttpClientResponseException::reason` methods.

In order to be custom status codes compatible is better for people to use `HttpResponse::code` instead of `HttpResponse::getStatus`.

For example, we hit an issue in Micrometer https://github.com/micronaut-projects/micronaut-micrometer/pull/857

I have annotated the method as deprecated but not for removal. I don’t think we should remove it as it will break a lot of apps and tests.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
@@ -83,7 +83,7 @@ public class HttpClientResponseException extends HttpClientException implements 
 
     /**
      * @return The {@link io.micronaut.http.HttpStatus} returned
-     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}
+     * To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}.
      */
     @Deprecated(since = "4.7.2")
     public HttpStatus getStatus() {
--- a/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
@@ -85,7 +85,6 @@ public class HttpClientResponseException extends HttpClientException implements 
      * @return The {@link io.micronaut.http.HttpStatus} returned
      * To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}.
      */
-    @Deprecated(since = "4.7.2")
     public HttpStatus getStatus() {
         return getResponse().getStatus();
     }
--- a/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
@@ -83,9 +83,25 @@ public class HttpClientResponseException extends HttpClientException implements 
 
     /**
      * @return The {@link io.micronaut.http.HttpStatus} returned
+     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}
      */
+    @Deprecated(since = "4.7.2")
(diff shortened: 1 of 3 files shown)


graemerocher:
I think this will be a large and disruptive deprecation for a lot of codebases. Not sure it is worth deprecating to be honest

yawkat:
I think we should avoid getStatus in framework code, but for most users, it's not really a problem to use because they don't use custom status types in the first place.

sdelamo:
>  I think we should avoid getStatus in framework code. 

The problem is how to convey this to @micronaut-projects/core-developers without deprecating it.

http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
```diff
@@ -83,11 +83,27 @@ public HttpResponse<?> getResponse() {
 
     /**
      * @return The {@link io.micronaut.http.HttpStatus} returned
+     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}
```
sdelamo:
```suggestion
     * To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}.
```

http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientResponseException.java
```diff
@@ -83,11 +83,27 @@ public HttpResponse<?> getResponse() {
...
      */
+    @Deprecated(since = "4.7.2")
```
sdelamo:
```suggestion
```

http/src/main/java/io/micronaut/http/HttpResponse.java
```diff
@@ -36,7 +36,9 @@ public interface HttpResponse<B> extends HttpMessage<B> {
 
     /**
      * @return The current status
+     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}
      */
+    @Deprecated(since = "4.7.2")
```
sdelamo:
```suggestion
```

http/src/main/java/io/micronaut/http/HttpResponse.java
```diff
@@ -36,7 +36,9 @@ public interface HttpResponse<B> extends HttpMessage<B> {
 
     /**
      * @return The current status
+     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}
```
sdelamo:
```suggestion
     * To support custom status codes. Use {@link #code()} instead of {@link #getStatus()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #getStatus()} and {@link HttpStatus#getReason()}
```

http/src/main/java/io/micronaut/http/HttpResponse.java
```diff
@@ -68,7 +70,9 @@ default HttpResponse<B> setAttribute(CharSequence name, Object value) {
 
     /**
      * @return The HTTP status
+     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #status()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #status()} and {@link HttpStatus#getReason()}
      */
+    @Deprecated(since = "4.7.2")
```
sdelamo:
```suggestion
```

http/src/main/java/io/micronaut/http/HttpResponse.java
```diff
@@ -68,7 +70,9 @@ default HttpResponse<B> setAttribute(CharSequence name, Object value) {
 
     /**
      * @return The HTTP status
+     * @deprecated To support custom status codes. Use {@link #code()} instead of {@link #status()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #status()} and {@link HttpStatus#getReason()}
```
sdelamo:
```suggestion
     * To support custom status codes. Use {@link #code()} instead of {@link #status()} and {@link HttpStatus#getCode()}  and {@link #reason()} instead of {@link #status()} and {@link HttpStatus#getReason()}
```

sdelamo:
@graemerocher @yawkat I have removed the deprecation but kept the javadoc clarifications. Can you re-review?

n0tl3ss (approved):

yawkat (approved):

graemerocher (approved):

PR #11288, opened 2024-10-29
Migrate netty HttpClient to ExecutionFlow

yawkat:
Also adds cancellation support to ExecutionFlow.


core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
```diff
@@ -135,6 +137,34 @@ public ImperativeExecutionFlow<T> tryComplete() {
...
+            this.onCancel = () -> {
+                prev.run();
+                hook.run();
+            };
+        } else {
+            this.onCancel = hook;
+        }
+    }
+
     private abstract static class Step<I, O> {
```
graemerocher: can this be sealed?

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
```diff
@@ -334,4 +364,13 @@ ExecutionFlow<E> apply(ExecutionFlow<E> executionFlow) {
             return executionFlow;
         }
     }
+
+    private static final class Cancel<E> extends Step<E, E> {
+
+        @Override
+        ExecutionFlow<E> apply(ExecutionFlow<E> input) {
+            input.cancel();
+            return ExecutionFlow.error(new Exception("Cancelled"));
```
graemerocher: shouldn't this be `java.util.concurrent.CancellationException`

graemerocher (changes requested):
I assume these changes won't break context propagation and distributed tracing?

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlowImpl.java
yawkat: ive changed it. this exception should actually never surface, cancel must always be the last step in a DelayedExecutionFlow. the job of the Cancel step is to cancel the upstream steps, it can't handle downstreams.

yawkat:
The code is mostly the same so I don't think it should break anything.

graemerocher (approved):

sdelamo (commented):
Why are we migration to `ExecutionFlow`? Is that more efficient that using Project reactor code?

yawkat:
Yep. Especially for cases where the result is immediately available

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
```diff
@@ -84,6 +88,30 @@ public ExecutionFlow<Object> putInContext(String key, Object value) {
...
+                return context;
+            }
+        });
+        return this;
+    }
+
+    @Override
+    public void cancel() {
+        List<Subscription> stc;
+        synchronized (this) {
```
dstepanov, line 106: Should we avoid `synchronized`?

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
```diff
@@ -102,7 +130,20 @@ public Context currentContext() {
             @Override
             public void onSubscribe(Subscription s) {
                 this.subscription = s;
-                s.request(1);
+                boolean cancel;
+                synchronized (ReactorExecutionFlowImpl.this) {
```
dstepanov, line 134: I don't understand what is happening. Should we avoid a subscription if the flow is canceled?

dstepanov (changes requested):

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
yawkat: synchronized is fine here because the block contains no other method calls

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
yawkat: ExecutionFlow downstream APIs (onComplete and cancel) are generally not thread safe. My assumption with this API is that cancel is always called strictly after onComplete is called. In particular, DelayedFlow will throw an exception if you call onComplete after cancel. That means that if the flow is cancelled, the user shouldn't be subscribing afterwards anyway (though there is no check in ReactiveExecutionFlow).

This check is only necessary because onSubscribe can happen some time delayed after onComplete is called, so by that time the flow *can* be cancelled.

dstepanov (approved):

PR #11294, opened 2024-10-31
Remove context propagation filter

yawkat:
Implement "proper" context propagation in the client to get rid of the old ClientServerContextFilter.

Because context propagation uses its own key for the reactor context instead of ServerRequestContext.KEY, I've deprecated ServerRequestContext.KEY and added a more general API that can use the propagated context from the ContextView.


http/src/main/java/io/micronaut/http/context/ServerRequestContext.java
```diff
@@ -99,5 +106,20 @@ public static <T> T with(@Nullable HttpRequest<?> request, @NonNull Callable<T>
...
+
+    /**
+     * Retrieve the current server request context from the given reactor context view.
+     *
+     * @param context The reactor context view
+     * @param <T> The body type
+     * @return The request context if it is present
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> Optional<HttpRequest<T>> currentRequest(ContextView context) {
```
graemerocher: add `@since` and nullability annotation to the parameter

graemerocher (approved):

dstepanov (approved):

PR #11297, opened 2024-11-01
Optimizations in reactor <-> ExecutionFlow transformations

yawkat:
- In toPublisher, return a specialized FlowAsMono that (a) directly extends Mono so does not need a reactor wrapper and (b) supports SYNC fusion so that when the flow completes after the toPublisher call but before subscribe, we can take advantage of reactor fusion operations.
- Introduce a new method fromPublisherEager that can unwrap FluxAsMono directly (to short-circuit the direct ExecutionFlow->Publisher->ExecutionFlow transform) but can also transform Monos that complete immediately (e.g. Mono.just(…).map(…)) directly to ImperativeExecutionFlow.


http/src/main/java/io/micronaut/http/reactive/execution/ReactiveExecutionFlow.java
```diff
@@ -91,4 +108,8 @@ static <K> ReactiveExecutionFlow<K> fromFlow(@NonNull ExecutionFlow<K> flow) {
     static <K> Publisher<K> toPublisher(Supplier<ExecutionFlow<K>> flowSupplier) {
         return (Publisher<K>) ReactorExecutionFlowImpl.toMono(flowSupplier);
     }
+
+    static <K> Publisher<K> toPublisher(ExecutionFlow<K> flow) {
```
graemerocher: javadoc and nullability annotations

graemerocher (approved):

dstepanov:
Can you please merge after https://github.com/micronaut-projects/micronaut-core/pull/11317

graemerocher:
please address the sonar blocking issue

PR #11299, opened 2024-11-01
Improve displaying of circular dependency error

andriy-dmytruk:
For some classes `ClassA`, `ClassB`, `ClassC` and `ClassD` that have a dependency cycle, the current message is:
```
Failed to inject value for field [propA] of class: io.micronaut.inject.failures.ctorcirculardependency.MyClassB
 
Message: Circular dependency detected
Path Taken: 
new MyClassD(MyClassB propB) --> new MyClassD([MyClassB propB]) --> MyClassB.propA --> new MyClassA([MyClassC propC]) --> new MyClassC([MyClassB propB])
^                                                                                                                                                     |
|                                                                                                                                                     |
|                                                                                                                                                     |
+-----------------------------------------------------------------------------------------------------------------------------------------------------+
```

After the change it would be:
```
Failed to inject value for field [propA] of class: io.micronaut.inject.failures.ctorcirculardependency.MyClassB
 
Message: Circular dependency detected
Path Taken: 
new MyClassD(MyClassB propB)
      \---> new MyClassD([MyClassB propB])
            \---> MyClassB.propA
                  ^  \---> new MyClassA([MyClassC propC])
                  |        \---> new MyClassC([MyClassB propB])
                  |              |
                  +--------------+
```

This improves it by:
* Showing the actual cycle, as the cycle does not always include the first element in the path. In this case it is `ClassB->ClassA->ClassC->ClassB`, so `ClassD` is ommitted
* Displaying each entry on a new line to improve readability. This is important for classes with long class names and multiple injection parameters. I changed the test class names to longer, as `A`, `B`, `C` did not seem realistic.

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/failures/ConstructorCircularDependencyFailureSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/failures/ConstructorCircularDependencyFailureSpec.groovy
@@ -31,36 +31,66 @@ class ConstructorCircularDependencyFailureSpec extends Specification {
         ApplicationContext context = ApplicationContext.run()
 
         when:"A bean is obtained that has a setter with @Inject"
-        B b =  context.getBean(B)
+        MyClassB b =  context.getBean(MyClassB)
 
         then:"The implementation is injected"
         def e = thrown(CircularDependencyException)
         e.message.normalize() == '''\
-Failed to inject value for field [a] of class: io.micronaut.inject.failures.ConstructorCircularDependencyFailureSpec$B
+Failed to inject value for field [propA] of class: io.micronaut.inject.failures.ConstructorCircularDependencyFailureSpec$MyClassB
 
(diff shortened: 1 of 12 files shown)


andriy-dmytruk:
The non-cyclic path display could be changed to a similar, as well, but I am not sure if that would be an improvement there.

graemerocher:
looks great! Just have to adjust the original tests

graemerocher (approved):

graemerocher:
Another rendering anomaly that may be worth looking at is when you have a Factory bean in the middle that causes the loop I am not sure we are rendering the factory name correctly

andriy-dmytruk:
> Another rendering anomaly that may be worth looking at is when you have a Factory bean in the middle that causes the loop I am not sure we are rendering the factory name correctly

The display in the test I added looks good. But that's a great idea to add the test, I caught an edge case in my implementation where the circle did not point to correct element.

PR #11300, opened 2024-11-01
Improve event loop locality of client requests

yawkat:
If a request is made on an event loop that is part of the same event loop group that the client is configured to use, prefer making the connection on that same event loop. This keeps all the relevant processing on the same thread and may improve performance.


graemerocher (approved):

sdelamo:
@yawkat does this PR still apply?

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -1062,6 +1065,57 @@ public int getMaxConcurrentHttp2Connections() {
...
             this.maxConcurrentHttp2Connections = maxConcurrentHttp2Connections;
         }
+
+        /**
+         * Optimize locality of client connections depending on which event loop makes a request.
+         * [available in the Netty HTTP client]
+         *
+         * @return The locality configuration
+         */
+        public @NonNull HttpClientConfiguration.ConnectionPoolConfiguration.ConnectionLocality getConnectionLocality() {
```
dstepanov, line 1076: Please add since to the new methods and classes

http-client/src/main/java/io/micronaut/http/client/netty/PoolResizer.java
```diff
@@ -250,11 +310,72 @@ final void markConnectionAvailable() {
...
+    private static final class ConnectionList {
+        private static final ResizerConnection[] EMPTY = new ResizerConnection[0];
+
+        private final Lock lock = new ReentrantLock();
+
+        /**
+         * Copy of {@link #safeItems} <i>only</i> for use in {@link #doSomeWork()}, without lock.
+         * {@link #doSomeWork()} may shuffle and reorder this array in-place as needed.
+         */
+        private volatile ResizerConnection[] unsafeItems = EMPTY;
```
dstepanov, line 331: Just use `CopyOnWriteArrayList`

dstepanov (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/PoolResizer.java
yawkat: i want to avoid lock contention in doSomeWork, and it also gives me the option to sort the connections in-place there

PR #11306, opened 2024-11-04
Minor server optimizations

yawkat:
NameUtilsTest takes too long to run as part of the suite, but if we need to optimize another of its methods, it may be useful again.

### DIFF
--- a/core/src/test/java/io/micronaut/core/naming/NameUtilsTest.java
+++ b/core/src/test/java/io/micronaut/core/naming/NameUtilsTest.java
@@ -1,0 +1,118 @@
+package io.micronaut.core.naming;
+
+import io.micronaut.core.util.StringUtils;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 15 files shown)


graemerocher (approved):

core/src/test/java/io/micronaut/core/naming/NameUtilsTest.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.LongAdder;
+import java.util.regex.Pattern;
+import java.util.stream.IntStream;
+
+@Disabled
```
dstepanov, line 18: Disabled?

dstepanov (changes requested):

core/src/test/java/io/micronaut/core/naming/NameUtilsTest.java
yawkat: yep, see PR text :P

PR #11309, opened 2024-11-05
Add lost SERVICE_ID request attribute with jdk http client.

### LINKED ISSUE #11041: HttpAttributes.SERVICE_ID not set when using jdk http client
### Expected Behavior

Service ID should be propagated in attributes

### Actual Behaviour

nothing is there - meaning that the OAuth2 token flow is not invoked.

### Steps To Reproduce

Have a service defined with @Client using OAuth2, and attempt to match with service-regex

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.2.3

altro3:
Fixed #11041


graemerocher:
@yawkat do we have a TCK now for the client? Would be good to add a test to that if there is

yawkat:
agreed

yawkat:
Thanks!

PR #11311, opened 2024-11-05
Specific `RequestArgumentBinder`

dstepanov:
I haven't added a specific binder because I think it's also overridden in servlet.


yawkat (approved):

dstepanov:
The other specific can be create from the route but HTTP module doesn't depend on the route module.

graemerocher (approved):

PR #11317, opened 2024-11-06
Improvements to `ExecutionFlow`

dstepanov:
Please check if they make sense

### DIFF
--- a/core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlow.java
+++ b/core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlow.java
@@ -18,13 +18,16 @@ package io.micronaut.core.execution;
 import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.annotation.Nullable;
 
+import java.util.concurrent.CompletionException;
+import java.util.function.BiConsumer;
+
 /**
  * {@link ExecutionFlow} that can be completed similar to a
  * {@link java.util.concurrent.CompletableFuture}.
  *
  * @param <T> The type of this flow
  */
-public sealed interface DelayedExecutionFlow<T> extends ExecutionFlow<T> permits DelayedExecutionFlowImpl {
+public sealed interface DelayedExecutionFlow<T> extends ExecutionFlow<T>, BiConsumer<T, Throwable> permits DelayedExecutionFlowImpl {
     static <T> DelayedExecutionFlow<T> create() {
         return new DelayedExecutionFlowImpl<>();
     }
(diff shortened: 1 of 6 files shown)


graemerocher (approved):

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlow.java
```diff
@@ -18,13 +18,16 @@
...
+import java.util.function.BiConsumer;
+
 /**
  * {@link ExecutionFlow} that can be completed similar to a
  * {@link java.util.concurrent.CompletableFuture}.
  *
  * @param <T> The type of this flow
  */
-public sealed interface DelayedExecutionFlow<T> extends ExecutionFlow<T> permits DelayedExecutionFlowImpl {
+public sealed interface DelayedExecutionFlow<T> extends ExecutionFlow<T>, BiConsumer<T, Throwable> permits DelayedExecutionFlowImpl {
```
yawkat: i dont like making this part of the API. how about a method `BiConsumer completer() { return this }` instead

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
```diff
@@ -114,6 +114,14 @@ public void cancel() {
 
     @Override
     public void onComplete(BiConsumer<? super Object, Throwable> fn) {
+        if (value instanceof Fuseable.ScalarCallable callable) {
+            try {
+                fn.accept(callable.call(), null);
```
yawkat: only put the .call inside the try, in case the consumer throws

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
```diff
@@ -164,6 +172,45 @@ public void onComplete() {
...
+        }
+        value.subscribe(new CoreSubscriber<>() {
+
+            Subscription subscription;
+            Object value;
+
+            @Override
+            public void onSubscribe(Subscription s) {
+                this.subscription = s;
+                s.request(1);
```
yawkat: use Long.MAX_VALUE, and dont request again in onNext

yawkat (approved):

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlow.java
dstepanov: Moved to implementation

http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
dstepanov: Originally copied from the other method, corrected both

core/src/main/java/io/micronaut/core/execution/DelayedExecutionFlow.java
yawkat: thanks

PR #11319, opened 2024-11-06
Fix lifecycle of HttpClient and BlockingHttpClient

### LINKED ISSUE #11307: Closed HTTP client reports incorrect error on request
### Issue description

It reports 'Cannot send HTTPS request. SSL is disabled' when it should be a specific error

glorrian:
#### Fix a bug  when a closed client issued an incorrect error as mentioned in #11307
Add special exceptions to simple requests and web sockets
Make tests for it
#### Add Lifecycle to the BlockingHttpClient interface as well as HttpClient interface
Implement new methods at BlockingHttpClient anonymous class at DefaultHttpClient
Implement new methods at JdkBlockingHttpClient
#### Fix tests where an undisclosed client was used
  JsonStreamSpec: delete "cleanup" with closing clients on each test since the client is shared between all tests and closes automatically at the end(@AutoClosable)
  HttpHeadSpec: replace "close" with "refresh" in "cleanup" of one test cause this instance of client is used by other tests
  ClientHostNameSpec: replace "close" with "refresh" in "cleanup" of one test cause HttpClient.create() caching clients instances so actually all tests use one instance of client
  ByteBufferSpec: make client field @Shared cause @Inject fields caching so all test actually use one instance of client


http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -193,6 +193,8 @@ public class ConnectionManager {
         this.clientCustomizer = from.clientCustomizer;
         this.informationalServiceId = from.informationalServiceId;
         this.nettyClientSslBuilder = from.nettyClientSslBuilder;
+
+        refresh();
```
yawkat: I don't think this is necessary

http-client/src/test/groovy/io/micronaut/http/client/ClientHostNameSpec.groovy
```diff
@@ -15,7 +15,7 @@ class ClientHostNameSpec extends Specification {
         e.message.contains('Connect Error: foo_bar') || e.message.contains('Connect Error: No such host is known (foo_bar)')
 
         cleanup:
-        client.close()
+        client.refresh()
```
yawkat: why these changes?

http-client/src/test/groovy/io/micronaut/http/client/HttpHeadSpec.groovy
```diff
@@ -364,7 +364,7 @@ class HttpHeadSpec extends Specification {
         body == "success"
 
         cleanup:
-        client.close()
+        client.refresh()
```
yawkat, line 367: this one too

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
glorrian: This is necessary cause this constructor is used for copying ConnectionManager to new instance so we need to bootstrap new instance to setup states. For example AtomicBoolean running is false by default and without refresh() isRunning() will return false. I add this line cause i had failed tests. Hope you will get me, thanks for your review.

http-client/src/test/groovy/io/micronaut/http/client/ClientHostNameSpec.groovy
glorrian: Cause if close this instance other tests will get an already closed client and then throw an exception when trying to use it

http-client/src/test/groovy/io/micronaut/http/client/HttpHeadSpec.groovy
glorrian: Same as previous comment

http-client/src/test/groovy/io/micronaut/http/client/ClientHostNameSpec.groovy
yawkat: the client in this test is not shared

glorrian: Not shared but cached inside factory i guess. When trying to run these tests, I get “closed client” exceptions, when I replace them with refresh, all the tests actually pass. I made quick look up inside the impl of factory client method, and it seems to me that this is caching code, but maybe I'm wrong. I'll check it again and come back with a final answer.

glorrian: You're right! these tests failed because of the vpn on my mac. I removed refresh() in this spec.

http-client/src/test/groovy/io/micronaut/http/client/HttpHeadSpec.groovy
glorrian: this one actually use one shared instance by hashCode
Screenshot 2024-11-07 at 23 46 33 copy

glorrian:
@graemerocher @yawkat can you restart build? I think this one failed test is flaky cause it passed on 21 java and on all my local environments

yawkat:
Thanks!

PR #11322, opened 2024-11-07
Minor perf optimizations

dstepanov:


http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -959,12 +959,19 @@ public String getExtension() {
...
+     * @return The parameters of the media type
      */
     public OptionalValues<String> getParameters() {
         return OptionalValues.of(String.class, parameters);
     }
 
+    /**
+     * @return The parameters map of the media type
+     */
+    public Map<CharSequence, String> getParametersMap() {
```
graemerocher, line 973: add `@since` and nullability annotations

http/src/main/java/io/micronaut/http/util/HttpUtil.java
```diff
@@ -54,32 +54,51 @@ public static boolean isFormData(HttpRequest<?> request) {
...
+    }
+
+    /**
+     * Resolve the {@link Charset} to use for the request.
+     *
+     * @param request The request
+     * @return An {@link Optional} of {@link Charset}
+     */
+    @SuppressWarnings("Duplicates")
+    public static Charset getResolveCharset(HttpMessage<?> request) {
```
graemerocher: add `@since` and nullability annotations

graemerocher (changes requested):

graemerocher (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -116,8 +116,49 @@
  * @since 1.0
  */
 @Internal
-public class NettyHttpRequest<T> extends AbstractNettyHttpRequest<T> implements HttpRequest<T>, PushCapableHttpRequest<T>, io.micronaut.http.FullHttpRequest<T>, ServerHttpRequest<T> {
+public final class NettyHttpRequest<T> extends AbstractNettyHttpRequest<T> implements HttpRequest<T>, PushCapableHttpRequest<T>, io.micronaut.http.FullHttpRequest<T>, ServerHttpRequest<T> {
     private static final Logger LOG = LoggerFactory.getLogger(NettyHttpRequest.class);
+    private static final BiConsumer<String, Object> CLEANUP_ATTRIBUTES = new BiConsumer<>() {
```
yawkat: I think it's possible to do this without a variable. put the lambda body in a static method and make it a method reference, then i believe it's cached.

http/src/main/java/io/micronaut/http/util/HttpUtil.java
```diff
@@ -54,32 +55,53 @@ public static boolean isFormData(HttpRequest<?> request) {
...
+    /**
+     * Resolve the {@link Charset} to use for the request.
+     *
+     * @param request The request
+     * @return An {@link Optional} of {@link Charset}
+     * @since 4.8
+     */
+    @SuppressWarnings("Duplicates")
+    @NonNull
+    public static Charset getResolveCharset(@NonNull HttpMessage<?> request) {
```
yawkat: Either call it get or call it resolve

yawkat (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -432,7 +391,48 @@ public void release() {
             formRouteCompleter.release();
         }
         if (attributes != null) {
-            attributes.forEach(CLEANUP_ATTRIBUTES);
+            attributes.forEach(cleanup());
```
yawkat: no it needs to be a method ref, something like NettyHttpRequest::cleanup

yawkat: Actually, looking at LambdaMetafactory, I think this change is altogether unnecessary: You only need to make releaseIfNecessary static, and the lambda won't capture anything, and will be cached.

yawkat:
@dstepanov please merge 4.8.x, there were some compilation issues on 4.8.x that i just fixed.

PR #11323, opened 2024-11-07
Support `RouteInfo` in filters, add `createSpecific` to `ArgumentBinder`

dstepanov:


core/src/main/java/io/micronaut/core/bind/ArgumentBinder.java
```diff
@@ -47,6 +48,17 @@
...
 public interface ArgumentBinder<T, S> {
 
+    /**
+     * Create a specific binder.
+     *
+     * @param argument The bound argument
+     * @return The specific binder
+     * @since 4.8
+     */
+    default ArgumentBinder<T, S> createSpecific(Argument<T> argument) {
```
graemerocher: nullability annotations

http-server/src/main/java/io/micronaut/http/server/binding/RouteInfoArgumentBinder.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Argument binder for {@link io.micronaut.web.router.RouteInfo} objects.
+ *
+ * @since 4.8.0
+ * @author Denis Stepanov
+ */
+@Singleton
+@Internal
+public final class RouteInfoArgumentBinder implements TypedRequestArgumentBinder<RouteInfo<?>>, FilterArgumentBinderPredicate {
```
graemerocher: does it need to be public?

graemerocher (changes requested):
Are there more argument binder that can benefit from createSpecific?

dstepanov:
> Are there more argument binder that can benefit from createSpecific?

Considering it's in the core, other implementations can use it.

graemerocher (approved):

yawkat (approved):

PR #11324, opened 2024-11-07
Replace HttpAttributes with accessors

yawkat:
Attributes are a suboptimal way of adding metadata to requests and responses, both because they need a string lookup, and because they require conversion.


http-client-core/src/main/java/io/micronaut/http/client/ClientAttributes.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Optional;
+
+/**
+ * Client-related attribute accessors.
+ *
+ * @author Jonas Konrad
+ * @since 4.8.0
+ */
+@SuppressWarnings("removal")
+public final class ClientAttributes {
```
graemerocher, line 32: is this regarded public API?

altro3:
@yawkat I think it would be better to add a fix for the missing `SERVICE_ID` attribute in the JDK http client at this point. See: https://github.com/micronaut-projects/micronaut-core/pull/11309

yawkat:
it's not super important that core code is migrated from HttpAttributes immediately, so #11309 can stay as-is. the main concern is users and non-core micronaut projects.

http/src/main/java/io/micronaut/http/HttpAttributes.java
```diff
@@ -15,70 +15,106 @@
...
 /**
  * Common HTTP attributes.
  *
  * @author graemerocher
  * @since 1.0
+ * @deprecated In order to make future optimizations possible, standard attributes should be
+ * accessed through their static method accessors (such as those in {@link BasicHttpAttributes})
+ * instead of directly through {@link io.micronaut.core.attr.AttributeHolder}.
  */
+@Deprecated(forRemoval = true)
```
graemerocher: add `@since` to all deprecated annotations

graemerocher (changes requested):

http-client-core/src/main/java/io/micronaut/http/client/ClientAttributes.java
yawkat: Yes

graemerocher (approved):

dstepanov (approved):

PR #11326, opened 2024-11-08
Update MicronautMetaServiceLoaderUtils.java to fix single threaded service loader #11229 

tbashour:
Fix for service loader #11229 

@andriy-dmytruk @graemerocher @dstepanov

I believe the change is not complete and still has a logical error. Another way to change the parallelisim would be to set -XX:ActiveProcessorCount. This option is used in dockerized environments to limit resources used by the application.

In my tests, wenn I set the option a value smaller than 3 (parallelism is active processor count -1), then micronaut application will not start.

I believe the condition to add the services in the class MicronautMetaServiceLoaderUtils (line 260) is not correct:

is:

if (val != null && predicate != null && !predicate.test(val)) {

but should be:

if (val != null && !collection.contains(val) && (predicate == null || predicate.test(val))) {

### DIFF
--- a/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
+++ b/core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
@@ -257,7 +257,7 @@ public final class MicronautMetaServiceLoaderUtils {
                 List<S> collection = new ArrayList<>(serviceEntries.size());
                 for (String serviceEntry : serviceEntries) {
                     S val = instantiate(serviceEntry, classLoader);
-                    if (val != null && predicate != null && !predicate.test(val)) {
+                    if (val != null && !collection.contains(val) && (predicate == null || predicate.test(val))) {
                         collection.add(val);
                     }
                 }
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
```diff
@@ -257,7 +257,7 @@ public List<S> collect(boolean allowFork) {
                 List<S> collection = new ArrayList<>(serviceEntries.size());
                 for (String serviceEntry : serviceEntries) {
                     S val = instantiate(serviceEntry, classLoader);
-                    if (val != null && predicate != null && !predicate.test(val)) {
+                    if (val != null && !collection.contains(val) && (predicate == null || predicate.test(val))) {
```
andriy-dmytruk: `collection.contains(val)` might be a costly operation, so I wonder if it is required. Do the services implement a correct equals method, so that this would actually make a difference?

If it is required,can we use a set instead of a list and don't do this check?

dstepanov:
Right, I don't think we need `collection.contains(val)`

core/src/main/java/io/micronaut/core/io/service/MicronautMetaServiceLoaderUtils.java
tbashour: The parallel option has it (line 312)
`if (result != null && !values.contains(result)) {`

That's why I added it. If the check is not required, I would remove it ... I would prefer to remove it from both places to avoid confusion.

dstepanov: I think we an remove it there

tbashour: okak, I have committed the change

dstepanov (approved):

graemerocher:
Thanks for the contribution!

PR #11327, opened 2024-11-08
Normalization of bean definition class names for more correct logs

### LINKED ISSUE #11320: NoSuchBeanException shouldn't show internally generated class names
### Expected Behavior

The message should be clear of extraneous noise and punctuation that could discourage people from reading it:

```
NoSuchBeanException: No bean of type Foo exists.
* FooImpl, a candidate of Foo, is disabled because:
 - Class Bar is not present
```

For bonus points, in the standard case of a single failed constraint:

```
NoSuchBeanException: No bean of type Foo exists. FooImpl is disabled because class Bar is not present
```

### Actual Behaviour

```
NoSuchBeanException: No bean of type [Foo] exists.
* [$FooImpl$Definition$Intercepted] a candidate of [Foo] is disabled because:
 - Class [Bar] is not present
```

### Steps To Reproduce

Depend on a bean that depends on a missing class (or whatever).

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.6.2

altro3:
Fixed #11320


dstepanov:
The correct fix would be for to replace `definition.asArgument().getTypeString(true)` with a check if it's `ProxyBeanDefinition` and get the proper type

altro3:
@dstepanov uhh... need more details. I didn't quite understand how to do it correctly in the end

dstepanov:
Something like `definition instanceof ProxyBeanDefinition` and use `getTargetType` probably

altro3:
What do you suggest? Add new methods to MessageUtil with arguments:
`normalizedBeanType(Argument<?> arg)`
`normalizedBeanType(BeanDefinition<?> arg)`
`normalizedBeanType(Class<?> arg)`

and add this check inside every method?

Are you sure this solution is better than string normalization?

dstepanov:
I think the problem is only in `BeanDefinition#asArgument`

altro3:
@dstepanov done. But I don't know how to write correct test for it

altro3:
@dstepanov ping

altro3:
@dstepanov ping 2025

altro3:
@dstepanov @graemerocher ping

graemerocher:
as part of https://github.com/micronaut-projects/micronaut-core/pull/11449 I added new methods under the `TypeInformation` interface changes the need for some of this PR. I think the PR should be updated to use these methods instead of adding more methods.

PR #11345, opened 2024-11-15
Upgrade Netty to 4.1.115.Final

dradosevic:
Upgrade Netty to 4.1.115.Final to address https://nvd.nist.gov/vuln/detail/CVE-2024-47535


dradosevic:
Also applied the change from https://github.com/micronaut-projects/micronaut-core/pull/10990 to fix the missing grolifant version which failed the build.

PR #11359, opened 2024-11-20
Update creatingClient.adoc

TbirdDuncan:
documentation is sparse on adding the http-client implementation and the hello world cli command creates an app with the dependency only on compile/testImplementation creating issues with requestInterceptors

### DIFF
--- a/src/main/docs/guide/quickStart/creatingClient.adoc
+++ b/src/main/docs/guide/quickStart/creatingClient.adoc
@@ -1,4 +1,6 @@
 As mentioned previously, the Micronaut framework includes both an <<httpServer,HTTP server>> and an <<httpClient,HTTP client>>. A <<lowLevelHttpClient,low-level HTTP client>> is provided which you can use to test the `HelloController` created in the previous section.
+Important note: you'll need to add 
+dependency:micronaut-http-client[]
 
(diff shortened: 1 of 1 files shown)


sdelamo (commented):
I am not sure about this change. You are highlighting the Netty based implementation of the http client but users may want to use the JDK based implementation. The paragraph above links to the http client implementation where users can find the necessary dependencies.

TbirdDuncan:
I think thats fair, however I implore folks to try to get a new user to build a micronaut client/server. It is quite the hassle with things are configured atm. I'm open to exploring new ideas. IMO the best solution is to better the error messaging that arises due when you try to inject the client around your app. Which is mentioned in the op (requestInterceptors). There's not enough guidance, if that makes sense. However, it still in no way states that it's required to have these on your implementation path to proceed.

glorrian:
dependency that you mentioned has strong binding to the netty based impl of the client, so i think the documentation should explain all case of the MK usage and you'd better be able to tell users how to choose  between a client implementation for real impact of this pr.

sdelamo:
I am closing this in favor of keeping just the links to the HTTP client sections. Thanks for the contribution, nonetheless.

Issue #11343, opened 2024-11-14
Zstd compression in MN4 fails for large response payloads


Jonathan-Hill:
### Expected Behavior

A large response payload (>32MB) should be compressed successfully when `zstd` compression is used, just as it is successful when other compression algorithms such as `gzip` are chosen.

### Actual Behaviour

When Zstd is used, payloads larger than 32 MB fail to compress. The same payload works fine if gzip is used, eg. the client sends a request header of `Accept-Encoding: gzip, deflate, br` but fails if the header is `Accept-Encoding: gzip, deflate, br, zstd`
This seems to be caused by the `ZstdOptions` class passing in a value for maxEncodeSize of 32MB (as defined in the `ZstdConstants` class).
This has caused some requests for clients to start failing since we upgraded to MN4 (presumably because Zstd wasn't supported in MN and/or Netty prior to this?).
The following error is thrown, and no response is returned to the client at all:

```
io.netty.handler.codec.EncoderException: requested encode buffer size (33701888 bytes) exceeds the maximum allowable size (33554432 bytes)
	at io.netty.handler.codec.compression.ZstdEncoder.allocateBuffer(ZstdEncoder.java:120)
	at io.netty.handler.codec.compression.ZstdEncoder.allocateBuffer(ZstdEncoder.java:37)
	at io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:893)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:875)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:984)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:868)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:863)
	at io.netty.channel.DefaultChannelPipeline.write(DefaultChannelPipeline.java:959)
	at io.netty.channel.AbstractChannel.write(AbstractChannel.java:295)
	at io.netty.channel.embedded.EmbeddedChannel.write(EmbeddedChannel.java:754)
	at io.netty.channel.embedded.EmbeddedChannel.writeOutbound(EmbeddedChannel.java:432)
[...]

PR #11361, opened 2024-11-20
Add maxEncodeSize configuration for zstd compression

glorrian:
As mentioned in #11343, Zstd encoder in Netty has a unique parameter of maxEncoded size, and if users of Micronaut want to use Zstd with files more than ≈31Mb Netty will throw an exception like "requested encode buffer size exceeds the maximum allowable size) so I thing we should enable change this parameters for MK's users if they would like to handle bigger files at MK server

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpCompressionStrategy.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpCompressionStrategy.java
@@ -39,4 +39,9 @@ public interface HttpCompressionStrategy extends Toggleable {
     default int getCompressionLevel() {
         return StandardCompressionOptions.gzip().compressionLevel();
     }
+
+    /**
+     * @return The maximum size of data that can be encoded using the zstd algorithm.
+     */
+    int getMaxZstdEncodeSize();
 }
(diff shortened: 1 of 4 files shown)


glorrian:
@yawkat @graemerocher @dstepanov I hope that I did everything right and I will be glad to receive a review. I can't write tests because its need to add a library https://github.com/luben/zstd-jni and i can't do it, because it's not in netty-bom and I can't connect it transitively because it's declared with optional=true in netty poms, I'm not sure about its license, so I didn't add it to libs.versions.toml

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpCompressionStrategy.java
```diff
@@ -39,4 +39,9 @@ public interface HttpCompressionStrategy extends Toggleable {
     default int getCompressionLevel() {
         return StandardCompressionOptions.gzip().compressionLevel();
     }
+
+    /**
+     * @return The maximum size of data that can be encoded using the zstd algorithm.
+     */
+    int getMaxZstdEncodeSize();
```
yawkat, line 46: Please add a default implementation for compatibility

yawkat (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpCompressionStrategy.java
glorrian: it already there

sdelamo (commented):
Could we add a test?

glorrian:
> Could we add a test?

I'm not sure about this dependency, but formally we can 
> I can't write tests because its need to add a library https://github.com/luben/zstd-jni and i can't do it, because it's not in netty-bom and I can't connect it transitively because it's declared with optional=true in netty poms, I'm not sure about its license, so I didn't add it to libs.versions.toml

glorrian:
Also i've tested it manually and it works on my local but i didn't run any auto tests

glorrian:
@sdelamo

PR #11369, opened 2024-11-22
feat: allow endpoint path customization

sdelamo:
For example, [kamal](https://kamal-deploy.org) expects health check to be in /up

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/EndpointConfiguration.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/EndpointConfiguration.java
@@ -17,6 +17,7 @@ package io.micronaut.management.endpoint;
 
 import io.micronaut.context.annotation.EachProperty;
 import io.micronaut.context.annotation.Parameter;
+import io.micronaut.core.annotation.Nullable;
 
 import java.util.Optional;
 
@@ -37,6 +38,7 @@ public class EndpointConfiguration {
 
     private Boolean enabled;
     private Boolean sensitive;
+    @Nullable
     private String path;
     private final String id;
 
@@ -101,8 +103,8 @@ public class EndpointConfiguration {
      * Endpoint's path. If not set the endpoint name is used as the path.
      * @param path Endpoint's path
      */
-    public void setPath(String path) {
-        this.path = path.startsWith(SLASH)
+    public void setPath(@Nullable String path) {
+        this.path = path != null && path.startsWith(SLASH)
             ? path.substring(1)
             : path;
     }
@@ -111,6 +113,7 @@ public class EndpointConfiguration {
      * Endpoint's path. If not set the endpoint name is used as the path.
      * @return Endpoint's path
      */
+    @Nullable
     public String getPath() {
         return path;
     }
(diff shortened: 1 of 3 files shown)


management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
```diff
@@ -132,10 +134,17 @@ protected Optional<String> resolveActiveEndPointId(Class<?> declaringType) {
...
+                        String path = beanContext.getBean(EndpointConfiguration.class, Qualifiers.byName(id)).getPath();
+                        if (StringUtils.isNotEmpty(path)) {
+                            return Optional.of(path);
+                        }
+                    }
                     if (id == null || !ENDPOINT_ID_PATTERN.matcher(id).matches()) {
                         id = NameUtils.hyphenate(beanDefinition.getName());
                     }
 
+
```
sdelamo:
```suggestion
```

management/src/main/java/io/micronaut/management/endpoint/EndpointConfiguration.java
```diff
@@ -94,4 +96,22 @@ public void setEnabled(Boolean enabled) {
     public void setSensitive(Boolean sensitive) {
         this.sensitive = sensitive;
     }
+
+    /**
+     * Endpoint's path. If not set the endpoint name is used as the path.
+     * @param path Endpoint's path
+     */
+    public void setPath(String path) {
```
yawkat: Nullability annotations

yawkat (approved):

management/src/main/java/io/micronaut/management/endpoint/EndpointConfiguration.java
graemerocher: also add `@since`

management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
```diff
@@ -132,6 +134,12 @@ protected Optional<String> resolveActiveEndPointId(Class<?> declaringType) {
                 BeanDefinition<?> beanDefinition = opt.get();
                 if (beanDefinition.hasStereotype(Endpoint.class)) {
                     String id = beanDefinition.stringValue(Endpoint.class).orElse(null);
+                    if (beanContext.containsBean(EndpointConfiguration.class, Qualifiers.byName(id))) {
```
graemerocher: probably better to use `beanContext.getProvider` and then `isPresent()` followed by `get()`

graemerocher (changes requested):

management/src/main/java/io/micronaut/management/endpoint/processors/AbstractEndpointRouteBuilder.java
sdelamo: why is better? Is it more performant to retrieve a bean via the `BeanProvider` API vs the `BeanContext` API?

sdelamo: I changed to use BeanProvider here https://github.com/micronaut-projects/micronaut-core/pull/11369/commits/85076fe304e9252bbe4d290e1e524682fd94301b

management/src/main/java/io/micronaut/management/endpoint/EndpointConfiguration.java
sdelamo: done

graemerocher (approved):

PR #11374, opened 2024-11-23
Add header-less methods to MessageBodyReader and ChunkedMessageBodyReader 

glorrian:
I think the interfaces we have now are too demanding. Most of the implemented methods of these interfaces don't actually use the headers param, so I think we should provide new interfaces without the headers param.

We also added `MessageBodyReader` as a replacement for`MediaTypeCodec`, but we will not be able to replace MediaTypeCodec interface such as `<T> T decode(Argument<T> type, InputStream inputStream)` method because in many cases where these interface are used there are no headers instance and we also cannot pass it as null reference because headers param is marked with an annotation by `@NonNull`

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/body/NettyByteBufMessageBodyHandler.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/body/NettyByteBufMessageBodyHandler.java
@@ -62,12 +62,22 @@ public final class NettyByteBufMessageBodyHandler implements TypedMessageBodyHan
     }
 
     @Override
+    public Publisher<ByteBuf> readChunked(Argument<ByteBuf> type, MediaType mediaType, Publisher<ByteBuffer<?>> input) {
+        return Flux.from(input).map(bb -> (ByteBuf) bb.asNativeBuffer());
+    }
+
+    @Override
     public ByteBuf read(Argument<ByteBuf> type, MediaType mediaType, Headers httpHeaders, ByteBuffer<?> byteBuffer) throws CodecException {
         return (ByteBuf) byteBuffer.asNativeBuffer();
     }
 
     @Override
     public ByteBuf read(Argument<ByteBuf> type, MediaType mediaType, Headers httpHeaders, InputStream inputStream) throws CodecException {
+        return read(type, mediaType, inputStream);
+    }
+
+    @Override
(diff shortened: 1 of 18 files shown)


glorrian:
@graemerocher @yawkat @dstepanov @sdelamo i'll glad to receive your reviews

glorrian:
and also i don't know why the build is failed. its smth with checkstyle but i checked it all on my local and I dug logs and it all has no errors from files that I changed. all tests passed so I don't have a single clue what's wrong....

yawkat:
What is your use case?

glorrian:
@yawkat The usecase is that I decided to see how I could replace deprecated MediaTypeCodec with a new MessageBodyReader class in the internal code and found out that bodyreader cannot be used in many cases because it requires ‘@NonNull headers’  which typecodec did not require before. So if we want to get rid of typecodec and use bodyreader, we logically need a new method.

yawkat:
What are you using MediaTypeCodec for?

glorrian:
> What are you using MediaTypeCodec for?

You can look at jackson impl of the messageBodyReader and its not using headers param at all for example. I just mean that codec interface was like decode(type, bytes) and if we want replace it with messagebodyreader we need same interface like read(type, @nullable mediatype, bytes)

glorrian:
POC of this pr it is a fact that we will NOT be able to replace deprecated mediatypedecoder until we provide an interface that does not require the use of headers, simply because in most usecases headers do NOT exist when we use mediatypecodec.

yawkat:
I understand that MessageBodyReader is not a perfect substitute for MediaTypeCodec. MessageBodyReader was designed *specifically* for HTTP messages. There are in-framework use cases where this approach does not work (multipart, specifically), and I am considering additional APIs for those cases. For that I want to understand what *you* are using it for.

glorrian:
I got you. But in the code MediaTypeCodec is marked as deprecated for removal, so we need the replacement to it. If messagebodyreader is not fits to this usecase maybe we need especially new interface?

dstepanov:
You can use `new SimpleHttpHeaders` if you don't have headers

dstepanov:
You can create your own interface that adds a default method with `new SimpleHttpHeaders` and introduces a method without headers

glorrian:
@dstepanov can i create a new pr with a new interface without headers?

dstepanov:
I think you can have it as an internal one for your project

Issue #11375, opened 2024-11-23
Preflight requests handled by `@Filter`s blocked by CorsFilter


ThomasIommi:
Hello!

As mentioned in issue #7934, it seems that the router might not recognize paths defined using patterns like those used by `@Filter`s. 
I'm afraid that checking for known paths reintroduce the same issue, causing all requests to a proxy filter to be blocked by the CORS filter with a 403 Forbidden error.

This situation seems to prevent us from implementing solutions suggested in the official documentation, such as the one found here: https://docs.micronaut.io/latest/guide/#proxyClient

Here's a snippet of the code in question:

```java
@Nullable
private boolean validatePreflightRequest(@NonNull HttpRequest<?> request,
                                         @NonNull CorsOriginConfiguration config) {
    ...
    List<HttpMethod> availableHttpMethods = router.findAny(request).stream().map(UriRouteMatch::getHttpMethod).toList();
    if (availableHttpMethods.stream().noneMatch(method -> method.equals(methodToMatch))) {
        return false;
    }    
    ...
    return true;
}
```

_Originally posted by @ThomasIommi in https://github.com/micronaut-projects/micronaut-core/issues/11097#issuecomment-2495529232_

yawkat:
The core issue is that filters are not recognized by the CORS filter, and the solution will be the same (if there is one), so I'm closing this as duplicate of #7934.

PR #11376, opened 2024-11-23
Fix preflight failure on unknown router paths

ThomasIommi:
Hi!

I've tried to propose a fix for this issue #11375 and to update tests.
Let me know if there are any suggestions or feedback. I'm open to any improvements or discussions.

Thanks!

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -314,21 +314,6 @@ class CorsFilterSpec extends Specification {
         HttpStatus.OK == response.status()
     }
 
-    @Property(name = "micronaut.server.cors.configurations.foo.allowed-origins", value = "http://www.foo.com")
-    @Property(name = "micronaut.server.cors.configurations.foo.allowed-methods", value = "GET")
-    void "A preflight request is rejected for a route that does exist but doesn't handle the requested HTTP Method"() {
-        given:
-        HttpRequest request = HttpRequest.OPTIONS("/example")
-                .header(HttpHeaders.ORIGIN, 'http://www.foo.com')
-                .header(HttpHeaders.ACCESS_CONTROL_REQUEST_METHOD, 'POST')
-
-        when:
-        HttpResponse<?> response = execute(request)
-
-        then:
-        HttpStatus.FORBIDDEN == response.status()
-    }
-
     @Requires(property = "spec.name", value = "CorsFilterSpec")
     @Controller
     static class TestController{
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -301,7 +301,7 @@ class CorsFilterSpec extends Specification {
 
     @Property(name = "micronaut.server.cors.configurations.foo.allowed-origins", value = "http://www.foo.com")
     @Property(name = "micronaut.server.cors.configurations.foo.allowed-methods", value = "GET")
-    void "A preflight request is rejected for a non-existing route"() {
+    void "A preflight request is NOT rejected for a non-existing route if CORS configuration is valid"() {
         given:
         HttpRequest request = HttpRequest.OPTIONS("/doesnt-exists-route")
                 .header(HttpHeaders.ORIGIN, 'http://www.foo.com')
@@ -311,7 +311,7 @@ class CorsFilterSpec extends Specification {
         HttpResponse<?> response = execute(request)
 
         then:
-        HttpStatus.FORBIDDEN == response.status()
+        HttpStatus.OK == response.status()
     }
 
     @Property(name = "micronaut.server.cors.configurations.foo.allowed-origins", value = "http://www.foo.com")
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
@@ -315,7 +315,7 @@ class CorsFilterSpec extends Specification {
     }
 
     @Property(name = "micronaut.server.cors.configurations.foo.allowed-origins", value = "http://www.foo.com")
-    @Property(name = "micronaut.server.cors.configurations.foo.exposed-headers", value = "Foo-Header,Bar-Header")
+    @Property(name = "micronaut.server.cors.configurations.foo.allowed-methods", value = "GET")
     void "A preflight request is rejected for a route that does exist but doesn't handle the requested HTTP Method"() {
         given:
         HttpRequest request = HttpRequest.OPTIONS("/example")
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsVersionSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsVersionSpec.groovy
@@ -112,7 +112,7 @@ class CorsVersionSpec extends Specification {
 
         when:
         request = HttpRequest.OPTIONS("/new-not-allowed-from-private")
-        preflightHeaders(null, true).each { k, v -> request.header(k, v)}
+        preflightHeaders("x-api-version", true).each { k, v -> request.header(k, v)}
         client.exchange(request)
 
         then:
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsVersionSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsVersionSpec.groovy
@@ -95,8 +95,7 @@ class CorsVersionSpec extends Specification {
         client.exchange(request)
 
         then:
-        HttpClientResponseException ex = thrown()
-        ex.status == HttpStatus.FORBIDDEN
+        noExceptionThrown()
     }
 
     void "preflight for version routed from private network"() {
--- a/http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
+++ b/http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
@@ -89,7 +89,7 @@ public class CorsFilter implements Ordered, ConditionalFilter {
     /**
      * @param corsConfiguration The {@link CorsOriginConfiguration} instance
      * @param httpHostResolver  HTTP Host resolver
-     * @deprecated use {@link CorsFilter(HttpServerConfiguration, HttpHostResolver, Router)} instead.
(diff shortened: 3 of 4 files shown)


glorrian:
what the motivation for this pr? i think server shouldn't response 200 on non existing route

ThomasIommi:
Hi!

In the current implementation, OPTIONS requests are being blocked with `403 Forbidden` errors, even when the routes do exist but are not explicitly known by the router. This occurs, for example, in scenarios where the server acts as a proxy using a `@Filter`, as outlined in the official documentation ((https://docs.micronaut.io/latest/guide/#proxyClient)).

I believe (though I might be mistaken) that it’s not the responsibility of the CorsFilter to block such requests at this stage. The original CORS request should be blocked if the server fails to handle it, not the preflight OPTIONS request.

I encountered this issue after upgrading from Micronaut 4.6.X to 4.7.X and traced it back to this PR: #11097.

Let me know if you disagree. If there’s another approach to account for routes handled by `@Filters`, I’d be happy to adjust this PR accordingly.

glorrian:
I am confused by the tests that you deleted "A preflight request is rejected for a non-existing route" and "A preflight request is rejected for a route that does exist but doesn't handle the requested HTTP Method" because they must pass. if you fix some functionality, you should at least leave all previous tests valid, but it's better to add new ones

ThomasIommi:
Have you reviewed the PR I mentioned (#11097)?

I believe it introduced a regression. Additionally, there were changes to the tests in the PR itself, and I don't think I can make those tests pass since I'm essentially saying they don't represent the correct behavior.

I'm specifically referring to parts like this one:

image

Sorry for the screenshot I couldn't figure out how to link to this diff :sweat_smile:.

glorrian:
yeah, thanks to the link on pr, i got you.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsVersionSpec.groovy
```diff
@@ -113,7 +112,7 @@ class CorsVersionSpec extends Specification {
 
         when:
         request = HttpRequest.OPTIONS("/new-not-allowed-from-private")
-        preflightHeaders(null, true).each { k, v -> request.header(k, v)}
+        preflightHeaders("x-api-version", true).each { k, v -> request.header(k, v)}
```
glorrian, line 115: what the reason of "x-api-version" here?

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -494,20 +494,15 @@ private MutableHttpResponse<?> handlePreflightRequest(@NonNull HttpRequest<?> re
 
     @Nullable
     private boolean validatePreflightRequest(@NonNull HttpRequest<?> request,
-                                                @NonNull CorsOriginConfiguration config) {
+                                             @NonNull CorsOriginConfiguration config) {
         Optional<HttpMethod> methodToMatchOptional = validateMethodToMatch(request, config);
         if (methodToMatchOptional.isEmpty()) {
```
glorrian:
```
if (validateMethodToMatch(request, config).isEmpty() ||
            !CorsUtil.isPreflightRequest(request) ||
            !hasAllowedHeaders(request, config)) {
            return false;
        }
```

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsVersionSpec.groovy
ThomasIommi: While debugging, I discovered that without the "Access-Control-Request-Headers" header, the test failed to target the correct `TestController` endpoint (annotated with `@CrossOrigin(allowPrivateNetwork = false)` and `@Version("2")`).

Before the proposed changes, I believe the test passed for the wrong reason — the `403 FORBIDDEN` response occurred because the route was not found, not due to the `@CrossOrigin(allowPrivateNetwork = false)` configuration.

I think the issue is related to the `@Version("2")` annotation since temporarily removing this annotation resolved the problem, but I then decided not to remove the annotation and add the header instead.

glorrian: ive seen to the code. this behaviour happens by the reason that matchesOrigin method at CorsFilter returns false if we defined Version annotation but request will not support this header cause accessControlRequestHeaders is null and then it responce 200 cause all cors requests return 200 by this pr. so i guess it test right, thanks

glorrian (approved):

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
```diff
@@ -311,22 +311,7 @@ class CorsFilterSpec extends Specification {
         HttpResponse<?> response = execute(request)
 
         then:
-        HttpStatus.FORBIDDEN == response.status()
-    }
-
-    @Property(name = "micronaut.server.cors.configurations.foo.allowed-origins", value = "http://www.foo.com")
-    @Property(name = "micronaut.server.cors.configurations.foo.exposed-headers", value = "Foo-Header,Bar-Header")
-    void "A preflight request is rejected for a route that does exist but doesn't handle the requested HTTP Method"() {
```
yawkat, line 319: We need this feature.

yawkat:
I'm not sure I like this change. It is not against spec from what I can tell, but it does weaken the filtering we have.

A simple workaround would be to make your filter pre-routing as well, since you clearly don't need route information.

Other options for this PR would be to either add a config option to disable CORS preflight handling for some paths, or a config option to disable the route requirement entirely like you do in this PR.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CorsFilterSpec.groovy
ThomasIommi: Perhaps we could handle it as a special case: if the route exists, we perform an additional check on the HTTP methods it supports. However, if the route is not found, we make no assumptions. Honestly, I find this approach a bit trickier. For me, it’s simpler to treat it as if the route it is not known / found.

dstepanov:
I think we should extract the `is route exists` logic to some interface and allow the addition of an extra checker.

ThomasIommi:
@yawkat 
Doesn't the workaround you propose still invalidate the use of `@Filter` to implement a gateway proxy?

In a similar scenario to this use case (https://docs.micronaut.io/latest/guide/#proxyClient), if I move the proxy filter before the CORS filter and find a way to bypass it, I believe I would then need to handle CORS in every backend behind the gateway. That doesn’t seem ideal. I hope this makes sense, and that I'm not overlooking something.

I still think the responsibility of the CORS filter for preflight requests is to simply validate the CORS configuration, not the routes. Of course, I could be wrong, but I thought we were on the same page based on the comments in the PR I'm trying to address; did I misunderstand, or has something changed your perspective?

Just curious! Maybe I'm missing some background. :stuck_out_tongue:

ThomasIommi:
I like the idea of using a configuration to enforce or disable route requirements. I can try to update the code accordingly as soon as possible.

dstepanov:
I would not disable validation; I would add a new check to validate the route's validity.

yawkat:
Sorry, I don't think changing CORS validation to accommodate filters is a good idea.

ThomasIommi:
It still feels to me that the CORS filter (as it was when this PR was opened) violates the SRP by handling both CORS filtering AND a pre-route check, without allowing the option to disable only one of these functionalities.

It's been a while since I've followed this topic, so I'm not sure if things have changed in the meantime.

I also wanted to apologize. I had hoped to be able to focus and make the requested changes, but I couldn't due to other things going on and my lack of experience in adding a configuration without disrupting the constructor's compatibility (though maybe I'm just being paranoid about that :sweat_smile:).

Anyway, I still think that @dstepanov's proposal to separate these two concerns is really good. Is there any chance it could still be considered?

By the way, you never really replied to this comment, and I'm still curious about your opinion and whether I'm missing something.

> @yawkat Doesn't the workaround you propose still invalidate the use of `@Filter` to implement a gateway proxy?
> 
> In a similar scenario to this use case (https://docs.micronaut.io/latest/guide/#proxyClient), if I move the proxy filter before the CORS filter and find a way to bypass it, I believe I would then need to handle CORS in every backend behind the gateway. That doesn’t seem ideal. I hope this makes sense, and that I'm not overlooking something.
> 
> I still think the responsibility of the CORS filter for preflight requests is to simply validate the CORS configuration, not the routes. Of course, I could be wrong, but I thought we were on the same page based on the comments in the PR I'm trying to address; did I misunderstand, or has something changed your perspective?
> 
> Just curious! Maybe I'm missing some background. 😛

dstepanov:
I'm fine with extracting the logic that checks if route exists and allowing a separate implementation.

yawkat:
The purpose of the CORS filter is to control access to routes. Knowledge of the target route is important in order to:

- properly honor route-specific CORS information
- properly respond with route metadata (`Access-Control-X`, for preflight)

This information is necessary for both the normal request handling, and the preflight request handling. For preflight handling, we also need to be in the pre matching stage, because the request method may not match that of the route (`Access-Control-Request-Method`). I don't think these are really separate issues, though the fact that they are conceptually so far apart (one pre-routing, one post-routing) is not great. Maybe that alone justifies some separate configuration for the pre-flight handling.

The approach we take to get this route information is not very flexible and does not accommodate filters that may programmatically "emulate" a route, such as a proxy filter. I can't think of an easy way to fix this yet, and I don't know if there will be one. Maybe there should be a third filter type (beyond RequestFilter and ResponseFilter) that can intercept route metadata without worrying about side effects from normal request execution.

What this PR does is modify the cors filter in a fairly minor way to "carve out" a workaround for filters behind the cors filter to use. I'm not sure about whether the implications of this on normal cors handling are a big deal, but in the end, my issue with it is that it's just a workaround. That's what I mean by "changing CORS validation to accommodate filters". It does not solve the fundamental issue.

And yes, until there is a broader solution to making "filter routes" visible to the cors filter, that means that proper cors-capable filters:

- need to be pre-routing, which is not the default, so may be suprising to some folks, and 
- need to implement preflight responses themselves (or delegate them to the proxy target or whatever)

ThomasIommi:
Of course I agree that preflight checks should examine routes for specific CORS configurations. However, I still disagree with returning a 404 (or 403) for an OPTIONS request simply because a route isn't found during this specific pre-route check, since, as we've noted, the pre-route check doesn't have complete visibility and doesn't take into account filter routes. It acts as a circuit breaker and effectively prevents the browser from dispatching the original request entirely, potentially leading to premature rejection even when a generic CORS configuration is present, for example in the application properties.

To clarify, my pull request was submitted with the best intentions, and was not intended as a workaround solely for my specific use case. I even researched how Spring handles this to ensure its potential validity.

Anyway thank you for your response and for providing clarification. I haven't had the opportunity yet, but if I need to update to a version beyond 3.6.x, I will explore a local workaround, potentially by reimplementing the CORS filter or by contributing to this project by attempting to make the pre-route check aware of filters, if feasible.

PR #11378, opened 2024-11-24
Add ability to configure NettyThreadFactory and set `daemon` and `priority` flags

### LINKED ISSUE #6102: Configurable NettyThreadFactory daemon parameter
### Feature description

I would like to configure an application with daemon = false.

I used Micronauts to generate HTTP client and it prevents JVM from shutting down without daemon = false.  For now, I override NettyThreadFactory bean.

altro3:
Fixed #6102


altro3:
@yawkat Review it, pls

yawkat:
Thanks!

PR #11379, opened 2024-11-24
Fix check for configuration properties

### LINKED ISSUE #10943: @ConfigurationProperties not functioning correctly on nested configuration when using KSP
### Expected Behavior

The @ConfigurationProperties annotation should apply to both the piece of configuration specified and any configuration nested under it.

### Actual Behaviour

If we have a config item uses @ConfigurationProperties to override what the property is called in the application.yml (ie to something that wouldn't be the default) then it will be correctly set, but any values nested underneath this item will not be set. Strangely this does not affect nested objects, just nested values.

If an additional config line is added to the application.yml that matches the name of the property then the values will be successfully looked up, but from the overridden path, not the path that matches the name. If the path that matches the name is removed then the lookup will fail.

### Steps To Reproduce

Reproduction steps are available in the README of the example application below, but generally:

1. Define a 2 level configuration structure in the application.yml file like the following: 
 ```
example-project:
  level-one:
    level-one-value: ONE
    level-two:
      levelTwoValue: TWO
```
2. Create a corresponding configuration class in the test directory like the following:
```
@ConfigurationProperties("example-project")
class Configuration {
    @Inject internal var levelOne = LevelOne()
    @ConfigurationProperties("level-one")
    internal class LevelOne {
        internal lateinit var levelOneValue: String
        @Inject internal var levelTwo = LevelTwo()
        @ConfigurationProperties("level-two")
        internal class LevelTwo {
            lateinit var levelTwoValue: String
        }
    }
}
```
3. Create a unit test that uses the @MicronautTest annotation, injects the configuration and prints the values of levelOneValue and levelTwoValue.
4. Run the test to observe that it passes.
5. Refactor the var levelOne to be called something else i.e notLevelOne, this should not affect anything as @ConfigurationProperties should point to the correct piece of configuration in the .yml file.
6. Rerun the test and levelOneValue and levelTwoValue will not have been initialised.

### Environment Information

- Kotlin 2.0
- JDK 17

### Example Application

https://github.com/cainecridland/micronaut-nested-config-problem-example

### Version

4.4.0

altro3:
Fixes #10943

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/beans/MapperIntroduction.java
+++ b/context/src/main/java/io/micronaut/runtime/beans/MapperIntroduction.java
@@ -563,7 +563,7 @@ final class MapperIntroduction implements MethodInterceptor<Object, Object> {
         @NonNull Argument<?>[] getBuilderArguments();
 
         /**
-         * Get the argument index based on its name
+         * Get the argument index based on its name.
          *
          * @param name The argument name
          * @return The index
--- a/core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationMetadataBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationMetadataBuilder.java
@@ -178,6 +178,11 @@ public class ConfigurationMetadataBuilder {
         metadata.name = name;
         metadata.path = propertyType.stringValue(ConfigurationReader.class, ConfigurationReader.PREFIX)
             .orElseGet(() -> NameUtils.hyphenate(buildPropertyPath(owningType, declaringType, name), true));
+        if (propertyType.hasStereotype(ConfigurationReader.class)) {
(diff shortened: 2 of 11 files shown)


altro3:
@dstepanov look to this, pls. I've added tests for inject-java and inject-kotlin. As you see, test with kotlin (KSP) is failed. Wrong introdpection `@ConfigurationProperties` bean. `injectibleFields` list is empty, bust must contains `levelOnez` property. With inject java all works fine

altro3:
@dstepanov It doesn't look like it's a checkstyle error :-)

{036D2E05-CDDF-4EF4-AA24-06A9D48A72E4}

altro3:
@dstepanov looks like the `4.8.x` branch is broken

dstepanov:
Looks like my fix is not correct

graemerocher (approved):

PR #11381, opened 2024-11-25
Remove LocalDate and LocalDateTime converters

radovanradic:
I think this is causing the issue in micronaut-data https://github.com/micronaut-projects/micronaut-data/issues/2096 and is discussed here https://github.com/micronaut-projects/micronaut-data/discussions/3241


context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
```diff
@@ -209,8 +209,8 @@ public void register(MutableConversionService conversionService) {
         addTemporalToDateConverter(conversionService, ZonedDateTime.class, ZonedDateTime::toInstant);
         // these two are a bit icky, but required for yaml parsing compatibility
         // TODO Micronaut 4 Consider deletion
-        addTemporalToDateConverter(conversionService, LocalDate.class, ld -> ld.atTime(0, 0).toInstant(ZoneOffset.UTC));
```
radovanradic, line 212: This line is converting `LocaleDate` to `Date` using `UTC` and micronaut-data is using default timezone when converting back so we are getting date diff issues. I believe this should use default timezone as well. 
There is also comment about removal of this conversion in Micronaut4. Can it be removed? If yes, then we can add converter needed in micronaut-data using default timezone and the issue would be fixed.

radovanradic: @yawkat Can we remove these conversions above, or change it to not use `UTC` but default timezone? It is causing issue in micronat-data explained in the description of this PR.

yawkat: I disagree that this should be default TZ. Yes they should be removed, but using system TZ is worse.

Why does data need to convert these at all? java.sql.Date has actual well-defined conversions from and to LocalDate

radovanradic: micronaut-data registers various conversions for dates and is always using timezones.
With this converter defined in micronaut-core, even if we register this in micronaut-data
```
conversionService.addConverter(java.sql.Date.class, LocalDate.class, date -> date.toLocalDate());
```
we will get wrong date when converted back.

I removed these converters from micronaut-core, hopefully nothing will break.

yawkat:
So the java.util.Date converter is taking priority over the java.sql.Date converter, even when using java.sql.Date? Seems like a bug in ConversionService

radovanradic:
> So the java.util.Date converter is taking priority over the java.sql.Date converter, even when using java.sql.Date? Seems like a bug in ConversionService

No, I think this micronaut-data block is causing that
```
if (value instanceof Date date) {
                    return setDate(statement, index, date);
                } else {
                    return setDate(statement, index, convertRequired(value, Date.class));
                }
```
it is converting to `java.util.Date` and not `java.sql.Date` and this is why it finds this converter in core.

yawkat:
Then data should convert to the right type, or add special cases where it needs different conversions from core. It should not require a core change

radovanradic:
> Then data should convert to the right type, or add special cases where it needs different conversions from core. It should not require a core change

Yes, I will close this PR as not needed and fix issue in micronaut-data. Thanks

PR #11387, opened 2024-11-28
Add correct message, when set duration / period value in wrong format.

### LINKED ISSUE #11385: TimeConverterRegistrar silenly ignores incovertible config properties on injection
### Expected Behavior

A message like the following should appear in the log/exception messages:
```
Unable to convert value [60 seconds] to target type [Duration] due to: Unparseable date format (60 seconds). Should either be a ISO-8601 duration or a round number followed by the unit type
```
which is informative & points out the root cause.

### Actual Behaviour

The following message/exception apears:
```
Error resolving property value [${health-check-cache-expiration}]. Property doesn't exist 
```

The `TimeConverterRegistrar` rejects other malformed specifications, like `60us` with an appropriate message, but completely misformatted values are silently ignored and misreported as missing. Other malformed inputs, like  integers in i.e. `micronaut.server.port` report a conversion error / invalid property format.

I believe the reason is that when `TimeConverterRegistrar` rejects the text with the initial regexp check, it does not call `context.reject()` unlike the other validations.

`TimeConverterRegistrar` also accepts some malformed inputs without any errors:
- 60hs - interprets as 60 hours
- 60ds - interprets as 60 days

### Steps To Reproduce

1. Have a configuraton bean  with 
```
@Value("${health-check-cache-expiration}")
private Duration healthCheckCacheExpiration;
```

2. Have an application with 
```
healthCheckCacheExpiration: 60 seconds
```
(converting a service from a dropwizard-based implementation, so different conversion formats were used, and remained in the config)

3. attempt to inject (otherwise materialize) the configuration bean, so the value is injected and converted from the configuration.

### Environment Information

- Ubuntu Linux 22.04
- OpenJDK Runtime Environment (build 21+35-2513)

### Example Application

_No response_

### Version

4.6.5

altro3:
Fixed #11385

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -355,13 +355,19 @@ public class TimeConverterRegistrar implements TypeConverterRegistrar {
                                 }
                                 context.reject(
                                     value,
-                                    new DateTimeParseException("Unparseable date format (" + value + "). Should either be a ISO-8601 duration or a round number followed by the unit type", value, 0));
+                                    new DateTimeParseException("Unparseable date format (" + value + "). Should either be a ISO-8601 duration or a round number followed by the unit type", value, 0)
+                                );
                                 return Optional.empty();
                             }
                         }
                     } catch (NumberFormatException e) {
                         context.reject(value, e);
                     }
+                } else {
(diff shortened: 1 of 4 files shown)


graemerocher:
@altro3 maybe this should target 4.7.x thanks

graemerocher (approved):

altro3:
This bug is clearly not critical, I think it's okay if we leave it until release 4.8.0

PR #11390, opened 2024-11-29
Fix processing HttpClientResponseException with custom Http Status codes

### LINKED ISSUE #11383: HttpClient should not fail with non-standard http codes
### Expected Behavior

I have to call a rest service that sometimes answer with weird http code like 551.
I expect to get a HttpClientResponseException with getResponse().code() == 551 and the body of the error.

### Actual Behaviour

Some part of code handle this HTTP 551 wrong (in my opinion).

[This line](https://github.com/micronaut-projects/micronaut-core/blob/4.7.x/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L622) produce an IllegalArgumentException because getStatus call HttpStatus.valueOf. And then the original HttpClientResponseException is lost.

Same thing [here](https://github.com/micronaut-projects/micronaut-core/blob/4.7.x/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java#L622).

Since the goal is to verify if the exception comes from an HTTP 404, the test could be `HttpStatus.NOT_FOUND.getCode() == ex.getResponse().code()` instead of using getStatus

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.6.5

altro3:
Fixed #11383

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
@@ -288,7 +288,7 @@ public class HttpClientIntroductionAdvice implements MethodInterceptor<Object, O
                 }
 
                 if (t instanceof HttpClientResponseException e) {
-                    if (e.getStatus() == HttpStatus.NOT_FOUND) {
+                    if (e.code() == HttpStatus.NOT_FOUND.getCode()) {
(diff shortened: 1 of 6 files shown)


graemerocher:
can you add a test?

altro3:
@graemerocher done

sdelamo (approved):

PR #11402, opened 2024-12-03
Fix for ServiceLoaderFeature

andriy-dmytruk:
`ServiceLoaderFeature` did not include all the `@Requires` annotations, as some were added with the `@Requirements`. I included such in the native-image `beforeAnalysis`. As an example, this fixes the usage of condition in `io.micronaut.configuration.metrics.binder.web.ClientRequestMetricRegistryFilter`.


graemerocher:
@andriy-dmytruk could you target 4.7.x?

graemerocher (approved):

PR #11405, opened 2024-12-03
Improve the dependency path exception message

andriy-dmytruk:
This improves the readability of message for long paths, large class names or classes with many parameters, as each class from the path is immediately visible. I changed the tests to use longer class names, which is more realistic.

The type names are also in fully-qualified shortened form, which makes them clickable in IDEs.

As a comparison this message
```java
Message: No bean of type [io.micronaut.inject.failures.NestedDependencyFailureSpec$MyClassD] exists. 
| Path Taken: new MyClassB() --> MyClassB.propA --> new MyClassA([MyClassC propC]) --> new MyClassC([MyClassD propD])
```
will change to
```java
Message: No bean of type [io.micronaut.inject.failures.NestedDependencyFailureSpec$MyClassD] exists.
Path Taken:
new i.m.i.f.N$MyClassB()
\---> i.m.i.f.N$MyClassB#propA
      \---> new i.m.i.f.N$MyClassA([MyClassC propC])
            \---> new i.m.i.f.N$MyClassC([MyClassD propD])
```

Part of the change is similar to visualization change for circular dependencies: https://github.com/micronaut-projects/micronaut-core/pull/11299.

### DIFF
--- a/core/src/main/java/io/micronaut/core/naming/NameUtils.java
+++ b/core/src/main/java/io/micronaut/core/naming/NameUtils.java
@@ -18,6 +18,7 @@ package io.micronaut.core.naming;
 import io.micronaut.core.annotation.AccessorsStyle;
 import io.micronaut.core.annotation.Experimental;
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.util.ArgumentUtils;
 import io.micronaut.core.util.StringUtils;
 
@@ -48,7 +49,7 @@ public class NameUtils {
      * @param name The name
      * @return True if it is
      */
-    public static boolean isHyphenatedLowerCase(String name) {
+    public static boolean isHyphenatedLowerCase(@Nullable String name) {
         if (name == null || name.isEmpty() || !Character.isLetter(name.charAt(0))) {
             return false;
         }
@@ -68,7 +69,7 @@ public class NameUtils {
      * @param suffixes The suffix to remove
      * @return The decapitalized name
      */
-    public static String decapitalizeWithoutSuffix(String name, String... suffixes) {
+    public static @NonNull String decapitalizeWithoutSuffix(@NonNull String name, String... suffixes) {
         String decapitalized = decapitalize(name);
         return trimSuffix(decapitalized, suffixes);
     }
@@ -80,7 +81,7 @@ public class NameUtils {
      * @param suffixes The suffixes
      * @return The trimmed string
      */
-    public static String trimSuffix(String string, String... suffixes) {
+    public static @NonNull String trimSuffix(@NonNull String string, String... suffixes) {
         if (suffixes != null) {
             for (String suffix : suffixes) {
                 if (string.endsWith(suffix)) {
@@ -97,7 +98,7 @@ public class NameUtils {
      * @param name The property name
      * @return The class name
      */
-    public static String capitalize(String name) {
+    public static @NonNull String capitalize(@NonNull String name) {
         final String rest = name.substring(1);
 
         // Funky rule so that names like 'pNAME' will still work.
@@ -114,7 +115,7 @@ public class NameUtils {
      * @param name The name
      * @return The hyphenated string
      */
-    public static String hyphenate(String name) {
+    public static @NonNull String hyphenate(@NonNull String name) {
         return hyphenate(name, true);
     }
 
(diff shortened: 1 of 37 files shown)


sdelamo:
maybe this should target 4.8.x

dstepanov:
Consider adding a full class name. I think IntelliJ will provide a link to the class

dstepanov:
How does it look with a factory method / field?

inject-groovy/src/test/groovy/io/micronaut/inject/failures/NestedDependencyFailureSpec.groovy
```diff
@@ -28,44 +28,49 @@ class NestedDependencyFailureSpec extends Specification {
...
-Failed to inject value for parameter [d] of class: io.micronaut.inject.failures.NestedDependencyFailureSpec$C
+        e.message.normalize() == """\
+Failed to inject value for parameter [propD] of class: io.micronaut.inject.failures.NestedDependencyFailureSpec\$MyClassC
 
-Message: No bean of type [io.micronaut.inject.failures.NestedDependencyFailureSpec$D] exists.''')
-        e.message.normalize().contains('Path Taken: new B() --> B.a --> new A([C c]) --> new C([D d])')
+Message: No bean of type [io.micronaut.inject.failures.NestedDependencyFailureSpec\$MyClassD] exists.$space
+Path Taken:$space
+new MyClassB()
+\\---> MyClassB.propA
```
andriy-dmytruk: There is an example of field here

andriy-dmytruk:
> How does it look with a factory method / field?

For factory there is a circular dependency test: https://github.com/micronaut-projects/micronaut-core/blob/da67f65f446a2c75c6e1048139e88d8650332f2b/inject-groovy/src/test/groovy/io/micronaut/inject/failures/FactoryCircularDependencyFailureSpec.groovy#L41-L48

graemerocher:
> Consider adding a full class name. I think IntelliJ will provide a link to the class

this is a good point, with the new lines maybe we can include qualified names. Can you check this link works in IntelliJ?

andriy-dmytruk:
Yes, the classnames are clickable:
image

This is more readable, though:
image

I think both are good. Wdyt?

graemerocher:
@dstepanov WDYT?

dstepanov:
Intellij supports even shortened names, so we can reduce it if it's too large:

try:
```
        System.out.println("io.m.r.Micronaut!");
    }
```

graemerocher:
@andriy-dmytruk can you try that?

andriy-dmytruk:
Yes, this is actually amazing! Now there are actually 2 options:
1. Shorten type owner names: 
    image
  When clicked:
    image

2. Do not shorten:
    image

I would say the first one is the overall best.

dstepanov:
Maybe we can have something like 200-300 characters per name and start reducing when it's more.

graemerocher:
the first one is the best yes

andriy-dmytruk:
I went for using the first form always, as it is shortest, but is still clickable. We can also extend it at a later point if we experience issues with it.

> Maybe we can have something like 200-300 characters per name and start reducing when it's more.

We can start a GitHub discussion, but I think shortest should be good in most cases. The chances of conflict are pretty low. If we do need this, the typical line limits are 80, 100, 120, which would mean around 60 chars limit for the name alone.

core/src/main/java/io/micronaut/core/naming/NameUtils.java
```diff
@@ -214,6 +215,41 @@ public static String getSimpleName(String className) {
...
+     * For example, {@code com.example.Owner$Inner} would become {@code c.e.O$Inner}.
+     * IDEs would still be able to recognize these types, but they would take less space
+     * visually.
+     *
+     * @since 4.8.x
+     * @param typeName The fully-qualified type name
+     * @return The shortened type name
+     */
+    @Experimental
+    public static String getShortenedName(String typeName) {
```
graemerocher: add nullability annotations to return type and arguments

core/src/main/java/io/micronaut/core/naming/NameUtils.java
```diff
@@ -214,6 +215,41 @@ public static String getSimpleName(String className) {
...
+        int nameStart = typeName.lastIndexOf('$');
```
graemerocher, line 232: check type name doesn't == null

graemerocher (approved):
just two small tweaks before merging.

core/src/main/java/io/micronaut/core/naming/NameUtils.java
andriy-dmytruk: That is not checked in other methods, so I assume it is not typical for `typeName` to be `null`. Added `@NonNull` though.

andriy-dmytruk:
@yawkat I don't understand the issue. Can you take a look?

https://ge.micronaut.io/s/szoo56vrxva66
44C5523F-E220-4E09-B7D3-AB5A84F2010C

yawkat:
that test is just flakey. i dont know why

Issue #11410, opened 2024-12-04
Expression Language - Getting bean which has `env` in path


hrothwell:
### Expected Behavior

When using expression langauge coupled with bean ctx retrieval, I expect to be able to access beans such as `io.micronaut.context.env.Environment`

### Actual Behaviour

The project fails to compile stating: 
`[ksp] /path/to/bad/file/File.kt: Failed to parse evaluated expression [#{ ctx[io.micronaut.context.env.Environment].activeNames.contains('test') }]. Cause: Unexpected token: env. Expected: 'IDENTIFIER'`

### Steps To Reproduce

1. try getting an Environment bean from ctx in an expression on a method
2. try to compile the project 

### Environment Information

- tested on JDK 17
- My test branch is on Microanut 4.3.2 and I've also tested 4.6.x, assuming this happens across multiple versions

### Example Application

https://github.com/hrothwell/micronaut-demo/tree/expression-language-environment-bean-issue

[Specific class](https://github.com/hrothwell/micronaut-demo/blob/0826eb3d9d75293d0167fab2a32d4d8b504ae240/src/main/kotlin/com/github/hrothwell/Bean.kt#L7)

### Version

4.3.2

graemerocher:
does `ctx.environment.activeNames` work?

hrothwell:
> ctx.environment.activeNames

`"#{ ctx.environment.activeNames.activeNames.contains('test') }"` does not work also for a similar reasoning:

`Failed to parse evaluated expression [#{ ctx.environment.activeNames.activeNames.contains('test') }]. Cause: Unexpected token: .. Expected: 'L_SQUARE'`

hrothwell:
FWIW I am taking a deeper look into the `SingleEvaluatedExpressionParser` and how it works, my first guess is that there is only one branch/means of processing after matching a specific token for ctx and env

Jarvx:
The following test was generated by AI to help validate and reproduce the behavior described here.
Please note that while the test is automatically generated, it has been verified to compile and execute against the current project (v4.10x).


## Test files(s)


`src/test/kotlin/io/micronaut/expression/EnvInFqcnExpressionKspTest.kt`
```kotlin
package io.micronaut.expression

import io.micronaut.context.ApplicationContext
import io.micronaut.context.annotation.Value
import jakarta.inject.Singleton
import org.junit.jupiter.api.Assertions.assertTrue
import org.junit.jupiter.api.Test

@Singleton
[...]

PR #11417, opened 2024-12-06
Fix env package name in expression langauge type

hrothwell:
Working on fixing #11410 

The root of the issue seems to be from `Tokenizer.getNextToken()` coupled with how `SingleEvaluatedExpressionParser.typeIdentifier()` loops through the identifiers. 

`Tokenizer.getNextToken()` has an order in which it checks for token types, when this is called to find the lookahead it doesn't seem to take into account what the previous token was which might be important for things like this type identifier. 

While the solution of checking for multiple token types specifically for `typeIdentifier()` works, it doesn't feel very optimal. Right now it only checks `IDENTIFIER` or `ENVIRONMENT` for this specific example, but it should likely also include things like `ctx`, or any other expected package naming that could trigger another token type


hrothwell:
I messed up my commits (joys of work laptop vs personal etc) so will close this and reopen with properly signed commits for CLA

hrothwell:
#11418 same PR changes with fixed CLA / commit author mumbo jumbo

PR #11422, opened 2024-12-08
KSP: Fixed method `isEnum()` for KotlinTypeArgumentElement

altro3:


altro3:
@dstepanov Review it, pls. Found this problem in simple test in micronaut-openapi:
```kotlin
@Controller("/hello")
class HelloController {

    @Get
    fun index(@Nullable @QueryValue("channels") channels: Collection<Channel?>?) = ""

    @Introspected
    enum class Channel {
        @JsonProperty("mysys")
        SYSTEM1,
        SYSTEM2
    }
}
```
When I get `typeArgument` for `Collection` type, I get `KotlinTypeArgumentElement` element, and default `isEnum()` implemention is
```java
    default boolean isEnum() {
        return this instanceof EnumElement;
    }
```

That's why method `isEnum()` returns `false`, but must be `true`

graemerocher (approved):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
```diff
@@ -523,6 +523,8 @@ internal open class KotlinClassElement(
         super<AbstractKotlinElement>.isStatic()
     }
 
+    override fun isEnum() = declaration.classKind == ClassKind.ENUM_CLASS
```
dstepanov, line 526: This shouldn't be needed. The fix should properly create a`KotlinEnumElement`

dstepanov (changes requested):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
altro3: I don't quite understand how you propose to do this? Now when I request a generic for a collection, I get an instance of the `KotlinTypeArgumentElement` class - and this is correct, but I can't figure out that this is an enum, because the check for an enum is `this instanceof EnumElement;`

So it's not clear at what point you propose to get `KotlinEnumElement`

dstepanov: Ah in this case, this is correct, but anyway, I think you should always need to call `resolved`

dstepanov (approved):

inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/KotlinClassElement.kt
altro3: Sorry, I didn't understand your last sentence either. Maybe it's because of the holidays :-)

dstepanov: `GenericElement` class has `getResolved`/`resolved` methods to get the actual type represented as generic

altro3: In any case, without this fix, the isEnum method may return incorrect value, as in my case.

PR #11439, opened 2024-12-13
Fix spaces encoding inside DefaultUriBuilder

### LINKED ISSUE #11434: HttpClient incorrectly encodes query params containing spaces with `+` and not `%20`
### Expected Behavior

When using Micronaut's HttpClient to make a request to an endpoint using query params that contain spaces, the spaces are correctly encoded as `%20` per [RFC3986 (URI Generic Syntax)](https://www.rfc-editor.org/rfc/rfc3986). 

### Actual Behaviour

The spaces are encoded as `+`. That's because DefaultUriBuilder [uses java.net.URLEncoder](https://github.com/micronaut-projects/micronaut-core/blob/9cddffcc463db93004afb7c32633e6a2166f7a79/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java#L418) for query parameter encoding, which in turn is based off of the [W3C's HTML 4 specification](https://www.w3.org/TR/html4/) and [subsequently encodes spaces as `+`](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/net/URLEncoder.java#L224-L226)

See [this StackOverflow answer](https://stackoverflow.com/a/47188851/71066) for a better understanding of some of the confusion. For reference, Spring's [UriUtils is correctly based on RFC3986](https://github.com/spring-projects/spring-framework/blob/3db1b94465ae21ecca8adccb9bdc1cfba286a958/spring-web/src/main/java/org/springframework/web/util/UriUtils.java#L305-L318)

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.7.1

glorrian:
fix: #11434
I have implemented a hardcoded solution to address the current issue because I am concerned about completely rewriting the functionality of java.net.URLEncoder, but i think it should be own MK uri encoder like it implemented in [Spring](https://github.com/spring-projects/spring-framework/blob/43ff6d9711540a8ca048867a27fb42db4a232c70/spring-core/src/main/java/org/springframework/util/StringUtils.java#L821)

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
@@ -39,7 +39,8 @@ class ClientFormatSpec extends Specification {
         given:
         var cafe = new Cafe(name: "Pizza Garden", "address": "Home Street 2")
         expect:
-        client.pipesFormattedObject(cafe) == "param=name|Pizza+Garden|address|Home+Street+2"
+        printf "client.pipesFormattedObject(cafe) = %s\n", client.pipesFormattedObject(cafe)
+        client.pipesFormattedObject(cafe) == "param=name|Pizza Garden|address|Home Street 2"
     }
 
     void "test csv format all values"() {
@@ -59,9 +60,9 @@ class ClientFormatSpec extends Specification {
 
         where:
         param                                    | query
-        ["a", "b", "c", "d", "e,f,g"]            | "ssv=a+b+c+d+e,f,g"
-        ["name": "Bob", "age": 8, "height": 120] | "ssv=name+Bob+age+8+height+120"
-        new Cafe(name: "AAA", address: "here")   | "ssv=name+AAA+address+here"
+        ["a", "b", "c", "d", "e,f,g"]            | "ssv=a b c d e,f,g"
+        ["name": "Bob", "age": 8, "height": 120] | "ssv=name Bob age 8 height 120"
+        new Cafe(name: "AAA", address: "here")   | "ssv=name AAA address here"
     }
 
     void "test pipes format all values"() {
--- a/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
+++ b/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
@@ -415,6 +415,7 @@ class DefaultUriBuilder implements UriBuilder {
     }
 
     private String encode(String userInfo) {
-        return URLEncoder.encode(userInfo, StandardCharsets.UTF_8);
+        return URLEncoder.encode(userInfo, StandardCharsets.UTF_8)
+                         .replaceAll("\\+", "%20"); // to match RFC3986
     }
 }
--- a/http/src/test/groovy/io/micronaut/http/uri/UriBuilderSpec.groovy
+++ b/http/src/test/groovy/io/micronaut/http/uri/UriBuilderSpec.groovy
@@ -183,7 +183,7 @@ class UriBuilderSpec extends Specification {
         expect:
         uri == 'myurl?%24top=10&%24filter=xyz'
     }
-    
+
     @Issue("https://github.com/micronaut-projects/micronaut-core/issues/6246")
     void "test uri build parse query param"() {
         given:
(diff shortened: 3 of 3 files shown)


jamesdh:
You might be better off just using [Apache's URIBuilder](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java) rather than hardcoding a mini-workaround or completely reinventing the wheel. URIBuilder gives you the option to [specify how to handle spaces](https://github.com/apache/httpcomponents-core/blob/b435a0b002bb957c8598c9555f6357c1403092dc/httpcore5/src/main/java/org/apache/hc/core5/net/URIBuilder.java#L197-L222), which could be made configurable in case someone prefers Micronauts existing behavior.

glorrian:
firstly, MK doesn't have a dependency on apache httpcore yet, and I'm not sure about its license to add this. also, I am not sure about the behavior of apache decoder in other tests and cases.

glorrian:
I would wait for the review, if the decision on the new encoder is approved, then I will be glad to implement your proposal

jamesdh:
Yea, I realized after commenting that micronaut-core doesn't directly have a dependency on anything apache, so that's probably not viable.

http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
```diff
@@ -39,7 +39,8 @@ class ClientFormatSpec extends Specification {
         given:
         var cafe = new Cafe(name: "Pizza Garden", "address": "Home Street 2")
         expect:
-        client.pipesFormattedObject(cafe) == "param=name|Pizza+Garden|address|Home+Street+2"
+        printf "client.pipesFormattedObject(cafe) = %s\n", client.pipesFormattedObject(cafe)
```
altro3, line 42: I think this is unnecessary

http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
```diff
@@ -415,6 +415,7 @@ private String expandOrEncode(String value, Map<String, ? super Object> values)
     }
 
     private String encode(String userInfo) {
-        return URLEncoder.encode(userInfo, StandardCharsets.UTF_8);
+        return URLEncoder.encode(userInfo, StandardCharsets.UTF_8)
+                         .replaceAll("\\+", "%20"); // to match RFC3986
```
altro3, line 419: You can't just change the framework's behavior. I think the most reasonable solution is to add a setting - how to encode a space. Then you leave backward compatibility and whoever needs it will enable RFC support mode

altro3 (changes requested):

http-client/src/test/groovy/io/micronaut/http/client/format/ClientFormatSpec.groovy
glorrian: I missed it after debug, thanks ;)

http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
glorrian: Agree with you Im gonna add it

glorrian: Agree with you Im gonna add it

jamesdh: It's not a good look for the framework to be handling this incorrectly when competing frameworks get it right. I'd argue it should be changed to the correct behavior, with an option to revert to the prior (incorrect) encoding behavior if needed. The Apache URIBuilder I mentioned in my earlier comment gets this right, defaulting to the correct encoding of `%20` but providing an optional argument for encoding spaces as `+` if needed.

glorrian: @yawkat need your final decision

yawkat: Indeed we can't change it this way, at least not without a config option.

graemerocher:
Replaced by https://github.com/micronaut-projects/micronaut-core/pull/11513

PR #11440, opened 2024-12-13
Fix client backpressure for streaming requests

### LINKED ISSUE #11428: Large memory increase after bumping from 4.6.2 to 4.7.7
### Expected Behavior

No memory increase

### Actual Behaviour

Hello!
It seems there is something that is causing large memory increase after updating to 4.7.x (see the graph below where the high spike is when our pod is rolled out with the new Micronaut version.
Screenshot from 2024-12-11 14-21-07

I believe it's due to something in [this PR](https://github.com/micronaut-projects/micronaut-core/pull/11182).
We started to get a lot of OOM all of a sudden following the upgrade. From my searches (I'm not familiar with the whole Micronaut codebase so maybe I'm wrong), [this change](https://github.com/micronaut-projects/micronaut-core/commit/f74dae1db7405859f1b69dc4cc81b68150e7e06e#diff-2d6e881f8d13084a9ebabcf6d3fdaa88df24d15431df685539574d59d6232548R2018) leads to any content from an HTTP Client that exceeds the maxContentLength to be streamed instead with a **buffer size of maxContentLength**. In our case we've put the value of this prop to Integer.MAX_VALUE (I believe due to some large content we were downloading that was failing the client).
```
private BodySizeLimits sizeLimits() {
    return new BodySizeLimits(Long.MAX_VALUE, configuration.getMaxContentLength());
}
```
Our usecase is that we proxy to a docker registry (simple ProxyHttpClient) and it downloads something more than 2GB.

Playing a bit around and considering the change above, our config was originally setting the `micronaut.http.services.registry.max-content-length` to `2147483647` (Integer.MAX_VALUE) (I believe due to some prior bugs we had. However it seems to me that since we're now streaming and following above, it's taken as the max buffer size.
Lowering this param seems to lower the memory consumption but I had no good result doing a memory dump as Netty uses direct memory...

I know our configuration of max-content-length was a bit awkward but it can be misleading to introduce such behaviour change and the fact that it fallback to such parameter to choose buffer size upon streaming a response seems unclear.

WDYT? Is my analysis correct?

Thank you!

### Steps To Reproduce

a `proxyHttpClient.proxy(httpRequest)` on a content larger than `max-content-length` should fallback to streaming ence using this value as the max buffer size

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.7.7

yawkat:
ProxyBackpressureTest takes quite long so I made it parallel. Wasn't able to do that with junit unfortunately.

Might fix #11428


graemerocher:
I'm not confident with the might part, can we get the reporter to test the branch? Also why were tests deleted in the PR?

graemerocher (approved):

PR #11449, opened 2024-12-16
Support Tracing of Dependency Injection

graemerocher:
Adds the ability to trace the behaviour of dependency and configuration injection. Can be activated a number of different ways, the simplest being an environment variable:

```
MICRONAUT_INJECT_TRACE=.+ ./gradlew run
```

Which enables useful output such as what is the configuration profile of the application, where configuration is read from, what beans are created, how long the bean took to be created, the graph of the bean creation and where it was created from.

To support this, the PR also adds property origin tracking so that one can inspect where a property originated from. 

In addition, some of the changes in `DefaultBeanContext` are in order to reduce stack trace sizes which were starting to get out of hand.


core/src/main/java/io/micronaut/core/type/TypeInformation.java
```diff
@@ -284,11 +310,193 @@ default boolean isArray() {
...
+            @NonNull TypeFormat format,
+            @NonNull Class<?> type,
+            @NonNull Map<String, Argument<?>> generics,
+            @NonNull AnnotationMetadata annotationMetadata) {
+            String typeFormat = TypeFormat.getTypeString(
+                format,
+                type,
+                generics
+            );
+            Optional<String> q = annotationMetadata.getAnnotationNameByStereotype(AnnotationUtil.QUALIFIER)
```
dstepanov, line 456: Maybe you can use similar code to `Qualifiers.forArgument`

inject/src/main/java/io/micronaut/context/ApplicationContextBuilder.java
```diff
@@ -36,6 +36,27 @@
...
 public interface ApplicationContextBuilder {
 
+    /**
+     * Sets the trace mode for bean resolution.
+     * @param traceMode The debug mode
+     * @since 4.8.0
+     * @see BeanResolutionTraceMode
+     * @return This builder
+     */
+    default @NonNull ApplicationContextBuilder traceMode(@NonNull BeanResolutionTraceMode traceMode) {
```
dstepanov: Can this be `beanResolutionTrace(@NonNull BeanResolutionTraceMode traceMode, String... classPatterns)`

inject/src/main/java/io/micronaut/context/BeanResolutionTraceMode.java
```diff
@@ -0,0 +1,433 @@
...
+        String classes = Optional
+            .ofNullable(System.getProperty(CLASSES_SYS_PROP))
+            .orElseGet(() -> System.getenv(CLASSES_ENV_VAR));
+        if (classes != null) {
+            return Set.of(classes.split(","));
+        }
+        return Set.of();
+    }
+
+    void startTrace(
```
dstepanov, line 112: Can you move all the tracing code into some other class, not to have it in a public enum?

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1029,8 +1034,15 @@ public <T> T createBean(@NonNull Class<T> beanType, @Nullable Qualifier<T> quali
         ArgumentUtils.requireNonNull("beanType", beanType);
         Optional<BeanDefinition<T>> candidate = findBeanDefinition(Argument.of(beanType), qualifier);
         if (candidate.isPresent()) {
-            try (BeanResolutionContext resolutionContext = newResolutionContext(candidate.get(), null)) {
-                return doCreateBean(resolutionContext, candidate.get(), qualifier, argumentValues);
+            BeanDefinition<T> beanDefinition = candidate.get();
+            try (BeanResolutionContext resolutionContext = newResolutionContext(beanDefinition, null)) {
+                T bean;
+                if (beanDefinition instanceof InstantiatableBeanDefinition<T> instantiatableBeanDefinition) {
+                    bean = resolveByBeanFactory(resolutionContext, instantiatableBeanDefinition, qualifier, argumentValues);
```
dstepanov: Don't create a new out-scope variable, just call `return postBeanCreated` right after. Next else is not needed. Same pattern repeats a few times

inject/src/main/java/io/micronaut/context/env/Environment.java
```diff
@@ -284,7 +287,9 @@ public interface Environment extends PropertyResolver, LifeCycle<Environment>, M
      * @param name   The name
      * @param values The values
      * @return This environment
+     * @deprecated Use {@link #addPropertySource(PropertySource)}
      */
+    @Deprecated(forRemoval = true)
```
dstepanov: Add since

inject/src/main/java/io/micronaut/context/env/PropertySource.java
```diff
@@ -65,20 +75,69 @@ static PropertySource of(String name, Map<String, Object> map) {
...
+    /**
+     * Create a {@link PropertySource} from the given map.
+     *
+     * @param name       The name of the property source
+     * @param map        The map
+     * @param convention The convention type of the property source
+     * @return The {@link PropertySource}
+     * @deprecated Use {@link #of(String, Map, PropertyConvention, Origin)}
+     */
+    @Deprecated(forRemoval = true)
```
dstepanov: Add since

inject/src/main/java/io/micronaut/context/env/PropertySource.java
```diff
@@ -89,7 +148,9 @@ public PropertyConvention getConvention() {
      * @param values The values as an array of alternating key/value entries
      * @return The {@link PropertySource}
      * @since 2.0
+     * @deprecated Use {@link #of(String, Map, Origin)}
      */
+    @Deprecated(forRemoval = true)
```
dstepanov, line 154: Add since

inject/src/main/java/io/micronaut/context/env/PropertySource.java
```diff
@@ -160,4 +245,30 @@ enum PropertyConvention {
...
+
+    /**
+     * A property entry models
+     * @param property The normalized property key
+     * @param value The property value
+     * @param raw The raw property key that originated the property
+     * @param origin
+     */
+    @Experimental
+    record PropertyEntry(
```
dstepanov: Can you convert it to an interface? I don't like the idea of exposing a record as a public API

dstepanov (changes requested):

dstepanov:
Can you show an example of the output?

dstepanov:
Maybe one of the outputs can be a flame graph like the async profiler output Async profiler output

core/src/main/java/io/micronaut/core/type/TypeInformation.java
graemerocher: code here is in core which can't reference inject

inject/src/main/java/io/micronaut/context/BeanResolutionTraceMode.java
graemerocher: the trace methods are not currently public but package private. In the future yes it would be good to add a public API but for the moment it is package private / internal

graemerocher:
Example Output 

Screenshot 2025-01-08 at 08 45 59

graemerocher:
More complex bean resolution debug output of the creation of the DataSource from Micronaut-SQL

Screenshot 2025-01-13 at 14 41 05

dstepanov (approved):

Issue #10564, opened 2024-03-01
Micronaut is decoding the ‘+’ in  the path variable passed to the controller code


sxsubram:
### Expected Behavior

In my unit test with I have

URI = "/test/1.0/" +
                "024-02-07T00:30:48.014+00:00?time=024-02-07T00:30:48.014+00:00" 

The controller has
`@Get(uris = {“/test/{version}/{+resourcePath}“})`

The resourcePath variable and query param should have the raw format value 
**resource variable path = 024-02-07T00:30:48.014+00:00**
**query param = 024-02-07T00:30:48.014+00:00**
 for downstream processing to work.

### Actual Behaviour

`@Get(uris = {“/test/{version}/{+resourcePath}“})`

In my unit test with 

URI = "/test/1.0/" +
                "024-02-07T00:30:48.014+00:00?time=024-02-07T00:30:48.014+00:00" 

client setting unencoded PATH = /test/1.0/024-02-07T00:30:48.014+00:00

**resource variable path = 024-02-07T00:30:48.014 00:00** //'+' is replaced with ' '
 from http req : path = /test/1.0/024-02-07T00:30:48.014+00:00
 from http req : query param time = 024-02-07T00:30:48.014 00:00 //'+' is replaced with ' '

I would like to get the resourcePath and query param in the raw format for downstream processing.

### Steps To Reproduce

```
@Controller
public class HelloController {
    @Get(uris = {"/test/{version}/{+resourcePath}" })
    @Produces(MediaType.ALL)
    public HttpResponse<?> triggerGet(@PathVariable("version") @Nullable String version,
                                      @PathVariable("resourcePath") @Nullable String resourcePath,
                                      HttpRequest httpRequest) {

        String path = httpRequest.getPath(); //This is the actual path
        return HttpResponse.ok("resource variable path = " + resourcePath + "\n" +
                " from http req : path = " + httpRequest.getPath() + "\n" +
                " from http req : query param time = " + httpRequest.getParameters().get("time"));
    }
```

```
@Test
    public void testPathUrlUnEncoded() {
        HttpRequest<?> req = HttpRequest.GET("/test/1.0/" +
                "024-02-07T00:30:48.014+00:00?time=024-02-07T00:30:48.014+00:00");
        String path = req.getPath();
        System.out.println("client setting unencoded PATH = " + path + "\n");
        System.out.println(client.toBlocking().retrieve(req));
    }
```

### Environment Information

JDK 17
[...]

PR #11477, opened 2024-12-31
Add UriTemplate API to preserve raw path component

ChaimaaeROUAI:
issue #10564


http/src/main/java/io/micronaut/http/uri/UriTemplate.java
```diff
@@ -199,6 +199,18 @@ public int getRawSegmentLength() {
             .reduce(Integer::sum)
             .orElse(0);
     }
+    /**
+     * Gets the raw unencoded path component from a URI.
+     *
+     * @param uri The URI string
+     * @return The raw path component, preserving encoding of special characters
+     */
```
sdelamo, line 208:
```suggestion
     * @since 4.8.0
     */
```

sdelamo (changes requested):

sdelamo:
@ChaimaaeROUAI @yawkat is working on a new api related to this for 4.9. Thanks for the contribution lets close this for now.

PR #11481, opened 2025-01-02
Prefer constructor over builder if present in introspections

graemerocher:
Currently if there is a builder then it is preferred over the constructor. This is a regression from 4.6.x behaviour. This alters the behaviour to prefer the constructor if one is present when returning `getConstructorArguments()`

Fixes https://github.com/micronaut-projects/micronaut-data/issues/3267

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
+++ b/inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
@@ -167,6 +167,15 @@ public abstract class AbstractInitializableBeanIntrospection<B> implements Unsaf
     }
 
     /**
+     * Whether an accessible constructor exists.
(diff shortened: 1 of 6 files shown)


sdelamo (approved):

inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
```diff
@@ -166,6 +166,15 @@ public Builder<B> builder() {
         }
     }
 
+    /**
+     * Whether an accessible constructor exists.
+     * @return True if a default constructor exists
+     * @since 4.7.11
+     */
+    protected boolean hasConstructor() {
+        return false;
```
dstepanov, line 175: Would be nice to have some default that is compatible with previously compiled beans

Issue #9480, opened 2023-06-21
Setting the 'Content-Disposition' header for arbitrary content


ee-usgs:
There should be an easy way to annotate a Controller method to set the `Content-Disposition: attachment` header.  The `Content-Disposition` header is an independent concern:  There are common and reasonable needs to send any http response as either `attached` or `inline`.

Some examples: 
* Serving a Javascript file on a coding website:  The same js file is displayed (`inline`)and a button offers a link to download it (`attachment`).
* Serving a `Flux<String>` of generated content (my situation).  I am sending report data:  For some controller methods I want to display it in the browsers, other methods need to send the contents as a download.

It looks like the current options are:
* Return a HttpReponse object instead, but then there is not much point in using a framework like MN, which adds a lot of value in intelligently handling return types.
* Add a Filter which sets the `Content-Disposition`, but that is a lot of non-intuitive complication to add for what seems like a useful and simple feature.

### Discussed in https://github.com/micronaut-projects/micronaut-core/discussions/9463

<div type='discussions-op-text'>

<sup>Originally posted by **ee-usgs** June 17, 2023</sup>
I am streaming CSV data from a Controller method - How do I set the `Content-Disposition: attachment` header so that browsers download it instead of display it?

I see that for sending 'Files', [this is done automatically](https://docs.micronaut.io/latest/guide/#transfers), but I don't intend to use either of the two return types shown in that example, my method signature looks like this:

```java
@Get("/result{?values*}")
public Flux<String> result(@Nullable @QueryValue("values") Map<String, String> filterMap);
```

Isn't there a way to just annotate the content disposition?  It seems weird to me that it currently seems tied to specific response types - In general, any response of any type might be inline or attachment.  It's an independent concern.

Thank you in advance for the help.

--ee</div>

deepthought-64:
Upvoting this - I have the same problem. 
Having a method that returns a Flow of Strings (coincidentally also a CSV export). I woul like to stream the result to the user due to memory concerns and I like the solution using the Flow. 
But I need to set the content-disposition header and I agree that it should be decoupled from the response type.

deepthought-64:
Update: As solution/workaround you can return a HttpReponse that wraps the Flowable/Flux as body. You can  set arbitrary headers that way.

```
@Post
fun search(): HttpResponse<Flowable<String>> {
     return HttpResponse
        .ok(searchAsFlow(/*...*/)
[...]

PR #11342, opened 2024-11-14
Move body writing logic into http-server

yawkat:
The main goal of this PR is to finally move the last pieces of the response encoding logic from RoutingInBoundHandler into a server-independent class (ResponseLifecycle). To accomplish this, I did a few refactors, mostly related to streaming responses.

Before this patch, streaming responses (e.g. Publisher<MyBean>) were handled by using the MessageBodyWriter to encode each item into a netty HttpContent (fully buffered) and concatenating them into a single netty ByteBody, with optional JSON separators (start/end bracket and commas).

The new approach is a bit different. The individual pieces are written using a new ResponseBodyWriter.writePiece method that, unlike the normal write method used before, does not have mutable response headers, which don't make sense for this scenario (the headers are already sent at the time of serialization). Additionally, it returns a ByteBody, so the piece can actually be streamed instead of fully buffered.

Next, the ByteBodys are concatenated using a new ConcatenatingSubscriber that streams each piece and forwards the data into a streaming ByteBody depending on runtime (BaseSharedBuffer implementation, for netty that's StreamingNettyByteBody). This class is the most intricate new code of this PR.

In order for servlet to use this ConcatenatingSubscriber, it needs a BaseSharedBuffer implementation, so I've moved the existing ReactiveByteBufferByteBody from the JDK client module to the HTTP module. For netty this is not used.

After these changes, I was able to move all the encoding methods to ResponseLifecycle with few changes. There are only a few netty-specific overrides in NettyResponseLifecycle: Netty-based ConcatenatingSubscribers, and handling  for the netty-specific StreamedHttpResponse.

To replace some of the netty-specific optimizations in RoutingInBoundHandler (such as the NettyResponseBodyWriterWrapper which has been removed), I've created a new ByteBodyFactory class that expands on the ByteBufferFactory concept with body factory methods. Unlike ByteBufferFactory, each method has a clear default implementation that works just fine and will be used as-is by servlet, but the class is also designed so that each method has a slightly more efficient netty implementation. ResponseBodyWriter has been changed to use ByteBodyFactory instead of ByteBufferFactory, making NettyResponseBodyWriterWrapper obsolete and also making many ResponseBodyWriters netty-independent (though I haven't moved them to http-server yet).

PR #11488, opened 2025-01-06
add @ContentDisposition annotation for setting Content-Disposition he…

ChaimaaeROUAI:
issue #9480

### DIFF
--- a/http/build.gradle.kts
+++ b/http/build.gradle.kts
@@ -26,6 +26,8 @@ dependencies {
     testImplementation(libs.micronaut.test.junit5) {
         exclude(group= "io.micronaut")
     }
+    testImplementation("io.micronaut:micronaut-http-client")
+    testImplementation("io.micronaut:micronaut-http-server-netty")
 }
 
 tasks {
--- a/http/src/main/java/io/micronaut/http/annotation/ContentDisposition.java
+++ b/http/src/main/java/io/micronaut/http/annotation/ContentDisposition.java
@@ -16,6 +16,7 @@
 package io.micronaut.http.annotation;
 
 import java.lang.annotation.*;
+
 /**
  * Annotation to specify the Content-Disposition header value for a method.
  * This is typically used to indicate how the content should be handled by the client.
--- a/http/src/main/java/io/micronaut/http/annotation/ContentDisposition.java
+++ b/http/src/main/java/io/micronaut/http/annotation/ContentDisposition.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.annotation;
+
+import java.lang.annotation.*;
+/**
+ * Annotation to specify the Content-Disposition header value for a method.
+ * This is typically used to indicate how the content should be handled by the client.
+ *
+ * @since 4.8.0
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ContentDisposition {
+    String value();
+}
--- a/http/src/main/java/io/micronaut/http/filter/ContentDispositionFilter.java
+++ b/http/src/main/java/io/micronaut/http/filter/ContentDispositionFilter.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


http/build.gradle.kts
```diff
@@ -26,6 +26,8 @@ dependencies {
     testImplementation(libs.micronaut.test.junit5) {
         exclude(group= "io.micronaut")
     }
+    testImplementation("io.micronaut:micronaut-http-client")
```
yawkat, line 29: please move the test to the tck

http/src/main/java/io/micronaut/http/annotation/ContentDisposition.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.annotation;
+
+import java.lang.annotation.*;
+/**
```
yawkat, line 20:
```suggestion

/**
```

http/src/main/java/io/micronaut/http/filter/ContentDispositionFilter.java
```diff
@@ -0,0 +1,53 @@
...
+
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.MutableHttpResponse;
+import io.micronaut.http.annotation.Filter;
+import jakarta.inject.Singleton;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+
+/**
+ * A filter that adds the `Content-Disposition` header to HTTP responses for specific endpoints.
```
yawkat, line 26: Add `@since`

http/src/main/java/io/micronaut/http/filter/ContentDispositionFilter.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Applies the filter logic to the HTTP request and modifies the response if the request
+     * matches specific criteria.
+     *
+     * @param request The incoming HTTP request.
+     * @param chain   The filter chain to proceed with the request.
+     * @return A publisher that emits the modified or unmodified HTTP response.
+     */
+    @Override
+    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
```
yawkat, line 41: Add nullability annotations

http/src/main/java/io/micronaut/http/filter/ContentDispositionFilter.java
```diff
@@ -0,0 +1,53 @@
...
+import jakarta.inject.Singleton;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+
+/**
+ * A filter that adds the `Content-Disposition` header to HTTP responses for specific endpoints.
+ */
+@Singleton
+@Filter("/**")
+public class ContentDispositionFilter implements HttpServerFilter {
```
yawkat, line 30: Use `@ServerFilter` API

yawkat: In fact I'm not sure we can have a general filter like this. It will affect performance of all controllers, even those without the annotation. Potentially this change needs to go into ResponseLifecycle instead (after #11342 )

dstepanov:
I think you can simply implement `ContentDisposition` as an alias for `Header`

http/src/main/java/io/micronaut/http/filter/ContentDispositionFilter.java
```diff
@@ -0,0 +1,53 @@
...
+     * Applies the filter logic to the HTTP request and modifies the response if the request
+     * matches specific criteria.
+     *
+     * @param request The incoming HTTP request.
+     * @param chain   The filter chain to proceed with the request.
+     * @return A publisher that emits the modified or unmodified HTTP response.
+     */
+    @Override
+    public Publisher<MutableHttpResponse<?>> doFilter(HttpRequest<?> request, ServerFilterChain chain) {
+        if (request.getPath().startsWith("/report/result")) {
```
graemerocher, line 42: this class is in `src/main/java` but seems to be hard coded to some test classes (`ReportController`) this doesn't seem to belong here at all. What is this PR trying to achieve?

http/src/main/java/io/micronaut/http/filter/ReportController.java
```diff
@@ -0,0 +1,41 @@
...
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Get;
+import reactor.core.publisher.Flux;
+
+/**
+ * Controller responsible for handling report generation requests.
+ * Provides an endpoint to generate and download a CSV report.
+ */
+@Controller("/report")
+public class ReportController {
```
graemerocher, line 28: this should not be in `src/main/java`

graemerocher (changes requested):

ChaimaaeROUAI:
After reviewing some documentation, I think the simplest and most idiomatic approach would be to use `HttpResponse<T>` whenever we need custom headers , rather than adding a new annotation.This way, we can avoid any global filter overhead or lifecycle changes.
And in Micronaut, we can simply  stream data from an `HttpResponse<Flux<T>>` just as well as from a raw `Flux<T>`
For example 
```java
@Get("/csv")
public HttpResponse<Flux<String>> downloadCsv() {
    Flux<String> data = Flux.just(
        "data1,data2",
        "data3,data4"
    );
    return HttpResponse
        .ok(data)
        .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"data.csv\"")
        .contentType(MediaType.TEXT_PLAIN_TYPE);
}
```
One thing  is the controller method signature changes from `Flux<String>` to `HttpResponse<Flux<String>>` to add custom header.
I could be mistaken, of course—any feedback is welcome!

yawkat:
Hmm yea maybe this is a bit too niche for an annotation. An annotation is possible (probably similar to what we do with `@Status` today), but not worth it

yawkat:
@ChaimaaeROUAI can you make a PR to the docs instead? #9480 shows that there is at least some confusion about how to do this, so it's worth writing down

PR #11495, opened 2025-01-08
More type pollution tests + avoid some of the checks

dstepanov:
Reduced some of the reactive HTTP path type-pollution

### DIFF
--- a/core/src/main/java/io/micronaut/core/execution/ExecutionFlow.java
+++ b/core/src/main/java/io/micronaut/core/execution/ExecutionFlow.java
@@ -29,6 +29,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/micronaut/core/execution/ExecutionFlow.java
```diff
@@ -278,5 +277,36 @@ default CompletableFuture<T> toCompletableFuture() {
...
     }
+
+    /**
+     * The flat map function interface.
+     * @param <I> The input
+     * @param <O> The output
+     * @since 4.8
+     */
+    @FunctionalInterface
+    interface FlatMapFn<I, O> {
```
yawkat: I don't think we should add APIs just to avoid type pollution for reactor.

Also this would be a binary incompatible change, and I think ExecutionFlow is used in servlet at this point

dstepanov: I have removed those changes

yawkat (approved):

graemerocher (approved):

Issue #11430, opened 2024-12-11
Unable to select alias for https


viniciusxyz:
### Expected Behavior

When using one of the properties below I hope the correct alias is selected in PKCS12

```properties
micronaut.ssl.key.alias=alias2
micronaut.server.ssl.key.alias=alias2
```

### Actual Behaviour

When I have a PKCS12 with several aliases, I cannot select the desired alias with the properties mentioned above

The alias property "micronaut.ssl.key.alias" is set to "alias2", however when validating the http certificate we should see a CN "localhost2":

![image](https://github.com/user-attachments/assets/6de2d072-fa27-4ea8-bafc-87270b89904d)

### Steps To Reproduce

1. Download, compile and run the example application below:

https://github.com/viniciusxyz/https-micronaut

This application has a certificate at the root that was created using this commands:

```sh
keytool -genkeypair -v -keystore keystore.p12 -storetype PKCS12 -alias alias1 -keyalg RSA -keysize 2048 -validity 3650 -dname "CN=localhost, OU=Spring Boot, O=My Company, L=City, ST=State, C=BR" -storepass password
keytool -genkeypair -v -keystore keystore.p12 -storetype PKCS12 -alias alias2 -keyalg RSA -keysize 2048 -validity 3650 -dname "CN=localhost2, OU=Spring Boot, O=My Company, L=City, ST=State, C=BR" -storepass password
keytool -genkeypair -v -keystore keystore.p12 -storetype PKCS12 -alias alias3 -keyalg RSA -keysize 2048 -validity 3650 -dname "CN=localhost3, OU=Spring Boot, O=My Company, L=City, ST=State, C=BR" -storepass password
```

2.  Validate the certificate used for https in the browser, I am validating the CN field and as you can see, each alias has a keypair with a different CN field

### Example Application

https://github.com/viniciusxyz/https-micronaut

### Version

4.7.6

yawkat:
Yea I think it's just not implemented

viniciusxyz:
> Yea I think it's just not implemented

As soon as I have time I will carry out an implementation and send the PR. Thanks for the clarification

viniciusxyz:
Feature added in 4.8.0 [#11503](https://github.com/micronaut-projects/micronaut-core/pull/11503)

PR #11503, opened 2025-01-13
Added support for alias selection in the keystore

viniciusxyz:
The approach used is very simple, the content of the alias is extracted and a new object representing a keystore is created only with that alias, this way it is possible to load only the desired alias and no more errors occur when we have multiple aliases in the keystore each with a password

Related issue #11430


sdelamo (commented):
Could you add a test?

viniciusxyz:
> Could you add a test?

for sure, I will do it today.

yawkat (approved):

graemerocher:
Thanks for the contribution!

PR #11520, opened 2025-01-20
Fix logged exception when parsing invalid HTTP Host header

yawkat:
Discovered by fuzzing.


graemerocher (approved):

graemerocher (changes requested):
can you fix the bug detected by Sonar

PR #11525, opened 2025-01-22
Jackson Databind to 2.18.2

sdelamo:
@yawkat Could you look into this PR to see if we could get rid of [`ResilientBeanDeserializerFactory`](https://github.com/micronaut-projects/micronaut-core/pull/6770)? `CreatorCollectionState` no longer exists in the Jackson Databind codebase.


sdelamo:
@yawkat I have removed `ResilientBeanDeserializerFactory `

PR #11527, opened 2025-01-22
add support for programmatically adding bean configurations

graemerocher:
Allows adding bean configurations at runtime that allow, for example, disabling certain beans within a package. Later it may make sense to allow `@MicronautTest` to use this disable certain beans as necessary.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/configurations/RequiresBeanSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/configurations/RequiresBeanSpec.groovy
@@ -84,7 +84,7 @@ class RequiresBeanSpec extends Specification {
 
         then:
         def e = thrown(ConfigurationException)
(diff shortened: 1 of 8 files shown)


sdelamo (approved):

dstepanov:
Personally, I don't like this. People will end up disabling internal packages and then be upset if we modify something internally.

graemerocher:
I tried to exclude the critical packages and make the API difficult enough to use that it requires a bit of thought, but I understand the concern.

dstepanov:
We can add a check that will throw if one of the packages is `io.micronaut`

graemerocher:
there is a check already but it is limited to `io.micronaut.context` and `io.micronaut.inject`, you suggesting we not allow disabling of any Micronaut module?

dstepanov:
I would not allow to disable Micronaut packages

graemerocher:
ok will go with that

dstepanov (approved):

PR #11545, opened 2025-01-27
docs: add example for streaming data with Content-Disposition

ChaimaaeROUAI:
In this issue #9480 , it seems that there is some confusion around how to stream data (like CSV) while forcing a file download via Content-Disposition.

### DIFF
--- a/src/main/docs/guide/httpServer/transfers.adoc
+++ b/src/main/docs/guide/httpServer/transfers.adoc
@@ -44,19 +44,7 @@ or any `Publisher`) without buffering the entire response in memory. If you want
 force the client browser to download the streamed data (for example, CSV lines),
 you can set the `Content-Disposition: attachment` header.
 
-[source,java]
-----
-@Get("/csv")
-public HttpResponse<Flux<String>> downloadCsv() {
-    Flux<String> data = Flux.just(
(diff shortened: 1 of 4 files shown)


src/main/docs/guide/httpServer/transfers.adoc
```diff
@@ -38,6 +38,31 @@ The server supports returning `304` (Not Modified) responses if the files being
...
+@Get("/csv")
+public HttpResponse<Flux<String>> downloadCsv() {
+    Flux<String> data = Flux.just(
+        "data1,data2",
+        "data3,data4"
+    );
+    return HttpResponse.ok(data)
+        .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"data.csv\"")
+        .contentType(MediaType.TEXT_PLAIN_TYPE);
+}
```
graemerocher: could you move this to a source file that compiles that preferably has a test and use the `snippet::` macro to include with examples for Java, Kotlin and Groovy. Thanks.

graemerocher (changes requested):

graemerocher (approved):

sdelamo (approved):

PR #11547, opened 2025-01-28
Support reactive/future HTTP client bodies

dstepanov:


http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
```diff
@@ -894,16 +896,21 @@ private <I, O, E> Mono<HttpResponse<O>> exchange(io.micronaut.http.HttpRequest<I
         AtomicReference<ScheduledExecutorService> scheduler = new AtomicReference<>(connectionManager.getGroup());
         ExecutionFlow<HttpResponse<O>> mono = resolveRequestURI(request).flatMap(uri -> {
             MutableHttpRequest<?> mutableRequest = toMutableRequest(request).uri(uri);
+            Object body = mutableRequest.getBody().orElse(null);
+            if (body instanceof Publisher<?> publisher) {
+                return ReactiveExecutionFlow.fromPublisher(publisher).flatMap(resolvedBody -> {
```
yawkat, line 901: I think Publisher support also needs multi-element support.

yawkat:
I don't really see how this is useful for the single-element case, cant you just do `body.onComplete(b->client.xyz(b))` instead of `client.xyz(body)` directly?

dstepanov:
The problem that I’m trying to use an interface API definition, controller works fine but the client is missing this.

http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
dstepanov: Not sure how we should support it

yawkat: My vote is that we don't for now. Actually mirroring the server behavior is more complicated than this. But this implementation is not workable, e.g. `@Body Publisher<byte[]>` will support streaming on the server but not on the client.

graemerocher:
Are all the failures related?

Issue #1853, opened 2019-07-01
Form-url-encoded body with JsonProperty annotation and http client


sdelamo:
### Steps to Reproduce

1. create a sample `mn create-app example.micronaut --inplace`

2.  Create a controller which consumes `form-url-encoded` with a POJO with a field annotated with `@JsonProperty`. 

```java
package example.micronaut;

import io.micronaut.http.HttpStatus;
import io.micronaut.http.MediaType;
import io.micronaut.http.annotation.Body;
import io.micronaut.http.annotation.Consumes;
import io.micronaut.http.annotation.Controller;
import io.micronaut.http.annotation.Post;
import io.micronaut.http.annotation.Status;

import javax.validation.Valid;

@Controller
public class HomeController {

    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
    @Status(HttpStatus.OK)
    @Post
    public void index(@Valid @Body DeviceAuthentication device) {
    }
}
```

```java
package example.micronaut;

import com.fasterxml.jackson.annotation.JsonProperty;
import io.micronaut.core.annotation.Introspected;

import javax.annotation.Nonnull;
import javax.validation.constraints.NotBlank;
import java.util.HashMap;
import java.util.Map;

@Introspected
public class DeviceAuthentication {

    @JsonProperty("client_id")
    @Nonnull
    @NotBlank
    private String clientId;

    public DeviceAuthentication() {

    }

    public DeviceAuthentication(String clientId) {
        this.clientId = clientId;
    }

    @Nonnull
    public String getClientId() {
        return clientId;
    }

    public void setClientId(@Nonnull String clientId) {
        this.clientId = clientId;
    }

    public Map<String, Object> toMap() {
        Map<String, Object> m = new HashMap<>();
        m.put("client_id", getClientId());
        return m;
    }
}
```

```groovy
package example.micronaut

import io.micronaut.context.ApplicationContext
import io.micronaut.http.HttpRequest
import io.micronaut.http.HttpResponse
import io.micronaut.http.HttpStatus
import io.micronaut.http.MediaType
import io.micronaut.http.client.BlockingHttpClient
import io.micronaut.http.client.HttpClient
import io.micronaut.runtime.server.EmbeddedServer
import spock.lang.AutoCleanup
import spock.lang.Shared
import spock.lang.Specification

class HomeControllerSpec extends Specification {

    @AutoCleanup
    @Shared
[...]

PR #11558, opened 2025-01-30
doc:  HTTP Client cannot use Form & @JsonProperty

sdelamo:
See: https://github.com/micronaut-projects/micronaut-core/issues/1853

This pull-requests adds a TCK test, and it documents that although the server binding of a form URL encoded payload supports Jackson annotations such as @JsonProperty, the HTTP Client does not.

The JDK HTTP Client does not even support form URL encoded content type.


http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/forms/FormsJacksonAnnotationsTest.java
```diff
@@ -0,0 +1,97 @@
...
+    static class SaveController {
+        @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+        @Post("/save")
+        Book save(@Body Book book) {
+            return book;
+        }
+    }
+
+    @Introspected
+    record Book(@JsonProperty("title") @NonNull String title, @JsonProperty("paginas") @Nullable Integer pages) {
```
sdelamo, line 94: If I use: 

```
@Introspected
record Book(@NonNull String title, @JsonProperty("paginas") @Nullable Integer pages) {}
```
 
 without the redundant ` @JsonProperty("title")` in GraalVM Native Image it renders as: 
 
`{“paginas":100}` instead of `{"title":"Building Microservices","paginas":100}`

Any idea? The TCK uses Micronaut Jackson Databind

sdelamo:
ping @yawkat

yawkat (approved):

PR #11559, opened 2025-01-31
improvement: List<MessageSource> already ordered

sdelamo:
This PR adds a test to verify message source order is taken into account. Moreover, it simplifies the CompositeMessageSource constructor just passing a List of beans of type MessageSource. The list will be already ordered.


dstepanov:
A collection is not ordered by a list is?

sdelamo:
> A collection is not ordered by a list is?

I am unsure if a `Collection` of beans of a certain type gets injected or ordered. The fact that the previous code ordered the collection explicitly tells me the collection may not have been injected in order. 

Nonetheless, I think injecting a `List` of beans of a certain type is more correct, which clearly conveys that you want an ordered collection.

yawkat (approved):

PR #11570, opened 2025-02-05
Atomic operations for BeanDefinitionProducer

### LINKED ISSUE #11569: Race condition in DefaultBeanContext.BeanDefinitionProducer
### Expected Behavior

This code should work reliably under high concurrency.

### Actual Behaviour

A variety of crashes due to race conditions in this code, which currently uses four different volatiles in order to be lock free. An example would be (truncated):

```
Caused by: java.lang.IllegalStateException: The definition is not enabled
	at io.micronaut.context.DefaultBeanContext$BeanDefinitionProducer.getDefinition(DefaultBeanContext.java:4310)
	at io.micronaut.context.DefaultBeanContext.collectBeanCandidates(DefaultBeanContext.java:2156)
```

which can occur because this class expects all reads of `definitionEnabled` to go via `isDefinitionEnabled(...)` but inside `getDefinition` it doesn't do that, so it's possible that `definitionEnabled == null` causing `producer.isDisabled()` to return false whilst another thread is in the process of initializing it.

I thought about trying to patch this minimally but it isn't the right way to go, because lock-free programming is hard to get right and the code would probably still have races. It'd be better to just use a single lock to protect the data.

### Steps To Reproduce

Our test suite uses Micronaut in very concurrent ways because it runs every unit test in parallel. On some fraction of runs it fails in this way.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.6.5

yawkat:
Replace the BeanDefinitionProducer API with atomic operations. This removes patterns like `if (producer.isReferenceEnabled()) producer.getReference()` which could cause TOCTOU concurrency issues.

Should fix #11569

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -4201,9 +4201,19 @@ public class DefaultBeanContext implements InitializableBeanContext, Configurabl
             AtomicReferenceFieldUpdater.newUpdater(BeanDefinitionProducer.class, Object.class, "definition");
         private static final Object DEFINITION_DISABLED_SENTINEL = "";
 
+        /**
+         * Initially the reference, may be set to {@code null} by {@link #disable()} or if
+         * {@link #getReferenceIfEnabled} determines the ref is disabled.
+         */
         @Nullable
         @SuppressWarnings("java:S3077")
         private volatile BeanDefinitionReference reference;
+        /**
+         * Initially {@code null}. If the {@link #reference} is enabled, and the definition from it
+         * is also enabled, this is set to the {@link BeanDefinition}. If the definition is
+         * disabled (either through conditions or explicitly by {@link #disable()}), this is set to
+         * {@link #DEFINITION_DISABLED_SENTINEL}.
+         */
         @Nullable
         @SuppressWarnings("java:S3077")
         private volatile Object definition;
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -143,6 +143,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2078,7 +2080,7 @@ protected void initializeContext(
         if (CollectionUtils.isNotEmpty(parallelBeans)) {
             processParallelBeans(parallelBeans);
         }
-        ForkJoinPool.commonPool().execute(() -> beanDefinitionsClasses.forEach(p -> p.isReferenceEnabled(this)));
+        ForkJoinPool.commonPool().execute(() -> beanDefinitionsClasses.forEach(p -> p.getReferenceIfEnabled(this)));
```
mikehearn, line 2083: What is this line trying to do originally? It looks a bit odd.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2463,7 +2462,7 @@ protected <T> BeanDefinition<T> findConcreteCandidate(@NonNull Class<T> beanType
     protected void processParallelBeans(List<BeanDefinitionProducer> parallelBeans) {
         if (!parallelBeans.isEmpty()) {
             List<BeanDefinitionProducer> finalParallelBeans = parallelBeans.stream()
-                    .filter(p -> p.isReferenceEnabled(this))
+                .filter(p -> p.getReferenceIfEnabled(this) != null)
```
mikehearn: Indentation changed.

mikehearn (approved):
Looks OK though it'd be good to have a comment justifying this sort of very fine grained parallelism with performance numbers. There are some unintuitive states possible here (e.g. reference == null whilst definition is not the sentinel). I can see why bean management would be performance sensitive, but how much difference does all this really make?

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
yawkat: I think it's kicking off async loading of all bean definitions that were not initialized by the previous code

sdelamo (approved):
can we have tests for this?

dstepanov:
Is it going to affect the startup performance?

yawkat:
@sdelamo i don't know how to reproduce the original bug, it's some weird race condition

yawkat:
@dstepanov I think it should be no worse than before, there's no additional locks or extra work. maybe it's even slightly faster

sdelamo:
@dstepanov If you are ok with this, lets merge it and release 4.8.1 core

Issue #11577, opened 2025-02-07
java.time.temporal.UnsupportedTemporalTypeException is thrown when parsing property value [micronaut.session.http.cookie-max-age]


daria-kashperska:
### Expected Behavior

Cookie is configured with specified micronaut.session.http.cookie-max-age property value.

### Actual Behaviour

An exception is thrown with the following stack trace:
```
java.time.temporal.UnsupportedTemporalTypeException: Unsupported unit: Seconds
        at java.base/java.time.Period.get(Period.java:443)
        at io.micronaut.http.cookie.Cookie.maxAge(Cookie.java:259)
        at java.base/java.util.Optional.ifPresent(Optional.java:178)
        at io.micronaut.http.cookie.Cookie.configure(Cookie.java:239)
        at io.micronaut.http.cookie.Cookie.configure(Cookie.java:224)
        at com.example.CookieService.createCookie(CookieService.java:21)
        at com.example.CookieController.testCookie(CookieController.java:17)
        at com.example.$CookieController$Definition$Exec.dispatch(Unknown Source)
        at io.micronaut.context.AbstractExecutableMethodsDefinition$DispatchedExecutableMethod.invokeUnsafe(AbstractExecutableMethodsDefinition.java:461)
        at io.micronaut.context.DefaultBeanContext$BeanContextUnsafeExecutionHandle.invokeUnsafe(DefaultBeanContext.java:4354)
        at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:231)
        at io.micronaut.web.router.DefaultUriRouteMatch.execute(DefaultUriRouteMatch.java:38)
        at io.micronaut.http.server.RouteExecutor.executeRouteAndConvertBody(RouteExecutor.java:488)
        at io.micronaut.http.server.RouteExecutor.callRoute(RouteExecutor.java:478)
        at io.micronaut.http.server.RequestLifecycle.callRoute(RequestLifecycle.java:182)
        at io.micronaut.http.server.RequestLifecycle.executeRoute(RequestLifecycle.java:170)
        at io.micronaut.http.server.RequestLifecycle$2.provideResponse(RequestLifecycle.java:390)
        at io.micronaut.http.filter.FilterRunner.provideResponseAndHandleErrors(FilterRunner.java:372)
        at io.micronaut.http.filter.FilterRunner.filterRequest(FilterRunner.java:307)
[...]

PR #11581, opened 2025-02-10
Fixed conversion string to temporalAmount - Duration or Period

altro3:
Fixed https://github.com/micronaut-projects/micronaut-core/issues/11577

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
+++ b/context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
@@ -89,7 +89,7 @@ import static java.time.temporal.ChronoField.YEAR;
 @Internal
 public class TimeConverterRegistrar implements TypeConverterRegistrar {
 
-    private static final Pattern PERIOD_MATCHER = Pattern.compile("^(-?\\d+)([unywmd])(s?)$");
+    private static final Pattern PERIOD_MATCHER = Pattern.compile("^(-?\\d+)([unywmMdD])(s?)$");
     private static final Pattern DURATION_MATCHER = Pattern.compile("^(-?\\d+)([unsmhd])(s?)$");
     private static final int MILLIS = 3;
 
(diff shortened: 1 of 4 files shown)


graemerocher:
@altro3 can you target 4.8.x?

sdelamo:
> @altro3 can you target 4.8.x?

I asked @altro3 to target 4.7.x as it was a bug introduced in 4.7.x. I will merge up @graemerocher

dstepanov:
Why `conversionService.convert("string", Duration)` is triggering `TemporalAccessor` converter?

sdelamo (approved):

altro3:
@dstepanov @graemerocher @sdelamo I would like to clarify that there is currently a bug in the Cookie interface in the maxAge method.

The argument is of the wrong type - TemporalAmount , but it should be Duration , because we are interested in the value in seconds, and Period stores the timestamp in days. From the TemporalAmount documentation:
```java
* There are two common implementations.
* {@link Period} is a date-based implementation, storing years, months and days.
* {@link Duration} is a time-based implementation, storing seconds and nanoseconds,
* but providing some access using other duration based units such as minutes,
* hours and fixed 24-hour days.
```
I don't understand why TemporalAmount is there.

sdelamo:
> I would like to clarify that there is currently a bug in the Cookie interface in the maxAge method.
> The argument is of the wrong type - TemporalAmount , but it should be Duration , because we are interested in the value in seconds, and Period stores the timestamp in days. From the TemporalAmount documentation:

but this should be a different pr right?

altro3:
@sdelamo as you wish - I can add this correction to this PR too. :-) . Current fix - solve the problem. I just wanted to say that it is actually not correct to use TemporalAmount in this case

altro3:
@sdelamo added this fix here. Don't think that we need to create different PR for it

altro3:
@sdelamo can't fix it here, becuse also need to change micronaut-session module for it

PR #11587, opened 2025-02-11
Environment refresh issues

msupic:
Added a test which can be used to reproduce two micronaut environment refresh related issues:
1. the `diffMap` method from `DefaultEnvironment` class doesn't record a change when the `newMap` doesn't have a value which exists in the `map`
2. the `refresh` method stops `bootstrap` and `runtime` environments but starts only the `runtime` environment which causes all property sources to be removed from the `bootstrap` env but not re-added later when the `start` method is called. Because of that, the `bootstrap` properties don't exist in the `runtime` catalog after refresh is done.

The added test currently doesn't fail because of the first issue. When the first issue gets fixed, the test will fail and require the fix of the second issue.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
@@ -959,6 +959,7 @@ public class DefaultApplicationContext extends DefaultBeanContext implements Con
         public Environment stop() {
             if (bootstrapEnvironment != null) {
                 bootstrapEnvironment.stop();
+                bootstrapEnvironment = null;
             }
             return super.stop();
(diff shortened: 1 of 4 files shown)


sdelamo (changes requested):
If this is a bug, can we target for 4.7.x ?

graemerocher:
A back port would be complicated at this stage because the property catalog stuff has been rewritten. Would need a different PR for 4.7.x

graemerocher:
@sdelamo I created a backport PR https://github.com/micronaut-projects/micronaut-core/pull/11593 note it is likely to cause issues when merging up

sdelamo (approved):

PR #11601, opened 2025-02-19
Permit whatwg URLs for requests

yawkat:
This PR changes the URL parsing a bit.

- Introduce a fast path for "boring" valid relative URLs. This is always enabled and lets us skip the expensive URI parsing with default settings.
- Add a config option that escapes certain characters that are valid in whatwg URLs (e.g. `[]|`, browsers will happily send these) but invalid in java.net.URI. This allows those characters to be handled normally by the server.
- Remove the option to disable URI validation. The normal path should be fast enough now, and for those that need the whatwg-only characters, there's the new escape config option.

There's also a giant WhatwgParser class in the tests that I implemented while researching a while ago. I don't plan to move it to production code atm, but it remains useful for tests.


yawkat:
i might have to move this to micronaut-router for use there, putting this into draft for now.

router/src/test/java/io/micronaut/web/router/uri/PercentDecoder.java
```diff
@@ -0,0 +1,85 @@
+package io.micronaut.web.router.uri;
+
+import io.micronaut.core.annotation.Nullable;
+
+final class PercentDecoder {
```
graemerocher, line 5: annotate `@Internal`, add javadoc for maintainers, and license etc.

router/src/test/java/io/micronaut/web/router/uri/UriUtilTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+import com.code_intelligence.jazzer.junit.FuzzTest;
+import org.junit.jupiter.api.Assertions;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+class UriUtilTest {
+    @FuzzTest(maxDuration = "30m")
```
graemerocher, line 11: in addition to fuzz tests can we add some additional tests that run as part of the regular test suite? I couldn't find these.

graemerocher (changes requested):

router/src/test/java/io/micronaut/web/router/uri/UriUtilTest.java
yawkat: It's covered by existing tests (the malformed uri ones mostly), but I just added a check that request.uri works as expected

dstepanov (approved):

graemerocher (approved):

PR #11610, opened 2025-02-21
Fix disabling eager each-property beans

dstepanov:
Caused by changes in bean definition loading

Fixes https://github.com/micronaut-projects/micronaut-sql/pull/1545


yawkat:
I removed some code that is unused after this change

yawkat (approved):

graemerocher (approved):

PR #11636, opened 2025-03-01
adding jitter parameter will help in avoiding thundering herd problem

gargsuraj12:


sdelamo (commented):
You would need to sign the CLA.

retry/src/main/java/io/micronaut/retry/annotation/Retryable.java
```diff
@@ -81,6 +81,12 @@
     @Digits(integer = 2, fraction = 2)
     String multiplier() default "1.0";
 
+    /**
+     * @return The random deviation percentage between retry attempts
+     */
+    @Digits(integer = 2, fraction = 2)
+    String jitter() default "0.25";
```
graemerocher, line 88: simply adding the `jitter()` member to the annotation is not enough to implement the functionality

graemerocher (changes requested):

PR #11651, opened 2025-03-10
fix getBoundGenericTypes() not returning resolved type arguments

graemerocher:
Currently if you use `withTypeArguments` on a `ClassElement` then `JavaClassElement` does not factor in this mutated types. The issue doesn't seem to impact Kotlin or Groovy.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaClassElement.java
@@ -643,6 +643,9 @@ public class JavaClassElement extends AbstractJavaElement implements ArrayableCl
     @NonNull
     @Override
     public List<ClassElement> getBoundGenericTypes() {
+        if (resolvedTypeArguments != null) {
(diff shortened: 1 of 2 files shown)


dstepanov (approved):

graemerocher:
@yawkat this change seems to break something, not sure what the behaviour is supposed to be here but it seems like the types should not be resolved

yawkat:
The old behavior is intentional and documented in the superclass javadoc, why are you changing it?

graemerocher:
See the test https://github.com/micronaut-projects/micronaut-core/pull/11651/files#diff-fff17b27de970e53a38735ba099375698f2725fde1f9feb317016b2a984ca27fR17 

After a call to `withTypeArguments` then `getBoundGenericTypes` still returns the old types

yawkat:
Yes and that's what the javadoc says it should do:

> In particular, this method reflects the declaration type: If there is a class Test<T> { T field; }, this method will return T as the field type, even if the field type was obtained through a Test\<String>.

graemerocher:
ok then I think this is a bug in SourceGen's `TypeDef` method. @dstepanov do you agree?

Issue #2123, opened 2019-09-15
Support skipping retry for 4xx status code HTTP responses


binkley:
### Steps to Reproduce

1. Create a web client with a Micronaut `@Client` annotation, endpoint irrelevant (see below)
2. Mark the client with `@Retry`
3. Create a data class with `@Introspected` on it, including simple type fields (ie, `String` or `int`)
4. Add `javax.validation.constraints` annotations on fields of the data class
5. Add a method to the client which consumes the data class (ex: `@Post`)
6. Create a listener for `RetryEvent`
7. Create a test using the client, and invoking the method, with the test using invalid values in the data class (based on the validation constraints)
8. Verify the client retried the remote call after the first invocation

### Expected Behaviour

A. The client fails to call the remote endpoint as the data payload is invalid [PASS]
B. The client does not retry the call, as it can never succeed [FAIL]

### Actual Behaviour

The client retries the method up to the default limit or the configured retry limit, regardless of validation.

As of release 1.2.0.M1, the client did not call the remote endpoint at all as the validation failed.  I tried releases 1.2.1 and 1.2.2, and both retry up to the limit (though do not try to call the remote endpoint).  This operation can never succeed.

### Example Application

https://github.com/binkley/spikes/blob/e2021e1d864ed0351c9d799946e5cf58ecef3a59/micronaut-hello-world/src/test/java/hello/world/HelloControllerSpec.java#L73

Please note the commit hash.  The line number this link points to is a `@Disabled` annotation so the commit passes local build.  To reproduce the issue, remove the annotation, and see the test fail.

jameskleeh:
This is the expected behavior. The retryable functionality is not aware of HTTP and has no special logic for 4xx status codes.

debrutal:
While i think that this general approach is definitely a plus for fitting more needs, we also suffer from having a context free retry.
Especially, when i do not want to spoil the logs with error messages or waste ressources.

A: One Option would be to enable to provide micronaut retry a context, in order to make an informed decision on when to retry and when not. Option B shall be as an inspiration for this and a temporary workaround.

B: Another would be to follow the contract of retry, which is what we did. 
While http status code is quite nice to have, we also want to react on connection resets, so the code does more than @binkley was asking. 
We created
- a custom Exception
```java
public class CustomRetryableException extends RuntimeException {
    public CustomRetryableException(Throwable cause) {
        super(cause);
    }
}
```
- a RetryableWrapperInterceptor annotation
```java
[...]

Issue #1920, opened 2019-07-22
Feature request: Retryable annotation should be able to ignore certain HTTP Status codes


geemang2000:
I'm using the declarative HTTP client and invoking a remote authentication service. This service returns a 401 and the Retryable annotation loops (three times).  In this and other HTTP Status codes e.g. ill-formed payload (400) I don't want a retry. 

It may save lots of try/catch code checking for certain status codes to add some type of Status Code selection.

markjfisher:
Agreed this would be useful. We're also seeing it open Circuit Breaker whenever a request to an external service gets 400 (e.g. bad parameters in request), which then locks out any valid requests until the circuit is closed again.

mast7860:
Hi , Any update on  this enhancement ?

frodeo:
With the new predicate feature added in Micronaut 2 you can avoid retries in these cases:

```
@Client("${myEndpoint}")
@CircuitBreaker(attempts = "4", predicate = ServerErrorRetryPredicate.class)
public interface MyClient {
    @Get
    Single<MyItem> getItem(int itemId);
}

public class ServerErrorRetryPredicate implements RetryPredicate {
    @Override
    public boolean test(Throwable throwable) {
        if (throwable instanceof HttpClientResponseException) {
            HttpClientResponseException e = (HttpClientResponseException) throwable;
            return e.getStatus().getCode() >= 500;
        }
        return true;
    }
}
```

This code still opens the circuit breaker though, and that is not a good thing if you write REST services.

An option could be to make the **io.micronaut.retry.intercept.CircuitBreakerRetry** class support the predicate when opening circuit breakers as well (not only for retries as it is now):

```
@Override
public void close(@Nullable Throwable exception) {
    if (exception == null && currentState() == CircuitState.HALF_OPEN) {
        closeCircuit();
    } else if (exception != null
        && childState.canRetry(exception) // THIS LINE WAS ADDED TO USE PREDICATE FOR CIRCUIT BREAKERS
    ) {
        if (currentState() != CircuitState.OPEN) {
            openCircuit(exception);
        }
    } else {
        // reset state for successful operation
        time = System.currentTimeMillis();
        lastError = null;
        this.childState = (MutableRetryState) retryStateBuilder.build();
    }
}

@Override
public boolean canRetry(Throwable exception) {
    if (exception == null) {
        throw new IllegalArgumentException("Exception cause cannot be null");
    }
    try {
[...]

PR #11658, opened 2025-03-14
Document declarative client and retryable

joshua-m-baker:
Add a note to the documentation about combining the declarative Http client and `@Retryable` and ignoring 4XX responses

There are currently multiple issues about not retrying 4XX responses (see #1920 and #2123). I understand the behavior is intended to keep `@Retryable` agnostic to what it is retrying, but I think an extra call-out in the docs about what seems to be the recommended solution would help provide extra clarity, since disabling 4XX retries seems like a common desire. 

Additionally, I think an explicit note would make it clearer that with the declarative client and `@Retryable`, by default 404s will not be retried, but other 4XX responses will be. This is implied by the call out that 404 responses are not considered an error condition, but I think it's generally unexpected to see 4XXs being retried but 404s are 'excluded'


sdelamo (approved):
@joshua-m-baker Thanks for the contribution

PR #11670, opened 2025-03-21
extract HttpHeadersUtil::parseCharacterEncoding and parseAcceptCharset

sdelamo:
To allow the usage of those methods in runtimes where no HttpRequest or HttpMessage is available

### DIFF
--- a/http/src/main/java/io/micronaut/http/util/HttpHeadersUtil.java
+++ b/http/src/main/java/io/micronaut/http/util/HttpHeadersUtil.java
@@ -160,6 +160,7 @@ public final class HttpHeadersUtil {
      * Resolve the {@link Charset} to use for the request.
      *
      * @param contentType ContenType
+     * @param acceptCharset Accept Charset
      * @return An {@link Optional} of {@link Charset}
      * @since 4.8.8
      */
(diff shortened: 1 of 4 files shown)


graemerocher (approved):

http/src/main/java/io/micronaut/http/util/HttpHeadersUtil.java
```diff
@@ -135,4 +140,67 @@ public static String splitAcceptHeader(@NonNull String text) {
...
+        try {
+
+            if (contentType != null) {
+                String charset = contentType.getParametersMap().get(MediaType.CHARSET_PARAMETER);
+                if (charset != null) {
+                    try {
+                        return Charset.forName(charset);
+                    } catch (Exception e) {
+                        // unsupported charset, default to UTF-8
+                        return Charset.defaultCharset();
```
yawkat: change to utf-8

yawkat (changes requested):

http/src/main/java/io/micronaut/http/util/HttpHeadersUtil.java
sdelamo: https://github.com/micronaut-projects/micronaut-core/pull/11670/commits/43b3a2dfcfe58982a8ad3a6cb8004079b73dbf4a

yawkat (approved):

PR #11672, opened 2025-03-21
Netty 4.2

yawkat:


http-netty/src/main/java/io/micronaut/http/netty/channel/EventLoopGroupFactory.java
```diff
@@ -35,6 +38,7 @@
  * @author graemerocher
  * @since 1.2.0
  */
+@NextMajorVersion("This doesn't create event loops anymore. Rename to e.g. NettyTransport")
 public interface EventLoopGroupFactory {
```
graemerocher, line 42: Frankly I would just deprecate the whole class with `forRemoval` rather than leave it and make it do something it doesn't do anymore

graemerocher (approved):

http-netty/src/main/java/io/micronaut/http/netty/channel/EventLoopGroupFactory.java
yawkat: we still need all the implementations, so this class still has a purpose. it's just that the abstraction level has changed–it's not event loops being created anymore, it's IoHandlerFactories (and channel classes)

yawkat:
@sdelamo what can we do about the failure?

Netty 4.2 graduated the io-uring transport from incubator to the main netty repo, so the version catalog does not need the incubator version anymore and I removed it.

graemerocher:
> @sdelamo what can we do about the failure?
> 
> Netty 4.2 graduated the io-uring transport from incubator to the main netty repo, so the version catalog does not need the incubator version anymore and I removed it.

probably you need to do something similar to https://github.com/micronaut-projects/micronaut-tracing/blob/7.0.x/tracing-bom/build.gradle#L8

graemerocher:
@melix can confirm

melix:
That's correct, you need to add suppressions. Something like:

```
micronautBom {
    suppressions {
        acceptedVersionRegressions.add("netty-iouring")
    }
}
```

yawkat:
Thanks!

PR #11675, opened 2025-03-21
Fix: updating method name to match rfc (RFC5987)

### LINKED ISSUE #11674: Wrong RFC in the  method name for File Attachement
### Expected Behavior

During investigation for how the file attachment names are being handled in micronaut. It seems like the method name referencing the RFC is not correct.

Correct RFC as mentioned in the code comment: [RFC#6266](https://httpwg.org/specs/rfc6266.html)
Method name in code: `encodeRfc6987` 

### Actual Behaviour

The method name should match the referenced RFC

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.8.7

vabs:
Fixes #11674

This is a small fix. The method name for sanitizing the filename was incorrectly mentioning the RFC.


yawkat (approved):

vabs:
@yawkat The merging is blocked here. Can you please tell me which is the correct branch I should target this PR too?

yawkat:
There's test failures

PR #11677, opened 2025-03-23
Fix shutdown hooks

dzmauchy:
This PR adds

- protection against JVM receiving a SIGINT just before the application starts
- a flag to disable adding the Micronaut shutdown hook (described in the Javadoc)

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/Micronaut.java
+++ b/context/src/main/java/io/micronaut/runtime/Micronaut.java
@@ -135,7 +135,7 @@ public class Micronaut extends DefaultApplicationContextBuilder implements Appli
                             } catch (Throwable stopError) {
                                 LOG.error("Embedded Application shutting down", stopError);
                             }
-                            LOG.warn("Embedded Application: {}", e.getMessage());
+                            LOG.warn("Failed to register shutdown hook", e);
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


graemerocher:
the commits need to be done with the same GitHub account as created PR for CLA approval to work.

dzmauchy:
> the commits need to be done with the same GitHub account as created PR for CLA approval to work.

Fixed

context/src/main/java/io/micronaut/runtime/Micronaut.java
```diff
@@ -113,18 +113,31 @@ protected Micronaut() {
...
+                                    }
+                                }
+                            }));
+                        } catch (IllegalStateException e) {
+                            try (applicationContext) {
+                                embeddedApplication.stop();
+                            } catch (Throwable stopError) {
+                                LOG.error("Embedded Application shutting down", stopError);
                             }
+                            LOG.warn("Embedded Application: {}", e.getMessage());
```
yawkat:
```suggestion
                            LOG.warn("Failed to register shutdown hook", e);
```

yawkat (approved):
LGTM, but @graemerocher please check if EmbeddedApplication is the right place for this.

graemerocher (approved):

PR #11680, opened 2025-03-24
Fix accessing package protected fields from a different package

dstepanov:


graemerocher (approved):

core-processor/src/main/java/io/micronaut/inject/writer/DispatchWriter.java
```diff
@@ -514,19 +526,57 @@ public TypedElement getDeclaringType() {
...
+        }
+
+        @Override
+        public TypedElement getDeclaringType() {
+            return null;
+        }
+
+        @Override
+        public ExpressionDef dispatchExpression(ExpressionDef bean) {
+            return TYPE_REFLECTION_UTILS.invokeStatic(
```
yawkat, line 574: Would be nice to move this to method handles eventually. Might even work with native image, then

yawkat (approved):

PR #11685, opened 2025-03-25
ThreadLocal bean improvements

### LINKED ISSUE #11293: Performance: @ThreadLocal annotated bean causes ClassUtils.forName() upon every access.
### Expected Behavior

Access to thread-local values should be typically as fast as possible and avoid expensive operations, such as proxy bean creation or invoking a ClassLoader. The observed repetitive call to io.micronaut.core.reflect.ClassUtils.forName() on every access to a @ThreadLocal bean method seems unintentional to me.

**I would have assumed accesses to @ThreadLocal annotated beans to perform comparably to classic java.lang.ThreadLocal. However, using a classic java.lang.ThreadLocal is significantly faster.**

AbstractInitializableBeanDefinition.getScope() has a specific optimization  for SINGLETON_SCOPE. Possibly similar is conceivable for @ThreadLocal. Or hopefully the generated $DateUtils$Definition$Intercepted.getStdDateFormat()
 could avoid the whole getProxyTargetBean() chain entirely.

### Actual Behaviour

Referring to below example code, every access to the injected @ThreadLocal annotated bean via
`dateUtils.getStdDateFormat();`
attempts to
* DefaultApplicationContext(DefaultBeanContext).getProxyTargetBean
* DefaultApplicationContext(DefaultBeanContext).findCustomScope
* eventually load "io.micronaut.runtime.context.scope.ThreadLocal.class" from the ClassLoader.

### Steps To Reproduce

Given a minimal bean
```
@io.micronaut.runtime.context.scope.ThreadLocal
public class DateUtils {

	private final StdDateFormat stdDateFormat = new StdDateFormat();

	public DateFormat getStdDateFormat() {
		return stdDateFormat;
	}
}
```

consumed via

```
	@Inject
	DateUtils dateUtils;
	
	DateFormat df = dateUtils.getStdDateFormat();
```

sees the sequences of logs like

```
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Successfully loaded class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Successfully loaded class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Successfully loaded class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Successfully loaded class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Successfully loaded class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Successfully loaded class io.micronaut.runtime.context.scope.ThreadLocal
2024-10-31 00:47:42 [default-nioEventLoopGroup-1-4] DEBUG i.micronaut.core.reflect.ClassUtils - Attempting to dynamically load class io.micronaut.runtime.context.scope.ThreadLocal
```

if debug logging is configured for io.micronaut.core.reflect.ClassUtils.

Note that the invocations are all on the same thread.

Stacktrace
```
Thread [pool-2-thread-1] (Suspended (breakpoint at line 276 in ClassUtils))	
	ClassUtils.forName(String, ClassLoader) line: 276	
	$DateUtils$Definition(AbstractInitializableBeanDefinition<T>).lambda$getScope$0(String) line: 311	
	0x00007fbf4c3aa9a0.apply(Object) line: not available	
	Optional<T>.flatMap(Function<? super T,Optional<? extends U>>) line: 289	
	$DateUtils$Definition(AbstractInitializableBeanDefinition<T>).getScope() line: 307	
	DefaultApplicationContext(DefaultBeanContext).findCustomScope(BeanResolutionContext, BeanDefinition<T>, boolean, boolean) line: 3033	
	DefaultApplicationContext(DefaultBeanContext).resolveBeanRegistration(BeanResolutionContext, BeanDefinition<T>, Argument<T>, Qualifier<T>) line: 3005	
	DefaultApplicationContext(DefaultBeanContext).getProxyTargetBean(BeanResolutionContext, BeanDefinition<T>, Argument<T>, Qualifier<T>) line: 1505	
	$DateUtils$Definition$Intercepted.interceptedTarget() line: not available	
	$DateUtils$Definition$Intercepted.getStdDateFormat() line: not available	
	GridRepository.writeLiveMeasurement(IEvent<Named,String>, JsonNode) line: 304	
...
```

with
	name	"io.micronaut.runtime.context.scope.ThreadLocal" (id=179)	
	classLoader	ClassLoaders$AppClassLoader  (id=103)	

### Environment Information

* Operating System: Gentoo Linux
* JDK: openjdk 21.0.5 2024-10-15 LTS
OpenJDK Runtime Environment Temurin-21.0.5+11 (build 21.0.5+11-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.5+11 (build 21.0.5+11-LTS, mixed mode, sharing)
* Micronaut 4.6.3

### Example Application

n/a

### Version

4.6.3

yawkat:
Improve performance and lifecycle management of ThreadLocal beans.

- Add a benchmark
- Cache AbstractInitializableBeanDefinition.getScope. This is the majority of the performance improvement
- Implement ThreadLocalCustomScope without the AbstractConcurrentCustomScope superclass. This leads to a minor performance improvement (no more locks)
- Add optional advanced lifecycle handling. Can be turned on with the ThreadLocal.lifecycle flag. There are two cleanup paths: When a thread terminates, its beans are destroyed through a Cleaner on GC. Separately, when the app context terminates, all those cleaner tasks are invoked immediately.

There's still some performance left on the table, see the benchmark, but it's much better than before.

Fixes #11293

### DIFF
--- a/context/src/main/java/io/micronaut/runtime/context/scope/ThreadLocal.java
+++ b/context/src/main/java/io/micronaut/runtime/context/scope/ThreadLocal.java
@@ -17,11 +17,12 @@ package io.micronaut.runtime.context.scope;
 
 import jakarta.inject.Scope;
 
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
  * A {@link io.micronaut.context.scope.CustomScope} that stores objects in thread local storage.
@@ -35,4 +36,13 @@ import java.lang.annotation.Target;
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Scope
 public @interface ThreadLocal {
+    /**
+     * If enabled, track bean life cycle. This means that the bean will be stopped, destroy
+     * listeners will be called etc., when the application context is closed or when the associated
+     * thread terminates. Note that this adds some overhead, so it's off by default.
+     *
+     * @return Whether to enable lifecycle support for this bean
+     * @since 4.9.0
+     */
+    boolean lifecycle() default false;
 }
--- a/benchmarks/src/jmh/java/io/micronaut/context/scope/ThreadLocalScopeBenchmark.java
+++ b/benchmarks/src/jmh/java/io/micronaut/context/scope/ThreadLocalScopeBenchmark.java
@@ -1,0 +1,80 @@
+package io.micronaut.context.scope;
+
+import io.micronaut.context.ApplicationContext;
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.runtime.context.scope.ThreadLocal;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+public class ThreadLocalScopeBenchmark {
+    ApplicationContext ctx;
+    Holder holder;
+
+    @Setup
+    public void setup() {
+        ctx = ApplicationContext.run(Map.of("spec.name", "ThreadLocalScopeBenchmark"));
+        holder = ctx.getBean(Holder.class);
+    }
+
+    @TearDown
+    public void tearDown() {
+        ctx.close();
+    }
+
+    @Benchmark
+    public int bench() {
+        return holder.myThreadLocal.foo();
+    }
+
+    public static void main(String[] args) throws RunnerException {
+        if (false) {
+            ThreadLocalScopeBenchmark b = new ThreadLocalScopeBenchmark();
+            b.setup();
+            for (int i = 0; i < 100; i++) {
+                b.bench();
+            }
+            return;
+        }
+
+        Options opt = new OptionsBuilder()
+            .include(ThreadLocalScopeBenchmark.class.getName() + ".*")
+            .warmupIterations(10)
+            .measurementIterations(10)
+            .mode(Mode.AverageTime)
+            .timeUnit(TimeUnit.NANOSECONDS)
+            .forks(1)
+            .build();
+        new Runner(opt).run();
+    }
+
+    @Singleton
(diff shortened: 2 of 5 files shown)


context/src/main/java/io/micronaut/runtime/context/scope/ThreadLocal.java
```diff
@@ -35,4 +36,13 @@
...
 public @interface ThreadLocal {
+    /**
+     * If enabled, track bean life cycle. This means that the bean will be stopped, destroy
+     * listeners will be called etc., when the application context is closed or when the associated
+     * thread terminates. Note that this adds some overhead, so it's off by default.
+     *
+     * @return Whether to enable lifecycle support for this bean
+     * @since 4.9.0
+     */
+    boolean lifecycle() default false;
```
graemerocher, line 47: is turning this off by default not a breaking change?

graemerocher (approved):

context/src/main/java/io/micronaut/runtime/context/scope/ThreadLocal.java
yawkat: It only worked for beans that were local to the thread that runs stop() previously, so I think it's more of a breaking change to set it to true than it is to set it to false

dstepanov (approved):

dstepanov:
IMHO thread scope should be deprecated, I don't think it fits in our async model.

It would be nice to add bechmarks for our events/listeners stuff... but that's unrelated...

yawkat:
I think it still makes sense for some use cases such as encapsulating non-thread-safe classes (eg SimpleDateFormat).

However with virtual threads this becomes more problematic

PR #11686, opened 2025-03-25
Fix Lombok builder with private access

dstepanov:


graemerocher:
Needs to target 4.7.x and merge up

graemerocher (approved):

PR #11692, opened 2025-03-27
Clarify the JavaDocs of PropagatedContext.

mikehearn:


core/src/main/java/io/micronaut/core/propagation/ThreadPropagatedContextElement.java
```diff
@@ -19,10 +19,11 @@
...
 
 /**
- * Special version of {@link PropagatedContextElement} that is thread-aware and can update the thread context on the initial propagation entry
- * and restore it on the propagation exist.
+ * A {@link PropagatedContextElement} that bridges to code that uses raw thread locals instead of
+ * {@link PropagatedContext}. The object returned by {@link #updateThreadContext()} will be given
+ * to {@link #restoreThreadContext(Object)} when the propagation scope is closed.
  *
- * @param <S> The restore state type
+ * @param <S> The type held by the wrapped {@link ThreadLocal}.
```
dstepanov: I wouldn't mention implementation details about `ThreadLocal` in the future we are most likely going to switch to the scoped variables

dstepanov (changes requested):

core/src/main/java/io/micronaut/core/propagation/ThreadPropagatedContextElement.java
mikehearn: Fixed.

dstepanov (approved):

PR #11714, opened 2025-04-04
Fix ordering of bean event listeners when @Order is used and not the Ordered interface.

### LINKED ISSUE #10221: BeanEvent listener triggering order does not respect @Order
### Expected Behavior

BeanEvent listener triggering order is determined by @Order annotations and is deterministic if no default order is provided.

### Actual Behaviour

BeanEvent listener triggering order seems to be non-deterministic and it seems that @Order annotations on the event listener beans have no effect.

### Steps To Reproduce

Create multible BeanCreatedEvent implementations or the same Bean.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.2.0

mikehearn:
Fixes #10221

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -1264,7 +1264,7 @@ public class DefaultBeanContext implements InitializableBeanContext, Configurabl
     @NonNull
     private <T> T triggerPreDestroyListeners(@NonNull BeanDefinition<T> beanDefinition, @NonNull T bean) {
         if (beanPreDestroyEventListeners == null) {
-            beanPreDestroyEventListeners = loadListeners(BeanPreDestroyEventListener.class);
+            beanPreDestroyEventListeners = loadBeanEventListeners(BeanPreDestroyEventListener.class);
         }
         if (!beanPreDestroyEventListeners.isEmpty()) {
             Class<T> beanType = getBeanType(beanDefinition);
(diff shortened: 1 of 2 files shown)


mikehearn:
The CI failure looks like it might be spurious:

```
  net.bytebuddy.pool.TypePool$Resolution$NoSuchTypeException: Cannot resolve type description for com.google.common.base.Optional
```

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -4157,32 +4157,47 @@ private final class EventListenerListenersSupplier<T extends EventListener> impl
...
-                    listeners.add(listener);
+                    registrations.add(registration);
+                }
+                // Avoiding the simpler Comparator.comparing(BeanRegistration::getOrder) to keep lambda setup machinery
+                // off the critical startup path.
+                registrations.sort(BeanRegistrationOrderComparator.INSTANCE);
+                for (BeanRegistration<T> registration : registrations) {
+                    listeners.add(registration.getBean());
                 }
-                OrderUtil.sort(listeners);
```
graemerocher, line 4181: unclear to me why this fix is needed. `BeanRegistration` implements `Ordered`  which intializaes the order either from the definition or the bean itself. Why is this not working?

mikehearn: Previously OrderUtil.sort() was being called on a list of the beans themselves, not the `BeanRegistration` objects. However it's a good point that a custom comparator isn't needed in this case, so I removed that.

graemerocher (approved):

PR #11722, opened 2025-04-07
Add support multipart request parts in `@RequestBean`

### LINKED ISSUE #11682: Add support multipart request parts in RequestBean
### Feature description

Test request:
Image

Example controller on spring boot:
```java
    @PostMapping(value = "/testMultipart/{pathVar}", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
    public void testMultipart(WrappedRq rq) {
        log.info("fileName: {}", rq.fileName.getBytes());
    }

    public record WrappedRq(
        @PathVariable
        String pathVar,
        @RequestParam
        String queryVar,
        // Part from multipart body
        String fileName,
        // Part from multipart body
        MultipartFile file
    ) {}
```
Request debug:

Image

As you see, attributes `String fileName` and `MultipartFile file` filled correctly.

Example on micronaut:
```java
    @Post(value = "/testMultipart/{pathVar}", consumes = MediaType.MULTIPART_FORM_DATA)
    public void testMultipart(@RequestBean WrappedRq rq) {
        log.info("fileName: {}", rq.fileName.getBytes());
    }

    @Introspected
    public record WrappedRq(
        @PathVariable
        String pathVar,
        @QueryValue
        String queryVar,
        // Part from multipart body
        String fileName,
        // Part from multipart body
        CompletedFileUpload file
    ) {
    }
```

Request debug:

Image

I see two problems that would be good to fix:

1. At the moment, even if you specify the `@Part` annotations to the fileName and file attributes, Micronaut still won't understand what to put there. That is, at the moment it is impossible to use RequestBean and multipartfirmdata normally.

2. Micronaut cannot correctly determine that the parameter of the endpoint method describes a `@RequestBean` unless you explicitly specify the annotation. Although this can be easily done if you analyze the `WrappedRq` class and see that the value is Consumes.

In addition, Spring Boot understands constructions with a list of multipart files and puts all the parsed files there

altro3:
Fixed #11682

### DIFF
--- a/http/src/main/java/io/micronaut/http/annotation/Part.java
+++ b/http/src/main/java/io/micronaut/http/annotation/Part.java
@@ -35,7 +35,7 @@ import java.lang.annotation.Target;
  */
 @Documented
 @Retention(RUNTIME)
-@Target({ElementType.PARAMETER})
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.ANNOTATION_TYPE})
 @Bindable
 @Inherited
 public @interface Part {


altro3:
@graemerocher @dstepanov @yawkat Could you help with this feature?

This change is just the first step to be able to put the Part annotation at the field level in the RequestBean class. But how to add correct processing is not yet clear. Multipart request processing is implemented differently... It is hard for me to figure out where and what needs to be fixed to add this feature

yawkat:
unfortunately as you say multipart is handled specially depending on server implementation, and is pretty complicated. this feature will probably have to wait for a unified multipart api

altro3:
Any plans to do it? Ithink this is important to people, who want to migrate from spring to micronaut. Currently I foun only this feature thats not supported by micronaut, but supported by spring boot

dstepanov:
@altro3 The parts are loaded async. Check `NettyPartUploadAnnotationBinder` and see that it returns `PendingRequestBindingResult` which will return `Optional.empty` if it's not yet resolved.
To implement this case you would need to convert the bind method of `RequestBeanAnnotationBinder` to return `PendingRequestBindingResult` with proper check for all the arguments `isPending`.

yawkat:
@altro3 multipart improvements are WIP but it's a bit stuck in review on the netty side atm. https://github.com/netty-contrib/codec-multipart/pull/25

maybe that situation will improve now that netty 4.2 is out.

altro3:
oh, understand...

@yawkat Unfortunately the project looks dead... :-(

yawkat:
It's not exactly dead, but there's just not very many people that are knowledgeable enough about netty and multipart to do such a big review. I've definitely not given up on it, it is a medium term priority for me.

PR #11723, opened 2025-04-07
PreMarching filter errors should be filtered as ordinary request filters errors

dstepanov:
When error occurs the exeption mapper response should be processed by the response filters

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ResponseFilterOnPreMatchExceptionHandlerTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/filter/ResponseFilterOnPreMatchExceptionHandlerTest.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2017-2023 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


yawkat:
This does not solve the issue I mentioned. With this change, normal response filters still run when a PreMatching request filter returns an early response, which is undesirable.

Response filters should run if and only if a request filter of the same priority also runs.

dstepanov:
I would say it’s desirable considering JaxRs is working in the same way, if the request pre matching commits a response the response filters run, the behavior matches the ordinary request filters.

yawkat:
No, normal filters work like I describe.

```java
@RequestFilter("/foo")
@Order(0)
HttpResponse<String> requestFilterA() {
    System.out.println("RequestFilterA");
    return HttpResponse.ok("foo");
}

@ResponseFilter("/foo")
@Order(-1)
void responseFilterB() {
    System.out.println("ResponseFilterB");
}

@ResponseFilter("/foo")
@Order(1)
void responseFilterC() {
    System.out.println("ResponseFilterC");
}
```

This will run filters A and B but not filter C.

With your change, if there is a PreRouting filter, B and C can run without A running at all.

dstepanov:
No it's not, in your example the prerouting filters work the same way, filter C will not run.

PR #11704, opened 2025-03-31
Event loop loom carrier

yawkat:
Adds support for running loom threads on the event loop.

PR #11736, opened 2025-04-10
New client pool implementation

yawkat:
#8100 overhauled the connection pooling, but its primary purpose was to improve the separation of concerns, and to extend connection pooling to non-exchange methods. From an efficiency standpoint, it is not very good. In particular, it uses a central "doSomeWork" method to dispatch requests onto connections, which is shared between all threads and only dispatches in a serial fashion. The connections were also shared with no regard to which event loop they were created on.

#11300 bolted on "client affinity" functionality: Requests arriving on an event loop would preferentially be dispatched to connections running on the same event loop, which can reduce context switches and improve latency. However, the actual dispatch still happens in the central `doSomeWork` method serially, now further weighed down by the client affinity algorithm.

Benchmarks connected to #11704 show that this model breaks down in scenarios with many connections serving many requests on multiple event loops. `doSomeWork` becomes a bottleneck, since only one event loop can execute it at a time, and the connection selection algorithm for client affinity makes this even heavier.

This PR introduces a new connection pool built with scalability in mind from the start. Instead of a global pool of connections, the pool is split into "local pools", one for each event loop. Requests that are created on one event loop can be dispatched purely within that loop, without coordination with other loops. Contention between loops is kept to a minimum.

If load is uneven between event loops, requests can still spill into a global request queue as a fallback, and be picked up by less busy loops. This ensures progress, but should not be a hot path. The logic for moving between event loops or into the global queue is unfortunately quite complex.


http-client-core/src/main/java/io/micronaut/http/client/loadbalance/FixedLoadBalancer.java
```diff
@@ -84,6 +87,11 @@ public URI getUri() {
         return uri;
     }
 
+    @Internal
+    public ServiceInstance getServiceInstance() {
```
graemerocher, line 96: add javadoc that this method is considered internal and shouldn't be used.

http-client/src/main/java/io/micronaut/http/client/netty/PoolResizer.java
```diff
@@ -22,410 +22,1083 @@
...
+                if (!toDispatch.tryCompleteExceptionally(e)) {
+                    // this is probably fine, log it anyway
+                    log.debug("Failure during connection dispatch operation, but dispatch request was already complete.", e);
                 }
-            });
-            if (endState == WorkState.IDLE) {
-                // nothing else to do \o/
-                break;
+            } catch (Exception f) {
+                log.error("Internal error", f);
```
graemerocher: improve this error message and include the message from the exception.

graemerocher (changes requested):
If we are getting a new design can we use this opportunity to expose some APIs to collect http client pool metrics?

yawkat:
i don't really want to expose structured internals of the connection pool, it constrains the design too much. unstructured data, sure, but i don't think we have anything in core for that?

maybe we need a metrics api in core

graemerocher (approved):

yawkat:
@graemerocher as discussed, i added a pool abstraction + the old pool implementation in case there are any issues with the new one

graemerocher:
can you address the sonar failure please?

http-client/src/main/java/io/micronaut/http/client/netty/Pool40.java
```diff
@@ -297,19 +352,20 @@ final void addPendingRequest(PendingRequest sink) {
...
     }
 
-    final void markConnectionAvailable() {
+    void markConnectionAvailable() {
         dirty();
     }
 
-    final void forEachConnection(Consumer<ResizerConnection> c) {
+    @Override
+    public void forEachConnection(Consumer<Pool49.ResizerConnection> c) {
```
yawkat:
```suggestion
    public void forEachConnection(Consumer<Pool.ResizerConnection> c) {
```

yawkat:
the old pool impl isnt covered by tests anymore, but i ran the test suite manually with it enabled and it worked fine.

Issue #10485, opened 2024-02-12
Update to Kotlin 2.0


sdelamo:
Micronaut Framework 5 should update to Kotlin 2.0. 

[Kotlin 2.0.0-Beta3﻿](https://github.com/JetBrains/kotlin/releases/tag/v2.0.0-Beta3) was released on January 2024. 

The biggest change is the [K2 Compiler](https://blog.jetbrains.com/kotlin/2023/02/k2-kotlin-2-0/)

rlconst:
Just in case even Micronaut 3 happily live in K2

sgammon:
@rlconst Not under latest KSP...

scprek:
Not critical, but The OTEL BOM packaged with micronaut platform causes a kotlin 2.X bump IF you use the OTEL kotlin extensions library. So we are downgrading it atm.

We did this because a while back we hit issues with OTEL micronaut and the OTEL Java agent (which we use) but I'm guessing there's a way to disable the micronaut auto configure aspect and just leverage the bean OTEL things.

scprek:
Also the micronaut test bom bumps up `io.mockk` to a version that bumps to `kotlin 2.x`. Should we continue to force downgrade these types of issues until micronaut 5 has first class support for kotlin 2.X?

scprek:
Ok I think this is only the OTEL kotlin extension not playing nice. Other libraries that pull in kotlin 2.x are not causing issue

Create any micronaut starter app and add the line

`implementation("io.opentelemetry:opentelemetry-extension-kotlin")` and 🤯 

```
Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The actual metadata version is 2.1.0, but the compiler version 1.9.0 can read versions up to 2.0.0.
The class is loaded from /Users/stevecprek/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.1.10/d3028429e7151d7a7c1a0d63a4f60eac86a87b91/kotlin-stdlib-2.1.10.jar!/kotlin/Unit.class
```

Not even doing anything with that library, simply including caused this.

colesnodgrass:
[...]

PR #11742, opened 2025-04-14
Support Kotlin 2 KSP 2

dstepanov:
PR doesn't add any Kotlin 2 / KSP 2 dependency or modify the existing. KSP and KSP 2 share the same code with some logical changes, so we don't need a new implementation of `inject-kotlin`.

Added new modules that are not supposted to be published:
- `test-inject-kotlin2-test` - just like `test-inject-kotlin-test` but with KSP2 support, somehow code compiled with Kotlin 1/2 cannot be used as a library with Kotlin 2/1.
- `test-suite-kotlin2-ksp2` - the copy of the kotlin suite that is using Kotlin 2 / KSP 2
- `test-inject-kotlin2-ksp2` - the copy of tests from `inject-kotlin` with modifications that are needed for KSP2

Those modules should replace code in equivalent modules in v5 and be deleted.

All the writers are changed to always generate the bytecode in the round it not in the final call. KSP had a problem of touching some AST elements after the round is completed. Source gen does some extra checks for generating the bytecode.

Fixes https://github.com/micronaut-projects/micronaut-core/issues/10485


sgammon:
@dstepanov Here to help with tests or other stuff if needed. Just tag me and we are happy to try a snapshot.

dstepanov:
Thanks, I have added Kotlin 2 KSP 2 tests, reporting bugs to the KSP project and trying if it's passing. Hopefully 4.9 and KSP 2.0.2 will work.

settings.gradle
```diff
@@ -1,5 +1,8 @@
 pluginManagement {
     repositories {
+        maven {
+            url = uri("https://oss.sonatype.org/content/repositories/snapshots")
+        }
```
sdelamo:
```suggestion
```

can we remove the snapshot repository?

test-inject-kotlin2-ksp2/build.gradle.kts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    testImplementation(libs.micronaut.validation.processor) {
+        exclude(group = "io.micronaut")
+    }
+    testImplementation(libs.javax.persistence)
+    testImplementation(projects.micronautRuntime)
+    testImplementation(libs.neo4j.bolt)
+//    testImplementation(libs.managed.kotlinx.coroutines.core)
+//    testImplementation(libs.managed.kotlinx.coroutines.jdk8)
+//    testImplementation(libs.managed.kotlinx.coroutines.rx2)
```
sdelamo:
```suggestion
```

sdelamo (approved):

graemerocher (approved):

altro3:
@dstepanov What about test for kotlin2, but ksp1 ? Sample release: https://github.com/google/ksp/releases/tag/2.1.20-1.0.32

I mean, people say that there are too many bugs in ksp2, so I don't think it's worth giving up support for ksp1 for kotlin2

dstepanov:
We don’t give up support for ksp 1. Not sure about that combination. You can try

PR #11746, opened 2025-04-15
doc: health indicators

sdelamo:


graemerocher:
might be good to link to the javadoc for each class in the table

graemerocher (approved):

PR #11756, opened 2025-04-16
Include Argument in BeanCreatedEvent

patrick96:
Provides the `Argument<T>` instance that was used to create the bean in the `BeanCreatedEvent`.

In many cases, this is just `Argument.of(beanType)`, but for `createRegistration` this is actually a proper `Argument` instance provided by the user or the dependency injector.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
+++ b/inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
@@ -16,6 +16,8 @@
 package io.micronaut.context.event;
 
 import io.micronaut.context.BeanContext;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.type.Argument;
 import io.micronaut.inject.BeanDefinition;
 import io.micronaut.inject.BeanIdentifier;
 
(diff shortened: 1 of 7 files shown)


inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
```diff
@@ -48,4 +57,8 @@ public BeanCreatedEvent(BeanContext beanContext, BeanDefinition<T> beanDefinitio
     public BeanIdentifier getBeanIdentifier() {
         return beanIdentifier;
     }
+
+    public Argument<T> getArgument() {
```
graemerocher: needs javadoc, but I would call it `getBeanType()`

inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
```diff
@@ -31,15 +32,23 @@ public class BeanCreatedEvent<T> extends BeanEvent<T> {
...
      * @param beanIdentifier The bean identifier
+     * @param argument       The {@link Argument} used to create the bean
      * @param bean           The bean
      */
-    public BeanCreatedEvent(BeanContext beanContext, BeanDefinition<T> beanDefinition, BeanIdentifier beanIdentifier, T bean) {
+    public BeanCreatedEvent(BeanContext beanContext,
+                            BeanDefinition<T> beanDefinition,
+                            BeanIdentifier beanIdentifier,
+                            Argument<T> argument,
+                            T bean) {
```
graemerocher, line 56: this is a breaking API change. You will need to add a new constructor with the argument (I would call the parameter beanType) and delegate to this new constructor. You can use `beanDefinition.asArgument()` as a default value to maintain compatibility.

inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
```diff
@@ -31,15 +32,23 @@ public class BeanCreatedEvent<T> extends BeanEvent<T> {
 
     private final BeanIdentifier beanIdentifier;
 
+    private final Argument<T> argument;
```
graemerocher: rename to `beanType`

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
patrick96: Thanks for catching that. Did not consider that user code could also create instances of this

patrick96:
Renamed variables to `beanType` everywhere

patrick96:
I also added some tests. This should now be ready for reviews

graemerocher:
@patrick96 can you re-target to the 4.9.x branch this is this is not a bug fix

inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
```diff
@@ -48,4 +64,11 @@ public BeanCreatedEvent(BeanContext beanContext, BeanDefinition<T> beanDefinitio
     public BeanIdentifier getBeanIdentifier() {
         return beanIdentifier;
     }
+
+    /**
+     * @return The argument used to create the bean
+     */
+    public Argument<T> getBeanType() {
```
graemerocher, line 74: please add `@since 4.9.0` and `@io.micronaut.core.annotation.NonNull` to the return type

graemerocher (changes requested):

inject/src/main/java/io/micronaut/context/event/BeanCreatedEvent.java
patrick96: Done

patrick96:
Re-targeted to 4.9.x

graemerocher (approved):

PR #11766, opened 2025-04-21
improve error message when multiple bean candidates are present and all are Secondary

edgao:
e.g. if you have this code: (kotlin syntax, hopefully the intent is clear)
```kotlin
@Singleton
class SomeOtherBean(val foo: Foo)

interface Foo

@Singleton
@Secondary
class Foo1 : Foo

@Singleton
@Secondary
class Foo2 : Foo
```

micronaut throws a cryptic `java.util.NoSuchElementException`, because it blindly calls `iterator.next()`.

This PR just adds a check for empty list, and throws a more informative exception.

(wasn't sure how to update the unit tests to verify this - DefaultBeanContextSpec didn't seem like the right place? I'd appreciate a pointer to the relevant test suite)

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -3278,13 +3278,13 @@ public class DefaultBeanContext implements InitializableBeanContext, Configurabl
         if (candidates.size() == 1) {
             return candidates.iterator().next();
         }
-        Collection<BeanDefinition<T>> candidates originalCandidates = candidates;
+        Collection<BeanDefinition<T>> originalCandidates = candidates;
         candidates = candidates.stream().filter(candidate -> !candidate.hasDeclaredStereotype(Secondary.class)).toList();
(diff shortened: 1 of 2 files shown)


frifri (approved):

graemerocher:
could you add a test please

edgao:
wasn't sure where to put tests 😅 is there an existing test suite? (DefaultBeanContextSpec didn't look right, but maybe I'm looking in the wrong place)

graemerocher:
probably can go in test-suite submodule

edgao:
done - verified locally that without the runtime diff, the test fails.

I couldn't figure out how to use `e.getPossibleCandidates()` - AFAICT it returned an empty iterator? So I just asserted on the exception message. Would appreciate a tip if there's a better way to do this, or if I'm instantiating the exception incorrectly.

edgao:
@graemerocher is there anything else I should do here?

edgao:
looks like CI is green! is this PR ready to go?

dstepanov (approved):

Issue #11655, opened 2025-03-13
Provide `java.time.Clock` as a bean and use it in every component that needs access to current time


bfg:
### Feature description

I was recently developing a rest service that uses `micronaut-security` and `micronaut-data` with several oauth2 providers to implement authentication.

Testing such endpoints with *real, but expired* jwt tokens was very hard, due to a fact that there's currently no way to inject a clock that points to a temporal in the past - jwt tokens do expire. I have my own `Clock` implementation for testing purposes that has ability to set a time and I've had to re-implement certain `micronaut-security` claims validators that query current time to support full end-to-end testing during build time.

So I propose that `micronaut-core` creates `Clock` singleton bean by default using `Clock.systemDefaultZone()` and all components that query current time inject clock bean and obtain current time using `Clock.instant()` instead of `Instant.now()` (or other `java.time.XXX.now()` friends).

This would also benefit `micronaut-security`'s `ExpirationJwtClaimsValidator` and `micronaut-data`'s `DateTimeProvider` that uses current time for `@DateCreated` and `@DateUpdated`.

This is just a feature request - I'm also willing to contribute code if maintainers think that this feature is worth including in a framework.

yawkat:
I think this would be a good feature, I don't see any downsides. InstantSource would be even better. Feel free to contribute!

bfg:
> I think this would be a good feature, I don't see any downsides. InstantSource would be even better. Feel free to contribute!

Ok, against which branch should I create PR?

yawkat:
4.9

aurelien-poscia-sonarsource:
@bfg I have exactly the same problem / need to mock time to be able to hard-code jwts for testing. Did you make any progresses on a PR ? I can also help and contribute if you bootstraped something :)

bfg:
> hard-core jwts for testing. Did you make any progresses on a PR ? I can also help and contribute if you bootstraped something :)

Well, I had to disable built-in `ExpirationJwtClaimsValidator`/`NotBeforeJwtClaimsValidator` and replace them with my own implementations with injected `Clock` implementation.

aurelien-poscia-sonarsource:
Thanks for your reply, what about your proposal to open a pull request in micronaut to make the Clock injection possible?

bfg:
> Thanks for your reply, what about your proposal to open a pull request in micronaut to make the Clock injection possible?

[...]

PR #11774, opened 2025-04-23
 Added factory class for InstantSource bean creation

bfg:
For motivation and rationale see #11655

### DIFF
--- a/context/src/main/java/io/micronaut/context/time/InstantSourceFactory.java
+++ b/context/src/main/java/io/micronaut/context/time/InstantSourceFactory.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context.time;
+
+import io.micronaut.context.annotation.Bean;
+import io.micronaut.context.annotation.Factory;
+import jakarta.inject.Singleton;
+
+import java.time.Clock;
+import java.time.InstantSource;
+
+/**
+ * Factory class that creates {@link InstantSource} bean.
+ */
+@Factory
+public final class InstantSourceFactory {
+    /**
+     * Creates {@link InstantSource} from a system clock.
+     *
+     * @return The system clock
+     */
+    @Bean
+    @Singleton
+    public InstantSource systemClock() {
+        return Clock.systemDefaultZone();
+    }
+}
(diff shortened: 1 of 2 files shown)


bfg:
Hey @yawkat, I've created a simple draft for this; I have few questions

* did I place factory class in a correct module and package?
* maybe we should consider creating full `Clock` type in the future so that client code can query timezone as well?

yawkat:
The location seems reasonable to me. @graemerocher WDYT

I don't think we should expose a full Clock. The server time zone is not something that microservices should touch anyway.

bfg:
> The location seems reasonable to me. @graemerocher WDYT
> 
> I don't think we should expose a full Clock. The server time zone is not something that microservices should touch anyway.

just FYI: I've searched the code and `java.time.ZoneId` is imported in the following classes:

```
./http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/DateTimeElement.java
./context/src/main/java/io/micronaut/scheduling/annotation/Scheduled.java
./context/src/main/java/io/micronaut/scheduling/NextFireTime.java
./context/src/main/java/io/micronaut/scheduling/ScheduledExecutorTaskScheduler.java
./context/src/main/java/io/micronaut/runtime/converters/time/TimeConverterRegistrar.java
./core/src/main/java/io/micronaut/core/reflect/ClassUtils.java
./http/src/main/java/io/micronaut/http/MutableHttpHeaders.java
./http/src/main/java/io/micronaut/http/HttpHeaders.java
./http-netty/src/main/java/io/micronaut/http/netty/NettyHttpHeaders.java
```

... I guess that `scheduling` package might need zone, because it's using system default. Dunno, we might be able to work with system default.

In addition it's being used in `micronaut-data` as well; system default zone is being used by default.

```
./data-r2dbc/src/main/java/io/micronaut/data/r2dbc/mapper/ColumnIndexR2dbcResultReader.java:import java.time.ZoneId;
./data-r2dbc/src/main/java/io/micronaut/data/r2dbc/mapper/ColumnNameR2dbcResultReader.java:import java.time.ZoneId;
./data-runtime/src/main/java/io/micronaut/data/runtime/convert/DataConversionServiceFactory.java:import java.time.ZoneId;
./data-document-model/src/main/java/io/micronaut/data/document/model/query/builder/MongoQueryBuilder.java:import java.time.ZoneId;
./data-document-model/src/main/java/io/micronaut/data/document/model/query/builder/MongoQueryBuilder2.java:import java.time.ZoneId;
```

mikehearn:
The PR looks good and I've implemented something similar in my own project that uses Micronaut. Could you maybe take it out of draft (and I guess now, rebase to 4.10.x as it's new API?). I think when a PR is marked as in draft, the maintainers don't look at it.

yawkat:
There's also a new Ticker API in netty that it would be cool to integrate with: https://github.com/netty/netty/pull/15026/files#diff-df3131658fad36f38d0be80c7870224d302810296ef4a42b79ffe02113443a8a

But since that is nanoTime-based, I'm not sure it's feasible.

bfg:
@mikehearn I was/still am waiting for feedback 🤷 In my project I've included `Clock` bean.
@yawkat I guess `nanoTime` is actually not needed in the framework

yawkat (approved):

graemerocher (approved):

yawkat:
Thanks!

bfg:
@yawkat thanks! I'll create PRs in other sub-projects so that this bean can become actually useful. Thanks again!

yawkat:
sorry it took so long :)

dstepanov:
I think it should be created only if there is one missing

PR #11787, opened 2025-04-28
Feature/Support for QueryValue as a object

MarianConstantinMarica:


yawkat:
@dstepanov can you think of a better way to do this? otherwise i think it's ok

yawkat (approved):

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,49 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+                                      Argument<T> argument) {
+        Optional<BeanIntrospection<T>> introspectionOpt = BeanIntrospector.SHARED.findIntrospection(argument.getType());
+        if (introspectionOpt.isEmpty()) {
+            return BindingResult.unsatisfied();
+        }
+
+        BeanIntrospection<T> introspection = introspectionOpt.get();
+        T instance;
+        try {
+            instance = introspection.instantiate();
```
altro3: Why no-args constructor is required? I'think you need more tests with `record` and required args constructor (without no-arg)

altro3 (changes requested):

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,49 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+                                      HttpRequest<?> source,
+                                      AnnotationMetadata annotationMetadata,
+                                      ConvertibleMultiValues<String> parameters,
+                                      Argument<T> argument) {
+        Optional<BeanIntrospection<T>> introspectionOpt = BeanIntrospector.SHARED.findIntrospection(argument.getType());
+        if (introspectionOpt.isEmpty()) {
+            return BindingResult.unsatisfied();
+        }
+
+        BeanIntrospection<T> introspection = introspectionOpt.get();
```
graemerocher, line 159: this should be altered to use `introspection.builder()` so that records, immutable types, types with custom constructors etc. are supported.

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
MarianConstantinMarica: I will update!

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
graemerocher: When using `IntrospectionBuilder` API and finally call the `build()` method any exception should b caught and the conversion rejected with `context.reject`

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
MarianConstantinMarica: I made the changes with the introspection.builder(). Seems to work for object with custom constructor but I made a test for record and don't work. Am I missing something?

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,51 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+        BeanIntrospection.Builder<T> introspectionBuilder = introspection.builder();
```
graemerocher, line 160: to maximise performance you should call [getBuilderArguments](https://docs.micronaut.io/latest/api/io/micronaut/core/beans/BeanIntrospection.Builder.html#getBuilderArguments())() here

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,51 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+
+        for (BeanProperty<T, Object> property : introspection.getBeanProperties()) {
```
graemerocher: iterative here over the [getBuilderArguments](https://docs.micronaut.io/latest/api/io/micronaut/core/beans/BeanIntrospection.Builder.html#getBuilderArguments())() using array iteration to obtain the index

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,51 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+            String propertyName = property.getName();
+            Optional<String> value = parameters.getFirst(propertyName);
+
+            if (value.isEmpty()) {
+                value = property.asArgument()
```
graemerocher: remove the call to `asArgument` and just use the argument from builder arguments above

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,51 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+                    .getAnnotationMetadata()
+                    .stringValue(Bindable.class, "defaultValue");
+            }
+
+            if (value.isPresent()) {
+                Optional<Object> converted = conversionService.convert(value.get(), context.with(property.asArgument()));
```
graemerocher: remove this call to `asArgument()` and just use builder argument as above

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -130,6 +146,51 @@ public BindingResult<T> bind(ArgumentConversionContext<T> context, HttpRequest<?
...
+                if (converted.isPresent()) {
+                    try {
+                        introspectionBuilder.with(propertyName, converted.get());
```
graemerocher: invoke with passing the index from the array iteration https://docs.micronaut.io/latest/api/io/micronaut/core/beans/BeanIntrospection.Builder.html#with(int,io.micronaut.core.type.Argument,A)

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
MarianConstantinMarica: I tried with getBuilderArguments() iteration, still not working for the records

http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
graemerocher: this is because records are not supported in Groovy and you a rewriting the test in Groovy. Probably will be fixed by https://github.com/micronaut-projects/micronaut-core/pull/11832

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/PaginationRequest.java
```diff
@@ -0,0 +1,6 @@
+package io.micronaut.http.server.netty.binding;
+
+import io.micronaut.core.annotation.Introspected;
+
+@Introspected
+public record PaginationRequest(Integer page, Integer size) {}
```
altro3, line 6: @MarianConstantinMarica just rename `PaginationRequest.java` to `PaginationRequest.groovy` and after this all must work correctly

graemerocher (approved):

PR #11795, opened 2025-05-07
Fix HttpClientException serviceId

### LINKED ISSUE #11794: ReadTimeoutException is missing "serviceId"
### Expected Behavior

To be able to detect which service did not respond within timeout, attribute "serviceId" should be set for ReadTimeoutException.

### Actual Behaviour

 attribute "serviceId" is null

### Steps To Reproduce

From reading the code https://github.com/micronaut-projects/micronaut-core/blob/4.9.x/http-client-core/src/main/java/io/micronaut/http/client/exceptions/ReadTimeoutException.java
you can see that the only existing constructor of ReadTimeoutException calls the constructor of its superclass HttpClientException with parameter "shared" true.

See 
https://github.com/micronaut-projects/micronaut-core/blob/2395c237f96706bef3d9c0648b01f57462501abd/http-client-core/src/main/java/io/micronaut/http/client/exceptions/ReadTimeoutException.java#L29
and 
https://github.com/micronaut-projects/micronaut-core/blob/2395c237f96706bef3d9c0648b01f57462501abd/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java#L57

This set attribute "serviceIdLocked" to true, which then never allows to set the "serviceId".
See
https://github.com/micronaut-projects/micronaut-core/blob/2395c237f96706bef3d9c0648b01f57462501abd/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java#L79

### Environment Information

_No response_

### Example Application

_No response_

### Version

Micronaut Core 4.8.11
Micronaut Platform 4.8.2

altro3:
Fixed #11794

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
@@ -54,7 +54,6 @@ public class HttpClientException extends HttpException {
         if (!shared) {
             throw new IllegalArgumentException("shared must be true");
         }
-        serviceIdLocked = true;
     }
 
     /**


graemerocher:
is there a test we can add for this?

http-client-core/src/main/java/io/micronaut/http/client/exceptions/HttpClientException.java
```diff
@@ -54,7 +54,6 @@ public HttpClientException(String message, Throwable cause, boolean shared) {
         if (!shared) {
             throw new IllegalArgumentException("shared must be true");
         }
-        serviceIdLocked = true;
```
yawkat, line 57: can't do this, the exception is shared.

altro3: And is it possible to fix it somehow?

PR #11798, opened 2025-05-09
fix config properties refresh for each property

graemerocher:
The refresh endpoint doesn't work for `@EachProperty` beans. This means that datasource properties are not refreshed from the `/refresh` endpoint and other periodic refresh mechanisms. The reason this is broken is because:

* The `@ConfigurationProperties` stereotype is used instead of `@ConfigurationReader`
* The each property prefix is not handled
* The `inject` method of the bean context doesn't set the path.


runtime/src/test/groovy/io/micronaut/runtime/context/scope/RefreshScopeSpec.groovy
```diff
@@ -138,13 +140,15 @@ class RefreshScopeSpec extends Specification {
```
dstepanov, line 144: Add          bean.testEachProps() == 'test'

runtime/src/test/groovy/io/micronaut/runtime/context/scope/RefreshScopeSpec.groovy
```diff
@@ -138,13 +140,15 @@ class RefreshScopeSpec extends Specification {
...
         System.setProperty("foo.bar", "bar")
+        System.setProperty("foos.test.bar", "bar")
         Environment environment = beanContext.getEnvironment()
         Map<String, Object> previousValues = environment.refreshAndDiff()
         beanContext.publishEvent(new RefreshEvent(previousValues))
 
         then:
         bean.testValue() == 'bar'
         bean.testConfigProps() == 'bar'
+        bean.testEachProps() == 'bar'
```
dstepanov, line 156: Will this work for `@EachBean(EachConfig) class EachConfigBeans` ?

dstepanov (changes requested):

runtime/src/test/groovy/io/micronaut/runtime/context/scope/RefreshScopeSpec.groovy
dstepanov: And `@EachProperty(list=`?
The other use-cases might be combinations with `@Requires`

graemerocher: it is unrelated, they don't config

graemerocher: I don't think we can support requires without a context restart.

dstepanov (approved):

PR #11800, opened 2025-05-13
Add JFR event for HTTP requests

yawkat:
This was useful for loom-on-netty debugging.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/HttpRequestEvent.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/HttpRequestEvent.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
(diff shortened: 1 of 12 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/HttpRequestEvent.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * JFR event that tracks HTTP request lifetime.
+ *
+ * @since 4.9.0
+ * @author Jonas Konrad
+ */
+@Internal
+@StackTrace(false)
+@Enabled(false)
+abstract class HttpRequestEvent extends Event {
```
graemerocher: make `sealed`

graemerocher (approved):

dstepanov (approved):

PR #11802, opened 2025-05-14
Fix environment refreshAndDiff method populating old value

radovanradic:
`Environment.refreshAndDiff()` method returns previous value in map of changed values. I thought it should return new value, but not sure if there was reason it was implemented this way or it is actually a bug.


radovanradic:
Closing since it is a breaking change and probably this was intentional way to work since it is here for long time and someone would file a bug already.

PR #9696, opened 2023-08-09
Fix SSE Content-Type header

Katchahoy:
resolves #9627

PR #11812, opened 2025-05-20
fix: always return `Content-Type: text/event-stream` on SSE endpoints

### LINKED ISSUE #11810: SSE Endpoints Serving JSON Data Sporadically Return `Content-Type: application/json`
### Expected Behavior

Endpoints with `@Produces(TEXT_EVENT_STREAM)` always return the header `Content-Type: text/event-stream`.

### Actual Behaviour

An endpoint with `@Produces(TEXT_EVENT_STREAM)` that returns complex objects in the SSE event data (serialised to JSON) _sometimes_ returns the header `Content-Type: application/json`. In a resource-restricted Docker container (1.1 CPUs), this happens with a probability of roughly 1%.

When `Content-Type: application/json` is returned, everything else about the response is exactly as expected. Only the `Content-Type` header is wrong.

### Steps To Reproduce

 1. Create a micronaut application with an SSE endpoint returning JSON data. For example like [the Controller in `ServerSentEventSpec`](https://github.com/micronaut-projects/micronaut-core/blob/70247927df1b6f02f7d43a7e64d59575a09cdfc7/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/sse/ServerSentEventSpec.groovy#L132).
 2. Run the micronaut application in a resource-restricted environment. For example, run it in a docker container and assign it only 1.1 CPUs.
 3. Execute a test running many requests against the SSE endpoint and verify that the `Content-Type` header is `text/event-stream`.
 4. ~1% of tests will fail because the `Content-Type` header is `application/json`

### Environment Information

- Operating System: `Ubuntu 24.04.2 LTS`
- JDK Version: 23.0.2

### Example Application

_No response_

### Version

4.8.2

jGleitz:
fixes #11810

The root cause for #11810 was that the body writer for the content of SSE events received `MediaType.APPLICATION_JSON_TYPE` as `mediaType` when called from `TextStreamBodyWriter`. The called body writer would set that `mediaType` on the `MutableHeaders`. #9696 added a line after the call to the body writer that set the `Content-Type` back to `MediaType.TEXT_EVENT_STREAM_TYPE`. However, this fix left in a brief period during which the `Content-Type` header was set to `application/json`. If, by chance, the response headers were written to the wire precisely during this period, the `application/json` was written to the response headers.

This fix passes the `mediaType` received by `TextStreamBodyWriter` on to the inner body write, ensuring that the inner body writer doesn’t change the desired `Content-Type`.

I failed to find a way to reproduce the exact race condition for #11810 in a unit test. However, we were able to reproduce the race condition reliably using tens of thousands of requests against a resource-restricted docker container. These tests failed ~1% of the time, and are not failing at all after this fix.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/sse/ServerSentEventSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/sse/ServerSentEventSpec.groovy
@@ -36,6 +36,7 @@ import jakarta.inject.Inject
 import org.reactivestreams.Publisher
 import reactor.core.publisher.Flux
 import spock.lang.Specification
+import spock.lang.Unroll
 
 import java.time.Duration
 import java.time.temporal.ChronoUnit
@@ -91,9 +92,10 @@ class ServerSentEventSpec extends Specification {
         ex.message == "Client '/sse': Internal Server Error"
     }
 
+    @Unroll
     void "test Content-Type header is correct"() {
         def httpRequest = HttpRequest
-                .GET(server.getURI().toString() + "/sse/object")
+                .GET(server.getURI().toString() + "/sse" + path)
                 .accept(MediaType.TEXT_EVENT_STREAM)
 
         when:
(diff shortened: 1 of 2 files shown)


jGleitz:
I have signed the CLA (see screenshot below) but the status check doesn’t seem to pick that up.

grafik

yawkat (approved):

PR #11832, opened 2025-06-03
Add support records for groovy

altro3:
See https://github.com/micronaut-projects/micronaut-openapi/issues/2168

Groovy AST already supported groovy / java records, but now `getBeanProperties()` returns empty list for every record, because `GroovyClassElement` don't understand that class is record.

This PR added this lost functionality


altro3:
@graemerocher @dstepanov @sdelamo Look to this, pls

altro3:
@dstepanov Same problem with kotlin: can't use java records - getBeanProperties returns always empty list. But looks like this is bug in KSP

dstepanov (approved):

PR #11833, opened 2025-06-04
Test for processing kotlin with java records

altro3:


altro3:
@dstepanov @graemerocher reproducer for problem wiht java records and kotlin. The same as here: https://github.com/micronaut-projects/micronaut-core/pull/11832

altro3:
@dstepanov coukd you look to this? Is it possible to use java records with kotlin ?

PR #11846, opened 2025-06-10
Remove dynamic postfix for dummy file

altro3:
Replace `nanoTime()` postfix to random long for deterministic build (See: https://github.com/micronaut-projects/micronaut-openapi/issues/2180).

This fix solves old problem with KSP: when I want to create a directory in `META-INF` and call `visitMetaInfFile("swagger\dummy" + new Random().nextLong(Long.MAX_VALUE))` dummy file created, but after annotaion processing finised, dummy file still exists. 

This problem ONLY with KSP.

Example: how it works now:
{D4E1AF48-CA94-437F-873F-D50E5456DCF4}

After this fix and applied to micronaut openapi:
{D6955CB4-9491-4876-ADF1-96E4866D4C0F}

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
+++ b/core-processor/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
@@ -32,7 +32,6 @@ import io.micronaut.inject.writer.GeneratedFile;
 import java.net.URI;
 import java.net.URL;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
@@ -89,7 +88,7 @@ public interface VisitorContext extends MutableConvertibleValues<Object>, ClassW
      *
      * @return The annotation metadata builder
      *
-     * @since  4.0.0
+     * @since 4.0.0
      */
     @Internal
     @NonNull
@@ -176,13 +175,13 @@ public interface VisitorContext extends MutableConvertibleValues<Object>, ClassW
             return projectDir;
         }
         // let's find the projectDir
-        Optional<GeneratedFile> dummyFile = visitGeneratedFile("dummy" + System.nanoTime());
+        Optional<GeneratedFile> dummyFile = visitGeneratedFile("dummy");
         if (dummyFile.isPresent()) {
             URI uri = dummyFile.get().toURI();
             // happens in tests 'mem:///CLASS_OUTPUT/dummy'
             if (uri.getScheme() != null && !uri.getScheme().equals("mem")) {
                 // assume files are generated in 'build' or 'target' directories
-                Path dummy = Paths.get(uri).normalize();
+                Path dummy = Path.of(uri).normalize();
                 while (dummy != null) {
                     Path dummyFileName = dummy.getFileName();
                     if (dummyFileName != null && ("build".equals(dummyFileName.toString()) || "target".equals(dummyFileName.toString()))) {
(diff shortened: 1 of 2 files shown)


altro3:
@graemerocher @sdelamo @dstepanov look to this, pls

core-processor/src/main/java/io/micronaut/inject/writer/GeneratedFile.java
```diff
@@ -96,6 +96,11 @@ public interface GeneratedFile {
      */
     Writer openWriter() throws IOException;
 
+    /**
+     * Delete generated file object.
+     */
+    void delete();
```
dstepanov: Can you implement it without a new method just by deleting the path?

altro3: @dstepanov Yeah, you're right. But need add this logic to default implementation for mehtods `getClassesOutputPath` and `getProjectDir`. Fixed

dstepanov: BTW Is there a way not to create random files which will break caching? Maybe try to access the dummy file and if it exists calculate the pass from it?

altro3: How can this break caching? We create an empty file and immediately delete it. By the way, this is your solution, not mine. All I did was add a postfix to minimize the potential problem when a file with such a name exists.

But, again, how does this affect caching? Before running the method, we don't have a dummy file, after executing the method, we don't have one either. How does caching suffer from this?

graemerocher: creates random outputs vs inputs

dstepanov: @altro3 I bet Gradle tracks all the files created to create a cache key. A random file will cause a cache miss. I don't remember was what the old issue, but it would be nice to find a way not to create a random file, just try to access an existing one to get the path.

altro3: ok, here is the final solution:

1. Dynamic postfixes for dummy files: yes, they are not required (at least I did not see any problems in their absence). I remember that I added this because one of my processors created this file and when I tried to call the getProjectDir method, I got an error because of it. Well, let's try to remove the postfix, we'll see what happens.

2. About deleting files: unfortunately, KSP does not provide any tools to delete the generated file. @dstepanov this is about your suggestion. Therefore, it is better not to create dummy files, because you will not be able to delete them

3. Limitations of KSP: unfortunately, KSP does not provide any way to find out the current directory of the project. The only way is to create that dummy file, which then cannot be deleted.

So what I did:

1. In this module, I just removed the postfix - I still don't believe that it somehow affects the cache, because the dummy file is created once for the entire process, the path to the project is calculated and this calculated value is saved in the context. The dummy file is not created anymore. Sorry, but I don't see where the caching problem could be here. And the gradle did not have problems with the cache if you ran the build several times. Looks like a far-fetched problem.

2. The main changes are here: https://github.com/micronaut-projects/micronaut-openapi/pull/2183

3. Now I try to calculate the path to the project directory and the output path whenever possible, so as not to create dummy files again later.

4. The main improvements concern KSP: now I can only create a single dummy file to calculate the project directory. All other paths are calculated automatically.

5. In the documentation, I added that for KSP it is highly desirable to specify the micronaut.openapi.project.dir property - in this case, dummy files are not created at all and the build will be clean.

I think this is the maximum that can be done in the difficult conditions that KSP provides.

altro3:
@sdelamo this fix is very little. I don't think that we need to wait release 4.10.0 for it

PR #11848, opened 2025-06-10
Fix two buffer leaks

yawkat:
One from OSS-Fuzz, one from manual debugging. Latter has no test case because it's apparently a race condition.


graemerocher (approved):

graemerocher:
Remember this will need to be merged up

dstepanov (approved):

graemerocher:
some test seems to be failing on this one

PR #11849, opened 2025-06-10
Move to main HTTP/3 version

yawkat:


yawkat:
@melix to work around the dependency mess, I've created a "meta" module http-netty-http3 that people can depend on instead of directly on the broken netty dependencies. WDYT?

graemerocher (approved):

http-client/build.gradle.kts
```diff
@@ -1,4 +1,4 @@
-import org.apache.tools.ant.taskdefs.condition.Os
+metaimport org.apache.tools.ant.taskdefs.condition.Os
```
melix: Looks like a typo

melix (approved):
sgtm!

PR #11856, opened 2025-06-12
Disable Loom support on JVMs that support the virtual threads API but don't really implement it.

mikehearn:
This avoids a crash when a thread is cast to VirtualThread but is in fact a ThreadBuilder.BoundVirtualThread that uses a platform thread instead.

Espresso is an example of such a JVM.

@yawkat

### DIFF
--- a/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
+++ b/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
@@ -26,7 +26,6 @@ import java.lang.invoke.MethodType;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 
 /**
--- a/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
+++ b/context/src/main/java/io/micronaut/scheduling/LoomSupport.java
@@ -26,6 +26,7 @@ import java.lang.invoke.MethodType;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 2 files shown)


context/src/main/java/io/micronaut/scheduling/LoomSupport.java
```diff
@@ -71,10 +72,18 @@ public final class LoomSupport {
             isVirtual = MethodHandles.lookup()
                 .findVirtual(Thread.class, "isVirtual", MethodType.methodType(boolean.class));
 
-            // invoke, this will throw an UnsupportedOperationException if we don't have --enable-preview
-            ofVirtual.invoke();
+            // This will throw if this Java doesn't support Loom, or if it does but only with
+            // --enable-preview.
+            unstarted1("virtual threads support probe", null, () -> {
```
yawkat: the static fields `unstarted1` uses are unset here. How can this work?

context/src/main/java/io/micronaut/scheduling/LoomSupport.java
```diff
@@ -71,10 +72,18 @@ public final class LoomSupport {
...
+                try {
+                    // Check if the JVM supports a new enough standard library version yet doesn't
+                    // really implement virtual threads internally (e.g. Espresso as of June 2025).
+                    sup.set(Class.forName("java.lang.ThreadBuilders$BoundVirtualThread").isInstance(Thread.currentThread()));
```
yawkat: shouldn't it be the opposite? also looks like a race condition

mikehearn:
@yawkat Yes sorry, I pushed this code way too early intending to mark it as a draft but you got there too fast for me. The current version should pass tests and also fixes my project.

yawkat (approved):

sdelamo (changes requested):
checkstyle fails

sdelamo:
> checkstyle fails

I pushed a fix. @mikehearn run `./gradlew checkstyleMain` before pushing please 🙏

PR #11862, opened 2025-06-13
Remove managed Kotlin 2 KSP 2 from BOM

dstepanov:


melix (approved):

sdelamo:
without this change [source-gen](https://github.com/micronaut-projects/micronaut-sourcegen) fails

> The problem is with Kotlin, we compile with Kotlin 1, kotlinpoet brings dependecies from kotlin 2 but not it puls newest version of kotlin libs and breakes

melix:
For more context: a BOM cannot contain dependencies with the same group/artifact but different versions: build tools cannot "choose". Or, more precisely, Maven will use the first one it sees, and Gradle will perform optimistic upgrade. So we have to choose which version of the library we manage.

PR #11863, opened 2025-06-13
fix: Allow other executor instrumentations than context propagation

### LINKED ISSUE #11653: Thread propagation on executor services doesnt work with micrometer
### Expected Behavior

If you are using micronaut micrometer, thread context is propagated down any task that is run using an executor service 

### Actual Behaviour

if the [ExecutorServiceMetricsBinder](https://github.com/micronaut-projects/micronaut-micrometer/blob/5.10.x/micrometer-core/src/main/java/io/micronaut/configuration/metrics/binder/executor/ExecutorServiceMetricsBinder.java) bean is called before the [ExecutorServiceInstrumenter](https://github.com/micronaut-projects/micronaut-core/blob/d3b0c386962d3499539550544cba522ca772d75b/context-propagation/src/main/java/io/micronaut/context/propagation/instrument/execution/ExecutorServiceInstrumenter.java#L38), the executor service does not get instrumented with the PropagatedContext. When you move to a thread from the executor service, you lose the thread context like the `ServerRequestContext.currentRequest()`. If the metrics binder is run after the service instrumenter, the thread context is progated properly. 

I believe this is a unintended regression introduced in [this pr](https://github.com/micronaut-projects/micronaut-core/pull/11490) because `ExecutorServiceMetricsBinder` bean [wraps the executor service bean in an `InstrumentedExecutorService`](https://github.com/micronaut-projects/micronaut-micrometer/blob/5.10.x/micrometer-core/src/main/java/io/micronaut/configuration/metrics/binder/executor/ExecutorServiceMetricsBinder.java#L110).  If the `ExecutorServiceInstrumenter` is called _after_ the `ExecutorServiceMetricsBinder`, it now [checks that the executor service is already an `InstrumentedExecutorService`](https://github.com/micronaut-projects/micronaut-core/blob/d3b0c386962d3499539550544cba522ca772d75b/context-propagation/src/main/java/io/micronaut/context/propagation/instrument/execution/ExecutorServiceInstrumenter.java#L53-L55) so it does not apply its thread propagation.

### Steps To Reproduce

1. In the example repo, run the tests (`./gradlew test --rerun`) a few times. The test `PropagationControllerWithMicrometerSpec` would intermitatmly fail because the `PropagationStatusResponse.onIoThread` field will be false which means it cant access the current server request on the io thread.
2. The test `PropagationControllerWithoutMicrometerSpec` should alway pass because it disables the `ExecutorServiceMetricsBinder` making sure the io executor service is always instrumented with the propagated contexted

### Environment Information

- macos 14.3
- adoptjdk 23.0.2

### Example Application

https://github.com/RooHoo33/micronaut-thread-propagation

### Version

repo uses micronaut 4.7.6 but it affects versions 4.7.4 and above

hauke-d:
fixes #11653

The `InstrumentedExecutorService` interface is also used in micrometer.

The current instrumentation only checks if a given executor service is instrumented, but there is no notion of which specific instrumentation it is instrumented with.

This means that if the [micrometer instrumentation](https://github.com/micronaut-projects/micronaut-micrometer/blob/56217245d1cc33cf6894c925ebed71eaf990dc28/micrometer-core/src/main/java/io/micronaut/configuration/metrics/binder/executor/ExecutorServiceMetricsBinder.java#L91-L125) is applied first, the context propagation instrumentation is skipped.

To solve this for context propagation, we provide an implementation of the `InstrumentedExecutorService` and traverse the targets to find that specific implementation. It's not perfect, but should be better than before.

### DIFF
--- a/context-propagation/src/main/java/io/micronaut/context/propagation/instrument/execution/ContextPropagatingExecutorService.java
+++ b/context-propagation/src/main/java/io/micronaut/context/propagation/instrument/execution/ContextPropagatingExecutorService.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context.propagation.instrument.execution;
+
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.core.propagation.PropagatedContext;
+import io.micronaut.scheduling.instrument.InstrumentedExecutorService;
+
+import java.util.Optional;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+
+/**
+ * Wraps {@link ExecutorService} to instrument it for propagating the {@link PropagatedContext}
+ * across threads.
+ *
+ * @author Hauke Lange
+ * @since 4.9.0
+ */
+@Internal
+public class ContextPropagatingExecutorService implements InstrumentedExecutorService {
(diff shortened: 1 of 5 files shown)


hauke-d:
At least one of the workflow failures looks transient. Can you help me understand if the failing test is just unstable or if I need to have a closer look? To me it looks like a synchronization thing and unrelated as well.

hauke-d:
@graemerocher @dstepanov Could I get a review? We would love to see this fixed 😊

dstepanov (approved):

graemerocher:
Thanks for the contribution!

PR #11865, opened 2025-06-13
Check error element type. Fixes #11859

### LINKED ISSUE #11859: Postponed warning message fails with class cast exception
### Expected Behavior

```
Caused by: java.lang.ClassCastException: class io.micronaut.annotation.processing.visitor.JavaClassElement cannot be cast to class javax.lang.model.element.Element (io.micronaut.annotation.processing.visitor.JavaClassElement is in unnamed module of loader java.net.URLClassLoader @3ac99c68; javax.lang.model.element.Element is in module java.compiler of loader 'platform')
        at io.micronaut.annotation.processing.BeanDefinitionInjectProcessor.process(BeanDefinitionInjectProcessor.java:234)
        at org.gradle.api.internal.tasks.compile.processing.DelegatingProcessor.process(DelegatingProcessor.java:62)
        at org.gradle.api.internal.tasks.compile.processing.DynamicProcessor.process(DynamicProcessor.java:52)
        at org.gradle.api.internal.tasks.compile.processing.DelegatingProcessor.process(DelegatingProcessor.java:62)
        at org.gradle.api.internal.tasks.compile.processing.TimeTrackingProcessor.access$401(TimeTrackingProcessor.java:37)
```

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.8.x

graemerocher:
Adds type checking to fix #11859

Not sure how to reproduce yet.

### DIFF
--- a/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
+++ b/inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
@@ -15,6 +15,7 @@
  */
 package io.micronaut.annotation.processing;
 
+import io.micronaut.annotation.processing.visitor.AbstractJavaElement;
 import io.micronaut.annotation.processing.visitor.JavaClassElement;
 import io.micronaut.annotation.processing.visitor.JavaNativeElement;
 import io.micronaut.context.annotation.ClassImport;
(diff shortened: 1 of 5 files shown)


dstepanov (approved):

inject-java/src/main/java/io/micronaut/annotation/processing/BeanDefinitionInjectProcessor.java
```diff
@@ -277,6 +282,20 @@ public final boolean process(Set<? extends TypeElement> annotations, RoundEnviro
         return false;
     }
 
+    private static Element resolvedFailedElement(Object errorElement) {
```
yawkat: would it be possible to do this wherever the PostponeToNextRoundException is created? the other use of getErrorElement seems to be doing something similar

graemerocher: I improved the logic here

yawkat (approved):

PR #11874, opened 2025-06-17
Improve bean ordering semantics

### LINKED ISSUE #11871: Bean event listeners need to be ordered relative to all others
### Expected Behavior

`BeanCreationEventListener` objects are currently only ordered within the scope of the type for which they listen. This can lead to non-deterministic behavior when one listener listens for a supertype of another and which runs first ends up being random.

Example:

```
BeanCreationEventListener<Object> listener;

BeanCreationEventListener<DataSource> dataSourceReplacer;
```

The latter replaces data source beans with a compatible wrapper. Does the listener see the replacement or the original `DataSource`? Answer: undefined.

The `EventListenerListenersSupplier` (could this have a better name?) does do an ordering, but because there is one supplier per type, and the two listeners have different types, it doesn't help.

I think all bean creation listeners should be ordered relative to each other regardless of type.

### Actual Behaviour

Random ordering.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.9.0

mikehearn:
The default order of an object in OrderUtil.getOrder should be zero not LOWEST_PRECEDENCE, as otherwise there is no way to request a bean be last in the list. The intention here is to leave a bean that doesn't request an order un-ordered relative to other beans, so zero is the correct value.

Apply a global ordering to all bean event listeners regardless of requested generic type. Previously, listeners were ordered only relative to other listeners that specified the same generic type, meaning the final ordering could be unexpected in cases where supertypes were matched. Ordering is now done after all the listeners are aggregated.

Fixes #11871

### DIFF
--- a/core/src/main/java/io/micronaut/core/order/OrderUtil.java
+++ b/core/src/main/java/io/micronaut/core/order/OrderUtil.java
@@ -36,10 +36,30 @@ import java.util.stream.Stream;
 public class OrderUtil {
 
     /**
-     * Provide a comparator for collections.
+     * Orders objects using {@link #getOrder(Object)}, i.e. objects that don't implement
+     * {@link Ordered} and which are not {@link AnnotationMetadata} will be placed
+     * non-deterministically in the final position.
+     *
+     * <p>
+     * You probably want to use {@link #COMPARATOR_ZERO} instead, as with this comparator
+     * no object can sort itself to the back of the collection.
      */
     // Keep as an anonymous class to avoid lambda overhead during the startup
     public static final Comparator<Object> COMPARATOR = new Comparator<Object>() {
+        @Override
+        public int compare(Object o1, Object o2) {
+            int order1 = getOrder(o1);
+            int order2 = getOrder(o2);
+            return Integer.compare(order1, order2);
+        }
+    };
+
+    /**
+     * Orders objects using {@link #getOrderWithDefaultPrecedence(Object, int)} using zero as the
+     * default precedence.
+     */
+    // Keep as an anonymous class to avoid lambda overhead during the startup
+    public static final Comparator<Object> COMPARATOR_ZERO = new Comparator<Object>() {
         @Override
         public int compare(Object o1, Object o2) {
             int order1 = getOrder(o1);
@@ -66,11 +86,25 @@ public class OrderUtil {
 
     /**
      * Provide a comparator, in reversed order, for collections.
+     *
+     * <p>
+     * You probably want to use {@link #REVERSE_COMPARATOR_ZERO} instead, as with this
+     * comparator no object can sort itself to the front of the collection.
      */
     public static final Comparator<Object> REVERSE_COMPARATOR = Collections.reverseOrder(COMPARATOR);
 
     /**
-     * Sort the given list.
+     * Provide a comparator, in reversed order, for collections.
+     */
+    public static final Comparator<Object> REVERSE_COMPARATOR_ZERO = Collections.reverseOrder(COMPARATOR_ZERO);
+
+    /**
+     * Sort the given list using {@link #COMPARATOR}.
+     *
+     * <p>
+     * You should probably not use this method. Prefer calling
+     * {@code list.sort(OrderUtil.COMPARATOR_ZERO)} instead, which offers more intuitive behavior
+     * for beans that don't expose an ordering.
      *
      * @param list The list to sort
      */
@@ -80,6 +114,11 @@ public class OrderUtil {
 
     /**
      * Sort the given list.
+     *
+     * <p>
+     * You should probably not use this method. Prefer calling
+     * {@code stream.sorted(OrderUtil.COMPARATOR_ZERO)} instead, which offers more intuitive behavior
+     * for beans that don't expose an ordering.
      *
      * @param list The list to sort
      * @param <T>  The stream generic type
@@ -140,6 +179,11 @@ public class OrderUtil {
     /**
      * Sort the given list.
      *
+     * <p>
+     * You should probably not use this method. Prefer calling
+     * {@code list.sort(OrderUtil.REVERSE_COMPARATOR_ZERO)} instead, which offers more intuitive
+     * behavior for beans that don't expose an ordering.
+     *
      * @param list The list to sort
      */
     public static void reverseSort(List<?> list) {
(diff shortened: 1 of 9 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1263,27 +1255,31 @@ protected <T> void destroyLifeCycleBean(LifeCycle<?> cycle, BeanDefinition<T> de
...
-                        } catch (Exception e) {
-                            throw new BeanDestructionException(beanDefinition, e);
-                        }
+                    for (ListenersSupplier.ListenerAndOrder<BeanPreDestroyEventListener> listener : entry.getValue().get(null)) {
+                        listeners.add(listener);
                     }
-
+                }
+            }
+            OrderUtil.sort(listeners);
```
dstepanov: Why do you need to sort them? They should be sorted once

dstepanov (changes requested):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
mikehearn: How do you mean? Listeners can be ordered relative to one another so they must be sorted. It's not enough to sort them within the `Class<?>` key, because there might be two different class keys that both accept the same type and you need to order between those.

dstepanov: Listeners are sorted once, and I don't see a reason to resort them all the time.

mikehearn: The issue is that beans are both created lazily and selected per-type. Each event that's fired can select a different subset of listeners, which might in turn create them (so we can't pre-compute the sorts as the objects don't exist at initialization). In the common case there are zero or one listeners, so we can avoid the sort easily then. 

Do you see a good way to avoid the sort in every case?

core/src/main/java/io/micronaut/core/order/OrderUtil.java
```diff
@@ -174,18 +174,19 @@ public static void sort(Object[] objects) {
...
+     * of the {@link AnnotationMetadata}, or zero if the parameter is neither of those.
      */
     public static int getOrder(Object o) {
         if (o instanceof Ordered ordered) {
             return ordered.getOrder();
         } else if (o instanceof AnnotationMetadata metadata) {
             return getOrder(metadata);
         }
-        return Ordered.LOWEST_PRECEDENCE;
+        return 0;
```
graemerocher: this is a breaking change that has the potential to break existing applications

mikehearn: Possibly; to break an app bean would have to be trying to give itself a low precedence and failing due to this bug.

What we could do is use a separate comparator to fix the case of bean event listeners specifically, and leave the default ordering alone with a note in the javadoc that it's buggy. It can be marked as deprecated and then things re-unified in Micronaut 5.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
dstepanov: What it can create? A listener? By registering it manually?

mikehearn: The listener beans. I'm assuming there's a reason the existing code has this lazy supplier abstraction: the actual bean is created only when there's an event to send it. We could change that to create them all up front but I guess changing the timing of when beans are created is also backwards incompatible.

core/src/main/java/io/micronaut/core/order/OrderUtil.java
dstepanov: Maybe we can leave this method and only change the internal use-cases like listeners.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
dstepanov: Creating anything up front would hurt the startup. Not sure about your change, do you need this change for anything?

mikehearn: Yes, we have a listener that records information about beans that might be found in the heap. It must therefore run last, as otherwise the listener will record info about an intermediate bean that might be replaced by a later event listener.

Is this a real concern? A sort of a small array is extremely fast. I can add a branch to skip if .size() <= 1 but I'd be surprised if sorting event listener beans ever came up in profiles unless someone implements `Ordered` in an expensive way.

core/src/main/java/io/micronaut/core/order/OrderUtil.java
mikehearn: Yes, I think having a separate comparator and a note to clean it up in 5 is the way to go. I'll do that.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
dstepanov: Please create a test

mikehearn:
@graemerocher @dstepanov I've added a test, avoided changing the existing OrderUtil API and skip the sort if the list isn't big enough to need one. I also renamed the test file. PTAL.

mikehearn:
CI failure looks like a flake rather than a result of this change. I see quite a few flakes locally related to the HTTP/Netty tests.

mikehearn:
Everything seems to pass except the flake I already fixed in 4.9.x, so once it's merged up that test would have passed too. I think this is ready for merge.

graemerocher:
this needs to target 4.10.x

Change is too high risk to go into 4.9.x this late

mikehearn:
@graemerocher It does target 4.10.x now (I rebased and retargeted the PR already).

mikehearn:
I tweaked the test. Note that the reason this is so complex is that event listener dispatch works in an odd way - it has a `List<Map.Entry<Class<?>, List<Listener>>>` type for the listener registrations. First we iterate over this outer list and select only the inner lists (map.entry values) that for which the key is assignable from the bean. This looks like it might be an optimization to reduce the number of `Class.isAssignableFrom` calls so I didn't want to remove it, but there's no comment in the code stating that explicitly.

A simpler approach would be to just have a `List<Listener>` and sort it whenever a new listener is added. Then dispatch can be just a loop over all listeners that does an isAssignableFrom test on each element. But I'm not sure if that'd be faster or slower, overall.

dstepanov (approved):

PR #11893, opened 2025-06-23
Jackson 2.19.1

dstepanov:
Can we still use the latest Jackson? If not we need to remove the bump and leave the fix


dstepanov:
Considering that lang4j is using Jackson 2.19 and leaking it into the BOM

jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
```diff
@@ -422,14 +422,18 @@ public Object getValue(Object pojo) throws IllegalArgumentException {
...
                             final UnsafeBeanProperty<Object, Object> beanProperty = (UnsafeBeanProperty<Object, Object>) property.get();
-                            newProperties.set(i, new BeanIntrospectionPropertyWriter(
-                                    existing,
-                                    beanProperty,
-                                    existing.getSerializer(),
-                                    config.getTypeFactory(),
-                                    existing.getViews()
-                                )
-                            );
+                            if (existing instanceof AnyGetterWriter anyGetterWriter) {
```
yawkat: add to the if statement just above this one instead, which does the same check for XmlBeanPropertyWriter

yawkat (approved):

graemerocher:
tests fail on upgrade

PR #11917, opened 2025-07-02
throw exception if encoded cookie exceeds 4096 byte limit

sdelamo:
[Cookie byte limit is 4096](https://datatracker.ietf.org/doc/html/rfc6265#section-6.1). A user may create a cookie which exceeds the limit, for example a big JWT saved in a cookie,  and  we don't warn the user he is exceeding the limits.


core/src/main/java/io/micronaut/core/util/StringUtils.java
```diff
@@ -500,6 +502,10 @@ public static Iterator<String> splitOmitEmptyStringsIterator(final CharSequence
         return new SplitOmitEmptyIterator(sequence, splitCharacter);
     }
 
+    public static int byteCount(@NonNull String str) {
```
yawkat: You can copy netty `utf8ByteCount`.

Even better would be a specialized check that the size doesnt exceed the limit. e.g. if `str.length() * 3 < limit`, we don't need to check every byte, and if `str.length() * 2 > limit` we also don't need to

http-client/src/main/java/io/micronaut/http/client/netty/NettyStreamedHttpResponse.java
```diff
@@ -140,7 +142,10 @@ public boolean isStream() {
     @Override
     public synchronized MutableHttpResponse<B> cookie(Cookie cookie) {
         ServerCookieEncoder.INSTANCE.encode(cookie)
-                .forEach(cookieEncoded -> headers.add(HttpHeaderNames.SET_COOKIE, cookieEncoded));
+                .forEach(cookieEncoded -> {
```
yawkat: I think an exception would be better. Maybe with a configurable limit?

sdelamo: the limit of 4096 is enforced by browsers. Not sure what a configurable limit will do for users.

yawkat: it apparently varies by browser: http://browsercookielimits.iain.guru/

and ofc there's non-browser clients

core/src/main/java/io/micronaut/core/util/StringUtils.java
sdelamo: > You can copy netty utf8ByteCount.

Done via: https://github.com/micronaut-projects/micronaut-core/pull/11917/commits/1cd1f1bd3ad27ab4424a24cd901b84555ef4046a

sdelamo: > Even better would be a specialized check that the size doesnt exceed the limit. e.g. if str.length() * 3 < limit, we don't need to check every byte, and if str.length() * 2 > limit we also don't need to

Are you recommending this to avoid invoking only `utf8ByteCount` if we "think" we will exceed the limit?

I would prefer to avoid guesses. This will be only invoked when populating the Set-Cookie header and thus I think we can pay the performance toll of invoking `utf8ByteCount`

yawkat: You can avoid having to guess entirely for certain sizes, and that helps with worst-case performance.

http-client/src/main/java/io/micronaut/http/client/netty/NettyStreamedHttpResponse.java
sdelamo: What are you thinking about? Making it configurable in [HttpServerConfiguration](https://github.com/micronaut-projects/micronaut-core/blob/4.10.x/http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java) and [HttpClientConfiguration](https://github.com/micronaut-projects/micronaut-core/blob/4.10.x/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java)? That it is a bit hairy to do as HTTP responses classes are not beans, and to trickle down the configuration setting is a bit ugly. 

A simpler thing would be to add a default method to `MutableHttpResponse`  e.g. `MutableHttpResponse::cookie(Cookie cookie, Integer maxCookieSize)`, but I don't really like that API.

I personally think 4096 is a sensible default. I don't really see users wishing to change that. 

I have added [the size limit as a method parameter](https://github.com/micronaut-projects/micronaut-core/pull/11917/commits/848cd279067daab083ba3782ea2a5dabb33dcbbc) to make it more flexible.

sdelamo: > I think an exception would be better. Maybe with a configurable limit?

@yawkat I [have changed the implementation to thrown an exception instead of logging](https://github.com/micronaut-projects/micronaut-core/pull/11917/commits/6c233380eed2a248452186e45aee95ee927e4c96).

sdelamo:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=micronaut-projects_micronaut-core&pullRequest=11917) **Quality Gate failed**
> 
> Failed conditions  [51.3% Coverage on New Code](https://sonarcloud.io/component_measures?id=micronaut-projects_micronaut-core&pullRequest=11917&metric=new_coverage&view=list) (required ≥ 70%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=micronaut-projects_micronaut-core&pullRequest=11917)

sonar is complaining about test coverge. However, I have covered the exception handler via a test in `test-suite` and the othe methods coverage complains are from the code ported from Netty.

graemerocher (approved):

yawkat (approved):

PR #11919, opened 2025-07-02
4.7.x -> 4.8.x merge

yawkat:


core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationUtils.java
```diff
@@ -107,6 +107,9 @@ private static String getPath(AnnotationMetadata annotationMetadata) {
 
     @NonNull
     private static String computeIterablePrefix(AnnotationMetadata annotationMetadata, String prefix) {
+        if (prefix == null) {
+            prefix = StringUtils.EMPTY_STRING;
+        }
```
yawkat, line 112: is this supposed to be 4.7 only? looks a bit weird

PR #11937, opened 2025-07-09
Add string: and base64: resource loaders

yawkat:
Used for future PEM loading PR.

### DIFF
--- a/http/src/main/java/io/micronaut/http/resource/ResourceLoaderFactory.java
+++ b/http/src/main/java/io/micronaut/http/resource/ResourceLoaderFactory.java
@@ -15,17 +15,19 @@
  */
 package io.micronaut.http.resource;
 
-import io.micronaut.core.annotation.NonNull;
 import io.micronaut.context.annotation.BootstrapContextCompatible;
 import io.micronaut.context.annotation.Factory;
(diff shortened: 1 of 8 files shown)


graemerocher (approved):

http/src/main/java/io/micronaut/http/resource/ResourceLoaderFactory.java
```diff
@@ -71,6 +73,24 @@ public ResourceLoaderFactory(Environment environment) {
         return new DefaultFileSystemResourceLoader();
     }
 
+    /**
+     * @return The string resource loader
+     */
+    @Singleton
+    @BootstrapContextCompatible
+    protected @NonNull ResourceLoader getStringResourceLoader() {
```
dstepanov, line 81: I wonder if we can remove those as beans, and simply have some collection filled without the bean context

dstepanov:
Do we need those classes to be public?

yawkat:
Since there's a few places where all the resource loaders are explicitly listed, it seems reasonable to have all the default ones public. It's not a huge API commitment anyway.

dstepanov (approved):

PR #11939, opened 2025-07-10
Support loading keys and trusts from PEM files

yawkat:
This PR adds support for loading PEM files in place of key and trust stores.

- Introduce a PemParser class that can load various PEM types: PKCS#8 private keys (possibly encrypted), PKCS#1 private keys, SEC.1 elliptic curve keys, X.509 certificates. This is a bit complicated because we need to do some content sniffing.
- If key store loading fails from a path, also try loading it as PEM. For key stores, the expected format is a single PRIVATE KEY followed by some CERTIFICATEs, while for a trust store it expects a set of CERTIFICATEs.
- Add properties to KeyStoreConfiguration for separate key and certificate files, since keeping them separate is more common in practice (incl. in the docs)
- Update docs


gradle/libs.versions.toml
```diff
@@ -165,6 +165,7 @@ managed-netty-transport-native-epoll = { module = "io.netty:netty-transport-nati
 managed-netty-transport-native-kqueue = { module = "io.netty:netty-transport-native-kqueue", version.ref = "managed-netty" }
 managed-netty-transport-native-iouring = { module = "io.netty:netty-transport-native-io_uring", version.ref = "managed-netty" }
 managed-netty-transport-native-unix-common = { module = "io.netty:netty-transport-native-unix-common", version.ref = "managed-netty" }
+managed-netty-pkitesting = { module = "io.netty:netty-pkitesting", version.ref = "managed-netty" }
```
graemerocher, line 168: does this need to be `managed` if it is only for testing?

http/src/main/java/io/micronaut/http/ssl/SslBuilder.java
```diff
@@ -137,33 +152,269 @@ protected Optional<KeyStore> getKeyStore(SslConfiguration ssl) throws Exception
...
+                        throw f;
+                    }
+                } else {
+                    throw e;
+                }
+            }
             return store;
+        } else if (spec instanceof PemBasedCertificateSpec pem) {
+            List<Object> keyItems;
+            try (InputStream s = resourceResolver.getResourceAsStream(pem.keyPath).orElseThrow()) {
```
graemerocher: throw a better exception here with a message that includes the keyPath mentioned it is missing

http/src/main/java/io/micronaut/http/ssl/SslBuilder.java
```diff
@@ -137,33 +152,269 @@ protected Optional<KeyStore> getKeyStore(SslConfiguration ssl) throws Exception
...
+                keyItems = new PemParser(pem.provider, pem.password)
+                    .loadPem(new String(s.readAllBytes(), StandardCharsets.UTF_8));
+            }
+            List<Object> certItems;
+            try (InputStream s = resourceResolver.getResourceAsStream(pem.certificatePath).orElseThrow()) {
```
graemerocher: throw a better exception here with a message that includes the certificatePath mentioned it is missing

http/src/main/java/io/micronaut/http/ssl/SslBuilder.java
```diff
@@ -137,33 +152,269 @@ protected Optional<KeyStore> getKeyStore(SslConfiguration ssl) throws Exception
...
+            store.load(null, null);
+            store.setKeyEntry(
+                "key",
+                pk,
+                null,
+                certificates(certItems).toArray(new X509Certificate[0])
+            );
+            return store;
+        } else {
+            throw new AssertionError("Weird CertificateSpec");
```
graemerocher, line 280: throw `ConfigurationException` with a message saying it is an invalid or unsupported certificate type

http/src/main/java/io/micronaut/http/ssl/SslBuilder.java
```diff
@@ -137,33 +152,269 @@ protected Optional<KeyStore> getKeyStore(SslConfiguration ssl) throws Exception
...
+        }
+    }
+
+    private static @NonNull KeyStore createEmptyKeyStore(@Nullable String provider, String type) throws KeyStoreException, NoSuchProviderException {
+        return provider == null ? KeyStore.getInstance(type) : KeyStore.getInstance(type, provider);
+    }
+
+    private void loadPem(@NonNull String resource, @Nullable String password, @Nullable String provider, KeyStore store) throws IOException, GeneralSecurityException, PemParser.NotPemException {
+        List<Object> items;
+        try (InputStream s = resourceResolver.getResourceAsStream(resource).orElseThrow()) {
```
graemerocher: throw a better exception here that includes a path to the resource that failed to load

http/src/main/java/io/micronaut/http/ssl/SslConfiguration.java
```diff
@@ -438,6 +446,54 @@ public void setType(String type) {
...
+
+        /**
+         * A path to a PEM file containing the private key. Can use
+         * {@code classpath:}, {@code file:}, {@code string:} or {@code base64:}. Cannot be set at
+         * the same time as the {@code path} property.
+         *
+         * @param keyPath The key path
+         * @since 4.10.0
+         */
+        public void setKeyPath(String keyPath) {
```
graemerocher: add nullability annotations to the setters

http/src/main/java/io/micronaut/http/ssl/SslConfiguration.java
```diff
@@ -438,6 +446,54 @@ public void setType(String type) {
...
+
+        /**
+         * A path to a PEM file containing the private key of the server. Can use
+         * {@code classpath:}, {@code file:}, {@code string:} or {@code base64:}. Cannot be set at
+         * the same time as the {@code path} property.
+         *
+         * @return The key path
+         * @since 4.10.0
+         */
+        public String getKeyPath() {
```
graemerocher, line 460: If this is required specify `@NonNull` and validate before returning.

http/src/main/java/io/micronaut/http/ssl/SslConfiguration.java
```diff
@@ -438,6 +446,54 @@ public void setType(String type) {
...
+
+        /**
+         * A path to a PEM file containing the certificate chain. Can use
+         * {@code classpath:}, {@code file:}, {@code string:} or {@code base64:}. Cannot be set at
+         * the same time as the {@code path} property.
+         *
+         * @return The certificate chain path
+         * @since 4.10.0
+         */
+        public String getCertificatePath() {
```
graemerocher, line 485: if this is required specify `@NonNull` and validate before returning

http/src/main/java/io/micronaut/http/ssl/SslConfiguration.java
```diff
@@ -438,6 +446,54 @@ public void setType(String type) {
...
+
+        /**
+         * A path to a PEM file containing the certificate chain. Can use
+         * {@code classpath:}, {@code file:}, {@code string:} or {@code base64:}. Cannot be set at
+         * the same time as the {@code path} property.
+         *
+         * @param certificatePath The certificate chain path
+         * @since 4.10.0
+         */
+        public void setCertificatePath(String certificatePath) {
```
graemerocher: check argument is not `null`

http/src/main/java/io/micronaut/http/ssl/SslBuilder.java
```diff
@@ -137,33 +152,269 @@ protected Optional<KeyStore> getKeyStore(SslConfiguration ssl) throws Exception
...
+     * Certificate spec based on {@link SslConfiguration.KeyStoreConfiguration#getPath()} or
+     * {@link SslConfiguration.TrustStoreConfiguration#getPath()}. Note that the path can still
+     * point to a PEM.
+     *
+     * @since 4.10.0
+     */
+    protected static final class KeyStoreBasedCertificateSpec extends CertificateSpec {
+        final String path;
+
+        KeyStoreBasedCertificateSpec(String type, String password, String provider, String path) {
```
graemerocher: `private` like the other constructors

http/src/test/java/io/micronaut/http/ssl/PemParserTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+import java.math.BigInteger;
+import java.security.interfaces.ECPrivateKey;
+import java.security.interfaces.RSAPrivateKey;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+
+class PemParserTest {
```
graemerocher, line 13: could you add more tests for the error conditions and assert we are producing good errors for invalid config

graemerocher (changes requested):

http/src/main/java/io/micronaut/http/ssl/SslBuilder.java
yawkat: This code path can't happen because the class is sealed

gradle/libs.versions.toml
yawkat: not necessarily, but since all the other netty deps are managed and users may want to use this module too, it seems reasonable to me

http/src/main/java/io/micronaut/http/ssl/PemParser.java
```diff
@@ -0,0 +1,534 @@
...
+import java.util.List;
+
+/**
+ * Generic PEM file parser with type detection. <b>Note that this class does not defend against DER
+ * parser differential vulnerabilities, so it should not be used to verify untrusted PEMs.</b>
+ *
+ * @param provider The security provider to use for creating cryptographic data structures
+ * @param password The configured password for encrypted private keys
+ */
+record PemParser(
```
dstepanov, line 45: I think this can be usefull in core

yawkat: WDYM?

graemerocher (approved):

Issue #11950, opened 2025-07-17
CorsFilter should log on a higher level when it rejects a request


voronaam:
### Expected Behavior

When a request is blocked by the internals of the Micronaut Framework at least a single line is logged about it.

### Actual Behaviour

Nothing is logged, unless the log level is set to TRACE.
The request never even got to the programmers code and they have no idea why. This log line should be at least on the INFO level.

### Steps To Reproduce

1. Configure a normal HTTP application with `micronaut.server.cors.enabled` set to `true`
2. Do not configure `micronaut.server.cors.localhostPassThrough` (leave it to default `false`)
3. Attempt to make a call to the applicaiton running on a `localhost:8080` from inside of a Web Browser extension
4. (Alternatively) Just add `Origin: chrome-extension://bfohobmbnhjmppckaeodeopikmammoee` header in the test request you make via curl/Postman/etc
5. Observe: 403 Denied response. Nothing in the logs.
The request is rejected because of the header `Origin: chrome-extension://bfohobmbnhjmppckaeodeopikmammoee`. It is technically still localhost, but the CorsFilter does not know it.

There is a line logged on the TRACE level:

```
TRACE i.m.http.server.cors.CorsFilter - The request specifies an origin different than localhost. To prevent drive-by-localhost attacks the request is forbidden
```

### Version

4.7.6

yawkat:
We will not use a higher log level because it is a potential DoS vector.

PR #11951, opened 2025-07-17
Make CorsFilter log on Info level when denying localhost requests

voronaam:
This fixes the issue #11950 

I do not think many people run with trace logs enabled and it was frustrating to me when a valid request was denied with absolutely nothing in the logs. See the issue for more details.

Short description of the problem: requests from localhost coming from inside of a browser's extension are not detected as coming from localhost. I do not think it is worth the effort to upgrade the detection of localhost and it is better to just log a line on the info. There is already `micronaut.server.cors.localhostPassThrough` to resolve this problem for a developer, but the developer should at least know where to look for it.


yawkat:
Logging as > debug by default is dangerous due to denial of service. I'll have to see if this code is reachable in production environments.

sdelamo:
thanks for the contribution @voronaam but we have to keep logs by default in INFO to minimum.

PR #11984, opened 2025-08-11
Introduce ReadBuffer API

yawkat:
This PR adds a new ReadBuffer API that aims to replace the current ByteBuffer.

The main advantage is a clear ownership system based on Closeable instead of the optional ReferenceCounted. This simplifies user code, and gets rid of potentially slow `instanceof` checks. It's also necessary for netty 5 compatibility.

As the name suggests, a `ReadBuffer` is read-only. The ByteBuffer write methods are rarely used by micronaut projects, so they only add unnecessary complexity for the non-netty implementation. The few write use cases will be covered by the ByteBodyFactory buffer methods.

In addition to the new API, I also moved `AvailableByteBody` to use `ReadBuffers`. In the future, we will only need one implementation of AvailableByteBody based on ReadBuffer, instead of the array and netty based ones that exist right now. Finally, since some of the places where AvailableNettyByteBody was used now use AvailableByteBody, those additional interface type checks introduced a type pollution issue that I've resolved by making InternalByteBody an abstract class instead of an interface. **This is a separate commit for easier review**.

My plan is to use ReadBuffers to also merge some of the more complex ByteBody implementations, or make them more efficient. For example, ReactiveByteBufferByteBody could use ReadBuffers instead of nio ByteBuffers, and BaseSharedBuffer could be simplified using ReadBuffers instead of having to accommodate both netty and nio buffers.


buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBufferFactory.java
```diff
@@ -0,0 +1,224 @@
...
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Netty-based {@link ReadBufferFactory}. Also has additional utilities for dealing with netty
+ * buffers.
+ *
+ * @author Jonas Konrad
+ * @since 4.10.0
+ */
+public final class NettyReadBufferFactory extends ReadBufferFactory {
```
graemerocher, line 44: Consider `@Experimental` or `@Internal`

core/src/main/java/io/micronaut/core/io/buffer/ReadBufferFactory.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /**
+     * Wrapper around a {@link OutputStream} that buffers into a
+     * {@link ReadBuffer}. Must be closed after use, even if
+     * {@link #finishBuffer()} has not been called (e.g. on error), to avoid resource leaks.
+     */
+    public interface BufferingOutputStream extends Closeable {
+        /**
+         * Get the stream you can write to.
```
graemerocher: can you document semantics in terms of what happens if you call this multiple times, one time etc.

http-netty/src/main/java/io/micronaut/http/netty/body/AvailableNettyByteBody.java
```diff
@@ -43,8 +44,10 @@
  *
  * @since 4.5.0
  * @author Jonas Konrad
+ * @deprecated Use {@link NettyByteBodyFactory}
  */
 @Internal
+@Deprecated(forRemoval = true) // still used by micronaut-oracle-cloud
```
graemerocher: add `since` to the `@Deprecated` annotation

http-netty/src/main/java/io/micronaut/http/netty/body/AvailableNettyByteBody.java
```diff
@@ -56,16 +59,12 @@ public AvailableNettyByteBody(@NonNull ByteBuf buffer) {
     }
 
     public static CloseableAvailableByteBody empty() {
-        return new AvailableNettyByteBody(Unpooled.EMPTY_BUFFER);
+        return AvailableByteArrayBody.create(NettyReadBufferFactory.of(ByteBufAllocator.DEFAULT).createEmpty());
```
graemerocher: split this statement onto multiple lines to improve readability

http-netty/src/main/java/io/micronaut/http/netty/body/StreamingNettyByteBody.java
```diff
@@ -220,9 +221,14 @@ public static final class SharedBuffer extends BaseSharedBuffer<ByteBufConsumer,
...
         private ByteBuf addingBuffer;
 
-        public SharedBuffer(EventLoop loop, BodySizeLimits limits, Upstream rootUpstream) {
+        public SharedBuffer(EventLoop loop, NettyByteBodyFactory byteBodyFactory, BodySizeLimits limits, Upstream rootUpstream) {
             super(limits, rootUpstream);
             this.eventLoop = loop;
+            this.byteBodyFactory = byteBodyFactory;
+        }
+
+        public EventLoop eventLoop() {
```
graemerocher, line 231: is this public API? if so needs javadoc

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/io/buffer/ReadBufferFactory.java
yawkat: It's just an accessor. I'll add a comment.

http-netty/src/main/java/io/micronaut/http/netty/body/StreamingNettyByteBody.java
yawkat: StreamingNettyByteBody is Internal

buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBufferFactory.java
yawkat: I want to start using it immediately in micronaut-oracle-cloud

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
```diff
@@ -0,0 +1,169 @@
...
+     * buffer. This is <i>not</i> a <a href="#consuming">consuming operation</a>, even if
+     * {@code n == readable()}.
+     *
+     * @param n How many bytes to read
+     * @return A new, independent buffer that reads the first {@code n} bytes
+     * @throws IllegalStateException     If this buffer is already closed or consumed
+     * @throws IndexOutOfBoundsException If {@code n > readable()}
+     */
+    @NonNull
+    public abstract ReadBuffer readSome(int n) throws IndexOutOfBoundsException;
```
dstepanov: `some` sounds strange, maybe `readOnly` or just `read`. Please have a better name for the parameter.

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
```diff
@@ -0,0 +1,169 @@
...
+ * <h2 id="consuming">Consuming operations</h2>
+ *
+ * <p>Certain operations on ReadBuffers are <i>consuming</i>, meaning that they read all the
+ * remaining bytes in the buffer. Those operations are called out in the javadoc. After such an
+ * operation, the buffer does not need to be closed (though it's still allowed to do so).</p>
+ *
+ * @author Jonas Konrad
+ * @since 4.10.0
+ */
+public abstract class ReadBuffer implements AutoCloseable {
```
dstepanov, line 40: Why an abstract and not an interface?

http/src/main/java/io/micronaut/http/body/AvailableByteBody.java
```diff
@@ -112,7 +127,19 @@ default OptionalLong expectedLength() {
...
+        try (ReadBuffer rb = toReadBuffer()) {
+            return rb.toByteBuffer();
+        }
+    }
+
+    @Override
+    @NonNull
+    default InputStream toInputStream() {
+        try (ReadBuffer rb = toReadBuffer()) {
+            return rb.toInputStream();
```
dstepanov, line 140: It looks a bit strange that you can close the buffer and read it from IS

http/src/main/java/io/micronaut/http/body/AvailableByteBody.java
```diff
@@ -112,7 +127,19 @@ default OptionalLong expectedLength() {
      * @return The bytes
      */
     @NonNull
-    ByteBuffer<?> toByteBuffer();
+    default ByteBuffer<?> toByteBuffer() {
+        try (ReadBuffer rb = toReadBuffer()) {
+            return rb.toByteBuffer();
```
dstepanov, line 132: Here I would expect that `toByteBuffer` will close the buffer

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
```diff
@@ -0,0 +1,169 @@
...
+     * Create a new buffer with the same content as this one, consuming this buffer in the process.
+     * If this buffer is later closed, the returned buffer remains usable.
+     *
+     * <p>This is a <a href="#consuming">consuming operation</a>.
+     *
+     * @return a new ReadBuffer containing all the bytes that were previously in this ReadBuffer
+     * @throws IllegalStateException If this buffer is already closed or consumed
+     */
+    @NonNull
+    public abstract ReadBuffer move();
```
dstepanov, line 89: It's not clear to me the purpose of this method. Consume and create a new buffer, why? Does it read all the data?

dstepanov (changes requested):

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
yawkat: it's slightly faster than interface calls

http/src/main/java/io/micronaut/http/body/AvailableByteBody.java
yawkat: toInputStream doesn't really read from the same buffer, it transfers ownership of the buffer to the stream so now the stream must be closed

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
yawkat: It makes edge cases with exception handling easier, you can avoid having to do something like this:

```java
ReadBuffer rb = ...;
boolean saved = false;
try {
  ...
  this.buffer = rb;
  saved = true;
  ...
} finally {
  if (!saved) rb.close();
}
```

And instead do this:

```java
try (ReadBuffer rb = ...) {
  ...
  this.buffer = rb.move();
  ...
}
```

yawkat: we have this pattern in ByteBody already fwiw

http/src/main/java/io/micronaut/http/body/AvailableByteBody.java
yawkat: It does but closing twice is not an issue so it's a good idea to always use try-with-resources. Also makes IDE warnings shut up

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
dstepanov: Why not simply `this.buffer = rb`?

yawkat: not sure what you mean

dstepanov: I don't understand why something needs to be moved

yawkat: Because `rb` is closed

core/src/main/java/io/micronaut/core/io/buffer/ReadBuffer.java
yawkat: I renamed it to `split`

graemerocher (approved):

dstepanov (approved):

PR #11985, opened 2025-08-11
Add ResponseBodyWriter for ByteBody

yawkat:
Lets controllers return ByteBody directly.


http/src/main/java/io/micronaut/http/ServerHttpRequest.java
```diff
@@ -37,4 +39,15 @@ public interface ServerHttpRequest<B> extends HttpRequest<B> {
...
     ByteBody byteBody();
+
+    /**
+     * Get a {@link ByteBodyFactory} that can be used to construct an efficient response to this
+     * request.
+     *
+     * @return The body factory
+     */
+    @NonNull
+    default ByteBodyFactory byteBodyFactory() {
```
dstepanov, line 50: Wouldn't be better to have the factory access added to `ByteBody`?

yawkat: It doesn't always make sense. For example an AvailableByteBody doesn't care where it's created. On the other hand, a streaming ByteBody can be bound to an EventLoop, so it makes sense to construct it on the same event loop as the request for performance. To make that work, I'd have to attach the ByteBodyFactory to the AvailableByteBody without a real reason.

graemerocher (approved):

dstepanov (approved):

PR #11987, opened 2025-08-12
Add explicit event loop transport configuration

yawkat:
prefer-native-transport only sets a preference. If the native transport fails to load for any reason (e.g. wrong architecture), the event loop silently falls back to nio.

This PR adds a new property to set an explicit preference list of transports to use. If the user wants to be sure they aren't using nio by accident, they can simply remove the nio entry from the list and there will be an exception.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/channel/EventLoopGroupConfiguration.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/channel/EventLoopGroupConfiguration.java
@@ -16,9 +16,12 @@
 package io.micronaut.http.netty.channel;
 
 import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.NextMajorVersion;
+import io.micronaut.core.annotation.NonNull;
 import io.micronaut.core.naming.Named;
 
 import java.time.Duration;
+import java.util.List;
 import java.util.Optional;
 
 /**
@@ -87,7 +90,25 @@ public interface EventLoopGroupConfiguration extends Named {
     /**
      * @return Whether to prefer the native transport
      */
+    @Deprecated(since = "4.10.0", forRemoval = true)
     boolean isPreferNativeTransport();
+
+    /**
+     * The transports to use for this event loop, in order of preference. Supported values are
+     * {@code io_uring,epoll,kqueue,nio}. The first available transport out of those listed will
+     * be used (nio is always available). If no listed transport is available, an exception will be
+     * thrown.
+     * <p>By default, only {@code nio} is used, even if native transports are available. If the
+     * legacy {@link #isPreferNativeTransport() prefer-native-transport} property is set to
+     * {@code true}, this defaults to {@code io_uring,epoll,kqueue,nio}.
+     *
+     * @return The available transports, in order of preference
+     */
+    @NonNull
+    @NextMajorVersion("Change default to all transports")
+    default List<String> getTransport() {
+        return isPreferNativeTransport() ? DefaultEventLoopGroupFactory.FACTORY_PRIORITY : List.of(NioEventLoopGroupFactory.NAME);
+    }
 
     /**
      * @return The shutdown quiet period
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
@@ -1302,6 +1302,7 @@ public class NettyHttpServerConfiguration extends HttpServerConfiguration {
         private Integer ioRatio;
         private String executor;
         private boolean preferNativeTransport = false;
+        private List<String> transport;
         private Duration shutdownQuietPeriod = Duration.ofSeconds(DEFAULT_SHUTDOWN_QUIET_PERIOD);
         private Duration shutdownTimeout = Duration.ofSeconds(DEFAULT_SHUTDOWN_TIMEOUT);
         private String name;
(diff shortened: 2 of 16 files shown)


dstepanov:
The downside that you need to know the name of the transport. Can we have something like `native` and that will resolve to `io_uring,epoll,kqueue`?

zdegner:
Just a comment -- I found the existing documentation a bit confusing for enabling native transports. Whenever I enabled them I'd see this error:

```
io.micronaut.http.server.exceptions.ServerStartupException: Unable to start Micronaut server on *:-1
        at app//io.micronaut.http.server.netty.NettyHttpServer.bind(NettyHttpServer.java:672)
        at app//io.micronaut.http.server.netty.NettyHttpServer.start(NettyHttpServer.java:314)
        at app//io.micronaut.http.server.netty.NettyHttpServer.start(NettyHttpServer.java:124)
        at app//io.micronaut.test.extensions.AbstractMicronautExtension.beforeClass(AbstractMicronautExtension.java:350)
        at app//io.micronaut.test.extensions.junit5.MicronautJunit5Extension.beforeAll(MicronautJunit5Extension.java:84)
        at java.base@21.0.8/java.util.ArrayList.forEach(ArrayList.java:1596)

        Caused by:
        java.lang.IllegalStateException: incompatible event loop type: io.netty.channel.SingleThreadIoEventLoop
            at io.netty.channel.AbstractChannel$AbstractUnsafe.register(AbstractChannel.java:332)
            at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:119)
            at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:113)
            at io.netty.channel.MultithreadEventLoopGroup.register(MultithreadEventLoopGroup.java:86)
            at io.netty.bootstrap.AbstractBootstrap.initAndRegister(AbstractBootstrap.java:339)
            at io.netty.bootstrap.AbstractBootstrap.doBind(AbstractBootstrap.java:288)
            at io.netty.bootstrap.AbstractBootstrap.bind(AbstractBootstrap.java:284)
            at io.netty.bootstrap.AbstractBootstrap.bind(AbstractBootstrap.java:262)
            at io.micronaut.http.server.netty.NettyHttpServer.bind(NettyHttpServer.java:596)
            ... 5 more
```

It took quite a bit of debugging (into a Linux container, since it would not appear when running on Windows where the native transports are not available) in order to realize that the "parent" event loop needs to be set as well as the default:

```yaml
micronaut:
  netty:
    event-loops:
      parent:
        prefer-native-transport: true
      default:
        prefer-native-transport: true
```

I didn't really see anything regarding this "parent" loop in the documentation.

yawkat:
I want to move to a single event loop model by default for 5.0, so hopefully it'll be easier then

yawkat:
@dstepanov The problem with a `native` alias is that you could have epoll and io_uring available at the same time, so it's ambiguous... imo since users have to depend on the transport anyway, it's not a big deal to make them also list those transports in the config by name

http-netty/src/main/java/io/micronaut/http/netty/channel/EventLoopGroupConfiguration.java
```diff
@@ -87,8 +90,26 @@ default double getThreadCoreRatio() {
...
+     * thrown.
+     * <p>By default, only {@code nio} is used, even if native transports are available. If the
+     * legacy {@link #isPreferNativeTransport() prefer-native-transport} property is set to
+     * {@code true}, this defaults to {@code io_uring,epoll,kqueue,nio}.
+     *
+     * @return The available transports, in order of preference
+     */
+    @NonNull
+    @NextMajorVersion("Change default to all transports")
+    default List<String> getTransport() {
```
graemerocher, line 109: can this be an `EnumSet` or `List<TransportType>`?

http-server-netty/src/main/java/io/micronaut/http/server/netty/configuration/NettyHttpServerConfiguration.java
```diff
@@ -1438,6 +1439,26 @@ public boolean isPreferNativeTransport() {
...
+         * {@code io_uring,epoll,kqueue,nio}. The first available transport out of those listed will
+         * be used (nio is always available). If no listed transport is available, an exception will be
+         * thrown.
+         * <p>By default, only {@code nio} is used, even if native transports are available. If the
+         * legacy {@link #isPreferNativeTransport() prefer-native-transport} property is set to
+         * {@code true}, this defaults to {@code io_uring,epoll,kqueue,nio}.
+         *
+         * @param transport The available transports, in order of preference
+         */
+        public void setTransport(@NonNull List<String> transport) {
```
graemerocher, line 1458: handle `null` and invalid values in the setter

graemerocher (changes requested):

http-netty/src/main/java/io/micronaut/http/netty/channel/EventLoopGroupConfiguration.java
yawkat: It has to be ordered. Also I don't want to use an enum because users can still define their own transports

graemerocher: how would a user define their own transport? Can you document this somewhere?

yawkat: In principle you can define your own EventLoopGroupFactory at the moment, nothing's stopping you and the APIs are public. I don't want to break that even if I can't really see the use case. Maybe some people implement it with local transport.

graemerocher (approved):

PR #11991, opened 2025-08-13
Support JSpecify

### LINKED ISSUE #11855: Support JSpecify annotations
### Feature description

Investigate the integration with https://jspecify.dev annotations

dstepanov:
Closes https://github.com/micronaut-projects/micronaut-core/issues/11855

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/NullMarked.java
+++ b/core/src/main/java/io/micronaut/core/annotation/NullMarked.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.MODULE;
+import static java.lang.annotation.ElementType.PACKAGE;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Marks the annotated declaration and all code transitively enclosed by it as <b>null-marked</b>:
+ * within that scope, type usages are non-null by default unless explicitly annotated as
+ * {@code @Nullable}.
+ *
+ * <p>This annotation is intentionally similar in purpose and semantics to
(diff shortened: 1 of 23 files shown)


sdelamo (changes requested):
Can you please add documentation? 

We have to document it in the [Nullability annotations](https://docs.micronaut.io/4.9.9/guide/#nullabilityAnnotations) section and we should tell users what annotation to use in their projects. Should users still be using the Micronaut Nullability annotations or should they be using the JSpecify annotations? 

Moreover, we should document the `@NullMark` annotation. Should users use the Micronaut's `@NullMark` or the JSpecify `@NullMark` Is there any drawback in using the JSpecify one? 

Will the JSpecify `@Nulllable` annotation also work for [Nullable injection](https://docs.micronaut.io/4.9.9/guide/#nullableInjection) ?

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaPropertyElement.java
```diff
@@ -76,6 +76,12 @@ final class JavaPropertyElement extends AbstractJavaElement implements PropertyE
         this.annotationMetadata = new PropertyElementAnnotationMetadata(this, getter, setter, field, null, false);
     }
 
+    @Override
+    protected AnnotationMetadata getTypeAnnotationMetadata() {
+        // TODO: aggregate type annotations from all the accessors
```
sdelamo: did you forget to resolve this `TODO`?

dstepanov:
Rewrote the section a bit, yes the annotations work in the same way.

sdelamo (commented):
are we affected by https://jspecify.dev/docs/whether/#annotation-processors ?

inject-java/src/main/java/io/micronaut/annotation/processing/visitor/JavaPropertyElement.java
dstepanov: > did you forget to resolve this `TODO`?

Rewote a bit, mentioning that we might need to add new checks to properly check for the nullability of reader / writer, that's nothing new

dstepanov:
> are we affected by https://jspecify.dev/docs/whether/#annotation-processors ?

Yes, annotation will be missing, but those are very specific scenarious

sdelamo (approved):
aside from https://github.com/micronaut-projects/micronaut-core/pull/11991#discussion_r2276383377 it looks good to me.

core/src/main/java/io/micronaut/core/annotation/NullMarked.java
```diff
@@ -0,0 +1,63 @@
...
+ *       annotations to express intent clearly.</li>
+ * </ul>
+ *
+ * @author Denis Stepanov
+ * @see 4.10
+ */
+@Documented
+@Target({MODULE, PACKAGE, TYPE, METHOD, CONSTRUCTOR})
+@Retention(RUNTIME)
+public @interface NullMarked {
```
graemerocher, line 62: do we need this? Why can't we just rely on the jspecify version?

graemerocher (commented):
The Spring folks deprecated their nullable/non-null annotations and went with JSpecify by default everywhere. Should we do the same?

dstepanov:
I wouldn't deprecate anything in v4 otherwise that will add to many deprecated warnings (Maybe we can do it in the last minor release before v5). Internally we don't really care which annotation is used, now we remap all of them to the `jakarta.annotation.N*` we might want to remap it to `io.micronaut.core.annotation.Nullable/NonNull` in the next version.

Not sure about `NullMarked`, I can delete it and remap it to a dummy `io.micronaut.core.annotation.NullMarked`. This means we will not not support the same behaviour as JSpecify without it. I don't expect many people will want to use it anyway.

graemerocher (approved):

PR #12001, opened 2025-08-19
Support `@Introspected` on packages

dstepanov:
- New visitor `PackageElementVisitor`
- Converted `@Configuration` use case to use the listener


core-processor/src/main/java/io/micronaut/inject/visitor/PackageElementVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param context The visitor context
+     */
+    default void visitPackage(@NonNull PackageElement element, @NonNull VisitorContext context) throws ProcessingException {
+    }
+
+    /**
+     * @return The supported default annotation names.
+     */
+    @NonNull
+    default Set<String> getSupportedPackageAnnotationNames() {
```
graemerocher: would rename to just `getSupportedAnnotationNames()` it is implicit from the name of the type this relates to packages. Also add nullability annotation to return type

core-processor/src/main/java/io/micronaut/inject/visitor/PackageElementVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+        Class<?>[] classes = GenericTypeUtils.resolveInterfaceTypeArguments(getClass(), PackageElementVisitor.class);
```
graemerocher: you can probably use:

```java
Optional<Class<?>> resolveInterfaceTypeArgument(..)
```
to make this simpler

src/main/docs/guide/ioc/introspection/atIntrospected.adoc
```diff
@@ -11,3 +11,7 @@ snippet::io.micronaut.docs.ioc.beans.IntrospectionSpec[tags="usage", indent=0]
 <3> A api:core.beans.BeanProperty[] can be retrieved from the introspection
 <4> Use the `set` method to set the property value
 <5> Use the `get` method to retrieve the property value
+
+It's possible to apply the introspection to all the classes in one package. Simply create `package-info` file and annotate the package with ann:core.annotation.Introspected[].
+
+snippet::io.micronaut.docs.ioc.introspection.pck.foobar.package-info[tags="usage", indent=0]
```
graemerocher, line 17: add a `WARNING:` here that the package should only contain classes that can be introspected. Also clarify here whether this behaviour is recursive for sub packages or a new package-info is required for each.

graemerocher (changes requested):

graemerocher (approved):

PR #12004, opened 2025-08-20
Add method name and owning class name to error messages

sosguthorpe:
Add details of method causing errors during AOP introduction.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
+++ b/core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
@@ -426,15 +426,15 @@ class DeclaredBeanElementCreator extends AbstractBeanElementCreator {
             InterceptedMethodUtil.hasDeclaredAroundAdvice(methodAnnotationMetadata) && !classElement.isAbstract()) {
             if (methodElement.isFinal()) {
                 if (InterceptedMethodUtil.hasDeclaredAroundAdvice(methodAnnotationMetadata)) {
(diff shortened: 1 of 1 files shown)


graemerocher:
needs a test

dstepanov:
This needs a different solution, usually javac will pinpoint to the class that caused the issue, but of the class is not part of the compiled classes that is hard to understand where is the problem

sosguthorpe:
OK. But it didn't print the class and we were left with an error message that mentioned a private unnamed method, and this was javac. I'm not sure what you mean by part of the compiled classes. It was a  bean-class in the codebase, that  extended another, in a different package, where the parent had an annotated package private method. I don't understand why that wouldn't be part of the classes to compile? Given the processor was visiting the class in the first place it has to be no?

Incidentally the Eclipse JDT compiler worked fine. Presumably because visitor processed the method in the same package as the declaring class and not the owning class, although I am just guessing.

Happy for this pr to be closed as won't do. But the fact still remains I had to pull down core, alter it and build it to get simple sign-pointing to which class was causing the breakage.

dstepanov:
Can you pls create a sample project with a problem?

sosguthorpe:
Yes, of course, no problem:
https://github.com/sosguthorpe/javac-error-replicator

```
./gradlew clean assemble

> Task :compileJava FAILED
error: Method annotated as executable but is declared private. To invoke the method using reflection annotate it with @ReflectiveAccess
1 error
```

sosguthorpe:
I'd be happy to help with a different solution. Perhaps if there is no "location" then the ProcessingException could be printed differently? It gets the Element that caused the error so that should be possible?

PR #12010, opened 2025-08-26
Support for reading env from .env files

### LINKED ISSUE #11994: Support for reading env from .env files
### Feature description

It would be nice if we could use https://github.com/cdimascio/dotenv-java to read env variables

iamsb97:
Closes #11994.


altro3:
This is not full support for dot env format. Need to imrove this integration. Also need to add support of multiple environments

Links:

https://hexdocs.pm/dotenvy/dotenv-file-format.html
https://www.dotenv.org/docs/
https://dotenvx.com/docs/env-file
https://github.com/motdotla/dotenv
https://github.com/radprogrammer/dotenv-RFC

sdelamo (changes requested):
Thanks for the contribution. Can you add documentation?

sdelamo (commented):
Could you clarify why .env files would be preferred instead of using .properties files?

iamsb97:
This PR needs improvements as suggested by altro3. I will add documentation along with those changes in the next update.

.env files are preferred for sensitive information like API keys and database passwords that shouldn't be committed to the git repo, unlike .properties files which are typically version controlled.

PR #12024, opened 2025-09-02
Use ReadBuffer to reduce code duplication

yawkat:
Lots of ByteBody code is duplicated for netty and non-netty. The common ReadBuffer API makes a refactor to merge much of that code possible.

- The reactive implementation of AbstractBodyAdapter is merged into the base class, which becomes non-abstract. The netty implementation retains some small thread safety additions.
- BaseSharedBuffer is refactored to work with ReadBuffers directly. Much of the subclass code is merged into the base class. Subclasses are retained due to different approaches to thread safety.
- BaseSharedBuffer.AsFlux subclasses are removed.
- A common class BaseStreamingByteBody is introduced that is based on BaseSharedBuffer. Some of the code from the subclasses is moved there.
- BufferConsumer gets an add(ReadBuffer) method. Subinterfaces with add(ByteBuf) or add(ByteBuffer) become obsolete and are removed.
- ByteBodyFactory gets adapt methods for converting publishers to ByteBodys, previously done through NettyBodyAdapter (mostly emptied of code) and ByteBufferBodyAdapter (deprecated).
- ConcatenatingSubscriber deals with ReadBuffers directly. ByteBuffer subclasses are removed. Netty subclass remains for concurrency control. Json-specific subclasses are replaced by a Separators record.
- PublisherAsBlocking is ungenerified and only deals with ReadBuffer now. Subclasses with specialized resource management are removed.
- PublisherAsStream is moved to ReadBuffers and is moved to micronaut-http.
- micronaut-http-client-jdk ByteBodySubscriber is mostly cleared and uses ByteBodyFactory.adapt to transform the reactive stream to a ByteBody.
- NettyByteBody is removed because it didn't do very much anymore, so that StreamingNettyByteBody can extend BaseStreamingByteBody instead.


graemerocher:
doesn't compile?

yawkat:
maybe the merge broke it

graemerocher (approved):

dstepanov (approved):

PR #12029, opened 2025-09-03
4.8.x -> 4.9.x

yawkat:


graemerocher:
should this wait for Netty 4.2.6 which is now out?

PR #12042, opened 2025-09-09
Fix compilation error for interface conf properties

dstepanov:


graemerocher (approved):

yawkat:
I don't think this fix is correct. When we add properties to configuration interfaces, we need them to be default methods. This patch makes it so those methods won't be injected at all

graemerocher:
true it would be good to return the default value and allow overriding if not specified

core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
```diff
@@ -189,6 +189,9 @@ public void visitClass(ClassElement classElement, VisitorContext context) {
             if (!methodElement.isStatic() && isInjectPointMethod(methodElement)) {
                 applyConfigurationInjectionIfNecessary(classElement, methodElement, context);
             } else if (anInterface && !processed.contains(methodElement)) {
+                if (methodElement.isDefault()) {
+                    return;
```
yawkat, line 194: can we make it a warning?

yawkat (approved):

PR #12045, opened 2025-09-10
Improvements to the bean context in v5

dstepanov:
- `DefaultBeanContext` and `DefaultApplicationContext` sealed / final
- `DefaultApplicationContext` is package protected
- `DefaultBeanContext` will be packaged protected in v6 to prevent forcing recompilation of definitions
-  `DefaultEnvironment` made final and package protected, instantiation replaced by `Environment.create(...)`
- Extracted deduction of environments and packages logic from `DefaultEnvironment`
- Bootstrap implemented as a separate property sources provider
- New configuration options added to achieve bootstrap like context (disable eager beans, events). Serde should use it to create static bean context instead of doing overrides
- Startup optimalizations


inject/src/main/java/io/micronaut/context/env/ConstantPropertySourceLoader.java
```diff
@@ -0,0 +1,63 @@
...
+import io.micronaut.core.optim.StaticOptimizations;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class ConstantPropertySourceLoader implements PropertySourceLoader {
```
graemerocher: please document this class. Can it be final?

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -682,421 +528,27 @@ private Optional<Map<String, Object>> readPropertiesFromLoader(String fileName,
...
+            propertyPlaceholderResolver.close();
         } catch (Exception e) {
             throw new RuntimeException("Failed to close!", e);
         }
+        for (PropertySourcesLocator propertySourcesLocator : configuration.getPropertySourcesLocators()) {
+            if (propertySourcesLocator instanceof Closeable closeable) {
+                try {
+                    closeable.close();
+                } catch (Exception e) {
+                    throw new RuntimeException("Failed to close!", e);
```
graemerocher: throw a more specific exception here that doesn't extend from RuntimeException and include the error message

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -682,421 +528,27 @@ private Optional<Map<String, Object>> readPropertiesFromLoader(String fileName,
...
-        return "digitalocean".equalsIgnoreCase(readFile(DO_SYS_VENDOR_FILE));
-    }
-
     @Override
     public void close() {
         try {
-            super.close();
+            propertyPlaceholderResolver.close();
         } catch (Exception e) {
             throw new RuntimeException("Failed to close!", e);
```
graemerocher: same here

inject/src/main/java/io/micronaut/context/env/DefaultEnvironmentAndPackageDeducer.java
```diff
@@ -0,0 +1,457 @@
...
+import static io.micronaut.context.env.Environment.DIGITAL_OCEAN;
+import static io.micronaut.context.env.Environment.ENVIRONMENTS_ENV;
+import static io.micronaut.context.env.Environment.ENVIRONMENTS_PROPERTY;
+import static io.micronaut.context.env.Environment.GOOGLE_COMPUTE;
+import static io.micronaut.context.env.Environment.IBM;
+import static io.micronaut.context.env.Environment.ORACLE_CLOUD;
+import static io.micronaut.context.env.Environment.TEST;
+
+@Internal
+final class DefaultEnvironmentAndPackageDeducer implements EnvironmentNamesDeducer, EnvironmentPackagesDeducer {
```
graemerocher, line 71: include a description of the role of this class and what it does. Even though it is internal we need documentation trail.

inject/src/main/java/io/micronaut/context/env/DefaultEnvironmentAndPackageDeducer.java
```diff
@@ -0,0 +1,457 @@
...
+
+        // deduce heroku
+        if (StringUtils.isNotEmpty(CachedEnvironment.getenv(HEROKU_DYNO))) {
+            environments.add(Environment.HEROKU);
+            environments.add(Environment.CLOUD);
+            deduceComputePlatform = false;
+        }
+
+        // deduce GAE
+        if (StringUtils.isNotEmpty(CachedEnvironment.getenv(GOOGLE_APPENGINE_ENVIRONMENT))) {
```
graemerocher, line 227: can we split this cloud specific logic into the modules somehow?

inject/src/main/java/io/micronaut/context/env/DefaultEnvironmentAndPackageDeducer.java
```diff
@@ -0,0 +1,457 @@
...
+                int nextIndex = i + 1;
+                if (nextIndex < len) {
+                    StackTraceElement next = stackTrace[nextIndex];
+                    if (!next.getClassName().startsWith("io.micronaut")) {
+                        packageName = NameUtils.getPackageName(next.getClassName());
+                    }
+                }
+            }
+
+            if (stackTraceElement.getMethodName().contains("$spock_")) {
```
graemerocher, line 303: again if we are undertaking this refactoring can we split this logic into the spock/junit modules?

inject/src/main/java/io/micronaut/context/env/Environment.java
```diff
@@ -213,6 +215,18 @@ public interface Environment extends PropertyResolver, LifeCycle<Environment>, M
...
 
+    /**
+     * Creates and returns a new instance of {@link Environment} using the provided
+     * application context configuration.
+     *
+     * @param configuration the application context configuration, must not be null
+     * @return a new instance of {@link Environment}, never null
+     */
+    @NonNull
+    static Environment create(@NonNull ApplicationContextConfiguration configuration) {
```
graemerocher, line 227: add since to all new public API

inject/src/main/java/io/micronaut/context/env/EnvironmentNamesDeducer.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context.env;
+
+import java.util.Set;
+
+public interface EnvironmentNamesDeducer {
```
graemerocher, line 35: needs javadoc and nullability annotations.

inject/src/main/java/io/micronaut/context/env/EnvironmentPackagesDeducer.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context.env;
+
+import io.micronaut.core.annotation.NonNull;
+
+import java.util.List;
+
+public interface EnvironmentPackagesDeducer {
```
graemerocher, line 33: needs javadoc

inject/src/main/java/io/micronaut/context/env/PropertySource.java
```diff
@@ -142,44 +141,6 @@ public Origin getOrigin() {
...
-    }
-
-    /**
-     * Create a {@link LinkedHashMap} of configuration from an array of values.
-     *
-     * @param values The values
-     * @return The created map
-     * @since 2.0
-     */
-    static Map<String, Object> mapOf(Object... values) {
```
graemerocher, line 166: this is not deprecated but what removed?

inject/src/main/java/io/micronaut/context/ApplicationContextConfigurationDelegate.java
```diff
@@ -0,0 +1,187 @@
...
+ * A delegate implementation of {@link ApplicationContextConfiguration} that forwards
+ * all operations to another {@link ApplicationContextConfiguration} instance.
+ * This class effectively acts as a wrapper or decorator, allowing interception or extension
+ * of the behavior of the underlying configuration.
+ *
+ * @author Denis Stepanov
+ * @since 5.0
+ */
+@Internal
+public class ApplicationContextConfigurationDelegate implements ApplicationContextConfiguration {
```
graemerocher: does this need to be public?

graemerocher (changes requested):
Could you update the breaks.adoc breaking changes section with examples of how to migrate code that used the old approach of constructing the context to the new approach.

inject/src/main/java/io/micronaut/context/env/DefaultEnvironmentAndPackageDeducer.java
dstepanov: I can make both deducers as a list which can be loaded from the `META-INF/services`?

inject/src/main/java/io/micronaut/context/env/PropertySource.java
dstepanov: It can be replaced with `Map.of` but I can keep the original method and deprecate it

graemerocher (approved):

PR #12047, opened 2025-09-11
feat(http-client-netty): configurable response decompression

### LINKED ISSUE #11995: Support for disabling decompression for Netty client
### Feature description

In my use-case I am trying to make a proxy server via micronaut where I will make a request to another server via micronaut app.

The goal is to test if the other server is sending gzip message correctly, but since netty decompresses and removes the content-encoding header, I am not able to check this in my test case.

For now I am using the micronaut-http-client-jdk variant to mitigate this issue

yawkat:
- Add HttpClientConfiguration.decompressionEnabled (default: true)
- Netty client: only add HttpContentDecompressor for HTTP/1 and HTTP/2 when enabled
- TCK: add DecompressionConfigTest covering default behavior and disabled mode
- JDK TCK: exclude DecompressionConfigTest from JdkHttpMethodTests

Use-case: allow proxies/tests to observe raw compressed payloads and preserve Content-Encoding.

Fixes #11995


yawkat:
(This PR is AI-generated)

graemerocher (approved):

PR #12057, opened 2025-09-19
Certificate provider API

yawkat:
Currently, certificates are configured as part of the `SslConfiguration`, separately for the client and server. Loading is done by SslContextBuilder and its client and server implementations. Downstream code such as micronaut-acme and micronaut-oracle-cloud has to extend these classes in order to implement custom certificate loading, but this has several disadvantages:

- It is difficult to use multiple certificate loading strategies in the same application
- Certificate loading conflicts with other customizations of the SSL context
- Certificate loading code has to be implemented twice, once for the server and once for the client
- It is netty-specific
- Certificate reloading has to happen through refresh events or complicated delegations

This PR decouples certificate loading from SSL context instantiation and customization. 

I introduce a new CertificateProvider interface to load key and trust stores. To support refreshing, this interface returns a `Publisher` that can emit new key stores as they become available, e.g. if the key file is changed. I implement a self-signed certificate provider (based on netty-pkitesting instead of the deprecated `SelfSignedCertificate` API), a file-based provider (with automatic FileWatcher-based reloading), and a resource-based provider.

Certificate providers are global and referred to by name using new `key-name` and `trust-name` SSL config properties. This makes it possible to use the same certificate in multiple places. In particular, you can define a single self-signed certificate and then use it as the key for a HTTPS server and as the trust for a HTTPS client, which means we could get rid of `insecure-trust-all-certificates` in many tests (though this is not a priority).

I've also added the option to set the key and trust on a per-listener basis for the HTTP server.

Customization of the netty SSL context builder is possible through the new public, extensible `NettySslContextBuilder` API. Users could extend this class to e.g. define their own more complex trust manager. The `NettySslContextBuilder` class is shared between the server and the client, though to override it users would have to replace the `NettyServerSslFactory` or `NettyClientSslFactory` classes respectively. These classes also have experimental builder methods that are listener- and service-specific, so customization could be applied only to some clients or listeners.


http-client/src/main/java/io/micronaut/http/client/netty/ssl/NettyClientSslFactory.java
```diff
@@ -0,0 +1,57 @@
...
+ * Factory for creating client-side Netty SSL context builders.
+ * Used by the Micronaut HTTP client to construct TCP/HTTP and QUIC/HTTP/3
+ * SSL contexts based on client configuration.
+ *
+ * @author Jonas Konrad
+ * @since 4.10.0
+ */
+@Singleton
+@BootstrapContextCompatible
+public class NettyClientSslFactory {
```
graemerocher, line 35: should this whole type be experimental? Should be be internal for now?

http-client/src/main/java/io/micronaut/http/client/netty/ConnectionManager.java
```diff
@@ -1779,4 +1798,57 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
             }
         }
     }
+
+    private class ClientContextWrapper extends SslContextAutoLoader {
```
graemerocher: can be `final`

http-netty/src/main/java/io/micronaut/http/netty/NettySslContextBuilder.java
```diff
@@ -0,0 +1,379 @@
...
+
+/**
+ * Builder for Netty {@link SslContext} (TCP/HTTP/1.1 and HTTP/2) and {@link QuicSslContext} (HTTP/3).
+ * Consumes {@link java.security.KeyStore} and trust store material and applies Micronaut SSL configuration
+ * such as ciphers, protocols, ALPN, client authentication, and provider selection (JDK vs OpenSSL).
+ *
+ * @author Jonas Konrad
+ * @since 4.10.0
+ */
+public class NettySslContextBuilder {
```
graemerocher, line 51: should this be experimental?

http-netty/src/main/java/io/micronaut/http/netty/SslContextAutoLoader.java
```diff
@@ -0,0 +1,227 @@
...
+        current = holder;
+    }
+
+    /**
+     * Obtain the current SSL context holder and retain the underlying Netty contexts.
+     *
+     * @return the retained holder, or {@code null} if no context is currently available
+     */
+    @Nullable
+    public final synchronized SslContextHolder takeRetained() {
```
graemerocher: is there a way to avoid synchronisation with read/write locks?

http/src/main/java/io/micronaut/http/ssl/AbstractCertificateFileConfig.java
```diff
@@ -0,0 +1,86 @@
...
+import io.micronaut.core.naming.Named;
+
+/**
+ * Base configuration for certificate material loaded from files.
+ * Holds common options shared by concrete file-backed providers.
+ *
+ * @author Jonas Konrad
+ * @since 4.10.0
+ */
+abstract class AbstractCertificateFileConfig implements Named {
```
graemerocher: can this be `sealed`? Should probably annotate with `@Internal`

http/src/main/java/io/micronaut/http/ssl/CertificateProvider.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Provides access to certificate material as {@link KeyStore} instances that can be
+ * consumed by SSL context builders. Implementations may actively refresh and emit
+ * new keystores when underlying sources change.
+ *
+ * @author Jonas Konrad
+ * @since 4.10.0
+ */
+public interface CertificateProvider extends Named {
+    String CONFIG_PREFIX = "micronaut.certificate";
```
graemerocher, line 33: would rename to `micronaut.ssl.certificate` to not create a new group

http/src/main/java/io/micronaut/http/ssl/ResourceCertificateProvider.java
```diff
@@ -0,0 +1,88 @@
...
+import io.micronaut.context.exceptions.ConfigurationException;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.async.publisher.Publishers;
+import io.micronaut.core.io.ResourceResolver;
+import org.reactivestreams.Publisher;
+
+import java.io.InputStream;
+import java.security.KeyStore;
+
+@EachBean(ResourceCertificateProvider.Config.class)
```
graemerocher, line 37: needs javadoc

http/src/main/java/io/micronaut/http/ssl/SelfSignedCertificateProvider.java
```diff
@@ -0,0 +1,179 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+@EachBean(SelfSignedCertificateProvider.Config.class)
+@Requires(classes = X509Bundle.class)
+@BootstrapContextCompatible
+public class SelfSignedCertificateProvider implements CertificateProvider {
```
graemerocher: needs javadoc

graemerocher (changes requested):

http-client/src/main/java/io/micronaut/http/client/netty/ssl/NettyClientSslFactory.java
yawkat: This, the server factory, and NettySslContextBuilder are the new extension points for SSL context customizations, so I want them non-internal immediately. They are flexible enough that they don't need to be experimental.

http/src/main/java/io/micronaut/http/ssl/CertificateProvider.java
yawkat: Isn't it cleaner as a separate group? All the other micronaut.ssl properties are duplicated for different clients and the server, while this new config is global. You could for example set `micronaut.ssl.enabled=false` but still use the certificates, as long as theres `micronaut.server.ssl.enabled=true` elsewhere

graemerocher (approved):

n0tl3ss (approved):

PR #12058, opened 2025-09-20
Added ability to set version in Controller method-endpoint annotation

altro3:
This is first step of preparation to Spring 7.0 (see https://spring.io/blog/2025/09/16/api-versioning-in-spring)

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/VersionTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/VersionTest.java
@@ -65,5 +65,10 @@ public class VersionTest {
         String pingV2() {
             return "pong v2";
         }
+
+        @Get(value = "/ping", version = "3")
+        String pingV3() {
+            return "pong v3";
+        }
     }
 }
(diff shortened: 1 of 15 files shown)


altro3:
@graemerocher @dstepanov @yawkat @sdelamo Also neeed to add support versioning strategy by path segment and media type. See: https://docs.spring.io/spring-framework/reference/7.0-SNAPSHOT/web/webmvc/mvc-config/api-version.html

Also it would be nice to do something like this: https://github.com/spring-projects/spring-framework/issues/35424 - Add ability to inject version to controller version

graemerocher:
so this feature exists a lot longer in Micronaut than it does in Spring and there are already different version resolution strategies, using the media type or path segment would just be another resolver. Injecting the version could be done with an argument binder. 

See https://github.com/micronaut-projects/micronaut-core/blob/4.10.x/router/src/main/java/io/micronaut/web/router/version/resolution/HeaderVersionResolver.java

graemerocher:
In fact I don't see the point of this change. We already have `@Version` which does the same thing and I am against duplicate ways to do the same thing.

altro3:
But why you added `consumes` and `produces` parameters? I thougth the reason, that you want to create an easy way to migrate from spring.

sdelamo:
Thanks for the contribution. 

@altro3 closing this for now

We want to release 4.10.0 core soon and I want to limit the scope.

graemerocher:
Spring only just added support for versioning to their annotations and they decided in their design to use the existing `GetMapping`, `PostMapping` etc. annotations. API versioning as a feature has existed in Micronaut for over 5 years (see https://github.com/micronaut-projects/micronaut-core/pull/4629). 

It is a non-goal to mirror everything the Spring team does.

PR #12072, opened 2025-09-29
Fix support for abstract and inner classes for `@Mixin` and `@ClassImport`

dstepanov:


sdelamo (approved):
should not this target 4.9.x?

graemerocher (approved):

PR #12080, opened 2025-10-01
fix release build

graemerocher:


graemerocher:
https://github.com/micronaut-projects/micronaut-core/pull/12081 looks better

PR #12086, opened 2025-10-02
Precalculate replaces at compilation time

dstepanov:


core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -2486,6 +2527,72 @@ private StatementDef addGetExposedTypes() {
         return StatementDef.multi(statements);
     }
 
+    private StatementDef addReplacesDefinition() {
+        AnnotationMetadata producingAnnotationMetadata = annotationMetadata;
+//        if (beanProducingElement instanceof MethodElement methodElement) {
```
graemerocher: if this code is not needed delete it and don't leave it commented out

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -2486,6 +2527,72 @@ private StatementDef addGetExposedTypes() {
...
+
+        classDefBuilder.addField(replacesField);
+
+        AnnotationClassValue<?> replacesBean = replacesAnnotationValue.annotationClassValue(Replaces.MEMBER_BEAN).orElse(null);
+        String named = replacesAnnotationValue.stringValue(Replaces.MEMBER_NAMED).orElse(null);
+        AnnotationClassValue<?> qualifier = replacesAnnotationValue.annotationClassValue(Replaces.MEMBER_QUALIFIER).orElse(null);
+        AnnotationClassValue<?> replacesFactory = replacesAnnotationValue.annotationClassValue(Replaces.MEMBER_FACTORY).orElse(null);
+
+        if (named != null && qualifier != null) {
+            visitorContext.fail("Both \"named\" and \"qualifier\" should not be present", beanProducingElement);
```
graemerocher: return here after failing or alternatively throw processing exception.

Is there a test for this failure case?

inject/src/main/java/io/micronaut/context/DefaultReplacesDefinition.java
```diff
@@ -0,0 +1,138 @@
...
+ * @param beanType          The type of current bean definition.
+ * @param beanTypeToReplace The type of the bean being replaced.
+ * @param qualifier         The qualifier of the bean being replaced.
+ * @param factoryClass      The factory class of the bean being replaced.
+ * @param <T>               The type of the bean being replaced.
+ * @author Denis Stepanov
+ * @since 5.0
+ */
+@Internal
+public record DefaultReplacesDefinition<T>(@NonNull Class<T> beanType,
```
graemerocher, line 45: does this need to be public?

graemerocher (changes requested):

core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
dstepanov: No, before it was just a configuration exception

inject/src/main/java/io/micronaut/context/DefaultReplacesDefinition.java
dstepanov: It is accessed from multiple places

graemerocher (approved):

PR #12088, opened 2025-10-03
Support dynamic proxies for `TypeHint`

### LINKED ISSUE #11837: Add annotation for GraalVM native + dynamic proxies
### Feature description

When working on demos for Micronaut GraalPy compiled to native images, I realized that we have the `@ReflectiveAccess` and `@TypeHint` annotations to describe that a type is accessed reflectively, but there's no equivalent for dynamic proxies.

In the example below, the type could be annotated with something like `@DynamicProxy` or `@TypeHint(dynamicProxy=true)`:

```
@GraalPyModule("pygal")
public interface PygalModule {
    Bar Bar();

    interface Bar {
        void add(String series, int[] values);
        Svg render();
    }

    interface Svg {
        String decode();
    }
}
```

The underlying reason is that the code does something like this:

```
var result= context.eval("python", """
                import pygal
                pygal
                """);
            var bar =  result.as(PygalModule.class);
```

the `as` method will generate a dynamic proxy. The current workaround is to declare:

```
[
  { "interfaces": [ "mn.graalpy.PygalModule" ] },
  { "interfaces": [ "mn.graalpy.PygalModule$Bar" ] },
  { "interfaces": [ "mn.graalpy.PygalModule$Svg" ] }
]
```

in `META-INF/native-image/demo/demo/proxy-config.json`

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/11837

### DIFF
--- a/core/src/main/java/io/micronaut/core/graal/GraalReflectionConfigurer.java
+++ b/core/src/main/java/io/micronaut/core/graal/GraalReflectionConfigurer.java
@@ -62,12 +62,15 @@ public interface GraalReflectionConfigurer extends AnnotationMetadataProvider {
                 if (t == null) {
                     return;
                 }
-                context.register(t);
+                context.registerDynamicProxy(t);
                 TypeHint.AccessType[] accessTypes = reflectConfig.enumValues("accessType", TypeHint.AccessType.class);
                 // DO NOT change to Set.of(..) which disallows duplicates
                 final Set<TypeHint.AccessType> accessType = CollectionUtils.setOf(
                     accessTypes
                 );
+                if (accessType.contains(TypeHint.AccessType.DYNAMIC_PROXY)) {
+                    context.registerDynamicProxy(t);
+                }
                 if (accessType.contains(TypeHint.AccessType.ALL_PUBLIC_METHODS)) {
                     final Method[] methods = t.getMethods();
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/micronaut/core/graal/GraalReflectionConfigurer.java
```diff
@@ -62,12 +62,15 @@ default void configure(ReflectionConfigurationContext context) {
                 if (t == null) {
                     return;
                 }
-                context.register(t);
+                context.registerDynamicProxy(t);
```
melix: Is this intentional change?

melix (commented):
This isn't exactly the same as #11837, right? As far as I understand this uses the dynamic runtime registration. I think ideally we also want a `@DynamicProxy` annotation similar to `@ReflectiveAccess` which generates `proxy-config.json`.

core/src/main/java/io/micronaut/core/graal/GraalReflectionConfigurer.java
dstepanov: No :(

graemerocher (approved):

melix (approved):

Issue #11829, opened 2025-06-02
When using KSP, Kotlin classes aren't injected properly if they inherit from Java classes using field injection


mikehearn:
### Expected Behavior

```java
// Java
@Bean
public class Parent {
    @Inject ApplicationContext context;
}
```

```kotlin
// Kotlin
@Bean
class Child(private val something: SomeSingleton) : Parent() {
    @PostConstruct fun init() {
        check(context != null)
    }
}
```

I expect this code to pass.

### Actual Behaviour

The generated `$Definition` class doesn't get the necessary code to do the field injection. Note that this works if Parent is also written in Kotlin, because then it seems to be detected as a `setContext` method that needs to be invoked.

I suspect this is some kind of limitation or problem with KSP when it's looking up the inheritance chain. Kotlin doesn't use fields in the same way Java does, classes always expose properties via methods, so I guess the KSP gets confused and doesn't notice the field needs to be injected.

### Version

4.9

graemerocher:
The report has useful snippets but no runnable build or assertion, so I could not validate it end to end from the issue alone. A minimized KSP regression should contain Parent.java with an @Inject ApplicationContext field, Child.kt extending Parent and declaring a constructor-injected dependency, and a post-construct assertion that the inherited field is non-null; keep a second all-Kotlin Parent variant as the control. The likely fault line is KSP's inherited-member model: the Kotlin class element or Core injection-point collection is dropping an @Inject field declared by a Java superclass instead of exposing it as a field injection point. Please attach that minimal project or generated definition output so the fix can be placed confidently in inject-kotlin versus core-processor.

PR #12102, opened 2025-10-05
fix open issue#11829

Jarvx:
This is an AI-generated fix for issue https://github.com/micronaut-projects/micronaut-core/issues/11829. The patch has been tested against version 4.9.0 (as suggested by the branch name).

I’m happy to discuss any improvements or suggestions!


graemerocher:
for each of these AI generated patches we will need tests to verify that the bug that is being fixed is actually fixed. Without tests we can't accept the patch

Jarvx:
> for each of these AI generated patches we will need tests to verify that the bug that is being fixed is actually fixed. Without tests we can't accept the patch

Thanks. I'll push tests soon so that the team can review.

Jarvx:
> for each of these AI generated patches we will need tests to verify that the bug that is being fixed is actually fixed. Without tests we can't accept the patch

I have uploaded the test scripts (groovy scripts) for all the PRs I've made.

graemerocher:
ok but these test scripts are not actual tests and are not integrated with the rest of the build. For the PRs to be in an acceptable state actual tests need to be included

Jarvx:
> ok but these test scripts are not actual tests and are not integrated with the rest of the build. For the PRs to be in an acceptable state actual tests need to be included

Do you mean an actual micronaut app  or a junit test ?   Would you be able to give some initial comments on the tests and patches I have submitted? This helps with improvements from my side.

graemerocher:
I would expect for an issue like this which addresses something related to Kotlin that the PR would include a test similar to this one https://github.com/micronaut-projects/micronaut-core/blob/4.10.x/inject-kotlin/src/test/groovy/io/micronaut/kotlin/processing/beans/BeanDefinitionSpec.groovy that verifies the compiled code is correct the same way as your Groovy script does.

Having a random Groovy script in the root that reproduces doesn't help as it will never be executed and doesn't count as a test.

Jarvx:
> I would expect for an issue like this which addresses something related to Kotlin that the PR would include a test similar to this one https://github.com/micronaut-projects/micronaut-core/blob/4.10.x/inject-kotlin/src/test/groovy/io/micronaut/kotlin/processing/beans/BeanDefinitionSpec.groovy that verifies the compiled code is correct the same way as your Groovy script does.
> 
> Having a random Groovy script in the root that reproduces doesn't help as it will never be executed and doesn't count as a test.

I'd say this isn't a random script : 
https://github.com/micronaut-projects/micronaut-core/blob/6a8c8cc01d7761f624673d3039d13fda970bed9e/reproduce_issue.groovy#L83
https://github.com/micronaut-projects/micronaut-core/blob/6a8c8cc01d7761f624673d3039d13fda970bed9e/reproduce_issue.groovy#L22

The script contains source and it creates the tests in a specific folder.
Regardless,  I'll add some fixes later to extract them into the target folder.

dstepanov:
@Jarvx Simply take a look how tests are written in the project and create a test in a similar style.

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/reproduce/reproduce.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.jupiter.api.assertThrows
+
+class ReproduceIssueSpec {
+
+    @Test
+    fun `Kotlin child inherits field from Java parent with KSP`() {
+        val ctx = ApplicationContext.run()
+
+        // Correctly reference the class for Micronaut to create a bean
+        assertThrows<IllegalStateException> {
```
altro3, line 14: @Jarvx  If you fix this bug, Why do you expect an error? It turns out that after your fix there shouldn't be an error.

altro3 (changes requested):

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/reproduce/reproduce.kt
Jarvx: Thanks. I am working on a new fix currently.  I'll close my existing PRs first.

PR #12104, opened 2025-10-06
Add `getDocumentation(parseContent)` to get the raw javadoc + corrected tests

dstepanov:


graemerocher:
Can you confirm the default behaviour is the same as 4.9.x?

graemerocher (approved):

Issue #12105, opened 2025-10-07
Invalid query value with @Format("csv")


Myshkouski:
### Expected Behavior

When using endpoint definition like this:

```java

@Controller("/formatted")
interface FormattedController {
        @Get("csv")
        String csvList(@QueryValue @Format("CSV") List<String> param)
}

```
... and sending the following request:

```http
GET /formatted/csv?param=a
```

... binded query value for `param` should be equal to:

```json
["a"]
```

The same behavior should be expected for the following multi-value formats - `"ssv"` and`"pipes"`.

### Actual Behaviour

Actual binded query value for `param` equals to `null`. 
This behaviour caused by incorrect implementation of [MultiValuesConverterFactory::splitByDelimiter](https://github.com/micronaut-projects/micronaut-core/blob/433faeeb2a0f4054423b6b28aa4f6aaab17c2f32/core/src/main/java/io/micronaut/core/convert/converters/MultiValuesConverterFactory.java#L188) method.

### Steps To Reproduce

1. Clone repo:
```shell
git clone https://github.com/Myshkouski/micronaut-core.git micronaut-core-fix-issue-1210
cd micronaut-core-fix-issue-1210
```
2. Checkout commit `d331e97 test: update test cases for binding formatted query parameters`:
```shell
git checkout d331e97
```
3. Run tests for `QueryParameterFormattingSpec`:
```shell
./gradlew :micronaut-http-server-netty:test --tests "io.micronaut.http.server.netty.binding.QueryParameterFormattingSpec"
```

### Example Application

https://github.com/Myshkouski/micronaut-core.git

### Version

4.9.10

PR #12106, opened 2025-10-07
Fixes Invalid query value with @Format("csv") #12105

Myshkouski:
See [#12105](https://github.com/micronaut-projects/micronaut-core/issues/12105).


altro3:
@dstepanov @graemerocher This is good fix. I guess you didn't notice it.

graemerocher:
Thanks!

PR #12114, opened 2025-10-10
Correct `ConfigurationMetadataWriterVisitor`

dstepanov:


core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
```diff
@@ -79,6 +80,11 @@ public VisitorKind getVisitorKind() {
         return VisitorKind.AGGREGATING;
     }
 
+    @Override
+    public Set<String> getSupportedAnnotationNames() {
+        return Collections.singleton("*");
```
graemerocher: This should be a Micronaut pattern no? Why everything?

graemerocher (approved):

core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
dstepanov: This might be all variations of `@ConfigurationReader` like `@ConfigurationProperties` `@EachX` but it will not work for custom annotations annotated with `@ConfigurationProperties`

graemerocher: yeah I don't think we should incrementally process all annotations by default

graemerocher: when you get a chance please try this change

dstepanov: `return Set.of(ConfigurationProperties.class.getName(), EachProperty.class.getName(), ConfigurationReader.class.getName(), "io.micronaut.management.endpoint.annotation.Endpoint");
` Should include only use-case we have

Issue #12124, opened 2025-10-15
Regression in dependency injection when inherited package private fields when upgrading from 4.7.x to 4.9.x


graemerocher:
### Expected Behavior

Should inject field, instead does nothing

Example:

[GDK-demo.zip](https://github.com/user-attachments/files/22921440/GDK-demo.zip)

### Actual Behaviour

Injects nothing

### Version

4.9.12

PR #12125, opened 2025-10-15
Fix injecting a field with the same name

dstepanov:


graemerocher:
does this fix https://github.com/micronaut-projects/micronaut-core/issues/12124 ?

graemerocher (approved):

dstepanov:
Yes

PR #12139, opened 2025-10-17
New leak detection for tests

yawkat:
Netty 4.2.7 introduced a new leak detection mechanism for tests that is more reliable and has lower overhead. https://github.com/micronaut-projects/micronaut-test/pull/1291 introduces a micronaut-test module that uses this new mechanism. However, the new detector is more strict about resource lifetimes, so tests need various small adjustments:

- Contexts must be closed at the end of the test
- Event loops must be shut down
- Tests must use only threads created in those tests (no Finalizers, no reactive boundedElastic)
- SSL contexts must be released immediately

There is also one minor leak fix in this PR, related to h2c support, in Http2ServerHandler.

Draft because this requires the new micronaut-test module.


gradle/libs.versions.toml
```diff
@@ -43,7 +43,7 @@ micronaut-build-plugins="7.6.4"
 micronaut-groovy = "4.7.0"
 micronaut-session = "4.6.0"
 micronaut-sql = "5.3.0"
-micronaut-test = "4.6.2"
+micronaut-test = "4.10.1"
```
sdelamo:
```suggestion
micronaut-test = "4.10.2"
```

https://github.com/micronaut-projects/micronaut-test/releases/tag/v4.10.2

sdelamo (approved):

sdelamo:
@yawkat CI Failed

PR #12141, opened 2025-10-17
Make @Retryable capture Exception by default

### LINKED ISSUE #12085: `@Retryable` should capture Exception by default not RuntimeException
### Feature description

By default `@Retryable` lets through any subclass of `Exception` that's not a subclass of `RuntimeException`. I can sorta see the logic of this when dealing with checked exceptions in Java but it makes no sense for any language without checked exceptions, like Kotlin. Arguably I would not expect this in Java either.

I think for Micronaut 5 the default should be changed to Exception.

iamsb97:
Closes #12085 .


graemerocher:
can you get the tests passing and document this change in `breaks.adoc`

graemerocher (approved):

iamsb97:
@graemerocher, I have resolved some conflicts and all the tests are passing. Please check.

Issue #11758, opened 2025-04-17
Nondeterministic behaviour when replacing @Context bean


sdedic:
### Expected Behavior

In my code, I wanted to replace a hikari DataSource, which is a `@Context` bean; this is by definition a singleton auto-initialized with the context start. My factory replaced it as
```
@Replaces(DataSource.class)
@Primary
@EachBean(DataSourceConfiguration.class)
```

Note the missing `@Context`qualifier in this replacement; I would expect that the created bean definition has `@Singleton` (default) scope. This kind of replacement usually works - as the BeanDefinition is replaced incl. its construction procedure.
The expected outcome was that my replacement Bean will be seen as DataSource; subject to BeanCreation listeners etc. and that the original DataSource implementation will not be instantiated at all.

### Actual Behaviour

The observed behaviour with `@Context` scoped beans is that:
- both replacement, but also the original bean instances are created. Any unwanted side effect of the replaced bean's init code will happen.
- sometimes, the original (not replacement) bean is injected
- whether one or two instanceof of the bean are created depends on bean initialization order.

The example application demonstrates, that `@Replaces` changes scope, under regular circumstances. Not documented explcitly, but is implied the concept of "replacing the bean definition" (that includes the scope spec). This can be seen in `ConsumerBean` that injects `Prototype` beans that replace `Singleton` original definition. If it is not indended to work that way, it should be prominently documented and checked during the replacement process with at least warning log.

The example application demonstrates, in the default provided configuration, that both the replacement (`ReplacementMultiBean`) AND original (`MultiBean`) instances are created during startup. Only the proper (`ReplacementMultiBean`) is however injectable. The `MultiBean` should have not been created at all.

Next, if the example app is modified so that `ReplacementConsumer` is `@Order(100)` (instead of -1000), the **original** bean is injected and startup fails on assertion. The `ReplacementMultiBean` is never created:
- `beanCandidateCache` is populated with `ReplacementBeanFactory` definition by `findBeanCandidatesInternal` (OK)
- `DefaultBeanContext.BeanKey` equals for both `ReplacementBeanFactory` and `MultiBeanFactory` methods
- the MultiBean (earlier initialized singleton) is already in the `singletonByArgumentAndQualifier` map

Therefore, even though `SingletonScope.findBeanRegistration` is called with `ReplacementBeanFactory` bean definition, the OTHER factory's bean instance is actually returned.

[...]

PR #12144, opened 2025-10-21
[Gen AI] Fix 11758:  Ensure @Replaces beans are detected globally 

Jarvx:
Issue: https://github.com/micronaut-projects/micronaut-core/issues/11758

#### Bug description

The bug occurs because replacement beans (`@Replaces`) are only discovered among the candidate collection being considered, not across the global registry. As a result, replacements declared elsewhere in the registry (e.g., a non-`@Context` replacement for a `@Context bean`) can be missed, and already-cached singleton instances for the replaced bean can remain in singleton caches, leading to nondeterministic behavior.

#### Solution / fix
The test defines an @Context-scoped bean (`OriginalContextApi`) and a `@Singleton` replacement (`ReplacementSingletonApi`) annotated with `@Replaces(OriginalContextApi)`. It asserts that after starting the ApplicationContext, the original `@Context` bean is not instantiated and that injecting ReplacedApi yields the replacement. On the original code, the test fails because the original `@Context` bean is eagerly created, reproducing the reported issue where replacing a `@Context` bean with a non-@Context bean still instantiates the original. This is aligned with the issue description. 2) 

Patch effectiveness: The patch makes two key changes. In `DefaultBeanContext`, it scans all bean definitions for replacements and prunes any eager (`@Context`) beans from `eagerInit` if they are replaced, regardless of the replacement's scope. This ensures the original @Context bean is not eagerly created when there is any replacement, even if the replacement is not `@Context`. In `SingletonScope`, it removes the fallback from `findBeanRegistrationByExactDefinition` to the type/qualifier-based cache, preventing returning a previously created instance of a different bean definition. This addresses the nondeterministic case where an original singleton could be returned because it matched by type/qualifier. Together, these changes fix both parts of the issue: preventing unwanted eager instantiation and ensuring correct bean selection by exact definition. 

#### Steps to reproduce
```bash
```

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -2032,6 +2032,20 @@ public class DefaultBeanContext implements InitializableBeanContext, Configurabl
                     throw new BeanInstantiationException(MSG_BEAN_DEFINITION + (ref == null ? "<ref discarded>" : ref.getName()) + MSG_COULD_NOT_BE_LOADED + e.getMessage(), e);
                 }
             }
+            // Prune eager beans that are replaced by any replacement bean in the application,
+            // regardless of the replacement scope.
+            List<BeanDefinition<Object>> replacementTypes = new ArrayList<>(4);
+            for (BeanDefinitionProducer producer : beanDefinitionsClasses) {
+                BeanDefinition<Object> def = producer.getDefinitionIfEnabled(this);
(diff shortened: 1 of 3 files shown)


Jarvx:

dstepanov:
For the core functionality always add tests for Java. (Not Kotlin like in this PR, nor Groovy). In this case the test should go to `inject-java` module.

dstepanov:
I don't understand why do you need to scan all the classes. The method has candidates all the possible candidates for this injection.

Jarvx:
This PR is improved by a new test script and a new patch. Hope you find it useful.

Issue #11799, opened 2025-05-13
Allow to disable via configuration the `packages` and `propertySources` keys of the environment endpoint response


sdelamo:
### Feature description

By default the [`env` endpoint](https://docs.micronaut.io/latest/guide/#environmentEndpoint) returns a JSON Object such as: 

```
{
"activeEnvironments": ...
"packages": 
"propertySources": 
}
```

I would like to be able to return just: 
```
{
"activeEnvironments": ...
}
```

For example to be able to know which environments my application is using without having to check the logs. 

It would be good to be able to have configuration property `endpoints.env.keys` with a List of Strings wit the default value `['activeEnvironments','packages','propertySources']`

Users will be able to expose only `activeEnvironments by setting: 

```properties
endpoints.env.keys[0]=activeEnvironments
```

sdelamo:
A workaround is to create a custom endpoint to do this: 

```java
package example.micronaut;

import io.micronaut.context.env.Environment;
import io.micronaut.management.endpoint.annotation.Endpoint;
import io.micronaut.management.endpoint.annotation.Read;
import java.util.LinkedHashMap;
import java.util.Map;

@Endpoint(id = "environments", defaultSensitive = false)
class EnvironmentsEndpoint {
    private final Map<String, Object> result;

    EnvironmentsEndpoint(Environment environment) {
        this.result = new LinkedHashMap<>();
        result.put("activeEnvironments", environment.getActiveNames());
    }

    @Read
    Map<String, Object> getEnvironmentInfo() {
        return this.result;
    }
}
```

ayagmar:
Hello @sdelamo 
This issue is well explained, so thank you 
I have a [PR](https://github.com/micronaut-projects/micronaut-core/pull/11801) open, please let me know what you think and if any adjustments are needed
It's my first contribution so feel free to share "nitpicks"
BR

graemerocher:
Triage: this is completed by merged PR #11801, which adds endpoints.env.activeKeys and tests/documentation for selecting activeEnvironments, packages, and propertySources. The issue can now be closed as completed; the custom endpoint workaround is no longer necessary for this use case.

PR #12145, opened 2025-10-21
[GenAI] Fix 11799:  binding a list to endpoints.env.keys to ensure only 'activeEnvironments' remains.

Jarvx:
Issue:  https://github.com/micronaut-projects/micronaut-core/issues/11799

#### Bug description
To allow selective fields in env endpoint via endpoints.env.keys configuration

#### Solution
What the test file do is configuring the env endpoint via `endpoints.env.keys` to only include 'activeEnvironments' and exclude 'packages' and 'propertySources'. On the original code, the test failed because the implementation only honored endpoints.env.active-keys, not the new endpoints.env.keys alias. 

The patch adds `getKeys()/setKeys(List<String>)` as aliases to the existing activeKeys property, allowing Micronaut's configuration binding to map endpoints.env.keys to the same underlying list. This maintains backward compatibility with endpoints.env.active-keys and satisfies the new configuration option. 

#### Steps to reproduce

Please use this command to observe the output from  provided test case.
```
```

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
@@ -78,25 +78,43 @@ public class EnvironmentEndpoint {
     }
 
     /**
-     * Gets the keys to be displayed by the environment endpoint.
-     * Defaults to ["activeEnvironments", "packages", "propertySources"] if not configured.
-     * Configurable via {@code endpoints.env.active-keys}.
-     *
-     * @return The list of active sections.
-     */
-    public List<String> getActiveKeys() {
-        return activeKeys;
-    }
-
-    /**
-     * Sets the sections to be displayed by the environment endpoint.
-     * Example: {@code endpoints.env.active-keys=activeEnvironments,packages}
-     *
-     * @param activeKeys The list of sections. If an empty list is provided, no sections will be displayed.
-     */
-    public void setActiveKeys(List<String> activeKeys) {
-        this.activeKeys = activeKeys;
-    }
+      * Gets the keys to be displayed by the environment endpoint.
+      * Defaults to ["activeEnvironments", "packages", "propertySources"] if not configured.
+      * Configurable via {@code endpoints.env.active-keys} or {@code endpoints.env.keys}.
+      *
+      * @return The list of active sections.
+      */
+     public List<String> getActiveKeys() {
+         return activeKeys;
+     }
+ 
+     /**
+      * Alias accessor for configuration binding from {@code endpoints.env.keys}.
+      *
+      * @return The list of active sections.
+      */
+     public List<String> getKeys() {
+         return getActiveKeys();
+     }
+ 
+     /**
+      * Sets the sections to be displayed by the environment endpoint.
+      * Example: {@code endpoints.env.active-keys=activeEnvironments,packages}
+      *
+      * @param activeKeys The list of sections. If an empty list is provided, no sections will be displayed.
+      */
+     public void setActiveKeys(List<String> activeKeys) {
+         this.activeKeys = activeKeys;
+     }
+ 
+     /**
+      * Alias mutator for configuration binding from {@code endpoints.env.keys}.
+      *
+      * @param keys The list of sections. If an empty list is provided, no sections will be displayed.
+      */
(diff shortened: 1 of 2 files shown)


Jarvx:

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/reproduce/EnvEndpointKeysConfigTest.kt
```diff
@@ -0,0 +1,151 @@
...
+import io.micronaut.context.ApplicationContext
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpResponse
+import io.micronaut.http.client.HttpClient
+import io.micronaut.runtime.server.EmbeddedServer
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+class EnvEndpointKeysConfigTest {
```
graemerocher: should be in Java or Groovy like all our tests and be placed in `io.micronaut.management.endpoint` since that is the functionality under test

management/src/main/java/io/micronaut/management/endpoint/env/EnvironmentEndpoint.java
```diff
@@ -91,13 +91,24 @@ public List<String> getActiveKeys() {
...
         this.activeKeys = activeKeys;
     }
 
+    /**
+     * Alias for {@link #setActiveKeys(List)} to allow binding from configuration property {@code endpoints.env.keys}.
+     * Example: {@code endpoints.env.keys=activeEnvironments,packages}
+     *
+     * @param keys The list of sections. If an empty list is provided, no sections will be displayed.
+     */
+    public void setKeys(List<String> keys) {
```
graemerocher: this fix seems unnecessary and the functionality already implemented by `setActiveKeys` so it appears this is just a documentation issue and no fix is needed.

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/reproduce/EnvEndpointKeysConfigTest.kt
```diff
@@ -0,0 +1,151 @@
...
+        val props = HashMap<String, Any?>()
+        props["endpoints.env.enabled"] = true
+        // ensure the endpoint is exposed over HTTP for newer Micronaut versions
+        props["management.endpoints.web.exposure.include"] = "env"
+        // try to avoid security blocking the endpoint in test environments
+        props["micronaut.security.enabled"] = false
+        props["micronaut.server.port"] = 0
+
+        server = ApplicationContext.run(EmbeddedServer::class.java, props)
+        client = HttpClient.create(server!!.url)
```
graemerocher: the client here is never closed and will leak resources

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/reproduce/EnvEndpointKeysConfigTest.kt
```diff
@@ -0,0 +1,151 @@
...
+        // fallback: try to locate a known env endpoint bean and serialize it
+        try {
+            val ctx = server!!.applicationContext
+            val candidateClassNames = listOf(
+                "io.micronaut.management.endpoint.env.EnvironmentEndpoint",
+                "io.micronaut.management.endpoint.env.EnvEndpoint",
+                "io.micronaut.management.endpoint.env.EnvironmentController"
+            )
+
+            for (className in candidateClassNames) {
```
graemerocher: this is very weird logic, why is reflection being used?

test-suite-kotlin-ksp/src/test/kotlin/io/micronaut/reproduce/EnvEndpointKeysConfigTest.kt
graemerocher: tests in `EnvironmentEndpointSpec` already test this functionality in a better way, so this test is largely useless

Jarvx:
I do not have the permission to add a label. So I'll just add a tag in the title to indicate this PR has used AI.

Jarvx:
I have pushed a new fix and test scripts. Hope this can be an improvement than the first cut.

Issue #11755, opened 2025-04-16
HttpClientConfiguration.getConnectionPoolConfiguration() is abstract


mikehearn:
### Expected Behavior

The user guide gives this sample:

```java
@Named("twitter")
@Singleton
class TwitterHttpClientConfiguration extends HttpClientConfiguration {
   public TwitterHttpClientConfiguration(ApplicationConfiguration configuration) {
        super(configuration);
    }
}
```

### Actual Behaviour

But it doesn't compile, because you have to implement the `getConnectionPoolConfiguration()` method. The JavaDoc for this method is redundant with the method name and could be deleted, as it doesn't explain how to get or create such a configuration or how to return a default.

### Version

4.8

graemerocher:
Triage: valid documentation/API usability issue with an open direct fix in [#12146](https://github.com/micronaut-projects/micronaut-core/pull/12146). The PR provides a default `ConnectionPoolConfiguration` from `HttpClientConfiguration`, so the guide’s `TwitterHttpClientConfiguration` example compiles while existing concrete overrides remain unchanged, and it adds a reproducer test. Keep open pending review and merge.

PR #12146, opened 2025-10-21
[GenAI] Fix 11755:  Providing a sensible default implementation in the base class that returns a new `ConnectionPoolConfiguration` instance

Jarvx:
Issue: https://github.com/micronaut-projects/micronaut-core/issues/11755

#### Bug description

`TwitterHttpClientConfiguration` does not compile because `getConnectionPoolConfiguration()` must be implemented. 

#### Root cause & Fix

The issue is that `HttpClientConfiguration.getConnectionPoolConfiguration()` is declared abstract, forcing subclasses (like user examples) to implement it. To fix this without changing the behavior of existing concrete classes, we can provide a sensible default implementation in the base class that returns a new `ConnectionPoolConfiguration` instance. Subclasses that already override the method (e.g. `DefaultHttpClientConfiguration`, `ServiceHttpClientConfiguration`) will continue to work unchanged. This is a minimal, backward-compatible change.

#### Steps to reproduce

Please use this command to see the test result:

```bash
```

### DIFF
--- a/test-suite-kotlin-ksp/src/test/java/io/micronaut/reproduce/ReproduceIssueTest.java
+++ b/test-suite-kotlin-ksp/src/test/java/io/micronaut/reproduce/ReproduceIssueTest.java
@@ -1,0 +1,19 @@
+package io.micronaut.reproduce;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Trivial test whose only purpose is to ensure the test sources are compiled.
+ * If HttpClientConfiguration#getConnectionPoolConfiguration() is abstract,
+ * compilation will fail because TwitterHttpClientConfiguration does not
+ * implement that method (matching the user-guide example).
+ */
+public class ReproduceIssueTest {
+
+    @Test
+    void compileOnlySmoke() {
+        // Intentionally empty. Successful compilation of the test sources
+        // (which include TwitterHttpClientConfiguration) indicates the
+        // framework does not declare getConnectionPoolConfiguration() as abstract.
+    }
+}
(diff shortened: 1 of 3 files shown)


test-suite-kotlin-ksp/src/test/java/io/micronaut/reproduce/ReproduceIssueTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Trivial test whose only purpose is to ensure the test sources are compiled.
+ * If HttpClientConfiguration#getConnectionPoolConfiguration() is abstract,
+ * compilation will fail because TwitterHttpClientConfiguration does not
+ * implement that method (matching the user-guide example).
+ */
+public class ReproduceIssueTest {
```
graemerocher, line 11: it doesn't make sense to place test code into a kotlin submodule that is Java. All tests should be collocated near the functionality they address, in this case in the http-client-core module

test-suite-kotlin-ksp/src/test/java/io/micronaut/reproduce/TwitterHttpClientConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+ * io.micronaut.http.client.HttpClientConfiguration that does not
+ * override getConnectionPoolConfiguration().
+ *
+ * If HttpClientConfiguration#getConnectionPoolConfiguration() is declared
+ * abstract in the framework (as reported in the issue), this class will
+ * fail to compile because it does not implement that method.
+ */
+@Named("twitter")
+@Singleton
+public class TwitterHttpClientConfiguration extends HttpClientConfiguration {
```
graemerocher, line 20: what does this even do? It isn't referenced anywhere

graemerocher (changes requested):

Issue #11819, opened 2025-05-26
Multiple `Cookie` headers on request


kristofdho:
### Expected Behavior

Calling `MutableHttpRequest.cookie(s)` keeps a single cookie header up to date.

### Actual Behaviour

Multiple Cookie headers are created depending on the specific request implementation.

### Steps To Reproduce

Specifically ran into this when adding additional cookies in a proxy filter, so calling `request.mutate().cookie(...)` in a `HttpServerFilter`.

This JUnit test shows the issue, based on the fix provided in https://github.com/micronaut-projects/micronaut-core/issues/6356:

```java
class CookieTest {

    private static final Set<Cookie> COOKIES = new LinkedHashSet() {{
        add(Cookie.of("a", "1"));
        add(Cookie.of("b", "2"));
        add(Cookie.of("a", "3"));
    }};
    private static final String EXPECTED_VALUE = "a=3; b=2"; // Comment in NettyHttpClientRequest claims a space is needed :shrug:

    static List<Arguments> httpRequests() {
        MutableHttpRequest<?> simpleRequest = new SimpleHttpRequestFactory().get("/");
        MutableHttpRequest<?> nettyClientRequest = new NettyClientHttpRequestFactory().get("/");
        MutableHttpRequest<?> nettyHttpRequest = new NettyHttpRequest<>(
                new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/"),
                AvailableNettyByteBody.empty(),
                Mockito.mock(ChannelHandlerContext.class),
                Mockito.mock(ConversionService.class),
                Mockito.mock(HttpServerConfiguration.class)).mutate();
        return List.of(
                Arguments.argumentSet(simpleRequest.getClass().getSimpleName(), simpleRequest),
                Arguments.argumentSet(nettyClientRequest.getClass().getSimpleName(), nettyClientRequest),
                Arguments.argumentSet(nettyHttpRequest.getClass().getSimpleName(), nettyHttpRequest)
        );
    }

    @ParameterizedTest
    @MethodSource("httpRequests")
    void requestCookiesOneByOne(MutableHttpRequest<?> request) {
        COOKIES.forEach(request::cookie);

        assertAll(
                () -> assertEquals(1, request.getHeaders().get(HttpHeaders.COOKIE).length()),
                () -> assertEquals(EXPECTED_VALUE, request.getHeaders().get(HttpHeaders.COOKIE))
        );
    }

[...]

PR #12147, opened 2025-10-21
[GenAI] Fix 11819: updating SimpleHttpRequest.updateCookies to remove existing cookie headers

Jarvx:
Issue: https://github.com/micronaut-projects/micronaut-core/issues/11819

#### Bug description

`Calling MutableHttpRequest.cookie(s)` is expected to keep a single Cookie header updated. However, multiple Cookie headers are created depending on the request implementation. 

#### Root cause & Fix
The test reproduces the problem for `SimpleHttpRequest` by  adding cookies, it expects exactly one Cookie header value with the latest value for duplicate names and correct semantics. 
On the original code, `SimpleHttpRequest.updateCookies used headers.add(...)` per cookie, which produced multiple Cookie header entries (one per cookie), causing the test to fail. 

The patch fixes this by building a single combined Cookie header string (joined with \"; \") and setting it once via `headers.set(...)`, thus preventing multiple header entries. 

#### Steps to reproduce

Please use this command to see the output from provided test:
```bash
```

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/NettyClientHttpRequest.java
@@ -131,7 +131,7 @@ public class NettyClientHttpRequest<B> implements MutableHttpRequest<B>, NettyHt
                 String value = ClientCookieEncoder.INSTANCE.encode(cookie);
                 this.cookies.put(cookie.getName(), value);
             }
-            headers.set(HttpHeaderNames.COOKIE, String.join(";", this.cookies.values()));
+            headers.set(HttpHeaderNames.COOKIE, String.join("; ", this.cookies.values()));
         } else if (!cookies.isEmpty()) {
             cookie(cookies.iterator().next());
         }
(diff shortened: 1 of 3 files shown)


Jarvx:

http/src/test/java/io/micronaut/http/cookie/CookieHeaderSingleValueTest.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CookieHeaderSingleValueTest {
```
graemerocher: test should be placed in where other cookie related tests are keep in the `io.micronaut.http.cookie` package of the `micronaut-http` module

Jarvx: Okay. I'll move it to the folder `http/src/test/ `. Please understand that it's not easy to find perfect test case for me at the moment. I'll need your feedback for this information.

Jarvx:
I have improved the logic in the test file and the bug fix. Also added more comments.

Issue #11815, opened 2025-05-22
Requires for bean disabled with DisabledBeanException does not work


cutz:
### Expected Behavior

Conditionally disabling a bean by throwing a `DisabledBeanException` from a factory method should be honored by other beans/factories that `@Require` the disabled bean.

### Actual Behaviour

An exception is raised rather than the requiring bean being disabled.

### Steps To Reproduce

Consider the following factory. 

```
@Factory
public class MyFactory {

    @Singleton
    public MyBean myBean() {
        throw new DisabledBeanException("MyBean Disabled");
    }

    @Singleton
    @Requires(bean = MyBean.class)
    public AnotherBean anotherBean(MyBean myBean) {
        return new AnotherBean(myBean);
    }
}
```

Attempting to find a bean for `AnotherBean` should return an empty optional. 

```
@Test
    void testConditionalMakeAnotherBeanUnavailable() {
        Assertions.assertTrue(application.getApplicationContext().findBean(AnotherBean.class).isEmpty());
    }
```

Instead an exception is raised.

```
Failed to inject value for parameter [myBean] of method [anotherBean] of class: com.example.MyFactory

Message: MyBean Disabled
Path Taken: 
@j.i.Singleton c.e.AnotherBean c.e.MyFactory.anotherBean#anotherBean(MyBean myBean)
\---> @j.i.Singleton c.e.AnotherBean c.e.MyFactory.anotherBean#anotherBean([MyBean myBean])
io.micronaut.context.exceptions.DependencyInjectionException: Failed to inject value for parameter [myBean] of method [anotherBean] of class: com.example.MyFactory

Message: MyBean Disabled
Path Taken: 
@j.i.Singleton c.e.AnotherBean c.e.MyFactory.anotherBean#anotherBean(MyBean myBean)
\---> @j.i.Singleton c.e.AnotherBean c.e.MyFactory.anotherBean#anotherBean([MyBean myBean])
	at io.micronaut.context.AbstractInitializableBeanDefinition.resolveBean(AbstractInitializableBeanDefinition.java:2136)
	at io.micronaut.context.AbstractInitializableBeanDefinition.getBeanForConstructorArgument(AbstractInitializableBeanDefinition.java:1352)
	at com.example.$MyFactory$AnotherBean1$Definition.instantiate(Unknown Source)
	at io.micronaut.context.DefaultBeanContext.resolveByBeanFactory(DefaultBeanContext.java:2335)
[...]

PR #12148, opened 2025-10-21
[GenAI] Fix 11815:   Modifications on AbstractInitializableBeanDefinition.resolveBean

Jarvx:
Issue: https://github.com/micronaut-projects/micronaut-core/issues/11815

#### Bug description

When a bean is conditionally disabled by throwing a DisabledBeanException from a factory method, other beans that @Require the disabled bean should also be considered unavailable.

#### Root Cause & Solutions
The test reproduces the reported issue: a factory method throws DisabledBeanException to disable a bean, and another bean annotated with `@Requires(bean = MyBean.class)` should therefore not be created.

In the original code, the `DisabledBeanException` thrown during dependency resolution was wrapped in a DependencyInjectionException, causing `ctx.findBean`(AnotherBean) to throw instead of returning an empty Optional, and the test failed as shown in the logs. 

The patch modifies `AbstractInitializableBeanDefinition.resolveBean` to rethrow `DisabledBeanException` directly (except for the special EachBean case which remains a `DisabledBeanException`), instead of wrapping it in `DependencyInjectionException`. This allows the upstream context to recognize the disabled dependency and honor @Requires, resulting in `findBean` returning an empty Optional and the test passing.

#### Steps to reproduce

Please use this command to execute the provided test:

```bash
```


Jarvx:
The PR is updated.

Issue #11780, opened 2025-04-25
Allow declarative httpclient to throw on 404


gnarroway-bullish:
### background

- at present, the declarative httpclient by default throws exceptions for any status code >=400 _except_ 404 
- for 404s, it returns null / empty (as documented [here](https://docs.micronaut.io/latest/guide/index.html#clientError))
- this is not intuitive nor idiomatic, and causes confusion (e.g. https://github.com/micronaut-projects/micronaut-core/issues/3104, https://github.com/micronaut-projects/micronaut-core/issues/11778)
- in particular, the behavior differs from popular clients in every language, and is also inconsistent with micronaut's own non-declarative client.
- the fact that the docs require a warning for this edge case is indicative that it is a smell

### impact

- requires code to null check at every point of call, creating risk of unexpected behavior if missed
- makes it difficult to differentiate between a not-found and an intentional empty response
- makes it difficult to swap implementations with other clients safely

### potential fixes

- Ideally it would work like any other client (option 1 below), but that is a breaking change. 
- The alternative option 2 is smelly in that it only applies to the niche case of declarative client + 404 response.

#### 1. Respect existing property

- make it respect `micronaut.http.client.exception-on-error-status` as documented, in line with the other exceptional statuses, and in line with the non-declarative client

#### 2. Add a new config property

- provide a configuration property to force 404 to also throw exception for declarative httpclient, in line with the other exceptional statuses, and in line with the non-declarative client

graemerocher:
First of all `this is not intuitive nor idiomatic` and `differs from popular clients in every language` is not correct. The behaviour is the same in OpenFeign which the client is based on.

See https://github.com/OpenFeign/feign/issues/476 and the rationale https://github.com/OpenFeign/feign/commit/24885fe9620ed620af43f4d2d6ffcfc82980e097#diff-b54b942c0c272042049c1fbdbe6084aa

I agree with making the behaviour configurable however.

graemerocher:
I think making it respect `micronaut.http.client.exception-on-error-status` makes sense

ayagmar:
Hello @graemerocher 
I just forked micronaut-core and started looking into this. I think I found two spots that could be relevant  :
either adding `HttpClientConfiguration `to `HttpClientIntroductionAdvice` and modifying `handleBlockingCall `/ `doOnError`, or maybe it's possible to do the changes directly in `DefaultHttpClient` 

[...]

PR #12149, opened 2025-10-21
Allow 404 error code to throw exception for declarative clients

iamsb97:
Addresses #11780.

Changes:
1. Added a new configuration property `exception-on-404-status` to control whether an HTTP client throws an exception or returns null when a 404 status is received.
2. Introduced a new interface `ConfiguredHttpClient` that exposes the `HttpClientConfiguration` associated with an individual client instance. This allows `HttpClientIntroductionAdvice` to access the client’s configuration in order to respect the new `exception-on-404-status` setting.
3. The default value for the new property is `false`, preserving existing behavior for backward compatibility.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/ConfiguredHttpClient.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/ConfiguredHttpClient.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017-2025 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.client;
+
+import io.micronaut.core.annotation.NonNull;
+
+/**
+ * All HTTP client implementations provide access to their
+ * configuration through this interface.
+ *
+ * @since 4.10.x
+ */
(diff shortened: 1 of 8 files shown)


iamsb97:
@graemerocher, please review this when you get a chance. Thanks!

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -288,17 +290,20 @@ protected void doOnError(Throwable t) {
...
-                            future.complete(e.getResponse());
-                        } else {
-                            future.complete(null);
-                        }
-                        return;
+                if (t instanceof HttpClientResponseException e && e.code() == HttpStatus.NOT_FOUND.getCode()) {
+                    HttpClientConfiguration config = httpClient.getConfiguration();
+                    if (config.isExceptionOn404Status()) {
+                        future.completeExceptionally(t);
+                    }
```
yawkat, line 297: missing an else here

http-client-core/src/main/java/io/micronaut/http/client/ConfiguredHttpClient.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.micronaut.core.annotation.NonNull;
+
+/**
+ * All HTTP client implementations provide access to their
+ * configuration through this interface.
+ *
+ * @since 4.10.x
+ */
+public interface ConfiguredHttpClient {
```
yawkat, line 26: Not sure I like having to add this interface…

iamsb97: The rationale behind adding this interface is that all HttpClient implementations already have the configuration field, so this would just be exposing it. 

1. Adding the getConfiguration() directly to HttpClient interface didn't seem appropriate as it only contains functional methods. 
2. Couldn't use `instanceof` as http-client-core does not depend on http-client.

To make this configurable without breaking any other behavior, I think the logic needs to be in `HttpClientIntroductionAdvice`, but currently there's no way to retrieve the client's configuration there.

Any suggestions on a better approach?

PR #12158, opened 2025-10-24
Introduce TaskQueueInterceptor

### LINKED ISSUE #12157: No Matching Netty Transport after upgrade to Micronaut 4.10.0
### Expected Behavior

After upgrading to Micronaut 4.10.0, a Micronaut HTTP service should start without error.

### Actual Behaviour

One of our services fails to start due to the following error:

> Caused by: java.util.NoSuchElementException: No matching transport was found. Configured transports are [nio], available transports are [instrumentedNioEventLoopGroupFactory, epoll, native]

This only occurs in one of our services.  All services are containerized and none of them contain any Micronaut configuration to change/modify the Netty event loop configuration.  It is also worth noting that we can only reproduce this in one of our clusters.

### Steps To Reproduce

1. Update to Micronaut 4.10.0 (from 4.9.x)
2. Start the service
3. Service fails on startup due to the `No matching transport was found. Configured transports are [nio], available transports are [instrumentedNioEventLoopGroupFactory, epoll, native]` error.

### Environment Information

 Service running in Docker container (airbyte-workload-api-service)

### Example Application

https://github.com/airbytehq/airbyte-platform

### Version

4.10.0

yawkat:
With the new Netty event loop APIs, we need a new approach for intercepting the event loop task queue.

This is required to fix #12157, so the PR is against 4.10.x

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/channel/DefaultEventLoopGroupRegistry.java
@@ -30,8 +30,10 @@ import io.micronaut.core.util.ArgumentUtils;
 import io.micronaut.http.netty.channel.loom.LoomCarrierGroup;
 import io.micronaut.inject.qualifiers.Qualifiers;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.IoEventLoop;
 import io.netty.channel.IoHandlerFactory;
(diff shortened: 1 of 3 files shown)


sdelamo (approved):

sdelamo (commented):
Can we add documentation for this?

graemerocher (approved):

pditommaso:
> Can we add documentation for this?

Yes, please 🙏

Issue #11656, opened 2025-03-13
Postconstruct is called twice for programmaticaly created bean


loicmathieu:
### Expected Behavior

When a bean is created manually then registed via regiserSingleton, if it has a method annoted with `@PostConstruct` this method must be called once.

### Actual Behaviour

Take as example the following bean:

```java
@Singleton
public class MyBean {

    @PostConstruct
    void init() {
        System.out.println("MyBean init " + this);
    }

    public String hello() {
        return "hello";
    }
}
```

When constructed then registered programmatically like that:

```java
public class Application {

    public static void main(String[] args) {
        ApplicationContext applicationContext = Micronaut.run(Application.class, args);
        MyBean bean = applicationContext.createBean(MyBean.class);
        applicationContext.registerSingleton(bean);

        System.out.println(bean.hello());
    }
}
```

The postconstruct method is called twice.

### Steps To Reproduce

[double-postconstruct.zip](https://github.com/user-attachments/files/19230362/double-postconstruct.zip)

### Environment Information

- OS: Linux 6.11.0-19-generic #19-Ubuntu SMP PREEMPT_DYNAMIC Wed Feb 12 21:43:43 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
- Java: openjdk version "21.0.5" 2024-10-15 LTS

### Version

4.7.6

graemerocher:
I don't think this is a supported pattern (ie. calling `createBean` and then `registerSingleton`). Changing this behaviour my break existing users as well.

loicmathieu:
The documentation of `createBean` says that the singleton is not registered so I had to register it manually if I want to be able to inject it.

I create beans conditionally, in a command, with parameters passed as command line options so I had to use `createBean`.

Maybe a solution would be to create the bean with `new`, manually do injection (applicationContext.init()) then register it?

graemerocher:
`createBean` is normally not used in combination with `@Singleton` because a singleton is... well singleton, there is only one

graemerocher:
Triage: the observed double callback is a consequence of combining two lifecycle operations: createBean constructs and initializes the object, then registerSingleton registers an already initialized instance and may initialize it again. The supported alternatives are to use createBean with constructor arguments and retain/pass the returned instance, or let Micronaut create the singleton through a factory/conditional bean definition. Since the reported sequence is outside the supported lifecycle contract, closing as not a bug.

PR #12159, opened 2025-10-24
[GenAI] Fixing 11656:  Handling the issue with @PostConstruct annotated bean

Jarvx:
**Issue link**:  https://github.com/micronaut-projects/micronaut-core/issues/11656

### Issue description

1) The test reproduces the issue: it creates a bean via `ApplicationContext.createBean` (which should invoke `@PostConstruct once`), then registers the same instance via r`egisterSingleton` and asserts that `@PostConstruct` is not invoked a second time. On the original code, the test failed with initCount=2, confirming the double @PostConstruct invocation. 

2) The patch resolves the issue by tracking instances created by the context (via `createBean`) in an identity-based set. When registerSingleton is called with `inject=true`, it checks this set: if the instance was already initialized by the context, it performs injection only (without initialization), avoiding a second `@PostConstruct` call. The test passes after the patch, and the full test suite builds successfully. 

**Note**: this approach introduces a global set of initialized instances, which may have implications (e.g., strong references and non-concurrent collection); 

### Solution
please run the command 

```bash
```

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -199,7 +199,10 @@ public class DefaultBeanContext implements InitializableBeanContext, Configurabl
     private final Map<Argument, Collection<BeanDefinition>> beanCandidateCache = new ConcurrentLinkedHashMap.Builder<Argument, Collection<BeanDefinition>>().maximumWeightedCapacity(30).build();
 
     private final Map<Class<?>, Collection<BeanDefinitionProducer>> beanIndex = new ConcurrentHashMap<>(12);
-
+    
+    // Tracks instances that have already been initialized (e.g., via createBean) to avoid double @PostConstruct on registration
+    private final Set<Object> initializedInstances = Collections.newSetFromMap(new IdentityHashMap<>());
+    
     private final ClassLoader classLoader;
     private final Set<Class<?>> thisInterfaces = CollectionUtils.setOf(
             BeanDefinitionRegistry.class,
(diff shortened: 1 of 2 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -691,7 +699,19 @@ public <T> BeanContext registerSingleton(@NonNull Class<T> type, @NonNull T sing
         if (beanDefinition != null && !(beanDefinition instanceof RuntimeBeanDefinition<T>) && beanDefinition.getBeanType().isInstance(singleton)) {
             try (BeanResolutionContext context = newResolutionContext(beanDefinition, null)) {
                 if (inject) {
-                    doInjectAndInitialize(context, singleton, beanDefinition);
+                    // If this instance was created by this context earlier, it may already have been injected/initialized.
+                    // In such case skip duplicate injection/initialization. Use the beansCreatedByContext set to detect this.
+                    boolean createdByThisContext;
+                    synchronized (beansCreatedByContext) {
```
altro3: why you use synchronizedSet together with synchronized block?

Jarvx: Oh.  `Collections.synchronizedSet(...)` already ensures that all accesses 

so the block is redundant. I’ll remove the extra synchronization. Thanks.

dstepanov:
It shouldn't matter if the bean is created by the bean context or not, the triggering of the post construct should act the same if the bean is create manually or not. Not sure for the current behaviour. I would expect that `createBean` will trigger all the life-cycle methods and register will not.

test-suite/src/test/java/io/micronaut/reproduce/DoublePostConstructTest.java
```diff
@@ -0,0 +1,27 @@
...
+import io.micronaut.context.ApplicationContext;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertSame;
+
+public class DoublePostConstructTest {
+
+    @Test
+    void postConstructCalledOnceWhenCreateAndRegisterSingleton() {
```
graemerocher: This PR may result in a memory leak since each call to `createBean` will create a new instance to be tracked forever. This is not the current behaviour of `createBean` which relies on the user to invoke post construct hooks.

It is actually debatable whether this is even a bug

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -656,28 +697,40 @@ public <T> BeanContext registerSingleton(@NonNull Class<T> type, @NonNull T sing
...
         if (beanDefinition != null && !(beanDefinition instanceof RuntimeBeanDefinition<T>) && beanDefinition.getBeanType().isInstance(singleton)) {
-            if (inject) {
-                try (BeanResolutionContext context = newResolutionContext(beanDefinition, null)) {
-                    doInjectAndInitialize(context, singleton, beanDefinition);
+            try (BeanResolutionContext context = newResolutionContext(beanDefinition, null)) {
+                if (inject) {
+                    // If this instance was created by this context earlier, it may already have been injected/initialized.
+                    // In such case skip duplicate injection/initialization. Use the beansCreatedByContext set to detect this.
+                    boolean createdByThisContext;
+                    synchronized (beansCreatedByContext) {
```
graemerocher: introduction of synchronisation on the `createBean` path will likely result in performance degradation

graemerocher (changes requested):

Jarvx:
I have updated this PR by adding new fixes and new reproducer test file.

dstepanov:
As I wrote before, tests need to be in Java for the core functionality

PR #12165, opened 2025-11-03
Update to jackson 3

### LINKED ISSUE #12152: Update to Jackson 3
https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0

iamsb97:
Closes #12152.

Json key ordering was updated in the tests because jackson 3 has `SORT_PROPERTIES_ALPHABETICALLY` set to true by default.

**Potential issue**: Lombok integration test breaks because Jacksonised annotation still refers to Jackson 2 (I think). Would probably have to wait until https://github.com/projectlombok/lombok/pull/3960 is merged.


iamsb97:
Hi @yawkat, I could use some help with a couple of issues.

1. `"creator property that doesn't have a getter"` test failure
This test in `BeanIntrospectionModuleSpec` fails when `ignoreReflectiveProperties` is `false`. The args passed to `createFromObjectsWith` in `BeanIntrospectionModule.java` are null, which seems to originate from Jackson. It looks like a change in Jackson 3, but I’m not certain if I’m missing something here.
2. `:benchmarks:typeCheckTest` failure
This fails after the change in `JacksonDatabindMapper` (around line 238) where I switched to using an `ObjectWriter` with view information since `ObjectMapper` is now immutable. If I revert that change, the benchmarks succeed. I compared the two code paths and they seem equivalent, so I’m not sure what’s causing the difference.

Would you mind taking a look at these?

discovery-core/src/main/java/io/micronaut/discovery/cloud/digitalocean/DigitalOceanMetadataResolver.java
```diff
@@ -112,7 +112,7 @@ public DigitalOceanMetadataResolver() {
      * @param configuration Digital Ocean Metadata configuration
      */
     public DigitalOceanMetadataResolver(ObjectMapper objectMapper, DigitalOceanMetadataConfiguration configuration) {
-        this(configuration, objectMapper.getFactory(), new JacksonDatabindMapper(objectMapper));
+        this(configuration, (JsonFactory) objectMapper.tokenStreamFactory(), new JacksonDatabindMapper(objectMapper));
```
yawkat, line 118: maybe we should just move to JsonMapper on the bean context

iamsb97: Do you mean moving Micronaut’s codebase to use Jackson’s JsonMapper instead of ObjectMapper more generally?

yawkat: yep

iamsb97: Okay, sounds good! Would you like me to make that change in this PR, or would you prefer it in a separate one?

yawkat: This PR is fine. I'll look at it when I have time

benchmarks/src/typeCheckTest/java/example/TypeThrashingTest.java
```diff
@@ -33,7 +33,17 @@ public class TypeThrashingTest {
...
     static void setupAgent() {
-        TypePollutionTransformer.install(net.bytebuddy.agent.ByteBuddyAgent.install());
+        java.lang.instrument.Instrumentation inst = net.bytebuddy.agent.ByteBuddyAgent.install();
+        net.bytebuddy.agent.builder.AgentBuilder.Transformer transformer = io.micronaut.test.typepollution.TypePollutionTransformer.create();
+        new net.bytebuddy.agent.builder.AgentBuilder.Default()
+            // Ignore Gradle and Byte Buddy internals to avoid resolution issues
+            .ignore(net.bytebuddy.matcher.ElementMatchers.nameStartsWith("org.gradle.")
+                .or(net.bytebuddy.matcher.ElementMatchers.nameStartsWith("net.bytebuddy.")))
+            // Only transform Micronaut classes (the code under test)
+            .type(net.bytebuddy.matcher.ElementMatchers.nameStartsWith("io.micronaut."))
```
yawkat: no, we can't have this restriction.

jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
```diff
@@ -237,7 +237,12 @@ public <T> T readValue(@NonNull ByteBuffer<?> byteBuffer, @NonNull Argument<T> t
     @Override
     public void writeValue(@NonNull OutputStream outputStream, @Nullable Object object) throws IOException {
         if (specializedWriter != null) {
-            specializedWriter.writeValue(outputStream, object);
+            if (object == null) {
+                specializedWriter.writeValue(outputStream, null);
+            } else {
+                // Bind root type to retain static typing while preserving active view to avoid excessive dynamic lookups
+                specializedWriter.forType(object.getClass()).writeValue(outputStream, object);
```
yawkat: i don't understand why this is necessary, `.forType(object.getClass()).writeValue(outputStream, object);` should be equivalent to `.writeValue(outputStream, object);`

yawkat (approved):

yawkat:
Thanks!

PR #12172, opened 2025-11-06
Make http header list size configurable in HTTP client for HTTP2

Pavlov7:
Description

This PR adds configurable maxHeaderListSize property to HTTP client configuration for HTTP2.

    Added maxHeaderListSize configuration property with getter/setter methods.
    Updated Http2FrameCodec instantiation to use the configurable header list size.

Considerations

    Can be configured via micronaut.http.client.http2.max-header-list-size property.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -1274,11 +1274,19 @@ public abstract class HttpClientConfiguration {
          */
         public static final String PREFIX = "http2";
 
+        /**
+         * The default max header list size in bytes.
+         */
+        @SuppressWarnings("WeakerAccess")
+        public static final int DEFAULT_MAX_HEADER_LIST_SIZE = 8192;
+
         private Duration pingIntervalRead = null;
 
         private Duration pingIntervalWrite = null;
 
         private Duration pingIntervalIdle = null;
+
+        private int maxHeaderListSize = DEFAULT_MAX_HEADER_LIST_SIZE;
 
         /**
          * For HTTP/2 connections, the interval from the last inbound message to when an automated ping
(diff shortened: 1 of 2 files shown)


yawkat:
Please take a look at Http2Settings and implement it similar to that config.

Pavlov7:
> Please take a look at Http2Settings and implement it similar to that config.

The main difference I see between Http2ClientConfiguration and Http2Settings is that Http2Settings is holding an instance of io.netty.handler.codec.http2.Http2Settings and getters and setters are reading from/writing to that instance.
I am not sure whether it would be better for Http2ClientConfiguration to also start holding an instance of io.netty.handler.codec.http2.Http2Settings, because then the micronaut-http-client-core module has to start depending on micronaut-http-netty so that it transitively starts depending on io.netty.codec.http2

Pavlov7:
> > Please take a look at Http2Settings and implement it similar to that config.
> 
> The main difference I see between Http2ClientConfiguration and Http2Settings is that Http2Settings is holding an instance of io.netty.handler.codec.http2.Http2Settings and getters and setters are reading from/writing to that instance. I am not sure whether it would be better for Http2ClientConfiguration to also start holding an instance of io.netty.handler.codec.http2.Http2Settings, because then the micronaut-http-client-core module has to start depending on micronaut-http-netty so that it transitively starts depending on io.netty.codec.http2

^ @yawkat

graemerocher:
@yawkat please review the feedback

yawkat:
Yea you're right. My idea doesn't work.

jdeex:
@Pavlov7 I think this one broke the consul client. It doesn't get any HTT2Configuration and tries to override the defaults causing a nullpointer. I set the 3 configuration options for max-header-list-size but HTT2Configuration is always null. Also did a quick fix to avoid the nullpointer on #12389, but not sure if that's the way you would like to go... We cannot upgrade to latest versions because of this

PR #12173, opened 2025-11-09
Pass HttpRequest to getLoadBalancerDiscriminator()

### LINKED ISSUE #8823: LoadBalancer.java: provide the outgoing http request during select()
### Feature description

I would like to write my own LoadBalancer which selects the right ServerInstance based on the outgoing HTTP request. Sadly, this is currently impossible. While a LoadBalancer has an `Object discrminiator` on which the LB can rely on:

https://github.com/micronaut-projects/micronaut-core/blob/3adad0002a19c6288f508ae00b01c8b2596203fe/http-client-core/src/main/java/io/micronaut/http/client/LoadBalancer.java#L42

 there is no way one could specify the discriminator.

DefaultHttpClient's resolveURI invokes `getLoadBalancerDiscriminator()`:

https://github.com/micronaut-projects/micronaut-core/blob/3adad0002a19c6288f508ae00b01c8b2596203fe/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java#L1504

but `getLoadBalancerDiscrminator()` is always empty:

https://github.com/micronaut-projects/micronaut-core/blob/3adad0002a19c6288f508ae00b01c8b2596203fe/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java#L1235-L1237

And although it has a protected access modifier, it does not give me any context about the outgoing request, making an extension DefaultHttpClient pointless.

---

It looks like there was an attempt to give context to the LoadBalancer. Could you help me understand why it has been left this way, and whether there are plans which helps us derive our own discriminator value?

Thank you.

iamsb97:
Closes #8823. 

Changes:
1. Modified `getLoadBalancerDiscriminator()` to accept `HttpRequest` parameter.
2. Default implementation now passes the request as the discriminator.
3. Applied to both `DefaultHttpClient` and `AbstractJdkHttpClient`

This provides users with the flexibility to:
1. Override `getLoadBalancerDiscriminator(HttpRequest)` to create custom discriminator values using the request.
2. Implement custom `LoadBalancer.select()` logic that routes based on request properties, rather than receiving null.

Existing LoadBalancer implementations that ignore the discriminator parameter continue to work as before.

### DIFF
--- a/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
+++ b/http-client-jdk/src/main/java/io/micronaut/http/client/jdk/AbstractJdkHttpClient.java
@@ -358,13 +358,21 @@ abstract class AbstractJdkHttpClient {
         return resolveURI(request);
     }
 
+    /**
+     * @param request The request object
+     * @return The discriminator to use when selecting a server for the purposes of load balancing (defaults to {@link io.micronaut.http.HttpRequest})
+     */
+    protected Object getLoadBalancerDiscriminator(io.micronaut.http.HttpRequest<?> request) {
+        return request;
+    }
+
     private <I> Mono<URI> resolveURI(io.micronaut.http.HttpRequest<I> request) {
         URI requestURI = request.getUri();
         if (loadBalancer == null) {
             return Mono.error(populateServiceId(new NoHostException("Request URI specifies no host to connect to"), clientId, configuration));
         }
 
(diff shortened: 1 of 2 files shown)


graemerocher:
there are some test failures

iamsb97:
Can you please re-run the CI checks? The previous run got stuck and didn’t complete.

iamsb97:
All tests have passed after syncing with the latest 5.0.x.

PR #12175, opened 2025-11-10
Fix lookup of annotation processors in Java 25

melix:
Our test fixtures for compiling java sources use the classpath, which is no longer scanned by default in Java 25. This commit adds the `-proc:full` compiler flag so that they are discovered again.


melix:
PR doesn't pass because of other tests failures on Java 25 which also need to be fixed, but not in that PR.

dstepanov (approved):

PR #12176, opened 2025-11-10
Adjust test for Java 25

melix:
The PR will not pass because there are other test failures (see #12175) but this fixes an independent problem.


yawkat (approved):

melix:
@yawkat also had to update the ciphers list, does it make sense to you?

yawkat:
hmm that part of the test is supposed to match `micronaut.server.ssl.ciphers`, not sure why it's being ignored

melix:
Ok I'll leave this one for you to investigate then :)

PR #12180, opened 2025-11-12
executable method processor and eager initialization

sdelamo:
This PR reproduces the issue with eager initialization and executable method processors. 

I think that when you have eager initialization of singletons Micronaut should:

A)  initialize singletons of type ExecutableMethodProcessor`
B) Execute process method,
C) Initialize the rest of the singletons

This is related to https://github.com/micronaut-projects/micronaut-mcp/issues/97


sdelamo:
I think we should backport this to Micronaut 4. Thoughts?

graemerocher (approved):

PR #12191, opened 2025-11-17
build(http-client): default to JDK client via transitive API dependency

araysse:
- Add api(projects.micronautHttpClientJdk) so depending on micronaut-http-client brings in the lightweight JDK client by default
- Demote Netty from API to optional: compileOnly(projects.micronautHttpNetty) and keep testImplementation for tests
- Netty beans stay disabled by @Requires when Netty is absent; if Netty is present, Netty client beans are enabled and used

Change summary in http-client/build.gradle.kts:

- Replaced:

  - api(projects.micronautHttpNetty)
  - api(libs.managed.netty.handler.proxy)

- With:

  - compileOnly(projects.micronautHttpNetty)
  - testImplementation(projects.micronautHttpNetty)
  - api(projects.micronautHttpClientJdk)
  - compileOnly(libs.managed.netty.handler.proxy)

### DIFF
--- a/.oca/custom_code_review_guidelines.txt
+++ b/.oca/custom_code_review_guidelines.txt
@@ -1,0 +1,24 @@
+# Sample guideline, please follow similar structure for guideline with code samples
+# 1. Suggest using streams instead of simple loops for better readability.
+#          <example>
+#          *Comment:
+#          Category: Minor
+#          Issue: Use streams instead of a loop for better readability.
+#          Code Block:
+#
+#        ```java
+#            // Calculate squares of numbers
+#            List<Integer> squares = new ArrayList<>();
+#            for (int number : numbers) {
+#              squares.add(number * number);
+#            }
+#          ```
+#          Recommendation:
+#
+#        ```java
+#            // Calculate squares of numbers
+#            List<Integer> squares = Arrays.stream(numbers)
+#              .map(n -> n * n) // Map each number to its square
+#              .toList();
+#          ```
+#          </example>
--- a/buildSrc/gradle.properties
+++ b/buildSrc/gradle.properties
@@ -1,0 +1,3 @@
+#
+#Thu Nov 13 11:09:14 WEST 2025
+micronaut-build-version=8.0.0-M9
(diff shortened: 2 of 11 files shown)


http-client-jdk/src/main/java/io/micronaut/http/client/jdk/cookie/DefaultCookieDecoder.java
```diff
@@ -41,6 +41,6 @@ public Optional<Cookies> decode(HttpRequest<?> request) {
 
     @Override
     public int getOrder() {
-        return NettyCookieDecoder.ORDER + 1;
+        return 2;
```
sdelamo, line 44: why change this? Please, rever to the previous value.

buildSrc/gradle.properties
```diff
@@ -0,0 +1,3 @@
+#
+#Thu Nov 13 11:09:14 WEST 2025
+micronaut-build-version=8.0.0-M9
```
sdelamo, line 3: delete this file

.oca/custom_code_review_guidelines.txt
```diff
@@ -0,0 +1,24 @@
+# Sample guideline, please follow similar structure for guideline with code samples
```
sdelamo, line 1: delete this file. if add such files, they should be added in the project template repository.

http-client/build.gradle.kts
```diff
@@ -16,8 +16,10 @@ dependencies {
     api(projects.micronautContext)
     api(projects.micronautHttpClientCore)
     api(projects.micronautWebsocket)
-    api(projects.micronautHttpNetty)
-    api(libs.managed.netty.handler.proxy)
+    compileOnly(projects.micronautHttpNetty)
+    testImplementation(projects.micronautHttpNetty)
+    api(projects.micronautHttpClientJdk)
```
sdelamo, line 21: this is incorrect. You want to add the micronaut-http-client dependency not the micronaut-http-client-jdk

sdelamo (changes requested):

PR #12192, opened 2025-11-17
Introduce a Micronaut module descriptor

melix:
This commit introduces a facility to list the Micronaut modules which are found at runtime. This will be useful diagnostics information, which can for example be used at runtime by the Micronaut Control Panel.

It is implemented 2-fold. First of all, an independent module named `micronaut-module-info`, standalone without any dependencies, defines the base interface and abstract class that can be used for such descriptors. The goal is to load the descriptors using service loading, independently of the framework. The reason we do this is because this module can then be used by the build plugins to either:
   - generate an implementation of a descriptor, which will then be loaded by Micronaut
   - statically reason about modules, for example for linting at build time.

The second module, `micronaut-module-info-runtime`, maps these services into Micronaut beans. A singleton, `MicronautRuntimeModules` can be used to fetch information about modules, which are then resolved as a hierarchy of modules.

For now, the implementation is limited, provides minimal information such a name for the module, its version, the Maven coordinates of the module and some tags. Eventually, we may enrich this with more information such as the dependencies of the module, or compatibility with other modules. For example this could become possible to express runtime requirements such as "module A requires at least one implementation of B at runtime, but not more than 1".

Or, it could become possible for a module to explain which test resources modules it requires, or even supply a list of MCP servers that could be used for coding with AI...


module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautModuleRuntimeInfoFactory.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashMap;
+
+/**
+ * Factory that creates runtime bean definitions for {@link MicronautRuntimeModule} instances.
+ * This factory loads all {@link MicronautModuleInfo} instances and converts them to
+ * resolved {@link MicronautRuntimeModule} instances with proper parent/child relationships.
+ */
+@Internal
+@Factory
+class MicronautModuleRuntimeInfoFactory {
```
graemerocher: can be `final`

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautModuleRuntimeInfoFactory.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Singleton
+    public List<MicronautRuntimeModule> getAllModules(ApplicationContext context) {
```
graemerocher: you could just define `List<MicronautModuleInfo> moduleInfos` as a method argument

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautModuleRuntimeInfoFactory.java
```diff
@@ -0,0 +1,124 @@
...
+        var moduleInfos = context.getBeansOfType(MicronautModuleInfo.class);
```
graemerocher: we might want to a GraalVM optimisation to load this statically at build time

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautRuntimeModule.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.Set;
+
+/**
+ * Represents a resolved Micronaut module at runtime with parent and children references.
+ * Instead of IDs, this runtime bean uses references to model the hierarchy and replaces
+ * MicronautModuleInfo which uses IDs for hierarchical relationships.
+ */
+@Experimental
+@NonNull
+public final class MicronautRuntimeModule {
```
graemerocher, line 35: add nullability annotations to API contract

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautRuntimeModule.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * A set of tags assigned to this module.
+     * @return the set of tags for this module
+     */
+    public Set<String> getTags() {
+        return info.getTags();
+    }
+
+    public MicronautRuntimeModule getRoot() {
```
graemerocher, line 123: javadoc

module-info/src/main/java/io/micronaut/module/info/MicronautModuleInfo.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Represents information about Micronaut modules which are available at runtime.
+ * This can be used for diagnostics. Information is loaded via service loading.
+ * Each module is recommended to extend the abstract class instead of directly implementing this interface.
+ */
+public interface MicronautModuleInfo {
```
graemerocher, line 26: add nullability annotations to API contract. Add `@Experimental`

graemerocher (changes requested):
How is it anticipated the module info will be populated? By some build tool?

graemerocher:
I management endpoint would be useful I think to see what modules are active and what beans each provide.

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautModuleRuntimeInfoFactory.java
melix: There is an optimization in Micronaut AOT for service loading, which proves to be redundant nowadays.

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautRuntimeModule.java
melix: What do you mean? `@NonNull` isn't enough?

module-info/src/main/java/io/micronaut/module/info/MicronautModuleInfo.java
melix: We can't here: this module intentionally doesn't depend on anything Micronaut related.

melix:
> How is it anticipated the module info will be populated? By some build tool?

I am working on a Micronaut Build change to automatically generate classes for each module.

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautRuntimeModule.java
alvarosanchez: I didn't know you could use `@NonNull` at a type level. What's the behaviour then?

It's own Javadoc says:

> Should be used at parameter, return value, and field level

alvarosanchez: What I wonder, on the other hand, is? can this be a record?

module-info-runtime/src/test/java/io/micronaut/module/info/runtime/ServiceLoaderDefinedModule.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.module.info.runtime;
+
+import io.micronaut.module.info.AbstractMicronautModuleInfo;
+
+import java.util.Set;
+
+public class ServiceLoaderDefinedModule extends AbstractMicronautModuleInfo {
```
alvarosanchez, line 22: This is more or less what you will generate with micronaut-build right?

module-info/src/main/java/io/micronaut/module/info/AbstractMicronautModuleInfo.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Base class for Micronaut module information.
+ * It is recommended, though not required, to extend this class.
+ */
+public abstract class AbstractMicronautModuleInfo implements MicronautModuleInfo {
```
alvarosanchez, line 26: Can be a record?

alvarosanchez (approved):

module-info-runtime/src/test/java/io/micronaut/module/info/runtime/ServiceLoaderDefinedModule.java
melix: correct

module-info/src/main/java/io/micronaut/module/info/AbstractMicronautModuleInfo.java
melix: No because each generated class will have to be different and a record cannot be extended. In addition module descriptors may add some logic later (e.g compatibility with other modules).

module-info-runtime/src/test/java/io/micronaut/module/info/runtime/ServiceLoaderDefinedModule.java
melix: It cannot be a record unless you want to give away the `Optional` methods (null safety).

module-info-runtime/src/main/java/io/micronaut/module/info/runtime/MicronautRuntimeModule.java
melix: mmm, I thought this worked like in JSpecify... @graemerocher what do you think?

graemerocher (approved):

PR #12193, opened 2025-11-18
Support deserializing json via jackson into kotlin classes that have default properties and use the default property values if missing

RooHoo33:
I found an bug recently in a Micronaut application I was writing where I was using the custom Micronaut object mapper, if a Kotlin class is annotated with `@Introspected` and the json payload is missing a primitive field, the class property's default value is ignored and the Jackson default primitive value is used. The Micronaut bean instantiation feature already supports instantiating a Kotlin class that has optional/default properties which this fix uses. If the parameter's constructor argument is not required and there is no value from the json payload in the value buffer, instead of having Jackson determine the default the primitive value, we just use null and have the bean instantiation functionality take over and use the default property value.

This is my first pr to the project and I wasnt sure if I needed to create an issue before submitting a pr. If that is required, I can totally do that! Thanks!

### DIFF
--- a/jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
+++ b/jackson-databind/src/main/java/io/micronaut/jackson/modules/BeanIntrospectionModule.java
@@ -54,6 +54,7 @@ import com.fasterxml.jackson.databind.deser.NullValueProvider;
 import com.fasterxml.jackson.databind.deser.SettableBeanProperty;
 import com.fasterxml.jackson.databind.deser.ValueInstantiator;
 import com.fasterxml.jackson.databind.deser.impl.MethodProperty;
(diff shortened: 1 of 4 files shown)


dstepanov:
@yawkat Did you say that you want to remove the bean introspection support for Jackson?

yawkat:
Yes.

graemerocher:
maybe target 4.10.x it can be considered there

RooHoo33:
Will do, I originally targeted 4.10.x before realizing 5.0.x was suppose to be released soon. Should I close this pr and open anther one for 4.10.x instead?

RooHoo33:
[opened a new pr targeting 4.10.x](https://github.com/micronaut-projects/micronaut-core/pull/12231)

PR #12195, opened 2025-11-18
test: isolate ConstantPropertySourceSpec in its own module

sdelamo:
test `ConstantPropertySourceSpec` passes when run on isolation. It fails when run in a suite. This PR moves the test to its own module. Not ideal but at least we remove the flakiness of the test:

https://ge.micronaut.io/scans/tests?search.relativeStartTime=P90D&search.timeZoneId=Europe%2FMadrid&tests.container=io.micronaut.context.env.ConstantPropertySourceSpec&tests.help=tags


melix (approved):
We could also use a separate source set in the same module, but this is fine for now.

PR #12197, opened 2025-11-18
test: ignore LoomCarrierSpec for MacOS

sdelamo:
LoomCarrierSpec fails on MacOS passes for Linux

https://ge.micronaut.io/scans/tests?search.startTimeMax=1763506799999&search.startTimeMin=1762815600000&search.tags=not:CI&search.timeZoneId=Europe%2FMadrid&tests.container=io.micronaut.http.server.netty.LoomCarrierSpec&tests.test=sticky%20on%20poller%20thread&tests.unstableOnly=false


melix (approved):
We should probably create a ticket because this would need to pass, right?

sdelamo:
> We should probably create a ticket because this would need to pass, right?

It passes in linux (both in ci and in developer machines).

melix:
Yes but it should pass on MacOS too?

PR #12202, opened 2025-11-19
Add support for less buggy multipart decoder

yawkat:
The standard netty multipart decoder is quite buggy. A new API is available at https://github.com/netty-contrib/codec-multipart . That implementation also has compatible drop-in replacement for the old netty APIs.

This PR checks whether the new implementation is available on the classpath, and if so, uses that instead of standard netty. There is also a config option to enable "quirks" that restore various bugs of the old implementation for compatibility.

This is experimental and undocumented because the new API is in alpha. I've added it to 4.10 to let people resolve long-standing issues with form data parsing.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -82,6 +82,7 @@ managed-kotlin-coroutines = "1.8.1"
 managed-methvin-directory-watcher = "0.19.1"
 managed-netty = "4.2.7.Final"
 managed-netty-tcnative = "2.0.74.Final"
+managed-netty-contrib-multipart = "1.0.0.Alpha1"
 managed-reactive-streams = "1.0.4"
 # This should be kept aligned with https://github.com/micronaut-projects/micronaut-reactor/blob/master/gradle.properties from the BOM
 managed-reactor = "3.7.12"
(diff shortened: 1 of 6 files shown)


graemerocher (changes requested):
can you add documentation for the quirks thing and what it means and how a user should use it?

yawkat:
Quirks are described in the readme here: https://github.com/netty-contrib/codec-multipart/?tab=readme-ov-file#multipart-core

I don't really want to document this yet, I'm adding this support to 4 so people can test it. The quirks are a fallback option that allows for partial compatibility.

graemerocher:
in the linked docs it says "Each quirk has a description of what behavior it restores." 

Where are these quirks documented?

yawkat:
In the DecoderQuirk enum: https://github.com/netty-contrib/codec-multipart/blob/1.0/multipart-core/src/main/java/io/netty/contrib/multipart/DecoderQuirk.java

I can't use that enum for the config property unfortunately, because it won't work for a compileOnly dependency. So I have to convert from String to the enum manually.

graemerocher (approved):

PR #12219, opened 2025-11-24
build: accept version and library regiression from netty-contrib-mult…

sdelamo:
…ipart


sdelamo:
@yawkat not sure why this is now necessary

melix (approved):

Issue #12211, opened 2025-11-21
Remove ByteBuffer


yawkat:
### Issue description

The ReadBuffer API is a suitable replacement. We should remove the ByteBuffer API in 5.0.

graemerocher:
Triage: valid planned breaking-change task; no runtime reproducer is applicable. The current codebase still exposes and consumes the ByteBuffer API in HTTP client, body, codec, and response paths. Suggested approach: inventory public API references, migrate implementation and integrations to ReadBuffer with compatibility adapters where needed, deprecate before removal, and add binary/source compatibility and pooled-buffer lifecycle tests. No duplicate or open PR was found.

PR #12236, opened 2025-12-02
Prepare JsonMapper for Jackson 3 and ReadBuffer

yawkat:
- Introduce a new ReadBuffer.useFastHeapBuffer method to directly access a buffer's backing array, used for fast deserialization.
- Migrate ReadBuffer.transferTo to useFastHeapBuffer
- Add a readValue(ReadBuffer) method to replace the ByteBuffer one
- Implement this method using useFastHeapBuffer
- Add a method to JacksonCoreParserFactory that takes an ObjectReadContext (Jackson 3)

In the short term, micronaut-serialization can use the new JacksonCoreParserFactory method. In the long term, callers should migrate to the new ReadBuffer readValue which makes JacksonCoreParserFactory obsolete.

### DIFF
--- a/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBuffer.java
+++ b/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBuffer.java
@@ -16,16 +16,18 @@
 package io.micronaut.buffer.netty;
 
 import io.micronaut.core.annotation.Internal;
-import org.jspecify.annotations.NonNull;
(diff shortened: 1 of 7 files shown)


sdelamo (approved):

sdelamo:
@yawkat address the sonar feedback regarding of `forRemoval=true` and `since` in the deprecated annotation.

yawkat:
I think forRemoval is not appropriate yet. If #12211 makes it into 5.0.0 those elements will be deleted immediately anyway, if not they will stick around until #12211 is done.

PR #12240, opened 2025-12-03
Fix property cache issue when environment is refreshed

### LINKED ISSUE #12190: Property cache issue when environment is refreshed
### Expected Behavior

A property should contain a new value after an environment is refreshed.

### Actual Behaviour

When the `refreshAndDiff` method (`DefaultEnvironment` class) is called, the `stop` method removes all property values from `catalog` (but not from `rawCatalog`) and then clears caches. If the `getProperty` method (`PropertySourcePropertyResolver` class) is executed by another thread after the `stop` method and before new properties are added to the catalog, the `getProperty` method will not find anything in the property cache and `catalog`, but will find the old value in the `rawCatalog` and add it to the cache, so the cache will contain old value. That causes all future executions of getProperty method to return old value from the cache instead of new value from the catalog.

### Steps To Reproduce

This test reproduces the issue:
https://github.com/micronaut-projects/micronaut-core/commit/c110899a80f22ec92bdd727b7d10086121b35097

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.x

iamsb97:
Closes #12190. 

Change summary:
1. Reset `nonGenerated` and `rawCatalog` along with `catalog` to keep all three structures in sync.
2. Skip cache writes while a refresh is in progress, using a volatile `refreshing` flag. This prevents stale values from being written into the cache after refresh starts.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -99,6 +99,8 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
     private final Map<ConversionCacheKey, Object> resolvedValueCache = new ConcurrentHashMap<>(20);
     private final EnvironmentProperties environmentProperties = EnvironmentProperties.fork(CURRENT_ENV);
 
+    private volatile boolean refreshing = false;
+
     /**
      * Creates a new, initially empty, {@link PropertySourcePropertyResolver} for the given {@link ConversionService}.
      *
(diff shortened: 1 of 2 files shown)


dstepanov:
Considering you are reseting different catalogues can you validate that the values are cleared

iamsb97:
I have added the requested test. I also verified that, without the changes to `reset()`, the assertion at line 150 in the Groovy test fails, since `rawCatalog` was not being cleared earlier. I couldn't find any usage of the `nonGenerated` catalog in any code path, so I’m not sure it can be verified through a test.

PR #12246, opened 2025-12-05
Refactor multipart

yawkat:
* Introduce low-level, server-independent form API
* Refactor high-level form API for better resource management and usability
* Implement high-level form API binders in a less server-dependent way, may become to server-independent in the future
* Implement low-level form API using new netty-contrib multipart codec


yawkat:
@graemerocher still some bugs left to fix, but the API is ready. check out the modified multipart classes and the docs on how they are bound.

yawkat:
@graemerocher PTAL at the API

core/src/main/java/io/micronaut/core/io/buffer/LeakTracker.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+}
+
+class LeakTrackerFactoryHolder {
+    static boolean nettyAvailable = true;
+}
+
+@SuppressWarnings({"ReturnValueIgnored", "ResultOfMethodCallIgnored"})
+class NettyLeakTrackerFactory<T> implements LeakTracker.Factory<T> {
```
graemerocher: should be final

core/src/main/java/io/micronaut/core/io/file/TemporaryFileResource.java
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Closeable resource class representing a temporary file. When this resource is closed, the file
+ * is deleted.
+ *
+ * @since 5.0.0
+ * @author Jonas Konrad
+ */
+@Internal
+public final class TemporaryFileResource implements Closeable {
```
graemerocher, line 38: does this need to be `public`?

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
```diff
@@ -710,6 +675,84 @@ private static ByteBuffer<?> toByteBuffer(AvailableByteArrayBody immediateByteBo
         return immediateByteBody.peek().toByteBuffer();
     }
 
+    @Override
+    public @NonNull Publisher<RawFormField> getRawFormFields() throws IllegalStateException {
+        return getRawFormFields(byteBody());
+    }
+
+    public @NonNull Flux<RawFormField> getRawFormFields(ByteBody byteBody) {
```
graemerocher, line 688: if this is public needs documentation

http/src/main/java/io/micronaut/http/form/FormCapableHttpRequest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.reactivestreams.Publisher;
+
+/**
+ * A request that is <i>capable</i> of parsing form data, though the request does not necessarily
+ * have to contain form data.
+ *
+ * @param <B> The body type
+ */
+public interface FormCapableHttpRequest<B> extends ServerHttpRequest<B>, LifecycleHttpRequest<B> {
+    /**
```
graemerocher, line 32: would add `@Experimental` here

src/main/docs/guide/httpServer/form/binding.adoc
```diff
@@ -0,0 +1,86 @@
+For a typical web application, the low-level form API is unnecessary. Instead, you will _bind_ individual form fields to controller parameters of different types.
+
+When you bind at least one form field to a controller parameter, the framework will take over the FormCapableHttpRequest, making the body only available through controller argument binding.
+
+There are various buffered and streaming types you can bind form fields to.
```
graemerocher, line 5: this is a bit technical sounding as a description. The user doesn't care about the implementation details, what they want to know is about the user facing API. Document the parameter binding approach and then mention later if necessary that there is a low-level form API if that is something we want to support

src/main/docs/guide/httpServer/form/binding.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+=== Standard Types
+
+The simplest binding is a `String`. The framework will wait for the form field to arrive, and then call your controller with the full value of the form field.
+
+snippet::io.micronaut.docs.server.form.FormController[tags="imports,class,String,endclass", title="String controller example"]
+
+It is also possible to bind to other simple types, such as `int` or `byte[]`.
+
+Note that if there are multiple form fields of the same name, only the first one is bound. Any later fields are discarded.
```
graemerocher, line 15: What if if I have a `List<String>` and multiple of the same name? Would it populate the list?

src/main/docs/guide/httpServer/form/binding.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+=== CompletedFileUpload
+
+A api:io.micronaut.http.multipart.CompletedFileUpload[] is similar to a `byte[]` parameter, except that the upload can be buffered to disk instead of memory. This allows you to receive even large files without consuming a lot of memory. This behavior needs to be configured, though.
+
+// TODO: example with `mixed` config properties
```
graemerocher: address this TODO

src/main/docs/guide/httpServer/form/binding.adoc
```diff
@@ -0,0 +1,86 @@
...
+NOTE: The uploaded file is automatically deleted when the controller terminates, so if you want to keep it around, you need to move or copy it elsewhere beforehand. You can also call `CompletedFileUpload.moveResource()` to get a new `CompletedFileUpload` instance that sticks around after the controller lifetime.
+
+snippet::io.micronaut.docs.server.form.FormController[tags="imports,class,CompletedFileUpload,endclass", title="CompletedFileUpload controller example"]
+
+<1> `transferTo` may block, so we need to run this controller on an IO thread
+<2> Transfer the upload to the temp file. If this upload is already on disk (depending on configuration), this is a simple move operation. If the upload is still in memory, this will write the data to disk
+
+=== StreamingFileUpload
+
+A api:io.micronaut.http.multipart.StreamingFileUpload[] _lazily_ streams a form field. Your controller can be called before the full content is received. Similar to `RawFormField`, you must consume the input data so that the upload does not stall.
```
graemerocher, line 42: this is the first time `RawFormField` is mentioned and yet it is referred to here?

src/main/docs/guide/httpServer/form/binding.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+The advantage of `StreamingFileUpload` over `CompletedFileUpload` is that you do not need any disk space, and that you can start streaming the data immediately to e.g. e.g. forward it to a backend service. If the backend is slow to receive the uploaded data, this backpressure will also apply to the browser upload, and that upload will slow down – there is no (significant) buffering in the Micronaut service, unlike with `CompletedFileUpload`.
```
graemerocher, line 44: can these types be used in filter arguments and for example gateway / proxy use cases?

graemerocher (changes requested):

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpRequest.java
yawkat: NettyHttpRequest is `@Internal`

src/main/docs/guide/httpServer/form/binding.adoc
yawkat: Not at the moment. It would require more rewriting of the argument binder structure

src/main/docs/guide/httpServer/form/binding.adoc
yawkat: RawFormField is described in formCapable… Not sure what the order should be

src/main/docs/guide/httpServer/form/binding.adoc
yawkat: No, it shouldn't. The reason for this behavior is that usually there is only one instance of a form field, and when we've seen that instance, we want to "finish" binding the parameter. If there is a possibility of future instances being bound, we have to wait until the end of the form body to be certain. That can interfere with streaming parameters like StreamingFileUpload.

graemerocher: right, I understand that, but note that in the Servlet world this "works" and you can do `String[] lines = request.getParameterValues("line");` where `line` is multiple `<input>` of the same name. The fact that this doesn't work would be surprising behaviour.

yawkat: Fair enough, but I think it's out of scope for this PR to change this behavior.

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
```diff
@@ -157,44 +162,13 @@ Optional<T> transform(NettyHttpRequest<?> nhr, ArgumentConversionContext<T> cont
...
-            boolean allFormData = true;
-            for (Object object : data) {
-                if (!(object instanceof MicronautHttpData<?>)) {
-                    allFormData = false;
-                    break;
-                }
+        if (reader == null && nhr.hasFormBody()) {
+            Map<String, List<CloseableByteBody>> bodies = new LinkedHashMap<>();
+            for (RawFormField rff : toListNow(nhr.getRawFormFields(imm))) {
+                bodies.computeIfAbsent(rff.metadata().name(), k -> new ArrayList<>(1)).add(rff.byteBody());
```
dstepanov, line 170: Can we receive a form part with the same name?

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
```diff
@@ -210,12 +184,41 @@ Optional<T> transform(NettyHttpRequest<?> nhr, ArgumentConversionContext<T> cont
...
+                error = t;
+                complete = true;
+            }
+
+            @Override
+            public void onComplete() {
+                complete = true;
+            }
+        };
+        flux.subscribe(sub);
```
dstepanov, line 217: Should this wait / be reactive somehow?

http/src/main/java/io/micronaut/http/multipart/RawFormField.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.Closeable;
+
+/**
+ * A raw, streamed form field. Note that processing of other form fields may be stalled until you
+ * consume the {@link #byteBody}!
+ *
+ * @param metadata The field metadata provided by the user
+ * @param byteBody The field bytes
+ */
+public record RawFormField(FormFieldMetadata metadata,
```
dstepanov: Not sure I like records in the public api

http/src/main/java/io/micronaut/http/multipart/PartData.java
```diff
@@ -25,40 +27,51 @@
 /**
  * Represents a chunk of data belonging to a part of a multipart request.
  *
+ * @param fieldMetadata The field metadata (name, file name, etc.)
+ * @param readBuffer The buffered part data
  * @author James Kleeh
  * @since 1.0
  */
-public interface PartData {
+public record PartData(FormFieldMetadata fieldMetadata, ReadBuffer readBuffer) implements Closeable {
```
dstepanov: Records in public API can be very limiting what can be done in the future

dstepanov (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
yawkat: Yes, e.g. when selecting multiple files in a file upload, or having multiple inputs with the same name

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyBodyAnnotationBinder.java
yawkat: This works because the body in `transform` is an `AvailableByteBody`, so we don't need to wait for input. There is also some special logic in FormDemuxer to make this work without reactive

http/src/main/java/io/micronaut/http/multipart/PartData.java
yawkat: I don't have a strong opinion on this, WDYT @graemerocher

graemerocher (approved):

sdelamo (commented):
this pull-request contains breaking changes to `StreamingFileUpload` but it doesn't contains information in `breaks.adoc`about how to deal with those.

PR #12258, opened 2025-12-09
Refactor: Deprecate TaskExecutors.BLOCKING (fixes #12217)

### LINKED ISSUE #12217: Remove or deprecate TaskExecutors.BLOCKING
### Issue description

TaskExecutor.BLOCKING is not required on JDK 21+ since virtual threads are always available. Just use TaskExecutors.VIRTUAL.

vijaykriishna:
Fixes: #12217

### DIFF
--- a/buildSrc/gradle.properties
+++ b/buildSrc/gradle.properties
@@ -1,3 +1,3 @@
 #
-#Thu Nov 20 11:29:35 CET 2025
-micronaut-build-version=8.0.0-M9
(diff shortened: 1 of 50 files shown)


sdelamo (commented):
I think `BLOCKING` is more idiomatic. The usage of virtual threads behind the scenes is an implementation detail. We support many server runtimes not sure if virtual threads will be used in all of them.

PR #12270, opened 2025-12-15
Add `AbstractTypeElementTest` that can be used without Spock + some inject-java tests converted to Java

dstepanov:
AI did conversion to Java. Git has some problems displaying it as a change


graemerocher:
I am confused as to why `AbstractTypeElementTest` extends from Spock specification?

dstepanov:
forgot to remove

PR #12276, opened 2025-12-16
Fix context beans are not singleton in some cases

dstepanov:


core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionVisitor.java
```diff
@@ -125,11 +125,6 @@ void visitDefaultConstructor(
      */
     boolean isInterface();
 
-    /**
```
dstepanov, line 128: Unused

graemerocher (approved):

PR #12277, opened 2025-12-16
Add bean configurations predicate support

dstepanov:
For serde static context we also need to filter out bean configurations


alvarosanchez (approved):

alvarosanchez:
@dstepanov this still fails. See https://github.com/micronaut-projects/micronaut-control-panel/pull/430

PR #12280, opened 2025-12-17
Reproduce missing type annotation from compiled class

dstepanov:
Reproducing a Javac bug that not able to read type annotations of a compiled class.
The bug is fixed in Java 22 https://bugs.openjdk.org/browse/JDK-8323093


graemerocher:
can you add a description to the PR that describes what this is for?

PR #12301, opened 2025-12-30
Issue #8411 | Retry Exhausted Callback

stevenlmcgraw:
Issue #8411

DRAFT
> Modeled after `RecoveryInterceptor` and `Recoverable`/`Fallback` annotations.

Extend `micronaut-retry` system with callback invoked on retries exhausted:
- Add `RetryExhaustedActionable.java` annotation
- Add `RetryExhaustedCallback.java`
- Add `RetryExhaustedActionInterceptor.java`
- Add `RetryExhaustedCallbackException.java`

### DIFF
--- a/buildSrc/gradle.properties
+++ b/buildSrc/gradle.properties
@@ -1,3 +1,3 @@
 #
-#Thu Nov 20 11:29:35 CET 2025
-micronaut-build-version=8.0.0-M9
+#Mon Dec 29 23:28:16 CST 2025
+micronaut-build-version=8.0.0-M13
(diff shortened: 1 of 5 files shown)


stevenlmcgraw:
I was poking around old issues with the `info: good first issue` label and found #8411, it seems like something that could be useful in some scenarios. 

[Based on this comment](https://github.com/micronaut-projects/micronaut-core/pull/9183#issuecomment-1613175346) it seemed like something along these lines is what is preferred.

If there's any interest in pursuing #8411, and this seems like the right way to do it, let me know and I can add tests and what not (or changes of course).

PR #12316, opened 2026-01-09
fix: semgrep-unsafe-reflection

sumeetkulkarni11:
**Pull Request — Semgrep Rule Fix**

- **Rule ID:** unsafe-reflection
- **Rule Message:** If an attacker can supply values that the application then uses to determine which class to instantiate or which method to invoke, the potential exists for the attacker to create control flow paths through the application that were not intended by the application developers. This attack vector may allow the attacker to bypass authentication or access control checks or otherwise cause the application to behave in an unexpected manner.
- **File Path:** /tools/scanResult/unzipped-2378829019/core/src/main/java/io/micronaut/core/reflect/ClassUtils.java
- **Line:** 311


graemerocher:
please don't submit PRs without tests, they won't be accepted. Thanks.

sumeetkulkarni11:
Noted! Will take care and add tests!

PR #12321, opened 2026-01-09
Return error on QueryValue conversion failures

### LINKED ISSUE #12238: QueryValue conversion errors are discarded since 4.10.0
### Expected Behavior

Providing an inconvertible value to a `@QueryValue` results in an error message showing the conversion error, e.g. attempting to convert a string to an integer shows the following error message: "Failed to convert argument [count] for value [a] due to: For input string: \"a\"".

### Actual Behaviour

Providing an inconvertible value to a `@QueryValue` results in an error message indicating that the input is missing: "Required QueryValue [count] not specified".

### Steps To Reproduce

MCVE:
```java
@Controller
public class TestController {

    @Get
    public String helloWorld(@QueryValue(value = "count") Integer count) {
        return "Hello World";
    }
}
```

Hitting this endpoint with `?count=a` results in the error message "Failed to convert argument [count] for value [a] due to: For input string: \"a\"" in Micronaut 4.9.4, but upgrading Micronaut to the 4.10 minor version results in the error usually shown for missing values: "Required QueryValue [count] not specified".

This change was caused by https://github.com/micronaut-projects/micronaut-core/pull/11787, which discards the result of `bindSimpleResult` [here](https://github.com/micronaut-projects/micronaut-core/pull/11787/files#diff-a1c471ddc267ada68aa722685e4003e13625a2b223a3557827047893980388d7R100-R104) if a value was not found, even though it might contain a conversion error.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.3

iamsb97:
Fixes #12238.

Change summary:
1. Return `BindingResult` with error on conversion failures instead of `BindingResult.unsatisfied()`.
2. Move error cases into separate tests and verify error messages.
3. Add new tests.


iamsb97:
@graemerocher : Please review this PR when you get a chance. Thanks!

graemerocher:
@iamsb97 can you check if this simpler fix passes your tests https://github.com/micronaut-projects/micronaut-core/pull/12359

iamsb97:
@graemerocher: It doesn't raise conversion error for POJO binding. In particular, for `void "test conversion error for object binding"()` test, it returns `Required QueryValue [book] not specified`.

graemerocher (approved):

binoysankar:
We are facing this error for "@Nullable @QueryValue(value = X_ABC_ID_QUERY_PARAM) String abc". Earlier we used to null but now it returns "".
I tried the latest snapshot version "4.10.8-SNAPSHOT" also. Would like to know which micronaut version I can see the changes?

mikehearn:
@binoysankar I faced the same problem. In my case it looked like a regression in Micronaut because it was triggered by an upgrade, but the actual bug was in the project code which had a line `@SerdeImport(String.class)`. This caused String to be registered as introspectable, which in 4.10 then causes it to be default instantiated if missing. String shouldn't be a bean but this mistake isn't caught by the Micronaut core engine at this time. I'll submit a PR to fix that soon.

You may want to check that String isn't being registered accidentally for introspection.

binoysankar:
> 

Thank you very much. Yes I was using serdeimport on Application class. Removed these and the build was successful.

```
@SerdeImport(String.class)
@SerdeImport(Integer.class)
```

PR #12350, opened 2026-01-15
Fix duplicate content-type header bug, add test

wfhartford-wordly:
Micronaut adds a second `Content-Type` header to an HttpResponse when the result of the controller function is a `HttpResponse<Flux<String>>` (or other nested data type), and the controller explicitly sets the content type on the `HttpResponse`. This can be demonstrated with the following simple controller:
```kotlin
@Controller
class VariableContentTypeController {
	@Get("/")
	fun index(): HttpResponse<Flux<String>> {
		return HttpResponse.ok(Flux.just("Hello World!")).contentType("text/plain")
	}
}
```
See https://github.com/wfhartford-wordly/micronaut-duplicate-content-type-headers for a full project demonstrating the bug.

Calling this function from curl shows that two content type headers are present in the response.
```
$ curl localhost:8080 -v
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Type: text/plain
< Content-Type: text/plain
< date: Thu, 15 Jan 2026 23:45:04 GMT
< transfer-encoding: chunked
< 
* Connection #0 to host localhost left intact
Hello World!
```

This PR fixes the bug by changing the `RouteExecutor` to call the `contentType` function which _sets_ the header rather than the `header` function which _adds_ it. I've also included a unit test which verifies the fix.

I first encountered this bug in Micronaut version 4.3.8 and have confirmed that it is present in version 4.10.7 and on the 5.0.x branch.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -744,7 +744,7 @@ public final class RouteExecutor {
         ).contextWrite(cv -> ReactorPropagation.addPropagatedContext(cv, propagatedContext).put(ServerRequestContext.KEY, request));
 
         return Mono.<MutableHttpResponse<?>>just(response
-            .header(HttpHeaders.CONTENT_TYPE, mediaType)
+            .contentType(mediaType)
             .body(ReactivePropagation.propagate(propagatedContext, bodyPublisher)))
             .contextWrite(context -> ReactorPropagation.addPropagatedContext(context, propagatedContext).put(ServerRequestContext.KEY, request));
     }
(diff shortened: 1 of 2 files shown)


graemerocher:
do you wish to target the 4.10.x branch or keep this in 5.0.x?

wfhartford-wordly:
> do you wish to target the 4.10.x branch or keep this in 5.0.x?

I'd love to see this fixed in 4.10.x. Would you like me to create another PR targeting that branch?

graemerocher:
@wfhartford-wordly a PR targeting 4.10.x would make sense. Thanks

wfhartford-wordly:
#12376 Is the same change targeting the 4.10.x branch.

yawkat:
I think we will just merge this up from 4.10.x

PR #12357, opened 2026-01-19
Support Authentication as a valid WebSocket handler argument

zManu3k:
Updates WebSocket validation to allow Authentication parameters in @OnOpen, @OnClose, @OnError, and @OnMessage handlers.

This enables WebSocket handlers to receive Authentication directly, aligning WebSocket argument binding with HTTP controllers.

Please note: this code won't work alone, you should accept the commit on micronaut-security too to make it work: [link](https://github.com/micronaut-projects/micronaut-security/pull/2116).


graemerocher:
do yo want to target 5.x or 4.10.x for this change?

zManu3k:
Hi, sorry for the late response. I migrated to 5.x and targeted that version for this change, as it’s not a bug fix. It should work on 4.10.x as well, but I haven’t tested it

Issue #12262, opened 2025-12-11
introspection fails when class implements same field from multiple interfaces


vekonypeter:
### Expected Behavior

When I have:
- an introspected Kotlin class implementing multiple interfaces
- both interfaces declare a field with the same name and type
- the interfaces does not share a parent where they would inherit the common field

I expect that introspection still works without introducing additional artificial interfaces.

### Actual Behaviour

The following minimal class structure:
```
interface A {
    val foo: String
}

interface B {
    val foo: String
}

@Introspected
class C(
    override val foo: String
) : A, B
```

causes this exception at application startup:

```
Error starting Micronaut server: Failed to load a service: java.lang.ClassFormatError: Duplicate method name "$property$foo$metadata" with signature "()Lio.micronaut.inject.beans.AbstractInitializableBeanIntrospection$BeanPropertyRef;" in class file com/example/$C$Introspection
io.micronaut.core.io.service.SoftServiceLoader$ServiceLoadingException: Failed to load a service: java.lang.ClassFormatError: Duplicate method name "$property$foo$metadata" with signature "()Lio.micronaut.inject.beans.AbstractInitializableBeanIntrospection$BeanPropertyRef;" in class file com/example/$C$Introspection
	at io.micronaut.core.io.service.ServiceScanner$ServiceInstanceLoader.collect(ServiceScanner.java:244)
	at io.micronaut.core.io.service.ServiceScanner$DefaultServiceCollector.collect(ServiceScanner.java:157)
	at io.micronaut.core.io.service.SoftServiceLoader.collectDynamicServices(SoftServiceLoader.java:199)
	at io.micronaut.core.io.service.SoftServiceLoader.collectAll(SoftServiceLoader.java:174)
	at io.micronaut.core.beans.DefaultBeanIntrospector.getIntrospections(DefaultBeanIntrospector.java:117)
	at io.micronaut.core.beans.DefaultBeanIntrospector.findIntrospection(DefaultBeanIntrospector.java:88)
	at io.micronaut.core.reflect.InstantiationUtils.tryInstantiate(InstantiationUtils.java:170)
	at io.micronaut.context.conditions.MatchesCustomCondition.matches(MatchesCustomCondition.java:58)
	at io.micronaut.context.AbstractInitializableBeanDefinitionAndReference.matches(AbstractInitializableBeanDefinitionAndReference.java:111)
[...]

PR #12358, opened 2026-01-19
Fix failed introspection for duplicated properties

davidsonnabend:
Initial fix for #12262.

I’m curious if you think this is the best path forward or if you’d prefer a different implementation.

Please let me know if anything can be improved or needs to be solved differently.


dstepanov:
I think the correct way is to fix it on `AstBeanPropertiesUtils#resolveBeanProperties`. Please also check setting/getting value works.

davidsonnabend:
I think I found the root cause. `AstBeanPropertiesUtils` isn't involved for this specific case.

The problem is that `KSPropertyDeclaration::findOverridee()` always returns the first declared interface. Thus, the check for overridden properties only works for the first interface (`A::foo`). The same property of the second interface (`B::foo`) is considered as a new property. This results in the list of duplicated properties and the described bug.

davidsonnabend:
Is there anything I can do to get this PR merged or at least reviewed?

dstepanov:
I’m a but concerned about the performance, this is a very perf consuming call. Can we make this checks only for some cases, only for interfaces?

davidsonnabend:
I added an additional check to prevent calling `getAllSuperTypes()` if the overridden declaration is not an interface.
Do you think that's enough?

davidsonnabend:
Is there a chance to get this merged in the near future? I'm open for further optimization if required.

davidsonnabend:
@dstepanov / @graemerocher Is there anything I can do to make progress here?

PR #12359, opened 2026-01-19
Returning simple binding errors early if they occur

### LINKED ISSUE #12238: QueryValue conversion errors are discarded since 4.10.0

graemerocher:
## Fix QueryValue conversion errors being discarded since 4.10.0

Fixes #12238

### Changes (rebased on 4.10.x):
- **QueryValueArgumentBinder.java**: Modified `bind()` to preserve conversion errors
- **QueryValueConversionErrorSpec.groovy**: Added test cases validating proper error messages

### Fix Details:
Now checks for conversion errors before discarding binding result:
```java
if (bindSimpleResult.isSatisfied() || !bindSimpleResult.getConversionErrors().isEmpty()) {
    return bindSimpleResult;
}
```

This ensures conversion errors like "Failed to convert argument [count] for value [a] due to: For input string: 'a'" are properly shown instead of the misleading "Required QueryValue [count] not specified" message.

**This PR targets the 4.10.x branch.**

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/QueryValueConversionErrorSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/QueryValueConversionErrorSpec.groovy
@@ -1,0 +1,78 @@
+package io.micronaut.http.server.netty.errors
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.HttpStatus
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.annotation.QueryValue
+import io.micronaut.http.client.BlockingHttpClient
+import io.micronaut.http.client.HttpClient
+import io.micronaut.http.client.exceptions.HttpClientResponseException
+import io.micronaut.runtime.server.EmbeddedServer
+import spock.lang.AutoCleanup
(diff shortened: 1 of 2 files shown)


graemerocher:
Replaced by https://github.com/micronaut-projects/micronaut-core/pull/12321

PR #12371, opened 2026-01-29
Remove deprecated URI template

dstepanov:


graemerocher:
please make sure we document this change and what is the alternative to `UriTemplate`

PR #12383, opened 2026-02-09
Fail fast on duplicate configuration resources (with configurable merge/first-match)

### LINKED ISSUE #11703: application.yml from Libraries Can Override or Be Silently Ignored — No Merging or Warning
### Expected Behavior

In this scenario, where a **Micronaut application defines `application.yaml`** and a **library dependency provides its own `application.yml`**, the expected behavior is:

1. **The application's configuration (`application.yaml`) should always take precedence.**  
   This aligns with developer expectations — the application's own config should never be overridden silently by a transitive library dependency.

2. **Library configuration files should be loaded optionally and with transparency.**  
   If a library provides its own `application.yml`, it may contain useful default values. These should be:
   - Merged *after* the application config (lower precedence)
   - Only if key conflicts don't exist
   - Accompanied by a **log message** indicating their source and loading behavior

3. **When multiple `application.yml` or `.yaml` files are found**, Micronaut should:
   - **Merge** them (with deterministic ordering and documented rules), or
   - Emit a **warning** stating that only one was used, and which one

4. Ideally, Micronaut would support a **configurable loading strategy**, so developers can explicitly opt into:
   - `FIRST_MATCH` (current behavior)
   - `MERGE_ALL`
   - `FAIL_ON_DUPLICATE`
   - Possibly others (e.g., `APP_ONLY`, `APP_OVERRIDES_LIBS`)

This approach ensures application authors retain full control over configuration while allowing library authors to provide sensible defaults **without risk of overriding or being silently ignored**.

### Actual Behaviour

We created a reproducible Gradle multi-module Micronaut project with:

- A main `app` module that defines `application.yaml`
- A `library` module that defines its own `application.yml`

**Expected:** The application's config (`application.yaml`) should be loaded and take precedence.  
**Actual:** The application's config is silently **ignored** — only the library's `application.yml` is used.

There are **no logs, warnings, or exceptions** indicating this override has occurred.

---

#### 🔬 Why This Happens: Deep Dive into Micronaut’s `YamlPropertySourceLoader`

The root cause lies in the `YamlPropertySourceLoader`, which extends `AbstractPropertySourceLoader`. Here’s the key method in `AbstractPropertySourceLoader`:

```java
private Optional<PropertySource> load(ResourceLoader resourceLoader, String fileName, int order) {
    if (this.isEnabled()) {
        Set<String> extensions = this.getExtensions(); // e.g., ["yml", "yaml"]
        Iterator<String> var5 = extensions.iterator();

        while (var5.hasNext()) {
            String ext = var5.next();
            String fileExt = fileName + "." + ext; // Tries application.yml then application.yaml

            Map<String, Object> finalMap = this.loadProperties(resourceLoader, fileName, fileExt);
            if (!finalMap.isEmpty()) {
                return Optional.of(this.createPropertySource(fileName, finalMap, order)); // First match wins
            }
        }
    }

    return Optional.empty();
}

### Steps To Reproduce


> **Prerequisite:** Ensure Java 21 is installed and configured as the active JDK.

1. **Clone the Repository:**
   ```bash
   git clone <repository-url>
   cd <repository-directory>
   ```

2. **Build the Project:**
   ```bash
   ./gradlew build
   ```

3. **Run the Application:**
   ```bash
   ./gradlew :app:run
   ```

4. **Observe the Output:**

   Notice that the configuration value from  
   `app/src/main/resources/application.yaml` is **not** printed.  
   Instead, the value from  
   `utilities/src/main/resources/application.yml` is used.

5. **Modify the Project:**

   Open `app/build.gradle` and comment out the following line:
   ```groovy
   implementation(project(":utilities"))
   ```

6. **Rebuild and Rerun the Application:**
   ```bash
   ./gradlew build
   ./gradlew :app:run
   ```

7. **Observe the Output Again:**

   Now, the configuration value from  
   `app/src/main/resources/application.yaml` **is** printed as expected.

---

These steps demonstrate the issue where the `application.yml` from the `utilities` module overrides the `application.yaml` from the `app` module.

### Environment Information

MacOS: 15.3.2 (24D81)
Java 21
Built tool: gradle

### Example Application

https://github.com/tmellanxt/micronaut-yaml-bug

### Version

4.7.6

### 🙋‍♂️ Contributor Statement

I am happy to contribute a fix for this issue and would appreciate guidance from the Micronaut team on the preferred direction. Specifically:

- I am in favor of introducing a **configurable loading strategy**, such as:
  - `micronaut.config.load-strategy=FIRST_MATCH | MERGE_ALL | FAIL_ON_DUPLICATE`, etc.
- I’d be glad to prototype a solution, update documentation, and write tests — but would like confirmation on which strategy or direction aligns with project goals before submitting a PR.

Additionally, it’s worth noting that this issue affects not just applications being overridden by libraries, but also cases where **multiple libraries** provide their own `application.yml`. In those cases, **only one is loaded**, and the rest are silently ignored, which can also break expected behavior unless libraries explicitly configure environment profiles or load configs differently.

### PR DESCRIPTION
**Pull request by graemerocher**
Fixes #11703

AI was used in the production for this PR.

### Summary
- Introduces a configurable strategy for handling duplicate config files on the classpath (e.g. `application.yml`, `application.properties`).
- Changes the default behavior to **fail fast** with a clear `ConfigurationException` listing all conflicting resource locations.
- Adds options to keep prior behavior (**FIRST_MATCH**) or **MERGE_ALL** duplicate resources (with optional `mergeOrder` by jar name patterns).

### Key changes
- New API: `ConfigurationLoadStrategy` / `ConfigurationLoadStrategyType`
- New builder hook: `ApplicationContextBuilder#configurationLoadingStrategy(...)` (propagated through `Micronaut`)
- Environment loading updates to detect duplicates and apply the selected strategy
- Tests covering default fail-fast, FIRST_MATCH, MERGE_ALL (+ `mergeOrder`), and env-specific resources
- Docs updates + breaking change note (v5) using snippet-backed examples

### Migration / behavior change
If you relied on “first match wins” when duplicates exist, configure:
```java
ApplicationContext.builder()
    .configurationLoadingStrategy(s -> s.type(ConfigurationLoadStrategyType.FIRST_MATCH))
    .start();
```

- **Expose configuration load strategy on builders**
- **Fix constant property source loader initialization**
- **Handle duplicate configuration resources**
- **Add tests for configuration loading strategy**
- **Document duplicate configuration resource handling**


graemerocher:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/micronaut-projects/micronaut-core/pull/12383#pullrequestreview-3772758009)

graemerocher:
the test failure is unrelated

dstepanov:
I would prefer for this to be implemented in the same way as Spring has it with `classpath*:`

classpath: - Locates only the first resource found at the specified path within the classpath.

classpath*: - Locates all resources that match the given name across all classpath entries (directories and JARs).

And this can be used with `.overrideConfigLocations("classpath*:/", "file:config/")`. 
The logic should be put somewhere in `DefaultClassPathResourceLoader` with something like `Stream<URL> getResources(String name, boolean includeAll);`

graemerocher:
I don't understand, in this case the user doesn't specify where or how config is loaded. If there are duplicate `application.properties` it is unknown which is loaded which causes weird behaviour like having to search through your jar files for the duplicate. We need a solution to this in 5.x that is easier to debug

inject/src/main/java/io/micronaut/context/env/ConfigurationLoadStrategy.java
```diff
@@ -0,0 +1,109 @@
...
+            // Always create a defensive copy to prevent external mutation
+            mergeOrder = new ArrayList<>(mergeOrder);
+        }
+
+        if (!mergeOrder.isEmpty() && type != ConfigurationLoadStrategyType.MERGE_ALL) {
+            throw new ConfigurationException("mergeOrder is only supported when configuration loading strategy type is MERGE_ALL");
+        }
+
+        mergeOrder = Collections.unmodifiableList(mergeOrder);
+    }
```
sdelamo: I don't think this constructor is used at all. 

```suggestion
```

sdelamo (changes requested):
I think the feature is useful but the PR does not contain tests for the new logic in `DefaultEnvironment::loadPropertySourceFromAbstractLoader`

inject/src/main/java/io/micronaut/context/env/ConfigurationLoadStrategy.java
graemerocher: It is used, it is the record initialiser and it is always used

inject/src/main/java/io/micronaut/context/env/ConfigurationLoadStrategy.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Configuration resource loading strategy.
+ *
+ * @param type             The strategy type. Defaults to {@link ConfigurationLoadStrategyType#FAIL_ON_DUPLICATE}.
+ * @param warnOnDuplicates Whether to warn when duplicates are found. Applies only to {@link ConfigurationLoadStrategyType#FIRST_MATCH}.
+ * @param mergeOrder       Artifact name regex patterns used to order resources before merging. Applies only to {@link ConfigurationLoadStrategyType#MERGE_ALL}.
+ * @since 5.0.0
+ */
+@NullMarked
+public record ConfigurationLoadStrategy(
```
dstepanov: Put `@NullMarked` to packages if needed. Not sure about records in API, might be problematic for the future

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -525,17 +529,240 @@ private Collection<PropertySourceLoader> evaluatePropertySourceLoaders() {
...
+            .thenComparingInt(IndexedUrl::originalIndex));
+
+        List<URL> ordered = new ArrayList<>(urls.size());
+        for (IndexedUrl idx : indexed) {
+            ordered.add(idx.url());
+        }
+        return ordered;
+    }
+
+    private static String artifactName(URL url) {
```
dstepanov, line 737: This logic should be somewhere in `ClassPathResourceLoader`

inject/src/main/java/io/micronaut/context/ApplicationContextBuilder.java
```diff
@@ -109,6 +111,17 @@ default ApplicationContextBuilder enableDefaultPropertySources(boolean areEnable
...
     }
 
+    /**
+     * Configure how Micronaut loads configuration resources when duplicates exist on the classpath.
+     *
+     * @param builderConsumer The strategy builder customizer
+     * @return This builder
+     * @since 5.0.0
+     */
+    default ApplicationContextBuilder configurationLoadingStrategy(Consumer<ConfigurationLoadStrategy.Builder> builderConsumer) {
```
dstepanov: Why Consumer?

dstepanov (changes requested):

sdelamo (approved):

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
```diff
@@ -525,17 +529,228 @@ private Collection<PropertySourceLoader> evaluatePropertySourceLoaders() {
...
+                                                                  ConfigurationLoadStrategy strategy) {
+        if (propertySourceLoader instanceof AbstractPropertySourceLoader abstractPropertySourceLoader) {
+            String envName = name + "-" + activeEnvironment.getName();
+            int order = abstractPropertySourceLoader.getOrder() + 1 + activeEnvironment.getPriority();
+            return loadPropertySourceFromAbstractLoader(envName, abstractPropertySourceLoader, resourceLoader, order, strategy);
+        }
+        return propertySourceLoader.loadEnv(name, resourceLoader, activeEnvironment);
+    }
+
+    private Optional<PropertySource> loadPropertySourceFromAbstractLoader(String fileName,
```
dstepanov, line 574: Can you move all of it into a new method to ClassPathResourceLoader with an option to merge of fail? I don't think it should be included here

dstepanov (changes requested):

inject/src/main/java/io/micronaut/context/env/DefaultEnvironment.java
graemerocher: I don't understand, `ClassPathResourceLoader` is in the `core` module and has nothing to do with environment configuration. Why would it go there?

dstepanov: Because the main issue is about the classpath files being loaded and the strategies resolving the conflicts etc. I don't think the DefaultEnvironment should be responsible for that logic. Considering there might be different consumers of the classpath files that might also want to react to similar problems. One of them being reported is actually the database flyway migrations being in multiple jars. Not sure if it's mostly about the properties or if some of the frameworks what to load multiple different files/configurations of the same name from the classpath.

graemerocher: I still don't understand what you want here. The core module knows nothing about how files can be merged, only DefaultEnvironment knows how to merge configuration files. At best we could move the logic to find resources and fail if there are duplicates, but all of the merging logic needs to stay in DefaultEnvironment. 

Is it you want some strategy interface that exists in ClasspathResourceLoader that has callbacks to react to different events like duplicate, merge, whatever? 

Perhaps you can take over this PR and implement what you had in mind.

PR #12385, opened 2026-02-10
fix: MediaType can decode quoted charset

### LINKED ISSUE #12247: HttpClient doesn't handle Content-Type charset values with quotation marks.
### Expected Behavior

As per [HTTP Semantics](https://httpwg.org/specs/rfc9110.html#media.type), `text/html; charset="utf-8"` is a valid specification of the charset value and the Micronaut `HttpClient` should be able to decode a response with such a body.

### Actual Behaviour

The HttpClient fails to decode a response with a `Content-Type` header where the charset is specified in quotation marks. For example, `content-type: text/html; charset="utf-8"`.

### Steps To Reproduce

Something like:

```
@MicronautTest
public class Test {

    @Inject
    @Client("/")
    HttpClient httpClient;

    @Test
    void testHelloWorldEndpoint() {
        String response = httpClient.toBlocking().retrieve("/hello");
    }

    @Controller("/")
    static class HelloWorldController {

        @Get("/hello")
        @Produces("text/plain; charset=\"utf-8\"")
        public String hello() {
            return "Hello, World!";
        }
    }

}
```

### Environment Information

- Operating System: Linux 
- JDK Version: OpenJDK-21

### Example Application

_No response_

### Version

4.10.3

valtercorreia:
This PR fixes #12247.

Opted for a simple removal of the quotes in the MediaType::getCharset() method as it looked cleared than doing it on the constructor.
Let me know what you think.

### DIFF
--- a/benchmarks/src/jmh/java/io/micronaut/http/MediaTypeParameterParserBenchmark.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/MediaTypeParameterParserBenchmark.java
@@ -1,0 +1,235 @@
+package io.micronaut.http;
+
+import io.micronaut.core.util.StringUtils;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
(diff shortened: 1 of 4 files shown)


yawkat:
The allowed formats for content types are much more complicated than this

yawkat (approved):
i expanded the parsing using the ParmParser code from netty-codec-multipart

graemerocher (approved):

graemerocher:
need to address the CI failures

yawkat:
Thanks!

Issue #12364, opened 2026-01-23
Relative redirects don't work in HttpClient


arnost:
### Expected Behavior

If HttpClient receives a redirect to a relative URL, it should apply that relative URL to the original request URL in order to get an absolute URL.

### Actual Behaviour

If the server returns a relative URL in the redirect, HttpClient treats it as an absolute URL.

### Steps To Reproduce

1. Clone:  https://github.com/arnost/example.micronaut.relative.redirect.fails
2. run ./mvnw mn:run
3. run curl -L  http://localhost:8080/redirect/byClientRelative
4. see log lines

DEBUG i.m.h.client.netty.DefaultHttpClient - Sending HTTP GET to http://localhost:8080/redirect/relative
DEBUG i.m.h.client.netty.DefaultHttpClient - Sending HTTP GET to http://localhost:8080/./target

### Environment Information

- Linux
- OpenJdk 21

### Example Application

https://github.com/arnost/example.micronaut.relative.redirect.fails

### Version

4.10.7 (4.8.0)

Tomucha:
Fixed in: https://github.com/micronaut-projects/micronaut-core/pull/12395

graemerocher:
Triage: this valid HTTP client bug is resolved by merged PR #12395, which resolves relative redirect locations against the original request URI and adds coverage for same-directory, parent-directory, root-relative, absolute, and multi-hop redirects. The issue’s external reproducer is sufficient; no additional attachment or Core work is needed. Closing as completed.

PR #12395, opened 2026-02-17
Fix: Relative redirects not resolved against request URI

ArtemLikhitskiy:
Problem
When DefaultHttpClient receives a redirect response (301/302/307/308) with a relative Location header, it should resolve it against the original request URI per [RFC 7231 §7.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.2):

The field value consists of a single URI-reference. When it has the form of a relative reference, the final value is computed by resolving it against the effective request URI.

Root Cause
In DefaultHttpClient.java, the method that processes the redirect Location header calls URI.create(location) directly — which for a relative URI like ./target creates a URI with no scheme/host. When this is then used to construct the next request, the path ./target is sent as-is without normalization.
The fix from 2.x (#4757) was lost during the large Netty HTTP client refactoring that happened between 2.x and 4.x.

java.net.URI.resolve() implements RFC 3986 §5.2 reference resolution exactly, so all edge cases are handled.

Tests
Added RelativeRedirectSpec.groovy in:
http-client/src/test/groovy/io/micronaut/http/client/RelativeRedirectSpec.groovy
Covers:

./target — same-directory relative redirect
../sibling — parent-directory relative redirect
/root-relative — root-relative redirect (was already working, regression guard)
Absolute URL redirect — must still work
Multi-hop chain with mixed relative + absolute redirects

### DIFF
--- a/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
+++ b/http-client/src/main/java/io/micronaut/http/client/netty/DefaultHttpClient.java
@@ -1274,12 +1274,12 @@ public class DefaultHttpClient implements
             if (parentRequest == null || parentRequest.getUri().getHost() == null) {
                 return resolveURI(request, false);
             } else {
-                URI parentURI = parentRequest.getUri();
-                UriBuilder uriBuilder = UriBuilder.of(requestURI)
-                        .scheme(parentURI.getScheme())
-                        .userInfo(parentURI.getUserInfo())
-                        .host(parentURI.getHost())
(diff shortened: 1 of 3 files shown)


Tomucha:
Hi, could you please review and merge this PR? We are facing issues in production where relative redirects are not resolved against the request URI, causing broken requests. This is a regression from 2.x ... Thanks!

See also:
https://github.com/micronaut-projects/micronaut-core/issues/12364

Thanks @ArtemLikhitskiy for the fix!

yawkat:
Thanks!

PR #12397, opened 2026-02-18
don’t cache failures until the context is running

sdelamo:
[The Test in this PR passed in MN 4](https://github.com/micronaut-projects/micronaut-core/pull/12396). This regression breaks Micronaut AWS Lambda support and other Serverless Functions.

Fixed with the help of AI

>  Two changes in DefaultBeanDefinitionService.java BeanDefinitionProducer:
> 1. `getReferenceIfEnabled (line 659)`: Only permanently null out `this.reference` when `context.isRunning()`. Before startup, the reference is preserved so conditions can be re-evaluated later.
>  2. `getDefinitionIfEnabled (line 690)`: Only set `this.definition = DEFINITION_DISABLED_SENTINEL` when `context.isRunning()`. Same rationale.
> Why this works: Before start(), the environment hasn't processed property sources into the resolver catalog, so property-based `@Require` conditions fail incorrectly. By not caching these failures permanently, the conditions are re-evaluated after `start()` when properties are available. The performance impact is negligible — the checkEnabledBeans ForkJoinTask and normal bean lookups after startup will still permanently cache results as before.

### DIFF
--- a/buildSrc/gradle.properties
+++ b/buildSrc/gradle.properties
@@ -1,3 +1,0 @@
-#
-#Mon Jan 12 10:59:53 CET 2026
-micronaut-build-version=8.0.0-M14
--- a/function/build.gradle.kts
+++ b/function/build.gradle.kts
@@ -10,4 +10,7 @@ dependencies {
 
     testAnnotationProcessor(projects.micronautInjectJava)
     testImplementation(projects.micronautJacksonDatabind)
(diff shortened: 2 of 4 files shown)


alvarosanchez (approved):

sdelamo:
@dstepanov check this fix is correct once you are back.

PR #12401, opened 2026-02-20
Fix executor fallback by allowing null-named UserExecutorConfiguration

lissam1:
allow an unnamed (null-qualified) default UserExecutorConfiguration to be instantiated without explicit properties under `micronaut.executors.*`, providing a fallback for ThreadFactory and executors.


lissam1:
i add some nullability because i had some errorprone failures

sdelamo (approved):

PR #12406, opened 2026-02-24
bugfix(test): fix SslRefreshSpec test by dynamically validating available ciphers

### LINKED ISSUE #12381: Investigate SslRefreshSpec failures
### Expected Behavior

SslRefreshSpec should pass

### Actual Behaviour

Currently `SslRefreshSpec` fails on CI for an unexplained reason. We need to investigate why

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.x

Aliaksie:
This PR addresses test instability related to SSL refresh spec execution across different environments.

Root Cause: 
The test was requesting a predefined list of cipher suites, but on some environments (depending on OS and OpenSSL version), not all of those ciphers are actually available. As a result, the test fails even though the SSL configuration itself is valid.

What This PR Changes: 

- Adds logic to dynamically check whether the requested cipher suites are actually supported by the active SSL provider.
- Filters the requested cipher list against the provider’s available cipher suites.
- Skips the test if the resulting filtered cipher list is empty (i.e., none of the requested ciphers are supported in the current environment).

### DIFF
--- a/http-client/src/test/groovy/io/micronaut/http/client/SslRefreshSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/SslRefreshSpec.groovy
@@ -20,6 +20,7 @@ import io.netty.handler.ssl.util.SelfSignedCertificate
 import spock.lang.Shared
 import spock.lang.Specification
 
+import javax.net.ssl.SSLServerSocketFactory
 import java.nio.file.Files
 import java.nio.file.Path
 import java.security.KeyStore
@@ -52,6 +53,8 @@ class SslRefreshSpec extends Specification {
             println "Using OpenSSL provider. Supported ciphers: $ciphers"
         } else {
             // JDK will be used
+            def supportedCiphers = SSLServerSocketFactory.default.getSupportedCipherSuites() as List
+            ciphers = requiredCiphers.findAll { supportedCiphers.contains(it) }
             println "Using JDK provider. Supported ciphers: $ciphers"
         }
     }
--- a/http-client/src/test/groovy/io/micronaut/http/client/SslRefreshSpec.groovy
+++ b/http-client/src/test/groovy/io/micronaut/http/client/SslRefreshSpec.groovy
@@ -14,23 +14,23 @@ import io.micronaut.http.server.netty.NettyHttpRequest
 import io.micronaut.runtime.context.scope.refresh.RefreshEvent
 import io.micronaut.runtime.server.EmbeddedServer
 import io.netty.handler.ssl.SslHandler
+import io.netty.handler.ssl.SslProvider
+import io.netty.handler.ssl.OpenSsl
 import io.netty.handler.ssl.util.SelfSignedCertificate
-import spock.lang.Ignore
-import spock.lang.IgnoreIf
-import spock.lang.Retry
 import spock.lang.Shared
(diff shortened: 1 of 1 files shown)


yawkat (approved):

Aliaksie:
hi @yawkat @graemerocher , 

Thanks for you review!

I’ve noticed the `SslRefreshSpec` test still failed in CI -  sorry for missing this earlier. I investigated further and found that the test’s behavior is highly dependent on the SSL provider and environment:

Locally on my Mac (GraalVM JDK 21, macOS):
- The test passes when OpenSSL is available and selected as provider and using my filter logic to determine supported ciphers.
- The test also passes if I explicitly set the SSL provider to JDK.

On CI (GraalVM JDK 21):
The JDK provider is selected (OpenSSL is not available or not used).
The test fails with a handshake error: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure

The filtering logic for JDK that I previously added does not actually guarantee that the ciphers are truly available for negotiation, it only checks if the cipher name is accepted by the engine, not if it can be successfully used in a handshake. I plan to remove this code from the test, as it does not provide a reliable check for cipher suite availability.

Proposal:
To ensure reliable test results, we have at least two options:
- Skip the test if OpenSSL is not available (i.e., when SslProvider.isAlpnSupported(SslProvider.OPENSSL_REFCNT) is false).
- Ensure that all cipher suites used in the test are correctly configured and supported in the CI JDK environment (which may require updating the JDK security policy or test configuration).

Please let me know which option you’d like to proceed with, or if you have other suggestions!

Aliaksie:
> hi @yawkat @graemerocher ,
> 
> Thanks for you review!
> 
> I’ve noticed the `SslRefreshSpec` test still failed in CI - sorry for missing this earlier. I investigated further and found that the test’s behavior is highly dependent on the SSL provider and environment:
> 
> Locally on my Mac (GraalVM JDK 21, macOS):
> 
> * The test passes when OpenSSL is available and selected as provider and using my filter logic to determine supported ciphers.
> * The test also passes if I explicitly set the SSL provider to JDK.
> 
> On CI (GraalVM JDK 21): The JDK provider is selected (OpenSSL is not available or not used). The test fails with a handshake error: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
> 
> The filtering logic for JDK that I previously added does not actually guarantee that the ciphers are truly available for negotiation, it only checks if the cipher name is accepted by the engine, not if it can be successfully used in a handshake. I plan to remove this code from the test, as it does not provide a reliable check for cipher suite availability.
> 
> Proposal: To ensure reliable test results, we have at least two options:
> 
> * Skip the test if OpenSSL is not available (i.e., when SslProvider.isAlpnSupported(SslProvider.OPENSSL_REFCNT) is false).
> * Ensure that all cipher suites used in the test are correctly configured and supported in the CI JDK environment (which may require updating the JDK security policy or test configuration).
> 
> Please let me know which option you’d like to proceed with, or if you have other suggestions!

After further consideration, I’d like a propose to try when the JDK provider is selected do a filter the required ciphers against the list returned by SSLServerSocketFactory.getDefault().getSupportedCipherSuites(), which may be more accurately reflects the ciphers that can be negotiated :)

PR #12412, opened 2026-02-27
test: query value nullable parameter binding

sdelamo:
attempt to reproduce a regression. 

It relates to : https://github.com/micronaut-projects/micronaut-core/pull/12321#issuecomment-3847296347


mikehearn:
This can be closed. The root cause is String being accidentally registered for introspection. In Micronaut 4.9 that didn't matter. In Micronaut 4.10 it does, but the better fix is to catch this error as String should never be an introspectable bean.

PR #12413, opened 2026-02-27
feat(jpms): add Java Module System support to Micronaut

graemerocher:
## Summary
- add JPMS module descriptors for Micronaut packages so the platform exposes Java Module System metadata
- keep the existing module requirements while updating the descriptors to the 2026 copyright year and enforced doc comment so build tooling stays happy

## Why
- Micronaut needs explicit JPMS support so it can participate in modular applications without losing existing transitive exposure
- keeping these descriptors formatted and documented ensures Spotless/Checkstyle do not block downstream pushes

See https://github.com/micronaut-projects/micronaut-core/issues/6395

### DIFF
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
@@ -46,6 +46,106 @@ tasks.withType(JavaCompile).configureEach {
     options.forkOptions.memoryMaximumSize = "2g"
 }
 
+def moduleInfoJava = file("src/main/java/module-info.java")
+
+if (moduleInfoJava.exists()) {
+    pluginManager.withPlugin("java") {
+        def sourceSets = extensions.findByType(org.gradle.api.tasks.SourceSetContainer)
+        if (sourceSets != null) {
+            sourceSets.named("main") {
+                java {
+                    exclude("module-info.java")
+                }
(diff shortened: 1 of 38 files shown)


yawkat:
Don't we still have split packages?

buildSrc/src/main/groovy/io.micronaut.build.internal.convention-base.gradle
```diff
@@ -46,6 +46,106 @@ tasks.withType(JavaCompile).configureEach {
...
+                            f.name.startsWith('micronaut-rxjava2') ||
+                            f.name.startsWith('micronaut-rxjava3') ||
+                            f.name.startsWith('micronaut-core-processor') ||
+                            f.name.contains('sourcegen-model') ||
+                            f.name.contains('sourcegen-bytecode-writer')
+                        )
+                    }
+                    .collect { it.absolutePath }
+                    .join(File.pathSeparator)
+            }
```
yawkat, line 100: Not a great solution imo

graemerocher: the split packages are between the compiler modules and the runtime modules which are not on the same classpath typically (runtime vs annotation processor) so shouldn't really be an issue.

This PR is in draft because at some point @melix should have a look at the Gradle build changes which are problematic.

PR #12419, opened 2026-03-02
fix @EachProperty env entry lookup for hyphenated prefixes

### LINKED ISSUE #10726: Object Storage doesn't work with environment variables
### Expected Behavior

I should be able to specify object storage properties as env vars, e.g. MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_BUCKET for micronaut.object-storage.oracle-cloud.default.bucket and MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_NAMESPACE for micronaut.object-storage.oracle-cloud.default.namespace

### Actual Behaviour

Values are only resolved from config files.

I discussed this with Alvaro and although this affects Object Storage, it's likely a core bug.

### Steps To Reproduce

Create an app

    mn create-app --build=gradle --features=object-storage-oracle-cloud com.example.objstoragebug

Add printlns to Application.java to display resolved values (see attached project)

Specify bucket and namespace as env vars, either with

    export MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_BUCKET=foo
    export MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_NAMESPACE=bar
    ./gradlew run

or

    MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_BUCKET=foo MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_NAMESPACE=bar ./gradlew run

In both cases the properties are correctly read and available from the ApplicationContext, but there is no enabled ObjectStorageOperations bean (should be io.micronaut.objectstorage.oraclecloud.OracleCloudStorageOperations)

If you hard code the values in application.properties, or in application-foo.properties and run with the "foo" env enabled, it works.

One other weird combination works. If I specify env vars and have one or more related properties in application.properties, it works. Uncomment

    micronaut.object-storage.oracle-cloud.default.enabled=true

in application.properties in the attached app and run with env vars and everything works as expected. Note that any of the 3 properties will cause this, it's not related to the value being true since the default is true.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.3.8

yawkat:
## Summary
- Switch the default `PropertySourcePropertyResolver#getPropertyEntries(String)` catalog from `NORMALIZED` to `GENERATED` so environment-variable based keys are discovered consistently.
- Add a regression test for `MICRONAUT_OBJECT_STORAGE_ORACLE_CLOUD_DEFAULT_BUCKET` style keys to verify default lookup behavior matches generated entries.
- Fixes #10726.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/env/DefaultConfigurationPath.java
+++ b/inject/src/main/java/io/micronaut/context/env/DefaultConfigurationPath.java
@@ -208,6 +208,19 @@ final class DefaultConfigurationPath implements ConfigurationPath {
         switch (kind) {
             case MAP -> {
                 Collection<String> entries = propertyResolver.getPropertyEntries(thisPath.prefix(), thisPath.propertyCatalog());
+                if (entries.isEmpty() && thisPath.propertyCatalog() == PropertyCatalog.NORMALIZED) {
+                    Collection<String> generatedEntries = propertyResolver.getPropertyEntries(thisPath.prefix(), PropertyCatalog.GENERATED);
+                    if (!generatedEntries.isEmpty()) {
+                        Collection<String> filteredEntries = new LinkedList<>();
+                        String prefix = thisPath.prefix();
+                        for (String key : generatedEntries) {
+                            if (!propertyResolver.getPropertyEntries(prefix + '.' + key, PropertyCatalog.GENERATED).isEmpty()) {
+                                filteredEntries.add(key);
+                            }
+                        }
+                        entries = filteredEntries;
+                    }
+                }
                 for (String key : entries) {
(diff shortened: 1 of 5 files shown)


graemerocher (approved):

sdelamo (changes requested):
CI Failed

yawkat:
Not sure I like the latest version of the fix…

graemerocher:
need to increase the coverage to cover the different cases

graemerocher (approved):

PR #12425, opened 2026-03-02
Check that basic Java types aren't being registered for introspection.

mikehearn:
This is a programming error that can cause bugs elsewhere in Micronaut, such as in the HTTP server where missing query params may get mapped to default values accidentally.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
+++ b/core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
@@ -24,6 +24,7 @@ import io.micronaut.core.annotation.AnnotationValueBuilder;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Introspected;
 import io.micronaut.core.annotation.NonNull;
+import io.micronaut.core.reflect.ClassUtils;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.inject.annotation.AnnotationMetadataHierarchy;
@@ -112,6 +113,7 @@ public class IntrospectedTypeElementVisitor implements TypeElementVisitor<Object
         final String targetPackage = introspected.stringValue("targetPackage").orElse(element.getPackageName());
 
         if (!classes.isEmpty()) {
+            classes.forEach(className -> validateIntrospectedType(className, element));
             AtomicInteger index = new AtomicInteger(0);
             classes.stream().flatMap(className -> context.getClassElement(className).stream()).forEach(ce -> {
                 if (isIntrospected(context, ce)) {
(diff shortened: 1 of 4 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1692,6 +1693,7 @@ public BeanContext registerBeanConfiguration(BeanConfiguration configuration) {
     public <B> BeanContext registerBeanDefinition(@NonNull RuntimeBeanDefinition<B> definition) {
         Objects.requireNonNull(definition, "Bean definition cannot be null");
         Class<B> beanType = definition.getBeanType();
+        RuntimeBeanTypeGuard.validateBeanType(beanType);
```
dstepanov: I believe we can have basic types as beans.

core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
```diff
@@ -196,6 +200,12 @@ private void processIntrospected(ClassElement element, VisitorContext context, A
         }
     }
 
+    private static void validateIntrospectedType(String className, ClassElement originatingElement, VisitorContext context) {
+        if (ClassUtils.isJavaBasicType(className)) {
+            context.fail(ClassUtils.basicJavaTypeRegistrationMessage(className), originatingElement);
```
dstepanov: throw ProcessingException

dstepanov (changes requested):

core/src/main/java/io/micronaut/core/reflect/ClassUtils.java
```diff
@@ -271,6 +271,19 @@ public static boolean isJavaBasicType(@Nullable String name) {
...
+     * Format the error message used when a basic JDK value type is registered in bean/introspection infrastructure.
+     *
+     * @param name The basic type name
+     * @return The formatted error message
+     * @since 4.10.17
+     */
+    @NonNull
+    public static String basicJavaTypeRegistrationMessage(@NonNull String name) {
+        return ("Basic JDK value type [%s] cannot be registered as introspectable, serializable, " +
+            "or as a bean. Register a dedicated wrapper/holder type instead.").formatted(name);
```
graemerocher: basic types can be beans for dependency injection purposes so remove `or as a bean`

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
graemerocher: remove `RuntimeBeanTypeGuard` and related changes. Basic types can be beans for dependency injection

inject-java/src/test/groovy/io/micronaut/inject/requires/RequiresSpec.groovy
```diff
@@ -459,14 +461,15 @@ package test;
 
 import io.micronaut.context.annotation.*;
 
-@Requires(beans=String.class)
+@Requires(beans=Runnable.class)
```
graemerocher: revert these changes

graemerocher (changes requested):

mikehearn:
Thanks. Comments applied, PTAL. What's the use case for beans of basic Java types that isn't met by configuration? The definition of bean given in the docs is an object whose lifecycle is managed by the injector, but basic/value types don't have a lifecycle, so this was a bit surprising.

graemerocher:
not sure how useful it is, we added it when we were working on CDI support. The CDI specification supports beans of basic types, primitives, arrays of primitives etc.

graemerocher (approved):

sdelamo:
@mikehearn can you merge 4.10.x into this branch. @dstepanov can you re-review?

PR #12427, opened 2026-03-04
Docs: Add note about YAML reserved words in ConfigurationProperties

### LINKED ISSUE #12409: @ConfigurationProperties fails to bind properties named `yes` or `no`
### Expected Behavior

Properties named `yes` and `no` should bind correctly to data class fields.

### Actual Behaviour

The values are not injected; fields retain their default values or remain null.

OR if this is intended some documentation that dictates non-binding property names.

### Steps To Reproduce

Kotlin configuration class:

    @ConfigurationProperties("foo.bar")
    data class FooBar(
        val yes: String = "",
        val no: String = "",
    )

YAML configuration:

    foo:
      bar:
        yes: "1"
        no: "2"

The `yes` and `no` fields do not receive the configured values.

### Environment Information

Local: macOS Sequoia 15.7.4
Server: Linux
JDK: 21
Kotlin: 2.0.21

### Example Application

_No response_

### Version

4.10.2

nikitanagar08:
## Summary

Added documentation explaining YAML reserved boolean words (yes/no, true/false, on/off) and how they are automatically converted to boolean values when used as unquoted keys in YAML configuration.

## Problem

When property names like yes or no are used in YAML configuration without quotes, SnakeYAML parses them as boolean values (true/false) instead of strings. This causes the property binding to fail.

## Solution

Added a comprehensive note in the configurationProperties.adoc documentation that:

1. Lists all YAML reserved boolean words
2. Explains the parsing behavior
3. Shows the incorrect and correct YAML syntax
4. Provides a workaround using quoted keys
5. References the YAML 1.2 specification

## Changes

Modified:
- src/main/docs/guide/config/configurationProperties.adoc

## Test Plan

- Documentation built successfully
- Checked for correct AsciiDoc syntax
- Verified code examples are clear

Fixes #12409

### DIFF
--- a/src/main/docs/guide/config/configurationProperties.adoc
+++ b/src/main/docs/guide/config/configurationProperties.adoc
@@ -32,6 +32,40 @@ Note for more complex configurations you can structure ann:context.annotation.Co
 
 For example creating a subclass of `EngineConfig` with `@ConfigurationProperties('bar')` will resolve all properties under the path `my.engine.bar`.
 
+[NOTE]
+====
+YAML Reserved Words
+~~~~~~~~~~~~~~~~~~~~
+
+When using YAML configuration files, certain words are reserved by the YAML specification and will be automatically converted to boolean values. This includes:
+
+* `yes` / `no`
+* `true` / `false`
+* `on` / `off`
+
+If you use these words as unquoted property keys in YAML, they will be parsed as booleans instead of strings:
+
+[source,yaml]
+----
+foo:
+  bar:
+    yes: "1"     # yes is parsed as true (boolean)
+    no: "2"      # no is parsed as false (boolean)
+----
+
+To use these words as property names, you must quote the keys:
+
+[source,yaml]
+----
+foo:
+  bar:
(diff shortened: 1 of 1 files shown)


src/main/docs/guide/config/configurationProperties.adoc
```diff
@@ -32,6 +32,40 @@ Note for more complex configurations you can structure ann:context.annotation.Co
 
 For example creating a subclass of `EngineConfig` with `@ConfigurationProperties('bar')` will resolve all properties under the path `my.engine.bar`.
 
+[NOTE]
+====
+YAML Reserved Words
+~~~~~~~~~~~~~~~~~~~~
+
+When using YAML configuration files, certain words are reserved by the YAML specification and will be automatically converted to boolean values. This includes:
```
graemerocher:
```suggestion
When using YAML configuration files, certain words are reserved by the YAML specification and will be automatically converted to boolean values. These include:
```

graemerocher (changes requested):

graemerocher (approved):

PR #12436, opened 2026-03-05
test: add regression for RequestBean record path variable binding

### LINKED ISSUE #8551: Using records with a required @PathVariable causes a Bad Request response
### Expected Behavior

Given a controller like this:
```
@Controller("/api/example")
public class ExampleController {
    @Post("/{id}")
    @Produces(MediaType.APPLICATION_JSON)
    public HttpResponse<?> example(@Valid @RequestBean Bean bean) {
        return HttpResponse.ok(
            Map.of(
                "bean", bean.toString()
            )
        );
    } //example
}
```
I expect `bean` to be filled with the value of the specified path variable, such as `1` in the case of `/api/example/1`. `Bean` is defined as follows:
```
@Introspected
public record Bean(HttpRequest<?> httpRequest, @PathVariable @Positive Integer id) {
}
```

### Actual Behaviour

Instead, the sender of the request receives the following response:
```
{
    "message": "Bad Request",
    "_links": {
        "self": {
            "href": "/api/example/1",
            "templated": false
        }
    },
    "_embedded": {
        "errors": [
            {
                "message": "Required argument [Integer id] not specified",
                "path": "/id"
            }
        ]
    }
}
```

### Steps To Reproduce

1. Clone the example application
2. Run the example application
3. Make a `POST` request to `/api/example/1` (`curl -X POST -L http://localhost:8080/api/example/1`)
4. See that the server has responded with `Required argument [Integer id] not specified` even though an ID was specified in the path

### Environment Information

- Operating System: macOS Ventura 13.1
- JDK (I presume this issue will also be reproducible in JDK 17, 19, etc.):
```
openjdk 21-ea 2023-09-19
OpenJDK Runtime Environment (build 21-ea+3-124)
OpenJDK 64-Bit Server VM (build 21-ea+3-124, mixed mode, sharing)
```

### Example Application

https://github.com/lbkulinski/record_bug_example

### Version

3.8.0

yawkat:
## Summary
- add regression coverage for issue #8551 on 5.0.x
- verify @PathVariable binds correctly inside an @Introspected record used as @RequestBean

## Issue
Fixes #8551


graemerocher (approved):

graemerocher:
should this not go in the HTTP TCK?

yawkat:
Yea seems better

PR #12438, opened 2026-03-05
fix: bind foo-123-bar style keys in configuration properties

### LINKED ISSUE #10215: @ConfigurationProperties does not load values if the property name uses snake_case or kebab-case and contains a number
### Expected Behavior

@ConfigurationProperties should be able to load configuration values when the property name uses `kebab-case` or `snake_case` containing (but not beginning with) a number.

For instance, values for the following property names should be loaded correctly:
* `snake_case_1234_prop`
* `kebab-case-1234-prop`
* `camelCase1234Prop`

### Actual Behaviour

@ConfigurationProperties does _not_ load values for `kebab-case` or `snake-case` property names containing numbers.

For example, values are *not* loaded for the following property names:
* `snake_case_1234_prop`
* `kebab-case-1234-prop`

However, values are loaded for property name `camelCase1234Prop`.

### Steps To Reproduce

The issue can be reproduced with the `application.yml` and `ConfigurationExample.kt` files copied below. 

### `application.yml` configuration file
```
configuration-example:
  camelCaseProp: 1
  camelCase1234Prop: 2
  PascalCaseProp: 3
  PascalCase1234Prop: 4
  snake_case_prop: 5
  snake_case_1234_prop: 6
  kebab-case-prop: 7
  kebab-case-1234-prop: 8
```

### ConfigurationExample.kt using @ConfigurationProperties
```
package com.example

import io.micronaut.context.annotation.ConfigurationProperties

@ConfigurationProperties("configuration-example")
class ConfigurationExample {
    var camelCaseProp: Number = 0;
    var camelCase1234Prop: Number = 0;
    var PascalCaseProp: Number = 0;
    var PascalCase1234Prop: Number = 0;
    var snakeCaseProp: Number = 0;
    var snakeCase1234Prop: Number = 0;
    var kebabCaseProp: Number = 0
    var kebabCase1234Prop: Number = 0
}
```

## IntelliJ Debugger with a loaded `ConfigurationExample` bean
Screenshot 2023-12-05 at 8 30 05 AM

### Environment Information

- JDK Version: openjdk 17.0.9 2023-10-17

### Example Application

_No response_

### Version

4.1.5

yawkat:
## Summary
- reproduce issue #10215 with the concrete `foo-123-bar`, `snake_case_1234_prop`, and `kebab-case-1234-prop` examples and assert they bind to the Java fields `foo123Bar`, `snakeCase1234Prop`, and `kebabCase1234Prop`
- normalize non-environment property source keys with a dehyphenate→hyphenate canonical alias so standalone numeric kebab segments resolve to the same canonical key as generated bean property names
- keep existing camelCase / snake_case / kebab-case digit binding coverage in the same regression test

## Why
`@ConfigurationProperties` bean fields are normalized to canonical kebab names (for example `foo123Bar` -> `foo123-bar`).
A user-supplied key like `foo-123-bar` previously stayed as-is and failed to match that canonical form.

## Verification
- `./gradlew :micronaut-inject-java:test --tests "io.micronaut.inject.configproperties.ConfigurationPropertiesSpec.test configuration properties binding for kebab/snake case names containing digits" --no-daemon`
- `./gradlew :micronaut-inject-java:test --tests "io.micronaut.inject.configproperties.ConfigurationPropertiesSpec" --no-daemon`
- `./gradlew :micronaut-inject:test --tests "io.micronaut.context.env.PropertySourcePropertyResolverSpec" --no-daemon`

Resolves #10215


inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
```diff
@@ -977,9 +977,19 @@ private List<String> resolvePropertiesForConvention(String property, PropertySou
...
+        //   hyphenate("foo-123-bar")               -> "foo-123-bar" (already hyphenated)
+        //   dehyphenate("foo-123-bar")             -> "Foo123Bar"
+        //   hyphenate(dehyphenate("foo-123-bar"))  -> "foo123-bar"
+        // Generated @ConfigurationProperties metadata for field foo123Bar uses "foo123-bar",
+        // so storing both aliases allows user-supplied "foo-123-bar" to resolve correctly.
+        String dehyphenated = NameUtils.hyphenate(NameUtils.dehyphenate(property), true);
+        if (hyphenated.equals(dehyphenated)) {
+            return Collections.singletonList(hyphenated);
+        }
+        return Arrays.asList(hyphenated, dehyphenated);
```
graemerocher, line 992: ideally it should try the hyphenated and if it doesn't resolve fall back to the dehyphenated instead of paying the cost of both

PR #12439, opened 2026-03-06
Expand thread-selection behavior

yawkat:
Thread-selection configuration should apply to filters and event listeners as well.

Also adds an option to only use the thread-selection parameter when necessary. For example, if thread-selection=blocking, it won't start virtual threads for each filter and controller, it will only start one virtual thread when it's first needed during request processing.


context/src/main/java/io/micronaut/scheduling/executor/DefaultExecutorSelector.java
```diff
@@ -66,24 +68,27 @@ protected DefaultExecutorSelector(
...
             } catch (NoSuchBeanException e) {
                 throw new SchedulerConfigurationException(
-                        method,
+                        methodReference,
                         "No executor configured for name: " + name
                 );
             }
         } else if (threadSelection == ThreadSelection.AUTO) {
-            if (method.hasStereotype(NonBlocking.class)) {
+            if (method != null && method.hasStereotype(NonBlocking.class)) {
```
yawkat: can collapse with the == null branch

core/src/main/java/io/micronaut/core/execution/ImmediateExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.concurrent.Executor;
+
+/**
+ * An executor that immediately executes its tasks.
+ *
+ * @since 5.0.0
+ * @author Jonas Konrad
+ */
+public final class ImmediateExecutor implements ConditionalExecutionExecutor {
```
graemerocher, line 29: Internal?

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -238,6 +249,15 @@ public void setThreadSelection(@Nullable ThreadSelection threadSelection) {
         }
     }
 
+    /**
+     * Sets whether configured executors are used only for calls from non-blocking threads.
+     *
+     * @param redispatchNonBlockingOnly {@code true} to redispatch only from non-blocking threads
+     */
+    public void setRedispatchNonBlockingOnly(boolean redispatchNonBlockingOnly) {
```
graemerocher, line 273: please explain in more detail the behavioural implication of this setting from reading it will be unclear to the standard user

graemerocher (changes requested):
Document in breaks.adoc this change of behaviour and the implications for users

graemerocher:
also fix the tests :)

core/src/main/java/io/micronaut/core/execution/ImmediateExecutor.java
yawkat: I think this can be public API with no risk.

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
yawkat: done

src/main/docs/guide/appendix/breaks.adoc
```diff
@@ -105,6 +105,19 @@ Executable methods annotated with `@Parallel` will not trigger parallel executio
 
 `@Scheduled` is no longer processed in parallel, this reduced the complexity of the method processor and removed unnecessary overhead.
 
+==== HTTP server thread selection now applies across more request stages
```
yawkat, line 108: @graemerocher WDYT? more verbose than I'd have written it, but it is correct and useful

core/src/main/java/io/micronaut/core/execution/ImmediateExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void execute(Runnable command) {
+        try {
+            command.run();
+        } catch (Exception e) {
+            LOG.error("Error in immediate executor", e);
+        }
+    }
```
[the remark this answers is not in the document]
yawkat: intentional

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -184,27 +204,68 @@ public void accept(ChannelHandlerContext ctx, io.netty.handler.codec.http.HttpRe
...
+            return ExecutionFlow.empty();
         }
+        return ExecutionFlow.async(getRequestEventExecutor(), () -> {
+            try (PropagatedContext.Scope ignore = PropagatedContext.getOrEmpty()
+                .plus(new ServerHttpRequestContext(request))
+                                .propagate()) {
+                receivedPublisher.publishEvent(new HttpRequestReceivedEvent(request));
+            }
+            return ExecutionFlow.empty();
+        });
```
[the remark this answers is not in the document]
yawkat: not workable, we have to assume the executor works.

graemerocher (approved):

sdelamo (changes requested):
Please, add thread selection documentation and link it from `breaks.adoc` instead of documenting in that section.

Issue #10851, opened 2024-05-23
StreamingFileUpload with Multipart Form Data materialises in RAM when QueryParam is used in Controller and query param is not sent


lmoric-vendavo:
### Expected Behavior

Subscribing to StreamingFileUpload should produce small enough chunks so large file uploads can be streamed/processed without a need to be all materialised in RAM.

### Actual Behaviour

Only 2 chunks are present, where second chunk contains whole file content.

### Steps To Reproduce

Working example controller:
```
    @Post("/testMultipartUpload")
    @Consumes(MediaType.MULTIPART_FORM_DATA)
    fun testMultipartUpload(
        @PathVariable version: ApiVersion,
        httpRequest: HttpRequest<Any>,
        datasetFile: StreamingFileUpload
    ): Mono<HttpResponse<String>> {
        println("MD5: " + httpRequest.parameters.get("md5"))
        val mySub = MySub()
        datasetFile.subscribe(mySub)
        return Mono.just(ok("Uploaded"))
    }
```
MD5 parameter is obtained from HttpRequest. MySub() is a subscriber, which can get chunks from the StreamingFileUpload, in this case working as expected, large files can be consumed by thousands of ~65KB chunks.

When we put any kind of query parameter, things got broken:
```
    @Post("/testMultipartUpload{?md5}")
    @Consumes(MediaType.MULTIPART_FORM_DATA)
    fun testMultipartUpload(
        @PathVariable version: ApiVersion,
        @QueryValue md5: String? = null,
        datasetFile: StreamingFileUpload
    ): Mono<HttpResponse<String>> {
        println("MD5: " + md5)
        val mySub = MySub()
        datasetFile.subscribe(mySub)
        return Mono.just(ok("Uploaded"))
    }
```
In this example, where we provide `md5` parameter as QueryValue, MySub() will obtain just two chunks, where second chunk can be several gigabytes (whole uploaded file). This of course can lead to JVM Out Of Memory for large files.

However, the issue is present only in case `md5` is not send in API call, when `md5` is sent, everything works as expected again.

### Environment Information

Amazon Corretto 17.0.11
io.micronaut:micronaut-http-server-netty:4.4.9

### Version

4.4.0

graemerocher:
[...]

PR #12454, opened 2026-03-10
test: add regression coverage for optional query value multipart streaming

yawkat:
## Summary
- add a regression test for issue #10851 in `ConcurrentFormTransferSpec`
- cover multipart streaming behavior for `StreamingFileUpload` with optional `@QueryValue` both missing and present
- assert that uploads stay chunked (instead of materializing into one large chunk) in both request variants on 5.0.x

## Testing
- `./gradlew :micronaut-http-server-netty:test --tests "io.micronaut.http.server.netty.binding.ConcurrentFormTransferSpec"`

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/ConcurrentFormTransferSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/ConcurrentFormTransferSpec.groovy
@@ -1,9 +1,13 @@
 package io.micronaut.http.server.netty.binding
 
 import io.micronaut.context.ApplicationContext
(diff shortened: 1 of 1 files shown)


graemerocher (approved):

yawkat:
Addressed Copilot review comments in 93573abf19:\n- switched multipart request construction to InputStream-based addPart with fresh stream per request\n- changed nullable annotation to jakarta.annotation.Nullable for test-package consistency\n- replaced side-effect map with doOnNext in streaming body processing\n- explicitly close HttpClient in cleanup\n- relaxed brittle chunk-count assertions while keeping max-chunk-size guard against full materialization

PR #12461, opened 2026-03-10
Fix request context availability in coroutine context factories

### LINKED ISSUE #12373: ServerRequestContext.currentRequest() return null in HttpCoroutineContextFactory when use CompletedFileUpload argument
### Expected Behavior

I expect HttpCoroutineContextFactory to always behave the same way.
But when invoking a controller method with a file upload using CompletedFileUpload, the following call returns null:

``` kotlin
ServerRequestContext.currentRequest<Any>()
```

In other scenarios, it does not return null.

### Actual Behaviour

ServerRequestContext.currentRequest<Any>() inside HttpCoroutineContextFactory.create
returns null when call controller method with CompletedFileUpload argument

In other scenarios, it does not return null.

### Steps To Reproduce

1. Create new micronaut application
2. Add HttpCoroutineContextFactory 
3. Call ServerRequestContext.currentRequest<Any>() inside create method
4. Add controller method with CompletedFileUpload argument
5. Check that current reques is null in case to call upload file

### Environment Information

_No response_

### Example Application

https://github.com/pelenthium/micrnonaut-coroutine-factory

### Version

4.10.7

yawkat:
## Summary
- move coroutine context setup into `executeRouteAndConvertBody(...)` and run it before `requestArgumentSatisfier.fulfillArgumentRequirementsAfterFilters(...)`
- for suspended routes, pass `isKotlinCoroutine=true` and `ContextView` into `executeRouteAndConvertBody(...)`
- keep `ContinuationArgumentBinder` simple (no extra factory-time propagation wrapper)
- retain suspend + `CompletedFileUpload` regression coverage for request context availability

## Why this is necessary
The timing issue remains the root cause:
- suspended routes previously called `setupCoroutineContext(...)` before entering the request-scoped propagation in `executeRouteAndConvertBody(...)`
- `HttpCoroutineContextFactory.create()` can call `ServerRequestContext.currentRequest()`
- at that moment, request context was not yet in scope

This alternative fix applies the suggestion from PR comments by moving coroutine setup into the request-scoped execution path, so factory creation naturally sees the request context.

## Verification
- `./gradlew :test-suite-kotlin:test --tests 'io.micronaut.docs.server.suspend.SuspendControllerSpec' --no-daemon -q`
- `./gradlew :micronaut-http:compileKotlin :micronaut-http-server:compileJava :test-suite-kotlin:test --tests 'io.micronaut.docs.server.suspend.SuspendControllerSpec' --no-daemon -q`

Fixes #12373

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
+++ b/http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
@@ -71,6 +71,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
+import reactor.util.context.ContextView;
 
 import java.io.IOException;
 import java.nio.channels.ClosedChannelException;
@@ -434,9 +435,8 @@ public final class RouteExecutor {
         if (executorService != null) {
             if (routeInfo.isSuspended()) {
                 executeMethodResponseFlow = ReactiveExecutionFlow.fromPublisher(Mono.deferContextual(contextView -> {
-                        coroutineHelper.ifPresent(helper -> helper.setupCoroutineContext(request, contextView, propagatedContext));
                         return Mono.from(
-                            ReactiveExecutionFlow.fromFlow(executeRouteAndConvertBody(propagatedContext, routeMatch, request)).toPublisher()
+                            ReactiveExecutionFlow.fromFlow(executeRouteAndConvertBody(propagatedContext, routeMatch, request, true, contextView)).toPublisher()
                         );
                     }));
             } else if (routeInfo.isReactive()) {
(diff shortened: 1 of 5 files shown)


yawkat:
@dstepanov PTAL, I'm not super happy with the proposed fix. Maybe we can move the request context set to be earlier so this change isnt necessary?

dstepanov:
Maybe, cannot tell why is this caused

yawkat:
@dstepanov I wondered this too but the AI explanation makes sense to me: The coroutine ctx is created in setupCoroutineContext, which is called before `propagatedContext.plus(new ServerHttpRequestContext` in RouteExecutor

dstepanov:
Try to move ` coroutineHelper.ifPresent...` to `executeRouteAndConvertBody` before `requestArgumentSatisfier.fulfillArgumentRequirementsAfterFilters`

Add a new boolean isKotlinCoroutine to executeRouteAndConvertBody and default it to false

dstepanov (approved):

PR #12467, opened 2026-03-11
Fix request scope propagation after suspend client calls

### LINKED ISSUE #12399: Calling a "RequestScoped" bean after a suspending function call throws an Exception "No request present"
### Expected Behavior

_RequestScope_ beans should remain accessible throughout the entire request handling, including after suspend calls to HTTP clients.

Workaround
Wrapping _RequestScope_ bean access in launch() { } or async() { } restores access, suggesting the context is somehow re-established on a new coroutine/thread.

### Actual Behaviour

After a declarative HTTP client suspend call completes, accessing @RequestScope beans throws:
java.lang.IllegalStateException: No request present

### Steps To Reproduce

```
@Controller
class TestCoroutineController(
    private val requestScopedBean: TestRequestScopedBean,
    private val testClient: TestClient,
) {
    private val logger = LoggerFactory.getLogger(javaClass)

    /**
     * Demonstrates that @RequestScope beans work before any HTTP client call.
     */
    @Get("/get-path")
    suspend fun beforeClientCall() {
        val path = requestScopedBean.getValue() // Works fine
        logger.info("Current path: $path")

        testClient.callEndpoint()
        val path2 = requestScopedBean.getValue() // Throws "No request present"
        logger.info("Current path: $path2")


        coroutineScope {
            launch(Dispatchers.IO) {
                val asyncPath = requestScopedBean.getValue()    // Works again
                logger.info("Async Current path: $asyncPath")
            }
        }
    }
}

/**
 * Simple @RequestScope bean for demonstrating context loss.
 */
@RequestScope
class TestRequestScopedBean : RequestAware {
    private var requestPath: String? = null

    override fun setRequest(request: HttpRequest<*>?) {
        requestPath = request?.path
    }

    fun getValue(): String = "Request path: $requestPath, Thread: ${Thread.currentThread().name}"
}

@Client("https://www.google.com")
interface TestClient {

    @Get("/")
    suspend fun callEndpoint(): Any
}
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.8

yawkat:
## Summary
- add a Kotlin suspend regression in `test-suite-kotlin` that reproduces request-scope loss after a declarative suspend HTTP client call when Micrometer context propagation is on the classpath
- preserve `PropagatedContext` while resuming `CompletionStage` results in `KotlinInterceptedMethodHelper`, so request scope remains available after the client call resumes
- wire Micrometer context-propagation through the version catalog (`libs.versions.toml`) and consume it via `libs.micrometer.context.propagation`

## Verification
- `./gradlew :test-suite-kotlin:test --tests '*SuspendControllerSpec' -q`

Fixes #12399

### DIFF
--- a/aop/src/main/kotlin/io/micronaut/aop/util/KotlinInterceptedMethodHelper.kt
+++ b/aop/src/main/kotlin/io/micronaut/aop/util/KotlinInterceptedMethodHelper.kt
@@ -20,7 +20,6 @@ import io.micronaut.core.annotation.Internal
 import io.micronaut.core.propagation.PropagatedContext
 import java.util.concurrent.CompletionException
 import java.util.concurrent.CompletionStage
-import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
 
 /**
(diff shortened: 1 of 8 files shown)


aop/src/main/kotlin/io/micronaut/aop/util/KotlinInterceptedMethodHelper.kt
```diff
@@ -34,17 +35,25 @@ internal object KotlinInterceptedMethodHelper {
...
+                            throwable
+                        }
+                        continuation.resumeWithException(exception)
                     }
-                    continuation.resumeWithException(exception)
+                }
+                if (propagatedContext == null) {
+                    resume()
+                } else {
+                    propagatedContext.propagate().use { resume() }
```
dstepanov: Pls avoid try-resources propagate

dstepanov (changes requested):

PR #12468, opened 2026-03-11
Update graal version to 25.0.2

sdelamo:
Updated graal-svm version from 23.1.10 to 25.0.2.


sdelamo:
@graemerocher I think we should make the graalvm version and `org.graalvm.sdk:nativeimage` dependency managed in `micronaut-core+ so that they are defined in the core BOM.  What do you think?

graemerocher:
Ok

PR #12471, opened 2026-03-11
Propagate context for default error fallback logging

### LINKED ISSUE #12268: RequestLifecycle onErrorNoFilter default logging MDC propagation
### Discussed in https://github.com/micronaut-projects/micronaut-core/discussions/12267

<div type='discussions-op-text'>

<sup>Originally posted by **andrew-pirkl** December 14, 2025</sup>
Hey all, I have a question about the RequestLifecycle and the way default error handling works. When a request goes through the `onErrorNoFilter` function and has no `errorRoute` or `ExceptionHandler` defined, then the exception eventually logged through the call to [createDefaultErrorResponseFlow](https://github.com/micronaut-projects/micronaut-core/blob/2a53b7e4bbb6d17e0a0f39266fc30601073a7dca/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java#L260C35-L260C47) doesn't include the `PropagatedContext`  and so doesn't include any of the propagated MDC items.

Is this by design, and is there a suggested way to make the MDC items available in that log? I tried creating a generic `ExceptionHandler` to handle all exceptions and log it there so it goes through [handlerExceptionHandler](https://github.com/micronaut-projects/micronaut-core/blob/2a53b7e4bbb6d17e0a0f39266fc30601073a7dca/http-server/src/main/java/io/micronaut/http/server/RequestLifecycle.java#L253) and that correctly includes the MDC items in the logged exception, but I'd prefer to not have to have that custom handler.

I did see there was a related bug fixed here: https://github.com/micronaut-projects/micronaut-core/pull/11847</div>

yawkat:
## Summary
- propagate request `PropagatedContext` through `RequestLifecycle` fallback default error response flow so default error logging runs with the request context
- add a regression test that verifies default fallback error logging includes MDC values from `MdcPropagationContext`
- keep propagation in lambda style via `propagatedContext.propagate(...)`

Fixes #12268

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/handler/DefaultErrorResponsePropagatedContextTest.java
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/errors/handler/DefaultErrorResponsePropagatedContextTest.java
@@ -1,0 +1,102 @@
+package io.micronaut.http.server.netty.errors.handler;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
+import io.micronaut.context.ApplicationContext;
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.context.propagation.slf4j.MdcPropagationContext;
+import io.micronaut.core.propagation.MutablePropagatedContext;
+import io.micronaut.http.HttpRequest;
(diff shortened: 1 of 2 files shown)


yawkat:
will retarget to 5.0.x

graemerocher (approved):

Issue #12203, opened 2025-11-19
@Pattern bean validation should not be applied to a null query parameter annotated with @Nullable


asamal:
### Expected Behavior

When a query parameter is annotated with `@Nullable` and `@Pattern` and `null` is passed, it is expected that the parameter is not validated to match the regex. This behavior was in Micronaut 4.9.4

### Actual Behaviour

The `@Pattern` bean validation is applied to a `null` value and fails with `must match "{regexp}"` `ValidationException`.

### Steps To Reproduce

Run `MyControllerTest#sayHello_null` in the attached example application
⚠️ `String.class` should be annotated with `@Introspected` to reproduce the error

### Environment Information

- Operating System: macOS Tahoe 26.1
- JDK Version: 25

### Example Application

https://github.com/asamal/micronaut-core-issue-12203

### Version

4.10.1

graemerocher:
Triage: valid validation regression with a complete external reproducer. Jakarta Validation treats null as valid for constraints such as Pattern unless a separate not-null constraint is present, so the current failure indicates that nullable metadata is being lost or validation is being invoked with the wrong argument state. Open PR #12475 directly adds a focused regression test for the introspected String case; keep this issue open for review and merge. Suggested fix is to preserve Nullable/optional parameter semantics through route argument binding and validation metadata synthesis.

PR #12475, opened 2026-03-12
test: reproduce nullable @Pattern query regression with introspected String

yawkat:
## Summary
- add a Java test route combining `@QueryValue(\"name\")`, nullable `String`, and `@Pattern(regexp = \"^(alice|bob)$\")`
- add `@Introspected(classes = { String.class })` test metadata to mirror the attached reproducer setup
- extend `io.micronaut.http.server.netty.java.ParameterBindingSpec` with `/java/parameter/nullable-pattern` expecting `200` and `Parameter Value: null`

## Verification
- ran: `./gradlew :micronaut-http-server-netty:test --tests 'io.micronaut.http.server.netty.java.ParameterBindingSpec' --stacktrace`
- result on `5.0.x`: fails in `test bind HTTP parameters for URI /java/parameter/nullable-pattern` with `body == null` (non-200 path), reproducing the regression

Related: #12203


yawkat:
@graemerocher @dstepanov the linked issue seems to happen due to an introspection for java.lang.String, is that something we support?

PR #12476, opened 2026-03-12
fix: preserve compatible SequencedSet @Parameter values in createBean

### LINKED ISSUE #12232: Cannot create bean with SequencedSet parameter
### Expected Behavior

When creating a bean that has `@Parameter SequencedSet<Integer> set` in its constructor argument list, I expect passing a `LinkedHashSet` to `createBean` to create a new instance of that bean.

### Actual Behaviour

The bean context throws an exception that it can't convert the object to a `SequencedSet`:

```
io.micronaut.context.exceptions.BeanInstantiationException: Error instantiating bean of type  [com.example.Application$A]

Message: Invalid bean @Argument [SequencedSet<Integer E> set]. Cannot convert object [[]] to required type: interface java.util.SequencedSet
Path Taken:
new c.e.A$A([@Parameter SequencedSet<Integer> set])
	at io.micronaut.context.DefaultBeanContext.lambda$resolveArgumentValues$15(DefaultBeanContext.java:1117)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.micronaut.context.DefaultBeanContext.resolveArgumentValues(DefaultBeanContext.java:1116)
	at io.micronaut.context.DefaultBeanContext.doCreateBeanWithArguments(DefaultBeanContext.java:1080)
	at io.micronaut.context.DefaultBeanContext.createBean(DefaultBeanContext.java:1063)
	at io.micronaut.context.BeanContext.createBean(BeanContext.java:157)
	at com.example.Application.run(Application.java:26)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2469)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2461)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2423)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2425)
	at picocli.CommandLine.execute(CommandLine.java:2174)
	at io.micronaut.configuration.picocli.PicocliRunner.run(PicocliRunner.java:146)
	at io.micronaut.configuration.picocli.PicocliRunner.run(PicocliRunner.java:123)
	at com.example.Application.main(Application.java:19)
```

### Steps To Reproduce

1. Create a bean class with a `SequencedSet` parameter:

```java
    @Bean
    static class A {
        public A(@Parameter SequencedSet<Integer> set) {
            System.out.println(set.getClass());
        }
    }
```
Note that with `SequencedSet<Object>` as the argument type, this works fine.

2. Call `BeanContext#createBean` with a `LinkedHashSet`:

```java
context.createBean(A.class, new LinkedHashSet<>());
```

---

I attached a debugger and apparently both the `DefaultBeanContext` and `DefaultMutableConversionService` insist on doing conversion if the object matches the target type but is a `Map` or `Iterable`.

Tangentially, if the parameter in `A` above was changed to `Set`, the bean creation works, but the constructor would get a `HashSet` as its argument instead of the originally passed `LinkedHashSet` (with the same elements though). The same thing happens seemingly for all other `Iterable` and `Map` types: the values in the map or iterable are repackaged in a new instance of potentially different type.
Not sure what the contract here actually is, but intuitively, I would have expected reference equality between the argument passed to `createBean` and the argument passed to the bean constructor (if they're compatible to the argument type)

### Environment Information

- JDK Version: OpenJDK 21.0.2. OpenJDK 25.0.1
- Ubuntu 24.04

### Example Application

_No response_

### Version

4.10.2, 4.10.3-SNAPSHOT, 5.0.0-SNAPSHOT

yawkat:
## Summary
- Preserve constructor arguments in `DefaultBeanContext#createBean` when the runtime value already matches the declared argument type, instead of forcing conversion for iterable/map raw types.
- Add a regression case in `ParametrizedFactorySpec` that reproduces `LinkedHashSet` -> `@Parameter SequencedSet<Integer>` and verifies identity preservation.
- Add `SequencedSetParamBean` test fixture used by the new regression.

## Potential impact
- **Behavioral change scope:** only affects `createBean(...args)` paths where a provided argument is already assignable to the declared parameter type.
- **Positive impact:** avoids unnecessary collection conversion and fixes valid Java 21 `SequencedSet` inputs that previously failed with "Cannot convert object ... to required type: SequencedSet".
- **Compatibility risk:** low-to-moderate; code that implicitly depended on forced conversion side-effects (e.g., wrapping/copying assignable iterable/map instances) will now keep the original instance.
- **Release targeting:** because this is a bug fix with focused scope and regression coverage, it is suitable for `5.0.x`.

## Verification
- `./gradlew :micronaut-inject-java:test --tests io.micronaut.inject.factory.parameterizedfactory.ParametrizedFactorySpec -q --no-daemon --max-workers=1 -Dorg.gradle.warning.mode=none -Dorg.gradle.jvmargs='-Xmx2g -XX:MaxMetaspaceSize=768m'`
- New test in `ParametrizedFactorySpec` passes and reproduces the issue on old logic.

Fixes #12232

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -1058,7 +1058,7 @@ public sealed class DefaultBeanContext implements ConfigurableBeanContext permit
                 if (args.length > i) {
                     Object val = args[i];
                     if (val != null) {
-                        if (argumentType.isInstance(val) && !CollectionUtils.isIterableOrMap(argumentType)) {
+                        if (argumentType.isInstance(val)) {
                             argumentValues.put(requiredArgument.getName(), val);
                         } else {
                             argumentValues.put(requiredArgument.getName(), conversionService.convert(val, requiredArgument).orElseThrow(() ->
--- a/inject-java/src/test/groovy/io/micronaut/inject/factory/parameterizedfactory/ParametrizedFactorySpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/factory/parameterizedfactory/ParametrizedFactorySpec.groovy
@@ -19,6 +19,8 @@ import io.micronaut.context.ApplicationContext
 import io.micronaut.context.exceptions.BeanInstantiationException
 import spock.lang.Specification
 
+import java.util.LinkedHashSet
+
 /**
  * @author Graeme Rocher
  * @since 1.0
@@ -83,4 +85,22 @@ class ParametrizedFactorySpec extends Specification  {
         cleanup:
         context.close()
     }
+    void "test createBean accepts compatible sequenced set parameter"() {
+        given:
+        ApplicationContext context = ApplicationContext.run()
+        def input = new LinkedHashSet<Integer>()
+        input.add(1)
+        input.add(2)
+
+        when:
+        SequencedSetParamBean bean = context.createBean(SequencedSetParamBean, input)
+
+        then:
+        bean != null
+        bean.set.is(input)
+
+        cleanup:
+        context.close()
+    }
+
 }
--- a/inject-java/src/test/groovy/io/micronaut/inject/factory/parameterizedfactory/SequencedSetParamBean.java
+++ b/inject-java/src/test/groovy/io/micronaut/inject/factory/parameterizedfactory/SequencedSetParamBean.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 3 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1058,7 +1058,7 @@ private <T> Map<String, Object> resolveArgumentValues(BeanResolutionContext reso
                 if (args.length > i) {
                     Object val = args[i];
                     if (val != null) {
-                        if (argumentType.isInstance(val) && !CollectionUtils.isIterableOrMap(argumentType)) {
```
graemerocher, line 1061: not sure if this will break other things

yawkat: I made the AI look into git history for that reason, but no test breaks and I didn't find a compelling reason why this check existed

graemerocher (approved):

PR #12481, opened 2026-03-13
Add FileCustomizableResponseType body writer

### LINKED ISSUE #12298: Returning a subclass results in incorrect ResponseBodyWriter being used
### Expected Behavior

Testcase below should work fine.

### Actual Behaviour

Fails because `StreamedFile` isn't serializable. The cause is that the RouteInfo has `NettyJsonHandler` as the body writer and it accepts anything with JSON mime-type, even though the body should be handled by `StreamFileBodyWriter`.

NOTE: The return type is set by micronaut-openapi, we didn't pick it.

### Steps To Reproduce

```java
@Get
public HttpResponse<FileCustomizableResponseType> GET() {
    String serializedResponse = "{\"hello\": \"world\"}";
    InputStream inputStream = new ByteArrayInputStream(serializedResponse.getBytes(StandardCharsets.UTF_8));
    
    return HttpResponse.ok(new StreamedFile(inputStream, MediaType.APPLICATION_JSON));
}
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.9.12

yawkat:
## Summary
- add a `FileCustomizableResponseType` response body writer that delegates to the writer for the runtime subtype
- keep the route-level declared-type writer selection flow intact while avoiding incorrect JSON serialization for `StreamedFile`, `SystemFile`, and custom file response types
- add regression coverage for `HttpResponse<FileCustomizableResponseType>` with JSON media type

## Why
For routes declared as `HttpResponse<FileCustomizableResponseType>`, Micronaut can preselect a writer from the declared type before the runtime subtype is known. With JSON media type, that can lock in the JSON writer and bypass the file-specific handling for `StreamedFile` and `SystemFile`.

This change makes the declared type select a dedicated dispatcher writer for `FileCustomizableResponseType`, and that writer then resolves the actual runtime subtype writer when writing the response.

## Tests
- added `FileTypeHandlerSpec` coverage for:
  - `StreamedFile` returned as `HttpResponse<FileCustomizableResponseType>` with JSON media type
  - `SystemFile` returned as `HttpResponse<FileCustomizableResponseType>` with JSON media type
  - custom `FileCustomizableResponseType` with a custom writer and JSON media type
- fail-before / pass-after was verified for those new cases
- full `FileTypeHandlerSpec` passed locally

## Validation notes
A broader `:micronaut-http-server-netty:test` run was not fully green locally because of an unrelated failure in:
- `io.micronaut.http.server.netty.stream.StreamPressureSpec > consumer pressure`

That failure was:
- `java.io.IOException: Read end dead`

This appears to be outside the file response path touched by this patch.

## Semantics note
This is intentionally narrow in scope: the declared response type still selects the route writer, but for `FileCustomizableResponseType` that selected writer is now a dispatcher that performs runtime subtype selection.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/body/FileCustomizableResponseTypeBodyWriter.java
+++ b/http-server/src/main/java/io/micronaut/http/server/body/FileCustomizableResponseTypeBodyWriter.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.server.body;
+
+import io.micronaut.core.annotation.Experimental;
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.Order;
+import io.micronaut.core.type.Argument;
+import io.micronaut.core.type.MutableHeaders;
+import io.micronaut.http.ByteBodyHttpResponse;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpResponse;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.MutableHttpResponse;
+import io.micronaut.http.body.ByteBodyFactory;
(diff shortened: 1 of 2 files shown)


http-server/src/main/java/io/micronaut/http/server/body/FileCustomizableResponseTypeBodyWriter.java
```diff
@@ -0,0 +1,100 @@
...
+import io.micronaut.http.server.types.files.FileCustomizableResponseType;
+import jakarta.inject.Singleton;
+
+import java.io.OutputStream;
+
+@Internal
+@Experimental
+@Singleton
+@Order(-11)
+public final class FileCustomizableResponseTypeBodyWriter implements ResponseBodyWriter<FileCustomizableResponseType> {
```
graemerocher: does it need to be public? Either way add javadoc for maintainers that include a description of this purpose of this class as summarised in the PR

graemerocher (changes requested):

graemerocher (approved):

PR #12487, opened 2026-03-13
Fix nested each property camelCase config prefix

### LINKED ISSUE #11988: @EachProperty with nested @Configuration only accepts kebab-case
### Expected Behavior

@ConfigurationProperties under @EachProperty should also recognise non kebab case ways

### Actual Behaviour

If we have @EachProperty with nested configuration properties @ConfigurationProperties accepts only kebab-case

While if the nesting is done in a bean which is not part of @EachProperty it works without kebab-case as well.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

https://github.com/anuragagarwal561994/micronaut-issues/tree/main/issue-11988

### Version

4.9.9

yawkat:
$## Summary\n- normalize calculated nested configuration prefixes in `ConfigurationUtils` so runtime bean metadata matches relaxed property naming\n- add a regression test for nested `@ConfigurationProperties("bloomFilter")` under `@EachProperty(list = true)`\n\n## Verification\n- `./gradlew :micronaut-inject-java:test --tests "io.micronaut.inject.foreach.EachPropertyNestedCamelCaseIssue11988Spec" --no-daemon`\n- `./gradlew :micronaut-inject-java:test --tests "io.micronaut.inject.foreach.EachPropertyInterfaceSpec" --no-daemon`\n\n
Fixes #11988

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationUtils.java
+++ b/core-processor/src/main/java/io/micronaut/inject/configuration/ConfigurationUtils.java
@@ -19,6 +19,7 @@ import io.micronaut.context.annotation.ConfigurationReader;
 import io.micronaut.context.annotation.EachProperty;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.naming.NameUtils;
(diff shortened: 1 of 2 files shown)


yawkat:
Not sure if this is a good solution…

PR #12509, opened 2026-03-17
Propagation context - scope values

dstepanov:
- Switching to scoped values propagation by default
- Introducing static and property based switch to use use thread-local propagation
- Deprecating try-resources propagate
- Switched some of the tests that don't support it to thread-local propagation - Coroutines, Reactor


core/src/main/java/io/micronaut/core/propagation/PropagatedContext.java
```diff
@@ -259,11 +251,24 @@ default <V> Supplier<V> wrap(Supplier<V> supplier) {
...
+
+    /**
+     * Executes the given supplier with this context in scope, restoring the previous context when execution completes.
+     *
+     * @param callable The supplier
+     * @param <V>      The supplier return type
+     * @return the result of calling {@link Callable#call()}.
+     * @throws Exception if an exception occurs
+     */
+    <V> V propagateCall(Callable<V> callable) throws Exception;
```
yawkat, line 268: either call this method `propagate` or rename the supplier/runnable methods to match

graemerocher (approved):

core/src/main/java/io/micronaut/core/util/ExceptionUtils.java
```diff
@@ -0,0 +1,55 @@
...
+     * Throws a given {@link Exception} without requiring the caller to handle it explicitly,
+     * bypassing checked exception restrictions at compile time.
+     *
+     * @param <T> the type of the throwable being thrown
+     * @param <R> the return type of the method
+     * @param t the throwable instance to be thrown
+     * @return this method does not return normally, as it always throws the supplied throwable
+     * @throws T the throwable passed as the argument
+     */
+    public static <T extends Exception, R> R sneakyThrow(Exception t) throws T {
```
yawkat: why two methods

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -78,6 +78,7 @@
 import java.time.LocalDateTime;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
```
yawkat: ?

build.gradle
```diff
@@ -48,3 +48,8 @@ configurations.jacocoAggregation.dependencies.removeIf {
 configurations.javadocAggregatorBase.dependencies.removeIf {
     it instanceof ProjectDependency && it.path.startsWith(":test-")
 }
+
+tasks.withType(JavaCompile).configureEach {
+    options.compilerArgs.add("--enable-preview")
```
yawkat: I think this breaks compatibility with non-preview applications.

core/src/main/java/io/micronaut/core/propagation/PropagatedContext.java
dstepanov: Having the same methods brings conflicts. We already have `propagate(Supplier<V> supplier)` so I don't want to make additional breaking changes

yawkat: oh right when you pass a lambda. fair enough.

core/src/main/java/io/micronaut/core/util/ExceptionUtils.java
dstepanov: yeah not sure why I did that

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
dstepanov: Removed

build.gradle
dstepanov: This was from the original PR when I was fighting Java version, not it should be removed

yawkat (approved):

PR #12523, opened 2026-03-18
Fix parsing of fully qualified type identifiers in ctx[...] expressions

### LINKED ISSUE #11410: Expression Language - Getting bean which has `env` in path

yawkat:
## Summary
- Fix `SingleEvaluatedExpressionParser` parsing for fully qualified type identifiers used in `ctx[...]` expressions.
- Add Java and Groovy regression tests covering `ctx[io.micronaut.context.env.Environment]`.
- Addresses the parsing failure reported in #11410.

## Validation
- Passed `./gradlew :micronaut-inject-java:test --tests io.micronaut.expressions.BeanContextAccessExpressionsSpec -q`
- Passed `./gradlew :micronaut-inject-groovy:test --tests io.micronaut.expressions.BeanContextAccessExpressionsSpec -q`
- Broader build validation was attempted, but unrelated failures occurred, and later reruns hit Gradle cache lock contention.
- This PR does not claim a clean full-build result from the current environment.

Fixes #11410

### DIFF
--- a/core-processor/src/main/java/io/micronaut/expressions/parser/SingleEvaluatedExpressionParser.java
+++ b/core-processor/src/main/java/io/micronaut/expressions/parser/SingleEvaluatedExpressionParser.java
@@ -507,16 +507,28 @@ public final class SingleEvaluatedExpressionParser implements EvaluatedExpressio
         }
 
         List<String> parts = new ArrayList<>();
-        parts.add(eat(IDENTIFIER).value());
+        parts.add(typeIdentifierPart().value());
         while (lookahead != null && lookahead.type() == DOT) {
             eat(DOT);
-            parts.add(eat(IDENTIFIER).value());
+            parts.add(typeIdentifierPart().value());
         }
 
         if (wrapped) {
             eat(R_PAREN);
         }
         return new TypeIdentifier(String.join(".", parts));
+    }
+
+    private Token typeIdentifierPart() {
+        if (lookahead == null) {
+            throw new ExpressionParsingException("Unexpected end of input. Expected: 'IDENTIFIER'");
+        }
+        if (!lookahead.type().isOneOf(IDENTIFIER, ENVIRONMENT)) {
+            throw new ExpressionParsingException("Unexpected token: " + lookahead.value() + ". Expected: 'IDENTIFIER'");
+        }
+        Token token = lookahead;
+        lookahead = tokenizer.getNextToken();
+        return token;
     }
 
     // ParenthesizedExpression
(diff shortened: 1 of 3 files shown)


yawkat:
Implementation note for reviewers: the failure happens because `ctx[...]` parses an unwrapped type name via `typeIdentifier(...)`, but the tokenizer reserves `env` as `ENVIRONMENT` globally, so `io.micronaut.context.env.Environment` stops parsing at the `env` package segment. This change keeps tokenization intact and only broadens what counts as an identifier segment while parsing type references.

yawkat:
Updated the branch to tighten the parser change and expand regression coverage:

- narrowed `typeIdentifierPart()` back down to the one extra token we actually need for this issue (`ENVIRONMENT`)
- added coverage for both `ctx[io.micronaut.context.env.Environment]` and `ctx[T(io.micronaut.context.env.Environment)]`

Validation update:
- `:micronaut-inject-java:test --tests io.micronaut.expressions.BeanContextAccessExpressionsSpec` passed with an isolated `GRADLE_USER_HOME`
- the analogous Groovy test path is currently blocked in this environment by an unrelated `:micronaut-buffer-netty:compileJava` failure (`io/micronaut/inject/ast/annotation/ElementAnnotationMetadataFactory`) before it reaches the spec

yawkat:
Validation update: I reran the previously suspicious tasks with an isolated `GRADLE_USER_HOME`, and the earlier failures did not reproduce.

Confirmed passing in the isolated environment:
- `:micronaut-buffer-netty:compileJava`
- `:micronaut-function-web:test --tests io.micronaut.function.web.WebFunctionSpec`
- `:micronaut-context:test --tests io.micronaut.logging.PropertiesLoggingLevelsConfigurerSpec`
- `:micronaut-inject-java:test --tests io.micronaut.expressions.BeanContextAccessExpressionsSpec`

I also reran the broader `spotlessApply check -x japiCmp -x checkVersionCatalogCompatibility` validation with the isolated Gradle home. It progressed substantially further and did not hit the earlier module failures before the local tool timeout; Gradle was still running at that point, so this now looks like a time-budget/environment limitation rather than a reproducible failure caused by this patch.

graemerocher (approved):

PR #12529, opened 2026-03-18
Expose additional Netty HTTP client parser limits

### LINKED ISSUE #11629: Property max-header-size appears to only apply to request headers sent to app
### Expected Behavior

Micronaut allows overriding the max header size via the following property: **micronaut.server.netty.max-header-size**

Documentation description of this property is: "Sets the maximum size of any one header. Default value (8192)."

Expected/hoped behavior was that the value of this property would apply to all http requests and responses managed by the running application.

### Actual Behaviour

My observation is that the value of micronaut.server.netty.max-header-size is being applied only to the request headers sent to the running application, but not the following:
1. response headers returned from the app 
2. request headers sent from the app to another http service
3. response headers returned from another http service to the app

### Steps To Reproduce

To reproduce the behavior described in scenario 1 above:
1. Set the value of micronaut.server.netty.max-header-size = 80
2. Create a simple app that contains a Controller that returns an HttpResponse including a header that is > 80bytes
3. Run the app
4. Via Curl, send request to the app. Response code = 200 OK
5. Conclusion drawn (and also validated via debugging and watching the execution flow) is that the default value of max-header-size (8192) is being used to measure response headers size against, instead of the custom value of max-header-size

To reproduce the behavior described in scenarios 2 and 3 above:
1. Set the value of micronaut.server.netty.max-header-size = 80
2. Create an app that contains a Client that makes a request to another http service, including a header that is > 80bytes.   Also ensure that http service returns a response header that is > 80 bytes.
3. Run the app
4. Via Curl, send request to the app. Response code = 200 OK
5. Conclusion drawn (and also validated via debugging and watching the execution flow) is that the default value of max-header-size (8192) is being used to measure headers sent to and received from another http service, instead of the custom value of max-header-size

### Environment Information

JDK 11

### Example Application

_No response_

### Version

4.7.14

yawkat:
## Summary
- expose `micronaut.http.client.max-initial-line-length`
- expose `micronaut.http.client.max-chunk-size`
- wire both properties through all `HttpClientCodec` construction sites and cover binding in config tests

## Context
Micronaut already exposed the server-side HTTP/1 parser limits and the client `max-header-size`, but the client still hardcoded Netty's initial-line and chunk-size limits. This change makes those existing Netty HTTP/1 parser knobs configurable on the client side as well.

## Verification
- `./gradlew :micronaut-http-client-core:compileJava :micronaut-http-client:test --tests 'io.micronaut.http.client.config.DefaultHttpClientConfigurationSpec' -q`

Fixes #11629

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -170,6 +170,10 @@ public abstract class HttpClientConfiguration {
 
     private int maxHeaderSize = DEFAULT_MAX_HEADER_SIZE;
 
+    private int maxInitialLineLength = DEFAULT_MAX_INITIAL_LINE_LENGTH;
+
+    private int maxChunkSize = DEFAULT_MAX_CHUNK_SIZE;
+
     private Proxy.Type proxyType = Proxy.Type.DIRECT;
 
     @Nullable
@@ -256,6 +260,8 @@ public abstract class HttpClientConfiguration {
             this.loggerName = copy.loggerName;
             this.maxContentLength = copy.maxContentLength;
             this.maxHeaderSize = copy.maxHeaderSize;
+            this.maxInitialLineLength = copy.maxInitialLineLength;
+            this.maxChunkSize = copy.maxChunkSize;
             this.proxyAddress = copy.proxyAddress;
             this.proxyPassword = copy.proxyPassword;
             this.proxySelector = copy.proxySelector;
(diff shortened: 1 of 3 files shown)


http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
```diff
@@ -698,6 +722,24 @@ public void setMaxHeaderSize(@ReadableBytes int maxHeaderSize) {
...
+    public void setMaxInitialLineLength(@ReadableBytes int maxInitialLineLength) {
+        this.maxInitialLineLength = maxInitialLineLength;
+    }
+
+    /**
+     * Sets the maximum chunk size the client can handle. Default value ({@value io.micronaut.http.client.HttpClientConfiguration#DEFAULT_MAX_CHUNK_SIZE}).
+     *
+     * @param maxChunkSize The maximum chunk size the client can handle
+     */
+    public void setMaxChunkSize(@ReadableBytes int maxChunkSize) {
```
graemerocher, line 743: add `@since` declarations to the javadoc

graemerocher (approved):

http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
yawkat: Added `@since 5.0.1` tags to the newly introduced `maxInitialLineLength` / `maxChunkSize` accessors and mutators, and re-ran the focused client-core/client verification command.

PR #12535, opened 2026-03-19
Allow KSP visitors to traverse Kotlin inner classes

### LINKED ISSUE #10941: Micronaut is unable to inject dependencies into nested tests when using KSP
### Expected Behavior

If a @MicronautTest test and its injected dependencies are within a nested class, the dependencies are injected when the test in the nested class runs.

### Actual Behaviour

The dependencies are not injected, leading to the following error: `kotlin.UninitializedPropertyAccessException: lateinit property has not been initialized`

### Steps To Reproduce

Reproduction steps are available in the README of the example application below, but generally:

1.  Create a test class that uses the @MicronautTest annotation
2. Create a nested inner class (with the @Nested annotation) containing dependencies injected using the @Inject annotation
3. Create a test within the inner class that uses the injected dependencies.
4. The test will fail, but will pass if you move everything outside of the nested inner class.

### Environment Information

- Kotlin 2.0
- JDK 17

### Example Application

https://github.com/cainecridland/micronaut-nested-test-injection-problem-example

### Version

4.4.0

yawkat:
## Summary
- allow KSP visitor traversal of public Kotlin inner classes
- remove the static-only restriction that skipped Kotlin `inner` nested test classes
- keep bean-definition recursion behavior unchanged

## Context
Issue #10941 reports `@MicronautTest` dependency injection failing for Kotlin JUnit `@Nested` inner classes under KSP. The KSP traversal in `TypeElementSymbolProcessor` filtered inner classes with `onlyStatic()`, which excludes Kotlin `inner` classes by definition.

This change removes that static-only filter from visitor traversal while preserving the existing `PUBLIC` filter. `BeanDefinitionProcessor` still excludes `Modifier.INNER`, so this does not enable inner-class bean generation.

## Verification
- reproduced the issue as KSP-only on 4.10.x, with equivalent KAPT behavior passing
- confirmed the same problematic traversal logic exists on 5.0.x
- applied the minimal 5.0.x patch in `inject-kotlin`

Fixes #10941

### DIFF
--- a/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/beans/BeanDefinitionProcessor.kt
+++ b/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/beans/BeanDefinitionProcessor.kt
@@ -83,7 +83,7 @@ internal class BeanDefinitionProcessor(private val environment: SymbolProcessorE
                                 isVetoed(ksAnnotation)
                             }
                         }
-                        .filter { !it.modifiers.contains(Modifier.INNER) }
+                        .filter { declaration -> declaration.classKind == ClassKind.INTERFACE || !declaration.modifiers.contains(Modifier.INNER) || declaration.annotations.any { annotation -> annotation.annotationType.resolve().declaration.qualifiedName?.asString() == "org.junit.jupiter.api.Nested" } }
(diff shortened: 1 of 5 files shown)


dstepanov:
Can you add a test?

PR #12540, opened 2026-03-20
Prevent direct bean introspection instantiation for abstract types

### LINKED ISSUE #10889: Instantiation methods in introspection of abstract class tries to instantiate the abstract class

yawkat:
## Summary
- prevent direct constructor-based bean introspection instantiation for abstract classes and interfaces
- preserve creator-backed instantiation paths by allowing static `@Creator` methods
- add a regression test for abstract `@Introspected` classes

## Verification
- ran targeted `BeanIntrospectionSpec` regressions covering abstract classes, plain interfaces, and interface static creator methods
- reran `BeanIntrospectionSpec` after refining the generation rule

Fixes #10889

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
@@ -550,6 +550,8 @@ final class BeanIntrospectionWriter implements OriginatingElements, ClassOutputW
 
     private byte[] generateIntrospectionClass() {
         boolean isEnum = beanClassElement.isEnum();
+        boolean hasStaticCreator = constructor != null && constructor.isStatic();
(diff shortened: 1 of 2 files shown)


yawkat:
AI keeps failing, I don't want to spend more time on this.

PR #12543, opened 2026-03-20
Fix constructor-order check for proxied factory beans

### LINKED ISSUE #10834: Reordering constructors breaks intercepted factory bean compilation
### Expected Behavior

When compiling a factory that creates a bean with 2 constructors, it wrongly fails

Given the factory (not `@NotNull` is important for interceptor creation, which is another bug #10836)
```java
@Factory
public class BrokenFactory {
    @Bean
    public @NotNull TwoConstructorClass twoConstructorClass() {
        return new TwoConstructorClass();
    }
}
```

And `TwoConstructorClass` being:
```java
public class TwoConstructorClass {
    public TwoConstructorClass(String age) {
    }
    public TwoConstructorClass() {
        this("10");
    }
}
```
I would expect compilation to work

### Actual Behaviour

Compilation fails with
```
error: The produced type from a factory which has AOP proxy advice specified must define an accessible no arguments constructor. Proxying types with constructor arguments can lead to unexpected behaviour. See the javadoc for for Around.ProxyTargetConstructorMode for more information and possible solutions: classCallingInConstructor
    public @NotNull TwoConstructorClass classCallingInConstructor() {
```

but (only) when switching the constructors (so that the parameterless one is first) the error goes away
```
public class TwoConstructorClass {
    public TwoConstructorClass() {
        this("10");
    }
    public TwoConstructorClass(String age) {
    }
}
```

### Steps To Reproduce

1. `git clone https://github.com/nbrugger-tgm/micronaut-reproductions`
2. `cd micronaut-reproductions/micronaut-4-factory-regression-2`
3. `gradle classes`

### Environment Information

- OS: Kubuntu
- JDK: 17

### Example Application

https://github.com/nbrugger-tgm/micronaut-reproductions

### Version

4.4.2

yawkat:
## Summary
- validate proxied factory-produced beans against the presence of an accessible default constructor instead of the selected primary constructor
- add Java and Groovy regression coverage for classes that declare a parameterized constructor before an accessible no-arg constructor
- preserve the existing proxy writer constructor visitation path while removing the false constructor-order dependency

## Verification
- attempted targeted runs for `FactoryWithScopedProxySpec` in `micronaut-inject-java` and `micronaut-inject-groovy`
- verification in this environment is blocked by a pre-existing unrelated compile failure in `core/src/main/java/io/micronaut/core/propagation/ScopedValues.java` (`ScopedValue` preview/toolchain issue), so those targeted tests could not be executed to completion here

Resolves #10834

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/processing/FactoryBeanElementCreator.java
+++ b/core-processor/src/main/java/io/micronaut/inject/processing/FactoryBeanElementCreator.java
@@ -212,14 +212,19 @@ final class FactoryBeanElementCreator extends DeclaredBeanElementCreator {
             }
 
             MethodElement constructorElement = producedType.getPrimaryConstructor().orElse(null);
-            if (!producedType.isInterface() && constructorElement != null && constructorElement.getParameters().length > 0) {
+            MethodElement defaultConstructor = producedType.getDefaultConstructor().orElse(null);
+            if (!producedType.isInterface() && defaultConstructor == null) {
                 final String proxyTargetMode = producedAnnotationMetadata.stringValue(AnnotationUtil.ANN_AROUND, "proxyTargetMode").orElse("ERROR");
                 switch (proxyTargetMode) {
                     case "ALLOW":
-                        allowProxyConstruction(constructorElement);
+                        if (constructorElement != null) {
+                            allowProxyConstruction(constructorElement);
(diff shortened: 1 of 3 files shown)


yawkat:
not sure if this makes sense

graemerocher (approved):

PR #12557, opened 2026-03-23
Fix unqualified @Replaces matching qualified beans

### LINKED ISSUE #10506: Micronaut mixes up a qualified bean from unqualified one when @Replace is used
### Expected Behavior

When a factory provides a qualified bean and an unqualified bean, Micronaut can't resolve the qualified one.

Example:
```
@Factory
@Prototype
public class MyFactory {

  @Named("blah")
  @Prototype
  CountDownLatch blah() {
    return new CountDownLatch(1);
  }

  @Singleton
  @Replaces(CountDownLatch.class)
  CountDownLatch foo() {
    return new CountDownLatch(1);
  }
}

@Prototype
public class MyClass {

  @Inject
  public MyClass(@Named("blah") CountDownLatch latch) { }
}
```

Note there's a Micronaut provided `@Factory`, which is why `@Replaces` is used:
```
@Factory
public class MicronautFactory {
  @Bean
  CountDownLatch foo() {
    return new CountDownLatch(7);  //  7 is really bad for my app :-)
  }
}
```

This should work but does not.

### Actual Behaviour

Even though `@Replaces` is not qualified, it eliminates the qualified bean, so this code will not run and complain:
`No bean of type [java.util.concurrent.CountDownLatch] exists for the given qualifier: @Named('blah')`

Same thing with a qualifier annotation.

### Steps To Reproduce

See above code snippet.

### Environment Information

JDK 21 / Linux

### Example Application

_No response_

### Version

3.10.1

yawkat:
## Summary
- prevent unqualified `@Replaces` from matching bean definitions with declared qualifiers
- add a regression test covering a named bean surviving an unqualified replacement

## Verification
- `./gradlew :micronaut-inject-java:test --tests 'io.micronaut.inject.qualifiers.replaces.ReplacesSpec'`
- `./gradlew :micronaut-inject:test --tests 'io.micronaut.context.DefaultBeanContextSpec'`

Resolves #10506

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/qualifiers/replaces/ReplacesSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/qualifiers/replaces/ReplacesSpec.groovy
@@ -17,6 +17,8 @@ package io.micronaut.inject.qualifiers.replaces
 
 import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
 import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Prototype
+import io.micronaut.inject.qualifiers.Qualifiers
 import spock.lang.Specification
 import io.micronaut.context.env.PropertySource
 /**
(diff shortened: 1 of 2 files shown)


yawkat:
not happy with the fix

PR #12570, opened 2026-03-26
fix: flush HTTP/2 decompressor when END_STREAM arrives on trailing HEADERS

adytzu2007:
When an HTTP/2 request ends with trailing HEADERS carrying END_STREAM (e.g. from Envoy decompressor), the DelegatingDecompressorFrameListener never receives onDataRead(..., endOfStream=true) and so never flushes its decompressor tail bytes. This causes requests to either hang (uncompressed) or complete with a truncated body (compressed).

Add Http2TrailerFlushListener, a frame listener decorator that sits between the decompressor and Http2ServerHandler. When it detects trailing HEADERS with END_STREAM, it routes the signal through the onDataRead path, ensuring the decompressor flushes before the request body is completed.


yawkat:
Why a separate listener? Seems like a hack

adytzu2007:
> Why a separate listener? Seems like a hack

Agreed but not sure where else I could add this. This needs to run before the decompressor but it also needs to run for non-compressed requests (my actual bug was that envoy decompresses the request then adds the end of stream on a trailer).

If you think there's a better way, can you point me in that direction?

Issue #12326, opened 2026-01-12
Suspend functions ignore `ExecuteOn` and use the event loop instead


jtgeorge3:
### Expected Behavior

When a `suspend` controller function is annotated with `@ExecuteOn(TaskExecutors.IO)`, it should execute on threads from IO executor pool, and thus keep blocking operations off the event loop. 

Maintaining this behavior from Micronaut 3 is important because some some downstream workflows of `suspend`-able controller methods have both non-blocking and blocking operations (e.g. async HTTP plus JDBC calls in the same call stack). So being able to explicitly select a blockable thread pool is important for ensuring no performance degradations during a Micronaut 4 migration upgrades

### Actual Behaviour

`suspend` controller functions ignore the `@ExecuteOn` annotation and use the default event loop. 
In fact, removing the `suspend` keyword results in better performance with blocking behavior, as developers can easily switch to other pools. This makes it dangerous to add `suspend` to controllers, and adds a significant layer of complexity and confusion when upgrading from Micronaut 3 to 4

### Steps To Reproduce

Clone [this git repo](https://github.com/jtgeorge3/micronaut-4-executeon-bug/blob/main/README.md) and follow the README

OR

1. Use Micronaut 4 (any version SHOULD work, but tested extensively with 4.10.6)
2. Set `thread-selection: BLOCKING` to use the blocking pool by default
3. Add `suspend` to a controller method that has a blocking operation (`Thread.sleep`, jdbc call, etc)
4. Mark the controller method with `ExecuteOn` and use a blocking thread pool of your choice (i.e. `TaskExecutors.BLOCKING/IO`)
5. Use a separate service to make concurrent calls to that endpoint (simulating realistic API traffic)
6. Simply remove the `suspend` keyword and try again
7. You'll see that the concurrent requests are handled more quickly after removing `suspend`

### Environment Information

- MacOS Tahoe 26.2
- openjdk 21.0.9 2025-10-21
- Gradle 8.12
- Kotlin 2.1.10

### Example Application

https://github.com/jtgeorge3/micronaut-4-executeon-bug

### Version

4.10.6

jtgeorge3:
I added [a branch for a Micronaut 3 test](https://github.com/jtgeorge3/micronaut-4-executeon-bug/tree/micronaut-3) to give a clear comparison of 3 vs 4 behaviors. 

Probably the clearest expression of the issue is found in comparing the load tests:

## Micronaut 3 - all load tests take around 1s
[...]

PR #12577, opened 2026-03-26
Fix @ExecuteOn ignored for Kotlin suspend functions

jpbempel:
When a Kotlin controller method is a suspend function and annotated with `@ExecuteOn(TaskExecutors.IO)`, the handler runs on the Netty event loop thread instead of the IO thread pool. The same annotation works correctly for Java methods and Kotlin non-suspend methods.

In `RouteExecutor.callRoute()`, the `executorService` retrieved from `@ExecuteOn` is correctly obtained but silently ignored in the `isSuspended()` branch. Also, `ContinuationArgumentBinder.setupCoroutineContext` hardcodes `Dispatchers.Default` as the coroutine dispatcher, so post-suspension resumptions always run on the default pool.

To fix we are passing the `executorService` to an overloaded version of `setupCoroutineContext` and call `suscribeOn(scheduler)`

we are fixing also the scoped-value propagation mode that throw an `IllegalStateException` because of a case missing.

### DIFF
--- a/core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
+++ b/core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
@@ -214,7 +214,10 @@ final class PropagatedContextImpl implements PropagatedContext {
     @Nullable
     public static PropagatedContext getOrNull() {
         return switch (PropagatedContextConfiguration.get()) {
-            case SCOPED_VALUE -> ScopedValues.get();
+            case SCOPED_VALUE -> {
+                PropagatedContext fromScopedValue = ScopedValues.get();
+                yield fromScopedValue != null ? fromScopedValue : ThreadContext.get();
+            }
             case THREAD_LOCAL -> ThreadContext.get();
         };
     }
@@ -238,11 +241,7 @@ final class PropagatedContextImpl implements PropagatedContext {
 
     @Override
     public Scope propagate() {
-        return switch (PropagatedContextConfiguration.get()) {
-            case THREAD_LOCAL -> ThreadContext.propagate(ThreadContext.get(), this);
-            case SCOPED_VALUE ->
-                throw new IllegalStateException("Scope propagation requires thread-local support. Set 'micronaut.propagation' to 'thread-local'.");
-        };
+        return ThreadContext.propagate(ThreadContext.get(), this);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/micronaut/core/propagation/PropagatedContextImpl.java
```diff
@@ -214,7 +214,10 @@ public boolean isEmpty() {
     @Nullable
     public static PropagatedContext getOrNull() {
         return switch (PropagatedContextConfiguration.get()) {
-            case SCOPED_VALUE -> ScopedValues.get();
+            case SCOPED_VALUE -> {
+                PropagatedContext fromScopedValue = ScopedValues.get();
+                yield fromScopedValue != null ? fromScopedValue : ThreadContext.get();
```
dstepanov, line 219: We don't want to mix scoped value and thread local

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/server/suspend/ScopedValueCoroutineFailureTest.kt
```diff
@@ -27,17 +22,9 @@ class ScopedValueCoroutineFailureTest {
...
-            client.toBlocking().retrieve(HttpRequest.GET<Any>("/coroutine/failure"))
-        }
-        assertEquals(500, exception.status.code)
-        val details = buildString {
-            exception.response.getBody(String::class.java).ifPresent { append(it) }
-            exception.cause?.message?.let { append(it) }
-            append(exception.message)
-        }
-        assertTrue(details.contains("Scope propagation requires thread-local support"))
+    fun `suspend endpoint works when propagation mode is scoped value`() {
```
dstepanov, line 25: Please don't modify existing tests

http-server/src/main/java/io/micronaut/http/server/RouteExecutor.java
```diff
@@ -434,27 +434,28 @@ ExecutionFlow<HttpResponse<?>> callRoute(PropagatedContext propagatedContext, Ro
...
         if (executorService != null) {
             if (routeInfo.isSuspended()) {
+                Scheduler scheduler = Schedulers.fromExecutor(executorService);
                 executeMethodResponseFlow = ReactiveExecutionFlow.fromPublisher(Mono.deferContextual(contextView -> {
                         return Mono.from(
-                            ReactiveExecutionFlow.fromFlow(executeRouteAndConvertBody(propagatedContext, routeMatch, request, true, contextView)).toPublisher()
+                            ReactiveExecutionFlow.fromFlow(executeRouteAndConvertBody(propagatedContext, routeMatch, request, true, contextView, executorService)).toPublisher()
                         );
-                    }));
+                    }).subscribeOn(scheduler));
```
dstepanov, line 442: Why do you change threads twice? Once here and once in the setup method?

dstepanov (changes requested):

jpbempel:
Closing this for 5.x
the handling of Scoped Value seems complicated for this case.
fixing it fr 4.10.x branch instead in #12594

PR #12579, opened 2026-03-27
ensure consistent use of context class loader

graemerocher:
If an application is loaded with the alternative context loader there are cases where this context loader is not used. This addresses those cases.


graemerocher:
@copilot apply changes based on the comments in [this thread](https://github.com/micronaut-projects/micronaut-core/pull/12579#pullrequestreview-4021549698)

PR #12593, opened 2026-04-01
Improve Config import

dstepanov:


inject/src/main/java/io/micronaut/context/DefaultApplicationContext.java
```diff
@@ -746,7 +746,7 @@ private static final class BootstrapPropertySourcesLocator implements PropertySo
 
         private final ApplicationContextConfiguration bootstrapConfiguration;
         private final ApplicationContext parentContext;
-        private @Nullable ApplicationContext bootstrapContext;
+        private final List<ApplicationContext> bootstrapContexts = new ArrayList<>();
```
graemerocher, line 749: why was it necessary to have multiple bootstrap contexts now when we are trying to get rid of the one we have?

dstepanov: because now we store property source loaders in DefaultEnv, and before we were always access them from the configuration which will always create a new bootstrap

PR #12594, opened 2026-04-02
Fix @ExecuteOn ignored for suspend functions

jpbempel:
When a Kotlin controller method is a suspend function and annotated with @ExecuteOn(TaskExecutors.IO), the handler runs on the Netty event
 loop thread instead of the IO thread pool. The same annotation works
 correctly for Java methods and Kotlin non-suspend methods.

In RouteExecutor.callRoute(), the executorService retrieved from @ExecuteOn is correctly obtained but silently ignored in the isSuspended() branch. Also,
ContinuationArgumentBinder.setupCoroutineContext hardcodes Dispatchers.Default as the coroutine dispatcher, so post-suspension resumptions always run on the default pool.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
+++ b/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
@@ -25,6 +25,7 @@ import jakarta.inject.Singleton;
 import reactor.util.context.ContextView;
 
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 
 /**
  * Coroutines helper.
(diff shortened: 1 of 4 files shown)


graemerocher:
seems to break a test

dstepanov:
Thanks for this @jpbempel, and sorry it sat for so long — your diagnosis was right and it's the basis of the fix that landed.

#13032 is now merged into `5.2.x`, with you credited as a co-author on the commit. On top of your change it caches the coroutine dispatcher per executor (dispatcher equality is identity-based, so building one per request allocates on the request path and defeats `withContext` short-circuiting), builds it from the raw executor rather than the `ContextPropagatingExecutorService` wrapper, and documents the `thread-selection` behaviour change in the breaking-changes appendix.

Closing this one as superseded. Note the fix landed on `5.2.x` rather than `4.10.x` — partly because making a suspend route obey `AUTO`/`IO`/`BLOCKING` thread selection is a runtime behaviour change that's easier to justify on a minor line than in a patch.

PR #12605, opened 2026-04-08
Change propagated context default to thread local

graemerocher:


sdelamo (approved):

sdelamo:
@graemerocher CI fails

PR #12614, opened 2026-04-17
Change default for CrossOrigin.allowCredentials

yawkat:
Defaulting to `true` may be surprising to users, especially those switching from spring where the default is `false`.


sdelamo (approved):

yawkat:
still some test failures

PR #12627, opened 2026-04-21
fix(cors): replace reference equality with value equality in isAny and isAnyMethod

### LINKED ISSUE #12626: CorsFilter#isAny and isAnyMethod use reference equality (==) instead of value equality, causing CORS to fail when configured via application.yml
### Expected Behavior

CORS preflight (OPTIONS) requests succeed when `allowed-headers` or `allowed-methods` 
are configured with wildcard values in application.yml.

### Actual Behaviour

CORS preflight requests fail because `isAny(allowedHeaders)` and `isAnyMethod(allowedMethods)` 
always return false when the configuration is loaded from application.yml, even when 
allowed-headers is set to ["*"] or allowed-methods is set to [].

The root cause is that both methods use reference equality (==) to compare against 
the sentinel constants CorsOriginConfiguration.ANY and CorsOriginConfiguration.ANY_METHOD:

    private static boolean isAny(List<String> values) {
        return values == CorsOriginConfiguration.ANY; // always false after deserialization
    }

    private static boolean isAnyMethod(List<HttpMethod> allowedMethods) {
        return allowedMethods == CorsOriginConfiguration.ANY_METHOD; // always false after deserialization
    }

When application.yml is deserialized, new List instances are created that are never 
reference-equal to the constants, so the wildcard check silently fails.

### Steps To Reproduce

1. Add CORS configuration to application.yml:
```yaml
   micronaut:
     server:
       cors:
         enabled: true
         configurations:
           default:
             allowed-headers:
               - "*"
```
2. Define any POST endpoint, e.g. `@Post("/api/login")`

3. Send a CORS preflight request:
   curl -X OPTIONS http://localhost:8080/api/login \
     -H "Origin: http://localhost:3000" \
     -H "Access-Control-Request-Method: POST" \
     -H "Access-Control-Request-Headers: Content-Type"

4. Observe the request is rejected by CORS filter instead of returning 200

### Environment Information

- Java Version: 21

### Example Application

_No response_

### Version

4.10.12

fatekingsama:
## Summary
Fixes #12626

## Changes
Replace reference equality (`==`) with value equality in `CorsFilter#isAny` 
and `isAnyMethod`.

When CORS configuration is loaded from `application.yml`, deserialization 
creates new `List` instances that are never reference-equal to the sentinel 
constants `CorsOriginConfiguration.ANY` and `ANY_METHOD`, causing wildcard 
checks to silently return false.

## Fix
- `isAny`: also returns true when the list contains a single element equal to 
  `CorsOriginConfiguration.ANY.getFirst()`
- `isAnyMethod`: also returns true when the list is empty


http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -419,11 +419,13 @@ private static boolean matchesOrigin(String originRegex, String requestOrigin) {
     }
 
     private static boolean isAny(List<String> values) {
-        return values == CorsOriginConfiguration.ANY;
+        return values == CorsOriginConfiguration.ANY
+            || (values != null && values.size() == 1 && values.getFirst().equals(CorsOriginConfiguration.ANY.getFirst()));
```
yawkat: just use Objects.equals here

yawkat:
Please add a test

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -42,30 +36,17 @@
...
+import org.jspecify.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
```
sdelamo: don't use wildcard imports

sdelamo (changes requested):

fatekingsama:
This was caused by IntelliJ automatically optimizing imports. I've updated the configuration and replaced wildcard imports with explicit ones.

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -419,11 +409,12 @@ private static boolean matchesOrigin(String originRegex, String requestOrigin) {
     }
 
     private static boolean isAny(List<String> values) {
-        return values == CorsOriginConfiguration.ANY;
+        return Objects.equals(values, CorsOriginConfiguration.ANY);
     }
 
     private static boolean isAnyMethod(List<HttpMethod> allowedMethods) {
-        return allowedMethods == CorsOriginConfiguration.ANY_METHOD;
+        return allowedMethods == CorsOriginConfiguration.ANY_METHOD
```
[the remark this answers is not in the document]
yawkat: just use Objects.equals for the entire method

fatekingsama:
Friendly ping 👋 @sdelamo  @yawkat 
Would appreciate another review when you have a chance. Thanks!

yawkat (approved):

fatekingsama:
Hi @sdelamo , just following up on this one. The latest changes were approved on Jun 8, and this PR is still awaiting your review. Since Micronaut Core has moved on to 5.1.2, could you please take another look when you have a chance?

If anything else is needed from my side, I’m happy to update it. Thanks!

fatekingsama:
Hi @sdelamo , all checks are green now—could you please take another look when you have a chance? Thanks!

fatekingsama:
Hi @graemerocher , sorry to ping you directly. This PR is still awaiting a follow-up review after the latest changes, and all checks are green. Should I retarget it from 5.0.x to 5.1.x, or is 5.0.x still the appropriate target? Could you please advise whether anything else is needed and when it might be reviewed? Thanks!

fatekingsama:
Hi @sdelamo, thanks again for merging this fix into `5.0.x`.

I noticed that `5.1.8`, `5.1.9`, and `5.1.10` have since been released, but this fix does not appear to have been forward-merged into `5.1.x` yet.

Is another `5.0.x` → `5.1.x` merge planned, or would you prefer that I open a separate PR targeting `5.1.x`?

Thanks!

sdelamo:
@fatekingsama I will check if it has not been merged up and trigger a release if it has not.

PR #12628, opened 2026-04-21
Avoid iterating lazy property sources for config imports

graemerocher:
## Summary
- skip config import normalization for non-Map property sources
- avoid touching lazy test-resources property sources during config import discovery
- add a regression test that proves neither get() nor iterator() is called

## Verification
- ./gradlew :micronaut-inject:test --tests io.micronaut.context.env.ConfigImportDeclarationsTest


dstepanov:
The fix looks like way specific. Can you explain the reasoning?

graemerocher:
I am going to change the fix slightly, the issue is that the current impl breaks test resources

graemerocher:
I altered the PR, so basically the test resources is broken without this change since it iterates the property sources and calls get(..) on them which triggers early init of test resource resolvers. The issue is complex but if we want test resources not to be broken we need this change

sdelamo (approved):

PR #12629, opened 2026-04-22
Async Http Client

dstepanov:
Async Http client that can be used without any kind of reactive framework.


graemerocher:
address the copilot feedback and get CI passing

graemerocher (approved):
fix sonar before merge

PR #12632, opened 2026-04-24
Instantiate introspected beans only when request values are present

### LINKED ISSUE #12608: Introspected types are always instantiated when binding from request
### Expected Behavior

Given the following Controller and introspected type:
```java
@Controller
public class MyController {

  @Introspected
  public record MyPojo() {}

  @Get("/query")
  public void queryValueNotNull(@QueryValue @NotNull MyPojo queryValue) {}


  
  @Introspected
  public record MyRequestBean(@Body @NotNull MyPojo body) {}

  @Get("/body")
  public void bodyNotNull(@RequestBean MyRequestBean requestBean) {}
}
```

When requesting `GET /query` or `GET /body`, without sending query values or a body, I would expect the server to respond with `HTTP 400 - Bad Request` since the `@NotNull` constraints should be violated.

This was the behavior **before** updating to Micronaut `4.10.0`.

### Actual Behaviour

After updating to Micronaut `4.10.0` the server unexpectedly responds with `HTTP 200 - OK` in both cases.

Because `MyPojo` is annotated with `@Introspected`, it will always be instantiated and is never `null`.

### Steps To Reproduce

1. Have a type annotated with `@Introspected`
2. Use this type to bind as `@QueryValue` (or `@Body` when wrapped inside a `@RequestBean`)
3. Request this endpoint without providing any further information
4. Observe that the bound value is never `null`

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.0

iamsb97:
Fixes #12608.

Avoid instantiating `@Introspected` beans for `@QueryValue` and `@RequestBean` when no request values are present. Return null instead of an empty object.

Existing behavior for required arguments (e.g. HTTP 400) is unchanged; this applies when the argument is treated as optional/nullable.

This change currently covers `@QueryValue`, `@RequestBean` and `@Body` as specified in the issue. Please let me know if there are other annotations that should follow the same behavior.


iamsb97:
@dstepanov, please review this PR when you get a chance.

PR #12663, opened 2026-05-05
Raw generic match argument qualifier

dstepanov:
Adding the fix from https://github.com/micronaut-projects/micronaut-serialization/pull/1279


dstepanov:
@copilot Fix checkstyle

graemerocher (approved):

PR #12636, opened 2026-04-28
Introduce Introspected.Property annotation to support explicit bean properties

dstepanov:
- Added @Introspected.Property annotation with external name and read/write access support.
  - Mapped Jackson @JsonProperty, @JsonGetter, and @JsonSetter to @Introspected.Property.

Fixes https://github.com/micronaut-projects/micronaut-serialization/issues/691

PR #12668, opened 2026-05-11
Jakarta peristence Access annotation visitor

ZARAYACH:


graemerocher:
this likely belongs in the micronaut data repo

ZARAYACH:
> this likely belongs in the micronaut data repo

So this pr is based on the Pr that introduces @introspected.Property [#12636](https://github.com/micronaut-projects/micronaut-core/pull/12636) which also maps the Jakarta.persistence.Access annotation but it doesn't do that correctly cause The access can be targeted to type/method/field but it's always mapped to the Introcpected.Property which only is applicable to method/field so i'm not sure if that mapping as whole should be moved to the data repo ?

graemerocher:
ok I see, will let @dstepanov comment

graemerocher:
this overlaps with https://github.com/micronaut-projects/micronaut-core/pull/12636

PR #12671, opened 2026-05-11
Optimize DateTimeFormatter cache with Lock-free reads and FIFO eviction

altro3:
## Summary
This PR optimizes DateTimeFormatter caching in TimeConverterRegistrar by replacing the heavy LRU approach with a high-performance Fixed-size Ring Buffer (FIFO).
## Problem
Current LRU implementations (like ConcurrentLinkedHashMap) require a "write-on-read" operation to update access order on every get() call. This causes cache-line contention and unnecessary memory overhead for metadata.
## Solution: Lock-free FIFO
I introduced a lightweight FIFO eviction using AtomicReferenceArray and a simple incrementing cursor. This ensures the "hot path" is a pure read operation.
## Comparison (for 100 entries)

* Speed (Throughput):
* Old (LRU): ~50-80ns per lookup (due to write-on-read metadata updates).
   * New (FIFO): ~30-40ns per lookup. (approx. 2x faster on hits).
* Structural Memory Overhead (Metadata only):
* Old (LRU): ~4,000 bytes (for Node objects, prev/next pointers in the linked list).
   * New (FIFO): ~440 bytes (array of references + atomic primitives). (~90% less overhead).
* Concurrency:
* Old (LRU): Constant CAS/Lock contention to reorder elements on every hit.
   * New (FIFO): Zero contention on hits. Atomic updates only on cache misses.
   
## Efficiency for real-world usage (e.g., 10 entries)
While the limit is 100, most applications use fewer than 10 unique locale/pattern combinations. In this typical scenario, the overhead difference is even more significant:

* Structural Latency:
* Old (LRU): ~50ns (requires event recording and metadata updates on every hit).
   * New (FIFO): ~15ns. Pure L1-cache hit with zero state updates on hits. ~3x faster.
* Metadata Memory Footprint:
* Old (LRU): ~500 bytes (Node objects, pointers, and event buffers).
   * New (FIFO): ~72 bytes (Small array + atomic primitives). ~7x less memory.
* CPU Cache Locality:
* Old (LRU): Metadata is scattered across the heap (multiple objects).
   * New (FIFO): Metadata fits entirely within a single 64-byte CPU cache line, ensuring maximum hardware efficiency.

## Implementation Details

* Uses AtomicReferenceArray<String> evictionQueue for safe visibility.
* Uses (evictionIndex.getAndIncrement() & Integer.MAX_VALUE) % MAX_CACHE_SIZE to safely handle integer overflows and ensure a positive index.
* Retains original MAX_CACHE_SIZE and method signatures.


sdelamo:
@altro3 target 5.1.x please.

altro3:
@sdelamo Created new PR here: https://github.com/micronaut-projects/micronaut-core/pull/12674

PR #12678, opened 2026-05-14
Add CDI integration hooks

graemerocher:
## Summary

This PR adds a small set of CDI-oriented extension hooks needed by the ODI CDI Lite implementation without making the existing internal APIs public again.

The CDI Lite integration needs to reuse Micronaut's bean resolution, generated bean definitions, generated proxies, and annotation metadata model, but CDI has a few observable semantics that cannot be implemented cleanly from outside the context today:

- CDI normal scoped beans use client proxies, so target resolution has to be delayed and the target resolution path has to be preserved separately from proxy construction.
- CDI dependent objects created during a resolution need integration-controlled destruction behavior.
- CDI producer/null-bean behavior needs a fallback path when a resolved Micronaut bean definition produces `null`.
- CDI assigns candidate compatibility and array injection semantics differently in a few cases.
- CDI extensions need compile-time access to annotation metadata details, such as defaults and retention policy, without using reflection.

The public API added here is intentionally narrow and experimental. `BeanResolutionCustomizer` is installed through `ApplicationContextBuilder` / `BeanContextConfiguration`, has default no-op behavior for all methods, and is annotated with `@Experimental`. It is not a functional interface so future bean resolution customization points can be added without replacing the type.

## Changes

- Added `BeanResolutionCustomizer` and wired it through `ApplicationContextBuilder`, `BeanContextConfiguration`, `DefaultApplicationContextBuilder`, and `DefaultBeanContext`.
- Added resolution hooks for array-as-bean lookup, bean lookup fallback arguments, candidate matching, null bean replacement, dependent bean destruction, bean initialization gating, and lazy proxy target resolution-path preservation.
- Added lazy scoped proxy support needed by CDI client-proxy semantics, including generated proxy/bean-definition handling and tests for circular dependencies.
- Added bean creation metadata to `BeanCreatedEvent`: the root bean definition and the direct dependent beans created for the bean.
- Added compile-time annotation metadata hooks to expose annotation default values, annotation retention policy, and declared annotation value construction from the visitor metadata builder.
- Added wildcard/type reconstruction support needed for CDI build-time generic type handling, including Java and Kotlin wildcard element behavior.
- Preserved generic field types in generated bean definitions.
- Ignored marker interceptor bindings during interceptor selection so CDI marker metadata does not select interceptors by itself.
- Reported inherited method injection declaring bean information so diagnostics and CDI metadata can distinguish the inherited method owner from the resolved bean type.
- Fixed regressions caught by the focused core test runs:
  - Factory bean descriptions now use the declaring bean type rather than the generated declaring type.
  - Root bean creation paths now include constructor arguments for dependency/circular-dependency diagnostics.
  - Java wildcard elements are no longer treated as raw types.
  - Type reconstruction now collapses raw placeholder arguments without collapsing real wildcard or nested generic arguments.

## Verification

- `./gradlew :micronaut-inject-java:test --no-daemon`
  - `SUCCESS: Executed 4819 tests in 2m 29s (7 skipped)`
- `./gradlew :micronaut-inject-groovy:test --tests io.micronaut.inject.failures.ConstructorDependencyFailureSpec --tests io.micronaut.inject.failures.ConstructorCircularDependencyFailureSpec --tests io.micronaut.inject.failures.FactoryDependencyFailureSpec --tests io.micronaut.inject.failures.FactoryCircularDependencyFailureSpec --no-daemon`
  - `SUCCESS: Executed 5 tests in 2.7s`
- `./gradlew :micronaut-http-server-netty:test --tests io.micronaut.http.server.netty.stack.InvocationStackSpec --no-daemon`
  - `SUCCESS: Executed 11 tests in 2.4s`

The earlier full `check` attempt was stopped while investigating the targeted `micronaut-inject-java:test` failures, so this PR records the focused verification that was rerun after the fixes.


inject/src/main/java/io/micronaut/context/AbstractInitializableBeanDefinition.java
```diff
@@ -603,6 +603,24 @@ public T inject(BeanResolutionContext resolutionContext, BeanContext context, T
...
+     * should receive field/method injection and initialization callbacks.
+     *
+     * @param resolutionContext The resolution context
+     * @param context The bean context
+     * @param bean The bean instance
+     * @return True if the bean should be initialized
+     */
+    @Internal
+    @UsedByGeneratedCode
+    protected final boolean shouldInitializeBean(BeanResolutionContext resolutionContext, BeanContext context, Object bean) {
```
dstepanov: Maybe this method can be added to BeanResolutionContext instead of to the BD

dstepanov: There we can do more optimizations in the future if we want

inject/src/main/java/io/micronaut/context/BeanContextConfiguration.java
```diff
@@ -135,4 +138,21 @@ default Predicate<BeanConfiguration> beanConfiguraionsPredicate() {
...
+    default BeanResolutionCustomizer beanResolutionCustomizer() {
+        return BeanResolutionCustomizer.DEFAULT;
+    }
+
+    /**
+     * @return The custom scope registry factory.
+     * @since 5.1.0
+     */
+    @Nullable
+    default Function<@NonNull BeanContext, @NonNull CustomScopeRegistry> customScopeRegistryFactory() {
```
dstepanov: Maybe this should be an interface with create method

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -2403,20 +2435,53 @@ private <T> BeanRegistration<T> resolveBeanRegistration(@Nullable BeanResolution
...
+                                                                Argument<T> resolvedBeanType,
+                                                                BeanRegistration<T> registration) {
+        Optional<?> resolvedNullBean = beanResolutionCustomizer.resolveNullBean(beanType, resolvedBeanType, registration.beanDefinition);
+        if (resolvedNullBean.isPresent()) {
+            return BeanRegistration.of(this, registration.identifier, registration.beanDefinition, (T) resolvedNullBean.get());
+        }
+        return registration;
+    }
+
+    private static boolean isNullableBeanDefinition(BeanDefinition<?> beanDefinition) {
```
dstepanov, line 2480: In what sense bean definition can be annotated as nullable?

inject/src/main/java/io/micronaut/inject/proxy/InterceptedMethodProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import io.micronaut.inject.ExecutableMethod;
+
+/**
+ * Internal interface implemented by generated proxies that keep bridge-aware executable methods.
+ *
+ * @param <T> The bean type
+ * @since 5.1.0
+ */
+@Internal
+public interface InterceptedMethodProvider<T> extends InterceptedBean {
```
dstepanov: Do we need a new interface for this?

inject/src/main/java/io/micronaut/context/BeanResolutionCustomizer.java
```diff
@@ -0,0 +1,144 @@
...
+     * Returns whether a newly constructed bean instance should receive field/method injection and initialization
+     * callbacks.
+     *
+     * @param resolutionContext The current resolution context
+     * @param beanDefinition The bean definition
+     * @param bean The newly constructed bean instance
+     * @return True if the bean instance should be injected and initialized
+     * @since 5.1.0
+     */
+    default boolean shouldInitializeBean(BeanResolutionContext resolutionContext, BeanDefinition<?> beanDefinition, Object bean) {
```
dstepanov, line 128: Use BeanRegistration here or dont use it in shouldDestroyDependentBeanAfterResolution

core-processor/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
```diff
@@ -92,6 +93,30 @@ public interface VisitorContext extends MutableConvertibleValues<Object>, ClassW
...
 
+    /**
+     * Resolves the default values declared by the given annotation type.
+     *
+     * @param annotationName The annotation type name
+     * @return The default values, or an empty map if the annotation type cannot be resolved
+     * @since 5.1.0
+     */
+    @Experimental
+    default Map<CharSequence, Object> getAnnotationDefaultValues(String annotationName) {
```
dstepanov, line 104: It's always better to use the actual element to read info like this and below, I don't think this works for Kotlin

dstepanov (changes requested):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
graemerocher: CDI allows returning null beans and then injecting null, we don't we have to compensate for that

core-processor/src/main/java/io/micronaut/inject/visitor/VisitorContext.java
graemerocher: we don't have an AnnotationClassElement API where a method like this would naturally fit, there appears to be Kotlin coverage in this PR

inject/src/main/java/io/micronaut/inject/proxy/InterceptedMethodProvider.java
graemerocher: no I guess it could be collapsed into InterceptedBean

dstepanov (approved):

PR #12681, opened 2026-05-20
Fix heap ByteBuf release in NettyReadBuffer.useFastHeapBuffer

### LINKED ISSUE #12680: NettyReadBuffer.useFastHeapBuffer does not release heap-backed ByteBuf
### Expected Behavior

`NettyReadBuffer.useFastHeapBuffer` should release the underlying `ByteBuf` after the callback runs for a heap-backed buffer.

### Actual Behaviour

In the heap-backed branch, `buf` is set to `null` before the callback is called, so the original `ByteBuf` is never released. `close()` cannot release it later because `buf` is already `null`.

I noticed this while looking at `JacksonDatabindMapper.readValue(ReadBuffer, Argument)`, which calls `ReadBuffer.useFastHeapBuffer`.

### Steps To Reproduce

Add this test under `buffer-netty`:

```java
@Test
void useFastHeapBufferReleasesHeapByteBuf() {
    ByteBuf byteBuf = Unpooled.wrappedBuffer(new byte[]{1, 2, 3});
    ReadBuffer readBuffer = NettyReadBufferFactory.of(ByteBufAllocator.DEFAULT).adapt(byteBuf);

    assertEquals(3, readBuffer.useFastHeapBuffer(java.nio.ByteBuffer::remaining));
    readBuffer.close();

    assertEquals(0, byteBuf.refCnt());
}
```
### Environment Information

_No response_

### Example Application

_No response_

### Version

5.0.0

shblue21:
Fixes #12680

`NettyReadBuffer.useFastHeapBuffer` consumes heap-backed buffers, so it should also release the underlying `ByteBuf` once the callback has run.

Added a small regression test for that path.

Verified with:

- `./gradlew :micronaut-buffer-netty:test`
- `./gradlew :micronaut-buffer-netty:check`

### DIFF
--- a/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBuffer.java
+++ b/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBuffer.java
@@ -140,7 +140,11 @@ final class NettyReadBuffer extends ReadBuffer {
         ByteBuf b = getBuf();
         if (b.hasArray()) {
             buf = null;
-            return function.apply(java.nio.ByteBuffer.wrap(b.array(), b.arrayOffset() + b.readerIndex(), b.readableBytes()));
+            try {
+                return function.apply(java.nio.ByteBuffer.wrap(b.array(), b.arrayOffset() + b.readerIndex(), b.readableBytes()));
+            } finally {
+                b.release();
+            }
         } else {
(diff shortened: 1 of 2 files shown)


yawkat (approved):

yawkat:
Thanks!

PR #12690, opened 2026-05-28
fix(ksp): resolve generic type arguments for suspend functions returning non-generic typealiases (#12686)

### LINKED ISSUE #12686: `inject-kotlin` KSP visitor drops generic type argument for `suspend fun` returning a typealias to `List<T>`
### Expected Behavior

For a Micronaut `@Client` interface method `suspend fun foo(...): Alias` where `typealias Alias = List<Bar>`, the declarative client should deserialize the JSON array into `List<Bar>` (typed DTO elements). This works under KAPT and works on the same source if the typealias is replaced with an inline `List<Bar>`.

### Actual Behaviour

The KSP-generated bean-definition metadata for the method emits `Argument.of(java.util.List, "T")` — the element-type slot holds the literal type-variable name `"T"` with no nested `Argument` carrying `Bar`. At runtime `micronaut-jackson-databind` deserializes against this `Argument`, falls back to `List<Object>`, and each JSON object becomes a `LinkedHashMap`. The list returns successfully; the failure surfaces at the caller as:

```
java.lang.ClassCastException: class java.util.LinkedHashMap cannot be cast to class ...Bar
```

on the first typed property access.

### Steps To Reproduce

1. Declare the trigger shape — a typealias to a bare generic `List`, plus a declarative client whose `suspend fun` returns that alias:

   ```kotlin
   data class Bar(val a: String, val b: String)
   data class Req(val ids: List<String>)

   // The trigger — typealias to a bare generic List
   typealias Alias = List<Bar>

   @Client("svc")
   interface DemoClient {
       @Post("/x")
       suspend fun broken(@Body body: Req): Alias

       // Concrete generic class works correctly, for contrast:
       // @Post("/y")
       // suspend fun ok(@Body body: Req): BulkQueryResult<Bar>
   }
   ```

2. Stub the server to return `[{"a":"1","b":"2"}]`.
3. Call `runBlocking { client.broken(Req(listOf("x"))) }` and access `result.first().a`.
4. `ClassCastException: LinkedHashMap cannot be cast to Bar`.

#### Smoking gun (bytecode)

Disassemble `build/generated/ksp/main/.../$DemoClient$Intercepted$Definition$Exec.class`, method `$metadata$broken()`:

**KSP-generated (broken)** — the DTO class is never `ldc`-loaded in this initializer; the element-type slot is the literal type-variable name `"T"`:

```
ldc           class java.util.List
ldc           "T"
invokestatic  io/micronaut/core/type/Argument.of(Ljava/lang/Class;Ljava/lang/String;)Lio/micronaut/core/type/Argument;
```

**KAPT-generated (works) on the same source** — the DTO class is `ldc`-loaded and passed through `Argument.ofTypeVariable` as the bound for `T`:

```
ldc           class java.util.List
ldc           class .../Bar
invokestatic  Argument.of(Ljava/lang/Class;Ljava/lang/String;)Lio/micronaut/core/type/Argument;     ; outer
...
ldc           class java.util.List
ldc           class .../Bar
invokestatic  Argument.of(Ljava/lang/Class;Ljava/lang/String;)Lio/micronaut/core/type/Argument;
invokestatic  Argument.ofTypeVariable(Ljava/lang/Class;Ljava/lang/String;
                                      Lio/micronaut/core/annotation/AnnotationMetadata;
                                      [Lio/micronaut/core/type/Argument;)Lio/micronaut/core/type/Argument;   ; T bound to Bar
```

The compiled `DemoClient.class` JVM `Signature` attribute on the suspend method preserves the parameterization:

```
(Lpkg/Req;Lkotlin/coroutines/Continuation<-Ljava/util/List<Lpkg/Bar;>;>;)Ljava/lang/Object;
```

So the Kotlin compiler is innocent — the substitution `T = Bar` is lost only inside the KSP visitor of `micronaut-inject-kotlin`. Concrete generic classes (e.g. `BulkQueryResult<Bar>`) are NOT affected — KSP visits a `KSClassDeclaration` with concrete `arguments`. The fault is specific to the **typealias-to-`List<T>` + suspend** shape, which routes through `KSTypeAlias` resolution where the substitution map is dropped.

### Environment Information

- Operating System: macOS (Darwin 25.4.0)
- JDK Version: 21
- Kotlin: 2.3.20
- KSP: 2.3.6 (also reproduces on this line; 2.3.7 / 2.3.8 release notes show no related fix)
- JSON binder: `io.micronaut:micronaut-jackson-databind:4.10.20` + `jackson-databind:2.21.2`
- Annotation processing: KSP only (no KAPT)

### Version

4.10.20 (also reproduces on the 4.10.23 release line; release notes for 4.10.21 – 4.10.23 and v5.0.0 contain no related `inject-kotlin` fix)

---

### Workaround

Express the concrete `List<Bar>` directly on the method signature; drop the typealias:

```kotlin
suspend fun broken(@Body body: Req): List<Bar>
```

KSP then visits a `KSClassDeclaration` for `List` with concrete arguments and emits the correct nested `Argument`.

### Related

- KSP [#1011](https://github.com/google/ksp/issues/1011) *Generic TypeAliases Parameter are not consistent resolved* — closed 2022, narrow scope (inheritance via `getAllFunctions/Properties`); current case is a different traversal but the same KSP API surface.
- KSP [#1544](https://github.com/google/ksp/issues/1544) *Unexpected containing declaration while working with typealias of typealias* — open, different sub-case.

idsoke:
**Summary**

Fixes #12686

KSP loses concrete generic type information when a @Client suspend function returns a non-generic typealias wrapping a generic type, for example:

typealias BarList = List<Bar>

As a result, generated argument metadata incorrectly contains a literal "T" instead of the resolved Bar type, causing Jackson to deserialize elements as LinkedHashMap rather than Bar.

**This eventually results in:**

ClassCastException: LinkedHashMap cannot be cast to Bar

KAPT is unaffected because javac expands typealiases before annotation processing.

**Root Cause**

resolveTypeArguments(owner, KSType, ...) iterated over:

type.declaration.typeParameters

For non-generic typealiases, the alias declaration itself has no type parameters, so the underlying generic binding (Bar) was never resolved.

**Changes**
AbstractKotlinElement.kt

When type.declaration is a KSTypeAlias without type parameters, resolve the alias to its underlying type using:

KSTypeAlias.type.resolve()

before processing type arguments.

Generic typealiases remain unaffected and continue through the existing resolution path.

**test-suite-kotlin**

Added regression coverage for a suspend @Client method returning:

typealias BarList = List<Bar>

The response must deserialize into properly typed Bar instances.

Test

**Before fix:**

ClassCastException: LinkedHashMap cannot be cast to Bar

**After fix:**

testSuspendClientReturnsTypealiasToList PASSED

**Notes for Reviewer**
Only resolveTypeArguments(owner, KSType, ...) was modified.
No behavioral changes for:
generic typealiases
direct List<Bar> return types
non-alias generic resolution paths

**Attachment before and after** 

[SuspendClientSpec-after-fix-full.txt](https://github.com/user-attachments/files/28356785/SuspendClientSpec-after-fix-full.txt)
[SuspendClientSpec-before-fix-full.txt](https://github.com/user-attachments/files/28356786/SuspendClientSpec-before-fix-full.txt)


idsoke:
Hi maintainers 👋 

The CI workflows (`Java CI` and `GraalVM Latest CI`) are currently in `action_required` state and waiting for approval to run, since this PR comes from a first-time/external contributor. Could someone please approve the workflow runs so CI can execute?

Thanks!

dstepanov (approved):

PR #12694, opened 2026-05-30
Clean up NoSuchBeanException error message (closes #11320)

### LINKED ISSUE #11320: NoSuchBeanException shouldn't show internally generated class names

szabelin:
Closes #11320.

## Summary

Cleans up the `NoSuchBeanException` error message so it no longer wraps bean type names and class identifiers in `[…]`, and resolves generated proxy class names (`$Foo$Definition$Intercepted`) back to the user-facing simple type.

**Before:**
```
NoSuchBeanException: No bean of type [Foo] exists.
* [$FooImpl$Definition$Intercepted] a candidate of [Foo] is disabled because:
 - Class [Bar] is not present
```

**After:**
```
NoSuchBeanException: No bean of type Foo exists.
* FooImpl, a candidate of Foo, is disabled because:
 - Class Bar is not present
```

## Production changes

- `NoSuchBeanException` — drop the `[`/`]` characters from the message constants used in the top line.
- `DefaultBeanContext.resolveDisabledBeanMessage` and `DefaultApplicationContext.appendEachPropertyMissingBeanMessage` / `appendMissingEachBeanMessage`:
  - Strip the `[`/`]` decoration around bean type names.
  - Render the candidate label using `beanDefinition.getBeanType().getSimpleName()` instead of `beanDefinition.asArgument().getTypeString(true)`, so intercepted/AOP-proxied beans show up as the user-facing type (`FooImpl`) rather than the generated proxy class name (`$FooImpl$Definition$Intercepted`).
  - Use commas around `, a candidate of <type>,` so the line reads naturally.
- The recursive parser that walks transitive `No bean of type … present within context` failures is updated to match the new bracketless format, so the nested-failure tree still works.
- Conditions that emit failure reasons containing a single identifier drop the `[]` decoration: `MatchesPresenceOfClassesCondition`, `MatchesAbsenceOfClassesCondition`, `MatchesAbsenceOfClassNamesCondition`, `MatchesPresenceOfBeansCondition`, `MatchesAbsenceOfBeansCondition`.

### Scope note

Other condition classes that embed *values* (not just identifiers) in their failure reasons — `MatchesPropertyCondition`, `MatchesConfigurationCondition`, `MatchesEnvironmentCondition`, `MatchesCurrentOsCondition`, `MatchesSdkCondition`, etc. — keep their bracket formatting in this PR. The brackets there serve as light value delimiters (`Property [foo] with value [bar baz]`), so the trade-off is different from a single class name. Happy to extend the cleanup in a follow-up if maintainers prefer.

## Tests

- New focused spec `NoSuchBeanMessageFormatSpec` (tagged with `@Issue` for #11320) that asserts the issue scenario directly:
  - top line has no surrounding brackets
  - candidate label has no `[]`, no `$Definition`, no `$Intercepted`
  - `Class … is not present` condition reason has no `[]`
- Existing tests across `inject`, `inject-java`, `inject-groovy`, `inject-kotlin`, and `test-inject-kotlin2-ksp2` that asserted on the old bracket-decorated message format are updated to the new format. The semantic content of the assertions is unchanged — only the bracket / generated-name decoration was updated.

## Test plan

- [x] `./gradlew :micronaut-inject:test` — all green (212 tests, including the new spec)
- [x] `./gradlew :micronaut-inject-java:test` — all green (4760 tests)
- [x] `./gradlew :micronaut-inject-groovy:test` — all green
- [x] `./gradlew :micronaut-inject-kotlin:test` — all green
- [x] Repo-wide grep confirms no remaining occurrences of `No bean of type [`, `a candidate of [`, `requires the presence of a bean of type [`, `Class [.] is not present`, `Class [.] is not absent`, `No bean of type [.] present within context`, `Existing bean [`, or `Configuration requires entries under the prefix: [` outside of the new spec's negative-assertion lines

## Branch

Targeting `4.10.x` per CONTRIBUTING.md. Code is identical on `master` so a forward-port is mechanical; happy to do it if maintainers prefer.

### DIFF
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/configproperties/InterfaceConfigurationPropertiesSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/configproperties/InterfaceConfigurationPropertiesSpec.groovy
@@ -233,7 +233,7 @@ interface MyConfig {
 
         then:"we expect a bean resolution"
         def e = thrown(NoSuchBeanException)
-        e.message.contains("No bean of type [test.MyConfig\$ChildConfig] exists")
+        e.message.contains("No bean of type test.MyConfig\$ChildConfig exists")
 
         cleanup:
         context.close()
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/failures/ConstructorDependencyFailureSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/failures/ConstructorDependencyFailureSpec.groovy
@@ -38,7 +38,7 @@ class ConstructorDependencyFailureSpec extends Specification {
         e.message.normalize() == """\
 Failed to inject value for parameter [propA] of class: io.micronaut.inject.failures.ConstructorDependencyFailureSpec\$MyClassB
 
-Message: No bean of type [io.micronaut.inject.failures.ConstructorDependencyFailureSpec\$MyClassA] exists.$space
+Message: No bean of type io.micronaut.inject.failures.ConstructorDependencyFailureSpec\$MyClassA exists.$space
 Path Taken:$space
 new i.m.i.f.C\$MyClassB(MyClassA propA)
 \\---> new i.m.i.f.C\$MyClassB([MyClassA propA])"""
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/failures/FieldDependencyMissingFailureSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/failures/FieldDependencyMissingFailureSpec.groovy
@@ -38,7 +38,7 @@ class FieldDependencyMissingFailureSpec extends Specification {
         e.message.normalize() == """\
 Failed to inject value for field [propA] of class: io.micronaut.inject.failures.FieldDependencyMissingFailureSpec\$MyClassB
 
-Message: No bean of type [io.micronaut.inject.failures.FieldDependencyMissingFailureSpec\$MyClassA] exists.$space
+Message: No bean of type io.micronaut.inject.failures.FieldDependencyMissingFailureSpec\$MyClassA exists.$space
 Path Taken:$space
 new i.m.i.f.F\$MyClassB()
 \\---> i.m.i.f.F\$MyClassB#propA"""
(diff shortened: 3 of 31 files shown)


szabelin:
Heads-up on context: I opened this alongside #12692. We hit the JAX-RS `@Deprecated` crash (#640 on `micronaut-jaxrs`) in production, and the root cause turned out to be in `AnnotationMetadataSupport` — that's #12692. While I was in the inject module I noticed this issue had been sitting open for a while with a clear before/after spec and no claimants, so I figured I'd pick it up too rather than leave it for next time.

The two PRs are independent — different files, no shared dependencies — but if it's easier to review them together that's also fine. Happy to rebase or split further if you'd prefer.

graemerocher:
I think this should maybe target 5.0.x for 4.x we only really want critical bug fixes at this stage

szabelin:
Totally fair — this is a UX cleanup, not a critical bug fix, so `4.10.x` is the wrong target. The fix we're actually running in production is #12692 (the JAX-RS `@Deprecated` crash), so that's the one we'd love eyes on first. For this PR, happy to either:

1. Close it now and re-open against `5.0.x` once #12692 is in and there's bandwidth, or
2. Re-target the branch to `5.0.x` right now and let it sit until you have time.

Slight preference for (1) — keeps your queue clean — but either way, no rush on this one. Let me know which you prefer.

graemerocher:
1. is fine

szabelin:
Closing per #12694 (comment) — agreed, will re-open against 5.0.x with the same diff when there's bandwidth on your end to review the broader cleanup. No urgency on this one. Thanks!

Issue #930, opened 2018-11-19
OpenAPIDefinition annotation not used for Kotlin class


markjfisher:
### Task List

- [X] Steps to reproduce provided
- [ ] Stacktrace (if present) provided
- [ ] Example that reproduces the problem uploaded to Github
- [X] Full description of the issue provided (see below)

### Steps to Reproduce

1. Create an application with swagger-kotlin feature enabled
2. Add an OpenAPIDefinition annotation
2. Change object to class and wrap main in companion object
3. Build Project

### Expected Behaviour

The swagger file should include the open api annotation information.

### Actual Behaviour

The information is missing.

### Environment Information

- **Operating System**: Linux
- **Micronaut Version:** 1.0.1
- **JDK Version:** 1.8

### Comments

When using the generated "object" class, the swagger is generated, ie. this works:

    @OpenAPIDefinition(
        info = Info(
            title = "Swagger Test"
        )
    )

    object Application {
        @JvmStatic
        fun main(args: Array<String>) {
            Micronaut.build()
                    .packages("test.swagger")
                    .mainClass(Application.javaClass)
                    .start()
        }
    }

And generates the following swagger file:

    openapi: 3.0.1
    info:
      title: Swagger Test
    paths:
      /swagger:
        get:
          operationId: index
    ...

But the following does not:

    @OpenAPIDefinition(
        info = Info(
            title = "Swagger Test"
        )
    )

    class Application {
        companion object {
            @JvmStatic
            fun main(args: Array<String>) {
                Micronaut.build()
                        .packages("test.swagger")
                        .mainClass(Application::class.java)
                        .start()
            }
        }
    }

The info section is no longer in the swagger file:

    openapi: 3.0.1
    paths:
      /swagger:
        get:
          operationId: index
    ...

markjfisher:
I am able to work around this by putting

    object Fake

immediately after the OpenAPIDefinition

glazzari:
Hi, I'm facing a similar issue. I'm not able to use the OpenApiDefinition annotation in the Application.kt gerenated by micronaut CLI. I can workaround this by adding object Fake, as suggested above.

```
@OpenApiDefinition --> In IntelliJ, this returns "This annotation is not applicable to target 'top level function'"
fun main(args: Array<String>) {
	build()
	    .args(*args)
		.packages("com.example")
		.start()
}
```

PR #12695, opened 2026-06-01
Add language-neutral extension points for Scala integration

graemerocher:
## Summary

This PR is the minimized Core prerequisite for extracting Scala support into `micronaut-scala`. It intentionally contains only language-neutral processor and runtime extension points plus the generic fixes required by those extension points.

## What belongs in Core

- Add the experimental `VisitorContext.Language.SCALA` marker.
- Add the experimental `BeanDefinitionInjectionPointResolver` SPI and a no-op default `VisitorContext` accessor, allowing another language model to classify collection, map, Option-like, registration, and other custom injection points.
- Add the experimental `EnumElement` enum-constant lookup extension so language-specific enum models can provide their static lookup method.
- Make `BeanInjectionUtils` delegate unsupported types to the resolver after applying Core standard Java rules.
- Add neutral `Object`-returning generated lookup helpers that use `ConversionService` to convert values to declared targets, while preserving the existing generated methods for binary compatibility.
- Support object-backed enum constant references supplied through `EnumElement` without embedding Scala checks in `BeanIntrospectionWriter`.
- Make container expansion conditional on a successful `Iterable` conversion.
- Retain the generic `Map` conversion check and the language-neutral fixes for factory-produced configuration properties, custom container candidate matching, and inherited default-method AOP.

The Core production implementation contains no Scala package-name checks and no `getScala`-named methods; the language enum is the sole explicit Scala identifier. No Scala compiler, Scala library, GraalPy, or Scala implementation module is added here.

## Rationale

Scala compiler models and compiler-plugin APIs are compiler-specific and do not belong in the Core processor/runtime. Keeping Core limited to stable, experimental, language-neutral SPIs lets micronaut-scala own the Scala element model, compiler plugin, collection and Option classification, enum implementation, and compiler-coupled behavior. The neutral generated helpers and conversion changes let those implementations integrate with existing Core bean definitions without introducing a Scala-specific runtime API.

Compiler-plugin artifact compatibility is handled in the standalone repository with full Scala compiler-version suffixes, following the [Scala compiler-plugin binary-compatibility guidance](https://docs.scala-lang.org/overviews/plugins/index.html#binary-compatibility-of-compiler-plugins).

## Staged follow-up PRs

1. [micronaut-build PR #930](https://github.com/micronaut-projects/micronaut-build/pull/930) adds Scala source discovery and language filtering to documentation snippets.
2. [micronaut-scala PR #2](https://github.com/micronaut-projects/micronaut-scala/pull/2) moves the Scala compiler plugin, Scala visitor model, shared tests, and Scala 3.3.8 compiler-coupled publications into the independent micronaut-scala repository.
3. [Core documentation follow-up PR #12981](https://github.com/micronaut-projects/micronaut-core/pull/12981) restores `test-suite-scala` and the user-facing Scala documentation using the micronaut-scala snapshot.

The branches remain intentionally unmerged until all three stages build together. The intended release order is micronaut-build, this Core SPI, a Core release containing the SPI, micronaut-scala 1.0.0, then the documentation follow-up updated from the snapshot to 1.0.0.

## Verification

- Core processor and inject compilation passed.
- Targeted resolver, enum, processor, inject, and API compatibility checks passed: 4,960 tests, 8 skipped.
- The standalone implementation passed its full check with 224 tests and 2 skipped.
- The broad Core check was attempted; two concurrency-sensitive integration tests failed in that run (stream connection reset and SSL temporary-file server bind), while both tests passed when run serially.

Related implementation has no GraalPy dependency; Scala 2 remains out of scope.


SCALA3_SUPPORT_PLAN.md
```diff
@@ -0,0 +1,488 @@
+# Scala 3 Support for Micronaut Core
```
WojciechMazur: One things that's not in the plan is the long term maintance plan - we cannot expect Micronaut team to know Scala, and especially about compiler plugin knowledge, so if it's merged there should be somebody responsible for this part. 

As a compiler team member, and release officer I can help with opening PRs for upgrade to new minor, but I have no domain knowledge about this ecosystem, so I'd rather limit my involvement to version upgrades and critical compiler fixes. However, I'm involved in multiple different OSS areas, so my time might be really limited

SCALA3_SUPPORT_PLAN.md
```diff
@@ -0,0 +1,488 @@
...
+
```
WojciechMazur: Other part is publishing and maintance of the plugin - compiler plugins have neither binary or source binary compatibility. The sources typically dont't change much, but compiler plugin is strongly coupled with compiler itself - we're not operating on abstraction over compiler, we're touching it directly, so any change compiler, but introduce source incompatibility for maintianers, or when using 2 different versions, binary issues for users. 

Every compiler plugin should be published per supported Scala version using full-cross version, so we need a plan for this part as well: 
- all Scala versions since 3.3 LTS? 
- only latest versions in series LTS and Next series 
- only LTS versions?
- a single version? 

Personally I'd lean toward: latest 3.3 LTS (supported for 1 more year) and latest 3.9 LTS (first release in 6-9 weeks)

https://docs.scala-lang.org/overviews/plugins/index.html#binary-compatibility-of-compiler-plugins 

Example of such an issue (an links to other known to me historical problems in the ecosystem): https://github.com/scalus3/scalus/issues/225

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
```diff
@@ -0,0 +1,1659 @@
...
+      hasFlag(symbol, Flags.PackageClass) ||
+      hasFlag(symbol, Flags.Synthetic) ||
+      hasFlag(symbol, Flags.Artifact)
+
+  private def skipMethod(symbol: Symbol)(using Context): Boolean =
+    symbol == Symbols.NoSymbol ||
+      symbol.denot.isConstructor ||
+      hasFlag(symbol, Flags.Synthetic) ||
+      hasFlag(symbol, Flags.Artifact) ||
+      hasFlag(symbol, Flags.Accessor)
```
WojciechMazur: hasFlag takes a FlagSet, yet you're invoking it multiple times
```suggestion
  private def skipMethod(symbol: Symbol)(using Context): Boolean =
      symbol.denot.isConstructor ||
      hasFlag(symbol, Flags.Synthetic | Flags.Artifact | Flags.Accessor)
``` 

I think `NoSymbol.denot` returns `NoDenotation` which is safe to use directly to cehck for isConstructor of flags directly. It might only throw when trying to access owner

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
```diff
@@ -0,0 +1,1659 @@
...
+final class MicronautScalaCompilerPlugin extends StandardPlugin:
+
+  override val name: String = "micronaut-scala"
+  override val description: String = "Generates Micronaut metadata for Scala 3 sources"
+  override val optionsHelp: Option[String] = None
+
+  override def initialize(options: List[String])(using Context): List[PluginPhase] =
+    val delegate = MicronautScalaCompilerPlugin.delegate(getClass)
+    val initializeMethod = delegate.getClass.getMethod("initialize", classOf[List[?]], classOf[Context])
+    initializeMethod.invoke(delegate, options, summon[Context]).asInstanceOf[List[PluginPhase]]
```
WojciechMazur: I don't quite understand why reflective invocation is required to start the compiler plugin, and why we can't do it directly. It should be documented in there.

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
```diff
@@ -0,0 +1,1659 @@
...
+                      case literal: tpd.Literal =>
+                        annotationConstantValue(literal.const.value).orNull
+                      case select: tpd.Select if isEnumConstant(select.symbol) =>
+                        select.name.toString
+                      case ident: tpd.Ident if ident.name.toString == "_" =>
+                        null
+                      case ident: tpd.Ident if isEnumConstant(ident.symbol) =>
+                        ident.name.toString
+                      case _ =>
+                        renderedClassLiteralValue(tree.show).map(name => classValueData(name)).getOrElse(tree.show)
```
WojciechMazur: Using LLMs is fine for contributing, but let's keep some redability standards...  
We're represeting the Scala community here against multiple JVM languages 

```suggestion
    arrayLiteralValues(tree)
    .orElse(classLiteralValue(tree))
    .orElse(nestedAnnotationValue(tree))
...
```

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
```diff
@@ -0,0 +1,1659 @@
...
+          case ident: tpd.Ident =>
+            symbolConstantValue(ident.symbol)
+          case _ =>
+            None
+      }
+
+  private def constantTypeValue(tpe: Type)(using Context, AnnotationDefaults): Option[Object] =
+    tpe match
+      case constant: ConstantType =>
+        annotationConstantValue(constant.value.value)
```
WojciechMazur: we can easily drop most the the helper method 

```suggestion
      case ConstantType(constant) => 
        constant.value match 
          case _: Unit => None
          case value: AnyVal => Some(value) 
          case value: String => ???
          case _ => null
```

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
```diff
@@ -0,0 +1,1659 @@
...
+          case value: String => value
+          case value: java.lang.Boolean => value
+          case value: java.lang.Byte => value
+          case value: java.lang.Short => value
+          case value: java.lang.Integer => value
+          case value: java.lang.Long => value
+          case value: java.lang.Float => value
+          case value: java.lang.Double => value
+          case value: java.lang.Character => value
+          case _ => null
```
WojciechMazur: Wouldn't that be easier to flip the check? Literal can be `AnyVal | Class | String | Nul`, AnyVal includes `Unit` 
`Constant` has `isAnyVa` check so only remaining check guarding against Unit.

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
```diff
@@ -0,0 +1,1659 @@
...
+    "scala.Boolean" -> "boolean",
+    "scala.Byte" -> "byte",
+    "scala.Char" -> "char",
+    "scala.Double" -> "double",
+    "scala.Float" -> "float",
+    "scala.Int" -> "int",
+    "scala.Long" -> "long",
+    "scala.Short" -> "short",
+    "scala.Unit" -> "void"
+  )
```
WojciechMazur: generaly you should match on symbol or type, not the string representation

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
gbrennon: that a comment that im happy to read :)
its important to keep readability something that acts as a characteristic of that framework :+1:

SCALA3_SUPPORT_PLAN.md
graemerocher: some good points here, it might make sense to have an external micronaut-scala project where this lives so it can have a different versioning lifecycle

inject-scala/src/main/scala/io/micronaut/scala/processing/MicronautScalaCompilerPlugin.scala
graemerocher: Micronaut itself supports compiler extensions to the main processor, so I guess this is trying to allow dynamic load of these extensions, but will review if it is necessary

SCALA3_SUPPORT_PLAN.md
graemerocher: hopefully this can be low overhead, but it is great if you can be a contact if we have a blocking change we can't work out ourselves

SCALA3_SUPPORT_PLAN.md
graemerocher: Moved the code outside of Micronaut Core and into a separate module that can be versioned based on the Scala compiler version

dstepanov:
Reviewed the SPI surface against what [micronaut-scala#2](https://github.com/micronaut-projects/micronaut-scala/pull/2) actually consumes. Most of it holds up — the `AopProxyWriter` reorder in particular is a real fix, since `defaultMethodInterfaceTypes` is populated inside `addInterceptedIfNeeded`, which ran *after* the old superinterface block, so inherited default-method interfaces were never added to the proxy.

Four things needed changing. I pushed them to [`claude/micronaut-scala-extensions-m7fw6g`](https://github.com/micronaut-projects/micronaut-core/compare/core.scala...claude/micronaut-scala-extensions-m7fw6g) (branched off `core.scala`, two commits) so they can be cherry-picked or merged in.

### 1. `PropertySourcePropertyResolver` — regression for existing Java code

```java
Map.class.isAssignableFrom(requiredType) || conversionService.canConvert(Map.class, requiredType)
```

`canConvert` walks the source hierarchy up to `Object`, so this is true for every type with a generic `Object -> X` internal converter: `List`, `String`, `Optional`, `Object[]`, `ConvertibleMultiValues`. A `@Value("${target}") List<Foo>` with no direct value but with `target.*` sub-entries previously resolved to nothing; with this branch it resolves the sub-map and `Object -> List` wraps it into a singleton list. The branch was also placed ahead of the `PropertyResolver` branch, changing precedence.

Fixed by requiring a converter registered for a `Map` source specifically (`canConvert(Map, X) && !canConvert(Object, X)`) and moving the new branch after `Map` and `PropertyResolver`, so the addition is strictly additive. Two regression tests added; the two tests on this branch still pass.

### 2. `BeanIntrospectionWriter.validateEnumValueOfMethod` — inverted assignability check

```java
!beanClassElement.isAssignable(method.getReturnType())
```

asks whether the *enum* is assignable to the return type — it accepts a lookup method returning `Object` and rejects a covariant return, which is the opposite of the stated constraint. Core's own convention (`ClassElement#getAccessibleStaticCreators`) is `method.getReturnType().isAssignable(this)`. Flipped, and added an `isPrivate()` rejection since the generated introspection is a separate class and would otherwise fail with `IllegalAccessError`. Also dropped the branches in `newEnumConstantRef`/`enumValueExpression` that became unreachable in the two-shape refactoring.

### 3. Six dead `@UsedByGeneratedCode` methods

`getBeanMapForConstructorArgument` / `ForMethodArgument` / `ForField` and `getBeanRegistrationCollectionForConstructorArgument` / `ForMethodArgument` / `ForField`, plus `resolveBeanMapOfType` and `resolveBeanRegistrationCollection`, are emitted by no writer — not here, and not in micronaut-scala#2, which uses only `BeanDefinitionInjectionPointResolver`, `getEnumValueOfMethod`, `EnumConstantObjectRef` and the `*Object` lookups. Removed: ~196 lines off the supertype of every generated bean definition.

### 4. Bytecode growth for Java, Groovy and Kotlin

The injection point records always carry a bean type, so `genericType != null` is always true and *every* collection, map, `Optional` and bean-registration injection point in every language now emits a full `ArgumentExpUtils.pushCreateArgument(...)` instead of reading the type parameter off the `Argument` that is already generated for that injection point. Only `scala.collection.Map` genuinely needs it — `isAssignable(Map.class)` is false there, so the old derivation would pick the key type; for `Seq[Foo]` and `Option[Foo]` the existing derivation lands on the right type already.

Changed to write the declared generic type out only when the existing derivation would not arrive at it (`requiresDeclaredGenericType`). Java keeps the bytecode it emits today, Scala still gets the explicit argument where it needs one.

### Smaller ones in the same commits

- A conversion failure in `resolveOptionalBean` now raises a `DependencyInjectionException` naming both types instead of a bare `ConversionErrorException`.
- `asIterable` does one converter lookup instead of `canConvert` followed by `convert`.
- `FactoryBeanElementCreator` skips bean-property enumeration for primitive and array produced types.
- Reverted a stray blank line in `BeanInjectionUtils`.

### Two things left for you

**`FactoryBeanElementCreator`** now binds configuration to factory-produced `@ConfigurationReader` beans in every language — a user-visible change for Java that had no test. I added `FactoryConfigurationPropertiesSpec` covering it, but note what it documents: a factory that `@Replaces` a `@ConfigurationProperties` bean and sets values by hand will now have those values overwritten by configuration where a matching property exists. `Neo4jPropertiesFactory` in the existing suite is exactly that shape and only passes because nothing is configured for it. Worth confirming that is intended.

**`BeanDefinition.isCandidateBean`** — the added `!beanClass.isAssignableFrom(getBeanType())` clause has no test, and it cannot get a Java one: for the Java model `getContainerElement()` is only non-empty for container types, all of which are `Iterable`, so the first clause already short-circuits and the new clause is a no-op. It is only reachable from a language model with a non-`Iterable` container, so its coverage has to live in micronaut-scala.

Verified on JDK 25: `:micronaut-inject:test` (317), `:micronaut-inject-java:test` (4965, 8 skipped), `:micronaut-inject-groovy:test` (1063, 17 skipped), `:micronaut-inject-kotlin:test` (807, 11 skipped), checkstyle on `core-processor` and `inject` — all green.

---

dstepanov (approved):
Approving — the review fixes are now pushed onto `core.scala` itself (`6f68244..4fbb09b`), rather than sitting on a side branch as before.

Two commits:

- **`a11847f`** — the four defects from my earlier review comment: the `PropertySourcePropertyResolver` map-convertible check (it matched any type reachable through a generic `Object -> X` converter, so `@Value("${x}") List<Foo>` started consuming sub-maps); the inverted assignability check in `validateEnumValueOfMethod`; the six dead `@UsedByGeneratedCode` methods on `AbstractInitializableBeanDefinition`; plus the smaller items (`DependencyInjectionException` instead of a bare `ConversionErrorException` for an unconvertible optional bean, a single converter lookup in `asIterable`, primitive/array guard in `FactoryBeanElementCreator`).
- **`4fbb09b`** — `requiresDeclaredGenericType`, so the declared generic type is only written out when the existing runtime derivation would not arrive at it. Only `scala.collection.Map` genuinely needs it; `Seq[Foo]` and `Option[Foo]` derive correctly already, so Java, Groovy and Kotlin keep the bytecode they emit today instead of gaining a `pushCreateArgument(...)` at every collection, map, `Optional` and registration injection point.

Verified on JDK 25: `:micronaut-inject:test` (317), `:micronaut-inject-java:test` (4965, 8 skipped), `:micronaut-inject-groovy:test` (1063, 17 skipped), `:micronaut-inject-kotlin:test` (807, 11 skipped), checkstyle on `core-processor` and `inject` — all green.

Two things I could not close out, so flagging rather than blocking:

**Factory-produced configuration properties.** The `FactoryBeanElementCreator` change binds configuration to factory-produced `@ConfigurationReader` beans in every language. `FactoryConfigurationPropertiesSpec` (added in `a11847f`) documents what that means: a factory that `@Replaces` a `@ConfigurationProperties` bean and sets values by hand now has those values overwritten wherever a matching property is configured. `Neo4jPropertiesFactory` in the existing suite is exactly that shape and only passes because nothing is configured for it. Please confirm that is the intent.

**Enum lookup validation against the real Scala model.** The corrected check now requires the lookup method's return type to be assignable to the enum and the method to be non-private. `ScalaEnumElement#getEnumValueOfMethod` in micronaut-scala#2 selects a static `valueOf(String)` without filtering on return type, so it relies on the Scala model reporting that return type as the enum. That should hold for a Scala 3 `enum`, but I could not run their suite here (it needs Scala 3 plus a Core snapshot carrying this branch), so it wants a green micronaut-scala build before the Core release. Worth adding the return-type filter on their side too, mirroring the Core contract.

---

PR #12714, opened 2026-06-09
Introduce primitive ops to introspection

dstepanov:
Allows to get/set values without boxing


core-processor/src/main/java/io/micronaut/inject/writer/DispatchWriter.java
```diff
@@ -363,6 +374,252 @@ public MethodDef buildDispatchOneMethod() {
...
+
+    @Nullable
+    private static ClassElement firstParameterType(MethodElement methodElement) {
+        ParameterElement[] parameters = methodElement.getParameters();
+        return parameters.length == 1 ? parameters[0].getGenericType() : null;
+    }
+
+    private static String primitiveName(TypeDef.Primitive primitiveType) {
+        if (primitiveType.equals(TypeDef.Primitive.BOOLEAN)) {
+            return "boolean";
```
graemerocher: should probably make all of these constants in ClassUtils where we already have these constants referenced. They are also referenced in PrimitiveElement

core-processor/src/main/java/io/micronaut/inject/writer/DispatchWriter.java
```diff
@@ -363,6 +374,252 @@ public MethodDef buildDispatchOneMethod() {
...
+            return "float";
+        }
+        if (primitiveType.equals(TypeDef.Primitive.DOUBLE)) {
+            return "double";
+        }
+        throw new IllegalStateException("Unsupported primitive dispatch type: " + primitiveType);
+    }
+
+    private static ExpressionDef defaultPrimitiveValue(TypeDef.Primitive primitiveType) {
+        if (primitiveType.equals(TypeDef.Primitive.BOOLEAN)) {
```
graemerocher: maybe this should be some map lookup instead of lots of if branches

graemerocher (changes requested):

graemerocher (approved):

dstepanov:
@copilot fix build

PR #12718, opened 2026-06-11
Add public-only method visibility option for @Around AOP

bogdanu:
Introduce Around.methodVisibility() with a PUBLIC_ONLY option for class-level AOP, preserving existing behavior by default. Apply the selector during compile-time proxy method selection so public-only class advice skips non-public helper methods unless the method has explicit advice.

Add compile-time coverage for public-only interception, constructor helper safety, method-level advice precedence, default behavior, and unchanged factory-produced bean selection.

### Use Case

 Class-level `@Around` advice can currently cause Micronaut to generate intercepted overrides for non-public methods that are still overridable by the generated subclass, such as package-private and protected methods.

  This can be problematic when an advised class calls one of those helper methods from its constructor. Java virtual dispatch can enter the generated AOP override before the generated subclass has finished initializing its interceptor state, which can fail during bean construction.

  `methodVisibility = PUBLIC_ONLY` lets applications apply class-level AOP to the bean’s public API while leaving non-public helper methods unproxied. This is useful for constructor helpers, internal decomposition methods, and package-private/protected test seams that should not become part of the AOP interception surface.
  
  Checking `Modifier.isPublic(context.getTargetMethod().getModifiers())` inside an interceptor is too late for this problem. That check can decide whether the interceptor should apply its behavior, but it does not prevent Micronaut from generating an AOP override for the non-public method. If a constructor calls that helper, Java virtual dispatch still enters the generated override first, before the generated subclass has finished initializing its interceptor/proxy method state. `PUBLIC_ONLY` fixes this at compile time by not generating the override or AOP dispatch entry for inherited non-public helper methods in the first place.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,28 @@ repositories {
     mavenCentral()
 }
 
+def pistolMavenPublishing = providers.gradleProperty("pistolMavenPublishing")
+    .map { it.toBoolean() }
+    .orElse(false)
+
+if (pistolMavenPublishing.get()) {
+    allprojects {
+        plugins.withId("maven-publish") {
+            publishing {
+                repositories {
+                    maven {
+                        name = "Pistol"
+                        url = uri("s3://pistol-maven/maven2")
+                        authentication {
+                            create("awsIm", org.gradle.authentication.aws.AwsImAuthentication)
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
 tasks.named("updateVersionCatalogs") {
     // we set the list to empty because we accept upgrades which improve the status
     rejectedQualifiers = []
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -14,6 +14,28 @@ repositories {
     mavenCentral()
 }
 
+def pistolMavenPublishing = providers.gradleProperty("pistolMavenPublishing")
+    .map { it.toBoolean() }
+    .orElse(false)
+
+if (pistolMavenPublishing.get()) {
```
graemerocher, line 21: please remove this

graemerocher (changes requested):

graemerocher:
you will need to sign the CLA

PR #12728, opened 2026-06-17
Fix raw generic argument qualifier fallback

dstepanov:


graemerocher (approved):

graemerocher:
seems Copilot has some good feedback

PR #12732, opened 2026-06-18
Fix byte range handling for File responses

### LINKED ISSUE #12731: File responses do not handle bytes=0-0 or suffix ranges correctly
### Expected Behavior

File and SystemFile responses should handle single-byte and suffix ranges.

- Range: bytes=0-0 should return 206 Partial Content with the first byte.
- Range: bytes=-10 should return 206 Partial Content with the last 10 bytes.

### Actual Behaviour

- Range: bytes=0-0 is ignored and returns the full body with 200 OK.
- Range: bytes=-10 returns bytes from the start of the file instead of the last 10 bytes.

### Steps To Reproduce

1. Add an endpoint that returns a file with known contents.
2. Request it with Range: bytes=0-0.
3. Observe 200 OK with the full body.
4. Request it with Range: bytes=-10.
5. Observe bytes from the start of the file instead of the last 10 bytes.

### Environment Information

_No response_

### Example Application

_No response_

### Version

5.1.x, 5.0.x

shblue21:
## Summary

Fixes #12731

Range handling for File/SystemFile responses had two small gaps from #8553:

- `Range: bytes=0-0` was ignored because single-byte ranges did not pass the range check.
- `Range: bytes=-10` returned bytes from the start of the file instead of the last 10 bytes.

Invalid and unsupported ranges still fall back the same way as before.

## Testing

- `./gradlew :micronaut-http-server-netty:test --tests io.micronaut.http.server.netty.types.FileTypeHandlerSpec`

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/body/SystemFileBodyWriter.java
+++ b/http-server/src/main/java/io/micronaut/http/server/body/SystemFileBodyWriter.java
@@ -110,8 +110,8 @@ public final class SystemFileBodyWriter extends AbstractFileBodyWriter implement
                 ) {
                     IntRange range = parseRangeHeader(rangeHeader, fileLength);
                     if (range != null // A server that supports range requests MAY ignore or reject a Range header field that contains an invalid ranges-specifier (Section 14.1.1)
-                        && range.firstPos < range.lastPos // A server that supports range requests MAY ignore a Range header field when the selected representation has no content (i.e., the selected representation's data is of zero length).
-                        && range.firstPos < fileLength
+                        && range.firstPos <= range.lastPos
+                        && range.firstPos < fileLength // A server that supports range requests MAY ignore a Range header field when the selected representation has no content (i.e., the selected representation's data is of zero length).
                         && range.lastPos < fileLength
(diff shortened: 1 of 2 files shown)


http-server/src/main/java/io/micronaut/http/server/body/SystemFileBodyWriter.java
```diff
@@ -110,8 +110,8 @@ public ByteBodyHttpResponse<?> write(ByteBodyFactory bodyFactory, HttpRequest<?>
                 ) {
                     IntRange range = parseRangeHeader(rangeHeader, fileLength);
                     if (range != null // A server that supports range requests MAY ignore or reject a Range header field that contains an invalid ranges-specifier (Section 14.1.1)
-                        && range.firstPos < range.lastPos // A server that supports range requests MAY ignore a Range header field when the selected representation has no content (i.e., the selected representation's data is of zero length).
-                        && range.firstPos < fileLength
+                        && range.firstPos <= range.lastPos
```
yawkat, line 113: why support this if the spec says it's not required?

PR #12747, opened 2026-06-24
Handle invalid zone id conversion

kali834x:
the charsequence to zoneid converter in `TimeConverterRegistrar` only catches `DateTimeParseException`, but `ZoneId.of` throws `ZoneRulesException` for an unknown region and a plain `DateTimeException` for a malformed id, neither of which extends `DateTimeParseException`. a bound request value such as `?zone=foobar` therefore leaks the exception straight out of `ConversionService.convert` instead of being rejected, so a bad value turns into a 500 rather than a 400. widening the catch to `DateTimeException` routes every parse failure through `context.reject` and resolves the conversion to `Optional.empty`, the same broader type the temporal format converters in this class already catch.


kali834x:
any update?

graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #12753, opened 2026-06-25
reject header-injection characters in default cookie encoders

kali834x:
the pure-java DefaultServerCookieEncoder and DefaultClientCookieEncoder build the Set-Cookie and Cookie header values by concatenating the cookie name, value, path and domain directly with no character validation. the netty-backed encoders that replace them on the netty runtime route through netty's STRICT/LAX encoders, which reject these characters, so the pure-java path used on non-netty runtimes is the one place where a value containing ';' silently injects an extra cookie attribute and a value containing CR/LF splits the header (CWE-113). add CookieUtils.verifyCookieComponent and call it for name, value, path and domain in encodeCookie and in the client encoder's encode so both reject control characters and ';' the way the netty path already does.


kali834x:
any update?

yawkat:
it's in my queue, i'm just not sure we want to start validating this sort of input because that implies security guarantees this API might not need

yawkat:
@kali834x I've decided to accept this due to the risk of "cookie value escape". however i've asked AI to make the validation more strict, to match upstream netty code and forestall future changes. PTAL

kali834x:
looks good to me. the stricter octet checks and the balanced-quote unwrapping line up with netty's CookieUtil, and snapshotting name/value/path/domain before validating means the validated strings are the ones that get appended. the inline review nits (empty name, hex codepoints in the messages) are already covered by the follow-up commit. ran the http cookie tests on the branch locally and they're green.

kali834x:
took a look, the stricter validation looks good to me. matching netty's token/cookie-octet rules is better than the blocklist I had, and reading each component once before building the header plus the balanced-quote handling covers cases mine missed. ran the cookie encoder tests locally on this branch and they're green, thanks for taking it.

yawkat:
Thanks!

PR #12758, opened 2026-06-27
test: add regression tests for DefaultBeanContext

amirdeljouyi:
I added two regression tests for DefaultBeanContext around bean lookup behavior.

Impact on coverage:

* Covers the relevant DefaultBeanContext logic, including [runtime singleton registration](https://github.com/micronaut-projects/micronaut-core/blob/5.1.x/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java#L651-L688) and [BeanKey identity logic](https://github.com/micronaut-projects/micronaut-core/blob/5.1.x/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java#L3670-L3741)
* Line coverage for DefaultBeanContext increases by 17%.

Why:

DefaultBeanContext relies on BeanKey equality for qualified and generic lookups, and it clears several lookup caches when runtime singletons are registered. The existing tests did not directly lock down those two paths together. These tests verify that BeanKey identity includes qualifiers and type arguments, and that runtime singleton registration is immediately visible through direct, optional, collection, stream, map, and registration lookups.

I hope you find these tests useful. 😄 Please let me know if you have any questions or concerns.

Thanks,

### DIFF
--- a/test-suite/src/test/java/io/micronaut/context/DefaultBeanContextTest.java
+++ b/test-suite/src/test/java/io/micronaut/context/DefaultBeanContextTest.java
@@ -2,10 +2,21 @@ package io.micronaut.context;
 
 import io.micronaut.context.annotation.Secondary;
 import io.micronaut.context.exceptions.NonUniqueBeanException;
+import io.micronaut.core.type.Argument;
+import io.micronaut.inject.qualifiers.Qualifiers;
 import jakarta.inject.Singleton;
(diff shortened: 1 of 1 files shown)


dstepanov:
Thanks @amirdeljouyi! Rebased onto `5.2.x` and continued in #13056, with your commit kept as-is and a small follow-up tightening a few assertions. Closing this one as superseded.

PR #12759, opened 2026-06-28
Handle null instances in QueryValueArgumentBinder

### LINKED ISSUE #12667: Controller Validation regression in Micronaut 4.10.x
Hi there, stumbled across this when updating micronaut.

When there is a Controller which takes an @Nullable @QueryValue enum argument, an exception occurs deep in the validation stack somewhere.

What I'm trying to do:

Perform a GET /enum with no query parameters on an HTTP endpoint defined like this - 
```
  @Get("enum")
  public String enumGet(@Nullable @QueryValue("status") StatusEnum status) {
    return Optional.ofNullable(status).map(StatusEnum::toString).orElse("NULL");
  }
```

Expected outcome:
The string "NULL" is returned with a 200

Actual outcome:
The controller method is never reached, and a 500 is returned

I've put together a sample project that demonstrates the issue - https://github.com/starnose/brokenvalidation

Test show that:
GET /enum?status=Active works fine
GET /enum fails

And for a similar endpoint that uses a Nullable String -
GET /string?status=Active works fine
GET /string also works fine

I've tried 4.10.1 and 4.10.13 and both demonstrate the issue.
If you change the micronaut version back to 4.9.4 it's all happy.

MaheshAravindV:
Closes #12667 
Handle null instances created by the creator in `QueryValueArgumentBinder` explicitly. Error if the argument is not marked `@Nullable` and forward null if it is.

### DIFF
--- a/http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
+++ b/http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
@@ -211,6 +211,15 @@ public class QueryValueArgumentBinder<T> extends AbstractArgumentBinder<T> imple
 
         try {
             T instance = introspectionBuilder.build();
+
+            if (instance == null) {
+                if (argument.isNullable()) {
+                    return BindingResult.empty();
+                }
+
+                return BindingResult.unsatisfied();
+            }
+
             return () -> Optional.of(instance);
         } catch (Exception e) {
             return BindingResult.unsatisfied();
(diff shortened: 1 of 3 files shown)


http/src/main/java/io/micronaut/http/bind/binders/QueryValueArgumentBinder.java
```diff
@@ -205,12 +195,17 @@ private BindingResult<T> bindPojo(ArgumentConversionContext<T> context,
...
         }
 
-        if (!hasAnyValue) {
-            return BindingResult.empty();
-        }
-
         try {
             T instance = introspectionBuilder.build();
+
+            if (instance == null) {
```
MaheshAravindV, line 201: This shows a warning as `introspectionBuilder.build()` is marked as `@NonNull` but actually it can return null. Do not want to mark it with `@Nullable` as that is leading to a cascade of methods needing to be marked `@Nullable`

dstepanov (approved):

PR #12769, opened 2026-06-30
Ensure ByteBuf release during writes

sdelamo:
Use ByteBufInputStream with releaseOnClose in a try-with-resources block when writing Netty ByteBuf bodies. This guarantees the buffer is released when the stream closes, including when transferTo throws, and removes the manual release call.


yawkat (approved):
not sure if this is safe…

PR #12800, opened 2026-07-11
Inherit HTTP/2 settings for service clients

### LINKED ISSUE #12791: http2.max-header-list-size cannot be set for service-scoped HTTP clients
### Expected Behavior

Disclaimer: This issues has been mostly AI generated, but we verified that the workaround actually works.

For a client bound to a `micronaut.http.services.*` entry, it should be possible to raise the
HTTP/2 response header limit (`SETTINGS_MAX_HEADER_LIST_SIZE`, default 8192), either globally:

```yaml
micronaut.http.client.http2.max-header-list-size: 81920
```

or per service:

```yaml
micronaut.http.services.my-service.http2.max-header-list-size: 81920
```

### Actual Behaviour

Both settings are silently ignored for service-scoped clients. Responses with headers over 8 KiB
fail:

```
HttpClientException: Client 'my-service': Error occurred reading HTTP response:
  Header size exceeded max allowed size (8192)
```

Two causes:

1. The per-service key never binds: `ServiceHttpClientConfiguration.ServiceHttp2ClientConfiguration`
   is annotated with `WebSocketCompressionConfiguration.PREFIX` (`"ws.compression"`) instead of
   `Http2ClientConfiguration.PREFIX` (`"http2"`). Fixed on `5.0.x` by #12625 — **please backport
   to 4.10.x**.

2. The global key doesn't reach service clients: the `HttpClientConfiguration` copy constructor
   propagates `maxHeaderSize` etc., but not the nested `Http2ClientConfiguration` — every service
   client gets a fresh one with defaults. **Not fixed by #12625, affects 5.0.x too.**

Combined, on 4.10.x there is no way to configure this value for a service-scoped client. (We hit
this as a production regression: adding a service id to a working client — just for the metrics
`serviceId` tag — silently discarded the global `http2.max-header-list-size` fix.)

### Workaround

We opted the service out of HTTP/2 entirely and raised the HTTP/1.1 limit instead
(`max-header-size` does propagate to service clients, and is honored since #11971):

```yaml
micronaut:
  http:
    client:
      max-header-size: 81920
    services:
      my-service:
        alpn-modes:
          - http/1.1


### Steps To Reproduce

1. Configure as in "Expected Behavior", with `url` pointing at any HTTPS endpoint that negotiates
   h2 and returns > 8 KiB of response headers.
2. Inject `@Client("my-service") HttpClient` and make any request.
3. Observe the 8192 error despite both settings.

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.25 (cause 2 also on 5.0.x)

hutiefang76:
Closes #12791

A service-scoped HTTP client currently falls back to a new `Http2ClientConfiguration` when no service-specific `http2.*` property is present. That skips the global HTTP/2 settings, so values such as `micronaut.http.client.http2.max-header-list-size` are lost for named clients.

This change copies the global HTTP/2 configuration into the service configuration first, then lets service-specific properties override individual values. It covers both paths: when a nested service HTTP/2 bean exists and when the outer configuration has to create the fallback.

Tests cover inheriting all four global HTTP/2 values and overriding only selected values for one service.

Validation:
- `ManualHttpServiceDefinitionSpec`: 6 tests passed
- Spotless and Checkstyle passed
- `japiCmp` passed

I also ran both HTTP client module checks. The affected tests passed; four existing external-network tests timed out (two public WebSocket echo tests and two SSL/redirect tests), including on focused reruns, so they are not related to this configuration change.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -1453,6 +1453,26 @@ public abstract class HttpClientConfiguration {
         private int maxHeaderListSize = DEFAULT_MAX_HEADER_LIST_SIZE;
 
         /**
+         * Default constructor.
+         */
+        public Http2ClientConfiguration() {
+        }
+
+        /**
+         * Copy constructor.
+         *
+         * @param copy The HTTP/2 configuration to copy
+         */
+        protected Http2ClientConfiguration(@Nullable Http2ClientConfiguration copy) {
+            if (copy != null) {
+                this.pingIntervalRead = copy.pingIntervalRead;
+                this.pingIntervalWrite = copy.pingIntervalWrite;
+                this.pingIntervalIdle = copy.pingIntervalIdle;
+                this.maxHeaderListSize = copy.maxHeaderListSize;
+            }
+        }
+
+        /**
          * For HTTP/2 connections, the interval from the last inbound message to when an automated ping
          * should be sent. This can be used to keep low-traffic connections alive.
          *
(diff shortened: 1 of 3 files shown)


hutiefang76:
recheck

yawkat (approved):

PR #12802, opened 2026-07-12
Document immutable configuration with record-like types

MariusVolkhart:
The immutable-configuration guide predated Java records and Kotlin data classes. It framed the interface style as the blanket "preferred" approach — which also contradicted its own note that records are more efficient — and documented only Java records.

This rewrite:

- Reframes the choice around the behaviour that actually differs between the styles: an interface reflects a refreshed `Environment` automatically (each getter reads it on every call), whereas constructor-based styles capture their values once at construction and need `@Refreshable` to observe a refresh.
- Generalises the "Java Record Classes" section to "record-like types", covering Java records and Kotlin data classes. Both have their canonical/primary constructor treated as `@ConfigurationInject`, so no annotation is required (`@JvmRecord` is not needed for Kotlin).
- Adds a "Default values" section: prefer `@Bindable(defaultValue = …)` over a language-level default parameter, because only `@Bindable` is written to the generated configuration metadata that IDE completion and JSON schema consume.
- Fixes a stale sample output (`rodLength=7B.0` → `7.0`).

The Kotlin data-class behaviour is covered by a new test in `test-suite-kotlin`.

### DIFF
--- a/src/main/docs/guide/config/immutableConfig.adoc
+++ b/src/main/docs/guide/config/immutableConfig.adoc
@@ -2,43 +2,41 @@ Since 1.3, Micronaut framework supports the definition of immutable configuratio
 
 dependency::micronaut-context[]
 
-`micronaut-context` is a transitive dependency of `micronaut-http`. If you use a Micronaut HTTP runtime, your project already includes the `Micronaut-context` dependency. 
+`micronaut-context` is a transitive dependency of `micronaut-http`. If you use a Micronaut HTTP runtime, your project already includes the `micronaut-context` dependency.
 
-There are two ways to define immutable configuration. The preferred way is to define an interface annotated with ann:context.annotation.ConfigurationProperties[]. For example:
+Micronaut framework offers several styles of immutable configuration. They bind values the same way but differ in how a bean behaves when the configuration is <<refreshable, refreshed>> at runtime.
 
-snippet::io.micronaut.docs.config.itfce.EngineConfig[tags="imports,class",indent=0,title="@ConfigurationProperties Example"]
+== Interface-based configuration
+
+One approach is to declare an interface annotated with ann:context.annotation.ConfigurationProperties[]. It reflects refreshed configuration automatically, without any additional annotation:
+
+snippet::io.micronaut.docs.config.itfce.EngineConfig[tags="imports,class",indent=0,title="@ConfigurationProperties Interface"]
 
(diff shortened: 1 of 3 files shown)


src/main/docs/guide/config/immutableConfig.adoc
```diff
@@ -2,72 +2,85 @@ Since 1.3, Micronaut framework supports the definition of immutable configuratio
...
-<2> Use the configuration properties
+Java records and Kotlin data classes can be used directly. Micronaut framework treats their canonical (record) or primary (data class) constructor as though it were annotated with ann:context.annotation.ConfigurationInject[], so no annotation is needed on the constructor:
 
-Configuration values can then be supplied when running the application. For example:
+snippet::io.micronaut.docs.config.immutable.ValueAddedTaxConfiguration[tags="imports,class",title="Record-like Configuration"]
 
-snippet::io.micronaut.docs.config.immutable.VehicleSpec[tags="start",indent=0,title="Supply Configuration"]
+<1> The `percentage` component defines the `vat.percentage` configuration property
+
+Like all constructor-based configuration, record-like types capture their values at construction, so annotate them with ann:runtime.context.scope.Refreshable[] if they must update on a <<refreshable, refresh>>.
```
graemerocher, line 57: I am not sure if this is true for records because they are `final` therefore `@Refreshable` will not work since it relies on AOP

dstepanov:
Thanks @MariusVolkhart — taken over in #13057, rebased onto `5.2.x`.

Part of this has already landed on `5.2.x` via #13034 (the `micronaut-context` casing, the stale `rodLength=7B.0` output, the generalization to record classes, and equivalent Kotlin tests), so #13057 carries the remainder: the section restructure, the refresh-behaviour reframing, and the new "Default values" section.

On @graemerocher's review comment — I checked it by compiling and running all four styles against a real `RefreshEvent`. Records are not special: `@Refreshable` on a record compiles fine and behaves exactly as on a non-final class, because no AOP proxy is generated for a `@ConfigurationProperties` class at all (`ConfigurationReaderBeanElementCreator` passes `aopProxyType = false`). The concern was still worth raising, though, because it surfaced an error older than this PR: `@Refreshable` only means the next lookup rebuilds the bean; a singleton that took the configuration in its constructor keeps the stale instance. #13057 documents that limitation explicitly.

PR #12803, opened 2026-07-12
fix(http): @QueryValue object binding for @JsonProperty aliases with underscores

### LINKED ISSUE #12746: Http Query objects's don't work with fields containg underscores
### Expected Behavior

Whenever binding a Query object with a field which contains a JsonProperty with an underscore
Given:
```
        @Introspected
        class Book {
            private String title
            private String extraDetails
            Book(@JsonProperty("title") String title,
                 @JsonProperty("extra_details") @Nullable String extraDetails
                ) {
                this.title = title
                this.extraDetails = extraDetails
            }

            String getExtraDetails() {
                return extraDetails
            }

            String getTitle() {
                return title
            }
        }
---
        @Get('/query-object')
        String queryObject(@QueryValue Book book) {
            "Parameter Value: $book.title $book.extraDetails"
        }

```
when:
```
        HttpMethod.GET  | '/parameter/query-object?age=30&title=JavaBook&extra_details=Details' 
```
Expect:
```
Parameter Value: JavaBook Details
``` 

### Actual Behaviour

The properties with underscore arent being bound to
```
Parameter Value: JavaBook null
``` 

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

https://github.com/KDanisme/micronaut-core/tree/underscore-query-binding-bug

### Version

5.0.3

samikshathorat21:
## Description

Fixes #12746.

### Summary

This PR fixes an issue where `@QueryValue` object binding ignored aliases defined using `@JsonProperty` for constructor parameters.

Previously, query parameters were resolved only by their Java property names. As a result, properties annotated with aliases such as `@JsonProperty("extra_details")` were not bound correctly from incoming query parameters.

---

## Problem

Given the following request object:

```java
@Introspected
class Book {

    private final String title;
    private final String extraDetails;

    Book(
        @JsonProperty("title") String title,
        @JsonProperty("extra_details") @Nullable String extraDetails
    ) {
        this.title = title;
        this.extraDetails = extraDetails;
    }

    String getTitle() {
        return title;
    }

    String getExtraDetails() {
        return extraDetails;
    }
}
```

and the following request:

```http
GET /query-object?title=JavaBook&extra_details=Details
```

the `extraDetails` property was not populated because the binder only matched the Java property name (`extraDetails`) and ignored the alias (`extra_details`).

---

## Solution

This change updates `QueryValueArgumentBinder` to:

- Resolve the introspected property alias when available.
- Fall back to the Java property name when no alias is present.
- Preserve the existing behavior for applications already using Java property names.

---

## Tests

Added regression tests covering:

- Normal property binding.
- `@JsonProperty` aliases containing underscores.
- Multiple aliased properties.
- Nullable constructor parameters.
- Missing optional parameters.
- Alias precedence when both alias and Java property name are present.
- Fallback to the Java property name when the alias is absent.

---

## Verification

The following checks were executed successfully:

```text
./gradlew :micronaut-http:compileJava
./gradlew :micronaut-http-server-tck:test
./gradlew spotlessApply
./gradlew spotlessCheck
```

This change is minimal, backward compatible, and does not introduce any public API changes.

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/binding/HttpQueryObjectsUnderscoresTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/binding/HttpQueryObjectsUnderscoresTest.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.server.tck.tests.binding;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.core.annotation.Introspected;
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

graemerocher:
Thanks for the contribution!

Issue #12804, opened 2026-07-12
Micronaut ksp access to invalid KotlingAwalsAccessibleLifetimeToken


gs-kug:
### Expected Behavior

Compile a project migrated from micronaut 4 to 5

### Actual Behaviour

I'm tring to migrate a project from micronaut 4 to 5. Ksp version 1 is no longer availble we must use ksp version 2 and there is a problem micronaut plugins, that prevents the project compilation:

`
Access to invalid ksp.org.jetbrains.kotlin.analysis.api.platform.lifetime.KotlinAlwaysAccessibleLifetimeToken@e62cd44: PSI has changed since creation
ksp.org.jetbrains.kotlin.analysis.api.lifetime.KaInvalidLifetimeOwnerAccessException: Access to invalid ksp.org.jetbrains.kotlin.analysis.api.platform.lifetime.KotlinAlwaysAccessibleLifetimeToken@e62cd44: PSI has changed since creation
	at ksp.org.jetbrains.kotlin.analysis.api.fir.symbols.KaFirFileSymbol.getPsi(KaFirFileSymbol.kt:65)
	at com.google.devtools.ksp.impl.symbol.kotlin.KSFileImpl.getPsi(KSFileImpl.kt:45)
	at com.google.devtools.ksp.impl.symbol.kotlin.KSFileImpl.filePath_delegate$lambda$0(KSFileImpl.kt:59)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:86)
	at com.google.devtools.ksp.impl.symbol.kotlin.KSFileImpl.getFilePath(KSFileImpl.kt:58)
	at ksp.com.google.devtools.ksp.common.impl.CodeGeneratorImpl.associate(CodeGeneratorImpl.kt:200)
	at ksp.com.google.devtools.ksp.common.impl.CodeGeneratorImpl.createNewFile(CodeGeneratorImpl.kt:171)
	at ksp.com.google.devtools.ksp.common.impl.CodeGeneratorImpl.createNewFile(CodeGeneratorImpl.kt:69)
	at io.micronaut.kotlin.processing.visitor.KotlinVisitorContext$KspGeneratedFile.<init>(KotlinVisitorContext.kt:305)
	at io.micronaut.kotlin.processing.KotlinOutputVisitor.visitMetaInfFile(KotlinOutputVisitor.kt:55)
[...]

PR #12805, opened 2026-07-12
Fix KSP2 crash writing aggregating outputs

### LINKED ISSUE #12677: KSP fails on @ConfigurationProperties when an empty Kotlin file exists
### Expected Behavior

The KSP should not fail, and the project should be built.

### Actual Behaviour

`./gradlew build` fails when a `@ConfigurationProperties` annotation is used and an empty Kotlin file is present.

```terminaloutput
> Task :kspKotlin FAILED
e: [ksp] Error finalizing type visitor  [io.micronaut.context.visitor.ConfigurationMetadataWriterVisitor@263ecd75]: Access to invalid ksp.org.jetbrains.kotlin.analysis.api.platform.lifetime.KotlinAlwaysAccessibleLifetimeToken@62a1220d: PSI has changed since creation
e: [ksp] ksp.org.jetbrains.kotlin.analysis.api.lifetime.KaInvalidLifetimeOwnerAccessException: Access to invalid ksp.org.jetbrains.kotlin.analysis.api.platform.lifetime.KotlinAlwaysAccessibleLifetimeToken@62a1220d: PSI has changed since creation

FAILURE: Build failed with an exception.
```

Removing either the `MyConfiguration` class or `Empty.kt` file makes the build succeed.

---

This behavior is not a showstopper for us (we will update our codegen to not produce empty files). But the error message is very cryptic, and it took us a few hours to get to the root cause. It would be wonderful if it were possible to fix it. Thank you for looking into it.

### Steps To Reproduce

Clone [this repo with a reproducible setup](https://github.com/krystofrezac/micronaut-ksp-configuration-properties-issue). Then run `./gradlew build`

### Environment Information

- Java: openjdk 21.0.5 2024-10-15 LTS
- Kotlin: 2.3.10
- KSP: 2.3.8
- Micronaut gradle plugin: 4.6.2
- Micronaut: 4.10.23

### Example Application

https://github.com/krystofrezac/micronaut-ksp-configuration-properties-issue

### Version

4.10.23

kamilkrzywanski:
Fixes #12677

### Problem

Aggregating visitors (config metadata is the usual one) write META-INF files from `finish()`. We were passing `Dependencies.ALL_FILES` for those outputs.

With KSP2 + incremental processing that walks every `KSFile` and reads `filePath`. If PSI is already invalidated — empty `.kt` files are a reliable trigger — you get:

```
KaInvalidLifetimeOwnerAccessException: PSI has changed since creation
```

### Fix

Use `Dependencies(aggregating = true, sources = emptyArray())` instead of `ALL_FILES`.

Still aggregating (KSP dirties the output on any change via `anyChangesWildcard`), but we don't touch user source files after PSI may already be gone. Same idea as the no-originating-elements path in `visitGeneratedFile`.

Related: #12804 (same underlying issue, OpenAPI finish path).

### Verification

- #12677 reproducer (`@ConfigurationProperties` + empty `Empty.kt` + incremental KSP): fails stock, succeeds with this change
- `:micronaut-inject-kotlin:test` for beans / visitor / configproperties: 334 tests passed

### DIFF
--- a/inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/KotlinCompiler.java
+++ b/inject-kotlin-test/src/main/groovy/io/micronaut/annotation/processing/test/KotlinCompiler.java
@@ -99,6 +99,17 @@ public class KotlinCompiler {
         return toClassLoader(resultPair);
     }
 
+    /**
+     * Creates a Kotlin source file for use with {@link #compile(List, Consumer, List)}.
+     *
+     * @param name source file name
+     * @param source Kotlin source content
+     * @return Kotlin source file
+     */
+    public static SourceFile kotlinSource(String name, @Language("kotlin") String source) {
+        return SourceFile.Companion.kotlin(name, source, true);
+    }
+
     @NotNull
     private static URLClassLoader toClassLoader(Pair<Pair<KotlinCompilation, JvmCompilationResult>, Pair<KotlinCompilation, JvmCompilationResult>> resultPair) {
         try {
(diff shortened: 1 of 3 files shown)


dstepanov:
Added a reproducer

dstepanov (approved):

PR #12806, opened 2026-07-13
Fix #12742: Introspection fails for private JsonProperty fields

### LINKED ISSUE #12742: Introspected.Property issue: the field is not accessible for visibility
### Expected Behavior

_No response_

### Actual Behaviour

The compileJava gradle task fails when a property from an openapi generated classes starts with '$' like in this example:
```
    @Nullable
    @JsonProperty(JSON_PROPERTY_$_REF)
    @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
    private String $ref;
```

### Steps To Reproduce

The `micronaut-kubernetes-client-openapi-common` module uses the openapi gradle plugin to create a kubernetes client from the spec. One of created classes contains:
```
    @Nullable
    @JsonProperty(JSON_PROPERTY_$_REF)
    @JsonInclude(JsonInclude.Include.USE_DEFAULTS)
    private String $ref;
```
After I upgraded micronaut version from 5.0.3 to 5.1.2 in micronaut-kubernetes, I started getting this error in builds:
```
micronaut-kubernetes/kubernetes-client-openapi-common/build/generated/openapi/generateClientOpenApiModels/src/main/java/io/micronaut/kubernetes/client/openapi/model/V1JSONSchemaProps.java:134: error: Element annotated with @Introspected.Property cannot be used as an introspected property: the field is not accessible for visibility [DEFAULT]
    private String $ref;
```

### Environment Information

_No response_

### Example Application

_No response_

### Version

5.1.2

samikshathorat21:
## Summary

This PR fixes a compiler validation issue where private fields annotated with `@JsonProperty` (mapped to `@Introspected.Property`) were incorrectly rejected with the error `the field is not accessible for visibility [DEFAULT]`, even when valid public getter/setter methods were present.

The validation now considers method-based property access before reporting a visibility error, allowing valid bean properties to be introspected while continuing to reject private annotated fields that have no accessible read/write path.

### Changes

- Update `validateIntrospectedPropertyField(...)` to consider `canRead` and `canWrite` during visibility validation.
- Add regression tests covering an OpenAPI-style `$ref` property with public accessors.
- Add a negative regression test to verify that private annotated fields without accessors still fail validation.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/utils/AstBeanPropertiesUtils.java
@@ -514,7 +514,7 @@ public final class AstBeanPropertiesUtils {
         if ((hasReadAccess && canRead) || (hasWriteAccess && canWrite)) {
             return;
         }
-        if (!canReadField && !canWriteField) {
+        if (!canReadField && !canWriteField && !canRead && !canWrite) {
             failInvalidIntrospectedProperty(
                 fieldElement,
                 "the field is not accessible for visibility [" + visibility + "]"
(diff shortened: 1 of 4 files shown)


dstepanov:
@samikshathorat21  Can you pls target 5.1?

samikshathorat21:
Done, I have opened a new PR targeting the 5.1.x branch: #12810.
Thanks!

graemerocher:
Other PR merged, thanks

PR #12812, opened 2026-07-14
Handle health status for endpoint base paths

### LINKED ISSUE #12637: Health endpoint path misconfiguration is allowed
### Expected Behavior

if `endpoints.all.path` yaml configuration is supplied with a missing trailing slash - e.g. `/myservice`, either:
1. the application startup fails with an appropriate message to the user; or
2. the framework handles the absence of the slash(es), and all framework classes function as expected

(also applies to when leading slashes are omitted)

### Actual Behaviour

Omitting the trailing slash does not result in any startup issues, but the path is incorrect in `io.micronaut.management.endpoint.health.filter.HealthResultFilter` resulting in all health endpoint calls returning a 200 response, regardless of the health status. 

There may also be other classes that are affected.

### Steps To Reproduce

1. specify a `endpoints.all.path` configuration without a trailing slash, and call the health endpoint. Note the 200 HTTP response status. 
2. update the health endpoint response to return a DOWN status and call it again. Note the 200 HTTP response status instead of the expected 503 HTTP status. 

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.9.2

sakshichitnis27:
## Summary

- Match health responses using the resolved endpoint route instead of constructing the filter path from configuration.
- Ensure DOWN health responses return the configured HTTP status when `endpoints.all.path` omits a leading or trailing slash.
- Add regression coverage for the affected path variants.

## Testing

- `./gradlew :micronaut-management:test --tests "io.micronaut.management.endpoint.EndpointsBasePathSpec"`
- `./gradlew :micronaut-management:check spotlessCheck`

Fixes #12637

### DIFF
--- a/management/src/main/java/io/micronaut/management/endpoint/health/filter/HealthResultFilter.java
+++ b/management/src/main/java/io/micronaut/management/endpoint/health/filter/HealthResultFilter.java
@@ -18,17 +18,14 @@ package io.micronaut.management.endpoint.health.filter;
 import io.micronaut.context.annotation.Requires;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.health.HealthStatus;
-import io.micronaut.http.HttpRequest;
 import io.micronaut.http.HttpStatus;
 import io.micronaut.http.MutableHttpResponse;
-import io.micronaut.http.annotation.Filter;
 import io.micronaut.http.annotation.ResponseFilter;
 import io.micronaut.http.annotation.ServerFilter;
+import io.micronaut.http.filter.FilterPatternStyle;
 import io.micronaut.management.endpoint.EndpointDefaultConfiguration;
 import io.micronaut.management.endpoint.health.HealthEndpoint;
 import io.micronaut.management.health.indicator.HealthResult;
-import io.micronaut.web.router.MethodBasedRouteMatch;
-import io.micronaut.web.router.RouteAttributes;
 
 /**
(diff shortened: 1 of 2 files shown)


management/src/main/java/io/micronaut/management/endpoint/health/filter/HealthResultFilter.java
```diff
@@ -33,11 +37,7 @@
  * @author graemerocher
  * @since 1.0
  */
-@ServerFilter({
-    HealthResultFilter.DEFAULT_MAPPING,
-    HealthResultFilter.LIVENESS_PROBE_MAPPING,
-    HealthResultFilter.READINESS_PROBE_MAPPING
-})
+@ServerFilter(Filter.MATCH_ALL_PATTERN)
```
graemerocher: this will make the filter match all requests which adds overhead

graemerocher (changes requested):

management/src/main/java/io/micronaut/management/endpoint/health/filter/HealthResultFilter.java
sakshichitnis27: Thanks, addressed in 3ea4dbf. 
The filter now uses path-specific regex mappings for the health, liveness, and readiness endpoints, with optional boundary slashes, instead of matching all requests. I also added coverage for all leading/trailing slash combinations.

graemerocher (approved):

PR #12816, opened 2026-07-14
Copy all HTTP client configuration fields

yawkat:
## Summary
- Copy newer top-level `HttpClientConfiguration` fields in the copy constructor
- Preserve protocol selection fields such as `alpnModes` and `plaintextMode`
- Preserve `pcapLoggingPathPattern` for named service client configurations
- Add regression coverage for copied HTTP client settings and named service pcap inheritance

## Verification
- `./gradlew :micronaut-http-client:test --tests io.micronaut.http.client.config.DefaultHttpClientConfigurationSpec --rerun-tasks -q` (26 tests, 0 failures; warnings only)
- `./gradlew :micronaut-http-client-core:check -q -x japiCmp -x checkVersionCatalogCompatibility` (passed)
- `git diff --check` (passed)

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -38,6 +38,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
@@ -310,7 +311,7 @@ public abstract class HttpClientConfiguration {
             this.threadFactory = copy.threadFactory;
             this.httpVersion = copy.httpVersion;
             this.plaintextMode = copy.plaintextMode;
-            this.alpnModes = copy.alpnModes;
+            this.alpnModes = new ArrayList<>(copy.alpnModes);
             this.allowBlockEventLoop = copy.allowBlockEventLoop;
             this.dnsResolutionMode = copy.dnsResolutionMode;
             this.addressResolverGroupName = copy.addressResolverGroupName;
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

dstepanov (commented):
Rebased onto `5.2.x` and fixed the `build (25)` failure. Reviewed the whole diff; one issue found, now fixed in the branch.

### The CI failure

`:micronaut-management:test` — 4 specs in `ServiceHttpClientHealthIndicatorSpec` died with:

```
java.lang.NullPointerException: Cannot invoke "java.util.Collection.toArray()" because "c" is null
    at io.micronaut.http.client.HttpClientConfiguration.<init>(HttpClientConfiguration.java:314)
    at io.micronaut.http.client.ServiceHttpClientConfiguration.<init>(ServiceHttpClientConfiguration.java:164)
```

That spec builds a `ServiceHttpClientConfiguration` from a `Mock(HttpClientConfiguration)`. Spock instantiates mocks without running field initializers, so every field on the copy source is null — including `alpnModes`, which the new `new ArrayList<>(copy.alpnModes)` dereferenced unguarded.

### The wider issue

Three of the newly copied fields have a non-null contract, and copying null over them is wrong in each case:

- `alpnModes` — `setAlpnModes` does `Objects.requireNonNull`; NPEs directly in the copy constructor.
- `plaintextMode` — `setPlaintextMode` does `Objects.requireNonNull`, and `ConnectionManager:1347` does `switch (httpVersion.getPlaintextMode())`.
- `dnsResolutionMode` — `getDnsResolutionMode()` is not `@Nullable` and `ConnectionManager.getResolver` (line 394) switches on it, reached from line 260 whenever `addressResolverGroupName` is null, which is the common path. This one is the subtler of the three: it trades the NPE in the configuration for one on the first connection instead.

All three now only overwrite the default when the source actually carries a value. The remaining new fields are fine as plain assignments — `addressResolverGroupName` and `pcapLoggingPathPattern` are genuinely nullable (null disables), and `allowBlockEventLoop`/`decompressionEnabled` are primitives with nothing to guard.

### Tests

Added a case copying from a source with none of these set, asserting all three defaults survive. Verified it is not vacuous: dropping any one guard fails it.

Also checked the tests already in the PR are not vacuous — removing the `pcapLoggingPathPattern` copy line fails 2 of them.

`DefaultHttpClientConfigurationSpec` and `ServiceHttpClientHealthIndicatorSpec` both pass locally. I only ran those two specs, not the full job — CI covers the rest.

PR #12823, opened 2026-07-19
Fix bean introspection for OpenAPI generated $ref properties

### LINKED ISSUE #12742: Introspected.Property issue: the field is not accessible for visibility

samikshathorat21:
## Summary

Fixes #12742.

OpenAPI-generated models may contain fields such as `$ref` with corresponding JavaBean accessors `getRef()` and `setRef()`. During bean property resolution, the field name `$ref` was not matched with the accessor property `ref`, causing the field to be treated as a separate property and failing visibility validation.

This change updates `resolvePropertyNameForField` to also check the property name without the leading `$` when resolving fields.

## Changes

- Update `resolvePropertyNameForField` to resolve `$ref` to `ref` when appropriate.
- Add a regression test covering an OpenAPI-style `$ref` field with `getRef()` / `setRef()` accessors.

## Testing

- Added regression test in `BeanIntrospectionSpec`.
- Ran:

```bash
./gradlew :micronaut-inject-java-test:test --tests "io.micronaut.inject.visitor.beans.BeanIntrospectionSpec"

### CODE REVIEW

**Comment by dstepanov** [comment]
Please check if it's not yet fixed in 5.1 branch.
I believe it's fixed with:
```
        if (fieldPropertyName.length() > 1 && fieldPropertyName.charAt(0) == '$') {
            String accessorPropertyName = "$" + Character.toUpperCase(fieldPropertyName.charAt(1)) + fieldPropertyName.substring(2);
            if (props.containsKey(accessorPropertyName)) {
                return accessorPropertyName;
            }
        }
```
If not then this code should be combined somehow.

samikshathorat21:
I checked the 5.1.x branch locally and the issue is still reproducible there.

The current code only checks for $Ref, while the generated accessors resolve to ref, so the lookup still fails.
I kept the existing logic and added the fallback to ref in the same $ handling block as suggested.

dstepanov (approved):

PR #12827, opened 2026-07-21
Fix processing replaced executable methods

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-gcp/pull/1519


sdelamo (approved):

graemerocher (approved):

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -1870,7 +1870,10 @@ protected boolean exec() {
 
     private void processExecutableMethodsProcessAtStartup() {
         Map<Class<? extends Annotation>, Collection<ExecutableMethodProcessor>> processorsByAnnotation = CollectionUtils.newLinkedHashMap(10);
-        for (BeanDefinition<Object> definition : beanDefinitionProvider.getProcessedBeans(this)) {
+        List<BeanDefinition<Object>> processedBeans = new ArrayList<>();
```
graemerocher: can you size the array to the same size as the processed beans?

dstepanov: It's an iterator, I have put 100 so it doesn't resize much

PR #12837, opened 2026-07-23
Cache successful message lookups in ResourceBundleMessageSource

### LINKED ISSUE #12832: ResourceBundleMessageSource does not cache successful message lookups
### Expected Behavior

Successful raw message resolutions should be cached by (locale, code). 

### Actual Behaviour

`ResourceBundleMessageSource.getRawMessage()` reads from `messageCache`, but successful resolutions return before the result is inserted into the cache.

Current code:

```
Optional<String> opt = messageCache.get(messageKey);
if (opt == null) {
    try {
        final Optional<ResourceBundle> bundle = resolveBundle(locale);
        if (bundle.isPresent()) {
            return bundle.map(b -> b.getString(code));
        } else {
            return resolveDefault(code);
        }
    } catch (MissingResourceException e) {
        opt = resolveDefault(code);
    }
    messageCache.put(messageKey, opt);
}
return opt;
```

Both normal paths return early:

- A message found in the locale-specific bundle is returned without caching.
- A message resolved from the default bundle is returned without caching.

`messageCache.put(...)` is reached only when `ResourceBundle.getString()` throws `MissingResourceException`. Consequently, the message cache primarily caches missing-key fallback results, while successful messages are resolved repeatedly.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

https://github.com/micronaut-projects/micronaut-petclinic/tree/main

### Version

5.2.x

el-psy-kongroo-d:
## Problem

`ResourceBundleMessageSource.getRawMessage()` has a `messageCache`, but the two
success paths — a message found in the locale-specific bundle, and one resolved
from the default bundle — return before `messageCache.put(...)`. Only the
`MissingResourceException` fallback populates the cache, so successful lookups are
resolved on every call and never cached.

## Fix

Assign the resolved value to `opt` in both success paths so control falls through
to `messageCache.put(messageKey, opt)` before returning. Behaviour for callers is
unchanged; successful lookups (and empty results from an unresolvable bundle) are
now cached, consistent with the existing `MissingResourceException` fallback.

## Tests

Adds two Spock cases to `ResourceBundleMessageSourceSpec`: a successful lookup is
stored in `messageCache`, and an unresolvable-bundle lookup caches the empty
result. (The existing "repeated to exercise cache" assertions pass whether or not
the value is cached, so they didn't catch this.) The cache is read via reflection,
mirroring `test-suite`'s `ResourceBundleMessageSourceTest`.

`:micronaut-inject:check` and the `test-suite` `ResourceBundleMessageSourceTest`
pass.

Closes #12832

### DIFF
--- a/inject/src/main/java/io/micronaut/context/i18n/ResourceBundleMessageSource.java
+++ b/inject/src/main/java/io/micronaut/context/i18n/ResourceBundleMessageSource.java
@@ -112,9 +112,9 @@ public class ResourceBundleMessageSource extends AbstractMessageSource {
             try {
                 final Optional<ResourceBundle> bundle = resolveBundle(locale);
                 if (bundle.isPresent()) {
-                    return bundle.map(b -> b.getString(code));
+                    opt = bundle.map(b -> b.getString(code));
                 } else {
-                    return resolveDefault(code);
+                    opt = resolveDefault(code);
                 }
             } catch (MissingResourceException e) {
                 opt = resolveDefault(code);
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #12840, opened 2026-07-24
fix: encode illegal characters in UriBuilder path

### LINKED ISSUE #12578: Getting a io.micronaut.http.exceptions.UriSyntaxException when trying to build a URI with a url which have {} in the path
### Expected Behavior

Getting a io.micronaut.http.exceptions.UriSyntaxException when trying to build a URI with 

`UriBuilder.of(url).build()`

where url I am trying to build is **https://abc.com/XXX/YYY;abcd=${ABCD};ab_cd=1** 

Error seems to be related to having {} in path.

### Actual Behaviour

Throwing

Illegal character in path at index 35: https://abc.com/XXX/YYY;abcd=${ABCD};ab_cd=1

### Steps To Reproduce

String url = "https://abc.com/XXX/YYY;abcd=${ABCD};ab_cd=1";

URI uri = UriBuilder.of(url).build();

### Environment Information

MacOS 26.2
JDK 17

### Example Application

_No response_

### Version

4.10.1

kamilkrzywanski:
Fixes #12578

### Problem

`UriBuilder.of(url).build()` throws `UriSyntaxException` when the path contains characters that are illegal in a URI path, for example curly braces:

```java
UriBuilder.of("https://abc.com/XXX/YYY;abcd=${ABCD};ab_cd=1").build();
// Illegal character in path ... ${ABCD}
```

`build()` was reconstructing the URI string and handing it to `new URI(String)` without encoding the path. Query params were already encoded; the path was not (unless you went through `expand()` with a template).

### Fix

When the path is not treated as a URI template, percent-encode illegal path characters on build. Valid `%XX` sequences are left as-is so already-encoded paths are not double-encoded (that was the concern that blocked #7302).

```java
UriBuilder.of("https://abc.com/XXX/YYY;abcd=${ABCD};ab_cd=1").build().toString()
// https://abc.com/XXX/YYY;abcd=$%7BABCD%7D;ab_cd=1
```

Also covers spaces and other non-pchar characters. URI template expansion via `expand()` is unchanged.

### Tests

- #12578 repro (`${...}` in matrix-style path)
- spaces in path
- already percent-encoded path not double-encoded
- unicode path segments

```bash
./gradlew :micronaut-http:test --tests 'io.micronaut.http.uri.UriBuilderSpec'
```

### DIFF
--- a/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
+++ b/http/src/main/java/io/micronaut/http/uri/DefaultUriBuilder.java
@@ -341,6 +341,8 @@ class DefaultUriBuilder implements UriBuilder {
             String pathStr = path.toString();
             if (isTemplate(pathStr, values)) {
                 pathStr = UriTemplate.of(pathStr).expand(values);
+            } else {
+                pathStr = encodePath(pathStr);
             }
 
             builder.append(pathStr);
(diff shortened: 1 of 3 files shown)


yawkat:
i dont think we can do this, it doesnt match browser behavior.

PR #12841, opened 2026-07-26
Make error body buffering for streaming clients configurable

### LINKED ISSUE #12771: ReactorSseClient discards JSON error response body when SSE endpoint returns an HTTP error
### Expected Behavior

When a Server-Sent Events (SSE) endpoint returns a non-2xx HTTP error status (e.g., 400 Bad Request or 500 Internal Server Error) containing a standard application/json error payload (like Micronaut's JsonError), the HttpClientResponseException thrown by the ReactorSseClient should retain this error payload. Calling exception.getResponse().getBody(JsonError.class) **should return the parsed error**, consistent with the behavior of the standard blocking or reactive HttpClient.

### Actual Behaviour

The HttpClientResponseException thrown by ReactorSseClient has an **empty** body. Calling exception.getResponse().getBody() returns Optional.empty(). Inspecting the response reveals it is a NettyStreamedHttpResponse with a null body. It appears that because the client expects a stream of Event<T> chunks, it drops or fails to buffer the standard JSON error payload when an error status code is encountered, making it impossible to read the server's error message.

### Steps To Reproduce

1. Create a Micronaut controller with an SSE endpoint that throws an exception (which Micronaut translates to a 500 Internal Server Error with a JsonError body).

```
@Controller("/sse")
public class SseController {
    @Get(produces = io.micronaut.http.MediaType.TEXT_EVENT_STREAM)
    public Flux<Event<String>> streamEvents() {
        throw new RuntimeException("Intentional exception from SSE endpoint");
    }
}
```

2. Create a client test that calls this endpoint using ReactorSseClient.

```
@MicronautTest
public class SseClientTest {
    @Inject
    @Client("/")
    ReactorSseClient sseClient;

    @Test
    void testSseEndpointException() {
        Flux<Event<String>> eventFlux = sseClient.eventStream("/sse", String.class);
        
        HttpClientResponseException exception = assertThrows(HttpClientResponseException.class, () -> {
            eventFlux.collectList().block();
        });
        
        // This assertion fails because the body is empty!
        assertTrue(exception.getResponse().getBody(JsonError.class).isPresent(), "Body should contain the JsonError");
    }
}
```

3. Run the test and observe that the response body is missing, despite the server sending a valid JSON error payload.

### Environment Information

- Operating System: macOS (reproducible on any OS)
- JDK Version: JDK 17 / JDK 21

### Example Application

https://github.com/bertdw-1985/sse-null-body-bug

### Version

4.3.7 (Also verified on latest)

adityaanikam:
Fixes #12771

Streaming clients (`SseClient`, `ReactorSseClient`, `dataStream`, `jsonStream`) drop the error response body when no explicit `errorType` is given, so `getResponse().getBody(..)` is empty on an HTTP error. `readBodyOnError` in `NettyHttpClient` only buffers the body when `errorType != DEFAULT_ERROR_TYPE`, which is never true for the default overloads.

Adds `micronaut.http.client.buffer error body for streaming` on `HttpClientConfiguration` to control this, per @graemerocher's comment that the default should be configurable.

The default is `false`, preserving current behaviour: buffering an arbitrarily large error body for every streaming client is a memory tradeoff worth opting into, and the existing behaviour is asserted by `ClientStreamSpec` ("test a stream that produces an error without an error type"). Happy to flip the default to `true` if you would prefer the bug fixed out of the box.

Adds a test that mirrors that existing one with the flag enabled, asserting the body is present.

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/HttpClientConfiguration.java
@@ -131,6 +131,11 @@ public abstract class HttpClientConfiguration {
     /**
      * The default value.
      */
+    public static final boolean DEFAULT_BUFFER_ERROR_BODY_FOR_STREAMING = false;
+
+    /**
+     * The default value.
+     */
     @SuppressWarnings("WeakerAccess")
     public static final boolean DEFAULT_ALLOW_BLOCK_EVENT_LOOP = false;
 
@@ -221,6 +226,7 @@ public abstract class HttpClientConfiguration {
     private int maxRedirects = DEFAULT_MAX_REDIRECTS;
 
     private boolean exceptionOnErrorStatus = DEFAULT_EXCEPTION_ON_ERROR_STATUS;
+    private boolean bufferErrorBodyForStreaming = DEFAULT_BUFFER_ERROR_BODY_FOR_STREAMING;
     private boolean decompressionEnabled = true;
 
     private SslConfiguration sslConfiguration = new ClientSslConfiguration();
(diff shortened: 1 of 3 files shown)


dstepanov:
Thanks @adityaanikam! I've rebased this onto current `5.2.x` and continued it in #13055, with your commit and authorship preserved, plus a follow-up fix so a non-JSON error body isn't forced through `JsonError` once the flag is on. Closing this one in favour of that.

PR #12857, opened 2026-08-01
fix(ksp): include Kotlin internal types in bean $EXPOSED_TYPES

### LINKED ISSUE #12854: KSP generates incomplete exposed types in bean definitions
### Expected Behavior

The precompiled definition in `build/generated/ksp/main/classes/<package>/<definition class>` should have all implemented interfaces in `$EXPOSED_TYPES`.

### Actual Behaviour

For classes in deeply nested packages the KSP plugin fails to add the implemented interfaces to the exposed types property. Classes in less nested packages (which have the exact same structure as the faulty classes) do not trigger this bug.

### Steps To Reproduce

1. check out the reproducer code
2. run the `assemble` Gradle task
3. compare the decompilation of the classes in the KSB build dir. Both should contain the line `$EXPOSED_TYPES = Set.of(Working.class, Intf.class);`, but only one adds the interface type to this set.

### Environment Information

- OS: Linux
- JDK: GraalVM CE 25

### Example Application

https://github.com/blued-gear/demo-micronaut-classdefinitionbug

### Version

5.1.0

arimu1:
## Summary
- Treat Kotlin `Visibility.INTERNAL` as public in `AbstractKotlinElement.isPublic()`.
- Kotlin `internal` types are public on the JVM; after #12783, `isAccessibleFromBeanDefinition` omitted internal interfaces from `$EXPOSED_TYPES` when the bean was in another package (same package still worked).
- Add KSP regression coverage matching the [reproducer](https://github.com/blued-gear/demo-micronaut-classdefinitionbug) from #12854.

Fixes #12854

## Testing
- `./gradlew :micronaut-inject-kotlin:test --tests 'io.micronaut.kotlin.processing.beans.BeanDefinitionSpec.test exposed types include kotlin internal interface from another package' --tests 'io.micronaut.kotlin.processing.beans.BeanDefinitionSpec.test limit the exposed bean types*' --tests 'io.micronaut.kotlin.processing.inject.ast.ClassElementSpec' --no-daemon` (JDK 25)
- `./gradlew :micronaut-inject-java:test --tests 'io.micronaut.inject.beans.BeanDefinitionSpec.test exposed types omit package-private supertypes from another package' --no-daemon` (JDK 25) — Java package-private filtering unchanged

### DIFF
--- a/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
+++ b/inject-kotlin/src/main/kotlin/io/micronaut/kotlin/processing/visitor/AbstractKotlinElement.kt
@@ -99,7 +99,14 @@ internal abstract class AbstractKotlinElement<T : KotlinNativeElement>(
     }
 
     override fun isPublic() = if (annotatedInfo is KSDeclaration) {
-        annotatedInfo.getVisibility() == Visibility.PUBLIC
+        // Kotlin `internal` is public on the JVM (module-wide), so Class constants and
+        // bean exposed-type references are valid from any package in the compilation.
+        // Treating only Visibility.PUBLIC as public caused incomplete $EXPOSED_TYPES for
+        // internal interfaces implemented by beans in other packages (issue #12854).
+        when (annotatedInfo.getVisibility()) {
+            Visibility.PUBLIC, Visibility.INTERNAL -> true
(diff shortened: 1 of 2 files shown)


dstepanov (approved):
Not sure if this will work correctly with more complex cases, but I guess we can deal with it later

PR #12858, opened 2026-08-01
fix(http-client): preserve PropagatedContext for client filters with micrometer context-propagation

### LINKED ISSUE #12851: Context propagation to HTTP Client Filters with `io.micrometer:context-propagation`
### Expected Behavior

A `PropagatedContext` should always be propagated to client filters, even when `io.micrometer:context-propagation` is on the classpath.

### Actual Behaviour

When `io.micrometer:context-propagation` is added as a dependency, the `PropagatedContext` is intermittently not propagated to `ClientFilter`s - some outgoing client requests see an empty context instead of the one set for the current request. Removing the dependency makes propagation consistent again.

We add `io.micrometer:context-propagation` because the [Micronaut docs](https://docs.micronaut.io/4.8.18/guide/#contextPropagation) recommend it for improved context propagation, even though our code works fine without it.

### Steps To Reproduce

1. Clone https://github.com/krystofrezac/micronaut-http-client-context-propagation-bug
2. Run `./gradlew test` (with io.micrometer:context-propagation in build.gradle.kts, as committed)- the test fails intermittently since `TestClientFilter` doesn't always see the expected context value.
3. Remove `implementation("io.micrometer:context-propagation")` from `build.gradle.kts` and re-run `./gradlew test` - the test passes consistently.

### Environment Information

- Operating System: macOS 26
- Micronaut version: 4.10.17
- Kotlin version: 2.3.10
- JDK: 21
- Build tool: Gradle (Kotlin DSL)
- Runtime: Netty

### Example Application

https://github.com/krystofrezac/micronaut-http-client-context-propagation-bug

### Version

4.10.17

arimu1:
## Description

Fixes #12851

When `io.micrometer:context-propagation` is on the classpath, Micronaut Reactor enables automatic context propagation. Across sequential Kotlin `suspend` HTTP client calls, the `PropagatedContext` ThreadLocal can be cleared even though the value remains in the Kotlin coroutine context. Declarative clients then capture an empty context for filters, so some outbound requests intermittently (or every other call) see no server-set context elements.

### Fix

In `HttpClientIntroductionAdvice`, for Kotlin coroutine client methods, re-apply `PropagatedContext` from the coroutine context for the synchronous client setup path (where context is captured for the filter runner). Kotlin types are isolated behind a presence check so pure-Java clients without kotlinx-coroutines are unaffected.

### Reproduction

Matches https://github.com/krystofrezac/micronaut-http-client-context-propagation-bug:
- Server filter adds a `PropagatedContextElement`
- Suspend controller issues multiple sequential `@Client` calls
- Client filter reads `MutablePropagatedContext`

Without this fix, filter sees alternating/missing values with micrometer context-propagation enabled.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Related tests have been included
- [x] Documentation related to the change is present in the PR (N/A – behavior restored to documented expectation)

## Testing

```bash
./gradlew :micronaut-context-propagation:test --tests 'io.micronaut.context.propagation.client.ClientFilterPropagatedContextSpec'
```

Result: **PASSED** (JDK 25)

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
@@ -19,6 +19,7 @@ import io.micronaut.aop.InterceptedMethod;
 import io.micronaut.aop.InterceptorBean;
 import io.micronaut.aop.MethodInterceptor;
 import io.micronaut.aop.MethodInvocationContext;
+import io.micronaut.aop.kotlin.KotlinInterceptedMethod;
 import io.micronaut.context.annotation.BootstrapContextCompatible;
 import io.micronaut.context.exceptions.ConfigurationException;
 import io.micronaut.core.annotation.AnnotationMetadata;
@@ -40,7 +41,6 @@ import io.micronaut.core.type.MutableArgumentValue;
 import io.micronaut.core.type.ReturnType;
 import io.micronaut.core.util.ArrayUtils;
 import io.micronaut.core.util.CollectionUtils;
-import io.micronaut.core.util.KotlinUtils;
 import io.micronaut.core.util.StringUtils;
 import io.micronaut.core.version.annotation.Version;
 import io.micronaut.http.BasicHttpAttributes;
@@ -114,25 +114,6 @@ public class HttpClientIntroductionAdvice implements MethodInterceptor<Object, O
      * The default Accept-Types.
      */
     private static final MediaType[] DEFAULT_ACCEPT_TYPES = {MediaType.APPLICATION_JSON_TYPE};
-
-    /**
-     * Whether Kotlin coroutine {@link PropagatedContext} recovery is usable.
-     * Resolved like {@link KotlinUtils}: try to touch optional helpers at class
-     * init and treat {@link NoClassDefFoundError} as unavailable (no reflective
-     * {@code ClassUtils.isPresent} strings that need GraalVM metadata).
-     */
-    private static final boolean KOTLIN_CLIENT_PROPAGATION_AVAILABLE;
-
-    static {
-        boolean available;
-        try {
-            available = KotlinUtils.KOTLIN_COROUTINES_SUPPORTED
-                && KotlinClientPropagatedContext.isAvailable();
-        } catch (NoClassDefFoundError e) {
-            available = false;
-        }
-        KOTLIN_CLIENT_PROPAGATION_AVAILABLE = available;
-    }
 
     private final List<ReactiveClientResultTransformer> transformers;
     private final HttpClientBinderRegistry binderRegistry;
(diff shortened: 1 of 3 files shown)


graemerocher:
can you address the feedback from Copilot and fix check style?

arimu1:
@graemerocher Addressed:

- Copilot hot-path note: only allocate the capturing `Supplier` when Kotlin coroutine propagation can apply; non-Kotlin path runs `dispatchClientCall` directly.
- Checkstyle `InnerTypeLast`: `KotlinClientPropagatedContext` moved after methods; static availability flag with other fields.

Local: `:micronaut-http-client-core:checkstyleMain` green; `ClientFilterPropagatedContextSpec` PASSED (JDK 25).

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -113,6 +115,15 @@ public class HttpClientIntroductionAdvice implements MethodInterceptor<Object, O
...
     private static final MediaType[] DEFAULT_ACCEPT_TYPES = {MediaType.APPLICATION_JSON_TYPE};
 
+    private static final boolean KOTLIN_CLIENT_PROPAGATION_AVAILABLE =
+        ClassUtils.isPresent(
+            "io.micronaut.core.async.propagation.KotlinCoroutinePropagation",
+            HttpClientIntroductionAdvice.class.getClassLoader()
+        ) && ClassUtils.isPresent(
+            "io.micronaut.aop.kotlin.KotlinInterceptedMethod",
+            HttpClientIntroductionAdvice.class.getClassLoader()
+        );
```
graemerocher: think this needs to be rewritten with a similar approach to https://github.com/micronaut-projects/micronaut-core/blob/5.2.x/core/src/main/java/io/micronaut/core/util/KotlinUtils.java

either that or these entries need GraalVM metadata

graemerocher (changes requested):

arimu1:
@graemerocher Rewrote per your review (tip `2a7f636eef`):

- Dropped `ClassUtils.isPresent` string checks (GraalVM metadata concern).
- Availability now uses a **KotlinUtils-style** static init: touch optional helpers via direct class refs; catch `NoClassDefFoundError`.
- No hot-path `Supplier` allocation — open a `PropagatedContext.Scope` (or null) then `dispatchClientCall` in try/finally.

Local: `:micronaut-http-client-core:compileJava` + `checkstyleMain` green (JDK 25).

graemerocher (approved):

http-client-core/src/main/java/io/micronaut/http/client/interceptor/HttpClientIntroductionAdvice.java
```diff
@@ -185,17 +204,24 @@ public Object intercept(MethodInvocationContext<Object, Object> context) {
...
-                                interceptedMethod, annotationMetadata, errorType, declaringType);
-                };
+                // When io.micrometer:context-propagation enables Reactor automatic context
+                // propagation, ThreadLocals may be cleared around coroutine resumes even though
+                // the PropagatedContext is still present in the Kotlin coroutine context.
+                // Re-apply it for the synchronous client setup so filters capture the right context.
+                // Scope open is only attempted for Kotlin suspend clients (no Supplier alloc).
+                PropagatedContext.Scope kotlinScope = null;
+                try {
+                    if (KOTLIN_CLIENT_PROPAGATION_AVAILABLE) {
```
dstepanov: Check for KotlinInterceptedMethod

dstepanov (changes requested):

arimu1:
@dstepanov Addressed: replaced `KOTLIN_CLIENT_PROPAGATION_AVAILABLE` (custom classpath probe touching Kotlin types eagerly at static init) with a direct `interceptedMethod instanceof KotlinInterceptedMethod` check at the call site — same pattern `InterceptedMethodUtil`/`KotlinInterceptedMethodImpl` already use in `aop`. `KotlinClientPropagatedContext.maybePropagate` now takes `KotlinInterceptedMethod` directly instead of doing its own internal instanceof.

Pushed to `2a7f636eef..e976acf77c`.

Local: `:micronaut-http-client-core:compileJava` + `checkstyleMain` green; `:micronaut-context-propagation:test --tests ClientFilterPropagatedContextSpec` PASSED (JDK 25).

PR #12862, opened 2026-08-03
Add an optional JEP-290 ObjectInputFilter to JdkSerializer

### LINKED ISSUE #12861: Optionally apply a JEP-290 ObjectInputFilter in JdkSerializer
### Feature description

`JdkSerializer` is the default `ObjectSerializer` (exposed as `ObjectSerializer.JDK`) and the default value serializer used by `micronaut-redis` for cache and HTTP session storage. Its `deserialize(...)` path calls `ObjectInputStream.readObject()` on bytes read back from a process-external store (Redis, and other cache or session backends).

Today `createObjectInput(...)` overrides only `resolveClass(...)` for classloader resolution and installs no `java.io.ObjectInputFilter` (JEP-290). `resolveClass(...)` resolves a class name but does not constrain which classes may be deserialized, so any type on the classpath can be instantiated from that input, and there is no supported way to apply a JEP-290 filter to this specific path without subclassing `ObjectInputStream`.

For deployments that read serialized values from a shared or otherwise untrusted store, it would help to be able to apply a scoped JEP-290 filter here as opt-in defense-in-depth. Installing a JEP-290 `ObjectInputFilter` on this path is the standard mitigation for this class of deserialization exposure.

### Suggested implementation

Add an optional `ObjectInputFilter` to `JdkSerializer`:

- a new constructor `JdkSerializer(ConversionService, ObjectInputFilter)`, and
- an optional system property (an `ObjectInputFilter.Config.createFilter(String)` pattern) so the filter can be enabled without subclassing.

The filter would be applied to the stream created in `createObjectInput(...)`. Default behaviour would be unchanged (no filter) so existing deployments are unaffected.

I have a small, tested patch ready and can open a PR for this.

Nexory:
Closes #12861

## What / Why

`JdkSerializer` is Micronaut's default `ObjectSerializer` (`ObjectSerializer.JDK`) and the default value serializer used by `micronaut-redis` for cache and HTTP session storage, so its `deserialize(...)` path reads Java-serialized bytes back from a process-external store (Redis, and other cache/session backends). Today `createObjectInput(...)` overrides only `resolveClass(...)` for classloader resolution and installs no `java.io.ObjectInputFilter` (JEP-290), so any class on the classpath is deserialized from that untrusted input. Installing a JEP-290 `ObjectInputFilter` on this path is the standard mitigation for this class of deserialization exposure.

This PR adds an optional, opt-in `ObjectInputFilter` to `JdkSerializer`:

- A new constructor `JdkSerializer(ConversionService, ObjectInputFilter)` to install a scoped filter programmatically.
- An optional system property `micronaut.serializer.jdk.serial-filter`, taking a standard `ObjectInputFilter.Config.createFilter(String)` pattern, so a deployment can enable filtering on this path without subclassing `ObjectInputStream`.
- The filter is applied to the stream in `createObjectInput(...)`. `resolveClass(...)` is left unchanged.

## Non-breaking

The default is unchanged. With no property set and no filter argument the filter is `null` and behaviour is byte-for-byte identical to before. Existing callers, including the shared `ObjectSerializer.JDK` instance, are unaffected.

## Tests

`JdkSerializerSpec` adds two cases: a restrictive filter makes deserialization of a disallowed class fail (surfaced as `SerializationException`), and a filter that allows the required type still round-trips. The existing round-trip and null cases are unchanged.

## Notes

Kept intentionally minimal and opt-in so it cannot affect existing deployments. Happy to adjust the surface to whatever the team prefers: default the filter on with a conservative resource-limit pattern (`maxdepth` / `maxrefs` / `maxbytes` / `maxarray`), bind it through `@ConfigurationProperties` instead of a system property, or keep only the constructor and drop the property. Whichever default is chosen, the single choke point stays in one place.

### DIFF
--- a/core/src/main/java/io/micronaut/core/serialize/JdkSerializer.java
+++ b/core/src/main/java/io/micronaut/core/serialize/JdkSerializer.java
@@ -80,7 +80,11 @@ public final class JdkSerializer implements ObjectSerializer {
         if (pattern == null || pattern.isBlank()) {
             return null;
         }
-        return ObjectInputFilter.Config.createFilter(pattern);
+        try {
+            return ObjectInputFilter.Config.createFilter(pattern);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Invalid " + SERIAL_FILTER_PROPERTY + " value [" + pattern + "]: " + e.getMessage(), e);
+        }
     }
 
     @Override
@@ -151,7 +155,7 @@ public final class JdkSerializer implements ObjectSerializer {
     /**
      * @param inputStream  The input stream
      * @param requiredType The required type
-     * @return A {@link ObjectOutputStream}
+     * @return A {@link ObjectInputStream}
      * @throws IOException if there is an error
      */
     private ObjectInputStream createObjectInput(InputStream inputStream, Class<?> requiredType) throws IOException {
--- a/core/src/main/java/io/micronaut/core/serialize/JdkSerializer.java
+++ b/core/src/main/java/io/micronaut/core/serialize/JdkSerializer.java
@@ -23,6 +23,7 @@ import org.jspecify.annotations.Nullable;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputFilter;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.ObjectStreamClass;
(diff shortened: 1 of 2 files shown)


Nexory:
Pushed. Four cases were missing from your list; allowed graphs and malformed filters
were already covered.

- the `Argument` overload rejects a class the filter disallows, and accepts one it
  allows, so the filter is shown to apply on that path too
- a filter that permits the declared type but not the type of one of its fields
  rejects the graph, which is the nested case
- a recording filter shows that the class handed to the filter is the one the
  `resolveClass` override returned

That last one has a limit worth measuring rather than glossing, so I measured it:
replacing the override body with a plain `super.resolveClass(desc)` leaves all nine
cases green. In a test JVM with one class loader the two resolve identically, so the
case establishes that the filter sees the resolved class, not that the override is
preferred over the default. Separating them needs a second class loader, which felt
like more machinery than this is worth. Say the word if you want it.

Measured in a pinned `eclipse-temurin:25-jdk` image, matching the java 25 line in
gradle.yml:

    JdkSerializerSpec                       9 tests run, 0 failed
    removing only the setObjectInputFilter   9 tests run, 4 failed
    micronaut-core suite, with and without   7417 tests run, 0 failed, both

The four that fail on removal are the three new negative cases plus the existing one;
the `Argument` positive and the malformed-property case stay green either way, which
is what I wanted from them. checkstyleMain and spotlessCheck are clean.

Precedence is documented on both constructors and on the constant: the constructor
argument wins over the system property, and neither touches the process-wide policy.
I also moved the two `@since` tags to 5.2.2, since 5.2.1 has shipped.

The base-branch question from my last comment is still open.

yawkat:
I don't see the value in this. We are planning to move away from the jdk serializer anyway. The people that would opt into a filter should just switch off the jdk serializer.

Nexory:
That's fair. `JdkSerializer` is already deprecated for removal, so this would add API
to a class on its way out, and for anyone who would opt into a filter, moving off JDK
serialization is the better fix.

If the team would rather not take it, I'm happy to close this PR and the issue. Just let
me know which way you want it.

PR #12865, opened 2026-08-03
fix(cors): resolve preflight config from the requested method's route

kali834x:
getAnyConfiguration resolves the preflight policy by walking router.findAny, which returns every route on the path regardless of http method, and takes the first one whose @CrossOrigin matches the origin. a GET route annotated for one origin therefore approves an Access-Control-Request-Method of POST aimed at a sibling POST route annotated for a different origin, because the annotation leaves allowedMethods at any and availableHttpMethods only checks the method exists somewhere on that path. the preflight comes back 200 with Access-Control-Allow-Methods: POST, so the browser goes ahead and sends the real non-simple request and the per-endpoint policy the docs advertise stops holding. skipping route matches whose getHttpMethod differs from methodToMatch keeps the lookup on the route that would actually serve the request; getConfiguration already resolves from the matched route and is unaffected.

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/cors/CrossOriginTest.java
@@ -134,6 +134,26 @@ public class CrossOriginTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


kali834x:
any update?

PR #12868, opened 2026-08-06
Backport fix for NPE when SSE events are received with no `data` field

kevin-wise:
Auth0 event streams include events with no `data` field, which results in a NullPointerException when `byteBufferFactory.wrap(currentEvent.data)` is called. This is already fixed in 5.x by https://github.com/micronaut-projects/micronaut-core/pull/12363/changes#diff-2d6e881f8d13084a9ebabcf6d3fdaa88df24d15431df685539574d59d6232548R2241. This PR makes the minimal changes from that PR to solve the exception.

Let me know if you're no longer accepting bug fixes to 4.x.


kevin-wise:
@graemerocher when do you think you'll be able to release this?

graemerocher:
next 4.x release will be mid September if I recall

kevin-wise:
@graemerocher Any chance of getting a micronaut-core release earlier than that? The two bugs I've fixed are preventing us from upgrading from 4.7 to 4.10

kevin-wise:
looks like it just released. Thanks @sdelamo !

PR #12869, opened 2026-08-06
applyExecutorToPublisher for suspend functions

rawilder:
Built upon https://github.com/micronaut-projects/micronaut-core/pull/12594

When a function is suspending, apply the executor, and add tests that check the different ways an executor can be selected.

### DIFF
--- a/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
+++ b/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
@@ -25,6 +25,7 @@ import jakarta.inject.Singleton;
 import reactor.util.context.ContextView;
 
 import java.util.List;
+import java.util.concurrent.ExecutorService;
 
 /**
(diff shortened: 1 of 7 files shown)


dstepanov:
Thanks for this @rawilder, and @jpbempel for #12594 — the diagnosis and the core of the fix are yours.

I've opened #13032, which carries this forward onto `5.2.x` with you both credited as co-authors on the commit. On top of your change it caches the coroutine dispatcher per executor rather than building one per request (dispatcher equality is identity-based, so a fresh one each time allocates on the request path and defeats `withContext` short-circuiting), builds it from the raw executor rather than the `ContextPropagatingExecutorService` wrapper, and documents the `thread-selection` behaviour change. Your tests are in there almost verbatim, plus two covering those two points.

Closing this one in favour of #13032. Note that means the fix lands on `5.2.x` first — if it's wanted in a 4.10 patch too, that's a separate back-port.

PR #12871, opened 2026-08-10
fix(cookie): parse request Cookie header as name/value pairs

kali834x:
DefaultServerCookieDecoder.decode parses an inbound Cookie request header on runtimes without http-netty, but delegates to java.net.HttpCookie.parse, a Set-Cookie response parser. on a Cookie header with several pairs it returns only the first and drops the rest (SID=a; JSESSIONID=b yields just SID), throws IllegalArgumentException on a malformed name so a bad request turns into a 500, and treats Path/Domain/Max-Age as attributes on the inbound cookie, letting a client set cookie.getPath() which NettyCookies uses to filter which cookies reach the app. the NettyLaxServerCookieDecoder sibling backed by netty ServerCookieDecoder.LAX already parses it as a plain name/value list, so the two SPI implementations disagree on identical input. rewrite decode to split on ';', unwrap balanced quotes, and skip empty or unparsable pairs instead of interpreting attributes or throwing.

### DIFF
--- a/http/src/main/java/io/micronaut/http/cookie/DefaultServerCookieDecoder.java
+++ b/http/src/main/java/io/micronaut/http/cookie/DefaultServerCookieDecoder.java
@@ -16,11 +16,14 @@
 package io.micronaut.http.cookie;
 
 import io.micronaut.core.annotation.Internal;
-import java.net.HttpCookie;
+
+import java.util.ArrayList;
 import java.util.List;
 
 /**
- * {@link ServerCookieDecoder} implementation backed by {@link HttpCookie#parse(String)}.
+ * {@link ServerCookieDecoder} implementation that parses an HTTP {@code Cookie} request header into
+ * its individual name/value pairs, as defined by
+ * <a href="https://datatracker.ietf.org/doc/html/rfc6265#section-4.2.1">RFC 6265, Section 4.2.1</a>.
  *
  * @author Sergio del Amo
  * @since 4.3.0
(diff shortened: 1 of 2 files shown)


HDPark95 (commented):
I found one malformed-value case where the two server decoders still disagree:

```java
String header = "a=\"unterminated; b=2";
```

At `a915b6c`, `NettyLaxServerCookieDecoder` returns only `[b]`, while the new `DefaultServerCookieDecoder` returns `[a, b]` (with `a`'s value still starting with the unmatched quote). Netty's decoder drops a value that starts with `"` but does not end with `"`; this implementation's `unwrapValue` instead returns that value unchanged.

That leaves the runtime-dependent behavior this change is trying to remove, and accepts an unparsable pair rather than skipping it. Could `addCookie`/`unwrapValue` signal and drop an unbalanced opening quote? I tested a minimal pre-unwrapping guard: the full `:micronaut-http:test` task passed 1,697 tests, and a cross-decoder spec covering this input passed both assertions.

kali834x:
Good catch. Pushed df0455c: unwrapValue now returns null when a value opens with a quote that's never closed and addCookie drops that pair, same as Netty LAX. `a="unterminated; b=2` decodes to just `[b]` now, a bare `a="` is dropped, and `a=""` still yields an empty value. Added a regression test covering those. While in there I also bounded the `=` scan to the current segment (it previously searched the rest of the header) and renamed the unused catch variable, per the other review comments.

kali834x:
gentle ping

dstepanov:
Superseded by micronaut-projects/micronaut-core#13054, which is this change rebased onto the current 5.2.x — same two commits, unchanged. Thanks @kali834x, authorship is preserved on both commits. Closing this one in favour of that.

PR #12875, opened 2026-08-11
fix(aop): preserve toString behavior for lazy proxies

### LINKED ISSUE #12873: ClassFormatError: duplicate toString() in generated $Intercepted proxy when the proxied bean declares toString()
### Expected Behavior

The application context starts and the AOP proxy delegates `toString()` to the target bean.

### Actual Behaviour

The context fails at startup with a `ClassFormatError` because the generated `$Definition$Intercepted` class contains two `toString()` methods:

```
Exception in thread "main" io.micronaut.core.io.service.SoftServiceLoader$ServiceLoadingException: Failed to load a service: Duplicate method name "toString" with signature "()Ljava.lang.String;" in class file repro/$RefreshableBeanWithToString$Definition$Intercepted
	at io.micronaut.core.io.service.MicronautMetaServiceLoaderUtils$ServiceInstanceLoader.collect(MicronautMetaServiceLoaderUtils.java:315)
	...
Caused by: java.lang.ClassFormatError: Duplicate method name "toString" with signature "()Ljava.lang.String;" in class file repro/$RefreshableBeanWithToString$Definition$Intercepted
	at repro.$RefreshableBeanWithToString$Definition$Intercepted$Definition.<init>(Unknown Source)
```

### Steps To Reproduce

Any AOP-proxied bean that declares its own `toString()` triggers it. Minimal case — one `@Refreshable` bean:

```java
// src/main/java/repro/RefreshableBeanWithToString.java
package repro;

import io.micronaut.runtime.context.scope.Refreshable;

@Refreshable
public class RefreshableBeanWithToString {

    @Override
    public String toString() {
        return "RefreshableBeanWithToString";
    }
}
```

```java
// src/main/java/repro/Application.java
package repro;

import io.micronaut.context.ApplicationContext;

public class Application {
        try (ApplicationContext context = ApplicationContext.run()) {
            System.out.println(context.getBean(RefreshableBeanWithToString.class));
        }
    }
}
```

```kotlin
// build.gradle.kts
plugins {
    java
    application
}

repositories {
    mavenCentral()
}

dependencies {
    annotationProcessor(platform("io.micronaut.platform:micronaut-platform:5.1.0"))
    annotationProcessor("io.micronaut:micronaut-inject-java")
    implementation(platform("io.micronaut.platform:micronaut-platform:5.1.0"))
    implementation("io.micronaut:micronaut-runtime")
}

application {
    mainClass = "repro.Application"
}
```

`./gradlew run` fails with the error above.

During a large Micronaut 4 → 5 migration we hit the same `ClassFormatError` from several directions: `@Refreshable` beans, `@RequestScope` beans with Lombok `@Data` (which generates `toString()`), and `@ConfigurationProperties` classes proxied by `@MockBean` in tests. The workaround in every case is removing the `toString()` declaration from the proxied class, which is not always possible (third-party classes must instead be replaced via a `@Factory` to avoid the proxy).

### Environment Information

- Micronaut platform 5.1.0 (resolves micronaut-inject 5.1.10)
- JDK: Zulu 25.0.1
- macOS 15 (arm64); also reproduced on Linux x64 CI runners

### Example Application

_No response_

### Version

5.1.10

Boulea7:
Fixes #12873

## Summary

- avoid emitting a duplicate lazy-proxy `toString()` when the target declares an interceptable override
- avoid overriding an inherited final `toString()` while preserving default and inherited non-final behavior
- add annotation-processing and runtime coverage for `@Refreshable` proxies

## Verification

- focused `ScopedProxyToStringSpec`: 4 tests passed
- related AOP and final-method tests: 41 tests passed
- `RefreshScopeSpec`: 9 tests passed
- full `micronaut-inject-java` tests: 4,880 tests, 8 skipped, 0 failed
- `spotlessCheck`
- `micronaut-core-processor` and `micronaut-inject-java` Checkstyle, build, and `japiCmp` tasks

Local verification used OpenJDK 26 because the GraalVM for JDK 25 version specified by `CONTRIBUTING.md` was not available in the test environment. The affected module builds completed successfully while reporting existing Javadoc diagnostics from unchanged files. Upstream CI has not yet run for this pull request.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
+++ b/core-processor/src/main/java/io/micronaut/aop/writer/AopProxyWriter.java
@@ -675,7 +675,9 @@ public class AopProxyWriter extends ProxyingBeanDefinitionWriter {
                             .invoke(COPY_BEAN_CONTEXT_FOR_LAZY_PROXY_TARGET_METHOD, aThis.field(proxyBeanDefinitionField))
                     )
                 ));
-                proxyBuilder.addMethod(getLazyProxyTargetToStringMethod());
+                if (shouldGenerateLazyProxyTargetToStringMethod(interceptedMethods)) {
+                    proxyBuilder.addMethod(getLazyProxyTargetToStringMethod());
+                }
             } else {
                 if (hotswap) {
                     proxyBuilder.addSuperinterface(TypeDef.parameterized(HotSwappableInterceptedProxy.class, targetType));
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

dstepanov (approved):

baztian:
@graemerocher or @dstepanov Any chance that this will be merged into 5.1.x release?

PR #12880, opened 2026-08-14
Use management refresh endpoint in CORS TCKs

sdelamo:
Update CORS TCK scenarios to exercise the management refresh endpoint, configure endpoint sensitivity consistently, and verify refresh events are triggered only for allowed requests.

Add the Micronaut management module to the JDK and Netty HTTP server TCK test suites, and simplify test setup by removing the embedded refresh controllers.


sdelamo:
@yawkat This PR does not change the test behaviour. It just updates the tck tests to use the real refresh endpoint instead of a custom refresh controller mimicking the refresh endpoint.

yawkat (approved):

PR #12882, opened 2026-08-14
Fix handle inaccessible configuration properties in metadata

### LINKED ISSUE #12879: Configuration metadata generation fails for @EachProperty configurations with inaccessible properties since 5.1
### Expected Behavior

Configuration metadata and JSON schema generation should not fail for an `@EachProperty` configuration class solely because it or one of its inherited properties is not visible under the default introspection visibility.

### Actual Behaviour

After upgrading from Micronaut 4.x to Micronaut 5.1.10, compilation fails while generating configuration metadata/schema:

```text
error: Element annotated with @Introspected.Property cannot be used as an introspected property:
the field is not accessible for visibility [DEFAULT]
```

The exception originates from `JsonSchemaConfigurationMetadataWriter`, which calls `getBeanProperties()` while writing configuration schema metadata.

### Context

Our affected classes are `@EachProperty` configuration roots that extend external configuration types and also use `@ConfigurationBuilder`. Initially, we believed that package-private `@ConfigurationBuilder` fields caused the problem. However, making those fields `public` did **not** resolve the compilation failure.

The failure is therefore not solely a `@ConfigurationBuilder` field-visibility issue. It appears to be caused by configuration-schema metadata generation resolving inaccessible/inherited properties with the default introspection visibility.

This did not fail with Micronaut 4.x and started after upgrading to Micronaut 5.1.10. It appears related to the explicit `@Introspected.Property` validation introduced in https://github.com/micronaut-projects/micronaut-core/commit/22cdeff40d

### Confirmed Workaround

Add explicit introspection with `ANY` visibility to each affected configuration root:

```java
@EachProperty("example")
@Introspected(visibility = Introspected.Visibility.ANY)
class ExampleConfiguration extends SomeConfigurationBase {
    // may contain @ConfigurationBuilder fields
}
```

This resolved the compilation failure in our application.

### Questions

1. Is `@Introspected(visibility = Introspected.Visibility.ANY)` the intended workaround for configuration classes in this situation?
2. Should `JsonSchemaConfigurationMetadataWriter` honor the configuration class introspection visibility, or avoid failing when a property cannot be represented in the schema?
3. If this behavior is intentional, should it be documented as a Micronaut 5 migration change?

dstepanov:
Fixes https://github.com/micronaut-projects/micronaut-core/issues/12879

### DIFF
--- a/core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
+++ b/core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
@@ -16,6 +16,7 @@
 package io.micronaut.context.visitor;
 
 import io.micronaut.context.annotation.Bean;
+import io.micronaut.context.annotation.BeanProperties;
 import io.micronaut.context.annotation.ConfigurationBuilder;
 import io.micronaut.context.annotation.ConfigurationInject;
 import io.micronaut.context.annotation.ConfigurationProperties;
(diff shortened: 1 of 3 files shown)


graemerocher:
probably needs to target 5.1.x

dstepanov:
Targeting 5.1

graemerocher (approved):

PR #12886, opened 2026-08-15
Fix FormDemuxer event loop race

### LINKED ISSUE #12860: `FormDemuxer` constructor can run `devolveToStreaming` off the event loop — intermittent multipart 500s under concurrency
### Expected Behavior

A `multipart/form-data` POST binding a `CompletedFileUpload` should succeed regardless of which
thread the filter chain resumed on before the route's arguments are fulfilled.

### Actual Behaviour

> **Disclosure: this report, the reproducer and the suggested fix were all written by an AI coding
> agent (Claude), reading the 5.1.10 sources.** I ran everything myself and every number quoted is
> from a real run on the machine described below, I put quite some effort and tokens into it and read through everything myself.

About 1 request in 15,000 fails with a 500 from inside
`FormDemuxer$OptimisticBufferingContent.devolveToStreaming`. Three signatures, **no application
frames in any of them**:

| `FormDemuxer.java` | statement | exception |
|---|---|---|
| `:413` | `tracker.subtract(…)` | `IllegalArgumentException: bytes > value` |
| `:415` | `sc.add(buffer)` | `IllegalStateException: Must only be called on event loop` |
| `:422` | `emit(…)` | `AssertionError` — the `assert …inEventLoop()` at `:207`, so only under `-ea` |

Three consecutive statements of one method: one bug, `devolveToStreaming` running on a thread that
is not the channel's event loop.

#### Cause

```java
// FormDemuxer.<init>, :100-111
this.eventLoop = channel.eventLoop();
try (var s = byteBodyFactory.toStreaming(byteBody)) {
    this.upstream = s.primary(this);                     // :105
}
// we delay streaming until the upstream is set, so let's do that now.
if (state instanceof OptimisticBufferingContent opt) {   // :108
    opt.devolveToStreaming(false);                       // :109
}
```

`s.primary(this)` reaches `SharedBuffer.subscribe`, which does
`eventLoop.execute(() -> subscribe0(…))` when the caller is off the loop
(`StreamingNettyByteBody.java:205-210`). The constructor then reads `state` at `:108` — non-`volatile`,
declared at `:78` — while the event loop is concurrently in `subscribe0` → `forwardInitialBuffer` →
`add` → `forwardOutput()` writing it. No happens-before edge, so if that read observes
`OptimisticBufferingContent` the devolution runs on the calling thread.

Everything else in the class is confined: the other methods either
`assert eventLoop == null || eventLoop.inEventLoop()` or hop (`cancel()` `:125`,
`onBytesConsumed` `:468`, `allowDiscard` `:488`). This fix-up does neither.

Without `-ea` the `emit` assert never fires and the same interleaving proceeds — into inconsistent
`totalTracker` accounting and possible double-release, since both threads walk the same
unsynchronised `buffers` list.

#### Suggested fix

Hop the fix-up onto the event loop, like the other entry points:

```diff
             // we delay streaming until the upstream is set, so let's do that now.
-            if (state instanceof OptimisticBufferingContent opt) {
-                opt.devolveToStreaming(false);
-            }
+            if (eventLoop.inEventLoop()) {
+                devolveIfStillBuffering();
+            } else {
+                eventLoop.execute(this::devolveIfStillBuffering);
+            }
         }
     }
+
+    private void devolveIfStillBuffering() {
+        assert eventLoop == null || eventLoop.inEventLoop();
+        if (state instanceof OptimisticBufferingContent opt) {
+            opt.devolveToStreaming(false);
+        }
+    }
```

The `state` read moves onto the event loop, serialized with `subscribe0` instead of racing it.
Netty's task queue is FIFO and both tasks are submitted from this thread, so the hop runs after
`subscribe0`: either that already devolved and the hop is a no-op, or it saw `upstream == null`,
skipped, and the hop devolves. Exactly one devolution, always on the event loop. The synchronous
path is unchanged.

Measured with the patched class shadowing the released one on the classpath (verified via
`getProtectionDomain()`):

| build | requests | 500s |
|---|---|---|
| released 5.1.10 | 228,540 | 14 |
| **patched** | **40,026,747** | **0** |

All patched requests returned 200 with the upload bound correctly. A second pair of runs with the
filter kept *on* the event loop (33M requests each, released and patched, both clean) covers the
fix's `inEventLoop()` branch, which the async path never takes.

A suggestion, not a PR: exercised only by that harness, never against micronaut-core's own test
suite.

### Steps To Reproduce

Reproducer: <https://github.com/iribaar77/micronaut-formdemuxer-race-repro>

`./gradlew run` — exit code 0 means it reproduced. First failure in ~2 s on a 32-core box, ~5 s
pinned to 4 cores (`taskset -c 0-3`). `./gradlew runPatched` is the same harness against the
suggested fix. No Docker, no database, no micronaut-security.

Two conditions have to hold:

**1. The filter chain must resume off the event loop**, which puts
`RequestLifecycle.fulfillArguments` → `FormRouteCompleter.start` → `new FormDemuxer(…)` on a
non-Netty thread. The reproducer uses `Mono.fromCallable(…).subscribeOn(pool).flatMap(x ->
continuation.proceed())` in a `@ServerFilter`; in my application it is `micronaut-security`'s
`SecurityFilter` with an async `AuthenticationFetcher`. `@ExecuteOn(BLOCKING)` on the controller is
*not* the cause — that applies in `RouteExecutor.callRoute`, after `fulfillArguments`.

**2. The request head and the body must arrive in separate TCP segments** — same box, same build,
only the client's write pattern changed:

| request written as | requests | 500s |
|---|---|---|
| head, then body — two `write()`s | 198,663 | 14 |
| head + body in a single `write()` | 18,509,828 | **0** |

With the body in the head's segment the server has a complete body before routing starts and the
window never opens. That is why a naive concurrent-upload loop misses it; every real client,
Micronaut's included, writes a multipart body separately from the head.

Stack, `Must only be called on event loop` variant (the other two differ only above
`devolveToStreaming`):

```
java.lang.IllegalStateException: Must only be called on event loop
    at io.micronaut.http.netty.body.StreamingNettyByteBody$SharedBuffer.add(StreamingNettyByteBody.java:248)
    at io.micronaut.http.server.netty.multipart.FormDemuxer$StreamingContent.add(FormDemuxer.java:447)
    at io.micronaut.http.server.netty.multipart.FormDemuxer$OptimisticBufferingContent.devolveToStreaming(FormDemuxer.java:415)
    at io.micronaut.http.server.netty.multipart.FormDemuxer.<init>(FormDemuxer.java:109)
    at io.micronaut.http.server.netty.NettyHttpRequest.getRawFormFields(NettyHttpRequest.java:719)
    at io.micronaut.http.server.multipart.FormRouteCompleter.start(FormRouteCompleter.java:121)
    at io.micronaut.http.server.RequestLifecycle.fulfillArguments(RequestLifecycle.java:603)
```

### Environment Information

- micronaut-core **5.1.10** (via `io.micronaut.platform:micronaut-platform:5.1.0`)
- Netty 4.2.16.Final; `io.netty.contrib:netty-codec-multipart-core:1.0.0.Final`
- `micronaut.server.netty.server-type`: default (`streamed`); assertions on (`-ea`)
- OpenJDK 25.0.4, Fedora Linux 44, x86_64, 32 cores

First seen on 5.0.6 (2026-07-18). `FormDemuxer.java` is byte-identical between the 5.0.6 and 5.1.10
sources jars, so nothing in the 5.0.x → 5.1.x line has touched it. No equivalent code in 4.x.

`server-type: full_content` does avoid it — the body is then an `AvailableByteBody` and takes the
first constructor branch, whose comment at `:88-90` explicitly allows off-event-loop callers. I
cannot use it: it also rejects oversize uploads with a bare framework 413 before my own error
mapping runs, which is a breaking change for my API clients.

### Example Application

https://github.com/iribaar77/micronaut-formdemuxer-race-repro

### Version

5.1.10

yawkat:
## Summary

- serialize the constructor's optimistic-buffering fix-up on the channel event loop
- recheck the demuxer state when deferred work runs, avoiding stale devolution after another state transition
- preserve the existing synchronous path when construction already occurs on the event loop

## Testing

- `./gradlew :micronaut-http-server-netty:spotlessCheck :micronaut-http-server-netty:test --tests io.micronaut.http.server.netty.multipart.FormDemuxerTest`
- 13 `FormDemuxerTest` tests passed

Resolves #12860

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/FormDemuxer.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/multipart/FormDemuxer.java
@@ -105,9 +105,18 @@ public final class FormDemuxer implements BufferConsumer {
                 this.upstream = s.primary(this);
             }
             // we delay streaming until the upstream is set, so let's do that now.
-            if (state instanceof OptimisticBufferingContent opt) {
-                opt.devolveToStreaming(false);
+            if (eventLoop.inEventLoop()) {
+                devolveIfStillBuffering();
+            } else {
+                eventLoop.execute(this::devolveIfStillBuffering);
             }
+        }
+    }
+
+    private void devolveIfStillBuffering() {
+        assert eventLoop == null || eventLoop.inEventLoop();
+        if (state instanceof OptimisticBufferingContent opt) {
+            opt.devolveToStreaming(false);
         }
     }
 


yawkat:
It's too racy, I don't think a test is possible.

graemerocher (approved):

HDPark95 (commented):
`StreamingContent.allowDiscard` looks like the same pattern: L493 reads the non-volatile `state` four lines above its hop at L497, and `StreamingNettyByteBody.close` (L133) calls it with no hop. Driving the demuxer from a real event loop instead of `EmbeddedChannel`, closing an unconsumed field body enters `allowDiscard` on the caller thread with `inEventLoop()` false, so that read races the loop's writes. Should the shortcut sit below the hop?

`FormDemuxerTest` can't cover this: on `EmbeddedChannel` `inEventLoop()` is always true, so the new `else` branch never runs there.

yawkat:
StreamingContent.allowDiscard should be fine, it's covered by the happens-before semantics.

PR #12856, opened 2026-08-01
Fix quadratic response body assembly in NettyReadBufferFactory.compose

jjathman-ch:
Fixes #12855

`NettyReadBufferFactory.compose(Iterable<ReadBuffer>)` received the complete `Iterable` of body
chunks but added them to the `CompositeByteBuf` one at a time via
`CompositeByteBuf.addComponent(boolean, ByteBuf)`, which calls `consolidateIfNeeded()` after every
single component. Each consolidation copies **everything accumulated so far**, making aggregation
of a body of size `S` in chunks of size `c` cost roughly `S² / (2 · maxNumComponents · c)` bytes of
copying — about 19 GB for a 200 MB response at the default 16 components and 64 KB socket reads,
entirely on the Netty event loop. See #12855 for the full analysis and two production stack traces.

### Change

- Extract all `ByteBuf`s from the input `ReadBuffer`s into a `List<ByteBuf>` first, then hand the
  complete list to `CompositeByteBuf.addComponents(boolean, Iterable<ByteBuf>)`, which consolidates
  at most once, at the end, instead of after every component.
- The extraction and handover are split into two separate `try` blocks because ownership of each
  `ByteBuf` moves in stages:
  - If extraction fails partway, the already-extracted buffers have no owner (since
    `NettyReadBuffer.toByteBuf()` nulls its field) and must be released explicitly.
  - `addComponents` takes ownership of everything it is given, releasing anything it did not add
    in its own `finally`, so after it is called only the composite itself needs releasing.
- `allocator.compositeBuffer()` (no-arg) is kept as-is, along with the existing size-0/size-1
  shortcuts, so the shape of the returned buffer (single component) is unchanged. This keeps the
  change safe for the server and proxy paths that write the composite back to a socket.

### Tests

- `composeManyKeepsContent` (in `AbstractReadBufferTest`, covering both the Netty and NIO
  factories): composes 50 small buffers and asserts the result matches the exact concatenation of
  the inputs.
- `composeConsolidatesOnlyOnce`: uses a counting allocator to assert that only one large
  (consolidation-sized) allocation happens, instead of `ceil(n / 16)`.
- `composeReleasesAllInputsOnFailure`: composes a list containing an already-consumed buffer and
  asserts every input `ByteBuf` ends at `refCnt() == 0`, guarding against a leak on the failure
  path.

### Verification

- `:micronaut-buffer-netty:test`, `checkstyleMain`/`checkstyleTest`: all green.
[...]

PR #12891, opened 2026-08-16
Fix quadratic response body assembly in NettyReadBufferFactory.compose

graemerocher:
This pull request replaces #12856 with the rebased change.

## Summary

* Fix quadratic response body assembly in `NettyReadBufferFactory.compose`.
* Handle GraalPy current-context views correctly in Python bridge execution frames.

The original PR's Python CI failure was caused by calling `Context.enter()` on a context returned by `Value.getContext()`/`Context.get()`, which is not enterable. The runtime now only leaves contexts it successfully entered and preserves execution tracking for already-current contexts.

## Verification

* `:micronaut-context-python:check`
* `:micronaut-context-python-netty:check`
* `:micronaut-inject-python:check`
* Formerly failing GraalPy regression test
* Python async/await documentation tests

Closes #12856.

### DIFF
--- a/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBufferFactory.java
+++ b/buffer-netty/src/main/java/io/micronaut/buffer/netty/NettyReadBufferFactory.java
@@ -24,6 +24,7 @@ import io.netty.buffer.ByteBufOutputStream;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.util.ReferenceCountUtil;
 import org.jspecify.annotations.Nullable;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


graemerocher:
Addressed both Copilot comments in commit 61fef609cd:\n\n* Removed the GraalVM exception-message matching; the context-entry fallback no longer depends on unstable/localized text, and leave remains conditional on successful entry.\n* Pre-sized the Netty component list when the input is a Collection.\n\nVerified with the formerly failing Python regression test, Netty compose test, and Spotless checks.

graemerocher:
SonarCloud identified the only PR-specific failure: new coverage was 76.47% in NettyReadBufferFactory.java because the non-Collection Iterable branch was untested. Added composeWithNonCollectionIterable() in commit 74abf9e3e2; the isolated Netty suite now passes 27 tests and Spotless passes.

yawkat (approved):

PR #12893, opened 2026-08-16
Fix native Python Throwable subclasses

n0tl3ss:
## Summary

GraalPy native cannot create a Python subclass of a concrete Java Throwable at runtime. Keep the Java superclass in compile-time transformed source for Micronaut metadata, but strip only concrete Java Throwable bases from executable runtime bytecode.

This leaves normal Java concrete inheritance unchanged and avoids broad native-image preservation flags.

## Verification

- `./gradlew :micronaut-inject-python:test --tests io.micronaut.python.processing.PythonAstParserTest.testRuntimeTransformStripsConcreteJavaThrowableBases -Ppython-ci --no-daemon --console=plain`
- Test passed with the guide `from java.lang import RuntimeException` syntax.

### DIFF
--- a/inject-python/src/main/resources/GRAALPY-VFS/io.micronaut/micronaut-inject-python/src/micronaut_transformer.py
+++ b/inject-python/src/main/resources/GRAALPY-VFS/io.micronaut/micronaut-inject-python/src/micronaut_transformer.py
@@ -614,14 +614,13 @@ def micronaut_annotation(name, repeated=None, annotationTypeTarget=False):
         Micronaut metadata and handler matching remain unchanged; only the
         executable runtime class is made a regular Python exception.
         """
-        base_name = self._base_name(base)
-        if not base_name:
-            return False
-        class_element = self.java_class_elements.get(base_name)
+        class_name = self._java_type_name(base)
+        class_element = self.callback_get_class_element(class_name) if class_name else None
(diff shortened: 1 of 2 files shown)


n0tl3ss:
Updated PR 12893 in commit a94c292358.\n\n- Runtime ImportFrom Java classes are tracked for executable runtime transformation.\n- Java Throwable bases are replaced with builtins.Exception, so imported Java Exception aliases cannot shadow the Python base.\n- Existing Python Exception bases are preserved without duplicate-base errors.\n- Added regression coverage for imported aliases, java.type(...), duplicate bases, and Java Exception shadowing.\n\nValidation: PythonAstParserTest 40/40 passed.

graemerocher (approved):

n0tl3ss:
Crema/native rationale: GraalPy native executes Python Throwable subclasses inside a closed-world image, so the generated Java throwable base must be retained and converted explicitly. JVM mode resolves that Java base dynamically and did not reproduce the failure. This keeps Python exception behavior correct without moving Python code to Java or adding broad native-image preserves.

Observed native error: `GraalPy native cannot create a Python subclass of a concrete Java Throwable at runtime` for the guide form `from java.lang import RuntimeException`; the focused parser regression uses that exact import.

PR #12895, opened 2026-08-17
Fix Python bridge casts for response subtypes

n0tl3ss:
The Python bridge currently casts every converted HttpResponse result to HttpResponse, even when the generated override declares a subtype such as MutableHttpResponse.

This produces a generated-source compilation error in Python guides that override Micronaut handlers returning MutableHttpResponse. Cast to the resolved declared return type instead.

Validated with :micronaut-inject-python:compileJava and :micronaut-inject-python:test.


graemerocher:
Needs a test

n0tl3ss:
Crema/native rationale: the native Python-to-Java bridge must return concrete HttpResponse subtypes through generated facades without relying on JVM cast/reflection fallback. JVM class loading handles these subtype conversions dynamically, so the defect was native-only. The fix keeps the behavior in Core rather than adding a guide workaround.

Observed native/compiler error: generated Python handlers declaring a subtype such as `MutableHttpResponse` were converted through the bridge as the base `HttpResponse`, causing the generated-source return conversion to fail. The JVM path resolves the declared subtype through normal class loading.

PR #12896, opened 2026-08-18
Diagnose concrete Java inheritance in native Python

n0tl3ss:
## Summary
- emit a clear native-runtime error before GraalPy attempts to subclass a concrete Java class
- retain supported Java-interface adaptation and existing Throwable handling
- add parser coverage for concrete Java classes and interfaces

## Validation
- `:micronaut-inject-python:compileTestJava`
- `:micronaut-inject-python:test --tests io.micronaut.python.processing.PythonAstParserTest` (task is disabled by this checkout; Gradle build succeeds)


n0tl3ss:
Crema/native rationale: GraalPy native cannot extend arbitrary concrete Java classes, while JVM mode supports that inheritance. This diagnostic makes the native limitation explicit at the bridge boundary so guides fail clearly and can use supported interfaces; it does not alter normal JVM behavior.

Observed native error: `SystemError: Java Class can be extended only in JVM mode.` This was raised when a Python class attempted to use a concrete Java superclass; interface bases remain supported.

PR #12899, opened 2026-08-19
Add Sec-Fetch-* HTTP Headers

sdelamo:
see. https://www.w3.org/TR/fetch-metadata/


graemerocher:
should this go into 5.2.x?

sdelamo:
@graemerocher I target it to 5.1.x because it can be used for security.

graemerocher (approved):

PR #12908, opened 2026-08-24
Reduce critical wait times in PythonPool

yawkat:
When many requests come in at the same time, they could all reach borrow() simultaneously, fail to find a pooled context, and then pile up in createBorrowedPooledContext(). This could lead to many seconds of stalled requests, even if contexts are returned to the pool in the meantime.

This change reworks the pooling mechanism to use a single lock for both the pooledQueue and for limiting the context creation. This means that while one thread is creating a context, other threads can still reuse newly released contexts and continue.


graemerocher:
There are failures?

PR #12909, opened 2026-08-24
A bean introspection describes only the constructor it instantiates beans with

dstepanov:
## What

`BeanIntrospection` exposes one constructor. Everything a caller can learn about the constructors of an introspected type comes from `getConstructor()`, which is synthesised from `getConstructorArguments()` and describes the single constructor the introspection instantiates beans with. Nothing in a generated introspection carries the annotations of any other constructor, or the annotations of its parameters.

A specification that *describes* a type rather than instantiating it needs all of them. Jakarta Validation names a constructor by its parameter types and describes each one: its parameters, its cross-parameter constraints, its return value, and the group conversions on both. `jakarta.validation.metadata.BeanDescriptor` has `getConstrainedConstructors()` and `getConstraintsForConstructor(Class<?>...)`; with one constructor available the validator returns `null` for any constructor the introspection did not pick, and the TCK then fails with a `NullPointerException` on `ConstructorDescriptor`.

This adds `BeanIntrospection.getConstructors()`, returning every described constructor with `getConstructor()` first:

```java
/**
 * @return Every described constructor of the type, getConstructor() first
 */
default List<BeanConstructor<T>> getConstructors() {
    return List.of(getConstructor());
}
```

The `default` implementation keeps every existing introspection valid, and callers that do not opt in see no change.

## Opting in

Writing every constructor of every introspected type is not free, so it is gated. There are two ways in, and they compose:

**Per constructor, with `@Executable`** — the annotation now accepts `ElementType.CONSTRUCTOR`, so a single constructor can ask to be described:

```java
@Introspected
class Order {
    Order() {}
    @Executable Order(String name) {}   // described
    Order(String name, int quantity) {} // not described
}
```

The gate reads a *declared* stereotype, so a type-level `@Executable` — which means "all public methods are executable" — does not sweep every constructor into the introspection. There is a test pinning that.

**Per type, with `@Introspected(constructors = true)`** — describes every declared constructor. This is what a module that needs the full picture (the validation processor, via `IntrospectedValidationIndexesVisitor`) asks for, and nobody else pays for it.

## How it works

- **API** — `getConstructors()` on `BeanIntrospection`, and the `constructors()` member on `@Introspected`.
- **Runtime** — a `BeanConstructorRef` holder in `AbstractInitializableBeanIntrospection` carrying annotation metadata, arguments and a dispatch index, mirroring the existing `BeanMethodRef` pattern, plus an `instantiateConstructorInternal(int, Object[])` hook that generated code overrides with a switch. `BeanConstructor` already had everything else needed, so no new type was introduced.
- **Writer** — `BeanIntrospectionWriter` emits one `BeanConstructorRef` per described constructor, each with its own `AnnotationMetadata` and its own `Argument[]`, built through the same `pushBuildArgumentsForMethod` path bean properties already use so parameter and type-use annotations survive.
- **Cost when unused** — when nothing opts in the writer emits no extra field, no extra methods, and calls the original super constructor. The generated bytecode is byte-for-byte what it was.

## Ordering

`getConstructor()` comes first, per the contract. When the introspection is built from a static creator rather than a constructor, that creator is prepended so the contract still holds there too.

## Reviewer notes

**The opt-in is purely additive metadata.** The risk worth reviewing is whether describing extra constructors could shift which constructor the introspection actually builds beans with. The second commit pins that it does not: each test builds the same type twice, with and without the opt-in, and asserts `getConstructor()`, `getConstructorArguments()`, `isBuildable()` and the result of `instantiate()` are identical. The `instantiate()` check reads a property off the constructed object, so it proves the right constructor body ran rather than only that the metadata looks unchanged. Covered for an explicit `@Creator`, a Kotlin secondary constructor, and Kotlin default parameter values.

**The tests were checked for teeth.** I temporarily broke the ordering logic in the writer and re-ran: all five Java invariance tests failed, along with one earlier test. Two "described first" assertions were also vacuous on first writing — the instantiating constructor happened to be first in declaration order anyway — so those types now declare a non-public constructor first, and the primary genuinely has to be moved to the front for the assertion to hold.

**`@since` tags** are set to `5.2.0` based on the current branch version. If this lands in a different release they need updating in three places: `BeanIntrospection.getConstructors()`, `Introspected.constructors()`, and the new `AbstractInitializableBeanIntrospection` members.

**Out of scope.** `ReflectiveIntrospection` — which already carries this exact contract on the reflective side of the bridge — does not exist on this branch, so the two micronaut-validation special cases (`IntrospectedBeanDescriptor.constructors()` and `ReflectiveExecutables.beanConstructor`) are downstream of this change rather than part of it.

## Tests

Java, Groovy and Kotlin, covering: every declared constructor described; a constructor carrying its own annotations; a constructor carrying its parameters' annotations, including type-use annotations on type arguments; instantiating through a described constructor; an introspection that did not opt in still describing exactly one; the `@Executable` gate and the type-level `@Executable` non-regression; and the constructor-selection invariance above. Kotlin adds secondary constructors, data classes and default parameter values; Groovy adds its own constructor shapes.

Full rerun across `core`, `inject`, `inject-java`, `inject-java-test`, `inject-groovy` and `inject-kotlin`: 14,621 tests, no failures.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/beans/visitor/BeanIntrospectionWriter.java
@@ -33,6 +33,7 @@ import io.micronaut.inject.annotation.AnnotationMetadataHierarchy;
 import io.micronaut.inject.annotation.AnnotationMetadataReference;
 import io.micronaut.inject.annotation.MutableAnnotationMetadata;
 import io.micronaut.inject.ast.ClassElement;
+import io.micronaut.inject.ast.ConstructorElement;
 import io.micronaut.inject.ast.Element;
 import io.micronaut.inject.ast.ElementQuery;
 import io.micronaut.inject.ast.EnumConstantElement;
@@ -98,6 +99,7 @@ final class BeanIntrospectionWriter implements OriginatingElements, Buildable<Li
     private static final String FIELD_CONSTRUCTOR_ARGUMENTS = "$CONSTRUCTOR_ARGUMENTS";
     private static final String FIELD_BEAN_PROPERTIES_REFERENCES = "$PROPERTIES_REFERENCES";
     private static final String FIELD_BEAN_METHODS_REFERENCES = "$METHODS_REFERENCES";
+    private static final String FIELD_BEAN_CONSTRUCTORS_REFERENCES = "$CONSTRUCTORS_REFERENCES";
     private static final String FIELD_ENUM_CONSTANTS_REFERENCES = "$ENUM_CONSTANTS_REFERENCES";
     private static final java.lang.reflect.Method FIND_PROPERTY_BY_INDEX_METHOD =
         ReflectionUtils.getRequiredInternalMethod(AbstractInitializableBeanIntrospection.class, "getPropertyByIndex", int.class);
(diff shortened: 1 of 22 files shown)


dstepanov:

**Downstream revisions**

Caveats:

graemerocher (approved):

PR #12910, opened 2026-08-24
Add a reflection bridge: Micronaut metadata over java.lang.reflect

dstepanov:
## What this is

A specification that has to handle any class — Jakarta Validation, Jakarta REST, CDI — ends up reading annotations, parameters and properties reflectively for the types the annotation processors never saw. micronaut-validation, micronaut-jaxrs and ODI each wrote that conversion themselves, and each wrote it differently: one adds stereotypes, another none, one expands repeatable containers, another does not, and none of them serves defaults the way the generated metadata does.

This is one implementation in `micronaut-inject`, shaped like the metadata the processors generate, so that code written against generated metadata also works for a type that has none. Everything is `@Experimental`.

| Class | Lines | What it is |
| --- | ---: | --- |
| `inject/annotation/ReflectionAnnotationMetadataBuilder` | 453 | `AnnotationMetadata` from an `AnnotatedElement` |
| `inject/annotation/ReflectionAnnotationCustomizer` | 57 | the runtime counterpart of the processors' mappers and transformers |
| `context/AnnotationReflectionUtils` | +120 | the `Argument` factories over `AnnotatedType`, `Parameter`, `Field`, `Executable` |
| `inject/reflection/ReflectionBeanIntrospection` | 838 | a `BeanIntrospection` over a `Class` |
| `inject/reflection/SupplementedBeanIntrospection` | 382 | a generated introspection completed by a reflective one |
| `inject/reflection/ReflectionBeanIntrospector` | 113 | generated introspections first, reflection only for the rest |
| `inject/reflection/ReflectiveIntrospection` | 120 | what a reflective introspection knows beyond the contract |
| `inject/reflection/ReflectionExecutableMethod` | 86 | an `ExecutableMethod` over a `Method` |
| `inject/reflection/ReflectionExecutables` | 149 | `Method`/`Constructor` → the best metadata available |
| `inject/reflection/IntrospectedExecutableMethod` | 68 | an `ExecutableMethod` over a `BeanMethod` |
| `inject/reflection/MethodHierarchy` | 429 | what each level of a method hierarchy declares, apart and merged |
| `inject/annotation/AnnotationMetadataSupport` | +9 | a registered annotation type resolved through the caller's loader |

`MethodHierarchy` answers the question every consumer asks straight after "describe this class". A specification that describes a method reads it together with the methods it overrides, and has rules about which level may declare what — so it needs the levels **apart**, not merged. The walk asks each level's introspection what that level declares itself (`findDeclaredMethod` on a `ReflectiveIntrospection`, the bean method filtered by declaring type on a generated one) and returns the local declaration, the exact one where a level can be told apart, the inherited ones nearest first, and the three merged into one view of the annotations, the parameters and the return type.

A declaration read reflectively is `exact`: Java reflection does not inherit the annotations of a method. One read from a generated introspection is not, because that metadata already merges what the method overrides. When #12917 lands that branch answers exactly too and nothing else moves — the declaring-type filter it applies becomes redundant rather than wrong.

The annotation mappers, transformers and remappers of the processors are **not** applied: they live in `micronaut-core-processor`, which is not on a runtime classpath. `ReflectionAnnotationCustomizer` is the runtime hook for the extensions that need one — a processor extension deriving a member from the annotation type, such as the validator classes copied from a `@Constraint` meta-annotation, has to be applied to the reflective metadata too, or code reading the member behaves differently for a type the processor never saw.

`BeanIntrospector.SHARED` is unchanged. `ReflectionBeanIntrospector` serves the generated introspections of a delegate first and reflects only for the others, so a caller opts into reflection explicitly.

## Why

Two consumers are waiting on it:

- **micronaut-validation** deletes its `validation-reflection` module — `ReflectionValidator` and eleven supporting classes, 9852 deletions — and runs one `DefaultValidator` over a supplementing `ReflectionBeanIntrospector`. The Jakarta Validation TCK passes on that branch: `jakartaTck` 1054, `jakartaTckIntrospection` 961, `jakartaTckReflection` 1019, no failures.
  It also drops the three classes it wrote in core types with no validation concept in them: `ReflectiveExecutables` (147 lines, now `ReflectionExecutables`), `IntrospectedExecutableMethod` (63, now the copy here) and the traversal half of `ExecutableHierarchy` (~180 of 486, now `MethodHierarchy`). What stays there is the Jakarta declaration rules over the levels this PR resolves.
- **micronaut-jaxrs** needs the same thing for its reflective path (micronaut-jaxrs#641). Three sites in `jaxrs-reflection` reflect over methods and constructors directly today and would otherwise reimplement `ReflectionExecutables`.

## Tests

68 cases in `micronaut-inject`, all green; the module's suite is 380.

| Spec | Cases | Covers |
| --- | ---: | --- |
| `ReflectionBridgeSpec` | 11 | the metadata builder and a walk through each class |
| `ReflectionArgumentFactoriesSpec` | 5 | the six `AnnotationReflectionUtils` factories |
| `ReflectionBeanIntrospectionSpec` | 12 | constructor selection, instantiation, property members, the builder |
| `SupplementedBeanIntrospectionSpec` | 11 | generated-plus-reflective merging |
| `ReflectionExecutablesSpec` | 8 | the resolution strategy |
| `MethodHierarchySpec` | 15 | the levels of a method hierarchy |
| `ReflectionAnnotationCustomizerSpec` | 3 | the service hook |
| `AnnotationTypeClassLoaderSpec` | 3 | the annotation-type/class-loader fix |

Each case asserts the reflective metadata is shaped **like the generated metadata**, which is the point of the whole thing: code written against one has to work against the other.

<details>
<summary>The cases in full</summary>

**The metadata builder** — a class's declared annotations, their stereotypes, the repeated ones and the inherited ones; annotations composing another flattened next to the declared one; a meta-annotation as a stereotype with its values; a super class's annotation present but not *declared*; every occurrence of a repeated annotation with its values; a member equal to its default carried in the defaults rather than stored; class, enum and nested annotation members stored as `AnnotationClassValue`, the enum name and a nested `AnnotationValue`; a package-private annotation type's members still read.

**The argument factories** — `argumentOf(Field)` carrying the type-use annotations of *every* level, so `Map<String, @Tag List<@Tag String>>` keeps both; `argumentOf(Parameter)` carrying the parameter's own annotations and those of its type arguments; `argumentsOf(Executable)` over methods and constructors alike, returning `Argument.ZERO_ARGUMENTS` for an executable with no parameter; `returnArgumentOf(Method)` reading the annotated return type; `argumentOf(name, AnnotatedType)` letting the caller name the argument.

**The reflective introspection** — constructor selection following the processor's four rules, one case each (a `@Creator` wins even when not public; else the only public constructor; else the no-arg one among several public; else the widest declared one); every declared constructor listed, the selected one first; what instantiation refuses (wrong argument count, a null for a non-nullable argument checked strictly) and what it lets through; an interface having no constructor to invoke; `isIntrospectable` accepting a class and an interface, refusing a primitive, an array, an enum, an annotation and anything under `java.`; the indexed properties by stereotype and by annotation value; a property made of every member declaring it, most specific first, so a field shadowing a super class's field yields both with their own annotations; a member reading the value it holds, and a setter throwing because it yields none; the builder by name, by index, from an existing bean and by conversion.

**Generated plus reflective** — the type, metadata and indexed properties coming from the generated introspection; a generated method served as generated but reporting the class reflection knows declares it; a method the processor left out (public but not `@Executable`) coming from reflection, and a method on both sides appearing once; `findDeclaredMethod` answering what one level declared, which a generated introspection cannot because it merges the annotations of the methods a method overrides; a described constructor returned untouched; a constructor described with no metadata gaining the arguments reflection read; every constructor listed, the generated one first; a generated property carrying the type arguments reflection reads from every member, so an interface getter declaring `List<@Tag String>` over a plain `List<String>` field keeps the annotation; the property members and the supplementing introspector.

**The resolution strategy** — a method of a type with a bean definition resolving to that definition's `ExecutableMethod`; one of an introspected type without a bean definition resolving through the `BeanMethod`; one of a type with neither resolving to `ReflectionExecutableMethod`; a method declared by a super type resolving against the type declaring it rather than against a sub type bean overriding it; a constructor of an introspected type resolving to the introspection's `BeanConstructor`; one of a non-introspected type to a constructor over the reflective metadata; a constructor the introspection does not describe read from the constructor itself; one a `ReflectiveIntrospection` knows coming from its `getConstructors()`.

**The method hierarchy** — a reflective hierarchy reporting every level nearest first, super classes before the interfaces of each level, every declaration exact; the merged views holding the method annotations, the parameter annotations and the type-use annotations of the return type arguments of all four levels, the local one winning; a type declaring the method in two branches being `parallel` while one with a single super declaration is not; a type that inherits without overriding reading the declaration of the type that declares it, and the walk starting *there*, so an interface the sub type does not implement is not reported; an interface reached by two branches reported once; a method inherited as an interface default found on the interface declaring it; an interface hierarchy reporting only what the interface declares, returning the local metadata and arguments identically rather than merging one level; a `java.lang.reflect.Method` resolving against the type declaring it; a type with no such method rejected; a generated introspection's declaration **not** exact, so the declared view falls back to the local one while the merged view still holds both levels; the same types under a supplementing introspector telling the declarations apart, the child carrying only its own annotation; a bean definition's `ExecutableMethod` read without the annotations of its class; `declaredBy` empty for a type the shared introspector does not know and present for one the reflective introspector does.

**The service hook** — a registered customizer deriving a member of the annotation it supports; seeing the values as the builder converts them, so a member left at its default is not among them; an unsupported annotation untouched. The customizer is registered as a real service in the test module and supports exactly one annotation type, so it changes nothing else in the suite.

**The class-loader fix** — `AnnotationMetadataSupport.getAnnotationType` caches annotation types by name. A child-first deployment loader defines its own copy of the same annotation and the caller compares against *that* copy, so returning the cached one silently fails every comparison. Once registered from one loader, asking with a loader that defines its own copy returns that loader's copy while the original loader still gets the original; a loader with no copy falls back to the registered type; a type the bootstrap loader defines cannot be shadowed.

</details>

### Verified, not assumed

The class-loader case was checked against a reverted fix: it fails without the nine-line change, so it is a regression test rather than a description.

Two shapes are worth a reviewer's attention, because the obvious test passes vacuously for both:

- `SupplementedBeanIntrospection.DescribedBeanConstructor` engages only when the generated constructor's metadata is empty, and the processor records constructor annotations whenever it sees them — including for `@Introspected(classes = …)`. Its value is therefore not the metadata it substitutes (empty on both sides in that case) but the **arguments** it merges: the type-use annotations on constructor parameters that the generated introspection drops. The test is built on exactly that shape.
- An inherited, non-overridden method already reports the super type as its `getDeclaringType()`, so only an **overriding** sub type exercises the declaring-type guard in `ReflectionExecutables.executableMethod`. A fixture that merely inherits proves nothing.

### One known javadoc slip

`ReflectionBeanIntrospection.isIntrospectable`'s javadoc says the type must not be "a primitive, an array, an interface, an annotation, an enum or a type of the JDK", but interfaces are deliberately allowed — the comment on the line below says so, and an interface is introspected for its declarations. The javadoc should drop "an interface". Left as is pending review.

## How to run

```
./gradlew :micronaut-inject:test \
  --tests 'io.micronaut.inject.reflection.*' \
  --tests '*ReflectionAnnotationCustomizer*' --tests '*AnnotationTypeClassLoader*'
```

## Rebased onto 5.2.x

`BeanIntrospection.getConstructors()` landed in 5.2.x (#12909), so `ReflectiveIntrospection` now narrows it rather than declaring it — one `@Override`, in its own commit.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
+++ b/inject/src/main/java/io/micronaut/context/AnnotationReflectionUtils.java
@@ -18,7 +18,7 @@ package io.micronaut.context;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.reflect.ClassUtils;
-import io.micronaut.core.reflect.ReflectionUtils;
+import io.micronaut.inject.annotation.ReflectionAnnotationMetadataBuilder;
 import io.micronaut.core.type.Argument;
 import io.micronaut.inject.annotation.AnnotationMetadataHierarchy;
 import io.micronaut.inject.annotation.MutableAnnotationMetadata;
@@ -31,8 +31,11 @@ import java.lang.reflect.AnnotatedParameterizedType;
 import java.lang.reflect.AnnotatedType;
 import java.lang.reflect.AnnotatedWildcardType;
 import java.lang.reflect.Array;
+import java.lang.reflect.Executable;
+import java.lang.reflect.Field;
 import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
@@ -95,6 +98,99 @@ public final class AnnotationReflectionUtils {
      * @param superType The supertype that we want to get the type information for
      * @return The annotated generic supertype, with the same raw type as {@code superType}
      */
+    /**
+     * Converts an annotated type to an {@link Argument}: the raw type, the type arguments recursively, and the
+     * type-use annotations of each level as metadata built by
+     * {@link io.micronaut.inject.annotation.ReflectionAnnotationMetadataBuilder}.
+     *
+     * @param annotatedType The annotated type
+     * @return The argument
+     * @since 5.1
+     */
+    public static Argument<?> argumentOf(AnnotatedType annotatedType) {
+        return toArgument(null, annotatedType, Map.of());
+    }
+
+    /**
+     * Converts an annotated type to a named {@link Argument}.
+     *
+     * @param name          The name of the argument, can be {@code null}
+     * @param annotatedType The annotated type
+     * @return The argument
+     * @since 5.1
+     */
+    public static Argument<?> argumentOf(@Nullable String name, AnnotatedType annotatedType) {
+        return toArgument(name, annotatedType, Map.of());
+    }
+
+    /**
+     * Converts a parameter to an {@link Argument} named after it, whose metadata holds the annotations of the
+     * parameter and the type-use annotations of its type.
+     *
+     * @param parameter The parameter
+     * @return The argument
+     * @since 5.1
+     */
+    public static Argument<?> argumentOf(Parameter parameter) {
+        return withElementAnnotations(toArgument(parameter.getName(), parameter.getAnnotatedType(), Map.of()), parameter);
+    }
+
+    /**
+     * Converts a field to an {@link Argument} named after it, whose metadata holds the annotations of the
+     * field and the type-use annotations of its type.
+     *
+     * @param field The field
+     * @return The argument
+     * @since 5.1
+     */
+    public static Argument<?> argumentOf(Field field) {
+        return withElementAnnotations(toArgument(field.getName(), field.getAnnotatedType(), Map.of()), field);
+    }
+
+    /**
+     * Converts the parameters of a method or constructor to {@link Argument}s.
+     *
+     * @param executable The method or constructor
+     * @return The arguments, in the order of the parameters
+     * @since 5.1
+     */
+    public static Argument<?>[] argumentsOf(Executable executable) {
+        Parameter[] parameters = executable.getParameters();
+        if (parameters.length == 0) {
+            return Argument.ZERO_ARGUMENTS;
+        }
+        Argument<?>[] arguments = new Argument[parameters.length];
+        for (int i = 0; i < parameters.length; i++) {
+            arguments[i] = argumentOf(parameters[i]);
+        }
+        return arguments;
+    }
+
+    /**
+     * Converts the return type of a method to an {@link Argument} carrying the type-use annotations of the
+     * return type. The annotations of the method itself belong to the method, not to its return type.
+     *
+     * @param method The method
+     * @return The argument
+     * @since 5.1
+     */
+    public static Argument<?> returnArgumentOf(Method method) {
+        return toArgument(null, method.getAnnotatedReturnType(), Map.of());
+    }
+
+    private static Argument<?> withElementAnnotations(Argument<?> argument, AnnotatedElement element) {
+        AnnotationMetadata own = ReflectionAnnotationMetadataBuilder.build(element);
+        if (own == AnnotationMetadata.EMPTY_METADATA) {
+            return argument;
+        }
+        return Argument.of(
+            argument.getType(),
+            argument.getName(),
+            combine(argument.getAnnotationMetadata(), own),
+            argument.getTypeParameters()
+        );
+    }
+
     @Nullable
     private static AnnotatedType findAnnotatedSupertype(AnnotatedType subType, Class<?> superType) {
         Class<?> raw = getRawType(subType.getType());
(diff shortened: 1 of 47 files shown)


dstepanov:

I built this PR's merge-into-`5.1.x` commit, published core to an isolated local Maven repository, and built both downstream projects against it.

**Downstream revisions**

**What was run** (baseline for comparison: unmodified `origin/5.1.x`)

Caveats:

dstepanov:
Replaced by https://github.com/micronaut-projects/micronaut-core/pull/12977

PR #12911, opened 2026-08-24
Extend @AliasFor to support OverridesAttribute-style member overrides

dstepanov:
## Summary

Extends core's annotation alias machinery so that constraint-composition semantics like `jakarta.validation.OverridesAttribute` can be expressed with `@AliasFor` and computed at build time in the annotation metadata:

- **Member-annotation remapping**: when an annotation member carries no literal `@AliasFor`/`@Aliases`, the registered `AnnotationTransformer`s are applied to the member's annotations, and any produced `@AliasFor`/`@Aliases` values are processed as if declared directly. A produced alias without a `member` value defaults to the annotated member's own name (`OverridesAttribute`'s empty-`name` semantics, which a transformer cannot know). This lets annotations with no Micronaut dependency participate in alias resolution.
- **New `@AliasFor` members**:
  - `applyDefault()` — the alias also applies when the member is not explicitly set, using the member's default value. Opt-in, so existing alias behavior is unchanged. Deliberately uncached: a `hasAnnotations(member)` short-circuit keeps the per-usage cost low for members without annotations.
  - `index()` — for a repeatable aliased annotation declared multiple times as a stereotype, selects the occurrence (declaration order) the alias applies to; `-1` (default) applies to all occurrences; an out-of-range index is dropped.
- **Stereotype reconciliation**: alias-introduced annotations are reconciled with the declared stereotypes in `addStereotypes`. When the target is declared, the aliased members override the declared values in place — for repeatable targets the occurrence selected by `index` — instead of accumulating another occurrence in the repeatable container. Targets not declared as stereotypes are still contributed as new stereotypes with higher priority, as before.

## Motivation

Enables implementing Jakarta Validation constraint composition attribute overrides without runtime reflection: a ~50-line transformer in micronaut-validation remaps `@OverridesAttribute` to `@AliasFor` and the composed constraint's usage metadata carries the overridden values. Verified end-to-end against the Jakarta Validation 3.1.1 TCK `ConstraintCompositionTest` via a composite build with micronaut-validation: all 8 OverridesAttribute-related methods pass (repeated-`@Pattern` `constraintIndex` targeting, defaults propagation, default-name resolution, explicit `List` containers), full TCK suite green.

## Notes for reviewers

- For non-repeatable declared targets the resulting metadata is identical to the previous first-writer-wins merge; the behavior change is for repeatable declared targets, where an alias-introduced annotation previously accumulated as an extra occurrence with only the aliased member set — arguably never the intended outcome.
- Overrides merge after the declared stereotype has been processed, so overriding a member of a composed-of-composed annotation does not cascade into that annotation's own stereotype tree.
- `RemapMemberAnnotationToAliasForSpec` covers eight scenarios (override precedence, explicit names, defaults, `index` on repeated and `List`-container stereotypes, index-less all-occurrence overrides, out-of-range indexes, undeclared targets). Docs added to the "Aliasing / Mapping Annotations" guide section.

### DIFF
--- a/src/main/docs/guide/ioc/annotationMetadata.adoc
+++ b/src/main/docs/guide/ioc/annotationMetadata.adoc
@@ -98,19 +98,14 @@ Two additional members refine this behaviour:
 
 For example:
 
-.Overriding a repeatable stereotype member
-[source,java]
-----
-@Pattern(regexp = ".....")
-@Pattern(regexp = "bar")
-@interface ComposedZip {
+snippet::io.micronaut.docs.inject.aliasfor.ComposedZip[tags="imports, class", indent=0, title="Overriding a Member of a Declared Stereotype"]
 
-    @AliasFor(annotation = Pattern.class, member = "regexp", index = 1, applyDefault = true)
-    String regex() default "\\d*";
-}
-----
+<1> `@Size` is declared as a stereotype with `max` set to `5`
+<2> The `max()` member aliases `Size.max`, overriding the declared value
 
-Here any use of `@ComposedZip` produces metadata containing two `@Pattern` stereotypes with the `regexp` members `"....."` and `"\\d*"` — the second occurrence is overridden by the `regex()` default, and by an explicit value such as `@ComposedZip(regex = "[0-9]+")` when set.
+Any use of `@ComposedZip` therefore produces `@Size(min = 5, max = 10)` in the metadata — `max` comes from the `max()` default because `applyDefault` is `true` — while `@ComposedZip(max = 20)` produces `@Size(min = 5, max = 20)`.
+
+NOTE: `index()` is only computed for annotations processed from Java and Groovy sources. The Kotlin (KSP) annotation metadata builder does not extract repeatable stereotype containers, so a repeatable annotation declared multiple times on a Kotlin annotation class cannot be targeted by index.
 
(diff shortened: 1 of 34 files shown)


dstepanov:

**Downstream revisions**

Caveats:

src/main/docs/guide/ioc/annotationMetadata.adoc
```diff
@@ -87,6 +87,33 @@ include::http-client-core/src/main/java/io/micronaut/http/client/annotation/Clie
...
+
+Two additional members refine this behaviour:
+
+* `applyDefault()` - by default an alias only applies when the member is explicitly set at the use site. With `applyDefault = true` the alias also applies the member's *default* value. This allows modelling semantics such as `jakarta.validation.OverridesAttribute`, where the overriding member's default replaces the value declared on the composed constraint.
+* `index()` - when the aliased annotation is repeatable and declared multiple times as a stereotype, `index` selects which occurrence (in declaration order, starting at zero) the alias overrides. The default of `-1` applies the alias to every declared occurrence, and an index outside the declared occurrences is ignored.
+
+For example:
+
+.Overriding a repeatable stereotype member
+[source,java]
```
graemerocher: please make this a multi language snippet with all supported languages

graemerocher (changes requested):

src/main/docs/guide/ioc/annotationMetadata.adoc
dstepanov: Done in 540e319ccf — the inline Java block is now a `snippet::` macro backed by real sources in `test-suite`, `test-suite-groovy` and `test-suite-kotlin`, so it renders Java/Groovy/Kotlin tabs.

One change to the example itself: the previous illustration used `index` on a repeatable stereotype, which cannot be expressed in Kotlin — the KSP builder does not extract repeatable stereotype containers from a Kotlin annotation class, so the alias lands as a separate stereotype rather than overriding an occurrence. The snippet now overrides a member of a non-repeatable declared stereotype (`annotation` + `member` + `applyDefault`), which behaves identically in all three languages and is asserted by a `ComposedZipSpec` in each suite. `index()` remains documented in prose with a note about the Kotlin limitation.

graemerocher (approved):

PR #12915, opened 2026-08-24
Update actions/setup-java to v6

brunoborges:
Updates all active workflow references to the pinned actions/setup-java v6.0.0 release.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #12916, opened 2026-08-25
fix(processor): stop generating an imported type's introspection twice

dstepanov:
## What

Naming a type on `@Introspected(classNames = …)` or `@ClassImport` could generate its introspection twice and fail the build. The second `visitServiceDescriptor` call asks the `Filer` to recreate a resource it already created for this compilation:

```
error: Failed to generate class: 'test.$test_Outer$Nested$Introspection':
  Unable to generate Bean entry at path:
  META-INF/micronaut/io.micronaut.core.beans.BeanIntrospectionReference/test.$test_Outer$Nested$Introspection
```

This matters because a nested type carrying no annotation of its own is never handed to the processor by the round, so `classNames`/`@ClassImport` is the documented way to reach it — and that is exactly where the duplicate appears.

Two independent defects contributed.

### 1. Imported elements were visited once per root element

`TypeElementVisitorProcessor.process` accumulated imports into `extraClasses` and re-added the **whole accumulated list** to the visit list on every iteration:

```java
List<ClassElement> extraClasses = new ArrayList<>();
for (JavaClassElement javaClassElement : new ArrayList<>(javaClassElements)) {
    extraClasses.addAll(VisitorUtils.collectImportedElements(...));
    javaClassElements.addAll((Collection) extraClasses);
}
```

A file with three top-level types made an imported class visit **three times** — measured, not inferred. Only newly discovered imports are added now, and a class already queued for this round is not queued again. The equivalent Python processor already guarded against this; the Java one did not.

### 2. Introspections written on behalf of another element were untracked

`isIntrospected` looks for `<package>.$<Simple>$Introspection`. An introspection generated on behalf of another element is written as `$<originating class>$<Simple>$Introspection`, so the lookup never matches and the guard falls back to the in-memory `processed` set — which is keyed on the introspected type. Not every write path records its target there: `handleBuilder` writes an introspection for the **builder** type but records the type being **built**.

The introspections written during a compilation are now tracked by generated class name alongside the type they were generated for. Writing the same one twice is a no-op; a genuine name collision between two *different* types raises a readable `ProcessingException` instead of the cryptic Filer error, per the narrower of the two options in the issue.

### 3. The test harness could not see any of this

`InMemoryJavaFileManager.createResource` collapsed duplicate writes with `computeIfAbsent`, so no `AbstractTypeElementSpec` test could ever observe a double write. It now refuses to recreate a resource, the way the real javac `Filer` does. This is what turns the shared-builder case below into a genuine regression test.

## Tests

`BeanIntrospectionSpec` — nested type via `classNames`, nested type via `@ClassImport`, one type imported by two importers, and two `@Introspected` types sharing a `builderClass`. The last **fails without the visitor fix**, reproducing the reported error verbatim:

```
error: Failed to generate class: 'test.$test_SharedBuilder$Introspection': Unable to generate Bean entry at path: META-INF/micronaut/io.micronaut.core.beans.BeanIntrospectionReference/test.$test_SharedBuilder$Introspection
```

`ImportTypeElementSpec` — two visit-count cases. Both **fail without the processor fix** (`3 != 1`).

## Notes for reviewers

- **The strict Filer is the change most worth a second look.** 7018 tests across `inject-java`, `inject-java-test`, `context`, `http-validation`, `inject-groovy` and `inject-kotlin` pass with it, so nothing in the codebase was relying on the silent overwrite — but it will now surface duplicate metadata writes anywhere else they exist.
- **One correction to the issue's root-cause analysis.** The `isIntrospected` long-form blindness is real, but I could not make it produce a duplicate on its own — `processed` covers it within a compilation. The failure needs a write path that bypasses `processed` entirely, which is what the builder path does.
- **`isIntrospected` was deliberately left alone.** Broadening its classpath lookup to guess long-form names risks *skipping* introspections on incremental recompiles, where the previous compilation's output is still on the classpath. A missing introspection is a worse failure than a duplicate.
- The `@Introspected` visitor fix lives in `core-processor`, so it covers Java, Groovy and Kotlin. The processor fix is Java-only; Groovy and Kotlin have no `@ClassImport` collection path.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
+++ b/core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
@@ -49,8 +49,10 @@ import java.io.OutputStream;
 import java.lang.annotation.Annotation;
 import java.lang.annotation.RetentionPolicy;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -76,6 +78,15 @@ public class IntrospectedTypeElementVisitor implements TypeElementVisitor<Object
     private static final String ANN_LOMBOK_BUILDER = "lombok.Builder";
 
     private final Set<String> processed = new HashSet<>();
+    /**
+     * The introspections written during this compilation, keyed by the generated introspection class
+     * name and holding the name of the type the introspection was generated for. An introspection
+     * generated on behalf of another element (via {@link Introspected#classNames()},
+     * {@link Introspected#classes()} or {@link io.micronaut.context.annotation.ClassImport}) is not
+     * named after the type it introspects, so this is the only reliable way to detect that the same
+     * introspection is about to be written twice.
+     */
+    private final Map<String, String> writtenIntrospections = new HashMap<>();
 
     @Override
     public int getOrder() {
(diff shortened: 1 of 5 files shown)


dstepanov:

**Downstream revisions**

Caveats:

graemerocher (approved):

PR #12917, opened 2026-08-25
Make the declared only annotation metadata usable

dstepanov:
## What

`AnnotationMetadata.getDeclaredMetadata()` returns a view of what an element declares, without what it inherits. `DefaultAnnotationMetadata` built that view with `allAnnotations` and `allStereotypes` set to `null`:

```java
return new DefaultAnnotationMetadata(declaredAnnotations, declaredStereotypes, null, null, ...);
```

Every lookup that reads those maps returns nothing on the result — `getAnnotationValuesByType`, `getAnnotationValuesByName`, `stringValue`, and the `Foo$List` repeatable container mapping — and `isEmpty()` reports the view as empty even when the element declares annotations. The declared maps are always a subset of the all maps (`addDeclaredAnnotation` writes into both), so the view is now built with `all` = `declared`.

Three smaller things come with it:

- `hasEvaluatedExpressions` is carried over to the view; it was dropped.
- `annotationsByStereotype` is narrowed to the declared annotations, so a method that declares nothing no longer reports `Constraint -> [NotNull]` from what its super method declared — and `getAnnotationValuesByStereotype` no longer collects `null` entries for annotations absent from the declared maps.
- `MutableAnnotationMetadata` overrides `getDeclaredMetadata()`. The build time metadata resolves repeatable containers from its own `annotationRepeatableContainer` map, for annotations that aren't on the compiled sources' classpath, so its view is a clone: it keeps that state and stays detached from later mutations of the element.

`getAnnotationValuesByName` also ended with `return List.of()` instead of the values it had just resolved and cached in its non-repeated fallback branch — `getAnnotationValuesByType` returns them from the identical branch.

`MethodElement` gains `getDeclaredMethodAnnotationMetadata()`, next to the existing `getMethodAnnotationMetadata()` which is documented as including the super methods. No new API is needed for parameters: the builder already tags both super-method and overridden-parameter annotations as not declared (`isDeclared = currentElement == element`), so `parameter.getAnnotationMetadata().getDeclaredMetadata()` answers correctly once the view works.

## Why

Given a method that overrides another, there was no way to ask for the annotations *this* declaration carries while keeping repeatable annotations readable. Jakarta Validation needs exactly that in four places — `ConstraintFinder.lookingAt(Scope.LOCAL_ELEMENT)`, `declaredOn(ElementType...)`, the implicit group of a constraint, and the declarations the specification forbids across a hierarchy. Without it, micronaut-validation raises `ConstraintDeclarationException: Parameter constraints cannot be added in overriding or implementing methods` for overriding methods that declare nothing, because "declared" reads as more than the type declares.

## Tests

`DeclaredMethodAnnotationMetadataSpec` covers the reproduction (an overriding method that declares nothing reports nothing, at the method and at the parameter), repeatable annotations surviving the narrowing, and the same end-to-end through a generated `BeanIntrospection` — the runtime path a validator takes. Reverting the two `inject` files makes the repeatable and the introspection tests fail.

## For the reviewer

Compatibility was checked on three axes:

- **API**: `japiCmp` against the real `5.1.12` baseline reports **0 errors** for `core`, `inject` and `core-processor`. Both visible additions are source-compatible warnings (`MethodElement.getDeclaredMethodAnnotationMetadata()` — "Method now provides default implementation"; `MutableAnnotationMetadata.getDeclaredMetadata()` — an override of an inherited method). A `javap` signature diff before/after shows 4 additions, 0 removals or changes. No `accepted-api-changes.json` entry is needed. japicmp does not catch one real source incompatibility, though: an existing implementor that declares a same-named method it cannot override breaks. `PythonMethodElement` had a private `getDeclaredMethodAnnotationMetadata()` and stopped compiling — fixed in the second commit by renaming that helper to `getOwnMethodAnnotationMetadata()`. Out-of-tree `MethodElement` implementations could hit the same thing.
- **Generated code**: `management`, `http-server-netty`, `router` and `http-client` compiled with the changed processor and with the baseline, `--rerun-tasks` both times — 671 class files, `diff -r` clean. Bean definitions, `$$AnnotationMetadata` classes, AOP proxies and introspections are unchanged.
- **Runtime**: both runtime callers of `getDeclaredMetadata()` in core — `QualifiedBeanType.getDeclaredQualifier()` and `AbstractInterceptorChain.resolveInterceptorValues()` — are guarded by `instanceof AnnotationMetadataHierarchy`, and `AnnotationMetadataHierarchy.getDeclaredMetadata()` is untouched. The compile time sites are likewise hierarchy-guarded, except three builder paths (`DeclaredBeanElementCreator` for `@Adapter`, `AbstractBeanDefinitionBuilder`, the Java/Groovy `BeanDefinitionBuilder`s) where the old view was `isEmpty() == true` with every value lookup blank — there the change replaces silently empty results with correct ones.

~10,700 tests green: inject 296, inject-java 4883, inject-java-test 323, inject-groovy 1053, inject-kotlin 790, inject-kotlin-test 200, context 48, aop 111, http-server-netty 1128, http-client 680, runtime/management/router/retry/jackson-databind/http-server/websocket 433, test-suite 226, test-suite-groovy 483.

Two things worth a second opinion:

1. **The `getAnnotationValuesByName` fix is a real behavior change**, not only a declared-view fix. Its fallback branch fires when the annotation is repeatable, the container is absent, and the annotation is present under its own name — the same branch from which `getAnnotationValuesByType` already returns the value, and which the old code already cached. Two runtime consumers may now see values they previously missed: `InterceptorBindingQualifier` and the GraalVM `ServiceLoaderFeature`. Existing tests assert non-empty results from this method and pass; none asserted the empty behavior. Happy to split it out if you'd rather ship the declared view alone.
2. `pushNewAnnotationMetadataOrReference` throws `IllegalStateException` for a non-empty metadata that isn't a `MutableAnnotationMetadata`. A plain `DefaultAnnotationMetadata` declared view is now non-empty, so *if* one reached that writer it would throw instead of silently emitting `EMPTY_METADATA`. Compile time metadata is always `MutableAnnotationMetadata` — which is why the generated output is byte-identical — so I left the writer alone rather than widening the diff; a `MutableAnnotationMetadata.of(...)` conversion there would close it.

Pre-existing and unchanged: `AbstractEnvironmentAnnotationMetadata` doesn't override `getDeclaredMetadata()`, so for bean definition metadata containing `${...}` expressions it returns the full metadata. Introspections aren't wrapped that way, so the validation use case above is unaffected.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/ast/MethodElement.java
+++ b/core-processor/src/main/java/io/micronaut/inject/ast/MethodElement.java
@@ -53,6 +53,7 @@ public interface MethodElement extends MemberElement {
      * The method will only return annotations defined on a method or inherited from the super methods,
      * while {@link #getAnnotationMetadata()} for a method combines the class and the method annotations.
      * NOTE: For a constructor {@link #getAnnotationMetadata()} will not combine the class annotations.
+     * See {@link #getDeclaredMethodAnnotationMetadata()} for the annotations of this declaration only.
      *
      * @return The method annotation metadata
      * @since 4.0.0
@@ -64,6 +65,22 @@ public interface MethodElement extends MemberElement {
                 return MethodElement.this.getAnnotationMetadata();
             }
         };
+    }
+
+    /**
+     * Returns the annotations of this method declaration only.
+     * Unlike {@link #getMethodAnnotationMetadata()} the annotations inherited from the super methods are
+     * excluded, and unlike {@link #getAnnotationMetadata()} the annotations of the class are excluded as well.
+     *
+     * @return The annotations declared by this method
+     * @since 5.2.0
+     */
+    default AnnotationMetadata getDeclaredMethodAnnotationMetadata() {
+        // the first call narrows a hierarchy of the class and the method metadata to the method metadata,
+        // the second one narrows the method metadata to what this declaration carries
+        return getMethodAnnotationMetadata().getAnnotationMetadata()
+            .getDeclaredMetadata()
+            .getDeclaredMetadata();
     }
 
     /**
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationMetadata.java
@@ -74,6 +74,12 @@ public interface AnnotationMetadata extends AnnotationSource {
 
     /**
      * Gets the declared metadata without inherited metdata.
+     *
+     * <p>A metadata that combines multiple elements, such as the class and the method annotations, is narrowed
+     * to the metadata of the element it is declared on; a subsequent call narrows that metadata to the
+     * annotations the element itself declares, excluding the ones inherited from the super elements.
+     * A metadata that is already declared only is returned unchanged.</p>
+     *
      * @return The declared metadata
      * @since 3.0.0
      */
--- a/inject-java/src/test/groovy/io/micronaut/inject/annotation/DeclaredMethodAnnotationMetadataSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/annotation/DeclaredMethodAnnotationMetadataSpec.groovy
@@ -1,0 +1,177 @@
+package io.micronaut.inject.annotation
+
+import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
+import io.micronaut.core.beans.BeanIntrospection
+import io.micronaut.inject.ast.ClassElement
+import io.micronaut.inject.ast.MethodElement
+import io.micronaut.inject.ast.ParameterElement
+
+class DeclaredMethodAnnotationMetadataSpec extends AbstractTypeElementSpec {
+
+    void "test the declared method annotations exclude the ones of the overridden method"() {
+        given:
+        ClassElement classElement = buildClassElement('''
+package test;
+
+import java.lang.annotation.*;
+
+class Test implements Contract {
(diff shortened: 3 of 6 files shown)


dstepanov:
Follow-up commit: `micronaut-inject-python` did not compile.

`PythonMethodElement` already declares a **private** `getDeclaredMethodAnnotationMetadata()` returning `ElementAnnotationMetadata`, which cannot implement the new default method of the same signature — so adding it to `MethodElement` is source-incompatible with any existing implementor that has a same-named method it cannot override. japicmp classifies the addition as source compatible and does not catch this; the in-repo implementor did.

Fixed by renaming the private helper to `getOwnMethodAnnotationMetadata()` — it returns the element's own method annotations before they are merged with the inherited ones, so the name now says what it does. `PythonMethodElement` inherits the default `getDeclaredMethodAnnotationMetadata()`, which resolves to the same declaration-only metadata through its merged hierarchy.

I kept the interface method name: it mirrors `getMethodAnnotationMetadata()`, and the collision was with a private helper rather than an incompatible public contract. Happy to rename it instead if you would rather not carry that risk for out-of-tree `MethodElement` implementations.

Verified with `./gradlew classes testClasses --continue` over the whole repo — all 809 tasks, zero compile errors. `:micronaut-inject-python-test:test` is SKIPPED locally (it needs the Python toolchain), so that suite is covered by CI rather than by me.

graemerocher (approved):

PR #12919, opened 2026-08-25
fix(validation): only advise methods that declare @RequiresValidation

dstepanov:
## What

When an introduction proxy is created, the methods to advise with the validation interceptor are selected in `DefaultElementBeanDefinitionBuilderFactory`:

```java
for (MethodElement methodElement : target.getEnclosedElements(
        ElementQuery.ALL_METHODS.annotated(am -> am.hasAnnotation(ANN_REQUIRES_VALIDATION)))) {
    methodElement.annotate(ANN_VALIDATED);
}
```

The predicate receives the method's `AnnotationMetadata`, which for a `MethodElement` **combines the class and the method** annotations. `ValidationVisitor` puts `@RequiresValidation` on the class as soon as *any* member of it needs validation, so `am.hasAnnotation(...)` was true for every method of the type and every method was annotated `@Validated`.

The fix narrows the predicate to `hasDeclaredAnnotation`, which is what `DeclaredBeanElementCreator#makeInterceptedForValidationIfNeeded` already uses for ordinary beans.

## Why

An introduction bean with one constrained method routed *all* of its methods through `ValidatingInterceptor`. Nothing was validated incorrectly — the interceptor finds nothing and returns — but every call paid for an interception it did not need. It also makes `@ValidateOnExecution(ExecutableType.NONE)` inexpressible: a method has to be *described* by `BeanDescriptor.getConstraintsForMethod` while its execution is not intercepted, which needs `@Executable` without `@RequiresValidation`. With the class-level annotation leaking through the query, the method was re-advised regardless.

Ordinary `@Singleton` beans were never affected — that path already used `hasDeclaredAnnotation`. The leak was confined to introduction proxies.

## Two gaps this uncovered

Removing the stray `@Validated` broke `MyRepoIntroductionSpec."test tx default repo methods"` and `MyAbstractRepoSpec."test default interceptor method"`. Both assert that a **type-level** `@Around` advice intercepts an interface **default** method — the behaviour added in 47455d9741 for micronaut-projects/micronaut-data#2943. They were passing only because the leaked `@Validated` gave those default methods a declared around advice. Two real gaps:

1. **`IntroductionInterfaceBeanElementCreator`** pre-filtered out every non-abstract method lacking *declared* around advice, stricter than `visitIntrospectedMethod`, which already honours around advice declared by the type. Default methods now survive the filter when the type declares around advice, matching how abstract classes behave.

2. **`DefaultRuntimeProxyDefinition#introduction`** resolved *introduction* interceptors for every executable method. `resolveIntroductionInterceptors` returns around + introduction with the introduction interceptor terminating the chain, so a concrete method's real implementation was never invoked (`findByIdWithDefault` returned `null`). Abstract methods now resolve introduction interceptors and concrete ones around interceptors, matching what the compile-time `AopProxyWriter` does.

`RuntimeProxyRepositoryTest#testMethodsCount` expected 1 executable method; with `@ByteBuddyStacktraceVerified` (type-level `@Around`) on `CustomCrudRepo`, the default method is now legitimately intercepted, so it is updated to 2 with the names asserted.

## Compatibility

No API break — `japiCmp` is green for `:micronaut-aop` and `:micronaut-core-processor`, and no signature changed.

Behaviour, against released 5.1.x:

- **Abstract methods of introduction proxies**: no functional change. They are proxied regardless and `ValidatingInterceptor` was a no-op for them. Visible only as metadata (`hasAnnotation("io.micronaut.validation.Validated")` is now false) and a shorter interceptor chain.
- **Default/concrete methods of such a type**: previously proxied only because of the stray annotation, now proxied only when the type or the method declares around advice. `getExecutableMethods()` can list fewer methods. Return values are unchanged — the old chain was `[ValidatingInterceptor] → proceed → real impl`.
- **Interface default methods with a type-level `@Around`** are now intercepted where before that happened only if the type incidentally had a constrained member. This *adds* interception and is the item most worth a release note.
- **Concrete methods on runtime introduction proxies** now run their real implementation instead of being terminated by the introduction interceptor. Every around interceptor is kept, in the same order.
- Constructor, field and injection-point validation are untouched — those go through `BeanDefinitionWriter#applyConfigurationInjectionIfNecessary`, keyed on the constructor's own declared `@RequiresValidation`.

## Tests

New `ValidationAdviceSpec` in `inject-java` (5 cases: concrete bean, introduction interface, introduction abstract class, method inherited from a super interface, overriding method) and in `inject-groovy` (3 cases). All fail on the unfixed compiler and pass with it.

Green locally: `:micronaut-inject-java:test`, `:micronaut-inject-groovy:test`, `:micronaut-inject-kotlin:test`, `:micronaut-aop:test`, `:micronaut-context:test`, `:micronaut-inject:test`, `:micronaut-core-processor:test`, `:test-suite:test`, `:test-suite-groovy:test`, `:test-suite-kotlin:test`, plus checkstyle.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/runtime/DefaultRuntimeProxyDefinition.java
+++ b/aop/src/main/java/io/micronaut/aop/runtime/DefaultRuntimeProxyDefinition.java
@@ -135,7 +135,11 @@ public record DefaultRuntimeProxyDefinition<T>(BeanDefinition<T> proxyBeanDefini
         ));
         List<InterceptedMethod<T>> interceptedMethods = new ArrayList<>(executableMethods.size());
         for (ExecutableMethod<T, ?> executableMethod : executableMethods) {
-            Interceptor<T, ?>[] methodInterceptors = InterceptorChain.resolveIntroductionInterceptors(interceptorRegistry, executableMethod, interceptors);
+            // Only the abstract methods are implemented by the introduction advice,
+            // the concrete ones are intercepted and proceed to the actual implementation
+            Interceptor<T, ?>[] methodInterceptors = executableMethod.isAbstract()
+                ? InterceptorChain.resolveIntroductionInterceptors(interceptorRegistry, executableMethod, interceptors)
+                : InterceptorChain.resolveAroundInterceptors(interceptorRegistry, executableMethod, interceptors);
             if (methodInterceptors.length > 0) {
                 interceptedMethods.add(new InterceptedMethod<>((ExecutableMethod) executableMethod, (Interceptor[]) methodInterceptors));
             }
--- a/core-processor/src/main/java/io/micronaut/inject/processing/IntroductionInterfaceBeanElementCreator.java
+++ b/core-processor/src/main/java/io/micronaut/inject/processing/IntroductionInterfaceBeanElementCreator.java
@@ -73,8 +73,16 @@ final class IntroductionInterfaceBeanElementCreator<R> extends AbstractBeanEleme
         return proxyBuilder.build();
     }
 
+    /**
+     * Checks if the method is an interface default method.
+     * Only invoked for a non-abstract method of an interface, where a static and a private method
+     * are the only alternatives to a default method; an interface method can never be final.
+     *
+     * @param method The non-abstract method of the interface
+     * @return true if the method is a default method
+     */
     private static boolean isDefaultMethod(MethodElement method) {
-        return !method.isAbstract() && !method.isStatic() && !method.isPrivate() && !method.isFinal();
+        return !method.isStatic() && !method.isPrivate();
     }
 
     private void handlePropertyMethod(ElementProxyBuilder<R> proxyBuilder, List<MethodElement> methods, @Nullable MethodElement method) {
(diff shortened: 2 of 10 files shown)


dstepanov:

**Downstream revisions**

Caveats:

graemerocher:
can you add a test for Python and Kotlin?

dstepanov:
@graemerocher added in b6b1764854.

**Kotlin** — `ValidationAdviceSpec` in `inject-kotlin`, 5 cases mirroring the `inject-java` spec: concrete bean (control), introduction interface, introduction abstract class, method inherited from a super interface, overriding method.

**Python** — `ValidationAdviceSpec` in `inject-python-test`, 2 cases against the `$RuntimeProxy$Definition` the Python processor emits: introduction ABC, and a constrained method inherited from a parent ABC.

The Python module needed `micronaut-validation-processor` on its test classpath for the tests to be meaningful. `PythonMethodElement` annotates a *method* with `@RequiresValidation` on its own, but nothing annotates the *class*, and `DefaultElementBeanDefinitionBuilderFactory` gates the whole validation block on the class-level annotation. Without `ValidationVisitor` running, the Python tests took the `DeclaredBeanElementCreator` path and passed identically on the fixed and the unfixed compiler — they never reached the changed line. I diffed a full `:micronaut-inject-python-test:test` run with and without the dependency: same result set, so it changes nothing else.

Both specs were checked against a temporarily reverted fix (`hasDeclaredAnnotation` → `hasAnnotation`): 4 of 5 fail in Kotlin (the concrete-bean case is the control — that path already used `hasDeclaredAnnotation`) and 2 of 2 in Python. Green with the fix, re-run after the rebase onto ed209b9236.

One case I deliberately left out: a Python method that *overrides* a constrained method. In Java and Kotlin the override inherits the parent's `@NotBlank` and stays validated; in Python it does not, because `PythonMethodElement.requiresValidation()` looks only at the method's own parameters. That looks like a real gap on the Python side, but it is unrelated to this PR, so I did not pin the behaviour down in a test here.

PR #12922, opened 2026-08-27
Reuse one interceptor instance across a bean's lifecycle

dstepanov:
## Summary

A non-singleton interceptor is already one instance per intercepted bean for constructor and method interception, but `@PostConstruct` and `@PreDestroy` each resolved a fresh one. An interceptor could therefore not release in `@PreDestroy` what it acquired in `@PostConstruct`.

For one advised singleton and one prototype interceptor bound to all kinds, `5.2.x` creates three instances:

```
1:AROUND_CONSTRUCT   1:AROUND   2:POST_CONSTRUCT   3:PRE_DESTROY
```

This makes it one, reused by every phase, and destroyed as a dependent of the target after the target's own `@PreDestroy`.

## What changed

- `InterceptedBeanDefinition` and `ParameterizedInterceptedBeanDefinition` resolve one registration set covering `AROUND_CONSTRUCT`, `POST_CONSTRUCT` and `PRE_DESTROY`, pass it to the constructor chain, and make it visible to post-construct interception of the same bean. A bean with constructor advice runs post-construct interception from *inside* the constructor chain, so the set is pushed before construction and popped after it, keyed by definition so a bean created while another is being constructed reads its own.
- A proxy whose target has lifecycle bindings retains its registrations and exposes them through `Intercepted`. That is the only route surviving from creation to destruction, since `dispose` runs with a fresh resolution context.
- `DefaultBeanContext` hands the registrations a bean owns to the dispose call, so a bean with lifecycle advice and **no** proxy can reuse them for `@PreDestroy`.

Each interception point still filters by its own binding and kind, so the wider binding cannot over-apply an interceptor.

## Also fixed

An interceptor bound to an `@Introduction` annotation was compiled into its own introduction proxy carrying the target's `POST_CONSTRUCT` binding. Creating it ran post-construct interception, which resolved interceptors by binding, found the proxy again and recursed until the stack was exhausted — such a bean could not be created at all on `5.2.x`. `isAopProxyType` already excluded `Interceptor` beans from around advice for this reason; the same guard now applies on the introduction path. This is independent of the rest and could be split out.

## Compatibility

- `initialize`/`dispose` keep their existing signatures and delegate to new overloads that take the registrations.
- `Intercepted.$interceptorRegistrations()` has an empty default, so proxies generated by earlier versions and non-proxy beans fall through to the previous resolve-by-binding path.
- The sharing lives in default methods of the definition interfaces, so definitions compiled by earlier versions pick it up **without recompiling**. Only the proxy-retained path needs a recompile.
- `japiCmp` green on `micronaut-aop`, `micronaut-inject` and `micronaut-core-processor`.

## Known limitation

A prototype created with `createBean` and destroyed with `destroyBean(Object)` has no tracked registration — `findBeanRegistration(Object)` searches only the singleton and custom scopes — so nothing links its destruction back to its creation and `@PreDestroy` resolves a new interceptor. It shares one instance for construction and post-construct. A test pins this. Prototypes injected into other beans are unaffected.

## Verification

| Module | Tests | Failures |
|---|---|---|
| micronaut-inject-java | 4903 | 0 |
| micronaut-inject-groovy | 1053 | 0 |
| micronaut-inject-kotlin | 790 | 0 |
| micronaut-inject | 296 | 0 |
| test-suite | 483 | 0 |

Plus `:micronaut-aop:check`, `:micronaut-core-processor:check` and `validateAssembleDocs`.

21 new specs cover reuse across every phase; ordering and kind selection with several interceptors; the target shapes that reach lifecycle interception by different routes (around, `proxyTarget = true`, introduction, factory-produced, `@AroundConstruct` with no proxy); binding isolation between beans; and that the registrations a proxy exposes are the same list instance it was constructed with, holding the very instances that performed the interception.

## Docs

New **Interceptor Life Cycle** page: the five kinds and which interceptor interface each selects, singleton versus per-target scope, the sequence one bean's interceptor sees, ordering, selection rules, and a table of behaviour per target shape.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/beandefinition/SharedInterceptorRegistrations.java
+++ b/aop/src/main/java/io/micronaut/aop/beandefinition/SharedInterceptorRegistrations.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.micronaut.aop.chain;
+package io.micronaut.aop.beandefinition;
 
 import io.micronaut.aop.Interceptor;
 import io.micronaut.core.annotation.Internal;
@@ -58,6 +58,9 @@ import java.util.List;
  * <p>Destruction happens later with a fresh resolution context, so nothing is shared with it here; see
  * {@code MethodInterceptorChain} for how pre-destroy reaches the interceptors a bean owns.</p>
  *
+ * <p>Everything here is package-private except {@link #store}, which the runtime proxy path in
+ * {@code io.micronaut.aop.runtime} calls from another package.</p>
+ *
  * @author Denis Stepanov
  * @since 5.2.0
  */
@@ -85,7 +88,7 @@ public final class SharedInterceptorRegistrations {
      * @since 5.2.0
      */
     @SuppressWarnings("unchecked")
-    public static void push(BeanResolutionContext resolutionContext,
+    static void push(BeanResolutionContext resolutionContext,
                             BeanDefinition<?> definition,
                             @Nullable List<?> registrations) {
         if (registrations == null || registrations.isEmpty()) {
@@ -112,7 +115,7 @@ public final class SharedInterceptorRegistrations {
      * @since 5.2.0
      */
     @SuppressWarnings("unchecked")
-    public static void pop(BeanResolutionContext resolutionContext,
+    static void pop(BeanResolutionContext resolutionContext,
                            BeanDefinition<?> definition,
                            @Nullable List<?> registrations) {
         if (registrations == null || registrations.isEmpty()) {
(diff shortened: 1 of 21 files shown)


aop/src/main/java/io/micronaut/aop/beandefinition/SharedInterceptorRegistrations.java
```diff
@@ -0,0 +1,153 @@
...
+ * bean created while another is being constructed reads its own.</p>
+ *
+ * <p>Destruction happens later with a fresh resolution context, so nothing is shared with it here; see
+ * {@code MethodInterceptorChain} for how pre-destroy reaches the interceptors a bean owns.</p>
+ *
+ * @author Denis Stepanov
+ * @since 5.2.0
+ */
+@Internal
+public final class SharedInterceptorRegistrations {
```
graemerocher, line 68: does this need to be public?

graemerocher (approved):

PR #12929, opened 2026-08-30
Leave @Primary out of composite qualifiers

dstepanov:
`BeanDefinitionWriter.getQualifier` writes `@Primary` as a literal null qualifier, since primary is the same as no qualifier. That is fine while `@Primary` is the only qualifier on the element, but when it is declared beside another qualifier the null lands in the array passed to `Qualifiers.byQualifiers(...)`, and `CompositeQualifier.filterQualified` dereferences it. The result is an NPE at resolution time — for example a `@Primary` `@Factory` bean that also carries a custom qualifier throws as soon as one of its produced beans is looked up, because resolving the produced bean resolves the factory by its (composite) qualifier.

This change filters `@Primary` out of the qualifier names before the simple/composite decision: `@Primary` breaks ties between candidates that already qualify, so it narrows nothing and takes no part in a composite. If nothing but `@Primary` was declared, no qualifier is written at all, matching the existing single-qualifier behaviour.

Includes a regression spec resolving a produced bean from a factory that is both `@Primary` and qualified.


graemerocher (approved):

dstepanov:
Superseded by #12945, which carries the identical commit from a branch in this repository.

Recreated in-repo because the required "SonarCloud Code Analysis" status is never posted for a pull request from a fork: the workflow gates the analysis step on `secrets.SONAR_TOKEN`, and GitHub withholds repository secrets from fork pull requests, so the check stays permanently in the expected state and blocks the merge.

PR #12930, opened 2026-08-30
Honour the argument of RuntimeBeanDefinition.Builder.singleton(boolean)

dstepanov:
`DefaultRuntimeBeanDefinition.Builder.singleton(boolean isSingleton)` assigned `this.singleton = true` unconditionally, ignoring its argument. As a result a runtime bean definition built with `singleton(false)` was still registered in the singleton scope: every `getBean` call returned the same instance, and the bean was never tracked as a dependent of the bean that resolved it (so it was not destroyed with it either).

This assigns the argument instead, and adds a regression spec verifying that `singleton(false)` yields a distinct instance per lookup while `singleton(true)` keeps returning the same one.


graemerocher (approved):

dstepanov:
Superseded by https://github.com/micronaut-projects/micronaut-core/pull/12946, which carries the identical commit from a branch in this repository (fork pull requests never receive the required "SonarCloud Code Analysis" status).

PR #12931, opened 2026-08-30
Remove a bean from the scope map when a custom scope destroys it

dstepanov:
`AbstractConcurrentCustomScope.remove(BeanIdentifier)` looked the bean up with `scopeMap.get(identifier)` and closed it, but never took the entry out of the scope map. The destroyed instance therefore stayed in the map and kept being served: any subsequent resolution through the scope — e.g. a client proxy dereferencing a scoped bean — returned the already destroyed instance forever, instead of triggering the creation of a fresh one.

This changes `remove` to take the entry out of the map (`scopeMap.remove(identifier)`) before closing it, matching what `destroyScope` does (close all + clear) and what `CustomScope` implementations are expected to do.

Includes a regression spec for a custom scope extending `AbstractConcurrentCustomScope`: after `remove(id)` the destroyed instance is evicted and the next resolution creates a new instance, and removing an unknown identifier leaves the scope untouched.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/scope/AbstractConcurrentCustomScope.java
+++ b/inject/src/main/java/io/micronaut/context/scope/AbstractConcurrentCustomScope.java
@@ -179,7 +179,7 @@ public abstract class AbstractConcurrentCustomScope<A extends Annotation> implem
             }
             if (CollectionUtils.isNotEmpty(scopeMap)) {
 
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

dstepanov:
Superseded by https://github.com/micronaut-projects/micronaut-core/pull/12947, which carries the identical commit from a branch in this repository (fork pull requests never receive the required "SonarCloud Code Analysis" status).

PR #12933, opened 2026-08-30
Resolve a null proxy target through the bean resolution customizer

dstepanov:
`DefaultBeanContext.getProxyTargetBean` returned `registration.bean` directly. A `@Nullable` factory method is allowed to produce `null`, but when the produced bean sits behind a lazy proxy, the proxy was handed that `null` as its target and the first intercepted call failed with a `NullPointerException`.

The ordinary bean lookup path already routes a null bean through `resolveNullBeanRegistration`, where a `BeanResolutionCustomizer` can substitute a replacement value via `resolveNullBean`. The proxy target path now does the same: the `Argument`-based overload delegates to the `BeanDefinition`-based one, which consults the customizer when the resolved bean is `null`, so both the generated proxy code paths and direct `getProxyTargetBean` calls behave consistently with regular lookups.

Adds a regression spec: a lazy `proxyTarget` bean produced as `null` by a `@Nullable` factory resolves the customizer's substitute on the intercepted call instead of throwing an NPE. The spec fails with an NPE without the fix.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/aop/lazyproxytarget/ProxyTargetNullBeanSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/aop/lazyproxytarget/ProxyTargetNullBeanSpec.groovy
@@ -1,0 +1,104 @@
+package io.micronaut.aop.lazyproxytarget
+
+import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
+import io.micronaut.context.ApplicationContextBuilder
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

dstepanov:
Superseded by https://github.com/micronaut-projects/micronaut-core/pull/12951, which carries the identical commit from a branch in this repository (fork pull requests never receive the required "SonarCloud Code Analysis" status).

PR #12935, opened 2026-08-30
Process a class annotated only @ClassImport

dstepanov:
`BeanDefinitionInjectProcessor` keeps an annotation from the round when it has one of `ANNOTATION_STEREOTYPES` as a stereotype or `isProcessedAnnotation` matches its name. `ClassImport` is listed among the stereotypes, but an annotation type never carries itself as a stereotype, so that entry can never admit it: the import trigger survives today only because the `isProcessedAnnotation` fallback's `io.micronaut.*` pattern happens to cover it. This admits `ClassImport` by name instead, since it is processed for the classes it names rather than for a stereotype it carries.

Two specs lock the trigger: a class annotated only `@ClassImport` produces bean definitions for the imported classes, both when compiled from source (`ClassImportOnlySpec`) and when a visitor generates the importer into a later compilation round (`ClassImportGeneratedRoundSpec`).

A related trap: a generated importer must not be annotated `io.micronaut.core.annotation.Generated` — `ModelUtils.resolveTypeElements` skips elements carrying it and the imports are silently dropped (verified: the generated-round spec fails when the importer carries it). Noted in the `ClassImport` javadoc.


graemerocher (approved):

dstepanov:
Superseded by https://github.com/micronaut-projects/micronaut-core/pull/12952, which carries the identical commit from a branch in this repository (fork pull requests never receive the required "SonarCloud Code Analysis" status).

PR #12936, opened 2026-08-30
Let the customizer allow a bean to instantiate to null

dstepanov:
Continues the `BeanResolutionCustomizer` integration-hook series (#12678) by completing the customizer's say over null instantiation.

A customizer can already substitute a value for a bean that produced `null` via `resolveNullBean`, but `DefaultBeanContext.resolveByBeanFactory` threw `InstantiatableBeanDefinition [...] returned null` before that hook was ever consulted, unless the definition itself was declared `@Nullable`. A factory/provider may legitimately produce `null`, with the integration deciding what the null becomes — a default value, a substitute, or a meaningful exception.

This adds `BeanResolutionCustomizer.shouldAllowNullBean(BeanResolutionContext, BeanDefinition)` (default `false`) and consults it in the null check of `resolveByBeanFactory`, letting the null flow through to the existing `resolveNullBeanRegistration` path.

Tests cover both directions: a customizer allowing null for a factory returning null gets its `resolveNullBean` substitute, and without the customizer's permission the existing instantiation error still throws.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/dependent/BeanResolutionCustomizerSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/dependent/BeanResolutionCustomizerSpec.groovy
@@ -4,6 +4,7 @@ import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
 import io.micronaut.context.ApplicationContextBuilder
 import io.micronaut.context.BeanResolutionContext
 import io.micronaut.context.BeanResolutionCustomizer
+import io.micronaut.context.exceptions.BeanInstantiationException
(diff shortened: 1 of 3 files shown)


dstepanov:
Closing: this hook duplicates an extension point the definition metadata already provides. A compile-time processor (or a runtime registration via RuntimeBeanDefinition's annotation metadata) can put the Nullable stereotype on the definition, and `isNullableBeanDefinition` then lets the null through to `resolveNullBeanRegistration`/`resolveNullBean` — the same flow this hook gated, with no new API. Verified by switching the integration that motivated this to the annotation and re-running its full test suite. The related proxy-path fix in #12933 is independent and still applies.

PR #12938, opened 2026-08-30
Fire destruction listeners when closing a context-created bean registration

dstepanov:
## Problem

`BeanRegistration.of(BeanContext, BeanIdentifier, BeanDefinition, bean, dependents)` only returns a `BeanDisposingRegistration` — whose `close()` destroys the bean through `BeanContext.destroyBean(registration)` — when the definition is a `DisposableBeanDefinition`, the bean is a `LifeCycle`, or there are dependents. For any other bean it returns a plain `BeanRegistration` whose `close()` is a no-op.

Closing such a registration therefore silently skips `BeanPreDestroyEventListener` / `BeanDestroyedEventListener` (and the cache purging `destroyBean` performs). The main real-world path is custom scopes, which destroy their beans via `CreatedBean.close()` (`AbstractConcurrentCustomScope.destroyScope()`, `DefaultCustomScopeRegistry`, `RefreshScope`): a scoped bean with no `@PreDestroy` of its own never reaches the destruction listeners when the scope is destroyed.

A bean with no destruction logic of its own can still matter to a `BeanPreDestroyEventListener`/`BeanDestroyedEventListener` — the listener contract should not depend on whether the bean happens to implement its own dispose logic.

## Change

When a bean context is provided, `BeanRegistration.of(...)` now always returns the disposing registration, so `close()` always destroys through the context and the destruction listeners fire consistently.

## Call-site audit / behavior notes

- `DefaultBeanContext.destroyBean(registration)` handles `BeanDisposingRegistration` via an `instanceof` branch and never calls `close()` on it, so there is no recursion; the context shutdown path (`stop()`) also goes through `destroyBean` directly and is unchanged.
- Destroying a bean through `destroyBean(...)` and then also closing its registration (e.g. a scope tearing down later) runs `destroyBean` twice and fires the listeners twice. This is not new: registrations of beans with dispose logic (and any registration with dependents) already behaved this way — there was never an idempotency guard on `close()`.
- `BeanRegistration.equals()` compares `getClass()`: a manually constructed `new BeanRegistration(...)` no longer equals an `of()`-created registration for a bean without dispose logic (it previously could). Equality between `of()`-created registrations is unchanged. No code in the repository relies on the old mixed-class equality.
- Registrations for such beans now carry two extra reference fields (context + dependents); footprint impact is negligible.
- Binary compatibility is unaffected: no signatures changed and `BeanDisposingRegistration` is package-private/`@Internal`.

## Tests

- Closing a registration of a bean with no `@PreDestroy`/`LifeCycle` fires `BeanPreDestroyEventListener` and `BeanDestroyedEventListener`.
- Closing a registration destroys dependent beans and fires their listeners too.
- A `BeanPreDestroyEventListener` can replace the bean instance per its contract; the `BeanDestroyedEvent` sees the replacement.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/lifecycle/registrationclose/DependentBean.java
+++ b/inject-java/src/test/groovy/io/micronaut/inject/lifecycle/registrationclose/DependentBean.java
@@ -1,0 +1,7 @@
+package io.micronaut.inject.lifecycle.registrationclose;
+
+import io.micronaut.context.annotation.Prototype;
+
+@Prototype
+public class DependentBean {
+}
--- a/inject-java/src/test/groovy/io/micronaut/inject/lifecycle/registrationclose/DependentBeanDestroyedListener.java
+++ b/inject-java/src/test/groovy/io/micronaut/inject/lifecycle/registrationclose/DependentBeanDestroyedListener.java
@@ -1,0 +1,19 @@
+package io.micronaut.inject.lifecycle.registrationclose;
+
+import io.micronaut.context.event.BeanDestroyedEvent;
+import io.micronaut.context.event.BeanDestroyedEventListener;
+import jakarta.inject.Singleton;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Singleton
+public class DependentBeanDestroyedListener implements BeanDestroyedEventListener<DependentBean> {
+
+    final List<DependentBean> destroyed = new ArrayList<>();
+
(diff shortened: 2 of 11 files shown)


graemerocher (approved):

dstepanov:
Superseded by #12943 (same change, opened from a branch on micronaut-projects/micronaut-core instead of a fork; also addresses container-element registration handling).

PR #12939, opened 2026-08-30
Fix Kotlin 2.4 compiler test compatibility

skrcode:
Completes #12815 by fixing the Kotlin compiler-test compatibility edges exposed by Kotlin 2.4.10.

## What changed

- update the Kotlin compile-testing fork to 0.13.0, whose release aligns with Kotlin 2.4 and KSP 2.3.9
- give each test compilation its own Kotlin/KSP workspace so a failed compilation cannot leak state into later tests
- preserve Micronaut's Kotlin 2.0 test semantics explicitly while suppressing the compiler's version warning and retaining the legacy annotation-target behavior
- exclude Kotlin's new synthetic covariant bridge from the reflection assertion that counts logical repository methods

## Why

The Renovate branch fails before its compiler tests can run because compile-testing 0.12.1 passes a null `optIn` value into Kotlin 2.4's non-null compiler API. After aligning compile-testing, reused compiler state and Kotlin's new synthetic bridge were the remaining compatibility edges.

## Validation

- `JAVA_HOME=<JDK 25> ./gradlew --no-daemon :micronaut-inject-kotlin:test` — 790 tests passed, 11 skipped
- `JAVA_HOME=<JDK 25> ./gradlew --no-daemon :micronaut-inject-kotlin-test:check :micronaut-inject-kotlin:check` — passed; 141 tasks including Spotless, Checkstyle, POM validation, tests, and binary compatibility
- `git diff --check` — passed
- repository-wide `./gradlew --no-daemon check` progressed through 285 tasks and stopped only at `:micronaut-inject-python:compileVfsPythonBytecode`; that task requires the documented GraalVM JDK 25, while the local JDK is Temurin 25

JAIPilot Remote was also attempted, but its Gradle plugin mirror could not resolve `io.micronaut.build.shared.settings:8.0.1`; no remote result is claimed as test evidence.

Built and verified with [JAIPilot](https://github.com/JAIPilot/jaipilot):
- Skills: `jaipilot-maintainer-intent`, `jaipilot-fast-execution`, and `jaipilot-review-diff`
- Execution profile: `gpt-5.6-sol · xhigh · fast`


graemerocher (approved):

graemerocher:
Thanks for the contribution!

PR #12942, opened 2026-08-31
Fix Python pooled request-body context conversion

graemerocher:
## Summary

- Add a permanent end-to-end JMH benchmark for pooled Python JSON request bodies and a no-body control endpoint.
- Convert reconstructible `@Introspected` Python wrappers in the controller’s borrowed GraalPy context, including recursive nested, collection, map, array, enum, mutable, and frozen properties.
- Integrate the PythonPool concurrency lifecycle changes from #12908, with deterministic pool and context-release regressions.
- Preserve ordinary host-bridge and non-pooled compatibility; request-wide context leasing remains out of scope.

## Evidence

The pre-fix and fixed revisions were benchmarked consecutively on the same host with the same benchmark source and JMH settings: 8 workers, pool size 8, 5 × 2 s warmup, and 5 × 2 s measurement.

| Endpoint | Pre-fix | Fixed | Change |
| --- | ---: | ---: | ---: |
| JSON request body | 1,124 ops/s | 33,987 ops/s | ~30.2× |
| No-body control | 38,214 ops/s | 37,842 ops/s | −1.0% |

The pre-fix body path stabilized around 1.1k ops/s while the fixed path stabilized around 34k ops/s. The unchanged control result confirms that the context-aware conversion removes primary-context serialization without imposing a request-wide pooled-controller tax. The matching reports are `python-request-body-baseline-review3.txt` and `python-request-body-fixed-review3-clean.txt` under `benchmarks/build/reports/jmh/`; async-profiler outputs remain under the same report directory.

The complete 1/2/4/8 worker × 1/2/4/8 pool-size sweep also compiles and completes successfully. Its shorter per-trial results are retained as regression data, while the table above uses the longer like-for-like run requested during review.

## Verification

- Full affected checks pass for `micronaut-context-python`, `micronaut-inject-python`, `micronaut-inject-python-test`, and `test-suite-python` with `-Ppython-ci` (122 + 104 + 583 + 196 tests).
- Checkstyle and Spotless pass as part of those module checks.
- The end-to-end JMH benchmark compiles and both the full scaling sweep and longer matched comparison complete successfully.

### DIFF
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     id("io.micronaut.build.internal.convention-base")
+    id("io.micronaut.build.internal.python")
     id("me.champeau.jmh") version "0.7.3"
 }
 
@@ -48,6 +49,8 @@ val jmhWarmupIterations = providers.gradleProperty("jmh.warmupIterations").map(S
 val jmhProfilers = providers.gradleProperty("jmh.profilers")
     .map { it.split(",").map(String::trim).filter(String::isNotEmpty) }
     .getOrElse(emptyList())
+val jmhPoolSizes = providers.gradleProperty("jmh.poolSizes")
+    .map { it.split(",").map(String::trim).filter(String::isNotEmpty) }
 val jmhHumanOutput = providers.gradleProperty("jmh.humanOutput")
     .map(layout.projectDirectory::file)
 
@@ -57,6 +60,11 @@ jmh {
     iterations = jmhIterations
     warmupIterations = jmhWarmupIterations
     profilers = jmhProfilers
+    providers.gradleProperty("jmh.warmupTime").orNull?.let(warmup::set)
+    providers.gradleProperty("jmh.timeOnIteration").orNull?.let(timeOnIteration::set)
+    jmhPoolSizes.orNull?.let { sizes ->
+        benchmarkParameters.put("poolSize", objects.listProperty(String::class.java).value(sizes))
+    }
     humanOutputFile.set(jmhHumanOutput)
     duplicateClassesStrategy = DuplicatesStrategy.WARN
 }
@@ -69,7 +77,6 @@ tasks {
     named<Jar>("jmhJar") {
         isZip64 = true
         manifest.attributes["Multi-Release"] = "true"
-        exclude("GRAALPY-VFS/**")
     }
 }
 
--- a/benchmarks/src/jmh/java/io/micronaut/python/benchmark/PythonRequestBodyPoolBenchmark.java
+++ b/benchmarks/src/jmh/java/io/micronaut/python/benchmark/PythonRequestBodyPoolBenchmark.java
@@ -55,7 +55,8 @@ import java.util.concurrent.TimeUnit;
  *
  * <p>Run a smoke benchmark with
  * {@code ./gradlew :benchmarks:jmh -Pjmh.includes=io.micronaut.python.benchmark.PythonRequestBodyPoolBenchmark
- * -Pjmh.warmupIterations=1 -Pjmh.iterations=1}.</p>
+ * -Pjmh.warmupIterations=1 -Pjmh.iterations=1}. Run the full pool-size sweep by adding
+ * {@code -Pjmh.poolSizes=1,2,4,8}.</p>
  *
  * @since 5.2.0
  */
@@ -70,7 +71,7 @@ public class PythonRequestBodyPoolBenchmark {
         """;
 
     /** Configured Python context pool size. */
-    @Param({"1", "2", "4", "8"})
+    @Param({"4"})
     public int poolSize;
 
     private ApplicationContext applicationContext;
--- a/benchmarks/src/jmh/java/io/micronaut/python/benchmark/PythonRequestBodyPoolBenchmark.java
+++ b/benchmarks/src/jmh/java/io/micronaut/python/benchmark/PythonRequestBodyPoolBenchmark.java
@@ -1,0 +1,232 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.python.benchmark;
+
+import io.micronaut.context.ApplicationContext;
+import io.micronaut.context.python.PythonContextExecutor;
+import io.micronaut.runtime.server.EmbeddedServer;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * End-to-end reproduction for pooled Python controller request-body conversion contention.
+ *
+ * <p>Each invocation sends a real HTTP request through Netty JSON decoding, body binding,
+ * generated Python wrapper conversion, and pooled controller execution. The control endpoint
+ * isolates the general cost of the same pooled controller without a data-class body.</p>
+ *
+ * <p>Run a smoke benchmark with
+ * {@code ./gradlew :benchmarks:jmh -Pjmh.includes=io.micronaut.python.benchmark.PythonRequestBodyPoolBenchmark
+ * -Pjmh.warmupIterations=1 -Pjmh.iterations=1}.</p>
+ *
+ * @since 5.2.0
+ */
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class PythonRequestBodyPoolBenchmark {
+
+    private static final String EXPECTED_BODY = "Ada:compiler:3:True:benchmark";
+    private static final String JSON = """
+        {"customer":"Ada","product":"compiler","quantity":3,"priority":true,"note":"benchmark"}
+        """;
+
+    /** Configured Python context pool size. */
+    @Param({"1", "2", "4", "8"})
+    public int poolSize;
+
+    private ApplicationContext applicationContext;
+    private HttpClient client;
+    private HttpRequest bodyRequest;
+    private HttpRequest controlRequest;
+
+    /** Starts the server and verifies both benchmark paths. */
+    @Setup(Level.Trial)
+    public void setUp() throws IOException, InterruptedException {
+        try {
+            applicationContext = ApplicationContext.builder()
+                .properties(Map.of(
+                    "micronaut.server.port", -1,
+                    "micronaut.python.pool.enabled", true,
+                    "micronaut.python.pool.size", poolSize,
+                    "micronaut.python.pool.warn-wait", "60s"
+                ))
+                .start();
+            warmPool();
+            EmbeddedServer server = applicationContext.getBean(EmbeddedServer.class).start();
+            URI root = server.getURI().resolve("/python-request-body-benchmark/");
+            client = HttpClient.newBuilder()
+                .connectTimeout(Duration.ofSeconds(30))
+                .build();
+            bodyRequest = HttpRequest.newBuilder(root.resolve("body"))
+                .header("Content-Type", "application/json")
+                .POST(HttpRequest.BodyPublishers.ofString(JSON))
+                .build();
+            controlRequest = HttpRequest.newBuilder(root.resolve("control"))
+                .POST(HttpRequest.BodyPublishers.noBody())
+                .build();
+
+            verify(exchange(bodyRequest), EXPECTED_BODY);
+            verify(exchange(controlRequest), "control");
+        } catch (IOException | InterruptedException | RuntimeException e) {
+            tearDown();
+            throw e;
+        }
+    }
+
+    /** Stops the benchmark application. */
+    @TearDown(Level.Trial)
+    public void tearDown() {
+        if (applicationContext != null) {
+            applicationContext.close();
+        }
+    }
+
+    /** Measures request-body conversion with one JMH worker. */
+    @Benchmark
+    @Threads(1)
+    public void bodyThreads1(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(bodyRequest));
+    }
+
+    /** Measures request-body conversion with two JMH workers. */
+    @Benchmark
+    @Threads(2)
+    public void bodyThreads2(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(bodyRequest));
+    }
+
+    /** Measures request-body conversion with four JMH workers. */
+    @Benchmark
+    @Threads(4)
+    public void bodyThreads4(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(bodyRequest));
+    }
+
+    /** Measures request-body conversion with eight JMH workers. */
+    @Benchmark
+    @Threads(8)
+    public void bodyThreads8(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(bodyRequest));
+    }
+
+    /** Measures the pooled control endpoint with one JMH worker. */
+    @Benchmark
+    @Threads(1)
+    public void controlThreads1(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(controlRequest));
+    }
+
+    /** Measures the pooled control endpoint with two JMH workers. */
+    @Benchmark
+    @Threads(2)
+    public void controlThreads2(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(controlRequest));
+    }
+
+    /** Measures the pooled control endpoint with four JMH workers. */
+    @Benchmark
+    @Threads(4)
+    public void controlThreads4(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(controlRequest));
+    }
+
+    /** Measures the pooled control endpoint with eight JMH workers. */
+    @Benchmark
+    @Threads(8)
+    public void controlThreads8(Blackhole blackhole) throws IOException, InterruptedException {
+        blackhole.consume(exchange(controlRequest));
+    }
+
+    private HttpResponse<String> exchange(HttpRequest request) throws IOException, InterruptedException {
+        return client.send(request, HttpResponse.BodyHandlers.ofString());
+    }
+
+    private void warmPool() throws InterruptedException {
+        PythonContextExecutor contextExecutor = applicationContext.getBean(PythonContextExecutor.class);
+        ExecutorService executor = Executors.newFixedThreadPool(poolSize);
+        CountDownLatch acquired = new CountDownLatch(poolSize);
+        CountDownLatch release = new CountDownLatch(1);
+        List<Future<?>> futures = new ArrayList<>(poolSize);
+        try {
+            for (int i = 0; i < poolSize; i++) {
+                futures.add(executor.submit(() -> contextExecutor.withContext(context -> {
+                    acquired.countDown();
+                    await(release);
+                    return null;
+                })));
+            }
+            if (!acquired.await(2, TimeUnit.MINUTES)) {
+                throw new IllegalStateException("Timed out warming the Python context pool");
+            }
+            release.countDown();
+            for (Future<?> future : futures) {
+                future.get(2, TimeUnit.MINUTES);
+            }
+        } catch (java.util.concurrent.ExecutionException | java.util.concurrent.TimeoutException e) {
+            throw new IllegalStateException("Failed to warm the Python context pool", e);
+        } finally {
+            release.countDown();
+            executor.shutdownNow();
+        }
+    }
+
+    private static void await(CountDownLatch latch) {
+        boolean interrupted = false;
+        while (true) {
+            try {
+                latch.await();
+                break;
+            } catch (InterruptedException e) {
+                interrupted = true;
+            }
+        }
+        if (interrupted) {
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private static void verify(HttpResponse<String> response, String expected) {
+        if (response.statusCode() != 200 || !expected.equals(response.body())) {
+            throw new IllegalStateException("Unexpected benchmark response: " + response.statusCode() + " " + response.body());
+        }
+    }
+}
--- a/context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
+++ b/context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -35,6 +36,7 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
+import java.util.function.Supplier;
 
 import io.micronaut.core.async.publisher.Publishers;
 import io.micronaut.core.annotation.Internal;
@@ -72,6 +74,7 @@ public final class GraalPyRuntimeUtil {
     private static final String TO_PYTHON_STANDARD_TYPE = "__micronaut_to_python_standard_type";
     private static final String UTC_OFFSET = "__micronaut_utc_offset";
     private static final AsyncMemberAdapter ASYNC_MEMBER_ADAPTER = new AsyncMemberAdapter();
+    private static final ScopedValue<ContextConversion> CURRENT_CONTEXT_CONVERSION = ScopedValue.newInstance();
     private static final Source PUT_MEMBER_SOURCE = Source.newBuilder(PYTHON, """
         def __micronaut_put_member(target, name, value):
             setattr(target, name, value)
@@ -250,10 +253,7 @@ public final class GraalPyRuntimeUtil {
     @UsedByGeneratedCode
     public static @Nullable Object coerceValue(@Nullable Object value) {
         return switch (value) {
-            case PooledValueCoercible pooledValueCoercible when value instanceof Enum<?> ->
-                pooledValueCoercible.asPolyglotValue();
-            case ValueCoercible valueCoercible when !(value instanceof PooledValueCoercible) ->
-                valueCoercible.asPolyglotValue();
+            case ValueCoercible valueCoercible -> valueCoercible.asPolyglotValue();
             case null, default -> value;
         };
     }
@@ -266,6 +266,10 @@ public final class GraalPyRuntimeUtil {
      * @return The coerced value
      */
     public static @Nullable Object coerceToContext(@Nullable Object value, Context context) {
+        return withContextConversion(context, () -> coerceToContext0(value, context));
+    }
+
+    private static @Nullable Object coerceToContext0(@Nullable Object value, Context context) {
         Object standardType = coerceStandardTypeToContext(value, context);
         if (standardType != value) {
             return standardType;
@@ -283,7 +287,22 @@ public final class GraalPyRuntimeUtil {
                 return null;
             }
             case PooledValueCoercible pooledValueCoercible -> {
-                return pooledValueCoercible.asPolyglotValue(context);
+                return coercePooledValue(pooledValueCoercible, context);
+            }
+            case ValueCoercible valueCoercible -> {
+                Value polyglotValue = valueCoercible.asPolyglotValue();
+                if (polyglotValue == null || isValueInContext(polyglotValue, context)) {
+                    return polyglotValue;
+                }
+                throw new IllegalArgumentException(
+                    "Python wrapper " + value.getClass().getName() + " cannot be reconstructed in the target context"
+                );
+            }
+            case Value polyglotValue -> {
+                if (isValueInContext(polyglotValue, context)) {
+                    return polyglotValue;
+                }
+                throw new IllegalArgumentException("Cannot pass a polyglot Value to a different context");
             }
             case List<?> list -> {
                 List<@Nullable Object> result = new ArrayList<>(list.size());
@@ -327,6 +346,10 @@ public final class GraalPyRuntimeUtil {
      * @return The coerced value
      */
     public static @Nullable Object coerceToContext(@Nullable Object value, Context context, Class<?> declaredType) {
+        return withContextConversion(context, () -> coerceToContext0(value, context, declaredType));
+    }
+
+    private static @Nullable Object coerceToContext0(@Nullable Object value, Context context, Class<?> declaredType) {
         if (value == null) {
             return null;
         }
@@ -342,7 +365,8 @@ public final class GraalPyRuntimeUtil {
         }
         return switch (value) {
             case PooledValueCoercible pooledValueCoercible ->
-                pooledValueCoercible.asPolyglotValue(context);
+                coercePooledValue(pooledValueCoercible, context);
+            case ValueCoercible _, Value _ -> coerceToContext0(value, context);
             case List<?> _ when List.class.equals(declaredType) ->
                 coerceToContext(value, context);
             case Map<?, ?> _ when Map.class.equals(declaredType) ->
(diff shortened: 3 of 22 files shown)


context-python/src/main/java/io/micronaut/context/python/PythonPool.java
```diff
@@ -423,26 +416,77 @@ private Context getOrCreateEventLoopContext(PythonEventLoop eventLoop) {
...
+                    }
+                    if (creatingContext == null && (!pooled || size.get() < targetSize)) {
+                        // we can create a context.
+                        creatingContext = Thread.currentThread();
+                        break;
+                    }
+                    try {
+                        // another thread is creating a context, wait for it or wait for the queue to get a new item.
+                        if (warnThreshold == null || !warnThreshold.isPositive() || !LOG.isWarnEnabled()) {
+                            wait();
```
dstepanov: This wait loop never checks `closed`, so a waiter can block here forever after the pool is closed — and it can wedge graceful shutdown.

`closePool()` and `shutdownGracefully()` both set `closed = true` and `notifyAll()`, but:

- `closePool()` clears `pooledContexts` and never resets `size`, so `size.get() < targetSize` stays false;
- `release()` returns early when closed without re-queueing, so the queue stays empty.

A woken waiter therefore polls an empty queue, fails the create condition, and re-enters `wait()` permanently.

The shutdown path makes this concrete: `shutdownGracefully()` flips `closed` immediately and then waits on `onNoActiveExecutions(...)`. Any in-flight request that reaches `borrow()` after that point with an empty queue blocks forever, its execution never finishes, and `gracefulShutdown` never completes.

This is compounded by the interrupt handling below — `InterruptedException` is swallowed (`interrupted = true`, loop continues) and only re-asserted after the loop exits, so the thread is now uninterruptible. The previous `borrow()` threw `IllegalStateException` on interrupt, which at least let the thread die.

Suggested fix — bail out at the top of the loop:

```java
while (true) {
    if (closed) {
        throw new IllegalStateException("Pool closed");
    }
    Context polled = pooledQueue.poll();
    ...
```

(`createBorrowedPooledContext` already throws `IllegalStateException("Pool closed")` for the create path, so this just makes the wait path consistent with it.)

---

context-python/src/main/java/io/micronaut/context/python/PythonPool.java
```diff
@@ -423,26 +416,77 @@ private Context getOrCreateEventLoopContext(PythonEventLoop eventLoop) {
         if (existing != null) {
             return existing;
         }
-        synchronized (eventLoopContexts) {
-            existing = eventLoopContexts.get(eventLoop);
-            if (existing != null) {
-                return existing;
+        return eventLoopContexts.computeIfAbsent(eventLoop, _ -> borrow0(false));
```
dstepanov: `borrow0(false)` can `wait()` on the pool monitor *and* build a full GraalPy context, and here all of that runs inside `ConcurrentHashMap.computeIfAbsent`, holding a bin lock the whole time.

`ConcurrentHashMap` documents that the mapping function must be short and must not attempt to update any other mapping of the map. Two consequences:

- unrelated `PythonEventLoop` keys that hash to the same bin serialize behind a context build;
- any `GraalPyContextCustomizer` that reached back into the pool from `createContext()` would deadlock (a blocked thread holds the bin lock while waiting on the `PythonPool` monitor).

The previous `synchronized (eventLoopContexts)` version serialized too, but it didn't nest a monitor wait inside a map lock. Computing outside the map and using `putIfAbsent` (closing the loser) — or keeping an explicit lock — would avoid the shape entirely.

---

context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
```diff
@@ -249,6 +250,8 @@ public static boolean isNone(@Nullable Value value) {
     @UsedByGeneratedCode
     public static @Nullable Object coerceValue(@Nullable Object value) {
         return switch (value) {
+            case PooledValueCoercible pooledValueCoercible when value instanceof Enum<?> ->
+                pooledValueCoercible.asPolyglotValue();
```
dstepanov: This branch produces a `Value` bound to the **primary** context: the no-arg `asPolyglotValue()` on a generated enum calls `PythonContextRuntime.enumValue(ref, name)`, which resolves via `getContext()`.

Nothing downstream re-homes it. `coerceStandardTypeToContext` has no `Value` case, so `putMember` → `coerceToContext(v, targetCtx)` falls through to `default -> value` and hands a foreign-context `Value` straight to the target context. That's reachable whenever `graalpyInternalValue` came from `fromPolyglotValue(...)` in a borrowed pooled context and a setter or the sync path then writes an enum property — i.e. exactly the cross-context failure this PR is fixing.

If the branch were dropped, enums would fall through to `default -> value` and the existing `case PooledValueCoercible -> asPolyglotValue(context)` in `coerceToContext` would resolve them in the correct context — which is what the new `enumValue(Context, ...)` overload is for.

I assume the shim exists so call sites that *don't* re-coerce (the generated setters' direct `Value.putMember`, `invokePythonMethod`) don't receive a raw host enum. If so, it's trading a raw host object for a wrong-context `Value`, and a comment saying which of the two is the lesser evil here would help the next reader.

Minor: the guard tests `value` rather than the pattern variable — `when pooledValueCoercible instanceof Enum<?>` reads better.

---

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
```diff
@@ -602,6 +611,66 @@ && requiresValidationIntrospection(element, beanProperties)) {
                                     return reconstructedValue.returning();
                                 })
                             ));
+                        if (isReconstructibleBean) {
```
dstepanov, line 619: The generated `asPolyglotValue(Context)` never caches its result and has no re-entrancy guard, unlike the no-arg version right above it. From the expected source in `GenerateToStringEqualsSpec`:

```java
public Value asPolyglotValue(Context arg1) {
    if (GraalPyRuntimeUtil.isValueInContext(this.graalpyInternalValue, arg1)) {
        return this.asPolyglotValue();
    } else {
        return PythonContextRuntime.newUninitializedInstance(arg1, ..., AnnotationUtil.mapOf(...));
    }
}
```

The no-arg path assigns `graalpyInternalValue` *before* populating members (plus the `graalpyInternalValueSyncing` flag) precisely so re-entrancy terminates. This path builds the whole property map first and stores nothing, so:

1. every call reconstructs a fresh Python object. Passing the same wrapper as two arguments gives Python two distinct objects (`is` comparisons, mutation visibility), and a sub-object shared by two properties is duplicated per reference;
2. a cyclic Python object graph (built on the Python side and passed back, e.g. `a.next = b; b.prev = a`) recurses until `StackOverflowError`, because each `coerceToContext(child, ctx, Child.class)` re-enters `child.asPolyglotValue(ctx)` with nothing memoized.

JSON bodies can't be cyclic, so the new specs won't catch either. A test with a diamond — two properties referencing the same nested instance — would be cheap and would pin down the intended identity semantics. If per-call reconstruction is deliberate (immutable request bodies), a note on the method saying so would help.

---

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
```diff
@@ -285,22 +287,28 @@ public void visitClass(ClassElement element, VisitorContext context) {
...
                     final boolean isIntroductionBean = element.hasStereotype(Introduction.class);
                     boolean isJunit5Test = element.getEnclosedElement(ElementQuery.ALL_METHODS.onlyInstance().annotated(ann -> ann.hasDeclaredAnnotation(JUNIT_TEST))).isPresent();
                     boolean isConfigurationBuilderType = isConfigurationBuilderType(element);
 
                     List<PropertyElement> beanProperties = element.getBeanProperties();
                     boolean hasDynamicBeanProperties = beanProperties.stream().anyMatch(PythonStubGenerator::isDynamicBeanProperty);
+                    // Declared attributes are fully represented by generated Java fields. Custom
+                    // Python properties may depend on hidden state that cannot be reconstructed in
+                    // another context from the BeanIntrospection alone.
+                    boolean isReconstructibleBean = isIntrospectedBean && !hasDynamicBeanProperties;
```
dstepanov, line 311: `isReconstructibleBean` looks only at the class's own properties, so reconstructibility isn't transitive.

Take the `Person`/`Address` pair from `GenerateToStringEqualsSpec`: if `Address` gains a custom `@property`, it stays a plain `ValueCoercible` while `Person` is still marked `PooledValueCoercible`. `Person.asPolyglotValue(ctx)` then emits `coerceToContext(this.address, arg1, python.Address.class)`, which hits `default -> value` in the three-arg `coerceToContext` (it only special-cases `PooledValueCoercible`), so the raw Java wrapper is set as the Python attribute. Python sees a host object and `address.city` fails at runtime — the wrapper only exposes `getCity()` — with nothing pointing at the cause.

Either check property types transitively when computing this flag, or fail loudly at compile time when a reconstructible bean has a non-reconstructible Python property.

---

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
```diff
@@ -1024,23 +1136,41 @@ public static Value newInstance(PythonClassReference classReference, Map<String,
...
-                    );
-                }
-            }
+            populateProperties(instance, props);
             return instance;
         });
     }
 
+    private static void populateProperties(Value instance, Map<String, Object> props) {
+        if (!props.isEmpty()) {
```
dstepanov: Dropping the `props != null` guard that the extracted callers had makes the public `@UsedByGeneratedCode` overloads (`newInstance(…, Map)`, `newFrozenDataclassInstance(…, Map)`, `newUninitializedInstance(…, Map)`) NPE on a null map instead of no-op'ing. Generated code always passes `AnnotationUtil.mapOf(...)`, but these are public entry points on an exported class — keeping the null check is cheap.

---

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
```diff
@@ -2721,14 +2779,16 @@ static void coerceParameterToPolyglotValue(
...
+        if (mapOfPython) {
             parameter = RUNTIME_UTIL.invokeStatic("coerceMap", TypeDef.of(Map.class), methodParam);
-        } else if (genericType.isAssignable(List.class) && genericType.getTypeArguments().get("E") instanceof PythonClassElement) {
+        } else if (listOfPython) {
             parameter = RUNTIME_UTIL.invokeStatic("coerceList", TypeDef.of(List.class), methodParam);
         } else {
             parameter = methodParam;
         }
-        if (targetContext != null) {
+        if (targetContext != null && !(genericType instanceof PythonClassElement) && !mapOfPython && !listOfPython) {
```
dstepanov: Worth a comment explaining these three exclusions, because they turn off the target-context coercion for exactly the types this PR just taught to convert themselves.

As far as I can tell this is *not* a regression — pre-PR these wrappers were plain `ValueCoercible`, and the three-arg `coerceToContext` only special-cases `PooledValueCoercible`, so they fell through to `default -> value` anyway. And the pooled call paths re-coerce at the boundary (`invokePooledScript` / `invokePooledMethod` → `coerceArgumentsToContext`), which is why `/pool/body` passes.

But two generated call sites have no such boundary: `GraalPyRuntimeUtil.invokePythonMethod` passes `arguments` straight through to the helper, and the property setters call `Value.putMember` directly rather than `GraalPyRuntimeUtil.putMember`. For those, a `Dataclass`, `list[Dataclass]` or `dict[str, Dataclass]` parameter now reaches the target context as a raw host wrapper with the new `asPolyglotValue(Context)` never consulted. If the exclusion is here to avoid double conversion after `coerceList`/`coerceMap`, passing the original parameter into `coerceToContext` (rather than the already-coerced one) would get both.

---

context-python/src/main/java/io/micronaut/context/python/PythonPool.java
```diff
@@ -423,26 +416,77 @@ private Context getOrCreateEventLoopContext(PythonEventLoop eventLoop) {
...
+                            wait();
+                        } else {
+                            long now = System.nanoTime();
+                            Duration timeUntilWarn = Duration.ofNanos(lastWarned + warnThreshold.toNanos() - now);
+                            if (!timeUntilWarn.isPositive()) {
+                                LOG.warn("No available Python contexts; waiting {} so far (pool target: {}, current: {}, queue: {})", Duration.ofNanos(now - start), targetSize, size.get(), pooledQueue.size());
+                                lastWarned = now;
+                                timeUntilWarn = warnThreshold;
+                            }
+                            wait(timeUntilWarn.toMillis(), timeUntilWarn.toNanosPart() % 1_000_000);
```
dstepanov: Two readability notes on this block:

- The single `creatingContext` slot means only one thread may build a context at a time, so N cold concurrent requests pay N × context-build latency serially. I assume that's deliberate (bounding the memory/CPU spike of concurrent GraalPy context builds), and it's why both new specs and the JMH benchmark needed a `warmPool` helper — but the comment on the `break` only says "we can create a context". One line saying *why* creation is serialized would save the next reader the deduction.

- `wait(timeUntilWarn.toMillis(), timeUntilWarn.toNanosPart() % 1_000_000)` is correct — `toNanosPart()` is the nano-of-second and the modulo extracts the sub-millisecond remainder that `toMillis()` already excludes — but it reads like an off-by-1000 bug. Worth a short comment.

---

inject-python-test/src/test/groovy/io/micronaut/python/annotation/processing/test/web/PooledControllerSpec.groovy
```diff
@@ -92,4 +198,25 @@ def pair() -> str:
             System.setProperty("micronaut.python.context-id.enabled", previousContextIdProperty)
         }
     }
+
+    private static void warmPool(ApplicationContext context, def executor, int size) {
```
dstepanov: This helper is duplicated verbatim in `PooledClassSpec` — it could move to `AbstractPythonTypeElementSpec`.

It also assumes the *default* pool size is at least 4. `computeDefaultSize()` is `max(1, availableProcessors * 2)`, so on a single-CPU runner only 2 of the 4 submitted tasks can acquire a context, the other two block in `borrow0`, and `assert acquired.await(30, TimeUnit.SECONDS)` fails after a 30s stall. Pinning `micronaut.python.pool.size` in the spec's application context would make both this and the downstream `toSet().size() >= 2` assertions deterministic.

---

benchmarks/src/jmh/java/io/micronaut/python/benchmark/PythonRequestBodyPoolBenchmark.java
```diff
@@ -0,0 +1,232 @@
...
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class PythonRequestBodyPoolBenchmark {
+
+    private static final String EXPECTED_BODY = "Ada:compiler:3:True:benchmark";
+    private static final String JSON = """
+        {"customer":"Ada","product":"compiler","quantity":3,"priority":true,"note":"benchmark"}
+        """;
+
+    /** Configured Python context pool size. */
+    @Param({"1", "2", "4", "8"})
```
dstepanov: 4 param values × 8 `@Benchmark` methods = 32 trials for a plain `:benchmarks:jmh` run, each booting an embedded server and warming a pool of GraalPy contexts. Since `-Pjmh.poolSizes` now exists to drive the sweep, defaulting this to a single value (`@Param({"4"})`) would keep the permanent benchmark cheap while still letting the 1/2/4/8 comparison be reproduced on demand.

---

benchmarks/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     id("io.micronaut.build.internal.convention-base")
+    id("io.micronaut.build.internal.python")
```
dstepanov, line 3: Applying the Python convention plugin here means every `:benchmarks` build now pulls in GraalPy, including for the many benchmarks unrelated to Python. Together with dropping `exclude("GRAALPY-VFS/**")` from `jmhJar` below, that's a noticeable build-time and jar-size cost for people running the HTTP/routing benchmarks. Both changes are needed for this benchmark to work, so no objection — just confirming that's the intended trade-off rather than a side effect.

---

dstepanov (commented):
Reviewed both commits against `5.2.x` at `02ee217`. All 15 checks are green.

The core idea looks right: request bodies were being materialized in the primary context eagerly in the `@JsonCreator` constructor and then handed to a borrowed pooled context. Dropping that eager `newUninitializedInstance` and adding a context-aware `asPolyglotValue(Context)` on reconstructible beans is the correct shape of fix, and the generated-source assertions in `GenerateToStringEqualsSpec` / `PyronautCompilerTest` pin it down well. `releasedContextCanBeBorrowedWhileAnotherContextIsBeingCreated` and `conversionFailureReleasesBorrowedContext` are good deterministic regressions for the pool rewrite.

Findings inline, roughly in priority order:

1. **`PythonPool.borrow0` can hang forever after close, and can wedge graceful shutdown.** The wait loop never checks `closed`, `closePool()` doesn't reset `size`, and `release()` drops contexts once closed — so a waiter woken by the shutdown `notifyAll()` re-enters `wait()` permanently. Swallowing `InterruptedException` makes it uninterruptible too, where the old `borrow()` threw. This is the one I'd want addressed before merge.
2. **`borrow0(false)` runs inside `ConcurrentHashMap.computeIfAbsent`**, holding a bin lock across a monitor wait and a full GraalPy context build.
3. **The new `coerceValue` enum branch returns a primary-context `Value`** that nothing downstream re-homes — reintroducing the cross-context hazard this PR fixes, for enum properties.
4. **`asPolyglotValue(Context)` neither caches nor guards re-entrancy**, unlike the no-arg version: repeated calls yield distinct Python objects, shared sub-objects get duplicated, and a cyclic graph recurses until `StackOverflowError`.
5. **`isReconstructibleBean` isn't transitive** — a reconstructible bean with a non-reconstructible Python property silently passes a raw host wrapper into the target context.

Plus smaller notes on the `coerceParameterToPolyglotValue` exclusions, the dropped `props != null` guard, serialized context creation, the duplicated `warmPool` helper and its implicit ≥2-CPU assumption, and the benchmark's 32-trial default matrix.

The `warnWaitMs` → `warnWait` rename is fine — 5.2.0 is still `-SNAPSHOT`, so no config compatibility concern.

Nothing here undermines the central fix.

---

context-python/src/main/java/io/micronaut/context/python/PythonPool.java
graemerocher: Fixed. `borrow0` now checks `closed` on every loop iteration, closure notifies waiters and resets pool state, and interruption re-asserts the flag and fails immediately. Added tests for both close-while-waiting and interruption.

context-python/src/main/java/io/micronaut/context/python/PythonPool.java
graemerocher: Fixed. Context construction now happens outside the `ConcurrentHashMap` callback, followed by `putIfAbsent`; a losing context is removed from the cache and closed.

context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
graemerocher: Fixed. Target-context conversion now routes enums through the context-aware pooled reconstruction path. Foreign `Value` instances are rejected instead of being passed into another context.

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
graemerocher: Fixed with conversion-scoped identity tracking and early registration immediately after allocating the target object. Added repeated-argument identity, diamond identity, and cyclic graph tests.

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
graemerocher: Fixed. Reconstruction now recursively context-coerces nested wrappers, collections, maps, and arrays. A non-reconstructible wrapper or raw `Value` from another context fails loudly instead of leaking a host wrapper into Python. The support now applies to any representable `@Introspected` Python class, not only dataclasses.

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
graemerocher: Fixed. The public generated-code overloads accept nullable property maps again, and `populateProperties` preserves the prior null/no-op behavior. Added a runtime test.

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
graemerocher: Clarified and fixed the boundary. Ordinary generated Python-class arguments intentionally remain host bridges so Python can call their generated Java accessors. Pooled stubs now pass raw arguments to `invokePooled`/`invokePooledScript`, which performs all recursive conversion only after borrowing the execution context. Full Python HTTP regression tests pass.

context-python/src/main/java/io/micronaut/context/python/PythonPool.java
graemerocher: Added comments documenting why context creation is serialized and how the millisecond/nanosecond split for `Object.wait` is calculated.

inject-python-test/src/test/groovy/io/micronaut/python/annotation/processing/test/web/PooledControllerSpec.groovy
graemerocher: Fixed. `warmPool` is now shared in `AbstractPythonTypeElementSpec`, and the relevant contexts explicitly configure a pool size of 4 for deterministic behavior on small runners.

benchmarks/src/jmh/java/io/micronaut/python/benchmark/PythonRequestBodyPoolBenchmark.java
graemerocher: Fixed. The default is now a single pool size (`4`), and the benchmark Javadoc documents `-Pjmh.poolSizes=1,2,4,8` for the full sweep.

benchmarks/build.gradle.kts
graemerocher: Confirmed: this is an intentional benchmark-only trade-off. The convention plugin and VFS resources are required for the end-to-end Python HTTP fixture; they increase the benchmark build/jar cost but do not affect production artifacts.

context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
```diff
@@ -249,8 +253,7 @@ public static boolean isNone(@Nullable Value value) {
     @UsedByGeneratedCode
     public static @Nullable Object coerceValue(@Nullable Object value) {
         return switch (value) {
-            case ValueCoercible valueCoercible when !(value instanceof PooledValueCoercible) ->
-                valueCoercible.asPolyglotValue();
+            case ValueCoercible valueCoercible -> valueCoercible.asPolyglotValue();
```
dstepanov: Dropping the enum special case is right, but collapsing the guard also removed the `!(value instanceof PooledValueCoercible)` exclusion, and I think that part over-shoots.

Pooled wrappers now resolve through the **no-arg** `asPolyglotValue()` here, which returns the stored value or builds one in the primary context. The only consumer is `putMember`, which then calls `coerceToContext(value, target.getContext())` — and that now hits the new `case Value polyglotValue ->` branch and throws `"Cannot pass a polyglot Value to a different context"` when they differ.

Concretely, in the legacy no-arg `asPolyglotValue()` sync path:

```java
GraalPyRuntimeUtil.putMember(this.graalpyInternalValue, "address", GraalPyRuntimeUtil.coerceValue(this.address));
```

If `this.graalpyInternalValue` came from `fromPolyglotValue(...)` on a value returned by a pooled call, it lives in a pooled context. If `this.address` was then set from Java and has no stored value, `coerceValue` materializes it in the *primary* context and `putMember` throws. Before this commit the pooled exclusion returned the raw wrapper, `putMember` → `coerceToContext` → `coercePooledValue(…, pooledCtx)` reconstructed it in the right context, and it worked.

Keeping the exclusion gives enums the correct-context resolution I was asking for *and* keeps every pooled wrapper deferred to the context-aware path:

```java
return switch (value) {
    case ValueCoercible valueCoercible when !(value instanceof PooledValueCoercible) ->
        valueCoercible.asPolyglotValue();
    case null, default -> value;
};
```

`nonReconstructibleWrapperFromAnotherContextFailsLoudly` covers the plain-`ValueCoercible` case, which is the one that genuinely should throw; I don't see coverage for a pooled wrapper reached through the no-arg sync path with a non-primary stored value. Worth a check either way — if the collapse is deliberate, a test pinning it would help.

---

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
```diff
@@ -2719,18 +2829,32 @@ static void coerceParameterToPolyglotValue(
         List<ExpressionDef> parameters,
         VariableDef.MethodParameter methodParam,
         @Nullable ExpressionDef targetContext) {
+        if (targetContext != null) {
+            // Generated Python-class arguments stay as host bridges for ordinary methods because
+            // Python code can call their generated Java methods. Pooled bridges instead pass raw
+            // arguments to invokePooled, which converts them after borrowing the target context.
+            parameters.add(param.getGenericType() instanceof PythonClassElement
+                ? methodParam
```
dstepanov: Thanks — the rationale comment answers the question I had. One asymmetry it doesn't cover, though: a bare Python-class argument stays a host bridge, but `list[Dataclass]` / `dict[str, Dataclass]` now fall into the `coerceToContext` arm and their *elements* get reconstructed as real Python objects.

That's a behavior change from the previous commit, not just a cleanup. Before, list/map params went through `coerceList`/`coerceMap`, whose `coerced instanceof PooledValueCoercible ? v : coerced` handed the raw wrappers back, so elements stayed host bridges like the bare case. Now:

```python
def f(o: Order):        # o is a host bridge  -> o.getCustomer()
def f(os: list[Order]): # os[0] is a Python object -> os[0].customer
```

So the same Python code shape works or breaks depending on whether the parameter is wrapped in a list. If converting the elements is the intent, the comment above is a good place to say why lists differ, and a spec exercising a `list[Dataclass]` parameter on the ordinary (non-pooled) bridge would pin it — I don't see one today, so this reads as unintentional fallout from moving the coercion ahead of the `coerceList`/`coerceMap` branch.

Same question for the async injection path: `PythonPooledStubGenerator` now passes injected values raw, which is right for `injectPooledScript` (it re-coerces per context via `coerceInjectedValue`, and storing raw actually fixes a latent cross-context reuse), but the `async` branch takes `asyncMemberValue(script, value)` instead, which wraps the raw host object rather than a converted `Value`. Worth confirming injected *Python* beans still adapt correctly there.

---

dstepanov (commented):
Re-reviewed at `8788768`. Thanks — every point from the previous round is addressed, and several are addressed more thoroughly than I suggested.

Verified fixed:

- **Shutdown hang** — `closed` is now checked at the top of the wait loop, `InterruptedException` re-asserts the flag and throws, and `closePool()` resets `size`. `waitingBorrowFailsWhenPoolCloses` and `waitingBorrowRespondsToInterruption` cover both.
- **`computeIfAbsent`** — replaced with borrow-outside + `putIfAbsent` under the monitor, closing the loser and dropping its cache entry.
- **Enum branch** — gone; see the one inline comment below on how far the collapse went.
- **Identity and cycles** — the `ScopedValue`/`ContextConversion` memo plus `rememberPooledValue` before property population is a better answer than the identity map I suggested: the frozen-dataclass path now allocates first via `setInstanceProperty` too, so cycles work there as well rather than throwing. `coerceArgumentsPreservesPooledWrapperIdentity`, `pooledConversionSupportsCyclesAfterTargetAllocation` and the `Node`/`Pair` end-to-end spec (`left is right and left.next is left`) pin it down.
- **Non-transitive reconstructibility** — the new `ValueCoercible` arm turns the silent host-object degradation into a named `IllegalArgumentException`, with `nonReconstructibleWrapperFromAnotherContextFailsLoudly`.
- **Null props** — restored, `@Nullable`-annotated, and covered by `uninitializedInstanceAcceptsNullProperties`.
- **`warmPool`** — moved to `AbstractPythonTypeElementSpec`, and both specs now pin `micronaut.python.pool.size: 4`, so the ≥2-CPU assumption is gone.
- **Benchmark** — `@Param({"4"})` with the sweep documented via `-Pjmh.poolSizes`.
- Comments added for the serialized-creation rationale and the `wait(millis, nanos)` arithmetic.

Two things from this commit worth a second look, both inline: the `coerceValue` collapse may have gone one step further than needed for pooled wrappers on the legacy sync path, and `list[Dataclass]` parameters now convert their elements while a bare `Dataclass` parameter stays a host bridge.

One process note: the headline 14.4x in the description was measured on the original implementation. This commit puts a `ScopedValue` binding and two `IdentityHashMap`s on the per-conversion hot path, so a re-run of the benchmark before merge would be worth having.

Nit: the new `closeContext` helper delegates straight to `GraalPyContextFactory.closeContext` without the cancellation-tolerant handling that `createBorrowedPooledContext` does inline, so a cancelled-context `PolyglotException` during shutdown will propagate out of `getOrCreateEventLoopContext` — and in the pool-closed branch it would replace the `IllegalStateException` you meant to throw.

CI is still running on this head (`audit` and `build_matrix` green so far); I haven't seen the Java and Python CI results yet.

---

context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
```diff
@@ -333,20 +361,90 @@ public static boolean isNone(@Nullable Value value) {
...
+        CURRENT_CONTEXT_CONVERSION.get().values.put(source, value);
+    }
+
+    private static <T> T withContextConversion(Context context, Supplier<T> operation) {
+        if (CURRENT_CONTEXT_CONVERSION.isBound() && CURRENT_CONTEXT_CONVERSION.get().context.equals(context)) {
+            return operation.get();
+        }
+        return ScopedValue.where(CURRENT_CONTEXT_CONVERSION, new ContextConversion(context)).call(operation::get);
+    }
+
```
dstepanov, line 453: I think this, rather than machine variance, explains the control-endpoint regression in the updated Evidence table (27,688 → 13,146 ops/s).

`withContextConversion` binds the `ScopedValue` and constructs a `ContextConversion` unconditionally, and `ContextConversion`'s two `IdentityHashMap` fields are eagerly initialized:

```java
private final IdentityHashMap<PooledValueCoercible, Value> values = new IdentityHashMap<>();
private final IdentityHashMap<PooledValueCoercible, Boolean> inProgress = new IdentityHashMap<>();
```

So every call through `coerceArgumentsToContext` pays a scope binding plus two map allocations **even when `args` is empty or entirely primitives**, and `putMember` pays the same on every individual member set.

`/control` takes no body and does no conversion work at all, yet it goes through `invokePooledMethod` → `coerceArgumentsToContext` on every request. A uniform per-call tax on a path that previously did nothing is exactly the shape that shows up as a control-endpoint regression while the body endpoint still improves hugely against a pathological baseline. Machine noise would be unlikely to move the control ~2× in one direction while leaving the body result plausible.

Two cheap ways out, either of which keeps the identity semantics intact:

- Skip the binding entirely when there is nothing to memoize — e.g. return early from `coerceArgumentsToContext` when `args.length == 0`, and from `withContextConversion` when the value can't contain a `PooledValueCoercible` (`isInteropPrimitive`, `String`, etc.).
- Allocate the maps lazily on the first `coercePooledValue` call, so a conversion that never touches a pooled wrapper costs only the binding.

Worth re-measuring the control endpoint after either — if it returns to ~27k, that confirms the cause and gives the Evidence table a trustworthy control again.

Two notes on the description while you're in there:

- The Baseline column looks like it came from the original long run, while Current is a 2 × 1 s warmup / 3 × 1 s measurement run. For a JIT- and GraalPy-heavy workload that's far from steady state, and it isn't like-for-like — a same-session run of both sides with a real warmup would make the multiplier meaningful.
- The Verification section has empty inline-code spans that swallowed the names: "reports are  and ;" and "Full affected checks pass for , , , and  with  (121 + 104 + 583 + 196 tests)."

---

dstepanov (commented):
Re-reviewed at `a4b81eb`. Both open points from the last round are fixed, and the list/map one landed in the better direction than the alternative I floated:

- **`coerceValue`** — the `!(value instanceof PooledValueCoercible)` exclusion is restored, with `putMemberDefersPooledWrapperConversionToTargetContext` pinning that pooled wrappers reach the target context through `coerceToContext` rather than the primary-context no-arg path.
- **List/map asymmetry** — resolved by keeping host bridges *everywhere* rather than converting elements, so `list[Item]` and `dict[str, Item]` match the bare-argument case. The new spec exercising `describe(self, item: Item, items: list[Item], items_by_name: dict[str, Item])` in one signature is exactly the right shape for this, and the updated comment now states the rule.
- **Async injection** — `asyncMemberValue` routes `PooledValueCoercible`/`Value` through `coerceToContext` instead of wrapping the raw host object, with a test.
- **`closeContext`** — cancellation-tolerant handling is now in the shared helper and used by both call sites.

One new thing inline, and I'd treat it as worth resolving before merge rather than after: I don't think the control-endpoint regression in the updated Evidence table is machine variance. `withContextConversion` binds the `ScopedValue` and eagerly allocates two `IdentityHashMap`s on every `coerceArgumentsToContext` and every `putMember`, including for zero-argument calls that convert nothing — which is precisely what `/control` does on every request. Two cheap fixes suggested; either should restore the control number and, more importantly, give the table a control that actually controls.

No concerns left on the correctness side. CI is running fresh on this head; Python CI hasn't reported yet.

---

context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
graemerocher: Fixed. `coerceValue` again excludes `PooledValueCoercible`, so `putMember` defers reconstruction to its target context instead of materializing a primary-context value. `putMemberDefersPooledWrapperConversionToTargetContext` verifies the no-arg overload is never called.

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
graemerocher: Fixed. Ordinary bare, `list[...]`, and `dict[..., ...]` Python arguments now consistently remain host bridges; the end-to-end `Consumer.describe` regression exercises all three shapes. Async injection now reconstructs only `PooledValueCoercible` beans in the event-loop context while leaving non-reconstructible client proxies as async host bridges. The focused runtime test and all 11 `PythonAsyncioSpec` tests pass.

graemerocher:
Also addressed the review-summary cleanup nit: all event-loop loser and pool-closed context cleanup now uses the cancellation-tolerant close helper, so an already-cancelled PolyglotException cannot replace the intended pool-closed failure. The full benchmark sweep was rerun after these changes and the PR evidence was updated to 10,271 ops/s versus the unchanged 1,137 ops/s request-body baseline at 8 workers/pool 8, with the observed control-run variance called out explicitly.

dstepanov (approved):

context-python/src/main/java/io/micronaut/context/python/GraalPyRuntimeUtil.java
graemerocher: Fixed in 25e51aec34. coerceArgumentsToContext now returns immediately for empty or entirely interop-primitive arguments, direct primitive conversion bypasses the ScopedValue session, and ContextConversion allocates both identity maps lazily. A regression test pins the primitive-argument fast path. I also reran pre-fix and fixed revisions consecutively with identical settings (8 workers, pool 8, 5 x 2 s warmup, 5 x 2 s measurement): body 1,124 -> 33,987 ops/s (~30.2x), control 38,214 -> 37,842 ops/s (-1.0%). The PR description now contains these matched results and the corrected report/module names. Full affected checks pass (122 + 104 + 583 + 196 tests), including Spotless and Checkstyle.

PR #12949, opened 2026-08-31
Add @ContentDisposition annotation for declarative response headers

### LINKED ISSUE #9480: Setting the 'Content-Disposition' header for arbitrary content

ShubhJ46:
Closes #9480.

Right now the only ways to send a `Content-Disposition` header are to return an `HttpResponse` directly or add a `Filter`, even though it's an independent concern from the return type (as described in the issue). This PR adds a `@ContentDisposition` annotation you can put on a controller method or class:

```java
@Get("/report")
@ContentDisposition(filename = "report.csv")
Flux<String> report() {
    ...
}
```

`type` defaults to `attachment`; set it to `inline` to have the browser display the response instead of downloading it. `filename` is optional.

It's wired the same way `@Status` already is - `RouteInfo` resolves the header value once from the annotation metadata, and `RouteExecutor` applies it in `forStatus`, which is shared by every response-building path (direct body, reactive/streamed body, empty response, etc). That also means it applies to `Flux`/`Publisher` return types, which was the original reporter's use case.

I also moved the RFC 6266 filename-encoding logic that already existed in `StreamedFile` (for the `attach()` API) into a small shared `ContentDispositionUtils` in the `http` module, so both that code path and the new annotation build headers the same way instead of duplicating the encoding logic.

Since this is a new public annotation and a new default method on `RouteInfo`, I'd rather get feedback on the shape of the API before polishing further - happy to adjust `type` to something more type-safe (e.g. an enum) or change the defaulting behavior if that's preferred.

Tested:
- `./gradlew :micronaut-http:check :micronaut-router:check :micronaut-http-server:check :micronaut-http-server-tck:check :test-suite-http-server-tck-netty:check :test-suite-http-server-tck-jdk:check` all pass
- Added a TCK test (`ContentDispositionTest`) that runs against both the Netty and JDK server implementations


ShubhJ46:
The failing \`build (25)\` job is unrelated to this change - it's \`io.micronaut.http.client.jdk.ClientProxySpec > test downloading via http proxy using proxy-address\`, which does a live proxied request to https://micronaut.io/ and asserts the response body contains a specific HTML fragment. The site's homepage has since been redesigned, so the fragment no longer matches. This PR doesn't touch \`micronaut-http-client-jdk\` at all, only \`http\`, \`router\`, \`http-server\`, and \`http-server-tck\`.

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
```diff
@@ -0,0 +1,121 @@
...
+     * @since 4.10.0
+     */
+    public static String toHeaderValue(String type, @Nullable String filename) {
+        if (filename == null || filename.isEmpty()) {
+            return type;
+        }
+        // https://httpwg.org/specs/rfc6266.html#advice.generating
+        // 'filename' parameter is the fallback for legacy browsers, 'filename*' is the supported approach.
+        return type + "; filename=\"" + sanitizeAscii(filename) + "\"; filename*=utf-8''" + encodeRfc6987(filename);
+    }
```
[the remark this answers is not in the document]
ShubhJ46: Addressed this by making `type` a `ContentDisposition.Type` enum (`ATTACHMENT`/`INLINE`) instead of adding a runtime check — that makes an invalid or CRLF-containing value structurally impossible to construct at the annotation layer. Didn't add a separate check in the utility itself either: any value set via `MutableHttpResponse.header(...)` already goes through `CaseInsensitiveMutableHttpHeaders`'s RFC 7230 field-content validation before it reaches a response, so a second check here would just duplicate that.

http-server/src/test/groovy/io/micronaut/http/util/ContentDispositionUtilsEncodingSpec.groovy
```diff
@@ -0,0 +1,21 @@
...
+            sb.appendCodePoint(codePoint)
+            String s = sb.toString()
+
+            def encoder = new QueryStringEncoder('')
+            encoder.addParam("foo", s)
+            def encoded = encoder.toString().substring("?foo=".length())
+
+            assert ContentDispositionUtils.encodeRfc6987(s) == encoded
+        }
+    }
```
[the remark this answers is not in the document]
ShubhJ46: Left this as-is for now. It's pre-existing coverage relocated from StreamedFileSpec rather than something new in this PR, and it runs in about a second, so trading exhaustive BMP coverage for speed didn't seem worth it. I did add a dedicated supplementary-plane case alongside it in the latest commit. Open to switching to sampling if CI time becomes a real concern.

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
```diff
@@ -0,0 +1,123 @@
...
+package io.micronaut.http.util;
+
+import io.micronaut.core.util.StringUtils;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Utility class to build RFC 6266 compliant {@code Content-Disposition} header values.
+ *
+ * @author Shubham Jain
+ * @since 4.10.0
```
yawkat: 5.2.0

yawkat: please keep the comment that this is copied from netty

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public final class ContentDispositionUtils {
```
yawkat, line 29: `@Internal`

router/src/main/java/io/micronaut/web/router/RouteInfo.java
```diff
@@ -292,6 +292,18 @@ default boolean isImperative() {
      */
     HttpStatus findStatus(@Nullable HttpStatus defaultStatus);
 
+    /**
+     * Finds the {@code Content-Disposition} header value declared via
+     * {@link io.micronaut.http.annotation.ContentDisposition}, if any.
+     *
+     * @return The header value, or {@code null} if none was declared
+     * @since 4.10.0
```
sdelamo:
```suggestion
     * @since 5.2.0
```

sdelamo (approved):

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
sdelamo:
```suggestion
 * @since 5.2.0
```

sdelamo (changes requested):
We need a section in the documentation about this new annotation.

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
```diff
@@ -0,0 +1,123 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.util;
+
+import io.micronaut.core.util.StringUtils;
+import org.jspecify.annotations.Nullable;
```
sdelamo, line 20:
```suggestion
import org.jspecify.annotations.Nullable;
import io.micronaut.core.annotation.Internal;
```

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
sdelamo:
```suggestion
@Internal
public final class ContentDispositionUtils {
```

http/src/main/java/io/micronaut/http/util/ContentDispositionUtils.java
ShubhJ46: Done — all the `@since` tags on this feature are now `5.2.0` (there were a couple more spots I'd missed beyond the one you and sdelamo caught).

ShubhJ46: That comment was never removed — it's still there above encodeRfc5987 (just renamed from encodeRfc6987 per your other comment). Let me know if you meant a different spot.

test-suite/src/test/java/io/micronaut/docs/server/response/ContentDispositionController.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.ContentDisposition;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Get;
+
+@Requires(property = "spec.name", value = "contentdisposition")
+@Controller("/content-disposition")
+public class ContentDispositionController {
```
graemerocher, line 26: please add snippets for the other languages (Kotlin, Groovy and Python)

graemerocher (changes requested):

test-suite/src/test/java/io/micronaut/docs/server/response/ContentDispositionController.java
ShubhJ46: Added Kotlin, Groovy and Python versions, following the same StatusController/StatusControllerSpec pattern each language module already uses (plus the missing toc-python.yml entry for the Python guide).

One caveat on the Python one: I couldn't get the exact GraalVM JDK 25 toolchain running locally to actually execute the Python test - the GraalVM CE build I could install (25.3.4) is newer than what this repo's pinned Truffle/polyglot library version (25.2.4) supports, so `test-suite-python:test` gets skipped on my machine with a version-check error rather than actually running. The code is a direct port of the Java/Kotlin/Groovy versions (which I did verify run and pass) using the same nested-enum access pattern as `EngineConfig.py`'s `MapFormat$MapTransformation`, so I'm fairly confident in it, but flagging that it hasn't actually executed on my end - happy to fix anything CI turns up.

ShubhJ46:
Hi @graemerocher @sdelamo, thanks again for taking the time to review this. The Kotlin, Groovy and Python snippets have been pushed as requested. It looks like the workflows on the latest commit are still waiting for approval, so CI has not had a chance to run yet. Whenever one of you has a moment, could you kick it off? Happy to rebase if needed, and glad to make any further changes you would like.

graemerocher:
could you retarget 5.3.x since it is too late for that the 5.2.x release 

thanks

yawkat (approved):

PR #11546, opened 2025-01-28
Cleanup DelayedExecutionFlowImpl after completion

dstepanov:
Attempt to reduce memory consumption for https://github.com/micronaut-projects/micronaut-core/issues/10677

PR #12950, opened 2026-09-01
Avoid sharing incomplete flows in Reactor defuse

### LINKED ISSUE #12940: DelayedExecutionFlowImpl memory leak persists in 4.10.x — defuse() returns shared flow on keep-alive connections
### Expected Behavior

# DelayedExecutionFlowImpl memory leak persists in 4.10.x — `defuse()` returns shared flow on keep-alive connections

Image

## Expected Behavior

On a long-lived HTTP/1.1 keep-alive connection, each request should use an independent `ExecutionFlow`. Memory usage should remain stable regardless of how many requests are served on a single connection.

## Actual Behavior

`DelayedExecutionFlowImpl` step nodes (`Map`, `FlatMap`, `OnErrorResume`, `OnComplete`) accumulate indefinitely on keep-alive connections, growing by ~8 steps per request until OOM. This is the same symptom reported in #10677.

**Heap histogram from production (Micronaut 4.10.26, ~20 million requests served):**

```
  59,821,130  io.micronaut.core.execution.DelayedExecutionFlowImpl$Map
  59,821,124  io.micronaut.core.execution.DelayedExecutionFlowImpl$OnErrorResume
  39,880,756  io.micronaut.core.execution.DelayedExecutionFlowImpl$FlatMap
  19,940,378  io.micronaut.core.execution.DelayedExecutionFlowImpl$OnComplete
```

## Relationship to #10677 and PR #11546

PR #11546 added cleanup of `DelayedExecutionFlowImpl` internal state **after completion**. This helps when a flow completes normally. However, in the scenario described below, the shared flow **never completes**, so the cleanup never runs, and step nodes accumulate without bound.

## Root Cause Analysis

The root cause is in `ReactorExecutionFlowImpl.defuse()` ([source](https://github.com/micronaut-projects/micronaut-core/blob/4.10.x/core/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java)):

```java
// ReactorExecutionFlowImpl.defuse()
if (publisher instanceof FlowAsMono<T> flowAsMono) {
    return flowAsMono.flow;  // <-- returns the SAME underlying flow every time
}
```

When a route handler returns a reactive type (e.g., `Mono<HttpResponse>`), Micronaut's request pipeline converts it through:

```
Mono → ReactorExecutionFlowImpl → (filter chain) → FlowAsMono → defuse() → underlying flow
```

The problem is that on a keep-alive connection, `defuse()` unwraps `FlowAsMono` and returns the **same** underlying `DelayedExecutionFlowImpl` from a previous request. This flow was created for the first request and never completed (the completion signal was consumed by the subscriber path, not the flow path).

Each subsequent request then appends its filter chain steps to this shared flow:

```
Request 1: flow → Map → OnErrorResume → Map → OnErrorResume → Map → OnComplete → Map → FlatMap
Request 2:      → Map → OnErrorResume → Map → OnErrorResume → Map → OnComplete → Map → FlatMap
Request 3:      → Map → OnErrorResume → ...
...
Request N: chain grows by ~8 nodes per request, never GC'd until connection closes
```

The responses are still written correctly through the Reactor subscriber path (`FlowAsMono.subscribe()`), so the server functions normally — it's a silent memory leak.

## Proposed Fix

In `ReactorExecutionFlowImpl.defuse()`, when unwrapping a `FlowAsMono`, only return the underlying flow if it has already completed (immutable, safe to reuse). Otherwise, fall through to the subscriber path which creates an independent flow:

```java
if (publisher instanceof FlowAsMono<T> flowAsMono) {
    ImperativeExecutionFlow<?> completed = flowAsMono.flow.tryComplete();
    if (completed != null) {
        return (ExecutionFlow<T>) completed;
    }
    // fall through to subscriber path — creates independent flow
}
```

This preserves the optimization for completed flows while preventing step accumulation on incomplete/shared flows.

## Workaround

We added a Netty `ChannelDuplexHandler` that limits max requests per keep-alive connection (default 10,000). When the limit is reached, it adds `Connection: close` to the response, forcing the client to create a new connection and breaking the step accumulation chain.

## Environment Information

- Micronaut: 4.10.26
- JDK: 21
- OS: Linux (Kubernetes)
- Application: [Kestra](https://github.com/kestra-io/kestra) fork, webserver mode
- Production traffic: ~78K req/s across multiple pods, keep-alive connections from nginx upstream

### Actual Behaviour

_No response_

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.10.26

swaran21:
## Problem

`ReactorExecutionFlowImpl.defuse()` directly unwraps every `FlowAsMono`. When the wrapped `ExecutionFlow` is incomplete and is observed more than once, this exposes the same mutable delayed flow to each caller. Downstream `map`, `flatMap`, `onErrorResume`, and `onComplete` operations can then be appended to the shared chain.

The cleanup in #11546 releases delayed-flow nodes after completion, but it cannot help while a shared source remains incomplete.

## Solution

- Reuse a `FlowAsMono` source only when `tryComplete()` returns a completed imperative flow.
- Fall through to the existing subscriber path for incomplete flows so every conversion receives an independent delayed flow.
- Preserve the completed-flow fast path.

## Tests

Added regression coverage that:

1. Defuses the same incomplete `FlowAsMono` twice and verifies that the resulting flows are independent.
2. Confirms that a completed `FlowAsMono` still returns an `ImperativeExecutionFlow` with the expected value.

The first test fails on the previous implementation because both conversions return the shared underlying flow.

## Verification

Using GraalVM CE 25.1.3 (JDK 25.0.3):

- `:micronaut-http:test --tests "io.micronaut.http.reactive.execution.ReactorExecutionFlowImplSpec" --rerun-tasks --no-build-cache`: passed, 11/11 tests.
- `:micronaut-http:check --rerun-tasks --no-build-cache`: passed.

I also attempted the repository-wide `check` on Windows. It reached `micronaut-http-client` but has two unrelated environment/platform failures:

- `HttpHeadSpec`: expects `a/b`, while `java.nio.file.Path` produces `a\b` on Windows.
- `PcapLoggingSpec`: expected generated PCAP files were absent.

Fixes #12940

### DIFF
--- a/http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
+++ b/http/src/main/java/io/micronaut/http/reactive/execution/ReactorExecutionFlowImpl.java
@@ -71,9 +71,12 @@ final class ReactorExecutionFlowImpl implements ReactiveExecutionFlow<Object> {
                 return ExecutionFlow.error(t);
             }
         } else if (publisher instanceof FlowAsMono<T> flowAsMono) {
-            // unwrap directly
-            //noinspection unchecked
-            return (ExecutionFlow<T>) flowAsMono.flow;
+            ImperativeExecutionFlow<?> completed = flowAsMono.flow.tryComplete();
+            if (completed != null) {
+                //noinspection unchecked
+                return (ExecutionFlow<T>) completed;
+            }
+            // An incomplete flow may be shared. Subscribe below to create an independent flow.
         }
 
         // special subscriber that (a) contains the propagated context and (b) can return an
--- a/http/src/test/groovy/io/micronaut/http/reactive/execution/ReactorExecutionFlowImplSpec.groovy
+++ b/http/src/test/groovy/io/micronaut/http/reactive/execution/ReactorExecutionFlowImplSpec.groovy
@@ -151,6 +151,48 @@ class ReactorExecutionFlowImplSpec extends Specification {
         flow.tryCompleteValue() == "foobar"
     }
 
+    @Issue("https://github.com/micronaut-projects/micronaut-core/issues/12940")
+    def 'defuse incomplete FlowAsMono creates independent flows'() {
+        given:
+        Hooks.resetOnOperatorDebug()
+        DelayedExecutionFlow<String> source = DelayedExecutionFlow.create()
+        def mono = ReactorExecutionFlowImpl.toMono(source)
+
+        when:
+        def first = ReactorExecutionFlowImpl.defuse(mono, PropagatedContext.empty())
+        def second = ReactorExecutionFlowImpl.defuse(mono, PropagatedContext.empty())
+
+        then:
+        !first.is(source)
+        !second.is(source)
+        !first.is(second)
+        first.tryComplete() == null
+        second.tryComplete() == null
+
+        when:
+        source.complete("foo")
+
+        then:
+        first.tryCompleteValue() == "foo"
+        second.tryCompleteValue() == "foo"
+    }
+
+    @Issue("https://github.com/micronaut-projects/micronaut-core/issues/12940")
+    def 'defuse completed FlowAsMono preserves immediate flow'() {
+        given:
+        Hooks.resetOnOperatorDebug()
+        DelayedExecutionFlow<String> source = DelayedExecutionFlow.create()
+        def mono = ReactorExecutionFlowImpl.toMono(source)
+        source.complete("foo")
+
+        when:
+        def flow = ReactorExecutionFlowImpl.defuse(mono, PropagatedContext.empty())
+
+        then:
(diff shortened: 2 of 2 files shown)


yawkat:
It is not permitted to reuse ExecutionFlows, including calling defuse twice like in the test case. I think this is the wrong fix. Will close for now.

swaran21:
Thanks for clarifying. I misunderstood the ExecutionFlow contract and used an unsupported scenario by calling defuse() twice on the same flow.
I understand now that ExecutionFlows are not reusable or intended for multiple subscribers, so the test does not represent a valid server execution path. I won’t pursue this implementation further.

PR #12953, opened 2026-09-01
Add JPMS support for core and inject

### LINKED ISSUE #6395: Add module-info to micronaut-inject and micronaut-core

dstepanov:
Closes #6395

## Summary

- add explicit JPMS descriptors for `micronaut-core` and `micronaut-inject`
- declare Micronaut's standard service-loader uses/providers in those descriptors
- discover `META-INF/micronaut` entries through named modules and instantiate non-exported generated definitions with a module-aware lookup
- add a named-module integration application that boots `ApplicationContext`, resolves a bean, exercises service loading and static optimizations, and builds a `jlink` image
- keep aggregated Javadoc in classpath mode because it combines source sets with split packages

## Reproducer and root cause

Adding descriptors alone reproduces the original failure: the application starts as a named module but bean resolution fails. `ClassLoader.getResources("META-INF/micronaut/")` does not enumerate metadata inside named modules, and a public method-handle lookup cannot construct generated definitions in a non-exported application package.

The new integration suite captures both cases. `MicronautMetaServiceLoaderUtils` supplements class-loader scanning with `ModuleReader` scanning and uses a private lookup after adding the required read edge. A named Micronaut application therefore opens its generated-code package to core, for example:

```java
opens com.example to io.micronaut.core;
```

## Compatibility note

This PR defines the modules as `io.micronaut.core` and `io.micronaut.inject`. Existing 5.1.x manifests advertise `io.micronaut.micronaut_core` and `io.micronaut.micronaut_inject` as automatic module names. Since module names are public JPMS API, this is intentionally a draft pending agreement on whether the canonical names are appropriate for 5.x/a future major, or whether the descriptors should preserve the existing automatic names.

## Verification

```text
./gradlew :micronaut-core:check :micronaut-inject:check :test-suite-javax-inject:test :test-suite-jpms:check
./gradlew docs
```

The core and inject jars also pass `jar --validate`, and the application runs from the generated linked runtime image.


graemerocher:
I started this a while ago since there are some challenges with class loading and netty. This PR is more complete I believe https://github.com/micronaut-projects/micronaut-core/pull/12413

PR #12958, opened 2026-09-02
Read a live annotation as an AnnotationValue

dstepanov:
## The problem

`Qualifiers.byAnnotation(Annotation)` (#12928) reads the members off a live annotation and stores them the way compiled metadata stores them: a `Class` as an `AnnotationClassValue`, an enum by the name of its constant, a nested annotation as an `AnnotationValue`, arrays of each likewise. That is what lets a qualifier built at runtime compare equal to one compiled. The conversion lived in private helpers of `AnnotationMetadataQualifier` (`resolveAnnotationBindingValues`, `readMember`, `asMemberValue`, `asAnnotationValue`, `resolveAllValues`).

Any framework built on Micronaut that is handed annotation instances at runtime needs exactly this conversion. The Jakarta CDI implementation ([micronaut-cdi](https://github.com/micronaut-projects/micronaut-cdi)) is handed them constantly (`Instance.select(Annotation...)`, `Event.select(...)`, `BeanContainer.isMatchingBean(Set<Annotation>)`, synthetic-bean qualifiers) and had to carry its own copy of the reader, in two places (`CdiAnnotations.valueOf`/`storedForm` and `SynthesisRunner.membersOf`). The copy drifted from core's: it did not convert a nested annotation member, so a qualifier such as `@Located(region = @Region("east"))` selected by its literal was unsatisfied, because the live `Region` proxy never equalled the stored `AnnotationValue`. That has since been fixed downstream, but no downstream should have to carry the copy at all.

## The API

In `core`, on `io.micronaut.core.annotation.AnnotationValue`:

```java
public static <T extends Annotation> AnnotationValue<T> of(T annotation)
```

It reads every zero-argument, non-synthetic member of `annotation.annotationType().getDeclaredMethods()` off the instance (via `ReflectionUtils.invokeInaccessibleMethod`), members left at their default included since a live instance answers every one of them, and returns an `AnnotationValue` named after `annotationType().getName()` with `RetentionPolicy.RUNTIME`.

Conversion rules, the same ones `asMemberValue` applied:

| Reflection returns | Recorded as |
|---|---|
| `Class` | `AnnotationClassValue` |
| `Enum` | its `name()` |
| nested `Annotation` | `AnnotationValue.of(nested)`, recursively, every member |
| `Class[]` / `Enum[]` / `Annotation[]` | an array of the converted form |
| anything else (primitives, `String`, arrays of either) | as is |

A member that cannot be read, because the annotation type is not open to this module or because the instance throws (or answers `null`), is reported with an `IllegalStateException` naming the member and the annotation type. It is never silently skipped.

`@NonBinding` is **not** applied by the factory. Whether a member takes part in the comparison of two qualifiers is the qualifier's concern, not the conversion's, and `NonBinding` lives in `inject` anyway.

## What changes in `inject`

`AnnotationMetadataQualifier.fromAnnotation` now calls `AnnotationValue.of(annotation)` and then drops the members whose declaring method carries `@NonBinding`. The private reader helpers are deleted. Its observable behaviour is preserved:

- a member that cannot be read: the factory's exception is caught and the qualifier falls back to comparing by type alone (`null`), as before;
- no binding member left: `null`, as before;
- `@NonBinding` on a member of a *nested* annotation is not honoured; the nested annotation is compared whole, as before (`resolveAllValues` read it whole; the factory does too). This is now covered by a test rather than assumed.

## What deliberately does not change

- `Qualifiers.byAnnotation(Annotation)` and its behaviour, including the specs added in #12928, which pass unchanged.
- `@NonBinding` handling: it stays in the qualifier.
- `AnnotationValue.equals`: still compares the annotation name and the declared `getValues()` maps (same size, `AnnotationUtil.areEqual` per member, so arrays element-wise and nested `AnnotationValue`s recursively). Defaults are not consulted by `equals`; that is what `matches` is for.

## Tests

- `core`: `AnnotationValueOfAnnotationSpec` covers primitive/String members, enum, class, nested annotation, arrays of enum/class/nested annotation, defaults being present, equality of two instances written the same way, and the unreadable-member exception and message.
- `inject-java`: `AnnotationValueOfAnnotationSpec` compiles a qualifier with nested-annotation, enum, class and array members, and asserts that `AnnotationValue.of(literal)` `equals` (both directions, same `hashCode`) the `AnnotationValue` the bean definition was compiled with, member by member and as a whole, and that a bean is resolved by the live literal. This is the exact equality the CDI bug turned on. A second feature confirms that `@NonBinding` on a nested annotation's member is not honoured.
- Existing `QualifierMemberSpec` (inject-java) and `QualifierByAnnotationSpec` (inject) from #12928 pass unchanged.

Ran locally:

- `./gradlew :micronaut-core:test :micronaut-inject:test :micronaut-core-processor:test`: core 7393 tests, 0 failures, 2 skipped, inject 312 tests, 0 failures, core-processor has no test sources (`NO-SOURCE`).
- `./gradlew :micronaut-inject-java:test`: 4962 tests, 0 failures, 8 skipped.
- `:micronaut-core:checkstyleMain :micronaut-inject:checkstyleMain` pass.

## Two corner cases that are not identical to before

Both arise only when readability differs *per member* — an annotation proxy that throws for one member and answers the rest. When the whole type is inaccessible, every path already fell back to comparing by type alone, exactly as before.

- An unreadable member annotated `@NonBinding` used to be skipped before it was read, so it did not force the type-only fallback. The factory now reads every member first, so it throws and the qualifier falls back to type-only.
- An unreadable member of a *nested* annotation used to be silently left out of the nested value. It now throws, and the qualifier falls back to type-only.

Both follow from the factory refusing to silently drop a member it cannot read, which is the behaviour a general-purpose reader should have. Flagged here so the choice is a reviewed one rather than an accident.

### DIFF
--- a/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
+++ b/core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
@@ -68,6 +68,26 @@ import static io.micronaut.core.reflect.ReflectionUtils.EMPTY_CLASS_ARRAY;
  */
 public class AnnotationValue<A extends Annotation> implements AnnotationValueResolver {
 
+    private static final String ANNOTATION_PARAMETER = "annotation";
+
+    /**
+     * The members of each annotation type, by name, in declaration order: the zero-argument methods it declares,
+     * looked up once per type rather than on every read of an instance.
+     */
+    private static final ClassValue<Map<String, Method>> ANNOTATION_MEMBERS = new ClassValue<>() {
+        @Override
+        protected Map<String, Method> computeValue(Class<?> annotationType) {
+            Method[] declared = annotationType.getDeclaredMethods();
+            Map<String, Method> members = CollectionUtils.newLinkedHashMap(declared.length);
+            for (Method member : declared) {
+                if (member.getParameterCount() == 0 && !member.isSynthetic()) {
+                    members.put(member.getName(), member);
+                }
+            }
+            return Collections.unmodifiableMap(members);
+        }
+    };
+
     private final String annotationName;
     private final ConvertibleValues<Object> convertibleValues;
     private final Map<CharSequence, Object> values;
@@ -1549,7 +1569,7 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
      * @since 4.0.0
      */
     public static <T extends Annotation> AnnotationValueBuilder<T> builder(AnnotationValue<T> annotation) {
-        ArgumentUtils.requireNonNull("annotation", annotation);
+        ArgumentUtils.requireNonNull(ANNOTATION_PARAMETER, annotation);
         return new AnnotationValueBuilder<>(annotation, annotation.getRetentionPolicy());
     }
 
@@ -1562,7 +1582,7 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
      * @return The builder
      */
     public static <T extends Annotation> AnnotationValueBuilder<T> builder(AnnotationValue<T> annotation, @Nullable RetentionPolicy retentionPolicy) {
-        ArgumentUtils.requireNonNull("annotation", annotation);
+        ArgumentUtils.requireNonNull(ANNOTATION_PARAMETER, annotation);
         return new AnnotationValueBuilder<>(annotation, retentionPolicy);
     }
 
@@ -1593,7 +1613,7 @@ public class AnnotationValue<A extends Annotation> implements AnnotationValueRes
      * @since 5.2.0
      */
     public static <T extends Annotation> AnnotationValue<T> of(T annotation) {
-        ArgumentUtils.requireNonNull("annotation", annotation);
+        ArgumentUtils.requireNonNull(ANNOTATION_PARAMETER, annotation);
         Class<? extends Annotation> annotationType = annotation.annotationType();
         Map<String, Method> members = ANNOTATION_MEMBERS.get(annotationType);
         Map<CharSequence, Object> values = CollectionUtils.newLinkedHashMap(members.size());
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -1564,6 +1566,114 @@ public static <T extends Annotation> AnnotationValueBuilder<T> builder(Annotatio
...
+     * @param <T>        The annotation type
+     * @return The annotation value, with {@link RetentionPolicy#RUNTIME} retention
+     * @throws IllegalStateException When a member cannot be read, because the annotation type is not open to this
+     *                               module or because the instance does not answer for it
+     * @since 5.2.0
+     */
+    public static <T extends Annotation> AnnotationValue<T> of(T annotation) {
+        ArgumentUtils.requireNonNull("annotation", annotation);
+        Class<? extends Annotation> annotationType = annotation.annotationType();
+        Method[] members = annotationType.getDeclaredMethods();
```
graemerocher: perhaps this should be cached in `Map<String, Method>` with key being the member name to avoid repeated retrieval.

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
```diff
@@ -1564,6 +1566,114 @@ public static <T extends Annotation> AnnotationValueBuilder<T> builder(Annotatio
...
+    }
+
+    /**
+     * A member value read off an annotation instance, in the form compiled metadata records it.
+     *
+     * @param value The value as reflection returns it
+     * @return The value as metadata records it
+     */
+    private static Object toMemberValue(Object value) {
+        if (value instanceof Class<?> type) {
```
graemerocher: this can probably be a switch expression

graemerocher (changes requested):

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
dstepanov: Done in f9fdfa1: the members are now looked up once per annotation type through a `ClassValue<Map<String, Method>>`, keyed by member name in declaration order; `of()` iterates that map.

core/src/main/java/io/micronaut/core/annotation/AnnotationValue.java
dstepanov: Done in f9fdfa1: `toMemberValue` is a switch expression now. One wrinkle: `Class<?>[]` and `Enum<?>[]` are not accepted as case patterns (javac reads the `<` as a comparison and reports "different case kinds"), so those two arrays are matched with a guarded `Object[]` case.

graemerocher (approved):

dstepanov:
Absorbed into #12977, which is where this work now lives: the five commits are on that branch unchanged, and `ReflectionAnnotations` in the new `micronaut-reflection` module builds on `AnnotationValue.of(Annotation)` rather than carrying a second copy of the conversion, which is what this PR set out to prevent. Closing in favour of that one.

PR #12959, opened 2026-09-02
Expose the target constructor to constructor interceptors of proxied beans

dstepanov:
### Problem

A bean with both around advice and constructor interception (`@Around` + `@AroundConstruct`) is constructed through the generated proxy definition. The proxy constructor declares the target constructor's parameters followed by five internal ones (resolution context, bean context, qualifier, interceptor registrations and interceptor registry).

`ConstructorInterceptorChain` already trimmed those internal values out of `getParameterValues()`, but `getConstructor()`, `getArguments()` and therefore `getDeclaringType()` still described the proxy constructor. A `ConstructorInterceptor` saw the proxy class as the declaring type and five more arguments than parameter values. A plain `@AroundConstruct` bean, which is not proxied, was consistent.

### Fix

Runtime only, no processor change. The chain keeps two constructors: the proxy constructor it invokes (unchanged, `proceed()` still appends the internal values) and a trimmed view it exposes to interceptors. When the bean definition is an `AdvisedBeanType` and internal parameters are present, the view reports:

- the intercepted type as the declaring type
- only the target constructor's arguments (the first N proxy arguments, which are the target's parameter elements)
- the constructor's annotation metadata
- `instantiate(values)` delegating to the proxy constructor with the internal values appended, so `ctx.getConstructor().instantiate(ctx.getParameterValues())` works

Trace logging uses the view's description (`MyBean(Alpha alpha,Beta beta)`). Interceptor selection still runs on the raw constructor before the chain is built, and `DefaultBeanContext` diagnostics keep describing the constructor that is actually invoked.

### Tests

New `AroundConstructTargetViewSpec` covers an advised bean with injected dependencies, instantiating through `getConstructor()` from the interceptor, a no-arg advised bean, and the unchanged plain `@AroundConstruct` case. Without the fix three of the four features fail. `AroundConstructCompileSpec`, `LifecycleInterceptorCompileSpec`, `LifecycleInterceptorOrderingSpec` and `LifecycleInterceptorReuseSpec` all pass.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
@@ -289,6 +289,10 @@ public final class ConstructorInterceptorChain<T> extends AbstractInterceptorCha
         private final BeanConstructor<T> proxyConstructor;
         private final Class<T> declaringBeanType;
         private final Argument<?>[] arguments;
+        /**
+         * The values of the internal parameters the proxy constructor declares after the bean's own ones. Never
+         * empty: this view only exists when the proxy constructor declares such parameters.
+         */
         private final @Nullable Object[] internalParameters;
 
         private InterceptedTargetConstructor(BeanConstructor<T> proxyConstructor,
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

dstepanov:
@copilot fix sonar

PR #12963, opened 2026-09-02
Resolve a repeatable annotation through its container in the by-name metadata queries

dstepanov:
## Symptom

A repeated annotation is stored under its container while the stereotype index names the member. For `@Q("a") @Q("b")` (with `@Repeatable(Qs.class) @Qualifier @interface Q`), `declaredAnnotations` and `allAnnotations` hold `Qs`, `annotationsByStereotype[jakarta.inject.Qualifier]` holds `Q`. The `Class` overloads bridge the two through `findRepeatableAnnotationContainerInternal`; the `String` overloads did not. Probe on unmodified `5.2.x`, identical for the compile-time `MutableAnnotationMetadata` and for the generated `$Definition` metadata, on a class, a field, a constructor parameter and a method parameter:

```
hasDeclaredAnnotationQs:true, hasAnnotationQs:true,
annotationNamesByStereotype:[repeatabledeclared.Q],
declaredAnnotationValuesByNameQ:[a, b], annotationValuesByNameQ:[a, b],
declaredAnnotationNamesByStereotype:[],      <- expected [repeatabledeclared.Q]
hasDeclaredAnnotationQ:false,                <- hasDeclaredAnnotation(Q.class) is true
hasDeclaredStereotypeQ:false,
hasAnnotationQ:false,                        <- hasAnnotation(Q.class) is true
hasStereotypeQ:false                         <- hasStereotype(Q.class) is true
```

`getDeclaredAnnotationNamesByStereotype` was the worst case: it filtered the index entries against `declaredAnnotations` keys, so every repeatable member was dropped and the result was always empty.

## Fix

One private helper in `DefaultAnnotationMetadata`, `containsAnnotation(map, name)`: present if the map has the name itself or, when `findRepeatableAnnotationContainerInternal(name)` is non-null, its container. It replaces the bare `containsKey` in every by-name presence check: `hasDeclaredAnnotation(String)`, `hasAnnotation(String)`, `hasStereotype(String)`, `hasDeclaredStereotype(String)`, the filter and the fallback of `getDeclaredAnnotationNamesByStereotype`, and the fallback of `getAnnotationNamesByStereotype`. This mirrors what the `Class` overloads already do rather than introducing a second lookup.

Cost: a direct hit is unchanged; only a miss pays one extra map lookup for the container. No new allocation on any path. `MutableAnnotationMetadata` overrides only `findRepeatableAnnotationContainerInternal`, so the compile-time and the deserialised runtime forms share the fix without changes there.

## Deliberately left alone

- `findDeclaredAnnotation(String)` / `findAnnotation(String)` / `getDeclaredAnnotation(String)`: a repeatable has no single value to return by name. The `Class` overloads (interface defaults) already return the first repeated value via `getDeclaredAnnotationValuesByType`; making the `String` form do the same needs the repeatable-resolution path, not the one-line presence rule.
- `getDeclaredAnnotationValuesByType(Class)` / `getDeclaredAnnotationValuesByName(String)`: already symmetric, both map through the container.
- Observable side effects of the widened presence checks, noted rather than changed: `synthesizeDeclared(Q.class)` now builds the first repeated value instead of returning `null` (it already did so for `synthesize(Q.class)`); `synthesize(X.class, "Q")` with the member's *name* as the source now yields a default-valued proxy instead of `null`; `getValue("Q", member, type)` now falls back to `Q`'s default value where it returned empty. All are the `String` form catching up with the `Class` form.

## Tests

`inject-java/src/test/groovy/io/micronaut/inject/annotation/repeatable/RepeatableDeclaredQueriesSpec.groovy` compiles one source with `@Q("a") @Q("b")` on a class, an injected field, a constructor parameter and an `@Executable` method parameter, and probes all ten by-name queries on each element in both forms:

- compile-time form via `buildClassElement` (`MutableAnnotationMetadata` behind the `ClassElement`, `FieldElement`, `ParameterElement`);
- generated form via `buildBeanDefinition` (`DefaultAnnotationMetadata` deserialised from the `$Definition`, its injected field, constructor argument and executable-method argument), plus the `Class` overloads `hasDeclaredAnnotation(Q.class)` / `hasDeclaredStereotype(Q.class)` as the reference.

Both feature methods fail on unmodified `5.2.x` with the probe output above and pass with the fix.

```
./gradlew :micronaut-inject:checkstyleMain :micronaut-inject:test :micronaut-inject-java:test :micronaut-core-processor:test
  :micronaut-inject:checkstyleMain   SKIPPED   (enabled = false in inject/build.gradle.kts)
      re-enabled via init script: 306 pre-existing violations in 61 files, none in the changed lines
      (the one hit in DefaultAnnotationMetadata.java, line 993 `newAnnotationValue`, predates this change)
  :micronaut-inject:test             312 tests, 0 failures
  :micronaut-inject-java:test        4962 tests, 0 failures, 8 skipped (pre-existing @Ignore/@PendingFeature)
  :micronaut-core-processor:test     NO-SOURCE
BUILD SUCCESSFUL
```

## Scope narrowed after CI

The first push also resolved a repeatable through its container in `hasDeclaredAnnotation`, `hasAnnotation`,
`hasStereotype` and `hasDeclaredStereotype`. That broke `inject-kotlin`'s `ConfigurationMetadataSpec` (three
features): the Kotlin element model asks `hasStereotype(Property)` to tell a data-class configuration property
from a parameter that already carries `@Property`, and once the container answered for the member, every
property of a `@ConfigurationProperties` data class read as already annotated and was left out of the
generated configuration metadata. Bisected to that one method; baseline `5.2.x` passes the spec locally.

So the by-name `has*` queries keep meaning "declared under its own name" — the compiler relies on it — and
**only `getDeclaredAnnotationNamesByStereotype` changes**, which is the query the finding is about. The spec
now states the kept answers explicitly.

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/annotation/repeatable/RepeatableByNameLookupSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/annotation/repeatable/RepeatableByNameLookupSpec.groovy
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.inject.annotation.repeatable
+
+import io.micronaut.annotation.processing.test.AbstractTypeElementSpec
+import io.micronaut.core.annotation.AnnotationMetadata
+import io.micronaut.inject.BeanDefinition
+
+import java.lang.annotation.Annotation
+
+/**
+ * Looking a repeatable annotation up by name must find the first repeated value, exactly like the
+ * {@code Class} overload does. A repeatable is stored under its container, so the plain map lookup
+ * these queries do misses it — even for a single {@code @Q("a")}, which is wrapped too.
+ */
+class RepeatableByNameLookupSpec extends AbstractTypeElementSpec {
+
+    private static final String Q = 'repeatablebyname.Q'
+
+    private static final String SOURCE = '''
+package repeatablebyname;
+
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import jakarta.inject.Qualifier;
+import jakarta.inject.Singleton;
+
+@Singleton
+@Q("a")
+@Q("b")
+class Multi {
+}
+
+@Singleton
+@Q("a")
+class Single {
+}
+
+@Retention(RetentionPolicy.RUNTIME)
+@Qualifier
+@Repeatable(Qs.class)
+@interface Q {
+    String value();
+}
+
+@Retention(RetentionPolicy.RUNTIME)
+@interface Qs {
+    Q[] value();
+}
+'''
+
+    void "test finding a repeatable annotation by name at compile time"() {
+        given:
+        AnnotationMetadata multi = buildClassElement(SOURCE).getAnnotationMetadata()
+
+        expect:
+        multi.findAnnotation(Q).get().stringValue().get() == 'a'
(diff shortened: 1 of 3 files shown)


graemerocher (approved):

dstepanov:
Narrowed in 54312719ab after CI: the first version also widened the by-name `has*` queries, which broke Kotlin data-class configuration metadata (`ConfigurationMetadataSpec`). Only `getDeclaredAnnotationNamesByStereotype` changes now — details added to the description. The remaining red checks on the previous head were netty/GraalPy timeouts (`HttpPipeliningSpec`, `ProxyBackpressureTest`, `NettyPythonAsyncioRuntimeTest`), not this change.

Issue #12129, opened 2025-10-15
@EventListener methods are not invoked in bean order


mikehearn:
### Expected Behavior

```kotlin
@Singleton class First {
    @EventListener fun onFoo(event: FooEvent) { println("first") }
}

@Singleton class Last : Ordered {
    override fun getOrder(): Int = 10
    @EventListener fun onFoo(event: FooEvent) { println("last") }
}
```

should always print `first, last`

### Actual Behaviour

Ordering is undefined and varies from run to run.

The reason is twofold:

1. There's a bug in `OrderUtils` (my bad, I'll send a PR). But also...
2. The generated ApplicationEventListener classes don't implement `Ordered` and forward it to the underlying class, nor copy `@Ordered`.

I think `AopProxyWriter` needs to be extended to notice if the target class implements `Ordered` or exposes `@Ordered` metadata and if so, duplicate that to the generated proxy.

### Version

5.0-SNAPSHOT

graemerocher:
Triage: credible and likely valid event-listener ordering bug, although the issue lacks a standalone reproducer. The stated failure has two independent candidates: OrderUtils behavior and generated ApplicationEventListener/AOP proxy definitions not preserving Ordered or order metadata. Current source has ordering tests for ordinary beans but no matching proxied event-listener regression. Suggested fix: add a minimal Kotlin or Java compile/runtime test with one un-ordered and one Ordered listener, verify generated proxy metadata and listener sorting, then fix the first failing layer. No duplicate or open PR was found.

zeguilherme99:
Hi @graemerocher, can I work on it?

graemerocher:
This might be fixed in 5.2.x already

zeguilherme99:
@graemerocher So is this task closed? 

I'm trying to find a first issue to start contributing here; do you have any advice?

graemerocher:
checkout the ones labelled good first issue https://github.com/micronaut-projects/micronaut-core/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22info%3A%20good%20first%20issue%22

PR #12965, opened 2026-09-02
Control the order in which singletons are destroyed

### LINKED ISSUE #6493: Micronaut doesn't give the application code control of the order in which resources are gracefully shutdown/closed
### Feature description

From conversations with @jameskleeh @jeffbrown it appears that micronaut uses bean dependencies to gracefully shutdown the beans created. This does not follow a functional order of resource termination in a service. Other frameworks do allow for shutdown hooks to be ordered but micronaut doesn't. 

Consider the example in the following link. (See: https://twitter.github.io/finatra/user-guide/getting-started/lifecycle.html#server-shutdown-lifecycle)

> For example, you have code which is reading from a queue (via a “subscriber”), transforming the data, and then publishing (via a “publisher”) to another queue. When the main application is exiting you most likely want to close the “subscriber” first to ensure that you transform and publish all available data before closing the “publisher”.
> 
> Assuming, that both objects are a c.t.util.Closable type, a simple way to close them would be:
> 
> ```closeOnExit(subscriber); closeOnExit(publisher)```

With micronaut there doesn't seem to be such control. If I defined a service class like this
```
   class ServiceClass(kafkaConsumer, kafkaProducer, dao, cache) {
   }
```
I have no idea in which order (kafkaConsumer, kafkaProducer, dao, and cache) are closed. I may want the following order
 - consumer to stop consuming
 - wait until dao is done with the job before shutting down db connection
 - cache to be updated before disconnecting cache
 - produce the last message and only then shutdown producer. 

and the needs may change in the future but right now, there doesn't seem to be a way to control that.

dstepanov:
Closes #6493. Also covers the `@Order` annotation half of #12129.

### Problem

Singletons are destroyed in dependency order, but the sort had gaps and there was no way to declare an order between beans that are not wired together:

- `getRequiredComponents()` checked the **bean class's** metadata for `@Inject` instead of the field's, so field-injected dependencies were never considered. A field-injected dependency could be destroyed before its dependent (reproduced in 3 of 4 naming variants).
- Beans not connected by a dependency edge were destroyed in `ConcurrentHashMap` order, which is identity-hash based and changed from run to run.
- The workaround recommended in the issue, `@Order` on `@EventListener` methods for `ShutdownEvent`, did not work because the generated adapter bean carried no `@Order` metadata (class-level `@Order` and `Ordered` did work).

### Changes

- **New `@DependsOn`** (`io.micronaut.context.annotation`): lists bean types that are created before the annotated bean is instantiated and destroyed after it, without injecting them. Works on classes and factory methods, accepts interfaces (every implementation is covered), and fails with a clear `BeanInstantiationException` when a listed type has no bean. Creation is enforced in `resolveByBeanFactory`; destruction falls out of the listed types being required components.
- **`AbstractInitializableBeanDefinition.getRequiredComponents()`** reads the field argument's metadata, so `@Inject` fields are dependencies (`@Value`/`@Property` fields stay excluded), and includes `@DependsOn` types.
- **`DefaultBeanContext.topologicalSort`** is now a Kahn-style sort. A bean is destroyed before every bean it requires; where dependencies leave the order open, beans are destroyed in bean name order so the sequence is deterministic. Cycles are broken at the first bean. Candidate lookup per required type is cached.
- **`DeclaredBeanElementCreator.visitAdaptedMethod`** copies a method-level `@Order` onto the adapter bean; a method-level value overrides a class-level one.
- **Docs**: new "Destruction Order" section in the lifecycle guide with a multi-language `MessageConsumer`/`MessagePublisher` snippet backed by `DependsOnSpec` in the Java, Groovy and Kotlin test suites.

### Tests

`BeanDestructionOrderSpec` (13 cases): field injection across naming variants, `@Value` exclusion, constructor/method injection, `@DependsOn` on a class, on an interface and on a factory method, the missing-bean error, stable ordering, cycles, and `ShutdownEvent` listener ordering with method-over-class override. `DependsOnSpec` in all three language test suites verifies both creation and destruction order. Existing inject, inject-java (lifecycle/aop/context/field/constructor/factory/any), inject-groovy, inject-kotlin and test-suite lifecycle/event specs pass.

Note: `core-processor` checkstyle currently fails on `AopProxyWriter.java:188` (two blank lines). That is already on `5.2.x` and unrelated to this change.

### DIFF
--- a/src/main/docs/guide/ioc/lifecycle.adoc
+++ b/src/main/docs/guide/ioc/lifecycle.adoc
@@ -36,37 +36,17 @@ NOTE: Simply implementing the `Closeable` or `AutoCloseable` interface is not en
 
 When the context is closed, singletons are destroyed in dependency order: a bean is always destroyed before the beans injected into it, whether they were injected through the constructor, a method or a field. This lets a bean use its dependencies from its `@PreDestroy` method.
 
-Where the dependencies leave the order open, annotate the beans with ann:core.annotation.Order[] or implement api:core.order.Ordered[]. Beans with a lower value are destroyed first, so a consumer that should stop reading before the producer it feeds is closed can be declared as:
+Beans that are not injected into each other but still have to be created and destroyed in a particular order can declare the dependency with ann:context.annotation.DependsOn[]. The annotated bean is created after, and destroyed before, every bean of the listed types:
 
-[source,java]
-----
-@Singleton
-@Order(1) // <1>
-class MessageConsumer {
+snippet::io.micronaut.docs.lifecycle.MessageConsumer,io.micronaut.docs.lifecycle.MessagePublisher[tags="class", indent=0]
 
-    @PreDestroy
-    void stop() {
-        // stop consuming and flush pending messages
-    }
-}
+<1> The consumer depends on the publisher without injecting it
+<2> The consumer stops while the publisher is still open, so it can publish the messages it has in flight
+<3> The publisher is closed once the consumer has stopped
 
-@Singleton
-@Order(2) // <2>
-class MessagePublisher {
+The annotation also works on factory methods and accepts interfaces, in which case every bean implementing the interface is created before and destroyed after the annotated bean. Beans not connected by any dependency are destroyed in bean name order, so the destruction sequence is the same from run to run.
 
-    @PreDestroy
-    void close() {
-        // close the connection
-    }
-}
-----
-
-<1> The consumer has the lowest value and is destroyed first
-<2> The publisher is destroyed once the consumer has stopped
-
-A dependency always outlives its dependents, regardless of the `@Order` values. Beans of equal precedence are destroyed in bean name order, so the destruction sequence is the same from run to run.
-
-TIP: Shutdown logic that spans several beans can also be placed in ann:runtime.event.annotation.EventListener[] methods for api:context.event.ShutdownEvent[], which is published before any bean is destroyed. Such methods can be ordered with ann:core.annotation.Order[] as well.
+TIP: Shutdown logic that spans several beans can also be placed in ann:runtime.event.annotation.EventListener[] methods for api:context.event.ShutdownEvent[], which is published before any bean is destroyed. Such methods can be ordered with ann:core.annotation.Order[].
 
 == Dependent Beans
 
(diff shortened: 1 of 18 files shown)


src/main/docs/guide/ioc/lifecycle.adoc
```diff
@@ -32,6 +32,42 @@ snippet::io.micronaut.docs.lifecycle.Connection[tags="class", indent=0]
 
 NOTE: Simply implementing the `Closeable` or `AutoCloseable` interface is not enough for a bean to be closed with the context. One of the above methods must be used.
 
+=== Destruction Order
+
+When the context is closed, singletons are destroyed in dependency order: a bean is always destroyed before the beans injected into it, whether they were injected through the constructor, a method or a field. This lets a bean use its dependencies from its `@PreDestroy` method.
+
+Where the dependencies leave the order open, annotate the beans with ann:core.annotation.Order[] or implement api:core.order.Ordered[]. Beans with a lower value are destroyed first, so a consumer that should stop reading before the producer it feeds is closed can be declared as:
+
+[source,java]
```
graemerocher: should be a multi language snippet

src/main/docs/guide/ioc/lifecycle.adoc
```diff
@@ -32,6 +32,42 @@ snippet::io.micronaut.docs.lifecycle.Connection[tags="class", indent=0]
 
 NOTE: Simply implementing the `Closeable` or `AutoCloseable` interface is not enough for a bean to be closed with the context. One of the above methods must be used.
 
+=== Destruction Order
+
+When the context is closed, singletons are destroyed in dependency order: a bean is always destroyed before the beans injected into it, whether they were injected through the constructor, a method or a field. This lets a bean use its dependencies from its `@PreDestroy` method.
+
+Where the dependencies leave the order open, annotate the beans with ann:core.annotation.Order[] or implement api:core.order.Ordered[]. Beans with a lower value are destroyed first, so a consumer that should stop reading before the producer it feeds is closed can be declared as:
```
graemerocher: I am not sure we  should leave this down to numerical ordering. In Spring you can do can control this with `@DependsOn` which defines construction and destruction order so the below is expressed as:

```
----
@Singleton
@DependsOn(MessagePublisher.class)
class MessageConsumer {

    @PreDestroy
    void stop() {
        // stop consuming and flush pending messages
    }
}

@Singleton 
class MessagePublisher {

    @PreDestroy
    void close() {
        // close the connection
    }
}
```

So here the consumer is created after the publisher and destroyed before the publisher

graemerocher (changes requested):

graemerocher (approved):

PR #12968, opened 2026-09-02
Retain the annotations a composed annotation introduces

dstepanov:
Adds `@Retainable`, which keeps the annotations a composed annotation introduces attributed to it, instead of only flattening them into the element's name-keyed stereotype indexes.

## The feature

Mark an annotation `@Retainable` and every annotation composing it keeps that occurrence on its own `AnnotationValue`, with `@AliasFor` member overrides already applied:

```java
@Retainable
@interface Constraint { }

@Constraint
@interface Size { int min() default 0; }

@Size(min = 5)
@interface Username {
    @AliasFor(annotation = Size.class, member = "min")
    int min() default 5;
}
```

```java
// for an element annotated @Username(min = 3)
metadata.getAnnotation(Username.class).getStereotypes(); // [ @Size(min = 3) ]
```

The marker goes on the annotation being composed, so a framework annotation opts its whole family in at once: marking `jakarta.validation.Constraint` makes every constraint retained wherever it is composed, and no constraint author writes anything. An annotation is retainable when the marker is anywhere in its own stereotypes, and a retained occurrence keeps its own retainable stereotypes in turn, so the tree nests.

There is no new read API. `AnnotationValue.getStereotypes()` is already public and already documented as the stereotypes of the annotation; until now it was only populated during mapping and remapping.

## What it makes possible

`AnnotationMetadata` indexes stereotypes by name, so the association between an individual occurrence and the annotation that introduced it is not recoverable. For two composed constraints that each compose `@Size`:

```java
@Size(min = 5)  @interface ComposedA { @AliasFor(annotation = Size.class, member = "min") int min() default 5; }
@Size(max = 50) @interface ComposedB { @AliasFor(annotation = Size.class, member = "max") int max() default 50; }

@ComposedA(min = 3) @ComposedB(max = 9) class Test {}
```

```
getAnnotationValuesByName(Size)           -> [@Size(min=3), @Size(max=9)]   correct values, kept apart
getAnnotationNamesByStereotype(Size$List) -> [ComposedA, ComposedB]         a set union, keyed by the container
```

Nothing says which occurrence came from which composed annotation, and position is not a reliable proxy: occurrences are stored in a `LinkedHashSet` by `addRepeatableInternal`, so equal occurrences contributed by different parents collapse into one. Attribution becomes containment instead, one entry under each composed annotation, and repeatable composed annotations, `AliasFor.index` and nesting all fall out of the tree the builder already computes.

**Composed constraints.** `jakarta.validation.metadata.ConstraintDescriptor.getComposingConstraints()` is defined per composed constraint, so micronaut-validation reconstructs it with ~266 lines of annotation-type reflection in `DefaultConstraintDescriptor` — `getDeclaredAnnotations`, `getDeclaredMethods`, `Method#invoke`, `getDefaultValue`, `setAccessible(true)`. With the retained tree that becomes a walk over `parentAnnotationValue.getStereotypes()`, and micronaut-validation opts in by marking `jakarta.validation.Constraint` in the transformers it already ships.

**Interceptor bindings**, in this PR. `@InterceptorBinding` is `@Retainable`, so every annotation meta-annotated with it retains its binding occurrence. Two annotations composing the same binding annotation with different members on one element — `@NorthZone` and `@SouthZone` each composing `@Zone` — now resolve to both occurrences, which the flat index collapses to one.

## How the tree is stored

The tree is kept in a reserved annotation member, `AnnotationUtil.STEREOTYPES_MEMBER` (`"$stereotypes"`), as an `AnnotationValue[]` in the values map of the composing occurrence. This follows the existing `$nonBinding` convention. A member value is stored, cloned, merged across an `AnnotationMetadataHierarchy` and written by the code that already exists for every other member, so:

- `DefaultAnnotationMetadata`, `MutableAnnotationMetadata` and `AnnotationMetadataSupport` need no change, and no new constructors are added anywhere;
- the builder moves the retainable part of the computed tree into the member, in `addAnnotation`;
- the writer has one special case: the member is filtered to the runtime-retained annotations before it is emitted, and dropped when none is left, the same treatment source-retention annotations get at write time.

`AnnotationValue.getStereotypes()` reads the member when no stereotypes were computed for the value. The member is not an attribute: it is hidden from `getValues()`, `getMemberNames()`, `getConvertibleValues()` and `toString()`, so it takes no part in `equals`/`hashCode` and attaching stereotypes changes no existing comparison. Annotation proxies synthesised from the metadata are built from `getValues()`, so they never see it either.

## Transitive overrides now cascade

Before this change, `@Outer(shortest = 3)` overriding `Inner.min`, where `@Inner` overrides `Size.min`, yielded `@Inner(min = 3)` but `@Size(min = 1)`: `applyIntroducedAliases` rewrote the intermediate annotation's members after its subtree had been computed and did not re-run down the tree. The flat index reported the same value, so this was pre-existing #12911 behaviour, but it would have blocked the motivating consumer — micronaut-validation cascades today by re-applying `@OverridesAttribute` from each already-overridden parent, and the Jakarta Validation TCK covers constraint composition.

An override now re-applies the occurrence's own aliases with the overridden values to its existing subtree, recursively, so the retained tree and the flat index both report `@Size(min = 3)`. Re-deriving the aliases of a node further down needs that node's annotation type, and the tree holds only annotation values, so the type was looked up by name; where an implementation cannot resolve a type from its name alone — the builder already notes this for KSP and Groovy — the cascade stopped after one level. The builder now remembers every annotation type it saw with its native element and falls back to it, so the cascade reaches the leaves at any depth. Those entries are native elements of the session that produced them, so a frontend reusing one builder across processing rounds clears them at that boundary; KSP does, javac and Groovy do not need to.

## Interceptor bindings drop `$bindingValues`

Because `@InterceptorBinding` is retainable, the copy the compiler used to make is redundant. `InterceptorBindingMembers` no longer nests a copy of the binding annotation's members, minus the `@NonBinding` ones, inside every `@InterceptorBinding` it synthesises for a `bindMembers = true` binding. The binding keeps `bindMembers`, and `InterceptorBindingQualifier.resolveBoundOccurrences` compares the occurrences themselves on each side: the ones declared on the element, and the ones composed by another annotation and retained on it, with an occurrence declared on a method replacing the one on its class the way the bindings themselves are resolved. `DefaultInterceptorRegistry` matches through it, and so does the qualifier.

Two things surfaced on the way. The qualifier's injection-time member check compared the wrong object and had always matched by name only, leaving the registry to do the member matching per interception point; it now matches members wherever it has the metadata, which only removes interceptors the registry would have rejected. And the bindings the remapper skips (`@Around`, `@Introduction`, `@AroundConstruct`, `@InterceptorBinding` itself) keep only the marker among their stereotypes, so an annotation composing `@Around` is not made retainable and pays nothing.

## Cost and compatibility

Retention costs generated code proportional to the retainable occurrences at every use site, so the decision sits on the framework annotation that defines the contract. When nothing retainable is composed, the member is absent and the writer emits what it emitted before, so generated metadata classes are unchanged. `AnnotationValue.getValues()` gains one `containsKey` per call and copies the map only for occurrences that actually carry the member.

Metadata compiled by earlier versions keeps working. Bindings written before this change still carry `$bindingValues` and are read from it, which `BindingOccurrenceResolutionSpec` pins, including a binding read from a copy matching one read from the element.

This was also checked outside the build, by compiling with the **5.1.13** annotation processor and swapping in the core, inject, aop, context and core-reactive jars from this branch:

- an application whose generated metadata was verified to contain `$bindingValues` behaves identically to the 5.2.x baseline: member binding, a defaulted binding member, a method-level binding overriding a class-level one, and a name-only binding;
- both partial-upgrade directions work, since an application and its libraries are not recompiled together: a 5.1-compiled interceptor library against newly compiled beans, and newly compiled interceptors against 5.1-compiled beans, each including an annotation composing the binding annotation;
- **micronaut-data 5.1.3**, built against core 5.1.10, carries no `bindMembers` binding anywhere, so nothing of its precompiled metadata used the removed copy. Its 109 bean definitions and all four of its interceptor bindings resolve byte-for-byte identically on the 5.1.13 runtime and on this branch, and a context booted with its jars starts the same on both.

## Tests

`inject-java/src/test/groovy/io/micronaut/inject/annotation/StereotypeOriginSpec.groovy`, most assertions run against both the compile-time metadata and the written one:

- the flat index still cannot attribute the two occurrences, which is the behaviour being improved on;
- `ComposedA -> [@Size(min=3)]` and `ComposedB -> [@Size(max=9)]`, override applied per occurrence;
- a plainly composed `@NotNull` is reported under the annotation that introduced it and not under the other;
- the marker on a framework annotation makes every annotation meta-annotated with it retainable, without any of them declaring it, and the marker itself is never retained;
- nesting, `@Outer -> @Inner -> @Size`, with a transitive override cascading through it, through more than one intermediate annotation, and for annotations that retain nothing, since the cascade is not a retention feature;
- the shapes `@OverridesAttribute` needs: a constraint composing `@Size` twice with an index-selective override reaching each occurrence and neither leaking into the other, an override with no index reaching every occurrence, composing annotations written inside their container by hand, and a hand-rolled container that is not the declared `@Repeatable` one;
- written trees exclude `SOURCE` and `CLASS` retention annotations;
- the tree survives conversion to mutable metadata, cloning, merging a hierarchy, and being rebuilt through `mutate()` and merged back;
- the element shapes a validation integration reads: property metadata, container element metadata, inheritance from an interface, and a record component;
- an annotation composing nothing retainable retains nothing and behaves exactly as before.

`ValidationRetainedStereotypesSpec` marks `jakarta.validation.Constraint` through a transformer, the way micronaut-validation would. `core`'s `AnnotationValueSpec` covers the reserved member itself: read through `getStereotypes()`, hidden from the attributes and the convertible values, ignored by equality, and both halves of the seeded-builder contract. `ComposedBindingMembersSpec` covers the composed interceptor binding end to end, and `BindingMemberDefaultsSpec` reads the compared occurrences rather than the removed copy.

The change lives in the shared builder and writer, so all three languages get it; parity specs in `inject-groovy` and `inject-kotlin`, and the cascade has its own Groovy parity spec.

Green: `:micronaut-inject-java:test`, `:micronaut-inject:test`, `:micronaut-inject-groovy:test`, `:micronaut-inject-kotlin:test`, `:micronaut-core:test`, `:micronaut-aop:test`, japicmp and checkstyle on core/inject/aop/core-processor.


dstepanov:
## Where the opt-in marker should sit

I raised a concern about per-use-site cost under the `@Constraint`-based opt-in the Motivation section proposes, on the assumption that every constraint occurrence would carry `@Documented`, `@Retention`, `@Target` and friends. **That assumption was wrong** — those are in `AnnotationUtil.INTERNAL_ANNOTATION_NAMES` and are dropped in `processAnnotation` long before anything becomes a stereotype. I measured the real trees rather than guessing again.

For a leaf annotation that carries a family marker but composes nothing (the `@NotNull` / `@Size` shape):

```
leaf tree                  = [@Marker]
composed tree              = [@Marker, @Leaf]
Leaf subtree under Composed = [@Marker]
```

So under a transitive opt-in on `jakarta.validation.Constraint`, every `@NotNull` and `@Size` use site carries `$stereotypes = [@Constraint(validatedBy = […]), @Inherited]`. Two entries, not the pile I first claimed — but still paid at every constraint occurrence in an application, for annotations that compose nothing and will never be asked for composing constraints.

Narrowing the marker to the annotations that actually compose something is measurably cheaper, and it is where the marker naturally belongs — it is what `RetainStereotypes`' own javadoc shows (`@Username`, not `@Size`). `micronaut-validation` can decide that in `ValidationAnnotationRemapper`, which already reads `annotation.getStereotypes()` to find `@Constraint`: add the marker only when a stereotype is itself constraint-annotated.

There is one consequence worth stating explicitly, because it decides the design:

```
narrow opt-in: Leaf present in tree = true
narrow opt-in: Leaf subtree         = null
```

With the marker only on the composed annotation, the composing annotation **is** in the tree but keeps no subtree of its own, so `@Constraint` is not visible beneath it. A consumer walking the tree cannot tell from the tree alone which composing annotations are constraints — it needs `annotationType.isAnnotationPresent(Constraint.class)`, one cached reflective check per composing annotation type. That is cheap, and `micronaut-validation` already does exactly this kind of check in `ConstraintDefinitions`, so the narrow opt-in looks like the better trade: cost proportional to actual composition, and one reflective predicate on a path that is already reflective.

Worth noting either way that a leaf's retained tree is not empty but `[@Marker]` — the family marker itself is retained. It is not the `@RetainStereotypes` marker (that one is correctly excluded, and the spec asserts it), but under the transformer approach it would be `@Constraint`, which is at least useful provenance rather than noise.

## Two reproductions pushed

I pushed `8a9a8a1` with reproductions for the two gaps I could confirm. Both are `@PendingFeature`, so CI stays green and they fail the moment the behaviour is fixed without the marker being removed.

**1. A transitive override does not cascade** — `StereotypeOriginSpec`. This is the known limitation, but it is not pre-existing for the motivating consumer. `micronaut-validation` cascades *today*: `DefaultConstraintDescriptor` builds each composing descriptor from the already-overridden `AnnotationValue` of its parent, whose constructor re-applies `@OverridesAttribute` one level down. For `@Outer(shortest = 3)` → `@Inner` → `@Size` it yields `Size(min = 3)`; the tree yields `Size(min = 1)`. Constraint composition is covered by the Jakarta Validation TCK, so reading composing *values* off the tree would regress it. I split the nesting assertions that do hold into their own test and left the flat-index assertion alongside the tree one, so a fix moves both together.

**2. The reserved member leaks through `getConvertibleValues()`** — `AnnotationValueSpec`. `convertibleValues` is built in the constructor from the raw values map (`AnnotationValue.java:192`), before `getValues()` can hide the member, so `names()`, `values()`, `asMap()` and iteration all report `$stereotypes` as an attribute:

```
outer.values             = [shortest:3]
outer.memberNames        = [shortest]
outer.convertible        = [$stereotypes:[@Inner(min=3)], shortest:3]
outer.convertible.names  = [$stereotypes, shortest]
```

This is the one hole left in "the member is not an attribute", and it is the view anything walking members generically goes through, since it backs `AnnotationValueResolver`. `equals`/`hashCode`/`toString` and the annotation proxy path all go through `getValues()` and are unaffected — `buildAnnotation` strips `$nonBinding` explicitly and gets `$stereotypes` for free.

---

dstepanov:
## Composition shapes `@OverridesAttribute` needs — all four hold

Pushed `3f24155`. These are **passing coverage**, not defects. The spec exercised a repeatable *composed* annotation applied twice to an element, but not a constraint composing the same constraint twice — which is the shape `AliasFor.index` was added for, since those occurrences are exactly what `OverridesAttribute.constraintIndex` selects between and the flat index cannot address them at all.

| Shape | Result |
|---|---|
| `@Password` composing `@Size` twice, `index = 0` / `index = 1` overrides | each occurrence takes its own override, neither leaks |
| override with no `index` (the `-1` default) | reaches every occurrence, each keeping the member it was not overridden on |
| `@Size.List({@Size(min=1), @Size(max=5)})` written by hand | same tree as the repeated form javac folds |
| a container that is **not** the declared `@Repeatable` container | flattened too, occurrence addressable by index |

The last one I wrote expecting the opposite and corrected against the observed behaviour, and it is the most interesting of the four: `@Sizes({@Size(max = 5)})`, where `Sizes` is not `Size`'s declared container, is still unwrapped into a real occurrence. That happens to be exactly what `micronaut-validation` does at runtime — `ConstraintContainers` resolves the constraint a container holds "whatever its name", and `DefaultConstraintDescriptor.repeatedConstraintAnnotations` unwraps any member returning an array of annotations — so the tree and the reflective implementation agree on the exotic shape too. Worth having pinned, since nothing else asserts it.

## Cases still worth covering, in rough priority

Grouped by what each would prove. Nothing below is filed as a defect; these are the gaps I would want closed before the reflection in `DefaultConstraintDescriptor` is deleted rather than kept as a fallback.

**Sources that have no tree at all.** Each is a path where a tree-based `getComposingConstraints()` returns nothing, verified by reading the code:
- an XML-declared composed constraint — `XmlValidationMetadataProvider.parseConstraints` builds metadata by hand (`addDeclaredAnnotation` plus a synthetic `@Constraint` stereotype), so there is no subtree;
- a composed constraint on a class compiled without the processor — `ReflectionAnnotationMetadataBuilder` populates the stereotype indexes but neither `$stereotypes` nor `@AliasFor`, so attribution *and* override application are both absent;
- mixed sources, where XML `ignore-annotations` overrides one property of a class that otherwise has processor metadata.

**Propagation the tree does not do.** The tree carries structure and `@AliasFor` overrides only; `groups`, `payload`, `validationAppliesTo` and `@ReportAsSingleViolation` are applied per level by `composingConstraint` today and would become the walker's job — worth pinning at depth ≥ 2, where the current recursion gets them right for free.

**Exceptions a compile-time alias cannot raise.** An `@OverridesAttribute → @AliasFor` transformer applies silently, while these are `ConstraintDefinitionException` at runtime today and TCK-visible: overriding a member the composing constraint does not declare, overriding with the wrong type, a `constraintIndex` beyond the occurrence count, and composing constraints sharing no validation target.

**Plumbing the tree has to survive.** Untested and cheap: container-element constraints (`List<@MinimumLength String>`, i.e. type-argument metadata), interface → implementation inheritance rather than just the `AnnotationMetadataHierarchy.merge()` already covered, AOP-intercepted method parameters, and record/constructor parameters.

## One unrelated finding in micronaut-validation#631

While tracing this I found a bug that is independent of this PR: **XML `<constraint-definition>` validators are ignored for composing constraints.** `DefaultValidator.constraintValidatorClasses` consults `metadataProviders`, but `DefaultConstraintDescriptor.composingConstraint` resolves validators through `ConstraintContainers.withValidators`, which reads `constraintType.getAnnotation(Constraint.class).validatedBy()` and never sees a provider — it is a static method with no access to them. Since `validateConstrains` uses `constraint.getConstraintValidatorClasses()` off the descriptor, an XML-configured validator genuinely will not run. A constraint honours its XML definition when applied directly and ignores it when composed. Raising it here only because it turned up in this trace; it belongs on that PR.

---

dstepanov:
## Validated `9e51738` — both reproductions close, one gap remains

Ran `:micronaut-core:test`, `:micronaut-inject:test` and `:micronaut-inject-java:test` against the fix, since `applyIntroducedAliases` is on the path of every annotation the builder processes and the blast radius is wider than retention.

**4986 tests, everything pre-existing green.** The only failures were two new cases I added on top; one of them is a real gap, the other was my assertion being wrong. Pushed as `0ca3160`.

### The cascade stops after one intermediate level

Filed `@PendingFeature`. For `@A(shortest = 7)` over `@B` over `@C` over `@Size`:

```
@B(min = 7)   ✔
@C(min = 7)   ✔
@Size(min = 1)  �’ expected 7
```

The diagnosis is fairly well pinned down by the shape of the failure. `overrideMembers` overrides an occurrence's members *before* re-applying its aliases, and `@C`'s members **are** overridden while its aliases are not re-applied — so the recursion is not stopping on the `stereotypes == null || isEmpty()` guard, which leaves `processAliases` returning early on a null annotation type. The first level is a `ProcessedAnnotation` taken straight out of the declared stereotype list and carries its type; every level below is rebuilt by `toProcessedAnnotation`, which resolves the type by name through `getAnnotationMirror` and yields `null` when that lookup does not resolve:

```java
List<AnnotationValue<?>> overriddenStereotypes = applyIntroducedAliases(
    context,
    stereotypes.stream().map(this::toProcessedAnnotation).toList(),  // type resolved by name, may be null
    cascaded
)
```

Worth fixing rather than documenting: two levels is ordinary for the motivating consumer — a composed constraint composing another composed constraint — and `micronaut-validation` cascades to any depth today, because each composing descriptor re-applies `@OverridesAttribute` from its own already-overridden value. The depth-2 case this PR fixed is the shallowest instance of the same shape.

### The cascade is not tied to retention — passes

Pinned separately, and it is the part of `9e51738` with the widest reach. `applyIntroducedAliases` runs for every annotation, so the uncascaded value was reported by the flat index of annotations that retain nothing at all, and **every existing `@AliasFor` user sees this behaviour change** — not just opted-in consumers. A test that fixes no retention state now asserts the flat index takes the cascaded value, so a regression there is not misread as a retention regression. The broad run above is the evidence that nothing depended on the old behaviour.

I also corrected the javadoc on the cascade reproduction, which still described the defect in the present tense after the fix landed.

### State of the branch

| | |
|---|---|
| `getConvertibleValues()` leak | fixed, test passes |
| transitive override, one level | fixed, test passes |
| transitive override, two levels | **open**, `@PendingFeature` |
| composition shapes (4 cases) | pass |
| cascade outside retention | passes |

Nothing else from the [earlier list](https://github.com/micronaut-projects/micronaut-core/pull/12968#issuecomment-5512569968) has moved — the "no tree at all" paths (XML mapping, the reflection bridge), the propagation cases, and the `ConstraintDefinitionException` cases are all still open, and they only become decidable once you have settled whether `micronaut-validation` keeps the reflective implementation as a fallback.

---

dstepanov:
## Validation-side check: no new pending tests, four shapes covered

Pushed `132c0cb`.

**There is no integration to run yet.** micronaut-validation#631 pins core `5.1.11` through the core BOM and does not reference `@RetainStereotypes` anywhere, so building it against this branch would mean bumping it to a local `5.2.0-SNAPSHOT` *and* writing the opt-in and the `@OverridesAttribute` transformer — that is the migration itself, not a check on it. So I drove the validation-facing element shapes through core's own harness instead.

### The four shapes all hold

The specs drove the tree through a method parameter only. A validation integration also sees constraints on properties, container elements, inherited methods and record components, and each reaches the metadata by a different route — a property merges the field with the accessors, a container element lands on a type argument, an implementation inherits from its interface, a record component feeds both the property and the constructor argument. That matters because the reserved member travels in a values map through all of them, which is exactly where a member that is not an attribute can get dropped or duplicated.

| Shape | Result |
|---|---|
| property metadata (field + accessor merge) | `[@Size(min = 8)]` |
| container element, `List<@MinimumLength String>` | `[@Size(min = 8)]` |
| interface → implementation inheritance | `[@Size(min = 8)]` |
| record component (property and constructor argument) | `[@Size(min = 8)]` |

So: **nothing new to file as `@PendingFeature`.** The only open item remains the two-level cascade from `0ca3160`.

### Two things worth knowing, neither a defect in this PR

**`@Inherited` is load-bearing for the inheritance case, and it lands in the tree.** With the marker alone, the composed annotation does not reach an implementing class at all — only its flattened stereotypes do, so the implementation's metadata had `Size$List` at the right value but no `MinimumLength`. `micronaut-validation`'s `ValidationAnnotationRemapper` already adds `@Inherited` to every constraint, so this is handled in practice, and the fixture carries it. The consequence for the tree is that `@java.lang.annotation.Inherited` shows up as a composing annotation in every constraint's retained subtree:

```
[@java.lang.annotation.Inherited, @jakarta.validation.constraints.Size(min = 8)]
```

`@Documented`, `@Retention`, `@Target` and `@Repeatable` are filtered as `AnnotationUtil.INTERNAL_ANNOTATION_NAMES`, but `@Inherited` is not. A consumer filters it out trivially, and validation filters by `@Constraint` anyway, but it is retained noise at every use site under the family opt-in — arguably `@Inherited` belongs in that list too.

**`AbstractTypeElementSpec`'s local transformers do not reach the introspection path.** I first wrote these against `ValidationRetainedStereotypesSpec`, using the real `@Constraint` / `@OverridesAttribute` annotations and the transformer strategy, and all four "failed". They were not failing on the tree: neither `getLocalAnnotationTransformers` nor `getLocalAnnotationRemappers` is consulted when the source is compiled through `buildBeanIntrospection` / `buildClassLoader`, so the opt-in never happened and the override was never applied — the flat `Size` sat at its declared `min = 5` rather than `8`, and no transformer callback fired. Worth knowing before someone else writes a transformer-based test against an introspection and reads the result as a product bug; if transformer-strategy coverage of introspections is wanted, that harness needs the builder wired into that path too. I moved the tests to `StereotypeOriginSpec` and declared the marker directly, which tests the plumbing question without needing the transformer.

I also chased an apparent duplicate — the inherited flat index showing `[[min: 8], [min: 8]]` — and it is not inheritance or retention. That probe's fixture had `TYPE_USE` in the constraint's `@Target`, so the annotation applied to both the method and the return type use. With retention on and off the count is identical, and `micronaut-validation` already dedups this in `getConstraints0` via `ConstraintAnnotationKey`.

### Still open from the earlier list

The [remaining groups](https://github.com/micronaut-projects/micronaut-core/pull/12968#issuecomment-5512569968) are unchanged and none of them is testable in core: the sources that have no tree (XML mapping, the reflection bridge), `groups`/`payload`/`validationAppliesTo` propagation, and the `ConstraintDefinitionException` cases. All three are `micronaut-validation` behaviour rather than builder behaviour, so they belong on that side once the opt-in exists there.

---

dstepanov:
## One more defect: `mutate()` moves the tree out of the member the writer emits

Pushed `4b335e2` and `46dac2e`.

`AnnotationValueBuilder` seeds its values from `AnnotationValue#getValues()` — which hides the reserved member — and separately copies `getStereotypes()` into the transient `stereotypes` field:

```java
AnnotationValueBuilder(AnnotationValue<T> value, @Nullable RetentionPolicy retentionPolicy) {
    this.values.putAll(value.getValues());                       // the member is hidden here
    this.stereotypes = value.getStereotypes() == null ? null : new ArrayList<>(value.getStereotypes());
}
```

So `mutate()` and `AnnotationValue.builder(value)` move the tree into the field, which is the representation this PR set out to avoid because the writer never emits it:

```
original contains member = true
mutated  contains member = false
mutated  getStereotypes  = [@Size(min=8), @SourceOnly]   ← still reads fine
builder  contains member = false
```

Reads are unaffected, so a rebuilt value looks correct in memory. The existing assertion — `retaining.mutate().member("min", 4).build().getStereotypes() == [size]` — covers exactly that reading path, which is what masks it. It also leaves the two accessors disagreeing: `getStereotypes()` answers while `contains($stereotypes)` does not. Filed `@PendingFeature`.

**I could not demonstrate actual data loss from it, and that is worth stating plainly.** I expected metadata carrying a rebuilt value to write no tree, wrote that test, and it passed unexpectedly: `addDeclaredAnnotation` merges into the entry already present rather than replacing it, so the member the rebuilt value lost is still on the original and survives the writer. I kept it as passing coverage instead — that is the path a `TypeElementVisitor` adjusting an annotation takes, and were the merge to become a replace, every such visitor would silently drop the tree with nothing failing until the metadata was read back. So the finding is a broken invariant with no reachable victim in core today, not a live bug; it will bite the first consumer that rebuilds a value and uses it as the sole source.

## Checked and found sound — not filed

- **AOP-intercepted method arguments** — the tree reaches the intercepted method's argument metadata intact, member and all.
- **`AnnotationMetadataHierarchy` read without `merge()`** — `getAnnotation` and `getAnnotationValuesByName` both resolve through the hierarchy with the tree present.
- **Nested `SOURCE`-retention filtering** — an opted-in composing annotation that itself composes a `SOURCE` annotation has its subtree correctly emptied and the member dropped, so the depth-1 retention test was not hiding a depth-2 leak.
- **Groovy cascade parity** — the cascade is in the shared builder but each frontend resolves annotation types by name differently, and that resolution is what carries the override down. The Groovy parity spec covered attribution only; cascade now asserted, and it passes. Added as `46dac2e`.

## Suite status

`:micronaut-inject-groovy:test` is green on this branch. **`:micronaut-inject-kotlin:test` I could not run** — Maven Central returns 429 for `kotest-runner-junit5/6` and `micronaut-test-kotest5` through this environment's proxy, on repeated attempts. It is the one language frontend unvalidated since the cascade fix, and given the fix's depth limit traces to per-frontend mirror resolution, it is worth someone running locally.

## Open items

| | |
|---|---|
| two-level cascade (`0ca3160`) | open, `@PendingFeature` |
| `mutate()` drops the persisted member (`4b335e2`) | open, `@PendingFeature` |
| Kotlin suite | unvalidated here (proxy 429) |

Everything else from the earlier lists is either passing coverage on this branch or belongs to `micronaut-validation` once the opt-in exists there.

---

dstepanov:
## Validated `ccfad5c` — green, and the seeded-builder contract now has tests

Pushed `19393b9`.

| module | tests | failures | errors |
|---|---|---|---|
| `micronaut-core` | 7387 | 0 | 0 |
| `micronaut-inject` | 312 | 0 | 0 |
| `micronaut-inject-java` | 4991 | 0 | 0 |
| `micronaut-inject-groovy` | 1066 | 0 | 0 |

The `mutate()` reproduction passes with its `@PendingFeature` removed, and nothing regressed. **My stated cause for the deeper cascade was wrong** — thank you for instrumenting it. The type resolving at every level and the cascade computing `@C(min = 7)` carrying `@Size(min = 7)`, only for a later pass to rebuild `@B`'s subtree from a pre-override `@C` and win the last write, is a different defect from the one I described, and a more interesting one: the fix is ordering, not recursion.

### Two halves of the contract the fix states but does not assert

`ccfad5c`'s message makes two claims that nothing tested. Both now do, and both pass:

- **A `stereotype()` call on a seeded builder reaches the member, not only the field.** This is the shape an integration takes to opt an annotation in — `micronaut-validation`'s `ValidationAnnotationRemapper` does literally `annotation.mutate().stereotype(...).build()` to add `@Inherited`. Had the addition landed only in the transient field, the rebuilt value would write the tree it was seeded with and silently drop the addition, which is the same class of bug the fix just closed.
- **An unseeded builder gains no reserved member.** This is the mapping protocol the fix was careful to preserve: on a builder not seeded from a retaining value, the field's null and empty states mean "fill the stereotypes from the annotation definition" and "skip". Writing the member there would turn a mapping instruction into retained state. Worth a test precisely because the guard is a single `retainedStereotypes &&`, and nothing else would notice if it were dropped.

### Kotlin still unrun

`:micronaut-inject-kotlin:test` remains blocked in this environment — Maven Central returns 429 through the proxy, now for `symbol-processing-aa-embeddable` as well as the kotest artifacts, across four attempts spread over the session. It is the one frontend never exercised since `9e51738`, and since the cascade defect is about *when* subtrees get rebuilt during processing, a frontend with its own processing order is exactly where it could differ. Worth one local run before merge.

### State

| | |
|---|---|
| two-level cascade | open, `@PendingFeature`, cause now correctly identified |
| `mutate()` dropping the persisted member | fixed, test passes |
| `getConvertibleValues()` leak | fixed, test passes |
| one-level cascade, non-retention cascade | fixed, tests pass |
| composition shapes, element shapes, Groovy parity, seeded-builder contract | passing coverage |
| Kotlin suite | still unrun here |

---

dstepanov:
## Recheck on `19393b9` — CI fully green, one thing to fix in the description

No new commits since my last push; head is `19393b9`.

**All 14 checks pass**, and the branch is `mergeable_state: clean`. SonarCloud has gone from the 1 new critical issue it reported earlier to a pass. GraalVM native tests, Python CI, the build matrix and `license/cla` are all green.

**This closes the Kotlin gap I flagged.** I could never run `:micronaut-inject-kotlin:test` here — Maven Central kept returning 429 through this environment's proxy across five attempts. CI runs `./gradlew check jacocoReport --no-daemon --continue` at the root and `inject-kotlin` is in `settings.gradle`, so `build (25)` exercised it on this exact SHA and passed. Nothing outstanding on that front, and the description's "Green: … `:micronaut-inject-kotlin:test`" is now backed by CI rather than only by a local run.

### The description no longer mentions the open limitation

The rewrite replaced the old "Known limitation, pre-existing" section with "Transitive overrides now cascade". That section is accurate for what it describes — the one-level cascade is fixed and both views agree — but the two-level case is still open, and the branch still carries the test saying so:

```groovy
@PendingFeature(reason = "a later processing pass overwrites the cascaded subtree with one computed before the override")
void "a transitive override cascades through more than one intermediate annotation"() {
```

As it reads now, a reviewer would take cascading as fully solved. It is worth a short paragraph in that section: `@A` over `@B` over `@C` over `@Size` reaches `@B(min = 7)` and `@C(min = 7)` but leaves `@Size(min = 1)`, because a later processing pass rebuilds `@B`'s subtree from a `@C` computed before the override and the last write wins. That is a real constraint on the motivating consumer — a composed constraint composing another composed constraint is ordinary in validation, and `micronaut-validation` cascades to any depth today by re-applying `@OverridesAttribute` from each already-overridden parent. Reading composing *values* off the tree is safe at one level and not yet at two.

The test list in the description is also now a little behind the branch: it does not mention the element shapes a validation integration sees (property metadata, container elements, interface inheritance, record components), the seeded-builder contract in `AnnotationValueSpec`, or the Groovy cascade parity spec.

### Everything else stands

| | |
|---|---|
| `getConvertibleValues()` leak | fixed, test passes |
| one-level cascade, cascade outside retention | fixed, tests pass |
| `mutate()` dropping the persisted member | fixed, tests pass |
| two-level cascade | **open**, `@PendingFeature` |
| composition shapes, element shapes, seeded-builder contract, Groovy parity | passing coverage |
| Kotlin suite | green in CI |

---

dstepanov:
## Recheck on `b007ca0` — all reproductions closed, one risk in the new fallback

Validated locally: `micronaut-core` 7389, `micronaut-inject` 312, `micronaut-inject-java` 4991, `micronaut-inject-groovy` 1066 — **zero failures, and no `@PendingFeature` left in the branch.** Every defect filed in this thread is now closed. Pushed `d593693`, a doc correction only.

My depth-cascade diagnosis was wrong twice and `f2353f4` got it right: it is the name lookup after all, but the reason it fails is that the test harness builds with a parser that never entered the source, which is the same condition the builder already notes for KSP and Groovy. I had discarded that explanation on the strength of the instrumentation in `ccfad5c`. The comment on that test still carried the disproved version, so `d593693` rewrites it to the requirement and the real cause.

## The fallback map is not cleared across a KSP round

`f2353f4` adds an instance field on `AbstractAnnotationMetadataBuilder`:

```java
private final Map<String, T> processedAnnotationTypes = new HashMap<>();
```

keyed on the annotation name, holding native elements, populated from every `ProcessedAnnotation` constructed. On javac that is harmless. On KSP I think it is not, and `KotlinVisitorContext.updateResolver` says why in its own words:

```kotlin
// A new resolver means a new round and a new Analysis API session, and this context is
// reused across that boundary rather than rebuilt. Every memo below holds symbols from the
// old session, so they must go: reading one afterwards throws
// KaInvalidLifetimeOwnerAccessException ("PSI has changed since creation"). The
// node-keyed memos would merely miss, since the nodes are new objects, but classByNameCache
// is keyed on a String and would hit and hand back a dead declaration.
binaryNameCache.clear()
annotationTypeCache.clear()
repeatableContainerCache.clear()
jvmNameCache.clear()
classByNameCache.clear()
this.resolver = resolver
annotationMetadataBuilder.resolver = resolver
```

The new map is exactly the shape that comment warns about — String-keyed, holding symbols — and it lives on `annotationMetadataBuilder`, whose `resolver` is reassigned two lines below, but it is not in the clearing list. `TypeElementSymbolProcessor` builds the context once and calls `updateResolver` on every subsequent round rather than rebuilding, so the map survives the session boundary and can hand back a dead `KSDeclaration` on a later round. `BeanDefinitionProcessor` constructs a fresh context per round and is unaffected.

The fix looks like one line — clear it alongside the others in `updateResolver`, which needs a small hook on the builder since the field is private. Worth also deciding whether it should hang off `clearCaches()` for symmetry with `ANNOTATION_DEFAULTS`; as an instance field it dies with the builder on javac, so that is tidiness rather than correctness there.

**I cannot test this here** — `:micronaut-inject-kotlin:test` still will not resolve through this environment's proxy (429 from Maven Central, six attempts now), and it would need multi-round processing to trigger regardless. CI on `b007ca0` was still queued/in-progress when I checked, so it has not answered either. Flagging it as a reasoned risk with the evidence above, not a demonstrated failure — it deserves a look from someone who can run KSP locally, since this fallback exists specifically for the frontends where the name lookup fails.

## State

| | |
|---|---|
| `getConvertibleValues()` leak | fixed |
| one-level cascade, cascade outside retention | fixed |
| `mutate()` dropping the persisted member | fixed |
| two-level cascade | fixed in `f2353f4` |
| `processedAnnotationTypes` stale across a KSP round | **open risk, untested** |
| composition shapes, element shapes, seeded-builder contract, Groovy parity | passing coverage |

---

dstepanov:
## Review of `4a4bf58`

Green on everything that resolves here: `micronaut-core`, `micronaut-inject`, `micronaut-context`, `micronaut-aop`, `micronaut-inject-java` (4992), `micronaut-inject-groovy` (1066), zero failures. `:test-suite:test` and `:micronaut-inject-kotlin:test` are both blocked by Maven Central 429s in this environment (`micronaut-jdbc`, kotest/KSP), so neither ran. CI's `build (25)` was still in progress when I checked.

The rename and inversion are right, and they close the cost concern properly: the marker on the composing family, cost proportional to the retainable occurrences instead of the whole subtree. Better than the narrowing I suggested, because the framework annotation that defines the contract is the thing that knows.

Three things below. The first is the one I would not merge without.

### 1. The KSP stale-symbol risk got worse, not addressed

`processedAnnotationTypes` is still never cleared, and this commit adds a **second** consumer on a much hotter path:

```java
private boolean isRetainable(AnnotationValue<?> annotationValue) {
    List<AnnotationValue<?>> stereotypes = annotationValue.getStereotypes();
    if (stereotypes == null) {
        String annotationName = annotationValue.getAnnotationName();
        T annotationType = getAnnotationMirror(annotationName)
            .orElseGet(() -> processedAnnotationTypes.get(annotationName));
        return annotationType != null && lookupOrBuildForType(annotationType).hasStereotype(RETAINABLE);
    }
```

Previously the fallback was reached only while cascading an override. Now it is on the retention decision itself, for every occurrence whose stereotypes were not computed — and it does not merely read the element, it calls `lookupOrBuildForType` on it. `KotlinVisitorContext.updateResolver` states the hazard in its own comment: the context is reused across a round rather than rebuilt, symbols from the old session throw `KaInvalidLifetimeOwnerAccessException`, and a **String-keyed** cache "would hit and hand back a dead declaration". It clears five caches for exactly this and then reassigns `annotationMetadataBuilder.resolver` two lines later; this map is on that builder and is not among them. `TypeElementSymbolProcessor` reuses the context across rounds, `BeanDefinitionProcessor` does not.

Still untested — I cannot run KSP here, and it needs multi-round processing to bite — but the surface is wider now and the fallback exists precisely for the frontends where the name lookup fails.

### 2. Interceptor binding resolution is a much larger blast radius than the rest of the PR

`InterceptorBindingQualifier` (+230/-…), `DefaultInterceptorRegistry`, `InterceptorBindingMembers` and `@InterceptorBinding` itself are runtime DI/AOP resolution, not annotation-metadata plumbing. Two claims in the message carry real risk and I could not exercise either, because `test-suite` is exactly the coverage that did not run:

- **"Bindings written by earlier versions keep their copy and are read as before."** `resolveBoundOccurrences` prefers `$bindingValues` when present, so a mixed classpath — a library compiled on 5.1 against an application compiled on 5.2 — ends up comparing a legacy copy on one side against a retained occurrence on the other. That looks like it should hold, since both are `AnnotationValue`s of the binding annotation and `equals` goes through `getValues()`, but it is a compatibility path that nothing in the diff pins. A test with a hand-built `$bindingValues` binding on one side and a retained occurrence on the other would settle it cheaply.
- **"it now matches members wherever it has the metadata, which only removes interceptors the registry would have rejected."** This narrows qualification at injection time. If that equivalence is off in any case, an interceptor silently stops applying — the worst failure mode in this area, since nothing errors.

`ComposedBindingMembersSpec` is a good addition, but it is compile-time metadata. Worth confirming `:test-suite:test` and the AOP TCK suites are green before this lands, since I could not.

### 3. `@since` on `InterceptorBindingQualifier` is now wrong

```diff
- * @since 2.4.0
+ * @since 3.0.0
```

`@since` records when the type was introduced, not when it was last touched. The class has existed since 2.4.0; this states it appeared in 3.0.0. Unrelated to the change and worth reverting — happy to push that one-liner if you want it.

---

dstepanov:
## Two gaps filled in `BindingOccurrenceResolutionSpec`, and the KSP fix is now #12985

Pushed `f89c869`. `:micronaut-inject:test` is green at 320.

I had written a separate spec for `resolveBoundOccurrences` before `d30e92e` landed and four of its six cases duplicated yours, so I dropped it and added only what was missing to your spec:

- **A binding that binds members with nothing to read them from resolves to none** — no metadata at all, which is the shape a qualifier built from a plain collection of bindings has, and metadata carrying no occurrence of the binding annotation. Both return null rather than throwing.
- **The declared preference**, the `declaredOverrides` branch, which nothing exercised.

The second is worth a note, because my first attempt at it asserted the wrong thing and passed for the wrong reason. With a plain annotation declared on both levels, the hierarchy already answers `findAnnotation` with the declared one, so resolving with and without the preference agree and the flag proves nothing. The branch only shows where the inherited occurrence is **composed**: with the class carrying the binding annotation through another annotation and the method declaring it directly, resolving with the preference binds only what the method declares, and resolving without it binds the inherited occurrence too. That is now what the test sets up.

Still not covered here, and I could not run it: the injection-time narrowing. `:test-suite:test` is blocked by a Maven Central 429 on `micronaut-jdbc` in my environment, so the claim that matching members at injection time "only removes interceptors the registry would have rejected" is still resting on reasoning rather than a green suite. That one wants a runtime test — an interceptor that qualifies but whose members do not match, asserted to still be applied or not applied as intended — rather than another unit test on the qualifier.

## The KSP clearing is up as #12985

Raised as micronaut-projects/micronaut-core#12985 rather than pushed here, since it is a distinct change and you may prefer to fold it in. It targets this branch, because the map it fixes is introduced here.

It adds an `@Internal` `clearProcessedAnnotationTypes()` on the builder and calls it from `KotlinVisitorContext.updateResolver` alongside the five memos already cleared there. `javac` keeps one builder for the whole task but its elements stay valid across rounds, and Groovy builds a context per source unit, so neither needs it.

`:micronaut-core-processor:compileJava`, `:micronaut-inject-kotlin:compileKotlin` and checkstyle pass. The Kotlin **test** dependencies are what 429s, not the main compile, so I still could not run that suite — and the defect needs multi-round processing, which I do not think the current specs reach, so a green run would not have settled it either way.

---

dstepanov:
## The KSP clearing is folded in

#12985 is merged into this branch as `b5649c9` and closed. Squashed rather than rebased — the repository does not allow rebase merges — so the branch keeps its linear history and the PR is recorded as merged rather than abandoned.

What landed: an `@Internal` `clearProcessedAnnotationTypes()` on `AbstractAnnotationMetadataBuilder`, called from `KotlinVisitorContext.updateResolver` alongside the five memos already cleared there, so the String-keyed map of native elements does not survive a KSP round boundary and hand back a symbol of a session that has ended. javac keeps one builder for the whole task but its elements stay valid across rounds, and Groovy builds a context per source unit, so neither needs the call; the method is documented for any future implementation that reuses a builder across rounds.

Validated on the merged head: `:micronaut-core-processor:compileJava`, `:micronaut-core-processor:checkstyleMain`, `:micronaut-inject-kotlin:compileKotlin`, `micronaut-core`, `micronaut-inject` (320) and `micronaut-inject-java` (4992) all pass.

The caveat from that PR carries over unchanged: **the Kotlin test suite still has not run.** Maven Central 429s its test dependencies in my environment — the main compile succeeds, so this is not a code problem — and the defect needs multi-round processing, which I do not believe the current specs reach, so a green run would not have confirmed the fix either way. It wants a local KSP run, and ideally a spec that processes more than one round.

Two other things from the earlier reviews are still open and unaddressed:

- the injection-time narrowing in `InterceptorBindingQualifier` — "only removes interceptors the registry would have rejected" — which `:test-suite:test` would cover and which 429s here on `micronaut-jdbc`;
- `@since 2.4.0` → `@since 3.0.0` on `InterceptorBindingQualifier`, which misstates when the class was introduced.

---

graemerocher:
I am concerned how this will work with already compiled code that was compiled with an older version of Micronaut. We don't want to have to recompile all the modules again. What is the binary backwards compatibility story?

dstepanov:
Nothing has to be recompiled. Both halves of that — the API and the generated metadata — are covered, and I re-ran everything below against the current head `b5649c9`.

**API.** Nothing was removed. `japicmp` is green on `core`, `inject`, `aop` and `core-processor`. Every addition is additive: the `@Retainable` marker, the `AnnotationUtil.STEREOTYPES_MEMBER` constant, and two new methods on `InterceptorBindingQualifier`. `META_BINDING_VALUES` and `bindingValuesMatch` are both still there with the same signatures, so anything compiled against them still links.

**Metadata.** The compiler stops writing the `$bindingValues` copy, but the runtime still reads it wherever it is present, so previously compiled bindings resolve exactly as before. `resolveBoundOccurrences` takes the copy first and only falls back to looking for the occurrence on the element.

I checked that by compiling with the **real 5.1.13 annotation processor** out of the Gradle cache, confirming the generated metadata contains `$bindingValues`, and running it with the `core`, `inject`, `aop`, `context` and `core-reactive` jars from this branch swapped in:

| scenario | result |
| --- | --- |
| app compiled by 5.1.13, run on this branch | member binding, a defaulted binding member, a method-level binding overriding a class-level one, and a name-only binding all behave identically to the 5.2.x baseline |
| 5.1-compiled interceptor library + newly compiled beans | passes, including an annotation composing the binding annotation |
| newly compiled interceptors + 5.1-compiled beans | passes, same shapes |

The two mixed rows are the case you are really asking about, since an application and its dependencies are not recompiled together. A binding read from a 5.1 copy and one read from a newly compiled element match each other.

For the record, the one place 5.1.13 and this branch differ is a bean leaving a binding member at its default, and that is #12925, already on `5.2.x` before this PR. Against the `5.2.x` baseline the results are identical.

**micronaut-data 5.1.3**, built against core 5.1.10, as a real precompiled consumer:

- it uses `bindMembers` nowhere, so none of its metadata ever carried the copy — its advice (`@Repository`, `@Connectable`) binds by name;
- its 109 bean definitions and all four interceptor bindings resolve byte-for-byte identically on the 5.1.13 runtime and on this branch;
- a context booted with its jars starts the same on both: 89 definitions, 4 with bindings, interceptor registry resolvable.

**In CI**, `BindingOccurrenceResolutionSpec` pins the legacy shape directly: a binding read from its copy, one read from the element, an occurrence found through the retained tree, non-binding members excluded, name-only comparison, and a copy-shaped binding matching an element-shaped one.

The one thing that does change for code that *is* recompiled: an AOP use site gains the retained `@InterceptorBinding` occurrence and loses the copy. That is a smaller payload, since the copy duplicated the annotation's members at every use site, and it is what lets two annotations composing the same binding annotation with different members be told apart — which the name-keyed index cannot do at all.

Happy to run the harness against any other precompiled module you would like to see, or to fold it into the build as a test that compiles with a released processor and runs on the current runtime.

graemerocher (approved):

PR #12971, opened 2026-09-02
Key a parameter's and a field's mutated metadata by the declaring type, not the type it is read through

dstepanov:
## The problem

A `TypeElementVisitor` that mutates the annotations of a method parameter or a field (`annotate(...)` / `removeAnnotation(...)`) does so through the `MUTATED_ANNOTATION_METADATA` cache in `AbstractAnnotationMetadataBuilder`. That cache keyed a parameter's metadata as `(owningType, method, parameter)` and a field's as `(owningType, field)`, where `owningType` is the class the member was *reached through*, not the class that declares it.

An inherited member is the same member whichever class it is reached through, and its annotation hierarchy never includes the owner (a parameter's hierarchy is the overridden parameters plus the variable; a field's is the field alone). So a mutation made through the declaring class landed in a different cache entry from the one a later read through a subclass consulted, and the subclass silently saw the *unmutated* metadata.

Probe (now the regression test): `Base` declares `void inherited(@Marker String a)` and a field `@Marker String f`; `Sub extends Base` inherits both. A visitor on `Base` removes `@Marker` and adds `@Added` on the parameter and the field. A second visitor on `Sub` reads the same parameter and field through `Sub`. On 5.2.x it still saw `@Marker` and never saw `@Added`:

```
Condition not satisfied:

SubVisitor.parameterAnnotations.contains('test.Added')
|          |                    |
|          [test.Marker]        false
```

Both the removal and the addition are lost across owners, and the compiled `BeanDefinition` of `Sub` (whose injection points and executable methods are read through `Sub`) is written with the stale metadata.

## The change

The first component of the parameter and field cache keys becomes the member's **declaring** type instead of its owning type:

- `AbstractAnnotationMetadataBuilder.lookupOrBuildForParameter(T declaringType, T methodElement, T parameterElement)` — key `Key3(declaringType, method, parameter)`.
- `AbstractAnnotationMetadataBuilder.lookupOrBuildForField(T declaringType, T element)` — key `Key2(declaringType, field)`.
- `AbstractElementAnnotationMetadataFactory.lookupForParameter` / `lookupForField` pass `getMethodElement().getDeclaringType()` / `getDeclaringType()` instead of `getOwningType()`.
- `KotlinElementAnnotationMetadataFactory` builds its own keys rather than calling these methods and had the identical owner-keyed `Key3`/`Key2` for parameters and fields; it now keys on `declaringType` the same way, so the three languages behave alike.

The method signatures are unchanged in arity (only the parameter is renamed), so `GroovyAnnotationMetadataBuilder`'s override, the Groovy test helper and `PythonElementAnnotationMetadataFactory` (which already passed the declaring type) keep compiling; the Groovy override's parameter is renamed to match.

Why the declaring type rather than dropping the type from the key altogether: `AbstractAnnotationMetadataBuilder.clearMutated(Object)` walks every `Iterable` key and drops the entries that contain a class being cleared. `TypeElementVisitorProcessor`, `MixinVisitorProcessor` and `BeanDefinitionInjectProcessor` call it with a postponed class's native element to discard the partial mutations made to that class before it was postponed. Keeping the declaring type in the key preserves that: clearing `Base` still drops the entries for `Base`'s parameters and fields, exactly as before. For a member read through its own declaring class the key is byte-for-byte what it was, so nothing changes for the non-inherited case.

In Java the class component is the bare `TypeElement` (`JavaElementAnnotationMetadataFactory.getNativeElement` unwraps `JavaNativeElement.Class` to its element), in Groovy the `ClassNode`, and in Kotlin `getClassDefinitionCacheKey` strips the type, so the declaring-type component is stable however the declaring class was resolved (raw, or parameterised through the subclass).

## What deliberately does not change

`lookupOrBuildForMethod(T owningType, T element)` keeps its owning-type key. A method's metadata is where class-level annotations are merged in, so the same inherited method legitimately has different metadata when reached through different classes; that split is intentional and is untouched here.

Type-level per-use-site isolation (a type annotated at one parameter site not leaking to every other mention of that type) is unaffected: it lives in the type native elements, not in the parameter/field keys.

## Test

`inject-java/src/test/groovy/io/micronaut/visitors/InheritedMemberAnnotationMutationSpec.groovy` registers two local visitors: `BaseVisitor` (mutates the parameter and field through `Base`) and `SubVisitor` (records what the same members look like through `Sub`, including that `owningType` is `Sub`). The spec asserts `@Added` is present and `@Marker` absent both in the visitor's view through `Sub` and in the compiled `BeanDefinition` of `Sub` (executable-method argument and injected field). It fails on unmodified 5.2.x with the output quoted above and passes with this change.

Compile-time only; no runtime code is touched.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
@@ -283,15 +283,19 @@ public abstract class AbstractAnnotationMetadataBuilder<T, A> {
     }
 
     /**
-     * Build the metadata for the given element. If the element is a method the class metadata will be included.
+     * Build the metadata for the given parameter element. The metadata is keyed by the type that <em>declares</em>
+     * the method, not by the type the method is reached through: a parameter of an inherited method is the same
+     * parameter whether it is read through the declaring class or through a subclass, so a mutation made through
+     * one must be visible through the other. Unlike a method, a parameter's metadata never includes the owning
+     * type's annotations, so nothing distinguishes it per owner.
      *
-     * @param owningType       The owning type
+     * @param declaringType    The type declaring the method
      * @param methodElement    The method element
      * @param parameterElement The parameter element
      * @return The {@link AnnotationMetadata}
      */
-    public CachedAnnotationMetadata lookupOrBuildForParameter(T owningType, T methodElement, T parameterElement) {
-        return lookupOrBuild(new Key3<>(owningType, methodElement, parameterElement), parameterElement);
+    public CachedAnnotationMetadata lookupOrBuildForParameter(T declaringType, T methodElement, T parameterElement) {
+        return lookupOrBuild(new Key3<>(declaringType, methodElement, parameterElement), parameterElement);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


graemerocher:
need to check the Python implementation and see if the same change is needed there

graemerocher (approved):

PR #12982, opened 2026-09-03
Refactor CORS handling into injectable components

sdelamo:
Extract origin configuration lookup, response decoration, and preflight validation from CorsFilter into dedicated components.

Add a configuration option to disable the CORS filter entirely, preserve deprecated extension points for compatibility, and expand coverage for CORS utilities, preflight validation, and filter configuration.

Related to: https://github.com/micronaut-projects/micronaut-security/pull/2271


http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -173,17 +186,103 @@ public final HttpResponse<?> filterRequest(HttpRequest<?> request) {
...
+        return CORS_FILTER_ORDER;
+    }
+
+    /**
+     * @param config   The {@link CorsOriginConfiguration} instance
+     * @param response The {@link MutableHttpResponse} object
+     * @deprecated Unused. Use {@link CorsResponseDecorator} instead.
+     */
+    @Deprecated(forRemoval = true, since = "5.2.0")
+    protected void setAllowCredentials(CorsOriginConfiguration config, MutableHttpResponse<?> response) {
```
dstepanov, line 250: These eight `protected` methods (`setAllowCredentials`, `setAllowPrivateNetwork`, `setExposeHeaders`, `setVary`, `setOrigin`, `setAllowMethods`, `setAllowHeaders`, `setMaxAge`) are now never called by `CorsFilter` and their bodies throw `UnsupportedOperationException`. Since `filterRequest`/`filterResponse` are `final`, overriding these was the *only* extension mechanism the class offered.

Concretely: an app that extends `CorsFilter` and overrides `setOrigin(String, MutableHttpResponse)` to emit a fixed `Access-Control-Allow-Origin` instead of echoing the request origin will silently go back to echoing the caller's `Origin` — no compile error, no warning, and it's security-relevant. A subclass that overrides and calls `super.setOrigin(...)` instead gets an `UnsupportedOperationException` on every CORS response.

`setAllowPrivateNetwork` is `@since 4.3.0` and the rest predate that, so this is a break in released API rather than in something new here. Could these keep working implementations that delegate to the injected `CorsResponseDecorator`, so the deprecation is source-compatible and the runtime behaviour is preserved?

dstepanov (commented):
Reviewed the extraction against the original `CorsFilter` logic method by method — `DefaultCorsOriginConfigurationRetriever`, `DefaultPreflightRequestValidator` and `DefaultCorsResponseDecorator` reproduce the previous behaviour faithfully, including the route-level → server-wide fall-through and the `!corsConfiguration.isEnabled()` short-circuit. The two deliberate behaviour changes both look like fixes: `(AnnotationMetadata) rm` → `rm.getAnnotationMetadata()`, and gating `Access-Control-Allow-Private-Network` on `.filter(Boolean.TRUE::equals)` rather than header presence — which is what makes the TCK change meaningful, since that assertion was previously vacuous.

Three things worth a look, inline. Two non-blocking notes as well:

- The new public SPIs (`CorsResponseDecorator`, `PreflightRequestValidator`, `CorsOriginConfigurationRetriever`) point `@DefaultImplementation` at package-private `@Internal` classes, so a consumer can replace them but cannot subclass or delegate to the defaults — the `protected` methods on `DefaultCorsResponseDecorator` are unreachable from outside the package.
- `micronaut.server.cors.filter.enabled` gets no entry in `src/main/docs/guide/httpServer/serverConfiguration/cors.adoc`, which is where the distinction from `cors.enabled` most needs explaining.

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -46,71 +43,87 @@
...
  *
+ * <p>The filter is enabled by default. Set {@code micronaut.server.cors.filter.enabled}
+ * to {@code false} to disable all CORS request validation and response decoration, including
+ * handling configured through {@link CrossOrigin}.</p>
+ *
  * @author James Kleeh
  * @author Graeme Rocher
  * @since 1.0
  */
+@Requires(property = HttpServerConfiguration.PREFIX + ".cors.filter.enabled", value = StringUtils.TRUE, defaultValue = StringUtils.TRUE)
```
dstepanov, line 73: Gating the whole bean on `micronaut.server.cors.filter.enabled` also disables `Cross-Origin-Embedder-Policy` and `Cross-Origin-Resource-Policy`, since `setCrossOriginEmbedderPolicy`/`setCrossOriginResourcePolicy` only run inside `filterResponse`.

Those two are configured independently (`micronaut.server.cors.cross-origin-embedder-policy` / `...-resource-policy`) and aren't CORS request handling — they're response hardening that applies with no `Origin` header at all, which is exactly why `isEnabled(request)` returns `true` for them. So an app that sets `cross-origin-resource-policy: same-origin` and separately sets `filter.enabled: false` (CORS terminated at a gateway) silently loses both headers from every response.

The javadoc on `CorsFilterConfiguration` promises only that "CORS request validation and response decoration" is disabled, so this is a surprise. Worth either moving those two headers into an always-registered filter, or calling the interaction out explicitly in the javadoc and the config docs.

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -904,10 +904,32 @@ public static class CorsConfiguration implements Toggleable {
...
+         */
+        public CorsFilterConfiguration getFilter() {
+            return filter;
+        }
+
+        /**
+         * Sets the CORS filter configuration.
+         *
+         * @param filter The CORS filter configuration
+         * @since 5.4.0
```
dstepanov: `@since 5.4.0` here, but `getFilter()` just above says `@since 5.2.0`, and `projectVersion` is `5.2.0-SNAPSHOT`. Same tag on `CorsFilterConfiguration` itself and on `CrossOriginUtil.getCorsOriginConfigurationForAnnotationMetadataProvider`, while everything else new in the PR (the three interfaces, `CorsFilter`'s deprecations, the new `CrossOriginUtil` methods) is tagged 5.2.0. All four should read 5.2.0 — otherwise the published javadoc tells 5.2 users the property isn't available to them.

graemerocher:
I agree this is a breaking change and can't go into a minor release

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
sdelamo: I have created this PR with new APIs 

https://github.com/micronaut-projects/micronaut-core/pull/12992

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
sdelamo: resolved via  https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/1f3bf325ec1cd5b50cfaf231ec7ed828cf128738

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/6cc736871fa08394467efbe0875a255dc5b02190

sdelamo:
@graemerocher the PR should not longer be a breaking change.

http-server/src/main/java/io/micronaut/http/server/cors/DefaultCorsResponseDecorator.java
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * <p>Values supplied by a validated preflight request are echoed where required by the CORS
+ * protocol. Response headers with multiple values are written either as one comma-separated
+ * header or as repeated header fields according to the server CORS configuration.</p>
+ *
+ * @since 5.2.0
+ */
+@Singleton
+@Internal
+class DefaultCorsResponseDecorator implements CorsResponseDecorator {
```
graemerocher, line 56: can it be `final`? the methods that are not overridden don't need to be public

http-server/src/main/java/io/micronaut/http/server/cors/DefaultPreflightRequestValidator.java
```diff
@@ -0,0 +1,115 @@
...
+ * <p>The requested method must be permitted by the resolved CORS configuration and match an
+ * application route. When no route matches, a resolved static resource is treated as supporting
+ * {@link HttpMethod#GET}. Requested headers and private-network access are then checked against
+ * the same configuration.</p>
+ *
+ * @since 5.2.0
+ */
+@Singleton
+@Internal
+class DefaultPreflightRequestValidator implements PreflightRequestValidator {
```
graemerocher, line 47: can it be `final`?

http-server/src/main/java/io/micronaut/http/server/HttpServerConfiguration.java
```diff
@@ -1016,6 +1038,45 @@ public void setCrossOriginEmbedderPolicy(@Nullable CrossOriginEmbedderPolicy cro
...
+         * Configuration for enabling or disabling the CORS server filter.
+         *
+         * <p>Disabling the filter disables all server-side CORS processing, including rules
+         * declared with {@link io.micronaut.http.server.cors.CrossOrigin}. This differs from
+         * {@link CorsConfiguration#isEnabled()}, which controls server-wide CORS
+         * configurations but still permits route-level CORS annotations.</p>
+         *
+         * @since 5.2.0
+         */
+        @ConfigurationProperties("filter")
```
graemerocher, line 1052: why is the namespace here just "filter" and not "cors-filter"?

http-server/src/main/java/io/micronaut/http/server/cors/DefaultCorsOriginConfigurationRetriever.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public @Nullable CorsOriginConfiguration findCorsOriginConfiguration(@Nullable String origin,
+                                                                         List<? extends RouteMatch<?>> routeMatches) {
+        if (origin == null) {
+            return null;
+        }
+        for (RouteMatch<?> routeMatch : routeMatches) {
+            Optional<CorsOriginConfiguration> originConfigurationOptional =
```
graemerocher, line 57: optional adds overhead and this looks like some kind of internal API call so seems unnecessary

http-server/src/main/java/io/micronaut/http/server/cors/CrossOriginUtil.java
```diff
@@ -16,48 +16,228 @@
...
- * Utility classes to work with {@link CrossOrigin}.
+ * Utilities for resolving and validating CORS settings declared with {@link CrossOrigin}.
+ *
+ * <p>This class converts annotation metadata into {@link CorsOriginConfiguration} instances
+ * and provides the matching operations shared by the server CORS infrastructure.</p>
+ *
  * @author Sergio del Amo
  * @since 3.9.0
  */
 public final class CrossOriginUtil {
```
graemerocher, line 48: review this class and see what needs to be internal / non-public

adding lots of what seem to be internal focused methods to the public API surface doesn't seem wise

http-server/src/main/java/io/micronaut/http/server/cors/CrossOriginUtil.java
```diff
@@ -16,48 +16,228 @@
...
+    /**
+     * Conversion context for the HTTP method supplied by a CORS preflight request.
+     *
+     * @since 5.2.0
+     */
+    public static final ArgumentConversionContext<HttpMethod> CONVERSION_CONTEXT_HTTP_METHOD = ImmutableArgumentConversionContext.of(HttpMethod.class);
```
graemerocher, line 54: does this need to be public? Why?

graemerocher (changes requested):

http-server/src/main/java/io/micronaut/http/server/cors/DefaultPreflightRequestValidator.java
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/955ed47de1edbb656118cb0581a6692a9e7ca327 and https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/48a73d5e0fb2e52e59a06c9b1c45954b306d0dee

http-server/src/main/java/io/micronaut/http/server/cors/DefaultCorsResponseDecorator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/955ed47de1edbb656118cb0581a6692a9e7ca327 and https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/48a73d5e0fb2e52e59a06c9b1c45954b306d0dee

http-server/src/main/java/io/micronaut/http/server/cors/DefaultCorsOriginConfigurationRetriever.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12982/commits/376917d040991780fbf98f155510cdd3aed30e82

http-server/src/main/java/io/micronaut/http/server/cors/CrossOriginUtil.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12982/changes/9f36cc6a3c9c34846066e14ea0f3c5ab2154ea80

PR #12984, opened 2026-09-03
docs: document Python threading configuration

graemerocher:
## Summary

- Document GraalPy's GIL and current Java virtual-thread limitation.
- Recommend cached platform-thread `io` and `blocking` executors.
- Document experimental Python context pooling and tuning guidance.
- Add the new page to both Python table of contents files.

## Verification

- `git diff --check` passes.
- `./gradlew publishGuide` and `./gradlew docs` were attempted but stop before documentation rendering because existing inject Java, Groovy, and Kotlin compilation errors report unresolved framework classes.

### DIFF
--- a/src/main/docs/guide/languageSupport/python/pythonThreading.adoc
+++ b/src/main/docs/guide/languageSupport/python/pythonThreading.adoc
@@ -2,7 +2,7 @@
 
 === The Python GIL
 
-GraalPy currently uses a Global Interpreter Lock (GIL). The GIL allows only one thread to execute Python code at a time in a given GraalPy context. GraalPy can release the GIL while calling foreign code or performing I/O, but you should not assume that Python code in the same context executes in parallel. See the https://github.com/oracle/graalpython/blob/master/docs/user/Interoperability.md[GraalPy multithreading documentation] for more information.
+GraalPy currently uses a Global Interpreter Lock (GIL). The GIL allows only one thread to execute Python code at a time in a given GraalPy context. GraalPy can release the GIL while calling native code or performing I/O, but you should not assume that Python code in the same context executes in parallel. See the https://github.com/oracle/graalpython/blob/master/docs/user/Interoperability.md[GraalPy multithreading documentation] for more information.
 
 Micronaut can run Python code in multiple GraalPy contexts. Each context has its own GIL, so a pool of contexts can reduce GIL contention when several requests execute Python concurrently. Context pooling is experimental and should be enabled and tuned only after measuring the workload.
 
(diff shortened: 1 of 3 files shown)


src/main/docs/guide/languageSupport/python/pythonThreading.adoc
```diff
@@ -0,0 +1,59 @@
...
+----
+micronaut:
+  python:
+    pool:
+      enabled: true
+      size: 8
+      warn-wait: 2s
+----
+
+`enabled` defaults to `true`. `size` is the target number of additional GraalPy contexts; a value of `0` selects the default of twice the number of available processors. Contexts are created lazily and each context has a memory and warm-up cost. For CPU-bound Python code, start with a pool size near the number of available processors and increase it only when measurements show useful parallelism. For I/O-heavy Python code, a larger pool may improve throughput, but it should remain bounded by the concurrency the application can sustain. Set `warn-wait` to a useful threshold, such as `2s`, to log when requests wait for an available context and use that signal when tuning `size`.
```
fniephaus:
```suggestion
`enabled` defaults to `true`. `size` is the target number of additional GraalPy contexts; a value of `0` selects the default of twice the number of available processors. Contexts are created lazily and each one consumes memory and requires initialization work. Compiled Python code is shared across contexts where possible. For CPU-bound Python code, start with a pool size near the number of available processors and increase it only when measurements show useful parallelism. For I/O-heavy Python code, a larger pool may improve throughput, but it should remain bounded by the concurrency the application can sustain. Set `warn-wait` to a useful threshold, such as `2s`, to log when requests wait for an available context and use that signal when tuning `size`.
```

src/main/docs/guide/languageSupport/python/pythonThreading.adoc
```diff
@@ -0,0 +1,59 @@
+== Python Threading
+
+=== The Python GIL
+
+GraalPy currently uses a Global Interpreter Lock (GIL). The GIL allows only one thread to execute Python code at a time in a given GraalPy context. GraalPy can release the GIL while calling foreign code or performing I/O, but you should not assume that Python code in the same context executes in parallel. See the https://github.com/oracle/graalpython/blob/master/docs/user/Interoperability.md[GraalPy multithreading documentation] for more information.
```
fniephaus:
```suggestion
GraalPy currently uses a Global Interpreter Lock (GIL). The GIL allows only one thread to execute Python code at a time in a given GraalPy context. GraalPy can release the GIL while calling native code or performing I/O, but you should not assume that Python code in the same context executes in parallel. See the https://github.com/oracle/graalpython/blob/master/docs/user/Interoperability.md[GraalPy multithreading documentation] for more information.
```

PR #12986, opened 2026-09-03
Register binder constructors for native images

n0tl3ss:
native guide runs fail when dynamically invoking AbstractArgumentBinder(ConversionService) and DefaultExecutableBinder(Map): these constructors are not reachable in the native image.

Add narrow Micronaut Core reachability metadata for the exact constructors and a packaging regression spec.

Validation:
- jq empty core/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core/reachability-metadata.json
- :micronaut-core:test was started but interrupted after dependency compilation; no native build was run.

### DIFF
--- a/core/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core/reachability-metadata.json
+++ b/core/src/main/resources/META-INF/native-image/io.micronaut/micronaut-core/reachability-metadata.json
@@ -1,0 +1,37 @@
+{
+  "reflection": [
+    {
+      "type": "io.micronaut.core.bind.annotation.AbstractArgumentBinder",
+      "methods": [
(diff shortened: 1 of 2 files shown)


n0tl3ss:
Crema/native rationale: the native closed-world analysis omitted the binder constructors used by generated Python HTTP code, whereas the JVM loads those constructors on demand. Registering the exact constructors in Core fixes native reachability without adding a broad preserve or changing guide code.

Observed native errors: `AbstractArgumentBinder(ConversionService)` and `DefaultExecutableBinder(Map)` were reported as not reachable during generated Python HTTP/binder execution. JVM class loading resolves these constructors on demand.

PR #12992, opened 2026-09-04
Add extensible response header populators

sdelamo:
Introduce a response filter that applies headers supplied by ordered populator beans while preserving headers already set by applications.

Move configured cross-origin policy headers out of CorsFilter into dedicated populators and cover the behavior with an HTTP server integration test.

This could be used here: https://github.com/micronaut-projects/micronaut-security/pull/2266 and https://github.com/micronaut-projects/micronaut-core/pull/12982#discussion_r3926489549


http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -302,38 +298,6 @@ protected void setAllowPrivateNetwork(CorsOriginConfiguration config, MutableHtt
...
     }
 
-    /**
-     * Sets the Cross-Origin-Embedder-Policy header from the CORS configuration when it is not already present.
-     *
-     * @param corsConfiguration CORS Configuration
-     * @param response The {@link MutableHttpResponse} object
-     * @since 5.2.0
-     */
-    protected void setCrossOriginEmbedderPolicy(HttpServerConfiguration.CorsConfiguration corsConfiguration, MutableHttpResponse<?> response) {
```
sdelamo, line 312: this is not a breaking change as this code was never released

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
```diff
@@ -302,38 +298,6 @@ protected void setAllowPrivateNetwork(CorsOriginConfiguration config, MutableHtt
...
-    }
-
-    /**
-     * Sets the Cross-Origin-Resource-Policy header from the CORS configuration when it is not already present.
-     *
-     * @param corsConfiguration CORS Configuration
-     * @param response The {@link MutableHttpResponse} object
-     * @since 5.2.0
-     */
-    protected void setCrossOriginResourcePolicy(HttpServerConfiguration.CorsConfiguration corsConfiguration, MutableHttpResponse<?> response) {
```
sdelamo, line 328: this is not a breaking change as this code was never released.

http-server/src/main/java/io/micronaut/http/server/filter/ResponseHeaderPopulatorFilter.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Adds headers produced by the registered populators when those headers
+     * have not already been set on the response.
+     *
+     * @param request The current HTTP request
+     * @param response The current HTTP response
+     */
+    @ResponseFilter
+    @Internal
+    void filterResponse(HttpRequest<?> request, MutableHttpResponse<?> response) {
```
graemerocher, line 61: I am not sure about this to be honest, it seems of lots of complexity for something simple and adds a global filter that seems suboptimal from a performance perspective 

What does this gain over just registering another filter?

sdelamo: We need the ability for features to contribute response headers. I thought that having a single filter, `ResponseHeaderPopulatorFilter`, which is disabled by default unless there is a bean of type `ResponseHeaderPopulator`, would have better performance than registering multiple filters for header population.

We need to populate these two headers, `Cross-Origin-Embedder-Policy` and `Cross-Origin-Resource-Policy`. [We will also need to populate the `Content-Security-Policy` response header](https://github.com/micronaut-projects/micronaut-security/blob/0e2007d882ea10f7ed6cc4f41cff27bafe052e3d/security-csp/src/main/java/io/micronaut/security/csp/filters/ContentSecurityPolicyFilter.java), among others.

This approach tries to avoid having multiple filters for the same use case: populating a response header if it does not already exist.

graemerocher: but the way you contribute response headers already exists, you define a filter 

also it adds yet another way to do the same thing.

I would like to hear @dstepanov and @yawkat opinion on this one. In the end yes it is true you end up with multiple filters, but you are still going to end up with multiple classes contributing to the stack somewhere

graemerocher: and we then have to document somewhere else how ordering is handled with this new API

sdelamo: > and we then have to document somewhere else how ordering is handled with this new API

[I added docs.](https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/1ac3bd6fd44ee30380c2cd553db38c91c6ec339e)

The motivation for this PR is that I am seeing the need to add HTTP response headers in multiple places, mostly security-related HTTP response headers. I prefer to have a single filter. 

This API is useful internally, but it is also useful externally. It removes the need for users to implement a filter (decide whether they want/need to use `HttpServerFilter` or a method-annotated filter—the recommended approach, etc.) to populate a HTTP Response header.

[@dstepanov's PR review](https://github.com/micronaut-projects/micronaut-core/pull/12982#discussion_r3926489549) raises concerns about mixing the CORS logic (preflight, etc.) with the population of CORS-related HTTP response headers in the same filter. I am happy to leave the response header population in the `CorsFilter`, but I think it is cleaner to have an API such as the one this PR adds.

http-server/src/main/java/io/micronaut/http/server/filter/ResponseHeaderPopulatorFilter.java
```diff
@@ -0,0 +1,74 @@
...
+        for (ResponseHeaderPopulator responseHeaderPopulator : responseHeaderPopulators) {
+            HttpHeaderTuple httpHeader = responseHeaderPopulator.findHttpHeader(request);
+            if (httpHeader != null && !response.getHeaders().contains(httpHeader.name())) {
+                response.getHeaders().add(httpHeader.name(), httpHeader.value());
+            }
+        }
+    }
+
+    @Override
+    public int getOrder() {
```
dstepanov, line 71: Ordering looks inverted for what the contract promises. `FilterRunner` runs request filters in ascending order and response filters in the *reverse* order (see the javadoc on `FilterRunner#filters`, and `sortReverse`). With `ServerFilterPhase.LAST.after()` (59250, the maximum) this filter's `@ResponseFilter` runs *before* every other filter's response filter.

That's fine for headers set by the controller/response itself (already present by then), but it means any other filter that sets the same header later — e.g. the micronaut-security filters this SPI is meant to serve — will not be seen by the `contains()` check, and will end up appending a second value. If the intent is "everything else wins, we only fill in the gaps", this should have the *lowest* order (e.g. `ServerFilterPhase.FIRST.before()`) so its response filter runs last.

A test with a second `@ServerFilter` that sets one of the populated headers in its own response filter would pin this down.

---

http-server/src/main/java/io/micronaut/http/server/cors/CorsResponseHeaderPopulatorFactory.java
```diff
@@ -0,0 +1,86 @@
...
+     * header.
+     *
+     * @return A populator for the configured policy
+     * @throws DisabledBeanException If no resource policy is configured
+     */
+    @Named("crossOriginResourcePolicy")
+    @Singleton
+    @Requires(property = "micronaut.server.cors.cross-origin-resource-policy")
+    ResponseHeaderPopulator crossOriginResourcePolicy() {
+        CrossOriginResourcePolicy policy = corsConfiguration.getCrossOriginResourcePolicy();
```
dstepanov, line 62: Three things here:

1. The property names are hardcoded strings. `HttpServerConfiguration.PREFIX` (`micronaut.server`) exists — `@Requires(property = HttpServerConfiguration.PREFIX + ".cors.cross-origin-resource-policy")` keeps this in sync if the prefix ever moves.
2. `@Requires(property = ...)` plus the `DisabledBeanException` null-check is double gating. The null-check is only reachable when the property is present but unconvertible (e.g. `cross-origin-resource-policy: bogus`); in that case silently disabling the bean hides a config typo that previously also silently did nothing. Dropping the `@Requires` and keeping only the `DisabledBeanException` would be equivalent and simpler.
3. Behaviour change worth noting: the policy value is now read once at bean creation and captured in the lambda, whereas `CorsFilter` read it from `CorsConfiguration` per response. Anything mutating the configuration bean after startup (a `BeanCreatedEventListener<CorsConfiguration>`, a test that sets the value programmatically) no longer has any effect, and the `@Requires` means such a bean is not even created. Reading `corsConfiguration.getCrossOriginResourcePolicy()` inside the returned lambda would preserve the old semantics.

---

http/src/main/java/io/micronaut/http/HttpHeaderTuple.java
dstepanov, line 1: A few notes on this new public type:

- Copyright header says `2017-2206`.
- `@Introspected` — is it actually needed? Nothing here is serialized or bean-mapped, and it adds introspection metadata to every build that sees `io.micronaut.http`.
- `CharSequence` components in a record mean `equals`/`hashCode` are identity-based for anything that isn't a `String` (the `CrossOrigin*Policy` enums this PR passes in are `CharSequence`s, so `new HttpHeaderTuple("a", policy).equals(new HttpHeaderTuple("a", "same-site"))` is false). If the record is meant to have value semantics, `String` components would be safer; if it isn't, that's fine but worth being deliberate about.
- Naming: `HttpHeaderTuple` reads oddly for public API in `io.micronaut.http` — `HttpHeaderValue` or `HttpHeaderEntry` would say more. Also, `ResponseHeaderPopulator` is `@Experimental` but this type, which exists solely to serve it, isn't — they should probably match.

---

http-server/src/main/java/io/micronaut/http/server/filter/ResponseHeaderPopulator.java
dstepanov, line 1: On the API shape, since this is new public (if `@Experimental`) SPI and micronaut-security is meant to build on it:

- One header per bean means a module that wants to set several security headers has to register N beans, each re-deriving the same request state. `@Nullable Collection<HttpHeaderTuple> findHttpHeaders(HttpRequest<?>)` would cover both cases at no cost.
- The populator only sees the request. Security headers frequently depend on the response — status, content type (e.g. no `X-Frame-Options` on a non-HTML response). Worth deciding now whether that is deliberately out of scope, because adding the response later is a breaking signature change.

Minor: the class javadoc `{@link ResponseHeaderPopulatorFilter}` points at a package-private `@Internal` class, so it renders as a dead link in the published javadoc. Plain text would be better.

---

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
dstepanov, line 1: `isEnabled(HttpRequest)` (line 117) is now stale: it still returns `true` when `crossOriginEmbedderPolicy` or `crossOriginResourcePolicy` is configured, but with those two `set*` methods gone, `filterResponse` does nothing for a request without an `Origin` header (`getConfiguration(request)` is empty). So with either policy configured, `CorsFilter` is now run for *every* request purely to no-op. That whole block, and the trace log below it, should collapse to `return request.getOrigin().isPresent()`.

Related question: `filterPreFlightRequest` is `@PreMatching` and short-circuits with a response. Does the new `ResponseHeaderPopulatorFilter` response filter still run for that short-circuited preflight response, or do preflight responses now lose the configured COEP/CORP headers? Worth a TCK case for a preflight (`OPTIONS` + `Origin` + `Access-Control-Request-Method`) either way.

---

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CrossOriginPolicySpec.groovy
dstepanov, line 1: This spec is now a strict subset of the new `CrossOriginPolicyTest` in `http-server-tck` (same configuration, same two endpoints, same assertions), and the TCK runs against Netty too. Suggest deleting the spec rather than maintaining both.

---

test-suite-python/src/test/python/micronaut/docs/server/filters/SecurityHeaderPopulator.py
dstepanov, line 1: `import java` is unused here. Also, unlike `TraceFilter.py` next door there is no accompanying `SecurityHeaderPopulatorSpec.py`, so nothing exercises this file — the same is true of the Java/Groovy/Kotlin copies, which are gated on `spec.filter=SecurityHeaderPopulator` that no test ever sets. Since the docs only reference the Java/Groovy/Kotlin snippets, either add a small spec that asserts `X-Content-Type-Options: nosniff` (which would also cover the SPI end-to-end in each language) or drop the Python copy.

---

dstepanov (commented):
Reviewed the design and the behaviour changes. The direction — pulling the "set this header unless the app already set it" logic out of `CorsFilter` into an ordered SPI — is a good one, and it fixes the pre-existing wart that COEP/CORP were tied to the CORS filter's lifecycle. A few things I think need attention before this lands:

**Likely functional issue**
- `ResponseHeaderPopulatorFilter#getOrder` returns `ServerFilterPhase.LAST.after()`, but response filters run in *reverse* order, so this filter's response filter runs first, not last. Headers set by other filters (including the micronaut-security filters this is aimed at) won't be seen by the `contains()` check and will be duplicated. Details inline.
- `CorsFilter#isEnabled` still short-circuits on the two policies, which is now dead logic that makes `CorsFilter` run on every request for nothing.

**API shape (worth settling while it's still `@Experimental`)**
- One header per bean, and no access to the response. Both are hard to change later; comments inline.
- `HttpHeaderTuple`: `@Introspected` looks unnecessary, `CharSequence` record components give identity-based `equals`, and the copyright header says `2206`.

**Config/test**
- The populator factory captures the policy value at startup and gates on hardcoded property names — a small behaviour change vs. reading the configuration per response.
- The new Netty spec case duplicates the new TCK test; the docs example classes are never exercised by any test.

Nothing here is a blocker on the concept, just the ordering one on correctness.

---

http-server/src/main/java/io/micronaut/http/server/filter/ResponseHeaderPopulatorFilter.java
graemerocher: this is exactly the kind of ordering confusion I am concerned about with this PR

sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/180669830f77fd32d90f6b176851b0c3dd825fe2

http-server/src/main/java/io/micronaut/http/server/cors/CorsResponseHeaderPopulatorFactory.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/deb1b1794730619dca803d5e58f3583c86996c8d

http/src/main/java/io/micronaut/http/HttpHeaderEntry.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/0487806e0b57447959c0c3971a3a87d63e624d16

http-server/src/main/java/io/micronaut/http/server/cors/CorsFilter.java
sdelamo: > isEnabled(HttpRequest) (line 117) is now stale: it still returns true when crossOriginEmbedderPolicy or crossOriginResourcePolicy is configured, but with those two set* methods gone, filterResponse does nothing for a request without an Origin header (getConfiguration(request) is empty). So with either policy configured, CorsFilter is now run for every request purely to no-op. That whole block, and the trace log below it, should collapse to return request.getOrigin().isPresent().

Resolved via 
https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/8da8b2168f52cf91f4e1596f3f400fb313de90f0

test-suite-python/src/test/python/micronaut/docs/server/filters/SecurityHeaderPopulator.py
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/362116eb0b44428e30e0abd2bd2a52a3470cf841

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/cors/CrossOriginPolicySpec.groovy
sdelamo: it is ok. I don't mind the duplication to have a test closer to the code and the additional benefit of having the test in the TCK.

http-server/src/main/java/io/micronaut/http/server/filter/ResponseHeaderPopulator.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-core/pull/12992/commits/d2ff625957383aa691f141a6cbe478280c6109ce

graemerocher (approved):

PR #13011, opened 2026-09-04
Inherit the active environment in GraalPy contexts

n0tl3ss:
GraalPy contexts configured with python.Executable must inherit the host environment so the active virtualenv metadata is visible to embedded Python.

This enables EnvironmentAccess.INHERIT and adds a regression test for VIRTUAL_ENV propagation. It keeps the existing executable discovery and does not add native-image preserves.

Validation: -Ppython-ci :micronaut-context-python:test --tests io.micronaut.context.python.GraalPyContextFactoryTest (6/6 passed).

### DIFF
--- a/context-python/src/main/java/io/micronaut/context/python/GraalPyContextFactory.java
+++ b/context-python/src/main/java/io/micronaut/context/python/GraalPyContextFactory.java
@@ -29,6 +29,7 @@ import jakarta.inject.Named;
 import jakarta.inject.Singleton;
 import org.graalvm.polyglot.Context;
 import org.graalvm.polyglot.Engine;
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

n0tl3ss:
Crema/native rationale: the embedded GraalPy context used by Crema must inherit the active environment explicitly; a normal JVM launch inherits the process environment through its ordinary runtime path. This fixes native launcher behavior without changing application configuration.

Observed native symptom: embedded `python.Executable` contexts did not inherit `VIRTUAL_ENV`/host environment metadata, so the active GraalPy virtualenv was invisible to the native fixture. A normal JVM launch inherited the process environment.

PR #13018, opened 2026-09-05
Preserve map key spelling for indexed keys in the RAW property catalog

### LINKED ISSUE #13016: Property normalization hyphenates map keys after a list index in .properties and -D sources, but not in YAML sequences
### Expected Behavior

A `Map` key that follows a list index in a property key — `SOME_KEY` in `foo.items[0].SOME_KEY=x` — should reach the aggregated value spelled as written, the same as when the identical configuration is written as a YAML sequence.

Both of these:

```yaml
# application.yml
foo:
  items:
    - SOME_KEY: x
```

```properties
# application.properties
foo.items[0].SOME_KEY=x
```

should produce:

```java
env.getProperty("foo.items", Argument.listOf(Map.class));   // [{SOME_KEY=x}]
```

### Actual Behaviour

The two disagree. The YAML sequence keeps `SOME_KEY`. The property key form lower-cases it and converts the underscore to a hyphen:

```
YAML sequence             : [{SOME_KEY=x}]
foo.items[0].SOME_KEY     : [{some-key=x}]
```

`StringConvention.RAW` cannot recover the original spelling. `PropertyCatalog.RAW` is populated only with the verbatim flat entry — `expandProperty` and `collapseProperty` run exclusively against the GENERATED catalog — so RAW offers either the hyphenated `List` or the unexpanded flat `String`, never an aggregated `List` spelled `SOME_KEY`:

```java
env.getProperties("foo", StringConvention.RAW)
// {items[0].SOME_KEY=x}
```

`@MapFormat(keyFormat = StringConvention.RAW)` does not help either. For a `Map`-typed target it is honored and does consult the RAW catalog, but since RAW never received the expanded structure, the result is the unexpanded flat entry rather than the list.

**Affected sources** are those that supply the key as a flat string containing an index: `.properties` files, system properties (`-D`), command-line arguments, config trees, and `Map`-based sources such as `PropertySource.of(...)`. **Not affected:** YAML sequences and JSON, which supply an already-structured `List<Map<…>>` whose inner keys never pass through key normalization. Environment variables are also unaffected — `EnvironmentProperties` lower-cases the whole name before the resolver sees it, so no original spelling exists to preserve.

This matters wherever the `Map` key is data rather than a Java property name and its exact spelling is significant — environment variable names, HTTP header names, and any `Map<String, String>` whose keys are not Micronaut property names.

Root cause: `PropertySourcePropertyResolver.processPropertySource` normalizes the **entire** key through `resolvePropertiesForConvention` → `NameUtils.hyphenate(property, true)`, then passes the already-normalized remainder to `expandProperty`, which uses that string for both the list index and the `Map` key. Normalization intended for property *names* is applied to the `Map` *key* after the indexed segment. Hyphenating the prefix before `[` is intended and covered by tests; nothing covers segments after `]`.

### Steps To Reproduce

```java
import io.micronaut.context.env.MapPropertySource;
import io.micronaut.context.env.PropertySourcePropertyResolver;
import io.micronaut.core.naming.conventions.StringConvention;

import java.util.List;
import java.util.Map;

public class Repro {
        var indexedKey = new PropertySourcePropertyResolver(
            MapPropertySource.of("indexedKey", Map.of("foo.items[0].SOME_KEY", "x"))
        );

        // Structured, as a YAML sequence supplies it
        var structured = new PropertySourcePropertyResolver(
            MapPropertySource.of("structured", Map.of("foo.items", List.of(Map.of("SOME_KEY", "x"))))
        );

        System.out.println("structured : " + structured.getProperties("foo").get("items"));
        System.out.println("indexedKey : " + indexedKey.getProperties("foo").get("items"));
        System.out.println("RAW        : " + indexedKey.getProperties("foo", StringConvention.RAW));
    }
}
```

Output on 5.1.11:

```
structured : [{SOME_KEY=x}]
indexedKey : [{some-key=x}]
RAW        : {items[0].SOME_KEY=x}
```

### Environment Information

- Operating System: macOS 15.7.5
- JDK Version: Oracle GraalVM 25.0.4+7.1 (Java 25)

### Example Application

_No response_

### Version

5.1.11

MariusVolkhart:
Micronaut normalizes a property name to kebab case. That is correct for property names, but wrong for a Map key that a user wrote as data. A key supplied as a single flat string, such as foo.items[0].SOME_KEY, is normalized in full before it is split into structure, so the Map key becomes some-key. The same configuration written as a YAML sequence keeps SOME_KEY, because the loader supplies an already-structured List<Map<..>> whose inner keys never pass through normalization. The two authoring forms therefore disagree, and the spelling matters whenever the key is data rather than a property name: environment variable names, HTTP header names, and any Map<String, String> keyed by something outside Micronaut's naming rules.

StringConvention.RAW exists to recover the spelling a user wrote, but it could not help here. The RAW catalog only ever received the verbatim flat entry, never the aggregated structure, so it could return the unexpanded String but never the list carrying the original Map keys.

Run the indexed-property expansion into the RAW catalog as well, built from the verbatim key and stored under the verbatim base name. The GENERATED and NORMALIZED catalogs are untouched, so binding, which depends on the normalized spelling, is unaffected; RAW simply gains the view it was always meant to offer.

Skip the expansion for the ENVIRONMENT_VARIABLE convention. No user spelling survives to that point, and its rewritten keys, such as A[0]_B, would be misread as Map keys rather than as delimiters.

Give RAW a deep copy of any existing container before expanding into it. Catalogs store values by reference, so an aggregate may be the very instance GENERATED serves, or the property source's own value; mutating it in place would corrupt both.

Fixes #13016

### DIFF
--- a/inject-java/src/test/groovy/io/micronaut/inject/configproperties/ConfigurationPropertiesSpec.groovy
+++ b/inject-java/src/test/groovy/io/micronaut/inject/configproperties/ConfigurationPropertiesSpec.groovy
@@ -20,6 +20,7 @@ import io.micronaut.context.annotation.BeanProperties
 import io.micronaut.context.env.PropertySource
 import io.micronaut.core.util.CollectionUtils
 import spock.lang.Specification
+import spock.lang.Unroll
 
 class ConfigurationPropertiesSpec extends Specification {
 
@@ -106,6 +107,31 @@ class ConfigurationPropertiesSpec extends Specification {
         config.defaultPort == 9999
         config.defaultValue == 9999
         config.primitiveDefaultValue == 9999
+    }
+
+    @Unroll
+    void "test configuration properties list binding accepts every spelling of an indexed key inner segment #key"() {
+        given: "the same indexed key written with a differently cased or separated inner segment"
+        ApplicationContext applicationContext = ApplicationContext.builder("test").build()
+        applicationContext.environment.addPropertySource(PropertySource.of(
+            'test',
+            [(key): 123]
+        ))
+        applicationContext.start()
+
+        expect: "every spelling binds to the same property of the object in the List"
+        applicationContext.getBean(MyConfig).innerVals[0].expireUnsignedSeconds == 123
+
+        cleanup:
+        applicationContext.close()
+
+        where:
+        key << [
+            'foo.bar.innerVals[0].expireUnsignedSeconds',
+            'foo.bar.innerVals[0].expire-unsigned-seconds',
+            'foo.bar.innerVals[0].expire_unsigned_seconds',
+            'foo.bar.innerVals[0].EXPIRE_UNSIGNED_SECONDS',
+        ]
     }
 
     void "test configuration inner class properties binding"() {
--- a/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
+++ b/inject/src/main/java/io/micronaut/context/env/PropertySourcePropertyResolver.java
@@ -746,16 +746,13 @@ public class PropertySourcePropertyResolver implements PropertyResolver, AutoClo
                                 property,
                                 properties.getOrigin()
                             ));
-                            expandProperty(
+                            expandIndexedProperty(
+                                entries,
+                                propertyName,
                                 resolvedProperty.substring(i),
-                                val -> entries.put(propertyName, new DefaultPropertyEntry(
-                                    propertyName,
-                                    val,
-                                    property,
-                                    properties.getOrigin()
-                                )),
-                                () -> entries.getOrDefault(propertyName, NULL_ENTRY).value(),
-                                value
+                                value,
+                                property,
+                                properties.getOrigin()
                             );
                         }
                         if (first) {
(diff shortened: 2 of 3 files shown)


dstepanov (commented):
The implementation itself holds up — I traced the aliasing argument and couldn't break it, and `:micronaut-inject`, `:micronaut-context` and `:micronaut-inject-java` all pass on the branch. Two things I'd like to settle before this goes in, though.

## `Fixes #13016` overclaims

The issue's stated Expected Behavior is a `getProperty` call on the **default** catalog:

```java
env.getProperty("foo.items", Argument.listOf(Map.class));   // [{SOME_KEY=x}]
```

Running the issue's own reproducer against this branch:

```
structured GENERATED : [[SOME_KEY:x]]
indexed    GENERATED : [[some-key:x]]
indexed    getProperty listOf(Map) : [[some-key:x]]
indexed    RAW        : [items[0].SOME_KEY:x, items:[[SOME_KEY:x]]]
```

The first two lines — the disagreement the issue is titled after — are unchanged. Only the RAW view improves, and only for callers who opt into `StringConvention.RAW`.

Leaving GENERATED alone is the right call, and the description says so plainly. But `Fixes` will auto-close an issue whose reported symptom still reproduces. Could we make this `Related to #13016` and leave the issue open for the GENERATED-side decision? It also reframes what we're reviewing: as written this isn't a bug fix, it's a new capability for `StringConvention.RAW` landing on a patch branch.

## Every app pays for the deep copy

The copy at the `rawEntries.put` site is unconditional, so it runs for every container-valued property in every property source — including apps with no indexed key anywhere. `AbstractPropertySourceLoader.processMap` emits every `List` value whole, so each list-of-maps in a YAML config gets its full subtree duplicated and held permanently in the RAW catalog. RAW previously stored by reference and cost nothing.

The copy is only *needed* when a later indexed key expands into a base that a bare key already populated — the case the "structured base value plus an indexed key" test covers. That can be made lazy: keep a set of base names RAW has privatized, drop a name from it on the bare `put`, and in the RAW branch of the indexed path deep-copy the existing aggregate only when its base isn't in the set. Same invariant, O(1) on the common path, and the second copy folds into the same mechanism. Given this is 5.2.x, I'd rather have that up front than as a follow-up.

## Two smaller notes

`deepCopyForRawExpansion` rebuilds every `Map` as a `LinkedHashMap`, so a source supplying a `TreeMap` or a case-insensitive map silently loses that behaviour in RAW. It also passes arrays and `Set`s through uncopied — harmless today, since `expandProperty` only ever builds `List`/`Map`, but the javadoc's "every value entering the RAW catalog is one the catalog alone owns" isn't quite true for those. Either handle them or narrow the claim.

The user-visible RAW shape change is worth a release note. RAW submaps now carry the aggregate alongside the flat keys, mirroring GENERATED exactly:

```
RAW nested submap: [items[0]:[SOME_KEY:x], items:[[SOME_KEY:x], [OTHER:y]], items[1]:[OTHER:y]]
GEN nested submap: [items[0]:[some-key:x], items:[[some-key:x], [other:y]], items[1]:[other:y]]
```

So `getPropertyEntries(name, RAW)`, `@MapFormat(keyFormat = RAW)` bindings and the `/info` endpoint (`ReactiveInfoAggregator` uses `getAllProperties(RAW, NESTED)`) all gain entries. The symmetry with GENERATED is the right design; it just isn't called out anywhere.

## What's good

The ordering tests are the strongest part — structured-then-indexed, indexed-then-bare-then-indexed, and the nested re-alias at `t.env[1]` cover exactly the cases I tried to construct by hand, and they correctly use `LinkedHashMap` for deterministic iteration. The scalar-vs-container collisions I probed (`t.env=hello` then `t.env[0].A=1`, and the reverse) behave identically in RAW and GENERATED. Extracting `expandIndexedProperty` is a clean deduplication, and the ENVIRONMENT_VARIABLE skip is correct and unusually well explained.

One note on the `ConfigurationPropertiesSpec` addition: it passes on the parent commit unchanged, so it guards existing GENERATED behaviour rather than covering this change. Fine to keep, just worth knowing that all the change-driven coverage lives in `PropertySourcePropertyResolverSpec`.

dstepanov:
Correction on my second point — I tried the lazy copy I proposed and **your test caught it**, so the placement I suggested was wrong.

I had said to privatize "in the RAW branch of the indexed path". Doing that fails `test a structured base value plus an indexed key leaves no hyphenated key in RAW`:

```
RAW: [[A:1], [b:2, B:2]]   expected [[A:1], [B:2]]
```

The reason is the part I'd underestimated: the eager copy isn't only protecting GENERATED from RAW, it's **snapshotting the value before GENERATED mutates it**. For `t.env=[[A:1]]` followed by `t.env[1].B=2`, GENERATED's own `expandIndexedProperty` runs first in the iteration and appends `[b:2]` to the *source list in place* — the very instance RAW holds by reference. Privatizing after that inherits GENERATED's hyphenated `b`. So the copy has to happen before the GENERATED expansion, not after. Nice test; it's doing real work.

A lazy version does still work if the privatization is hoisted to the **top of the property iteration**, before `resolvePropertiesForConvention`. At that point the base is still pristine, inductively: the first indexed key for a base copies it before GENERATED touches it, and every later one finds RAW already owns it.

I implemented that and ran it — all 84 `PropertySourcePropertyResolverSpec` tests pass, plus `:micronaut-inject` and `:micronaut-context` in full, with output identical to this PR's:

```
RAW: [items[0].SOME_KEY:x, items:[[SOME_KEY:x]], myItems[0].K:v, myItems:[[K:v]]]
GEN: [items[0].some-key:x, items:[[some-key:x]], my-items[0].k:v, my-items:[[k:v]]]
```

and with no indexed keys anywhere, RAW goes back to sharing the source instance — zero copying for configs that don't use the feature.

The shape is: a `Set<String> rawOwnedBases` field cleared in `reset()`; the bare `rawEntries.put` stores `value` by reference again and does `rawOwnedBases.remove(property)`; and a new block at the top of the `for (String property : properties)` loop that, for a key with `[`, deep-copies the existing RAW aggregate once on `rawOwnedBases.add(base)`. The copy of the leaf `value` on the indexed path stays as you have it — that one is still needed for the `t.env[1]` re-alias case your third ordering test covers.

Happy to push it as a follow-up rather than hold this PR on it — the correctness of what you have isn't in question, it's purely that every app currently pays a deep copy of each list-valued config property for a RAW-only feature. Your call which way round.

My first point stands unchanged: I'd still like `Fixes #13016` softened to `Related to`, since the GENERATED-side disagreement the issue is titled after still reproduces on this branch.

dstepanov:
Thanks @MariusVolkhart — the analysis here was spot on, and the ordering tests are the best part of it. The one that catches a copy taken after GENERATED has already expanded in place saved me from shipping a broken version of my own suggestion.

I've continued this in #13039, rebased onto current `5.2.x`, with your commit credited. The only functional change is that the defensive copy is now taken lazily, on the first indexed key targeting a base, rather than for every container value entering RAW — that keeps configs with no indexed keys from copying anything (retained heap 11.0 MB → 8.0 MB on a synthetic structured config). Your tests all pass unmodified.

I also softened `Fixes #13016` to `Related to`, since the GENERATED-side disagreement that issue reports still reproduces; that part is a separate decision.

Closing this one in favour of #13039. Sorry for the CLA friction — nothing here was blocked on it, and the credit follows you into the new branch.

PR #13026, opened 2026-09-06
Rework the Python (GraalPy) support: correctness, lifecycle, asyncio over Netty, compile time

dstepanov:
A review-and-fix pass over the Python (GraalPy) support: `context-python`, `context-python-netty`, `inject-python`, `inject-python-test`, `test-suite-python`, the `PythonCompile` Gradle task in `buildSrc`, and the Python guide. The modules are unreleased, so nothing here keeps compatibility with earlier snapshots. 171 files, +15.7k / -8.7k lines, squashed from 72 commits.

The work went through ten rounds of an independent read-only review (Codex) plus profiling on OpenJDK 25 (Truffle fallback) and GraalVM CE 25.2.4. Every finding was either fixed with a regression test or recorded as a design decision; the sections below summarise them.

## Runtime correctness (`context-python`)

- **Context identity and entry.** `Value.getContext()` returns a view that is `equals` but not `==` to the creator context. Runtime state is keyed by `Context.equals` and entry goes through the creator instance.
- **Execution frames as the lifecycle unit.** A JVM-wide `PythonContextRegistry` counts every execution of a context in a `ScopedValue` frame. `closeWhenIdle` marks a context closing under the registry lock together with the idle check, so no new outermost execution can start between the check and the close, while calls nested in an execution that began earlier still complete. Every generated entry point now runs inside a frame: pooled lookups, `newInstance`, `newIntroduction`, `enumValue`, `newUninitializedInstance` (with and without properties), `newFrozenDataclassInstance`, `findClass`, `findScript`, the runtime proxy creations, Netty protocol callbacks, scheduled loop callbacks, executor workers and Java stage completions awaited from Python.
- **No monitor across guest calls.** The GraalPy GIL serialises guest execution; a Java monitor held across it could deadlock. Guest code is kept out of `ConcurrentHashMap` remapping functions (a module that re-enters the pool for itself while loading used to fail as a recursive update).
- **Pool lifecycle.** Closing the pool defers the real close until the active-execution counter reaches zero; graceful shutdown keeps serving in-flight borrows and only `ApplicationContext.close()` fails waiting borrows. A context created while `closePool` ran could previously be lost; registration now happens under the monitor. Script injections are versioned and applied by each context's owner instead of being broadcast to contexts another thread may be using.
- **Pooled coroutines keep their lease.** An async pooled bridge call borrowed a context, called the coroutine function, released the context and only then drove the coroutine, so another caller could borrow the same context while the coroutine ran on it. `invokePooledAsync` / `invokePooledScriptAsync` convert the coroutine to a stage while the context is leased and return it when the stage completes.
- **Coroutines without an event loop.** The fallback for a thread with no event loop created asyncio's selector loop, whose self-pipe is a socket pair. Application contexts have no host socket access, so the emulated POSIX layer refused it and the fallback only ever worked in tests with `allowAllAccess`. A socket-free `BaseEventLoop` now drives such coroutines to completion on the calling thread.
- **Memory leaks (found with GC logs, JFR old-object sampling and async-profiler).** GraalPy stores `threading.local` in Java thread locals, so asyncio's per-thread running-loop and policy state pinned every closed context for the life of the thread; the module keeps that state in a per-thread table of its own. Registry states of closed contexts and the event-loop admission sets are swept.
- **Conversion and exceptions.** Collection conversion no longer swallows failures into empty collections and no longer exhausts generators with a size pass; dicts convert to their keys like `list(d)`. Exceptions crossing a coroutine keep their Java type and message. An `async def` returning a Python class produced a bean definition whose dispatch signature did not match the generated `CompletionStage<T>` stub (`NoSuchMethodError` at request time).
- **Cleanup robustness.** No-active-executions listeners, pool closes and bootstrap cleanup continue past a failure, including an `Error`, and rethrow the first with its own type; a context whose main module fails is unregistered and closed.
- **Introductions and proxies.** `newIntroduction` stubbed every abstract method on each instantiation and checked a marker nothing set; the runtime module now prepares a class once per context, and abstract subclasses of a prepared introduction are prepared on their own. The proxy creator no longer carries its own copy of the stubbing.
- **Host class lookup.** `graalpy.context.host-class-lookup` lists the package prefixes Python code may look up; empty keeps every class visible, and the JDK, Jakarta and `io.micronaut` packages are always visible because generated code depends on them.

## asyncio over Netty (`context-python-netty`)

- **Every address family through Netty.** The socket-polling transport layer in `micronaut_asyncio.py` is gone; `create_connection`, `create_server`, `open_connection`, `start_server`, the Unix-domain variants and `create_datagram_endpoint` are Netty channels for IPv4, IPv6 and Unix sockets. Channel classes are chosen from the event loop's `IoHandler`, not from the loop's class name (Netty 4.2 loops are all `SingleThreadIoEventLoop`, so epoll and io_uring loops used to get NIO channels).
- **asyncio contract.** Name resolution follows `getaddrinfo` semantics (`family` narrows, addresses are tried in resolver order, `all_errors=True` raises an `ExceptionGroup`, `local_addr` is paired by family, `AI_NUMERICHOST` applies to every address that would be resolved). `create_server` binds one socket per resolved address; `host=None`/`""` binds one dual-stack IPv6 wildcard listener, or the IPv4 wildcard on a JVM without IPv6 sockets (decided by opening one, never by reading exception text), and any bind failure fails the call. Servers implement `start_serving()`, `wait_closed()` with the 3.12 semantics, `serve_forever()` cancellation, `async with`, `sockets`; transports are `asyncio.Transport` / `DatagramTransport` subclasses with write-buffer limits mapped to Netty's water marks and `pause_writing` / `resume_writing` from channel writability. Connected datagram transports write plain buffers (the native transports failed with `EISCONN` on addressed packets). Failures map to `ConnectionRefusedError`, `socket.gaierror`, `TimeoutError` and `OSError` with the Java throwable as `java_exception`. TLS handshake and shutdown timeouts default to asyncio's 60 and 30 seconds and must be positive numbers. `connection_lost` is called once. IPv6 addresses are 4-tuples and scope ids are preserved. Unsupported arguments (`sock=`, `happy_eyeballs_delay`, `interleave`, `keep_alive`, `set_protocol`) raise instead of taking a slower path.
- **Lifecycle.** Channels are owned by the provider whose loop opened them; shutdown waits for connections and binds in flight, refuses new ones, and a channel handed to `connect_accepted_socket` must belong to the adopting loop. Scheduled callbacks run in the `contextvars` context of the code that scheduled them. `close()` and `stop()` on the loop raise `RuntimeError`; callback failures reach the loop's exception handler, which logs through Micronaut's logger.
- **Per-event-loop contexts are bounded.** `micronaut.python.pool.max-event-loop-contexts` (default 0, no cap) admits the first loops that run Python and routes the others through the shared pool, with a warning naming each refused loop.
- **Throughput.** Received data is handed to protocols as a `ByteBuffer` (a `bytes(byte[])` conversion took 7.8 s per 4 MiB on the interpreted runtime versus 7 ms).

## Compile time (`inject-python`, `buildSrc`)

- **Structure.** The 4.5k-line `PythonStubGenerator` is split (bridge method spec, annotation stub generator, pooled stub generator, shared templates); `PythonAnnotationMetadataBuilder` is split into the builder, `PythonAnnotationValues` (normalisation) and `PythonAnnotationTypes`; `GraalPyUtil` became `PythonTypeResolver` plus value conversion; the processing model, elements and visitor live in `model`, `element` and `visitor` packages. Type annotations travel from Python to Java as structured `TypeRef`s instead of strings re-parsed with bracket matching, and the transformer's Java-type questions are answered from Java. The embedded Python helper snippets moved from Java string constants into a `micronaut_runtime.py` module.
- **Processor.** Attribute expressions are resolved with `ast.literal_eval` instead of `eval`; `except Exception: pass` guards and `print` diagnostics are replaced by narrow guards and visitor-context messages (warnings and info now reach javac's messager). Dead functions removed. Cross-annotation `@AliasFor` resolution is left to the base builder. Annotation literals are checked exactly against the member type (integral ranges via exact `BigDecimal`, finite floats within the float range, one-character strings for `char` and `char[]`, no silent narrowing), untyped decorator members keep strings as strings, annotations without `@Target` may target annotation types, and every `Annotated` spelling (bare, `typing.`, alias, `typing_extensions`) and every set spelling resolve. Array defaults in Python annotation declarations are converted per component.
- **Incremental compilation.** Dependency analysis records every clause of a direct import (multi-line, continued, after a semicolon, submodules imported from their package) with a scanner that respects strings, brackets and comments; only `getattr` on an imported module is a dynamic relationship instead of any `getattr(` anywhere forcing a full rebuild.
- **Gradle task.** `PythonCompile` runs through the Worker API in a process-isolated worker (`PythonCompileWorkAction` / `PythonCompileParameters`, which `.gitignore`'s `build/` rule had been hiding), compiles every source root in one invocation, and the parser shares one GraalPy engine per parser mode with its driver snippets as cached `Source`s. The compile-time Python code has pytest-style tests of its own (`test_micronaut_processor.py`, `test_micronaut_transformer.py`) run inside GraalPy from JUnit.

## Performance

Measured on the interpreted runtime; ratios hold on GraalVM.

- `isNone` is a null check (5.8 µs to 0.02 µs per call; it ran on every bridge return, list element and map entry).
- `invokePythonMethod` calls executable members directly (8.4 µs to 3.5 µs per bridge call).
- Collection sizing uses the array protocol before `__len__` (37 µs to 2.6 µs for five ints).
- Python classes are resolved once per context; dataclass rebuilds write all fields in one guest call; introspected stubs only re-sync changed fields; the generated `Value` constructor reads each member once.
- `PythonVisitorContext` caches its metadata builder and factory; the primary context shares its engine with the pool so compiled code is shared and one compiler queue exists.
- An engine option from `graalpy.engine.options` wins over the built-in `engine.CompilerThreads=1`.
- The JMH benchmark reports process CPU per iteration, since the HTTP path is latency bound.

## Observability and docs

- `PythonContextExecutor.statistics()` returns a `PythonPoolStatistics` snapshot (sizes, borrows, waits, longest wait); with `micronaut-management` the `pythonpool` endpoint serves it.
- Startup logs the active Truffle runtime; a new "Runtime Requirements" guide page explains which JDK gets a JIT for the bundled GraalPy version (`graalpyVersion` in `gradle.properties`), how to detect the fallback runtime, and the compiler-thread setting.
- The threading, asyncio networking and unsupported-features pages describe the pool, event-loop contexts, graceful shutdown, statistics and the exact asyncio contract above.

## Known limit

Generated bridge code resolves the application runtime installed last, so one JVM serves generated bridge code from one application at a time (tracked as D2, documented; `uninstall` is a compare-and-set so a replaced runtime cannot remove its successor).

## Verification

All suites pass on the rebased commit with `-Ppython-ci`, checkstyle clean: context-python 160, context-python-netty 53, inject-python 120, inject-python-test 598, test-suite-python 200. Every review finding has a named regression test.


context-python/src/main/java/io/micronaut/context/python/PythonHttpConversion.java
```diff
@@ -0,0 +1,158 @@
...
+     * @return The converted response
+     */
+    @SuppressWarnings({"unchecked", "NullAway"})
+    public static <T> HttpResponse<T> convertHttpResponse(Value value, Class<T> bodyType) {
+        HttpResponse<?> response = PythonConversion.convertValue(value, HttpResponse.class);
+        if (response == null) {
+            return null;
+        }
+        return convertHttpResponse(response, bodyType);
+    }
```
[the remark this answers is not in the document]
dstepanov: Fixed in 29613d079d57ae4ad0cc0c441964581d85d2fce5: the method now declares its result nullable and the javadoc says when it returns null (the value is not an HttpResponse).

buildSrc/src/main/groovy/io/micronaut/build/internal/python/PythonCompileWorkAction.java
```diff
@@ -0,0 +1,65 @@
...
+public abstract class PythonCompileWorkAction implements WorkAction<PythonCompileParameters> {
+
+    private static final String PYRONAUT_COMPILER_MAIN_CLASS = "io.micronaut.python.compiler.PyronautCompiler";
+
+    @Override
+    public void execute() {
+        String destinationDir = getParameters().getDestinationDir().get();
+        List<File> classpath = new ArrayList<>(getParameters().getClasspath().getFiles());
+        // one compiler run over every root: each run writes the launcher, the VFS file list and the
+        // package initialisers for the whole output, so a run per root would keep only the last root's
```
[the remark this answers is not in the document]
dstepanov: Fixed in 29613d079d57ae4ad0cc0c441964581d85d2fce5: the comment now finishes the sentence (a run per root would keep only the last root's files and drop the others).

context-python/src/main/java/io/micronaut/context/python/GraalPyContextConfiguration.java
```diff
@@ -100,6 +105,44 @@ void setOptions(@MapFormat(keyFormat = StringConvention.RAW, transformation = Ma
...
+    void setHostClassLookup(@Nullable List<String> hostClassLookup) {
+        this.hostClassLookup = hostClassLookup == null ? List.of() : List.copyOf(hostClassLookup);
+    }
+
+    /**
+     * The host class filter for {@link Context.Builder#allowHostClassLookup(Predicate)}.
+     *
+     * @return The filter
+     */
+    Predicate<String> hostClassFilter() {
```
graemerocher, line 139: I am concerned about this, what about user packages? Since we generate java classes from user packages won't this break user code? It may be passing in the test suite because our test packages uses "io.micronaut" but what about "example"

context-python/src/main/java/io/micronaut/context/python/GraalPyContextFactory.java
```diff
@@ -223,7 +248,7 @@ private static Context buildContext(HostAccess hostAccess,
             .hostClassLoader(classLoader)
             .engine(engine)
             .exceptionHandler(GraalPyExceptionHandler.RETHROW_HOST_RUNTIME_EXCEPTION)
-            .allowHostClassLookup(_ -> true);
+            .allowHostClassLookup(contextConfiguration.hostClassFilter());
```
graemerocher, line 258: explain why we need to restrict host class access

context-python/src/main/java/io/micronaut/context/python/PythonConversion.java
```diff
@@ -0,0 +1,556 @@
...
+     * object or the string representation of the value.</p>
+     *
+     * @param value The polyglot value
+     * @return Whether the value represents Java null or Python None
+     */
+    public static boolean isNone(@Nullable Value value) {
+        return value == null || value.isNull();
+    }
+
+    static LocalDate convertLocalDate(Value value) {
```
graemerocher, line 90: add maintainer focused javadoc to all methods of this class

src/main/docs/guide/languageSupport/python/asyncAwait/asyncioNetworking.adoc
```diff
@@ -10,9 +10,11 @@ Supported networking APIs include:
...
...d the callbacks scheduled with `call_soon`, `call_later` and `call_at`, run on the Netty event-loop thread inside a Python execution frame, so a graceful shutdown waits for a running callback the same way it waits for a controller. Once the runtime has selected a context for closing, no new callback or bridge call starts on it (a scheduled callback is skipped, a bridge call fails with `IllegalStateException`) while calls nested in an execution that began earlier still complete. A protocol callback runs only while its context is open; once a close is selected the channel's callbacks fail and the channel is closed. A scheduled callback that raises is reported to the loop's exception handler: the default handler logs it through Micronaut's logger, and a failing custom handler is reported the same way. Channels opened by a Netty event loop are closed by the shutdown of the provider whose loop opened them; the shutdown waits for connections and binds still in flight, and refuses new ones. Java stages awaited from Python complete on the loop inside an execution frame, so a completion arriving after the awaiting coroutine returned is still tracked, and skipped once the context is closing.
```
graemerocher: this is like a big hunk of text now, would be good to break it up to make it easier to read and comprehend

src/main/docs/guide/languageSupport/python/pythonRuntime.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+Python code can look up any class the application class loader can load, through `java.type("...")` and `from ... import ...`. Restrict the lookup to the packages the application needs with `graalpy.context.host-class-lookup`; the JDK, Jakarta and `io.micronaut` packages stay visible because the generated Python code depends on them:
+
+[configuration]
+----
+graalpy:
+  context:
+    host-class-lookup:
+      - com.example
+      - org.reactivestreams
```
graemerocher, line 49: this change is likely to break a lot of downstream guides which use other packages (kafka, mongo, etc.)

python-improvements.md
```diff
@@ -0,0 +1,554 @@
+# Micronaut Python (GraalPy) implementation: improvements list
```
graemerocher: delete this planning document

graemerocher (changes requested):

context-python/src/main/java/io/micronaut/context/python/GraalPyContextConfiguration.java
dstepanov: The restriction is opt-in: the default (property unset or empty) is the `className -> true` branch just above, so every class the application class loader can load is visible, user packages such as `example` included. The test suite runs with the default, which is why `example` works there. Only an application that sets `graalpy.context.host-class-lookup` narrows the lookup, and then it must list its own packages; a75118924720e9a896f3da2b8e6acb7202fea04b says so in the javadoc, the guide and a new test (`hostClassLookupIsUnrestrictedByDefault`).

context-python/src/main/java/io/micronaut/context/python/GraalPyContextFactory.java
dstepanov: Added a comment in a75118924720e9a896f3da2b8e6acb7202fea04b: Python reaches Java through `java.type` and imports, which by default exposes every loadable class, `java.lang.Runtime` and `ProcessBuilder` included. The filter is the knob for applications that run Python they trust less than their Java; unset, it accepts everything.

context-python/src/main/java/io/micronaut/context/python/PythonConversion.java
dstepanov: Done in a75118924720e9a896f3da2b8e6acb7202fea04b: a class-level note on the rules every converter follows (None, host objects, wrappers, no silent shrinking), the order of attempts on `convertValue`, and javadoc on each package-private and private method saying what calls it and why.

src/main/docs/guide/languageSupport/python/asyncAwait/asyncioNetworking.adoc
dstepanov: Split in a75118924720e9a896f3da2b8e6acb7202fea04b into sections (address families and name resolution, servers, transports, errors and validation, callbacks and shutdown, transport selection, TLS, transport extras) with bullet lists; no content dropped.

src/main/docs/guide/languageSupport/python/pythonRuntime.adoc
dstepanov: Nothing changes for downstream guides: the setting is opt-in and unset by default, so kafka, mongo and every other package stay visible. The guide now says that first and describes the restriction as something an application enables deliberately (a75118924720e9a896f3da2b8e6acb7202fea04b).

python-improvements.md
dstepanov: Deleted in a75118924720e9a896f3da2b8e6acb7202fea04b; the PR description no longer refers to it.

graemerocher (approved):

PR #13027, opened 2026-09-06
Do not register an implicit HEAD route when an explicit @Head maps to the same URI

### LINKED ISSUE #13020: GET and HEAD route conflict
### Expected Behavior

HEAD requests should be routed to the defined method annotated with @Head, while GET requests should be routed to the @Get method.
This is a requirement when implementing the IETF RPP draft : https://www.ietf.org/archive/id/draft-ietf-rpp-core-00.html#name-availability-for-creation.

### Actual Behaviour

400 errors on HEAD request, successful routing on GET. Simply changing the path on either the @Head or @Get annotation makes both work as expected.

### Steps To Reproduce

    @Head("/{id}/availability")
    HttpResponse<?> availableSimple(@PathVariable("id") String id) {
        return HttpResponse.ok();
    }

    @Get("/{id}/availability")
    HttpResponse<?> availableAdditional(@PathVariable String id) {
        return HttpResponse.ok();
    }

    @Test
    void testHttp(RequestSpecification spec) {
        var response = spec
            .when().head("/domainNames/example.dummy/availability")
            .then().statusCode(200)
            .extract()
            .response();
    }

### Environment Information

MacOS
Java 25

### Example Application

_No response_

### Version

5.1.13

UsmanEjaz10:
Fixes #13020

## Summary

A controller that declares both an explicit `@Head` and a `@Get` for the **same URI** served `GET`
requests fine but answered `HEAD` with **400 Bad Request**:

```java
@Head("/{id}/availability")  HttpResponse<?> availableSimple(...) { ... }
@Get ("/{id}/availability")  HttpResponse<?> availableAdditional(...) { ... }
```

This PR stops the duplicate `HEAD` route from being registered in the first place. The fix is at
route-build time in `AnnotatedMethodRouteBuilder`; no runtime resolution logic changes.

## Root cause

`AnnotatedMethodRouteBuilder` auto-registers an **implicit `HEAD` route** for every `@Get` method
(so a bare GET handler also answers HEAD), in addition to the developer's explicit `@Head` route.
When both target the identical resolved URI, the route table ends up with **two `HEAD` routes whose
path templates are literally identical**. At request time `DefaultRouter.findClosest()` filters by
method and content negotiation, then reaches `resolveAmbiguity()` to break the tie — but equal
templates have equal specificity, nothing breaks the tie, and resolution throws
`DuplicateRouteException`, which the server translates into the observed 400.

## The fix

`router/.../AnnotatedMethodRouteBuilder.java`:

1. **Resolve the URI once** per `@Get` iteration (it was resolved twice with identical inputs) and
   reuse it for the GET route and the implicit-HEAD decision.
2. **Guard the implicit HEAD route** so it is only registered when the controller does not already
   declare an explicit `@Head` for that exact URI:

```java
if (method.booleanValue(Get.class, "headRoute").orElse(true)
        && !hasExplicitHeadMapping(bean, resolvedUri)) {
    route = HEAD(resolvedUri, bean, method).produces(produces);
    ...
}
```

3. **`hasExplicitHeadMapping(bean, resolvedUri)`** scans the bean's executable methods for an
   `@Head`-stereotyped method whose *resolved* URI equals the candidate's, honouring the
   `uris = [...]` multi-URI form on both annotations.

Why this is the right place:

- **Build time, not request time.** The duplicate is synthetic and should never exist. Suppressing
  it at registration keeps the route table honest; a runtime ambiguity tie-break would paper over
  ill-formed state and add heuristics to every ambiguous match.
- **Order-independent.** It inspects the bean's complete statically-computed method list, so the
  result does not depend on whether `process()` visits the `@Get` or `@Head` method first.
- **Per-URI.** The check sits *inside* the `uris` loop, so `@Get(uris = ["/a", "/b"])` with
  `@Head("/a")` suppresses the implicit HEAD only for `/a`; `/b` keeps its implicit HEAD.
- **`@Get(headRoute = false)` is unchanged** — it short-circuits before this check.
- **Scoped to one controller.** Two *different* controllers claiming the same URI is a separate,
  pre-existing ambiguity and is deliberately out of scope here.
- No content-negotiation change was needed: removing the duplicate is sufficient; the suspected
  `produces` mismatch between the two routes was a red herring.

A note on the helper signature: it takes `BeanDefinition<?>`, not the raw `BeanDefinition` used
elsewhere in the file. This is a **compile requirement**: calling the generic `getExecutableMethods()`
on a raw type erases its return to a raw `List`, so the enhanced-for element types as `Object` and
the loop over `ExecutableMethod<?, ?>` does not compile.

## Tests

`router/src/test/groovy/io/micronaut/context/router/HeadGetSameUriRouteBuilderSpec.groovy` — new
spec exercising the router the way the HTTP server does (`Router.findClosest(...)`), so the failure
mode is precise and no embedded server is needed. Four features:

| Scenario | Asserts |
|---|---|
| Explicit `@Head` + `@Get`, same URI | HEAD resolves (previously threw `DuplicateRouteException`) and invokes the **explicit `@Head`** method |
| Same controller, `GET` to the shared URI | `GET` still routes to the `@Get` method |
| `@Get(uris = ["/a", "/b"])` + `@Head("/a")` | `/a` → explicit `@Head`; `/b` → implicit HEAD **still registered** and routes to `@Get` (suppression is per-URI) |
| `@Get(headRoute = false)` + sibling `@Head`, same URI | Only the explicit `@Head` answers HEAD; `GET` still hits `@Get` |

Each case pins the winning handler via a distinct `X-Handler` response header, so a wrong-route
match fails loudly instead of silently passing.

## Documentation

- `Get.java`: the `headRoute()` javadoc now states that when the controller declares an explicit
  `@Head` route for the same URI, the implicit HEAD route is not registered for that URI.

## Verification

- `./gradlew :micronaut-router:test` — **60 tests, 0 failures** (full router module).
- New spec: 4/4 green (`JAVA_HOME` pointing at a JDK 25; the build requires JVM 25+).
- Pre-fix behaviour confirmed: without the guard the new spec's first feature throws
  `DuplicateRouteException`.

### DIFF
--- a/http/src/main/java/io/micronaut/http/annotation/Get.java
+++ b/http/src/main/java/io/micronaut/http/annotation/Get.java
@@ -97,7 +97,9 @@ public @interface Get {
     boolean single() default false;
 
     /**
-     * @return True if a HEAD route should also be registered for the same method
+     * @return True if a HEAD route should also be registered for the same method. If the
+     * controller declares an explicit {@link Head} route for the same URI, the implicit
+     * HEAD route is not registered for that URI and the explicit one takes precedence.
      */
     boolean headRoute() default true;
 }
--- a/router/src/main/java/io/micronaut/web/router/AnnotatedMethodRouteBuilder.java
+++ b/router/src/main/java/io/micronaut/web/router/AnnotatedMethodRouteBuilder.java
@@ -81,9 +81,8 @@ public class AnnotatedMethodRouteBuilder extends DefaultRouteBuilder implements 
             Set<String> uris = this.resolveUrisMapping(Get.class, method);
             for (String uri: uris) {
                 MediaType[] produces = resolveProduces(method);
-                UriRoute route = GET(resolveUri(bean, uri,
-                        method,
-                        uriNamingStrategy),
+                String resolvedUri = resolveUri(bean, uri, method, uriNamingStrategy);
+                UriRoute route = GET(resolvedUri,
                         bean,
                         method).produces(produces);
 
@@ -93,10 +92,15 @@ public class AnnotatedMethodRouteBuilder extends DefaultRouteBuilder implements 
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("Created Route: {}", route);
                 }
-                if (method.booleanValue(Get.class, "headRoute").orElse(true)) {
-                    route = HEAD(resolveUri(bean, uri,
-                            method,
-                            uriNamingStrategy),
+                // Only auto-create an implicit HEAD route for this GET method if the
+                // controller does not already declare an explicit @Head method for the
+                // exact same URI. Otherwise, two HEAD routes would match the same request
+                // (this implicit one and the explicit one), and since neither is more specific
+                // than the other, the router cannot resolve the ambiguity and rejects the
+                // request as a duplicate route (see DuplicateRouteException / HTTP 400).
+                if (method.booleanValue(Get.class, "headRoute").orElse(true)
+                        && !hasExplicitHeadMapping(bean, resolvedUri)) {
+                    route = HEAD(resolvedUri,
                             bean,
                             method).produces(produces);
                     if (definition.port > -1) {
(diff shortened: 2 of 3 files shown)


graemerocher:
Resumed in https://github.com/micronaut-projects/micronaut-core/pull/13030

PR #13030, opened 2026-09-07
Prefer an explicit @Head route over the implicit one derived from @Get

### LINKED ISSUE #13020: GET and HEAD route conflict

graemerocher:
Fixes #13020. Alternative to #13027.

## Summary

A controller that declares both an explicit `@Head` and a `@Get` for the same URI served `GET`
fine but answered `HEAD` with `400 Bad Request`:

```java
@Head("/{id}/availability") HttpResponse<?> availableSimple(@PathVariable("id") String id) { ... }
@Get ("/{id}/availability") HttpResponse<?> availableAdditional(@PathVariable String id)   { ... }
```

`AnnotatedMethodRouteBuilder` registers an implicit `HEAD` route for every `@Get` method, so that a
bare `GET` handler also answers `HEAD`. Alongside the user's explicit `@Head` route, two `HEAD`
routes match the request with identical path templates and identical specificity. Nothing breaks the
tie, `DefaultRouter.findClosest` throws `DuplicateRouteException`, and the server turns that into the
observed 400.

## The fix

Mark the implicit route as implicit, and let route resolution prefer the explicit one when — and
only when — it is about to give up:

* `UriRouteInfo#isImplicitHead()` (new `default` method, `false`) reports whether a `HEAD` route was
  derived from a `@Get` mapping rather than declared with `@Head`. `AnnotatedMethodRouteBuilder`
  flags the route it synthesises; `DefaultUrlRouteInfo` carries the flag.
* `ImplicitHeadRoutes.preferExplicit(...)` drops implicit candidates from a set of otherwise
  indistinguishable matches, provided at least one explicitly declared route survives.
* It is applied in `DefaultRouter#findClosest` and in the `Router#findClosest` default (used by
  `FilteredRouter`) at the one place that would otherwise throw `DuplicateRouteException`.

`AnnotatedMethodRouteBuilder` also now resolves the URI once per `@Get` iteration instead of twice
with identical inputs.

## Why a tie-break rather than not registering the route

#13027 suppresses the implicit `HEAD` route at build time whenever the bean declares an explicit
`@Head` for the same resolved URI. Copilot flagged that this matches on the URI alone and may remove
implicit routes that differ by other qualifiers. That is not hypothetical — two `HEAD` routes for the
same URI are frequently **not** ambiguous, because earlier stages of resolution already choose
between them, and both must stay individually reachable.

Running this PR's specs against #13027 turns those cases into 404s:

| Controller | Request | main | #13027 | this PR |
|---|---|---|---|---|
| `@Head` produces `text/plain`, `@Get` produces `application/json` | `HEAD` + `Accept: application/json` | → `@Get` | **no route** | → `@Get` |
| same | `HEAD` + `Accept: text/plain` | → `@Head` | → `@Head` | → `@Head` |
| same | `HEAD`, no preference | **400** | → `@Head` | → `@Head` |
| `@Version("2") @Head`, `@Version("1") @Get` | `HEAD` + `X-API-VERSION: 1` | → `@Get` | **no route** | → `@Get` |
| same | `HEAD` + `X-API-VERSION: 2` | → `@Head` | → `@Head` | → `@Head` |

Content negotiation runs in `DefaultRouter.resolveAmbiguity` and version filtering runs in
`FilteredRouter` *after* matching, so neither can compensate for a route that was never registered.
Deferring the decision to the point of failure keeps the implicit route available to every earlier
discriminator, and the ambiguity is only ever broken when the request genuinely cannot be resolved.

Note that placing this *inside* `resolveAmbiguity` would reintroduce the versioning regression, since
`findAllClosest` feeds `FilteredRouter` — hence the two explicit call sites rather than one.

Copilot's second point, the startup-time scan per `@Get` route, does not arise: there is no scan.
The flag is set where the route is already being constructed, and the tie-break allocates nothing
unless a request would otherwise have been rejected as a duplicate.

## Behaviour that is deliberately unchanged

* `@Get(headRoute = false)` still registers no implicit `HEAD` route.
* Two genuinely duplicate routes still raise `DuplicateRouteException`; the tie-break does not apply
  when every remaining candidate is implicit (or none is).
* `findAllClosest` is untouched, so consumers that legitimately want every matching route — CORS
  preflight, `Allow` computation — keep seeing both.

## Tests

`router/src/test/groovy/io/micronaut/context/router/ExplicitHeadRouteSpec.groovy` (7 features) and
`ExplicitHeadRouteVersioningSpec.groovy` (3 features) exercise `Router.findClosest(...)` the way the
HTTP server does, so no embedded server is needed. Each case pins the winning handler with a distinct
`X-Handler` header, so a wrong-route match fails loudly rather than passing silently. Coverage:

* the reported case — explicit `@Head` + `@Get` on the same URI, both directions
* `@Get(uris = ["/a", "/b"])` + `@Head("/a")` — `/b` keeps its implicit `HEAD`
* `@Get(headRoute = false)` beside an explicit `@Head`
* differing `produces`, all three `Accept` outcomes
* differing `@Version`, both versions
* genuinely ambiguous routes still reported as duplicates
* only the `@Get`-derived route is flagged `isImplicitHead()`

Against unpatched `5.2.x` four of these fail; against #13027 the `produces` and version-1 cases fail.

## Verification

JDK 25, `5.2.x` at `3dc8ea0286`:

* `:micronaut-router:test` — 66 tests, 0 failures
* `:micronaut-router:checkstyleMain/checkstyleTest`, `:micronaut-http:checkstyleMain` — clean
* `:micronaut-http-server:test` — green
* `:micronaut-http-server-netty:test` — 1133 tests, 2 failures, both pre-existing and unrelated
  (`BroadcasterSpec` fails identically on a clean tree; `StreamPressureSpec` is timing-flaky)

---

https://claude.ai/code/session_01B7GKVDfQZ8vg11dGQhNCJt

### DIFF
--- a/router/src/main/java/io/micronaut/web/router/ImplicitHeadRoutes.java
+++ b/router/src/main/java/io/micronaut/web/router/ImplicitHeadRoutes.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.web.router;
+
+import io.micronaut.core.annotation.Internal;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Tie-break used as the very last step of route resolution, once every other discriminator
+ * (media types, versioning, URI specificity, ...) has been applied and more than one route still
+ * matches the request.
+ *
+ * <p>A {@code @Get} mapping also registers an implicit {@code HEAD} route so that a bare {@code GET}
+ * handler answers {@code HEAD} requests too. When the same controller additionally declares an
+ * explicit {@code @Head} route for the same URI, both routes match a {@code HEAD} request with equal
+ * specificity and the request would otherwise be rejected with a
+ * {@link io.micronaut.web.router.exceptions.DuplicateRouteException} (surfaced as a 400 response).
+ * The user's explicit declaration is the more specific intent, so it wins.</p>
+ *
+ * <p>Applying this only at the point of failure — rather than suppressing the implicit route when it
+ * is built — keeps the implicit route available to every earlier discriminator. Two routes for the
+ * same URI that differ by {@code produces} or by {@code @Version} are not ambiguous at all, and both
+ * remain individually reachable.</p>
+ *
+ * @author Graeme Rocher
+ * @since 5.2.0
+ */
+@Internal
+final class ImplicitHeadRoutes {
+
+    private ImplicitHeadRoutes() {
+    }
+
+    /**
+     * Drops implicit {@code HEAD} routes from a set of otherwise indistinguishable matches, provided
+     * at least one explicitly declared route remains.
+     *
+     * @param matches The ambiguous matches, never empty
+     * @param <T>     The target type
+     * @param <R>     The result type
+     * @return The explicitly declared matches, or {@code matches} unchanged if the tie-break does not apply
+     */
+    static <T, R> List<UriRouteMatch<T, R>> preferExplicit(List<UriRouteMatch<T, R>> matches) {
+        int implicitCount = 0;
+        for (UriRouteMatch<T, R> match : matches) {
+            if (match.getRouteInfo().isImplicitHead()) {
+                implicitCount++;
+            }
+        }
+        if (implicitCount == 0 || implicitCount == matches.size()) {
+            // nothing implicit to drop, or every candidate is implicit: leave the ambiguity untouched
+            return matches;
+        }
+        List<UriRouteMatch<T, R>> explicit = new ArrayList<>(matches.size() - implicitCount);
+        for (UriRouteMatch<T, R> match : matches) {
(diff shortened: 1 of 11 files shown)


router/src/main/java/io/micronaut/web/router/ImplicitHeadRoutes.java
```diff
@@ -0,0 +1,77 @@
...
+            return matches;
+        }
+        List<UriRouteMatch<T, R>> explicit = new ArrayList<>(matches.size() - implicitCount);
+        for (UriRouteMatch<T, R> match : matches) {
+            if (!match.getRouteInfo().isImplicitHead()) {
+                explicit.add(match);
+            }
+        }
+        return explicit;
+    }
```
yawkat, line 76: I think you can replace this with ArrayList.removeIf, it already has an implicitCount==0 optimization and it makes the code more clear.

yawkat: Hmm or maybe not, this method bails when all routes are implicit…

yawkat (approved):
Please add a TCK test

PR #13032, opened 2026-09-07
Honour the route's executor for Kotlin suspend functions

### LINKED ISSUE #12326: Suspend functions ignore `ExecuteOn` and use the event loop instead

dstepanov:
Supersedes #12869 and #12594, rebased onto `5.2.x` and extended. Fixes #12326.

Thanks to @rawilder and @jpbempel — the diagnosis and the core of the fix are theirs, and they are credited as co-authors on the commit.

## The bug

A Kotlin `suspend` route ignored `@ExecuteOn` and `micronaut.server.thread-selection`. `RouteInfo.getExecutor` picked an executor, `RouteExecutor.callRoute` then dropped it on the floor for the suspended branch, and the route ran on the event loop and resumed on `Dispatchers.Default`.

## The fix

Two changes are needed, and a partial fix looks plausible but doesn't work. A suspend function runs **synchronously on its caller** until the first suspension point, and its coroutine context governs only where it *resumes*:

- **`applyExecutorToPublisher`** around the deferred `Mono` moves the body off the event loop up to the first suspension point. Changing only the dispatcher would leave that part on the event loop.
- **The executor becomes the coroutine's dispatcher**, so the continuation — and the child coroutines that inherit it — stay there afterwards. `subscribeOn` alone would run the body on IO and then resume on `Dispatchers.Default`.

The no-executor path is unchanged: `dispatcher ?: Dispatchers.Default`, and the WebSocket caller keeps the 3-arg overload.

## Differences from #12869

- **The dispatcher is resolved once per executor, not per request.** `Executor.asCoroutineDispatcher()` returns a fresh `ExecutorCoroutineDispatcherImpl` on every call. Per request that allocates on the request path, re-runs the `ScheduledThreadPoolExecutor` reflection probe in its `init`, and — because dispatcher equality is identity — stops `withContext` from recognising the route's own dispatcher as the one already in effect. The cache lives on the `CoroutineHelper` singleton, so it is bound to the application context that owns the executors and cannot outlive them.
- **The dispatcher is built from the raw executor, not the instrumented wrapper.** `ExecutorServiceInstrumenter` wraps every `ExecutorService` bean, so what `getExecutor()` returns is a `ContextPropagatingExecutorService`. Coroutine propagation is already handled by `MicronautPropagatedContext` (a `ThreadContextElement`); dispatching through the wrapper would *additionally* capture the propagated context of whichever thread happened to resume the continuation and push it as an outer scope. `CoroutineHelper` now unwraps first, matching what `applyExecutorToPublisher` already does a few lines up.
- **`publishOn` is kept, and the reason is now in a comment.** It holds the same "the response is assembled off the event loop" invariant that the blocking branch gets from `ExecutionFlow.async`.
- **The `thread-selection` behaviour change is documented**, in the guide and in the breaking-changes appendix.
- License headers on the new test files, plus two new tests (below).

## Behaviour change

`@ExecuteOn` and `micronaut.server.thread-selection` resolve through the same `RouteInfo.getExecutor`, so this also makes a `suspend` route obey `AUTO`, `IO` and `BLOCKING` selection.

The alternative was to special-case suspend inside `DefaultExecutorSelector` and treat it as non-blocking under `AUTO`. I went the other way: the selector also feeds `@ExecuteOn` and the AOP `ExecuteOnInterceptor`, so Kotlin-specific knowledge doesn't belong in it, and it would leave `AUTO` offloading `fun foo(): String` but not `suspend fun foo(): String` — `suspend` is no compile-time promise the body doesn't block. `@NonBlocking` already exists for opting a route back onto the event loop, and the selector honours it.

The default is `MANUAL`, which selects no executor, so applications on the default are unaffected. Documented in `breaks.adoc`.

## Tests

`test-suite-kotlin/.../http/server/executeon/` — 14 tests:

- `SuspendExecuteOnTest` — `@ExecuteOn` on suspend and blocking routes, and that a route without it stays on the event loop.
- `SuspendResumptionTest` — thread identity *across* a suspension point, child coroutine inheritance, `withContext` winning and then restoring the route's dispatcher, that the dispatcher is the **same instance across two requests** (the caching fix), and that the request is **still propagated after resuming** on the executor (the unwrapping fix).
- `SuspendThreadSelectionTest` — `AUTO`/`IO`/`BLOCKING` offload both route shapes; `MANUAL` leaves both on the event loop.

Every thread-name constant is pinned by at least one *positive* assertion, so the negative assertions can't pass vacuously from a typo'd prefix.

Verified locally: 14/14 new tests pass, and a regression sweep over the existing suspend, coroutine and `ThreadSelectionSpec` tests is 85 tests / 0 failures.

## Note on 4.10.x

#12869 targeted `4.10.x`; this targets `5.2.x`. If the fix is wanted in a 4.10 patch, that needs a separate back-port — but note the `thread-selection` behaviour change above, which is easier to justify on a minor line than in a patch.

### DIFF
--- a/http-server/build.gradle.kts
+++ b/http-server/build.gradle.kts
@@ -15,6 +15,8 @@ dependencies {
     annotationProcessor(projects.micronautInjectJava)
 
     testImplementation(libs.managed.netty.codec.http)
+    testImplementation(libs.managed.kotlinx.coroutines.core)
+    testImplementation(projects.micronautContextPropagation)
 
     testAnnotationProcessor(projects.micronautInjectJava)
     testAnnotationProcessor(platform(libs.test.boms.micronaut.validation))
--- a/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
+++ b/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
@@ -76,7 +76,14 @@ public final class CoroutineHelper {
         ContinuationArgumentBinder.setupCoroutineContext(httpRequest, contextView, propagatedContext, coroutineContextFactories, dispatcher);
     }
 
-    private CoroutineContext dispatcherFor(ExecutorService executorService) {
+    /**
+     * Resolves, and caches, the dispatcher for an executor. Package private so that the caching and unwrapping
+     * can be asserted directly.
+     *
+     * @param executorService The executor the route runs on
+     * @return The dispatcher for that executor
+     */
+    CoroutineContext dispatcherFor(ExecutorService executorService) {
         // the coroutine's own propagation is handled by KotlinCoroutinePropagation, so dispatch on the raw
         // executor rather than the instrumented wrapper, which would otherwise capture the propagated context
         // of whichever thread happened to resume the continuation
--- a/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
+++ b/http-server/src/main/java/io/micronaut/http/server/CoroutineHelper.java
@@ -16,7 +16,9 @@
 package io.micronaut.http.server;
 
 import io.micronaut.context.annotation.Requires;
+import io.micronaut.context.propagation.instrument.execution.ContextPropagatingExecutorService;
 import io.micronaut.core.annotation.Internal;
+import io.micronaut.core.annotation.Nullable;
 import io.micronaut.core.propagation.PropagatedContext;
 import io.micronaut.http.HttpRequest;
 import io.micronaut.http.bind.binders.ContinuationArgumentBinder;
@@ -25,6 +27,10 @@ import jakarta.inject.Singleton;
 import reactor.util.context.ContextView;
 
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import kotlin.coroutines.CoroutineContext;
 
 /**
  * Coroutines helper.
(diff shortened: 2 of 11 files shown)


graemerocher (approved):

dstepanov:
Pushed 670b48d — the SonarCloud failure was the new-code coverage gate (51.9% vs 80%), not an issue with the change.

The cause is that `test-suite-kotlin` contributes no coverage: the root build drops every `:test-` project from the jacoco aggregation. So the lines only a Kotlin suspend route can reach counted as uncovered even though they are covered by the suite.

Added `CoroutineHelperTest` in `http-server`, asserting the two behaviours this change turns on directly — that the dispatcher is resolved once per executor, and that an instrumented executor resolves to its target's dispatcher rather than the wrapper's. `dispatcherFor` is package private so the test can reach it. That leaves only the four lines of the suspend branch in `RouteExecutor` uncounted, which genuinely need a Kotlin route to exercise.

PR #13034, opened 2026-09-07
docs: fix guide errors and fill in missing language snippets

graemerocher:
## What

A review of `src/main/docs` for errors and inconsistencies, plus snippet coverage for all four supported languages (Java, Groovy, Kotlin, Python).

## Errors fixed

**Broken snippet markup**

- `snippet::io.micronaut.docs.server.suspend.SuspendController.kt` in the Kotlin coroutines section carried a file extension on the FQCN, so it resolved in no language and the block rendered empty.
- Two macros in `routing.adoc` were missing a comma before `title=`; `contextEvents.adoc` had an unterminated quote in a `title` attribute.
- **43 sources across the four test suites had unbalanced snippet tags** — mostly `// tag::class[]` written where `// end::class[]` was meant, plus `// tag::end[]`, `// end::method` (no brackets) and `// end::endclass]`. Asciidoctor treats an unterminated tag as "include to end of file". No unbalanced tags remain anywhere in the repo.
- 25 snippets referenced tags that did not exist in their source, rendering an empty code block for that language.

**Table of contents**

- `httpServer/form/binding.adoc` — 7KB of finished content from #12246 — was never listed in `toc.yml`, so it has never been published. Now included, renamed to `formBinding` because `binding` was already taken and the guide build rejects duplicate section names.
- `deadlockedThreads.adoc` existed in two places after the health docs split (#11730); the stale copy is removed. `discoveryClientHealthIndicator` was missing from the TOC entirely.
- `toc-python.yml` referenced two long-deleted Netflix Hystrix/Ribbon pages, and omitted three Python-only pages (`standardTypes`, `moduleScripts`, `graalTools`) from the Python guide.
- Removed eight dead, unreferenced files: an empty `config/validation.adoc`, a `customTypeConverters.adoc` whose include target no longer exists, and obsolete Spring Loaded / RabbitMQ / MySQL / Spring stubs.

**Content**

- `endpoints.health.deadlocked-thread.enabled` → `deadlocked-threads.enabled`, verified against `DeadlockedThreadsHealthIndicator`.
- "All provided health indicators are exposed on `/health/readiness`" — the deadlocked-threads one is `@Liveness`.
- Broken `<<introspections>>` xref, `rodLength=7B.0` → `7.0`, a callout naming `AnnotationContextExample` instead of `AnnotationMemberContext`, "and pas" → "and passing", `Micronaut-context` → `micronaut-context`, two unclosed backticks, and a `---` used as an open block delimiter.

## Language coverage

Snippet targets carrying all four languages: **299 → 328 of 367**.

New sources cover Kotlin (injection, filter methods, CORS, plain text responses, task executors, forms, bean mappers, context propagation, introspection builders, `RouteMatch`, resources, byte body), Groovy (byte body, MDC propagation, forms, VAT config) and Python (`@DependsOn` lifecycle, introspection configuration).

Where a language genuinely cannot express the feature, the prose is scoped with language blocks rather than leaving an orphaned callout list — `package-info` (Java/Groovy only), the programmatic `RouteBuilder` API (Python declares routes as module functions) and field access kinds. The Python-flavoured guide no longer has a page whose snippet silently disappears, which previously left dangling callouts.

## Left open

Two gaps are framework limitations rather than documentation defects, so they are flagged rather than papered over:

- A Python `@Creator` classmethod produces an introspection that throws `NoSuchMethodError` on `instantiate()`. This affects the already-published `staticAtCreator` example (`Business.py`), confirmed with a probe test.
- KSP ignores `@AnnotationExpressionContext` on an annotation *member* (type level works), so `evaluatedExpressions` has no Kotlin tab.

The single remaining missing tag (`constructors` in Python's `IntrospectionSpec`) is downstream of the first: Python has one `__init__`, and the `@Creator` workaround hits exactly that error.

## Verification

`./gradlew :test-suite:test :test-suite-groovy:test :test-suite-kotlin:test publishGuide` and `./gradlew -Ppython-ci=true :test-suite-python:test` all pass from clean.

Reviewed with Codex, which raised two findings — new Python snippets whose tagged region excluded imports the displayed code needs (`@Introspected(classes=Person)` and `@DependsOn(MessagePublisher)` would both have been unbound in copied code). Both are fixed; unlike Java, Python needs an explicit relative import for a same-package name.


src/main/docs/guide/ioc/introspection/atIntrospected.adoc
```diff
@@ -12,10 +12,21 @@ snippet::io.micronaut.docs.ioc.beans.IntrospectionSpec[tags="usage", indent=0]
...
+[.lang-java.lang-groovy]
+--
 WARNING: The package should contain only classes that can be introspected.
 
 NOTE: Only classes located directly within the package are processed; subpackages are ignored.
+--
+
+[.lang-kotlin.lang-python]
+--
+NOTE: Package-level introspection requires a `package-info` declaration, which only Java and Groovy support. In Kotlin and Python, annotate each class with ann:core.annotation.Introspected[], or list the types with the `classes` member as shown in <<introspectedAnnotationOnAConfigurationClass, Use the @Introspected Annotation on a Configuration Class>>.
```
dstepanov, line 28: This xref is broken in the Python guide: `introspectedAnnotationOnAConfigurationClass` is in `toc.yml` but **not** in `toc-python.yml`, so for the `.lang-python` half of this block the link has no target — the dangling-reference failure mode this PR sets out to remove.

Same root cause hides two of the new Python sources added here:

- `test-suite-python/.../ioc/beans/PersonConfiguration.py` is the snippet for `introspectedAnnotationOnAConfigurationClass`
- `test-suite-python/.../ioc/beans/Vehicle.py` is the snippet for `atCreator`

Neither page is in `toc-python.yml`, so both files are written but never published.

Adding both keys under `ioc.introspection` in `toc-python.yml` fixes the xref and makes the two new sources reachable.

(Related, smaller: `accessKind.adoc` was reworded to "not possible in Kotlin or Python", but `accessKind` is also absent from `toc-python.yml`.)

---

src/main/docs/guide/ioc/introspection/atIntrospected.adoc
```diff
@@ -12,10 +12,21 @@ snippet::io.micronaut.docs.ioc.beans.IntrospectionSpec[tags="usage", indent=0]
 <4> Use the `set` method to set the property value
 <5> Use the `get` method to retrieve the property value
 
+[.lang-java.lang-groovy]
+--
 It's possible to apply the introspection to all the classes in one package. Simply create `package-info` file and annotate the package with ann:core.annotation.Introspected[].
+--
 
 snippet::io.micronaut.docs.ioc.introspection.pck.foobar.package-info[tags="usage", indent=0]
```
dstepanov, line 19: The prose above and below is now scoped to `[.lang-java.lang-groovy]`, but the snippet between them is not — and `package-info` sources exist only for Java and Groovy:

```
test-suite/.../ioc/introspection/pck/foobar/package-info.java
test-suite-groovy/.../ioc/introspection/pck/foobar/package-info.groovy
```

So for Kotlin/Python this renders an empty code block sitting between two blocks of prose that are correctly hidden — the same "snippet resolved in no language, block renders empty" symptom fixed elsewhere in this PR. Wrapping the snippet in the same `[.lang-java.lang-groovy]` block would make it consistent.

---

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/i18n/MessageSourceFactory.kt
```diff
@@ -1,14 +1,18 @@
...
 import jakarta.inject.Singleton
+//end::imports[]
 
+//tag::clazz[]
 @Factory
 internal class MessageSourceFactory {
     @Singleton
-    fun createMessageSource(): MessageSource = ResourceBundleMessageSource("io.micronaut.docs.i18n.messages")
+    fun createMessageSource(): MessageSource =
+        ResourceBundleMessageSource("io.micronaut.docs.i18n.messages", Ordered.HIGHEST_PRECEDENCE)
```
dstepanov, line 19: Aligning with the Java version on `Ordered.HIGHEST_PRECEDENCE` is right for the docs, but the Java source also carries a guard that wasn't ported here (or to the Groovy file):

```java
@Requires(property = "spec.name", value = "I18nSpec")
//tag::clazz[]
@Factory
class MessageSourceFactory {
```

Without it this is an unconditional `@Factory` publishing a highest-precedence `MessageSource` into every test in `test-suite-kotlin` (and `test-suite-groovy`). Only `I18nSpec` consumes it today so nothing breaks now, but it's a trap for the next test in either suite that resolves a `MessageSource`. Placing the annotation above `//tag::clazz[]` keeps it out of the published snippet, exactly as Java does.

---

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/ioc/injection/field/Vehicle.kt
```diff
@@ -0,0 +1,20 @@
+package io.micronaut.docs.ioc.injection.field
+
+import jakarta.inject.Inject
+import jakarta.inject.Singleton
+
+@Singleton
+internal class Vehicle {
```
dstepanov: Cosmetic: these four new files (`ctor`, `field`, `method`, `nullable`) are untagged, so the whole file — `internal` modifier included — is what readers see. The Java counterparts are package-private, which renders as nothing:

```java
@Singleton
class Vehicle {
    @Inject Engine engine; // <1>
```

`internal class Vehicle` in a first-look dependency-injection example is noise that the reader has to decide whether to copy. Dropping `internal` matches the Java output; otherwise these mirror the Java layout (two classes per file, no license header) exactly, which is good.

---

dstepanov (commented):
I verified the mechanical claims rather than taking the description at face value, by scripting the checks against both `5.2.x` and the branch.

**Reproduced and confirmed:**

| Claim | Result |
|---|---|
| No unbalanced snippet tags remain | 0 imbalances across all four suites; also 0 malformed tags (`end::method` with no `[]`, `end::endclass]`) |
| 25 snippets referenced non-existent tags | Exactly 25 on `5.2.x`, **1** on the branch — the documented Python `IntrospectionSpec[constructors]` gap |
| `deadlocked-thread` → `deadlocked-threads` | `DeadlockedThreadsHealthIndicator:48` |
| Deadlocked-threads is a liveness indicator | `@Liveness` at line 46; every other indicator in that table is `@Readiness` |
| `rodLength=7B.0` → `7.0` | matches `VehicleSpec.java:43` |
| `toc.yml` complete | every key resolves to an `.adoc`, no duplicate keys, no orphan `.adoc` files |
| Main-guide xrefs | all resolve (the two apparent misses are Asciidoctor auto-generated section ids) |

The `binding` → `formBinding` rename is the right call — `binding` is the request-binding page and the three existing `<<binding>>` references still land there. Dropping `endclass` from `transfers.adoc` is also correct: the Java/Groovy/Python `DownloadController` only ever had a `class` tag. The Kotlin `IssuesController` and `TraceFilter` restructuring now mirrors the Java tags and callout numbering exactly. Removals I spot-checked (`end::props[]` in `PaginationCommand.groovy`, `end::rxImport[]` in `LocalFunctionInvokeSpec.java`, `tag::start[]` in `config/itfce/VehicleSpec.kt`) were genuine orphans rather than the surviving half of a pair.

**Findings inline.** One is worth fixing before merge — a new xref in `atIntrospected.adoc` points at a page that isn't in `toc-python.yml`, so it dangles in exactly the Python guide the surrounding block was written for, and the same gap leaves two of this PR's new Python sources unpublished. The other three are optional.

---

src/main/docs/guide/ioc/introspection/atIntrospected.adoc
graemerocher: Fixed in c9739dd10c — added `introspectedAnnotationOnAConfigurationClass` to `toc-python.yml`. Verified against the actual generated Python guide (`./gradlew publishGuidePython`): the anchor is now present and `PersonConfiguration.py` renders, along with `MessageConsumer.py`.

On `atCreator` I went a different way rather than adding the key. The page opens "For classes with multiple constructors" — a premise Python cannot satisfy, since a class has one `__init__`. My `Vehicle.py` was really demonstrating `staticAtCreator` (already in the Python guide via `Business.py`), and via a pattern that throws `NoSuchMethodError` on `instantiate()` today. Publishing it would document something that does not work. So `Vehicle.py` is removed and `atCreator.adoc` gets a `[.lang-python]` note pointing at `staticAtCreator` — the same treatment as `package-info` and `accessKind`.

`accessKind` I left out of `toc-python.yml` deliberately: it documents a Java/Groovy-only feature, so its absence from the Python guide is correct. The reworded note still earns its place in the combined guide, where a reader with the Python tab selected does see the page.

One thing surfaced while checking this: the Python guide has 9 other dangling anchors on `5.2.x` — `annotationMetadata`, `graal`, `ideSetup`, `interceptorLifecycle`, `jspecify`, `kapt`, `ksp`, `proxy`, `serializationUsingJacksonDatabind`. All pre-existing and all pointing at Java/Kotlin-only pages that `toc-python.yml` omits by design. Left alone here as out of scope, but happy to take it as a follow-up.

src/main/docs/guide/ioc/introspection/atIntrospected.adoc
graemerocher: Fixed in c9739dd10c — the snippet now sits inside the same `[.lang-java.lang-groovy]` open block as the prose.

Confirmed against the rendered HTML rather than just the source, since a block macro inside an open block was the part I was unsure of. The generated structure is now:

```
div.openblock.lang-java.lang-groovy
  div.paragraph
  div.listingblock.multi-language-sample  -> code.language-java
  div.listingblock.multi-language-sample  -> code.language-groovy
  div.admonitionblock.warning
  div.admonitionblock.note
div.openblock.lang-kotlin.lang-python
  div.admonitionblock.note
```

Only Java and Groovy samples are emitted, so there is no empty block left for Kotlin/Python.

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/i18n/MessageSourceFactory.kt
graemerocher: Good catch — fixed in c9739dd10c for both Kotlin and Groovy, with the annotation above `//tag::clazz[]` so it stays out of the published snippet, exactly as Java does. The two specs now carry `@Property(name = "spec.name", value = "I18nSpec")` to match.

You are right that nothing breaks today and it was a trap rather than a live bug: the factory was `@Factory`-annotated with no condition, so both suites had a `HIGHEST_PRECEDENCE` `MessageSource` in every context. Worth noting the precedence only became unconditional-and-highest in this PR — I had added `Ordered.HIGHEST_PRECEDENCE` to align with Java, which made the missing guard matter more than it did before.

test-suite-kotlin/src/test/kotlin/io/micronaut/docs/ioc/injection/field/Vehicle.kt
graemerocher: Fixed in c9739dd10c — dropped `internal` from all four (`ctor`, `field`, `method`, `nullable`), so the rendered Kotlin now opens with `class Vehicle {` and lines up with the package-private Java output.

dstepanov (approved):

src/main/docs/guide/ioc/introspection/atIntrospected.adoc
graemerocher: Followed up on the 9 pre-existing dangling anchors in `393a02cd1e`.

First, a correction to what I reported earlier: my original count compared raw `href="#…"` against `id="…"` in the generated HTML, which over-counts. Anchors inside a `[.lang-…]` block that excludes Python are hidden by CSS, so they are not a visible break. Re-checking with an HTML parser that tracks the enclosing `lang-` classes gives 11 genuinely visible broken links across 8 anchors — the `ioc.adoc` reference to `annotationMetadata` was already correctly scoped and needed nothing.

They split into two groups.

**Published in `toc-python.yml`** — target is language-agnostic, carries no snippets, and was simply never carried over: `aop/interceptorLifecycle` (a plain sibling of `lifecycleAdvice`, which the Python guide already lists), `cli/proxy`, `httpServer/jsonBinding/serializationUsingJacksonDatabind` (linked 3× from `breaks.adoc`), and `languageSupport/graal` + `graalServices`.

**Fixed on the referencing side** — target is Java/Kotlin-only or deliberately excluded:

- `compilerArch.adoc`: KAPT and KSP now link to kotlinlang.org. The Java and Groovy bullets in that same list already use external links, so the three are now symmetric.
- `breaks.adoc`: JSpecify links to jspecify.dev. The sentence is about the standard, and `ioc/nullabilityAnnotations` opens *"In Java, you can use annotations"*, so it does not belong in the Python guide.
- `annotationArch.adoc`: dropped the cross-reference to `ioc/annotationMetadata`. That page cannot join the Python guide as it stands — 4 of its 8 snippets have no Python source and 2 of those carry callout lists that would be orphaned. Scoping is not an option either: the affected regions are whole `===` sections, and a heading cannot sit inside an open block. The bullet keeps its `@Retainable` API link.
- `quickStart.adoc`: scoped the IDE recommendation. `ideSetup` and its children are *commented out* of `toc-python.yml`, so that exclusion is deliberate — the Python lane now gets its own sentence rather than a dead link.

Verified by parsing the generated HTML for every guide variant and resolving each internal href against the ids actually present, ignoring anchors hidden by language scoping:

```
[all    ] visible dangling anchors: 0
[groovy ] visible dangling anchors: 0
[java   ] visible dangling anchors: 0
[kotlin ] visible dangling anchors: 0
[python ] visible dangling anchors: 0
```

The Java guide keeps its internal `#ideSetup` link, so nothing regressed for the other lanes.

PR #13040, opened 2026-09-07
Describe the members of a property by the type declaring them

dstepanov:
micronaut-validation ([micronaut-projects/micronaut-validation#631](https://github.com/micronaut-projects/micronaut-validation/pull/631)) validates over generated bean introspections and attributes a constraint to the element declaring it: an interface's constraints belong to the implicit group of that interface, `ElementDescriptor.getConstraintDescriptors()` is scoped to the declaring element, the `TraversableResolver` is asked with the member kind, and a field-level constraint is validated against the field's value. A generated introspection reported a property with the annotations of its field, its getter and every super type merged into one metadata, so the validator could read that only from the reflective `ReflectiveIntrospection` of the `micronaut-reflection` module, and the introspection profile of its Jakarta Validation TCK excluded the tests that need it.

### What changes

- **`@Introspected(members = true)` describes every declaration.** `BeanProperty.getMembers()` now lists the field of the property with the fields of the same name it hides in the super classes, and the read and write method of every type of the hierarchy declaring one - the bean type first, then its super classes, then its interfaces - each reported by `getDeclaringType()` and carrying the annotations of its own declaration only (`MethodElement.getDeclaredMethodAnnotationMetadata()`). The declarations of an accessor are the methods it overrides plus the methods of the same signature the hierarchy declares beside it, since an interface inheriting an accessor from two parent interfaces without redeclaring it overrides neither. A getter of a super type is read through the getter overriding it, the same virtual call, so the hierarchy adds no dispatch target; a hidden field is read through the class declaring it. The ordinary field dispatch keeps naming the owning type and names the field on it as javac names a field on its qualifying type, which fixes an `IllegalAccessError` when an introspection generated in another package read a field inherited from a package-private super class; a hidden field whose declaring class the introspection cannot name is listed but not readable, and a member's declaring type is carried as a class value falling back to the name so such an introspection loads (the previous `Class` constructor of the member reference is kept). The members are additive: a spec in the Java, the Groovy and the Kotlin module compiles the same hierarchy with and without the members and finds every metadata the previous API answers - the introspection, the properties and their arguments, the bean methods with their return types and arguments, the constructor - the same. `IntrospectionMetadataShape` of `micronaut-inject-test-utils` renders those answers, so the comparison is written once and no expected metadata is checked in.
- **`BeanIntrospection.separatesDeclarations()`** says whether an introspection tells the declarations apart: a generated one answers true when compiled with the members, a `ReflectiveIntrospection` always does. The validator's `ReflectionSupport.separatesDeclarations` can delegate to it.
- **`BeanMethod.getDeclaredMethodAnnotationMetadata()`** answers what a method declares itself, without the bean type and without the methods it overrides - the runtime counterpart of #12917's `MethodElement.getDeclaredMethodAnnotationMetadata()`; the generated metadata already retained the split.
- **`ReflectionArguments.returnOf(Method)`** folds the annotations of the method into the return argument, as `of(Parameter)` does for a parameter and as a generated executable method of a bean definition reports it; the validator's fold in `ExecutableHierarchy.Declaration` becomes a no-op for reflective declarations. The bean methods of a reflective introspection keep the argument of the return type alone, which is what a generated introspection writes, so the return parity specs hold.
- **`GroovyMethodElement.getOverriddenMethods()`** wrapped the method itself once per overridden method; it now wraps each overridden method, so the Groovy hierarchy of an accessor no longer collapses to one declaration.

`ReflectionExecutables.beanConstructor` already reads `BeanIntrospection.getConstructors()`, which answers for `@Introspected(constructors = true)` too, so nothing changed there.

### Measured on micronaut-validation `spec-compliance`

With `members = true` on the TCK archive types and `separatesDeclarations` delegating to the introspection (plus exact declarations for the bean methods of such an introspection through `getDeclaredMethodAnnotationMetadata()`): `:micronaut-validation:test` 534 tests, 0 failures; `jakartaTck` 1054 run, 0 failed; `jakartaTckIntrospection` with every exclusion removed 1054 run, 34 failed, none outside the former exclusions. 14 of the 44 exclusions pass now:

- `ValidationRequirementTest.testConstraintAppliedOnFieldAndProperty`, `testFieldAccess`
- `GroupTest.testImplicitGrouping`
- `GroupSequenceIsolationTest.testCorrectDefaultSequenceInheritance3`
- `ConstraintInheritanceTest.testValidationConsidersConstraintsFromSuperTypes`
- `ConstraintDescriptorTest.testGetGroupsWithImplicitGroup`
- `ContainerElementTypeDescriptorTest.testGetContainerElementMetaDataForRoles`
- `CrossParameterDescriptorTest.testFindConstraintsForMethodLookingAt` (needs the exact bean method declarations)
- `ElementDescriptorTest.testDeclaredOn`
- `TraversableResolverTest.testCorrectCallsToIsReachableAndIsCascadable`, `...ForParameterValidation`, `...ForReturnValueValidation`
- `ValueAccessStrategyTest.testValueFromFieldIsPassedToValidatorOfFieldLevelConstraint`
- `ValueExtractorDefinitionTest.valuePassedToExtractorRetrievedFromHost`

Of the 30 left, 20 fail with "Bean introspection not found" for nested classes of the TCK test classes - `ExecutableDescriptorTest.testGetParameterDescriptorsForConstructorOfInnerClass` among them: the Java annotation processing round hands the visitors the top-level types only, so the harness's `TestClassVisitor` never annotates `CustomerService.InnerClass`. The TCK expects the enclosing instance among the constructor parameters, as `getConstructors()` reports it, so that one is a harness matter, not the described constructor. The other 10 need the methods an interface or a super class declares that are not bean methods of the introspection.

Reviewed with `codex review` over three passes; every finding is addressed above: the field dispatch naming an inaccessible declaring class, the changed member-reference constructor shape, and the declarations of an accessor inherited from parallel parent interfaces. A fourth, a private nested super class in the same package, is a non-issue that a test now documents: javac compiles such a class as package-private in the class file, so the introspection names it and the hidden field is read. The final pass reported no actionable defect.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
+++ b/core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
@@ -30,6 +30,7 @@ import io.micronaut.inject.annotation.AnnotationMetadataHierarchy;
 import io.micronaut.inject.ast.ClassElement;
 import io.micronaut.inject.ast.ElementModifier;
 import io.micronaut.inject.ast.ElementQuery;
+import io.micronaut.inject.ast.FieldElement;
 import io.micronaut.inject.ast.ImportedClass;
 import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.ast.ParameterElement;
@@ -51,8 +52,11 @@ import java.lang.annotation.Annotation;
 import java.lang.annotation.RetentionPolicy;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -532,6 +536,9 @@ public class IntrospectedTypeElementVisitor implements TypeElementVisitor<Object
         List<PropertyElement> beanProperties = ce.getBeanProperties(propertyElementQuery).stream()
             .filter(p -> !p.isExcluded())
             .toList();
+        if (members) {
+            writer.describeMembers();
+        }
         Optional<MethodElement> constructorElement = ce.getPrimaryConstructor();
         constructorElement.ifPresent(constructorEl -> {
             if (ArrayUtils.isNotEmpty(constructorEl.getParameters())) {
@@ -568,7 +575,7 @@ public class IntrospectedTypeElementVisitor implements TypeElementVisitor<Object
                 beanProperty.getReadType().map(t -> t.withAnnotationMetadata(annotationMetadata)).orElse(null),
                 beanProperty.getWriteType().map(t -> t.withAnnotationMetadata(annotationMetadata)).orElse(null),
                 beanProperty.isReadOnly(),
-                members ? resolvePropertyMembers(beanProperty) : List.of()
+                members ? resolvePropertyMembers(ce, beanProperty) : List.of()
             );
 
             for (AnnotationValue<?> indexedAnnotation : indexedAnnotations) {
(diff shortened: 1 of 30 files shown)


inject-groovy/src/test/resources/io/micronaut/inject/visitor/introspection-metadata-shape.txt
```diff
@@ -0,0 +1,123 @@
+introspection:
```
graemerocher: is this file necessary? Seems a bit error prone to include this unstructured internal data in tests

inject-kotlin/src/test/resources/io/micronaut/kotlin/processing/visitor/introspection-metadata-shape.txt
```diff
@@ -0,0 +1,172 @@
+introspection:
```
graemerocher: in particular we definitely don't want to duplicate it all over the place

graemerocher (changes requested):

inject-groovy/src/test/resources/io/micronaut/inject/visitor/introspection-metadata-shape.txt
dstepanov: Removed. The three checked-in dumps and the specs reading them are gone; the invariant they were there for - `members = true` is additive - is kept by the test in each module's `BeanIntrospectionSpec` that compiles the same hierarchy twice, with and without the members, and compares the two introspections. The expected side is computed by the test rather than checked in, so there is no data file to go stale.

inject-kotlin/src/test/resources/io/micronaut/kotlin/processing/visitor/introspection-metadata-shape.txt
dstepanov: Agreed, and the rendering itself was duplicated in three specs too. It now lives once in `IntrospectionMetadataShape` of `micronaut-inject-test-utils`, which inject-java-test, inject-groovy and inject-kotlin all use; each spec is a single `IntrospectionMetadataShape.of(withMembers) == IntrospectionMetadataShape.of(plain)`.

inject/src/main/java/io/micronaut/inject/beans/AbstractInitializableBeanIntrospection.java
```diff
@@ -1667,7 +1668,17 @@ public ElementType getElementType() {
...
+            // constant, so the class value of the declaring type carries its name when the constant fails,
+            // and the class is loaded by that name through the loader of the bean type
+            AnnotationClassValue<?> declaringType = ref.declaringType();
+            Class<?> type = declaringType.getType().orElse(null);
+            if (type != null) {
+                return type;
+            }
+            return ClassUtils.forName(declaringType.getName(), getBeanType().getClassLoader())
+                .orElseThrow(() -> new IllegalStateException("The type declaring the member " + ref.name()
+                    + " of " + getBeanType().getName() + " cannot be loaded: " + declaringType.getName()));
```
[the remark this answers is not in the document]
dstepanov: Done in bfdbe70. The member memoises the resolved class in a lazily initialised volatile field, the way `BeanPropertyImpl` already memoises its members: the resolution is idempotent, so a benign race is fine. This matters more than it looks, because `ClassUtils.forName` caches nothing outside its common-type map, so the name path re-loaded the class and rebuilt the failure message on every call.

core-processor/src/main/java/io/micronaut/inject/beans/visitor/IntrospectedTypeElementVisitor.java
```diff
@@ -602,44 +609,175 @@ private AnnotationMetadata mergeAnnotations(AnnotationMetadata annotationMetadat
...
+        if (candidateParameters.length != parameters.length) {
+            return false;
+        }
+        for (int i = 0; i < parameters.length; i++) {
+            if (!candidateParameters[i].getType().getName().equals(parameters[i].getType().getName())) {
+                return false;
+            }
+        }
+        return true;
+    }
```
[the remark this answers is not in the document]
dstepanov: I checked this against a test rather than assuming, and the declarations are not missed. The signature match is only the second of two nets: the first is `method.getOverriddenMethods()`, which is javac's `Elements.overrides` and already resolves erasure and bridges, so `Impl.setValue(String)` reports `Holder<T>.setValue(T)` as overridden and the interface declaration is collected there. The name comparison exists to catch the case overriding cannot see, a parallel interface that declares an accessor the type inherits without redeclaring, which is why it is a supplement and not the primary path. bfdbe70 adds a spec pinning exactly your scenario: a `Holder<T>` and an `AlsoHolder<T>` implemented with `String`, where all six declarations are reported with the annotations each one carries. If you know of a shape that still slips through, I would rather see it as a failing case than widen the match speculatively.

core/src/main/java/io/micronaut/core/beans/BeanMethod.java
```diff
@@ -36,7 +38,22 @@ public interface BeanMethod<B, T> extends Executable<B, T>, Named {
...
+     * {@code MethodElement.getDeclaredMethodAnnotationMetadata()} of the annotation processors.
+     *
+     * @return The annotation metadata of this method declaration only
+     * @since 5.2.0
+     */
+    default AnnotationMetadata getDeclaredMethodAnnotationMetadata() {
+        // the first call narrows the metadata of the bean type and the method to the one of the method, the
+        // second one narrows the method metadata to what this declaration carries
+        return getAnnotationMetadata().getDeclaredMetadata().getDeclaredMetadata();
+    }
```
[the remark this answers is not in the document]
dstepanov: Leaving this as is, deliberately. The double call is the same expression, with the same explanatory comment, as `MethodElement.getDeclaredMethodAnnotationMetadata()` introduced in #12917, and this method is documented as its runtime counterpart. Giving the runtime side a differently shaped helper would make the two halves of one contract read differently, which costs more than the chained call does. The behaviour of `getDeclaredMetadata()` is specified on `AnnotationMetadata` itself: it narrows a combined metadata to the element it is declared on, and a second call narrows that to what the element itself declares. Worth revisiting only as a single named API on `AnnotationMetadata` covering both sides at once.

graemerocher (approved):

PR #13054, opened 2026-09-08
fix(cookie): parse request Cookie header as name/value pairs

dstepanov:
Rebase of #12871 (by @kali834x) onto the current `5.2.x`. Same two commits, no content changes.

`DefaultServerCookieDecoder.decode` parses an inbound `Cookie` request header on runtimes without http-netty, but delegated to `java.net.HttpCookie.parse`, a `Set-Cookie` *response* parser. On a `Cookie` header with several pairs it returned only the first and dropped the rest (`SID=a; JSESSIONID=b` yielded just `SID`), threw `IllegalArgumentException` on a malformed name so a bad request turned into a 500, and treated `Path`/`Domain`/`Max-Age` as attributes on the inbound cookie, letting a client set `cookie.getPath()` which `NettyCookies` uses to filter which cookies reach the app. The `NettyLaxServerCookieDecoder` sibling backed by Netty's `ServerCookieDecoder.LAX` already parses it as a plain name/value list, so the two SPI implementations disagreed on identical input.

`decode` now splits on `;`, unwraps balanced quotes, and skips empty or unparsable pairs instead of interpreting attributes or throwing. A value that opens with a double quote that is never closed drops the pair, matching `CookieUtil.unwrapValue` in Netty LAX.

Review comments on the original PR are all addressed: the `=` scan is bounded to the current segment (no O(n²) rescan of the rest of the header), the unused catch variable is named `ignored`, and the unbalanced-quote divergence @HDPark95 found is fixed.

Verified against Netty by running both decoders side by side over a set of edge-case headers: `a="unterminated; b=2`, `a="`, `a=""`, `a=1;;b=2; ;c=3`, `=noname; a=1`, `foo`, `a=`, `a=b=c` and `a=1,b=2` all now decode identically. Remaining differences are pre-existing and not touched here: `LAX.decode(String)` returns a `TreeSet`, so the Netty decoder emits cookies sorted by name and deduped while this one keeps header order and duplicates; Netty also skips the RFC 2965 `$Version` pair and does not trim whitespace around a name or value.

`:micronaut-http:test --tests '*Cookie*'` passes, 35 tests.

Closes #12871

### DIFF
--- a/http/src/main/java/io/micronaut/http/cookie/DefaultServerCookieDecoder.java
+++ b/http/src/main/java/io/micronaut/http/cookie/DefaultServerCookieDecoder.java
@@ -16,11 +16,14 @@
 package io.micronaut.http.cookie;
 
 import io.micronaut.core.annotation.Internal;
-import java.net.HttpCookie;
+
+import java.util.ArrayList;
 import java.util.List;
 
 /**
- * {@link ServerCookieDecoder} implementation backed by {@link HttpCookie#parse(String)}.
+ * {@link ServerCookieDecoder} implementation that parses an HTTP {@code Cookie} request header into
+ * its individual name/value pairs, as defined by
+ * <a href="https://datatracker.ietf.org/doc/html/rfc6265#section-4.2.1">RFC 6265, Section 4.2.1</a>.
  *
  * @author Sergio del Amo
  * @since 4.3.0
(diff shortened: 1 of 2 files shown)


dstepanov:
@kali834x Can you pls resign the CLA?

graemerocher (approved):

PR #13056, opened 2026-09-08
test: add regression tests for DefaultBeanContext

dstepanov:
Rebase of #12758 onto `5.2.x`, with a follow-up commit tightening the assertions. Original work by @amirdeljouyi; #12758 is superseded and closed in favour of this one.

Two tests on `DefaultBeanContext`:

* `beanKeyEqualityIncludesQualifierAndTypeArguments` — `BeanKey` identity accounts for both the qualifier and the argument's type arguments (`List<String>` is not `List<Integer>`), across all three constructors.
* `runtimeSingletonRegistrationInvalidatesLookupCachesAndResolvesNamedCollections` — a negative `containsBean` result cached before a runtime singleton is registered does not survive the registration, and the new singleton is immediately visible through direct, optional, collection, stream, map and registration lookups.

The second one has teeth: with the `containsBeanCache` purges in both `purgeCacheForBeanInstance` and `purgeCacheForBeanDefinition` removed, it fails. (Removing either one alone does not — the two paths are redundant for this scenario, so the test pins the observable contract rather than one line.)

The follow-up commit drops two `toString` assertions that re-implemented `BeanKey`'s own formula and the `equals(self)`/`equals(null)` branch padding, and compares the two `getBeanRegistrations` results by value instead of by reference so the test survives the collection cache handing out a copy.

### DIFF
--- a/test-suite/src/test/java/io/micronaut/context/DefaultBeanContextTest.java
+++ b/test-suite/src/test/java/io/micronaut/context/DefaultBeanContextTest.java
@@ -2,10 +2,21 @@ package io.micronaut.context;
 
 import io.micronaut.context.annotation.Secondary;
 import io.micronaut.context.exceptions.NonUniqueBeanException;
+import io.micronaut.core.type.Argument;
+import io.micronaut.inject.qualifiers.Qualifiers;
 import jakarta.inject.Singleton;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


dstepanov:
@amirdeljouyi Can you pls resign the CLA?

amirdeljouyi:
> @amirdeljouyi Can you pls resign the CLA?

It's done :)

graemerocher (approved):

PR #13066, opened 2026-09-08
Honour @Bean(preDestroy) on a bean class

dstepanov:
## The asymmetry

`io.micronaut.context.annotation.Bean` is `@Target({METHOD, ANNOTATION_TYPE, TYPE, FIELD})` and its `preDestroy()` member is documented as "The method to invoke to destroy the bean". Only `FactoryBeanElementCreator` ever read it. `DeclaredBeanElementCreator` registers pre-destroy callbacks from methods annotated `@PreDestroy` and nothing else, so on a managed bean the member did nothing at all — no callback, no error, no warning:

```java
@Singleton
@Bean(preDestroy = "close")   // silently ignored
class Closeable1 implements AutoCloseable {
    @Override public void close() { ... }   // never called on context.close()
}

@Factory
class ProducedFactory {
    @Singleton @Bean(preDestroy = "close")  // honoured
    Produced produced() { return new Produced(); }
}
```

Nothing in the code suggests the class case was left out deliberately — the member simply has a factory-scoped implementation, and the guide describes it as a factory feature while the javadoc does not. So this honours it rather than rejecting it: it is what a reader of the javadoc expects, and it gives an `AutoCloseable` bean a destroy callback without writing `@PreDestroy`.

## What changed

`DeclaredBeanElementCreator` resolves the named method on the bean type before its members are visited and registers it as a pre-destroy callback:

- Same validation as the factory case — a public, no-argument, accessible instance method of the bean type. A name that resolves to nothing (missing, private, or taking parameters) is now a `ProcessingException`, not silence.
- The callback is claimed as a lifecycle method, so it is excluded from around advice and from the bean's executable methods. That is what a method annotated `@PreDestroy` already gets, and it matches the exclusion `FactoryBeanElementCreator` applies to a produced bean's callback.
- A method that carries `@PreDestroy` itself is left to the member visitor, so naming it in `preDestroy` does not register the callback twice.

This is in the shared `core-processor` path, so it applies to Java, Groovy and Kotlin alike. The javadoc on `Bean#preDestroy()` and `src/main/docs/guide/ioc/lifecycle.adoc` were updated — the guide said "For factory beans, the `preDestroy` value…".

## Tests

New `PreDestroyOnBeanClassSpec` in `inject-java` covers the honoured case (plain, on context close, method inherited from a parent class, overloaded name), the no-duplicate-callback case, the AOP case, and three invalid-name cases (missing, private, takes parameters). Mirrored for Groovy (`inject-groovy`) and Kotlin (`inject-kotlin`) with the honoured and invalid-name cases.

`:micronaut-inject:test`, `:micronaut-inject-java:test`, `:micronaut-inject-groovy:test` and `:micronaut-inject-kotlin:test` all pass.

## Context

Found while working out what Jakarta CDI 5.0's `@AutoClose` would cost to implement in https://github.com/dstepanov/micronaut-cdi. That mapping is free if this member works on a class, and needs a compile-time rewrite of every auto-closeable bean if it does not.

### DIFF
--- a/src/main/docs/guide/ioc/lifecycle.adoc
+++ b/src/main/docs/guide/ioc/lifecycle.adoc
@@ -19,7 +19,24 @@ snippet::io.micronaut.docs.lifecycle.PreDestroyBean[tags="class", indent=0]
 <1> The `PreDestroy` annotation is imported
 <2> A method is annotated with `@PreDestroy` and will be invoked when the context is closed.
 
-For factory beans, the `preDestroy` value in the api:context.annotation.Bean[] annotation tells Micronaut framework which method to invoke.
+The `preDestroy` value in the api:context.annotation.Bean[] annotation names the method to invoke instead, which is useful when the destroy method cannot be annotated. It works both on a factory method and on the bean class itself:
+
+[source,java]
+----
+@Singleton
+@Bean(preDestroy = "close")
+public class Resource implements AutoCloseable {
+
+    @Override
+    public void close() {
+        // release the resource
+    }
+}
+----
+
+The named method must be a public, no-argument method of the bean type; a name that resolves to no such method is a compilation error.
+
+For factory beans, the value is set on the factory method:
 
 snippet::io.micronaut.docs.lifecycle.ConnectionFactory[tags="class", indent=0]
 
--- a/src/main/docs/guide/ioc/lifecycle.adoc
+++ b/src/main/docs/guide/ioc/lifecycle.adoc
@@ -19,24 +19,16 @@ snippet::io.micronaut.docs.lifecycle.PreDestroyBean[tags="class", indent=0]
 <1> The `PreDestroy` annotation is imported
 <2> A method is annotated with `@PreDestroy` and will be invoked when the context is closed.
 
-The `preDestroy` value in the api:context.annotation.Bean[] annotation names the method to invoke instead, which is useful when the destroy method cannot be annotated. It works both on a factory method and on the bean class itself:
+The `preDestroy` value in the api:context.annotation.Bean[] annotation names the method to invoke instead of annotating it. It can be declared on the bean class itself:
 
-[source,java]
-----
-@Singleton
-@Bean(preDestroy = "close")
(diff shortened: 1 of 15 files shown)


src/main/docs/guide/ioc/lifecycle.adoc
```diff
@@ -19,7 +19,24 @@ snippet::io.micronaut.docs.lifecycle.PreDestroyBean[tags="class", indent=0]
 <1> The `PreDestroy` annotation is imported
 <2> A method is annotated with `@PreDestroy` and will be invoked when the context is closed.
 
-For factory beans, the `preDestroy` value in the api:context.annotation.Bean[] annotation tells Micronaut framework which method to invoke.
+The `preDestroy` value in the api:context.annotation.Bean[] annotation names the method to invoke instead, which is useful when the destroy method cannot be annotated. It works both on a factory method and on the bean class itself:
+
+[source,java]
```
graemerocher: needs to be a multi language snippet with all langs

graemerocher (changes requested):

src/main/docs/guide/ioc/lifecycle.adoc
dstepanov: Done in 2ddf276 — the inline block is now `snippet::io.micronaut.docs.lifecycle.Cache[tags="class"]`, with the bean added to test-suite (Java), test-suite-groovy, test-suite-kotlin and test-suite-python, and each suite's existing lifecycle spec asserting the callback fires.

The Python one turned up a related gap: `PythonStubGenerator` bridges a method into the generated Java stub when it carries `@PreDestroy`/`@Executable`/etc., plus an explicit hook for the method a factory's `@Bean(preDestroy)` names — the same asymmetry one layer down. Without the class-level counterpart the generated definition invoked a method the stub did not declare and disposal failed with a `NoSuchMethodError`. That is fixed in the same commit; `:test-suite-python:test` and `:micronaut-inject-python:test` pass with `-Ppython-ci`.

graemerocher (approved):

PR #13068, opened 2026-09-08
Let construct advice enforce constraints on a container-invoked constructor

dstepanov:
Draft. Makes it possible for micronaut-validation to enforce constraints declared on a constructor the
container invokes — Jakarta Validation §5.1.2, `ExecutableType.CONSTRUCTORS` — which today is never checked:

```java
@Singleton
class Event {
    Event(@NotNull String title) {}      // parameters not validated on construction
}
```

Describing a constrained constructor and validating one explicitly already work (#12909). What was missing is
the container enforcing them on invocation.

## The premise that made this look big was wrong

An earlier write-up said a constructor cannot be advised, and rated closing this as *large*. Constructors
**are** advisable, and the machinery is mature: `@AroundConstruct`, `ConstructorInterceptor`,
`ConstructorInterceptorChain`, and — since #12959 — a target-constructor view so an interceptor of a proxied
bean sees the bean's own constructor rather than the generated proxy one.

Probing it in this branch, everything a Jakarta `ExecutableValidator` needs is already reachable from a
`ConstructorInterceptor`: constructor-level metadata (cross-parameter and return-value constraints) via
`getConstructor().getAnnotationMetadata()`, per-parameter metadata via `getArguments()`, the resolved
`getParameterValues()`, and `getDeclaringType()` reporting the target type even under a proxy.

So micronaut-validation can declare construct advice on a constrained constructor from its own
`ValidationVisitor` — which already visits constructors and already marks the constrained ones with
`@RequiresValidation`. **Core needs no processor change for that**, and this PR deliberately does not add one:
core supplies the mechanism and stays ignorant of the library's annotation.

Two things in core did stand in the way.

## 1. An exception thrown by construct advice was swallowed

Advice around a *method* reaches its caller as it was thrown. Advice around a *constructor* runs while the
container is creating the bean, where `DefaultBeanContext` wraps any throwable in a
`BeanInstantiationException`. Validation advice that rejects a construction was therefore invisible as itself
— and the TCK asserts `catch (ConstraintViolationException e)`.

`ConstructorInterceptorChain` now distinguishes a throwable from the constructor *body* — still wrapped,
behaviour unchanged — from one thrown by the *advice* around it, which is carried in an `@Internal`
`ConstructorAdviceException` that `DefaultBeanContext` unwraps at the single instantiation choke point. Both
the before-proceed case (parameter constraints) and the after-proceed case (return-value constraints) are
covered.

**This is the part I would most like a second opinion on**, and it is a separate commit so it can be dropped.
The alternative is to keep today's wrapping, which preserves the `BeanInstantiationException` contract but
leaves the TCK methods red.

## 2. `@RequiresValidation` did not allow `ElementType.CONSTRUCTOR`

`ValidationVisitor` already applies it to constructors synthetically; the declaration just did not say so,
which makes it impossible to write in source or in a test.

## Behaviour this enables

Measured with an `@AroundConstruct` annotation on a constructor:

| Case | Behaviour |
|---|---|
| plain `@Singleton` / `@Prototype`, records | advised on every container instantiation |
| constructor injection | advice runs after dependencies resolve, before the body |
| around-advised type | intercepted once, on the `$Intercepted` construction; the advice sees the target constructor |
| `@Around(proxyTarget = true)` | intercepted on the target's construction |
| `@Factory` method | not advised — and correctly so: §5.1.2 asks for constructors *the container* invokes, and a factory body's `new` is user code. A constrained factory method is already advised as a *method* |
| abstract / `@Introduction` type, inherited superclass constructor | not advised; construct advice is not inherited |
| plain `new Event(...)` in user code | untouched |
| failure during eager startup | `BeanInstantiationException` as today; a lazy `getBean` sees the advice's own exception |

## Follow-up, not in this PR

- **micronaut-validation:** a `@ValidatedConstructor` (`@AroundConstruct`, deliberately separate from
  `@Validated`, which is `@Around` — putting that on a constructor would drag around-advice semantics onto its
  metadata), applied from `ValidationVisitor.visitConstructor`, and a `ValidatingConstructorInterceptor` at
  `InterceptPhase.VALIDATE` calling the existing `validateConstructorParameters` /
  `validateConstructorReturnValue`. Return-value constraints must be evaluated against `getDeclaringType()`,
  not `instance.getClass()` — under `proxyTarget=false` the latter is the `$Intercepted` subclass.
  `@ConfigurationProperties` should be skipped there: those already validate after construction, and advising
  them too would double-validate.
- Then un-exclude the TCK classes and re-measure. Note the TCK has **four** constructor methods, not two, and
  `ExecutableValidationTest` / `ExecutableTypesTest` are excluded wholesale rather than per-method.
- **Separate defect found while probing:** a bean with **type-level** `@Around` **and type-level**
  `@AroundConstruct` gets no construct interception, silently. Constructor-level construct advice on the same
  bean works, so it does not block this. Worth its own issue.

## Tests

`ConstructorValidationAdviceSpec` in `inject-java` exercises the whole path a validation library would take,
with nothing core-internal short-circuited. The beans under test carry real constraints — `@Size(min = 5)` on
a constructor parameter, and a `jakarta.validation.Constraint`-meta-annotated `@ValidEvent` on the constructor
itself — a local `TypeElementVisitor` stands in for `ValidationVisitor`, deriving `@RequiresValidation` and the
construct advice from those constraints, and a stand-in `ValidatingConstructorInterceptor` reads the
constraints off the invocation context and throws a real `jakarta.validation.ConstraintViolationException`.

Covered: a violated parameter constraint surfacing from `getBean` as a `ConstraintViolationException` and not a
wrapper; a satisfied one constructing the bean; a constraint on the constructor rejecting the instance it
returned (the after-proceed case); an unconstrained constructor not advised at all; an around-advised bean
validated once on its proxy's construction, in both the violating and the passing case; and a constructor-body
exception still being wrapped.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
@@ -23,6 +23,7 @@ import io.micronaut.aop.InvocationContext;
 import io.micronaut.context.BeanContext;
 import io.micronaut.context.BeanRegistration;
 import io.micronaut.context.BeanResolutionContext;
+import io.micronaut.context.exceptions.ConstructorAdviceException;
 import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.AnnotationValue;
 import io.micronaut.core.annotation.Internal;
@@ -107,7 +108,13 @@ public final class ConstructorInterceptorChain<T> extends AbstractInterceptorCha
             } else {
                 finalParameters = getParameterValues();
             }
-            return beanConstructor.instantiate(finalParameters);
+            try {
+                return beanConstructor.instantiate(finalParameters);
+            } catch (RuntimeException | Error e) {
+                // Marks the throwable as coming from the constructor body rather than from advice, so that it
+                // keeps the wrapping an unadvised constructor's throwable gets. Unwrapped again in instantiate.
+                throw new ConstructorBodyException(e);
+            }
         } else {
             interceptor = this.interceptors[index++];
             if (LOG.isTraceEnabled()) {
@@ -197,13 +204,47 @@ public final class ConstructorInterceptorChain<T> extends AbstractInterceptorCha
         final InterceptorRegistry interceptorRegistry = beanContext.getBean(InterceptorRegistry.ARGUMENT);
         final Interceptor<T1, T1>[] resolvedInterceptors = interceptorRegistry
             .resolveConstructorInterceptors(constructor, interceptors);
-        return Objects.requireNonNull(new ConstructorInterceptorChain<>(
+        ConstructorInterceptorChain<T1> chain = new ConstructorInterceptorChain<>(
             definition,
             constructor,
             resolvedInterceptors,
             additionalProxyConstructorParametersCount,
             parameters
-        ).proceed(), "Constructor interceptor chain illegally returned null for constructor: " + constructor.getDescription());
+        );
+        T1 bean;
+        try {
+            bean = chain.proceed();
+        } catch (ConstructorBodyException e) {
+            throw e.rethrowCause();
+        } catch (ConstructorAdviceException e) {
+            throw e;
+        } catch (RuntimeException | Error e) {
+            // Anything else escaped the advice itself rather than the constructor it wraps. Advice around a
+            // method reaches its caller as it was thrown; carry this one so construction advice does too.
+            throw new ConstructorAdviceException(e);
+        }
+        return Objects.requireNonNull(bean, "Constructor interceptor chain illegally returned null for constructor: " + constructor.getDescription());
+    }
+
+    /**
+     * Carries a throwable out of the intercepted constructor's own body, so that
+     * {@link #instantiate} can tell it apart from one thrown by the advice around it.
+     */
+    private static final class ConstructorBodyException extends RuntimeException {
+
+        private final transient Throwable bodyCause;
+
+        private ConstructorBodyException(Throwable cause) {
+            super(cause.getMessage(), cause, false, false);
+            this.bodyCause = cause;
+        }
+
+        private RuntimeException rethrowCause() {
+            if (bodyCause instanceof Error error) {
+                throw error;
+            }
+            return (RuntimeException) bodyCause;
+        }
     }
 
     private static @Nullable Object[] resolveConcreteSubset(BeanDefinition<?> beanDefinition,
--- a/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
+++ b/aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
@@ -65,6 +65,12 @@ public final class ConstructorInterceptorChain<T> extends AbstractInterceptorCha
      */
     private final BeanConstructor<T> interceptedConstructor;
     private final @Nullable Object[] internalParameters;
+    /**
+     * The exception the intercepted constructor's own body threw, if it threw one. Kept so that
+     * {@link #instantiate} can tell it apart from an exception thrown by the advice around it: the two are
+     * propagated differently.
+     */
+    private @Nullable RuntimeException bodyFailure;
 
     /**
      * Default constructor.
@@ -110,10 +116,9 @@ public final class ConstructorInterceptorChain<T> extends AbstractInterceptorCha
             }
             try {
                 return beanConstructor.instantiate(finalParameters);
-            } catch (RuntimeException | Error e) {
-                // Marks the throwable as coming from the constructor body rather than from advice, so that it
-                // keeps the wrapping an unadvised constructor's throwable gets. Unwrapped again in instantiate.
-                throw new ConstructorBodyException(e);
+            } catch (RuntimeException e) {
+                bodyFailure = e;
+                throw e;
             }
         } else {
             interceptor = this.interceptors[index++];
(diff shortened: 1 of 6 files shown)


aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
```diff
@@ -197,13 +204,47 @@ public static <T1> T1 instantiate(
...
+            throw new ConstructorAdviceException(e);
+        }
+        return Objects.requireNonNull(bean, "Constructor interceptor chain illegally returned null for constructor: " + constructor.getDescription());
+    }
+
+    /**
+     * Carries a throwable out of the intercepted constructor's own body, so that
+     * {@link #instantiate} can tell it apart from one thrown by the advice around it.
+     */
+    private static final class ConstructorBodyException extends RuntimeException {
```
graemerocher: looks like this is used for flow control in that case it should override fillstackTrace() for performance

inject/src/main/java/io/micronaut/context/exceptions/ConstructorAdviceException.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Rethrows the exception the advice threw.
+     *
+     * <p>Advice cannot throw a checked exception, so the cause is always a {@link RuntimeException} or an
+     * {@link Error}. An {@link Error} is thrown directly; a {@link RuntimeException} is returned so that the
+     * call site can {@code throw} it and keep the compiler informed that the flow ends there.</p>
+     *
+     * @return The runtime exception the advice threw
+     */
+    public RuntimeException rethrowCause() {
```
graemerocher: again if this is used for flow control override fillstackTrace() for performance

graemerocher (changes requested):
needs documentation

aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
dstepanov: Removed this class entirely. The chain now remembers the exception its constructor body threw in a field, so `instantiate` tells a body failure from an advice failure by identity — one carrier instead of two, and no second `fillInStackTrace` to override.

inject/src/main/java/io/micronaut/context/exceptions/ConstructorAdviceException.java
dstepanov: Done — `fillInStackTrace()` now returns `this`, following `DisabledBeanException`. Noted in the javadoc that it is flow control between the chain and the bean creation boundary and never reaches user code; the exception it carries keeps its own stack trace, and that is the one the caller sees.

inject/src/main/java/io/micronaut/context/exceptions/ConstructorAdviceException.java
```diff
@@ -0,0 +1,63 @@
...
+        if (adviceCause instanceof Error error) {
+            throw error;
+        }
+        return (RuntimeException) adviceCause;
+    }
```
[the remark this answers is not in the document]
dstepanov: Fixed by narrowing rather than by handling: the constructor now takes a `RuntimeException`, and the chain no longer catches `Error` at all — an `Error` keeps the `BeanInstantiationException` wrapping it has today. No cast, and the invariant is enforced by the type.

aop/src/main/java/io/micronaut/aop/chain/ConstructorInterceptorChain.java
```diff
@@ -197,13 +204,47 @@ public static <T1> T1 instantiate(
...
+        private ConstructorBodyException(Throwable cause) {
+            super(cause.getMessage(), cause, false, false);
+            this.bodyCause = cause;
+        }
+
+        private RuntimeException rethrowCause() {
+            if (bodyCause instanceof Error error) {
+                throw error;
+            }
+            return (RuntimeException) bodyCause;
```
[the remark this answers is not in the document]
dstepanov: The duplication is gone — that second carrier was deleted in favour of a field on the chain, and narrowing to `RuntimeException` removed the cast and the `Error` branch from the one that remains.

inject/src/main/java/io/micronaut/context/exceptions/ConstructorAdviceException.java
```diff
@@ -0,0 +1,63 @@
...
+@Internal
+public final class ConstructorAdviceException extends RuntimeException {
+
+    private final transient Throwable adviceCause;
+
+    /**
+     * @param cause The throwable the advice threw
+     */
+    public ConstructorAdviceException(Throwable cause) {
+        super(cause.getMessage(), cause, false, false);
```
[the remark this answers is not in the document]
dstepanov: Switched to the plain `(message, cause)` constructor plus a `fillInStackTrace()` override, so suppression stays enabled and only the carrier's own stack trace is skipped. The exception the advice threw keeps its stack trace intact, and it is that one the caller receives — the carrier is unwrapped in `DefaultBeanContext` and never logged or observed.

dstepanov:
Review addressed in 911c522.

**Documentation** — added an *Exceptions Thrown by Construct Advice* section to the Bean Life Cycle Advice guide (`src/main/docs/guide/aop/lifecycleAdvice.adoc`), covering the contract this PR establishes: an exception thrown by a `ConstructorInterceptor` reaches whoever asked for the bean as it was thrown, before or after `proceed()`; an exception from the constructor’\s own body is still wrapped; and so is a failure during eager initialization at startup. That last claim now has a test behind it rather than being asserted on faith.

**`fillInStackTrace`** — overridden on `ConstructorAdviceException`, following `DisabledBeanException`. The second carrier that would have needed the same treatment is gone: the chain remembers the exception its constructor body threw in a field, and `instantiate` tells a body failure from an advice failure by identity. That also removes the duplication Copilot flagged.

**The cast** — fixed by narrowing rather than by handling. The carrier takes a `RuntimeException`, and the chain no longer catches `Error`, so an `Error` keeps the `BeanInstantiationException` wrapping it has today. No cast, no `Error` branch, invariant enforced by the type.

8 specs pass; 7554 tests green across `inject-java`, `inject-groovy`, `inject-kotlin`, `inject` and `context`.

graemerocher (approved):

PR #13077, opened 2026-09-09
Keep reusable GraalPy context alive across app shutdown

n0tl3ss:
When Pyronaut reuses its primary GraalPy context, each test fixture can close a Micronaut ApplicationContext while the reusable context remains needed by the next fixture. PythonPool.onDestroyed() currently closes the pool unconditionally, marking that shared context as closing.

Guard pool destruction when context reuse is enabled, matching GraalPyContextFactory, and add a regression test that evaluates the primary context after ApplicationContext.close().

Tested: ./gradlew :micronaut-context-python:test --tests io.micronaut.context.python.PythonPoolShutdownTest --no-daemon --console=plain -Ppython-ci

### DIFF
--- a/context-python/src/main/java/io/micronaut/context/python/PythonPool.java
+++ b/context-python/src/main/java/io/micronaut/context/python/PythonPool.java
@@ -742,6 +742,9 @@ final class PythonPool implements PythonContextExecutor, BeanDestroyedEventListe
 
     @Override
     public void onDestroyed(BeanDestroyedEvent<Context> event) {
+        if (PythonContextRuntime.isReuseContext()) {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


graemerocher (approved):

n0tl3ss:
Crema/native rationale: Crema reuses one GraalPy context across tests, so closing each fixture ApplicationContext must not destroy the shared Python pool. JVM tests use ordinary independent JVM lifecycle behavior and did not reproduce this failure. This guard preserves the JVM lifecycle while making reusable native GraalPy contexts safe.

Observed native error: after the first fixture closed its ApplicationContext, later tests raised `KeyError` for `FruitRepository` followed by `java.lang.IllegalStateException: The Python context is closing`. JVM tests did not reuse the same GraalPy context.

Issue #13074, opened 2026-09-09
A raw type's compiled argument is indistinguishable from one written with a type variable


dstepanov:
### Feature description

#13063 puts a type variable's bounds on the argument it compiles to, which is what a consumer needs to match a
variable by what it was declared with. Trying to read a runtime argument with it shows the other half of the
question is still unanswerable: **a raw usage compiles to the declaring type's own variables, so nothing tells
it apart from a usage written with a type variable.**

Probed on `5.2.x`, three injected fields of `class Holder<T extends Number>`:

```java
@Inject public List raw;            // FIELD raw:      [placeholder(name=E, typeVariable=true, bounds=[Object])]
@Inject public List<T> variable;    // FIELD variable: [placeholder(name=T, typeVariable=true, bounds=[Number])]
@Inject public List<String> concrete; // FIELD concrete: [argument(String, typeVariable=false)]
```

`List` raw and `List<T>` differ only in the *name* of the placeholder — `E` borrowed from `java.util.List`'s own
declaration versus `T` from the class — and a name is not a discriminator: an application may name a variable
`E`, and the borrowed name changes with the declaring type.

### Why it matters

Jakarta CDI's assignability rules (4.0 §2.4.2.1) treat the two as different questions. A required type with a
type variable matches a bean type when the bean type is assignable to every bound of the variable; a *raw*
required type matches by erasure and says nothing about parameters. Implementing them on Micronaut, the
distinction has to come from somewhere, and today it comes from the annotation processor: a visitor asks
`ClassElement.isRawType()` while compiling and writes what it learned into an annotation of its own for the
runtime to read back. Everything else that recording carried is now on the argument thanks to #13063 and
#13023 — bounds and wildcards — so this is the one thing keeping it alive.

Reading the placeholders back and treating them as variables, which is what the argument now invites, turns a
raw injection point into a parameterized one: five scenarios of the CDI TCK go from passing to failing that
way, all of them about raw types (`AssignabilityOfRawAndParameterizedTypesTest`,
`ParameterizedTypesInjectionRawAmbiguousTest`, `ProducerFieldDefinitionTest#testProducerFieldWithTypeVariable`
among them).

### Suggestion

Any of these would answer it:

- compile a raw usage to an argument with **no** type parameters, which is what the source says and what
  erasure means;
- keep the parameters but mark the argument as raw — a `boolean isRawType()` on `Argument`, alongside
  `isTypeVariable()`;
[...]

Issue #12998, opened 2026-09-04
RuntimeBeanDefinition.getTypeArguments drops a type argument bound by an intermediate generic super type, so the bean satisfies every type-argument qualifier


dstepanov:
### Expected Behavior

`RuntimeBeanDefinition.getTypeArguments(Class)` should report the type arguments a runtime-registered bean gives
to a super type, whether the bean class declares them itself or inherits them from an intermediate generic type.
A `class IntegerHandler extends AbstractHandler<Integer>`, where `AbstractHandler<T> implements Handler<T>`,
gives `Handler` the argument `Integer`, so `Qualifiers.byTypeArguments(String.class)` should not accept it.

### Actual Behaviour

`getTypeArguments` answers an empty list, silently. Every type-argument qualifier reads an empty list as
"the type does not specify type arguments, so this is the equivalent of using `Object`" and accepts the
candidate, so the bean satisfies *every* type-argument qualifier and is never excluded:

```
getTypeArguments(Handler)                 = []          <- expected [Integer]
Qualifiers.byTypeArguments(String)        = true        <- expected false
Qualifiers.byTypeArgumentsClosest(String) = false
```

`getBeansOfType(Handler.class, Qualifiers.byTypeArguments(String.class))` therefore returns a bean that is a
`Handler<Integer>`.

The cause is in `io.micronaut.core.reflect.GenericTypeUtils`. `resolveInterfaceTypeArguments` collects the
generic interfaces of the hierarchy and matches one by its raw type, but never substitutes the type variables of
the levels in between, so the `Handler<T>` it finds on `AbstractHandler` carries the variable `T`, which does not
resolve to a `Class` and is dropped. `resolveSuperTypeGenericArguments` and `resolveInterfaceTypeArgument` walk
their hierarchies the same way and have the same gap:

```
class Leaf extends Mid<String>,  Mid<T> implements Iface<T>
  GenericTypeUtils.resolveInterfaceTypeArguments(Leaf, Iface) -> []              (expected [String])
  GenericTypeUtils.resolveInterfaceTypeArgument(Leaf, Iface)  -> Optional.empty  (expected String)

class Baz extends Bar<String>,   Bar<T> extends Foo<T>
  GenericTypeUtils.resolveSuperTypeGenericArguments(Baz, Foo) -> []              (expected [String])
```

A class that re-declares the super type with a concrete argument is unaffected, which is why this has not shown
[...]

PR #13090, opened 2026-09-09
Tell a resolved binding on a bean definition apart from a written type variable

### LINKED ISSUE #13089: A bean definition's type argument reports isTypeVariable() for a concrete binding, and a raw producer's is not raw
### Feature description

Follow-up to #13074 / #13086, which made a raw usage answerable with `Argument.isRawType()`. Trying to retire a
compile-time recording against it left one question still unanswerable: **`isTypeVariable()` is true both for an
argument written as a type variable and for a concrete binding on a bean definition**, so a consumer cannot tell
`Holder<T>` from `Holder<String>` once the argument comes from a definition.

Probed on `5.2.x` at `20260909.151215`, two producers of the same generic type:

```java
static class Box<T> {}

@Produces @Singleton Box<String> typed() { … }
@Produces @Singleton Box raw() { … }         // @SuppressWarnings("rawtypes")
```

```
definition of typed():  asArgument() raw=false  params=[String /var=true /raw=false]   <-- String, reported as a variable
definition of raw():    asArgument() raw=false  params=[Object /var=true /raw=false]   <-- and the raw one is not raw either
```

At an injection point the same two questions answer correctly — `@Inject Holder<T>` gives a placeholder whose
`getBounds()` are the declared bounds, `@Inject Holder<String>` gives a plain argument, and a raw `@Inject
Holder` is `isRawType()`. It is on a **bean definition's own argument** that both collapse: every type argument
is a binding for the declaration's parameter, so it is flagged `isTypeVariable()`, whatever was bound to it,
and a raw producer's argument is not marked raw.

### Why it matters

Jakarta CDI's assignability rules (4.0 §2.4.2.1) ask a different question of each: a bean type with a *variable*
argument matches a required type assignable to every bound of that variable, while a bean type with a
*concrete* argument matches only that type, and a raw bean type matches by erasure. Implementing them on
Micronaut, a consumer that trusts `isTypeVariable()` on a definition's argument turns `Box<String>` into
`Box<T>` and the bean starts matching everything — the same class of silent over-matching as #12998, one level
down.

Concretely: reading these back instead of a compile-time recording of what the source said fails five TCK
scenarios about raw and parameterized producers, and makes a deployment the specification requires to be
rejected validate instead.

### Suggestion

Make the two distinguishable on a definition's argument, the way #13086 made rawness distinguishable on a
usage:

- `isTypeVariable()` on a definition's type argument reports whether the *source* wrote a variable there, with
  a resolved binding reported as an ordinary argument (the variable it binds is still available through
  `GenericPlaceholder#getVariableName()` where one is needed); and
- a raw producer's argument reports `isRawType()`, as a raw usage at an injection point already does.

The second is arguably part of #13086 for the producer path — the writer marks the usage, and a factory method's
return type is a usage like any other.

Found in https://github.com/dstepanov/micronaut-cdi, where `@CdiGenericVariables` — a compile-time recording of
what the source declared — exists for exactly the two questions above. #13023, #13063 and #13086 have retired
three quarters of it; this is what is left.

dstepanov:
A bean definition's own type argument could not answer two questions that Jakarta CDI's assignability rules (4.0 §2.4.2.1) ask differently of each: whether the source wrote a type variable there, and whether the bean type was written raw. `isTypeVariable()` was true of the `String` in a `Box<String>` producer as much as of the `T` in `class Box<T>`, and a raw producer's bean type was not raw at all.

Follow-up to #13074 / #13086, which made a raw *usage* answerable with `Argument.isRawType()`.

Fixes #13089

## What was wrong

Both halves had a single narrow cause, so `isTypeVariable()` keeps the meaning it always had and no additive API was needed.

**A resolved binding read as a written variable.** The definition's `$TYPE_ARGUMENTS` were already right — the writer emits `Argument.of(String.class, "T")` for a `Box<String>` producer and `Argument.ofTypeVariable(Object.class, "T")` only for a parameter the bean leaves unbound. The loss was in `BeanDefinition.asArgument()`, which rebuilt the bean type from `getTypeParameters()`, a `Class[]`; `Argument.of(Class, AnnotationMetadata, Class[])` names each entry after the parameter the declaring type declares and makes it a type variable, so every argument came back a placeholder whatever had been bound to it. It now takes the arguments from `getTypeArguments()` directly.

**A raw producer's bean type was not raw.** A factory method's return type is a usage like any other, but the bean type of a definition is not written as an argument — it is rebuilt from the class and the recorded type arguments — and a raw usage compiles to the type arguments the declaring type declares, so `Box raw()` and `Box<T>` reached the runtime as the same shape. The writer already knows: `beanTypeElement` is the generic return type, and the same `ArgumentExpUtils.isRawType` the argument writer asks of a usage answers for it across the three languages. It is recorded on `PrecalculatedInfo` and `AbstractInitializableBeanDefinition.asArgument()` builds `Argument.ofRawType` from it. The type parameters stay the declared placeholders, which is what a raw injection point keeps, so the two read alike.

## Why not change what the flag means

The issue offered an additive answer if the blast radius of redefining `isTypeVariable()` was too wide. It was not needed: nothing about the flag changed. Every caller that reads a definition's type arguments — `ExactTypeArgumentNameQualifier`, `TypeArgumentQualifier`, `ClosestTypeArgumentQualifier`, `MatchArgumentQualifier`, `BeanDefinition#getTypeArguments(Class)` — goes through `getTypeArguments`, which is untouched, and those arguments were already distinguishing the two cases. Only `asArgument()` (and `getGenericBeanType()`, which delegates to it) was fabricating placeholders, and it now stops. The arguments keep the names they had, so `asArgument().getTypeVariables()` is keyed the same way and a placeholder's `getVariableName()` is still reachable wherever a definition leaves a parameter unbound.

One note in passing: `ExactTypeArgumentNameQualifier` matches on the *erasure's* class name, not on the variable name, so it was never affected either way.

## Reflection

`ReflectionBeanDefinition` recorded the arguments a class gives its generic super types but nothing for the class itself, where a generated definition writes its own parameters too — so `asArgument()` of a reflective `class OpenBox<T>` had no type parameters where the generated one had a placeholder. It now records them, and a new parity spec diffs the two whole descriptions of one class.

## Behaviour

The four combinations, read from a compiled definition:

| declaration | `asArgument()` raw | type argument |
| --- | --- | --- |
| `Box<Integer> typed()` | false | `Integer`, `isTypeVariable()` false |
| `Box raw()` | **true** | `Number` var=`T` |
| `class OpenBox<T extends Number> extends Box<T>` | false | `Number` var=`T` |
| `class ClosedBox extends Box<Integer>` | false | `Integer`, `isTypeVariable()` false |

## Compatibility

`PrecalculatedInfo` gains a tenth component. Both constructors it had are kept, so a definition generated before this still loads. Reviewers may want to confirm whether japicmp wants an entry in `config/accepted-api-changes.json` for the new record component — I could not check it here without the released baseline.

## Tests

New specs in `inject-java`, `inject-groovy` and `inject-kotlin` covering the combinations above (Kotlin has no raw types, so it pins the other three), each also asserting that the arguments stay keyed by the name of the parameter they bind and that beans still resolve by type argument; and `DefinitionTypeArgumentParitySpec` in `reflection`, which compares a generated definition with the reflective one for the same class and records the one shape reflection cannot express — a raw bean type, since a `Class` is a declaration rather than a usage.

`:micronaut-inject:test`, `:micronaut-core:test`, `:micronaut-reflection:test`, `:micronaut-inject-java:test`, `:micronaut-inject-groovy:test` and `:micronaut-inject-kotlin:test` all pass, as does checkstyle on the four changed main source sets.

## Motivation

https://github.com/dstepanov/micronaut-cdi answers CDI's assignability rules from `@CdiGenericVariables`, a compile-time recording of what the source declared. #13023, #13063 and #13086 retired three quarters of it; this is the rest. Read back as they were, a `Box<String>` matched everything — the silent over-matching of #12998, one level down — and a deployment the specification requires to be rejected validated instead.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/writer/ArgumentExpUtils.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/ArgumentExpUtils.java
@@ -625,7 +625,7 @@ public final class ArgumentExpUtils {
      * @param element The element
      * @return true if the type is raw
      */
-    private static boolean isRawType(TypedElement element) {
+    static boolean isRawType(TypedElement element) {
         if (element instanceof GenericPlaceholderElement || element instanceof WildcardElement
             || !(element instanceof ClassElement classElement)) {
             // a variable or a wildcard is written as such, it is never a raw usage of a type
--- a/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
@@ -523,7 +523,8 @@ public final class BeanDefinitionWriter implements BeanElement, Toggleable, Elem
         boolean.class, // isConfigurationProperties
         boolean.class, // isContainerType
         boolean.class,  // requiresMethodProcessing,
-        boolean.class // hasEvaluatedExpressions
+        boolean.class, // hasEvaluatedExpressions
+        boolean.class // isRawBeanType
     );
 
     private static final String FIELD_CONSTRUCTOR = "$CONSTRUCTOR";
@@ -2425,7 +2426,9 @@ public final class BeanDefinitionWriter implements BeanElement, Toggleable, Elem
                                 )
                             : ExpressionDef.constant(false),
                         // 9: hasEvaluatedExpressions
-                        ExpressionDef.constant(evaluatedExpressionProcessor.hasEvaluatedExpressions())
+                        ExpressionDef.constant(evaluatedExpressionProcessor.hasEvaluatedExpressions()),
+                        // 10: isRawBeanType - a producer of a type written without its type arguments
+                        ExpressionDef.constant(ArgumentExpUtils.isRawType(beanTypeElement))
 
                     )
                 )
--- a/inject-groovy/src/test/groovy/io/micronaut/inject/generics/DefinitionTypeArgumentSpec.groovy
+++ b/inject-groovy/src/test/groovy/io/micronaut/inject/generics/DefinitionTypeArgumentSpec.groovy
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.inject.generics
+
+import io.micronaut.ast.transform.test.AbstractBeanDefinitionSpec
+import io.micronaut.context.ApplicationContext
+import io.micronaut.core.type.Argument
+import io.micronaut.core.type.GenericPlaceholder
+import io.micronaut.inject.BeanDefinition
+import io.micronaut.inject.qualifiers.Qualifiers
+import spock.lang.AutoCleanup
+import spock.lang.Shared
+
+/**
+ * A bean definition's own type argument answers what the source wrote: a binding of a concrete type is an
+ * ordinary argument, a parameter the bean leaves unbound is a placeholder of that name, and a bean type
+ * written without its type arguments is raw.
+ */
+class DefinitionTypeArgumentSpec extends AbstractBeanDefinitionSpec {
+
+    private static final String SOURCE = '''
+package test
+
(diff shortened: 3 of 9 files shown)


dstepanov:
Withdrawing this: the consumer that motivated it no longer needs it.

micronaut-cdi answers "what did the source write here" from a recording its own processor makes at compile time (`@CdiGenericVariables`), and the rule that module is built to — what the compiler recorded first, its optional reflection module second, an error naming that module third — makes such a recording the design rather than a stopgap. So the definition-side flag is not a blocker there, and this PR was filed on the premise that it was.

The inconsistency it describes is real regardless: `isTypeVariable()` answers true for the `String` of a `Box<String>` on a definition's argument, where the same question at an injection point answers false, and #13086 has just made rawness answerable on both sides. If that fidelity is worth having for other consumers of the runtime model, reopen it on those grounds rather than on this one — I have no measurement showing anything else needs it today.

PR #13097, opened 2026-09-10
Update managed-netty version to 4.2.18.Final

sdelamo:
https://github.com/netty/netty/releases/tag/netty-4.2.18.Final

> Netty 4.2.18 also changes decompression buffer growth. Compressed requests with very small micronaut.server.max-request-buffer-
  > size values may now fail with DecompressionException. The cumulative body-limit test allows additional decoder allocation
  > headroom while retaining its 64-byte body limit.


graemerocher (commented):
**Verdict.** Code changes are correct. Two things should be fixed before merge, neither a code change:
the test commit's comment describes the wrong cause, and the PR carries a Netty behaviour change
in request decompression that is not mentioned anywhere.

## 1. Request decompression: Netty 4.2.18 changed what `maxAllocation` means (verified by execution)

Netty 4.2.18 ships netty/netty#17370 ("Fix JdkZlibDecoder silently truncating highly compressible
streams"). It floors the decoder's requested output buffer at 512 bytes and, unchanged from before,
calls `prepareDecompressBuffer` for that growth after *every* inflate call, including the final one
when no input remains. With a `maxAllocation` cap the growth request can fail, and then the decoder
throws `DecompressionException("Decompression buffer has reached maximum size")` even though all
output was already produced.

Micronaut passes `micronaut.server.max-request-buffer-size` straight through as that cap
(`PipeliningServerHandler.maxZlibAllocation()`), so this is a behaviour change for Micronaut.
Measured with a standalone harness against both versions, one gzip stream, sync-flushed per chunk:

| maxAllocation | chunk inflates to | 4.2.17 | 4.2.18 |
|---|---|---|---|
| 64 | 48 | ok | **fails on chunk 1** |
| 256 | 48 | ok | **fails on chunk 1** |
| 512 | 48 | ok | **fails on chunk 1** |
| 1024 | 48 | ok | ok |
| 1024 | 400 | ok | ok |
| 1024 | 600 | ok | **fails on chunk 1** |
| 4096 | 3500 | ok | ok |
| 4096 | 3700 | ok | **fails on chunk 1** |
| 0 (unlimited) | 48 | ok | ok |

So on 4.2.18: a cap of 512 bytes or less rejects every compressed body on its first chunk, and
above that the effective per-read inflate cap is `maxAllocation - 512`.

Production impact with the default 10 MB `max-request-buffer-size`: a 512-byte window at the top of
a 10 MiB per-read cap, and the same body would be rejected by the 10 MB `max-request-size` anyway.
Negligible. Anyone who set `max-request-buffer-size` to a few hundred bytes now gets every gzip or
deflate request rejected; that is an odd configuration but it worked before.

What this means for the PR:
- The test change (`BodySizeLimits(64, 64)` -> `(64, 1024)`) is the right way to keep the test on
  the cumulative limit, but its comment, "Keep the decoder allocation limit above the cumulative body
  limit tested here", is not the reason. The reason is that Netty's decoder now needs 512 bytes of
  headroom after every inflate call, so any cap up to 512 fails outright. Suggest the comment say
  that and link netty/netty#17370, so the next person does not "fix" it back.
- The behaviour change should be stated in the PR description and ideally in release notes.
- Worth an upstream issue: the 512-byte floor should only apply when the decode loop is going to
  continue (`pendingOutput || !inflater.needsInput()`); the growth call after the last inflate is
  where the spurious failure comes from. The Netty PR text says "a configured limit keeps its
  meaning", so this looks unintended.
- Optional Micronaut-side hardening: floor `maxZlibAllocation()` at a few KiB with a comment. A cap
  smaller than one socket read never bounded anything anyway. Not required for this PR.

Note the upgrade is still clearly right: the truncation bug it fixes affects `ZlibWrapper.NONE`,
which Micronaut uses for `Content-Encoding: deflate` via `ZLIB_OR_NONE`, so on 4.2.17 a highly
compressible deflate body can arrive silently short. 4.2.18 also carries nine codec-http security
fixes.

## 2. QUIC writability deferral: correct, and the premise is real

`QuicheQuicStreamChannel` is byte-identical between 4.2.17 and 4.2.18, so the transient is exposed
by the quiche 0.29.2 bump rather than by Java code. The transient is visible in the source:
`writeQueued()` calls `updateWritabilityIfNeeded(true)` after draining its queue, and the caller
`writable(long capacity)` then immediately calls `updateWritabilityIfNeeded(this.capacity > 0)`,
which flips it back to false when capacity is exhausted. Both flips happen synchronously inside one
`writable()` call. Micronaut reacting to the first flip inside `channelWritabilityChanged` would
call `writeSome()` re-entrantly from inside the QUIC channel's own drain.

Deferring `writeSome()` by one executor task is sound for this: the task runs after `writable()`
returns, so it sees the final state. Checked for lost wakeups: every real false-to-true transition
fires a fresh event, `quicWritePending` is only set while a task is queued, so a true-false-true
sequence during the window still results in a write. Checked for teardown: `ctx` is never nulled,
and after `handlerRemoved` the outbound queue is empty, so a late task is a no-op.

Restricting it to QUIC (yawkat's commit) is the right call; TCP writability events come from the
outbound buffer after flush and do not have this reentrancy, and deferring there would add a hop to
every transition. Passing the builder's `quic` flag rather than an `instanceof` check is fine.

The new `PipeliningServerHandlerSpec` case tests the deferral logic with a simulated transient on an
`EmbeddedChannel`, not the QUIC channel itself. That is acceptable for a unit test; the HTTP/3 test
suite is what exercises the real thing.

Nit: the PR body is a bare link. The commit messages carry the rationale; the description should
too, since the change is not just a version bump.

## 3. `ProxyBackpressureTest` message
Assertion-message-only. Harmless.

PR #13103, opened 2026-09-10
JSON handling correctness fixes

graemerocher:
## Summary

- **`ResponseLifecycle`: a reactive response body without a route is now framed as a JSON
  array.** When a `Publisher` body was encoded without a `RouteInfo` — a response produced by a
  server filter or an exception handler — `isJson` was hard-coded to `false`, so the items were
  written as concatenated JSON documents (`{...}{...}`) under `Content-Type: application/json`,
  which is not a valid JSON body. The flag is now derived from the resolved content type in that
  case too, mirroring `RouteInfo#isResponseBodyJsonFormattable` by excluding `byte[]` and
  `ByteBuffer` items, which keep being written through unchanged. Without a route there is no
  declared body type, so the item type is taken from the first item that is written;
  `LazySendingSubscriber` only completes the flow once that item has been produced.

- **`JsonMessageHandler.writePiece` honours the declared `Argument`.** `writeTo(..)` serializes
  with the declared type and passes an already serialized JSON document (a `CharSequence` with a
  declared type of `Object`) through unchanged, while `writePiece(..)` called
  `jsonMapper.writeValue(s, object)`, dropping both. The two entry points now share one code path.
  On the Netty server only `writePiece` runs, so this is where the two runtimes could disagree.

- **`JsonCounter`: the framing of an unwrapped top-level array is validated.** The comma between
  elements was optional and a comma where an element was expected became part of the next element,
  so `[1 2]`, `[{}{}]`, `["foo""bar"]`, `[1,,2]` and `[,1]` were all accepted as framing. The
  separator is now required and a comma in place of an element is rejected. Separately, the end of
  the input was never checked against the unwrapping state, so `[1,2`, `[{}` and a lone `[`
  completed normally; the new `JsonCounter#noMoreInput`, called by `JsonChunkedProcessor` when the
  input ends, fails unless the top-level array was closed. Each element was already parsed by
  Jackson, so this only concerns the framing itself.

- **`JsonCounter`: the UTF-16/UTF-32 check honours the reader index.** It read the first four bytes
  of the buffer by absolute index while every other access in the class — and
  `JsonChunkedProcessor.countLoop`, which explicitly supports a non-zero reader index — works
  relative to `readerIndex()`. Buffers handed to `feed()` currently always start at zero, so this
  was not reachable from the wire; it is now read from the reader index and only inspects bytes
  that map to the first four bytes of the input.

## Testing

- `./gradlew :micronaut-http-netty:test :micronaut-json-core:test` — 184 tests, all green.
- `./gradlew :micronaut-http-server-netty:test --tests '*Json*'` — 90 tests, all green.
- `./gradlew :micronaut-http-server-netty:test` — 1160 tests, all green (19 skipped).
- `./gradlew :micronaut-http-server:test :micronaut-jackson-databind:test` — 128 + 64 tests, all
  green.
- `checkstyleMain`/`checkstyleTest` for `http-server`, `json-core`, `http-netty`,
  `http-server-netty` and `jackson-databind`. The only reported violations are pre-existing and in
  untouched files: the file length of `NettyHttpServerConfiguration` and a javadoc violation in
  `NettyWebSocketSession`.

Fail-before evidence, each recorded by restoring the base version of the changed main source and
re-running the new test:

- `FilterJsonArrayResponseSpec` "reactive response body without a route is formatted as a json
  array": `response.body() == '[{"n":1},{"n":2}]'` failed with the body
  `{"n":1}{"n":2}` (the two sibling cases, `byte[]` items and a non-JSON content type, pass before
  and after and guard against over-framing).
- `JsonMessageHandlerWritePieceSpec`: "writePiece writes an already serialized document through
  unchanged" failed with `"{\"foo\":\"bar\"}"` instead of `{"foo":"bar"}`, and the
  `writePiece`/`writeTo` agreement case for a declared `Object` failed the same way.
- `JsonCounterSpec` "illegal inputs unwrapTopLevelArray": all nine new inputs (`[1 2]`, `[{}{}]`,
  `["foo""bar"]`, `[1,,2]`, `[,1]`, `[`, `[1,2`, `[{}`, `[{"foo":`) failed with "Expected exception
  of type 'io.micronaut.json.JsonSyntaxException', but no exception was thrown", in all three
  feeding modes including the new byte-by-byte one. (To run these against the old logic,
  `noMoreInput` was temporarily stubbed out as a no-op.)
- `JsonCounterSpec` "non-zero reader index" failed with `JsonSyntaxException: Input must be legal
  UTF-8 JSON` on the buffer whose JSON starts at reader index 4.
- `JsonBodyBindingSpec` "test truncated json array with publisher argument" failed with "Expected
  exception of type 'HttpClientResponseException', but no exception was thrown" — the truncated
  body completed as an empty stream with a `200`.

New coverage: `JsonCounterSpec` now also feeds every top-level-array case one byte at a time
through a helper that mirrors `JsonChunkedProcessor`, so chunk boundaries inside an element and
between an element and its separator are covered.

## Out of scope

- A truncated request body bound to a reactive type is rejected with a `500`, not a `400`. Errors
  raised while a request body is streamed reach `ResponseLifecycle#handleStreamingError`, which
  does not consult `ExceptionHandler` beans, so a malformed *element* (for example
  `[{"name":}]`) already produces a `500` today. The new test asserts the same status as that
  pre-existing case; routing streamed-body errors to the JSON exception handler is a separate
  change.
- A trailing comma (`[1,]`) is still accepted as framing. `JsonCounter` documents that it does not
  detect every syntax error, and rejecting it needs a third state in the unwrapping state machine.

## Behaviour change

A reactive response body produced **without a route** (a server filter or an exception handler returning `HttpResponse.ok(Flux.just(...))`) with a JSON content type is now written as a JSON array, as a routed `Flux<T>` always was. That includes custom `+json` types such as `application/problem+json` and `application/vnd.foo+json`; `application/x-json-stream` is unaffected. Single-value publishers (`Mono`, `Single`, `Maybe`) are not framed. Anyone streaming concatenated JSON documents from a filter under a `+json` content type will see `[...]` framing after this change.

A JSON syntax error in a streamed `Publisher<T>` request body is answered with 400 instead of 500.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/body/JsonCounter.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/body/JsonCounter.java
@@ -86,13 +86,13 @@ public final class JsonCounter {
             // If the input is utf-16 or utf-32, one of the first four bytes will be 0. Checking
             // this separately and only for four bytes allows us to avoid the work in the hot loops
             // below.
-            int r = buf.readableBytes();
-            if ((r >= 1 && buf.getByte(0) == 0)
-                || (r >= 2 && buf.getByte(1) == 0)
-                || (r >= 3 && buf.getByte(2) == 0)
-                || (r >= 4 && buf.getByte(3) == 0)) {
-
-                throw new JsonSyntaxException("Input must be legal UTF-8 JSON");
+            int i = buf.readerIndex();
+            // only look at the bytes of this buffer that are among the first four bytes of the input
+            int end = i + Math.min(buf.readableBytes(), 4 - (int) position);
+            for (; i < end; i++) {
+                if (buf.getByte(i) == 0) {
+                    throw new JsonSyntaxException("Input must be legal UTF-8 JSON");
+                }
             }
         }
         if (!isBuffering()) {
--- a/http-netty/src/main/java/io/micronaut/http/netty/body/JsonCounter.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/body/JsonCounter.java
@@ -59,7 +59,7 @@ public final class JsonCounter {
      *
      * @see #unwrapTopLevelArray()
      */
-    private boolean allowUnwrappingArrayComma;
+    private boolean expectUnwrappingArrayComma;
 
     /**
      * The region of the last complete top-level JSON node we have visited. Polled by the user.
@@ -100,6 +100,18 @@ public final class JsonCounter {
         }
         if (isBuffering()) {
             proceedUntilNonBuffering(buf);
+        }
+    }
+
+    /**
+     * Signal that the input has ended. A top-level array that is being
+     * {@link #unwrapTopLevelArray() unwrapped} must have been closed at this point.
+     *
+     * @throws JsonSyntaxException If the input ended inside the top-level array
+     */
+    public void noMoreInput() throws JsonSyntaxException {
+        if (unwrappingArray && state != State.AFTER_UNWRAP_ARRAY) {
+            throw new JsonSyntaxException("Unexpected end of input inside the top-level array");
         }
     }
 
@@ -219,12 +231,21 @@ public final class JsonCounter {
                 throw new JsonSyntaxException("UTF-8 BOM not allowed");
             }
 
-            // if we are unwrapping a top-level array, search for a comma
-            if (allowUnwrappingArrayComma) {
+            // if we are unwrapping a top-level array, the elements must be separated by commas
+            if (unwrappingArray) {
                 i = skipWs(buf, i, end);
-                if (i < end && buf.getByte(i) == ',') {
-                    allowUnwrappingArrayComma = false;
-                    i++;
+                if (i < end) {
+                    byte b = buf.getByte(i);
+                    if (expectUnwrappingArrayComma) {
+                        if (b == ',') {
+                            expectUnwrappingArrayComma = false;
+                            i++;
+                        } else if (b != ']') {
+                            failMissingComma();
+                        }
+                    } else if (b == ',') {
+                        failUnexpectedComma();
+                    }
                 }
             }
             i = skipWs(buf, i, end);
@@ -266,6 +287,7 @@ public final class JsonCounter {
             case ']' -> {
                 if (unwrappingArray) {
                     state = State.AFTER_UNWRAP_ARRAY;
+                    expectUnwrappingArrayComma = false;
                 } else {
                     failMismatchedBrackets();
                 }
@@ -311,7 +333,7 @@ public final class JsonCounter {
             position--;
             flushAfter();
             position++;
-            allowUnwrappingArrayComma = unwrappingArray;
+            expectUnwrappingArrayComma = unwrappingArray;
             state = State.BASE;
         } else if (unwrappingArray && (b == ',' || b == ']')) {
             position--;
@@ -322,7 +344,7 @@ public final class JsonCounter {
             } else {
                 state = State.AFTER_UNWRAP_ARRAY;
             }
-            allowUnwrappingArrayComma = false;
+            expectUnwrappingArrayComma = false;
         } else {
             failMissingWs();
         }
@@ -397,7 +419,7 @@ public final class JsonCounter {
         assert position >= bufferStart;
         lastFlushedRegion = new BufferRegion(bufferStart, position + 1);
         bufferStart = -1;
-        allowUnwrappingArrayComma = unwrappingArray;
+        expectUnwrappingArrayComma = unwrappingArray;
     }
 
     /**
@@ -450,6 +472,14 @@ public final class JsonCounter {
 
     private static void failMismatchedBrackets() throws JsonSyntaxException {
         throw new JsonSyntaxException("JSON has mismatched brackets");
+    }
+
+    private static void failMissingComma() throws JsonSyntaxException {
+        throw new JsonSyntaxException("Expected a comma between the elements of the top-level array");
+    }
+
+    private static void failUnexpectedComma() throws JsonSyntaxException {
+        throw new JsonSyntaxException("Expected an element of the top-level array, not a comma");
     }
 
     private static void failMissingWs() throws JsonSyntaxException {
--- a/http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java
+++ b/http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java
@@ -22,6 +22,7 @@ import io.micronaut.core.convert.ConversionService;
 import io.micronaut.core.execution.ExecutionFlow;
 import io.micronaut.core.io.buffer.ByteBuffer;
 import io.micronaut.core.type.Argument;
+import io.micronaut.core.convert.exceptions.ConversionErrorException;
 import io.micronaut.http.ByteBodyHttpResponse;
 import io.micronaut.http.ByteBodyHttpResponseWrapper;
 import io.micronaut.http.HttpMethod;
@@ -41,6 +42,7 @@ import io.micronaut.http.body.ResponseBodyWriter;
 import io.micronaut.http.codec.CodecException;
 import io.micronaut.http.exceptions.HttpStatusException;
 import io.micronaut.http.reactive.execution.ReactiveExecutionFlow;
+import io.micronaut.json.JsonSyntaxException;
 import io.micronaut.web.router.DefaultUrlRouteInfo;
 import io.micronaut.web.router.RouteAttributes;
 import io.micronaut.web.router.RouteInfo;
@@ -308,7 +310,15 @@ public abstract class ResponseLifecycle {
     protected final ExecutionFlow<? extends ByteBodyHttpResponse<?>> handleStreamingError(HttpRequest<?> request, Throwable t) {
         // limited error handling
         MutableHttpResponse<?> errorResponse;
-        if (t instanceof HttpStatusException hse) {
+        if (t instanceof ConversionErrorException cee && cee.getCause() instanceof JsonSyntaxException jse) {
+            // with delayed parsing, json syntax errors show up as conversion errors
+            t = jse;
+        }
+        if (t instanceof JsonSyntaxException) {
+            // a syntax error in a streamed request body is the client's fault, not the server's,
+            // the same way RequestLifecycle answers it for a fully buffered body
+            errorResponse = HttpResponse.badRequest().body(t.getMessage());
+        } else if (t instanceof HttpStatusException hse) {
             errorResponse = HttpResponse.status(hse.getStatus());
             if (hse.getBody().isPresent()) {
                 errorResponse.body(hse.getBody().get());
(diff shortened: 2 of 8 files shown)


dstepanov (commented):
Array framing for a multi-item `Flux` is right, but the new branch for responses without a route also frames single-value publishers. I ran a probe spec on the Netty server against this head and against the base `ResponseLifecycle`/`JsonMessageHandler`:

| Response from a filter, `application/json` | 5.2.x | this PR |
|---|---|---|
| `Mono.just([n:1])` | `{"n":1}` | `[{"n":1}]` |
| `Mono.empty()` | empty body | `[]` |
| `Flux.empty()` | empty body | `[]` |
| `Flux<Map>` as `application/problem+json` | `{..}{..}` | `[{..},{..}]` |
| `Flux<Map>` as `application/x-json-stream` | `{..}{..}` | `{..}{..}` |
| controller `Mono<Map>` (reference) | `{"n":1}` | `{"n":1}` |

1. **Bug: a `Mono`/`Single`/`Maybe` body gets wrapped in an array.** [`ResponseLifecycle.java#L236-L243`](https://github.com/micronaut-projects/micronaut-core/blob/d3e03fed7f25007f33167485b307e206d514083b/http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java#L236-L243)
   - Controllers never reach here with a `Mono` because `RouteExecutor` unwraps it first. The branch without a route has no `Publishers.isSingle(body.getClass())` check.
   - So a filter returning `HttpResponse.ok(Mono.just(dto))` now sends `[{...}]`.
   - Please skip framing for single-value publishers and add `Mono`/`Mono.empty()` rows to `FilterJsonArrayResponseSpec`.
2. **Whether to frame can depend on timing.** The flag is read once, when the stream completes ([#L259](https://github.com/micronaut-projects/micronaut-core/blob/d3e03fed7f25007f33167485b307e206d514083b/http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java#L259)), but items that aren't JSON-formattable can flip it while `concatMap` is already writing ([#L247-L248](https://github.com/micronaut-projects/micronaut-core/blob/d3e03fed7f25007f33167485b307e206d514083b/http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java#L247-L248)). With an async source, a mixed `Flux.just(map, bytes)` can come out as `[{"n":1},<raw bytes>]` or unframed depending on timing. Deciding from the first item would make it deterministic.
3. **Release-note material:** custom `+json` types without a route (`application/problem+json`, `application/vnd.foo+json`) now get array framing. It's consistent with the route path, but it's a visible change for anyone streaming concatenated documents from a filter.
4. **Malformed arrays get the wrong status.** `JsonCounter` now rejects bodies like `[{}{}]` and `[1,,2]`, which were accepted before, but with 500 rather than 400, and [`JsonBodyBindingSpec.groovy#L276`](https://github.com/micronaut-projects/micronaut-core/blob/d3e03fed7f25007f33167485b307e206d514083b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/binding/JsonBodyBindingSpec.groovy#L276) locks in the 500. A client syntax error should probably be 400.
5. **Not verified: `JsonMessageHandler.writePiece` now honours the declared type.** A direct caller holding an unspecialised handler with a supertype `Argument` would now get `writerFor(supertype)` and lose subtype-only properties. The new spec covers Jackson only, not micronaut-serialization.

The `JsonCounter` framing and charset changes look correct: I walked through `[]`, `[1 ]`, `[[1],[2]]` and a comma split across chunks.

graemerocher:

---

The primary changes are otherwise coherent, but routeless reactive responses containing raw `ByteBody` items can now have JSON delimiters inserted into their byte stream. This is a functional regression for supported raw response bodies.


- [P2] Exclude ByteBody from JSON array formatting — http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java:271-273
  When a short-circuit filter or exception handler returns a `Publisher<ByteBody>` with a JSON content type, this predicate classifies it as JSON-formattable. `ByteBodyWriter.writePiece` passes those raw bytes through unchanged, so `concatenateJson` then injects brackets and commas and corrupts a response whose items represent raw chunks; treat `ByteBody` and equivalent raw body types as non-formattable like `byte[]` and `ByteBuffer`.

---

**Disposition:** Verified. Fixed locally in `562c00bc19`: `ByteBody` is treated as raw like `byte[]` and `ByteBuffer`, with a spec row. `RouteInfo.isResponseBodyJsonFormattable` has the same pre-existing gap and is left alone.

graemerocher:
Thanks for the probe table, that was exactly the right check. Point by point:

1. **Single-value bodies wrapped in an array — fixed** (`312adeb297`). The routeless branch now checks `Publishers.isSingle(body.getClass())` like `RouteExecutor` does before the route path gets here, so `Mono.just(dto)` sends `{...}` and `Mono.empty()` an empty body again. `FilterJsonArrayResponseSpec` has `Mono.just`, `Mono.empty` and `Flux.empty` rows; `Flux.empty()` stays `[]`, matching a routed `Flux<T>`.

2. **Timing-dependent framing — fixed** (same commit). The formattable check now runs on the first item only, so a mixed stream comes out the same way whatever the source's timing. Two spec rows cover it with the second item delayed by 50 ms: object-then-bytes is `[{"n":1},{"n":2}]`, bytes-then-object is `{"n":2}{"n":1}`.

3. **Release-note material — agreed.** Added a "Behaviour change" section to the PR description covering custom `+json` types without a route.

4. **Malformed arrays answered 500 — fixed** (`665079b4e0`). `handleStreamingError` only knew `HttpStatusException`; it now treats a `JsonSyntaxException` (also when wrapped in a `ConversionErrorException` by delayed parsing) as 400, the same way `RequestLifecycle.onErrorNoFilter` already does for a buffered body. `JsonBodyBindingSpec` expects 400 for the truncated arrays now and gains the `[{}{}]` and `[1,,2]` shapes. One caveat, stated in the test: on a route that streams its response, the first element is already on the wire when the error arrives, so the only thing the server can still do is close the connection; the new cases therefore use a route that collects the body before answering.

5. **`writePiece` and subtype properties — verified, no change in behaviour.** I measured `Animal plain()`, `Mono<Animal>`, `Flux<Animal>` and `List<Animal>` controller methods returning a `Dog` with an extra property, on 5.2.x and on this branch: all four give `{"name":"rex"}` on both. On the route path the writer is already specialised for the declared type, so subtype-only properties were dropped before this PR on every path including `Flux<Animal>`. The commit only makes an *unspecialised* `writePiece` agree with `writeTo`, which has used the declared type all along. Not covered for micronaut-serialization here; its `writeValue(OutputStream, Argument, T)` contract is the same.

Also fixed while here (`de0aec123b`, from a second review pass): the routeless formattable check excluded `byte[]` and `ByteBuffer` but not `ByteBody`, whose writer passes bytes through unchanged, so a filter returning `Publisher<ByteBody>` under a JSON type would have had brackets and commas spliced into its byte stream. `ByteBody` is now treated as raw too, with a spec row. `RouteInfo.isResponseBodyJsonFormattable` has the same gap for a declared `Publisher<ByteBody>`; that is pre-existing and left alone.

http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java
```diff
@@ -277,7 +314,15 @@ protected CloseableByteBody concatenateJson(Publisher<ByteBody> items) {
...
         MutableHttpResponse<?> errorResponse;
-        if (t instanceof HttpStatusException hse) {
+        if (t instanceof ConversionErrorException cee && cee.getCause() instanceof JsonSyntaxException jse) {
+            // with delayed parsing, json syntax errors show up as conversion errors
+            t = jse;
+        }
+        if (t instanceof JsonSyntaxException) {
+            // a syntax error in a streamed request body is the client's fault, not the server's,
+            // the same way RequestLifecycle answers it for a fully buffered body
+            errorResponse = HttpResponse.badRequest().body(t.getMessage());
```
dstepanov: The same syntax error in a buffered body goes through `JsonExceptionHandler` and the `ErrorResponseProcessor`: "Invalid JSON" reason, the standard JSON error body (or problem+json when that module is present), and any user `ExceptionHandler<JsonSyntaxException>`. Here the client gets a bare text body with the raw exception message, so the error format depends on whether the body happened to be streamed. The new tests only check the status.

Could this go through `errorResponseProcessor` with `BAD_REQUEST` and the same `"Invalid JSON: " + message`, the way `createDefaultErrorResponse` does for the 500?

http-netty/src/main/java/io/micronaut/http/netty/body/JsonCounter.java
```diff
@@ -219,12 +231,21 @@ private void proceedUntilBuffering(ByteBuf buf) throws JsonSyntaxException {
...
+            if (unwrappingArray) {
                 i = skipWs(buf, i, end);
-                if (i < end && buf.getByte(i) == ',') {
-                    allowUnwrappingArrayComma = false;
-                    i++;
+                if (i < end) {
+                    byte b = buf.getByte(i);
+                    if (expectUnwrappingArrayComma) {
+                        if (b == ',') {
+                            expectUnwrappingArrayComma = false;
```
dstepanov: Nit: after consuming the expected comma, the method goes straight to reading the element and doesn't repeat the unexpected-comma check, so the result depends on chunking. Probe on this head:

| input | whole buffer | byte by byte |
|---|---|---|
| `[1 , ,2]` | elements `1`, `,2` | unexpected comma |
| `[{} , ,{}]` | missing whitespace | unexpected comma |
| `[1, ,2]` | unexpected comma | unexpected comma |

Jackson will probably still reject `,2`, so not a hole, but repeating the check after the comma (and adding `'[1 , ,2]'` to the illegal inputs) would make the framing check consistent.

dstepanov (approved):
All five points from the earlier review are addressed: single-value publishers are no longer framed, framing is decided from the first item, the `+json` behaviour change is in the description, streamed syntax errors are 400, and the `writePiece` subtype question was measured. `ByteBody` being treated as raw is a good catch too.

Two small things inline, neither blocking.

http-server/src/main/java/io/micronaut/http/server/ResponseLifecycle.java
graemerocher: Done in 478009dfde: the streamed JsonSyntaxException now goes through the ErrorResponseProcessor with BAD_REQUEST, the "Invalid JSON" title and "Invalid JSON: " + message, the same as JsonExceptionHandler builds it for a buffered body, and the two specs now check the body shape (`_embedded.errors`, `_links.self`, content type) as well as the status.

http-netty/src/main/java/io/micronaut/http/netty/body/JsonCounter.java
graemerocher: Done in 478009dfde: after consuming the expected comma the check for a stray comma runs again on the same buffer, so `[1 , ,2]` is rejected as an unexpected comma whether it arrives whole or byte by byte; `[1 , ,2]` and `[{} , ,{}]` are added to the illegal inputs.

PR #13105, opened 2026-09-10
Remove netty server classes left behind by earlier rewrites, keeping the two micronaut-spring still needs

graemerocher:
## Summary

Removes code in `http-netty` and `http-server-netty` that no longer has a single caller
anywhere in the repository. Each candidate was checked by hand (see *Verification* below)
before being touched; nothing here changes runtime behaviour.

- **`io.micronaut.http.netty.reactive.HandlerSubscriber` (deleted)** — `@Internal`, no
  references in main, test or documentation sources. Unused since the netty write path
  moved to the `ByteBody`/`NettyWriteContext` APIs. It is also incorrect as written:
  `channelWritabilityChanged` calls `maybeRequestMore()`, which dereferences
  `subscription` unconditionally, so it would fail in the `NO_SUBSCRIPTION` /
  `NO_CONTEXT` states if the handler were ever wired into a pipeline again.
- **`io.micronaut.http.server.netty.ByteBufDelegate` (deleted)** — `@Internal`, 1040 lines
  of `ByteBuf` plumbing with no callers.
- **`io.micronaut.http.server.netty.DefaultHttpContentProcessor` (removed)** — `@Internal` and unused here; micronaut-spring's `spring-web` still constructs it and migrates to `MessageBodyReader` on its side (its migration PR is in progress). First kept for that reason, removed on review since an internal class need not wait for the `HttpContentProcessor` API removal.
- **`io.micronaut.http.server.netty.HttpContentSubscriberFactory` (kept)** — already
  `@Deprecated(forRemoval = true, since = "4.5.0")` and unused, but `public` and not
  `@Internal`, so removing it is an API break that belongs in the next major per
  MAINTAINING.md. Left as is.
- **`io.micronaut.http.server.netty.multipart.NettyFileUploadInputStream` (deleted)** —
  package-private, a leftover of the `HttpPostRequestDecoder`-era form parsing (it still
  called `FileUpload.release()`).
- **`HttpData -> byte[]` and `HttpData -> CharSequence` converters in
  `NettyConvertersSpi` (kept)** — an earlier revision removed them; they were restored in
  34bcb23f43 with `ConverterRegistrySpec` covering the registrations, since user code can still
  hand an `io.netty.handler.codec.http.multipart.HttpData` to the conversion service even
  though the server itself no longer produces one.
- **`io.micronaut.http.server.netty.HttpContentProcessor` (kept, deprecation
  strengthened)** — this is the only candidate that is `public` *and* not `@Internal`, so
  it is **not** deleted here; its `@Deprecated` is upgraded to
  `@Deprecated(forRemoval = true, since = "4.0.0")` so users get a removal warning. A
  dangling javadoc `{@link}` to the long-gone
  `io.micronaut.http.server.netty.jackson.JsonContentProcessor` is dropped at the same
  time.

## Public API impact — please review before merging

Removing a `public` type is binary incompatible even when the type is unusable in
practice. The affected types are:

| Type | Module | Visibility | Status before this PR | Action |
| --- | --- | --- | --- | --- |
| `io.micronaut.http.netty.reactive.HandlerSubscriber` | http-netty | `public` | `@Internal` | deleted |
| `io.micronaut.http.server.netty.ByteBufDelegate` | http-server-netty | `public` | `@Internal` | deleted |
| `io.micronaut.http.server.netty.DefaultHttpContentProcessor` | http-server-netty | `public` | `@Internal` | removed (`fc09a8cd59`, on review: internal, so it need not wait for the `HttpContentProcessor` API); micronaut-spring migrates its one use to `MessageBodyReader` before taking 5.2.2 |
| `io.micronaut.http.server.netty.HttpContentSubscriberFactory` | http-server-netty | `public` | `@Deprecated(forRemoval = true, since = "4.5.0")` | **kept** (API break, next major) |
| `io.micronaut.http.server.netty.multipart.NettyFileUploadInputStream` | http-server-netty | package-private | `@Internal` | deleted |
| `io.micronaut.http.server.netty.HttpContentProcessor` | http-server-netty | `public` | `@Deprecated` (not `forRemoval`, not `@Internal`) | **kept**, now `forRemoval` |

The split is deliberate: `@Internal`, package-private and already-`forRemoval` types are
deleted outright, while the one plain-public interface is only flagged, so maintainers can
decide whether it goes in this release or the next major.

Two related points for the maintainers:

- `japiCmp` (baseline `5.2.0`) reports the four deleted classes as *warnings*
  ("Is not binary compatible / Class has been removed"), not errors, so the check passes
  as-is. If you want them recorded as accepted, note that
  `buildSrc`'s `AcceptedApiChangesRule` requires a `member` field on every entry — a
  memberless entry throws `Cannot invoke "String.equals(Object)" because "this.member" is
  null`, which is why no entries were added here. (The two existing memberless entries in
  `config/accepted-api-changes.json`, lines 1197-1203, would hit the same NPE if a change
  ever matched their types.)
- No entry was added to `src/main/docs/guide/appendix/breaks.adoc`; happy to add one if
  you want these removals called out for the release. The existing 4.0 note about
  `HttpContentProcessor` being superseded by `MessageBodyHandler` is left as is, since the
  interface itself is still present.

## Verification

For every candidate, the whole repository was searched — main and test sources in every
module (Java, Groovy, Kotlin), `src/main/docs`, `test-suite*`, `benchmarks`,
`META-INF/services`, GraalVM `native-image` configuration under `src/main/resources`, and
`module-info` files (there are none in this repo):

- `git grep -n -I -E "ByteBufDelegate|HandlerSubscriber"` — only the two files themselves.
- `git grep -n -I -E "HttpContentProcessor|HttpContentSubscriberFactory|NettyFileUploadInputStream"`
  — only the files themselves, plus:
  - `config/accepted-api-changes.json:1198,1202` — entries for
    `$DefaultHttpContentProcessorResolver$Definition` and
    `$JsonHttpContentSubscriberFactory$Definition`, i.e. *different* classes removed in an
    earlier release, not these types;
  - `src/main/docs/guide/appendix/breaks.adoc:400-403` — prose describing the 4.0
    replacement of the API. `HttpContentProcessor` is kept, so this stays accurate.
- `git grep` over `*.json`, `*/META-INF/*`, `*module-info*`, `*.properties`, `*.adoc`,
  `*.txt`, `*.yml`, `*.xml` for all six names: only the two items above, plus the
  unrelated `META-INF/services/io.micronaut.core.convert.TypeConverterRegistrar` entry for
  `NettyConvertersSpi`, which is retained.
- `http-netty/src/main/resources/META-INF/native-image/.../reflect-config.json` contains
  only netty/jctools entries; `http-server-netty` has no native-image configuration.
- For the `HttpData` converters:
  `git grep -n -I -E "codec\.http\.multipart|HttpPostRequestDecoder|DiskFileUpload|MemoryFileUpload|HttpData|DefaultHttpDataFactory"`
  shows the netty multipart codec is used only by the *client*
  (`http-client/.../NettyHttpClient.java`, which encodes with its own
  `HttpPostRequestEncoder`/`HttpDataFactory` and never routes through this converter) and
  by two test helpers that build multipart requests client-side
  (`MaxRequestSizeSpec.groovy:298,359`, `test-suite/.../AbstractMicronautSpec.groovy`
  configures `DiskFileUpload`). Server-side, the only producers/consumers of netty
  `HttpData` were the classes deleted here.

- The two `HttpData` → `byte[]`/`CharSequence` converters in `NettyConvertersSpi` stay registered for application code and are tagged `@NextMajorVersion` (review).

## Testing

No behavioural change, so no new tests. Instead:

- `./gradlew classes` — BUILD SUCCESSFUL (all modules, main sources).
- `./gradlew testClasses` — BUILD SUCCESSFUL (all modules, test sources: Java, Groovy,
  Kotlin, all `test-suite*` projects).
- `./gradlew :micronaut-http-netty:checkstyleMain :micronaut-http-server-netty:checkstyleMain`
  — BUILD SUCCESSFUL.
- `./gradlew :micronaut-http-netty:checkstyleTest :micronaut-http-server-netty:checkstyleTest`
  — BUILD SUCCESSFUL.
- `./gradlew :micronaut-http-server-netty:test :micronaut-http-netty:test :micronaut-http:test -q`
  — BUILD SUCCESSFUL. Totals from the JUnit XML: http-netty 148 tests / 0 failures,
  http-server-netty 1153 tests / 0 failures / 19 skipped, http 1857 tests / 0 failures.
- `./gradlew :micronaut-http-server-netty:test --rerun-tasks` — 1153 tests, 0 failures.
- Binary compatibility: `./gradlew :micronaut-http-netty:japiCmp :micronaut-http-server-netty:japiCmp`
  (the repo's japicmp-based check, baseline `5.2.0`) — BUILD SUCCESSFUL. The reports list
  the four deleted classes as removals at *warning* severity; see the note above.

One flaky failure was observed on the first full `:micronaut-http-server-netty:test` run:
`io.micronaut.http.server.netty.ssl.FileCertificateProviderTest "[3] format = PEM"` failed
with `TLSV1_ALERT_CERTIFICATE_REQUIRED` from the client's TLS handshake. It is unrelated to
this change (an mTLS certificate-provider test that touches none of the removed code), and
it passed when re-run on its own (`--tests
"io.micronaut.http.server.netty.ssl.FileCertificateProviderTest"`, 7 tests, 0 failures) and
again in the subsequent full `--rerun-tasks` run of the whole module.

## Out of scope

- `io.micronaut.http.server.netty.HttpContentProcessor` is not deleted — see the public
  API section.
- No `breaks.adoc` entry and no `config/accepted-api-changes.json` entries were added; both
  are maintainer decisions and the build is green without them.
- The pre-existing checkstyle violations mentioned in the contributor notes
  (`NettyHttpServerConfiguration` FileLength, `NettyHttpRequest` duplicate import) are
  untouched.


graemerocher:
Removing the two `HttpData` registrations is an observable runtime compatibility break, even if the server no longer produces `HttpData` internally. `NettyConvertersSpi` is loaded through `META-INF/services/io.micronaut.core.convert.TypeConverterRegistrar`, so its conversions are available to application code through the public `ConversionService` API; `@Internal` on the registrar implementation does not make those registrations invisible.

I verified this with a completed `MemoryFileUpload` in `ConverterRegistrySpec`: on this PR, `ctx.getBean(ConversionService).convert(upload, byte[])` returns `Optional.empty`; restoring the two removed registration blocks makes the same test return the original `test` bytes. Consequently the PR's statement that nothing changes runtime behavior is not true for callers that convert Netty multipart data explicitly.

Please retain these converter registrations, or handle their removal as a separately documented compatibility change rather than dead-code deletion.

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConvertersSpi.java
```diff
@@ -73,40 +70,6 @@ public void register(MutableConversionService conversionService) {
...
-                        return Optional.of(byteBuf.toString(context.getCharset()));
-                    } catch (Exception e) {
-                        context.reject(e);
-                        return Optional.empty();
-                    }
-                }
-        );
     }
 
 }
```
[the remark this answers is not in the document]
graemerocher: Both registrations are restored in 34bcb23f43, with a comment recording why they stay. On the deprecation alternative: it is not available here. The conversions are lambdas passed to `addConverter` inside a method body, so there is no API element for `@Deprecated` to attach to and no caller would ever see a warning. The only user-visible deprecation would be prose in the upgrade notes, which is a maintainer call, so it is recorded as a follow-up rather than done here.

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentProcessor.java
```diff
@@ -34,9 +34,10 @@
  *
  * @author Graeme Rocher
  * @since 1.0
- * @deprecated Use the {@link io.micronaut.http.body.MessageBodyReader} API instead
+ * @deprecated Use the {@link io.micronaut.http.body.MessageBodyReader} API instead. This interface
+ * is no longer used or implemented by the server and will be removed in a future release.
```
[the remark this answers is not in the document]
graemerocher: Updated in 34266098b5, with one correction to the premise: the repository has no timeline convention to align to. Only 4 of roughly 100 `forRemoval` javadocs mention a timeline, all in `RequestLifecycle`. The actual convention is `since = <deprecation version>` plus a replacement pointer, with the removal recorded in `breaks.adoc` when it happens. The javadoc now states the granularity the project does commit to, a future major rather than a 5.x minor, instead of the contentless "a future release".

graemerocher:
Confirmed and fixed in 34bcb23f43. You are right that this is a compatibility break rather than dead-code removal: `NettyConvertersSpi` is loaded through `META-INF/services/io.micronaut.core.convert.TypeConverterRegistrar`, so the registrations are reachable from application code via the public `ConversionService`, and `@Internal` on the registrar does not change that. Both registrations are restored, and `ConverterRegistrySpec` now covers a completed `MemoryFileUpload` converting to `byte[]` and `CharSequence`; on the previous code that test reproduces your result exactly, `Optional.empty` with `Completed: true`. I have also corrected the PR description, which claimed the change was behaviour-neutral. It now states explicitly what the PR does and does not remove.

dstepanov (commented):
Nothing in core uses these classes and none of them are beans, but `DefaultHttpContentProcessor` still has a caller outside this repo.

1. **Removing it breaks micronaut-spring `spring-web`.** [`MicronautNettyChannelResolver`](https://github.com/micronaut-projects/micronaut-spring/blob/6.2.x/spring-web/src/main/java/io/micronaut/spring/web/reactive/MicronautNettyChannelResolver.java#L68) is a `@Singleton @Primary` bean on both 6.2.x and 6.1.x. It imports `DefaultHttpContentProcessor` and `HttpContentProcessor`, and `resolveContentProcessor` calls `new DefaultHttpContentProcessor(nettyHttpRequest, serverConfiguration)`.
   - spring-web won't compile against a core release that includes this.
   - With existing jars, calling that method throws `NoClassDefFoundError`.
   - japiCmp stays green only because the class is `@Internal`.
   - Options: keep `DefaultHttpContentProcessor` in 5.2.x, or first change micronaut-spring to stop constructing it (and deprecate `ChannelResolver.resolveContentProcessor`).
2. **`HttpContentProcessor` marked `forRemoval`** also reaches micronaut-spring, because its public `ChannelResolver` returns `Optional<HttpContentProcessor>`. It deserves a heads-up or tracking issue over there.
3. **`HttpContentSubscriberFactory` is public and not `@Internal`.** It has been deprecated for removal since 4.5.0, and I found no users in the org. Removing it in a minor release still deserves a `breaks.adoc` entry or an accepted API change.
4. **The PR description is out of date.** It still lists the `HttpData` converters as removed, but they were restored in 34bcb23f43 (with the new `ConverterRegistrySpec` tests).

I also checked the other classes for `META-INF/services`, native-image config, `@TypeHint`, `@Requires(classes=...)`, string class names, and org-wide code search. No hits apart from the ones above.

graemerocher:

---

The patch removes a public interface in the 5.2.1 patch line, creating a source and binary compatibility break for consumers that still implement it.


- [P1] Retain the factory until the next major release — http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpContentSubscriberFactory.java:27-28
  When a 5.2.x application or dependency still declares a bean implementing this public interface, upgrading to 5.2.1 makes source recompilation fail and can make existing bytecode fail linkage with `NoClassDefFoundError`, even if the server no longer invokes it. This is a breaking API deletion in a patch branch, which `MAINTAINING.md:76-78` says must not leak into patch/minor releases; retain the deprecated interface until the next major release.

---

**Disposition:** Agreed, and it overrides my earlier remedy: I had added a `breaks.adoc` entry, but MAINTAINING.md is clear that breaking changes belong in the next major. `HttpContentSubscriberFactory` is restored in `bb75be18c9`, deprecated for removal as before.

graemerocher:
Thanks, the micronaut-spring caller is exactly what the org-wide search should have turned up and did not.

1. **`DefaultHttpContentProcessor` — restored** (`1334801de4`). It stays, `@Internal` and now `@Deprecated(forRemoval = true)` alongside `HttpContentProcessor`, with a javadoc note naming the spring-web caller so nobody deletes it again before micronaut-spring stops constructing it. The PR title and description are updated to match.
2. **`HttpContentProcessor` `forRemoval`** — kept. With the implementation restored, spring-web compiles; the `forRemoval` only produces a removal warning there, which is the intended heads-up. A tracking issue for micronaut-spring (deprecate `ChannelResolver.resolveContentProcessor`, stop constructing `DefaultHttpContentProcessor`) is drafted and will be opened when the current CI hold is lifted.
3. **`HttpContentSubscriberFactory`** — restored (`4ef3f29f35`). I first added a `breaks.adoc` entry, but a second review pass pointed at MAINTAINING.md: breaking changes target the next major and must not leak into a patch or minor, and a public non-`@Internal` interface removal is one however dead it is. It stays, deprecated for removal since 4.5.0, and goes with the next major.
4. **Description** — corrected: the `HttpData` converters are listed as kept (restored in 34bcb23f43 with `ConverterRegistrySpec`), and the summary table now shows `DefaultHttpContentProcessor` as kept.

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConvertersSpi.java
```diff
@@ -74,6 +74,8 @@ public void register(MutableConversionService conversionService) {
                 }
         );
 
+        // the server itself no longer produces netty HttpData instances, but these two conversions are
+        // reachable from application code through the public ConversionService API, so they stay registered
```
yawkat: `@NextMajorVersion`: Remove

http-server-netty/src/main/java/io/micronaut/http/server/netty/DefaultHttpContentProcessor.java
```diff
@@ -28,10 +28,17 @@
 /**
  * This class will handle subscribing to a stream of {@link io.netty.handler.codec.http.HttpContent}.
  *
+ * <p>Nothing in this repository uses this class any more, but it is still constructed directly by
+ * {@code MicronautNettyChannelResolver} in micronaut-spring (spring-web), so it has to stay until
+ * that module stops depending on it and the {@link HttpContentProcessor} API is removed.
+ *
  * @author Graeme Rocher
  * @since 1.0
+ * @deprecated Use the {@link io.micronaut.http.body.MessageBodyReader} API instead. Kept only for
```
yawkat: we can remove this one earlier since it's internal

yawkat (approved):

http-server-netty/src/main/java/io/micronaut/http/server/netty/converters/NettyConvertersSpi.java
graemerocher: Tagged in fc09a8cd59: the two registrations are in a private `registerHttpDataConverters` annotated `@NextMajorVersion("Remove the HttpData converters, the server no longer produces HttpData")`.

http-server-netty/src/main/java/io/micronaut/http/server/netty/DefaultHttpContentProcessor.java
graemerocher: Removed in fc09a8cd59 (it's `@Internal`). micronaut-spring's `MicronautNettyChannelResolver` still constructs it, so that module needs its migration to `MessageBodyReader` before it picks up 5.2.2; the migration is in progress on its side.

PR #13107, opened 2026-09-10
Netty server bootstrap and shutdown fixes

graemerocher:
## Summary

- **Size and name the acceptor ("parent") event loop group as an acceptor.** Unless
  `micronaut.netty.event-loops.parent` is configured, `NettyHttpServer` creates the acceptor group
  itself from a `DefaultEventLoopGroupConfiguration` whose thread count is `0`, which Netty expands
  to `max(1, 2 * availableProcessors)`. A group that only accepts connections was therefore started
  with 24 event loops (and 24 selectors) on a 12-core machine. The acceptor group now defaults to a
  single thread. An explicitly configured thread count still wins, both via
  `micronaut.netty.event-loops.parent.num-threads` (registry-provided group) and via the legacy
  `micronaut.server.netty.parent.threads`.
- **Route the thread count of `DefaultNettyEmbeddedServerFactory.createEventLoopGroup` through
  `DefaultEventLoopGroupRegistry.numThreads`**, so that a `threadCoreRatio` configured for a group is
  honoured on this path instead of being silently ignored.
- **Give non-default event loop groups their own thread pool name.** Groups created by
  `DefaultNettyEmbeddedServerFactory` all used the shared Netty thread factory, so acceptor threads
  were named `default-eventLoopGroup-*` and were indistinguishable from worker threads in a thread
  dump. They are now named after the group configuration (`parent-eventLoopGroup-*`), which matches
  what `DefaultEventLoopGroupRegistry` already does for registry-provided groups.
  `NettyThreadFactory.EventLoopCustomizableThreadFactory` (`@Internal`) gained a
  `customizeForEventLoop(String poolName)` overload for this; the no-arg overload is unchanged.
- **Close the listener channels in `NettyHttpServer.stopInternal`.** `stop()` and `stopServerOnly()`
  relied entirely on the event loop groups being shut down to release the listening sockets:
  - the acceptor group is only shut down when this server created it, and `stop()` does not wait for
    `shutdownGracefully()`, so the port stayed bound for the whole shutdown quiet period after
    `stop()` returned - `stop()` followed by `start()` on a fixed port could fail to bind;
  - when `micronaut.netty.event-loops.parent` is configured the acceptor group comes from the
    registry and is not owned by the server, so nothing closed the listening channel at all: the
    server kept accepting connections after `stop()`, and `stopServerOnly()` (used by CRaC to
    quiesce the server before a checkpoint) left the listening socket open.

  Each listener now closes its channel at the start of `stopInternal`, before any owned event loop
  group is shut down, and waits for the close. Completing the close promise is not the last step -
  the event loop still has to deregister the channel from its selector before the JDK releases the
  socket - so the close is followed by a round trip through that event loop. Without that round trip
  the port was observed to stay unbindable for another ~100-300 ms in roughly half of the runs when
  the server owns the acceptor group (the group shutdown that follows keeps the loop from getting
  back to its selector promptly).

## Testing

Fail-before evidence, produced by checking out the base version of the changed main sources, keeping
only the package-private `NettyHttpServer.getParentGroup()` test accessor, and running the new specs:

- `ParentEventLoopGroupSpec` (JDK 25, 12 cores):

  ```
  io.micronaut.http.server.netty.ParentEventLoopGroupSpec acceptor event loop group uses a single, distinctly named thread by default FAILED

    Condition not satisfied:

    executorCount(parentGroup) == 1
    |             |            |
    24            |            false
                  <io.netty.channel.MultiThreadIoEventLoopGroup@67e77f52 children=inaccessible readonlyChildren=inaccessible terminatedChildren=inaccessible terminationFuture=inaccessible chooser=inaccessible>
        at io.micronaut.http.server.netty.ParentEventLoopGroupSpec.acceptor event loop group uses a single, distinctly named thread by default(ParentEventLoopGroupSpec.groovy:31)
  ```

  The other two features of that spec (`micronaut.netty.event-loops.parent.num-threads` and
  `micronaut.server.netty.parent.threads` are honoured) guard the "explicit configuration still
  wins" requirement; the legacy one also fails on the base version on its thread-name assertion.

- `NettyStopSpec` - all three new features fail on the base version with the same shape:

  ```
  the port is free once stop returns FAILED
  the port is free once stop returns with a shared acceptor event loop group FAILED
  the port is free once stopServerOnly returns with a shared acceptor event loop group FAILED

    Condition failed with Exception:

    canBind(port)
    |       |
    |       53904
    java.net.BindException: Address already in use
        at java.base/sun.nio.ch.Net.bind(Net.java:522)
        ...
        at io.micronaut.http.server.netty.NettyStopSpec.canBind(NettyStopSpec.groovy:26)
        at io.micronaut.http.server.netty.NettyStopSpec.the port is free once stop returns(NettyStopSpec.groovy:67)
  ```

  The two "shared acceptor event loop group" features assert, before stopping, that the group the
  server uses is the one the `EventLoopGroupRegistry` provides, so the not-owned path is genuinely
  exercised.

What was run (all on JDK 25, macOS):

- `./gradlew :micronaut-http-server-netty:test --tests '*ParentEventLoopGroupSpec*'` - 3 tests, pass.
- `./gradlew :micronaut-http-server-netty:test --tests '*NettyStopSpec*' --rerun-tasks` - 5 tests,
  pass; repeated 4 times to confirm the port assertions are not flaky.
- `./gradlew :micronaut-http-server-netty:test --tests '*Stop*' --tests '*Shutdown*' --tests
  '*MultiServer*' --tests '*EventLoop*'` - 19 tests, pass (`GracefulShutdownSpec` 8,
  `NettyStopSpec` 5, `ParentEventLoopGroupSpec` 3, `NettyStartStopSpec` 1, `NettyMultiServerSpec` 1,
  `ExecutorServiceWithMultipleEventLoopsSpec` 1). Graceful shutdown is not regressed.
- `./gradlew :micronaut-http-server-netty:test` - full module suite, 204 classes / 1162 tests, all
  pass. A first run of the suite had 5 failures in `CustomStaticMappingLocalSpec` (4, all
  `ResponseClosedException` instead of `HttpClientResponseException`) and
  `FileCertificateProviderTest` (1, `ServerStartupException: Unable to start Micronaut server on
  *:0`). Both specs pass on their own
  (`--tests '*FileCertificateProviderTest*' --tests '*CustomStaticMappingLocalSpec*'`, 11 tests,
  pass) and both pass in a clean `--rerun-tasks` run of the whole suite, so they are pre-existing
  flakes under the module's parallel test forks and unrelated to these changes.
- `./gradlew :micronaut-http-netty:test` - full module suite, 25 classes / 148 tests, all pass.
- `./gradlew :micronaut-http-netty:checkstyleMain :micronaut-http-server-netty:checkstyleMain` -
  pass. The two reported violations are pre-existing and in files this PR does not touch
  (`NettyWebSocketSession.markClosing` DesignForExtension, `NettyHttpServerConfiguration` FileLength).

## Out of scope

- **Active child connections are not closed by `stop()`.** In-flight requests on accepted
  connections continue to be served by the (usually shared) worker group, exactly as before. Closing
  them from `stop()` would be a behavioural change of its own and overlaps with what
  `GracefulShutdownCapable` already provides, so it was left alone.
- **Bind failure of a second listener.** When one listener binds and a later one fails, `bind()`
  calls `stopInternal(false)` before `activeListeners` has been assigned, so the already bound
  listeners are still not closed on that path. Fixing it means restructuring where the bind failure
  is handled in `start()`, which is outside what this PR changes.
- **The residual deregistration round trip is per listener event loop.** It is skipped when the loop
  is already shutting down or when `stopInternal` runs on that loop itself, so in those cases the
  socket release still depends on the event loop group shutdown.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/channel/NettyThreadFactory.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/channel/NettyThreadFactory.java
@@ -122,6 +122,11 @@ public class NettyThreadFactory {
      */
     @Internal
     public static final class EventLoopCustomizableThreadFactory implements ThreadFactory {
+        /**
+         * The thread pool name used when no event loop group name is given.
+         */
+        public static final String DEFAULT_POOL_NAME = "default-eventLoopGroup";
+
         private final boolean daemon;
         private final int priority;
         private final boolean nonBlocking;
@@ -135,7 +140,19 @@ public class NettyThreadFactory {
         }
 
         public ThreadFactory customizeForEventLoop() {
-            return new CustomizedThreadFactory(daemon, priority, nonBlocking);
+            return customizeForEventLoop(DEFAULT_POOL_NAME);
+        }
+
+        /**
+         * Create a thread factory for a single event loop group, using the given thread pool name
+         * so that its threads can be told apart from those of other event loop groups.
+         *
+         * @param poolName The thread pool name
+         * @return The thread factory for the event loop group
+         * @since 5.2.0
+         */
+        public ThreadFactory customizeForEventLoop(String poolName) {
+            return new CustomizedThreadFactory(poolName, daemon, priority, nonBlocking);
         }
 
         @Override
@@ -147,8 +164,8 @@ public class NettyThreadFactory {
     private static final class CustomizedThreadFactory extends DefaultThreadFactory {
         private final boolean nonBlocking;
 
-        public CustomizedThreadFactory(boolean daemon, int priority, boolean nonBlocking) {
-            super("default-eventLoopGroup", daemon, priority);
+        public CustomizedThreadFactory(String poolName, boolean daemon, int priority, boolean nonBlocking) {
+            super(poolName, daemon, priority);
             this.nonBlocking = nonBlocking;
         }
 
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/DefaultNettyEmbeddedServerFactory.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/DefaultNettyEmbeddedServerFactory.java
@@ -28,6 +28,7 @@ import io.micronaut.core.order.OrderUtil;
 import io.micronaut.core.util.CollectionUtils;
 import io.micronaut.http.body.MessageBodyHandlerRegistry;
 import io.micronaut.http.codec.MediaTypeCodecRegistry;
+import io.micronaut.http.netty.channel.DefaultEventLoopGroupRegistry;
 import io.micronaut.http.netty.channel.EventLoopGroupConfiguration;
 import io.micronaut.http.netty.channel.EventLoopGroupFactory;
 import io.micronaut.http.netty.channel.EventLoopGroupRegistry;
@@ -289,10 +290,26 @@ public class DefaultNettyEmbeddedServerFactory
     @Override
     public EventLoopGroup createEventLoopGroup(EventLoopGroupConfiguration configuration) {
         return new MultiThreadIoEventLoopGroup(
-            configuration.getNumThreads(),
-            nettyThreadFactory,
+            DefaultEventLoopGroupRegistry.numThreads(configuration),
+            threadFactory(configuration),
             eventLoopGroupFactory.createIoHandlerFactory(configuration)
         );
+    }
+
+    /**
+     * Resolve the thread factory for the given event loop group configuration. Groups other than
+     * the default one get their own thread pool name so that their threads are distinguishable in
+     * a thread dump.
+     *
+     * @param configuration The event loop group configuration
+     * @return The thread factory to create the event loop threads with
+     */
+    private ThreadFactory threadFactory(EventLoopGroupConfiguration configuration) {
+        if (nettyThreadFactory instanceof NettyThreadFactory.EventLoopCustomizableThreadFactory customizable &&
+            !EventLoopGroupConfiguration.DEFAULT.equals(configuration.getName())) {
+            return customizable.customizeForEventLoop(configuration.getName() + "-eventLoopGroup");
+        }
+        return nettyThreadFactory;
     }
 
     @Override
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
@@ -81,6 +81,7 @@ import io.netty.channel.socket.ServerSocketChannel;
 import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.handler.codec.quic.QuicSslContext;
 import io.netty.handler.ssl.SslContext;
+import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.GlobalEventExecutor;
@@ -542,7 +543,8 @@ public class NettyHttpServer implements NettyEmbeddedServer {
             return source;
         }
         return new DefaultEventLoopGroupConfiguration(
-            NettyHttpServerConfiguration.Parent.NAME,
+            // keep a name the user configured; only the implicit acceptor group is called "parent"
+            parent == null ? NettyHttpServerConfiguration.Parent.NAME : source.getName(),
             DEFAULT_PARENT_THREADS,
             source.getThreadCoreRatio(),
             source.getIoRatio().orElse(null),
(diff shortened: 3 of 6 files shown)


graemerocher:
The listener close does not fully quiesce `stopServerOnly()`: an already accepted keep-alive connection can still dispatch brand-new requests after this method returns.

I reproduced this on `56552e68e` with a raw HTTP/1.1 socket: send `GET /ping` and receive `200 OK`, call `server.stopServerOnly()`, then send a second `GET /ping` on the same socket. The second request also receives `200 OK`. The new port assertion passes because the listener is closed, but `Listener.activeConnections` is left open on the shared worker group.

This is different from allowing a request that was already in flight to finish, as described in the out-of-scope note: clients holding persistent connections can initiate more application work after the server reports that it has stopped. In particular, that means the CRaC path is not actually quiescent and may reach the checkpoint with live TCP channels. Please drain/close the tracked child connections before `stopServerOnly()` completes (or otherwise prevent them from accepting new requests), and add a regression test that attempts a second request on an existing keep-alive connection.

graemerocher:
Confirmed and fixed in 705e0fcb5d. `stopInternal` now closes each listener's tracked child connections and awaits them, straight after the listener channels and before any owned event loop group shutdown, with the usual guard against waiting on a channel whose own event loop is the calling thread.

On the tension you might expect with graceful shutdown: there is not one, they compose. `shutdownGracefully()` remains the explicit draining phase and is untouched, and by the time it completes `activeConnections` is empty so the new close has nothing to do. `stop()` and `stopServerOnly()` are the abrupt end of the lifecycle and must not leave a channel that can start new work. Draining from `stopInternal` is not viable anyway for a synchronous method CRaC depends on, since a connection with an open streaming response only drains if the client cooperates and could block the checkpoint indefinitely.

The regression test is the one you described, driving a raw HTTP/1.1 socket: request, stop, second request on the same socket must not be served, run for both `stopServerOnly()` and `stop()`. It fails on the previous code with `HTTP/1.1 200 OK|pong` on the second request. I ran `NettyStopSpec` seven times to confirm it is not flaky, and `GracefulShutdownSpec` still passes. The now-obsolete out-of-scope note in the description has been struck through.

dstepanov (approved):
Approving. Connection bookkeeping has no leaks, stopping twice is idempotent, a shared acceptor group is never shut down, and the new specs fail on the old code.

On the 6-hour timeout: I don't think this PR caused it.
- In the cancelled run, `NettyStopSpec`, `ParentEventLoopGroupSpec` and `EventLoopGroupCreationSpec` all passed about 10 minutes in, and 22,109 tests ran with 0 failures.
- The log then went silent for 5h24m in the root `:jacocoTestReport` ("Generating aggregate Jacoco reports"), a step that takes about 0.1s on green runs.
- The unrelated Netty 4.2.18 bump stalled in the same step in the same window.

Please look at the first point before merging:

1. **`stop()` can now block forever.** [`closeServerChannel`](https://github.com/micronaut-projects/micronaut-core/blob/feba70822e36397d63035e70bda2fe729735db1e/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java#L1098-L1118) and [`closeConnections`](https://github.com/micronaut-projects/micronaut-core/blob/feba70822e36397d63035e70bda2fe729735db1e/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java#L1128-L1143) wait with `awaitUninterruptibly()` and no timeout.
   - If a request handler blocks a worker event loop (a sleep, a latch, a blocking call), the close task can't run there.
   - `stop()` then hangs, and with it JVM shutdown hooks or a CRaC checkpoint, and it can't be interrupted.
   - A bounded wait (e.g. the group's `shutdownTimeout`), then log and continue, avoids this.
2. **In-flight requests are cut off by default.** `closeConnections` closes busy connections too, and `GracefulShutdownListener` only runs with `micronaut.lifecycle.graceful-shutdown.enabled=true`, which is off by default.
   - A plain SIGTERM now drops requests still running on the blocking executor, and h2/h3 connections close without GOAWAY.
   - Before, the worker group's quiet period usually let a nearly finished response go out.
   - This is worth a release note, or closing only idle connections.
3. **A `/stop`-style endpoint loses its own response.** When `stop()` is called on the event loop, the deadlock guard skips the wait but still runs `channel.close()` inline. The caller's connection closes before its response is written, and the port isn't guaranteed free when `stop()` returns.
4. **The legacy ratio setting is ignored.** An explicit `micronaut.server.netty.parent.thread-core-ratio` can't be told apart from its 1.0 default, because of [`getNumThreads() != 0`](https://github.com/micronaut-projects/micronaut-core/blob/feba70822e36397d63035e70bda2fe729735db1e/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java#L539), so the acceptor gets 1 thread. This affects only the deprecated path.
5. **The keep-alive case can pass for the wrong reason.** In [`NettyStopSpec.groovy#L158`](https://github.com/micronaut-projects/micronaut-core/blob/feba70822e36397d63035e70bda2fe729735db1e/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/NettyStopSpec.groovy#L158), a 30s `SocketTimeoutException` is an `IOException` and maps to the "closed" value.

Missing tests: `stop()` from a request handler, or while an event loop is blocked; in-flight behaviour with graceful shutdown off; HTTPS, h2, h3 and multiple listeners.

graemerocher:
The Java CI job for this branch has now twice run far past its normal length after every module test task had completed (the earlier run was cancelled by the 6-hour runner timeout at the root Jacoco step; the re-run was cancelled today after sitting in the Gradle step for nearly two hours against a normal 65 minutes). The previous commit's job was green.

Locally, on the head commit, `:micronaut-http-server-netty:test` (1164 tests) and the HTTP server TCK (264 tests, forced past the build cache) both pass and Gradle exits promptly, so this does not reproduce on macOS. The only production code added since the green run is the pair of unbounded waits in `stop()` for the listener channel and the accepted connections to close.

`52dad3fe53` bounds those waits by the acceptor group's shutdown timeout (15 s by default) and logs a warning naming the channel if it is reached. That is right on its own terms, since `stop()` should never be able to block indefinitely on a channel that does not close, and if the waits are what stalls the runner this fixes it. If the job still stalls after this, the waits are ruled out and the cause is elsewhere; I will look at the run output either way.

graemerocher:

---

Shutdown can still leave a newly accepted connection alive when its worker registration races with the new connection-closing pass. The parent configuration rewrite also discards an explicitly configured group name on one creation path.


- [P1] Track accepted channels before worker registration — http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java:1130-1133
  With the normal shared worker group, a connection accepted immediately before shutdown can be queued for `childGroup.register` without having run `Listener.initChannel` yet. The parent-event-loop round trip does not drain the worker queue, so this iteration can miss the channel; it may initialize afterward and remain usable because the shared worker group is not shut down. Track channels before asynchronous registration or synchronize with the worker group so `stop()` and `stopServerOnly()` cannot return with a live connection.

- [P3] Preserve the configured parent group name — http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java:543-545
  When `micronaut.server.netty.parent.event-loop-group` selects a custom name, no matching registry group exists, and the thread count remains zero, this fallback replaces the configured name with `parent`. Consequently the new naming logic produces `parent-eventLoopGroup-*` rather than a name derived from the configured group; the same configuration preserves its name when a nonzero thread count is supplied. Use the source configuration's name here.

---

**Disposition:** Reviewed at the pre-push head. Both verified and fixed locally on top of `52dad3fe53`: `fe5f20ef52` does a round trip through every worker event loop before the closing pass so queued registrations have run, and keeps a configured acceptor group name (spec row fails on the previous commit).

PR #13112, opened 2026-09-10
Add JMH benchmarks for multipart, form, chunked and streaming JSON server scenarios

graemerocher:
## Summary
- Adds `ServerScenariosBenchmark` next to the existing `ControllersBenchmark`/`FullHttpStackBenchmark`, driving the real server pipeline through an `EmbeddedChannel`.
- Scenarios: multipart upload bound as `CompletedFileUpload` and as `byte[]` (content-length and chunked), url-encoded form bound to a record with a list field, ~30 KiB JSON array bound as `List<Item>` and as `Flux<Item>` (content-length and chunked), `Flux<Item>` response as JSON array and as `application/x-json-stream`, 64 KiB `byte[]` response.
- Establishes a baseline for the Netty server hardening work; baseline numbers are recorded in the review plan.
## Testing
- `./gradlew :benchmarks:jmhJar` and ran `ServerScenariosBenchmark` with `-Dmicronaut.python.enabled=false -Djmh.executor=CUSTOM -Djmh.executor.class=io.micronaut.http.server.stack.JmhFastThreadLocalExecutor`; every scenario verifies its response in `@Setup`.
## Out of scope / follow-ups
- The benchmark JVM needs `micronaut.python.enabled=false` on JDKs without the GraalPy language; consider defaulting that in the jmh task.

### DIFF
--- a/benchmarks/src/jmh/java/io/micronaut/http/server/stack/NettyUtil.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/server/stack/NettyUtil.java
@@ -5,6 +5,9 @@ import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.embedded.EmbeddedChannel;
 
+import java.util.ArrayList;
+import java.util.List;
+
 final class NettyUtil {
     static ByteBuf readAllOutboundContiguous(EmbeddedChannel clientChannel) {
         ByteBuf requestBytes = PooledByteBufAllocator.DEFAULT.buffer();
@@ -16,6 +19,23 @@ final class NettyUtil {
             requestBytes.writeBytes(part);
         }
         return requestBytes;
+    }
+
+    /**
+     * Read every outbound buffer as its own part, in order. Unlike
+     * {@link #readAllOutboundContiguous} this keeps the boundaries the codec produced, so a caller
+     * can deliver each part as a separate read.
+     */
+    static List<ByteBuf> readAllOutboundParts(EmbeddedChannel clientChannel) {
+        List<ByteBuf> parts = new ArrayList<>();
+        while (true) {
+            ByteBuf part = clientChannel.readOutbound();
+            if (part == null) {
+                break;
+            }
+            parts.add(part);
+        }
+        return parts;
     }
 
     static ByteBuf readAllOutboundComposite(EmbeddedChannel channel) {
--- a/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java
@@ -94,7 +94,13 @@ public class ServerScenariosBenchmark {
 
         ApplicationContext ctx;
         EmbeddedChannel channel;
-        ByteBuf requestBytes;
+        /**
+         * The encoded request, one buffer per message the client codec wrote. Each is delivered in
+         * its own read cycle, so a chunked request reaches the server as separate reads and takes
+         * its streaming path, instead of being coalesced into one read that the server would treat
+         * as a fully available body.
+         */
+        List<ByteBuf> requestParts;
         ByteBuf responseBytes;
 
         @Setup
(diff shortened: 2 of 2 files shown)


dstepanov (approved):
Approving: the measurement mechanics are sound.
- **Compiles:** `:benchmarks:compileJmhJava` succeeds.
- **Runs and verifies on the calling thread:** I drove every scenario 2–3k times through `Holder`. All ten completed on the caller thread, so the spin-wait never fired, and each response is byte-verified.
- **No per-iteration pooled-memory growth:** `retainedDuplicate()` reuse is correct and outbound parts are released.
- **Bindings match the names:** `CompletedFileUpload`, `byte[]`, and the `Person` record with its list.

Please look at the first point before this becomes the baseline:

1. **The three `*_CHUNKED` scenarios never reach the streaming request path.**
   - **Where:** [`send()`](https://github.com/micronaut-projects/micronaut-core/blob/2159b8b325c24e9eeeef52b2cbadd872ded6278c/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java#L132-L144) writes the whole encoded request with a single `writeInbound`.
   - **Why:** HTTP/1.1 has no flow-control handler, so the full body, `LastHttpContent` included, arrives before `readComplete`. `PipeliningServerHandler`'s optimistic handler then hands the controller an already buffered body. My probe confirmed this for `MULTIPART_FILE_CHUNKED`, `JSON_POST_LARGE_LIST_CHUNKED` and `JSON_POST_LARGE_LIST_REACTIVE_CHUNKED` ([#L244](https://github.com/micronaut-projects/micronaut-core/blob/2159b8b325c24e9eeeef52b2cbadd872ded6278c/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java#L244), [#L297](https://github.com/micronaut-projects/micronaut-core/blob/2159b8b325c24e9eeeef52b2cbadd872ded6278c/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java#L297), [#L312](https://github.com/micronaut-projects/micronaut-core/blob/2159b8b325c24e9eeeef52b2cbadd872ded6278c/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java#L312)).
   - **Effect:** chunked vs. content-length differs only in chunk-framing parsing. The javadoc's "streaming path through the form demuxer" and "incremental JSON parsing" claims aren't exercised, so changes to streaming, backpressure or incremental parsing won't show up in this baseline.
   - **Fix:** split the request across at least two `writeInbound` calls (head plus first chunk, then the rest), or rename the scenarios.
2. **Verification may dominate the small scenarios.** [`exchangeAndVerify`](https://github.com/micronaut-projects/micronaut-core/blob/2159b8b325c24e9eeeef52b2cbadd872ded6278c/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java#L173-L198) compares every response byte on every call. For `LARGE_TEXT_RESPONSE` (64 KiB, about 9 µs unwarmed), that is plausibly a big share of the measured time and dilutes server-side gains. Checking length only, or every Nth call, would help. This is an estimate; I didn't measure the split.
3. **Not the same setup as `FullHttpStackBenchmark`:** no `channel.freezeTime()`, and `main()` doesn't pass the FastThreadLocal thread-factory args that `FullHttpStackBenchmark` does. Running via `main` silently uses plain threads, which changes pooled-allocator behaviour.
4. **The description overstates coverage.** There is no chunked `byte[]` multipart and no content-length `Flux` scenario.
5. **Pre-existing, affects all benchmarks:** `NettyUtil.readAllOutboundContiguous` doesn't release the parts it copies, so paranoid leak detection reports leaks from setup.

graemerocher:

---

The benchmark compiles conceptually, but its chunked cases coalesce the entire wire request into one read cycle. This bypasses the server streaming request path central to several newly added scenarios and makes their measurements misleading.


- [P2] Preserve read cycles for chunked request scenarios — benchmarks/src/jmh/java/io/micronaut/http/server/stack/ServerScenariosBenchmark.java:132-134
  When any `*_CHUNKED` scenario runs, the encoded request has already been coalesced into one `ByteBuf`, and this single `writeInbound` call produces only one channel read cycle. Consequently, `PipeliningServerHandler.OptimisticBufferingInboundHandler` receives `LastHttpContent` before `channelReadComplete` and treats the request as a fully available body, so these benchmarks do not exercise the streaming/devolution paths they claim to measure. Keep the encoded chunks separate and deliver them in separate read cycles.

---

**Disposition:** Verified and important for every number measured against this harness: the chunked scenarios were coalesced into one read, so the server took its fully-available path. Fixed locally in `f85f137a15`: one codec buffer per `writeInbound`, and setup refuses a `*_CHUNKED` scenario that would arrive in fewer than two reads. The chunked baseline rows need re-taking once this lands.

PR #13113, opened 2026-09-10
Send a streamed element and its separator as one message

graemerocher:
## Summary

- `ConcatenatingSubscriber` pushed the `[`, `,` and `]` of a streamed JSON array to the shared
  buffer as `ReadBuffer`s of their own, separate from the element bytes. On the netty side every
  buffer that reaches `StreamingOutboundHandler.add0` becomes its own `DefaultHttpContent` and asks
  for a flush, so a stream of N elements produced **2N+2** outbound messages: a separator message
  and an element message per element, then the closing bracket, then an empty `LastHttpContent`.
  Under chunked transfer encoding that is also a five byte `1\r\n,\r\n` chunk on the wire per
  element. `ConcatenatingSubscriber.forward` now takes the leading separator and, when the element
  is an `AvailableByteBody` (the case for anything the message body writers have already
  serialised), composes the two into one buffer via `ReadBufferFactory.compose`. A streaming
  component body still gets its separator as a separate buffer, as before.
- The trailing bytes now travel with the completion signal rather than as a message of their own.
  `BufferConsumer` gains a default `addAndComplete(ReadBuffer)` that is semantically `add` followed
  by `complete`; `BaseSharedBuffer` forwards it to its subscribers, and
  `PipeliningServerHandler.StreamingOutboundHandler` overrides it to write a
  `DefaultLastHttpContent` carrying those bytes instead of a content message plus an empty
  terminator. Every other `BufferConsumer` inherits the default and behaves exactly as before.
  `BaseSharedBuffer.add` was refactored into `add0(rb, completeAfter)` behind a new
  `addGuarded` hook so that the two subclasses that wrap the add in a concurrency guard
  (`StreamingNettyByteBody.SharedBuffer`'s event-loop check and `ReactiveByteBufferByteBody.SharedBuffer`'s
  `submit`) keep exactly the scope they had.
- Net effect: **N+1** outbound messages instead of 2N+2, and the bytes on the wire are byte for
  byte unchanged (a `DefaultLastHttpContent` carrying `]` encodes to the same `1\r\n]\r\n0\r\n\r\n`
  as a content message plus an empty terminator).

### How the flush-per-item behaviour is preserved

Nothing is ever buffered waiting for a later element:

- The separator that is merged is the one *in front* of the element, which is already known the
  moment the element arrives — the element is written in the same call it would have been before,
  just with one byte of prefix. No element waits for its successor.
- The trailing bytes travel with a completion that has *already happened*, so they add no latency
  either.
- Server-sent events use `Separators.NONE`: `beforeFirst`, `between` and `afterLast` are all
  `null`, so the SSE path takes exactly the same code as before — one buffer, one message, one
  flush per event.

`PipeliningServerHandlerSpec`'s `'streaming responses flush after every item'` feature pins the
contract one level below this change: it feeds raw `ByteBuf`s straight into the response body and
asserts one flush per buffer emitted. That intent is unchanged and the assertion is untouched — the
change is about how many buffers a JSON array element turns into, not about whether a buffer is
flushed when it arrives. The new `StreamingResponseChunkSpec` pins the same guarantee at the level
this change does touch: an SSE event and an asynchronously produced array element each still
produce exactly one message and one flush at the moment they are produced.

## Measurement

`ServerScenariosBenchmark` does not exist on `5.2.x`, so the numbers below come from
`StreamingResponseBenchmark`, added in the second commit of this branch: 200 small records
(`{"id":0,"message":"message-0","value":0}`) streamed through the embedded channel, once as a
single JSON array and once as a JSON stream. Absolute values are therefore not comparable with the
recorded baseline for the equivalent scenarios (309830 ns/op / 335437 B/op and 203672 ns/op /
243590 B/op), but they are the same shape and the comparison below is base against branch in one
session on one machine.

JMH: `-f 1 -wi 5 -w 3s -i 5 -r 3s -t 1 -bm avgt -tu us -prof gc`, three runs each, alternating base
and branch. The 1-minute load average was between 2.47 and 2.97 immediately before every run, and
no java process other than the benchmark itself was running.

Time, us/op (median of three runs; each run's own JMH error in brackets):

| scenario | base | branch | change |
|---|---|---|---|
| JSON array, 200 records | **352.5** (352.0 ±14.0, 352.5 ±5.0, 355.4 ±4.2) | **244.8** (247.9 ±3.5, 229.1 ±3.3, 244.8 ±3.8) | −107.7 us/op, −31% |
| JSON stream, 200 records (control, no separators) | **220.3** (216.5 ±2.4, 220.3 ±13.3, 220.8 ±1.1) | **220.2** (222.2 ±1.2, 220.2 ±1.7, 217.8 ±3.1) | no change |

Allocation, `gc.alloc.rate.norm` in B/op (median of three runs):

| scenario | base | branch | change |
|---|---|---|---|
| JSON array, 200 records | **486660** (486660, 493429, 486637) | **395602** (395702, 395447, 395602) | −91058 B/op, −19% |
| JSON stream, 200 records (control) | **302600** (313503, 298938, 302600) | **299720** (318947, 298925, 299720) | no change |

The array improvement is roughly twelve times the sum of the two error bars, so it is a result. The
JSON stream row is the control: it uses `Separators.NONE`, the code path is untouched, and it does
not move — the spread there is run-to-run noise.

Per element that is about 0.54 us and 455 bytes.

### Outbound message and flush counts

More direct than the timings, measured with an `EmbeddedChannel` and a handler sitting between
`PipeliningServerHandler` and the codec (`StreamingResponseChunkSpec`):

| case | base messages | branch messages | base flushes | branch flushes |
|---|---|---|---|---|
| 3-element JSON array, produced inside the read cycle | 8 (`[`, `1`, `,`, `2`, `,`, `3`, `]`, ``) | 4 (`[1`, `,2`, `,3`, `]`) | 1 | 1 |
| empty JSON array | 2 (`[]`, ``) | 1 (`[]`) | 1 | 1 |
| JSON array, first element produced asynchronously | 2 (`[`, `1`) | 1 (`[1`) | 2 | 1 |
| SSE, per event | 1 | 1 | 1 | 1 |

For N elements that is 2N+2 messages before and N+1 after; at the benchmark's N=200, 402 messages
become 201. Flushes only differ when the element is produced outside a read cycle, because
`PipeliningServerHandler` already defers flushes to `channelReadComplete` while reading — but that
asynchronous case is the normal one for a real reactive endpoint, and there the flushes per element
halve too.

## Testing

New: `http-server-netty/src/test/groovy/io/micronaut/http/server/netty/StreamingResponseChunkSpec.groovy`,
four features — the wire form and message split of a streamed JSON array, an asynchronously
produced array (message and flush per element), an empty array, and an SSE stream where events must
arrive one at a time.

Fail-before evidence, with `http/src/main`, `http-netty` and `http-server-netty/src/main` checked
out at `origin/5.2.x` and the new spec in place:

```
a streamed json array sends one content message per element FAILED
  Condition not satisfied:
  mon.contents == ['[1', ',2', ',3', ']']
  |   [[, 1, ,, 2, ,, 3, ], ]

an asynchronously produced json array element is written and flushed as one message FAILED
  Condition not satisfied:
  mon.contents == ['[1']
  |   [[, 1]

an empty streamed json array sends a single content message FAILED
  Condition not satisfied:
  mon.contents == ['[]']
  |   [[], ]

server sent events are not held back waiting for the next one PASSED
```

The SSE feature passing on both the base and the branch is the point: it is there to pin what must
*not* change.

Ran:

- `./gradlew :micronaut-http:test` — green.
- `./gradlew :micronaut-http-netty:test` — green.
- `./gradlew :micronaut-http-server-netty:test --tests '*Json*' --tests '*Sse*' --tests '*ServerSentEvent*' --tests '*PipeliningServerHandlerSpec*' --tests '*StreamingResponseChunkSpec*'` — green.
- `./gradlew :micronaut-http-server-netty:test` (full) — green, 1138 tests.
- `./gradlew :micronaut-http-client:test` — not a module this change touches, but it consumes
  `BaseSharedBuffer` and `ReactiveByteBufferByteBody`, so it was run for confidence. One failure,
  `ProxyBackpressureTest.backpressureLocal[6] ssl=true version=3 endpoint=/proxy`
  (`AssertionFailedError: expected: <true> but was: <false>` at `ProxyBackpressureTest:124`), a
  timing-sensitive backpressure assertion that ran while the machine was loaded. Re-running
  `--tests '*ProxyBackpressureTest*'` alone passes all six parameterisations.
- `./gradlew :micronaut-http:checkstyleMain :micronaut-http-netty:checkstyleMain :micronaut-http-server-netty:checkstyleMain`
  and the matching `checkstyleTest` tasks, plus `:benchmarks:checkstyleJmh` — no new violations. The
  two reported are pre-existing and in files this branch does not touch
  (`NettyWebSocketSession` DesignForExtension, `NettyHttpServerConfiguration` FileLength).

## Out of scope

- **Bounded look-ahead.** `ConcatenatingSubscriber` still requests the next element only once the
  previous one has been written *and* acknowledged as consumed, so serialisation never overlaps
  with the write. A small look-ahead gated on the channel's write-buffer high-water mark would let
  the two overlap, but it changes the back-pressure contract and belongs in its own change.
- **HTTP/2.** `MultiplexedServerHandler` inherits the default `addAndComplete`, so an h2 response
  still sends the trailing bytes as their own DATA frame followed by an empty end-of-stream frame.
  The separator merge does help there, since it happens above the transport; combining the last
  DATA frame with end-of-stream is a separate change.

## Note on the benchmark in this PR

This PR adds a small `StreamingResponseBenchmark` so the change can be measured on its own. A
broader harness covering the same ground, `ServerScenariosBenchmark`, is proposed separately in
#13112. If that one lands first, the benchmark commit here (`Add a JMH benchmark for streaming a
Flux of records`) can be dropped and the measurement repeated with it; the commit is kept separate
for exactly that reason.

### DIFF
--- a/http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
+++ b/http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
@@ -336,6 +336,36 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
      */
     @Override
     public void add(ReadBuffer rb) {
+        addGuarded(rb, false);
+    }
+
+    /**
+     * Add a given buffer to this {@link BaseSharedBuffer} and complete it, in one operation. This
+     * allows subscribers that implement {@link BufferConsumer#addAndComplete(ReadBuffer)} to
+     * combine the final bytes and the completion signal into a single downstream message.<br>
+     * Not thread safe, caller must handle concurrency.
+     */
+    @Override
+    public void addAndComplete(ReadBuffer rb) {
+        boolean subscribersCompleted = addGuarded(rb, true);
+        complete0(!subscribersCompleted);
+    }
+
+    /**
+     * Hook for subclasses that need to apply a concurrency guard around the {@link #add} portion
+     * of {@link #add(ReadBuffer)} and {@link #addAndComplete(ReadBuffer)}. Subclasses must call
+     * {@code super.addGuarded(rb, completeAfter)} and return its result.
+     *
+     * @param rb           The buffer to add
+     * @param completeAfter Whether the subscribers should be completed together with this buffer
+     * @return {@code true} iff the subscribers have been completed as part of this call
+     */
+    protected boolean addGuarded(ReadBuffer rb, boolean completeAfter) {
+        return add0(rb, completeAfter);
+    }
+
+    private boolean add0(ReadBuffer rb, boolean completeAfter) {
+        boolean subscribersCompleted = false;
         try (rb) {
             assert !working;
 
@@ -349,7 +379,7 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
 
             // drop messages if we're done with all subscribers
             if (complete || error != null) {
-                return;
+                return false;
             }
             if (expectedLength == -1) {
                 Exception totalSizeException = sizeLimitTrackers.totalSize().add(rb.readable());
@@ -357,14 +387,19 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
                     // for maxBodySize, all subscribers get the error
                     error(totalSizeException);
                     rootUpstream.allowDiscard();
-                    return;
+                    return false;
                 }
             } // else, already checked the Content-Length
 
             working = true;
             if (subscribers != null) {
+                subscribersCompleted = completeAfter;
                 for (BufferConsumer consumer : subscribers) {
-                    consumer.add(rb.duplicate());
+                    if (completeAfter) {
+                        consumer.addAndComplete(rb.duplicate());
+                    } else {
+                        consumer.add(rb.duplicate());
+                    }
                 }
             }
             if (reserved > 0 || fullSubscribers != null) {
@@ -391,6 +426,7 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
             }
             working = false;
         }
+        return subscribersCompleted;
     }
 
     /**
@@ -399,12 +435,16 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
      */
     @Override
     public void complete() {
+        complete0(true);
+    }
+
+    private void complete0(boolean notifySubscribers) {
         if (expectedLength > lengthSoFar) {
             throw new IncorrectContentLengthException("Received fewer bytes than specified by Content-Length");
         }
         complete = true;
         expectedLength = lengthSoFar;
-        if (subscribers != null) {
+        if (notifySubscribers && subscribers != null) {
             for (BufferConsumer subscriber : subscribers) {
                 subscriber.complete();
             }
--- a/http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
+++ b/http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
@@ -347,8 +347,21 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
      */
     @Override
     public void addAndComplete(ReadBuffer rb) {
-        boolean subscribersCompleted = addGuarded(rb, true);
-        complete0(!subscribersCompleted);
+        // The final bytes are stored and this buffer is marked complete before any streaming
+        // subscriber learns of the completion. A subscriber may consume a reserved split from
+        // inside its completion callback (the servlet integration does), and that consumer has
+        // to find the final bytes in the buffer and a completed state, or it would miss them and
+        // then wait for a completion that is never delivered.
+        List<ReadBuffer> deferred = addGuarded(rb, true);
+        complete0(deferred == null);
+        if (deferred != null) {
+            // a snapshot: a subscriber added reentrantly by a callback below is completed by
+            // subscribe0 itself, since the buffer is complete by now
+            List<BufferConsumer> targets = new ArrayList<>(subscribers);
+            for (int i = 0; i < targets.size(); i++) {
+                targets.get(i).addAndComplete(deferred.get(i));
+            }
+        }
     }
 
     /**
@@ -358,14 +371,18 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
      *
      * @param rb           The buffer to add
      * @param completeAfter Whether the subscribers should be completed together with this buffer
-     * @return {@code true} iff the subscribers have been completed as part of this call
+     * @return With {@code completeAfter}, the copies of {@code rb} still to be delivered to the
+     * streaming subscribers together with their completion, one per subscriber in order, or
+     * {@code null} if there are none to deliver. Always {@code null} without {@code completeAfter}.
      */
-    protected boolean addGuarded(ReadBuffer rb, boolean completeAfter) {
+    @Nullable
+    protected List<ReadBuffer> addGuarded(ReadBuffer rb, boolean completeAfter) {
         return add0(rb, completeAfter);
     }
 
-    private boolean add0(ReadBuffer rb, boolean completeAfter) {
-        boolean subscribersCompleted = false;
+    @Nullable
+    private List<ReadBuffer> add0(ReadBuffer rb, boolean completeAfter) {
+        List<ReadBuffer> deferred = null;
         try (rb) {
             assert !working;
 
@@ -379,7 +396,7 @@ public abstract class BaseSharedBuffer implements BufferConsumer {
 
             // drop messages if we're done with all subscribers
             if (complete || error != null) {
-                return false;
+                return null;
             }
             if (expectedLength == -1) {
                 Exception totalSizeException = sizeLimitTrackers.totalSize().add(rb.readable());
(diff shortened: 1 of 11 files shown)


graemerocher:

---

The new combined add-and-complete path exposes completion before the shared buffer has finalized its own state, breaking supported reentrant consumption and potentially deadlocking requests.


- [P1] Finalize shared state before notifying subscribers — http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java:398-399
  When a trailing buffer reaches an existing subscriber, `consumer.addAndComplete(...)` emits completion while `working` is still true and before the buffer is saved or `complete` is set. A synchronous `onComplete` callback that consumes a reserved split—an explicitly supported servlet pattern in `ReactiveByteBufferByteBodyTest.reentrancy`—therefore reenters `subscribe0`; assertions fail, or blocking consumption waits for completion that cannot run until the callback returns. Buffer the final data and mark the shared state complete before delivering the downstream terminal callback.

---

**Disposition:** Verified by test: with the combined add-and-complete, a streaming subscriber's completion callback that reads a reserved split (the servlet pattern) tripped the reentrancy guard, and without assertions would have been handed the buffer minus the closing bracket and then blocked forever. Fixed locally in `e701fa89af`: the final piece is stored and the buffer marked complete before any streaming subscriber is told. The test also exposed a call-order dependency in core's `LeakTracker.Factory.staticInitializer` when Netty is absent, guarded in `41f58d8592`.

http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
```diff
@@ -336,6 +336,53 @@ private void discardBuffer() {
...
+        // subscriber learns of the completion. A subscriber may consume a reserved split from
+        // inside its completion callback (the servlet integration does), and that consumer has
+        // to find the final bytes in the buffer and a completed state, or it would miss them and
+        // then wait for a completion that is never delivered.
+        List<ReadBuffer> deferred = addGuarded(rb, true);
+        complete0(deferred == null);
+        if (deferred != null) {
+            // a snapshot: a subscriber added reentrantly by a callback below is completed by
+            // subscribe0 itself, since the buffer is complete by now
+            List<BufferConsumer> targets = new ArrayList<>(subscribers);
```
dstepanov: `targets` is snapshotted after `complete0`, but `deferred` was sized in `add0`. `complete0` completes `fullSubscribers` synchronously, and a `buffer()` callback can subscribe another reserved split right there. `ReactiveByteBufferByteBody.SharedBuffer.submit` uses a `ReentrantLock`, so the nested `subscribe0` runs inline; the netty `SharedBuffer` also subscribes inline on the event loop once `adding` is back to false. `subscribers` then has more entries than `deferred`, and `deferred.get(i)` throws `IndexOutOfBoundsException`. That escapes from `onComplete`, so the streaming subscribers before it never get the `]` or their completion (the response hangs), and the remaining deferred duplicates leak. `add0` can hit the same thing: `fullSubscriber.completeExceptionally(bufferSizeExceeded)` runs after `deferred` is built.

Reproduced on this branch with a `JDK_JSON` concatenation and two `FASTEST` splits: one split is streamed, the primary is `buffer()`ed, and the `buffer()` callback streams the second split. Emitting one element and completing throws `Index 1 out of bounds for length 1` at `BaseSharedBuffer.java:362`. The same test passes on the merge base with only the `LeakTracker` commit applied.

Limiting the snapshot to the subscribers `deferred` was built for fixes it. A subscriber added later already gets the buffered bytes and its completion from `subscribe0`, because the buffer is complete by then (plus `import java.util.Objects;`):

```suggestion
            List<BufferConsumer> targets = new ArrayList<>(Objects.requireNonNull(subscribers).subList(0, deferred.size()));
```

(NullAway rejects a bare `subscribers.subList`.) With this change the reproducer and the `ConcatenatingSubscriber*` tests pass locally.

dstepanov (commented):
Nice reduction: 2N+2 messages down to N+1. I went through the JSON array, NDJSON and SSE paths (how messages are split and when they flush), the empty array, an error mid-stream (that path is unchanged), compression when a `LastHttpContent` carries content, and demand accounting (forwarded and consumed totals are unchanged; the trailing bytes are only counted at completion). I found nothing wrong in any of those, and `StreamingResponseChunkSpec`, `PipeliningServerHandlerSpec` and `JsonStreamSpec` pass locally.

One reentrancy problem, in `BaseSharedBuffer.addAndComplete`: see the inline comment.

http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
graemerocher: Thanks, that's exactly right. Applied your suggestion in 807bb8a0c3 (snapshot limited to `subscribers.subList(0, deferred.size())`, with the `Objects.requireNonNull` for NullAway) and added your reproducer as `ConcatenatingSubscriberReentrancyTest.splitSubscribedFromABufferingCallbackDuringCompletion`: streamed split, buffered primary whose callback streams the second split, one element then completion. It fails with `Index 1 out of bounds for length 1` without the change and all three consumers see `[{}]` with it.

yawkat:
please split your PRs

graemerocher:
split them how? this is one logical fix

graemerocher:
Split as requested:

- #13190 — the `addAndComplete` API, `BaseSharedBuffer` (including both reentrancy fixes from this review) and `ConcatenatingSubscriber` (`http` module)
- #13196 — the Netty handler half that writes the trailing bytes in the terminating message (stacked on #13190)
- #13191 — the `LeakTracker.Factory.staticInitializer` fix
- #13192 — the `StreamingResponseBenchmark` JMH class

Closing this one; the review threads here are addressed in the split PRs.

Issue #13110, opened 2026-09-10
Update all Micronaut Platform 5.2 projects to Micronaut Core 5.2


dstepanov:
Tracking issue for moving every project managed by the Micronaut Platform 5.2 onto **Micronaut Core 5.2.x** (latest: [5.2.0](https://central.sonatype.com/artifact/io.micronaut/micronaut-core-bom/5.2.0), released 2026-09-10; the platform catalog still pins `5.1.14`).

**Progress:** 40/71 projects on core 5.2 · 69/71 tested against 5.2 (54 ✅ pass · 15 ⚠️ issues found · 0 ❌ open core regression · 0 ⏳ running)

For each project:
1. Merge a core 5.2.x bump on its active branch.
2. Test it against core 5.2.x.
3. Fix any regressions in core or in the module.

**Legend:** ⬜ not on core 5.2 yet · ✅ on core 5.2 · Test column: ✅ passes · ⚠️ issues found (fixed or module-side) · ❌ open core regression · ⏳ running · — not tested yet

**Core under test:** `5.2.x` at c6158f9e0e (5.2.1-SNAPSHOT, includes the fixes below) or the equivalent 5.2.0 release code. Core issues found so far:
- [x] micronaut-projects/micronaut-core#13096: iterate introspections in a stable order again (broke micronaut-data H2 tests)
- [x] micronaut-projects/micronaut-core#13099: keep a property's annotations to the type it is read through (broke serde `@JsonIgnoreProperties` inheritance)
- [ ] `/metrics` management endpoint of released micronaut-micrometer 6.0.2 (compiled against core 5.1) returns 404 on core 5.2, which breaks micronaut-sql `DataSourcePoolMetadataSpec`, 6 micronaut-kafka metrics specs and 2 micronaut-eclipsestore metrics specs. Bisected to micronaut-projects/micronaut-core#12960 (d44e268e3b, enumerate beans by their `@Indexed` types); fix in micronaut-projects/micronaut-core#13111
- [ ] micronaut-projects/micronaut-core#13109: only look up introspections in a class loader that shares the core types (breaks micronaut-oracle-cloud `oraclecloud-function-http`; verified 27/27 with the fix)
[...]

PR #13114, opened 2026-09-10
Serve the registered annotation type to a caller naming no class loader

dstepanov:
Fixes the 32 Jakarta Validation TCK failures micronaut-validation hits on 5.2.x, tracked in #13110.

## Cause

`AnnotationMetadata.getAnnotationType(name)` names no class loader, and `AnnotationMetadataSupport` answered it with the loader of the support class itself. Since #12977, a registered annotation type defined by another loader is resolved again through the loader asked with. So a type registered by the generated metadata of a child-first deployment was answered with the application loader's copy of the same name.

The validation TCK runs each test in such a deployment loader, with the TCK classes on the application classpath as well. `DefaultValidator` reads constraint types through `getAnnotationTypesByStereotype(Constraint.class)`, which reaches the one-arg lookup. It got the application's copies, so violations carried a constraint type that is not the class the test compares with. The assertions print identical types, e.g. `expected [interface ...$ValidTestBean] but found [interface ...$ValidTestBean]`. Class-level and cross-parameter constraints are the ones read that way.

A first-parent bisect over the validation TCK names 2c9161c43f (#12977) as the first bad commit: its predecessor d106b03832 (#13019) passes 384/384, and 2c9161c43f fails all 32.

## Fix

A caller naming no loader is answered for the thread context loader, the loader of the deployment it runs in, never for the loader of the support class:

- the registered type is kept when the context loader defines it, is one of its parents (a thread of the container a deployment runs in), or is absent;
- a context loader apart from it resolves the name again, as a loader the caller names does, so another deployment defining its own copy is served that copy rather than the first registration;
- a name nothing registered is loaded through the context loader before the loader of the support class.

The two-arg lookup is unchanged. The validation TCK sets the deployment loader as the context loader around each test, which is the loader that registered its constraint types.

## Verification

- `AnnotationTypeClassLoaderSpec`: two new cases (the context loader being the registering deployment, its parent, absent, or another deployment; and an unregistered name). Both fail on 5.2.x and pass here; the existing cases pass both ways.
- micronaut-validation e65b135f against core published from this branch: `:micronaut-tests:micronaut-jakarta-validation-tck:test` 384 tests, 0 failed (32 failed against c6158f9e0e, exactly the known set).
- `:micronaut-validation:test` 500 tests, 1 failed: `ValidatorSpec` "test @Introspected is required to validate the bean". It is a separate, already known 5.2 change and fails without this PR too.
- `:micronaut-inject:test` 407/0, `:micronaut-reflection:test` 397/0, `:micronaut-inject-java:test --tests 'io.micronaut.inject.annotation.*'` 195/0.

### DIFF
--- a/inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
+++ b/inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
@@ -342,14 +342,48 @@ public final class AnnotationMetadataSupport {
      * @return The annotation
      */
     static Optional<Class<? extends Annotation>> getAnnotationType(String name) {
-        // a caller naming no loader has no copy of its own to be served: the registered type is the one the
-        // generated metadata of its deployment registered, and resolving the name again through the loader of
-        // this class would answer the application's copy where a child-first deployment loader defines another
+        // a caller naming no loader is answered for the thread context loader, the loader of the deployment it
+        // runs in, never for the loader of this class: that one sees the application's copy of a type where a
+        // child-first deployment loader defines another
+        final ClassLoader contextLoader = Thread.currentThread().getContextClassLoader();
         final Class<? extends Annotation> type = ANNOTATION_TYPES.get(name);
         if (type != null) {
-            return Optional.of(type);
+            // the registered type is kept for a context that defines it or sits above the loader defining it,
+            // a thread of the container a deployment runs in; only a context loader apart from it - another
+            // deployment, which may define a copy of its own - resolves the name again
+            if (contextLoader == null || isSelfOrAncestor(contextLoader, type.getClassLoader())) {
+                return Optional.of(type);
+            }
+            return getAnnotationType(name, contextLoader);
         }
-        return getAnnotationType(name, AnnotationMetadataSupport.class.getClassLoader());
+        final ClassLoader ownLoader = AnnotationMetadataSupport.class.getClassLoader();
+        if (contextLoader != null && contextLoader != ownLoader) {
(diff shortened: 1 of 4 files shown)


inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
```diff
@@ -342,6 +342,13 @@ public static String getRepeatableAnnotation(String annotation) {
      * @return The annotation
      */
     static Optional<Class<? extends Annotation>> getAnnotationType(String name) {
+        // a caller naming no loader has no copy of its own to be served: the registered type is the one the
+        // generated metadata of its deployment registered, and resolving the name again through the loader of
+        // this class would answer the application's copy where a child-first deployment loader defines another
+        final Class<? extends Annotation> type = ANNOTATION_TYPES.get(name);
```
graemerocher, line 349: this probably needs to be aware of the thread context loader otherwise it could break that case

graemerocher (changes requested):

inject/src/main/java/io/micronaut/inject/annotation/AnnotationMetadataSupport.java
dstepanov: Made it aware of the thread context loader in 3def59e089: a lookup naming no loader keeps the registered type when the context loader defines it, is one of its parents or is absent, and otherwise resolves the name through the context loader (so another deployment gets its own copy); an unregistered name is loaded through the context loader before the support class's own. Two spec cases cover it, and the validation TCK still passes 384/384.

graemerocher (approved):

PR #13118, opened 2026-09-11
Cache the Date header and access log timestamp per second; clear the access log request attribute

graemerocher:
## Summary

- **`Date` header** (`http-server`): `RouteExecutor.applyConfiguredHeaders` and `AbstractFileBodyWriter` formatted the header from scratch for every response (`LocalDateTime.now()`, `ZonedDateTime.of(...)`, `withZoneSameInstant(GMT)`, `DateTimeFormatter.RFC_1123_DATE_TIME.format(...)`). The text only changes once per second. New `@Internal` `PerSecondCache` (one `volatile` reference to an immutable `(second, text)` record) and `HttpDateHeader` (the RFC 1123 GMT text through that cache) replace the per-response formatting; both call sites now `add(DATE, HttpDateHeader.now())`. The file writer also derives `Expires` from the same `currentTimeMillis()` reading instead of a second `LocalDateTime`. There was no existing per-second cache to reuse: Netty's `DateFormatter` formats on every call, and the file writers used the same `LocalDateTime.now()` path as `RouteExecutor`.
- **Access log timestamp** (`DateTimeElement`): formatted a `ZonedDateTime` per request. Unless the pattern has an unquoted sub-second field (`S`, `n`, `N`, `A`) its output is a function of the instant truncated to the second, so it now goes through a `PerSecondCache`; the default `[dd/MMM/yyyy:HH:mm:ss Z]` pattern and any custom pattern without sub-second fields are cached, a pattern with sub-second precision keeps formatting `ZonedDateTime.now()` exactly as before. Custom time zones (`pattern, Zone/Id`) are unaffected: the formatter's zone override is applied to the cached instant the same way it was applied to `now()`.
- **Access log connection metadata** (`HttpAccessLogHandler`, `Http2AccessLogManager`): `ConnectionMetadata.ofNettyChannel(ctx.channel())` was called per request. The metadata only wraps the channel and reads its addresses on demand, so one instance per connection is equivalent; it is now kept in the per-channel `AccessLogHolder` and in the connection-scoped HTTP/2 manager.
- **Access log request attribute** (`RoutingInBoundHandler`), the correctness part of this PR: with the access logger enabled, the current `NettyHttpRequest` was stored in a channel attribute named `NettyHttpRequest` (Micronaut Session's `SessionLogElement` reads it by that name) and never removed, so an idle keep-alive connection kept its last request reachable, after `release()`, until the channel closed. It also resolved the `AttributeKey` by name on every request. The key is now a static field, and `responseWritten` clears the attribute with `compareAndSet(request, null)`, so that with pipelining a newer request already stored for the same connection is left alone.
- **Benchmark harness** (`benchmarks`): `ControllersBenchmark` and `FullHttpStackBenchmark` disable the `Date` header (so the response bytes are identical each iteration) and never enable access logging, so neither of the paths above could be measured. `-Dmicronaut.bench.date-header=true` keeps the header on and relaxes the response check to a length comparison; `-Dmicronaut.bench.access-log=true` enables the access logger, routed by `logback.xml` to a NOP appender so the line is built but not written. Both default to off, so the existing measurements are unchanged. The harness also gains `micronaut-http-netty-http3` on its classpath; see "Out of scope" for why.

### Thread safety of the cache

`PerSecondCache` is shared by every event loop (the `Date` cache is static; `DateTimeElement.copy()` returns `this`, so one element serves every `AccessLog` on every loop). It holds a single `volatile` reference to an immutable `record Entry(long second, String text)`. A reader loads the reference once; if `entry.second` equals `floorDiv(now, 1000)` it returns `entry.text`, otherwise it formats the text for its own second and publishes a new record. Two loops that miss in the same second both format the same text and publish equivalent records. A loop that is descheduled after formatting second *N* and publishes after another loop has already published *N+1* leaves a stale record, which only makes the next reader miss and republish. In every interleaving a reader returns the text for the second it read from the clock, and because the record is immutable and published through a `volatile` write there is no torn or mixed value. Cost on the hot path: one `System.currentTimeMillis()`, one volatile read, one `long` compare.

The `ConnectionMetadata` fields are written and read on the channel's event loop only (`channelRead` / the HTTP/2 frame listener), like the rest of the holder's state.

## Benchmarks

Base is `origin/5.2.x` at 6d53fc7bb4 (the parent of this branch) plus the harness commit only (063c17a5d3), so both sides run the same harness; branch is 95e8f096d8. Both jars were built in the same session from the same JDK 25 and run with `-f 1 -wi 5 -w 3s -i 5 -r 3s -t 1 -bm avgt -tu ns -prof gc -jvmArgs "-Dmicronaut.python.enabled=false -Djmh.executor=CUSTOM -Djmh.executor.class=io.micronaut.http.server.stack.JmhFastThreadLocalExecutor"` (plus the harness switches per mode). Every run was started only when the 1-minute load average was below 3 and no other java process was above 40% CPU; start loads (base/branch) were 2.04/2.40, 2.60/2.14 and 2.05/2.83 for the three `ControllersBenchmark` modes and 2.94/2.66, 2.74/2.74 and 2.84/2.92 for `FullHttpStackBenchmark`. "Within noise" means the difference is smaller than the sum of the two error bars (99.9% CI).

### Date header off, no access log (harness defaults)

None of the changed code runs in this mode, so this is a no-regression check.

| ControllersBenchmark | TFB_LIKE | 2,610 ± 74 ns/op | 2,587 ± 30 ns/op | -23 ns/op (-0.9%) (within noise) | 6,408 ± 1 | 6,408 ± 1 | -0 (-0.0%) (within noise) |
| ControllersBenchmark | TFB_STRING | 2,394 ± 49 ns/op | 2,425 ± 47 ns/op | +31 ns/op (+1.3%) (within noise) | 6,456 ± 1 | 6,456 ± 1 | +0 (+0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_BEANS1 | 1,932 ± 31 ns/op | 1,850 ± 34 ns/op | -81 ns/op (-4.2%) | 3,856 ± 1 | 3,856 ± 1 | -0 (-0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_BEANS2 | 1,860 ± 20 ns/op | 1,878 ± 70 ns/op | +18 ns/op (+1.0%) (within noise) | 3,888 ± 1 | 3,888 ± 1 | +0 (+0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS1 | 1,909 ± 24 ns/op | 1,899 ± 27 ns/op | -10 ns/op (-0.5%) (within noise) | 3,896 ± 1 | 3,896 ± 1 | -0 (-0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS2 | 1,865 ± 39 ns/op | 1,842 ± 23 ns/op | -23 ns/op (-1.3%) (within noise) | 3,928 ± 1 | 3,928 ± 1 | -0 (-0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_MAP | 1,689 ± 55 ns/op | 1,766 ± 65 ns/op | +77 ns/op (+4.5%) (within noise) | 3,984 ± 1 | 3,984 ± 1 | +0 (+0.0%) (within noise) |
| ControllersBenchmark | MISSING_QUERY_PARAMETER | 1,987 ± 61 ns/op | 2,089 ± 266 ns/op | +102 ns/op (+5.1%) (within noise) | 4,920 ± 1 | 4,856 ± 1 | -64 (-1.3%) |
| FullHttpStackBenchmark | MICRONAUT | 2,763 ± 74 ns/op | 2,718 ± 218 ns/op | -45 ns/op (-1.6%) (within noise) | 6,496 ± 2 | 6,480 ± 2 | -16 (-0.2%) |

The -81 ns on `TFB_LIKE_BEANS1` and the -64/-16 B/op on the last two rows are single-fork variance (no changed code executes with both switches off; the previous perf PR measured same-jar fork-to-fork differences of the same size), not an effect of this PR.

### Date header on (`-Dmicronaut.bench.date-header=true`)

| ControllersBenchmark | TFB_LIKE | 2,585 ± 39 ns/op | 2,493 ± 26 ns/op | -93 ns/op (-3.6%) | 6,816 ± 1 | 6,392 ± 1 | -424 (-6.2%) |
| ControllersBenchmark | TFB_STRING | 2,945 ± 50 ns/op | 2,546 ± 103 ns/op | -399 ns/op (-13.5%) | 6,832 ± 1 | 6,416 ± 1 | -416 (-6.1%) |
| ControllersBenchmark | TFB_LIKE_BEANS1 | 2,187 ± 68 ns/op | 1,732 ± 23 ns/op | -455 ns/op (-20.8%) | 4,264 ± 1 | 3,856 ± 1 | -408 (-9.6%) |
| ControllersBenchmark | TFB_LIKE_BEANS2 | 2,243 ± 87 ns/op | 1,795 ± 64 ns/op | -448 ns/op (-20.0%) | 4,344 ± 1 | 3,888 ± 1 | -456 (-10.5%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS1 | 2,186 ± 121 ns/op | 1,790 ± 39 ns/op | -396 ns/op (-18.1%) | 4,240 ± 1 | 3,856 ± 1 | -384 (-9.1%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS2 | 2,076 ± 68 ns/op | 1,702 ± 16 ns/op | -374 ns/op (-18.0%) | 4,272 ± 1 | 3,888 ± 1 | -384 (-9.0%) |
| ControllersBenchmark | TFB_LIKE_MAP | 1,939 ± 38 ns/op | 1,674 ± 52 ns/op | -265 ns/op (-13.7%) | 4,392 ± 1 | 3,968 ± 1 | -424 (-9.7%) |
| ControllersBenchmark | MISSING_QUERY_PARAMETER | 2,216 ± 82 ns/op | 2,091 ± 39 ns/op | -125 ns/op (-5.6%) | 5,264 ± 1 | 4,880 ± 1 | -384 (-7.3%) |
| FullHttpStackBenchmark | MICRONAUT | 3,508 ± 355 ns/op | 3,024 ± 400 ns/op | -484 ns/op (-13.8%) (within noise) | 6,960 ± 2 | 6,544 ± 2 | -416 (-6.0%) |

With the header on, the branch is back to the "header off" numbers (compare `After` here with `Base` above: `TFB_LIKE_BEANS1` 1,732 vs 1,932, alloc 3,856 vs 3,856). The per-response cost of the header on 5.2.x was roughly 400 B/op and 250-450 ns; on the branch it is one `add` of a shared `String`.

### Date header on, access log on (`-Dmicronaut.bench.access-log=true`, NOP appender)

| ControllersBenchmark | TFB_LIKE | 4,010 ± 187 ns/op | 3,054 ± 185 ns/op | -956 ns/op (-23.8%) | 7,728 ± 2 | 6,928 ± 2 | -800 (-10.4%) |
| ControllersBenchmark | TFB_STRING | 3,838 ± 225 ns/op | 2,690 ± 130 ns/op | -1,148 ns/op (-29.9%) | 7,696 ± 2 | 6,976 ± 1 | -720 (-9.4%) |
| ControllersBenchmark | TFB_LIKE_BEANS1 | 2,955 ± 91 ns/op | 2,153 ± 70 ns/op | -803 ns/op (-27.2%) | 5,200 ± 1 | 4,440 ± 1 | -760 (-14.6%) |
| ControllersBenchmark | TFB_LIKE_BEANS2 | 3,100 ± 118 ns/op | 2,110 ± 120 ns/op | -989 ns/op (-31.9%) | 5,256 ± 2 | 4,432 ± 1 | -824 (-15.7%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS1 | 2,917 ± 70 ns/op | 2,202 ± 136 ns/op | -715 ns/op (-24.5%) | 5,168 ± 1 | 4,432 ± 1 | -736 (-14.2%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS2 | 3,094 ± 97 ns/op | 2,183 ± 119 ns/op | -911 ns/op (-29.4%) | 5,200 ± 2 | 4,464 ± 1 | -736 (-14.2%) |
| ControllersBenchmark | TFB_LIKE_MAP | 2,682 ± 128 ns/op | 2,225 ± 176 ns/op | -457 ns/op (-17.0%) | 5,328 ± 1 | 4,544 ± 1 | -784 (-14.7%) |
| ControllersBenchmark | MISSING_QUERY_PARAMETER | 3,328 ± 171 ns/op | 2,440 ± 84 ns/op | -889 ns/op (-26.7%) | 6,144 ± 2 | 5,512 ± 1 | -632 (-10.3%) |
| FullHttpStackBenchmark | MICRONAUT | 4,168 ± 398 ns/op | 3,551 ± 428 ns/op | -617 ns/op (-14.8%) (within noise) | 7,824 ± 3 | 7,080 ± 2 | -744 (-9.5%) |

Subtracting the "header on" rows, access logging itself cost about 900 B/op and 1.4 µs per request on 5.2.x (`TFB_LIKE`: 7,728 - 6,816 B/op, 4,010 - 2,585 ns) and costs about 540 B/op and 0.56 µs on the branch (6,928 - 6,392 B/op, 3,054 - 2,493 ns). What remains is the log line itself (`AccessLog` string building, the logback event) plus the remaining per-request elements, which this PR does not touch.

**Reading the numbers.** The time and allocation deltas in the two switched-on modes are outside the error bars on every `ControllersBenchmark` row; `FullHttpStackBenchmark` time is within its (wide) error bars in all modes while its allocation delta matches the `ControllersBenchmark` rows. With both switches off, nothing changes. The default Micronaut configuration has the `Date` header on, so the "Date header on" table is the one that applies to a stock application.

## Testing

- New `PerSecondCacheTest` (http-server, 4 tests): same second is formatted once and returns the same instance, value changes at the 999 to 1000 ms boundary, negative instants round toward minus infinity, `now()` is within the current second.
- New `HttpDateHeaderTest` (http-server, 11 tests): the cached value equals the pre-PR `ZonedDateTime.of(LocalDateTime, systemDefault).withZoneSameInstant(GMT).format(RFC_1123_DATE_TIME)` text for seven instants, fixed expected strings (`Tue, 14 Nov 2023 22:13:20 GMT` for 1,700,000,000,000 and ,999 ms; `:21` for 1,700,000,001,000), same instance within a second, change at the boundary, `now()` parses to the current second.
- New `DateHeaderSpec` (http-server-netty): two requests on one embedded connection, the second after sleeping into the next wall-clock second; both `Date` values are RFC 1123 GMT, each within the second it was produced, and the second is strictly later than the first. Also that `micronaut.server.date-header=false` still omits the header.
- New `DateTimeElementSpec` (http-server-netty, 17 tests): default pattern is cached (same instance within a second, new value across the boundary) and equal to the uncached format; `dd/MMM/yyyy:HH:mm:ss Z, UTC` gives `14/Nov/2023:22:13:20 +0000` / `:21 +0000`; `end:yyyy-MM-dd HH:mm:ss, Asia/Tokyo` gives `2023-11-15 07:13:20` and selects the last-response-write event; `HH:mm:ss.SSS, UTC` is not cached and gives `.000`, `.001`, `.999`, `.000` for consecutive milliseconds; live values from `onRequestHeaders` are current; 12 cases for the sub-second pattern scan including quoted `'S'` and escaped quotes.
- New `AccessLogRequestAttributeSpec` (http-server-netty): with the access logger enabled, a controller sees its own `NettyHttpRequest` in the channel attribute during routing (`same`), and the attribute is `null` after the response for two consecutive requests on the same connection; with the logger disabled the attribute is never set. Fail-before verified by restoring the 5.2.x `RoutingInBoundHandler` and running the spec: `Condition not satisfied: server.attr(KEY).get() == null` with the attribute still holding `GET /access-log-attribute` at `AccessLogRequestAttributeSpec.groovy:48`. The `Date`/timestamp caches preserve behaviour by design, so their tests guard equivalence rather than failing before.
- `./gradlew :micronaut-http-server:test :micronaut-http-server:checkstyleTest`: 143 tests, 0 failures.
- `./gradlew :micronaut-http-netty:test :micronaut-http-netty:checkstyleTest`: 148 tests, 3 skipped (pre-existing), 0 failures.
- `./gradlew :micronaut-http-server-netty:test`: 1177 tests, 22 skipped (pre-existing), 0 failures on the first run; `checkstyleTest` is not configured for this module (task SKIPPED).
- `checkstyleMain` clean for http-server and http-server-netty apart from the pre-existing `NettyHttpServerConfiguration` FileLength violation; `japiCmp` clean for both modules.

## Out of scope

- Enabling the access logger without `micronaut-http-netty-http3` on the classpath fails every request on 5.2.x with `NoClassDefFoundError: io/netty/handler/codec/quic/QuicChannel` from `ConnectionMetadataImpl.<clinit>`: the initializer guards `QuicChannel.class` with `catch (Exception)`, which does not catch a `NoClassDefFoundError`. That is what the first access-log benchmark run hit (the benchmarks module only had `http-server-netty`, which declares the HTTP/3 module `compileOnly`). This PR adds the module to the benchmark classpath so the switch works; the initializer itself is a separate fix.
- `RoutingInBoundHandler.prepareRequest` still calls `ctx.pipeline().get(HANDLER_ACCESS_LOGGER)` per request. The answer differs per pipeline (HTTP/1 pipelines carry the handler, HTTP/2 connection pipelines use `Http2AccessLogManager` instead, and a configuration reload can remove it), so replacing the scan needs a per-pipeline flag threaded through `HttpPipelineBuilder`; it is a ~10-node name comparison and left alone.
- With the default (non-legacy) HTTP/2 connection handler the `NettyHttpRequest` attribute is never set, because that connection pipeline carries `Http2AccessLogManager` rather than a `HANDLER_ACCESS_LOGGER` handler, so `SessionLogElement` reports `-` there. Pre-existing and unrelated to the caches.
- Access log elements other than the timestamp still allocate per request (`Optional`s from `ConnectionMetadata.remoteAddress()`, the log line string), as does the logback event; visible as the residual ~540 B/op in the third table.
- `ServerScenariosBenchmark` was not run: its scenarios exercise bodies and multipart, not the header/access-log paths, and the harness switches only cover the two benchmarks above.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/DateTimeElement.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/DateTimeElement.java
@@ -15,10 +15,12 @@
  */
 package io.micronaut.http.server.netty.handler.accesslog.element;
 
+import io.micronaut.http.server.util.PerSecondCache;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.handler.codec.http.HttpHeaders;
 import org.jspecify.annotations.Nullable;
 
+import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
@@ -29,6 +31,10 @@ import java.util.Set;
 
 /**
  * DateTimeElement LogElement.
+ * <p>
+ * Unless the pattern prints a sub-second field, the formatted text only changes once per second
+ * and is cached in a {@link PerSecondCache}. The element is shared by every access log instance
+ * ({@link #copy()} returns {@code this}), so the cache is shared by all event loops.
  *
  * @author croudet
  * @since 2.0
@@ -48,6 +54,12 @@ final class DateTimeElement implements LogElement {
     private final Set<Event> events;
     @Nullable
     private final String dateFormat;
+    /**
+     * {@code null} when the pattern has a sub-second field, in which case every value is
+     * formatted from the exact current time.
+     */
+    @Nullable
+    private final PerSecondCache cache;
 
     /**
      * Create a DateTimeElement.
@@ -81,7 +93,53 @@ final class DateTimeElement implements LogElement {
         } else {
             formatter = DateTimeFormatter.ofPattern(formatSplit[0], Locale.US).withZone(ZoneId.of(formatSplit[1].strip()));
         }
+        cache = hasSubSecondField(formatSplit[0]) ? null : new PerSecondCache(this::formatSecond);
         events = fromStart ? Event.REQUEST_HEADERS_EVENTS : LAST_RESPONSE_EVENTS;
+    }
+
+    /**
+     * Whether the pattern contains an unquoted fraction-of-second ({@code S}), nano-of-second
+     * ({@code n}), nano-of-day ({@code N}) or milli-of-day ({@code A}) field. Every other pattern
+     * letter is a function of the instant truncated to the second.
+     *
+     * @param pattern The {@link DateTimeFormatter} pattern
+     * @return {@code true} if the output can change within a second
+     */
+    static boolean hasSubSecondField(String pattern) {
+        boolean quoted = false;
+        for (int i = 0; i < pattern.length(); i++) {
+            char c = pattern.charAt(i);
+            if (c == '\'') {
+                quoted = !quoted;
+            } else if (!quoted && (c == 'S' || c == 'n' || c == 'N' || c == 'A')) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private String formatSecond(long epochSecond) {
+        return ZonedDateTime.ofInstant(Instant.ofEpochSecond(epochSecond), ZoneId.systemDefault()).format(formatter);
+    }
+
+    private String now() {
+        if (cache != null) {
+            return cache.now();
+        }
+        return ZonedDateTime.now().format(formatter);
+    }
+
+    /**
+     * The value this element produces for the given instant.
+     *
+     * @param epochMillis The instant, in milliseconds since the epoch
+     * @return The formatted text
+     */
+    String value(long epochMillis) {
+        if (cache != null) {
+            return cache.get(epochMillis);
+        }
+        return ZonedDateTime.ofInstant(Instant.ofEpochMilli(epochMillis), ZoneId.systemDefault()).format(formatter);
     }
 
     @Override
@@ -92,7 +150,7 @@ final class DateTimeElement implements LogElement {
     @Override
     public String onRequestHeaders(@Nullable SocketChannel channel, String method, HttpHeaders headers, String uri, String protocol) {
         if (events.contains(Event.ON_REQUEST_HEADERS)) {
-            return ZonedDateTime.now().format(formatter);
+            return now();
         } else {
             return ConstantElement.UNKNOWN_VALUE;
         }
@@ -101,7 +159,7 @@ final class DateTimeElement implements LogElement {
     @Override
     public String onLastResponseWrite(int contentSize) {
         if (events.contains(Event.ON_LAST_RESPONSE_WRITE)) {
-            return ZonedDateTime.now().format(formatter);
+            return now();
         } else {
             return ConstantElement.UNKNOWN_VALUE;
         }
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -25,6 +25,8 @@ dependencies {
     api(projects.micronautContextPython)
     api(projects.micronautHttpServer)
     api(projects.micronautHttpServerNetty)
+    // the access logger's ConnectionMetadata resolves the QUIC channel class when it is initialized
+    api(projects.micronautHttpNettyHttp3)
     api(projects.micronautHttpClient)
     api(projects.micronautJacksonDatabind)
     api(projects.micronautRouter)
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -55,6 +55,15 @@ val jmhPoolSizes = providers.gradleProperty("jmh.poolSizes")
     .map { it.split(",").map(String::trim).filter(String::isNotEmpty) }
 val jmhHumanOutput = providers.gradleProperty("jmh.humanOutput")
     .map(layout.projectDirectory::file)
+// Benchmark switches read by BenchOptions in the forked JMH JVM. A -D on the Gradle command line
+// only reaches Gradle itself, so they are exposed as -P properties and forwarded as JVM arguments:
+// -Pjmh.dateHeader=true, -Pjmh.accessLog=true
+val jmhBenchSwitches = listOf(
+    "jmh.dateHeader" to "micronaut.bench.date-header",
+    "jmh.accessLog" to "micronaut.bench.access-log"
+).mapNotNull { (property, systemProperty) ->
+    providers.gradleProperty(property).orNull?.let { "-D$systemProperty=$it" }
+}
 
 jmh {
     includes = jmhIncludes
@@ -68,6 +77,7 @@ jmh {
         benchmarkParameters.put("poolSize", objects.listProperty(String::class.java).value(sizes))
     }
     humanOutputFile.set(jmhHumanOutput)
+    jvmArgsAppend.addAll(jmhBenchSwitches)
     duplicateClassesStrategy = DuplicatesStrategy.WARN
 }
 
--- a/benchmarks/src/jmh/java/io/micronaut/http/server/stack/BenchOptions.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/server/stack/BenchOptions.java
@@ -1,0 +1,45 @@
+package io.micronaut.http.server.stack;
+
+import io.netty.buffer.ByteBuf;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * System-property switches shared by the HTTP stack benchmarks.
+ */
+final class BenchOptions {
+    /**
+     * {@code -Dmicronaut.bench.date-header=true} keeps the {@code Date} response header on. The
+     * response then changes once per second, so only its length is verified.
+     */
+    static final boolean DATE_HEADER = Boolean.getBoolean("micronaut.bench.date-header");
+    /**
+     * {@code -Dmicronaut.bench.access-log=true} enables the access logger. {@code logback.xml}
+     * routes the {@code bench-access-log} logger to a NOP appender, so the log line is built but
+     * never written.
+     */
+    static final boolean ACCESS_LOG = Boolean.getBoolean("micronaut.bench.access-log");
+
+    private BenchOptions() {
+    }
+
+    static Map<String, Object> serverProperties(String specName) {
+        Map<String, Object> properties = new HashMap<>();
+        properties.put("spec.name", specName);
+        // with the date header off the response is identical each time
+        properties.put("micronaut.server.date-header", DATE_HEADER);
+        if (ACCESS_LOG) {
+            properties.put("micronaut.server.netty.access-logger.enabled", true);
+            properties.put("micronaut.server.netty.access-logger.logger-name", "bench-access-log");
+        }
+        return properties;
+    }
+
+    static void verifyResponse(ByteBuf expected, ByteBuf actual) {
+        boolean mismatch = DATE_HEADER ? expected.readableBytes() != actual.readableBytes() : !expected.equals(actual);
+        if (mismatch) {
+            throw new AssertionError("Response did not match");
+        }
+    }
+}
--- a/benchmarks/src/jmh/java/io/micronaut/http/server/stack/BenchOptions.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/server/stack/BenchOptions.java
@@ -9,6 +9,9 @@ import java.util.Map;
  * System-property switches shared by the HTTP stack benchmarks.
  */
 final class BenchOptions {
+    // When running through Gradle rather than the jar, pass -Pjmh.dateHeader=true and
+    // -Pjmh.accessLog=true: a -D on the Gradle command line does not reach the forked JMH JVM,
+    // and the build forwards those properties as the system properties below.
     /**
      * {@code -Dmicronaut.bench.date-header=true} keeps the {@code Date} response header on. The
      * response then changes once per second, so only its length is verified.
--- a/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ControllersBenchmark.java
+++ b/benchmarks/src/jmh/java/io/micronaut/http/server/stack/ControllersBenchmark.java
@@ -68,9 +68,7 @@ public class ControllersBenchmark {
     @Benchmark
     public void test(Holder holder) {
         ByteBuf response = holder.exchange();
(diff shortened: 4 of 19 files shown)


graemerocher:

---

The patch introduces duplicate Date headers on common 304 file responses, and its new benchmark switches are ineffective through the repository's Gradle JMH execution path.


- [P2] Replace the copied Date header on 304 responses — http-server/src/main/java/io/micronaut/http/server/body/AbstractFileBodyWriter.java:120-120
  For conditional file requests returning 304, `originalResponse` normally already contains the Date header added by `RouteExecutor`, and `copyNonEntityHeaders` copies it into the new response. Calling `add` here appends another Date value, whereas the previous Netty `headers.date(...)` implementation replaced it, so default-configured 304 file responses now contain duplicate Date headers. Replace or set the existing value instead of appending.

- [P2] Forward benchmark switches into the JMH JVM — benchmarks/src/jmh/java/io/micronaut/http/server/stack/BenchOptions.java:16-22
  When these switches are supplied to the normal Gradle entry point, such as `./gradlew :benchmarks:jmh -Dmicronaut.bench.access-log=true`, they are visible only to Gradle. The JMH plugin launches `org.openjdk.jmh.Main` in a separate Java process without forwarding arbitrary system properties, so both constants remain false and the requested overhead is not measured. Wire the values through JMH's JVM arguments or expose them as the existing `-Pjmh.*` options.

---

**Disposition:** Both verified. Fixed locally: `6234b4a751` replaces instead of appends the Date header on a 304 (the copied headers already carried one; spec asserts exactly one), and `1f1b94aab9` forwards the switches to the forked JMH JVM as `-Pjmh.dateHeader` / `-Pjmh.accessLog`, verified in the fork's VM options line. The PR's own numbers were measured via the jar with `-jvmArgs`, so they stand.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/DateTimeElement.java
```diff
@@ -81,9 +93,55 @@ final class DateTimeElement implements LogElement {
         } else {
             formatter = DateTimeFormatter.ofPattern(formatSplit[0], Locale.US).withZone(ZoneId.of(formatSplit[1].strip()));
         }
+        cache = hasSubSecondField(formatSplit[0]) ? null : new PerSecondCache(this::formatSecond);
```
dstepanov: Nit: `",".split(",")` returns an empty array, so a comma-only format such as `%{,}t` now throws `ArrayIndexOutOfBoundsException` here, where it used to log a literal `,`. Degenerate input, but checking `format` when `formatSplit.length < 2` (the same string the formatter uses) avoids it.

dstepanov (approved):
LGTM. The per-second cache is race-free (immutable entry, a late stale write only causes a reformat), the Date/Expires output matches the old set semantics, and clearing the attribute can't drop a session value: it is read on response headers, which pass the access logger before `responseWritten`, and HTTP/2 and HTTP/3 keep it per stream channel.

The red CI looks unrelated: `StreamPressureSpec consumer pressure` also fails on 5.2.x, and the native TCK job hit `Connection reset by peer` before running tests. Worth a re-run before merging.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/DateTimeElement.java
graemerocher: Fixed in c87daff6bf: the sub-second check now uses the same string the formatter is built from (`format` when the split has fewer than two parts), and `DateTimeElementSpec` has a `%{,}t` case that expects a literal comma.

PR #13120, opened 2026-09-11
Look up optional netty types by name in the access log connection metadata

graemerocher:
## Summary

`ConnectionMetadataImpl` guards its references to `QuicChannel`, `DomainSocketAddress` and `DomainSocketChannel`, which come from optional netty modules, with a class literal inside a `catch (Exception)`. A class literal is resolved by the JVM, and a missing class surfaces as `NoClassDefFoundError`. That is an `Error`, so the catch never runs and the static initializer fails.

The QUIC codec is not on `http-server-netty`'s runtime classpath; it arrives with the HTTP/3 module. So with access logging enabled and no HTTP/3 module, any request whose channel is not a `SocketChannel` fails with

```
java.lang.NoClassDefFoundError: io/netty/handler/codec/quic/QuicChannel
    at ...accesslog.element.ConnectionMetadataImpl.<clinit>
```

Reproduced by loading the class with the QUIC codec removed from the runtime classpath.

**Who is affected:** access logging on a unix domain socket listener (or any other non-socket channel) without `micronaut-http-netty-http3`. TCP listeners over HTTP/1 and HTTP/2 are not affected, because `ConnectionMetadata.ofNettyChannel` returns early for a `SocketChannel` without touching the class. That is why this has not been reported. The test classpath always carries the HTTP/3 module, which is why no test caught it.

## Change

- Look the three types up with `Class.forName(name, false, loader)`, catching `ClassNotFoundException | LinkageError`, and keep `null` when absent.
- `ConnectionMetadata.getHostName` null-checks `DOMAIN_SOCKET_ADDRESS`; it was the one consumer that assumed the field non-null. (`netty-transport-native-unix-common`, which provides the domain socket types, is always on the runtime classpath, so in practice only the QUIC guard matters. All three now behave as the original code intended.)
- The fields are `@Nullable`, which NullAway requires for a `@NullMarked` package.

## Testing

`ConnectionMetadataOptionalClassesTest` loads the access log element package through a child-first class loader that refuses the QUIC package, then calls `ofNettyChannel` with an `EmbeddedChannel` (not a `SocketChannel`, so it reaches the optional-type checks) and `getHostName` with an unknown address type. On the unmodified code the test fails with the `NoClassDefFoundError` above; with the change it passes. It also asserts the optional types are present on the normal test classpath, so the happy path stays covered.

`:micronaut-http-server-netty:test`: 1159 tests, 0 failures. `checkstyleMain` clean.

Noticed while benchmarking with access logging enabled for #13118, where the JMH harness uses an `EmbeddedChannel`.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/ConnectionMetadata.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/ConnectionMetadata.java
@@ -76,7 +76,7 @@ public interface ConnectionMetadata {
             return Optional.of(addr.getPath().toString())
                 // remote address is empty string
                 .filter(s -> !s.isEmpty());
-        } else if (ConnectionMetadataImpl.DOMAIN_SOCKET_ADDRESS.isInstance(a)) {
+        } else if (ConnectionMetadataImpl.DOMAIN_SOCKET_ADDRESS != null && ConnectionMetadataImpl.DOMAIN_SOCKET_ADDRESS.isInstance(a)) {
             String path = ConnectionMetadataImpl.DomainSocketUtil.getPath(a);
             if (path.isEmpty() || path.equals("\0")) {
                 return Optional.empty();
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/ConnectionMetadataImpl.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/accesslog/element/ConnectionMetadataImpl.java
@@ -16,10 +16,10 @@
 package io.micronaut.http.server.netty.handler.accesslog.element;
 
 import io.micronaut.core.annotation.Internal;
+import org.jspecify.annotations.Nullable;
 import io.netty.channel.Channel;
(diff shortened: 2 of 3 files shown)


graemerocher:

---

The change safely handles absent optional Netty classes, guards nullable class references, and adds focused regression coverage. No correctness regressions were identified in the changed code.

---

**Disposition:** No findings.

dstepanov (approved):
LGTM. Verified locally:

- On the real `http-server-netty` runtime classpath (no QUIC codec), base code fails with `NoClassDefFoundError: io/netty/handler/codec/quic/QuicChannel` in `ConnectionMetadataImpl.<clinit>`; with this change `ofNettyChannel(EmbeddedChannel)` returns `Empty` and `getHostName(DomainSocketAddress)` resolves the path.
- Native image (GraalVM 25.0.3): `Class.forName(constant, false, X.class.getClassLoader())` is folded at build time with no reflection metadata. It returns the class with the QUIC jar present and `null` without it.
- Uses the same defining loader the class literal did, and the lookup runs once into `static final` fields, so nothing is repeated per request.
- Access log tests pass (`AccessLogSpec`, `AccessLogFormatParserSpec`, `ConnectionMetadataOptionalClassesTest`).

PR #13124, opened 2026-09-11
Answer 400 when a Publisher<CompletedFileUpload> part has no file name

graemerocher:
## Summary

`NettyPublisherPartUploadBinder` rejects a multipart part that carries the bound name of a `Publisher<CompletedFileUpload>` argument but has no `filename` parameter, since it cannot be a file upload. It did so with a bare `IllegalStateException`, which `RouteExecutor` logs as an unexpected error and answers with **500**. The same condition on a single `@Part CompletedFileUpload` argument is already answered by `FormFactory.completeFileUpload` with `HttpStatusException(BAD_REQUEST)`, which is the right answer: the request is malformed for that route, the server did not fail.

The binder now throws the same `HttpStatusException` with the same message, so the response is **400** and nothing is logged at error level. The field was already closed before the error, so nothing leaks; only the status and the log noise change.

## Testing

`PublisherPartKindMismatchTest` posts two 64 KiB parts named for the upload argument, once without and once with a `filename`. The attribute case expects 400 with the existing message (500 on 5.2.x); the upload case expects 200 with the summed size, unchanged. The raw-socket client reads on its own thread because the server answers the malformed request while the upload is still in flight.

`:micronaut-http-server-netty:test --tests '*Part*' --tests '*Upload*' --tests '*Multipart*' --tests '*Form*'`: 115 tests, 0 failures. `checkstyleMain` clean.

Codex review (gpt-5.6-sol, high): no findings.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyPublisherPartUploadBinder.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyPublisherPartUploadBinder.java
@@ -23,7 +23,9 @@ import io.micronaut.core.convert.ArgumentConversionContext;
 import io.micronaut.core.convert.ConversionService;
 import io.micronaut.core.io.buffer.ReadBuffer;
 import io.micronaut.core.type.Argument;
+import io.micronaut.http.HttpStatus;
 import io.micronaut.http.bind.binders.TypedRequestArgumentBinder;
+import io.micronaut.http.exceptions.HttpStatusException;
 import io.micronaut.http.multipart.CompletedAttribute;
 import io.micronaut.http.multipart.CompletedFileUpload;
 import io.micronaut.http.multipart.CompletedPart;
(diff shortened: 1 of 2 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyPublisherPartUploadBinder.java
```diff
@@ -105,7 +107,9 @@ public BindingResult<Publisher<?>> bindForNettyRequest(ArgumentConversionContext
                 .flatMap(f -> {
                     if (f.metadata().fileName() == null) {
                         f.close();
-                        return Flux.error(new IllegalStateException("Field was not a file upload (no filename parameter)"));
+                        // the request is malformed for this argument, not the server: same answer as
+                        // FormFactory.completeFileUpload gives for a single @Part CompletedFileUpload
+                        return Flux.error(new HttpStatusException(HttpStatus.BAD_REQUEST, "Field [" + f.metadata().name() + "] was expected to be a file upload, but is missing a file name"));
                     }
                     return ReactiveExecutionFlow.toPublisher(formFactory.get().completePart(request, f));
                 })));
```
dstepanov: Non-blocking: `completeFileUpload` already does this exact check, `close()` and `HttpStatusException`, and `completePart` only reaches it once the null case is gone. Calling it directly keeps the message in one place (the Groovy/Kotlin `UploadSpec`s assert it verbatim) and mirrors the `CompletedAttribute` branch:

```suggestion
                .flatMap(f -> ReactiveExecutionFlow.toPublisher(formFactory.get().completeFileUpload(request, f)))));
```

dstepanov (approved):
LGTM. Verified locally: the new test fails on the base binder (500 plus an ERROR log from RouteExecutor) and passes here. An empty `filename=""` still binds, and single `CompletedFileUpload` still answers the same 400. The multipart specs pass with no leak reports.

http-server-netty/src/main/java/io/micronaut/http/server/netty/binders/NettyPublisherPartUploadBinder.java
graemerocher: Took the suggestion in d19851daa7: the branch calls `completeFileUpload` directly and the two now-unused imports are gone. `PartKindMismatchSpec` and the upload specs still pass.

PR #13125, opened 2026-09-11
Add a reflection-free enum-keyed map to CollectionUtils

dstepanov:
Adds `CollectionUtils.newEnumMap(K[] constants)`, which returns an enum-keyed map that behaves like `java.util.EnumMap` without looking the enum's constants up by reflection. It also switches the three `new EnumMap<>(X.class)` calls in the repository to the new map.

## The reflection in `EnumMap`

Every `java.util.EnumMap` gets its key universe by reflection. In JDK 25, `EnumMap(Class)` calls `getKeyUniverse(keyType)`, which calls `SharedSecrets.getJavaLangAccess().getEnumConstantsShared(keyType)`, which calls `Class.getEnumConstantsShared()`:

```java
final Method values = getMethod("values");
values.setAccessible(true);
T[] temporaryConstants = (T[]) values.invoke(null);
enumConstants = constants = temporaryConstants;
```

That is:
- a `getMethod` lookup, which fills in the enum class's public-method reflection data
- `setAccessible`
- a reflective `invoke`
- a cache kept on the `Class` for as long as the class lives

`EnumMap(Map)` does the same when its argument is not already an `EnumMap`, and so does deserialization. Only `EnumMap(EnumMap)` skips it, by copying `keyUniverse` from a map that was itself built by reflection. `EnumMap` has no constructor that takes the constants. The same applies to `EnumSet.noneOf/allOf/of/range/copyOf` and to `Class.getEnumConstants`.

The `NoReflection` check in micronaut-projects/errorprone-no-reflection reports all of these under its `ENUM_CONSTANTS` category. To pass it, micronaut-projects/micronaut-jakarta-interceptors had to replace its `EnumMap`s with `LinkedHashMap`s filled in `values()` order.

## Why the new map avoids it

`EnumMap` only needs reflection because it is given a `Class`. At the call site the enum is known, so `MyEnum.values()` can hand over the constants directly. `values()` is a static method generated by the compiler, and the check does not report it. The rest of `EnumMap` is just an array of values indexed by `ordinal()`.

`EnumConstantsMap` is package-private in `io.micronaut.core.util`:

- **Storage:** it extends `AbstractMap` and holds the `values()` array, an `Object[]` of values and a size. An empty slot is `null`, so a `null` value is stored as a private sentinel.
- **Lookup:** a key is valid when `key instanceof Enum<?> e && e.ordinal() < universe.length && universe[e.ordinal()] == e`. The identity check rejects a constant of another enum without calling `getDeclaringClass()`.
- **Same behaviour as `EnumMap`:**
  - it accepts `null` values and rejects `null` keys with `NullPointerException`
  - `put` with a key from another enum throws `ClassCastException`
  - iteration follows ordinal order, and iterators are weakly consistent (no `ConcurrentModificationException`)
  - `Iterator.remove` works on all three views
  - `Entry.setValue` writes through to the map
- **Overrides:**
  - `get`, `getOrDefault`, `containsKey`, `containsValue`, `put`, `remove`, `size`, `clear` and `forEach`
  - `entrySet`, `keySet` and `values`
  - `putAll`, which copies the value array directly when the source is a map of the same enum
  - `equals` and `hashCode`: they compare arrays against a map of the same enum, and otherwise follow the `Map` contract, so the map equals an `EnumMap` or a `HashMap` with the same mappings
- **No `ENUM_CONSTANTS` calls.** The class uses no `EnumMap`, `EnumSet`, `Class.getEnumConstants`, `Enum.valueOf` or `describeConstable`.

The factory returns `Map<K, V>`. It checks that the array is non-null and that each element is a constant of one enum whose `ordinal()` equals its index, which is the shape `values()` returns. A mis-ordered array would otherwise corrupt lookups without any error. The map keeps the array rather than copying it, and the javadoc says it must not be mutated. Passing `values()` directly is always safe because each call returns a new array.

The javadoc of the existing `CollectionUtils.enumSet(E...)` now notes that it uses `EnumSet.noneOf`, which is reflective. Its behaviour is unchanged.

`@since 5.3.0`: this is new API, and `gradle.properties` on `5.3.x` says `5.3.0-SNAPSHOT`.

## Converted uses

`core`, `inject`, `aop` and `context` contain no `new EnumMap`. These are the only three in the repository's main sources, and all are now converted:

- `router` `DefaultRouter`: the two per-method route maps built in the constructor. The per-request lookup `methodRoutesByMethod.getOrDefault(httpMethod, EMPTY)` now does one slot read; the `Map` default method did `get` plus `containsKey`.
- `http-server-netty` `Http2ServerHandler`: the static `Http2Error` → exception map, built from the `values()` array its static initializer already iterates.
- `jackson-databind` `JacksonFeatures`: the serialization and deserialization feature maps created by each instance.

## Tests

`EnumConstantsMapSpec` covers:
- put, get, remove, `containsKey`, `containsValue`, size and clear
- `null` values, and rejection of `null` keys
- keys from another enum, both with the same ordinal and with an ordinal past the end: `get` returns null and `put` throws
- ordinal iteration order regardless of insertion order
- `Iterator.remove` on each view, and `Entry.setValue` writing through
- `equals`/`hashCode` against `EnumMap`, `HashMap` and a second map
- both `putAll` paths
- an enum whose constants have their own bodies
- the factory rejecting mis-ordered, sparse, `null`-containing and mixed-enum arrays
- a comparison against `java.util.EnumMap` over 10,000 random operations, for three seeds

### DIFF
--- a/core/src/main/java/io/micronaut/core/util/EnumConstantsMap.java
+++ b/core/src/main/java/io/micronaut/core/util/EnumConstantsMap.java
@@ -63,9 +63,9 @@ final class EnumConstantsMap<K extends Enum<K>, V extends @Nullable Object> exte
     private final @Nullable Object[] vals;
     private int size;
 
-    private @Nullable Set<Entry<K, V>> entrySet;
-    private @Nullable Set<K> keySet;
-    private @Nullable Collection<V> values;
+    private @Nullable Set<Entry<K, V>> entrySetView;
+    private @Nullable Set<K> keySetView;
+    private @Nullable Collection<V> valuesView;
 
     /**
      * @param universe The constants of the enum in ordinal order, already validated
@@ -182,30 +182,30 @@ final class EnumConstantsMap<K extends Enum<K>, V extends @Nullable Object> exte
 
     @Override
     public Set<Entry<K, V>> entrySet() {
-        Set<Entry<K, V>> es = entrySet;
+        Set<Entry<K, V>> es = entrySetView;
         if (es == null) {
             es = new EntrySet();
-            entrySet = es;
+            entrySetView = es;
         }
         return es;
     }
 
     @Override
     public Set<K> keySet() {
-        Set<K> ks = keySet;
+        Set<K> ks = keySetView;
         if (ks == null) {
             ks = new KeySet();
-            keySet = ks;
+            keySetView = ks;
         }
         return ks;
     }
 
     @Override
     public Collection<V> values() {
-        Collection<V> vs = values;
+        Collection<V> vs = valuesView;
         if (vs == null) {
             vs = new Values();
-            values = vs;
+            valuesView = vs;
         }
         return vs;
     }
--- a/core/src/main/java/io/micronaut/core/util/EnumConstantsMap.java
+++ b/core/src/main/java/io/micronaut/core/util/EnumConstantsMap.java
@@ -1,0 +1,496 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.core.util;
+
+import org.jspecify.annotations.Nullable;
+
+import java.util.AbstractCollection;
+import java.util.AbstractMap;
+import java.util.AbstractSet;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.BiConsumer;
+
+/**
+ * An enum-keyed map that behaves like {@link java.util.EnumMap} but is given the constants of the
+ * enum instead of looking them up reflectively from its class.
+ *
+ * <p>The values are kept in an array indexed by {@link Enum#ordinal()}. An empty slot is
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/micronaut/core/util/EnumConstantsMap.java
```diff
@@ -0,0 +1,496 @@
...
+                    return new SlotEntry(index);
+                }
+            };
+        }
+
+        @Override
+        public boolean contains(@Nullable Object o) {
+            return o instanceof Entry<?, ?> entry
+                && containsKey(entry.getKey())
+                && Objects.equals(get(entry.getKey()), entry.getValue());
```
graemerocher: `Set.contains` is specified in terms of `Objects.equals(candidate, stored)`, but this reverses the operands for the value. Consequently `entrySet().contains` (and `remove`, which delegates to it) can disagree with both the `Set` contract and `EnumMap`. I reproduced this with a stored value whose `equals(candidate)` returns true while the candidate value does not equal the stored value: this map returns true, whereas `EnumMap.entrySet().contains(candidate)` returns false. Please cache the key and compare `Objects.equals(entry.getValue(), get(key))` instead.

graemerocher (commented):
Reviewed the new map implementation and the three converted call sites, with particular attention to publication and concurrent access. I did not find a new concurrency regression: `HTTP2_ERRORS` is populated under class initialization, the router maps are built before final-field publication and then read-only, and `JacksonFeatures` retains the same mutable/non-thread-safe semantics it had with `EnumMap`. The focused `EnumConstantsMapSpec` suite passes (20 tests). I left one inline finding for a reproducible `entrySet().contains`/`remove` contract mismatch.

core/src/main/java/io/micronaut/core/util/EnumConstantsMap.java
dstepanov: Fixed in 1d2c166208. The entry set's `contains` and `remove` now read the key and value once and use `containsMapping(key, value)`: it finds the slot and compares `mask(candidate).equals(stored)`, the same way round as `EnumMap.containsMapping` and `Set.contains`. The new test `entrySet contains and remove compare the candidate value with the stored one, as EnumMap does` stores a value that equals anything; it fails without the fix and matches `EnumMap` with it. I also checked the other comparisons: `containsValue` already put the candidate first, and `equals` and `Entry.equals` put this map's value first, as `Map.equals` and `Map.Entry.equals` specify.

graemerocher (approved):

PR #13126, opened 2026-09-11
Report reflection with the NoReflection ErrorProne check

dstepanov:
Integrates [micronaut-projects/errorprone-no-reflection](https://github.com/micronaut-projects/errorprone-no-reflection) 1.0.1: the main Java sources of every library module are compiled with its `NoReflection` check, and reflection fails the compilation outside what the module allows.

## What changes

- **`convention-no-reflection`** applies `io.micronaut.errorprone.no-reflection` on top of `convention-nullaway`; `convention-library` applies it instead of `convention-nullaway`, so it runs in the same ErrorProne compilation as NullAway.
- **Allowances live in the module build files**, class by class with the categories each class uses: 240 classes in 27 modules, the reflection the code base has today. A class scope takes in its nested classes.
  ```kotlin
  noReflection {
      allowIn("io.micronaut.core.annotation.AnnotationMetadata", "ANNOTATIONS", "CLASS_LOADING", "ENUM_CONSTANTS", "REFLECTIVE_ACCESS")
  }
  ```
- **`micronaut-reflection`** allows its whole package, `io.micronaut.reflection.*`: reflection is its purpose.
- **`-PnoReflection.record`** reports as warnings, ignores the class allowances (package ones stay) and prints, for each module, the `noReflection` block allowing exactly what its code uses.
- **Versions:** the plugin needs the ErrorProne Gradle plugin 5.1.1 (was 4.4.0) and ErrorProne 2.50.0 (was 2.48.0).
- **`micronaut-module-info`** strips every dependency but ErrorProne's and NullAway's; it now keeps the check too.
- **`CONTRIBUTING.md`** has a *No reflection* section.

A new reflective call outside the allowances fails like this:

```
StringUtils.java:41: error: [NoReflection] Reflection is not allowed here [CLASS_NAMES]
    private static final String NR_PROBE = StringUtils.class.getSimpleName();
```

## Where the reflection is today

| Category | Reports |
|---|---|
| `REFLECTION_UTILS` | 256 |
| `REFLECTIVE_ACCESS` | 173 |
| `GENERIC_SIGNATURES` | 134 |
| `ANNOTATIONS` | 119 |
| `SERVICE_LOADING` | 107 |
| `CLASS_NAMES` | 91 |
| `CLASS_MEMBERS` | 81 |
| `HANDLES` | 67 |
| `CLASS_LOADING` | 62 |
| `ENUM_CONSTANTS` | 53 |
| `ANNOTATION_SYNTHESIS` | 46 |
| `INTERFACES`, `TARGET_MEMBERS`, `FIELD_UPDATERS`, `SERIALIZATION`, `PROXY` | 43 |

By module, most of it is in `core` (276), `core-processor` (266), `reflection` (199) and `inject` (185).

## Verification

- `compileJava testClasses --no-build-cache --rerun-tasks`: every one of the 66 Java compilations ran, with no report.
- `testClasses javadoc checkstyleMain` passes.
- A `getSimpleName()` call added to `StringUtils` fails `:micronaut-core:compileJava`.

## Known limitation

Record mode prints only what javac prints. javac stops after 100 warnings per compilation, and an up-to-date or cached task prints nothing, so a record run can miss classes: `core-processor` recorded 20 of its 43. Enforcement is not affected.

## Second commit: reflection removed from 23 runtime classes

Removes 54 of the 687 reports outside the `reflection` module and the annotation processors. 21 `allowIn` lines are removed and 2 lose a category.

| Category | Change |
|---|---|
| `ENUM_CONSTANTS` (20) | Lookups by name loop over `values()` and keep the exception type and message; enum-to-enum mappings (`NettyCookie`, `ConnectionManager`) are `switch`es; the fixed access-log event sets are `Set.of`; `Http2ServerHandler` indexes its prebuilt exceptions by `ordinal()` |
| `HANDLES` (23), `CLASS_LOADING` (2) | `LoomSupport` calls `Thread.ofVirtual()`, the builder, `Executors.newThreadPerTaskExecutor` and `Thread.isVirtual()` directly, as Java 25 has them; the `VirtualThread` class probe stays |
| `CLASS_NAMES` (9) | `getSimpleName()` of a fixed class becomes the literal (`Primary.SIMPLE_NAME`, `CronExpression.toString`, the retry operation labels, the access-log request attribute key) |

**Performance:** nothing on a request or thread path gets slower. `LoomSupport.isVirtual`, which is called on every virtual-thread scheduling decision, goes from a `static final` method handle to a direct call. The per-request attribute key and per-call retry labels no longer call `getSimpleName()`. The HTTP/2 error lookup and the SameSite cookie mapping get cheaper. The access-log sets are only read while the format is parsed, and the enum name loops run at configuration time.

**Verification:** `compileJava` passes with the smaller allowances. 103 targeted tests ran: 100 passed and 3 were skipped as OS-specific. They cover logging levels, cron, the environment deducer, retry, Netty cookies, client trace logging, the logger endpoint, access log, HTTP/2 errors, virtual threads, `@Primary` and SameSite conversion. Checkstyle is clean on the changed files.

Drafted with GPT-6 and reviewed. Most of the remaining reports are API contracts (`EnumSet` returns, synthesized annotations, `Method`/`Field` accessors), hot paths (`DefaultRouter`'s `EnumMap`s, atomic field updaters), conversion of arbitrary enums, optional-dependency probes, or `getSimpleName()` text that users can see.


context/src/main/java/io/micronaut/logging/LoggingConverterRegistrar.java
```diff
@@ -39,7 +39,13 @@ public void register(MutableConversionService conversionService) {
                 return Optional.of(LogLevel.NOT_SPECIFIED);
             }
             try {
-                return Optional.of(Enum.valueOf(LogLevel.class, object.toString().toUpperCase(Locale.ENGLISH)));
```
graemerocher, line 42: why is this necessary? This does work on Graal and is it not worse for performance?

context/src/main/java/io/micronaut/scheduling/cron/CronExpression.java
```diff
@@ -315,7 +315,7 @@ public String getExpression() {
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + "<" + expr + ">";
```
graemerocher, line 318: I don't think getClass().getSimpleName() can be regarded as reflection. Reflection is something that initialises reflective classData inside the java.lang.Class

core-processor/build.gradle.kts
```diff
@@ -16,3 +16,49 @@ dependencies {
     compileOnly(libs.managed.kotlin.stdlib.jdk8)
 }
 
+
+noReflection {
```
graemerocher, line 19: shoudln't it just be disabled for the processor module? Seems unnecessary

graemerocher (changes requested):

graemerocher (approved):

context/src/main/java/io/micronaut/logging/LoggingConverterRegistrar.java
dstepanov: Reverted to `Enum.valueOf(LogLevel.class, ...)` in f7057a65c0; `LoggingConverterRegistrar` is allowed `ENUM_CONSTANTS` in `context/build.gradle.kts` instead.

context/src/main/java/io/micronaut/scheduling/cron/CronExpression.java
dstepanov: Reverted to `getClass().getSimpleName()` in f7057a65c0; `CronExpression` is allowed `CLASS_NAMES` in `context/build.gradle.kts` instead.

core-processor/build.gradle.kts
dstepanov: Done in f7057a65c0: the recorded allow-list is gone and the check is disabled for the module with `checkedSourceSets = emptySet()`.

PR #13129, opened 2026-09-11
Keep buffered body pieces in place, size the JSON output buffer and parse the form type once

graemerocher:
## Summary

- **Composite bound** (`NettyReadBufferFactory.compose`): the composite was created with the allocator default of 16 components, so once a body arrived in more than 16 pieces (every body over ~128 KiB, since the codec delivers 8 KiB pieces) `addComponents` consolidated it: one contiguous allocation of the full size, a copy of every piece and the release of the pieces, on the event loop, before the body could be read. The composite is now created with 4096 components (32 MiB of 8 KiB pieces), so bodies of the usual sizes are read from their pieces in place. More pieces than that are consolidated once, at the end, as before. Nothing downstream assumes contiguity: a 2-to-16 piece body was already a composite, and the readers go through `hasArray()`, `ByteBufInputStream` or `getBytes`.
- **Fully buffered bodies in the handlers** (`PipeliningServerHandler`, `MultiplexedServerHandler`): when a body arrives completely before `channelReadComplete` (HTTP/1 pieces, HTTP/2 DATA frames), both handlers assembled it with a composite at the allocator default of 16 components and `addComponent` per piece, which consolidates after every 16th piece, copying everything accumulated so far each time, so a body of n pieces was copied about n/16 times. Both now call a shared package-private `PipeliningServerHandler.composeBody`, which creates the composite with 4096 components and adds all pieces in one `addComponents` call (at most one consolidation, only beyond the bound; `addComponents` also takes ownership of every piece if it fails, which the loop did not).
- **JSON output buffer** (`ReadBufferFactory`, `ByteBodyFactory`, `NettyReadBufferFactory`, `JsonMessageHandler`): `buffer(writer)` started at the allocator default of 256 bytes (32 for the JDK factory) and grew by reallocating and copying; Jackson's generator writes through an 8000 byte encoding buffer, so every JSON response over 256 bytes forced a reallocation on its first write, before any data was in the buffer. New `buffer(expectedSize, writer)` overloads on `ReadBufferFactory` and `ByteBodyFactory` (`@since 5.2.2`, additive) pass the hint to the allocator as the initial capacity; `JsonMessageHandler.write` (a whole response body) uses 8 KiB to match Jackson's flush block. Responses up to that size are written without growing the buffer, larger ones skip the first, empty reallocation. `writePiece`, which `ResponseLifecycle` also calls once per element of a streamed response, keeps the factory's default sizing: those elements reach the channel uncopied, and a hint that size would hold 8 KiB of capacity per unread element (review finding, `1b27caf783`). `buffer(writer)` keeps its behaviour.
- **Form type** (`NettyHttpRequest`): `hasFormBody()` and `getRawFormFields()` parsed the form type (url-encoded, multipart with its boundary, or neither) from the `Content-Type` on every call, and every binder that can bind a form argument asks `hasFormBody()` first, so a multipart route with several parts parsed the same header once per part plus once for the field stream. The parsed form type is now kept together with the `MediaType` instance it was parsed from; `NettyHttpHeaders` already caches the parsed content type and drops it when the headers are modified, so a changed `Content-Type` yields a new instance and a fresh parse.
- **byte[] parts** (`HttpConverterRegistrar`): converting a `CompletedFileUpload` to `byte[]` went through the message body reader for the part's content type, which read the upload's `InputStream` with `readAllBytes()` (16 KiB chunks, then a concatenation), so a 64 KiB part allocated about twice its size. `byte[]` is a raw view of the bytes like the `InputStream` and `ReadBuffer` targets the converter already serves directly, so it is now served from `CompletedPart.getBytes()`, one copy.

## Benchmarks

**These numbers replace the ones first posted on this PR.** The first set was taken with the `ServerScenariosBenchmark` from the initial version of #13112, which wrote the whole request into the `EmbeddedChannel` as one `ByteBuf`; Netty then handed the server a single body piece even for the `*_CHUNKED` scenarios, so those rows never exercised the multi-piece path this PR changes and the deltas were overstated (`BYTES_POST_LARGE` -56.6%, `BYTES_POST_LARGE_CHUNKED` -70.8%, `JSON_GET_LARGE_LIST` -14.9%). The merged #13112 harness writes each part with its own `writeInbound` and fails the run if a chunked scenario arrives in fewer than two pieces. The table below was measured on that harness.

Base is `origin/5.2.x` at 593c8d1012 (this branch rebased on it), built into a JMH jar in the same session as the branch jar, both on JDK 25. Flags: `-f 1 -wi 5 -w 3s -i 5 -r 3s -t 1 -bm avgt -tu us -prof gc -rf json -jvmArgs "-Dmicronaut.python.enabled=false -Djmh.executor=CUSTOM -Djmh.executor.class=io.micronaut.http.server.stack.JmhFastThreadLocalExecutor"`. `ServerScenariosBenchmark` is the merged file extended with `BYTES_POST_LARGE`, `BYTES_POST_LARGE_CHUNKED` (1 MiB `application/octet-stream` bound to `@Body byte[]`, with Content-Length and as 4 KiB chunks), `JSON_GET_LARGE_LIST` (500 items, ~30 KiB response), `JSON_UNTYPED_RESPONSE` (`HttpResponse<?>` route) and `LARGE_TEXT_RESPONSE_ACCEPT_IDENTITY` (64 KiB text with `Accept-Encoding: identity;q=1.0, foo;q=0.5, bar`); the extension is used for measurement only and not committed. The machine was otherwise idle (the other sessions were paused; 1-minute load below 3 at the start of every fork, 3-3.5 during a run from the benchmark itself). The base jar was run twice; "within noise" means the difference is smaller than the sum of the two error bars (99.9% CI) or smaller than the spread between the two base runs, whichever is larger. Allocation deltas under 1% are reported as within noise because same-jar forks differ by that much.

| ServerScenariosBenchmark | MULTIPART_FILE_COMPLETED | 36.7 ± 3.6 us/op | 37.9 ± 6.6 us/op | +1.3 us/op (+3.4%) (within noise) | 88,499 | 88,435 | -64 (-0.1%) (within noise) |
| ServerScenariosBenchmark | MULTIPART_FILE_BYTES | 39.7 ± 1.4 us/op | 37.4 ± 3.6 us/op | -2.2 us/op (-5.6%) (within noise) | 154,051 | 88,235 | -65,816 (-42.7%) |
| ServerScenariosBenchmark | MULTIPART_FILE_CHUNKED | 53.7 ± 2.6 us/op | 51.7 ± 2.5 us/op | -1.9 us/op (-3.6%) (within noise) | 97,660 | 96,204 | -1,456 (-1.5%) |
| ServerScenariosBenchmark | FORM_URLENCODED_POJO | 6.0 ± 0.3 us/op | 6.2 ± 0.3 us/op | +0.2 us/op (+3.3%) (within noise) | 14,560 | 14,536 | -24 (-0.2%) (within noise) |
| ServerScenariosBenchmark | JSON_POST_LARGE_LIST | 169.5 ± 2.5 us/op | 169.9 ± 5.4 us/op | +0.4 us/op (+0.2%) (within noise) | 247,684 | 247,764 | +80 (+0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_POST_LARGE_LIST_CHUNKED | 174.3 ± 1.8 us/op | 173.2 ± 3.1 us/op | -1.2 us/op (-0.7%) (within noise) | 249,564 | 249,492 | -72 (-0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_POST_LARGE_LIST_REACTIVE_CHUNKED | 284.8 ± 7.0 us/op | 295.8 ± 5.6 us/op | +11.0 us/op (+3.9%) (within noise) | 723,244 | 723,309 | +65 (+0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_STREAM_RESPONSE | 319.7 ± 3.0 us/op | 320.1 ± 12.1 us/op | +0.3 us/op (+0.1%) (within noise) | 342,998 | 342,975 | -24 (-0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_STREAM_RESPONSE_NDJSON | 207.1 ± 8.3 us/op | 212.2 ± 7.9 us/op | +5.1 us/op (+2.4%) (within noise) | 244,135 | 244,135 | +0 (+0.0%) (within noise) |
| ServerScenariosBenchmark | LARGE_TEXT_RESPONSE | 14.5 ± 0.3 us/op | 13.9 ± 0.1 us/op | -0.6 us/op (-4.1%) | 3,505 | 3,537 | +32 (+0.9%) (within noise) |
| ServerScenariosBenchmark | BYTES_POST_LARGE | 68.5 ± 0.6 us/op | 46.7 ± 0.5 us/op | -21.8 us/op (-31.9%) | 1,082,149 | 1,084,131 | +1,982 (+0.2%) (within noise) |
| ServerScenariosBenchmark | BYTES_POST_LARGE_CHUNKED | 230.6 ± 3.1 us/op | 201.6 ± 9.5 us/op | -29.0 us/op (-12.6%) | 1,160,424 | 1,145,943 | -14,481 (-1.2%) |
| ServerScenariosBenchmark | JSON_GET_LARGE_LIST | 74.9 ± 0.5 us/op | 74.0 ± 0.9 us/op | -0.9 us/op (-1.2%) (within noise) | 108,301 | 108,213 | -88 (-0.1%) (within noise) |
| ServerScenariosBenchmark | JSON_UNTYPED_RESPONSE | 1.7 ± 0.1 us/op | 1.7 ± 0.1 us/op | -0.0 us/op (-0.0%) (within noise) | 3,744 | 3,744 | +0 (+0.0%) (within noise) |
| ServerScenariosBenchmark | LARGE_TEXT_RESPONSE_ACCEPT_IDENTITY | 14.3 ± 0.2 us/op | 15.1 ± 0.2 us/op | +0.8 us/op (+5.4%) | 4,281 | 4,313 | +32 (+0.7%) (within noise) |

**Reading the numbers.** `BYTES_POST_LARGE` -21.8 us/op (-31.9%) and `BYTES_POST_LARGE_CHUNKED` -29.0 us/op (-12.6%) are the handler change: a 1 MiB body arriving as 128 pieces (256 as 4 KiB chunks) before `channelReadComplete` was consolidated 8 (16) times by `addComponent`, copying several MiB per request; it is now composed in place, and the only remaining copy is the `byte[]` the controller asked for (the ~1,080,000 B/op on both sides). The chunked row improves less in relative terms because, once each 4 KiB chunk is decoded separately, the codec's per-chunk work dominates the row (230 us base against 68 us for the same body with a Content-Length). `MULTIPART_FILE_BYTES` -65,816 B/op (-42.7%) at unchanged time is the `byte[]` part change: the 64 KiB part is copied once instead of through `readAllBytes`. `JSON_GET_LARGE_LIST` is within noise on this harness (-0.9 us); the 8 KiB JSON buffer saves one reallocation per response, which is not resolvable at this row's ±0.5-0.9 us error bars, and it is no longer claimed as a measured improvement. `MULTIPART_FILE_CHUNKED` is within noise on both time and allocation (-1.5%); the +78% allocation the coalescing harness showed on that row does not occur when the 17 pieces actually arrive separately, because the base handler then builds the same composite and consolidates it once at the 16th piece. `LARGE_TEXT_RESPONSE` -4.1% and `LARGE_TEXT_RESPONSE_ACCEPT_IDENTITY` +5.4% are opposite readings of two near-identical 14 us rows, which is the fork-to-fork variance seen between the two base runs on those rows (±5%); neither is a change from this PR, which does not touch the response text path.

`ControllersBenchmark` and `FullHttpStackBenchmark` are not affected by the harness fix (they do not use `ServerScenariosBenchmark`); their rows from the first run, base 5c4984bd5b, are kept for reference:

| ControllersBenchmark | TFB_LIKE | 2,625 ± 138 ns/op | 2,526 ± 70 ns/op | -99 ns/op (-3.8%) (within noise) | 6,320 ± 1 | 6,352 ± 1 | +32 (+0.5%) |
| ControllersBenchmark | TFB_STRING | 2,432 ± 17 ns/op | 2,490 ± 70 ns/op | +58 ns/op (+2.4%) (within noise) | 6,456 ± 1 | 6,456 ± 1 | +0 (+0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_BEANS1 | 1,831 ± 47 ns/op | 1,851 ± 226 ns/op | +21 ns/op (+1.1%) (within noise) | 3,792 ± 1 | 3,792 ± 1 | -0 (-0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_BEANS2 | 1,876 ± 18 ns/op | 1,904 ± 37 ns/op | +28 ns/op (+1.5%) (within noise) | 3,888 ± 1 | 3,912 ± 1 | +24 (+0.6%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS1 | 1,897 ± 55 ns/op | 1,923 ± 56 ns/op | +26 ns/op (+1.3%) (within noise) | 3,832 ± 1 | 3,896 ± 1 | +64 (+1.7%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS2 | 1,894 ± 101 ns/op | 1,969 ± 103 ns/op | +75 ns/op (+3.9%) (within noise) | 3,920 ± 1 | 3,928 ± 1 | +8 (+0.2%) |
| ControllersBenchmark | TFB_LIKE_MAP | 1,757 ± 36 ns/op | 1,763 ± 53 ns/op | +6 ns/op (+0.3%) (within noise) | 3,968 ± 1 | 3,920 ± 1 | -48 (-1.2%) |
| ControllersBenchmark | MISSING_QUERY_PARAMETER | 2,122 ± 201 ns/op | 2,029 ± 55 ns/op | -94 ns/op (-4.4%) (within noise) | 4,920 ± 1 | 4,856 ± 1 | -64 (-1.3%) |
| FullHttpStackBenchmark | MICRONAUT | 2,612 ± 170 ns/op | 2,629 ± 194 ns/op | +18 ns/op (+0.7%) (within noise) | 6,504 ± 2 | 6,128 ± 2 | -376 (-5.8%) |

`FullHttpStackBenchmark` allocation -376 B/op (-5.8%) with time within noise: the only change on that path is the 8 KiB initial buffer for the 27-byte JSON response, which the adaptive allocator serves from a different size class; it is reported as measured, not claimed.

## Testing

- `NettyReadBufferTest.composeKeepsPiecesInPlace` (new): composes 1280 pieces (a 10 MiB body in 8 KiB pieces) through a counting allocator and checks that no allocation larger than a piece happens, the content is intact and the pieces are released with the composite. Fail-before verified against the 5.2.x factory: `expected: <0> but was: <1>` at line 66. `composeConsolidatesOnlyOnce` (updated) composes 4097 pieces and checks that beyond the bound consolidation still happens exactly once and releases the pieces.
- `AbstractReadBufferTest.bufferWithExpectedSize` (new, runs for the Netty and the JDK factory): exact, smaller, zero and larger hints, a 100,000 byte body written in odd-sized pieces through an 8 KiB hint, and an empty write. `NettyReadBufferTest.bufferWithExpectedSizeRequestsThatCapacity` (new): the hint reaches the allocator as the initial capacity. These test the new API; there is no fail-before.
- `NettyHttpRequestFormTypeSpec` (new, 12 cases): url-encoded (with charset, upper case), multipart with unquoted and quoted boundary, multipart without boundary, other types and no header, stable across repeated calls, follows a `Content-Type` changed through `mutate()`, and `getRawFormFields()` still rejects a non-form request. Equivalence test by design.
- `CompletedFileUploadConverterTest` (new, 4 tests): in-memory upload to `byte[]` with and without a content type, to `String` through the reader, to `InputStream`. The `byte[]` cases pass before and after; the allocation change is in the multipart benchmark.
- `./gradlew :micronaut-buffer-netty:test :micronaut-buffer-netty:checkstyleTest`: 54 tests, 0 failures.
- `./gradlew :micronaut-http:test :micronaut-http:checkstyleTest`: 1861 tests, 0 failures.
- `./gradlew :micronaut-json-core:test`: 26 tests, 0 failures.
- `./gradlew :micronaut-http-netty:test`: 148 tests, 3 skipped (pre-existing), 0 failures.
- `PipeliningServerHandlerSpec` gains `a body that arrives in many pieces before read complete is composed without copying`: a 200 piece body written through a counting allocator before `flushInbound()`; no allocation larger than a piece, body intact, `received.close()` and `finishAndReleaseAll()` keep the spec's resource-scope check green. Fail-before verified against the 5.2.x handlers: `allocator.largeAllocations == 0` fails with `12`.
- `./gradlew :micronaut-http-server-netty:test`: 1169 tests, 22 skipped (pre-existing), 0 failures (run after the last commit; the run before the handler commit was 1168 tests, 0 failures).
- `./gradlew :test-suite-http2-server-tck-netty:test` (the HTTP/2 handler is touched): 263 tests, 3 skipped (pre-existing), 0 failures.
- `./gradlew :test-suite:test --tests 'io.micronaut.upload.*'` (multipart `byte[]`, `CompletedFileUpload`, disk and streaming uploads): 68 tests, 1 skipped (pre-existing), 0 failures.
- `checkstyleMain` clean for `core`, `buffer-netty`, `http`, `json-core` and `http-server-netty` apart from the pre-existing FileLength violation; `japiCmp` clean for all five.

## Out of scope

- `BaseSharedBuffer.add` still retains and releases each piece even when there is a single subscriber and no other reservation. That is buffer-ownership code shared by the streaming request and response paths; changing it needs its own review and tests and is left out of this PR.
- The `HttpData` to `byte[]`/`CharSequence` converters in `NettyConvertersSpi` belong to the same old multipart path as the removed class, but they are registered into the shared `ConversionService`, so removing them is a behaviour change without a performance benefit; they stay.
- `CompletedPart.getBytes()` itself is one copy of the upload, which a `byte[]` API cannot avoid. The remaining multipart allocation is the decoder's per-piece work.
- A body of more than 4096 pieces (over 32 MiB at 8 KiB per piece, or a smaller body sent in very small chunks) is still consolidated once.
- `FormDemuxer` hands an available (fully buffered) multipart body to the decoder as one buffer; when that buffer is a composite, the decoder's `readBytes` for each part goes through `CompositeByteBuf.nioBuffer`, which merges the components on the heap first (visible in a JFR allocation profile of the multipart rows: most of the `byte[]` allocation is under `CompositeByteBuf.nioBuffer <- AdaptiveByteBuf.setBytes <- MultipartDecoder.next`). Feeding the decoder one component at a time, as the streaming path already does, would avoid that merge for every buffered multipart body over 8 KiB, not only the ones this PR moves onto the composite path; it needs care with slice ownership and is left for a separate change.
- `JsonMessageHandler.writeTo` (the `OutputStream` path used by streaming responses) is unchanged; only `writePiece` sizes its buffer.

## Review follow-ups

- @dstepanov: the JSON size hint applied to streamed elements too (8.2 MB of capacity for 20 KB of data over 1,000 elements); now whole-body only, see above (`1b27caf783`).
- @dstepanov: the handlers cleared their piece lists only after `composeBody` returned, so a failed compose left released buffers reachable for a later `devolveToStreaming`/`discard`; cleared before composing (`1b27caf783`).
- Sonar: `composeBody`'s release guard is a `finally` instead of `catch (Throwable)` (`7ca523364a`).

## Codex review (gpt-5.6-sol, high)

One P1, verified and fixed on the branch: the new `byte[]` conversion routes a disk-backed part through `CompletedFileUpload.File.toReadBuffer()`, which opened `Files.newInputStream` and handed it to `copyOf(InputStream)` without closing it. That was a pre-existing leak in `toReadBuffer()` reachable from `getBytes()` and the in-memory copy; the new path made it reachable from `@Part byte[]` where the old reader path closed its stream. `toReadBuffer()` now closes the stream (`262265afd1`), which fixes every caller.


json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
```diff
@@ -72,6 +72,14 @@ public final class JsonMessageHandler<T> implements MessageBodyHandler<T>, Custo
      */
     public static final int ORDER = -10;
 
+    /**
+     * Starting capacity of the buffer a JSON response is serialised into. Jackson's generator
+     * writes through an 8000 byte encoding buffer, so a response of up to that size arrives in
+     * one write; with the allocator's 256 byte default that first write already forced the
+     * buffer to grow.
+     */
+    private static final int WRITE_BUFFER_SIZE = 8192;
```
graemerocher, line 81: should this be configurable through the Jackson / Netty layers

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
```diff
@@ -152,7 +160,7 @@ public void writeTo(Argument<T> type, @Nullable MediaType mediaType, T object, M
     @Override
     public CloseableByteBody writePiece(ByteBodyFactory bodyFactory, HttpRequest<?> request, HttpResponse<?> response, Argument<T> type, MediaType mediaType, T object) throws CodecException {
         try {
-            return bodyFactory.buffer(s -> jsonMapper.writeValue(s, object));
+            return bodyFactory.buffer(WRITE_BUFFER_SIZE, s -> jsonMapper.writeValue(s, object));
```
dstepanov: `writePiece` is also called once per element of a streamed response (`ResponseLifecycle` → `concatMap` → `writePieceAsync`), and those pieces reach the channel without being copied. So now every element takes 8 KiB of capacity, however small it is.

I measured a `Flux.range(0, 1000).map { [i: it] }` route with an outbound handler that sums `readableBytes()` and `capacity()` of each `ByteBuf` written to the socket:

| | readable | capacity |
|---|---|---|
| base `JsonMessageHandler` | 20,017 | 269,269 |
| this branch | 20,017 | 8,205,269 |

`ChannelOutboundBuffer` counts pending data by `readableBytes()` (`DefaultMessageSizeEstimator`). With a slow client, Netty's default 64 KiB high watermark therefore lets roughly 3,000 small elements queue: about 26 MiB of direct memory per connection instead of under 1 MiB. On the JDK factory each piece also holds a `byte[8192]`.

This ties into the configurability question above: a streaming caller wants no hint at all. Could the hint be applied only where a single whole body is written, or could a piece that ends up much smaller than its capacity be copied into a right-sized buffer?

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
```diff
@@ -635,11 +670,11 @@ void read(Object message) {
...
                     fullBody = buffer.getFirst().content();
                 } else {
-                    CompositeByteBuf composite = requiredCtx().alloc().compositeBuffer();
+                    List<ByteBuf> pieces = new ArrayList<>(buffer.size());
                     for (HttpContent c : buffer) {
-                        composite.addComponent(true, c.content());
+                        pieces.add(c.content());
                     }
-                    fullBody = composite;
+                    fullBody = composeBody(requiredCtx().alloc(), pieces);
```
dstepanov, line 680: Minor, and it only matters on an out-of-memory error: if `composeBody` throws (in practice only when consolidating more than 4096 pieces fails to allocate), `addComponents` has already released or taken every piece. But `buffer` still holds the `HttpContent`s, because `buffer.clear()` on the next line never runs. A later `handleUpstreamError` → `devolveToStreaming()` or `discard()` would then use or release them a second time. `bufferedContent` in `MultiplexedServerHandler.onDataRead` has the same problem. Clearing the list before composing would close this. The old loop had the same issue.

dstepanov (commented):
Went through the ByteBuf ownership, the composite bound, the form-type cache and the `byte[]` part conversion. `buffer-netty` tests, `PipeliningServerHandlerSpec` (49) and `NettyHttpRequestFormTypeSpec` (12) pass with paranoid leak detection. The composite paths look right: `addComponents` takes every piece, and the finally block releases the composite. The form-type cache is keyed by the `MediaType` instance, and `NettyHttpHeaders` resets that on `set`/`contentType`, so charset and boundary parameters parse exactly as before.

One real issue: the 8 KiB JSON buffer also applies to each element of a streamed response (inline). One minor ownership nit on the failure path.

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
graemerocher: Good catch, thank you for the measurement. Fixed in 1b27caf783 the first way you suggest: `JsonMessageHandler` now overrides `write()` (the whole-body path) with the 8 KiB hint and `writePiece` is back to the factory's default sizing, so streamed elements are right-sized as before. `JsonMessageHandlerBufferSizeTest` records which variant each path calls (`write` → 8192, `writePiece` → none). The constant's javadoc explains why the hint is whole-body only.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
graemerocher: Fixed in 1b27caf783: `PipeliningServerHandler` clears `buffer` before calling `composeBody`, and `MultiplexedServerHandler.onDataRead` nulls `bufferedContent` before handing the list over, so a failed compose leaves nothing for `devolveToStreaming`/`discard` to touch twice.

json-core/src/main/java/io/micronaut/json/body/JsonMessageHandler.java
graemerocher: With the hint now limited to a whole response body (see the thread below on streamed elements) I'd keep it a constant: it mirrors Jackson's own 8000-byte encoding buffer, which is where the first write comes from, so there is no separate knob that could be tuned independently of Jackson's without losing the point. The streaming case, where a caller genuinely wants no hint, is handled by not applying it there rather than by configuration. Happy to add a property if you'd still like one.

yawkat (approved):

PR #13130, opened 2026-09-11
Cache compression eligibility and Jackson readers and writers per type

graemerocher:
## Summary

- **Compression eligibility** (`MediaType`, `Compressor`): `MediaType.isTextBased()` ran up to five regular expressions against the type name on every call, and it is called for every response that reaches the compression threshold. It now uses the equivalent `startsWith`/`endsWith`/`equals` checks (name checks case-sensitive, subtype checks case-insensitive, exactly as the patterns behaved) and remembers the answer in a byte field of the `MediaType` instance; `MediaType.of(String)` returns shared constants for the usual response types, so those answer with one field read. `Compressor` used to split every `Accept-Encoding` value into a list of substrings before scoring them; it now walks the comma-separated values in place with token bounds, producing the same algorithm choice. `determineEncoding` and the `Algorithm` enum became package-private for the test.
- **Jackson reader/writer cache** (`JacksonDatabindMapper`): the mapper kept the last `ObjectReader`/`ObjectWriter` in a single slot keyed by `Argument` identity. The server never specialises the JSON handler per route, so every JSON body goes through one shared mapper: two routes with different response types evicted each other on every request, and a response whose type is resolved from the instance (`HttpResponse<?>` routes, error responses, streamed items: `Argument.ofInstance` creates a new `Argument` per call) always missed and ran `objectMapper.writerFor(constructType(...))` per response. The caches are now fixed-size arrays (64 slots each) indexed by type hash; an entry matches an argument with the same type, type arguments and `@JsonView` class (`Argument.equalsType`), with the identity check kept as the first, cheapest test. A viewed argument uses a slot derived from the view too, so viewed and plain arguments of one type do not evict each other. A slot collision replaces the entry, so the caches never hold more than 64 entries each and a lookup never allocates. `createReader`/`createWriter` became package-private for the test.

### Thread safety

`MediaType.textBased` is a plain byte written with a value that depends only on final fields; two threads racing write the same value. The mapper caches are `AtomicReferenceArray`s of immutable records; a stale read only costs a recomputation and a replacement of that slot.

## Benchmarks

Base is `origin/5.2.x` at 5c4984bd5b (the parent of both branches), built into a JMH jar in the same session as the branch jars, all from the same JDK 25. Every run: `-f 1 -wi 5 -w 3s -i 5 -r 3s -t 1 -bm avgt -prof gc -jvmArgs "-Dmicronaut.python.enabled=false -Djmh.executor=CUSTOM -Djmh.executor.class=io.micronaut.http.server.stack.JmhFastThreadLocalExecutor"`, `-tu ns` for `ControllersBenchmark` and `FullHttpStackBenchmark`, `-tu us` for `ServerScenariosBenchmark`. `ServerScenariosBenchmark` is the file from PR #13112 extended with `BYTES_POST_LARGE`, `BYTES_POST_LARGE_CHUNKED` (1 MiB `application/octet-stream` bound to `@Body byte[]`, with Content-Length and as 4 KiB chunks), `JSON_GET_LARGE_LIST` (500 items, ~30 KiB response), `JSON_UNTYPED_RESPONSE` (`HttpResponse<?>` route) and `LARGE_TEXT_RESPONSE_ACCEPT_IDENTITY` (64 KiB text with `Accept-Encoding: identity;q=1.0, foo;q=0.5, bar`); it is used for measurement only and not committed.

The machine was shared with other sessions running Gradle builds for most of the afternoon (1-minute load between 5 and 37). A runner script started each measurement only once the 1-minute load average was below 3 and no other java process was above 40% CPU, waiting up to 90 minutes for that; runs whose load rose during the measurement were discarded and repeated (a `ControllersBenchmark` base run with error bars up to ±1,197 ns, a `ServerScenariosBenchmark` base run with ±1,299 us on one row, and a `FullHttpStackBenchmark` run that failed with a `RejectedExecutionException` from the JMH executor when the load spiked to 24 mid-run). Start loads of the runs reported below: `FullHttpStackBenchmark` base 2.67 / A 2.96 / B 2.97, `ControllersBenchmark` base 2.63 / A 2.98 / B 2.82, `ServerScenariosBenchmark` base 2.93 / A 2.83 / B 2.90; the load rises to 3-3.5 during a run because of the benchmark itself. "Within noise" means the difference is smaller than the sum of the two error bars (99.9% CI). The ±8-88 B/op allocation differences on the `ControllersBenchmark` rows are the fork-to-fork variance seen on every previous PR (same-jar forks differ by that much); the rows below only claim the differences that are well outside it.

| ControllersBenchmark | TFB_LIKE | 2,625 ± 138 ns/op | 2,527 ± 41 ns/op | -97 ns/op (-3.7%) (within noise) | 6,320 ± 1 | 6,408 ± 1 | +88 (+1.4%) |
| ControllersBenchmark | TFB_STRING | 2,432 ± 17 ns/op | 2,510 ± 77 ns/op | +78 ns/op (+3.2%) (within noise) | 6,456 ± 1 | 6,456 ± 1 | +0 (+0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_BEANS1 | 1,831 ± 47 ns/op | 1,814 ± 34 ns/op | -17 ns/op (-0.9%) (within noise) | 3,792 ± 1 | 3,792 ± 1 | -0 (-0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_BEANS2 | 1,876 ± 18 ns/op | 1,894 ± 51 ns/op | +17 ns/op (+0.9%) (within noise) | 3,888 ± 1 | 3,896 ± 1 | +8 (+0.2%) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS1 | 1,897 ± 55 ns/op | 1,956 ± 53 ns/op | +59 ns/op (+3.1%) (within noise) | 3,832 ± 1 | 3,832 ± 1 | +0 (+0.0%) (within noise) |
| ControllersBenchmark | TFB_LIKE_ASYNC_BEANS2 | 1,894 ± 101 ns/op | 1,862 ± 41 ns/op | -32 ns/op (-1.7%) (within noise) | 3,920 ± 1 | 3,864 ± 1 | -56 (-1.4%) |
| ControllersBenchmark | TFB_LIKE_MAP | 1,757 ± 36 ns/op | 1,739 ± 58 ns/op | -18 ns/op (-1.0%) (within noise) | 3,968 ± 1 | 3,984 ± 1 | +16 (+0.4%) |
| ControllersBenchmark | MISSING_QUERY_PARAMETER | 2,122 ± 201 ns/op | 2,018 ± 42 ns/op | -104 ns/op (-4.9%) (within noise) | 4,920 ± 1 | 4,920 ± 1 | -0 (-0.0%) (within noise) |
| FullHttpStackBenchmark | MICRONAUT | 2,612 ± 170 ns/op | 2,943 ± 243 ns/op | +331 ns/op (+12.7%) (within noise) | 6,504 ± 2 | 6,536 ± 2 | +32 (+0.5%) |
| ServerScenariosBenchmark | MULTIPART_FILE_COMPLETED | 35.6 ± 0.6 us/op | 34.1 ± 0.5 us/op | -1.5 us/op (-4.2%) | 145,323 ± 22 | 145,275 ± 21 | -48 (-0.0%) |
| ServerScenariosBenchmark | MULTIPART_FILE_BYTES | 38.4 ± 0.5 us/op | 38.5 ± 0.8 us/op | +0.1 us/op (+0.1%) (within noise) | 211,067 ± 24 | 211,083 ± 24 | +16 (+0.0%) (within noise) |
| ServerScenariosBenchmark | MULTIPART_FILE_CHUNKED | 34.5 ± 1.0 us/op | 34.7 ± 0.5 us/op | +0.2 us/op (+0.7%) (within noise) | 82,155 ± 21 | 82,171 ± 21 | +16 (+0.0%) (within noise) |
| ServerScenariosBenchmark | FORM_URLENCODED_POJO | 6.1 ± 0.2 us/op | 5.9 ± 0.2 us/op | -0.2 us/op (-3.3%) (within noise) | 13,928 ± 4 | 14,032 ± 3 | +104 (+0.7%) |
| ServerScenariosBenchmark | JSON_POST_LARGE_LIST | 166.2 ± 3.1 us/op | 166.7 ± 1.7 us/op | +0.4 us/op (+0.3%) (within noise) | 245,956 ± 98 | 245,988 ± 98 | +32 (+0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_POST_LARGE_LIST_CHUNKED | 164.7 ± 3.0 us/op | 164.1 ± 1.0 us/op | -0.5 us/op (-0.3%) (within noise) | 246,580 ± 97 | 246,548 ± 98 | -32 (-0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_POST_LARGE_LIST_REACTIVE_CHUNKED | 409.1 ± 16.1 us/op | 406.0 ± 3.1 us/op | -3.2 us/op (-0.8%) (within noise) | 745,149 ± 239 | 725,165 ± 201 | -19,983 (-2.7%) |
| ServerScenariosBenchmark | JSON_STREAM_RESPONSE | 314.8 ± 8.6 us/op | 315.3 ± 3.6 us/op | +0.6 us/op (+0.2%) (within noise) | 342,838 ± 171 | 342,574 ± 183 | -264 (-0.1%) (within noise) |
| ServerScenariosBenchmark | JSON_STREAM_RESPONSE_NDJSON | 210.8 ± 16.9 us/op | 211.3 ± 5.5 us/op | +0.5 us/op (+0.2%) (within noise) | 244,070 ± 120 | 243,751 ± 125 | -319 (-0.1%) |
| ServerScenariosBenchmark | LARGE_TEXT_RESPONSE | 13.5 ± 0.2 us/op | 13.7 ± 0.0 us/op | +0.2 us/op (+1.4%) (within noise) | 3,545 ± 8 | 3,345 ± 8 | -200 (-5.6%) |
| ServerScenariosBenchmark | BYTES_POST_LARGE | 128.3 ± 9.7 us/op | 126.7 ± 1.2 us/op | -1.6 us/op (-1.2%) (within noise) | 1,062,926 ± 78 | 1,063,013 ± 76 | +88 (+0.0%) (within noise) |
| ServerScenariosBenchmark | BYTES_POST_LARGE_CHUNKED | 231.5 ± 13.1 us/op | 233.6 ± 1.8 us/op | +2.1 us/op (+0.9%) (within noise) | 1,072,192 ± 134 | 1,072,248 ± 136 | +56 (+0.0%) (within noise) |
| ServerScenariosBenchmark | JSON_GET_LARGE_LIST | 76.8 ± 1.0 us/op | 73.1 ± 0.6 us/op | -3.7 us/op (-4.9%) | 108,253 ± 45 | 107,901 ± 43 | -352 (-0.3%) |
| ServerScenariosBenchmark | JSON_UNTYPED_RESPONSE | 1.7 ± 0.1 us/op | 1.7 ± 0.0 us/op | +0.0 us/op (+2.2%) (within noise) | 3,688 ± 1 | 3,688 ± 1 | +0 (+0.0%) (within noise) |
| ServerScenariosBenchmark | LARGE_TEXT_RESPONSE_ACCEPT_IDENTITY | 15.2 ± 0.5 us/op | 14.0 ± 0.2 us/op | -1.2 us/op (-7.8%) | 4,297 ± 9 | 3,769 ± 8 | -528 (-12.3%) |

**Reading the numbers (this PR).** The compression eligibility change shows on the two rows that reach the compression threshold and carry a text type: `LARGE_TEXT_RESPONSE_ACCEPT_IDENTITY` -1.2 us/op (-7.8%) and -528 B/op (-12.3%) (the regex matchers plus the `Accept-Encoding` split), `LARGE_TEXT_RESPONSE` -200 B/op (-5.6%) (the regex matchers alone), and `JSON_GET_LARGE_LIST` -3.7 us/op (-4.9%), a 30 KiB `application/json` response that also goes through `isTextBased`. The Jackson cache shows on `JSON_POST_LARGE_LIST_REACTIVE_CHUNKED`: -19,983 B/op (-2.7%), about 100 B per element, the `ObjectReader` that was created for every `Flux<Item>` element. It does not show on `JSON_UNTYPED_RESPONSE` because buffered responses go through `JsonMessageHandler.writePiece`, which calls the untyped `jsonMapper.writeValue(stream, object)` and never resolves a writer for the argument; and the benchmark runs one route per fork, so the case of two routes evicting each other from the previous single slot is not measured here (it is covered by `severalTypesAreCachedAtOnce`). The registry bound does not change any hot path. Everything else is within noise, including `FullHttpStackBenchmark` (+331 ns with ±170/±243 error bars, and identical allocation apart from 32 B). `MULTIPART_FILE_COMPLETED` -1.5 us/op (-4.2%) is just outside its error bars but nothing in this PR runs on the multipart path beyond one cached `findReader` hit per request, so it is treated as run-to-run variance, not claimed.

## Testing

- New `MediaTypeTest.isTextBasedMatchesThePreviousRules` (19 cases): the string rules give the previous regex answers, including `TEXT/plain` and `application/vnd.api+JSON` being false (case-sensitive name checks) while `application/JSON` and `application/XML` are true (case-insensitive subtype checks), plus the cached second call and the static `isTextBased(String)`.
- New `CompressorTest` (28 header shapes): the in-place tokenizer picks the same algorithm as a copy of the previous split-based implementation, including empty tokens, leading/trailing commas, missing, malformed and repeated `q` values, wildcards, several header lines and unknown encodings. These two are equivalence tests by design and pass before and after.
- New `JacksonDatabindMapperTypeCacheTest` (7 tests): identity hits, type-equal hits for `Argument.ofInstance` and generic arguments, misses for different type arguments, four types cached at once across alternating calls, `@JsonView` as part of the key with the view actually applied to the output, and the specialised mapper bypassing the cache. Fail-before verified by running it against the 5.2.x mapper with only the visibility change: 4 of 7 fail (`typeEqualArgumentsHit` at line 38, `typeEqualGenericArgumentsHit` at line 50, `severalTypesAreCachedAtOnce` at line 75, `jsonViewIsPartOfTheKey` at line 95), each with `expected: <ObjectWriter@...> but was: <ObjectWriter@...>`.
- `./gradlew :micronaut-http:test :micronaut-http:checkstyleTest`: 1879 tests, 0 failures.
- `./gradlew :micronaut-jackson-databind:test :micronaut-jackson-databind:checkstyleTest`: 65 tests, 0 failures.
- `./gradlew :micronaut-json-core:test`: 26 tests, 0 failures.
- `./gradlew :micronaut-http-netty:test`: 148 tests, 3 skipped (pre-existing), 0 failures.
- `./gradlew :micronaut-http-server-netty:test`: 1184 tests, 22 skipped (pre-existing), 1 failure on the first run: `FileCertificateProviderTest [1] format = JKS` with `TLSV1_ALERT_CERTIFICATE_REQUIRED` from the TLS client; it is a certificate-rotation test unrelated to this change (nothing here touches TLS) and passed when re-run alone (7 tests, 0 failures).
- `checkstyleMain` clean for `http`, `jackson-databind` and `http-server-netty` apart from the pre-existing `NettyHttpServerConfiguration` FileLength violation; `japiCmp` clean for all three.

## Out of scope

- `MediaType.isTextBased()` for a name containing a line terminator: the previous `^text/.*$` style patterns could not match such a name (`.` excludes line terminators), the string checks can. A media type name with a newline cannot come from a parsed header, so this only changes the answer for a `MediaType` constructed from such a string in code.
- `DefaultHttpCompressionStrategy.shouldCompress` still calls the static `MediaType.isTextBased(String)`, which goes through `MediaType.of`. For the constants that is a `switch` on the string; for other content types it still parses a new `MediaType` per response, and that instance's cache is not reused.
- `JsonMessageHandler.writePiece` (buffered JSON responses) serialises with the untyped `jsonMapper.writeValue(stream, object)`, so it never uses the per-type writer; only the `writeTo` path and the readers do. Switching it to the typed writer would let `@JsonView` and the argument's type arguments apply to buffered responses, but that is a behaviour change and not part of this PR.

## Review follow-ups

- Copilot: the two 64-slot per-type caches were allocated by every specialized mapper (`createSpecific` makes one per route) although those answer every lookup with their own reader and writer. They are now only allocated by the general constructor (`cab67e8ee0`).
- @dstepanov: the `Accept-Encoding` tokenizer's `contains` and `=` searches ran to the end of the whole header, making it O(n²) in the header length (~85 ms for 8 KiB of commas, which fits the default `maxHeaderSize`). Both scans are bounded by the entry end now, and `CompressorTest.longHeadersAreTokenizedInLinearTime` asserts four 8 KiB headers stay in the microseconds per call and still match the split-based reference (`704ff1a4a0`).
- @dstepanov: `severalTypesAreCachedAtOnce` depended on four fixed types landing in distinct identity-hash-derived slots (fails about one run in eleven). `JacksonDatabindMapper.slotOf` is package-private for tests and the test picks non-colliding types from ten candidates (`704ff1a4a0`).

## Codex review (gpt-5.6-sol, high)

No findings.

---


graemerocher:
Addressed the Copilot note on `JacksonDatabindMapper`: the two 64-slot per-type caches are now only allocated by the general constructor; the specialized mappers (`createSpecific`, `cloneWithViewClass`) set them to `null` since they answer every lookup with their own reader and writer and never reach the caches. `:micronaut-jackson-databind:test` 65/65.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
```diff
@@ -184,7 +196,15 @@ private Algorithm determineEncoding(List<String> acceptEncoding) {
         return null;
     }
 
-    private enum Algorithm {
+    /**
+     * Whether {@code needle} occurs inside {@code s[start, end)}.
+     */
+    private static boolean contains(String s, int start, int end, String needle) {
+        int i = s.indexOf(needle, start);
```
dstepanov: `indexOf(needle, start)` searches to the end of the whole header rather than to `end`, and so does `header.indexOf('=', start)` at line 141. With six `contains` calls plus the `=` search per token, the loop is O(n²) in the header length where the split-based version was linear.

I copied the old and new code into a standalone driver (JDK 25):

| `Accept-Encoding` | old | new |
|---|---|---|
| `gzip, deflate, br, zstd` | 0.2 µs | 0.1 µs |
| `,` × 8000 | 78 µs | 84,865 µs |
| `a,` × 4000 | 102 µs | 30,001 µs |
| `x;q,` × 2000 | 131 µs | 15,435 µs |

An 8 KiB header fits the default `maxHeaderSize` of 8192, so any client can spend ~85 ms of server CPU per request on a route with a compressible response. Bounding the searches keeps it linear, e.g. `s.indexOf(needle, start, end) != -1` and `header.indexOf('=', start, end)` (JDK 21+), or a `regionMatches` scan within `[start, end)`. A `CompressorTest` case with a long comma-only header and a time/iteration bound would guard it.

jackson-databind/src/test/java/io/micronaut/jackson/databind/JacksonDatabindMapperTypeCacheTest.java
```diff
@@ -0,0 +1,133 @@
...
+        ObjectWriter otherWriter = mapper.createWriter(other);
+        ObjectWriter itemsWriter = mapper.createWriter(items);
+        ObjectWriter othersWriter = mapper.createWriter(others);
+        ObjectReader itemReader = mapper.createReader(item);
+        ObjectReader otherReader = mapper.createReader(other);
+        for (int i = 0; i < 3; i++) {
+            // alternate between the types: each keeps its own entry
+            assertSame(itemWriter, mapper.createWriter(item));
+            assertSame(otherWriter, mapper.createWriter(other));
+            assertSame(itemsWriter, mapper.createWriter(Argument.listOf(Item.class)));
```
dstepanov: This failed for me once, running this class together with another test class in `:micronaut-jackson-databind:test`:

```
expected: <tools.jackson.databind.ObjectWriter@675ffd1d> but was: <tools.jackson.databind.ObjectWriter@56303475>
  at JacksonDatabindMapperTypeCacheTest.severalTypesAreCachedAtOnce(JacksonDatabindMapperTypeCacheTest.java:77)
```

The next 13 runs (10 alone, 3 combined) passed. The slot is derived from `typeHashCode`, which uses `Class.hashCode()`, an identity hash, so whether `Item`, `Other`, `List<Item>` and `Map<String, Other>` land in four distinct slots of 64 depends on JVM state; for independent hashes that is about 63·62·61/64³ ≈ 91%, so roughly 1 in 11 hash assignments collide and the entries evict each other. Maybe assert only that equal types hit when there is no collision (compute `slot` for the four types and skip or pick others when two share one), or make the slot size injectable for the test.

dstepanov (commented):
The per-type Jackson cache key looks right to me: `equalsType` compares `getType()` and the type variables recursively, which is exactly what `constructType` reads (wildcards and placeholders included), and the view is part of the match. A scratch test with two different `@JsonView` classes on one type, a plain argument and `Argument.ofInstance`, including 8 threads interleaving them 20k times each, always produced the right shape. The caches are per mapper, the `ObjectMapper` is final, and specialized/cloned mappers get none, so there are no stale entries after reconfiguration. `MediaType.isTextBased` only depends on final fields.

Two things inline: the new `Accept-Encoding` tokenizer is quadratic in the header length (measured ~85 ms per request for an 8 KiB header), and `severalTypesAreCachedAtOnce` depends on identity-hash slot placement and failed once locally.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
graemerocher: Fixed in 704ff1a4a0: `contains` now scans with `regionMatches` bounded by the entry end and the `=` search is a bounded `charAt` loop, so the tokenizer is linear in the header length again. `CompressorTest.longHeadersAreTokenizedInLinearTime` runs four 8 KiB headers (`,`×8000, `a,`×4000, `x;q,`×2000, `gzip;q=0.5,`×700) and asserts the per-call time stays under 2 ms and the result matches the split-based reference.

jackson-databind/src/test/java/io/micronaut/jackson/databind/JacksonDatabindMapperTypeCacheTest.java
graemerocher: Fixed in 704ff1a4a0: `JacksonDatabindMapper.slotOf(Argument)` is package-private for tests and `severalTypesAreCachedAtOnce` now picks the first four of ten candidate types that land in distinct slots (skipping via `assumeTrue` only if fewer than four do, which needs seven or more collisions among ten). It also checks that an equal but distinct `Argument` instance hits the same entry.

yawkat:
please split your PRs

graemerocher:
Split as requested:

- #13193 — the in-place `Accept-Encoding` tokenizer with the bounded scans from this review
- #13194 — the `MediaType.isTextBased` cache
- #13195 — the Jackson per-type reader/writer cache (with the specialized-mapper and slot-collision follow-ups from this review)

Closing this one.

PR #13131, opened 2026-09-11
Do not retain lazy proxy targets as dependents of the proxy's context

dstepanov:
## Problem

A lazy AOP proxy (`@Around(proxyTarget = true, lazy = true)`) stores the context returned by `copyForLazyProxyTarget` in `$beanResolutionContext`. It keeps that context for its whole life and resolves its target through it on every intercepted call (see `AopProxyWriter#getLazyInterceptedTargetMethod`).

Sometimes the target is created on every call. That happens for a `@Prototype` target, or for a bean whose scope isn't registered. In that case `resolveBeanRegistration` falls through to `createRegistration(..., dependent = true)`, which calls `addDependentBean` on the retained context. The list grows by one registration per call and is never cleared. A benchmark hit `OutOfMemoryError` after a few million calls:

```
at io.micronaut.context.AbstractBeanResolutionContext.addDependentBean(AbstractBeanResolutionContext.java:427)
at io.micronaut.context.DefaultBeanContext.createRegistration(DefaultBeanContext.java:3432)
at io.micronaut.context.DefaultBeanContext.resolveBeanRegistration(DefaultBeanContext.java:3252)
at io.micronaut.context.DefaultBeanContext.getProxyTargetBean(DefaultBeanContext.java:1659)
```

Nothing reads those dependents. The proxy's own registration takes its dependents from the original context when the proxy is created, and `destroyProxyTargetBean` uses the registration's dependents. For a cached lazy target it also uses the cached instance. It never looks at the retained context.

That benchmark used `@io.micronaut.runtime.context.scope.ThreadLocal`. The scope wasn't found because `AbstractTypeElementSpec.buildContext(...)` only registers the compiled beans plus a few built-ins. `ThreadLocalCustomScope` is a bean, and `DefaultCustomScopeRegistry` looks scopes up as `CustomScope` beans. So the `@ThreadLocal` target behaved like a prototype. The same thing happens in an application whose scope bean is missing.

## Fix

- `AbstractBeanResolutionContext#copyForLazyProxyTarget` marks the copy it returns (`isLazyProxyTarget()`, package-private).
- `DefaultBeanContext#createRegistration` creates the bean in a `copy()` of a marked context. The new registration and any "destroy after resolution" dependents go to that throwaway copy, not the retained context. A side benefit: concurrent calls on the proxy no longer write to the same resolution path and dependent lists.
- `copyForLazyProxyTarget` also removes two mutable attribute values from the copy the proxy keeps: the `SharedInterceptorRegistrations` stack (its key is now `BeanResolutionContext.SHARED_INTERCEPTOR_REGISTRATIONS`) and the `INTERCEPTOR_REGISTRATIONS` map. `copy()` shares attribute values, so without this, concurrent target creations pushed onto the stack of the outer bean's creation and could post-construct a target with the other thread's prototype interceptor.

What stays the same:
- Scoped targets: `getOrCreateScopedRegistration` already creates through `resolutionContext.copy()`, so the scope hot path gets no extra allocation.
- Destruction semantics: a per-call lazy target was never destroyed through the retained context, and still isn't. Cached lazy targets are still destroyed with the proxy.
- `shouldPreserveLazyProxyTargetResolutionPath`: `copy()` keeps the path, so a preserved path still applies.
- Already compiled proxies: the fix is at runtime, so they get it without being regenerated.

## Tests

`LazyProxyTargetDependentBeansSpec` (inject-java):
- a `@Prototype` lazy target called 100 times leaves the retained context's dependents empty (**fails on 5.2.x** with 100 registrations)
- a `@ThreadLocal` lazy target in a context without the scope bean: the test checks that the scope isn't found, then that nothing is retained (**fails on 5.2.x**)
- the same target in a context that includes all beans resolves one scoped instance
- a cached lazy prototype target is created once and destroyed with the proxy
- two threads create a target through a proxy injected into a bean with `@AroundConstruct` / `@PostConstruct` advice, and each target is post-constructed by the interceptor that constructed it (13 of 40 mismatches before the attribute fix)

Local runs with the fix, all green:
- `:micronaut-inject-java:test` for `io.micronaut.aop.*`, `io.micronaut.inject.lifecycle.*`, `io.micronaut.inject.injectionpoint.*`, `io.micronaut.inject.scope.*`, `io.micronaut.inject.factory.*` and `io.micronaut.runtime.*`: 3,802 tests, 0 failures, 1 skipped
- `:micronaut-inject:test`: 407 tests, 0 failures
- `:micronaut-context:test`: 48 tests, 0 failures, 1 skipped

### DIFF
--- a/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
+++ b/inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
@@ -3369,6 +3369,12 @@ public sealed class DefaultBeanContext implements ConfigurableBeanContext permit
                                                      @Nullable Qualifier<T> qualifier,
                                                      BeanDefinition<T> definition,
                                                      boolean dependent) {
+        if (resolutionContext instanceof AbstractBeanResolutionContext abstractContext && abstractContext.isLazyProxyTarget()) {
+            // The context is retained by a lazy proxy, which resolves its target through it on every call. Create
+            // the bean in a copy, so that the created bean is not recorded as a dependent of the retained context
+            // and the resolution path is not shared by concurrent calls.
+            resolutionContext = resolutionContext.copy();
+        }
         try (BeanResolutionContext context = newResolutionContext(definition, resolutionContext)) {
             final BeanResolutionContext.Path path = context.getPath();
             final boolean isNewPath = path.isEmpty();
--- a/aop/src/main/java/io/micronaut/aop/beandefinition/SharedInterceptorRegistrations.java
+++ b/aop/src/main/java/io/micronaut/aop/beandefinition/SharedInterceptorRegistrations.java
@@ -67,8 +67,6 @@ import java.util.List;
 @Internal
 public final class SharedInterceptorRegistrations {
 
-    private static final String ATTRIBUTE = "io.micronaut.aop.sharedInterceptorRegistrations";
-
     private SharedInterceptorRegistrations() {
     }
 
@@ -94,10 +92,10 @@ public final class SharedInterceptorRegistrations {
         if (registrations == null || registrations.isEmpty()) {
             return;
         }
-        Deque<Entry> stack = (Deque<Entry>) resolutionContext.getAttribute(ATTRIBUTE);
+        Deque<Entry> stack = (Deque<Entry>) resolutionContext.getAttribute(BeanResolutionContext.SHARED_INTERCEPTOR_REGISTRATIONS);
         if (stack == null) {
             stack = new ArrayDeque<>(3);
-            resolutionContext.setAttribute(ATTRIBUTE, stack);
+            resolutionContext.setAttribute(BeanResolutionContext.SHARED_INTERCEPTOR_REGISTRATIONS, stack);
         }
         stack.push(new Entry(definition, registrations));
     }
@@ -181,7 +179,7 @@ public final class SharedInterceptorRegistrations {
 
     @SuppressWarnings("unchecked")
     private static @Nullable Deque<Entry> stack(BeanResolutionContext resolutionContext) {
-        return (Deque<Entry>) resolutionContext.getAttribute(ATTRIBUTE);
+        return (Deque<Entry>) resolutionContext.getAttribute(BeanResolutionContext.SHARED_INTERCEPTOR_REGISTRATIONS);
     }
 
     private record Entry(BeanDefinition<?> definition, List<?> registrations) {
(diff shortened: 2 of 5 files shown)


inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
```diff
@@ -3369,6 +3369,12 @@ final <T> BeanRegistration<T> createRegistration(@Nullable BeanResolutionContext
                                                      @Nullable Qualifier<T> qualifier,
                                                      BeanDefinition<T> definition,
                                                      boolean dependent) {
+        if (resolutionContext instanceof AbstractBeanResolutionContext abstractContext && abstractContext.isLazyProxyTarget()) {
+            // The context is retained by a lazy proxy, which resolves its target through it on every call. Create
+            // the bean in a copy, so that the created bean is not recorded as a dependent of the retained context
+            // and the resolution path is not shared by concurrent calls.
+            resolutionContext = resolutionContext.copy();
```
graemerocher, line 3376: [P1] Isolate mutable resolution-context attributes as well. `copy()` only clones the attributes map with `putAll`, so values remain shared. If this lazy proxy is injected while outer constructor/lifecycle advice is active, the retained context contains the mutable `ArrayDeque` used by `SharedInterceptorRegistrations`; every per-call copy then pushes and pops that same deque. A coordinated two-thread repro makes target A receive target B’s prototype interceptor during `@PostConstruct` (and vice versa). Please detach/reset per-resolution mutable attributes for lazy-target calls and add a concurrent regression where the proxy is created during lifecycle-advised bean resolution.

graemerocher (commented):
Review complete. I found one blocking concurrency issue: the new per-call `copy()` isolates the path and dependent list, but shallow-copies mutable resolution-context attribute values. A focused two-thread regression, with the lazy proxy created during lifecycle-advised bean resolution, consistently cross-wires prototype constructor/post-construct interceptor registrations between the two target instances. I added the detailed inline comment with the reproduction shape and suggested fix direction.

Validation performed:
- The PR’s `LazyProxyTargetDependentBeansSpec` scenarios pass.
- Ordinary concurrent lazy-target resolution passes.
- The nested lifecycle-advice concurrency regression fails consistently on this head.

inject/src/main/java/io/micronaut/context/DefaultBeanContext.java
dstepanov: Fixed in 4cc29dd429: `copyForLazyProxyTarget` now removes the `SharedInterceptorRegistrations` stack (key moved to `BeanResolutionContext.SHARED_INTERCEPTOR_REGISTRATIONS`) and `INTERCEPTOR_REGISTRATIONS` from the retained copy, so each target creation starts its own; added a two-thread regression with the proxy injected into a `@AroundConstruct`/`@PostConstruct`-advised bean (13/40 mismatches before, 0 after, 3 reruns).

graemerocher (approved):

PR #13133, opened 2026-09-11
Register the physical signature of Kotlin suspend functions in the Graal reflection config

dstepanov:
## Problem (observed on v5.2.0)

When a Kotlin `suspend` method is annotated `@ReflectiveAccess`, directly or because a processor adds it, `GraalTypeElementVisitor` registers the method with its logical parameter list. That list leaves out the `kotlin.coroutines.Continuation` parameter, so the registered method doesn't exist on the JVM.

This showed up in micronaut-jakarta-interceptors' Kotlin test suite, with `micronaut-graal` on the KSP processor path, in the generated `$SuspendingService$ReflectConfig`:

```
REFLECT-CONFIG io.micronaut.interceptor.test.SuspendingService accessType=[]
REFLECT-CONFIG ...SuspendingService.suspending[java.lang.String] -> NO SUCH METHOD
ACTUAL JVM: public java.lang.Object ...SuspendingService.suspending(java.lang.String,kotlin.coroutines.Continuation<? super java.lang.String>)
```

Cause: `GraalTypeElementVisitor.processMethodElement` uses `MethodElement.getParameters()`. For Kotlin, `getParameters()` hides the continuation and `getSuspendParameters()` returns the physical list. `ExecutableMethodsDefinitionWriter` already uses `getSuspendParameters()` for the executable method. `GraalReflectionConfigurer.configure` catches the `NoSuchMethodException` for the wrong signature and silently registers nothing.

## Fix

`processMethodElement` uses `getSuspendParameters()` when `MethodElement.isSuspend()`. It is the only place in the graal module that turns a `MethodElement`'s parameters into reflection metadata; constructors go through it too but are never suspend functions.

## Native image: what actually fails

Checked with GraalVM CE 25.0.2 in `test-suite-kotlin-graalvm` (`nativeTest`), before and after the fix:

| Native test | Before | After |
|---|---|---|
| `ExecutableMethod.getTargetMethod()` + `Method.invoke` on a `@ReflectiveAccess` suspend method | passes | passes |
| same on a suspend method with only `@Executable` (control) | passes | passes |
| `getDeclaredMethod(name, *types)` with the name and argument types taken from the `ExecutableMethod` at runtime | **`NoSuchMethodException`** | passes |

So `getTargetMethod()` does **not** fail in a native image. The generated `$Exec.getTargetMethodByIndex` calls `ReflectionUtils.getRequiredMethod` with constant arguments (`SuspendingService.class`, `"suspending"`, `String.class`, `Continuation.class`). native-image apparently folds that lookup and registers the method on its own: it works even without `@ReflectiveAccess`. The broken metadata matters for reflective lookups whose arguments are only known at runtime; those fail on 5.2.0 and pass with this change.

## Tests

- `GraalTypeElementVisitorKotlinSpec` (graal module) compiles Kotlin through KSP with the graal visitor. Core had no Kotlin test for this visitor, so the graal module now has test dependencies on `micronaut-inject-kotlin-test` / the Kotlin compiler. The spec asserts that the generated config names `java.lang.String, kotlin.coroutines.Continuation` for `String`-returning and `Unit`-returning suspend functions, and `java.lang.String` only for a regular function. It also runs `GraalReflectionConfigurer.configure` the way `ServiceLoaderFeature` does and checks that the real methods get registered, including a private `@Executable @ReflectiveAccess` suspend function of a bean. Both features fail without the fix.
- `SuspendingTargetMethodTest` in `test-suite-kotlin-graalvm` covers the three native cases above and also runs on the JVM.

`./gradlew :micronaut-graal:check :test-suite-kotlin-graalvm:check :test-suite-kotlin-graalvm:nativeTest` passes locally. The full `check` was not run locally.

### DIFF
--- a/graal/build.gradle.kts
+++ b/graal/build.gradle.kts
@@ -3,17 +3,18 @@ plugins {
 }
 dependencies {
     annotationProcessor(projects.micronautInjectJava)
+
     api(projects.micronautCoreProcessor)
+
+    testAnnotationProcessor(projects.micronautInjectJava)
 
     testImplementation(projects.micronautInject)
     testImplementation(projects.micronautHttp)
     testImplementation(projects.micronautInjectJavaTest)
-    testImplementation(libs.managed.groovy.json)
-    testImplementation(libs.javax.persistence)
-    testAnnotationProcessor(projects.micronautInjectJava)
-
     testImplementation(projects.micronautInjectKotlin)
     testImplementation(projects.micronautInjectKotlinTest)
+    testImplementation(libs.managed.groovy.json)
+    testImplementation(libs.javax.persistence)
     testImplementation(libs.managed.kotlin.compiler.embeddable)
     testImplementation(libs.managed.kotlin.stdlib)
 }
--- a/graal/build.gradle.kts
+++ b/graal/build.gradle.kts
@@ -11,4 +11,19 @@ dependencies {
     testImplementation(libs.managed.groovy.json)
     testImplementation(libs.javax.persistence)
     testAnnotationProcessor(projects.micronautInjectJava)
+
+    testImplementation(projects.micronautInjectKotlin)
+    testImplementation(projects.micronautInjectKotlinTest)
+    testImplementation(libs.managed.kotlin.compiler.embeddable)
+    testImplementation(libs.managed.kotlin.stdlib)
+}
+
+configurations.all {
+    resolutionStrategy.eachDependency {
+        if (requested.group == "org.jetbrains.kotlin") {
+            useVersion(libs.versions.managed.kotlin.asProvider().get())
+        } else if (requested.group == "com.google.devtools.ksp") {
+            useVersion(libs.versions.managed.ksp.get())
+        }
+    }
 }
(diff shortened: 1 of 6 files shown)


graemerocher (commented):
Reviewed the production change, the KSP and Java/KAPT `MethodElement` contracts, and the added regression coverage. I found no correctness defects.

`getSuspendParameters()` is the correct API for the physical JVM signature when `isSuspend()` is true: it preserves the logical parameters and appends the synthetic `kotlin.coroutines.Continuation`, while the existing `getParameters()` path remains unchanged for ordinary methods. The tests cover value- and `Unit`-returning suspend functions, a private executable method, generated reflection registration, and the runtime reflective lookup that failed before this patch.

Local verification:
- `git diff --check HEAD^ HEAD`
- `./gradlew :micronaut-graal:check` (21 tests passed, including the 2 new Kotlin cases)
- Targeted `SuspendingTargetMethodTest` (3 tests passed)

Recommendation: merge as-is. I did not rerun `nativeTest`; the PR documents that native-image verification separately.

graemerocher (approved):

PR #13134, opened 2026-09-11
Run injection and @PostConstruct after the constructor interceptor chain

dstepanov:
Superseded by #13137, which carries the same change against 5.2.x. Kept only so the review thread here stays in context.

## Problem

For a bean with constructor interception (`InterceptorKind.AROUND_CONSTRUCT`), the terminal call of the `ConstructorInterceptorChain` goes into the generated `doInstantiate`. That method creates the instance and then injects its fields and methods and runs `@PostConstruct` before returning. `DefaultBeanContext` then treats the result as fully built. As a result, a construction interceptor's `proceed()` wraps construction, injection and initialization together.

This was reproduced on v5.2.0 from micronaut-jakarta-interceptors, a Jakarta Interceptors implementation built on Core's constructor interception. The setup:
- two around-construct interceptors, A (outer) and B (inner)
- a `@Prototype` target that records its constructor, a field `@Inject`, a setter `@Inject` and its own `@PostConstruct`

Recorded order:
```
[A before proceed, B before proceed, target constructor, target setter injection, field injected: true, target postConstruct, B after proceed, A after proceed]
```
The Jakarta Interceptors 2.2 specification (around-construct lifecycle) requires that injection of the target instance is not complete until all around-construct interceptor methods have completed successfully. The required order is:
```
[A before proceed, B before proceed, target constructor, B after proceed, A after proceed, target setter injection, target postConstruct]
```
Consequences today:
- `@PostConstruct` can observe state before an outer construction interceptor has finished.
- `@PostConstruct` runs even if an outer construction interceptor throws after `proceed()`.

## Change

The chain's terminal call now only creates the instance. Injection and post-construct then run on the instance the chain returned, and no API is added to do it:

- The generated definition overrides the entry point it already has: `instantiate(resolutionContext, context)` for a non-parametrized bean, and for a parametrized one the `doInstantiate(…, Map)` override that was already generated. It calls the default method of the intercepted interface, which runs the interceptor chain and returns as soon as the chain has, then runs the injection and post-construct code that used to follow construction inside `doInstantiate`. That code is unchanged, including the `shouldInitializeBean` check.
- The generated `doInstantiate(…, Object[])`, the terminal call of the chain, now only builds the instance — except that an `@InjectScope` constructor or factory argument is still released there, as soon as the constructor has run. That is the contract of the annotation, and it keeps the release on the path where an outer interceptor throws after `proceed()` and the instance is never injected.
- `SharedInterceptorRegistrations.peek` falls through to the entry a bean's own construction completed with when another bean's construction is on top of the stack. Post-construct interception now runs after that bean's entry has been popped, and without this a bean created while another bean is being constructed would stop sharing one interceptor instance between its construction and its post-construct phase.
- Exception handling at the chain boundary is unchanged. An exception from an interceptor is still carried by `ConstructorAdviceException`. A failure during injection now happens outside the chain and is wrapped by `DefaultBeanContext` as before.
- An interceptor still returns the instance, and whatever the chain returns is what gets injected and initialized. If an interceptor calls `proceed()` more than once, only the returned instance is injected and initialized. Previously every instance was.

Kotlin (KSP), Groovy and Python share `BeanDefinitionWriter`, so they get the same ordering. Factory-produced beans take the same path: the factory call is the "constructor" and has nothing of its own to inject. `@Introduction` beans and `@Around` proxies are handled, and the proxy's own `instantiate` default still runs, so interceptors keep seeing only the bean's own constructor arguments. For `@Around(proxyTarget = true)`, the target definition is the intercepted one and gets the new order. The proxy's own super-constructor call and member injection were never intercepted and are unchanged.

## Default, not opt-in

This changes the observable lifecycle order for existing `AROUND_CONSTRUCT` users, and I made it the default:

- **It matches what Core already documents.** Bean Life Cycle Advice describes three separate points: construction, `@PostConstruct` and `@PreDestroy`. The interceptor lifecycle list says "`AROUND_CONSTRUCT` interception runs, then the bean's constructor. `POST_CONSTRUCT` interception runs…". The exception section describes construct advice that throws after `proceed()` as "rejecting the instance the constructor produced". Nothing says injection or `@PostConstruct` happen inside `proceed()`. Today a rejected instance has already had `@PostConstruct` run.
- **An opt-in member would sit at the wrong level.** The ordering belongs to the bean's whole construction chain, not to one binding. A member on `@InterceptorBinding` or `@AroundConstruct` would make the order depend on which bindings a bean combines, with no good answer when they disagree. It would also leave the Jakarta/CDI integrations to set it everywhere.
- **Users who need the injected bean already have a hook.** Code that wants the injected and initialized bean has a dedicated kind, `POST_CONSTRUCT`, which gets the same interceptor instance. The docs example `ProductInterceptors` only registers the returned instance and is unaffected.
- **Compiled beans don't change underneath users.** A bean definition compiled by an earlier version overrides neither entry point, so it keeps injecting inside `doInstantiate`. Behaviour only changes when a bean is recompiled.

The change is documented under Breaking Changes, and the lifecycle advice guide gets a "What Construct Advice Wraps" section. If maintainers would rather ship it as opt-in for a patch release, the natural switch is a bean-level member read by `BeanDefinitionWriter` when deciding whether to generate the entry-point override.

## API

No new API. Nothing is added to `InterceptedBeanDefinition` or `ParameterizedInterceptedBeanDefinition`; the only runtime change in `aop` is inside `SharedInterceptorRegistrations`, which is `@Internal`.

## Tests

- `inject-java` `AroundConstructInjectionOrderSpec` (15 cases) records the order around `proceed()` with two `ConstructorInterceptor`s bound through an `@AroundConstruct` annotation, on a bean with field injection, method injection and `@PostConstruct`. Cases:
  - a plain bean
  - private injected members (reflective path)
  - an `@Around` proxy
  - `@Around(proxyTarget = true)`
  - `@Introduction`
  - parametrized beans with and without a proxy
  - an outer interceptor throwing after `proceed()`: no injection, no `@PostConstruct`
  - an interceptor proceeding several times: only the returned instance is initialized
  - shared `@Prototype` construct/post-construct interceptor
  - a factory-produced bean
  - an `@InjectScope` constructor argument, released when the constructor has run and also when an interceptor rejects the instance afterwards
  - a bean constructed inside another bean's construction, which still shares its interceptor with its post-construct phase
  - a proxied bean with members, whose interceptor sees only the bean's own constructor arguments

  Without the change, the cases that involve injection or `@PostConstruct` fail with the old order. The factory case has nothing to inject and passes either way.
- `inject-kotlin` and `inject-groovy` `AroundConstructInjectionOrderSpec`: the same order for a plain bean, an `@Around` proxy and an outer interceptor throwing.
- Local full run on 5.3.x, no failures, 7925 tests: `inject` 405, `context` 48, `inject-java` 5292, `inject-groovy` 1129, `inject-kotlin` 868, `inject-python` 125 (`-Ppython-ci`), AOP docs tests `test-suite` 47 / `test-suite-kotlin` 7 / `test-suite-groovy` 4, plus checkstyle for `aop` and `core-processor`. `aop` has no tests of its own. The AOP and `@InjectScope` specs were rerun on this 5.2.x branch.
- A private constructor could not be covered: a bean whose injected constructor is private fails with `IllegalAccessError` even without any interception, which is a separate issue.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
+++ b/core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
@@ -641,6 +641,7 @@ public final class BeanDefinitionWriter implements BeanElement, Toggleable, Elem
     private static final Method INSTANTIATE_METHOD = ReflectionUtils.getRequiredMethod(InstantiatableBeanDefinition.class, "instantiate", BeanResolutionContext.class, BeanContext.class);
     private static final Method DO_INSTANTIATE_INTERCEPTED_METHOD = ReflectionUtils.getRequiredMethod(io.micronaut.aop.beandefinition.InterceptedBeanDefinition.class, "doInstantiate", BeanResolutionContext.class, BeanContext.class, Object[].class);
     private static final Method INTERCEPTED_DEFAULT_INSTANTIATE_METHOD = ReflectionUtils.getRequiredMethod(io.micronaut.aop.beandefinition.InterceptedBeanDefinition.class, "instantiate", BeanResolutionContext.class, BeanContext.class);
+    private static final Method INJECT_AND_INITIALIZE_INTERCEPTED_METHOD = ReflectionUtils.getRequiredMethod(io.micronaut.aop.beandefinition.InterceptedBeanDefinition.class, "injectAndInitialize", BeanResolutionContext.class, BeanContext.class, Object.class);
     private static final Method RESOLVE_INSTANTIATION_VALUES_METHOD = ReflectionUtils.getRequiredMethod(io.micronaut.aop.beandefinition.InterceptedBeanDefinition.class, "resolveInstantiationValues", BeanResolutionContext.class, BeanContext.class);
     private static final Method RESOLVE_PARAMETRIZED_INSTANTIATION_VALUES_METHOD = ReflectionUtils.getRequiredMethod(ParameterizedInterceptedBeanDefinition.class, "resolveInstantiationValues", BeanResolutionContext.class, BeanContext.class, Map.class);
     private static final Method INTERCEPTED_PARAMETRIZED_DEFAULT_INSTANTIATE_METHOD = ReflectionUtils.getRequiredMethod(ParameterizedInterceptedBeanDefinition.class, "instantiate", BeanResolutionContext.class, BeanContext.class);
@@ -1639,9 +1640,16 @@ public final class BeanDefinitionWriter implements BeanElement, Toggleable, Elem
                         .<ExpressionDef>mapToObj(index -> constructorValuesArray.arrayElement(index).cast(TypeDef.erasure(parameterElements[index].getType())))
                         .toList();
                     ExpressionDef newInstance = buildNewInstance(aThis, methodParameters, statements, extractedValues);
-                    statements.add(injectAndReturn(aThis, methodParameters, newInstance));
+                    // The terminal call of the constructor interceptor chain only creates the instance: injection and
+                    // post-construct run in injectAndInitialize once every construction interceptor has returned
+                    statements.add(newInstance.returning());
                     return StatementDef.multi(statements);
                 }));
+            if (needsInjectOrInitialize()) {
+                classDefBuilder.addMethod(MethodDef.override(INJECT_AND_INITIALIZE_INTERCEPTED_METHOD)
+                    .build((aThis, methodParameters) ->
+                        injectAndReturn(aThis, methodParameters, methodParameters.get(2).cast(beanTypeDef))));
+            }
         } else {
             MethodDef.MethodDefBuilder buildMethodBuilder;
             if (isParametrized) {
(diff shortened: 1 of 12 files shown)


core-processor/src/main/java/io/micronaut/inject/writer/BeanDefinitionWriter.java
```diff
@@ -1639,9 +1640,16 @@ private void addInstantiateMethod() {
                         .<ExpressionDef>mapToObj(index -> constructorValuesArray.arrayElement(index).cast(TypeDef.erasure(parameterElements[index].getType())))
                         .toList();
                     ExpressionDef newInstance = buildNewInstance(aThis, methodParameters, statements, extractedValues);
-                    statements.add(injectAndReturn(aThis, methodParameters, newInstance));
+                    // The terminal call of the constructor interceptor chain only creates the instance: injection and
+                    // post-construct run in injectAndInitialize once every construction interceptor has returned
+                    statements.add(newInstance.returning());
```
graemerocher, line 1645: [P1] Preserve `@InjectScope` cleanup when advice rejects the instance

Moving all of `injectAndReturn` after the chain also moves `destroyInjectScopedBeans` for constructor/factory arguments. If an outer interceptor throws after `proceed()`, `injectAndInitialize` is never called, so a prototype `@InjectScope` argument remains alive in the global inject scope; the old path destroyed it inside `doInstantiate` before returning to the interceptor. I reproduced this by giving the new rejected-instance test an `@InjectScope` prototype constructor argument with a `@PreDestroy` callback: the callback never ran. This violates `@InjectScope`'s after-constructor cleanup contract and can leak or late-destroy resources. Please split constructor-argument scope cleanup from member injection/post-construct, or otherwise guarantee cleanup on this failure path.

dstepanov: Confirmed and fixed: the @InjectScope release for constructor/factory arguments now stays in the generated `doInstantiate`, immediately after the constructor runs, so it happens even when an outer interceptor throws after `proceed()`; two new tests cover the success and the rejected path (the rejected one is your reproduction, asserting the argument's @PreDestroy still runs).

PR #13135, opened 2026-09-11
Share interceptor instances between a proxy target and its proxy

dstepanov:
## Problem (observed on v5.2.0)

Since 5.2, `InterceptedBeanDefinition.instantiate` shares interceptor registrations across construct, post-construct and pre-destroy of one bean definition (`SharedInterceptorRegistrations`). A bean that uses a target proxy has two definitions, though:

- the **proxy wrapper**, generated by `AopProxyWriter`, which resolves the method (AROUND) interceptor registrations as a constructor argument;
- the **target definition**, selected in `BeanDefinitionWriter#isInterceptedLifeCycleByType`, which resolves its construct/post-construct/pre-destroy registrations separately.

A target proxy is used for `@Around(proxyTarget = true)`, and for every bean a `@Factory` method produces with AOP advice (`FactoryBeanElementCreator` forces target-proxy mode). The same prototype interceptor bean is therefore instantiated twice for one target instance: once for lifecycle, once for business methods.

Reproduced from micronaut-jakarta-interceptors, a Jakarta Interceptors implementation built on these registrations. A prototype interceptor sets a field in its post-construct interceptor method; its around-invoke method reads the field and records `System.identityHashCode(this)`:

- `@Around(proxyTarget = true)` bean: `[postConstruct on 848958019, micronaut around, around on 670689666 sees initial]`
- `@Factory`-produced bean: `[postConstruct on 260638966, around on 504478805 sees initial]`
- the same bean without a target proxy: one instance, and around-invoke sees the value set in post-construct.

The Jakarta Interceptors spec requires one interceptor instance per target instance for all its interceptor methods.

The Core reproduction (`ProxyTargetInterceptorSharingSpec`, first commit) also shows a second defect on factory-produced beans. The target's own interceptor instance was **destroyed right after post-construct** and kept serving calls: `[2:AROUND_CONSTRUCT, 2:POST_CONSTRUCT, 2:DESTROYED, 1:AROUND, 1:AROUND]`. The cause is `BeanResolutionContext#markDependentAsFactory()`, which takes the *first* dependent as the factory. An intercepted factory method resolves its arguments and interceptors before it looks the factory up, and a singleton factory is not a dependent at all. The first non-singleton interceptor or argument was then destroyed as "the factory".

## Fix

**The proxy hands its interceptors to the target it creates.** A non-lazy proxy (including `hotswap`) resolves its target in its constructor, after its interceptor registrations were resolved as a constructor argument. It now calls a new `BeanResolutionContext#getProxyTargetBean(definition, type, qualifier, interceptorRegistrations)`, which puts the registrations on the resolution context (`PROXY_INTERCEPTOR_REGISTRATIONS`) for the duration of the call.

**The target adopts the non-singleton ones.** When `DefaultBeanContext#createRegistration` creates that target, it moves each non-singleton registration from the proxy's dependents to the target's dependents.
- Only a registration actually found among the proxy's dependents is moved. That is the proof the proxy owns it, so nothing is destroyed twice.
- The adopted instances are destroyed with the target, after its `@PreDestroy`, exactly once.
- The hand-off is narrowed to the adopted instances for the target's creation, then restored.

**Lifecycle interception reuses them.** `SharedInterceptorRegistrations#resolve` passes the adopted instances to a new internal `BeanResolutionContext#getBeanRegistrations(type, qualifier, reusable)`. It resolves the target's set by binding as before, but takes an existing registration instead of creating a non-singleton candidate of the same definition. This covers:
- construction and post-construct (`InterceptedBeanDefinition`, `ParameterizedInterceptedBeanDefinition`);
- a target with post-construct advice but no constructor advice (`InitializableIntercepted` now resolves and stores the set when it is a proxy target);
- pre-destroy: when no set was recorded, `MethodInterceptorChain` resolves by binding while reusing the interceptors the bean owns, instead of using only those. Using only the owned instances would drop an interceptor bound to pre-destroy by a different annotation.

**Factory dependents.** Generated factory definitions call a new `markDependentAsFactory(Object factoryBean)`, which removes the dependent holding that factory instance, or nothing when the factory is a singleton.

**Nothing global.** The hand-off lives on the resolution context of one creation. Ownership is carried by the ordinary dependents of the target's registration. There is no weak or strong map, and no idle second interceptor instance on the proxy.

**Unchanged.**
- Singleton interceptors: never adopted, still shared.
- Non-proxied beans and `@Introduction` beans: no hand-off.
- Lazy proxies (`lazy`, `cacheableLazyTarget`, scoped proxies): resolve their targets later, possibly many, so they keep intercepting with the proxy's own instances.
- After a `hotswap` swap: the proxy keeps the instance it shared with its original target.
- A singleton target that already existed when the proxy was created: keeps its own instances.
- A target created through a copied resolution context, as a custom scope does: the proxy's instances are not among its dependents, so nothing is adopted.

**Binary compatibility.** The new `BeanResolutionContext` methods are defaults that fall back to the previous behaviour, so beans compiled by earlier versions behave as before. The docs TIP says to recompile. This changes generated code (proxy constructors and factory definitions), so it could equally go to 5.3.x if it should not ship in a patch release.

## Relation to #13127

#13127, also on 5.2.x, addresses the same split, starting from scoped proxies. It goes the other way: the target creates its own instances, which are recorded in a weak identity table, and the proxy swaps them in per call. This PR keeps no table, and the proxy is not left with an idle instance. It also fixes the factory-dependent destruction above, which #13127 does not: its factory-produced singleton case still destroys the instance after post-construct. It does not cover scoped/lazy proxies, which #13127 does. The two overlap in `DefaultBeanContext`, `BeanResolutionContext`, `AopProxyWriter` and the guide, so they need reconciling before either is merged.

## Tests

- New `ProxyTargetInterceptorSharingSpec` (committed first; the original cases fail without the fix):
  - one prototype `MethodInterceptor` + `ConstructorInterceptor` instance serves AROUND_CONSTRUCT, POST_CONSTRUCT, business methods and PRE_DESTROY of a `proxyTarget = true` singleton and of a `@Factory`-produced singleton, and is destroyed once, after the target;
  - each target of a `proxyTarget` prototype and of a factory-produced prototype gets its own instance;
  - a target with post-construct advice and no constructor advice; a pre-destroy interceptor bound by another annotation still applies; a singleton interceptor is still one instance;
  - lazy and hotswap proxies still intercept every call;
  - a prototype argument of an intercepted factory method is no longer destroyed as the factory.
- On this branch (5.2.x): `ProxyTargetInterceptorSharingSpec` 7/7 and `LifecycleInterceptorReuseSpec` 14/14 pass.
- The full `inject`, `context`, `inject-java`, `inject-groovy` and `inject-kotlin` test suites are still running locally (against the same change on a 5.3.x base); results will follow in a comment. Checkstyle on `aop`, `inject` and `core-processor` reports nothing new.

### DIFF
--- a/inject/src/main/java/io/micronaut/context/ProxyInterceptorRegistrations.java
+++ b/inject/src/main/java/io/micronaut/context/ProxyInterceptorRegistrations.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.context;
+
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.inject.BeanDefinition;
+
+import java.util.List;
+
+/**
+ * The interceptor registrations a proxy with {@code proxyTarget = true} hands to its target, the value of
+ * {@link BeanResolutionContext#PROXY_INTERCEPTOR_REGISTRATIONS}.
+ *
+ * <p>While the proxy resolves its target the registrations are the ones the proxy was constructed with. While the
+ * context creates that target they are narrowed to the non-singleton registrations the target adopted.</p>
+ *
+ * @param target        The definition of the proxy target
+ * @param registrations The interceptor registrations
+ * @author Denis Stepanov
+ * @since 5.2.1
+ */
+@Internal
+public record ProxyInterceptorRegistrations(BeanDefinition<?> target,
+                                            List<? extends BeanRegistration<?>> registrations) {
+
+    /**
+     * Whether the registrations are handed to the given definition.
+     *
+     * @param definition The definition being created
+     * @return {@code true} if the definition is the proxy target the registrations are handed to
+     */
+    public boolean isFor(BeanDefinition<?> definition) {
+        return target == definition || target.equals(definition);
+    }
+}
--- a/aop/src/main/java/io/micronaut/aop/beandefinition/InitializableIntercepted.java
+++ b/aop/src/main/java/io/micronaut/aop/beandefinition/InitializableIntercepted.java
@@ -43,6 +43,11 @@ public interface InitializableIntercepted<T> extends InitializingBeanDefinition<
         // every @PostConstruct callback of the bean, superclass callbacks first. An interceptor that does not proceed
         // keeps all of them from running. The callbacks themselves are listed by getPostConstructExecutableMethods().
         Collection<BeanRegistration<Interceptor<?, ?>>> shared = SharedInterceptorRegistrations.peek(resolutionContext, this);
+        if (shared == null) {
+            // A proxy target without constructor advice that its proxy is creating: resolve the set now, with the
+            // interceptor instances adopted from the proxy, so the proxy's method calls use the same instances.
+            shared = SharedInterceptorRegistrations.resolveForProxyTarget(resolutionContext, this);
+        }
         return Objects.requireNonNull(MethodInterceptorChain.initialize(
             resolutionContext,
             context,
--- a/aop/src/main/java/io/micronaut/aop/beandefinition/InterceptedBeanDefinition.java
+++ b/aop/src/main/java/io/micronaut/aop/beandefinition/InterceptedBeanDefinition.java
@@ -21,14 +21,10 @@ import io.micronaut.context.BeanRegistration;
 import io.micronaut.context.BeanContext;
 import io.micronaut.context.BeanResolutionContext;
 import io.micronaut.core.annotation.Internal;
-import io.micronaut.core.annotation.AnnotationMetadata;
 import io.micronaut.core.annotation.AnnotationMetadataProvider;
-import io.micronaut.core.annotation.AnnotationUtil;
 import io.micronaut.inject.InstantiatableBeanDefinition;
-import io.micronaut.inject.qualifiers.Qualifiers;
 import org.jspecify.annotations.Nullable;
 
-import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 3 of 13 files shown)


dstepanov:
Local test results for this change (run against a 5.3.x base with the same two commits; the spec and `LifecycleInterceptorReuseSpec` were also run on this 5.2.x branch, 21/21):

| Module | Tests | Skipped | Failures |
|---|---|---|---|
| inject | 405 | 0 | 0 |
| context | 48 | 1 | 0 |
| inject-java (includes the aop specs) | 5285 | 9 | 0 |
| inject-groovy | 1126 | 17 | 0 |
| inject-kotlin (incl. KSP) | 865 | 12 | 0 |

Checkstyle on `aop` and `core-processor` reports nothing for the changed files (`inject` checkstyle is skipped by the build).

inject/src/main/java/io/micronaut/context/ProxyInterceptorRegistrations.java
```diff
@@ -0,0 +1,48 @@
...
+                                            List<? extends BeanRegistration<?>> registrations) {
+
+    /**
+     * Whether the registrations are handed to the given definition.
+     *
+     * @param definition The definition being created
+     * @return {@code true} if the definition is the proxy target the registrations are handed to
+     */
+    public boolean isFor(BeanDefinition<?> definition) {
+        return target == definition || target.equals(definition);
```
graemerocher, line 46: [P1] Match delegated target definitions here. Iterable targets such as an `@EachBean` factory product are represented by `BeanDefinitionDelegate`, and the qualified target lookup can wrap that definition again before `createRegistration`. The two wrappers are not equal, so `isFor` returns false and adoption is skipped. A focused test with two `@EachBean` products creates four prototype interceptor instances (separate POST_CONSTRUCT and AROUND instances for each product) instead of two. Unwrapping both delegates before this comparison makes that regression pass. Please compare the canonical target definitions while preserving qualifier isolation, and add an iterable proxy-target regression.

graemerocher (commented):
Review complete. I found one blocking gap in the interceptor hand-off for iterable/delegated proxy targets. An `@EachBean` factory product is wrapped in `BeanDefinitionDelegate` during candidate creation and can be wrapped again for qualified target resolution; `ProxyInterceptorRegistrations.isFor` compares the wrappers directly, so adoption is skipped and lifecycle versus method interception still use separate prototype instances. I added the detailed inline comment and regression result.

Validation performed:
- `ProxyTargetInterceptorSharingSpec` plus `LifecycleInterceptorReuseSpec`: 21/21 passed.
- Focused runtime `proxyTarget` lifecycle-sharing regression: passed.
- Focused two-product `@EachBean` regression: failed with 4 interceptor instances instead of 2.
- With only the target comparison changed to unwrap delegated definitions, the `@EachBean` regression passed.

PR #13136, opened 2026-09-11
Instantiate beans with a private injected constructor reflectively

dstepanov:
## Problem

A bean with a private injected constructor fails at runtime:

```java
@Prototype class Service { @Inject private Service() {} }
```

```
BeanInstantiationException ... IllegalAccessError: class probe.plain.$Service$Definition
tried to access private method 'void probe.plain.Service.<init>()'
```

Adding `@ReflectiveAccess`, or an `@AroundConstruct` binding, gave the same error. Java, Kotlin and Groovy all behaved this way.

**Cause.** `ClassElement#getPrimaryConstructor` uses `getAccessibleConstructors()`, which drops private constructors. `DefaultElementBeanDefinitionBuilderFactory#ofType` then found no constructor and made up a public no-arg `<init>` with `requiresReflection = false`, so `BeanDefinitionWriter` called a private constructor directly. When the class also had a public constructor, the `@Inject` one was silently ignored and the public one used.

A second bug hid behind the first: `createConstructorDefinition(MethodElement, VisitorContext)` computed `requiresReflection` as `!isAccessible()`. `isAccessible()` returns true for a private member carrying `@ReflectiveAccess`, so the flag was inverted in exactly that case.

## Behaviour

A private constructor is used only when it is annotated with `@ReflectiveAccess`, and it is then invoked reflectively:

```java
@Prototype class Service { @Inject @ReflectiveAccess private Service() {} }
```

Requiring the annotation keeps reflection opt-in, in line with the `@NextMajorVersion` note on `MemberElement#isAccessible`, and leaves an unannotated private constructor (including the implicit one of a private nested type) resolving exactly as before.

`@Inject` or `@Creator` on a private constructor without `@ReflectiveAccess` is now a compilation error rather than a silent no-op:

```
Constructor is declared private and is not accessible for the instantiation.
To instantiate the bean using reflection annotate the constructor with @ReflectiveAccess
```

An *unannotated* private constructor is left alone, so the implicit constructor of a private nested type and a `@Factory` with only static methods still compile.

## Changes

- `BeanInjectionUtils#findBeanConstructor(ClassElement)` does the primary constructor lookup and additionally considers `@ReflectiveAccess` private constructors:
  - one also annotated `@Inject` or `@Creator` wins over an unannotated accessible constructor;
  - if a type's only constructor is private and annotated, it is used;
  - otherwise the result is `getPrimaryConstructor()`, as before.
- `createConstructorDefinition` uses `isReflectionRequired()`.
- `BeanDefinitionCreatorFactory#isDeclaredBean` uses the same lookup, so a class declared a bean only by a private `@Inject` constructor is discovered without a scope annotation.
- `AbstractBeanDefinitionBuilder#initConstructor` uses it too, so `@Import` and other associated beans no longer fail with *"Cannot create associated bean with no accessible primary constructor"*.
- `aroundProxy` fails compilation when a class bean's constructor is private, since the generated proxy subclass has to call it. Reported on the constructor: *"Cannot apply AOP advice to a bean created with a private constructor. The constructor must be made non-private to support proxying: &lt;type&gt;"*. Factory-produced types are unaffected.
- `GraalTypeElementVisitor` uses the lookup for direct and imported beans alike, so the reflection config matches what the definition does. Its private-constructor registration was previously unreachable.
- `BeanInjectionUtils#validateBeanConstructor` raises the error above from `BeanDefinitionCreatorFactory`, the single entry point for bean creation, so it is reported even for a class that would not otherwise be a bean.
- The constructor injection guide said such a type is instantiated reflectively; it now describes the `@ReflectiveAccess` opt-in and the AOP restriction.

`ClassElement#getPrimaryConstructor` is unchanged, so introspections, mixins and other callers behave as before.

## Tests

`PrivateInjectConstructorSpec` in inject-java, inject-kotlin and inject-groovy, covering: a private `@Inject` constructor with no arguments; an unscoped bean declared only by one; a private constructor with arguments alongside a public one; a type whose only constructor is private; `@AroundConstruct`; the missing-`@ReflectiveAccess` compile error; and the AOP compile error. `BeanImportSpec` covers `@Import` of a pre-compiled type with a private injected constructor, and `GraalTypeElementVisitorSpec` checks the constructor reaches the reflection config.

Local run of core-processor, graal, inject-java, inject-groovy and inject-kotlin: 7322 tests, 0 failures.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/processing/definition/DefaultElementBeanDefinitionBuilderFactory.java
+++ b/core-processor/src/main/java/io/micronaut/inject/processing/definition/DefaultElementBeanDefinitionBuilderFactory.java
@@ -74,7 +74,7 @@ public class DefaultElementBeanDefinitionBuilderFactory implements ElementBeanDe
 
     @Override
     public ElementBeanDefinitionBuilder<OutputObjectDef> ofType(ClassElement classElement) {
-        MethodElement constructorElement = classElement.getPrimaryConstructor().orElse(null);
+        MethodElement constructorElement = BeanInjectionUtils.findBeanConstructor(classElement).orElse(null);
         if (constructorElement != null) {
             return constructor(
                 BeanInjectionUtils.createConstructorDefinition(constructorElement, visitorContext)
@@ -162,6 +162,12 @@ public class DefaultElementBeanDefinitionBuilderFactory implements ElementBeanDe
         BeanDefinitionWriter targetBeanWriter = (BeanDefinitionWriter) targetBeanDefinitionBuilder;
         MemberDefinition<ClassElement> elementProducerDefinition = targetBeanWriter.getElementProducerDefinition();
         boolean isFactoryMethod = !(elementProducerDefinition instanceof ConstructorDefinition<ClassElement, ?>);
+        if (elementProducerDefinition instanceof ConstructorDefinition<ClassElement, ?> constructorDefinition
+            && constructorDefinition.constructorElement() instanceof MethodElement constructor
+            && constructor.isPrivate()) {
+            // The generated proxy extends the bean type, and its constructor has to invoke the bean constructor
+            throw new ProcessingException(constructor, "Cannot apply AOP advice to a bean created with a private constructor. The constructor must be made non-private to support proxying: " + targetType.getName());
+        }
 
         Map<CharSequence, Boolean> settings = new LinkedHashMap<>();
         OptionalValues<Boolean> aroundSettings = aopElementAnnotationProcessor.getValues(AnnotationUtil.ANN_AROUND, Boolean.class);
--- a/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
+++ b/graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
@@ -33,6 +33,7 @@ import io.micronaut.inject.ast.ElementQuery;
 import io.micronaut.inject.ast.FieldElement;
 import io.micronaut.inject.ast.MethodElement;
 import io.micronaut.inject.ast.ParameterElement;
+import io.micronaut.inject.utils.BeanInjectionUtils;
 import io.micronaut.inject.visitor.TypeElementQuery;
 import io.micronaut.inject.visitor.TypeElementVisitor;
 import io.micronaut.inject.visitor.VisitorContext;
@@ -179,10 +180,6 @@ public class GraalTypeElementVisitor implements TypeElementVisitor<Object, Objec
                     element,
                     false
                 );
-                MethodElement me = element.getPrimaryConstructor().orElse(null);
-                if (me != null && me.isPrivate() && !me.hasAnnotation(ReflectiveAccess.class)) {
-                    processMethodElement(me, reflectiveClasses);
-                }
             }
 
             if (element.isInner()) {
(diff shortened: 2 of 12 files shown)


core-processor/src/main/java/io/micronaut/inject/processing/definition/DefaultElementBeanDefinitionBuilderFactory.java
```diff
@@ -74,7 +74,7 @@ public DefaultElementBeanDefinitionBuilderFactory(VisitorContext visitorContext)
 
     @Override
     public ElementBeanDefinitionBuilder<OutputObjectDef> ofType(ClassElement classElement) {
-        MethodElement constructorElement = classElement.getPrimaryConstructor().orElse(null);
+        MethodElement constructorElement = BeanInjectionUtils.findBeanConstructor(classElement).orElse(null);
```
graemerocher, line 77: [P1] Use the same lookup during bean discovery

This factory is reached only after `BeanDefinitionCreatorFactory` decides the class is a bean. `isDeclaredBean` still checks `classElement.getPrimaryConstructor()` for `@Inject`, so an otherwise unscoped `class Service { @Inject private Service() {} }` produces no bean definition and `getBean(Service)` throws `NoSuchBeanException`; the new lookup never runs. A public `@Inject` constructor is sufficient to declare a bean today, so the private equivalent should use `findBeanConstructor` during discovery as well, with a regression case that omits `@Prototype`/`@Singleton`.

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
```diff
@@ -294,7 +291,9 @@ private void processBeanMethods(Map<String, ReflectionConfigData> reflectiveClas
     }
 
     private void processBeanConstructor(Map<String, ReflectionConfigData> reflectiveClasses, ClassElement beanElement, boolean isImport) {
-        final MethodElement constructor = beanElement.getPrimaryConstructor().orElse(null);
+        final MethodElement constructor = isImport
+            ? beanElement.getPrimaryConstructor().orElse(null)
```
graemerocher: [P1] Route imported beans through the private-constructor lookup

`@Import(classes = Service.class)` with a private `@Inject` constructor still fails compilation because `AbstractBeanDefinitionBuilder.initConstructor` calls `getPrimaryConstructor()` and reports `Cannot create associated bean with no accessible primary constructor`. This branch also keeps the old lookup for imports, so the associated-bean builder and Graal registration remain inconsistent with direct beans. Please route imported/associated beans through `findBeanConstructor` too (and add an import regression test), or explicitly narrow and document the feature as excluding imported beans.

graemerocher (commented):
I found two constructor-selection paths that still bypass the new helper and leave supported private-injection cases broken. The submitted direct-bean regression tests and the module/API checks otherwise pass. I recommend addressing both gaps before merge.

core-processor/src/main/java/io/micronaut/inject/processing/definition/DefaultElementBeanDefinitionBuilderFactory.java
dstepanov: Fixed in 1ca1847. `BeanDefinitionCreatorFactory.isDeclaredBean` now asks `findBeanConstructor` instead of `getPrimaryConstructor`, so a class declared a bean only by a private `@Inject` constructor is discovered; regression cases without `@Prototype`/`@Singleton` were added in the inject-java, inject-kotlin and inject-groovy specs.

graal/src/main/java/io/micronaut/graal/reflect/GraalTypeElementVisitor.java
dstepanov: Fixed in 1ca1847. `AbstractBeanDefinitionBuilder.initConstructor` uses `findBeanConstructor`, so imported and associated beans resolve the private constructor instead of failing with "no accessible primary constructor", and `GraalTypeElementVisitor` now uses the same lookup for imports as for direct beans. `BeanImportSpec` covers `@Import` of a pre-compiled type with a private injected constructor.

graemerocher (approved):

PR #13138, opened 2026-09-11
Bind the type arguments of super types two or more levels up through each level

dstepanov:
`ClassElement.getAllTypeArguments()` merged each super type's map with `putAll`. The arguments of a direct super type are written in the variables of the type itself. The arguments of the types above it are written in the variables of the super type in between, and were never bound through what the type gives that super type. `BeanIntrospectionWriter` and `BeanDefinitionWriter` both write this map, so for

```java
@Introspected
class Reversed<A, B> extends HashMap<B, A> {}
```

`getTypeArguments(HashMap.class)` answered `[B, A]`, but `getTypeArguments(Map.class)` answered `[K, V]`, which are HashMap's own variables. Names collide across levels (`class Swapped<K, V> extends HashMap<V, K>`), so a consumer cannot correct the answer by name. The reflective resolver (`ReflectionArguments`) already carries bindings through intermediate levels.

The Java, Groovy and Kotlin elements already resolve each level through its bindings: the resolved type and bounds are right. Only the placeholder keeps the variable name of the upper type, and that name is what the writers record.

## Change

- Each entry a direct super type reports for the types above it is bound through that super type's own type arguments. All variables are replaced at once, so a variable introduced by a binding is never replaced again.
- Variables nested in type arguments, wildcard bounds and arrays are bound too.
- Direct entries are unchanged.
- Java, Groovy and Kotlin all go through this default, so `BeanIntrospection.getTypeArguments(Class)` and `BeanDefinition.getTypeArguments(Class)` are fixed for all three.

## Tests

New cases in `IntrospectionTypeArgumentsSpec` for inject-java, inject-groovy and inject-kotlin. Each checks both the introspection and the bean definition:

- a type two levels up with swapped variables: `Reversed<A, B> extends HashMap<B, A>` read as `Map` gives `[B, A]`
- a name collision: `Swapped<K, V> extends HashMap<V, K>` read as `Map` gives `[V, K]`
- an interface reached through an intermediate interface: `Holder<X, Y> implements Flipped<X, Y>`, `Flipped<F, E> extends Container<E, F>` gives `[Y, X]`
- a concrete binding one level up: `Strings extends ArrayList<String>` read as `Iterable` gives `[String]`, not a type variable

Without the fix, 3 Java, 3 Groovy and 5 Kotlin cases fail. Kotlin also reported `Strings` read as `Iterable` as a type variable.

In Kotlin, an `@Introspected` class extending `java.util.HashMap` or `ArrayList` generates an introspection that does not load (`ClassFormatError`, duplicate `$property$keys$metadata`). That already happens on 5.2.x. So the Kotlin JDK-collection cases are read through the bean definition, and the introspection cases use Kotlin-declared hierarchies.

Local runs with the change, all without failures:

| Module | Tests |
|---|---|
| core-processor | 6 |
| inject-java | 5284 |
| inject-groovy | 1131 |
| inject-kotlin | 871 |
| inject | 407 |
| test-suite | 493 |
| test-suite-groovy | 235 |
| test-suite-kotlin | 374 |

`inject-python` tests were skipped by the build. `PythonClassElement` overrides `getAllTypeArguments()` separately and is not touched here.


core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -964,4 +976,50 @@ static ClassElement of(String typeName, boolean isInterface,
...
+            ClassElement binding = bindings.get(placeholder.getVariableName());
+            if (binding == null) {
+                return type;
+            }
+            // A variable resolved through a binding already counts the dimensions of the binding
+            ClassElement bound = binding;
+            while (bound.getArrayDimensions() < placeholder.getArrayDimensions()) {
+                bound = bound.toArray();
+            }
+            return bound;
```
graemerocher, line 1015: [P1] Preserve type-use metadata from the placeholder occurrence

Returning the binding wholesale drops annotation metadata attached where the placeholder is used. I reproduced this with `AnnotatedMiddle<T> implements Container<@Marker T, T>` and `AnnotatedLeaf<X> extends AnnotatedMiddle<X>`: the direct `AnnotatedMiddle` introspection retains `@Marker`, but `AnnotatedLeaf.getTypeArguments(Container)[0]` has empty annotation metadata. This can strip validation, nullness, or other custom `TYPE_USE` metadata from the generated argument. Please merge/preserve the occurrence metadata when substituting and cover this transitive case.

core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
```diff
@@ -964,4 +976,50 @@ static ClassElement of(String typeName, boolean isInterface,
...
+        }
+        if (type instanceof WildcardElement) {
+            ClassElement folded = type.foldBoundGenericTypes(bound -> bound instanceof GenericPlaceholderElement ? bindTypeVariables(bound, bindings) : bound);
+            return folded == null ? type : folded;
+        }
+        Map<String, ClassElement> typeArguments = type.getTypeArguments();
+        Map<String, ClassElement> boundTypeArguments = bindTypeVariables(typeArguments, bindings);
+        return boundTypeArguments == typeArguments ? type : type.withTypeArguments(boundTypeArguments);
```
graemerocher, line 1023: [P1] Do not require every ClassElement to support copying

`withTypeArguments` explicitly permits implementations to throw `UnsupportedOperationException`, and the repository's own `SimpleClassElement` uses that default. I reproduced a simple `Leaf -> Middle<T=String> -> Ancestor<A=List<T>>` hierarchy built with the existing `ClassElement.of(String, ..., superType, interfaces)` factory: `leaf.getTypeArguments("Ancestor")` now throws here instead of returning the hierarchy map. This also risks breaking third-party `ClassElement` implementations that legitimately do not support copies. Please make the binding path compatible with non-copyable elements (or provide a supported reconstruction for `SimpleClassElement`) and add this case.

core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
dstepanov: Fixed in 27e94c4127. The bound type is now read through the use that annotates it: `TypeAnnotatedClassElement` / `TypeAnnotatedGenericPlaceholderElement` delegate to what the variable is bound to and merge the use-site type annotations (also on `getResolved()`, so the annotations survive the resolution the writers do). Annotating a copy of the binding was not an option - the metadata is cached per native element, so it leaked into the direct entry. Tests in inject-java `IntrospectionTypeArgumentsSpec`: your `AnnotatedMiddle<T> implements Container2<@Marker T, T>` / `AnnotatedLeaf<X>` case at element level (argument is `X` and keeps `@Marker`) and through the introspection and the bean definition, plus the concrete binding `AnnotatedStrings implements AnnotatedMiddle<String>` (`String` keeps `@Marker`). All three fail without the change. Note the written argument records no variable name for an annotated argument at all, in the direct case too (`buildArgumentWithGenerics` only passes it on the bounds branch) - that is untouched here, so the element-level test pins the variable.

core-processor/src/main/java/io/micronaut/inject/ast/ClassElement.java
dstepanov: Fixed in 27e94c4127. `SimpleClassElement` now implements `withTypeArguments(Map)`, so your `Leaf -> Middle<T=String> -> Ancestor<A=List<T>>` hierarchy answers `A=List<String>`; and the binding path no longer requires copying at all - `withTypeArguments` and `toArray` are called through a helper that answers the type unchanged when an implementation throws `UnsupportedOperationException`, leaving the variable unbound, which is what the type could say before. Covered by the new core-processor `AllTypeArgumentsSpec`: the `ClassElement.of(String, ..., superType, interfaces)` hierarchy, a variable bound directly (no copy needed), and a non-copyable type reference built with `ClassElement.of(Class, metadata, typeArguments)` that is reported unbound rather than throwing.

core-processor/src/main/java/io/micronaut/inject/ast/TypeAnnotatedClassElement.java
```diff
@@ -0,0 +1,469 @@
...
+ * <p>{@link ClassElement#getAllTypeArguments()} builds one when it binds the type argument of a super type it
+ * reaches through another type, so that what a type variable is bound to and the annotations written where that
+ * variable is used are both kept.</p>
+ *
+ * @author Denis Stepanov
+ * @since 5.2.1
+ */
+@Internal
+@Experimental
+public class TypeAnnotatedClassElement implements ClassElement {
```
graemerocher: having a delegate is error prone since if a new default method is introduced in the parent you have to remember to implement the delegate.

Is there a way to keep ClassElement and just use `withAnnotationMetadata`?

dstepanov: Agreed, and the delegate is gone in 0cfc42ba1c (so is the copy of it inject-python had to extend, which also clears the Sonar critical about the shadowed class name). `withAnnotationMetadata` on its own does not work: it presets the annotation metadata of the element, while the writers persist `getTypeAnnotationMetadata()`, which no implementation reads the preset for - I tried it and the generated argument still lost `@Marker`. What it turned out to buy was nothing the generated metadata keeps: the writers record no variable name for an annotated type argument at all (`buildArgumentWithGenerics` only passes one on the bounds branch, in the direct case too), and they resolve a variable before reading its annotations, so a variable bound to a concrete type never carried the annotations of the use on 5.2.x either. An annotated use is therefore answered with the element of the use itself, which already stands for what the variable is bound to and is the only element carrying those annotations; it keeps the variable name the intermediate type writes it with. Everything unannotated binds as before, Kotlin included, where the use carries no resolution of its own. Tests updated to state exactly that.

graemerocher:
are there performance implications to this change? Should we consider caching for this logic?

dstepanov:
Measured it rather than guessing.

**Per call.** Old vs new algorithm, interleaved in the same JVM so machine noise hits both equally — fresh elements every iteration for Java, which caches the result per element, and the same element for Groovy, which doesn't. Two rounds, consistent within a few percent:

| Hierarchy | Java old → new | Groovy old → new |
|---|---|---|
| `extends ArrayList<String>` | 4.30 → 5.62 µs (1.31×) | 4.07 → 6.08 µs (1.50×) |
| `<A, B> extends HashMap<B, A>` | 1.80 → 2.13 µs (1.18×) | 1.24 → 1.66 µs (1.34×) |
| `extends ConcurrentHashMap<String, List<Integer>>` | 2.22 → 2.79 µs (1.26×) | 3.31 → 4.02 µs (1.21×) |
| 8-level chain, a swap and a nested type at every level | 4.42 → 9.56 µs (2.16×) | 3.49 → 8.54 µs (2.45×) |
| `implements Runnable` | 0.31 → 0.33 µs (1.09×) | 0.16 → 0.21 µs (1.26×) |

So the binding does cost something: roughly 10–50% on realistic hierarchies, up to about 2.5× on a deliberately adversarial one, all in single-digit microseconds.

**In a real compile.** Temporary instrumentation (not part of this PR) summing the time spent in top-level `getAllTypeArguments()` calls with the new code:

| Compile | Top-level calls | Time in `getAllTypeArguments()` | Context |
|---|---|---|---|
| `:test-suite-kotlin:kspTestKotlin` | 12,762 | 28.0 ms | task 12.07 s, so 0.23% |
| `:test-suite:compileTestJava` (plus dependencies that recompiled) | 358 | 13.0 ms | build 50 s |
| `:test-suite-groovy:compileTestGroovy` | 51 | 1.6 ms | build 15 s |

That is the whole cost of the method. The part this PR adds is at most about a third of it at the realistic ratios above, so on the order of 10 ms for the Kotlin suite.

**Caching.** I don't think it's warranted for this change. Java already caches the result per element (`JavaClassElement.resolvedAllTypeArguments`). The figure that stands out is Kotlin's 12,762 calls: neither `KotlinClassElement` nor `GroovyClassElement` caches it, and the default `getTypeArguments(String)` rebuilds the whole hierarchy map on every call. That predates this PR. A per-instance cache in those two elements, mirroring the Java one, would remove most of those recomputations — happy to do that as a separate change if you want it.

PR #13140, opened 2026-09-11
Make the HTTP/2 connection receive window configurable

graemerocher:
## Summary
- `SETTINGS_INITIAL_WINDOW_SIZE` (`micronaut.server.netty.http2.initial-window-size`) only sizes the receive window of each stream. The window of the connection as a whole (stream 0) always starts at the protocol default of 65535 bytes, and netty's `Http2ConnectionHandler` never raises it. With the default server pipeline the combined upload throughput of all streams on one connection is therefore limited to about 64 KiB per round trip, whatever `initial-window-size` is set to. Netty's `Http2FrameCodec`, which the legacy multiplex pipeline uses, raises the connection window by twice the stream window delta on its own.
- Adds `micronaut.server.netty.http2.initial-connection-window-size`. Both pipelines now follow the `Http2FrameCodec` rule by default (connection window = 65535 + 2 × (stream window − 65535)), and the new property raises the connection window further when it is larger than that derived value. The derived value is a floor in both pipelines: the codec applies its rule before we get a chance to, and a window cannot be lowered once advertised, so the same rule is used for the `Http2ServerHandler` to keep the two pipelines consistent.
- `Http2ServerHandler.ConnectionHandler` sends the `WINDOW_UPDATE` on stream 0 right after the `SETTINGS` preface (in `handlerAdded`, or `channelActive` if the channel was not active yet). For the legacy `Http2FrameCodec` a small `Http2ConnectionWindow` handler is added after the codec in the three places it is installed (ALPN, h2c upgrade, h2c prior knowledge); it raises the window and removes itself.
- Increases below double the current window would only be advertised by netty's flow controller with the first refill (once the peer has used half the old window). Those are sent right away as well, by briefly moving the connection stream's update threshold; this also covers the `Http2FrameCodec`'s own increase for stream windows slightly above 65535.
- Values below 65535 are rejected by the setter; `@since 5.2.2`: 5.2.x was at `5.2.1-SNAPSHOT` when this was written and v5.2.1 was released in the meantime, so the property debuts in 5.2.2.

## Release note

On the default pipeline, a configured `micronaut.server.netty.http2.initial-window-size` now also raises the HTTP/2 *connection* receive window, to `65535 + 2 × (W − 65535)` (matching Netty's `Http2FrameCodec`), where it used to stay at the 64 KiB default. With a 16 MiB stream window that is about 32 MiB per connection, and since the connection window is only refilled as request bodies are consumed, a slow consumer can hold that much unread upload data per connection. The derived value is a floor: `initial-connection-window-size` can raise it further but setting it to 65535 does not restore the previous 64 KiB.

## Testing
- New `Http2ConnectionWindowSpec` (EmbeddedChannel, raw frame recorder on the client): the `effectiveWindowSize` table; no `WINDOW_UPDATE` without configuration; `SETTINGS` followed by `WINDOW_UPDATE` on stream 0 for the stream-window default, for an explicit window, for a small explicit window (100000), for a small stream window (70000), and for the explicit-vs-stream-window precedence; the legacy handler removing itself. Each pipeline case runs for both the `Http2ServerHandler` and the `Http2FrameCodec`.
- New `Http2ConnectionWindowServerSpec`: a raw netty h2c prior-knowledge client against a real `EmbeddedServer` checks the connection window it was granted for six property combinations, with and without `legacy-multiplex-handlers`.
- `NettyHttpServerConfigurationSpec`: property binding, unset default, and rejection of a value below 65535.
- Fail-before: with the window raise disabled, `Http2ConnectionWindowSpec` fails 5 of 19 cases (`explicit connection window is sent after the preface` and `explicit connection window wins over the stream window` for both pipelines, `connection window follows the stream window by default` for the `Http2ServerHandler`; the legacy variant of the latter passes on the base because the codec already does this). With only the threshold forcing disabled, `small explicit connection window is sent after the preface` fails for both pipelines.
- `./gradlew :micronaut-http-server-netty:test`: 1197 tests, 22 skipped (pre-existing), 0 failures on the final run. An earlier run under a 1-minute load average of 6 had two unrelated failures (`RequestDecompressionConfigSpec` connect error, `HttpResponseSpec "test date header turned off"`), both green when re-run alone.
- `./gradlew :test-suite-http2-server-tck-netty:test`: 263 tests, 3 skipped, 0 failures.
- `checkstyleMain` clean apart from the pre-existing `FileLength` violation in `NettyHttpServerConfiguration`; `japiCmp` clean.

## Review follow-ups

- @dstepanov: `Http2ConnectionWindowServerSpec` now covers the h2c upgrade wiring as well as prior knowledge (`1a60b0e568`); the release-note section above is from the same review.
- Sonar S107: the `ConnectionHandler` constructor reads the frame listener, access log manager and window from the builder (`59db24fb62`).

## Out of scope
- A throughput measurement of large uploads over a single HTTP/2 connection; this PR is about correctness of the advertised window.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpPipelineBuilder.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpPipelineBuilder.java
@@ -22,6 +22,7 @@ import io.micronaut.http.body.stream.BodySizeLimits;
 import io.micronaut.http.netty.SslContextHolder;
 import io.micronaut.http.netty.channel.ChannelPipelineCustomizer;
 import io.micronaut.http.server.netty.configuration.NettyHttpServerConfiguration;
+import io.micronaut.http.server.netty.handler.Http2ConnectionWindow;
 import io.micronaut.http.server.netty.handler.Http2ServerHandler;
 import io.micronaut.http.server.netty.handler.PipeliningServerHandler;
 import io.micronaut.http.server.netty.handler.RequestHandler;
@@ -453,6 +454,7 @@ final class HttpPipelineBuilder {
                 Http2FrameCodec http2FrameCodec = createHttp2FrameCodec();
                 pipeline.addLast(ChannelPipelineCustomizer.HANDLER_HTTP2_CONNECTION, http2FrameCodec);
                 specificGracefulShutdown = new Http2GracefulShutdown(pipeline.lastContext(), http2FrameCodec);
+                pipeline.addLast(createHttp2ConnectionWindow(http2FrameCodec));
                 pipeline.addLast(makeHttp2Handler());
             } else {
                 Http2ConnectionHandler http2ServerHandler = createHttp2ServerHandler(true);
@@ -477,14 +479,25 @@ final class HttpPipelineBuilder {
             return builder.build();
         }
 
+        /**
+         * The {@link Http2FrameCodec} raises the connection window only from the stream window.
+         * This handler applies an explicitly configured connection window on top of that.
+         */
+        private Http2ConnectionWindow createHttp2ConnectionWindow(Http2FrameCodec http2FrameCodec) {
+            NettyHttpServerConfiguration.Http2Settings http2 = server.getServerConfiguration().getHttp2();
+            return new Http2ConnectionWindow(http2FrameCodec, Http2ConnectionWindow.effectiveWindowSize(http2.http2Settings(), http2.getInitialConnectionWindowSize()));
+        }
+
         private Http2ConnectionHandler createHttp2ServerHandler(boolean ssl) {
+            NettyHttpServerConfiguration.Http2Settings http2 = server.getServerConfiguration().getHttp2();
             Http2ServerHandler.ConnectionHandlerBuilder builder = new Http2ServerHandler.ConnectionHandlerBuilder(makeRequestHandler(embeddedServices.getWebSocketUpgradeHandler(server), ssl))
                 .decompress(server.getServerConfiguration().isRequestDecompressionEnabled())
                 .compressor(embeddedServices.getHttpCompressionStrategy())
                 .bodySizeLimits(bodySizeLimits())
                 .accessLogManagerFactory(accessLogManagerFactory)
                 .validateHeaders(server.getServerConfiguration().isValidateHeaders())
-                .initialSettings(server.getServerConfiguration().getHttp2().http2Settings());
+                .initialSettings(http2.http2Settings())
+                .initialConnectionWindowSize(http2.getInitialConnectionWindowSize());
             server.getServerConfiguration().getLogLevel().ifPresent(logLevel ->
                 builder.frameLogger(new Http2FrameLogger(logLevel, NettyHttpServer.class)));
             return builder.build();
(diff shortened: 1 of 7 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ConnectionWindow.java
```diff
@@ -0,0 +1,138 @@
...
+     * raises the connection window by twice the amount the stream window exceeds the protocol
+     * default, so that a single stream cannot use up the whole connection window. The codec
+     * applies that rule on its own, so a configured value cannot go below it in the legacy
+     * pipeline; using the same floor here keeps both pipelines consistent.
+     *
+     * @param initialSettings The local settings sent in the preface
+     * @param configured      The explicitly configured connection window size, if any
+     * @return The connection window size, at least {@link Http2CodecUtil#DEFAULT_WINDOW_SIZE}
+     */
+    public static int effectiveWindowSize(Http2Settings initialSettings, @Nullable Integer configured) {
```
dstepanov, line 71: Not blocking, but worth a release note: on the default pipeline, an existing `initial-window-size` now also grows the connection window to 65535 + 2 × (W − 65535). With 16 MiB that is ~32 MiB instead of 64 KiB, and since the connection window is only refilled as bodies are consumed, a slow consumer can hold that much unread upload per connection. The derived value is a floor, so setting `initial-connection-window-size: 65535` doesn't bring the old limit back. Matching `Http2FrameCodec` makes sense to me, it's just a visible change for a patch release.

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpPipelineBuilder.java
```diff
@@ -589,7 +602,7 @@ public void upgradeTo(ChannelHandlerContext ctx, FullHttpRequest upgradeRequest)
                     if (frameCodec == null || multiplexHandler == null) {
                         return new Http2ServerUpgradeCodecImpl(connectionHandler);
                     } else {
-                        return new Http2ServerUpgradeCodecImpl(frameCodec, multiplexHandler);
+                        return new Http2ServerUpgradeCodecImpl(frameCodec, createHttp2ConnectionWindow(frameCodec), multiplexHandler);
```
dstepanov, line 605: The h2c upgrade wiring isn't covered: `Http2ConnectionWindowServerSpec` only uses prior knowledge. I tried a raw netty `HttpClientUpgradeHandler` + `Http2ClientUpgradeCodec` client against both pipelines (unset, 4 MiB, 100000, stream window 1 MiB) and it works, so maybe add an upgrade row to the spec to keep it that way.

dstepanov (approved):
LGTM. I checked the preface/WINDOW_UPDATE ordering against netty 4.2.17 for the active, channelActive and h2c upgrade paths, and the new specs pass locally. Two small notes inline, neither blocking.

http-server-netty/src/main/java/io/micronaut/http/server/netty/HttpPipelineBuilder.java
graemerocher: Added in 1a60b0e568: every row of `Http2ConnectionWindowServerSpec` now also runs through an HTTP/1.1 upgrade (`HttpClientUpgradeHandler` + `Http2ClientUpgradeCodec`; note that in Netty 4.2 the upgrade codec no longer adds the frame codec itself, so the codec is installed through the `upgradeToHandler`). 24/24 pass with paranoid leak detection.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Http2ConnectionWindow.java
graemerocher: Agreed; added a "Release note" section to the PR description with exactly this (an existing `initial-window-size` now also raises the connection window to 65535 + 2 × (W − 65535), a slow consumer can hold that much unread upload per connection, and the derived value is a floor so `initial-connection-window-size: 65535` does not restore the old limit).

PR #13142, opened 2026-09-11
Record per pipeline whether the access log handler is installed

graemerocher:
## Summary
- With the access logger enabled, `RoutingInBoundHandler.prepareRequest` called `ctx.pipeline().get(HANDLER_ACCESS_LOGGER)` for every request to decide whether to store the request in the `NettyHttpRequest` channel attribute that the access log (and Micronaut Session's `SessionLogElement`) reads. The answer is a property of the pipeline, not the request: HTTP/1 pipelines and the per-stream pipelines of the legacy HTTP/2 handlers carry the handler; the connection pipeline of the default HTTP/2 handler logs through `Http2AccessLogManager` and does not (a single channel attribute could not tell concurrent streams apart, so that is left as is).
- `HttpPipelineBuilder.StreamPipeline`, which is already attached to every channel the routing handler sees, now remembers whether its pipeline carries the handler. The pipeline is inspected once, on the first request, rather than at install time, so that a `NettyServerCustomizer` or `ChannelPipelineListener` that adds or removes the handler while the pipeline is built is still respected (Codex review finding, see below). Per request this is one channel-attribute read and a boolean instead of a pipeline walk. The `supportLoggingHandler` shortcut for servers without an access logger is unchanged, and a context without a `StreamPipeline` (a pipeline not built by `HttpPipelineBuilder`) falls back to the live lookup.
- Follows up on the "out of scope" note in #13118; it touches the same `if` in `prepareRequest`, so whichever merges second needs a trivial conflict resolution.

## Testing
- New `AccessLogPipelineFlagSpec`: a controller returns whether the channel attribute holds the request being routed. HTTP/1 with the logger enabled: set; logger disabled: not set; default HTTP/2 handler (h2c prior knowledge): not set, as before; legacy HTTP/2 handlers: set; a customizer that removes the handler in `onStreamPipelineBuilt`: not set.
- Fail-before: the change is meant to preserve behaviour, so the first four cases guard equivalence with the base rather than failing on it. The customizer case fails with an install-time flag (`has = accessLogHandler != null`: `HTTP_1_1 true` instead of `HTTP_1_1 false`), which is the Codex finding the first-request inspection addresses.
- `./gradlew :micronaut-http-server-netty:test`: 1163 tests, 22 skipped (pre-existing), 0 failures (1-minute load average 10 from other work on the machine, still green).
- `checkstyleMain` clean apart from the pre-existing `FileLength` violation in `NettyHttpServerConfiguration`.
- Codex review (`gpt-5.6-sol`, high): first pass flagged that an install-time flag diverges from the pipeline when a customizer removes or adds the handler; addressed by inspecting the pipeline on the first request. Second pass: no findings.

## Out of scope
- Setting the attribute for the default HTTP/2 handler (it never was; the connection channel is shared by concurrent streams).
- The `AttributeKey.valueOf` per request and clearing the attribute after the response, both handled in #13118.
- No benchmark: the path is only taken with the access logger enabled, which the harness on 5.2.x cannot switch on (#13118 adds that switch); the change replaces a ~10-node pipeline walk with an attribute read and is kept on its structural merits.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/accesslog/AccessLogPipelineFlagSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/accesslog/AccessLogPipelineFlagSpec.groovy
@@ -1,0 +1,105 @@
+package io.micronaut.http.server.netty.handler.accesslog
+
+import io.micronaut.context.ApplicationContext
+import io.micronaut.context.annotation.Requires
+import io.micronaut.context.event.BeanCreatedEvent
+import io.micronaut.context.event.BeanCreatedEventListener
+import io.micronaut.http.HttpRequest
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+import io.micronaut.http.client.HttpClient
+import io.micronaut.http.netty.channel.ChannelPipelineCustomizer
+import io.micronaut.http.server.netty.NettyHttpRequest
+import io.micronaut.http.server.netty.NettyServerCustomizer
+import io.micronaut.runtime.server.EmbeddedServer
+import io.netty.channel.Channel
+import io.netty.util.AttributeKey
+import jakarta.inject.Singleton
+import spock.lang.Specification
+
+/**
+ * The access log handler needs the current request in a channel attribute (Micronaut Session
+ * reads it from there). Whether a pipeline carries that handler is recorded when the pipeline is
+ * built, so this checks that the attribute is set exactly on the pipelines that have the handler.
+ */
+class AccessLogPipelineFlagSpec extends Specification {
+    private static final AttributeKey<NettyHttpRequest> KEY = AttributeKey.valueOf(NettyHttpRequest.class.getSimpleName())
+
+    def "request attribute is set only on pipelines with the access log handler"(Map<String, Object> properties, String expectedVersion, boolean expectAttribute) {
+        given:
+        def ctx = ApplicationContext.run([
+                'spec.name'                                       : 'AccessLogPipelineFlagSpec',
+                'micronaut.server.netty.access-logger.logger-name': 'http-access-log',
+        ] + properties)
+        def server = ctx.getBean(EmbeddedServer).start()
+        def client = ctx.createBean(HttpClient, server.URI)
+
+        when:
+        def response = client.toBlocking().retrieve('/access-log-flag')
+
+        then:
+        response == "$expectedVersion $expectAttribute"
+
+        cleanup:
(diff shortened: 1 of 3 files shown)


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/accesslog/AccessLogPipelineFlagSpec.groovy
```diff
@@ -0,0 +1,105 @@
...
+import io.micronaut.runtime.server.EmbeddedServer
+import io.netty.channel.Channel
+import io.netty.util.AttributeKey
+import jakarta.inject.Singleton
+import spock.lang.Specification
+
+/**
+ * The access log handler needs the current request in a channel attribute (Micronaut Session
+ * reads it from there). Whether a pipeline carries that handler is recorded when the pipeline is
+ * built, so this checks that the attribute is set exactly on the pipelines that have the handler.
```
dstepanov: Nit: this no longer matches the implementation. The flag is now worked out from the pipeline on the first request, not when the pipeline is built (the last case exists precisely because of that). Maybe: "Whether a pipeline carries that handler is remembered per pipeline, so this checks..."

dstepanov (approved):
Looks good. I traced every place `STREAM_PIPELINE_ATTRIBUTE` is set: HTTP/1 and h2c fallback (`insertHttp1DownstreamHandlers`), legacy multiplex and HTTP/3 streams (`insertHttp2DownstreamHandlers`), and the default HTTP/2 connection (`afterHttp2ServerHandlerSetUp`). In each case the cached answer matches what the old per-request `pipeline.get` returned. The h2c upgrade installs a fresh `StreamPipeline` in `upgradeTo`, and the upgraded request is never routed through the HTTP/1 pipeline first, so no stale `true` can carry over to the connection pipeline. Lazy inspection on the first request is the right answer to the customizer case, and state is only touched on the channel's event loop, so the non-volatile field is fine.

The red `Java CI / Test Report (25)` is `NettyHttpServerSpec.executionError` from an earlier run, which is unrelated; `Java CI` on the same head is green.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/accesslog/AccessLogPipelineFlagSpec.groovy
graemerocher: Reworded in 4b89ac66e7: the flag is worked out from the pipeline on the first request and remembered per pipeline, and the javadoc now names the late-added-handler case.

PR #13143, opened 2026-09-11
Keep the open WebSocket sessions next to the channel group

graemerocher:
## Summary
- `WebSocketSession.getOpenSessions()` on a server session collected the open sessions from the server's `ChannelGroup` through a `Stream.flatMap` pipeline with a capturing lambda on every call (chat-style handlers call it per message). It now loops over the group and reads the session attribute once per channel into a `HashSet`; the group only ever holds websocket channels and drops them when they close, so nothing else is needed.
- A first version of this PR added `WebSocketSessionRepository.getOpenSessions()` and a per-channel session map on `NettyHttpServer`. That put `WebSocketSession` into the generic signature of an interface every application's server implements while `micronaut-websocket` is only an optional dependency, and anything inspecting the server reflectively (Groovy's `MetaClass` in the `http-client` and `http-client-jdk` suites) failed with `TypeNotPresentException` when the module is absent. The API addition and the map are gone (`6be76be91e`); the change is now confined to `NettyServerWebSocketHandler`.
- The second half of the LC-15 remainder, caching the `LoomBranchSupport.currentScheduler()` lookup, is not done: the `MethodHandle` lookups already happen once in the static initializer, what remains per call is the unwrap of the JDK scheduler to our proxy; a cache keyed by the JDK scheduler object cannot be bounded because `LoomCarrierGroup` installs a new `StickyScheduler` per virtual thread; and the code path cannot run on JDK 25, so no test could exercise a change.

## Testing
- `OpenSessionsSpec` (new): two clients on a real server; each server session reports its own id and the ids of all open sessions (both see both), after the first client closes only the second remains (checked through the next reply and the channel group), then none.
- `./gradlew :micronaut-http-server-netty:test --tests 'io.micronaut.http.server.netty.websocket.*'`: 26 tests, 0 failures. `:micronaut-http-client:test --tests '*ServiceIdSpec' --tests '*SslSelfSignedSpec'` (two of the specs that failed on CI with the first version): 4 tests, 0 failures. `checkstyleMain`/`checkstyleTest` clean apart from the pre-existing violations.
- Codex review (`gpt-5.6-sol`, high) of the first version: (1) the new API exposed the `@Internal` `NettyWebSocketSession`; (2) a capturing lambda put `NettyWebSocketSession` into a synthetic method descriptor of `NettyHttpServer`; (3) the close listener was not detached by `removeChannel`; (4) direct mutation of the exposed `ChannelGroup` could diverge from the map. All four are moot now that the server is untouched.

## Out of scope
- `LoomBranchSupport.currentScheduler()`, as above.
- `NettyServerWebSocketBroadcaster` still uses the channel group with a matcher, which is what the group is for.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/websocket/WebSocketSessionRepository.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/websocket/WebSocketSessionRepository.java
@@ -15,8 +15,12 @@
  */
 package io.micronaut.http.netty.websocket;
 
+import io.micronaut.websocket.WebSocketSession;
 import io.netty.channel.Channel;
 import io.netty.channel.group.ChannelGroup;
+
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * Defines a ChannelGroup repository to handle WebSockets.
@@ -43,4 +47,22 @@ public interface WebSocketSessionRepository {
      * @return A {@link io.netty.channel.group.ChannelGroup}
      */
     ChannelGroup getChannelGroup();
+
+    /**
+     * Returns the sessions of the channels in this repository that are currently open, as a
+     * snapshot. The default implementation derives them from {@link #getChannelGroup()}.
+     *
+     * @return The open sessions
+     * @since 5.2.2
+     */
+    default Set<? extends WebSocketSession> getOpenSessions() {
+        Set<WebSocketSession> sessions = new HashSet<>();
+        for (Channel channel : getChannelGroup()) {
+            NettyWebSocketSession session = channel.attr(NettyWebSocketSession.WEB_SOCKET_SESSION_KEY).get();
+            if (session != null && session.isOpen()) {
+                sessions.add(session);
+            }
+        }
+        return sessions;
+    }
 }
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
@@ -38,6 +38,7 @@ import io.micronaut.http.netty.channel.DefaultEventLoopGroupRegistry;
 import io.micronaut.http.netty.channel.EventLoopGroupConfiguration;
 import io.micronaut.http.netty.channel.NettyChannelType;
 import io.micronaut.http.netty.channel.converters.ChannelOptionFactory;
+import io.micronaut.http.netty.websocket.NettyWebSocketSession;
 import io.micronaut.http.netty.websocket.WebSocketSessionRepository;
 import io.micronaut.http.server.HttpServerConfiguration;
 import io.micronaut.http.server.exceptions.ServerStartupException;
@@ -56,6 +57,7 @@ import io.micronaut.runtime.server.event.ServerShutdownEvent;
 import io.micronaut.runtime.server.event.ServerStartupEvent;
 import io.micronaut.scheduling.TaskExecutors;
 import io.micronaut.web.router.Router;
+import io.micronaut.websocket.WebSocketSession;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.Channel;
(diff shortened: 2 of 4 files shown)


http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
```diff
@@ -829,18 +843,39 @@ private void processOptions(Map<ChannelOption, Object> options, BiConsumer<Chann
...
+        }
     }
 
     @Override
     public ChannelGroup getChannelGroup() {
         return this.webSocketSessions;
     }
 
+    @Override
+    public Set<? extends WebSocketSession> getOpenSessions() {
```
dstepanov: Having websocket types only in the generic signature isn't enough. `java.beans.Introspector` sees `getOpenSessions()` as the `openSessions` property and resolves its generic return type `Set<? extends WebSocketSession>`. Without `micronaut-websocket` on the classpath that throws `TypeNotPresentException`.

Groovy's MetaClass introspects the class on any property access, so `embeddedServer.applicationContext` (`ClientProxySpec.groovy:102`) fails. That is where the http-client-jdk failures come from, and any Groovy/Spock test suite that uses a server without websockets would hit the same thing.

http-netty/src/main/java/io/micronaut/http/netty/websocket/WebSocketSessionRepository.java
```diff
@@ -43,4 +47,22 @@ public interface WebSocketSessionRepository {
...
     ChannelGroup getChannelGroup();
+
+    /**
+     * Returns the sessions of the channels in this repository that are currently open, as a
+     * snapshot. The default implementation derives them from {@link #getChannelGroup()}.
+     *
+     * @return The open sessions
+     * @since 5.2.2
+     */
+    default Set<? extends WebSocketSession> getOpenSessions() {
```
dstepanov: The default method has the same signature (`()Ljava/util/Set<+Lio/micronaut/websocket/WebSocketSession;>;` per javap), and `Class.getMethods()` includes it, so every implementation inherits the problem, including `NettyEmbeddedServer` implementations outside core. Removing the `NettyHttpServer` override alone wouldn't fix it.

Some options: a name that isn't a bean getter, a return type that doesn't name websocket types, or keeping the lookup in a websocket-only helper that the server session calls instead of on the repository interface.

dstepanov (commented):
The session bookkeeping looks right to me: every close path (`session.close()`, `handlerRemoved`, and a channel closing without either) removes the entry, and iterating the `ConcurrentHashMap` is safe while sessions come and go.

The new API is what breaks Java CI, though: 57 of 90 tests fail in `:micronaut-http-client-jdk:test` and 9 of 9 in `:test-suite-http-client-jdk-ssl:test`, all with `TypeNotPresentException: Type io.micronaut.websocket.WebSocketSession not present`. I reproduced it locally: `ServiceIdSpec` and `OptionsRequestAttributesSpec` fail 4/4 at f220518bc4 and pass on the merge base 2bb34f752f. Details inline.

http-server-netty/src/main/java/io/micronaut/http/server/netty/NettyHttpServer.java
graemerocher: Agreed, and thanks for pinning it to `Introspector`. Fixed in 6be76be91e by removing the API and the map altogether: `WebSocketSessionRepository` and `NettyHttpServer` are back to the merge base, and the server session collects the open sessions with a plain loop over the channel group inside `NettyServerWebSocketHandler`, which only loads with websocket support. `ServiceIdSpec` and `SslSelfSignedSpec` pass locally again.

http-netty/src/main/java/io/micronaut/http/netty/websocket/WebSocketSessionRepository.java
graemerocher: Went with the third option: no repository API at all (6be76be91e). The default method is gone with it, so nothing outside core inherits a websocket-typed signature.

PR #13146, opened 2026-09-11
Document where request event listeners run and report slow ones

graemerocher:
## Summary
- `HttpRequestReceivedEvent` and `HttpRequestTerminatedEvent` listeners are invoked on the executor the server selects for request handling (`ExecutorSelector.selectExecutor(null, config)`), which with the default `micronaut.server.thread-selection` (`MANUAL`, and also `AUTO`) is `ImmediateExecutor`, i.e. inline on the event loop of the connection: before the request is routed, and once the server is done with the request. The javadoc of both events said they are fired "in a non-blocking manner" / "asynchronously", which does not tell a listener author that blocking there stalls every connection of that event loop.
- Of the two options considered, this PR documents rather than re-dispatches. Running the listeners on the blocking executor whenever a listener bean exists would contradict the existing design and its test (`ThreadSelectionSpec` pins request event listeners to the loop for `MANUAL`/`AUTO` and to the blocking executor for `BLOCKING`/`IO`), and would add two thread hops per request (out to the executor and back to the loop via `executeOnEventLoopIfNeeded`) for every application that has any listener, including the many that only read headers.
- Javadoc of both events now states where listeners run, that they must not block, and the two ways out: `@Async` on the `@EventListener` method, or `micronaut.server.thread-selection=BLOCKING`. The guide's "Server Events" table gains both events and a `WARNING` with the same guidance (rendered with `publishGuide`; the `ann:scheduling.annotation.Async[]` link resolves).
- When debug logging is enabled for `io.micronaut.http.server.netty.RoutingInBoundHandler`, listeners that take longer than 100 ms while on the event loop are reported at debug level with the event type, the time and the request. With debug logging off the only cost is the `isDebugEnabled()` check; the timing only happens with it on, and the report is suppressed when the listeners ran off the loop (`BLOCKING`/`IO`).

## Testing
- New `SlowRequestEventListenerSpec`: a listener for both events sleeps for a configured time; with `MANUAL` (default) and `AUTO` and 250 ms, both listeners run on `default-eventLoopGroup-*` and one debug report per event is logged; with `MANUAL` and no delay, no report; with `BLOCKING`, the listeners run off the loop and no report is logged even though they take 250 ms.
- Fail-before: with the threshold effectively disabled (`TimeUnit.DAYS.toNanos(1)`), the 250 ms cases fail on the report count; the javadoc and guide changes are documentation.
- `./gradlew :micronaut-http-server-netty:test`: 1162 tests, 22 skipped (pre-existing), 0 failures. `:micronaut-http:test`: 1857 tests, 0 failures. `checkstyleMain` clean for `http` and, apart from the pre-existing `FileLength` violation in `NettyHttpServerConfiguration`, for `http-server-netty`. `./gradlew publishGuide` succeeds and the new section renders.
- Codex review (`gpt-5.6-sol`, high): (1) the terminated-event javadoc said "after the response has been written", which over-promises for streamed and discarded responses; reworded to "once the server is done with the request", with the guide saying the event does not imply a response was delivered. (2) The documentation named `AUTO` as the default thread selection; the default is `MANUAL` (`HttpServerConfiguration.threadSelection`); corrected in the javadoc and guide, and the test gained the `MANUAL` case. Third pass: no findings.

## Out of scope
- Making the threshold configurable; 100 ms is well beyond anything that belongs on an event loop and the report is debug-only.
- Changing `ThreadSelection` semantics.

### DIFF
--- a/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
+++ b/http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
@@ -58,6 +58,7 @@ import java.nio.channels.ClosedChannelException;
 import java.util.Optional;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 
@@ -78,6 +79,10 @@ public final class RoutingInBoundHandler implements RequestHandler {
      */
     private static final Pattern IGNORABLE_ERROR_MESSAGE = Pattern.compile(
         "^.*(?:connection (?:reset|closed|abort|broken)|broken pipe).*$", Pattern.CASE_INSENSITIVE);
+    /**
+     * Request event listeners that take longer than this on the event loop are reported at debug level.
+     */
+    private static final long SLOW_LISTENER_THRESHOLD_NANOS = TimeUnit.MILLISECONDS.toNanos(100);
 
     final StaticResourceResolver staticResourceResolver;
     final NettyHttpServerConfiguration serverConfiguration;
@@ -140,7 +145,7 @@ public final class RoutingInBoundHandler implements RequestHandler {
                     terminatedFlow = ExecutionFlow.async(getRequestEventExecutor(), () -> {
                         PropagatedContext.getOrEmpty()
                             .plus(new ServerHttpRequestContext(request))
-                            .propagate(() -> terminateEventPublisher.publishEvent(new HttpRequestTerminatedEvent(request)));
+                            .propagate(() -> publishRequestEvent(request, terminateEventPublisher, new HttpRequestTerminatedEvent(request)));
                         return ExecutionFlow.empty();
                     });
                 }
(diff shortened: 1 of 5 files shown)


http/src/main/java/io/micronaut/http/context/event/HttpRequestReceivedEvent.java
```diff
@@ -18,9 +18,18 @@
...
+ * URI, headers and so on but should not perform I/O.
+ * <p>
+ * Listeners are invoked on the thread the server selects for request handling, which with the
+ * default {@code micronaut.server.thread-selection} ({@code MANUAL}, and also with {@code AUTO})
+ * is the Netty event loop that received the request. Every listener runs to completion before the
+ * request is routed, so a listener that blocks or does slow work delays this request and every
+ * other connection on that event loop. Listeners that need to block must hand off, for example
+ * with {@code @Async} on the {@code @EventListener} method, or the server can be configured with
+ * {@code micronaut.server.thread-selection=BLOCKING} which moves the listeners (and controllers)
+ * off the event loop.
```
dstepanov, line 32: `BLOCKING` doesn't unconditionally move listeners off the loop. `redispatch-non-blocking-only` defaults to `true`, so `ExecutorSelector.selectExecutor` returns a `ConditionalExecutionExecutor` that only hands off when the current thread is a reactor `NonBlocking` thread (`ExecutorSelector.java:69-83`). Event-loop threads get that marker only when `NettyGlobalConfiguration.defaultThreadFactoryReactorNonBlocking` is on (`NettyThreadFactory.java:158`). If that's turned off, or an event loop uses its own thread factory, listeners still run inline with `BLOCKING`, and the new debug message then tells the user to set `BLOCKING`. Could this be qualified, or `@Async` recommended first? The same text is in `HttpRequestTerminatedEvent` and `serverEvents.adoc`.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/SlowRequestEventListenerSpec.groovy
```diff
@@ -0,0 +1,119 @@
...
+        client.close()
+        ctx.close()
+        logger.detachAppender(appender)
+        logger.level = previousLevel
+
+        where:
+        threadSelection | delayMillis | expectReport
+        'MANUAL'        | 250         | true
+        'AUTO'          | 250         | true
+        'MANUAL'        | 0           | false
```
dstepanov, line 74: This case expects zero reports, but it times the first publish in a fresh context, which also creates the `@EventListener` adapter beans, against a fixed 100 ms. A GC pause on a busy runner could produce a report. Sending one warm-up request before attaching the appender would make it deterministic.

http-server-netty/src/main/java/io/micronaut/http/server/netty/RoutingInBoundHandler.java
```diff
@@ -256,11 +261,32 @@ private ExecutionFlow<Void> executionFlowForReceivedEvent(NettyHttpRequest<?> re
...
+            publisher.publishEvent(event);
+            return;
+        }
+        long start = System.nanoTime();
+        try {
+            publisher.publishEvent(event);
+        } finally {
+            long taken = System.nanoTime() - start;
+            if (taken > SLOW_LISTENER_THRESHOLD_NANOS && request.getChannelHandlerContext().executor().inEventLoop()) {
+                LOG.debug("Listeners for {} took {} ms on the event loop for {}. Request event listeners must not block; use @Async on the listener or micronaut.server.thread-selection=BLOCKING", event.getClass().getSimpleName(), TimeUnit.NANOSECONDS.toMillis(taken), request);
```
dstepanov, line 285: Nothing limits how often this logs: a listener that is always slow produces two lines per request, each with the full URI including the query string. It's opt-in debug on one class, so not a big deal, but logging once per event type (or once per listener) would keep the signal without the volume.

dstepanov (commented):
The threading docs match the dispatch code: `MANUAL`/`AUTO` publish both events inline on the loop through `ImmediateExecutor`, routing waits for the received flow, and the terminated event fires from `responseWritten`. The debug check costs nothing measurable when there are no listeners (`isEmpty()` skips publishing) and just an `isDebugEnabled()` call otherwise. Three small comments inline.

Nit: the guide now names the internal logger class and the 100 ms threshold, which makes both part of the documented contract.

dstepanov (approved):
Approving; the inline comments above are non-blocking suggestions.

Issue #1126, opened 2019-01-17
[Kotlin] Json validation flow


raderio:
Usually on deserialization we want to collect all validation errors and return them in response to the client. 
In order to catch all error the validation should be done before object creation.

If it will be done after object creation, in `init` block, than validation will be separated in 2 phases, first one will be validation regarding non-nullable field, because we cannot create objects if they have non-nullable properties. 
The second one will be to do structural validation: min/max values, length, regexp pattern, etc.

Example
`class Data(val name: String, val age: Int, val email: String)`
when we send a json `{"email": "xyz@xyz.com", "name": "A"}` we should return a response where to indicate that `age` can not be null and `name` size must be greater than 2, at once to send all the errors.

Also, structural validation will be great to do based on annotations and not in `init` block, something like JSR 303.
This will be useful for documentation generation, it is far easier to generate based on annotations.

How I see the flow:
1. Read annotations and generate validation rules. Can be done by 3rd party library. As result Map<Class, List<Rules>>
2. Parse json as a map
3. Apply validation rules. Validation rules can be passed as parameter in configuration.
4. Create object

This flow also will fit validation for Java.

graemerocher:
I think we probably need more clarity on what you are trying to achieve. Maybe if you provide an example we can help solve your case.

graemerocher:
Needs more feedback, if provided please reopen

raderio:
I mean that is a solution to solve this issue https://github.com/FasterXML/jackson-module-kotlin/issues/58

Basically I tried to fix that for Kotlin validation should return all validation failure, not just first error regarding to null.

raderio:
@graemerocher what do you think about this?

Issue #1124, opened 2019-01-16
Using micronaut-session with GraalVM results in error


JWThewes:
I'm using Micronaut 1.0.3 and created a simple chat application.
When trying to build a native image using GraalVM RC 8 I get the following error on executing the native image:

{"message":"Internal Server Error: Error instantiating bean of type [io.micronaut.session.http.HttpSessionFilter]: Class io.micronaut.session.http.HttpSessionIdResolver[] is instantiated reflectively but was never registered. Register the class by using org.graalvm.nativeimage.RuntimeReflection"}

Tries with GraalVM RC 11 too but then the image creation fails.

Issue #1123, opened 2019-01-16
NoSuchMethodError when Parameter is Generated Class


RyanHoldren:
I am seeing a strange error. If I have a controller endpoint where one of the parameters is a generated class, the Definition class generated by Micronaut's annotation processor incorrectly skips the parameter as though there was no parameter at all.

For example:
```
@Controller("/email-drafts")
public class EmailDraftController {
	@Delete("/{draft}")
	public void delete(EmailDraft draft) {
		draft.delete();
	}
}
```
```
curl -X DELETE 'http://localhost/email-drafts/ZHJAZnSOvYifonFVn_OmM-hjZrY_22gYvsvLdvNatVE'
```
```
[nioEventLoopGroup-1-9] ERROR io.micronaut.http.server.netty.RoutingInBoundHandler - Unexpected error occurred: com.vercettie.switchboard.email.messages.drafts.EmailDraftController.delete()V
java.lang.NoSuchMethodError: com.vercettie.switchboard.email.messages.drafts.EmailDraftController.delete()V
at com.vercettie.switchboard.email.messages.drafts.$EmailDraftControllerDefinition$$exec12.invokeInternal(Unknown Source)
at io.micronaut.context.AbstractExecutableMethod.invoke(AbstractExecutableMethod.java:145)
at io.micronaut.context.DefaultBeanContext$BeanExecutionHandle.invoke(DefaultBeanContext.java:2413)
at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:237)
at io.micronaut.web.router.RouteMatch.execute(RouteMatch.java:123)
at io.micronaut.http.server.netty.RoutingInBoundHandler.lambda$buildResultEmitter$16(RoutingInBoundHandler.java:1284)
at io.reactivex.internal.operators.flowable.FlowableCreate.subscribeActual(FlowableCreate.java:71)
at io.reactivex.Flowable.subscribe(Flowable.java:14479)
at io.reactivex.internal.operators.flowable.FlowableMap.subscribeActual(FlowableMap.java:37)
at io.reactivex.Flowable.subscribe(Flowable.java:14479)
at io.reactivex.internal.operators.flowable.FlowableSwitchIfEmpty.subscribeActual(FlowableSwitchIfEmpty.java:32)
at io.reactivex.Flowable.subscribe(Flowable.java:14479)
at io.reactivex.Flowable.subscribe(Flowable.java:14429)
[...]

PR #13147, opened 2026-09-11
HTTP server TCK: body framing, request size limit, cookie attributes, Content-Length; Vary is only CORS when it names Origin

graemerocher:
Portable TCK tests for behaviour that differed between the Netty server and the servlet runtimes, written for the [servlet parity plan](https://claude.ai/code/artifact/47fda1ae-42c9-4117-b46d-d9018fe6e6a3) and validated on Netty here and on Jetty, Tomcat, Undertow and the JDK server in micronaut-servlet via `--include-build` (the fixes they drove are in micronaut-projects/micronaut-servlet#1123, #1124 and #1126).

- **BodyWithoutContentLengthTest**: an HTTP/1 request with neither `Content-Length` nor `Transfer-Encoding` has no body (RFC 9112); the servlet runtimes used to block waiting for one (micronaut-servlet#1097).
- **MaxRequestSizeTest**: `micronaut.server.max-request-size` applies to text, JSON and form bodies with 413, not only to multipart uploads. Found an oversized JSON body answering 400 on the servlet runtimes (Jackson 3 unchecked exception wrapping the stream failure), fixed in #1126.
- **CookieAttributesTest**: path, domain, Max-Age, Secure, HttpOnly and SameSite all reach `Set-Cookie`; several cookies give several headers.
- **ContentLengthTest**: String, byte[] and JSON responses announce their length.
- **CorsUtils / CorsSimpleRequestTest / SimpleRequestWithCorsNotEnabledTest**: the "no CORS headers" assertions failed on any `Vary` header at all. A server that compresses legitimately sends `Vary: Accept-Encoding` (Jetty's GzipHandler adds it to every compressible response whether or not it compresses this one), so the assertions now check that `Vary` does not name `Origin`. Until this ships, the Jetty suite in micronaut-servlet excludes the two affected tests.

**Downstream runners.** The function TCK suites select the whole `tests` package, so these tests reach micronaut-aws, -gcp and -azure as soon as they bump core, before a servlet release makes their server pass them (see the review: `BodyWithoutContentLengthTest`, `MaxRequestSizeTest` and, on aws, `CookieAttributesTest` fail on servlet 6.1.0). Each of those three classes carries a tag, `body-without-content-length`, `max-request-size` and `cookie-attributes`, so a runner can `@ExcludeTags` precisely the behaviour its server lacks, the way `multipart` is handled, and drop the exclusion when it bumps servlet.

Codex CLI review: no findings.

### DIFF
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/CorsUtils.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/CorsUtils.java
@@ -19,6 +19,8 @@ import io.micronaut.http.HttpHeaders;
 import io.micronaut.http.HttpMethod;
 import io.micronaut.http.HttpResponse;
 
+import java.util.Arrays;
+
 import static org.junit.jupiter.api.Assertions.assertFalse;
 
 /**
@@ -32,10 +34,25 @@ public final class CorsUtils {
     }
 
     /**
+     * Asserts that CORS did not mark the response as varying by {@code Origin}. A server may legitimately send a
+     * {@code Vary} header for other reasons, such as {@code Vary: Accept-Encoding} when it can compress the
+     * response, so the header's presence alone says nothing about CORS.
+     *
+     * @param response HTTP Response to run the assertion against
+     * @since 5.2.1
+     */
+    public static void assertVaryDoesNotNameOrigin(HttpResponse<?> response) {
+        assertFalse(response.getHeaders().getAll(HttpHeaders.VARY).stream()
+            .flatMap(value -> Arrays.stream(value.split(",")))
+            .map(String::trim)
+            .anyMatch(HttpHeaders.ORIGIN::equalsIgnoreCase));
+    }
+
+    /**
      * @param response HTTP Response to run CORS assertions against it.
      */
     public static void assertCorsHeadersNotPresent(HttpResponse<?> response) {
-        assertFalse(response.getHeaders().names().contains(HttpHeaders.VARY));
+        assertVaryDoesNotNameOrigin(response);
         assertFalse(response.getHeaders().names().contains(HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS));
         assertFalse(response.getHeaders().names().contains(HttpHeaders.ACCESS_CONTROL_MAX_AGE));
         assertFalse(response.getHeaders().names().contains(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN));
--- a/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyWithoutContentLengthTest.java
+++ b/http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyWithoutContentLengthTest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.http.server.tck.tests;
+
+import io.micronaut.context.annotation.Requires;
+import io.micronaut.core.annotation.NonNull;
+import io.micronaut.http.HttpRequest;
+import io.micronaut.http.HttpStatus;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Get;
+import io.micronaut.http.annotation.QueryValue;
+import io.micronaut.http.tck.AssertionUtils;
+import io.micronaut.http.tck.HttpResponseAssertion;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+
+import static io.micronaut.http.tck.TestScenario.asserts;
+
+/**
+ * A request that carries no body at all must still reach a route that declares a body type.
+ *
+ * <p>A GET has no {@code Content-Length} header, so the body length is reported as unknown rather than zero. That is
+ * indistinguishable from a body that has not arrived yet unless the server checks whether any bytes exist, and a
+ * server that assumes "unknown means present" fails to decode the absent body and answers 400.</p>
+ *
+ * <p>The shape comes from the OAuth 2.0 authorization code callback, which is a browser redirect carrying only query
+ * parameters into a controller whose parameter is a {@code HttpRequest<Map<String, Object>>}.</p>
+ */
+@SuppressWarnings({"java:S5960", "checkstyle:MissingJavadocType", "checkstyle:DesignForExtension"})
+public class BodyWithoutContentLengthTest {
+    public static final String SPEC_NAME = "BodyWithoutContentLengthTest";
+
+    @Test
+    void getWithABodyTypeAndNoBodyReachesTheRoute() throws IOException {
+        asserts(SPEC_NAME,
+            HttpRequest.GET("/body-without-content-length/callback?code=abc"),
+            (server, request) -> AssertionUtils.assertDoesNotThrow(server, request,
+                HttpResponseAssertion.builder()
+                    .status(HttpStatus.OK)
+                    .body("code=abc body=empty")
+                    .build()));
+    }
+
+    @Test
+    void deleteWithABodyTypeAndNoBodyReachesTheRoute() throws IOException {
+        asserts(SPEC_NAME,
+            HttpRequest.GET("/body-without-content-length/optional-body"),
+            (server, request) -> AssertionUtils.assertDoesNotThrow(server, request,
+                HttpResponseAssertion.builder()
+                    .status(HttpStatus.OK)
+                    .body("empty")
+                    .build()));
+    }
+
+    @Controller("/body-without-content-length")
+    @Requires(property = "spec.name", value = SPEC_NAME)
+    static class BodyWithoutContentLengthController {
+
+        @Get(value = "/callback", produces = MediaType.TEXT_PLAIN)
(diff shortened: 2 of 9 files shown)


http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MaxRequestSizeTest.java
```diff
@@ -0,0 +1,121 @@
...
+            .assertion((server, request) -> AssertionUtils.assertDoesNotThrow(server, request,
+                HttpResponseAssertion.builder()
+                    .status(HttpStatus.OK)
+                    .body("512")
+                    .build()))
+            .run();
+    }
+
+    @Test
+    void aTextBodyOverTheLimitIsRefusedWith413() throws IOException {
```
dstepanov, line 60: Three of the four cases here fail on the function runners. servlet-core (6.1.0, and 6.2.x until micronaut-servlet#1126 merges) has no max-request-size check for non-multipart bodies. On gcp function-http, the 4096-byte text and JSON bodies get 200 instead of 413, and the form body gets 400. On aws api-proxy, all three get 200.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyWithoutContentLengthTest.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * <p>The shape comes from the OAuth 2.0 authorization code callback, which is a browser redirect carrying only query
+ * parameters into a controller whose parameter is a {@code HttpRequest<Map<String, Object>>}.</p>
+ */
+@SuppressWarnings({"java:S5960", "checkstyle:MissingJavadocType", "checkstyle:DesignForExtension"})
+public class BodyWithoutContentLengthTest {
+    public static final String SPEC_NAME = "BodyWithoutContentLengthTest";
+
+    @Test
+    void getWithABodyTypeAndNoBodyReachesTheRoute() throws IOException {
```
dstepanov, line 54: Both cases fail on runners using a released servlet (6.1.0). The fix is only on servlet 6.2.x (#1123). On aws api-proxy, both answer 400 `Error decoding request body`. On gcp function-http, `/callback` answers 400 and `/optional-body` answers 500. These runners need the servlet bump or an exclusion before the core 5.2.2 bump reaches them.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyWithoutContentLengthTest.java
```diff
@@ -0,0 +1,87 @@
...
+            HttpRequest.GET("/body-without-content-length/callback?code=abc"),
+            (server, request) -> AssertionUtils.assertDoesNotThrow(server, request,
+                HttpResponseAssertion.builder()
+                    .status(HttpStatus.OK)
+                    .body("code=abc body=empty")
+                    .build()));
+    }
+
+    @Test
+    void deleteWithABodyTypeAndNoBodyReachesTheRoute() throws IOException {
```
dstepanov: Nit: the name says DELETE, but it sends `HttpRequest.GET` to a `@Get` route. Either rename it, or add a real `@Delete` route with a body type. A body-less DELETE is the other common case with no `Content-Length`.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/CookieAttributesTest.java
```diff
@@ -0,0 +1,114 @@
...
+                        assertEquals(2, setCookies.size(), () -> "expected two Set-Cookie headers: " + setCookies);
+                        assertTrue(setCookies.stream().anyMatch(c -> c.startsWith("first=1")), setCookies::toString);
+                        assertTrue(setCookies.stream().anyMatch(c -> c.startsWith("second=2")), setCookies::toString);
+                    })
+                    .build()));
+    }
+
+    private static void assertFullCookie(HttpResponse<?> response) {
+        List<String> setCookies = response.getHeaders().getAll(HttpHeaders.SET_COOKIE);
+        assertEquals(1, setCookies.size(), () -> "expected one Set-Cookie header: " + setCookies);
```
dstepanov, line 81: The assertions are right per RFC 6265 §3, but both cases fail on the AWS api-proxy runner (gcp passes both). There, `cookieAttributesAreEmitted` fails with `expected: <1> but was: <2>`: the value is split at the comma in `Expires=Mon, 14 Sep 2026 ...`. `severalCookiesProduceSeveralSetCookieHeaders` sees only `[first=1]`. That's a real aws bug the test surfaces, but it needs a fix or an exclusion in micronaut-aws before the core bump.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/RequestTerminatedEventTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * The event is published after the response has been sent, so the client can return before it fires.
+     */
+    private static void awaitTerminated(Counter counter, String path) {
+        long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(5);
+        while (counter.terminated.get() == 0 && System.nanoTime() < deadline) {
+            Thread.onSpinWait();
+        }
+        assertTrue(counter.terminated.get() == 1, () -> "expected one terminated event for " + path + " but saw " + counter.terminated.get());
```
dstepanov: This checks `== 1` as soon as the counter leaves zero, so a duplicate event published slightly later (e.g. from both the error path and the completion path) still passes. That's the "once" the Javadoc promises. Waiting a short grace period after the first event before asserting exactly 1 would catch it, and a sleep/poll would avoid spinning a core for up to 5 s.

http-server-tck/src/main/java/io/micronaut/http/server/tck/CorsUtils.java
```diff
@@ -31,11 +33,26 @@ private CorsUtils() {
 
     }
 
+    /**
+     * Asserts that CORS did not mark the response as varying by {@code Origin}. A server may legitimately send a
+     * {@code Vary} header for other reasons, such as {@code Vary: Accept-Encoding} when it can compress the
+     * response, so the header's presence alone says nothing about CORS.
+     *
+     * @param response HTTP Response to run the assertion against
+     * @since 5.2.1
```
dstepanov: `v5.2.1` is already tagged and 5.2.x is at `5.2.2-SNAPSHOT`, so this should be `@since 5.2.2`.

dstepanov (commented):
The new tests look spec-correct to me (RFC 9112 §6.3 framing, RFC 9110 413, RFC 6265 §3 no Set-Cookie folding), and the CORS `Vary` change is safe: `CorsFilter.setVary` only ever writes `Vary: Origin`, and the new assertion still catches that.

My main concern is timing for downstream runners. The function TCK suites select the whole `io.micronaut.http.server.tck.tests` package, so these tests start running there as soon as core 5.2.2 is bumped. I copied the new test classes into two runners (core 5.2.1, servlet 6.1.0) and ran them:

| Runner | Failed | Tests |
|---|---|---|
| micronaut-gcp 6.2.x `test-suite-http-server-tck-gcp-function-http` | 5 / 16 | `BodyWithoutContentLengthTest` x2, `MaxRequestSizeTest` x3 |
| micronaut-aws 5.2.x `test-suite-http-server-tck-function-aws-api-proxy-test` | 7 / 16 | `BodyWithoutContentLengthTest` x2, `MaxRequestSizeTest` x3, `CookieAttributesTest` x2 |

`RequestTerminatedEventTest`, `ServerSentEventsTest` and `ContentLengthTest` pass on both. Azure uses the same servlet-core, so I'd expect the same result there, but I didn't run it. On the servlet side, #1124 and #1126 are still open, and the size-limit fix is in #1126.

Since this targets a patch branch, could we either land the servlet release and add exclusions in aws/azure/gcp before 5.2.2, or tag these classes so runners can opt out (as `multipart` does)? Inline notes below.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/MaxRequestSizeTest.java
graemerocher: Tagged the class `@Tag("max-request-size")` in f8863d4a12 so a runner whose server does not yet enforce the limit on every body (servlet 6.1.0, and 6.2.x until micronaut-servlet#1126 lands) can exclude it with `@ExcludeTags`, as with `multipart`.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyWithoutContentLengthTest.java
graemerocher: Tagged `@Tag("body-without-content-length")` (f8863d4a12) so the aws/gcp/azure runners can exclude it until they pick up a servlet release with the framing fix.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/BodyWithoutContentLengthTest.java
graemerocher: Renamed the GET case to `getWithAStringBodyTypeAndNoBodyReachesTheRoute` and added a real `@Delete("/item/{id}")` route with a `HttpRequest<Map<String, Object>>` parameter; the DELETE case now sends `HttpRequest.DELETE` and expects `deleted 42 body=empty`. Passes on Netty and the servlet runners.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/CookieAttributesTest.java
graemerocher: Tagged `@Tag("cookie-attributes")` (f8863d4a12) so micronaut-aws can exclude it until the Set-Cookie splitting it surfaces is fixed there.

http-server-tck/src/main/java/io/micronaut/http/server/tck/tests/RequestTerminatedEventTest.java
graemerocher: Now polls with `LockSupport.parkNanos(10ms)` instead of spinning, and after the first event waits a further 200ms before asserting `assertEquals(1, ...)`, so a duplicate published later by another exit path fails the test.

http-server-tck/src/main/java/io/micronaut/http/server/tck/CorsUtils.java
graemerocher: Changed to `@since 5.2.2`.

yawkat (approved):

PR #13148, opened 2026-09-11
Only devolve the streams read in the current batch at read complete

graemerocher:
## Summary
- `Http2ServerHandler.ConnectionHandler.channelReadComplete` walked every active stream of the connection on every read batch (`connection().forEachActiveStream`) to call `devolveToStreaming()`. That call is a no-op for every stream except the ones whose request HEADERS (without END_STREAM) were read since the last read complete and whose body has not arrived in full; with `max-concurrent-streams` at its default of 100, a busy connection paid up to a hundred property lookups per batch, on the event loop, for nothing.
- Because the walk accepts every such stream, none survive a read complete, so the streams that can need the call are exactly the ones whose HEADERS frame without END_STREAM was read in the current batch. `Http2ServerHandler` now keeps them in a list filled by `onHeadersRead` and cleared at read complete. Streams that netty closed in the meantime (reset by the client, reset for a local stream error, connection error) are skipped by checking the netty stream state, which is what the walk over the *active* streams did implicitly. Streams that received a GOAWAY stay open in netty and are still devolved, as before (the `MultiplexedStream.reset` flag is deliberately not used for the check, because it is also set on GOAWAY). Trailers, the h2c upgrade request and push promises never enter the list: they are either ignored or accepted with a full body immediately.
- The list is iterated in place: `devolveToStreaming()` calls into the request handler, which can close a *later* pending stream synchronously (a full response to a request whose body is still open ends the stream and resets it), which the per-stream state check handles; nothing else modifies the list during read complete, because `onHeadersRead` is only called by the frame decoder during `channelRead`.
- Design notes and the termination-path table are in the review write-up (`h2-15-readcomplete.md` in the review scratchpad); `MultiplexedServerHandler` is untouched.

## Testing
- New `Http2PendingStreamSpec` (EmbeddedChannel, one flush = one read batch): headers plus part of the body are devolved at read complete and the rest of the body arrives on the streaming body; a full body within one batch is accepted without devolving; a stream reset by the client in the same batch is not accepted while a later stream on the connection is; a stream that received a GOAWAY in the same batch is still devolved; two pending streams in one batch where the request handler responds to the first synchronously (ending and resetting it) still get the second accepted with no unbound error; a stream closed locally between the read and the read complete is not accepted, and removing the handler with a stream pending (what a connection close does) accepts nothing and raises nothing.
- Fail-before: the change is meant to be behaviour-preserving, so the spec guards equivalence with the base rather than failing on it. It does catch the obvious wrong implementation: without the netty state check, `stream reset by the client before read complete is not accepted` and `stream closed locally before read complete is not accepted` fail because the closed streams are handed to the request handler.
- `./gradlew :micronaut-http-server-netty:test`: 1164 tests, 22 skipped (pre-existing), 0 failures. `./gradlew :test-suite-http2-server-tck-netty:test`: 263 tests, 3 skipped, 0 failures. `checkstyleMain` clean apart from the pre-existing `FileLength` violation in `NettyHttpServerConfiguration`.
- Codex review (`gpt-5.6-sol`, high): no findings.

## Out of scope
- Anything in `MultiplexedServerHandler`; the buffering and streaming logic is unchanged.
- No benchmark: the walk is not exercised by the harness (one stream per connection), and the change removes work rather than restructuring it.

### DIFF
--- a/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/Http2PendingStreamSpec.groovy
+++ b/http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/Http2PendingStreamSpec.groovy
@@ -1,0 +1,307 @@
+package io.micronaut.http.server.netty.handler
+
+import io.micronaut.http.body.AvailableByteBody
+import io.micronaut.http.body.CloseableByteBody
+import io.micronaut.http.netty.body.NettyByteBodyFactory
+import io.micronaut.http.server.netty.EmbeddedTestUtil
+import io.netty.buffer.ByteBuf
+import io.netty.buffer.Unpooled
+import io.netty.channel.ChannelHandlerContext
+import io.netty.channel.embedded.EmbeddedChannel
+import io.netty.handler.codec.http.DefaultHttpResponse
+import io.netty.handler.codec.http.HttpMethod
+import io.netty.handler.codec.http.HttpRequest
+import io.netty.handler.codec.http.HttpResponseStatus
+import io.netty.handler.codec.http.HttpVersion
+import io.netty.handler.codec.http2.DefaultHttp2DataFrame
+import io.netty.handler.codec.http2.DefaultHttp2GoAwayFrame
+import io.netty.handler.codec.http2.DefaultHttp2Headers
+import io.netty.handler.codec.http2.DefaultHttp2HeadersFrame
+import io.netty.handler.codec.http2.DefaultHttp2ResetFrame
+import io.netty.handler.codec.http2.Http2ChannelDuplexHandler
+import io.netty.handler.codec.http2.Http2Error
+import io.netty.handler.codec.http2.Http2FrameCodecBuilder
+import io.netty.handler.codec.http2.Http2FrameStream
+import io.netty.handler.codec.http2.Http2Stream
+import spock.lang.Specification
+
+import java.nio.charset.StandardCharsets
+
+/**
+ * Only streams whose request headers arrived in the current read batch, without their full body,
+ * are devolved to streaming at read complete. This checks the ways such a stream can end before
+ * the read complete.
+ */
+class Http2PendingStreamSpec extends Specification {
+    private static class Accepted {
+        HttpRequest request
+        CloseableByteBody body
+        OutboundAccess outboundAccess
+    }
+
+    private static class DuplexHandler extends Http2ChannelDuplexHandler {
+        @Override
+        void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+            ctx.fireChannelRead(msg)
+        }
+    }
+
+    private static class Harness {
+        EmbeddedChannel server
+        EmbeddedChannel client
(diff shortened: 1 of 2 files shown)


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/Http2PendingStreamSpec.groovy
```diff
@@ -0,0 +1,307 @@
...
+        def second = harness.newStream()
+
+        when: 'two streams are pending in the same batch'
+        harness.writeHeaders(first, false)
+        harness.writeData(first, "a", false)
+        harness.writeHeaders(second, false)
+        harness.writeData(second, "b", false)
+        harness.flushBatch()
+
+        then: 'both are accepted, the first is reset by the server without disturbing the second'
```
dstepanov, line 243: Responding on the first stream only closes the first stream, after it was accepted, so no later pending stream is closed while `devolvePendingStreams` iterates. The case described in its comment isn't covered: this test still passes with the state check replaced by `true`. The name also says skipped while both are accepted. Closing `second` from the first accept (e.g. `connection().stream(second.id()).close()`) and asserting a single accept would cover it.

http-server-netty/src/test/groovy/io/micronaut/http/server/netty/handler/Http2PendingStreamSpec.groovy
```diff
@@ -0,0 +1,307 @@
...
+        then:
+        accepted.isEmpty()
+
+        when: 'the handler goes away with a stream still pending, as it does when the connection closes'
+        def stream2 = client.pipeline().get(DuplexHandler).newStream()
+        client.writeOutbound(new DefaultHttp2HeadersFrame(headers, false).stream(stream2))
+        while ((packet = client.readOutbound()) != null) {
+            server.writeOneInbound(packet)
+        }
+        server.pipeline().remove(connectionHandler)
```
dstepanov, line 296: Once removed, the handler never sees `fireChannelReadComplete()`, so `accepted.isEmpty()` holds whatever `devolvePendingStreams` does (stream2 is still OPEN and would be accepted if read complete reached it). A real connection close also goes through `channelInactive`, which closes the streams. I'd drop this block.

dstepanov (approved):
Checked that the change is behaviour-preserving: the new spec passes against the old `forEachActiveStream` walk, and replacing the netty state check with `true` fails exactly the two closed-stream tests. `Http2ServerHandlerSpec`, the `http2` package specs, `MaxRequestSizeSpec` and `AccessLogSpec` pass on the head. The 5.2.x changes to `Http2ServerHandler` since the base (`userEventTriggered`, the `closeInput` RST code) don't overlap. Two small test notes inline.

Unrelated to this PR and also true with the old walk: when HEADERS, part of the body and RST_STREAM arrive in one read batch, the stream is closed before read complete, so its `bufferedContent` is never released (nothing else releases it). I'll follow that up separately.

PR #13153, opened 2026-09-12
Keep the WebSocket frame content alive until an @OnMessage handler completes

graemerocher:
## Summary

- **Frame content outlived by the handler** (`AbstractNettyWebSocketHandler`): when an `@OnMessage` method declares its payload as a Netty `ByteBuf` or a Micronaut `ByteBuffer<?>`, the bound argument is a view of the frame content, but the handler released that content right after binding and Netty's auto-release dropped the last reference when the read returned. A handler dispatched to another thread (`@ExecuteOn`, or the blocking executor that non-reactive handlers run on) therefore ran against a released buffer and failed with `IllegalReferenceCountException`, or, if the pooled buffer had already been handed out again, read unrelated bytes. Ownership of the content now passes to the handler invocation when the bound argument aliases it, and the buffer is released once the invocation's `ExecutionFlow` completes (or binding fails). `byte[]` and `String` parameters are copied at bind time and were never affected.
- **Fragmented messages decoded from the last fragment only**: for a parameter type that needs a codec or a `MessageBodyReader` (any POJO), the decoders were handed `msg.content()`, the final `ContinuationWebSocketFrame`, instead of the aggregated buffer. All decoders now operate on the aggregated content. A `ByteBuf` parameter is bound directly, since the conversion service refuses to pass a `CompositeByteBuf` (which is `Iterable`) through unchanged; a `MessageBodyReader` that throws no longer leaks the reference it was handed.
- **Release ordering with listeners**: `NettyServerWebSocketHandler.messageHandled` used to publish `WebSocketMessageProcessedEvent` from a task on the channel's executor, which would have freed the content before a listener read it; it now publishes inline in the completing thread and the callback releases afterwards, which also removes the executor hop from the per-message path (review findings by @dstepanov and @yawkat).
- Documentation: the WebSocket server guide now says that a `ByteBuf`/`ByteBuffer` parameter is a view of the frame that is released once the handler (and any publisher it returns) has completed and the `WebSocketMessageProcessedEvent` listeners have run; that `WebSocketSession.send` takes ownership, so echoing needs `session.send(buffer.retain())`; and it no longer lists a Java NIO `ByteBuffer` parameter, which is not supported (no converter, falls through to the JSON codec) — pre-existing, noted in review.

## Testing

- `WebSocketFrameOwnershipSpec` (new, 8 features) drives the server with a raw Netty client so frames can be fragmented as needed: a JSON POJO split across a text frame and a continuation frame is decoded from both; a 16 KB binary message split in two is bound to `ByteBuf` and to `ByteBuffer<?>` on the blocking executor, with the handler sleeping 100 ms before reading, and the complete payload is observed with no error and no leak report (the module runs with paranoid leak detection). Fail-before against the unpatched handler: all three time out waiting for a message (the POJO one decodes garbage from the last fragment, the binary ones hit the released buffer). Four more features: a `WebSocketMessageProcessedEvent` listener reads the bound `ByteBuf` (fails with the refCnt error without the inline publish); the bound buffer is echoed through `session.send(buffer.retain())`; a `String` is bound from two text fragments; a custom `MessageBodyReader` under `@Consumes("application/x-custom")` decodes the aggregated content and its rejection reaches `@OnError` as a `CodecException` with no leak (the reader path and its failure path); a parameter nothing can decode closes the session with `UNSUPPORTED_DATA`.
- `./gradlew :micronaut-http-netty:test :micronaut-http-netty:checkstyleMain`: 148 tests, 0 failures.
- `./gradlew :micronaut-http-server-netty:test --tests 'io.micronaut.http.server.netty.websocket.*' :micronaut-http-server-netty:checkstyleTest`: 37 tests, 0 failures.
- `./gradlew :test-suite-kotlin:test --tests '*WebSocketSuspend*'`: 2 tests, 0 failures (`WebSocketSuspendByteBufTest` times out on the previous head).

## Review follow-ups

- @dstepanov: Kotlin `suspend` handlers completed their flow immediately (the suspend branch tested the `ExecutionFlow` for `COROUTINE_SUSPENDED`, which it never is; pre-existing since #12439), so a suspend handler bound to a `ByteBuf` lost its content, the processed event fired early and its errors never reached `@OnError`. The coroutine check now runs inside the async flow on the invocation's own result and the flow completes with the continuation; `WebSocketSuspendByteBufTest` in test-suite-kotlin covers a delayed read and a failing handler (`e9b9ac1e73`).

- @dstepanov: content released before the asynchronously published `WebSocketMessageProcessedEvent` reached listeners — release now queued behind it (`34553c0f6e`).
- @dstepanov: echoing the bound buffer via `send` failed because `send` consumes the buffer — documented (`retain()` first) rather than changing the encoder, which would leak every user-allocated buffer passed to `send`.
- Sonar: coverage on new code was 68.5%; the four added features cover the reader, conversion and listener paths.

## Codex review (gpt-5.6-sol, medium)

No findings on the first version.

### DIFF
--- a/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
+++ b/http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
@@ -34,7 +34,6 @@ import io.micronaut.http.annotation.Consumes;
 import io.micronaut.http.bind.RequestBinderRegistry;
 import io.micronaut.http.body.MessageBodyHandlerRegistry;
 import io.micronaut.http.body.MessageBodyReader;
-import io.micronaut.http.codec.CodecException;
 import io.micronaut.http.codec.MediaTypeCodecRegistry;
 import io.micronaut.http.reactive.execution.ReactiveExecutionFlow;
 import io.micronaut.http.simple.SimpleHttpHeaders;
@@ -350,6 +349,8 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
             } else {
                 Argument<?> bodyArgument = this.getBodyArgument();
                 Object data;
+                // the aggregated message content when the bound argument aliases it; released once the handler completes
+                ByteBuf handlerOwnedContent = null;
 
                 if (WebSocketFrame.class.isAssignableFrom(bodyArgument.getType())) {
                     data = msg.retain();
@@ -381,32 +382,20 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
                         content = buffer;
                     }
 
-                    data = conversionService.convert(content, ByteBuf.class, bodyArgument).orElse(null);
-                    content.release();
-                }
-
-                if (data == null) {
-                    MediaType mediaType;
+                    boolean releaseContent = true;
                     try {
-                        mediaType = messageHandler.stringValue(Consumes.class).map(MediaType::of).orElse(MediaType.APPLICATION_JSON_TYPE);
-                    } catch (IllegalArgumentException e) {
-                        exceptionCaught(ctx, e);
-                        return;
-                    }
-                    try {
-                        data = mediaTypeCodecRegistry.findCodec(mediaType)
-                            .map(codec -> codec.decode(bodyArgument, new NettyByteBufferFactory(ctx.alloc()).wrap(msg.content())))
-                            .orElse(null);
-                    } catch (CodecException e) {
+                        data = decodeMessage(ctx, content, bodyArgument, messageHandler);
+                        if (data == content || (data instanceof ByteBuffer<?> byteBuffer && byteBuffer.asNativeBuffer() == content)) {
+                            // the argument is a view of the frame content: ownership passes to the handler invocation
+                            releaseContent = false;
+                            handlerOwnedContent = content;
+                        }
+                    } catch (Throwable e) {
                         messageProcessingException(ctx, e);
                         return;
-                    }
-                    if (data == null) {
-                        MessageBodyReader<?> reader = messageBodyHandlerRegistry.findReader(bodyArgument, mediaType)
-                            .orElse(null);
-                        if (reader != null) {
-                            ByteBuffer<ByteBuf> byteBuffer = new NettyByteBufferFactory(ctx.alloc()).wrap(msg.content().retain());
-                            data = reader.read((Argument) bodyArgument, mediaType, new SimpleHttpHeaders(), byteBuffer);
+                    } finally {
+                        if (releaseContent) {
+                            content.release();
                         }
                     }
                 }
@@ -418,6 +407,7 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
                             Collections.singletonMap(bodyArgument, data)
                     );
 
+                    ByteBuf finalHandlerOwnedContent = handlerOwnedContent;
                     try {
                         BoundExecutable boundExecutable = executableBinder.bind(
                                 messageHandler.getExecutableMethod(),
@@ -427,13 +417,22 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
 
                         Object finalData = data;
                         invokeExecutable(boundExecutable, messageHandler).onComplete((v, e) -> {
-                            if (e == null) {
-                                messageHandled(ctx, finalData);
-                            } else {
-                                messageProcessingException(ctx, e);
+                            try {
+                                if (e == null) {
+                                    messageHandled(ctx, finalData);
+                                } else {
+                                    messageProcessingException(ctx, e);
+                                }
+                            } finally {
+                                if (finalHandlerOwnedContent != null) {
+                                    finalHandlerOwnedContent.release();
+                                }
                             }
                         });
                     } catch (Throwable e) {
+                        if (finalHandlerOwnedContent != null) {
+                            finalHandlerOwnedContent.release();
+                        }
                         messageProcessingException(ctx, e);
                     }
 
@@ -492,6 +491,50 @@ public abstract class AbstractNettyWebSocketHandler extends SimpleChannelInbound
                     ctx,
                     CloseReason.UNSUPPORTED_DATA
             );
+        }
+    }
+
+    /**
+     * Decodes the aggregated content of a message into the type of the body argument. The caller keeps ownership
+     * of {@code content}; the returned value may alias it (see {@link #handleWebSocketFrame}).
+     *
+     * @param ctx            The context
+     * @param content        The aggregated message content
+     * @param bodyArgument   The body argument
+     * @param messageHandler The message handler
+     * @return The decoded message, or {@code null} if no converter, codec or reader can decode it
+     */
+    @Nullable
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private Object decodeMessage(ChannelHandlerContext ctx, ByteBuf content, Argument<?> bodyArgument, MethodExecutionHandle<?, ?> messageHandler) {
+        if (bodyArgument.getType().isInstance(content)) {
+            // CompositeByteBuf is Iterable, which the conversion service does not pass through unchanged
+            return content;
+        }
+        Object data = conversionService.convert(content, ByteBuf.class, bodyArgument).orElse(null);
+        if (data != null) {
+            return data;
+        }
+        MediaType mediaType = messageHandler.stringValue(Consumes.class).map(MediaType::of).orElse(MediaType.APPLICATION_JSON_TYPE);
+        NettyByteBufferFactory bufferFactory = new NettyByteBufferFactory(ctx.alloc());
+        data = mediaTypeCodecRegistry.findCodec(mediaType)
+            .map(codec -> codec.decode(bodyArgument, bufferFactory.wrap(content)))
+            .orElse(null);
+        if (data != null) {
+            return data;
+        }
+        MessageBodyReader<?> reader = messageBodyHandlerRegistry.findReader(bodyArgument, mediaType).orElse(null);
+        if (reader == null) {
+            return null;
+        }
+        // the reader takes ownership of the buffer it is given and releases it once the message has been read
+        ByteBuffer<ByteBuf> byteBuffer = bufferFactory.wrap(content.retain());
+        try {
+            return reader.read((Argument) bodyArgument, mediaType, new SimpleHttpHeaders(), byteBuffer);
+        } catch (Throwable e) {
+            // readers only release on success
+            content.release();
+            throw e;
         }
     }
 
--- a/src/main/docs/guide/httpServer/websocket/websocketServer.adoc
+++ b/src/main/docs/guide/httpServer/websocket/websocketServer.adoc
@@ -25,7 +25,7 @@ The ann:websocket.annotation.OnMessage[] method can define a parameter for the m
 
 * A Netty `WebSocketFrame`
 * Any Java primitive or simple type (such as `String`). In fact, any type that can be converted from `ByteBuf` (you can register additional api:core.convert.TypeConverter[] beans to support a custom type).
-* A `byte[]`, a `ByteBuf` or a Java NIO `ByteBuffer`.
+* A `byte[]`, a `ByteBuf` or a Java NIO `ByteBuffer`. A `ByteBuf` (or api:core.io.buffer.ByteBuffer[]) parameter is a view of the received frame content and is released once the `@OnMessage` method (and any publisher it returns) completes; call `retain()` or copy the data if you need it beyond that point.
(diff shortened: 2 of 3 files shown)


dstepanov (commented):
The fix works: I ran `WebSocketFrameOwnershipSpec` on the head and it passes, and the ownership hand-off, the reader error path (retain, then release twice) and the fragmented POJO decode all look right to me. None of the points below are regressions. I checked each one against the 5.1.x handler with the same reproducer and got the same result. They are places where the contract in the new doc line ("released once the `@OnMessage` method completes") still doesn't hold.

Out of scope, but related: the guide still lists a Java NIO `ByteBuffer` parameter as supported. No `ByteBuf -> java.nio.ByteBuffer` converter is registered, so it falls through to the JSON codec, and binary data fails with `CodecException: Illegal character ((CTRL-CHAR, code 1))`. That is the same on 5.1.x.

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -427,13 +417,22 @@ protected void handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame ms
...
-                            } else {
-                                messageProcessingException(ctx, e);
+                            try {
+                                if (e == null) {
+                                    messageHandled(ctx, finalData);
+                                } else {
+                                    messageProcessingException(ctx, e);
+                                }
+                            } finally {
+                                if (finalHandlerOwnedContent != null) {
```
dstepanov, line 421: `NettyServerWebSocketHandler.messageHandled` does not publish `WebSocketMessageProcessedEvent` inline. It only schedules it with `ctx.executor().execute(...)`, so this `finally` always releases the content before any listener runs. With `@OnMessage void onMessage(ByteBuf m)` and an `ApplicationEventListener<WebSocketMessageProcessedEvent<?>>` that reads `event.message`, the listener gets `IllegalReferenceCountException: refCnt: 0` for every message (reproduced on this head). Could the release move into the scheduled task, for example by retaining around the publish or by passing the release into `messageHandled`?

src/main/docs/guide/httpServer/websocket/websocketServer.adoc
```diff
@@ -25,7 +25,7 @@ The ann:websocket.annotation.OnMessage[] method can define a parameter for the m
 
 * A Netty `WebSocketFrame`
 * Any Java primitive or simple type (such as `String`). In fact, any type that can be converted from `ByteBuf` (you can register additional api:core.convert.TypeConverter[] beans to support a custom type).
-* A `byte[]`, a `ByteBuf` or a Java NIO `ByteBuffer`.
+* A `byte[]`, a `ByteBuf` or a Java NIO `ByteBuffer`. A `ByteBuf` (or api:core.io.buffer.ByteBuffer[]) parameter is a view of the received frame content and is released once the `@OnMessage` method (and any publisher it returns) completes; call `retain()` or copy the data if you need it beyond that point.
```
dstepanov: Echoing the bound buffer is probably the most common way to use it, and it still fails. `WebSocketMessageEncoder` sends a `ByteBuf` as `buf.slice()` without retaining it, so the write releases the frame content, and the handler-owned release then goes below zero. `Publisher<ByteBuf> onMessage(ByteBuf m, WebSocketSession s) { return s.send(m) }` sends `IllegalReferenceCountException: refCnt: 0, decrement: 1` to `@OnError` on every message. Either say here that you must `retain()` before passing the buffer to `send`, or have the encoder retain the slice (`retainedSlice()`) now that the handler owns a reference.

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
graemerocher: Fixed in 34553c0f6e: after `messageHandled` the release is queued on `ctx.executor()`, i.e. behind the task the server handler uses to publish `WebSocketMessageProcessedEvent` (inline only if the executor rejects it). New spec feature `message processed event listeners see the bound ByteBuf`: a listener reads the message bytes; it fails with the refCnt error on the previous head and passes now.

src/main/docs/guide/httpServer/websocket/websocketServer.adoc
graemerocher: Went with documenting it (34553c0f6e): changing the encoder to `retainedSlice()` would silently start leaking every user-allocated buffer passed to `send`, which has always been consumed by the write. The guide now says `send` takes ownership and to echo the bound buffer as `session.send(buffer.retain())`, and there is a spec feature doing exactly that. Also dropped the java.nio `ByteBuffer` claim from the guide, as you noted it never worked.

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -495,6 +492,103 @@ protected void handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame ms
...
+     * {@link #messageHandled} hands the message to have run. Implementations publish the
+     * {@code WebSocketMessageProcessedEvent} from a task on the channel's executor rather than
+     * inline, so the release is queued on that executor behind it.
+     *
+     * @param ctx     The context
+     * @param content The content owned by the handler invocation
+     */
+    private static void releaseAfterListeners(ChannelHandlerContext ctx, ByteBuf content) {
+        try {
+            ctx.executor().execute(content::release);
```
yawkat: this is relatively expensive, not sure it's a good idea. also it relies on how listeners work

graemerocher: Agreed, replaced in 15e006e889: no second task and no ordering assumption. `messageHandled` now takes the release as a `Runnable`; the server handler runs it in a `finally` inside the task that publishes the event, after the listeners have seen the message, and the default (`http-client`, the no-op) runs it right after the two-argument `messageHandled`. The listener spec feature still passes.

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -426,14 +407,23 @@ protected void handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame ms
                         );
 
                         Object finalData = data;
+                        Runnable release = finalHandlerOwnedContent == null ? () -> { } : finalHandlerOwnedContent::release;
                         invokeExecutable(boundExecutable, messageHandler).onComplete((v, e) -> {
                             if (e == null) {
-                                messageHandled(ctx, finalData);
+                                // the message may still be handed to listeners, messageHandled releases once that is done
+                                messageHandled(ctx, finalData, release);
```
dstepanov: This does not hold for Kotlin `suspend` handlers. In `NettyServerWebSocketHandler.invokeExecutable`, the suspend branch passes the return value of `invokeExecutable0` to `isKotlinCoroutineSuspended`. That value is an `ExecutionFlow`, never `COROUTINE_SUSPENDED` (the check is `==`), so the branch always returns `ExecutionFlow.empty()`, which completes at once. As a result, `suspend fun onMessage(buf: ByteBuf)` has its content released before the handler runs on its executor, or before it resumes after a suspension. That is the same `IllegalReferenceCountException` this PR fixes for other handlers.

This predates the PR (since #12439), and it also means the processed event fires early for suspend handlers and their errors never reach `@OnError`. Either exclude suspend handlers in the guide or follow up by making that branch return the flow from `invokeExecutable0`.

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
```diff
@@ -397,7 +397,7 @@ private Supplier<?> invokeWithContext(BoundExecutable boundExecutable, MethodExe
     }
 
     @Override
-    protected void messageHandled(ChannelHandlerContext ctx, Object message) {
+    protected void messageHandled(ChannelHandlerContext ctx, Object message, Runnable release) {
```
dstepanov: Minor: if `ctx.executor().execute(...)` rejects the task (event loop shutting down), `release` never runs and the content leaks. Before this PR the content was released at bind time, so this leak is new, though it only happens at shutdown. Catching `RejectedExecutionException` and running `release` would close it.

dstepanov (approved):
15e006e889 settles both threads: the release now runs in the task that publishes `WebSocketMessageProcessedEvent`, after the listeners, with no second task. The client handler gets the same through the default three-argument `messageHandled`. I traced every release path (decode failure, `bind` throwing, handler error, reader failure, nothing able to decode) and each releases exactly once. The ownership check covers every `ByteBuf` converter: the `ByteBuffer` wrapper hands back the same buffer, `String`, `CharSequence` and `byte[]` copy, and `Object`/`ReferenceCounted` hit the `isInstance` shortcut.

Two notes inline. The Kotlin `suspend` one predates this PR, so it can be a follow-up.

One behaviour change worth a line in the description: every decode failure now goes to `@OnError`. Before, a bad `@Consumes` media type went to `exceptionCaught`, and non-`CodecException` errors from a codec or reader propagated up the pipeline. The new behaviour is better.

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
graemerocher: Fixed in e9b9ac1e73 rather than documented away: the coroutine check now runs inside the async flow on the invocation's own result, and the flow completes with the continuation's future, so a suspend handler keeps its `ByteBuf` across suspensions, the processed event fires when it is done and its failure reaches `@OnError`. `WebSocketSuspendByteBufTest` (test-suite-kotlin) covers both and times out on the previous head.

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
graemerocher: Done in e9b9ac1e73: `RejectedExecutionException` from the executor runs `release` directly.

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
```diff
@@ -524,20 +517,18 @@ private DecodedMessage decodeContent(ChannelHandlerContext ctx, ByteBuf content,
...
+     * @param executor   The executor
+     * @param task       The task
+     * @param onRejected What to run instead when the executor rejects the task
      */
-    private static void releaseAfterListeners(ChannelHandlerContext ctx, ByteBuf content) {
+    protected static void executeOrElse(EventExecutor executor, Runnable task, Runnable onRejected) {
         try {
-            ctx.executor().execute(content::release);
+            executor.execute(task);
         } catch (RejectedExecutionException e) {
```
yawkat: i dont think we need to handle this, and it makes the code too complicated

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
```diff
@@ -397,17 +394,21 @@ private Supplier<?> invokeWithContext(BoundExecutable boundExecutable, MethodExe
     }
 
     @Override
-    protected void messageHandled(ChannelHandlerContext ctx, Object message) {
-        ctx.executor().execute(() -> {
+    protected void messageHandled(ChannelHandlerContext ctx, Object message, Runnable release) {
+        // if the event loop is shutting down no listeners will run, but the content must not leak
+        executeOrElse(ctx.executor(), () -> {
```
yawkat: i think this is a hot path so i would like to avoid an executor.execute. im still not sure what problem this is supposed to solve

http-netty/src/main/java/io/micronaut/http/netty/websocket/AbstractNettyWebSocketHandler.java
graemerocher: Gone in 833a606239, along with its spec: with the event published inline there is no task to be rejected.

http-server-netty/src/main/java/io/micronaut/http/server/netty/websocket/NettyServerWebSocketHandler.java
graemerocher: The `execute` was pre-existing (the event has always been published from a task on the channel's executor); what this PR had to solve is that once the handler invocation owns the frame content, releasing it in the completion callback freed the `ByteBuf` before a `WebSocketMessageProcessedEvent` listener on that task could read it (dstepanov's reproducer: every message gave the listener `refCnt: 0`). Passing the release into the task was the fix, but you're right that the hop itself isn't needed: in 833a606239 the event is published inline in the completing thread and the callback releases in a `finally` afterwards, so the per-message path has no `execute` at all and `messageHandled` is back to two arguments. The listener spec feature still passes.

PR #13155, opened 2026-09-12
Carry scheduled invocations as an interceptor chain attribute

dstepanov:
Lets an interceptor tell a `@Scheduled` method fired by the scheduler apart from a direct call of the same method by application code. The interceptor chain of a scheduled firing carries a `ScheduledInvocation` attribute; nothing else does.

### Why

Jakarta Interceptors separates business methods (`@AroundInvoke`) from timeout methods (`@AroundTimeout`), and `InvocationContext.getTimer()` is only non-null for a timeout invocation. micronaut-jakarta-interceptors uses `@Scheduled` as its timer service, but it can only flag the *method* as a timeout method at compile time, so a direct application call to a `@Scheduled` method is also run through `@AroundTimeout` advice and shown a timer. Core had nothing at the invocation level to distinguish the two.

### What

- `io.micronaut.aop.ScheduledInvocation` is the attribute value, stored under `ScheduledInvocation.ATTRIBUTE` in `InvocationContext.getAttributes()`. It exposes the fired `ExecutableMethod` and `getSchedule()`, the `@Scheduled` annotation that fired (evaluated, so expressions are resolved). A method with several `@Scheduled` annotations is fired once per schedule, and each firing carries its own annotation, which gives Jakarta's `Timer` something to describe.
- Interceptors read it with `ScheduledInvocation.find(context)`, which is empty for every other invocation. This is the direct counterpart of a null `getTimer()`, and it adds nothing to `MethodInvocationContext` itself.
- `ScheduledInvocation.invoke(method, schedule, target, args...)` fires a method as a scheduled task. `ScheduledMethodProcessor` now goes through it, and other timer integrations can use it too.
- `MethodInterceptorChain` puts the value into its attributes when it is built. Chains without it never allocate the attribute map.
- Documented in the Scheduled Tasks page (new section "Intercepting Scheduled Invocations") and cross-referenced from Around Advice, with a note that Jakarta-style around-timeout advice can rely on it.

### How the attribute gets into the chain

The scheduler invokes the bean's `ExecutableMethod` on the bean instance, which is the AOP proxy, so the chain is built inside the generated proxy method and the scheduler cannot hand it an attribute directly. `ScheduledInvocation.invoke` therefore binds a `ScopedValue` around the call (core targets JDK 25, where `ScopedValue` is final), and the chain constructor claims it:

- The claim matches the first chain built for the scheduled method: same method name and argument types, declared by a type the target is an instance of. The instance check rather than declaring-type identity makes this work for both subclass proxies (target is the proxy) and `proxyTarget = true` proxies (target is the real bean), and it excludes a same-named method of an unrelated bean.
- The value is claimed once, so a recursive call of the scheduled method and every other intercepted method the scheduled method calls in turn carry no attribute.
- A `ScopedValue` is not inherited by other threads, so a direct call to the same method on the same bean, on any thread, never carries it. When nothing is bound, the claim is a single `isBound()` check per chain construction.

No generated code changes, so this works for beans compiled by earlier versions. The change is additive and binary compatible: one new class and a private constructor on `MethodInterceptorChain`.

### Tests

- `context`: `ScheduledInvocationMarkerSpec` runs `@Scheduled(fixedDelay)` beans with `@Around` advice (subclass proxy and `proxyTarget = true`). It asserts the scheduler's firings carry the attribute with the right method and schedule, a method with two schedules sees both, the calls the task makes in turn carry nothing, and direct calls from the test thread carry nothing.
- `inject-java`: `ScheduledInvocationSpec` drives `ScheduledInvocation.invoke` deterministically and checks the carried method and schedule, the claim-once behaviour (recursion, a same-named method on another bean), the `proxyTarget` case, and that the value stays on the calling thread.

Full `context` and `inject-java` test suites and checkstyle pass locally.

### DIFF
--- a/aop/src/main/java/io/micronaut/aop/ScheduledInvocation.java
+++ b/aop/src/main/java/io/micronaut/aop/ScheduledInvocation.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.aop;
+
+import io.micronaut.core.annotation.Internal;
+import io.micronaut.inject.ExecutableMethod;
+import org.jspecify.annotations.Nullable;
+
+import java.util.Arrays;
+
+/**
+ * Invokes a method as a scheduled task, so that interceptors can tell the invocation apart from a direct call.
+ *
+ * <p>The scheduler fires a {@code @Scheduled} method through {@link #invoke(ExecutableMethod, Object, Object...)}.
+ * The invocation carries a thread-bound marker which the interceptor chain built for that method claims, and
+ * {@link MethodInvocationContext#isScheduled()} then returns {@code true} for every interceptor in that chain.
+ * A direct call to the same method from application code carries no marker, and neither do methods the scheduled
+ * method calls in turn: the marker is bound to one method and is claimed by the first chain built for it, so
+ * only the scheduled invocation itself observes it.</p>
+ *
+ * <p>Other schedulers or timer services can use the same entry point to fire methods, and interceptors that apply
+ * timer-specific advice, such as Jakarta Interceptors' around-timeout methods, can rely on the same flag.</p>
+ *
+ * @author Denis Stepanov
+ * @since 5.2.0
+ */
+public final class ScheduledInvocation {
+
+    private static final ScopedValue<ScheduledInvocation> CURRENT = ScopedValue.newInstance();
+
+    private final ExecutableMethod<?, ?> method;
+    private boolean claimed;
+
+    private ScheduledInvocation(ExecutableMethod<?, ?> method) {
+        this.method = method;
+    }
+
+    /**
+     * Invokes the method on the target as a scheduled task.
+     *
+     * <p>Interceptors of the method see {@link MethodInvocationContext#isScheduled()} return {@code true}.
+     * The marker applies only to the invocation of this method and only on the calling thread. Methods the
+     * invocation calls in turn, including a recursive call of the same method, are not marked.</p>
+     *
+     * @param method    The method to invoke
+     * @param target    The target bean, which may be an AOP proxy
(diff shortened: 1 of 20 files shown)


aop/src/main/java/io/micronaut/aop/ScheduledInvocation.java
```diff
@@ -0,0 +1,157 @@
...
+    public static @Nullable ScheduledInvocation claim(Object target, ExecutableMethod<?, ?> method) {
+        if (!CURRENT.isBound()) {
+            return null;
+        }
+        ScheduledInvocation current = CURRENT.get();
+        if (current.claimed) {
+            return null;
+        }
+        ExecutableMethod<?, ?> scheduled = current.method;
+        if (scheduled == method || (scheduled.getMethodName().equals(method.getMethodName())
```
graemerocher, line 144: [P1] Match the target invocation, not any invocation of a compatible type. This condition never records or compares the `target` passed to `invoke` (and the identity fast path ignores the chain target entirely). If `invoke` is given an unproxied instance, its method builds no chain; when that method calls the same executable on a different proxied instance, the peer's chain claims this value and `find` incorrectly reports that nested call as the scheduled firing. I reproduced this with a raw `PeerMarkedBean` delegating `run()` to the context's proxied `PeerMarkedBean`: the only intercepted record is `PeerMarkedBean.run:true`, where the contract requires `false`. Please retain the expected target and require identity, explicitly accounting for the real target behind a `proxyTarget` proxy, and add this regression case.

aop/src/main/java/io/micronaut/aop/ScheduledInvocation.java
```diff
@@ -0,0 +1,157 @@
...
+ * <p>The attribute is present only on the chain of the scheduled method itself. A direct call of the method from
+ * application code, a recursive call, and every method the scheduled invocation calls in turn do not carry it.
+ * Interceptors that apply timer-specific advice, such as Jakarta Interceptors' around-timeout methods, can use it
+ * where Jakarta Interceptors use a non-null timer.</p>
+ *
+ * <p>Code that fires methods on its own schedule, such as a timer service integration, marks its invocations the
+ * same way with {@link #invoke(ExecutableMethod, AnnotationValue, Object, Object...)}.</p>
+ *
+ * @author Denis Stepanov
+ * @since 5.2.0
```
graemerocher, line 49: [P2] Record the release in which this API actually debuts. `5.2.0` is already released, while the current `5.2.x` target declares `projectVersion=5.2.3-SNAPSHOT` (the PR's merge-base snapshot was already 5.2.2). This should be `@since 5.2.3` if the project accepts this stable additive API in a patch release; otherwise the SemVer-compatible option is to target the next minor release and use that version.

graemerocher (commented):
I found one invocation-correctness defect and one public API versioning issue. The supplied deterministic and live-scheduler specs pass, as do checkstyle and `:micronaut-aop:japiCmp`; an additional same-type delegation regression test reproduces the incorrect marker claim.

PR #13158, opened 2026-09-12
ReadTimeoutException to allow fresh, per-failure instances:

### LINKED ISSUE #12655: NettyHttpClient ReadTimeouts don't supply stacktrace
### Expected Behavior

A`@GET` function on a `@Client` should show a stacktrace pointing to the function invocation even when a ReadTimeout occured.

### Actual Behaviour

A stacktrace is formed without the invocation of the `@GET` method.

### Steps To Reproduce

Reproducer repo: https://github.com/KDanisme/micronaut-issue/tree/client-stacktrace
Simply run this projectd and observe the lack of a proper stacktrace

### Environment Information

_No response_

### Example Application

https://github.com/KDanisme/micronaut-issue/tree/client-stacktrace

### Version

4 and 5

pandareen:
Root cause:
`TIMEOUT_EXCEPTION` is a static singleton, so its stack trace is captured once at class-initialization time (not at the timeout site), and it's shared across all concurrent requests. It also can't be decorated with a service ID because it's locked as shared.

Fixes: #12655

### DIFF
--- a/http-client-core/src/main/java/io/micronaut/http/client/exceptions/ReadTimeoutException.java
+++ b/http-client-core/src/main/java/io/micronaut/http/client/exceptions/ReadTimeoutException.java
@@ -23,9 +23,26 @@ package io.micronaut.http.client.exceptions;
  */
 public final class ReadTimeoutException extends HttpClientException {
 
-    public static final ReadTimeoutException TIMEOUT_EXCEPTION = new ReadTimeoutException();
+    /**
+     * Shared instance.
+     *
+     * @deprecated This shared instance captures its stack trace at static initialization time
+     * and is reused across concurrent requests. Use {@link #ReadTimeoutException()} instead so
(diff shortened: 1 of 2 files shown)


pandareen:
@graemerocher could you please review this ?

yawkat:
The exception is static for performance reasons.

yawkat:
also, does this even fix #12655? The stack trace might be on the event loop and not include the caller like requested.

pandareen:
@yawkat 
> also, does this even fix #12655? The stack trace might be on the event loop and not include the caller like requested.

On your two challenges

1. Performance (static): valid in general, but this is the error path of a request that already spent seconds timing out — one allocation is negligible, and every other client exception on this path is already allocated fresh.
2. "Does it even fix #12655?": You were right — the fresh-instance change alone did not. The caller-stack capture is what actually closes it.

PR #13163, opened 2026-09-12
Add getSourceAnnotations(): the annotations of an element as the source wrote them

dstepanov:
### What this does

Adds an "as written" view of an element's annotations beside the annotation metadata:

```java
// io.micronaut.inject.ast.annotation.MutableAnnotationMetadataDelegate
@Experimental
default List<AnnotationValue<?>> getSourceAnnotations() { return List.of(); }
```

Because `Element extends MutableAnnotationMetadataDelegate<Element>`, it is reachable as `element.getSourceAnnotations()`, `classElement.getTypeAnnotationMetadata().getSourceAnnotations()` and `genericElement.getGenericTypeAnnotationMetadata().getSourceAnnotations()`.

The list is the declaration as the compiler sees it, in source order:
- a repeatable annotation written once is itself, a container the source wrote is the container; two or more repetitions arrive in the container the compiler synthesizes for them (javac and Groovy both hand that over, as the class file would; Python and KSP hand over the repetitions);
- nothing an `AnnotationMapper`, `AnnotationRemapper`, `AnnotationTransformer` or a visitor's `annotate(...)` added appears, nor any stereotype;
- each value carries the interface's retention and its defaults for the members the use left out, empty strings and arrays included (`readAnnotationDefaultValues(name, type, true)`);
- for a type variable it is what was written at this use (nothing if the use wrote nothing), or the type parameter's annotations when the element comes from `getDeclaredGenericPlaceholders()` / `getDeclaredTypeVariables()`;
- empty for an element no source backs (`ClassElement.of(...)`, reflection, synthesized properties).

### Why

The CDI Lite language model (`jakarta.enterprise.lang.model`) that micronaut-cdi builds on top of the AST promises exactly this, and the metadata record answers none of it by design (repetitions are folded into their container, mapper output and stereotypes are indistinguishable from what was written, empty defaults are dropped). Today the processor reads javac's mirrors for it, which leaves KSP, Groovy and Python compilations behind.

### How

- `AbstractAnnotationMetadataBuilder` gains `protected abstract List<? extends A> getWrittenAnnotations(T element)` (javac `getAnnotationMirrors()`, not `getAnnotationsForType`, which unwraps containers; KSP `annotations`; Groovy `getAnnotations()`; Python the decorators) and a public `readSourceAnnotations(T element)` that reads the raw member values and the defaults and skips `processAnnotation` entirely. Evaluated-expression strings stay strings, validation is not re-run. The list is cached lazily on the `CachedAnnotationMetadata` entry; `AbstractElementAnnotationMetadata` delegates to it.
- javac placeholders: `JavaElementAnnotationMetadataFactory.lookupTypeAnnotationsForGenericPlaceholder` keeps its fallback to the `TypeParameterElement` for the metadata; the source view reads the `TypeVariable` mirror for a use and the type parameter for a declaration. Declarations are marked on `JavaGenericPlaceholderElement` (`isDeclaration()`, now public and `@Internal`) by `getDeclaredGenericPlaceholders()` / `getDeclaredTypeVariables()`; cache keys are unchanged.
- javac arrays: `AbstractJavaElement.newClassElement` now builds the leaf with the leaf mirror and gives each dimension its own `JavaNativeElement.Class` holding that dimension's `ArrayType` (`JavaClassElement.toArray(ArrayType)`; `withArrayDimensions(n-1)` keeps taking `getComponentType()`). The compatibility swap in `lookupTypeAnnotationsForClass` ("component annotations stand for the array's unless JSpecify") is unchanged; because the dimensions used to share the innermost array mirror, the swap now first walks to the innermost array so multi-dimensional arrays answer exactly as before. Only `getSourceAnnotations()` is exact per dimension; an array made with `toArray()` from a component reports nothing for the synthetic dimension.
- KSP: `KotlinTypeParameterNativeElement` keeps the use's `KSType` (not part of equality, so metadata stays shared); the factory reads the use's annotations for the source view.
- Groovy: the type-annotation lookups already pass a synthetic node holding the type annotations, so the view works through `getWrittenAnnotations`.
- Python: the defaults of a Java annotation used from Python, which the Python builder never read, come from the Java builder for the source view (the three-argument `readAnnotationDefaultValues`, which also serves `getAnnotationDefaultValues`); the metadata path is unchanged. A type use built by `PythonTypeResolver` gets its source view through a new `AbstractElementAnnotationMetadataFactory.buildMutable(AnnotationMetadata, Supplier)` overload, and `TypeAnnotatedClassElement` forwards the element view to its delegate. Python attaches an `Annotated[...]` on a parameter or attribute to that declaration, and to the type use only inside a type argument; the view reports it where Python put it.

Not touched, as agreed: `JavaMethodElement.returnType`'s JSpecify `@NonNull` annotate call, `MutableAnnotationMetadata.addDeclaredRepeatable`, and KSP's boxing of annotated primitives.

### Tests

`SourceAnnotationsSpec` in `inject-java`, `inject-kotlin`, `inject-groovy` and `PythonSourceAnnotationsSpec` in `inject-python-test`: repetitions, a written container, defaults with empty values, stereotypes absent, `annotate(...)` absent, meta-annotations of an annotation interface present, type-use annotations, use versus declaration of a type variable, per-dimension array annotations with the metadata unchanged. The full `:micronaut-inject-java:test` (5286), `:micronaut-inject-groovy:test` (1131), `:micronaut-inject-kotlin:test` (866), `:micronaut-inject-python-test:test` (609) and `:micronaut-inject-python:test` (125) suites pass.

### DIFF
--- a/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
+++ b/core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
@@ -29,6 +29,7 @@ import io.micronaut.core.annotation.AnnotationValueBuilder;
 import io.micronaut.core.annotation.InstantiatedMember;
 import io.micronaut.core.annotation.Internal;
 import io.micronaut.core.annotation.Retainable;
+import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 import io.micronaut.core.expressions.EvaluatedExpressionReference;
 import io.micronaut.core.io.service.SoftServiceLoader;
@@ -57,6 +58,7 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -394,7 +396,7 @@ public abstract class AbstractAnnotationMetadataBuilder<T, A> {
         CachedAnnotationMetadata cachedAnnotationMetadata = MUTATED_ANNOTATION_METADATA.get(key);
         if (cachedAnnotationMetadata == null) {
             AnnotationMetadata annotationMetadata = buildInternal(element);
-            cachedAnnotationMetadata = new DefaultCachedAnnotationMetadata(annotationMetadata);
+            cachedAnnotationMetadata = new DefaultCachedAnnotationMetadata(annotationMetadata, () -> readSourceAnnotations(element));
             // Don't use `computeIfAbsent` as it can lead to a concurrent exception because the cache is accessed during in `buildInternal`
             MUTATED_ANNOTATION_METADATA.put(key, cachedAnnotationMetadata);
         }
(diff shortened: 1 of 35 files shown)


graemerocher:
needs Python tests

core-processor/src/main/java/io/micronaut/inject/annotation/AbstractAnnotationMetadataBuilder.java
```diff
@@ -477,6 +479,81 @@ protected final AnnotationMetadata buildInternal(T element) {
...
+        T annotationType = getTypeForAnnotation(annotationMirror);
+        RetentionPolicy retentionPolicy = getRetentionPolicy(annotationType);
+        Map<? extends T, ?> elementValues = readAnnotationRawValues(annotationMirror);
+        Map<CharSequence, Object> annotationValues = CollectionUtils.newLinkedHashMap(elementValues.size());
+        for (Map.Entry<? extends T, ?> entry : elementValues.entrySet()) {
+            T member = entry.getKey();
+            if (member == null) {
+                continue;
+            }
+            readAnnotationRawValues(element, annotationName, member, getAnnotationMemberName(member), entry.getValue(), annotationValues);
```
graemerocher, line 535: [P1] Preserve source semantics recursively for nested annotations. Values converted here still use `readNestedAnnotationValue`/`resolveAnnotationDefaults`, which builds the nested `AnnotationValue` without its native retention and calls the two-argument `readAnnotationDefaultValues`. For example, `@Outer(@Nested)` where `@Nested` has `RetentionPolicy.SOURCE`, `String description() default ""`, and `String[] tags() default {}` returns a nested value with `RUNTIME` retention and neither empty default. The same happens to annotations inside a repeatable container. This violates the documented guarantee that each value carries its interface retention and all defaults. Please make the source reader recursive (including `includeEmptyValues = true` and native retention) and cover nested members/container entries.

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/annotation/GroovyAnnotationMetadataBuilder.java
```diff
@@ -304,6 +304,12 @@ protected String getElementName(AnnotatedNode element) {
         throw new IllegalArgumentException("Cannot establish name for node type: " + element.getClass().getName());
     }
 
+    @Override
+    protected List<? extends AnnotationNode> getWrittenAnnotations(AnnotatedNode element) {
```
graemerocher, line 308: [P1] Include empty array defaults in the Groovy source view. This backend does not override `readAnnotationDefaultValues(..., boolean)`, and its existing two-argument implementation only records `ConstantExpression` defaults. Consequently `String[] tags() default []` is a `ListExpression` and `getSourceAnnotations()` omits `tags`; adding that member and asserting `getDefaultValues().containsKey("tags")` makes the new Groovy spec fail. Please add the inclusive default-reading path for Groovy and a regression case for empty arrays.

inject-groovy/src/main/groovy/io/micronaut/ast/groovy/annotation/GroovyAnnotationMetadataBuilder.java
```diff
@@ -304,6 +304,12 @@ protected String getElementName(AnnotatedNode element) {
...
+        // Unlike getAnnotationsForType, a repeatable container is kept as the source wrote it
+        return element.getAnnotations();
```
graemerocher, line 310: [P1] Do not expose Groovy's reordered annotation list as source order. The new test writes the two `@MyRepeatable` annotations before `@Singleton`, but expects this method to return `[Singleton, MyRepeatables]`; that directly contradicts the new API's promise of source order. The synthesized container should occupy the repeated annotation's source position (or the public contract must be changed consistently). Please order these nodes by their source positions and update the test to expect `[MyRepeatables, Singleton]`.

graemerocher (commented):
I found three correctness gaps in the new source-annotation contract. The four focused PR specs pass, but additional probes for nested annotation semantics and Groovy empty-array defaults fail; the Groovy spec also currently encodes an order that contradicts the API documentation.

PR #13167, opened 2026-09-12
Report annotation member defaults for Python as for the other languages

dstepanov:
Follows #13160, which settled the annotation member defaults contract for Java, Kotlin and Groovy and deliberately left Python out of scope. Two things were wrong there.

### An enum-constant default on a typed decorator member generated uncompilable Java

`extract_arg_defaults` hand-rolled its value conversion — `ast.literal_eval`, a special case for `ast.Name`, then `ast.dump` for everything else — while every member value given at a *usage* site goes through `convert_ast_value`, which understands attribute chains, lists, tuples, dicts and nested decorator calls. So an enum-constant default reached the stub generator as the AST repr and was written into the generated annotation type verbatim:

```
Pyronaut processing failed: incompatible types: java.lang.String cannot be converted to python.Colour
  Colour enumValue() default "Attribute(value=Name(id='Colour', ctx=Load()), attr='GREEN', ctx=Load())";
```

`extract_arg_defaults` now takes the visitor and delegates to `convert_ast_value`, and `PythonAnnotationStubGenerator` renders a default that has to be a reference rather than a literal as one — an enum constant as `Type.CONSTANT`, a `Class` member as `Type.class` — in scalar and in array position, matching what a usage-site value already produced. A scalar `type` member is now typed as `Class`, as a `list[type]` member already was.

Resolving the enum type through `getRawClassElement`, the way the usage path does, is not an option on the defaults path: the member type is a `PythonEnumElement`, which has no copy constructor, so the `TypeDef` the member was declared with is used directly.

### Python did not implement the empty-string distinction

`PythonAnnotationMetadataBuilder` only overrode the two-argument `readAnnotationDefaultValues`, so it always reported empty string defaults. It now implements the three-argument form and applies the rule documented on `JavaAnnotationMetadataBuilder.isValidDefaultValue`: written bean definition metadata omits empty *string* defaults, `VisitorContext.getAnnotationDefaultValues` includes them, and an empty list default is always kept. A decorator parameter with no default at all arrives as a null member value and is not a default in either mode; an explicitly set empty string is a usage value and is kept as before (`UntypedDecoratorMemberSpec` still passes).

`IntroductionAdviceSpec` asserted that an introduction returning the annotation value yields the empty string — that omission observed through the advice. Its decorator now declares a non-empty default, so the test keeps asserting what it is about.

### Polyglot list views leaking into metadata

A Python list reaching Java as a `java.util.List` is a view onto the guest object; reading it after the compilation context closes fails with `IllegalStateException: The Context is already closed`. `PythonValues` converted a polyglot `Value` but passed such a view through, so a `list` member default was written into annotation metadata as a live `PolyglotList`. It is now copied out eagerly, along with the equivalent `Map` view.

### Tests

`inject-python-test` gains `AnnotationMemberDefaultsSpec`, the Python sibling of the spec #13160 added to the three JVM suites, sharing its expectation literal for the visitor-context read. Two shapes still differ on the *written metadata* read, both because the Python class is visited before its generated Java annotation type exists, so member types are not yet known: a class value is written as the class name rather than an `AnnotationClassValue`, and a list value as a `List` rather than an array. A member value given at a usage site is written the same way, so the defaults are not a special case; the spec documents and asserts this.

`./gradlew :micronaut-inject-python:check :micronaut-inject-python-test:check :micronaut-context-python:check :micronaut-context-python-netty:check :test-suite-python:check -Ppython-ci --continue` passes.

### DIFF
--- a/inject-python-test/src/test/groovy/io/micronaut/python/annotation/processing/test/AnnotationMemberDefaultsSpec.groovy
+++ b/inject-python-test/src/test/groovy/io/micronaut/python/annotation/processing/test/AnnotationMemberDefaultsSpec.groovy
@@ -1,0 +1,224 @@
+/*
+ * Copyright 2017-2026 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.python.annotation.processing.test
+
+import io.micronaut.core.annotation.AnnotationClassValue
+import io.micronaut.core.annotation.AnnotationValue
+import io.micronaut.inject.ast.ClassElement
+import io.micronaut.inject.visitor.TypeElementVisitor
+import io.micronaut.inject.visitor.VisitorContext
+
+/**
+ * Asserts that annotation member defaults of every kind declared on a Python decorator are reported identically to
+ * the three JVM languages. The sibling copies of this spec live in the inject-java, inject-kotlin and inject-groovy
+ * test suites and share the expectation literals below.
+ */
+class AnnotationMemberDefaultsSpec extends AbstractPythonTypeElementSpec {
+
+    private void compileTestSource() {
+        buildBeanDefinition('python', 'Test', '''
+from enum import Enum
+from jakarta.inject import Singleton
+
+def micronaut_annotation(name, repeated=None, annotationTypeTarget=False):
+    def decorator(func):
+        return func
+    return decorator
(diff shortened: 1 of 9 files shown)


graemerocher (approved):

dstepanov:
Closing in favour of #13160, which now carries this work. Graeme asked for the Python support to be folded into that PR, and it retargets 5.2.x as this one does.

Merged in from here:
- the `PythonAnnotationStubGenerator` rendering of class and enum defaults as `Type.class` / `Type.CONSTANT`, including arrays, and the `type` → `Class` member mapping
- `isValidDefaultValue` skipping the `null` member value of a decorator parameter with no default
- `PythonValues.toJava` copying maps as well as lists
- `extract_arg_defaults` calling `convert_ast_value` directly
- the stronger `AnnotationMemberDefaultsSpec`, `DecoratorMemberDefaultSpec`, and the `test_micronaut_processor.py` unit test

Kept from #13160 where the two differed:
- the `resolveAnnotationMirror` fallback to the decorator registry, so `getAnnotationDefaultValues` answers while the declaring Python class is visited, not only once its generated Java type exists
- the guard dropping a default the processor can only render as an AST dump
- `IntroductionAdviceSpec` keeping the empty-string default and asserting `None`, rather than changing the default to `"stubbed"`, so the behaviour change stays covered

PR #13175, opened 2026-09-13
Describe a java.lang.reflect.Proxy through the interfaces it stands for

dstepanov:
## Summary

`ReflectionBeanIntrospection` now describes a `java.lang.reflect.Proxy` class through the interfaces it implements. Before, it refused a proxy class outright.

## Why

A proxy class gets a name the JDK generates, `jdk.proxyN.$ProxyN`. `ReflectionBeanIntrospection.isIntrospectable` skips everything in `jdk.*`, so a proxy got no introspection at all. Validating a proxied bean through micronaut-validation's reflective introspector then failed with:

```
Bean introspection not found for the class: class jdk.proxy2.$Proxy97
```

Hibernate Validator 9.1.3's own `ProxyTest.testValidateB` hit this: the proxy was of `B extends A`, with the constraints declared on `A`'s getters.

Collecting the getters an interface inherits from the interfaces it extends already worked; `ReflectionBeanIntrospection.of(B.class)` reported `A`'s properties and constraints. Only the proxy class itself was refused.

## What changed

- **`isIntrospectable`**: a proxy class is accepted when at least one interface it implements would be. A proxy of only JDK interfaces, such as `List`, is still refused.
- **Properties and methods**: a proxy is described only through its interfaces and the ones they extend. The proxy class's own methods are generated, carry no annotations, and would report `$ProxyN` as a member's declaring type. So every member is now the interface's declaration, with that declaration's annotations. The interfaces `java.lang.reflect.Proxy` itself implements are left out.
- **Type-level annotations**: a proxy takes the annotations of the interfaces it was created with. The proxy class has none of its own, and `@Introspected` and `@AccessorsStyle` decide what counts as a property.
- **Constructors**: a proxy class's only constructor takes its `InvocationHandler`. It is not selected, so `isBuildable()` is `false`, and `getConstructors()` does not list it. This is what the interfaces themselves report.
- **Member ranking**: `rankOf` uses the same class list, so a proxy's interfaces rank first. Ranking for ordinary classes and interfaces is unchanged.

## Reviewer notes

- `ReflectionBeanIntrospectionFallback` still skips proxy classes by package name. That's intended: it only serves types matched by the application's `micronaut.introspection.allow-reflection` patterns, and a generated proxy name can't match one. Code that describes any type goes through `ReflectionBeanIntrospector`, which now handles proxies.
- `BeanProperty.getDeclaringType()` is still the introspected type, the proxy class, as the `BeanIntrospection` contract requires. The interface shows up as the declaring type of each property member and of each bean method.

## Tests

- New `ProxyIntrospectionSpec`:
  - an interface that declares nothing itself reports the properties and annotations it inherits
  - a proxy of that interface is introspectable, with the same properties and annotation values
  - a property can be read through the proxy
  - members and methods name the interface as their declaring type
  - a proxy is not buildable and lists no handler constructor
  - a proxy of the declaring interface itself is described the same way
  - `ReflectionBeanIntrospector` returns an introspection for a proxy class
  - a proxy of only JDK interfaces is still refused
  - parity: an `@Introspected` Groovy sub-interface is checked against its generated introspection
- `./gradlew :micronaut-reflection:test`: 403 tests, 3 skipped (existing pending features), none failing; checkstyle and spotless clean.
- Hibernate Validator 9.1.3's engine tests, run against micronaut-validation on a classpath using this change with the package-private annotation fix: `ProxyTest.testValidateB` and `testValidateA` pass.


graemerocher (approved):

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -149,9 +150,12 @@ private ReflectionBeanIntrospection(Class<T> beanType,
...
         this.writePrefixes = prefixes(annotationMetadata, "writePrefixes", AccessorsStyle.DEFAULT_WRITE_PREFIX);
-        // a static @Creator factory is the instantiation route the processors select first, before any constructor
-        this.factory = selectFactoryMethod(beanType);
-        this.constructor = factory == null ? selectConstructor(beanType) : null;
+        // a static @Creator factory is the instantiation route the processors select first, before any
+        // constructor; a proxy declares neither - the constructor it carries takes its InvocationHandler and
+        // builds nothing a caller asked for - so it is not buildable, as the interfaces it stands for are not
+        boolean proxy = Proxy.isProxyClass(beanType);
+        this.factory = proxy ? null : selectFactoryMethod(beanType);
+        this.constructor = factory == null && !proxy ? selectConstructor(beanType) : null;
```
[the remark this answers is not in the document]
dstepanov: Builder support is now skipped for a proxy (`bebdf42e6c`): the builder method is a static method of the interface, not of the proxy class. The new spec case "a proxy is not built through the builder its interface names" covers it.

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -941,7 +1003,7 @@ private static String decapitalize(String name) {
 
     private List<BeanMethod<T, Object>> discoverMethods() {
         Map<String, Method> bySignature = new LinkedHashMap<>();
-        for (Class<?> type = beanType; type != null && type != Object.class; type = type.getSuperclass()) {
+        for (Class<?> type : classHierarchy(beanType)) {
```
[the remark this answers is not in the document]
dstepanov: `findDeclaredMethod` now returns empty for a proxy before the `getDeclaredMethod` fallback (`bebdf42e6c`), so the generated `$ProxyN` method is never reported. Covered by "a proxy declares no method of its own".

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -171,6 +175,19 @@ private ReflectionBeanIntrospection(Class<T> beanType,
         this.builderSupport = ReflectionIntrospectionBuilder.Support.of(beanType, annotationMetadata);
     }
 
+    /**
+     * The annotations of the described type: the ones the class declares, or, for a {@link Proxy}, the ones
+     * the interfaces it was created with declare. A proxy class carries no annotation of its own, and
+     * describing it as anything but the interfaces it stands for would lose what they say of themselves -
+     * {@link Introspected#accessKind()} and {@link AccessorsStyle} among it, which decide what a property is.
```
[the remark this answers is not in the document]
dstepanov: Reworded to "including `@Introspected#accessKind()` and `@AccessorsStyle`" (`bebdf42e6c`).

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -697,26 +736,40 @@ private List<BeanProperty<T, Object>> discoverProperties() {
...
         // properties, and whose default methods a type inherits as accessors of its own
+        // a proxy has no class of its own whose declarations the pass above reads: the interfaces are its
+        // declarations, so an @Introspected.Property there makes an accessor as it does on the interface itself
+        boolean proxy = Proxy.isProxyClass(beanType);
         for (Class<?> anInterface : allInterfaces(beanType)) {
             for (Method method : anInterface.getDeclaredMethods()) {
                 if (Modifier.isStatic(method.getModifiers()) || method.isSynthetic() || isGroovyObjectMethod(method)) {
                     continue;
                 }
+                Introspected.Property declared = proxy ? method.getAnnotation(Introspected.Property.class) : null;
```
graemerocher, line 747: Gating this on `proxy` makes `@Introspected.Property` on an *inherited* interface accessor visible only through a proxy, which breaks the invariant the PR description states ("a proxy is described only through its interfaces and the ones they extend… every member is the interface's declaration").

```java
interface A { @Introspected.Property @Tag("t") String value(); }
interface B extends A { }
```

- `ReflectionBeanIntrospection.of(B).getBeanProperties()` → empty. `B` declares nothing, so `classHierarchy` contributes nothing, and in this pass `declared` is `null`, so `value()` carries no read prefix and is dropped.
- `ReflectionBeanIntrospection.of(proxyOf(B).getClass()).getBeanProperties()` → `[value]`.

The same divergence hits an ordinary class implementing `A`: the override carries no annotation of its own (method annotations are not inherited), so the class pass sees `null` too.

The tests don't catch it because `ProxiedDeclared` puts the annotation on the interface that is introspected *directly* — `classHierarchy` returns `[I]` for an interface, so the first pass picks it up — and `ProxiedContract`/`ProxiedParityContract` use prefixed getters.

Reading the annotation unconditionally here (drop the `proxy ?`) makes both sides agree, and matches what the generated introspection of `@Introspected interface B extends A` reports.

---

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -697,26 +736,40 @@ private List<BeanProperty<T, Object>> discoverProperties() {
...
                     String property = accessorProperty(method.getName(), method.getReturnType());
+                    if (property == null && declared != null) {
+                        property = method.getName();
+                    }
                     if (property != null) {
-                        candidate(candidates, property).addGetter(method);
+                        if (declared != null) {
+                            candidate(candidates, property).addGetter(method, declared);
+                        } else {
+                            candidate(candidates, property).addGetter(method);
```
graemerocher, line 757: This branch collapses: `addGetter(Method)` is `getters.add(getter)` and `addGetter(Method, Property)` is the same plus `declare(getter, declaration)`, which returns immediately when the declaration is `null`. So the whole `if/else` is just

```java
candidate(candidates, property).addGetter(method, declared);
```

Related, a few lines down the setter branch re-reads `method.getAnnotation(Introspected.Property.class)` when `declared` already holds it. Note that branch passes the annotation whether or not the type is a proxy, while the name fallback above it is proxy-gated — the same asymmetry as the comment on the `declared` assignment.

---

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -168,7 +172,22 @@ private ReflectionBeanIntrospection(Class<T> beanType,
...
+    }
+
+    /**
+     * The annotations of the described type: the ones the class declares, or, for a {@link Proxy}, the ones
+     * the interfaces it was created with declare. A proxy class carries no annotation of its own, and
+     * describing it as anything but the interfaces it stands for would lose what they say of themselves,
+     * including {@link Introspected#accessKind()} and {@link AccessorsStyle}, which decide what a property is.
+     */
+    private static AnnotationMetadata typeMetadataOf(Class<?> beanType) {
+        if (Proxy.isProxyClass(beanType)) {
```
graemerocher, line 187: `ReflectionBeanIntrospectionFallback` needs this same lookup, and the reviewer note that it "still skips proxy classes by package name" only holds for proxies of *public* interfaces.

For a package-private interface the JDK names the proxy after the interface's own package. Verified on JDK 21:

```
class=com.example.$Proxy0   isProxy=true   superclass=java.lang.reflect.Proxy
```

`isPlatformType` is false for that name, `isIntrospectable` is now true, and `ReflectionIntrospectionPolicy.patterns()` compiles `com.example.*` to `\Qcom.example.\E.*`, which matches `com.example.$Proxy0`. So an application allowing its own package by pattern does reach the fallback with a proxy class.

There, `ReflectionBeanIntrospectionFallback#configured` builds "the annotations of the class" with `ReflectionAnnotations.metadataOf(type)`, which for a proxy class is empty, so `merge` returns the configured metadata alone — and it is then handed over as the metadata that *wins* in this constructor. That is exactly the precedence the method's own javadoc says it exists to avoid:

> the annotations of the class are therefore merged over the configured ones here, rather than handed to `ReflectionBeanIntrospection#of(Class, AnnotationMetadata)` as annotations that win

Effect: a configured `@Introspected(accessKind = FIELD)` or `@AccessorsStyle` overrides what the proxied interface declares, instead of losing to it. Having `configured` call this method (widening it past `private`) fixes it.

---

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -941,7 +1024,7 @@ private static String decapitalize(String name) {
 
     private List<BeanMethod<T, Object>> discoverMethods() {
         Map<String, Method> bySignature = new LinkedHashMap<>();
-        for (Class<?> type = beanType; type != null && type != Object.class; type = type.getSuperclass()) {
+        for (Class<?> type : classHierarchy(beanType)) {
```
graemerocher, line 1027: Now that a proxy contributes nothing here, all of its methods come from the interface pass below — which filters static, synthetic, bridge and `GroovyObject` methods but not visibility, while this pass requires `Modifier.isPublic`. Interfaces have had private methods since Java 9, so the two descriptions of one interface disagree:

```java
interface Svc {
    String getName();
    private String helper() { return "x"; }
}
```

- `of(Svc).getBeanMethods()` → `[getName]`
- `of(proxyOf(Svc).getClass()).getBeanMethods()` → `[getName, helper]`, and invoking `helper` on the proxy fails.

This is pre-existing — a class implementing `Svc` already reports `helper` through the same pass — but proxies depend on that pass exclusively, so it is worth the same `!Modifier.isPublic(method.getModifiers())` guard there.

---

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -1258,24 +1341,17 @@ private static boolean isIncompatibleSetter(Class<?> readType, Method setter) {
...
-                if (type == declaringType) {
-                    return rank;
-                }
-                rank++;
-                Class<?> superclass = type.getSuperclass();
-                if (superclass == null) {
-                    break;
-                }
-                type = superclass;
+            List<Class<?>> hierarchy = classHierarchy(beanType);
```
graemerocher, line 1348: Minor: `rankOf` runs inside the comparators of `selectGetter`/`selectSetter`, so every comparison now rebuilds the class list here on top of the recursive interface closure `allInterfaces` already walked (the latter pre-dates this PR). Both lists depend only on `beanType`, so hoisting them into fields alongside the other resolved state would remove the churn.

Same thought for `Proxy.isProxyClass(beanType)`, which the change adds at six call sites — the constructor already computes it once at line 156; a `private final boolean proxy` field would be cheaper and would read better than repeating the call in `classHierarchy`, `allInterfaces`, `findDeclaredMethod` and `getConstructors`.

I did check the rewrite itself against the old loop for a class, an interface and a declaring type that is in neither list — the ranks come out identical in each case.

---

reflection/src/test/groovy/io/micronaut/reflection/ProxyIntrospectionSpec.groovy
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * A {@link Proxy} stands for the interfaces it was created with and declares nothing of its own. A
+ * specification handed such an instance - Jakarta Validation validating a proxied bean - describes its class,
+ * and must read there what those interfaces declare, including what they inherit from the interfaces they
+ * extend.
+ */
+class ProxyIntrospectionSpec extends Specification {
+
+    private static final InvocationHandler HANDLER = { Object proxy, java.lang.reflect.Method method, Object[] args ->
```
graemerocher, line 18: A proxy routes `toString`, `hashCode` and `equals` to the handler too, so this returns `Integer` 42 for all three. `toString()` then fails with a `ClassCastException`, as does `equals()`.

It passes today because nothing calls them, but the first time an assertion in this spec fails, Spock renders the condition — including the proxy — and the `ClassCastException` replaces the real failure. Worth handling the three explicitly:

```groovy
private static final InvocationHandler HANDLER = { Object proxy, java.lang.reflect.Method method, Object[] args ->
    switch (method.name) {
        case "getString" -> "proxied"
        case "toString" -> "proxy"
        case "hashCode" -> System.identityHashCode(proxy)
        case "equals" -> proxy.is(args[0])
        default -> 42
    }
}
```

---

graemerocher (commented):
Reviewed at `bebdf42`, reading the change against the rest of the module. The approach looks right: describing a proxy through `allInterfaces` alone, dropping the `InvocationHandler` constructor, and taking type-level annotations off the proxied interfaces all line up with what those interfaces report on their own. Skipping the class walk also correctly keeps `java.lang.reflect.Proxy.h` out of the properties under `FIELD` access and `Serializable` out of the interface closure — I confirmed both against a real proxy.

Six inline comments. The two worth acting on before merge:

1. `@Introspected.Property` on an accessor a proxied interface *inherits* is honored only through the proxy, so `of(B)` and `of(proxyOf(B).getClass())` disagree for the same interface. That is the one place the "a proxy is described the way its interfaces are" invariant does not hold, and the tests miss it because the annotated accessor is always on the directly introspected interface.
2. `ReflectionBeanIntrospectionFallback` does reach proxy classes — for a package-private interface the JDK names the proxy `<pkg>.$Proxy0`, which an `allow-reflection` pattern matches — and its `configured()` helper reads the proxy class's own (empty) annotations, which inverts the precedence its javadoc describes.

The rest are a redundant branch, a pre-existing visibility gap in the interface pass of `discoverMethods` that proxies now depend on, a couple of allocations in a comparator, and a test handler that will mask the first failure in the new spec.

One thing outside the diff, for context rather than this PR: `ReflectionBeanDefinition.isDefinable` still accepts a proxy class — it is not an interface or abstract, and `selectConstructor` picks its single non-private constructor, the `InvocationHandler` one. That is the opposite conclusion from the one this PR reaches about the same constructor, and its javadoc points at `isIntrospectable` as the companion check.

I also checked the `rankOf` rewrite against the old loop for classes, interfaces and unknown declaring types — the ranks are identical, so non-proxy ranking is unchanged as claimed.

---

PR #13179, opened 2026-09-13
Compile generated source files in a Groovy compilation and run the Groovy visitors after Groovy's own transforms

dstepanov:
`VisitorContext.visitGeneratedSourceFile(packageName, name, originatingElements)` now works in a Groovy compilation: the source a `TypeElementVisitor` writes is compiled in the same compilation, its classes are visited by the type element visitors like every other class, and they end up in the same output.

### The problem

Under javac the generated file goes to the `Filer` and is processed in the next round; under KSP it goes to the code generator. The Groovy compiler has no rounds, and `GroovyVisitorContext.visitGeneratedSourceFile` delegated to the output visitor, where every Groovy implementation falls back to the interface default and returns `Optional.empty()`. A visitor that generates a small annotated class as source (micronaut-cdi's `BuildCompatibleExtensionVisitor` does, for `ExtensionContextRecordHolder` and `ScannedClassesImport`) therefore worked under javac and KSP and silently produced nothing under Groovy. Micronaut's own writers never noticed because they emit bytecode through `visitClass(...)`.

### What the compiler allows (Groovy 5.0.8, the version core builds with)

- `CompilationUnit.addSource(String name, String text)` queues a `SourceUnit` while the compilation runs. `CompilationUnit.compile(int)` checks the queue after the operations of every phase (`dequeued()`), and when something was queued it rewinds the unit to `Phases.INITIALIZATION`. Every phase operation skips the sources already through the current phase (`ISourceUnitOperation`/`IPrimaryClassNodeOperation.doPhaseOperation` compare `source.phase` with `unit.phase`), so only the queued source is brought up to the current phase, after which all sources continue together. The resolver relies on exactly this: `ClassNodeResolver.tryAsScript` adds a `.groovy` script found on the class path and `ResolveVisitor.resolveToOuter` throws `Interrupt` to restart.
- The per-source phase marker is only updated by `CompilationUnit.mark()`, which runs *after* the queue check. A source queued from the middle of a phase (our transform runs in `SEMANTIC_ANALYSIS`, after the built-in operations and possibly before other global transforms) would make the whole phase run again for every source already in it, transforms included. The resolver gets away with it because it interrupts from the first operation of the phase.
- `CompilationUnit.addNewPhaseOperation(op, phase)` registered for a *later* phase is appended after every existing operation of that phase (`recordPhaseOpsInAllOtherPhases`), and `SourceUnit.gotoPhase`/`completePhase`/`isPhaseComplete` are public (`ProcessingUnit`).
- `JavaAwareCompilationUnit` (what Gradle uses) runs javac once on the first `gotoPhase(SEMANTIC_ANALYSIS)` and clears `javaSources`, so the rewind does not run javac again and generates no stub for the new source.
- Adding a parsed `ClassNode` to a `ModuleNode` instead (`ModuleNode.addClass` plus `ASTTransformationVisitor.addNewPhaseOperation`) only re-runs the annotation transforms for the class; nothing runs the resolver, `StaticImportVisitor`, `InnerClassVisitor`, `GenericsVisitor` or the `CONVERSION` visitors on it, so every such step would have to be replicated by hand. Not viable in general.

### The implementation

`GroovyGeneratedSourceFiles` (new, internal, one per `CompilationUnit`, kept as node metadata on the `CompileUnit`) hands out in-memory `GeneratedFile`s. When one is written and closed it is put on a pending list and a phase operation is registered for the *following* phase. That operation runs after every other operation of its phase, marks the phase complete for the sources in the unit exactly as `CompilationUnit.mark()` would, and only then queues the pending sources with `addSource`. The rewind then processes the generated sources alone up to that phase and everything continues together, so no transform sees an existing class twice and the generated class is visited exactly once, by Micronaut's transforms and by any other global transform. A generated source may itself generate a source (the operation runs again for the new source and repeats). `GroovyVisitorContext.visitGeneratedSourceFile` uses this whenever it has a compilation unit and otherwise keeps delegating to the output visitor. Everything else, including what the existing accessors return, is unchanged.

Limits, documented on the method:

- The content must be Groovy source, as it must be Java under javac and Kotlin under KSP (`VisitorContext.getLanguage()` says which). micronaut-cdi's generated classes use the Java `{A.class, B.class}` annotation array syntax, which Groovy parses as a closure, so the cdi visitor will have to emit `[A, B]` under Groovy.
- The sources the compilation started with are resolved before the visitors run, so they cannot reference a generated class (the reverse works). javac defers such references to a later round; the Groovy compiler cannot.
- Nothing is left in the compilation to process a source written once class generation has started (a visitor's `finish()`), so the method returns `Optional.empty()` from then on and a file obtained earlier but written that late fails with an `IOException` rather than being lost.

### Why phase operations, and the alternatives considered

The compiler itself adds sources mid-compilation in exactly one place, `ClassNodeResolver.tryAsScript` (and `GroovyScriptEngine`'s resolver override), and it gets away without marking because it interrupts from the first operation of the phase. The `CompilationUnit` Javadoc explicitly offers `addSource` for adding units to a running compilation, and `dequeued()` documents that a unit already past a phase is skipped. Measured directly against Groovy 5.0.8 with a two-source unit: queuing from a semantic-analysis operation runs that phase's later operations twice on the existing source; queuing from the deferred operation with the phase marked runs every operation exactly once per source. Alternatives:

- `ModuleNode.addClass` + `ASTTransformationVisitor.addNewPhaseOperation` is the official route for a class a transform *builds as AST* (traits use it for their helper classes), and it only re-runs the annotation transforms for the class. Text needs parsing, resolution and the other conversion and semantic-analysis visitors, so every one of those would have to be replicated. Not viable.
- A nested `CompilationUnit` is Groovy's idiom for compiling generated text at runtime (`ProxyGeneratorAdapter`, `AstStringCompiler`, `GroovyClassLoader.parseClass`), but inside a running compilation it cannot see the unit's unfinished classes without a bridging `ClassNodeResolver`, its classes are outside the unit's AST (no duplicate-class checks, no attribution for incremental compilation) and the output has to be routed by hand. Heavier and less integrated.
- `setClassNodeResolver` is demand-driven: it adds a source only when something references the class, and nothing references a generated class.

### Incremental compilation

Gradle's incremental Groovy compilation registers a `CompilationCustomizer` at class generation that maps every class to a file via `sourceUnit.getSource().getURI().getPath()` and fails when the path is not under a source root. A `StringReaderSource` has an opaque `data:` URI. The generated unit therefore reports the URI of the source its first originating element came from (falling back to the source being visited), so the generated classes are tracked with the originating file, as javac tracks them with the originating elements. The target-directory test registers the same kind of customizer and asserts the attribution.

### Second fix: the visitors and bean definitions ran before Groovy's own transforms

Testing the generated sources with micronaut-sourcegen's Groovy suite (`@Builder` on a Groovy record) showed the generated builder calling `new Person()`. The cause is older than this PR: Groovy's `ASTTransformationVisitor.addPhaseOperations` registers the global transforms of a phase before the local ones, so `TypeElementVisitorTransform` (semantic analysis) saw a `record` before `RecordTypeASTTransformation` made it one, and `InjectTransform` (canonicalization) ran before `TupleConstructorASTTransformation` added the constructors of a record or of a `@TupleConstructor`, `@Canonical` or `@Immutable` class. `GroovyClassElement.possibleDefaultEmptyConstructor()` then invented a public no-arg constructor. Measured with a probe visitor: during visiting these classes reported no constructors and no methods; at `finish()` they reported the two-argument constructor and the generated methods. Consequences, both reproduced at runtime and now covered by `GroovyTransformOrderSpec`:

- `@Singleton record Rec(Dep dep)` got a definition calling `new Rec()` and failed with `IllegalAccessError: tried to access protected method java.lang.Record.<init>()`.
- `@Singleton @TupleConstructor class Svc { Dep dep }` got a no-arg constructor and no injection.

Both transforms keep their phase annotations but defer their work to a phase operation for `CANONICALIZATION` through `CompilationUnit.addNewPhaseOperation`, which the compiler appends after every operation of the phase, the local transforms included. The visitor operation is registered from semantic analysis and therefore precedes the injection operation registered from within canonicalization, preserving "visitors before bean definitions"; each is registered once per compilation unit and processes every source unit, the generated ones included. Generated sources are queued after both, at instruction selection, so their marking of a phase as complete never skips these operations.

Two consequences of seeing the real classes needed handling. Groovy expands a constructor with defaulted parameters into the shorter constructors only at class generation, right to left, so a `@Canonical` produced type showed one all-defaulted constructor and no no-arg one, and the factory proxy check in `FactoryBeanElementCreator` failed three existing AOP scope specs; `GroovyClassElement.getDefaultConstructor()` now derives the no-arg constructor Groovy will generate from such a constructor. And `@PackageScope`, itself a local semantic-analysis transform, turns a property into a package-private field without accessors before the visitors run, so `CustomVisitorSpec` no longer expects the two accessor methods that never existed in bytecode. Record accessors (`id()`, `name()`) still appear only at class generation; `getRecordMethods()` already synthesizes them for bean properties.

Also fixed while there: visitors left in the `loadedVisitors` thread-local by a compilation that failed after `TypeElementVisitorStart` and before `TypeElementVisitorEnd` were reused by the next compilation on the thread without `start()`; the start transform now replaces them when the compilation unit is a new one.

### Tests

`GeneratedSourceFileSpec` in `inject-groovy` has a `TypeElementVisitor` generate an annotated singleton as source and asserts, through the in-memory class loader the other Groovy specs use, that the class loads with its annotations, was visited exactly once, has a bean definition and is available from a bean context; that a generated source can generate a source; that several originating classes each get theirs; that a broken generated source fails the compilation naming the generated file; that an originating source cannot reference its generated class; and, through a `JavaAwareCompilationUnit` with a target directory set up as the Gradle Groovy compiler does, that the generated class and its bean definition are written next to the originating class.

`GroovyTransformOrderSpec` covers the record and tuple-constructor singletons at runtime, what a visitor sees of records and `@TupleConstructor`/`@Canonical`/`@Immutable` classes, and the replacement of stale visitors after a failed compilation. Full `:micronaut-inject-groovy:test`: 1137 tests, 0 failures.

Investigation notes: no Groovy test in core called `visitGeneratedSourceFile` before this, and the documentation makes no claim either way about Groovy support for generated sources.


core-processor/src/main/java/io/micronaut/inject/writer/ClassWriterOutputVisitor.java
```diff
@@ -137,6 +137,10 @@ default Optional<GeneratedFile> visitMetaInfFile(String path) {
     /**
      * Visit a source file that will be generated within the generated 'sources' directory.
      *
+     * <p>The file is compiled as part of the same compilation: javac and KSP process it in their next round,
+     * and the Groovy compiler queues it into the running compilation. The content must therefore be source in
+     * the language being compiled (see {@link io.micronaut.inject.visitor.VisitorContext#getLanguage()}).</p>
+     *
```
graemerocher, line 143: sounds like this means we will need `sourcegen-groovy`

graemerocher (approved):

core-processor/src/main/java/io/micronaut/inject/writer/ClassWriterOutputVisitor.java
dstepanov: Not necessarily: `sourcegen-generator-java` already ships `GroovyPoetSourceGenerator` (JavaPoet output declared as `Language.GROOVY`, registered in its services file) and a `test-suite-groovy` with `@Builder` on a Groovy record, commented out of the settings because core returned `Optional.empty()` under Groovy. Against this branch that suite compiles `PersonBuilder.groovy` from the JavaPoet text, since Groovy accepts that subset of Java syntax (including the `new P((Long) this.id, ...)` cast form), and its test passes once the record constructor fix on this PR is in. The known exception is Java's brace annotation arrays (`classes = {A.class, B.class}`), which Groovy parses as a closure; micronaut-cdi hand-writes those, so it needs `[A, B]` under Groovy. A Groovy-aware emitter would only be needed if the model ever generates Java-only constructs of that kind.

graemerocher: OK but it is probably not super optimal. The Groovy code should be annotated with `@CompileStatic` to get good performance

dstepanov: Agreed. The `PersonBuilder` the sourcegen Groovy suite produced against this branch confirms it: `javap` shows it implementing `GroovyObject` with the `metaClass` fields and invokedynamic call sites, so the JavaPoet text is compiled dynamically. Core cannot change that, since the text comes from `GroovyPoetSourceGenerator` in micronaut-sourcegen; the fix is for that generator to put `@groovy.transform.CompileStatic` on every class it emits (a `@CompileStatic` on a JavaPoet-shaped class compiles cleanly, as the generated code only uses declared types). I can open that sourcegen PR next.

PR #13180, opened 2026-09-13
Bind enum-keyed configuration Map with a bean value type

### LINKED ISSUE #12477: Configuration property binding fails for `Map<Enum, CustomClass>`
### Description

When using constructor injection with `@ConfigurationProperties` on a Kotlin data class, binding a YAML config to a `Map<String, CustomClass>` or a `Map<Enum, Int>` works perfectly. However, combining an Enum key and a Custom Class value as `Map<Enum, CustomClass>` fails.

Instead of binding the properties via the Property Binder, Micronaut attempts to resolve the map as a DI Bean dependency, resulting in a NoSuchBeanException.

### Steps to Reproduce

application.yml
```yaml
service:
  countries:
    France:
      population: 69081996
      area: 632702
    UAE:
      population: 11027129
      area: 83600

  countries2:
    France: 555
    UAE: 333
```

Application.kt
```kotlin
package com.example

import io.micronaut.context.annotation.ConfigurationProperties
import io.micronaut.context.annotation.Context
import io.micronaut.context.annotation.EachProperty
import io.micronaut.core.convert.format.MapFormat
import io.micronaut.core.naming.conventions.StringConvention
import io.micronaut.core.value.PropertyCatalog
import io.micronaut.runtime.Micronaut.run

fun main(args: Array<String>) {
    val ctx = run(*args)
    val config = ctx.getBean(AppConfig::class.java)
    println(config)
}

@Context
@ConfigurationProperties("service")
data class AppConfig(

    // !!!! THIS FAILS !!!!!
    @param:MapFormat(keyFormat = StringConvention.RAW)
    val countries: Map<Country, CountryConfig>, 

    // THIS WORKS PERFECTLY (if the previous line is commented out):
    // @param:MapFormat(keyFormat = StringConvention.RAW)
    // val countries: Map<String, CountryConfig>, 

    // THIS WORKS PERFECTLY:
    @param:MapFormat(keyFormat = StringConvention.RAW)
    val countries2: Map<Country, Int>  
)

@EachProperty("service.countries", catalog = PropertyCatalog.RAW)
data class CountryConfig(
    val population: Int,
    val area: Int
)

enum class Country {
    France, UAE
}
```

build.gradle.kts
```kotlin
plugins {
    id("org.jetbrains.kotlin.jvm") version "2.3.10"
    id("org.jetbrains.kotlin.plugin.allopen") version "2.3.10"
    id("com.google.devtools.ksp") version "2.3.6"
    id("io.micronaut.application") version "4.6.2"
    id("com.gradleup.shadow") version "9.3.2"
    id("io.micronaut.aot") version "4.6.2"
}

version = "0.1"
group = "com.example"


val kotlinVersion=project.properties.get("kotlinVersion")

repositories {
    maven("https://central.sonatype.com/repository/maven-snapshots/") {
        mavenContent { snapshotsOnly() }
    }
    mavenCentral()
    maven("https://s01.oss.sonatype.org/content/repositories/snapshots/") {
        mavenContent { snapshotsOnly() }
    }
}

dependencies {
    ksp("io.micronaut:micronaut-http-validation")
    ksp("io.micronaut.serde:micronaut-serde-processor")
    ksp("io.micronaut.validation:micronaut-validation-processor")

    implementation("io.micronaut.validation:micronaut-validation")
    implementation("io.micronaut.kotlin:micronaut-kotlin-runtime")
    implementation("io.micronaut.serde:micronaut-serde-jackson")
    implementation("org.jetbrains.kotlin:kotlin-reflect:${kotlinVersion}")
    implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:${kotlinVersion}")
    compileOnly("io.micronaut:micronaut-http-client")
    runtimeOnly("ch.qos.logback:logback-classic")
    runtimeOnly("tools.jackson.module:jackson-module-kotlin")
    runtimeOnly("org.yaml:snakeyaml")

    testImplementation("io.micronaut:micronaut-http-client")
    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}



application {
    mainClass = "com.example.ApplicationKt"
}

java {
    sourceCompatibility = JavaVersion.toVersion("25")
}

graalvmNative.toolchainDetection = false


micronaut {
    runtime("netty")
    testRuntime("junit5")
    processing {
        incremental(true)
        annotations("com.example.*")
    }
    aot {
        // Please review carefully the optimizations enabled below
        // Check https://micronaut-projects.github.io/micronaut-aot/latest/guide/ for more details
        optimizeServiceLoading = false
        convertYamlToJava = false
        precomputeOperations = true
        cacheEnvironment = true
        optimizeClassLoading = true
        deduceEnvironment = true
        optimizeNetty = true
        replaceLogbackXml = true
    }

}

tasks.named<io.micronaut.gradle.docker.MicronautDockerfile>("dockerfile") {

    baseImage = "eclipse-temurin:25-jre"
}
```

### Expected Behavior
The `Map<Country, CountryConfig>` should be populated from the YAML configuration, successfully converting the `String` keys to Enum values and mapping the nested objects, exactly as it does for `Map<String, CountryConfig>` and `Map<Country, Int>`.

### Actual Behavior
The application fails to start and throws a Bean resolution exception:
```
08:20:03.905 [main] ERROR io.micronaut.runtime.Micronaut - Error starting Micronaut server: Bean definition [com.example.AppConfig] could not be loaded: Failed to inject value for parameter [countries] of class: com.example.AppConfig

Message: No bean of type [java.util.Map<com.example.Country,com.example.CountryConfig>] exists. 
Path Taken: 
new @i.m.c.a.Context("service") c.e.AppConfig(Map<Country, CountryConfig> countries, Map<Country, Integer> countries2)
\---> new @i.m.c.a.Context("service") c.e.AppConfig([Map<Country, CountryConfig> countries], Map<Country, Integer> countries2)
io.micronaut.context.exceptions.BeanInstantiationException: Bean definition [com.example.AppConfig] could not be loaded: Failed to inject value for parameter [countries] of class: com.example.AppConfig

Message: No bean of type [java.util.Map<com.example.Country,com.example.CountryConfig>] exists. 
Path Taken: 
new @i.m.c.a.Context("service") c.e.AppConfig(Map<Country, CountryConfig> countries, Map<Country, Integer> countries2)
\---> new @i.m.c.a.Context("service") c.e.AppConfig([Map<Country, CountryConfig> countries], Map<Country, Integer> countries2)
	at io.micronaut.context.DefaultBeanContext.initializeEagerBeans(DefaultBeanContext.java:1901)
	at io.micronaut.context.DefaultBeanContext.initializeContext(DefaultBeanContext.java:1823)
	at io.micronaut.context.DefaultBeanContext.configureAndStartContext(DefaultBeanContext.java:3007)
	at io.micronaut.context.DefaultBeanContext.start(DefaultBeanContext.java:330)
	at io.micronaut.context.DefaultApplicationContext.start(DefaultApplicationContext.java:291)
	at io.micronaut.runtime.Micronaut.start(Micronaut.java:77)
	at io.micronaut.runtime.Micronaut.run(Micronaut.java:354)
	at io.micronaut.runtime.Micronaut.run(Micronaut.java:329)
	at com.example.ApplicationKt.main(Application.kt:17)
Caused by: io.micronaut.context.exceptions.DependencyInjectionException: Failed to inject value for parameter [countries] of class: com.example.AppConfig

Message: No bean of type [java.util.Map<com.example.Country,com.example.CountryConfig>] exists. 
Path Taken: 
new @i.m.c.a.Context("service") c.e.AppConfig(Map<Country, CountryConfig> countries, Map<Country, Integer> countries2)
\---> new @i.m.c.a.Context("service") c.e.AppConfig([Map<Country, CountryConfig> countries], Map<Country, Integer> countries2)
	at io.micronaut.context.AbstractInitializableBeanDefinition.resolveBean(AbstractInitializableBeanDefinition.java:2226)
	at io.micronaut.context.AbstractInitializableBeanDefinition.getBeanForConstructorArgument(AbstractInitializableBeanDefinition.java:1423)
	at com.example.$AppConfig$Definition.instantiate(Unknown Source)
	at io.micronaut.context.DefaultBeanContext.resolveByBeanFactory(DefaultBeanContext.java:2092)
	at io.micronaut.context.DefaultBeanContext.createRegistration(DefaultBeanContext.java:2803)
	at io.micronaut.context.SingletonScope.getOrCreate(SingletonScope.java:78)
	at io.micronaut.context.DefaultBeanContext.intializeEagerBean(DefaultBeanContext.java:2697)
	at io.micronaut.context.DefaultBeanContext.initializeEagerBean(DefaultBeanContext.java:2336)
	at io.micronaut.context.DefaultBeanContext.initializeEagerBeans(DefaultBeanContext.java:1895)
	... 8 common frames omitted
Caused by: io.micronaut.context.exceptions.NoSuchBeanException: No bean of type [java.util.Map<com.example.Country,com.example.CountryConfig>] exists. 
	at io.micronaut.context.DefaultBeanContext.newNoSuchBeanException(DefaultBeanContext.java:2451)
	at io.micronaut.context.DefaultApplicationContext.newNoSuchBeanException(DefaultApplicationContext.java:345)
	at io.micronaut.context.DefaultBeanContext.resolveBeanRegistration(DefaultBeanContext.java:2415)
	at io.micronaut.context.DefaultBeanContext.getBean(DefaultBeanContext.java:1642)
	at io.micronaut.context.AbstractBeanResolutionContext.getBean(AbstractBeanResolutionContext.java:236)
	at io.micronaut.context.AbstractInitializableBeanDefinition.resolveBean(AbstractInitializableBeanDefinition.java:2209)
	... 16 common frames omitted

Caused by: io.micronaut.context.exceptions.DependencyInjectionException: Failed to inject value for parameter [countries] of class: com.example.AppConfig

Caused by: io.micronaut.context.exceptions.NoSuchBeanException: No bean of type [java.util.Map<com.example.Country,com.example.CountryConfig>] exists. 
```

### Environment Information

- Micronaut Version: 5.0.0-M16
- Kotlin Version: 2.3.10
- JDK Version: temurin 25.0.2
- Build Tool: Gradle

elgamal-ahmed:
Fixes binding of a `@ConfigurationProperties` constructor parameter of type `Map` with an enum key and a configuration-bean value type. Closes #12477.

`Map<String, CountryConfig>` already works, and not by configuration binding. The value type is a bean, so `isPropertyParameter` returns false and the parameter is resolved as an injectable map, with the `@EachProperty` beans collected by their bean name. `BeanInjectionUtils#isInjectableMap` required the key to be a `CharSequence`, so `Map<Country, CountryConfig>` did not match, fell through to the plain single-bean lookup at the end of `getInjectionPoint`, and failed at startup with `NoSuchBeanException`.

Beans declared with `@EachProperty` or `@EachBean` exist once per configuration key or per parent bean, so their names form a closed set that an enum can model. `isInjectableMap` now accepts an enum key for those value types.

Whether a bean of the map type exists is not something the compiler can answer, so `resolveMapOfType` decides it. A map that is supplied directly, by a `@Factory` for instance, is resolved first and keeps the precedence it has today. Only when there is no such bean are the beans collected and their names converted to the key type.

That conversion applies to the names alone. The resolved beans are injected as they are, rather than passed through the general `Map` to `Map` converter, which drops entries whose key does not convert and rebuilds values that are themselves a `Map` or a `Collection`. A name that does not convert, or that collides with one already converted, fails the injection instead of quietly leaving a bean out of the map.

### Tests

`EnumKeyedMapInjectionSpec` and `EnumKeyedMapKeyConversionSpec` build real contexts and assert on values:

| test | merge base | this change |
|---|---|---|
| enum-keyed map of `@EachProperty` beans | `NoSuchBeanException` | populated, enum keys |
| enum-keyed scalar map | passes | passes |
| enum-keyed map of `@EachBean` beans | `NoSuchBeanException` | collected by bean name |
| `@Factory` map of plain beans | passes | passes |
| `@Factory` map of iterable beans | passes | passes |
| bean name with no matching constant | `NoSuchBeanException` | fails with a clear message |
| two names converting to one constant | `NoSuchBeanException` | fails with a clear message |
| bean value that is itself a `Map` | `NoSuchBeanException` | injected as the bean |

Three of those pass on the merge base, so the claim that nothing resolving today changes is checkable rather than asserted. `inject-kotlin` additionally carries the issue's Kotlin/KSP reproducer verbatim.

Verified on JDK 26: `:micronaut-inject-java:test` 5286, `inject-kotlin` 866, `inject` 405, `context` 48, `core` 7429. `inject-groovy` is green apart from 10 pre-existing `GroovyReconstructionSpec` `StackOverflowError`s that fail the same way on an unmodified checkout.


core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
```diff
@@ -341,7 +341,16 @@ private static boolean isPropertyParameter(ClassElement genericType, VisitorCont
             // Get the class with type annotations
             genericType = visitorContext.getClassElement(genericType.getCanonicalName()).orElse(genericType);
         } else if (genericType.isAssignable(Map.class)) {
-            ClassElement t = genericType.getTypeArguments().get("V");
+            Map<String, ClassElement> typeArguments = genericType.getTypeArguments();
```
graemerocher: `getTypeArguments()` returns the arguments declared on `genericType` itself, keyed by *its own* type-variable names. The branch condition is `isAssignable(Map.class)`, so `genericType` can be any `Map` subtype — e.g. `class Populations extends HashMap<Country, Integer>` or `interface CountryMap extends Map<Country, CountryConfig>`. For those, `getTypeArguments()` is empty, `keyType` is `null`, and the new branch is skipped, so they keep falling through to the bean-dependency path and keep failing the same way.

`BeanInjectionUtils#getInjectionPoint` resolves the same pair with `genericType.getTypeArguments(Map.class)`; using that here would make the two sides agree and would cover subtypes:

```java
Map<String, ClassElement> typeArguments = genericType.getTypeArguments(Map.class);
```

(The pre-existing `"V"` lookup below has the same limitation, so this would fix both.)

---

core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
```diff
@@ -341,7 +341,16 @@ private static boolean isPropertyParameter(ClassElement genericType, VisitorCont
...
+            ClassElement keyType = typeArguments.get("K");
+            if (keyType != null && !keyType.isAssignable(CharSequence.class)) {
+                // A Map injected as a bean dependency is collected by the bean qualifier name, so it
+                // only supports String keys (see BeanInjectionUtils#isInjectableMap). A Map with a
+                // non-String key cannot be a bean dependency and must be bound as a configuration
+                // property, even when its value type is itself a bean.
+                return true;
```
graemerocher: Two things here.

**Behaviour change beyond the reported case.** This makes *every* non-`CharSequence`-keyed `Map` parameter a configuration property, not just ones whose value type is a configuration bean. Such a map is not an *injectable* map (`isInjectableMap` rejects it), but it does not follow that it cannot be a bean dependency: it falls through to the final `new BeanInjectionPoint<>(genericType, ...)` in `BeanInjectionUtils#getInjectionPoint`, i.e. a plain single-bean lookup for the `Map` type itself. That lookup succeeds today when a `@Factory` exposes the map — e.g.

```java
@Factory
class Handlers {
    @Singleton Map<Status, Handler> handlers() { ... }
}

@ConfigurationProperties("app")
class AppConfig {
    @ConfigurationInject AppConfig(Map<Status, Handler> handlers) { ... }
}
```

After this change that parameter is annotated `@Property("app.handlers")` and bound from configuration instead. Worse, it fails quietly: `DefaultMutableConversionService`'s `Map` → `Map` converter `continue`s past any entry it cannot convert, so the constructor gets an empty map rather than an error. Is that trade-off intended? If so it's probably worth a note in the PR/issue, since it's a source-compatible but behaviour-breaking change on a 5.3.x line.

**Comment wording.** The code tests `CharSequence`, the comment says "String keys" / "non-String key". The code matches `isInjectableMap` (which also tests `CharSequence`), so it's the comment that's off — worth rewording to `CharSequence` so a reader doesn't go looking for a `String` check that isn't there.

---

inject-java/src/test/groovy/io/micronaut/inject/configproperties/EnumKeyedMapConfigurationPropertiesSpec.groovy
```diff
@@ -0,0 +1,121 @@
...
+        BeanDefinition beanDefinition = buildBeanDefinition('test.AppConfig', SOURCE)
+        def arguments = beanDefinition.constructor.arguments
+
+        then: '''both maps are treated as configuration properties.
+                 Before the fix the enum-keyed Map with a bean value was misclassified as a bean dependency
+                 (no @Property) and failed at runtime with a NoSuchBeanException.'''
+        arguments.length == 2
+        arguments[0].synthesize(Property).name() == 'service.countries'
+        arguments[1].synthesize(Property).name() == 'service.populations'
+    }
```
graemerocher: The case #12477 actually reports — `Map<Country, CountryConfig>`, a bean value type — is only asserted at the metadata level. The runtime test below exercises `populations` (the scalar map) and never touches `countries`, so nothing here shows the reported startup failure is actually resolved rather than converted into a different wrong answer.

That gap matters because the classification passing does not imply the binding works. `DefaultMutableConversionService`'s `Map` → `Map` converter skips (`continue`) any entry whose value cannot be converted, so with no `Map` → `CountryConfig` converter on the classpath the parameter binds to an **empty** map and the bean constructs without error. I looked for such a converter and didn't find one in core; `JacksonConverterRegistrar` registers `JsonNode` ↔ `Object`, `ArrayNode` → `Object[]`/`Iterable` and `Object` → `JsonNode`, but no `Map` → `Object`, and converter lookup walks the source/target hierarchies rather than chaining, so `Map` → `CountryConfig` wouldn't resolve through `Object` → `JsonNode` → `CountryConfig`. That makes the PR description's "population of the nested bean values is then handled by the serialization layer" the part most worth proving.

Could you add an end-to-end assertion for the bean-valued map — run the context with

```
'service.countries.France.population': 69081996,
'service.countries.France.area'      : 551695
```

and assert `config.countries[france].population == 69081996`? `micronaut-jackson-databind` is already on `inject-java`'s test runtime, so if the conversion does work it can be asserted right here. If it turns out it doesn't populate, that's worth knowing before this merges — binding an empty map is a quieter failure than the `NoSuchBeanException` it replaces.

(I wasn't able to run the suite locally to check either way — the build requires a JDK 25 toolchain and only 21 is available in this environment.)

---

graemerocher (commented):
Reviewed the change. The diagnosis looks right — `isPropertyParameter` only inspected the map's value type, so an enum-keyed map with a configuration-bean value was classified as a dependency even though `BeanInjectionUtils#isInjectableMap` can never collect it — and keying the decision off the map's key type is a reasonable place to fix it. CI is green.

Three things I'd like to see addressed, left inline:

1. `getTypeArguments()` misses `Map` subtypes (the branch guard is `isAssignable(Map.class)`); `getTypeArguments(Map.class)` is what `BeanInjectionUtils` uses for the same resolution.
2. The new branch also reclassifies non-`CharSequence`-keyed maps that are legitimately resolved today as a single-bean lookup (e.g. a `@Factory`-provided `Map<SomeEnum, SomeHandler>`), and they'd now bind silently to an empty map. Worth confirming that's intended on a 5.3.x line.
3. The bean-valued map — the case the issue reports — has no runtime assertion; only the scalar map is exercised end to end. Given that the `Map` → `Map` converter silently drops unconvertible values, that's the assertion that would actually prove the fix.

---

core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
elgamal-ahmed: Right, `getTypeArguments()` only returns what the element declares, so for a `Map` subtype it comes back empty and `keyType` is null.

Moot now, the fix no longer touches this file. The only lookup left is in `isInjectableMap`, which is gated on an exact type-name match against `Map`/`HashMap`/`LinkedHashMap`/`TreeMap`, so `K` and `V` are always declared.

Left the visitor's `"V"` lookup alone on purpose: `getTypeArguments(Map.class)` there would flip something like `class MyMap extends HashMap<String, SomeScopedBean>` from configuration property to bean dependency. Can do that separately if you want it.

core-processor/src/main/java/io/micronaut/context/visitor/ConfigurationMetadataWriterVisitor.java
elgamal-ahmed: Dropped that branch.

(Edited: an earlier version of this reply said the factory case was unaffected and pointed at a test for it. That was wrong on both counts, and the current revision fixes it. Leaving the note so the thread makes sense.)

It works differently now. Rather than reclassifying the parameter, it widens the injectable-map path that already makes `Map<String, CountryConfig>` work: `isInjectableMap` accepts an enum key when the value type is an iterable bean (`@EachProperty` or `@EachBean`).

Whether a bean of the map type exists isn't something the compiler can answer, so `resolveMapOfType` decides it. A map supplied directly, by a factory for instance, resolves first and keeps the precedence it has today. Only when there is no such bean are the beans collected and their names converted. So your `Map<Status, Handler>` still gets the factory's map, and so does a factory-supplied `Map<Country, CountryConfig>`, which the previous revision quietly took over. There's a test for that now, and it passes on the merge base as well as on this head.

inject-java/src/test/groovy/io/micronaut/inject/configproperties/EnumKeyedMapConfigurationPropertiesSpec.groovy
elgamal-ahmed: Checked it, and it doesn't work. The map comes back as an empty `LinkedHashMap`, for the reason you gave: no `Map` to `CountryConfig` converter, so the `Map` to `Map` converter skips every entry. The PR was swapping a `NoSuchBeanException` for a silently empty map.

Reworked it, details on the other thread. The spec builds a real context now and asserts `countries[France].population == 69081996` plus that the keys really are `Country` instances. The issue's Kotlin/KSP repro is in `inject-kotlin` as well.

Had JDK 26 here so I could run it: `:micronaut-inject-java:test` green, 5282 tests.

dstepanov (commented):
## Overall

The approach is sound and the layering is right: the compile-time gate only widens which injection points become a `MapOfBeansInjectionPoint`, and the "is there a bean of the map type?" question — which the compiler genuinely cannot answer — is deferred to `resolveMapOfTypeByKey`, which resolves a directly-supplied map first. Converting only the *names* rather than putting the whole map through the `Map` → `Map` converter is the correct call, and the `ShardMap extends LinkedHashMap` test pins exactly the behaviour that converter would have broken. Failing loudly on an unconvertible or colliding name, rather than dropping a bean, is the right default.

On the generated-code side: all four map injection sites (`BeanDefinitionWriter:3026/4034/4084`) emit the `...Object` variants, so the widened `Object` return of `resolveMapOfType` never reaches the legacy `Map<String, V>`-typed overloads. No binary-compatibility problem there.

## Main issue — the gate misses factory-declared `@EachBean` beans

`isKeyedByBeanName` decides via `isIterable(beanType)`, which reads `@EachProperty`/`@EachBean` off the *value type's own* annotation metadata. But the idiomatic way to declare an iterable bean is on a factory method:

```java
@Factory
class ShardClientFactory {
    @EachBean(ShardConfig.class)
    ShardClient client(ShardConfig config) { ... }
}
```

`ShardClient` the class carries no `@EachBean`, so `isIterable` is false, and `Map<Shard, ShardClient>` still falls through to the single-bean lookup and still fails with `NoSuchBeanException` — while `Map<String, ShardClient>` works today, because the String path never inspects the value type at all.

Both `@EachBean` tests here declare the annotation on the class, so this gap is not covered. Worth adding a factory-declared `@EachBean` case and either handling it or documenting the limitation explicitly: as it stands the fix will look to users like it works for some `@EachBean` beans and not others, with no obvious difference between them.

## Design question worth settling before merge

Restricting enum keys to iterable value types is defensible, but the runtime already resolves a supplied map bean first — so relaxing it to *any* bean value type would only change behaviour where no map bean exists and no beans are collected, i.e. today's `NoSuchBeanException` would become an empty map.

If that is the only thing the restriction buys, it may be cheaper to drop the `isIterable` condition and instead have `resolveMapOfTypeByKey` throw when the collected map is empty and no map bean was found. That preserves the failure, removes the factory gap above, and makes enum-keyed maps behave symmetrically with String-keyed ones. A semantics decision rather than a defect — but it decides how wide the fix is.

Related, and already true within the PR's current scope: for a configured-but-empty enum-keyed map of iterable beans, startup now silently injects an empty map where it previously failed. That is consistent with the String-keyed behaviour, so presumably intended — worth a line in the description.

## Minor

- `resolveMapOfType` keys off `getTypeVariable("K")` while the compile-time gate reads `getTypeArguments().get("K")`. Fine for the four map types `isInjectableMap` allows, and a missing `K` degrades safely to the existing path — but the two sides are coupled by a string literal across modules, so a comment on one pointing at the other would help.
- `beansByKey.put(key, entry.getValue()) != null` treats a null bean value as "no collision". Not reachable through `mapOfType`, just noting it.
- If the injection point is declared as `TreeMap<Country, X>` rather than `Map<...>`, `returnType.isInstance(beansByKey)` fails and the result goes through `convertRequired` — back through the `Map` → `Map` converter this method exists to avoid. Narrow, but it quietly undoes the guarantee the third `EnumKeyedMapKeyConversionSpec` test pins.
- Kotlin test: `@param:MapFormat(keyFormat = StringConvention.RAW)` on `countries` (the bean-valued map) has no effect on the bean-collection path and is absent from the Java equivalent. Dropping it would make the test assert the thing it claims to.

Nothing here undermines the correctness of what is implemented; the factory-`@EachBean` gap is the one I would want resolved — fixed or stated — before merge.


---

elgamal-ahmed:
* Factory `@EachBean` is real. `isIterable` reads the value type and the annotation is on the method, so `Map<Shard, X>` throws where `Map<String, X>` collects. Not fixed, pinned with a test and documented on `isKeyedByBeanName`.

* Dropping `isIterable`: the throw on empty would break an empty `@EachProperty` config, so it moves the asymmetry rather than removing it. Left as is, can widen if you prefer.

* Regression found while checking that, now fixed. Enum keyed maps of an inner configuration were broken here, not just unsupported: zones under regions are named `france-north`, so the names never convert. That shape resolves on the merge base, so my "nothing resolving today changes" was wrong. Fixed in the gate and in the qualifier used for the supplied map lookup, with a test each.

graemerocher (approved):

PR #13181, opened 2026-09-13
Read @Introspected.Property on the most derived declaration, for a proxy as for a class

dstepanov:
## Summary

Follow-up to #13175, addressing the review comments left after it merged. `ReflectionBeanIntrospection` now reads `@Introspected.Property` the way the processors do, for every type and not for a proxy alone; `ReflectionBeanIntrospectionFallback` gives a proxied interface's annotations the precedence a class's get; a private interface method is no longer a bean method of a proxy; and the class walk is resolved once instead of on every comparison.

## What changed

- **`@Introspected.Property` on an inherited accessor** (review comment on `declared`): the interface pass read the annotation only for a proxy, so `of(B)` and `of(proxyOf(B).getClass())` disagreed for `interface B extends A {}` with the declared accessor on `A`. It is now read for every type, with one rule: a declaration counts for the most derived one alone. An override hides the declaration it overrides and carries no annotation of its own, since method annotations are not inherited. That is what the processors read. Checked against the generated introspection, Java (compiled in memory) and Groovy, for a sub interface, a sub interface redeclaring the accessor, an abstract class inheriting it, a class overriding it, a class overriding a super class's declared accessor, and a prefixed getter carrying `accessKind = READ` on the interface, overridden in the class: the reflective side reports the same as the generated one in every case. Two of those differed before, both in the reflective side's favour: a class overriding a declared accessor of its super class reported the property, and an inherited `accessKind` was honoured for an override.
- **`ReflectionBeanIntrospectionFallback#configured`** (review comment on `typeMetadataOf`): the reviewer note on #13175 said the fallback never sees a proxy class. That holds for a proxy of public interfaces only. The JDK names the proxy of a package-private interface after the package of that interface, `io.micronaut.reflection.$Proxy0`, which an `allow-reflection` pattern for the package matches. `configured` then read the annotations of the proxy class, which has none, so what the application configured for the package won over what the interface declares, the opposite of the precedence the method exists to keep. It now reads `ReflectionBeanIntrospection.typeMetadataOf`, widened to package-private, which answers the annotations of the proxied interfaces.
- **Private interface methods** (review comment on `discoverMethods`): the interface pass filtered static, synthetic and bridge methods but not visibility, while the class pass requires `public`. A proxy, described through the interface pass alone, reported a private helper method as a bean method, and so did a class implementing the interface. Both passes are now one method, `addMethods`, public methods only.
- **Resolved once** (review comment on `rankOf`): `proxy`, `classHierarchy` and `interfaces` are fields, computed in the constructor and handed to `PropertyMembers`. `rankOf` rebuilt both lists on every comparison in `selectGetter` and `selectSetter`, and `Proxy.isProxyClass` was called at six sites. It is now called in the constructor and in the two static entry points, `isIntrospectable` and `typeMetadataOf`.
- **Redundant branch** (review comment on `addGetter`): the two accessor passes of `discoverProperties` are one method, `addAccessors`, which drops the `addGetter(Method)` overload, the branch on a null declaration, and the setter branch re-reading the annotation.
- **Spec handler** (review comment on `HANDLER`): it answers `toString`, `hashCode` and `equals` as an object does, and a `String` for a `String` return type. It returned `42` for everything but `getString`, and Spock renders the proxy when a condition fails, so a `ClassCastException` would have replaced the real failure.

## Reviewer notes

- `ReflectionBeanDefinition.isDefinable` still accepts a proxy class, as the review noted for context. Left out here on purpose: it is a bean definition question, not an introspection one.
- The suppressed Copilot note asking for a proxy test with a type-level setting that changes discovery is covered by the existing case on `ProxiedStyled`, which asserts both the properties and the `@AccessorsStyle` metadata through the proxy.

## Tests

- `ProxyIntrospectionSpec`, new cases:
  - an accessor a super interface declares `@Introspected.Property` is a property of the sub interface and of a proxy of it, and not of a class overriding it
  - a private method of an interface is not a method of the proxy, and its public default method is invocable through it
  - the fallback serves a proxy of a package-private interface, and the `accessKind` the interface declares wins over the one configured for the package
  - parity: the generated introspections of a Groovy sub interface, an abstract class inheriting the accessor and a class overriding it report what the reflective side does
- `./gradlew :micronaut-reflection:test`: 411 tests, 3 skipped (existing pending features), none failing. Checkstyle and spotless clean.


reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -797,7 +776,46 @@ private List<BeanProperty<T, Object>> discoverProperties() {
...
+     * @param candidates The candidate properties, by name
+     */
+    private void addAccessors(Class<?> type, Set<String> seen, Map<String, PropertyMembers> candidates) {
+        for (Method method : type.getDeclaredMethods()) {
+            if (Modifier.isStatic(method.getModifiers()) || method.isSynthetic() || method.isBridge() || isGroovyObjectMethod(method)) {
+                continue;
+            }
+            String name = method.getName();
+            int parameters = method.getParameterCount();
+            Introspected.Property declared = seen.add(signature(method)) ? method.getAnnotation(Introspected.Property.class) : null;
```
graemerocher: The premise behind this gate looks wrong, and I think it makes the reflective side diverge from the processors in the direction it is trying to fix.

"An override carries no annotation of its own, method annotations not being inherited" is true of `java.lang.reflect`, but it is not how the processors read a method. `JavaAnnotationMetadataBuilder#buildHierarchy` merges the overridden declarations into a method's metadata:

```java
} else if (element instanceof ExecutableElement executableElement) {
    // we have a method
    // for methods we merge the data from any overridden interface or abstract methods
    ...
    hierarchy.addAll(nativeElementsHelper.findOverriddenMethods(executableElement));
```

and `AstBeanPropertiesUtils` tests that merged view, not the declaration:

```java
private static boolean isIntrospectedPropertyMethod(MethodElement methodElement) {
    return methodElement.hasAnnotation(ANN_INTROSPECTED_PROPERTY);
}
```

`hasAnnotation` is true for an annotation inherited from an overridden method — that is exactly what `MethodElement#getDeclaredMethodAnnotationMetadata` exists to exclude ("Unlike `getMethodAnnotationMetadata()` the annotations inherited from the super methods are excluded"), and the property code does not use it. Nothing filters that inheritance by `@Inherited` either; `addAnnotations` records the parent's annotations as non-declared and `hasAnnotation` sees them.

So for a class overriding an accessor its interface declares `@Introspected.Property`, the generated introspection should still report the property, and this gate now drops it. Same for the `accessKind` case listed in the description — the processor's `resolveIntrospectedPropertyAccess` also reads `enumValues` off the merged metadata.

If that is right, the two behaviours the description gives up ("both in the reflective side's favour") were the matching ones, and the fix for the review comment is just to read the annotation in the interface pass as well — the `proxy ?` gate removed, no `seen` gate added. Worth confirming against the generated side before this goes further, since the new parity case asserts the opposite (see the comment on that assertion).

One more wrinkle for whatever rule you land on: `GroovyAnnotationMetadataBuilder#buildHierarchy` merges the overridden methods **only when the override carries `@Override`**, so Java and Groovy already disagree here, and no signature-based rule can reproduce both.

---

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
```diff
@@ -695,81 +724,31 @@ public String toString() {
 
     private List<BeanProperty<T, Object>> discoverProperties() {
         Map<String, PropertyMembers> candidates = new LinkedHashMap<>();
+        // the signatures of the methods seen so far, the most derived declaration of each first: a declaration
+        // hides the ones it overrides, and what @Introspected.Property says holds for the most derived one
+        // alone, as the processor reads it - an override carries no annotation of its own, method annotations
+        // not being inherited, so the declaration of a super class or of an interface makes an accessor of a
+        // type only where the type does not override it
+        Set<String> seen = new HashSet<>();
```
graemerocher: If the gate stays in some form, note that `signature()` is name plus erased parameter types, so "the same method" is decided on erasure and the answer differs between two shapes of the same override:

```java
interface Repo<T> {
    @Introspected.Property T item();       // covariant return
    @Introspected.Property void save(T t); // generic parameter
}
interface StringRepo extends Repo<String> {
    @Override String item();
    @Override void save(String t);
}
```

- `item()` — the override's signature is `item()`, the declaration's is `item()`, and the compiler's bridge is filtered as synthetic. The declaration is hidden.
- `save(String)` — the override's signature is `save(java.lang.String,)`, the declaration's is `save(java.lang.Object,)`. Different strings, so the declaration is *not* hidden and still makes the property.

Same relationship, opposite outcome. `MethodHierarchy#overriddenMethod` in this module already solves this ("the erasure of a generic declaration tolerated"), so there is something to reuse rather than a new rule to write.

Related: two unrelated interfaces of the same type declaring the same accessor are not in an override relationship at all, but the first one `interfaces` happens to list consumes the signature for the other. `MethodHierarchy#parallel` exists for that case.

---

reflection/src/test/groovy/io/micronaut/reflection/ProxyIntrospectionSpec.groovy
```diff
@@ -153,4 +216,22 @@ class ProxyIntrospectionSpec extends Specification {
...
+    void "an inherited @Introspected.Property accessor is the property the generated introspection reports"() {
+        given: "the types inheriting the declared accessor, and the one overriding it, described both ways"
+        def generatedSub = BeanIntrospector.SHARED.getIntrospection(ProxiedParityDeclaredSub)
+        def generatedBase = BeanIntrospector.SHARED.getIntrospection(ProxiedParityDeclaredBase)
+        def generatedBean = BeanIntrospector.SHARED.getIntrospection(ProxiedParityDeclaredBean)
+
+        expect: "the generated side reports the property to the types inheriting the accessor, not to the one overriding it"
+        generatedSub.beanProperties*.name == ["value"]
+        generatedBase.beanProperties*.name == ["value"]
+        generatedBean.beanProperties.empty
```
graemerocher, line 231: This is the assertion that should settle the question raised on the `seen` gate, and by my reading of the processors it should fail.

`ProxiedParityDeclaredBean.value()` carries `@Override`, which is what `GroovyAnnotationMetadataBuilder#buildHierarchy` requires before it merges the overridden declarations:

```java
if (!methodNode.getAnnotations(ANN_OVERRIDE).isEmpty()) {
    hierarchy.addAll(nativeElementHelper.findOverriddenMethods(methodNode.getDeclaringClass(), methodNode));
}
```

So `ProxiedParityDeclared.value()`'s `@Introspected.Property` is in the method's metadata, `AstBeanPropertiesUtils#isIntrospectedPropertyMethod` asks `hasAnnotation`, and the generated introspection should report a `value` property rather than none.

Line 131 of this spec (`overriding.beanProperties.empty`) asserts the same thing on the reflective side for the Java `ProxiedDeclaredBean`, so both would need revisiting together.

I could not run the suite to confirm — the build needs JDK 25 and I only have 21 here — and the checks on this PR are all `skipped` because it is a draft, so nothing has exercised these yet. Marking it ready (or running `:micronaut-reflection:test` again after the last edit) would give a definite answer. If the generated side does report the property, then dropping just the `proxy ?` gate, with no `seen` set, is the change that matches it.

Worth adding a case either way: drop `@Override` from a Groovy override of the same accessor. Java and Groovy disagree there, and it is the case a signature rule cannot see.

---

reflection/src/test/groovy/io/micronaut/reflection/ProxyIntrospectionSpec.groovy
```diff
@@ -103,6 +112,60 @@ class ProxyIntrospectionSpec extends Specification {
...
+        given:
+        def bean = proxyOf(ProxiedHelper)
+        def proxied = ReflectionBeanIntrospection.of(bean.getClass())
+
+        expect: "the interface reports its public methods alone"
+        ReflectionBeanIntrospection.of(ProxiedHelper).beanMethods*.name.toSet() == ["getName", "describe"].toSet()
+
+        and: "so does the proxy, which the private method is no method of"
+        proxied.beanMethods*.name.toSet() == ["getName", "describe"].toSet()
+        proxied.beanMethods.find { it.name == "describe" }.invoke(bean) == "proxied"
```
graemerocher: Small accuracy point: a JDK proxy routes a default method to the handler like any other, so this returns the handler's `"proxied"` and the body of `describe()` never runs — `helper()` is not exercised by the assertion. The method name and the spec text ("its public default method is invocable through it") read as though the default body is what answers.

`getName` and `describe` reaching the bean methods while `helper` does not is the point of the case and it makes that well; it is just the invocation half that proves less than it looks. Calling `ProxiedHelper.super`-style through a real implementation, or simply dropping the `invoke` line, would keep the case honest.

---

graemerocher (commented):
Thanks for the follow-up — four of the six points from the #13175 review are cleanly addressed and I have nothing to add on them:

- `configured()` now reading `typeMetadataOf` is the right fix, and the new fallback case discriminates properly: with the configured `FIELD` access winning, the interface has no fields and `getName` would not survive `isProperty`, so `beanProperties == ["name"]` really does prove `METHOD` from the interface won.
- `proxy`, `classHierarchy` and `interfaces` as fields removes the rebuild in the `rankOf` comparator, and `allInterfaces(List.of(beanType))` for a proxy is equivalent to the old special case.
- `addMethods` with the `public` filter applied to both passes fixes the private-interface-method leak for classes as well as proxies.
- The handler and the `addAccessors`/`addGetter` consolidation are both good. I checked the `proxyOf` change to `type.classLoader`: `Proxy.newProxyInstance(null, …)` for the bootstrap-loaded JDK interfaces works (`java.util.List` → `jdk.proxy1.$Proxy0` on 21), so the JDK case is unaffected.

The one thing I would not merge as-is is the `seen` gate, which goes beyond the review comment it answers and, I think, in the wrong direction. The comment asked for `@Introspected.Property` to be read on interface declarations for every type, not only for proxies; this reads it for every type but then suppresses it wherever a more derived type redeclares the method. The processors do not do that: `JavaAnnotationMetadataBuilder#buildHierarchy` merges the overridden declarations into a method's metadata, and `AstBeanPropertiesUtils` asks `hasAnnotation`, which sees them — `MethodElement#getDeclaredMethodAnnotationMetadata` is the API that excludes them and the property code does not use it. If that reading holds, the two behaviours the description gives up were the matching ones, and the minimal change is to drop the `proxy ?` gate and stop there. Details and the Groovy `@Override` wrinkle are in the inline comments.

Flagging the verification gap plainly: every check on this PR is `skipped` because it is a draft, so nothing has run these tests, and I could not run them either — the build wants JDK 25 and this environment has 21. The new parity assertion is the one that should decide it, and the inline comment on it says which way I expect it to go.

---

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
dstepanov: The gate stays, but as an override check rather than the signature set (`7798c931db`): `isOverridden` hides a declaration only behind a method that overrides it as a member of the bean type, a generic one matched through `MethodHierarchy.overrides`. On the premise: the builder does merge the overridden declarations into the hierarchy, but `AbstractAnnotationMetadataBuilder#addAnnotations` keeps, of the annotations that are not the method's own, only the ones meta-annotated `@Inherited` (`isDeclared || isStereotype || alwaysIncludeAnnotation || isInherited(...)`), and `Introspected.Property` is not. Compiled with the Java processor on JDK 25, `class Impl implements A { public String value() }` yields no property, and an inherited `accessKind = READ` is not honoured on an override either. `DeclaredAccessorParitySpec` now compiles thirteen shapes and checks the reflective side against each.

reflection/src/main/java/io/micronaut/reflection/ReflectionBeanIntrospection.java
dstepanov: Reworked as an override relation (`7798c931db`): a declaration is hidden only by a method overriding it as a member of the bean type, and a generic declaration is matched through `MethodHierarchy.overrides`, widened to package-private, so `save(T)` overridden by `save(String)` is hidden and the declaration of an unrelated interface is not. Both shapes are in `DeclaredAccessorParitySpec`, checked against the Java processor: `StringRepo extends Repo<String>` and `class Both implements Left, Right` come out as the generated side reports them.

reflection/src/test/groovy/io/micronaut/reflection/ProxyIntrospectionSpec.groovy
dstepanov: Ran it on JDK 25: the assertion holds, the generated introspection of `ProxiedParityDeclaredBean` reports no property, and the Java processor answers the same (`DeclaredAccessorParitySpec`, `class Impl implements A`). The merge you quote happens, but `AbstractAnnotationMetadataBuilder#addAnnotations` then keeps only the non-declared annotations that are `@Inherited`, which `Introspected.Property` is not, so the override carries nothing. The Groovy override without `@Override` is added as `ProxiedParityDeclaredPlainBean` (`7798c931db`): same answer on both sides. 424 tests pass locally.

reflection/src/test/groovy/io/micronaut/reflection/ProxyIntrospectionSpec.groovy
dstepanov: Dropped the `invoke` line and reworded the case (`7798c931db`): what it shows is `getName` and `describe` reaching the bean methods while `helper` does not.

PR #13184, opened 2026-09-14
Resolve Python compiler error dump directory lazily

graemerocher:
## What changed

The Java compiler in `inject-python` declared a static default for its error dump directory built from `System.getProperty("user.home")` (`~/.pyronaut/processor-error-dumps`). The `io.micronaut.python.compiler` package is initialised at build time in native images, so that path was baked in from the image build host and every processing failure at runtime reported `Full error details could not be written: <build host home>`.

This PR removes the static default and resolves the dump directory when a dump is actually written:

- An explicitly configured `errorDumpDirectory` (builder option / setter) still wins.
- Disk compilation defaults to a `processor-error-dumps` directory **next to** the target directory (e.g. `build/processor-error-dumps` for a target of `build/classes`), mirroring where the incremental cache already lives. It is deliberately kept out of the class output so a later successful build cannot package stale diagnostics.
- In-memory compilation (no target directory) writes to a freshly created private temporary directory via `Files.createTempDirectory`, rather than a shared predictable path under `java.io.tmpdir`.
- Nothing under `user.home` is used anywhere, and no static field in the compiler holds an environment-derived path.

Javadoc on `setErrorDumpDirectory` / `Builder.errorDumpDirectory` documents the new defaults.

## Tests

`PyronautCompilerErrorHandlingTest` gains:

- a disk compilation with no `errorDumpDirectory` asserting the dump lands next to the target directory, the thrown message names that path, and nothing is written inside the target directory;
- an in-memory compilation asserting the dump is written to a fresh private temp directory;
- a unit check of the default derivation for a relative target path;
- a reflective guard over the compiler's static fields that fails if any `File`/`Path`/`CharSequence` value contains `user.home` or `user.dir` (verified to fail against the previous code).

## Verification

- `:micronaut-inject-python:test` — 125 tests pass.
- `:micronaut-inject-python-test:test` — full suite (623 tests) passed before the last refactor; the error-handling class was re-run after it (10/10 pass). Note these tasks are gated behind `-Ppython-ci`.
- `checkstyleMain` passes for `inject-python`. There is one pre-existing `InnerTypeLast` warning (`closeClassLoader` after an inner class) that this PR does not touch.

## Review notes

An independent review flagged two issues in the first cut, both addressed here: the original default placed dumps inside the class output directory, and the in-memory fallback used a shared fixed path under `java.io.tmpdir`.


dstepanov (approved):
Approving — the core change is right. Dropping `DEFAULT_ERROR_DUMP_DIRECTORY` and resolving the path inside `writeDump` fixes the build-time-initialisation problem, and threading `outputDirectory` down from `compileJavaIsolated` is the minimal way to do it. The now-null default field is safe: `errorDumpDirectory` is read only in `resolveErrorDumpDirectory` (`PyronautJavaCompiler.java:775`), nowhere else.

Three follow-ups, none blocking the intent of the fix.

**1. The new default test will fail on macOS** — `PyronautCompilerErrorHandlingTest.java:159`

`TrackingJavaFileManager` canonicalises the target with `toRealPath()` (`TrackingJavaFileManager.java:46`), so `outputDirectory` — and therefore the dump path in the exception message — is the real path. The test builds `expectedDumpDirectory` from the raw `@TempDir`, and `assertConciseSyntaxError` asserts `message.contains(dumpDirectory.getAbsolutePath())` (line 236). On macOS `@TempDir` is under `/var/folders/…`, a symlink to `/private/var/folders/…`, so the message contains `/private/var/...` and that assertion fails. `countDumpFiles` still passes since it resolves through the symlink, so only the message check breaks. The existing tests are unaffected because they pass an explicit `errorDumpDirectory`, which never goes through `toRealPath`. Canonicalising the expected path in the test (`buildDirectory.toRealPath().resolve("processor-error-dumps")`) fixes it.

**2. In-memory compilation leaks a temporary directory per failure** — `PyronautJavaCompiler.java:783`

`Files.createTempDirectory(...)` runs on every dump, so a long-lived JVM (build daemon, test suite, IDE) that hits N in-memory failures leaves N directories under `java.io.tmpdir`, never cleaned; the previous behaviour reused one directory. Private-per-JVM is the right call — creating it once and caching it in a field (lazily, guarded) keeps the security property and bounds the leak at one directory.

**3. The "never inside the class output" guarantee does not hold for real build layouts** — `PyronautJavaCompiler.java:795`

`defaultErrorDumpDirectory` takes the target's parent. The example in the description (`build/classes` → `build/processor-error-dumps`) is the shallow case; with a conventional Gradle layout the target is `build/classes/java/main`, so dumps land in `build/classes/java/processor-error-dumps` — inside the class-output tree, which is what the javadoc says it avoids. Conversely, a caller passing the build directory itself as the target puts dumps in the project root as `./processor-error-dumps`, which survives `clean`. Only the shallow shape is covered by a test.

Worth deciding explicitly: either soften the javadoc to "next to the target directory" without the packaging guarantee, or prefer the configured `incrementalCacheDirectory` when one is set (already validated as non-overlapping with the target in `IncrementalCompilation.java:1288`) and fall back to the sibling. The description says this mirrors where the incremental cache lives, but that location is caller-supplied rather than defaulted to a sibling, so the justification does not quite hold as written.

**Minor**

- `javaCompilerHasNoStaticFieldDerivedFromUserHome` is a good guard, but it is brittle when `user.home` is very short: in containers where `HOME=/`, `text.contains("/")` matches every static `String` in the class. Skipping values shorter than a couple of characters, or asserting on `File`/`Path` fields only, would harden it.
- The dump directory now sits outside the target directory, so sandboxed builds that permit writes only under the declared output will silently fail to write dumps. That degrades gracefully via `DumpResult.failed`, but it is a behaviour change worth a line in the javadoc.

---

PR #13186, opened 2026-09-14
Make StreamPressureSpec independent of the reader thread

graemerocher:
Fixes the flaky `StreamPressureSpec.'consumer pressure'` feature seen on Java CI `build (25)` on 14 Sept 2026 across unrelated PRs (#13105, #13118, #13124, #13130), failing at `StreamPressureSpec.groovy:61` with `java.io.IOException: Read end dead` (sometimes `Pipe closed`).

### Mechanism

The feature fed the streamed response body through a `PipedInputStream`. Piped streams tie liveness to threads: `PipedInputStream` records the thread that last read from it, and `PipedOutputStream.write` throws `Read end dead` as soon as that thread has terminated.

The server reads `InputStream` response bodies via `InputStreamByteBody.toByteArrayPublisher`, which runs each `request(n)` on `Schedulers.fromExecutor(TaskExecutors.BLOCKING)`. On JDK 21+ the blocking executor is the thread-per-task **virtual thread** executor, so every read batch runs on a fresh virtual thread that dies once its batch is served. Instrumenting the stream showed **588 distinct virtual reader threads** for under 1 MiB. Whether the next write from the test thread saw a live or dead reader was a pure race — intermittent on Linux CI, and with `@IgnoreIf(macOs)` removed it fails on the very first iteration on macOS (`Read end dead` from `awaitSpace`).

The macOS "hang" was a second design flaw: the test wrote the whole 1 MiB payload into the 1 KiB pipe *before* reading anything from the client, so it could only complete if the entire payload fit into the netty + loopback socket buffers between server and client.

### Fix

- Replace the piped streams on both sides with a small queue-backed `InputStream` that does not care which thread reads from it; feeding the server never blocks the test thread, so no buffer-size dependency.
- The client side subscribes with a `BaseSubscriber` that requests **one chunk only when the test thread actually reads**, so a stalled consumer still translates into real backpressure on the server (the original got its "pressure" by blocking the client event loop in the 1 KiB pipe).
- Reads time out after 30 s instead of hanging, so a regression fails the build.
- Removed the macOS `@IgnoreIf`.

### Verification (macOS, JDK 25, virtual-thread blocking executor)

- Before (original spec, macOS ignore dropped, `@RepeatUntilFailure(maxAttempts = 25)`): fails on iteration #0 with `Read end dead`.
- After: `@RepeatUntilFailure(maxAttempts = 25)` → 25/25 pass (~1 s each); plus 20 separate `./gradlew :micronaut-http-server-netty:test --tests '*StreamPressureSpec*'` invocations → 20/20 pass.

Unrelated to #13185 (`ConcurrentFormTransferSpec` timeout).


http-server-netty/src/test/groovy/io/micronaut/http/server/netty/stream/StreamPressureSpec.groovy
```diff
@@ -37,57 +42,128 @@ class StreamPressureSpec extends Specification {
...
+        private static final byte[] EOF = new byte[0]
+
+        private final BlockingQueue<byte[]> chunks = new LinkedBlockingQueue<>()
+        private byte[] current
+        private int position
+        /**
+         * Invoked whenever the reader runs out of buffered data and is about to wait for the next
+         * chunk.
+         */
+        Runnable demand = {}
```
dstepanov, line 122: Nit: `demand` is written from the reactor subscription thread (in `hookOnSubscribe`, line 68) and read from the test thread inside `read()` without any shared synchronization/`volatile` on this field — `read()`'s `synchronized` doesn't help since the writer never synchronizes on this monitor. `clientError` right below is correctly made an `AtomicReference` for the same cross-thread-visibility reason; might as well mark `demand` `volatile` too so this test isn't trading one benign-in-practice race for another. Not blocking given this is test-only scaffolding and the reactor/netty handoff will almost certainly provide the necessary memory barrier in practice.

---

dstepanov (approved):
Solid fix for a real bug: `PipedInputStream`/`PipedOutputStream` tie liveness to the reading thread, and `InputStreamByteBody` reads response bodies via `request(n)` on the virtual-thread blocking executor, so every batch runs on a fresh thread — a pure race with piped streams. Replacing both sides with a queue-backed `InputStream` that doesn't care which thread reads, and driving client-side demand only when the test thread actually reads (so the server sees real backpressure instead of the old trick of blocking the client event loop against a 1 KiB pipe), correctly reproduces the "consumer pressure" scenario without the thread-liveness hazard. The 30s read timeout is a good addition too — a regression now fails fast instead of hanging.

I traced through `ChunkQueueInputStream.read(byte[], int, int)` including the EOF-sentinel handling and it's correct (the `while` loop re-checks after consuming a fresh chunk, so an empty EOF array is caught and returns -1 on the next iteration rather than being treated as a 0-length read). `close()` is idempotent since `current` is never reassigned once it holds the EOF sentinel.

Left one non-blocking nit inline about the `demand` field's cross-thread visibility. Given this is test-only scaffolding and matches the PR's own verification (25/25 and 20/20 passes locally), approving.

---

PR #13188, opened 2026-09-15
Do not apply the declaring class advice to a generated @Adapter bean

dstepanov:
## Problem

For a method annotated with `@Adapter` advice (`@EventListener` is the only one in Core), the compiler generates a separate bean that implements the adapted SAM interface and delegates to the method (`DeclaredBeanElementCreator#visitAdaptedMethod`). The adapter's annotation metadata was built from the declaring class merged with the adapted interface, so class-level interceptor advice on the declaring bean reached the generated adapter too.

Reproduced from micronaut-jakarta-interceptors: a bean carrying a lifecycle interceptor and declaring an `@EventListener` method ended up with **two** interceptor instances for one object, one for the bean and one for the generated adapter. The Jakarta Interceptors specification requires one interceptor instance per target instance. That module works around it by giving an empty chain to any metadata carrying `@Adapter`, an annotation lookup on every invocation that guesses at Core's generated shapes.

The Core-only reproduction in this PR (class-level `@Around` advice, prototype interceptor, one published event) showed, before the fix:
- 2 interceptor instances, one for `LoggedEventListenerBean`, one for `LoggedEventListenerBean$ApplicationEventListener$onEvent1`
- the interceptor invoked twice: on the adapter's `onApplicationEvent` and on the bean's `onEvent`
- the adapter's bean definition reporting the class-level advice annotation

`@Scheduled` is not affected: it is `@Executable` + `@Parallel` advice, not `@Adapter`, and generates no adapter bean. The spec covers it to pin that down.

## Change

When the adapter's metadata is built, the interceptor advice inherited from the declaring class (annotations with an `@Around`, `@Introduction` or `@InterceptorBinding` stereotype) is removed, together with the interceptor binding entries it contributed.

What the adapter keeps, and why:
- **Advice declared on the adapted method** was never copied to the adapter. It stays on the declaring bean and still runs, once, when the adapter delegates, so `@Retryable` on an `@EventListener` method keeps retrying. A test covers this.
- **Advice declared on the adapted SAM interface** is kept, since it describes the adapter itself.
- Everything else (scope, `@Adapter`, `@Indexed`, `@Order`, requirements) is unchanged.

Class-level advice now runs once, on the declaring bean, instead of wrapping both the adapter and the bean. Someone relying on it also wrapping the adapter's SAM method would see that change; the observable effect of the advice on the adapted method is otherwise the same.

## Contract considered and not added

A public way to ask whether a bean definition is a generated adapter (a utility or a `BeanDefinition` flag) was considered. With the advice no longer copied, the reported case needs no check at all, and nothing in Core would call such an API, so it is left out. It can be added when a module has a use for it.

## Tests

- New `ClassLevelAdviceOnAdaptedMethodSpec` (inject-java): class-level advice applied once on the declaring bean and absent from the adapter; method-level advice still runs when the adapter delegates; `@Scheduled` generates no adapter.
- inject-java, inject-groovy, inject, context and aop suites pass.
- `./gradlew check --continue` passes apart from HTTP tests that failed intermittently (`BinaryWebSocketSpec`, `StreamRequestSpec`, `ClientRedirectSpec`, JDK client TCK `RedirectTest`); a different set failed on each run and all pass when rerun alone.


core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -725,6 +740,55 @@ private void visitAdaptedMethod(MethodElement sourceMethod) {
...
+            .stream()
+            .filter(av -> av.stringValue().filter(adviceFromDeclaringClass::contains).isEmpty())
+            .map(av -> (AnnotationValue<Annotation>) av)
+            .toList();
+        proxyAnnotationMetadata.removeAnnotation(AnnotationUtil.ANN_INTERCEPTOR_BINDINGS);
+        proxyAnnotationMetadata.removeAnnotation(AnnotationUtil.ANN_INTERCEPTOR_BINDING);
+        proxyAnnotationMetadata.removeStereotype(AnnotationUtil.ANN_INTERCEPTOR_BINDINGS);
+        proxyAnnotationMetadata.removeStereotype(AnnotationUtil.ANN_INTERCEPTOR_BINDING);
+        for (AnnotationValue<Annotation> binding : remainingBindings) {
+            proxyAnnotationMetadata.addDeclaredRepeatable(AnnotationUtil.ANN_INTERCEPTOR_BINDING, binding);
```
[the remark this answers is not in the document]
dstepanov: Confirmed and fixed in b48c63bf5c: the remaining bindings are now re-added under `AnnotationUtil.ANN_INTERCEPTOR_BINDINGS`, and a new test puts advice on both the adapted SAM interface and the declaring class to check the interface advice is kept.

core-processor/src/main/java/io/micronaut/inject/processing/DeclaredBeanElementCreator.java
```diff
@@ -725,6 +740,55 @@ private void visitAdaptedMethod(MethodElement sourceMethod) {
...
+     * @param interfaceToAdapt        The adapted interface
+     */
+    private static void removeInheritedInterceptorAdvice(MutableAnnotationMetadata proxyAnnotationMetadata,
+                                                         ClassElement interfaceToAdapt) {
+        Set<String> adviceFromDeclaringClass = new LinkedHashSet<>();
+        for (String stereotype : ADVICE_STEREOTYPES) {
+            adviceFromDeclaringClass.addAll(proxyAnnotationMetadata.getAnnotationNamesByStereotype(stereotype));
+        }
+        // Anything the adapted interface declares itself describes the adapter and is kept
+        adviceFromDeclaringClass.removeAll(interfaceToAdapt.getAnnotationMetadata().getAnnotationNames());
```
graemerocher: This only excludes annotation names declared directly on the adapted interface. If the declaring class has `@Logged` and the interface has `@Composed`, where `@Composed` is meta-annotated with `@Logged`, the interface metadata names `Composed` but the merged advice set also contains `Logged`. `Logged` therefore remains in `adviceFromDeclaringClass`, and the subsequent `removeStereotype(Logged)` strips the interface's composed advice and interceptor binding. I reproduced this with an in-memory processor test: the generated adapter still has `@Composed`, but no longer has the `Logged`/`Around` stereotype. Please preserve all advice/stereotype names contributed by `interfaceToAdapt` (not only its direct annotation names) and add a composed-advice regression test.

dstepanov: Fixed in 0893605190: the interface's stereotype names (`getStereotypeAnnotationNames()`) are now kept as well as its direct annotation names, and `AdapterInheritedAdviceSpec` has a composed-advice regression test (interface `@Composed` meta-annotated with `@Logged`, class `@Logged @Timed`) that fails without the fix: the adapter keeps `Composed`, the `Logged`/`Around` stereotypes and the `Logged` binding, and still drops `Timed`.

graemerocher (approved):

PR #13190, opened 2026-09-15
Let a buffer consumer receive the final bytes and the completion as one signal

graemerocher:
Split out of #13113 at @yawkat's request; this is the `http` module half. The Netty handler half that turns it into fewer outbound messages is stacked on top as a separate PR, the `LeakTracker` fix and the JMH benchmark that were also in #13113 are their own PRs.

## Summary

- `BufferConsumer` gains a default `addAndComplete(ReadBuffer)` that is semantically `add` followed by `complete`. A consumer may override it to combine the two into a single downstream message; with the default nothing observable changes, and every existing `BufferConsumer` inherits it.
- `BaseSharedBuffer` implements it: the final bytes are stored and the buffer is marked complete *before* any streaming subscriber learns of the completion, because a subscriber may consume a reserved split from inside its completion callback (the servlet integration does) and has to find the final bytes and a completed state. `BaseSharedBuffer.add` is refactored into `add0(rb, completeAfter)` behind an `addGuarded` hook so that the two subclasses that wrap the add in a concurrency guard (`StreamingNettyByteBody.SharedBuffer`'s event-loop check and `ReactiveByteBufferByteBody.SharedBuffer`'s `submit`) keep exactly the scope they had.
- `ConcatenatingSubscriber` uses it for the trailing separator of a concatenation (the `]` of a streamed JSON array), and folds each leading separator (`[` or `,`) into the same buffer as the `AvailableByteBody` element it precedes via `ReadBufferFactory.compose`, so one element leads to one buffer downstream instead of two. A streaming component body still gets its separator as a separate buffer, as before. `NettyResponseLifecycle`'s override of `forwardComplete` follows the new signature.
- Nothing is ever buffered waiting for a later element: the merged separator is the one *in front* of the element, known the moment the element arrives, and the trailing bytes travel with a completion that has already happened. Server-sent events use `Separators.NONE` and take exactly the code path they took before.

## Review findings carried over from #13113

- @dstepanov: `targets` was snapshotted after `complete0`, but `deferred` was sized in `add0`; a `buffer()` callback subscribing another split reentrantly during completion ran the delivery loop off the end of `deferred` (`IndexOutOfBoundsException` escaping `onComplete`, response hang, leaked copies). The snapshot is limited to `subscribers.subList(0, deferred.size())`; a later subscriber is served by `subscribe0`, since the buffer is complete by then. Their reproducer is `ConcatenatingSubscriberReentrancyTest.splitSubscribedFromABufferingCallbackDuringCompletion`.
- Codex (`gpt-5.6-sol`, medium) on this split: notes that the Netty consumers do not override `addAndComplete` here. By design: that is the stacked PR.

## Testing

- `ConcatenatingSubscriberReentrancyTest` (2): a streaming consumer of a split reads the primary from its completion callback and sees the trailing bytes; a split subscribed from a buffering callback during completion gets the buffered bytes while the earlier streaming split still gets its closing bracket. Both fail without the ordering and snapshot changes.
- `./gradlew :micronaut-http:test`: 1875 tests, 0 failures. `:micronaut-http-server-netty:test` for the streaming, JSON stream and `PipeliningServerHandlerSpec` suites: 53 tests, 0 failures. Checkstyle clean on both modules.


http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
```diff
@@ -336,6 +337,55 @@ private void discardBuffer() {
...
+     */
+    @Override
+    public void addAndComplete(ReadBuffer rb) {
+        // The final bytes are stored and this buffer is marked complete before any streaming
+        // subscriber learns of the completion. A subscriber may consume a reserved split from
+        // inside its completion callback (the servlet integration does), and that consumer has
+        // to find the final bytes in the buffer and a completed state, or it would miss them and
+        // then wait for a completion that is never delivered.
+        List<ReadBuffer> deferred = addGuarded(rb, true);
+        complete0(deferred == null);
```
dstepanov, line 380: Nit, non-blocking: if `complete0` throws (the `IncorrectContentLengthException` for too few bytes, or a buffering subscriber's callback throws), the duplicates in `deferred` are never closed and the streaming subscribers never get the bytes. `ConcatenatingSubscriber` has no expected length, so it can't hit this today, but `addAndComplete` is public on the buffer. Closing whatever wasn't delivered in a `finally` would keep the leak detector quiet if another caller ever uses it.

dstepanov (approved):
LGTM. The store-then-complete-then-deliver ordering in `addAndComplete` covers the reentrant servlet consumer, and limiting delivery to `subList(0, deferred.size())` is safe because `subscribers` is only ever appended to. Both paths are pinned by `ConcatenatingSubscriberReentrancyTest`. Leaving `adding` out of the deferred delivery in `StreamingNettyByteBody` matches what `complete()` already does. One non-blocking nit inline.

http/src/main/java/io/micronaut/http/body/stream/BaseSharedBuffer.java
graemerocher: Done as a follow-up since #13190 had merged: the undelivered copies are closed in a `finally`, with a test that fails on the merged code. PR to follow in this thread.

graemerocher: The follow-up is #13253.

PR #13192, opened 2026-09-15
Add a JMH benchmark for streaming a Flux of records

graemerocher:
Split out of #13113 at @yawkat's request; measurement only, no production code.

## Summary

`StreamingResponseBenchmark` streams 200 small records (`{"id":0,"message":"message-0","value":0}`) through an `EmbeddedChannel`, once as a single JSON array and once as an NDJSON stream, alongside the existing `ServerScenariosBenchmark` scenarios. It is the benchmark behind the numbers in #13196.

Two points from the Codex review (`gpt-5.6-sol`, medium) of this split are addressed: the timed exchange compares each outbound part against the expected response as it arrives instead of composing them first (composite comparison costs more than the server work, and its cost depends on how the response is fragmented, which is exactly what #13196 changes), and the setup check asserts that all 200 records are present so a truncated stream cannot become the baseline.

## Testing

- `./gradlew :benchmarks:jmhJar` and a smoke run: `JSON_STREAM_RESPONSE` 335 us/op, `JSON_STREAM_RESPONSE_NDJSON` 232 us/op (1 warmup + 1 measurement iteration, loaded machine; not a measurement).


yawkat (approved):
fwiw EmbeddedChannel is probably not going to be very realistic for streaming

PR #13193, opened 2026-09-15
Tokenize Accept-Encoding in place with scans bounded to the entry

graemerocher:
Split out of #13130 at @yawkat's request; the compression half. The `MediaType.isTextBased` cache and the Jackson per-type cache from #13130 are their own PRs.

## Summary

- `Compressor.determineEncoding` examined the `Accept-Encoding` header through `split(",")` and `split(";")` on every response, allocating an array and a string per entry. The entries are now scanned in place: each comma-separated entry is examined for its `q` value and its encoding name without creating substrings. The result is unchanged for every input the split-based version accepted.
- @dstepanov found the first in-place version quadratic in the header length (the `'='` and encoding-name scans ran to the end of the whole header for every entry: an 8 KiB header of commas cost about 85 ms per request, within the default `maxHeaderSize`). Both scans are now bounded to the current entry, using the JDK's `String.indexOf(String, int, int)` / `indexOf(int, int, int)` overloads (review: no hand-written scanning in `Compressor`).

## Testing

- The selection is a static `determineEncoding(values, Set<Algorithm> available)`; the compressor derives the set once from its options (gzip, deflate, snappy always; brotli and zstd when their native libraries are present).
- `CompressorTest.determineEncodingMatchesTheSplitBasedTokenizer`: 28 header shapes (quality values, `*`, `identity`, empty entries, malformed `q`, several values) checked against a copy of the split-based reference for every one of the 32 subsets of the five algorithms, so each algorithm is both selected and skipped in the run whatever the test class path has; `everyAlgorithmIsSelectedForItsOwnName` pins the plain cases.
- `CompressorTest.longHeadersAreTokenizedInLinearTime`: compares the cost of a 1 KiB header with one eight times as long for four entry shapes. Linear scanning gives a ratio near 8; the unbounded scans gave 43–57 in the fail-before run. The check is a ratio so it does not depend on machine speed (a Codex finding on this split, which also caught that the test strategy returned a zero zstd size that netty rejects when `zstd-jni` is present).
- `./gradlew :micronaut-http-server-netty:test --tests '*Compress*'`: 83 tests, 0 failures, three consecutive runs. Checkstyle clean.


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
```diff
@@ -184,7 +196,33 @@ private Algorithm determineEncoding(List<String> acceptEncoding) {
...
+     * {@code end}, so that a header with many entries stays linear in its length.
+     */
+    private static boolean contains(String s, int start, int end, String needle) {
+        int n = needle.length();
+        for (int i = start; i + n <= end; i++) {
+            if (s.regionMatches(i, needle, 0, n)) {
+                return true;
+            }
+        }
+        return false;
```
yawkat: can this be replaced with some library method? really not something that should be in the Compressor, and there probably is already code for it somewhere

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
```diff
@@ -184,7 +196,33 @@ private Algorithm determineEncoding(List<String> acceptEncoding) {
...
+
+    /**
+     * Index of {@code ch} inside {@code s[start, end)}, or {@code -1}.
+     */
+    private static int indexOf(String s, char ch, int start, int end) {
+        for (int i = start; i < end; i++) {
+            if (s.charAt(i) == ch) {
+                return i;
+            }
+        }
```
yawkat: same here

http-server-netty/src/test/java/io/micronaut/http/server/netty/handler/CompressorTest.java
```diff
@@ -0,0 +1,187 @@
...
+     */
+    private static Compressor.Algorithm referenceDetermineEncoding(List<String> headerValues, boolean brotli, boolean zstd) {
+        float starQ = -1.0f;
+        float brQ = -1.0f;
+        float zstdQ = -1.0f;
+        float snappyQ = -1.0f;
+        float gzipQ = -1.0f;
+        float deflateQ = -1.0f;
+        for (String s : headerValues) {
+            for (String encoding : Arrays.asList(s.split(","))) {
```
yawkat: why asList

http-server-netty/src/test/java/io/micronaut/http/server/netty/handler/CompressorTest.java
```diff
@@ -0,0 +1,187 @@
...
+                    snappyQ = q;
+                } else if (encoding.contains("gzip") && q > gzipQ) {
+                    gzipQ = q;
+                } else if (encoding.contains("deflate") && q > deflateQ) {
+                    deflateQ = q;
+                }
+            }
+        }
+        if (brQ > 0.0f || zstdQ > 0.0f || snappyQ > 0.0f || gzipQ > 0.0f || deflateQ > 0.0f) {
+            if (brQ != -1.0f && brQ >= zstdQ && brotli) {
```
yawkat, line 82: don't we also have to check all the other algos, not just zstd

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
graemerocher: Replaced in 42da835deb with the JDK's own bounded overloads, `String.indexOf(String, int, int)` and `indexOf(int, int, int)` (JDK 21+); both helpers are gone.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/Compressor.java
graemerocher: Replaced in 42da835deb with the JDK's own bounded overloads, `String.indexOf(String, int, int)` and `indexOf(int, int, int)` (JDK 21+); both helpers are gone.

http-server-netty/src/test/java/io/micronaut/http/server/netty/handler/CompressorTest.java
graemerocher: No reason; iterating the array directly now (a3b1175068).

http-server-netty/src/test/java/io/micronaut/http/server/netty/handler/CompressorTest.java
graemerocher: Right. In a3b1175068 the selection is a static `determineEncoding(values, Set<Algorithm> available)` (the instance method passes the set it derives from its options once), and the parameterized test runs every header shape against all 32 subsets of the five algorithms, so each of br/zstd/snappy/gzip/deflate is both selected and skipped in the run regardless of which native codecs the test class path has. `everyAlgorithmIsSelectedForItsOwnName` pins the plain cases.

http-server-netty/src/test/java/io/micronaut/http/server/netty/handler/CompressorTest.java
```diff
@@ -0,0 +1,222 @@
...
+        String longHeader = entry.repeat(shortRepeats * 8);
+        // interleave the two so that JIT and GC affect both alike; take the fastest round for each
+        long shortBest = Long.MAX_VALUE;
+        long longBest = Long.MAX_VALUE;
+        for (int round = 0; round < 10; round++) {
+            shortBest = Math.min(shortBest, timePerCall(shortHeader, all));
+            longBest = Math.min(longBest, timePerCall(longHeader, all));
+        }
+        double ratio = (double) longBest / Math.max(shortBest, 1);
+        assertTrue(ratio < 24, "tokenizing " + longHeader.length() + " chars took " + ratio + " times as long as " + shortHeader.length() + " chars");
```
dstepanov, line 173: Non-blocking: this is still a wall-clock assertion. Taking the best of 10 rounds and allowing up to 24× against an expected 8× (64× if quadratic) gives it plenty of margin. The short header costs only a few µs per call, though, so a GC pause or a descheduled CI thread during all 10 short rounds could still trip it. If it ever flakes, counting `indexOf` calls or scanned chars (or tagging it so it stays out of the default run) would make it deterministic.

dstepanov (approved):
LGTM. Every scan is now bounded to its entry (`indexOf(ch/str, begin, end)`), so the total work is linear. The in-place loop matches `split(",")` on the edge cases: trailing and leading empty entries, multiple `=`, and whitespace around `q` (which `parseFloat` trims). Having the old tokenizer as the reference, run against all 32 subsets of the available algorithms, covers yawkat's point about checking every algorithm. One non-blocking remark on the timing test inline.

PR #13194, opened 2026-09-15
Cache MediaType.isTextBased

graemerocher:
Split out of #13130 at @yawkat's request; the `MediaType` half.

## Summary

`MediaType.isTextBased()` matched the type name against five regular expressions and compared the subtype against seven constants on every call, and the compression check calls it for every response. The answer only depends on final fields, so it is computed once per instance with plain `startsWith`/`endsWith` checks (case-sensitive on the name, case-insensitive on the subtype, as before) and cached in a nullable `Boolean` field; a racy write of the same value from two threads is harmless. The `Pattern` list and its static initializer are gone.

## Testing

- `MediaTypeTest`: the new checks are compared against the original regular expressions for every `MediaType` constant and a set of edge cases (`+json`/`+xml` suffixes, mixed-case subtypes, parameters, `text/` prefix without a subtype match, repeated access returning the cached value).
- `./gradlew :micronaut-http:test --tests '*MediaType*'`: 228 tests, 0 failures. Checkstyle clean. Codex (`gpt-5.6-sol`, medium): no findings.


http/src/main/java/io/micronaut/http/MediaType.java
```diff
@@ -1256,18 +1253,28 @@ public CharSequence subSequence(int start, int end) {
...
-        if (!matches) {
-            matches = subtype.equalsIgnoreCase("json")
-                    || subtype.equalsIgnoreCase("xml")
-                    || subtype.equalsIgnoreCase("yaml")
-                    || subtype.equalsIgnoreCase("graphql")
-                    || subtype.equalsIgnoreCase("yang")
-                    || subtype.equalsIgnoreCase("toml")
-                    || subtype.equalsIgnoreCase("x-cue")
-            ;
+        byte cached = textBased;
```
yawkat: I'd just use a nullable Boolean for this tbh

yawkat (approved):

http/src/main/java/io/micronaut/http/MediaType.java
graemerocher: Done in 1893630de3: a `@Nullable Boolean` field, the three byte constants are gone.

PR #13195, opened 2026-09-15
Cache Jackson readers and writers per type

graemerocher:
Split out of #13130 at @yawkat's request; the Jackson half.

## Summary

`JacksonDatabindMapper.createReader`/`createWriter` built an `ObjectReader`/`ObjectWriter` (through `JacksonConfiguration.constructType`) on every call, keyed by nothing: the `Argument` instance a route binds is new per request for anything generic. They are now cached per type in a bounded 64-slot array (one each for readers and writers; plain arrays, since an entry is an immutable record and a racy reader at worst recomputes an equal value), keyed by `Argument.typeHashCode()` with the `@JsonView` class mixed in, and matched with `equalsType` plus the view, so wildcards, placeholders and views are all part of the key. A collision evicts; nothing is ever wrong, only recomputed. The caches are only allocated by the general constructor: a specialized mapper (`createSpecific`, one per route; `cloneWithViewClass`) answers every lookup with its own reader and writer and never reaches them.

## Review findings carried over from #13130

- @dstepanov verified the key (`equalsType` compares what `constructType` reads, views included) with an 8-thread interleaving test, and that specialized/cloned mappers get no stale entries after reconfiguration.
- Copilot: specialized mappers allocated two unused 64-slot arrays per route; now `null` there.
- @dstepanov: `severalTypesAreCachedAtOnce` depended on identity-hash slot placement and failed about one run in eleven. The test now picks four candidate types whose slots differ (via a package-private `slotOf`) and also checks that an equal, distinct `Argument` instance hits the same entry.

## Testing

- `JacksonDatabindMapperTypeCacheTest` (7): same type hits, different type arguments miss, several types cached at once, `@JsonView` is part of the key, specialized mappers ignore the cache, `cloneWithFeatures` gets its own cache.
- `./gradlew :micronaut-jackson-databind:test`: 66 tests, 0 failures. Checkstyle clean on main and test. Codex (`gpt-5.6-sol`, medium): no findings.


jackson-databind/src/main/java/io/micronaut/jackson/databind/JacksonDatabindMapper.java
```diff
@@ -85,10 +94,16 @@ public final class JacksonDatabindMapper implements JsonMapper {
...
 
+    /**
+     * Per-type caches, only allocated for a general mapper. A specialized mapper (one with a
+     * {@link #specializedReader} and {@link #specializedWriter}) answers every lookup with those
+     * two and never touches the caches, and one such mapper is created per route, so they are
+     * {@code null} there.
+     */
     @Nullable
-    private TypeCache<ObjectReader> cachedReader;
+    private final AtomicReferenceArray<TypeCache<ObjectReader>> cachedReaders;
```
yawkat: i dont think it needs to be atomic

graemerocher: Agreed, plain arrays now (84a5c54086): the entries are immutable records and reference writes are atomic, so a racy reader at worst recomputes an equal value. The field javadoc says so.

yawkat (approved):

PR #13196, opened 2026-09-15
Write the trailing bytes of a streamed response as part of the terminating message

graemerocher:
Stacked on #13190 (the `addAndComplete` API); split out of #13113 at @yawkat's request. This is the part that changes what goes on the wire.

## Summary

- `PipeliningServerHandler.StreamingOutboundHandler` overrides `addAndComplete` to write the final bytes of a streamed response (the `]` of a JSON array) as the `DefaultLastHttpContent` that terminates the response, instead of a content message of their own followed by an empty terminator. Outside the steady state (response not started, connection gone) it falls back to `add0` + `complete0`.
- With the leading separators already folded into their elements by #13190, a streamed JSON array of N elements is **N+1** outbound messages instead of **2N+2**; the bytes on the wire are unchanged (a `LastHttpContent` carrying `]` encodes to the same `1\r\n]\r\n0\r\n\r\n` under chunked encoding).
- HTTP/2 needs no counterpart: Netty's encoder merges DATA frames queued in the same event loop tick, including the empty END_STREAM frame that follows the trailing bytes, so the stream already ends with the frame carrying them. `Http2ServerHandlerSpec` records the data frames of a concatenated response (`[1`, then `,2]` with END_STREAM) to keep it that way; Codex raised this as a gap on this split and the test is the answer.

### Outbound message and flush counts

Measured with an `EmbeddedChannel` and a handler between `PipeliningServerHandler` and the codec (`StreamingResponseChunkSpec`):

| case | base messages | branch messages | base flushes | branch flushes |
|---|---|---|---|---|
| 3-element JSON array, produced inside the read cycle | 8 (`[`, `1`, `,`, `2`, `,`, `3`, `]`, ``) | 4 (`[1`, `,2`, `,3`, `]`) | 1 | 1 |
| empty JSON array | 2 (`[]`, ``) | 1 (`[]`) | 1 | 1 |
| JSON array, first element produced asynchronously | 2 (`[`, `1`) | 1 (`[1`) | 2 | 1 |
| SSE, per event | 1 | 1 | 1 | 1 |

Flushes only differ when the element is produced outside a read cycle, because the handler already defers flushes to `channelReadComplete` while reading; that asynchronous case is the normal one for a reactive endpoint.

### Measurement

From `StreamingResponseBenchmark` (#13192, 200 small records through the embedded channel), base against the original branch, JMH `-f 1 -wi 5 -w 3s -i 5 -r 3s -t 1 -bm avgt -tu us -prof gc`, three runs each alternating base and branch on an idle machine (medians; each run's error in brackets):

| scenario | base | branch | change |
|---|---|---|---|
| JSON array, 200 records | **352.5** us/op (352.0 ±14.0, 352.5 ±5.0, 355.4 ±4.2) | **244.8** (247.9 ±3.5, 229.1 ±3.3, 244.8 ±3.8) | −31% |
| JSON array, 200 records, alloc | **486,660** B/op | **395,602** B/op | −19% |
| JSON stream, 200 records (control, `Separators.NONE`) | 220.3 us/op | 220.2 us/op | no change |

## Testing

- `StreamingResponseChunkSpec` (4 features): the cases in the table above, including that an asynchronously produced element is written and flushed as one message and that SSE events are not held back.
- `Http2ServerHandlerSpec` gains the concatenated-response feature (19 features, 0 failures); `PipeliningServerHandlerSpec` and the JSON stream specs pass (57 tests in the run). Checkstyle clean.


http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
```diff
@@ -1474,6 +1474,31 @@ private void add0(ReadBuffer buf) {
...
+                complete0();
+                return;
+            }
+
+            // the final bytes go out as the LastHttpContent that terminates the response, instead
+            // of a content message of their own followed by an empty terminator
+            outboundHandler = null;
+            writeCompressing(new DefaultLastHttpContent(NettyReadBufferFactory.toByteBuf(buf)), true, true);
+            writtenLast = true;
+            requestHandler.responseWritten(outboundAccess.attachment);
```
dstepanov, line 1498: Since #13102, `complete0` goes through `markResponseWritten()`, which sets `responseWritten` and forwards only the first call. This path calls `requestHandler.responseWritten` directly, so the flag stays false. I don't see a second call today, because `outboundHandler` is already null and `discardOutbound` won't reach this handler. Still, using `markResponseWritten()` here keeps the exactly-once guarantee in one place:

```suggestion
            markResponseWritten();
```

dstepanov (approved):
LGTM, merge after #13190. When the handler isn't in its steady state, the fallback runs `add0(buf); complete0()`, which does exactly what the default `addAndComplete` did, so nothing changes outside the fast path. `writeCompressing0` already pushes the content of a last message and then finishes, so compressed responses are fine too.

This stacked PR only gets the CLA check, so I ran it locally at 4ad3e90: `:micronaut-http-server-netty:test` filtered to `StreamingResponseChunkSpec`, `handler.*` (including `Http2ServerHandlerSpec`), `*Compress*` and `*Stream*`. 222 tests, 0 failures. One small consistency point inline.

http-server-netty/src/main/java/io/micronaut/http/server/netty/handler/PipeliningServerHandler.java
graemerocher: Applied in 806c3b6573.

PR #794, opened 2018-10-26
Issue-787: Do not auto configure DataSourceTransactionManagerFactory …

RaviH:
…if  is present.

PR #13198, opened 2026-09-16
Use the shared Python compiler Gradle plugin from micronaut-build

graemerocher:
Depends on micronaut-projects/micronaut-build#939 and #940 (both released in micronaut-build 8.1.1).

The Python compiler Gradle plugin (`PythonCompile`, `PythonVfsBytecodeCompile`, …) moved from `buildSrc` to the shared build plugins as `io.micronaut.build.internal.python`, so every Micronaut module can add a `test-suite-python` project for its documentation snippets.

Core keeps a thin `io.micronaut.build.internal.convention-python` script plugin which applies the shared plugin with `micronautBuild.python.compilerVersion = ''` and wires the compiler configuration to `projects.micronautInjectPython` / `projects.micronautContextPython`, i.e. the compiler built by this repository rather than the published one.

Requires micronaut-build **8.1.1** (released): the settings plugin and `micronaut-build-plugins` catalog version are bumped to 8.1.1.

Verified locally: `:test-suite-python:compileTestPython`, `:benchmarks:compilePython` and `:test-suite-python:test -Ppython-ci=true` (e.g. `PetControllerSpec`, `CustomScopeSpec`) pass.

Also replaces core's hand-written `python.yml` with the project-template one (micronaut-projects/micronaut-project-template#794), which runs the root `pythonCheck` task; the root build adds `inject-python`, `inject-python-test`, `context-python` and `context-python-netty` to it. `benchmarks` has no test sources, so it only contributes `:benchmarks:check` (a no-op) to `pythonCheck`.


benchmarks/build.gradle.kts
```diff
@@ -1,9 +1,16 @@
...
+    id("io.micronaut.build.internal.convention-python")
     id("me.champeau.jmh") version "0.7.3"
 }
 
+micronautBuild {
+    python {
+        // the benchmark tests are part of the regular CI, not of the Python CI
+        testsEnabled = true
+    }
+}
```
dstepanov: Nit: this block has no effect. `benchmarks` has only `src/jmh` and no test sources, and the Python CI log shows `:benchmarks:test NO-SOURCE`, so the comment about benchmark tests in the regular CI (and the same line in the PR description) is misleading. I'd drop it. Also note `:benchmarks:check` now joins `pythonCheck`, since the plugin adds every project that applies it. That's harmless here.

.github/workflows/python.yml
```diff
@@ -1,87 +1,122 @@
...
       - '[0-9]+.[0-9]+.x'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - master
+      - main
       - '[0-9]+.[0-9]+.x'
       # stacked Python fix PRs target each other's branches
       - 'python/**'
       - 'python-*'
```
dstepanov: The header says this file is synced from the project template, but these stacked-PR branch filters aren't in micronaut-projects/micronaut-project-template#794 (which is still open). The first sync after #794 merges will remove them, so either add them to #794 or accept losing them.

dstepanov (approved):
LGTM. The moved classes match `io.micronaut.build.python` in micronaut-build 8.1.1 apart from package, headers and javadoc, and Python CI's `pythonCheck` ran all five Python projects with 1509 tests, the same count as the latest 5.2.x run. Two non-blocking nits inline.

benchmarks/build.gradle.kts
graemerocher: Dropped the block (`benchmarks` has no test sources, so it was a no-op) and corrected the description — it now says `benchmarks` only contributes a no-op `:benchmarks:check` to `pythonCheck`.

.github/workflows/python.yml
graemerocher: Added the `python/**` / `python-*` base filters to the template's `python.yml` and `gradle.yml` in micronaut-projects/micronaut-project-template#794 (commit 918442c) and aligned this file's trigger block with the template text, so the first sync after #794 merges keeps them.

Issue #518, opened 2018-08-22
Extending Entities does not seem to work


vahidhedayati:
I have created a sample project which can be [downloaded here](https://github.com/vahidhedayati/micronaut-playground): 

I have setup a [few domain class](https://github.com/vahidhedayati/micronaut-playground/tree/master/shop/products/src/main/groovy/products/domain
) "entity classes"

```
package products.domain
import grails.gorm.annotation.Entity
@Entity
class Product {
    String name
    String description
    Date date
    BigDecimal price
    static mapping = {
        //cache true
        //tablePerConcreteClass false
        tablePerHierarchy false
    }
}
```

```
package products.domain
import grails.gorm.annotation.Entity
@Entity
class Generic extends Product {

    Integer width=0
    Integer height=0

    static mapping = {
        discriminator value: 'generic'
    }

}
```

```
package products.domain
import grails.gorm.annotation.Entity
@Entity
class Chair extends Generic {
    Boolean wheels=false
    static mapping = {
        discriminator value: 'chair'
    }
}
```

```
package products.domain
import grails.gorm.annotation.Entity
@Entity
class Table extends Generic {

    static mapping = {
        discriminator value: 'table'
    }

}
```
So a few classes 
Product 
Generic Extends product 
Chair/Table extend Generic

But when i try to start the app none of the DataLoader items appear to be created, it had been a lot more basic pre all the static mappings and did not appear to work. 

Apologies if it is an existing raised question

vahidhedayati:
This is absolutely my bad - there were a few issues with above project. 

1. I had a domain class called Table which in short is only going to clash with `create table table`

2. The main actual issue behind this is when you define DataLoader file this file must have `@Singleton` defined otherwise the data does not load

PR #13199, opened 2026-09-16
Write bean definitions, introspections and proxies as Java source

dstepanov:
The annotation processors build sourcegen models and write them as bytecode. This makes the same models writable as Java source, which javac compiles in a later round, so the generated bean definitions, introspections, executable methods and AOP proxies can be read and debugged as source.

## How it is switched on

`VisitorContext.visitObjectDef(ObjectDef, Element...)` writes a model, as bytecode by default. The Java implementation writes source instead when the sourcegen **bytecode writer is not on the annotation processor path**: a module that wants source replaces it there with a source generator.

```kotlin
listOf(configurations.annotationProcessor, configurations.testAnnotationProcessor).forEach { configuration ->
    configuration {
        exclude(group = "io.micronaut.sourcegen", module = "micronaut-sourcegen-bytecode-writer")
    }
}
```

So the choice is made by the dependencies of the module that opts in, with no flag and no global state. The presence of a generator is deliberately *not* the condition: a module using `@Builder` already has `sourcegen-generator-java` on its processor path and must keep getting bytecode. `sourcegen-generator-java` is a `compileOnly` dependency of `inject-java`, so it never reaches a consumer's regular compilation.

`BeanConfigurationWriter` now writes through `visitObjectDef` like every other generated class, instead of calling the bytecode writer itself.

## What changed in the writers

The models were shaped for the bytecode writer, which only has to satisfy the verifier. These changes make the same model satisfy javac as well, and the bytecode is unchanged:

- A bean definition declares the interfaces it implements with its bean type (`InstantiatableBeanDefinition<Foo>`), so `instantiate` and friends are typed rather than erased, including for a proxy definition and a definition built from an element.
- An `@Inject` method, a lifecycle callback, or an instantiation that throws a checked exception is wrapped and rethrown through `ExceptionUtils.sneakyThrow`.
- A value taken from the context outside a bean lookup is cast to the erasure of its type.
- A dispatch case rethrows the exceptions its method declares; an empty dispatcher's `getTargetMethodByIndex` throws unknown dispatch instead of falling through.
- A proxy target field is `final` only where it is assigned exactly once.
- A class in annotation metadata that is not accessible from the generated class falls back to an `AnnotationClassValue` of its name.
- `Set.of` with more than ten values gets a uniform component type, and `getRequiredArguments()` returns `Argument<Object>[]`.

## test-suite-java-sourcegen

A new module that compiles the sources of `test-suite` with the Java source generator, sharing the source directories rather than depending on `:test-suite`, whose jar holds the bytecode-generated definitions of the same classes. It generates 768 Java sources (bean definitions, introspections, executable methods, proxies and bean configurations) and runs the same tests.

## Reviewer notes

- **Blocked on [micronaut-sourcegen#518](https://github.com/micronaut-projects/micronaut-sourcegen/pull/518).** The Java generator needs those fixes. Until they are released, the last commit points the build at a local `2.2.1-SNAPSHOT` through `mavenLocal()`; drop that commit once a release contains them. CI cannot resolve the snapshot, so it will fail until then.
- `visitObjectDef` is `@Internal`; the processing option constant is `@Experimental`.

## Tests

| Run | Tests | Failing |
|---|---|---|
| `test-suite` (bytecode) | 493 | 0 |
| `test-suite-java-sourcegen` (Java source) | 493 | 0 |
| `inject-java` | 5,328 | 0 |
| `inject-groovy` | 1,142 | 0 |
| `inject-kotlin` | 896 | 0 |

The inject modules last ran against the sourcegen snapshot from before the final review fixes of #518; the two suites ran against the current one.


graemerocher:
Can you retarget 5.3.x?

PR #13201, opened 2026-09-16
Resolve Java packages under io.* other than io.micronaut

graemerocher:
Java packages under `io.*` other than `io.micronaut` (`io.swagger`, `io.kubernetes`, ...) could not be used from Python.

## Root cause

`io` is Python's built-in module, so a Java package under `io` can only be reached through the compiler's import rewriting, and that rewriting only knew `io.micronaut`:

- The compile-time transformer (`micronaut_transformer.py`) stripped the `io.` prefix of *every* rewritten import in the source it hands to the processor, but the processor (`micronaut_processor.py` / `PythonAstParser`) only restored it for `micronaut.*`. An explicit `from io.swagger.v3.oas.annotations import Operation` was therefore recorded as `swagger.v3.oas.annotations.Operation`, which no longer matched the generated decorator, and the annotation was silently dropped from the generated metadata (the OpenAPI spec lacked it). A star import happened to work at compile time because it relies on the generated decorators alone.
- The runtime transformer (`MicronautRuntimeTransformer`) only rewrote `from io.micronaut... import`, so no form of `io.swagger` / `io.kubernetes` import loaded at runtime: `ModuleNotFoundError: No module named 'io.swagger'; 'io' is not a package`.

## Fix

- The compile-time transformer keeps the Java module name in the source read by the processor; resolved imports are replaced by the generated decorators / `java.type()` aliases exactly as `from micronaut... import` already is. `from io.<x> import subpackage as alias` is treated as a module import (decorators generated for the package's annotation types, the import stays for the processor).
- The runtime transformer rewrites every `from io.<x> import ...` and `import io.<x> as alias` to the generated runtime package without the prefix (`swagger.v3.oas.annotations`), for any Java `io.*` package — the VFS generation already wrote the packages there.
- Since an `io.*` import can never be a Python module, one that names no class or package on the compile classpath is now a clear compilation error (`Cannot resolve Java import [io.swagger.v3.oas.annotations.Missing]: ...`) instead of a silently ignored annotation; `import io.<x>` without an alias is rejected with a hint (the name `io` cannot refer to the Java package at runtime). `import io` / `from io import StringIO` are untouched.
- Short section in the Python guide (`keywords.adoc`) documenting `io.*` package imports.

## Reproducer / tests

`IoPackageImportSpec` (inject-python-test, with `io.swagger.core.v3:swagger-annotations` added as a test dependency) compiles a bean using swagger annotations through explicit, aliased, star and `import ... as` imports, asserts `@Tag` / `@Operation` are present in the bean definition metadata, that a module using all import forms (plus an `io.*` Java enum) loads and runs in the GraalPy context, and the compile errors. Before the fix the explicit-import test failed on `definition.hasAnnotation(Tag)` and the runtime tests failed with `ModuleNotFoundError: No module named 'io.swagger'`. Python unit tests in `test_micronaut_transformer.py` cover the runtime rewriting and the error reporting.

Run: `:micronaut-inject-python:test` (125 tests), `:micronaut-inject-python-test:test` (619 tests) and `:test-suite-python:test --tests '*qualifiers*'` (compiles all doc examples) with `-Ppython-ci`, all green.

## Module tests that can drop their workaround

- micronaut-openapi (micronaut-projects/micronaut-openapi#2830, `docs-examples/DISABLED_TESTS.md`): every module importing `io.swagger.v3.oas.annotations.*` (`Application`, `controllers.annotations.HelloController`, `security.OrderController`, `schemas.*`, `generics.NamedResponseController`, `naming.*`, `resolution.*`, `config.Application`, `groups.Application`) can import the annotations directly and drop the `try: ... except ImportError: from swagger... import *` fallback.
- micronaut-kubernetes (micronaut-projects/micronaut-kubernetes#1155): the `java.type("io.kubernetes...")` aliases (e.g. `informer/ConfigMapInformer.py`) can become `from io.kubernetes.client.openapi.models import V1ConfigMap` imports.


graemerocher:
Superseded by #13238 (same change; the branch was renamed and GitHub closed this PR with it).

PR #13203, opened 2026-09-16
Python: initialize the Python runtime on demand before the @Context bean

graemerocher:
> Stacked on #13238 (part 2/32 of the Python fix stack; merge in order — GitHub retargets each PR when its base merges).

Python beans that Micronaut instantiates before the eager (`@Context`) beans failed with `GraalPy context has not been initialized`. Exposed by the Python test suites of micronaut-projects/micronaut-kafka#1404 and micronaut-projects/micronaut-gcp#1543 (worked around with a Java `PythonRuntimeInitializer` / `TypeConverterRegistrar`), and by tests disabled in micronaut-projects/micronaut-serialization#1425 (`Feature` `TypeConverter`), micronaut-projects/micronaut-security#2359 (`CustomPrincipalConverterTest`) and micronaut-projects/micronaut-kubernetes#1155 (`CustomTypeNameMapperTest`).

### Root cause

Generated Python bean stubs resolve the GraalPy runtime through the static `PythonContextRuntime` entry points (`newInstance(...)` etc.), which only knew the runtime once the `@Context` `@Named("python") org.graalvm.polyglot.Context` bean of `GraalPyContextFactory` had been created. `DefaultBeanContext` starts in the order `initializeTypeConverters()` (all `TypeConverter` beans are instantiated) → `processExecutableMethodsProcessAtStartup()` (`@Executable(processOnStartup = true)` processors such as the Kafka/Pub/Sub listener processors, `@Scheduled`, `ApiReactorExecMethodProcessor` instantiate the beans they process) → `initializeEagerBeans()` (`@Context` beans). A Python `TypeConverter`, a Python listener bean or a Python `TypeNameMapper` is therefore created before the runtime exists, and `PythonApplicationRuntime.require()` threw.

### Fix (`context-python`)

* `PythonRuntimeBootstrapConfigurer` (`@ContextConfigurer`) records the application context that is starting (`ApplicationContextConfigurer.configure(ApplicationContext)` runs in `configureContextInternal()`, before any bean is created).
* `PythonApplicationRuntime.require()` now creates that context's `@Named("python") Context` bean on first use when no runtime is installed yet. The factory method installs the runtime exactly as before, so early Python beans and the rest of the application share the one primary context, and the later eager-bean pass finds the bean already created (no second context, no ordering assumptions).
* `PythonRuntimeBootstrapShutdownListener` clears the record on `ShutdownEvent`, so a stopped application context is never used to build a runtime; a re-entrant request while the context bean is being built (e.g. from `main.py`) keeps the previous failure instead of building a second context.

No module-side initializer is needed any more; the Kafka `PythonRuntimeInitializer` and the GCP `TypeConverterRegistrar` workaround can be removed once this is released.

### Reproducer / regression test

`inject-python-test`: `PythonRuntimeInitOrderSpec` starts an application context with

* a Python `TypeConverter[str, Temperature]` bean (instantiated by `initializeTypeConverters()`), and
* a Python bean with a `@StartupProcessed` (`@Executable(processOnStartup = true)`) method processed by a Java `ExecutableMethodProcessor` fixture that instantiates the bean and invokes the method, the way the Kafka consumer processor does,

and checks that both work, that the Python objects live in the same context as the `@Named("python") Context` bean, that only one context bean exists, and that the stopped context no longer provides a runtime. Before the fix all three cases fail with `GraalPy context has not been initialized`.

One existing assertion in `PythonEventListenerSpec` (`getBeansOfType(ApplicationEventListener).isEmpty()`) was narrowed to the Python listener under test, since the runtime now contributes its shutdown listener bean.

### Module tests that can be re-enabled

* micronaut-serialization#1425: `example-python` `example.Feature` (`keys.adoc`) Python `TypeConverter`.
* micronaut-kubernetes#1155: `CustomTypeNameMapperTest` (`TypeNameMapper` collected by `ApiReactorExecMethodProcessor`).
* micronaut-kafka#1404 / micronaut-gcp#1543: the Java `PythonRuntimeInitializer` / `TypeConverterRegistrar` workarounds become unnecessary.
* micronaut-security#2359 `CustomPrincipalConverterTest`: the converter part is fixed; the test also starts a nested `ApplicationContext.run(...)` inside a `@MicronautTest`, which is a separate issue (the nested context replaces/closes the shared runtime).

### Not covered

A Python test class implementing `TestPropertyProvider` (Kafka `MyTest`): Micronaut Test calls `getProperties()` on the JUnit test instance before any `ApplicationContext` exists, so there is nothing to build the runtime from. That case is only solvable by bootstrapping a context ahead of JUnit (a test launcher can do that with `GraalPyContextFactory.bootstrapReusableContext`); doing it implicitly from the generated stub would switch the whole test JVM to context reuse, so it is left as is.

### Tests run

`:micronaut-context-python:checkstyleMain`, `:micronaut-context-python:test` (166 tests), `:micronaut-inject-python-test:test -Ppython-ci` (614 tests), `:test-suite-python:test -Ppython-ci` (207 tests): all passing.


context-python/src/main/java/io/micronaut/context/python/PythonApplicationRuntime.java
```diff
@@ -76,19 +89,72 @@ final class PythonApplicationRuntime {
...
     }
 
+    /**
+     * Record the application context that is starting, so the runtime can be installed on demand by
+     * creating its GraalPy context bean.
+     *
+     * @param beanContext The bean context
+     */
+    static void bootstrapFrom(BeanContext beanContext) {
+        BOOTSTRAP_CONTEXT.set(beanContext);
```
dstepanov: Nit: this record can go stale. `DefaultBeanContext.stop()` publishes `ShutdownEvent` only when `isRunning()`, so if `start()` throws (e.g. a failing startup in a test) the failed context stays referenced here until the next context replaces it, and `require()` could call `getBean` on a dead context. Also, the bootstrap context runs the configurer too (`ApplicationContextConfigurationDelegate` delegates `getContextConfigurer`); if a refresh recreates a bootstrap context, the record is replaced and `forgetBootstrap(main)` no longer clears it. Consider skipping bootstrap configurations and holding the context via a `WeakReference` (line 63).

dstepanov (approved):
LGTM; the ordering against `DefaultBeanContext` checks out. Nits: (1) `PythonContextRuntime#usePrimaryContext()` reads `current()`, which doesn't bootstrap, so a pooled entry point (`findPooledClass`/pooled calls) reached before the runtime exists goes to the primary context even when a pool gets configured — only `require()` bootstraps. (2) The static record is JVM-wide last-writer-wins: with two contexts starting in parallel an early bean of A can build B's GraalPy context (the static `CURRENT` had this already) — worth a Javadoc sentence.

context-python/src/main/java/io/micronaut/context/python/PythonApplicationRuntime.java
graemerocher: Both points taken: the record is now a `WeakReference` (a context whose `start()` threw is no longer kept alive by it), and `PythonRuntimeBootstrapConfigurer` skips the bootstrap context (`containsBean(BootstrapContextAccess)`): it cannot provide the GraalPy context bean and publishes no `ShutdownEvent`, so a refresh recreating it no longer replaces the main context's record. New spec `PythonRuntimeInitOrderSpec."test the bootstrap context of the application is never used to initialize the runtime"` starts with `bootstrapEnvironment(true)`, refreshes the environment and checks that closing the application clears the record.

graemerocher:
Addressed the review:

- `usePrimaryContext()` now goes through `require()`, so a pooled entry point reached before the runtime exists installs it on demand and routes to the pool once one is configured.
- Javadoc on `PythonApplicationRuntime` states that the bootstrap record is JVM-wide last-writer-wins, like the installed runtime.
- The record is a `WeakReference` and bootstrap contexts (`BootstrapContextAccess` present) are never recorded, so neither a failed start nor a refresh-created bootstrap context leaves a stale record. New spec with `bootstrapEnvironment(true)` + `environment.refresh()`.

`PythonRuntimeInitOrderSpec`, `PythonEventListenerSpec`, `:micronaut-context-python:test` and `:micronaut-inject-python-test:test` (`-Ppython-ci`) green.

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
```diff
@@ -176,10 +176,12 @@ public static boolean isReuseContext() {
 
     /**
      * Whether generated calls must use the primary context: no pool is registered or the context is reused.
+     * The runtime is installed on demand first, so a pooled entry point reached before the runtime exists
+     * routes to the pool once the application has configured one.
      */
     private static boolean usePrimaryContext() {
-        PythonApplicationRuntime runtime = PythonApplicationRuntime.current();
-        return runtime == null || runtime.pool() == null || PythonApplicationRuntime.isReuseContext();
+        PythonApplicationRuntime runtime = PythonApplicationRuntime.require();
```
dstepanov, line 204: Question (from reading, not reproduced): `require()` throws when there is no runtime and none can be bootstrapped — including while `BOOTSTRAPPING` is set, i.e. while `main.py` is evaluated, before `install` runs. For the overloads that already receive a `Context` (`findPooledClass(ref, context)` / `findPooledScript(..., context)`, lines 310/374), "no runtime" used to mean "use the given context" and now throws `GraalPy context has not been initialized`. Only pooled stubs call those, so it's probably unreachable from `main.py`, but falling back to `current() == null` there (or catching in those two overloads) would keep the old behaviour.

dstepanov (approved):
Thanks, all three addressed (weak record, bootstrap contexts skipped, `usePrimaryContext()` bootstraps, Javadoc). One small question inline, not blocking.

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
graemerocher: Good catch. The two overloads that already receive a context (`findPooledClass(ref, context)` / `findPooledScript(..., context)`) now decide through `useGivenContext()`, which reads `current()` as before (no runtime installed, no pool, or reuse → use the given context), so a call made while the context bean is still being built keeps the old behaviour; the entry points without a context keep going through `require()`. Pushed as 6353a0698d.

graemerocher:
Follow-up for the inline question: the two context-taking pooled lookups keep using the given context when no runtime is installed yet (`useGivenContext()`), pushed as 6353a0698d. `:micronaut-context-python:test`, `PythonRuntimeInitOrderSpec` and the pooled specs are green with `-Ppython-ci`.

dstepanov (approved):
Thanks — `useGivenContext()` restores the old behaviour for the two context-taking lookups while the other entry points keep bootstrapping through `require()`. LGTM (checked by reading the change on `6353a0698d`).

dstepanov (approved):
Re-approving `703a00ad00`: a clean rebase onto the updated #13238, commits unchanged (`git range-diff` shows only the new base).

dstepanov (approved):
Re-approving `45973db18f`: clean rebase onto the updated parent, commits unchanged (`git range-diff`).

dstepanov (approved):
Re-approving `6669d3a356`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `e446729fcc`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

PR #13204, opened 2026-09-16
Python: merge source and Java package initializers, report duplicate types and skipped sources

graemerocher:
> Stacked on #13203 (part 3/32 of the Python fix stack; merge in order — GitHub retargets each PR when its base merges).

Fixes three silent or opaque failures of the Python compiler exposed by the Python documentation snippets of micronaut-views#1195, micronaut-json-schema#412, micronaut-gcp#1543 and micronaut-security#2359.

### A Python package coinciding with an imported Java package

**Bug:** Python sources in a package that is also imported as a Java package (or a parent of one), e.g. sources in `micronaut/views/model/` while `from micronaut.views.model import ModelAndView` is used, or a `.py` module placed directly inside an imported Java package path, failed with

```
Failed to write Python code to [GRAALPY-VFS/micronaut-application/src/micronaut/views/model/__init__.py]: Output stream or writer has already been opened
```

**Root cause:** `PythonAnnotationProcessor` writes two kinds of `__init__.py` into the GraalPy VFS: one per application source package (`from .module import Type` ...) and one per imported Java package (`X = java.type(...)`, generated decorator imports, sub-package imports). Both were written independently, so a package that is both was written twice, and the second `openWriter()` on the same generated file failed.

**Fix:** the application source packages are collected and merged into the Java package shims in `writeAllToVFS`, written once per package. Java type assignments and decorator imports come first, so a module that imports a Java type from its own package (`micronaut/context/helper.py` doing `from micronaut.context import ApplicationContext`) finds it on the partially initialized package during the circular import; the application module imports follow and `__all__` lists both. Parent packages of a source package stay namespace packages as before. Any remaining double write of a VFS path (an application module named like a generated annotation module, e.g. `jakarta/inject/Singleton.py`) is now reported with the module name instead of the opaque `IOException`.

### Two same-named top-level classes in one package

**Bug:** two top-level Python classes of the same name in different modules of one package (`TurboFrameBuilderTest.py` and `TurboFrameTest.py` both defining `TurboFrameController`) generated the same Java class; `PythonAstParser.parse` keyed the class definitions by qualified name and the later one silently overwrote the earlier one.

**Fix:** the parser records the defining source of every generated type and reports a duplicate with both sources:

```
Duplicate Python type [app.Service] defined in [.../app/second.py] and [.../app/first.py]: a top-level class or a module with decorated functions generates a Java class named after it, so the two definitions would overwrite each other; rename one of them or move it to another package
```

A module of decorated functions (`service.py` generating `Service`) conflicts with a class `Service` of another module the same way. A module whose only module-level members are plain assignments (`NAME = "..."`) does not generate beans and keeps yielding to a class of that name, as before, so existing layouts with constants modules are unaffected.

### Files silently dropped from compilation

micronaut-views#1195 reported that `from org.slf4j import LoggerFactory` silently dropped the whole module from the compilation. That could not be reproduced: the `CustomSoyFileSetProvider` snippet with the direct import compiles and `SoyTest` passes with both the released 5.2.2 compiler and this branch (the shim `org/slf4j/__init__.py` is generated, the bean is present); the workaround `LoggerFactory = java.type("org.slf4j.LoggerFactory")` in that snippet is not needed. The regression test pins that behaviour. The two places where the compiler *could* skip a source without a diagnostic were hardened instead: a source not located in any Python source directory is now reported by the parser (`PythonAstParser.parse`) and by the VFS writer instead of being left out of the parsed environment or of the application VFS.

### Tests

`inject-python-test/src/test/groovy/io/micronaut/python/compiler/CompilerSilentFailureSpec.groovy` reproduces each case (the first three fail on `5.3.x` with the exact errors above, the duplicate-type cases pass silently there):
* a source package coinciding with an imported Java package with a sub-package (`micronaut/context/helper.py`), asserting the merged initializer order and starting the application with the bean injected with `ApplicationContext`;
* a module directly inside an imported Java package path (`jakarta/inject/registry.py`);
* an application module named like a generated annotation module;
* `from org.slf4j import LoggerFactory` compiles the module and its bean;
* duplicate classes, a decorated-function module vs. a class, and a constants module vs. a class (no conflict).

Ran `:micronaut-inject-python:test` (125 tests) and `:micronaut-inject-python-test:test -Ppython-ci` (618 tests): all pass.

### Module snippet workarounds this removes the need for

* micronaut-views#1195: `micronaut/views/model/ConfigViewModelProcessor.py` can import `ViewModelProcessor` from `micronaut.views.model` instead of `java.type(...)`; `CustomSoyFileSetProvider.py` can use `from org.slf4j import LoggerFactory`; the `TurboFrameBuilderController` rename is now enforced by a compile error rather than a silent overwrite.
* micronaut-json-schema#412: the runtime generation example can live in `io.micronaut.jsonschema.generator` next to `from micronaut.jsonschema.generator import SourceGenerator`.
* micronaut-gcp#1543: snippets in `micronaut.gcp.pubsub.bind`, `.serdes`, `.support` and `micronaut.gcp.secretmanager` can import their package's Java types.
* micronaut-security#2359: Python sources can live in `micronaut.security.oauth2.client.clientcredentials`.


inject-python/src/main/java/io/micronaut/python/processing/PythonAstParser.java
```diff
@@ -236,13 +236,22 @@ public PythonEnvironment parse(List<Source> sources, List<String> srcDirs, Visit
...
 
+        // Every top-level class and every script generates a Java class of its qualified name, so
+        // two definitions of one name in different sources would silently overwrite each other
+        Map<String, Definition> definitions = new LinkedHashMap<>();
+        String[] currentSource = new String[1];
         Value bindings = context.getBindings(PYTHON);
         bindings.putMember("callback", (Function<Object, Object>) o -> {
             if (o instanceof ClassDef classDef) {
                 String qualifiedName = resolveQualifiedName(classDef.packageName(), classDef);
+                checkUniqueDefinition(definitions, qualifiedName, new Definition(currentSource[0], true));
```
dstepanov: **Blocking regression:** every top-level `ClassDef` counts as a conflicting definition, including plain undecorated helpers. `app/a.py` and `app/b.py` each define `class Helper` plus a `@Singleton` service using it: compiles and runs on #13203 (services return "a" and "b"), fails here with `Duplicate Python type [app.Helper] defined in [.../b.py] and [.../a.py]`. Module-private helpers with common names (`Helper`, `Config`, `Handler`) are normal Python. Suggest only treating classes that produce a generated bean/introspection/reference as conflicting (warn or keep last-wins otherwise), and adding a spec for this case.

inject-python/src/main/java/io/micronaut/python/processing/PythonAstParser.java
```diff
@@ -236,13 +236,22 @@ public PythonEnvironment parse(List<Source> sources, List<String> srcDirs, Visit
...
                 classes.put(qualifiedName, classDef);
             } else if (o instanceof ScriptDef scriptDef) {
                 String qualifiedName = resolveScriptQualifiedName(scriptDef.packageName(), scriptDef);
+                // A script of module-level assignments only yields no beans, so a class of the same
+                // name in another module of the package takes precedence over it without conflict
+                boolean significant = !scriptDef.functions().isEmpty() || !scriptDef.decorators().isEmpty();
+                checkUniqueDefinition(definitions, scriptDef.qualifiedName(), new Definition(currentSource[0], significant));
```
dstepanov: Nit: `scriptDef.qualifiedName()` is `packageName + "." + name`, i.e. `.Service` in the root package, while classes are keyed with `resolveQualifiedName` (`Service`, line 246), so a class/script clash in the root package isn't detected. Use one key function for both.

inject-python/src/main/java/io/micronaut/python/processing/PythonAnnotationProcessor.java
```diff
@@ -749,6 +762,14 @@ private boolean writePythonSourceToVfs(StringBuilder filesList,
                                            String content,
                                            ClassElement originatingElement) {
         boolean unchanged = false;
+        if (!writtenVfsPaths.add(filePath)) {
+            // Surfaced here instead of as the opaque "Output stream or writer has already been opened" IOException
+            throw new ProcessingException(
+                originatingElement,
+                "Python source [" + filePath.substring(APPLICATION_SRC_PATH.length()) + "] is written twice: an application "
```
dstepanov: Nit: `substring(APPLICATION_SRC_PATH.length())` produces a wrong path for files outside `APPLICATION_SRC_PATH` (e.g. `APPLICATION_LAUNCHER_PATH`); guard with `startsWith`.

dstepanov (changes requested):
Merged initializers look correct (shims before module imports, source-package parents stay namespace packages). The duplicate-type check, though, rejects ordinary module-private helper classes — reproduced locally, details inline. Nit: the new `located` check in `PythonAnnotationProcessor` (~line 337) is effectively unreachable, since the parser's `startsWith` check throws first.

inject-python/src/main/java/io/micronaut/python/processing/PythonAstParser.java
graemerocher: Fixed. Definitions are now collected per qualified name and resolved after all sources are parsed (`resolveDefinition`): a class only counts as conflicting when it carries a Java annotation (resolvable on the classpath) or an application-defined decorator, a module when it has functions or decorators. Two plain `Helper` classes compile as before, the annotated one wins over a plain one regardless of module order (previously last-wins could drop the bean) and the package initializer only imports the winning definition, so the runtime resolves the same class (`PythonEnvironment.shadowedTypes`); shadowed definitions are logged as a note. A module and its same-named class from one source (`implementation.py` defining `Implementation`) are kept together as before. Specs: `"module-private helper classes of one name in several modules of a package compile and run"` (your `app/a.py` / `app/b.py` case, services return "a" and "b"), `"an annotated class of one name wins over a plain class of another module whatever the module order"`, `"a class annotated with an application-defined decorator conflicts with another annotated class"`.

inject-python/src/main/java/io/micronaut/python/processing/PythonAstParser.java
graemerocher: Fixed: scripts are keyed with `resolveScriptQualifiedName(...)`, the same function as the classes, so `service.py` + `class Service` in the root package is now reported (`"a class and a module of decorated functions of one name clash in the root package too"`).

inject-python/src/main/java/io/micronaut/python/processing/PythonAnnotationProcessor.java
graemerocher: Fixed: the prefix is only stripped when `filePath.startsWith(APPLICATION_SRC_PATH)`.

graemerocher:
Addressed the review:

- Duplicate-type check no longer rejects module-private helper classes: conflicts are resolved after parsing and only between definitions that generate a bean/introspection/bridged members (Java-annotated classes, modules with functions or decorators); the annotated definition wins over a plain one independent of module order and the package initializer imports only the winner so compile time and runtime agree; other duplicates keep last-wins with a note. Spec reproduces the `app/a.py` / `app/b.py` `Helper` case plus the ordering, application-decorator and root-package cases.
- Scripts and classes use the same qualified-name key (root package clash detected).
- `substring(APPLICATION_SRC_PATH.length())` guarded with `startsWith`.
- Removed the unreachable `located` check in `PythonAnnotationProcessor` (the parser rejects such sources first).

`CompilerSilentFailureSpec`, `:micronaut-inject-python:test` and `:micronaut-inject-python-test:test` (`-Ppython-ci`) green.

dstepanov (approved):
Thanks, all addressed. Verified on `b568a5447f` with `-Ppython-ci`: `CompilerSilentFailureSpec` (12) passes; in a local probe two modules with plain `class Helper` plus a `@Singleton` each compile and return "a"/"b", two `@Singleton class Service` still fail with `Duplicate Python type [app.Service]`, and an assignment-only `service.py` next to a plain `class Service` keeps the class in either module order.

dstepanov (approved):
Re-approving `3168487c9f`: a clean rebase onto the updated #13203, commits unchanged (`git range-diff` shows only the new base).

dstepanov (approved):
Re-approving `ae288b2576`: clean rebase onto the updated parent, commits unchanged (`git range-diff`).

dstepanov (approved):
Re-approving `7bfd03df26`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `96bb303afd`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `589bf20960`: rebase onto the updated parent, both commit bodies unchanged (`git range-diff` against `96bb303afd`: the two commits are identical, only the base moved).

---

PR #13206, opened 2026-09-16
Python: constructor-injected async beans and awaited Python beans in event-loop contexts

dstepanov:
Two failures of Python beans with `async def` methods that only show up when the pool gives each asyncio event loop its own GraalPy context. Found with the Pyronaut order-support-assistant example on 5.2.2. With context reuse on (the pytest harness), `asyncInstance` returns the startup instance, so neither failure appears there.

### Constructor injection
`PythonPool.getOrCreateClass` created the event-loop instance with `cls.newInstance()`, so a controller with constructor parameters failed with `TypeError: __init__() missing N required positional arguments`.

- `PythonContextRuntime.newInstance` now remembers the Java constructor arguments of instances whose class has coroutine methods. The map uses weak keys, and nothing is recorded when context reuse is on.
- `asyncInstance` replays `__init__` in the event-loop context with those arguments, resolved for that context:
  - Python beans become their event-loop instances.
  - Host beans get the async-member adapter.
  - Collections and standard types are converted as a normal constructor call converts them.
- The later member copy skips host objects the replay already passed to `__init__`.
- Without remembered arguments, behaviour is unchanged (no-arg `__init__`).

### Awaiting another Python bean
`asyncMemberValue` wrapped an injected Python bean (`ValueCoercible`) in `_MicronautAsyncMember`. Member access then reached the startup-context object, and `await` failed with `object polyglot.ForeignObject can't be used in 'await' expression`.

- An injected plain Python bean now resolves to its instance in the event-loop context: created, cached and member-copied the same way as the bean that owns the async method. In its own context it is assigned as its Python object, as constructor injection already passes it.
- Introductions and scoped proxies (detected on the Python side) keep the wrapper, so their Java-side interception still runs.
- Cycles through dependencies are guarded.
- The generator now adapts injected members for any class with an `async def`, not only classes whose async methods are bridged to Java. So a plain `@Singleton` with async methods can await its own injected clients.

### Tests
- `PythonAsyncioSpec` (real server, pool enabled, no reuse) gets 5 new cases:
  - constructor-injected async controller
  - awaiting a Python bean from it
  - a host `HttpClient` passed through `__init__`
  - an attribute-injected Python bean
  - that bean awaiting its own injected client

  All 5 fail on 5.2.x and pass with this change.
- `PythonAsyncioRuntimeTest` checks that the replayed `__init__` runs in the event-loop context with the converted arguments.
- context-python, context-python-netty, test-suite-python, inject-python and inject-python-test all pass with `-Ppython-ci` (1,181 tests).

### Reuse mode hiding this
In test-suite-python, `docs/asyncio` is the only suite with `async def`, and it enables the pool. The rest of the module runs with `micronaut.python.pool.enabled=false`. inject-python-test has no async Python beans. Pyronaut's `PytestTestEngine` turns on context reuse, so `pyronaut test` will still not catch this kind of failure.


context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
```diff
@@ -231,13 +252,150 @@ public static Value asyncInstance(Value fallback, PythonClassReference classRefe
...
+                                           Value source,
+                                           Set<Value> inProgress) {
+        if (!inProgress.add(source)) {
+            // a bean reached again through its own dependencies: its instance, unless it is still being created
+            Context context = pool.findEventLoopContext(eventLoop);
+            return context == null ? null : pool.findCachedClass(context, classReference);
+        }
+        try {
+            Object[] constructorArguments = rememberedConstructorArguments(source);
+            Value target = pool.getEventLoopClass(eventLoop, classReference, cls -> newEventLoopInstance(cls, constructorArguments));
```
graemerocher: P1: Cache event-loop instances per source bean, not just per class

getEventLoopClass ultimately caches by classReference.cacheKey(), so all plain instances of a class share one event-loop object. That conflates @Prototype beans (and multiple factory-produced instances) with different constructor arguments. Worse, an async invocation returns a coroutine before it runs; another source instance can copy its state into this shared target before the first coroutine executes. Please key the event-loop cache by the startup source identity as well as the class, or otherwise preserve bean-instance identity.

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
```diff
@@ -231,13 +252,150 @@ public static Value asyncInstance(Value fallback, PythonClassReference classRefe
...
+            copyTransferableMembers(source, target, constructorArguments);
```
graemerocher: P2: Preserve values produced by the replayed constructor

This copy runs immediately after __init__ executes in the event-loop context and copies every transferable primitive from the startup object. A constructor-derived string, number, or boolean that depends on context-local state or a context-local dependency is therefore replaced by the startup-context value. The new context-marker test misses this because the startup marker is None, which is not copied; if the startup marker is a different non-null string, it overwrites the event-loop marker. Please avoid copying members initialized by the replayed constructor (not merely raw host arguments), and add a regression with non-null differing markers.

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
dstepanov: Event-loop instances are now cached per startup instance (a weak map keyed by the startup Value, in the event-loop context's state) instead of by class cache key, so prototypes and factory-produced instances each get their own. The pool's class-instance cache is no longer used for this. Regression: two startup instances with different constructor arguments keep separate event-loop instances (`PythonAsyncioRuntimeTest`).

context-python/src/main/java/io/micronaut/context/python/PythonContextRuntime.java
dstepanov: The member names the event-loop `__init__` set are recorded when the instance is created, and later copies skip them; remembered setter-injected members are still applied. This replaces the raw host-argument skip. The test now sets a non-null startup marker ('startup') that differs from the event-loop one and checks the event-loop value survives a second `asyncInstance` call.

graemerocher (approved):

PR #13207, opened 2026-09-16
Python: bridge same-arity Java overloads, map *args to varargs, select Collection/byte[] overloads for list/bytes

graemerocher:
> Stacked on #13204 (part 4/32 of the Python fix stack; merge in order — GitHub retargets each PR when its base merges).

Fixes four ways the Python compiler (`micronaut-inject-python`) and the GraalPy runtime (`micronaut-context-python`) mishandled Java method overloads and varargs. They were found while porting module documentation snippets to Python: micronaut-projects/micronaut-views#1195 (`AuthorFetcher`), micronaut-projects/micronaut-kafka#1404 (`BookClient`, `arrays` tag), micronaut-projects/micronaut-security#2359 (`AuthenticationResponse.success`) and micronaut-projects/micronaut-serialization#1425 (`bytes` payloads).

## Same-arity Java overloads implemented by one Python method

A Python class implementing an interface such as `OptionFetcher<T>` (`generate(Class<T>)` / `generate(T)`) got a single stub method and javac failed with `is not abstract and does not override abstract method generate(java.lang.Integer)`.

**Root cause:** `PythonStubGenerator.addBridgeMethod` de-duplicated bridges by the *Python* method signature (`generate(Object)`), so the second interface overload, which resolves to the same Python method, was skipped.

**Fix:** de-duplicate on the Java signature the stub emits — the key is built from the emitted (resolved) parameter types, so a generic interface method (`handle(T)` implemented for `String`) and a plain interface method with the same erasure (`handle(String)`) are bridged once; the Python signature is recorded separately (with a `python:` prefix) so the later declared-method pass still does not bridge a method twice. The stub now declares both `generate(Class<Integer>)` and `generate(Integer)`, each dispatching to the Python `generate`.

## Python `*args` as Java varargs

`def send_books(self, *books: Book)` generated a method without parameters: the processor only visited `func_node.args.args`.

**Fix:**
* `micronaut_processor.py` emits an `ArgumentDef` for the vararg, marked `variadic`; `PythonParameterElement` types it as the array of the annotated element type (`Object[]` when unannotated) and `PythonMethodElement.isVarArgs()` reports it.
* The generated bridge spreads the trailing array into positional Python arguments (`PythonInvocation.withVarargs`), also when the Java interface method is the bridged signature (e.g. an unannotated `*messages` implementing `send(String, String...)`).
* Python callers of an introduced/advised method (`client.send_books(book1, book2)`) are handled by `PythonProxyCreator`: the method selector accepts a trailing-array method for the arity and the trailing values are packed into the array (a single list/array is still passed through unchanged).

Sourcegen has no varargs flag, so the stub declares `Book[] books` rather than `Book... books`; that is a valid override of a Java varargs method, and Python-side calls are dispatched by the runtime proxy rather than by host interop.

## Python `list` selected the `Map` overload

`AuthenticationResponse.success("user", ["ROLE"])` picked `success(String, Map)`.

**Root cause:** Truffle host interop accepts any guest value with members as a `Map` at "loose" priority, but only knows `List` for array-like values; a `Collection` parameter is only satisfiable by an interface proxy at a lower priority, so `Map` won.

**Fix:** `GraalPyHostAccessFactory` registers target type mappings (HIGH precedence) for `List`, `Collection` and `Iterable` accepting array-like Python values (and, for `Collection`/`Iterable`, the finite built-in containers `set`, `frozenset` and the dictionary views, copied into a list; generators and other lazy iterables keep the default lazy host view of an `Iterable` parameter). A list now selects the collection overload, and `describe(List)` vs `describe(Map)` is no longer ambiguous for a list. The mappings ignore host objects and bytes-like buffers, so a Java `byte[]` (or Java `List`) keeps its default host conversion and `Files.write(Path, byte[])` stays unambiguous against `Files.write(Path, Iterable)` (this was the cause of the `UploadControllerSpec` failures in the first Python CI run).

## Python `bytes` did not select `byte[]`

`read(b"...")` with `read(byte[])`, `read(String)`, `read(InputStream)`, `read(Object)` overloads picked `read(Object)` (a `byte[]` target only matches at "coerce" priority, below `Object`).

**Fix:** a `byte[]` target type mapping (HIGH precedence) for values with buffer elements (`bytes`, `bytearray`, `memoryview`), read with `Value.readBuffer`. `PythonCoercion.coerceToContext` also keeps primitive arrays intact instead of copying a `byte[]` into an `Object[]`, so a `byte[]` received from Java can be passed on to Java `byte[]` parameters.

## Tests

`MethodOverloadSpec` in `inject-python-test` reproduces the four cases (fixtures in `...processing/test/overloads`): same-arity overloads with a resolved type argument, `*args` on an executable method and on a Java varargs interface method (typed and unannotated), `*args` on an introduced method called from Python positionally / with a list / with no arguments, list/tuple/dict vs `Collection`/`Map`/`List` overloads, and `bytes`/`bytearray`/`str` vs `byte[]`/`String`/`Object` overloads.

Documentation: new "Java Overloads and Variadic Parameters" page in the Python guide.

## Verification

Run locally on the rebased branch (`-Ppython-ci`, `--max-workers=1`), all green:

* `:micronaut-inject-python:checkstyleMain`, `:micronaut-context-python:checkstyleMain`
* `:micronaut-inject-python:test` — 125 tests, 0 failed
* `:micronaut-context-python:test` — 169 tests, 0 failed, 2 skipped
* `:micronaut-inject-python-test:test` — 629 tests, 0 failed (includes the new `MethodOverloadSpec`, 6 tests)
* `:test-suite-python:test` — 205 tests, 0 failed (includes `micronaut.docs.server.upload.UploadControllerSpec`)

## Module snippets that can be re-enabled

* micronaut-views#1195: `io.micronaut.views.docs.fieldset.AuthorFetcher`, `BookAuthorSave`, `AuthorFetcherTest` (the `OptionFetcher` port).
* micronaut-kafka#1404: `io.micronaut.kafka.docs.consumer.batch.BookClient` `arrays` tag (`def send_books(self, *books: Book)`).
* micronaut-security#2359: the `java.util.List.of(...)` workaround for `AuthenticationResponse.success(String, Collection)` is no longer needed.
* micronaut-serialization#1425: Python `bytes` can be passed to `readValue(byte[], ...)` directly.

Note: the bean definition's executable-method metadata for a Python method that implements a resolved generic interface method (`generate(T)` with `T = Integer`) is unchanged by this PR; it follows the inherited-signature handling in `AbstractPythonClassElement` addressed separately.


inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
```diff
@@ -2696,13 +2707,21 @@ private void addBridgeMethod(BridgeMethodSpec spec, ObjectDefBuilder<?> builder,
...
-        String key = bridgeMethodKey(methodElement);
-        // Check if method name has already been added to avoid duplicates
-        if (addedMethodNames.contains(key)) {
+        // Duplicates are detected on the Java signature the stub emits, not on the Python method:
+        // a Java interface may declare same-arity overloads (generate(Class<T>) and generate(T))
+        // that Python, which has no overloading, implements with a single method. Each overload
+        // needs its own bridge or javac rejects the stub as not implementing the interface. The
+        // Python method key is recorded too so the later declared-method pass does not bridge a
+        // method again under its Python signature.
+        String key = bridgeMethodKey(signatureMethod);
```
dstepanov: **Blocking regression:** for interface bridges `signatureMethod` is the raw declaring method, so `handle(T)` keys as `handle(java.lang.Object;)` while the stub emits the resolved `handle(String)`. A second interface declaring `handle(String)` gets a different key and is bridged again:

```java
interface GenericHandler<T> { String handle(T v); }
interface StringHandler { String handle(String v); }
```
```python
class BothHandler(GenericHandler[str], StringHandler):
    def handle(self, value: str) -> str: ...
```

→ `Pyronaut processing failed: method handle(java.lang.String) is already defined in class python.BothHandler`. On #13204 this compiles (the Python-method key caught the duplicate). Suggest keying on the emitted signature (parameter types after `bridgeSignatureType`) and adding this case to `MethodOverloadSpec`.

context-python/src/main/java/io/micronaut/context/python/GraalPyHostAccessFactory.java
```diff
@@ -86,9 +86,86 @@ HostAccess hostAccess(Collection<TargetTypeMapping<?>> mappings) {
...
+        return value != null && !value.isNull() && !value.isHostObject() && !value.hasBufferElements() && value.hasArrayElements();
+    }
+
+    private static boolean isSequenceOrIterable(@Nullable Value value) {
+        if (value == null || value.isNull() || value.isHostObject() || value.hasBufferElements()) {
+            return false;
+        }
+        // sets, generators and other iterables, but not dictionaries (their keys are iterable) or strings
+        return value.hasArrayElements()
+            || value.hasIterator() && !value.hasHashEntries() && !value.isString();
```
dstepanov: Nit: this makes any non-host Python iterable eligible for `Collection`/`Iterable` parameters, and it's copied into an `ArrayList` at call time — a lazy/infinite iterable (e.g. `Flux.fromIterable(itertools.count())`) is consumed eagerly or hangs, and Java-side mutations of a Python `set` passed as a `Collection` no longer reach Python. Worth documenting, or restricting the `Iterable` mapping to finite built-in containers (set, frozenset, dict views).

dstepanov (changes requested):
The new bridge de-duplication regresses a class implementing a generic interface and a plain interface that resolve to the same method (compiles on #13204, fails here — reproduced locally with `-Ppython-ci`). Details inline.

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
graemerocher: Fixed in eefa1f0: the key is now built from the emitted parameter types (`bridgeParameters` computes the `ParameterDef`s first, `bridgeMethodKey(name, parameters, methodTypeVariables)` erases them, method type variables to their first bound), so `handle(T)` resolved for `String` and `StringHandler.handle(String)` share the key and are bridged once. Your case is in `MethodOverloadSpec` ("a generic and a plain interface resolving to the same Java method are bridged once", `GenericHandler`/`StringHandler` fixtures); it failed with "method handle(java.lang.String) is already defined" before the change.

context-python/src/main/java/io/micronaut/context/python/GraalPyHostAccessFactory.java
graemerocher: Restricted: `Collection`/`Iterable` now match sequences and the finite built-in containers only (`set`, `frozenset`, `dict_keys`/`dict_values`/`dict_items`, via `isSequenceOrContainer`); generators and other lazy iterables are no longer matched, so an `Iterable` parameter keeps the default lazy host view for them (`ResponseFactory.first(itertools.count(7))` returns `first:7` in `MethodOverloadSpec`). The copy semantics for sets/dict views (Java-side mutations do not reach Python) are documented in the factory javadoc and in `overloads.adoc`.

graemerocher:
Addressed the review in eefa1f0:

* Bridge de-duplication keys on the emitted (resolved) Java signature instead of the raw declaring method, so a class implementing a generic and a plain interface that resolve to the same method (`GenericHandler[str]` + `StringHandler`) is bridged once. Regression test in `MethodOverloadSpec` (failed before the fix with "method handle(java.lang.String) is already defined").
* `Collection`/`Iterable` mappings accept sequences and finite built-in containers (`set`, `frozenset`, dict views) only; generators and other lazy iterables keep the default lazy host view of an `Iterable` parameter (an infinite iterator no longer hangs). Copy semantics documented in the javadoc and `overloads.adoc`; tests added.

`:micronaut-inject-python:test` (125) and `:micronaut-inject-python-test:test -Ppython-ci` (653) pass.

dstepanov (approved):
Thanks — both addressed. Verified locally on `eefa1f0588` with `-Ppython-ci`: `MethodOverloadSpec` passes (7 tests, including the `GenericHandler`/`StringHandler` case and `ResponseFactory.first(itertools.count(7))`). Keying on the emitted signature and restricting the copy to sequences and finite built-in containers look right.

dstepanov (approved):
Re-approving `6ea7ac8dc1`: a clean rebase onto the updated parent, commits unchanged (`git range-diff` shows only the new base).

dstepanov (approved):
Re-approving `d99a9651cc`: clean rebase onto the updated parent, commits unchanged (`git range-diff`).

dstepanov (approved):
Re-approving `02a3018ee0`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `5536ab881a`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `f052e9faed`: rebase onto the updated parent, all four commit bodies unchanged (`git range-diff` against `5536ab881a`).

---

PR #13208, opened 2026-09-16
Python-native streaming: async for over publishers, async generators as publishers

dstepanov:
Lets Python consume a Java `Publisher` one item at a time with `async for`, and lets an async generator supply a Java `Publisher` or an HTTP streaming response, without writing Reactor pipelines. Demand, cancellation, ordering, errors and interpreter ownership are preserved across the boundary.

### Runtime

- `as_async_iterable(publisher)` and `as_publisher(factory)` are exported by the `micronaut_asyncio` module. The Java half, `PythonAsyncioStreams`, owns the Reactive Streams protocol: signals arriving on arbitrary threads are queued onto the Micronaut event loop inside an execution frame before any Python runs; without a Micronaut loop (a plain asyncio loop in a test) the Python side hops with `call_soon_threadsafe`.
- Java → Python: lazy single subscription, one `request(1)` per pending `__anext__`, at most one buffered item, overlapping `__anext__` refused, a publisher emitting more than requested is cancelled and the iteration fails. Leaving the `async with` block, `aclose()`, task cancellation and request cancellation cancel upstream exactly once. Java failures surface as `MicronautJavaException` with the cause in `java_exception`.
- Python → Java: cold publisher, a fresh iterator per subscription, advancement only with positive demand, saturating demand arithmetic, Reactive Streams 3.9 handling, one terminal signal. Cancellation cancels the pending advance and then closes the iterator on its loop, so `finally` blocks run on client disconnect and shutdown. A yielded `None` fails the stream. An open stream counts as an active execution, so graceful shutdown waits for its cleanup; a closing context terminates the stream instead of leaking it.
- Publishers returned by injected members are now awaited lazily (`MicronautPublisherAwaitable`): `await` keeps the one-item-then-cancel semantics, `as_async_iterable` unwraps the publisher. Nothing subscribes until one of the two happens.
- Both adapters require the asyncio bridge and a Micronaut loop (or a running Python loop) and fail with a clear message otherwise, rather than driving a generator synchronously.

### Compiler

An `async def` containing a `yield` is bridged as `Publisher<T>`, with `T` from `AsyncIterator[T]`, `AsyncIterable[T]`, `AsyncGenerator[T, S]` or `Publisher[T]` (unannotated: `Object`). Elements go through the same conversion as a returned `Publisher`, so yielded dataclasses serialize as their generated wrappers. Classless routes use new `invokePooledPublisher` / `invokePooledScriptPublisher` entry points. A plain `def` with `yield` is unchanged.

### Docs

New guide section "Python-Native Streaming" under Async/Await, with snippets from the new test-suite-python fixtures.

### Tests

- `PythonAsyncioStreamsTest` (29): ordering over 1,000 items, exact demand, zero demand, break, task and stage cancellation, final item plus completion, error plus cancellation, overflow, `request(0)`, saturation, separate generator instances, loop-thread execution, round trip between the directions.
- `PythonStreamingSpec` (8): JSON and SSE streaming from explicit adapters and async generator routes, first item observed before the last is generated, disconnect closing the generator, a 20-cycle connect/disconnect churn, an async controller consuming a streaming client.
- `PythonAsyncBridgeSpec` covers the generated bridge shapes.

Full runs of context-python, context-python-netty, inject-python, inject-python-test and test-suite-python are green (1,217 tests, with `-Ppython-ci`).

### Notes

- The Micronaut HTTP client drains a cancelled streaming body to reuse the connection instead of disconnecting, so server-side cancellation is tested with a raw socket (`StreamProbe.readAndDisconnect`). That client behaviour is unchanged here.
- The Reactive Streams TCK is TestNG-based and not on the build; the protocol tests above stand in for it.
- The `InnerTypeLast` checkstyle error in `PyronautJavaCompiler` predates this change.


inject-python/src/main/java/io/micronaut/python/processing/util/PythonTypeResolver.java
```diff
@@ -284,6 +294,11 @@ private static ClassElement resolvePythonTypeToJava(String typeAnnotation, Pytho
                 visitorContext.getClassElement(Set.class).orElse(ClassElement.of(Set.class));
             case "typing.Optional" ->
                 visitorContext.getClassElement(Optional.class).orElse(ClassElement.of(Optional.class));
+            case "AsyncIterator", "typing.AsyncIterator", "collections.abc.AsyncIterator",
```
graemerocher: This resolver handles every annotation, so globally mapping AsyncIterator/AsyncIterable/AsyncGenerator to Publisher also changes parameters, properties, synchronous returns, and ordinary coroutines that return an async iterator. For example, `async def source() -> AsyncIterator[str]: return other_source()` now advertises `CompletionStage<Publisher<String>>`, but the non-generator bridge only calls `toCompletionStage`, leaving the raw guest iterator unconverted. Please restrict this mapping to the detected async-generator return path, or explicitly convert non-generator results and handle the other annotation positions.

dstepanov: Moved the mapping out of the resolver into the async-generator return path (`PythonMethodElement.bridgeReturnTypeRef`): only an `async def` with a `yield` sees `AsyncIterator[T]`/`AsyncIterable[T]`/`AsyncGenerator[T, S]` as `Publisher[T]`; parameters, properties and a coroutine returning an iterator resolve as before (the spec now covers `async def numbers() -> AsyncIterator[str]: return other_source()` bridging as `CompletionStage<Object>`). Commit d48d439c62.

graemerocher:
@dstepanov can you investigate why the build hung? Maybe something to do with this PR?

dstepanov:
Looked into it. Both Python CI runs of this branch hung inside `:micronaut-context-python:test`: every other test task finished, but that worker stopped reporting right after `PythonContextRuntimeStateTest`, with `PythonContextRuntimeTest` (the next class in that JVM) never starting or finishing, until the job limit cancelled the run six hours later. The new `PythonAsyncioStreamsTest` runs a few classes earlier in the same JVM, so it is the prime suspect, but the whole module passes locally in the same class order, and the shutdown/lifecycle classes pass after it too.

Pushed 6c30f28980 so the next run explains itself instead of hanging: a JUnit extension dumps every thread stack once a single test has run for four minutes, the test task times out after thirty, and the streams test now stops its event-loop thread (and waits for it) before closing its GraalPy context, in case a queued callback was entering a context that was being closed under it. If the dump shows the stuck thread I will fix the cause on this PR.

context-python/src/main/resources/META-INF/GRAALPY-VFS/micronaut-application/src/micronaut_asyncio.py
```diff
@@ -1238,3 +1238,390 @@ def _micronaut_complete_future(future, value, throwable):
...
+        _dispatch_on(self._loop, self._micronaut_loop, self._deliver_terminal, _COMPLETED)
+
+    def _deliver(self, item):
+        if self._closed or self._terminal is not None:
+            return
+        waiter = self._waiter
+        if waiter is not None and not waiter.done():
+            waiter.set_result(item)
+            return
+        if self._buffered is _COMPLETED:
```
graemerocher: [P2] Reject the first item beyond outstanding demand

This buffer makes the adapter silently accept one `onNext` beyond the single outstanding `request(1)`. A publisher that emits exactly `one`, `two` from its first `request(1)` is not cancelled: the loop returns both values, and no overflow error is raised. The existing test only detects the third item because the second occupies this slot. That violates Reactive Streams rule 1.1 and the PR’s exact-demand guarantee. I reproduced it with a two-item publisher; the result was `[one, two, accepted-overflow]`. Please track outstanding demand on the Java subscriber and fail/cancel on the first signal for which no demand exists, rather than treating it as a valid prefetched item.

context-python/src/main/java/io/micronaut/context/python/PythonAsyncioStreams.java
```diff
@@ -0,0 +1,483 @@
...
+        }
+
+        private void fail(Throwable throwable) {
+            boolean emit = terminated.compareAndSet(false, true) && !cancelled.get();
+            if (emit) {
+                subscriber.onError(throwable);
+            } else {
+                LOG.debug("Dropping a Python stream failure after a terminal signal or cancellation", throwable);
+            }
+            if (driver == null) {
```
graemerocher: [P1] Release an already-started driver when dispatch is refused

When `closeWhenIdle` marks the context closing, the next positive `request` is refused by `dispatch` and calls `fail`. If the driver has already been created, this branch never releases the persistent execution entered in `start()`, so the close listener waits forever. I reproduced this with a generator paused after its first item: after marking its context closing, another request delivered `onError`, but `activeExecutions(context)` stayed at 1 and the idle callback never ran. Cancellation during closing takes the refusal fallback straight to `release()` and skips the generator’s `aclose`, so it also loses `finally` cleanup. The closing/refusal path needs to cancel/close the live driver and release its execution exactly once, rather than only releasing when `driver == null`.

graemerocher (commented):
Found two actionable streaming correctness issues and left both inline:

1. A live Python generator whose context starts closing can retain its active-execution lease after a refused dispatch, leaving `closeWhenIdle` blocked indefinitely; cancellation on the same path also skips async-generator cleanup.
2. Java-to-Python adaptation accepts and delivers the first `onNext` beyond outstanding demand, so a publisher emitting two items for one `request(1)` is neither cancelled nor failed.

Validation on head `7ac76262dddcd0b70e003d63a7cf8efe049a93af`:
- Added a temporary closing-context regression: `onError` arrived, but the idle callback timed out and the context still had one active execution.
- Added a temporary two-items-for-one-request regression: it returned `[one, two, accepted-overflow]` instead of cancelling on the second signal.
- `PythonAsyncBridgeSpec`: all 12 tests passed.
- Current GitHub Python, Java, GraalVM, vulnerability, and Sonar checks are green.

context-python/src/main/resources/META-INF/GRAALPY-VFS/micronaut-application/src/micronaut_asyncio.py
dstepanov: Confirmed and fixed in 04d1cb0b0f. The Java subscriber now tracks outstanding demand and refuses an item nobody requested, cancelling the publisher and failing the iteration before the item reaches Python; the Python buffer slot is gone, so an item with no pending `__anext__` is treated as the violation it is rather than held. Two regressions cover the boundary: two items for one `request(1)` yields `[one, <without demand>]` with the upstream cancelled, and a third item changes nothing. Both fail with the gate reverted.

context-python/src/main/java/io/micronaut/context/python/PythonAsyncioStreams.java
dstepanov: Confirmed and fixed in 04d1cb0b0f. Releasing only when `driver == null` was the wrong discriminator: the lease is now released by whichever path knows no further Python of the stream can run, so a refused dispatch of a request or a cancellation releases it instead of waiting for a `released()` that can never come. The driver still owns the release on the normal path. A cancellation while the context is closing genuinely cannot run `aclose` (the registry refuses the frame precisely because the context is going away), so the close of the context disposes the generator and that is now stated at the call site. Two regressions cover it: a generator paused after its first item, with its context marked closing, now fails and lets the close-when-idle gate run, for a later request and for a cancellation. The first fails with the fix reverted.

dstepanov:
Thanks, both were real. Fixed in 04d1cb0b0f, replies on each thread.

**1. Lease retained after a refused dispatch.** Releasing the execution only when `driver == null` was the wrong discriminator: once the context is closing, the dispatch of a later request or cancellation is refused, so the driver never reports `released()` and the gate waits forever. The lease is now released by whichever path knows that no further Python of the stream can run; the driver still owns the release on the normal path. Your second observation stands as a limitation rather than a fix: a cancellation during closing cannot run the generator's `aclose`, because the registry refuses the execution frame precisely because the context is going away, so the close of the context disposes the generator. That is now stated at the call site instead of being implied.

**2. First item beyond outstanding demand.** The Java subscriber now tracks outstanding demand and refuses an item nobody requested, cancelling the publisher and failing the iteration before the item reaches Python. The Python buffer slot is gone, so an item arriving with no pending `__anext__` is treated as the protocol violation it is. Your reproduction now returns `[one, <without demand>]` with the upstream cancelled.

**Verification.** I reverted each fix in turn to confirm the new tests catch the reported behaviour:

- lease fix reverted: the closing-context test fails, the close-when-idle gate never runs;
- demand gate reverted: both overflow tests fail (the Python side still catches it as a backstop, which is why the assertions pin the Java-side refusal).

Four regressions added: two items for one `request(1)`, a third item beyond that, and a generator paused after its first item whose context is marked closing, once for a later request and once for a cancellation.

Full runs with `-Ppython-ci` are green: context-python 230, context-python-netty 56, inject-python 127, inject-python-test 911, test-suite-python 231, checkstyle clean.

One thing to flag: I changed an existing assertion. The old overflow test asserted that the second item is drained from the buffer, which encoded exactly the behaviour you asked me to remove, so keeping it would have contradicted the fix.

graemerocher (approved):

PR #13209, opened 2026-09-16
Python: bridge overrides of inherited Java methods with the inherited signature

graemerocher:
> Stacked on #13207 (part 5/32 of the Python fix stack; merge in order — GitHub retargets each PR when its base merges).

A Python method that overrides a method of an implemented Java interface was mishandled in two ways. Both were exposed by the Python doc examples of the modules: Micronaut Data (micronaut-projects/micronaut-data#4049: `UserRepositorySpec`, `SaleRepository.findById` renamed `getById`), Micronaut Kafka (micronaut-projects/micronaut-kafka#1404: `RuntimeBootstrapServers`, `ConsumerSeekAware.onPartitionsRevoked`), Micronaut Views (micronaut-projects/micronaut-views#1195: `RewriteMapProvider.cssRenamingMap()`) and Micronaut JMS (micronaut-projects/micronaut-jms#786: `getOrder()`).

### Overriding an inherited abstract method with the same signature

```python
@JdbcRepository(dialect=Dialect.H2)
class UserRepository(CrudRepository[User, int]):

    @Query("UPDATE users SET userEnabled = false WHERE id = :id")
    def deleteById(self, id: int | None) -> None: ...
```

The Python `deleteById` was dropped from the generated bean definition, so Micronaut Data kept the inherited `deleteById(Integer)` with its default interceptor and the custom query never ran.

Root cause: `AbstractPythonClassElement.appendJavaInterfaceMethods` replaced the Python parameters with the inherited ones through an anonymous `PythonMethodElement` subclass. `withAnnotationMetadata` (called for every method by the bean element creators) copied that element through the constructor and lost the substituted parameters again, so `ProxyingBeanDefinitionWriter.findOverriddenBy` saw a `deleteById(Object)` method "overridden" by `deleteById(Integer)` and skipped it. With an `int` hint (instead of `int | None`) the stub generator additionally emitted a `deleteById(int)` overload next to the inherited `deleteById(Integer)` bridge, because `hasCompatibleBridgeParameter` compared `int` with `Integer` by name; a Java caller passing an `Integer` reached Python only through the inherited bridge.

### Overriding a default interface method

```python
@ContextConfigurer
class RuntimeBootstrapServers(ApplicationContextConfigurer):
    def configure(self, builder: ApplicationContextBuilder) -> None: ...
```

The Python body never ran unless the method was annotated `@Executable`: `PythonStubGenerator.addInterfaceAndHostBridges` skipped every `default` method of the implemented interfaces, so the Java default implementation was what Java callers reached.

### Fix

* `AbstractPythonClassElement`: a Python override of an interface method (abstract or default) now adopts the signature of the method it overrides - the resolved Java parameter types with the Python parameter names and annotations (a `@Query` refers to them by name), and the Java return type where its erasure differs from the Python one. Python methods are matched to inherited ones by boxed type (`int` ~ `Integer`) since Python has no overloading.
* `PythonMethodElement`: the adopted signature lives in fields that survive the copies made by `withAnnotationMetadata`/`withNewOwningType` (`withInheritedSignature`), instead of an anonymous subclass. `withParameters` uses the same mechanism.
* `PythonStubGenerator`: bridges a default interface method when the Python class declares an override — a method with matching type hints, or (after review) any method with the same name and arity, so an unannotated `def greet(self, who)` overrides `greet(String)` too (a default method with no override keeps its Java implementation); treats a primitive hint as compatible with the boxed Java parameter (same array dimensions) so a single bridge with the Java signature is emitted; converts the Python result to the Java return type when the hint names a different primitive (`-> int` for `long count()`), which previously failed with a `ClassCastException` at run time.
* `PythonJavaTypes`: boxing helpers shared by the two.

### Reproducer

`InheritedMethodOverrideSpec` in `inject-python-test`:

* a Python class implementing a Java interface with default methods (`greet`, `onEvent`, `Ordered.getOrder`) without `@Executable`; a Java caller through the interface reaches the Python bodies, also when the override has no type hints (`def greet(self, who)`, `def onEvent(self, *events)`), and a class that does not override them keeps the defaults (no bridge emitted);
* a Python `@Singleton` implementing `BoxedIdRepository<E, ID>` with `findById(self, id: int)` / `int | None`: a single `findById(Integer)` bridge, one executable method with an `Integer` argument that keeps the Python `@Executable`, and a Java caller passing an `Integer` reaches Python;
* the same for an `@Introduction` bean, and for a Micronaut Data `@JdbcRepository(CrudRepository[Product, int])` overriding `deleteById` with `@Query` and `findById` with `@Join`: the bean definition keeps the Python metadata and the inherited `Integer` signature.

### Module tests that can be re-enabled

* micronaut-data#4049 `jdbc-example-python`: `UserRepositorySpec` (`@Disabled`), and `SaleRepository.getById` can be renamed back to `findById`.
* micronaut-kafka#1404: `ConsumerSeekAware.onPartitionsRevoked` override in `seek.aware.ProductListener` is now invoked; `RuntimeBootstrapServers.configure` is now bridged (the configurer still runs before the GraalPy runtime exists, which is a separate issue).
* micronaut-views#1195: `@Executable` on `RewriteMapProvider.cssRenamingMap()` is no longer needed.
* micronaut-jms#786: `@Executable` on `AccumulatingSuccessHandler.getOrder()` is no longer needed.

### Tests run

`:micronaut-inject-python:test` (125 tests) and `:micronaut-inject-python-test:test -Ppython-ci` (618 tests) pass; the reproducer spec re-run after rebasing onto `5.2.x`. Review follow-up: `PythonJavaTypesTest` covers `isSameOrBoxedType` with array dimensions.


inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
```diff
@@ -439,16 +440,23 @@ private boolean addInterfaceAndHostBridges(AbstractPythonClassElement classEleme
...
             for (int i = 0; i < methods.size(); i++) {
                 MethodElement method = methods.get(i);
-                if (methodSet.contains(method) || method.isDefault()) {
+                if (methodSet.contains(method)) {
                     continue;
                 }
                 MethodElement resolvedMethod = resolvedInterfaceMethod(method, resolvedMethods, i);
                 MethodElement interfaceMethod = withOwningInterface(resolvedMethod, anInterface);
                 MethodElement bridgeMethod = resolveDeclaredBridgeMethod(element, interfaceMethod);
+                if (method.isDefault() && bridgeMethod == interfaceMethod) {
```
dstepanov: Should-fix: a default method is only bridged when the override's type hints match the Java parameters (`hasCompatibleBridgeParameter`), so an unannotated override is silently skipped:

```java
interface Greeter { String name(); default String greet(String who) { return "default:" + who; } }
```
```python
class LooseGreeter(Greeter):
    def greet(self, who): return "python:" + who
```

A Java call to `greet("x")` returns `default:x` (reproduced locally). Abstract methods don't have this problem since they fall back to the interface bridge. Suggest also bridging defaults when the class declares a Python method with the same name and arity — or document that hints are required — plus a test.

inject-python/src/main/java/io/micronaut/python/processing/util/PythonJavaTypes.java
```diff
@@ -67,4 +72,47 @@ public static boolean isThrowable(@Nullable ClassElement classElement) {
...
+    /**
+     * Whether two types are the same type once primitives are boxed. A Python {@code int} hint resolves to
+     * the primitive {@code int} while a Java {@code ID} argument resolves to the boxed {@code Integer}; Python
+     * has no overloading, so a method declared with the one overrides a method declared with the other.
+     *
+     * @param left  A type
+     * @param right Another type
+     * @return {@code true} if the types are the same, or a primitive and its wrapper
+     */
+    public static boolean isSameOrBoxedType(ClassElement left, ClassElement right) {
```
dstepanov, line 85: Nit: comparing `getName()` ignores array dimensions, so an `int` hint also matches `Integer[]` in `hasSameRawParameterTypes`/`hasCompatibleBridgeParameter`. Return false when `getArrayDimensions()` differ.

dstepanov (approved):
LGTM with two comments inline; the first one means the description ("the Python body now runs without `@Executable`") only holds for annotated overrides.

inject-python/src/main/java/io/micronaut/python/processing/PythonStubGenerator.java
graemerocher: Fixed in 678128e: a default method is now also bridged when the class declares a Python method with the same name and arity (or a `*args` method that accepts it), whatever its hints (`declaresOverride`); the interface signature is bridged and the Python method is looked up by name at run time, as for abstract methods. Your `LooseGreeter` case (`def greet(self, who)` returning `python:x`) plus an unannotated `def onEvent(self, *events)` are in `InheritedMethodOverrideSpec`; the description now says so.

inject-python/src/main/java/io/micronaut/python/processing/util/PythonJavaTypes.java
graemerocher: Fixed in 678128e: `isSameOrBoxedType` returns `false` when `getArrayDimensions()` differ (`int` vs `Integer[]`; `int[]` vs `Integer[]` stay distinct too, since they are distinct Java types). Covered in `PythonJavaTypesTest`.

graemerocher:
Addressed the review in 678128e:

* Default interface methods are bridged whenever the Python class declares a method with the same name and arity, not only when its type hints match the Java parameters, so an unannotated `def greet(self, who)` overrides `greet(String)` (the description overclaimed this before; fixed).
* `PythonJavaTypes.isSameOrBoxedType` returns `false` for different array dimensions.
* Tests: `InheritedMethodOverrideSpec` (unannotated and `*args` overrides of default methods), `PythonJavaTypesTest`.

dstepanov (approved):
Thanks, both addressed. Verified on `678128e1d9` with `-Ppython-ci`: `InheritedMethodOverrideSpec` (8, incl. `LooseGreeter` and the `*args` override), `PythonJavaTypesTest` (4) and inject-python checkstyle pass.

dstepanov (approved):
Re-approving `81bfbd7d8c`: a clean rebase onto the updated parent, commits unchanged (`git range-diff` shows only the new base).

dstepanov (approved):
Re-approving `2429a6363e`: clean rebase onto the updated parent, commits unchanged (`git range-diff`).

dstepanov (approved):
Re-approving `7c1eee02f6`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `7f5fe5a7b9`: rebase onto the updated parent, all commit bodies unchanged (diffed the old and new patches).

dstepanov (approved):
Re-approving `8b2d7b5e86`: the two reviewed commits are unchanged in the rebase (`git range-diff` against `7f5fe5a7b9`), plus the new Sonar commit, which I read closely because it rewrites the interface bridging loop:

- `BoxedIdRepository<E, ID>` → `<E, I>` is the S119 type-parameter naming rule, test-only.
- The `continue` guards in `addInterfaceAndHostBridges` become nesting (S135). Equivalent, with one semantic difference worth naming: `methodSet.add(method)` now also runs for a default method that keeps its Java implementation, where it was previously skipped along with the bridging. That only matters if `anInterface.getRawClassElement().getMethods()` could hold two equal `MethodElement`s at different indices (the index feeds `resolvedInterfaceMethod`), which it does not — so the generated bridges are the same. Recording every method it has seen is arguably the more honest reading of `methodSet` anyway.

Nit, and this goes for the Sonar pass across the stack rather than this PR alone: trading guard clauses for nesting to satisfy S135 costs a level of indentation in the hottest generator loops. Where a loop body ends up three levels deep, extracting the body into a private method would satisfy the rule and read better than the nesting.

---

